PR #6496, opened 2021-10-03
Updated Dockerfile

dmint789:
Fixed issue with Maria DB being an outdated version and bumped Node to version 14 LTS. Also removed empty line at the top.

### DIFF
--- a/WcaOnRails/Dockerfile
+++ b/WcaOnRails/Dockerfile
@@ -1,4 +1,3 @@
-
 FROM ruby:2.6.7
 EXPOSE 3000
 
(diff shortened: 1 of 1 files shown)


WcaOnRails/Dockerfile
```diff
@@ -7,9 +6,9 @@ WORKDIR /app
 
 RUN apt-get update && apt-get install -y curl gnupg
 
-# Install Node 10
+# Install Node 14 LTS
 # From: https://github.com/nodesource/distributions
-RUN curl -sL https://deb.nodesource.com/setup_10.x | bash
+RUN curl -fsSL https://deb.nodesource.com/setup_lts.x | bash -
```
gregorbg, line 11: For NodeJS, a lot of libraries depend on making sure that you're running a specific version of Node. As such, I'd prefer if you fixed this to be v14 instead of your current "download whatever happens to be LTS at the time of booting the Docker container" approach.

This is not a critical issue for the Ubuntu packages below (in particular, removing the mariadb-client version is fine) because there we don't depend on having a specific version be installed for our toolchain to work.

gregorbg (changes requested):
Thanks for taking care of our Docker environment :)
One minor remark about the NodeJS version, everything else looking good!

dmint789:
@gregorbg done

gregorbg (approved):
LGTM, thanks for taking care of the changes that I previously indicated :ship:

Issue #6375, opened 2021-07-19
MBLD Results with DNF/DNS look strange


Jambrose777:
**Describe the bug**
MBLD Results on competition pages have weird spacing, making it look like the DNF/DNS the time and the attempt # are seperate results / all the same result.

**To Reproduce**
1. Go to https://www.worldcubeassociation.org/competitions/BrizZonSideOpen2021/results/all?event=333mbf

**Expected behavior**
Should be able to tell which numbers belong to which attempts

**Screenshots**
![image](https://user-images.githubusercontent.com/11577241/126184015-1101ca3b-abb6-494c-a453-74481f5b1a8c.png)

**Additional context**
Also see #4426

kr-matthews:
Hey, I would like to work on this as my first issue.

gregorbg:
Closed as per #6510

PR #6510, opened 2021-10-24
Fix MBF spacing in results display.

kr-matthews:
See issue #6375.

Increase space between MBF attempt results in 'all-results' display,
to make delimitation between attempts clear.


gregorbg (approved):
LGTM!

PR #6511, opened 2021-10-30
Added papercard for merchandise

MibettyMeanie:


WcaOnRails/app/views/posts/_specialbox.html.erb
```diff
@@ -0,0 +1,28 @@
+<div class="ui card">
+  <div class="content">
+    <h3 class="center aligned header"><%= title %></h3>
```
gregorbg, line 3: Can you change it so that the `New` text does not push the actual title out of the center?
```
|_____CENTERED NEW_____|
```
should be more like
```
|_______CENTERED NEW___|
```

WcaOnRails/config/locales/en.yml
```diff
@@ -781,7 +781,7 @@ en:
       speedcubing_history: "Speedcubing History"
       tools: "Tools"
       logo: "Logo"
-      competitions: "Competitions"
+      competitions: "Competitions"
```
gregorbg, line 784: Superfluous space?

gregorbg (changes requested):
Thanks, a few minor comments before we can launch this :)

WcaOnRails/app/views/posts/_specialbox.html.erb
```diff
@@ -1,11 +1,11 @@
 <div class="ui card">
   <div class="content">
-    <h3 class="center aligned header"><%= title %></h3>
+    <h3 class="center aligned header"><%= (title+"<span class=\"blink\"><sup style=\"color: #ff0000;font-size: 10px;\">NEW</sup></span>").html_safe%></h3>
```
gregorbg, line 3: Why are you mangling the HTML tags into a Ruby string, instead of adding them in the HTML view directly?
`<h3><%= title + '<foo></foo>' ></h3>` should become `<h3><%= title ><foo></foo></h3>`

gregorbg (changes requested):

WcaOnRails/app/views/posts/_specialbox.html.erb
gregorbg: You're basically wrapping HTML tags as a Ruby string only to immediately unwrap them again

gregorbg: Also, while I'm at it: Don't do inline `style: xyz` directly in the HTML. Give a `class` like `red-footnote` to the `sup` and then in CSS do
```
.red-footnote {
   xyz
}
```

gregorbg (approved):
LGTM, thank you for taking care of the numerous changes :)

MibettyMeanie:
Thanks for being patient with all those superfluous spaces 🙂

Regards
Meenakshi

Issue #6505, opened 2021-10-16
Feature for knowing what the WCA dues are for a Competition


Jambrose777:
**Is your feature request related to a problem? Please describe.**
As a Delegate, with the WCA Dues System now based on a formula instead of a static amount, I would like to know how much the dues will be before submitting a Competition to WCAT.

**Describe the solution you'd like**
- Option 1 (probably preferable), When creating a competition after entering in the registration fee, a note appears under the input explaining what the approximated WCA Dues will be.
- Option 2, Have a page dedicated to enter a few fields and calculate the WCA Dues.

**Additional context**
- I believe the only fields needed to calculate the WCA Dues for a competition are the Country, Registration Fee, and Competitor limit. cc: @EdHollingdale if that is incorrect.
- Country Bands are stored in country_bands. Also see https://www.worldcubeassociation.org/wfc/country-bands. The values per competitor per band are hard coded in the CountryBand model.
- Countries in Band 0 have no dues regardless (so don't apply the 5% floor)
- The final Formula probably looks like this pseudo code:
`countryBand = CountryBand(Competition.Country)`
`pricePerCompetitor = countryBand === 0 ? 0 : max(countryBand.value, Competition.registration_fee.to_usd * .05)`
`estimatedWCADues = pricePerCompetitor * Competition.competitor_limit`

PR #6539, opened 2021-11-05
Add Feature for WCA Dues Estimate

notrichardpeng:
Addresses #6505
Still need to figure out a way to convert currencies to USD, and have currency formatting in the display - any clean ways to do that?


WcaOnRails/app/views/competitions/_competition_form.html.erb
```diff
@@ -205,6 +205,34 @@
...
+            }
+
+            <% iso2 = Country.where(id: @competition.countryId).map(&:iso2) %>
+            <% countryBand = CountryBand.where(iso2: iso2).map(&:number)[0] %>
+            <% pricePerCompetitor = 0 %>
+
+            <% if !countryBand.nil? && countryBand > 0 %>
+              <% pricePerCompetitor = CountryBand::BANDS[countryBand][:value] %>
+
+              //Need a way to convert whatever the currency is to USD
```
gregorbg, line 221: In line `205` above, the input field for `currency_code` is being created. You should be able to read the currency's iso_code from that field, and then use it to work with `Money::Currency` in Rails here.

But why do you want to convert into USD in the first place? For planning a competition budget, it would probably makes more sense to display the estimate in the local currency. You could optionally do something along the lines of:
`Estimated: 1234JPY (equivalent to 12.34USD as of %today)`.

WcaOnRails/app/views/competitions/_competition_form.html.erb
```diff
@@ -205,6 +205,34 @@
...
+              <% pricePerCompetitor = [pricePerCompetitor, @competition.base_entry_fee_lowest_denomination / 100 * 0.05].max %>
+            <% end %>
+
+            //Need better formatting of currency
+            <% if @competition.competitor_limit? && @competition.competitor_limit_enabled %>
+              <% estimateDues = pricePerCompetitor * @competition.competitor_limit %>
+              $('#due_estimate').html("<b>The estimated WCA dues for " + <%= @competition.competitor_limit %> + " competitors are " +  <%= estimateDues.round(2) %> + ".<b>")
```
gregorbg, line 228: Please use proper, translated strings by adding these texts to `en.yml` and then referencing `t(key.to.translated.string)` in Ruby template.

WcaOnRails/app/views/competitions/_competition_form.html.erb
```diff
@@ -205,6 +205,34 @@
...
+            <% else %>
+              $('#due_estimate').html("<b>The estimated WCA dues per competitor are " +  <%= pricePerCompetitor.round(2) %> + ".<b>")
```
gregorbg, line 230: Translated strings, see above

gregorbg (commented):
Thanks, I have added some remarks about working with money and currencies.
Our "money" library is at https://github.com/RubyMoney/money-rails which in turn is backed by https://github.com/Rubymoney/money

WcaOnRails/app/views/competitions/_competition_form.html.erb
```diff
@@ -205,6 +205,34 @@
         <%= f.input :currency_code, collection: Money::Currency.table.values, label_method: lambda { |c| "#{c[:name]} (#{c[:iso_code]})" }, value_method: lambda { |c| c[:iso_code] }, disabled: disable_money_input  %>
         <%= f.input :base_entry_fee_lowest_denomination, as: :money_amount, currency: @competition.currency_code, disabled: disable_money_input, currency_selector: "#competition_currency_code" %>
 
+        <script>
+          $('input[name="competition[base_entry_fee_lowest_denomination]"]').on('change', function() {
+            if(!$('#due_estimate').length){
+              $(this).siblings('.help-block').append('<br><p id="due_estimate"><b>XDXD</b></p>');
```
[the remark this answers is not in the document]
notrichardpeng: Oops, forgot to take out the informal placeholder text before pushing :P

WcaOnRails/app/views/competitions/_competition_form.html.erb
notrichardpeng: I think it's because the dedicated cost of each `countryBand` is in USD, the dues calculation would only be accurate if they're using the same currency. But yeah it would be convenient to convert it back as well to display in local money.

notrichardpeng:
Sorry, my local branches corrupted, didn't know so much stuff was pushed here while fixing it

PR #6540, opened 2021-11-05
Fixes #6017 - trainee delegate email report

### LINKED ISSUE #6017: Add ability for Trainee Delegates to receive reports by email
Currently Trainee Delegates are able to view Delegate reports on the website but do not receive them via email. This means that they also miss out on any discussion that takes place in the replies to reports (a key part of their training).

There should be a "I would like to receive Delegate Reports of WCA Competitions by email" tickbox on their user profile as with staff, that should be selected by default.

danieljames-dj:


WcaOnRails/app/models/user.rb
```diff
@@ -886,7 +887,7 @@ def editable_fields_of_user(user)
         email preferred_events results_notifications_enabled
       )
       fields << { user_preferred_events_attributes: [:id, :event_id, :_destroy] }
-      if user.staff?
+      if user.staff? || user.trainee_delegate?
```
gregorbg, line 890: Perhaps introduce a method like `staff_or_any_delegate?` that can be reused down the line

WcaOnRails/app/models/user.rb
```diff
@@ -931,22 +932,23 @@ def editable_fields_of_user(user)
     fields
   end
 
-  def self.clear_receive_delegate_reports_if_not_staff
-    User.where(receive_delegate_reports: true).reject(&:staff?).map { |u| u.update(receive_delegate_reports: false) }
+  def self.clear_receive_delegate_reports_if_not_eligible
+    User.where(receive_delegate_reports: true).reject(&:staff?).reject(&:trainee_delegate?).map { |u| u.update(receive_delegate_reports: false) }
```
gregorbg, line 936: re-usable method, see first comment.
`User.where(...).reject(&:staff_or_any_delegate?).map(...)`

WcaOnRails/app/models/user.rb
gregorbg: (including a comment explaining that trainee delegates are not staff, to understand why querying `staff?` is not enough)

WcaOnRails/app/views/users/edit.html.erb
```diff
@@ -133,7 +133,7 @@
           </div>
         <% end %>
 
-        <% if @user.staff? && (current_user.can_view_all_users? || current_user == @user) %>
+        <% if (@user.staff? || @user.trainee_delegate?) && (current_user.can_view_all_users? || current_user == @user) %>
```
gregorbg, line 136: re-usable method as per my first comment above.

gregorbg (changes requested):
Thanks for tackling this issue, just one remark by me on code reusability :)

gregorbg:
Kindly ping @danieljames-dj

WcaOnRails/app/models/user.rb
```diff
@@ -603,6 +604,9 @@ def senior_delegate?
     delegate_status == "senior_delegate"
   end
 
+  def staff_or_any_delegate?
+    staff? || trainee_delegate?
```
gregorbg, line 608: missing an `end` here. Maybe this is what confuses your RuboCop in the first place? (If the syntax parsing breaks here, the error might manifest several 10s or 100s lines downwards...)

gregorbg (changes requested):
Thank you for picking this up again! Unfortunately, there's a syntax error :(

WcaOnRails/app/models/user.rb
danieljames-dj: Oops yaa sorry for that... I'll fix that. Thank you.

danieljames-dj: And yes, that was the reason what confused RuboCop, even that worked well now.

PR #6552, opened 2021-11-18
Add Feature for WCA Dues Estimate

notrichardpeng:
Addresses #6505


WcaOnRails/config/initializers/ruby_money.rb
```diff
@@ -1,3 +1,7 @@
 # frozen_string_literal: true
 
 Money.locale_backend = :i18n
+
+eu_bank = EuCentralBank.new
+eu_bank.update_rates
```
gregorbg, line 6: Does putting this here mean that currency exchange rates only get updated whenever the server is getting deployed?

WcaOnRails/app/controllers/competitions_controller.rb
```diff
@@ -372,6 +372,13 @@ def time_until_competition
     }
   end
 
+  def dues_formatting
+    converted = Money.us_dollar(params[:value]).exchange_to(params[:to_currency])
+    render json: {
+      html: converted.format,
```
gregorbg, line 378: Is the result of the `.format` call actual HTML?

gregorbg: Otherwise, the name of the key is extremely misleading.

WcaOnRails/app/views/competitions/_competition_form.html.erb
```diff
@@ -205,6 +205,62 @@
...
+          $('input[name="competition[base_entry_fee_lowest_denomination]"]').on('change', function() {
+            if(!$('#dues_estimate').length){
+              $(this).siblings('.help-block').append('<br><p id="dues_estimate"><b></b></p>');
+            }
+            baseFee = this.value / 100 * 0.05; //TODO: check our currency subunit_to_unit
+            UpdateDues();
+          }).trigger('change');
+
+          $('select[name="competition[competitor_limit_enabled]"]').on('change', function() {
+            if(this.value === "false"){
```
gregorbg, line 246: Is the state of a checkbox really encoded as string in JS?!?

WcaOnRails/app/views/competitions/_competition_form.html.erb
```diff
@@ -205,6 +205,62 @@
         <%= f.input :currency_code, collection: Money::Currency.table.values, label_method: lambda { |c| "#{c[:name]} (#{c[:iso_code]})" }, value_method: lambda { |c| c[:iso_code] }, disabled: disable_money_input  %>
         <%= f.input :base_entry_fee_lowest_denomination, as: :money_amount, currency: @competition.currency_code, disabled: disable_money_input, currency_selector: "#competition_currency_code" %>
 
+        <script>
+          var bandFee = 0;
+          var baseFee = 0;
+          var compLimit = null;
+          var tempLimit = null;
```
gregorbg, line 212: What does this `tempLimit` variable do?

gregorbg (commented):
Just a few general comments, neither blocking nor approving because I am not sure if you consider this PR "done" yet. :smiley:

WcaOnRails/app/controllers/competitions_controller.rb
notrichardpeng: > 

nope, it isn't HTML :p

WcaOnRails/app/views/competitions/_competition_form.html.erb
notrichardpeng: Before disabling the competitor limit, I wanted to store it in a variable so that when it is enabled again, the dues field can immediately change back to displaying the locale with a specific competitor limit.

WcaOnRails/app/views/competitions/_competition_form.html.erb
notrichardpeng: yes... I see `=== "true"` and `=== "false"` used in other jquery files so I kept things consistent

WcaOnRails/config/initializers/ruby_money.rb
notrichardpeng: indeed. The `eu_bank` github readme also recommends updating it every time we convert, but I don't know if that would be too slow.

WcaOnRails/app/controllers/competitions_controller.rb
```diff
@@ -373,9 +373,13 @@ def time_until_competition
   end
 
   def dues_formatting
+    if !Money.default_bank.rates_updated_at || Money.default_bank.rates_updated_at < Time.now - 1.days
```
gregorbg, line 376: I like that! :smile: 
In Ruby, instead of calculating `Time.now - 1.days` you can do `1.days.ago`. But please check this, I'm writing this off the top of my head

notrichardpeng: However, locally `rates_updated_at` does not change after `update_rates` is called. And `rates_updated_at` gives year-month-date while Time.now gives year-month-date hour:minutes:seconds, can they still be compared?

I still added this change because `update_rates` only took 1-2 extra seconds to run on my machine 😬

notrichardpeng:
I think this PR is pretty much complete, except if we could address the minor issue that ajax requests get aborted when the page is just loaded up.

WcaOnRails/app/controllers/competitions_controller.rb
viroulep: Just passing by and I was curious about that PR; if updating the rates is some kind of routine jobs, I'd recommend putting this in a rake task instead of doing this in a controller method.

We do similar stuff [here](https://github.com/thewca/worldcubeassociation.org/blob/master/WcaOnRails/lib/tasks/work.rake) for instance, and it wouldn't be too hard to add an extra job.

gregorbg: Also, just a minor nit: Rails has a lot of convenience features for working with date and time. I believe you can do something like `1.days.ago` instead of `Time.now - 1.days`.

gregorbg:
Not a JQuery guru at all, but could your ajax problem be related to the `cancelPendingAjaxAndAjax` implementation?

Try looking at the source code of that method, it has an `if` check right at the top. Doing some `console.log('HEELLOOOO')`-style debugging within that if-condition should help to find out if the `.abort()` method is actually being invoked upon page reload.

WcaOnRails/app/controllers/competitions_controller.rb
notrichardpeng: Would I still need to use this though if I just put the currency update into the rake jobs

notrichardpeng:
> Not a JQuery guru at all, but could your ajax problem be related to the `cancelPendingAjaxAndAjax` implementation?
> 
> Try looking at the source code of that method, it has an `if` check right at the top. Doing some `console.log('HEELLOOOO')`-style debugging within that if-condition should help to find out if the `.abort()` method is actually being invoked upon page reload.
> 

Thanks, I'll be taking a closer look next week when I get back!

notrichardpeng:
Indeed, the ajax calls are being aborted in that `if` statement when first loading the page. The currency conversion will work upon page load if `window.wca._pendingAjaxById[id].abort();` is commented out.
 
However, I am sure the line is there for other reasons, so I am not confident that it will be the best if we remove that line just for the purpose of currency conversion upon page loading.

notrichardpeng:
Did I add the rake task correctly? @viroulep

viroulep:
I think so, you can actually test this locally by running `bin/rake jobs:work`!

WcaOnRails/app/views/competitions/_competition_form.html.erb
gregorbg: Okay. This fact seems really weird to me (not your fault at all) but I appreciate that you're trying to keep things consistent. So that's fine by me :D

WcaOnRails/config/locales/en.yml
```diff
@@ -496,7 +496,7 @@ en:
         end_date: ""
         generate_website: ""
         external_registration_page: "The page where the competitors have to register for the competition. Must be a valid http(s) url."
-        base_entry_fee_lowest_denomination: 'The base registration fee for the competition. It will be displayed on the competition information page and the registration page if used. If this is set to 0, a sentence will be displayed saying that registrations will be accepted for free.'
+        base_entry_fee_lowest_denomination: 'The base registration fee for the competition. It will be displayed on the competition information page and the registration page if used. If this is set to 0, a sentence will be displayed saying that registrations will be accepted for free. <br><br><b>DISCLAIMER: The following WCA Dues estimate is solely an approximation based on current exchange rates and is in no way legally binding for the Dues payment.</b><br>Click "Update Competition" after changing currency code to synchronised it with the estimation. If there are no available exchange rates for the selected currency, nothing will be displayed.'
```
gregorbg, line 499: Nit: It says `to synchronised it`. Should be `to synchronise it`.

WcaOnRails/app/jobs/update_currency_rates.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class UpdateCurrencyRates < SingletonApplicationJob
+  queue_as :default
+
+  def perform
+    if !Money.default_bank.rates_updated_at || Money.default_bank.rates_updated_at < 1.days.ago
```
gregorbg, line 7: Extract this `1.days.ago` into a constant along the lines of `CURRENCY_RATES_UPDATE_INTERVAL` maybe? Or is that too overkill?

gregorbg (commented):
Just a minor nitpick about the translation. Nothing seriously blocking. Plus another explanation on the jQuery will follow.

gregorbg:
> Indeed, the ajax calls are being aborted in that `if` statement when first loading the page. The currency conversion will work upon page load if `window.wca._pendingAjaxById[id].abort();` is commented out.
> 
> However, I am sure the line is there for other reasons, so I am not confident that it will be the best if we remove that line just for the purpose of currency conversion upon page loading.

Yes, I am also sure it is there for other reasons. In particular, that line clearly is what cancels any pending AJAX request, so it sits right at the "heart" of this method.

The question is not whether or not to alter that central method by removing the line, but the question is rather why there are two (virtually) simultaneous AJAX requests when the page is being loaded, so that one cancels the other.
Try loading the webpage with Dev Tools enabled and check the XHR requests on the networking tab to try and find out where those AJAX requests on page load are originating from.

WcaOnRails/app/jobs/update_currency_rates.rb
notrichardpeng: sure, that would be easier for others to edit

notrichardpeng:
Hi Gregor I just looked into the XHR request you talked about and found that three ajax requests of currency conversion were called at the same time at the start, because `.on('change')` is always triggered when the page is just loaded! Normally though only one of the three `.on('change')` is triggered, which wouldn't cause any problems (because we only input one field at a time). Now I need to find a way to make two of the three `.on('change')` not run upon start.

gregorbg (approved):
Thanks, that's very much in line with my initial suspicions about AJAX. The whole "cancel before retrying" notion upon page load makes this code order-dependent, which is not exactly nice. But it's a legacy problem of our jQuery codebase and totally not your fault!

Thanks for figuring this out and coming up with a relatively painless fix - LGTM!

PR #6557, opened 2021-11-26
Updated preferred formats to comply with the regulations

dmint789:
I have created a migration to delete old preferred formats to comply with the regulations and disallow organizers to select these. I list all of the formats I deleted from the db below.

Question: is it normal that the structure.sql file got automatically updated too? I only created a migration. There are a LOT of changes in that file for some reason.

List of deleted formats:
2x2x2 Best of 1,2,3
3x3x3 Best of 1,2,3
4x4x4 Best of 1,2,3
5x5x5 Best of 1,2,3
6x6x6 Best of 1,2
7x7x7 Best of 1,2
3x3x3 BLD Best of 1,2
4x4x4 BLD Best of 1,2
5x5x5 BLD Best of 1,2
3x3x3 OH Best of 1,2,3
clock Best of 1,2,3
megaminx Best of 1,2,3
pyraminx Best of 1,2,3
skewb Best of 1,2,3
square-1 Best of 1,2,3


WcaOnRails/db/migrate/20211126105123_update_preferred_formats_again.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+class UpdatePreferredFormatsAgain < ActiveRecord::Migration[6.0]
+  def change
+    PreferredFormat.where(event_id: "222", format_id: "1").delete_all
```
Jambrose777, line 5: More of a general question for WST: what's the difference between `.delete_all` and `.destroy_all` here?

WcaOnRails/db/migrate/20211126105123_update_preferred_formats_again.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+class UpdatePreferredFormatsAgain < ActiveRecord::Migration[6.0]
```
Jambrose777, line 3: Maybe instead of Again here, we can name it the regulations date. Possibly like: `RemoveEventFormatsFrom2020Regulations` or something (also not sure what year the regs updated this).

WcaOnRails/db/structure.sql
```diff
@@ -18,19 +18,19 @@ CREATE TABLE `Competitions` (
   `cityName` varchar(50) CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci NOT NULL DEFAULT '',
   `countryId` varchar(50) CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci NOT NULL DEFAULT '',
   `information` mediumtext CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci,
-  `year` smallint(5) unsigned NOT NULL DEFAULT '0',
```
Jambrose777, line 21: The structure.sql shouldn't be updated with this database migration and is likely pulling in local configs from your database. It'd be best to remove this file from the commit (and not commit it upon further changes when the structure is not updated).
`git checkout master -- WcaOnRails/db/structure.sql`

Jambrose777 (changes requested):

WcaOnRails/db/migrate/20211126105123_update_preferred_formats_again.rb
viroulep: There is a small paragraph about this in [`destroy_all`'s documentation](https://apidock.com/rails/ActiveRecord/Relation/destroy_all), but basically `delete_all` just fires the sql query, whereas `destroy_all` runs all the callbacks and destroys associated record if any.
In the case of `PreferredFormat` I think delete is just fine!

WcaOnRails/db/migrate/20211126105123_update_preferred_formats_again.rb
```diff
@@ -0,0 +1,47 @@
...
+  def change
```
viroulep, line 4: Does `change` actually works here?
It's meant to be a way to provided reversible actions so that we can use `db:migrate` or `db:rollback` (eg: `create_table` would automatically be transformed into a drop table when doing a rollback).
I realize previous similar migrations have this though...
One way of doing this "properly" is to provide an `up` and `down` methods which are used when migrating/rollback-ing.

viroulep:
This migration kind of relies a lot on what is already in the db; what do you think about making it so that it populates the db with exactly the valid data at the current time?

One possible way could be fixing the seeding file [here](https://github.com/thewca/worldcubeassociation.org/blob/master/WcaOnRails/db/seeds/preferred_formats.seeds.rb) with the explicit list of valid formats for each event, and then just make a migration that delete all preferred format and import back the seeds?

WcaOnRails/db/migrate/20211126105123_update_preferred_formats_again.rb
dmint789: I think it does actually work, as I've done a rollback and remigrated and it was working as expected

WcaOnRails/db/structure.sql
dmint789: I'll revert this

dmint789:
> This migration kind of relies a lot on what is already in the db; what do you think about making it so that it populates the db with exactly the valid data at the current time?
> 
> One possible way could be fixing the seeding file [here](https://github.com/thewca/worldcubeassociation.org/blob/master/WcaOnRails/db/seeds/preferred_formats.seeds.rb) with the explicit list of valid formats for each event, and then just make a migration that delete all preferred format and import back the seeds?

Thanks, that's actually a lot cleaner than what I did. I have now updated that file and reverted my previous changes. Will the website automatically reseed from this file though?

dmint789:
The new commit does the same thing I outlined in the PR and ALSO reverses the ranking of MBLD formats so that Best of 1 is the preferred format, as that is the most commonly used format

dmint789:
I don't understand why the tests failed

gregorbg:
A note regarding `structure.sql`: This file is automagically managed by Rails as part of the Migrations framework.
It was added to our Repo under MySQL 5, when a lot of defaults and column types were wildly different. Unfortunately, Rails is not capable of dynamically adjusting only the tables that were migrated within `structure.sql`. Instead, it either dumps the complete structure anew or it won't alter the file. (All-or-nothing scenario)

Nowadays when running a new migration, it re-generates the entire structure.sql based on MySQL 8 or MariaDB, depending on what you're running locally. This leads to the ludicrous amount of changes.

gregorbg:
Regarding the test failures: The first two are related.
Short version: The tests expect that Mo3 is a valid format for 3x3 and 5x5, which it isn't anymore. So you have to change the tests.
Long version: They both point to `competition_event:54` in the stack trace. That line `create`s an event in the database, and Rails alsways runs validations when entities are written to the database (that's just how Rails works). One particular validation is in `app/models/round.rb:50`, and it uses the `preferred_formats` table to verify whether a format is even allowed at all. With your changes, the Mo3 format `3` is not in that table anymore, so the validation fails.

The other failure is similar in nature, but very different in code. The test tries to click on "Mean of 3" in the dropdown although Mo3 isn't even a valid option anymore. The test has to be changed as well.

dmint789:
> Regarding the test failures: The first two are related. Short version: The tests expect that Mo3 is a valid format for 3x3 and 5x5, which it isn't anymore. So you have to change the tests. Long version: They both point to `competition_event:54` in the stack trace. That line `create`s an event in the database, and Rails alsways runs validations when entities are written to the database (that's just how Rails works). One particular validation is in `app/models/round.rb:50`, and it uses the `preferred_formats` table to verify whether a format is even allowed at all. With your changes, the Mo3 format `3` is not in that table anymore, so the validation fails.
> 
> The other failure is similar in nature, but very different in code. The test tries to click on "Mean of 3" in the dropdown although Mo3 isn't even a valid option anymore. The test has to be changed as well.

Am I supposed to fix `WcaOnRails/config/environments/test.rb`? Do I do it in this same PR?

gregorbg:
No, why would you want to fix the test environment?
There are a few specific tests that are failing, because they expect that a format is legal which actually isn't legal anymore. So you have to change the Ruby code of those very specific tests instead.

dmint789:
@gregorbg I think I finally found where the relevant bit of the code is located and possibly fixed it too

gregorbg:
Yup, that looks about right. But maybe deleting that one test altogether isn't the most elegant solution. Try and see if you can create a test scenario like "for 3x3x3, there exists no other option except Ao5 to choose from". (But if it takes too long, then just delete the existing test as you have done now and don't waste too much time on this)

dmint789:
> Yup, that looks about right. But maybe deleting that one test altogether isn't the most elegant solution. Try and see if you can create a test scenario like "for 3x3x3, there exists no other option except Ao5 to choose from". (But if it takes too long, then just delete the existing test as you have done now and don't waste too much time on this)

I have no idea how to code something like that. I have never worked with ruby tests before.

dmint789:
> > I have no idea how to code something like that. I have never worked with ruby tests before.
> 
> Which is why I suggested you try it out now to learn :) (Unless it would require a substantially unreasonable amount of time to come up with the test code)

It's just that I don't even know where to start with this issue, so it would probably take me a little while to figure it out

gregorbg:
Okay, then I leave it up to you to decide whether you want to invest the time :)
Irrespective of this decision about tests, you have to make sure there is a migration that actively loads the new seeds into the prod DB. Right now, the tests succeed because the test database is populated freshly every time tests are run. The prod DB is not.

dmint789:
> Okay, then I leave it up to you to decide whether you want to invest the time :) Irrespective of this decision about tests, you have to make sure there is a migration that actively loads the new seeds into the prod DB. Right now, the tests succeed because the test database is populated freshly every time tests are run. The prod DB is not.

I was actually wondering about how the prod db would even know to update that seeded data. I shall investigate this and make a migration.

viroulep:
@dmint789:
> Will the website automatically reseed from this file though?

Unfortunately not, seeds are only populated (and should only be populated) when creating a fresh database (ie: for tests, or for dev environment without the developer dump).
I've looked around a bit to see how people dealt with that, but [that SO post](https://stackoverflow.com/questions/7100739/loading-seed-data-for-a-rails-migration) seems to suggest to define the data in both places (migration and seeds).

Not very DRY, but you may want to create a migration that drop all preferred format and re-import all the valid preferred format for the current time.
I think you can execute raw sql in the migration, or use some Hash to feed the create method, something like that:

```ruby
PreferredFormat.delete_all
formats = [
{event_id:'333', format_id: 'a', rank: 1},
{event_id:'222', format_id: 'a', rank: 1},
{event_id:'444', format_id: 'a', rank: 1},
{event_id:'555', format_id: 'a', rank: 1},
#...
]
formats.each do |f|
  PreferredFormat.create!(f)
end
```

I definitely did not test this code, but I think you get the idea :)
It's worth noting this approach would also work in a seed file.

dmint789:
> @dmint789:
> 
> > Will the website automatically reseed from this file though?
> 
> Unfortunately not, seeds are only populated (and should only be populated) when creating a fresh database (ie: for tests, or for dev environment without the developer dump). I've looked around a bit to see how people dealt with that, but [that SO post](https://stackoverflow.com/questions/7100739/loading-seed-data-for-a-rails-migration) seems to suggest to define the data in both places (migration and seeds).
> 
> Not very DRY, but you may want to create a migration that drop all preferred format and re-import all the valid preferred format for the current time. I think you can execute raw sql in the migration, or use some Hash to feed the create method, something like that:
> 
> ```ruby
> PreferredFormat.delete_all
> formats = [
> {event_id:'333', format_id: 'a', rank: 1},
> {event_id:'222', format_id: 'a', rank: 1},
> {event_id:'444', format_id: 'a', rank: 1},
> {event_id:'555', format_id: 'a', rank: 1},
> #...
> ]
> formats.each do |f|
>   PreferredFormat.create!(f)
> end
> ```
> 
> I definitely did not test this code, but I think you get the idea :) It's worth noting this approach would also work in a seed file.

I dug around a bit and found that it's best to just update the seed file (which I have already done) and then simply run rails db:seed. Would this not be a more elegant solution? No need for a migration even

gregorbg:
The problem with your suggested solution (Philippe please shout if I'm wrong) would be that `rails db:seed` populates *all* seeds, not just the `preferred_formats.seeds.rb` that you're working on in this PR

dmint789:
It just doesn't seem very elegant to have a migration that is effectively useless if the db were to be restarted or something. Is there any issue with reseeding other seed files too? Or are you saying that that would also get rid of all data that was user generated over the years?

viroulep:
> Is there any issue with reseeding other seed files too?

Yes, we'll end up with duplicate preferred formats, teams, events, and so on.
And if we have actual "unique" constraints in the db, running db:seed will just fail :s

dmint789:
> > Is there any issue with reseeding other seed files too?
> 
> Yes, we'll end up with duplicate preferred formats, teams, events, and so on. And if we have actual "unique" constraints in the db, running db:seed will just fail :s

Okay, so do I need to do both the seed file update and the migration? If so, doesn't that mean there'd be overlap between the two?

gregorbg:
Correct, you will have to do both. And also correct that there will be overlap between the two. That's why the migration should say "drop everything from that one specific table, _then_ re-insert the same data that is also provided by the seeds".

The rationale is that the seeds should contain everything needed to spin up a new dummy DB (basically a reproducible testing environment, and also onboarding devs initially when you don't want to wait 30 minutes for the dev export straight away). Migrations on the other hand are for managing an _existing_ database that would be cluttered (database entries would be doubled when re-populating seeds that had already been populated before).

Technically speaking, adjusting the seeds isn't even necessary, because from a functional point of view the test environment really doesn't care if Mean of 2 is an allowed format or not. The migration is definitely more important. But this is such a substantial change that I'd say the particular importance of this Regulations update warrants touching the "unimportant" seed files as well, to keep things in sync :)

dmint789:
Thank you for all of the explanations! I have added a migration with up and down methods. I'm not really sure what to do for the downgrade method, so I just made it delete all of the entries, since the seed file doesn't include old preferred formats anyways. If needed, I can manually include the list of old formats in the downgrade, but that seems very unnecessary, as the only way to get that list moving forward would be to specifically migrate the db to its present state and then revert that migration, which makes no sense to me personally.

dmint789:
Could someone advise me on why this error occurred? Perhaps @viroulep would know.

viroulep:
Hum my guess is that you forgot to dump the migration number in the structure.sql: a lot of changes are not to be committed but the couple of lines at the end should!

dmint789:
I think I have finally ironed out the bugs. Please review.

dmint789:
Question: could there be anywhere on the website that would have an issue with removed events not having preferred formats?

WcaOnRails/db/migrate/20211209144129_repopulate_preferred_formats.rb
```diff
@@ -0,0 +1,40 @@
...
+      { event_id: '333mbf', format_id: '3', ranking: 3 },
+    ]
+
+    formats.each do |f|
+      PreferredFormat.create!(f)
+    end
+  end
+
+  def down
+    PreferredFormat.delete_all
```
gregorbg, line 38: This is technically not correct, although I'd argue that it would be too much of a hassle to reconstruct how to move from pre-2020 to post-2020.

Just mentioning it here, shouldn't block the merge really.

gregorbg (changes requested):
One remark about our beloved `structure.sql` and one non-critical comment.

Regarding certain parts that depend on certain `PreferredFormat`s: That's what we have staging for :smile: Please refer to the wiki to figure out how to deploy stuff to staging and try clicking through the website there, create a dummy comp, register for another dummy comp etc... to  see if the most "casual" workflows don't break

WcaOnRails/db/migrate/20211209144129_repopulate_preferred_formats.rb
dmint789: So do I just leave it as it is? As I mentioned above, I can repopulate the db with old values in the down function if needed

dmint789:
My commits were getting quite messy and out of date, so I did a rebase with master and redid the migration commit. Please review.

dmint789:
Cleaned up the commits as per @gregorbg's request. Please approve.

gregorbg (approved):
LGTM :ship:

PR #6559, opened 2021-11-28
Display Merchandise as normal card without NEW ad

gregorbg:
Retain blinking NEW feature via optional 'blinking_new' bool flag


gregorbg:
Merging this as the blinking `NEW` advert is starting to go on my nerves :sweat_smile:

PR #6564, opened 2021-11-29
Update the voters CSV file generation

emattos1723:
The software we use for elections (Heliosvoting) recently started requiring an extra field on the CSV file submitted. The first entry for each voter should now be "password" to accommodate for that.


gregorbg:
Thanks for your contribution! There is a test which expects the old three-column format at `WcaOnRails/spec/features/eligible_voters_spec.rb` which will have to be updated as well. (The test has to be changed to expect the four-column format)

Is there some documentation available for this change? Would it make sense for the WCA website to generate non-trivial passwords for people to use?

emattos1723:
Oh, now it makes sense, I was struggling to figure out what was causing the error, thanks!

I'm not aware of where we could find a documentation for that, as we found out about it when an error popped up while trying to set up the upcoming WCT Leader Election. I think I found the commit that added this change ([link](https://github.com/benadida/helios-server/commit/6e1aaafa77a516c72198a554bf9ff334c36f2d65)), but there doesn't seem to exist anything more than that.

> Would it make sense for the WCA website to generate non-trivial passwords for people to use?

What do you mean? The passwords used in elections are generated by Helios, we currently only generate the userid for each person, right?

emattos1723:
I tried updating the tests you mentioned. I will admit that the amount of repetition doesn't look pretty, but I'm not familiar with Ruby enough to know if there is a better way to write it. Let me know if I should have done it differently

gregorbg:
Ah, your PR indicates that the fourth column being added is about passwords, and "password" would just be a dummy password. But now I see that I was mistaken and looking at the Helios commit it seems that "voter_type" has to be set to "password".
You should definitely add a comment to your Ruby code explaining where that "password" literal comes from!

emattos1723:
I believe it should all be done now!  To be safe, I added comments both on the code that generates the file and on the tests, hopefully that is not too redundant. Let me know if there is anything else I have to do before this can be merged!

gregorbg:
I for one prefer having one comment too much over having no comment at all :smile: 
Thank you for your contribution and your prompt fixes/adjustments! Going live within the hour.

gregorbg (approved):

gregorbg:
Should be live now, please check by generating a fresh voters export.

emattos1723:
Both files seem to have generated correctly. Helios also didn't throw any error when I tried to use these, so everything should be working as it should. Thanks for your help!

PR #6565, opened 2021-11-30
Removed cutoff field from 3-5BLD on the Edit events page

dmint789:
Pretty simple fix

### DIFF
--- a/WcaOnRails/app/models/event.rb
+++ b/WcaOnRails/app/models/event.rb
@@ -62,6 +62,10 @@ class Event < ApplicationRecord
     !fewest_moves? && !multiple_blindfolded?
   end
 
+  def can_have_cutoff?
+    self.id != "333bf" && self.id != "444bf" && self.id != "555bf"
+  end
+
   # Events that are generally fast enough to never need to go over the default 10 minute time limit
   def fast_event?
     ['333', '222', '444', '333oh', 'clock', 'mega', 'pyram', 'skewb', 'sq1'].include?(self.id)
@@ -73,6 +77,7 @@ class Event < ApplicationRecord
       name: self.name,
       format_ids: self.formats.map(&:id),
       can_change_time_limit: self.can_change_time_limit?,
+      can_have_cutoff: self.can_have_cutoff?,
       is_timed_event: self.timed_event?,
       is_fewest_moves: self.fewest_moves?,
       is_multiple_blindfolded: self.multiple_blindfolded?,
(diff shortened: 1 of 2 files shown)


dmint789:
Wait, I think I removed the wrong thing. Do the regulations disallow cutoffs only for 3bld, 4bld and 5bld? I must have misunderstood something and removed cutoffs for the wrong events. Can someone please confirm that it's 3-5bld that need to have cutoffs removed?

gregorbg:
As per [9b4](https://www.worldcubeassociation.org/regulations/#9b4) and [9b5](https://www.worldcubeassociation.org/regulations/#9b5), MBLD and FMC *do* have cutoffs. This misunderstanding was partially caused by my confused remark on Slack, I am sorry.

Indeed only 3BLD, 4BLD and 5BLD do not have cutoffs as per [9b3](https://www.worldcubeassociation.org/regulations/#9b3). Please proceed to make the change accordingly.

gregorbg (changes requested):
In addition to the general confusion about which events have cutoffs, I have one pure software comment: You are "hijacking" another field `canChangeTimeLimit` which happens to fulfill the (now incorrect, as we discovered) purpose of detecting cutoff events.
As the selection of cutoff events will likely change in the future, we should add another `canAddCutoff` field. That field can then be backed by `canChangeTimeLimit` on the Ruby side of things, but for the JSON that the frontend operates on, changing time limits and adding cutoffs should be two completely distinct properties.

dmint789:
> In addition to the general confusion about which events have cutoffs, I have one pure software comment: You are "hijacking" another field `canChangeTimeLimit` which happens to fulfill the (now incorrect, as we discovered) purpose of detecting cutoff events. As the selection of cutoff events will likely change in the future, we should add another `canAddCutoff` field. That field can then be backed by `canChangeTimeLimit` on the Ruby side of things, but for the JSON that the frontend operates on, changing time limits and adding cutoffs should be two completely distinct properties.

I will add such a field to the backend and update the frontend

dmint789:
@gregorbg done

notrichardpeng (approved):

gregorbg (commented):
Cool, thanks for making the adjustments!
But does it actually work to provide the keys in `snake_case` on the Ruby side and parse them in `camelCase` on the JS side? That means our code has to do automated conversion between the two at some point...

Don't let this hold you back from merging, if it works that's totally fine by me!

dmint789:
> Cool, thanks for making the adjustments! But does it actually work to provide the keys in `snake_case` on the Ruby side and parse them in `camelCase` on the JS side? That means our code has to do automated conversion between the two at some point...
> 
> Don't let this hold you back from merging, if it works that's totally fine by me!

It's line 11 here: https://github.com/thewca/worldcubeassociation.org/blob/master/WcaOnRails/app/webpacker/javascript/wca/events.js.erb

PR #6567, opened 2021-12-01
Fixes #6566 - edge case of schedule in competition page

### LINKED ISSUE #6566: Edge case of schedule in competition page
In schedule of competition page, we can't add time beyond 11 PM.

danieljames-dj:


notrichardpeng (approved):

gregorbg:
Good catch, thank you for the fix!

Everything is fine in terms of the review and our general procedures. Just one "cherry on top" code review: Because one day has 24 hours, repeating `25` times seems very unintuitive at first glance. In edge cases like this, it is very helpful to include a comment in the code why `25` still makes sense here :)

PR #6585, opened 2021-12-06
Added message about the user's position in the waiting list

dmint789:
This is a small fix that just lets the user know what their position on the waiting list is to give more transparency. The issue with the current registration process is much bigger than this though, but perhaps this will improve things slightly, although the relevant text is very small and can be easily missed by most users.

One thing to change before this can be merged: display the number in an ordinal format (e.g. "7th" instead of "7")
EDIT: that's been implemented now

### DIFF
--- a/WcaOnRails/app/models/registration.rb
+++ b/WcaOnRails/app/models/registration.rb
@@ -171,6 +171,12 @@ class Registration < ApplicationRecord
     OpenStruct.new(index: index, length: pending_registrations.length)
   end
 
+  def paid_waiting_list_info
+    pending_paid_registrations = competition.registrations.pending.order(:last_payment_date)
+    index = pending_paid_registrations.index(self)
+    OpenStruct.new(index: index, length: pending_paid_registrations.length)
+  end
+
   def to_wcif(authorized: false)
     authorized_fields = {
       "guests" => guests,
--- a/WcaOnRails/app/models/registration.rb
+++ b/WcaOnRails/app/models/registration.rb
@@ -172,7 +172,7 @@ class Registration < ApplicationRecord
   end
 
   def paid_waiting_list_info
-    pending_paid_registrations = competition.registrations.pending.order(:last_payment_date)
+    pending_paid_registrations = competition.registrations.pending.order(registration_payments.map(&:created_at).max)
     index = pending_paid_registrations.index(self)
     OpenStruct.new(index: index, length: pending_paid_registrations.length)
   end
--- a/WcaOnRails/app/helpers/application_helper.rb
+++ b/WcaOnRails/app/helpers/application_helper.rb
@@ -267,4 +267,29 @@ module ApplicationHelper
                                        opts: options.to_json))
     component_container + script_tag
   end
+
+  # Converts given integer to ordinal number as string (e.g. 1 to "1st", 2 to "2nd", 6 to "6th", etc.)
+  def int_to_ordinal(num)
+    result = num.to_s
+    last_two_digits = num % 100
+
+    # Special case for 11th, 12th and 13th
+    if last_two_digits >= 11 && last_two_digits <= 13
+      result += "th"
+    else
+      last_digit = num % 10
+
+      if last_digit == 1
+        result += "st"
+      elsif last_digit == 2
+        result += "nd"
+      elsif last_digit == 3
+        result += "rd"
+      else
+        result += "th"
+      end
+    end
+
+    return result
+  end
 end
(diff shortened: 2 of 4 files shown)


KerrieJarman:
Thanks Deni, great start. Is there any differentiation between competitors that have paid and those that haven't?

dmint789:
> Thanks Deni, great start. Is there any differentiation between competitors that have paid and those that haven't?

There isn't, this simply shows the competitor's position on the waiting list.

KerrieJarman:
That is going to be problematic for us and for many other organisers but not all. For us we only accept competitors when they pay. But I recognise that not all communities do it like that.  Maybe there needs to be a way to differentiate between competitions that take payment via the website and those who use an external payment method then for those who do use the website payments only count paid competitors. 
There might also be a few other reasons a competitor is on the waiting list maybe the organizer hasn't got around to approve their registration yet, so they are not really on the waiting list just unapproved. Maybe it is a series competition and they can't be accepted to that competition as they have also registered for the other competition.

dmint789:
> That is going to be problematic for us and for many other organisers but not all. For us we only accept competitors when they pay. But I recognise that not all communities do it like that. Maybe there needs to be a way to differentiate between competitions that take payment via the website and those who use an external payment method then for those who do use the website payments only count paid competitors.

There is a field in the db that tells us if the competition uses Stripe payments, so that's easy to implement in terms of code.

> There might also be a few other reasons a competitor is on the waiting list maybe the organizer hasn't got around to approve their registration yet, so they are not really on the waiting list just unapproved.

I disagree with this, as their registration might not be approved even if they paid (e.g. if the competitor limit has been reached), so I think we should still display them as being on the waiting list, because no one is guaranteed to have their registration approved.

So it seems like we just need to change the way we display the competitor's position on the list for competitions with Stripe enabled and all other comps can use the implementation I made. How should it differ though? As far as I'm aware, the db doesn't differentiate between those who have already paid with stripe and those who haven't, they're just on the same waiting list. I also don't see why we'd want to show just the competitor's position on the list of those who have paid, because someone in front of them on the waiting list might just pay in the next minute and then it wouldn't make sense that your position on the list went up, as any queue should be expected to just have your position go down over time. Am I missing something?

Also, I'm wondering if involvement of the WCAT is necessary here.

dmint789:
There has been some internal communication on Slack where we decided to treat comps that use Stripe differently. If the competitor has paid, their position on the waiting list will be shown ordered by the "paid on" field instead of "registered on". If the competitor has not paid, they will be shown a message telling them to do so.

I haven't tested that this functions correctly for comps with Stripe though, as there is no information about registration payments in the development db. Please advise on how I could test this with mock data.

WcaOnRails/app/models/registration.rb
```diff
@@ -171,6 +171,12 @@ def waiting_list_info
     OpenStruct.new(index: index, length: pending_registrations.length)
   end
 
+  def paid_waiting_list_info
+    pending_paid_registrations = competition.registrations.pending.order(registration_payments.map(&:created_at).max)
+    index = pending_paid_registrations.index(self)
+    OpenStruct.new(index: index, length: pending_paid_registrations.length)
```
gregorbg, line 177: Avoid using `OpenStruct`, it is deprecated. With the newer RuboCop versions, the overcommit checks will also warn about this.

WcaOnRails/app/models/registration.rb
```diff
@@ -171,6 +171,12 @@ def waiting_list_info
     OpenStruct.new(index: index, length: pending_registrations.length)
   end
 
+  def paid_waiting_list_info
+    pending_paid_registrations = competition.registrations.pending.order(registration_payments.map(&:created_at).max)
```
gregorbg, line 175: I don't understand this. From what I can tell, `.order` is a method for retrieving data from the database using the `ORDER BY` SQL clause. Essentially, it sorts the data you're receiving.

You're passing in the **maximum** of the `created_at` dates, so it feels like "please order this list according to the 17th of January 1983" instead of saying "please order this list according to the creation date". What is your intention behind this line of code?

WcaOnRails/app/models/registration.rb
gregorbg: I understand that you did this only because the above `waiting_list_info` does the same. But if you pull in the latest changes, you will notice that I have already changed that :D

gregorbg (changes requested):
Some cosmetic changes, plus I'm very confused question about retrieving sorted paid registrations :sweat_smile:

dmint789:
I will close this draft PR, as it should probably be redone at this point, and we should wait to finalize internal discussions about how to treat payments and the waiting list order.

Issue #6378, opened 2021-07-20
Add the competition events to the emailed report


druidalek:
I would like to propose adding the events of a competition in the emailed report. 

Since we already have the date, location, number of competitors and equipment I don't see a reason not to include the events. You can argue that you are able to do this by opening the competition on the website, but I think this will save some clicks and time. 

The reason for adding the events is when I read the report I would like to know what type of a competition this was before I start reading the report. Was it only a few events, was it an all-event competition or maybe just fewest moves or only blind events, etc. As an organizer and delegate I would benefit from having this information directly in the email.

Jambrose777:
I've labeled this under WRC/WQAC as this is for the reports. I overall think that it would be a very nice addition to our reports. Even though the schedule is only a click away, so is other information presented. like date/location/competitors.

PR #6592, opened 2021-12-27
Add competition events to delegate report

kr-matthews:
Addresses #6378: events are added to the delegate report as html icons (with tooltips). I assume that will show up properly in the email version of the report? 

On a closely related note, I think it would be nice to show the competitor limit when the competitor count isn't available, since this also helps get an idea of the type of competition (as mentioned in the original issue). Something along the lines of:

**Competitors** Unknown, results are not posted yet. (Competitor limit was xy.)

Could I add that onto this PR, or would it be better to create a new issue and PR?

### DIFF
--- a/WcaOnRails/app/views/delegate_reports/_delegate_report.html.erb
+++ b/WcaOnRails/app/views/delegate_reports/_delegate_report.html.erb
@@ -41,7 +41,7 @@
   <div>
     <strong>Events</strong>
     (<%= competition.events.count %>)
-    <%= competition.events.map { |event| event.name }.join(", ") %>
+    <%= competition.events.map(&:name).join(", ") %>
   </div>
 
   <div>
--- a/WcaOnRails/app/views/delegate_reports/_delegate_report.html.erb
+++ b/WcaOnRails/app/views/delegate_reports/_delegate_report.html.erb
@@ -40,10 +40,8 @@
 
   <div>
     <strong>Events</strong>
-
-    <% competition.events.each do |event| %>
-      <%= cubing_icon(event.id, data: { toggle: "tooltip", placement: "top" }, title: event.name) %>
-    <% end %>
+    (<%= competition.events.count %>)
+    <%= competition.events.map { |event| event.name }.join(", ") %>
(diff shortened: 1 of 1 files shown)


gregorbg:
Better create a new PR for that competitor limit feature, because we need to discuss an issue with the icons separately

gregorbg (changes requested):
Thanks for the clean implementation! I'm blocking this PR with my review because I have doubts that it will show up properly in an email client.

Our current cubing icons are implemented by referencing the NPM `@cubing/icons` package which is installed into our frontend assets bundle. Emails and Email clients do not reference that bundle, nor do they know about the CSS provided by `@cubing/icons` that implements these icons.

I am not an expert on email, so I might be entirely wrong here; feel free to debate my standpoint. Probably the easiest way would be to send a dummy report from staging to yourself, or to install `Mailcatcher` for local development purposes.

kr-matthews:
I finally confirmed on staging that event icons won't show up in emails :frowning_face: (thanks @gregorbg for helping me to get that working). So I've replaced the icons with the names (plus event count), as pictured.

image


It would be nice to have the icons when viewing reports on the website, and only use the event names in emails, but I don't know if there's an easy way to do that.

WcaOnRails/app/views/delegate_reports/_delegate_report.html.erb
```diff
@@ -38,6 +38,12 @@
     <% end %>
   </div>
 
+  <div>
+    <strong>Events</strong>
+    (<%= competition.events.count %>)
+    <%= competition.events.map { |event| event.name }.join(", ") %>
```
viroulep, line 44: Just noticed this was on staging so I thought I'd checkout the diff to see if it can be applied easily on top of my work on adding/editing results; in ruby you can easily create a proc returning the attribute using `&:attribute`, ie you can do this:
```suggestion
    <%= competition.events.map(&:name).join(", ") %>
```
Not saying you should change it, just letting you know ;)

gregorbg (approved):
LGTM :ship:

Issue #6597, opened 2021-12-31
Add G5 Timers to Delegate Report template


Epride:
**Is your feature request related to a problem? Please describe.**
Now that Speedstacks Generation 5 timers are approved for use in WCA competition the Delegate report section related to equipment should add a line related to them

**Describe the solution you'd like**
Just adding the a line below Gen 4 Pro Timer for Gen 5 Pro Timer

PR #6598, opened 2021-12-31
Add line for Gen 5 Pro Timer to delegate reports

kr-matthews:
Addresses #6597.

1. When I run locally to test changes, all the reports are blank (makes sense). So technically I haven't confirmed that the changes work as expected, but it's an extremely simple change.
2. It seems a bit odd to add a default amount of 0 to each line, given that we ask people to delete unused lines.


gregorbg:
Thanks! Also, fair remark about the 0 default. My guess (I wasn't involved with the original decision) would be that if a Delegate forgets to delete the line, the 0 makes clear that the equipment was not used -- as opposed to leaving it blank by default, which could be ambiguous when accidentally left in the report.

gregorbg (approved):

PR #6617, opened 2022-01-10
Hide birthdays in registration by default.

### LINKED ISSUE #5160: Hide DoB by default in registrations
When registrations are viewed by organizers and/or Delegates the dates of birth are initially visible. Given that there is rarely a use for these to be seen (first timers ID verification is the main reason) and having them visible by default causes more problems than it solves I'd like to suggest the inverse of #3934 : Adding DoB to the list of columns that are hidden by default and have the button to show them at the top of the page.

kr-matthews:
Resolves #5160.

Like events and full email addresses, birthdays are now hidden by default and can be toggled via a button.

Previously, with just events and full emails, it was only possible to show one or the other. Now, any combination of the three is possible to show.

I also made the capitalization consistent on the buttons.

If there's a better way to do this, let me know and I can redo it. I'm still not very familiar with Rails at this point.

### DIFF
--- a/WcaOnRails/app/controllers/registrations_controller.rb
+++ b/WcaOnRails/app/controllers/registrations_controller.rb
@@ -37,7 +37,9 @@ class RegistrationsController < ApplicationController
 
   def edit_registrations
     @show_events = params[:show_events] == "true"
-    @show_full_mail = params[:show_email] == "true"
+    @show_full_emails = params[:show_full_emails] == "true"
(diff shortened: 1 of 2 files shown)


gregorbg (approved):
I'm not too wild about booleans being passed around as strings, but I acknowledge that this is the way it has been done before and you just kept consistent with the existing code in this PR.

LGTM!

PR #6620, opened 2022-01-16
Shared cumulative time limits fix

### LINKED ISSUE #6502: Cumulative Limit Shared Between Rounds Bug
**Describe the bug**
Whenever changing a cumulative limit that is shared with another round, the shared aspect glitches out.

**To Reproduce**
1. Go to '/competitions/{competitionId}/events/edit'
2. Add 2 events with 1 round each. Ie. Click 'Add Event' next to 4x4x4 Blindfolded and 5x5x5 Blindfolded. Then for each event select 1 Round in the dropdown menu.
3. Click the box for 'Time Limit'
4. Set up a Shared Cumulative Limit. Ie. Click the Cumulative Radio Button. Click 'Share with other Rounds'. Select the other Round created.
**Notice** Under the explaination text, 5x5x5 Blindfolded Round 1 is listed as a round that the cumulative limit is shared with (see Screenshot 1).
5. Click Ok.
**Notice** The time limit list the events and rounds for the shared cumulative limit.
6. Reclick the box for 'Time Limit' (Same as step 3).
7. Click Ok. - **This is where the Bug happens.**

- The Time Limit now no longer list that it is shared.
- If you open The Time limit editor box back up, 5x5x5 Blindfolded is no longer listed under the explanation text.
- If you then Click the 'share with other rounds' Button, 5x5x5 Blindfolded is already checked to share the limit with! (Screenshot 2)
- At that point you have to deselect 5x5x5 Blindfolded, press ok, then reopen the share with other rounds menu and reselect it in order to readd it.

**Expected behavior**
The cumulative Time limit should still be shared whenever editing the time limit for an event.

**Screenshots**
image
image

**Additional context**
For Delegates/Organizers Deselecting and Reselecting the event the time limit is shared with is easy enough, but this can easily slide under the radar and the person editing the event not even realize that the events page will no longer say the time limit is shared.

kr-matthews:
Resolves #6502.

## There were a few problems
1. When updating a shared cumulative time limit, those shared rounds (and the round itself) were not updated properly (as reported in #6502).
2. When changing from a shared cumulative time limit to a per solve time limit, those shared rounds were not updated.
3. When removing rounds which were part of a shared time limit, those shared rounds were not updated.
4. When deleting events which had rounds which were part of a shared time limit, those shared rounds were not updated.

## Example of current behaviour
If there are the following shared time limits:
- 2x2, 3x3, 4x4, 5x5
- 6x6, 7x7, clock, mega

And then 5x5 is changed to share with 4x4, 6x6, 7x7, and pyra, then the new shared time limits will be:
1. 2x2, 3x3
2. clock, mega
3. 4x4, 5x5, 6x6, 7x7, pyra

(Previously, this would show 2x2 and 3x3 in a group of 2x2, 3x3, 4x4, 5x5, and incompatibly show 4x4 and 5x5 in a group of 4x4, 5x5, 6x6, 7x7, pyra.)

Screenshot from 2022-01-15 16-44-52
Screenshot from 2022-01-15 16-53-39
Screenshot from 2022-01-15 17-06-47

### DIFF
--- a/WcaOnRails/app/webpacker/javascript/edit-events/modals/index.js
+++ b/WcaOnRails/app/webpacker/javascript/edit-events/modals/index.js
@@ -63,7 +63,6 @@ class EditRoundAttribute extends React.Component {
     wcifRound[this.props.attribute] = this.state.value;
 
     // TODO: still need to remove from other rounds if selecting per solve
-    // TODO: still need to remove from other rounds if deleting event
 
     // This is gross. timeLimit is special because of cross round cumulative time limits.
     // If you set a time limit for 3x3x3 round 1 shared with 2x2x2 round 1, then we need
--- a/WcaOnRails/app/webpacker/javascript/edit-events/modals/index.js
+++ b/WcaOnRails/app/webpacker/javascript/edit-events/modals/index.js
@@ -20,6 +20,8 @@ let RoundAttributeComponents = {
   advancementCondition: AdvancementConditionComponents,
 };
 
+// NOTE: this helper function modifies wcifEvents in place, and is used in other
+// React components, meaning it modifies their state/props for them
 export function removeRoundsFromSharedTimeLimits(wcifEvents, wcifRounds) {
   _.compact(_.flatMap(wcifEvents, 'rounds')).forEach(otherWcifRound =>
     _.pull(otherWcifRound.timeLimit.cumulativeRoundIds, ...wcifRounds)
--- a/WcaOnRails/app/webpacker/javascript/edit-events/modals/index.js
+++ b/WcaOnRails/app/webpacker/javascript/edit-events/modals/index.js
@@ -63,7 +63,7 @@ class EditRoundAttribute extends React.Component {
     wcifRound[this.props.attribute] = this.state.value;
 
     // TODO: still need to remove from other rounds if selecting per solve
-    // TODO: still need to remove from other rounds if deleting round/event
+    // TODO: still need to remove from other rounds if deleting event
 
     // This is gross. timeLimit is special because of cross round cumulative time limits.
     // If you set a time limit for 3x3x3 round 1 shared with 2x2x2 round 1, then we need
--- a/WcaOnRails/app/webpacker/javascript/edit-events/modals/index.js
+++ b/WcaOnRails/app/webpacker/javascript/edit-events/modals/index.js
@@ -62,22 +62,30 @@ class EditRoundAttribute extends React.Component {
     let wcifRound = this.getWcifRound();
     wcifRound[this.props.attribute] = this.state.value;
 
+    // TODO: still need to remove from other rounds if selecting per solve
+    // TODO: still need to remove from other rounds if deleting round/event
+
     // This is gross. timeLimit is special because of cross round cumulative time limits.
     // If you set a time limit for 3x3x3 round 1 shared with 2x2x2 round 1, then we need
     // to make sure the same timeLimit gets set for both of the rounds.
     if(this.props.attribute == "timeLimit") {
-      let timeLimit = this.state.value;
+      // the new time limit for this round
+      // we just want timeLimit = this.state.value, however the ids are required in the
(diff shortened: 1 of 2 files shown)


WcaOnRails/app/webpacker/javascript/edit-events/modals/index.js
```diff
@@ -20,11 +20,9 @@ let RoundAttributeComponents = {
...
 };
 
-function findRoundsSharingTimeLimitWithRound(wcifEvents, wcifRound) {
-  return _.compact(_.flatMap(wcifEvents, 'rounds')).filter(otherWcifRound =>
-    otherWcifRound !== wcifRound
-    && otherWcifRound.timeLimit
-    && otherWcifRound.timeLimit.cumulativeRoundIds.includes(wcifRound.id)
+export function removeRoundsFromSharedTimeLimits(wcifEvents, wcifRounds) {
+  _.compact(_.flatMap(wcifEvents, 'rounds')).forEach(otherWcifRound =>
+    _.pull(otherWcifRound.timeLimit.cumulativeRoundIds, ...wcifRounds)
```
gregorbg, line 25: Nitpick: As opposed to the previous `findRoundsSharingTimeLimitWithRound` function, you are now implicitly modifying the `wcifEvents` parameter in-place.

Please document that fact very clearly. If it's not too much of a hassle, try thinking of a way that the exported function doesn't `_.pull` by itself, but rather returns the events that have to be pulled, so that every React component can modify its own state/props based on the return value of this method.

gregorbg (commented):
Very thorough implementation, thanks a lot! Just one comment about the `removeRoundsFromSharedTimeLimits` method, though I'm not sure whether that's severe enough to block the entire PR. Let me know what you think!

WcaOnRails/app/webpacker/javascript/edit-events/modals/index.js
kr-matthews: I see what you mean. But the components using this helper already have the events that need to be pulled; they're passed in as the parameter `wcifRounds`. Doing what you suggest would basically amount to copy-pasting this helper function into any other component which uses it, unless I'm missing something.
I'm guessing it would be better to add a comment to the current state of the code, as opposed to having two copies of the same function in different files with some comments explaining why?

gregorbg: If the alternative would really just amount to copy-pasting, then adding a comment to this method and leaving as-is will be fine.

kr-matthews: Added comments here and with the import.

gregorbg: Thanks!
(a) Is there something like "documentation comments" in JavaScript? For example, in Java `// this is a general comment` and `/** This is a documentation comment */`.
(b) Please refrain from adding comments in the imports section, nobody ever reads these. Instead, add a comment at the call-site if necessary. I think changing the method comment to a proper documentation comment as per (a) will suffice.

kr-matthews: (a) Yes, seems like it, see [wikipedia](https://en.wikipedia.org/wiki/JSDoc).
(b) Ok makes sense. Something like this?
```
/**
 * Finds the cumulativeRoundIds of each event in wcifEvents and removes any
 * which are found in wcifRounds. Note that it modifies wicfEvents in place.
 *
 * @param {collection} wcifEvents Will be modified in place.
 * @param {Array}      wcifRounds Rounds to be removed from all cumulativeRoundIds.
 */
 ```

gregorbg: Yup, something along those lines :)

kr-matthews: Done. Thanks for bringing comments vs documentation to my attention.

gregorbg (approved):

Issue #6509, opened 2021-10-24
Change Vision & Strategy Upload to thewca/wca-documents


saranshgrover:
The vision and strategy document is currently uploaded as a raw pdf file on this repository - but Board would like to use the same format as all other documents and upload it to wca-documents instead. cc @gregorbg , making high priority since we need to upload the 2021 vision and i dont think this is a lot of work?

gregorbg:
Do you have a Markdown version of the Vision and Strategy document at hand?

The change will not be very complicated in terms of infrastructure, but someone :tm: will have to create the `.md` file for the Vision and Strategy document.

Epride:
I'm not sure if converting it to markdown is practical or even possible given the graphics involved in the document?

gregorbg:
It's certainly possible, at least as long as the graphics stand on their own. It is not possible to have text flow around a graphic in Markdown or to annotate a figure with labels and small arrows. But plain including of images is not a problem.

saranshgrover:
Im confused - why do we need a markdown file? There are pdf files in wca-documents already? Like [this](https://github.com/thewca/wca-documents/blob/master/documents/finances/2019%20Financial%20Report.pdf). cc @viroulep

viroulep:
IIRC pdf files in `wca-documents` are indeed copied as-is; you would just need to create a link to them on the website.
If this pdf document is bound to change regularly I would advise to convert it to markdown (to be able to easily follow the diff and review changes), but given it's the "Vision and Strategy" I would not expect it to change that often.

saranshgrover:
@thewca/software-team Can we process this ASAP? It is high priority

notrichardpeng:
@saranshgrover I have uploaded it to the wca-document repository as a pdf, should I remove it from here and re-link the locales to it?

gregorbg:
Closing this as per #6625

PR #6625, opened 2022-01-22
Move vision and strategy to wca-documents

notrichardpeng:
Addresses #6509 (second PR after fixing branching error in first)


gregorbg (approved):
LGTM, thanks for the thorough find/replace throughout all of the i18n files :+1:

Issue #720, opened 2016-06-13
Do not allow delegates to submit their delegate report until results are posted


pedrosino:
It's cool that we automatically get the number from the results, but in the event that results are not posted yet, we're stuck with "Competitors Unknown, results are not posted yet. ", which is kinda annoying.

~Two solutions:~
~- [ ] allow delegates to fill this if the results are not up~
~- [ ] once results are posted, come back and change this field~

As per the discussion below, we want to change things so it's not possible to submit the delegate report for a competition until the results have been posted.

jfly:
I think you're under the mistaken assumption that there's actually a field in the delegate_reports table for this?

pedrosino:
Ah, I am :(
Well, in that case I feel there should be a field for that. I remember we also discussed having the number of newcomers. As I said above, updating this once results are posted should be enough to keep information on the website correct. The discussion would be forever impaired/hindered, but we can live with that.

jfly:
I personally don't like the idea of new fields for each of these things. What I would support is requiring delegates to upload results to the website *before* they can lost their delegate report. They wouldn't have to wait for the results team to post the results, but that would let us compute these numbers for them.

pedrosino:
Ah, now that you say, it's not mandatory to have an actual field. It can just be updated on the report's text.
As of now, delegates don't upload results to the website (and given recent history, I'm not sure it's a good idea). I don't have a problem with allowing reports to be sent before results, since you may want to write the report when information is still fresh, while checking results can have other hiccups (missing DOBs, checking for newcomers, etc).
I'll try to take a look at the code for this and come up with a better proposal.

jfly:
> Ah, now that you say, it's not mandatory to have an actual field. It can just be updated on the report's text.

I'm not sure what you mean by this. Remember that we allow delegates to edit and save their report before posting them.

> As of now, delegates don't upload results to the website (and given recent history, I'm not sure it's a good idea)

We trust them to email a json file to results@, I'm just suggesting they upload the json file to the website instead of emailing them. I'm sure there will be some unexpected pain at first, but that always happens when you change things.

> I don't have a problem with allowing reports to be sent before results

I'm trying to make a distinction between a delegate *uploading* results and a results team member *posting* results. I suggest that we don't let a delegate post their delegate report until after they have uploaded their results (so we can compute the number of competitors).

> since you may want to write the report when information is still fresh, while checking results can have other hiccups (missing DOBs, checking for newcomers, etc).

[...]

PR #6628, opened 2022-01-31
Enforce submitting results before posting report

kr-matthews:
In delegate reports, added new function `can_see_submit_button?` and modified original `can_submit_delegate_report?` to also check whether results have been submitted (not necessarily posted) (plus rename to `can_submit?`).
If results haven't been posted, the button will be visible but disabled, with a tool tip explaining.
Also added an explanatory sentence to the 'report is not posted yet' box.
One test modified.

tooltip on disabled button

### DIFF
--- a/WcaOnRails/app/views/delegate_reports/edit.html.erb
+++ b/WcaOnRails/app/views/delegate_reports/edit.html.erb
@@ -26,8 +26,8 @@
     <%= f.input :remarks, input_html: { class: "markdown-editor markdown-editor-image-upload" } %>
 
     <%= f.button :submit, class: "btn-primary" %>
-    <% can_see_delegate_report_submit_button = DelegateReport.find(@delegate_report.id).can_see_delegate_report_submit_button?(@current_user) %>
-    <% can_submit_delegate_report = DelegateReport.find(@delegate_report.id).can_submit_delegate_report?(@current_user) %>
+    <% can_see_delegate_report_submit_button = @delegate_report.can_see_delegate_report_submit_button?(@current_user) %>
+    <% can_submit_delegate_report = @delegate_report.can_submit_delegate_report?(@current_user) %>
     <% if can_see_delegate_report_submit_button %>
       <%= button_tag(type: 'submit',
                      name: "delegate_report[posted]",
--- a/WcaOnRails/app/views/delegate_reports/edit.html.erb
+++ b/WcaOnRails/app/views/delegate_reports/edit.html.erb
@@ -26,13 +26,16 @@
     <%= f.input :remarks, input_html: { class: "markdown-editor markdown-editor-image-upload" } %>
 
     <%= f.button :submit, class: "btn-primary" %>
+    <% can_see_delegate_report_submit_button = DelegateReport.find(@delegate_report.id).can_see_delegate_report_submit_button?(@current_user) %>
     <% can_submit_delegate_report = DelegateReport.find(@delegate_report.id).can_submit_delegate_report?(@current_user) %>
-    <% if can_submit_delegate_report %>
+    <% if can_see_delegate_report_submit_button %>
       <%= button_tag(type: 'submit',
                      name: "delegate_report[posted]",
                      value: true,
+                     disabled: !can_submit_delegate_report,
                      class: "btn btn-danger",
-                     data: { confirm: "You are about to post your report, are you sure you want to do this? Once posted, a report cannot be changed." }) do %>
+                     data: { confirm: "You are about to post your report, are you sure you want to do this? Once posted, a report cannot be changed." },
+                     title: can_submit_delegate_report ? "" : "You must first submit results before you can post the report.") do %>
         Post the report
       <% end %>
     <% end %>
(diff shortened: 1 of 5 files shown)


kr-matthews:
I'm not sure why the first two tests are failing - they are about changing registration open dates of locked competitions, which I wouldn't think would be effected by my changes.
I can see why the third test is failing - the test competition doesn't have results submitted and so the 'post the report' button can't be clicked.

I'm not sure how to fix these and I'm having trouble running tests locally. If somebody could help me or point me in the right direction, that would be helpful.

gregorbg:
99.9% sure the two unrelated test fails were due to https://github.com/thewca/worldcubeassociation.org/pull/6636.
The third test is indeed related to your new feature; what exactly is the error you're getting when trying to run tests?

gregorbg (changes requested):
Some smaller remarks about the implementation.

My main concern is about the "procedural" implications for the whole WCA competition workflow. What is the reason for blocking DR submissions unless Results are posted? Who came up with this idea?
What if there is a legitimate problem with Results (Newcomer who forgot ID, controversial WRC decision that could affect results, .....) that keeps the Report from being posted although it's not strictly the Delegate's fault?

kr-matthews:
> My main concern is about the "procedural" implications for the whole WCA competition workflow. What is the reason for blocking DR submissions unless Results are posted? Who came up with this idea? What if there is a legitimate problem with Results (Newcomer who forgot ID, controversial WRC decision that could affect results, .....) that keeps the Report from being posted although it's not strictly the Delegate's fault?

These points were discussed in the linked issue (#720), and it seemed like it was agreed that this was worth doing. Two board members said they liked it or were ok with it. Maybe it's worth revisiting the discussion now, before merging this, since that discussion was a while ago?

Thanks for the implementation comments, I'll address them once we confirm we still want to move forward with this.

> 99.9% sure the two unrelated test fails were due to #6636. Rebase on top of `master` to pull the fixes into your feature branch.

Yes.

> The third test is indeed related to your new feature; what exactly is the error you're getting when trying to run tests?

`delegate_report_spec.rb` test attempts to view, edit, save, and submit the report, but it fails to find the 'post the report' button since it's disabled now (because results haven't been submitted). I assume the best way to fix the test is to mock submitting results before trying to post, and I'm not sure of the proper way to do that.

WcaOnRails/app/views/delegate_reports/edit.html.erb
```diff
@@ -26,13 +26,16 @@
     <%= f.input :remarks, input_html: { class: "markdown-editor markdown-editor-image-upload" } %>
 
     <%= f.button :submit, class: "btn-primary" %>
+    <% can_see_delegate_report_submit_button = DelegateReport.find(@delegate_report.id).can_see_delegate_report_submit_button?(@current_user) %>
```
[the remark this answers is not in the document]
kr-matthews: I was just copying the original definition of `can_submit_delegate_report` so I wasn't really thinking about it, but yes `@delegate_report` makes sense and it seems to work.

WcaOnRails/spec/factories/competitions.rb
```diff
@@ -86,6 +86,7 @@
     trait :with_valid_submitted_results do
       announced
       with_rounds { true }
+      results_submitted_at { Time.now }
```
gregorbg, line 89: Is there a specific reason why you dont't refer to the `results_posted` trait here, similar to the reference to `announced` two lines above?

gregorbg (commented):
Looking good now, just one general comment about your implementation

WcaOnRails/spec/factories/competitions.rb
kr-matthews: We only need results to be _submitted_, not necessarily _posted_, for the report to be submitted/posted. So just using the trait `:results_posted` in the test itself won't make it pass.

On a side note, it might be worth adding `results_submitted_at { ... }` to the `:results_posted` trait.

gregorbg: AH! I completely missed the fact that there's a difference between `results_submitted_at` and `results_posted_at`. Sorry, my bad, your explanation makes sense! But then we have to reconsider the change: Posting a Delegate Report has to work even after the Results were *posted*. Does your code support that case?

gregorbg (changes requested):
My unfortunate negligence about there being a difference between `results_submitted_at` and `results_posted_at` raised an interesting concern: Does this change support cases where the Results are actually posted (as in, published) before the Delegate Report is submitted?

WcaOnRails/spec/factories/competitions.rb
kr-matthews: I changed it to check whether results have been submitted or posted, just in case it's possible for results to be posted without being submitted.

I also realized that helper functions already exist for exactly these purposes, so I'm using those now. Although it seems odd that one is defined via `![...].nil?` and the other via `[...].present?`, given that these aren't the same in edge cases. (See [here](https://github.com/thewca/worldcubeassociation.org/blob/be4031eaf3e28aba027225f3823801441c271d85/WcaOnRails/app/models/competition.rb#L1147), and a few definitions below.)

gregorbg (approved):

PR #6683, opened 2022-02-06
Update GitHub token formats

gregorbg:
Diff looks like all secrets were updated because the newer Chef versions use a slightly newer encryption algorithm


FinnIckler (approved):
LGTM

PR #6684, opened 2022-02-07
Move Avatars storage to Amazon S3

gregorbg:
Up until now, user uploaded Avatars were stored in the `secrets` folder on the hard disk of our server. This continues to cause problems because (a) it costs more than S3 storage and (b) it takes ~25 minutes to spin up a new server because we have to sync over several GB worth of picture data.

By storing the Avatars on S3 and caching them via a CloudFront CDN we hope to save costs and we are also a big step closer to enabling multiple server instances with load balancing <3


saranshgrover:
Woo! So glad to see this being implemented!

KerrieJarman:
Can't comment on the change explicitly but happy to see this move forward

KerrieJarman (commented):
To my untrained eye, looks good

danieljames-dj (commented):
Thanks for the work @gregorbg. I've gone through the changes, I didn't understand completely as I'm not so fluent in Rails yet. So, I would like to give a neutral thumbs up.

gregorbg:
Having to merge this now due to the urgency of spinning up more servers without distributing 6GB worth of JPEG files among them

PR #6685, opened 2022-02-08
Make WST chore mailer text more general

gregorbg:


KerrieJarman:
Good idea! I generally just ignore the name but it makes more sense to make it generic

KerrieJarman (approved):

PR #6686, opened 2022-02-09
Disable public query for 1000 individual results

gregorbg:
...because we don't want to fry ourselves.


Jambrose777 (approved):

gregorbg:
Thank you for the lightning-fast review :D

PR #6687, opened 2022-02-10
Add .timeLimit check before pulling

kr-matthews:
[Recent PR](https://github.com/thewca/worldcubeassociation.org/commit/8c165dc4c3fa184626fae203ebe1f090f2d3b7ec) had a mistake. Need to check that the event has a time limit before modifying its time limit.


danieljames-dj (approved):
LGTM

PR #6689, opened 2022-02-11
Improve notifications for WCAT

DanielEgdal:
Include the amount of days until the competition, so WCAT can more easily see if there is a competition close to announcement deadline.

### DIFF
--- a/WcaOnRails/spec/helpers/notifications_helper_spec.rb
+++ b/WcaOnRails/spec/helpers/notifications_helper_spec.rb
@@ -70,7 +70,7 @@ RSpec.describe NotificationsHelper do
         notifications = helper.notifications_for_user(wcat_member)
         expect(notifications).to eq [
           {
-            text: "#{confirmed_competition.name} is pending announcement. The competition is happening in #{competition.days_until} days.",
+            text: "#{confirmed_competition.name} is pending announcement. The competition is happening in #{confirmed_competition.days_until} days.",
             url: admin_edit_competition_path(confirmed_competition),
           },
           {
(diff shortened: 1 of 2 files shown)


gregorbg (approved):

DanielEgdal:
Thanks! Should be updated now.

WcaOnRails/spec/helpers/notifications_helper_spec.rb
```diff
@@ -70,7 +70,7 @@
         notifications = helper.notifications_for_user(wcat_member)
         expect(notifications).to eq [
           {
-            text: "#{confirmed_competition.name} is waiting to be announced",
+            text: "#{competition.name} is pending announcement. The competition is happening in #{competition.days_until} days.",
```
gregorbg, line 73: Attention: You have accidentally renamed the variable during copy/paste. Please re-use `confirmed_competition` from the previous version of the test

gregorbg (changes requested):

WcaOnRails/spec/helpers/notifications_helper_spec.rb
DanielEgdal: Good catch - I had not noticed the the difference as I thought the context was identical.

gregorbg: You're still missing the second occurence of `competition.days_until`

DanielEgdal: .. thanks!

gregorbg (approved):

Issue #6589, opened 2021-12-20
Add Trainee Delegates role to WCIF


DanielEgdal:
It should be possible to extract trainee delegates from the WCIF files. 

Main motivation, solve this: https://github.com/thewca/wca-live/issues/74
I thought that should be easy to fix by adding trainees somewhere in this file (maybe I'm missing more) https://github.com/thewca/wca-live/blob/master/client/src/components/admin/AdminAccess/StaffMembersTable.js

But then I realised that the WCIF (which I think WCA live uses pretty much exclusively) does not include knowledge of trainees. 

I think the fix is to add a new line here https://github.com/thewca/worldcubeassociation.org/blob/master/WcaOnRails/app/models/user.rb#L1083 for trainees.
A change is probably also needed here. https://github.com/thewca/worldcubeassociation.org/blob/master/WcaOnRails/app/models/competition.rb#L1569

Is there more that needs to be done? If it is just the bottom two links I mentioned to fix the WCIF part that would be great. Otherwise please point directions or take over the implementation. :D

PR #5249, opened 2020-03-09
implementation of trainee delegates

danieljames-dj:
Fixes #5118.
Didn't write the test cases, so not sure if the test cases will pass. if it doesn't pass, I'll try writing the test cases as well 😬

PR #6690, opened 2022-02-11
Add Trainee Delegates role to WCIF

DanielEgdal:
Fixes https://github.com/thewca/worldcubeassociation.org/issues/6589

Again, not fully sure if this is complete, or I'm missing to account for something.

### DIFF
--- a/WcaOnRails/app/models/competition.rb
+++ b/WcaOnRails/app/models/competition.rb
@@ -1566,7 +1566,7 @@ class Competition < ApplicationRecord
       registration = registrations.find { |reg| reg.user_id == wcif_person["wcaUserId"] }
       # Note: person doesn't necessarily have corresponding registration (e.g. registratinless organizer/delegate).
       if registration && wcif_person["roles"]
-        roles = wcif_person["roles"] - ["delegate", "trainee_delegate", "organizer"] # These three are added on the fly.
+        roles = wcif_person["roles"] - ["delegate", "trainee-delegate", "organizer"] # These three are added on the fly.
         registration.update!(roles: roles)
       end
       if registration && wcif_person["assignments"]
--- a/WcaOnRails/app/models/competition.rb
+++ b/WcaOnRails/app/models/competition.rb
@@ -1566,7 +1566,7 @@ class Competition < ApplicationRecord
       registration = registrations.find { |reg| reg.user_id == wcif_person["wcaUserId"] }
       # Note: person doesn't necessarily have corresponding registration (e.g. registratinless organizer/delegate).
       if registration && wcif_person["roles"]
-        roles = wcif_person["roles"] - ["delegate","trainee_delegate", "organizer"] # These three are added on the fly.
+        roles = wcif_person["roles"] - ["delegate", "trainee_delegate", "organizer"] # These three are added on the fly.
         registration.update!(roles: roles)
       end
       if registration && wcif_person["assignments"]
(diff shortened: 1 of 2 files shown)


gregorbg:
We may need to update the WCIF specification: https://github.com/thewca/wcif/blob/master/specification.md?plain=1#L130
As this is a non-breaking change, I think we can proceed without a major version bump in WCIF.

DanielEgdal:
Yes, that makes sense. I have submitted a PR there as well https://github.com/thewca/wcif/pull/4 (Side note, I noticed the version was 1.0 despite previous changes. I just bumped the minor by 1 anyway).

WcaOnRails/app/models/competition.rb
```diff
@@ -1566,7 +1566,7 @@ def update_persons_wcif!(wcif_persons, current_user)
       registration = registrations.find { |reg| reg.user_id == wcif_person["wcaUserId"] }
       # Note: person doesn't necessarily have corresponding registration (e.g. registratinless organizer/delegate).
       if registration && wcif_person["roles"]
-        roles = wcif_person["roles"] - ["delegate", "organizer"] # These two are added on the fly.
+        roles = wcif_person["roles"] - ["delegate", "trainee_delegate", "organizer"] # These three are added on the fly.
```
jonatanklosko, line 1569:
```suggestion
        roles = wcif_person["roles"] - ["delegate", "trainee-delegate", "organizer"] # These three are added on the fly.
```

WcaOnRails/app/models/user.rb
```diff
@@ -1081,6 +1081,7 @@ def to_wcif(competition, registration = nil, registrant_id = nil, authorized: fa
     person_pb = [person&.ranksAverage, person&.ranksSingle].compact.flatten
     roles = registration&.roles || []
     roles << "delegate" if competition.delegates.include?(self)
+    roles << "trainee_delegate" if competition.trainee_delegates.include?(self)
```
jonatanklosko, line 1084:
```suggestion
    roles << "trainee-delegate" if competition.trainee_delegates.include?(self)
```

DanielEgdal: Could you elaborate on why to use a hyphen instead of underscore? I used an underscore as this was how Trainees were implemented in this PR https://github.com/thewca/worldcubeassociation.org/pull/5249

jonatanklosko: See https://github.com/thewca/wcif/pull/4#discussion_r805412204, it's just to align with the conventions in WCIF, since that's what we are building here :)

In all other places underscores are fine as they follow the conventions of Ruby code.

gregorbg: Simply put, there is a difference between what we do within our Ruby code (underscore) and what we publicly expose to the outside world (cleanly defined WCIF spec, which happens to use hyphens)

DanielEgdal: Alright, thanks! Should all be good now then.

jonatanklosko (approved):

DanielEgdal:
I guess it can be merged since it is approved now?

jonatanklosko:
Ah sure, sorry!

jonatanklosko:
I've just deployed this and the wca-live counterpart, so Trainee Delegates should have the expected access, thanks @DanielEgdal :)

PR #6693, opened 2022-02-14
Bump Chef versions to enable Tarsnap again

gregorbg:
Kinda useful having access to our old backups... :'D
Plus bonus PHP installation for WRT, didn't want to open a separate PR for that now that we desperately need to deploy tonight


FinnIckler:
LGTM!

PR #6703, opened 2022-02-18
Fix name warnings for numbers

DanielEgdal:
You get the error "Note: The competition id starts with a lowercase letter. We prefer to have it start with an uppercase." when a competition name starts with a number. This seems quite inaccurate, so here is a regex update.

### DIFF
--- a/WcaOnRails/app/models/competition.rb
+++ b/WcaOnRails/app/models/competition.rb
@@ -384,7 +384,7 @@ class Competition < ApplicationRecord
         warnings[:name] = I18n.t('competitions.messages.name_too_long')
       end
 
-      unless /^[[:upper:]]/.match(self.id)
+      unless /^[[:upper:]]|^\d/.match(self.id)
         warnings[:id] = I18n.t('competitions.messages.id_starts_with_lowercase')
       end
 
--- a/WcaOnRails/spec/models/competition_spec.rb
+++ b/WcaOnRails/spec/models/competition_spec.rb
@@ -431,6 +431,12 @@ RSpec.describe Competition do
       expect(competition.warnings_for(nil)[:id]).to eq I18n.t('competitions.messages.id_starts_with_lowercase')
     end
 
+    it "do not warn if competition id starts with a number" do
+      competition = FactoryBot.build :competition, id: "1stNumberedComp2021"
+      expect(competition).to be_valid
+      expect(competition.warnings_for(nil)[:id]).to eq nil
+    end
+
     it "warns if advancement condition isn't present for a non final round" do
       FactoryBot.create :round, competition: competition, event_id: "333", number: 1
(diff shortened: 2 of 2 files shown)


gregorbg:
Thank you for the PR submission! Can I assume this is green-lit by WCAT? Is there a concrete example of a valid competition name that starts with a number but is currently being rejected by our website?

DanielEgdal:
This competition gave the 'note'/warning (but wasn't rejected) by the website a those ~2 days ago when I made the PR. https://www.worldcubeassociation.org/competitions/1stSriLankanOpen2022
There is nothing wrong with competitions with a number and nothing from the WCRP prohibiting it.
I did send on this PR to WCAT and got some funny reaction of the message displayed didn't make sense for the above competition.

gregorbg:
Okay, thanks for the additional context! Would you mind adding a test that represents this new intended behavior? It can be similar to https://github.com/thewca/worldcubeassociation.org/blob/master/WcaOnRails/spec/models/competition_spec.rb#L428

DanielEgdal:
Not fully sure if I'm taking this the right way (as you know Ruby is not a language I have received education in). What you want is a new test case for a competition, and it should assertTrue in the sense that there is no warning. Does this mean that the last line in the new test case should be:
```rb
expect(competition.warnings_for(nil)[:id]).to eq nil
```
Or is there a different way to assert that the error message doesn't show up?

gregorbg:
Yup, that generally makes sense.

A word of advice: When dealing with a new language, try not to confuse yourself too much by the fact that you haven't worked with the language before.
Right now, the basic task is to establish test cases. The concept "this specific ID should generate a warning" and "this specific ID should not generate a warning" does not depend on whether you have received education in Ruby or not :smile: 
The next step will be to formalise it a little bit, like "if the ID is this and that, I expect a warning message that states 'hey dude dis wrong' " or "if the ID is like this and that, I expect there to be no warning message". Note that this is still fully independent of the programming language Ruby.
Only the last step will be to translate your assumptions into actual Ruby code. And in this case that should be easily possible by copy/pasting and modifying existing code. As you've shown with your suggested snippet, you managed to solve that pretty easily by yourself :wink:

gregorbg:
To let you in on a secret, I have no clue why you have to pass `nil` to the `warnings_for` method. I just accept that this is how that madness called "Ruby" decides to do things and go on with my day. So really I'm trying to motivate you not to be too intimidated by an obscure programming language :blush:

DanielEgdal:
Thanks! Yes, copying (or heavily being inspired from existing code) is how I have been able to help a little with website stuff. Thankfully this is somewhat an option thanks to a lot the useful concepts already being implemented (though the schema thing with the DB has tripped me up in the past xD)

gregorbg (approved):
Cool, thank you for the additional test!

PR #6706, opened 2022-02-20
Fix a bug with registration update where you can bring yourself back to the waiting list

viroulep:
When editing registrations is allowed, competitors could throw themselves back to the waiting list by editing their registration on the page loaded right after the first registration, but after the organizer accepted them.

This bug exists because we rely on the user-provided status in the params to check if we can allow updates to `accepted_at/by` and `deleted_at/by`; the status' param should only be used when competition's admins are editing the registration (which is correctly done in registration_params).

I went for the following changes:
  - prevent `accepted_at` and `deleted_at` from being permitted unless an admin is editing the registration
  - fill them as nil when a user update their registration from the deleted status (the only case they update them is when they want to register again)
  - don't rely on the params' status to check if the status has changed, instead rely on the accepted_at/deleted_at params

I also added a couple of tests for the registration edit features, nothing too fancy but at least it show the bug without the fix.

### DIFF
--- a/WcaOnRails/app/controllers/registrations_controller.rb
+++ b/WcaOnRails/app/controllers/registrations_controller.rb
@@ -373,12 +373,21 @@ class RegistrationsController < ApplicationController
       return
     end
     registration_attributes = registration_params
-    # Don't change status columns if the status is the same.
-    if @registration.checked_status.to_s == params[:registration][:status]
-      registration_attributes = registration_attributes.except(:accepted_at, :accepted_by, :deleted_at, :deleted_by)
-    end
     was_accepted = @registration.accepted?
     was_deleted = @registration.deleted?
+    # The only case we go to this endpoint if the registration was deleted is when we register again.
+    if was_deleted
+      # Set the accepted_at/deleted_at to nil iff it's not already set,
+      # which can happen when moving from deleted to accepted.
+      registration_attributes = { accepted_at: nil, deleted_at: nil }.merge(registration_attributes)
+    end
+    # Don't rely on the status in the params, compute the new status from the
+    # timestamps.
+    new_status = Registration.status_from_timestamp(registration_attributes[:accepted_at], registration_attributes[:deleted_at])
+    # Don't change status columns if the status is the same.
+    if @registration.checked_status == new_status
+      registration_attributes = registration_attributes.except(:accepted_at, :accepted_by, :deleted_at, :deleted_by)
+    end
     if current_user.can_edit_registration?(@registration) && @registration.update(registration_attributes)
       if !was_accepted && @registration.accepted?
         mailer = RegistrationsMailer.notify_registrant_of_accepted_registration(@registration)
@@ -617,14 +626,12 @@ class RegistrationsController < ApplicationController
     permitted_params = [
       :guests,
       :comments,
-      :accepted_at,
-      :deleted_at,
       registration_competition_events_attributes: [:id, :competition_event_id, :_destroy],
     ]
-    params[:registration][:deleted_at] = nil
-    params[:registration][:accepted_at] = nil
     if current_user.can_manage_competition?(competition_from_params)
       permitted_params += [
+        :accepted_at,
+        :deleted_at,
         :accepted_by,
         :deleted_by,
       ]
--- a/WcaOnRails/app/models/registration.rb
+++ b/WcaOnRails/app/models/registration.rb
@@ -45,14 +45,18 @@ class Registration < ApplicationRecord
     !accepted? && !deleted?
   end
 
-  def checked_status
-    if accepted?
+  def self.status_from_timestamp(accepted_at, deleted_at)
+    if !accepted_at.nil? && deleted_at.nil?
       :accepted
-    elsif pending?
+    elsif accepted_at.nil? && deleted_at.nil?
       :pending
     else
       :deleted
     end
+  end
+
+  def checked_status
+    Registration.status_from_timestamp(accepted_at, deleted_at)
   end
 
   def new_or_deleted?
--- a/WcaOnRails/spec/controllers/registrations_controller_spec.rb
+++ b/WcaOnRails/spec/controllers/registrations_controller_spec.rb
@@ -212,6 +212,63 @@ RSpec.describe RegistrationsController do
     end
   end
 
+  context "signed in as competitor where they can edit their registration" do
+    let!(:user) { FactoryBot.create(:user, :wca_id) }
+    let!(:competition) { FactoryBot.create(:competition, :registration_open, :editable_registrations, :visible) }
+    let(:threes_comp_event) { competition.competition_events.find_by(event_id: "333") }
+
+    before :each do
+      sign_in user
+    end
+
+    it "cannot change their own registration status" do
+      registration = FactoryBot.create :registration, :accepted, competition: competition, user_id: user.id
+      expect(registration.accepted?).to eq true
+      patch :update, params: {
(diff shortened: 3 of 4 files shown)


WcaOnRails/app/controllers/registrations_controller.rb
```diff
@@ -373,12 +373,21 @@ def update
       return
     end
     registration_attributes = registration_params
+    was_accepted = @registration.accepted?
+    was_deleted = @registration.deleted?
+    # The only case we go to this endpoint if the registration was deleted is when we regiter again.
```
gregorbg, line 378: Nit: Typo in `register`

WcaOnRails/app/controllers/registrations_controller.rb
```diff
@@ -373,12 +373,21 @@ def update
...
+    if was_deleted
+      # Set the accepted_at/deleted_at to nil iff it's not already set,
+      # which can happen when moving from deleted to accepted.
+      registration_attributes = { accepted_at: nil, deleted_at: nil }.merge(registration_attributes)
```
gregorbg, line 382: Using two lines to do
```
registration_attributes[:accepted_at] = nil
registration_attributes[:deleted_at] = nil
```
is more verbose but also more explicit. I think it would be better for people who are not well-acquainted with Ruby.

WcaOnRails/app/controllers/registrations_controller.rb
```diff
@@ -373,12 +373,21 @@ def update
...
+    end
+    # Don't rely on the status in the params, compute the new status from the
+    # timestamps.
+    new_status = Registration.new(registration_attributes.slice(:accepted_at, :deleted_at)).checked_status
```
gregorbg, line 386: I get the point of doing this, but constructing a dummy database entity with only two parameters, and only using it to compute one status check feels very wrong.

Is there any chance of outsourcing this into a static/module method? Something along the lines of `Registration.compute_status(accepted_at, deleted_at)` (which could then also be re-used within the `Registration` class)

gregorbg (changes requested):
Thank you for your PR and your keen observations about the registration status!
I hardly dare criticise the work of my predecessor in office, but I've still added a few remarks :)

WcaOnRails/app/controllers/registrations_controller.rb
viroulep: It would be more verbose, but unfortunately not semantically equivalent:
  - your suggestion says "take whatever is in `registration_attributes`, and override the two keys`.
  - mine says "take these two keys, and override with whatever is in `resgistration_attributes` (which is the behavior we want, given these keys may be set by `registration_params`).

Also I learnt ruby by looking at code that looked ruby-ish and not rubbish (mostly written by Jeremy, Jonatan, and al.), I'd argue that intentionally writing non ruby-ish code is actually a wrong bet for the readers; reading ruby-ish code they will certainly ask themselves what it does and hopefully learn some language feature from it.
Especially given that `Hash::merge` is a fairly common idiom among programming languages :)

WcaOnRails/app/controllers/registrations_controller.rb
viroulep: Thanks for spotting this!

WcaOnRails/app/controllers/registrations_controller.rb
viroulep: Right it did feel weird to me to, and it didn't fully convince myself because there is no guarantee that in the future constructing a registration from just this would work.
I can try the static way but I'm afraid it will be ugly too; I'll check what it looks like.

WcaOnRails/app/controllers/registrations_controller.rb
gregorbg: Okay... I accept your argument for Ruby-ish code from a perspective of motivating people to learn something new. But the fact that I screwed up trying to figure out what this line of code does kinda underlines my point that it is too packed/complex right now.

If there is no other concise way to solve this, feel free to discard my comment and resolve this conversation. But I just wanted to consider it in the first place.

viroulep: This is an alternative way of doing the same thing:
```suggestion
      unless registration_attributes.has_key?(:accepted_at)
        registration_attributes[:accepted_at] = nil
      end
      unless registration_attributes.has_key?(:deleted_at)
        registration_attributes[:deleted_at] = nil
      end
```

Now I'm generally used to receive comments like "your code does the thing, but did you know there is this construct in the language that does it for you?", but it's the first time I'm actually arguing to not change my code to something worse.

> But the fact that I screwed up trying to figure out what this line of code does kinda underlines my point that it is too packed/complex right now.

I don't agree with that:
  - there is a single method invocation on this line, which is definitely not "packed/complex" considering how often methods are chained in this repository.
  - it shows that until that PR the semantic of `a.merge(b)` was not clear for you.
It's worth noting that *specifically* to address this kind of situation and to make it clear to the reader, I wrote a comment that explain the intention of the line, and actually explain exactly what it does.

Since the suggestion I made at the beginning of this comment is semantically equivalent I'll let you pick whatever version you judge more suitable, my main concern is getting this merged asap since this bug is quite easy to trigger.

About the "write code readable for beginner point": if you really want this repository to follow this rule, I would suggest to make it clear somewhere what constructions we can or cannot use.
Indeed it's definitely hard to know where to draw the line between what construct is fine or not, and who draw that line.
For instance should we write this:
```ruby
new_array = []
for item in old_array do
  new_array << item.some_attribute
end
```
instead of `old_array.map(&:some_attribute)`, which is definitely not clear for someone who doesn't know ruby?

This map construct and the merge ones are quite used in our codebase (320 and 46 occurrences respectively), if we go this way and decide that some construct must be avoided, should we update all our codebase to get rid of them?

gregorbg: Wow, this is a very thorough explanation. I appreciate and thank you for elaborating your argument, but at the same time I also apologise if I got you on the wrong foot there. It's absolutely fantastic that you stepped in to fix this bug and by no means did I intend to disregard your effort.

This PR will be merged after I finish writing this comment, but I want to make a point why I brought this up in the first place. This is not about a fundamental discussion which attributes/methods are "allowed" in our codebase and which ones aren't, and neither is it my attempt to become a "benevolent dictator" who decides what's readable enough and what isn't. It's simply my observation that in the current WST there are a lot of motivated members who keep struggling with Ruby and the style of code that Ruby deems idiomatic.

That being said, there is a significant difference between (trivial) `map` invocations and this `merge` invocation. Your example about `old_array.map(&:some_attribute)` reads almost like an (English) sentence would, in terms of "subject - verb - object". Take a bunch of elements and _map_ them to _some\_attribute_.
The merge construct doesn't follow that logic in that it first lists a hash with (multiple!) deliberate `nil` values and then packs a pre-filled hash on top of that. Unless you know the specifc implementation of `merge` in whatever programming language you happen to be using (so this isn't even my usual Ruby-bashing!), you cannot immediately derive the expected behavior when keys clash.
Those are all difficulties that do not arise with simpler idioms like `map` or `select`/`filter`, which is also reflected by the numbers you gave: The usages of `map` exceed `merge` by a factor of roughly 7 to 1.

Anyways, this is already too much of a debate about personal taste and I really don't want this detail discussion to taint the fact that I am very grateful for your quick and thorough bugfix :heart: Your argument has a strong point and you're getting my approval as a peace offering now :smile:

gregorbg (approved):
This is ready to `merge` now :wink:

PR #6707, opened 2022-02-20
Fix credentials for AWS S3

### LINKED ISSUE #6705: Internal error when trying to connect to AWS S3 in development
**Describe the bug**

When running the website in development mode, opening the homepage lead to the posts index crashing, and trying to login crashes too with the following internal error:
```
Aws::Sigv4::Errors::MissingCredentialsError at /posts.json
missing credentials, provide credentials with one of the following options:
```

Can anyone actually work in dev mode with the current website state, and if yes how?
(having the private AWS secrets in the dev environment is not the answer I'm looking for :p)

Looking [here](https://github.com/thewca/worldcubeassociation.org/blob/master/WcaOnRails/config/initializers/carrierwave.rb#L20-L25) it seems we may want to provide some credentials in dev mode, but I doubt it's actually desirable by default unless we have dev credentials too.

Unless I'm missing something we have a couple of options here:
  - if we do have public credentials, we should put them in the environment.
  - if we don't, we should enable stub responses in dev mode too (this is what I did to get past the error).
  - if we want to enable temporary use of prod credentials for development purposes, I'd suggest to enable stubs if the aws key id or the aws secret is empty.

What do you think?
I can provide a patch if one of my suggestions sounds good.

viroulep:
Fixes #6705.

I also got an error running rspec that "Spring reloads, and therefore needs the application to have reloading enabled.
Please, set config.cache_classes to false in config/environments/test.rb.".

I think Spring is supposed to run in the test environment so I guess the setting should be set to false, but I'm not sure why it does work in the CI.

### DIFF
--- a/WcaOnRails/env_vars.rb
+++ b/WcaOnRails/env_vars.rb
@@ -8,7 +8,7 @@ EnvVars = Env::Vars.new do
     mandatory :SMTP_PASSWORD, string
   end
 
-  if Rails.env.development?
+  unless Rails.env.production?
     mandatory :DISABLE_BULLET, :bool
 
     optional :AWS_ACCESS_KEY_ID, :string, ''
--- a/WcaOnRails/config/initializers/carrierwave.rb
+++ b/WcaOnRails/config/initializers/carrierwave.rb
@@ -14,12 +14,13 @@ CarrierWave.configure do |config|
 
   aws_credentials = {
     region: EnvVars.S3_AVATARS_REGION,
-    stub_responses: Rails.env.test?, # Optional, avoid hitting S3 actual during tests
   }
 
   # only development needs explicit credentials to access AWS.
   # in production, access is derived implicitly through the IAM role of the machine
-  if Rails.env.development?
+  if EnvVars.AWS_ACCESS_KEY_ID.blank? || EnvVars.AWS_SECRET_ACCESS_KEY.blank?
+    aws_credentials[:stub_responses] = true
+  else
     aws_credentials['access_key_id'] = EnvVars.AWS_ACCESS_KEY_ID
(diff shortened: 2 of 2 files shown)


gregorbg:
Thanks for the PR! As for caching classes, Spring does actually **not** run in the tests. See the workflow YAML in the `.github` folder, it sets an environment variable `DISABLE_SPRING=1`.

That being said, I have no idea whether it's a good idea to run or not to run Spring in tests. I simply pasted the existing Travis CI when I migrated to GitHub Actions.

viroulep:
Ah thanks I got mislead because it is installed in tests in the Gemfile; I'll revert that part of the changes.

I guess it's mostly useful when testing locally as the application code changes a bit.

WcaOnRails/env_vars.rb
```diff
@@ -8,7 +8,7 @@
     mandatory :SMTP_PASSWORD, string
   end
 
-  if Rails.env.development?
+  unless Rails.env.production?
```
gregorbg, line 11: This change seems trivial enough, but I'm wondering whether you have any specific reason for including it in this PR?

viroulep: Yes, without it `EnvVars.AWS_ACCESS_KEY_ID` and `EnvVars.AWS_SECRET_ACCESS_KEY` are not declared and can't be used in the test environment; the only side effect is to include `DISABLE_BULLET` in the changes, but since I noticed the `.env.test` does define it I figured it was fine.

gregorbg: Okay, thanks :)

gregorbg (approved):
LGTM after clarification by Philippe.

PR #6709, opened 2022-02-23
Fix minor test nuisances

gregorbg:
- Dynamically determine whether or not to `cache_classes` during testing, depending on whether or not Spring is running
- Update to a version of `Apparition` JS driver that doesn't spam our logs


gregorbg:
Merging because the only two "failing" tests were two pending spec that I reverted just now.

PR #6712, opened 2022-02-28
Upgrade to Rails 6.1

gregorbg:
Prerequisite for using Ruby 3.

Although it's only a minor version bump, there are two non-trivial changes that I'd like some feedback on:
1. Some config values have new defaults (or some config settings simply didn't exist before): https://edgeguides.rubyonrails.org/configuring.html#default-values-for-target-version-6-1. Tests check out locally, but I'd like a manual review and cross-check of that list nonetheless
2. ActiveModel error models have changed significantly. In particular, we cannot simply concatenate `:id` errors to the `:name` errors for some entities like competitions. I have formulated an approach that _works_ in the sense that it should do the same thing as before, but it's not pretty. I'm very open to suggestions how to improve this (without using `concat`, because the errors array is immutable now)


danieljames-dj (commented):
I went through the code, from my very limited knowledge on rails, this looks good (I didn't understand many parts as well).

PR #6735, opened 2022-03-01
Manually update momentjs-rails to 2.29.1.1

gregorbg:
For some reason, Dependabot cannot figure this one out. Probably a missing Git tag? Whatever...
Treating this as a standard Dependabot PR, that is I will merge as soon as tests pass.


gregorbg:
Note to Kerrie: In the monthly digest you can treat this like a normal automated dependency update by Dependabot.

PR #6754, opened 2022-03-04
Align registration info table footer

kr-matthews:
When birthdays were hidden, it was misaligned.

Also add comments for clarity on other empty columns in footer.

image


gregorbg (approved):
Thanks for fixing this so promptly, and thanks for the bonus comments!
LGTM

PR #6765, opened 2022-03-13
Add link to main competition page in result submission mail

SAuroux:
Just a small addition so that there is a direct link to the competition main page in the result submission mail, implemented analogously to the competition confirmation mail for WCAT.


danieljames-dj (approved):
Changes looks good to me

PR #6767, opened 2022-03-14
Update README.md - table name RoundTypes

OJCubing:
The `RoundTypes` table is currently referred to as `Rounds` in the README.md file given by the public export.

### DIFF
--- a/webroot/results/admin/export/template.README.md
+++ b/webroot/results/admin/export/template.README.md
@@ -69,7 +69,7 @@ The database export consists of these tables:
 | Results                                 | WCA results per competition+event+round+person     |
 | RanksSingle                             | Best single result per competitor+event and ranks  |
 | RanksAverage                            | Best average result per competitor+event and ranks |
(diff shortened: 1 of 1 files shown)


danieljames-dj (approved):
Thanks for pointing this out and fixing, @OJCubing. From the `sorry_message` inside Rounds table, it's clear that the name was changed to RoundTypes and somehow we missed to update the same in the README.

OJCubing:
I just realised in the tsv export, the RoundTypes table is called `WCA_export_Rounds.tsv` - Does anyone know where I can find this file to rename it?

danieljames-dj:
Hi @OJCubing. I just downloaded the latest export and tested. Looks like the file `WCA_export_Rounds.tsv` is having only the `sorry_message` and couldn't find any issues. Could you please check again if it's fine?

Also, if you find any issue, it will be better if you create it as a separate issue in [issues page](https://github.com/thewca/worldcubeassociation.org/issues) because since this PR is already merged, there is very high chance that this can get skipped.

PR #6774, opened 2022-03-19
Server Infrastructure Changes: Moving From Tarsnap and add Server Classes

FinnIckler:
Hey, 
I hope making one PR for this is fine, because these are both quite minor changes code wise.

This PR changes the severs.sh and backup.sh scripts to move away from Tarnsap and give the user the opportunity to choose which instance type he wants to run.
The PR also includes some checks in the servers.sh scripts that were broken (user name check and 200 check [which triggers on stripe errors])

The non-code changes are:
- Created an S3 Bucket called wca-backups and configured Versioning and Lifecycle Policies
- Added permissions to our prod IAM instance profile to write to that bucket
- Did an initial sync to that bucket

Thanks,
Finn


chef/site-cookbooks/wca/recipes/base.rb
```diff
@@ -6,3 +6,4 @@
 package 'tmux'
 package 'tree'
 package 'ntp'
+package 'awscli'
```
gregorbg, line 9: No EOL at the end of the file. Please fix by adding a blank line at the end.

scripts/servers.sh
```diff
@@ -405,14 +412,6 @@ function passthetorch() {
     exit 1
   fi
 
-  server_status=`echo "$curl_result" | tail -1`
-  if [ "${server_status}" != "200" ]; then
```
gregorbg, line 409: I don't feel comfortable skipping an entire sanity check just because we don't like the way it behaves.
We should ideally change the behavior of the underlying check (a) not to return 50X when there are bad Stripe charges (b) change the check to accept 50X if the HTML body contains the word "Stripe" (c) actually care about the specific Stripe charges that are currently causing the error.

scripts/servers.sh
```diff
@@ -199,7 +204,9 @@ new() {
     --key-name $keyname \
     --instance-type $instance_type \
     --security-groups "SSH + HTTP + HTTPS" \
+    --iam-instance-profile "avatar_upload_role"
```
gregorbg, line 207: Why are you specifically mentioning Avatar upload here? Would it not make sense to have more general instance profiles like "production machine" that among others include the permission to upload avatars?

gregorbg (commented):
Thank you Finn! Some general comments from my side.

Also, it seems like you're pushing to a branch called `MoveFromTarsnap` on this repository. Please try to use your own fork if possible and merge from that.

scripts/servers.sh
FinnIckler: Just created a prod_role with the correct permissions

scripts/servers.sh
FinnIckler: Ok, I put it back in

gregorbg (approved):
LGTM :partying_face:

PR #6785, opened 2022-03-24
patch(gh-6783): modify RESULT_AFTER_DNS_WARNING

### LINKED ISSUE #6783: Improve RESULT_AFTER_DNS_WARNING result validator
WRT recently noticed that the [RESULT_AFTER_DNS_WARNING](https://github.com/thewca/worldcubeassociation.org/blob/master/WcaOnRails/lib/results_validators/individual_results_validator.rb#L7) result validator is only triggered if a DNS is followed by a valid result. However, a DNS followed by a DNF is just as strange. Looking at the [comment](https://github.com/thewca/worldcubeassociation.org/blob/master/WcaOnRails/lib/results_validators/individual_results_validator.rb#L168) in the code, this might even have been intended initially.

Change the implementation so that a DNS followed by a DNF also triggers the warning.

JoelGritter:
Resolves #6783 

- modify logic for RESULT_AFTER_DNS_WARNING to include instances of DNF following DNS
- update unit tests to include new case

### DIFF
--- a/WcaOnRails/lib/results_validators/individual_results_validator.rb
+++ b/WcaOnRails/lib/results_validators/individual_results_validator.rb
@@ -168,7 +168,7 @@ module ResultsValidators
       # Now let's try to find a DNS result followed by a non-DNS result
       first_index = all_solve_times.find_index(SolveTime::DNS)
       # Just use '5' here to get all of them
-      if first_index && all_solve_times[first_index, 5].select(&:complete?).any?
+      if first_index && all_solve_times[first_index, 5].select(&:valid_result?).any?
         competition_id, result, round_id, = context
         @warnings << ValidationWarning.new(:results, competition_id,
                                            RESULT_AFTER_DNS_WARNING,
--- a/WcaOnRails/lib/results_validators/individual_results_validator.rb
+++ b/WcaOnRails/lib/results_validators/individual_results_validator.rb
@@ -168,7 +168,7 @@ module ResultsValidators
       # Now let's try to find a DNS result followed by a non-DNS result
       first_index = all_solve_times.find_index(SolveTime::DNS)
       # Just use '5' here to get all of them
-      if first_index && all_solve_times[first_index, 5].select(&:valid_result?).any?
+      if first_index && all_solve_times[first_index, 5].select(&:started_attempt?).any?
         competition_id, result, round_id, = context
         @warnings << ValidationWarning.new(:results, competition_id,
                                            RESULT_AFTER_DNS_WARNING,
--- a/WcaOnRails/lib/solve_time.rb
+++ b/WcaOnRails/lib/solve_time.rb
@@ -135,6 +135,10 @@ class SolveTime
     !complete?
   end
 
+  def valid_result?
+    !dns? && unskipped?
+  end
+
   def time_seconds
     time_centiseconds / 100.0
   end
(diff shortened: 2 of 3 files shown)


JoelGritter:
Also, forewarning: I couldn't get the Docker build running on my local machine for some reason. Every time I tried to run the tests it would kill the container, without any visible warnings except `compilation failed`. Very peculiar.

danieljames-dj:
Thanks @JoelGritter for working on it. The changes looks good to me as well. But please wait for it to get reviewed from someone experienced as well. Meanwhile, could you please try to squash the two commits into one?

JoelGritter:
@danieljames-dj Pardon my ignorance, wouldn't squash and merge be done on merge?

danieljames-dj:
@JoelGritter  Sorry for the mistake, you are right. Squash will be done on merge. Maybe once someone experienced reviews it, this is good to be merged.

JoelGritter:
> I am only concerned about the terminology of `complete?` vs `valid_result?`. Your implementation considers `DNF` as a "valid" result, while DNS is not. It would make more sense to me to call it something like `started_attempt?` which could include DNF but explicitly rules out DNS.

`started_attempt?` makes more sense to me! I wasn't entirely sure what to name it either, but I think this is an improvement. 👍

gregorbg (approved):
LGTM, thank you for your contribution :)

danieljames-dj:
@JoelGritter Looks like there is some error in the 'run tests' part (unit tests). Could you please check whether you covered the unit tests for this warning?

JoelGritter:
@danieljames-dj I'm pretty sure the test failure is from the unit test I added to cover this case, rather than the code change causing the existing unit tests to fail. I haven't had the chance to take a closer look yet; I'm not sure why it's failing.

gregorbg:
Kindly ping :) Do you need support for looking into the failing tests?

JoelGritter:
@gregorbg Good callout! If there's anyone that could help with troubleshooting the unit test, that would be great!

Chances are I've made a simple error; this is my first time using Ruby. Could also be a data issue; that I didn't mock the data properly.

gregorbg:
Hm, here's my take on the issue...
For your tests, you added a round of `444` to `competition1` in the entire test method. Of course, all of the other validator checks (i.e. all the verifications that already existed before your PR) now also run on this round of 444. One of those "older" checks detects a problem with the new round of 444 that you haven't added to the list of "expected problems".

To solve this, you have to think (irrespective of Ruby or your new feature) what these feature checks should return/detect in general, and then implement those "expected errors" in the Ruby test.

Probably it will boil down to another `warns << something` invocation where you have to figure out the `something` part.

JoelGritter:
@gregorbg Thanks for the analysis! I didn't think about it that way before.

I've made a slight change to the PR: Adding a new 444 round to the competition2 instead should mean that other tests shouldn't warn about the 444 results, except for the specific check for this feature.

gregorbg:
The run failed again. Ignore the first error about `Claim WCA ID`, that's an inconsistent choke.

The second error is the actual problem. I think your moving of the 444 event from comp1 to comp2 changes nothing, the older validation checks still detect something that is not within the "expected detections" list.

JoelGritter:
I can't get the local Docker build working, and don't have the time to debug it either. I'll close this PR for now, if someone else wants to finish this off, go for it!

PR #6499, opened 2021-10-11
Reasons Menu for Registration Deletion

notrichardpeng:

PR #6786, opened 2022-03-31
Registration Deletion Reason Menu

notrichardpeng:
This time I'll be trying to implement this without mixing React and Ruby together, and only using Ruby :)
Previous closed PR: #6499


notrichardpeng:
Trying to figure out a way to pass different parameters into controllers based on different html element values...

WcaOnRails/app/assets/javascripts/registrations.js
```diff
@@ -28,6 +28,11 @@ onPage('registrations#edit_registrations', function() {
     if(!confirm("Delete the " + $selectedRows.length + " selected registrations?")) {
       e.preventDefault();
     }
+
+    var $reason = $('#delete-reason').val()
+    var $comment = $('#delete-comment').val()
+    console.log($reason);
```
gregorbg, line 34: Just noting that these two `console.log` should be removed before the feature goes live

gregorbg (commented):
Regarding submitting the choice of the user back to the controller, look into `simple_form_for`. You would have to create a model, sth along the lines of `RegistrationDeletion` which takes a reason and a freetext comment.

You can then use the convenience helpers that `simple_form` defines for you (the gem itself is well-documented and heavily used in our existing code for inspiration) and it will `POST` that back to the controller when submitting.

An additional benefit of creating a model (that might seem overkill in the first place) is that you can enumerate the deletion reasons with string constants and then add proper i18n localosation to those constants in the `en.yml` file.

notrichardpeng:
After discussing with Gregor, given that we are migrating to React and this PR is done with Ruby, we will close this PR and instead implement the feature later as part of the larger React migration.

Issue #6635, opened 2022-02-04
Scenario prohibiting Delegates from editing events 


Jambrose777:
**Describe the bug**
Setting up a competition a specific way causes the Delegate to be unable to edit a person's registration.

**To Reproduce**
- Create a Competition 
- Set "Deadline for competitors to update their registered events" to Blank
- Set "Registration edits" to "Competitors may edit their registration until the deadline for updating events"
- Have someone register for the competition.
- Accept the registration
- Attempt to edit the registration by removing/adding an event.
- See 500 Error

**Expected behavior**
Should be allowed to edit the registration successfully.

KerrieJarman:
I don't understand - why is "Deadline for competitors to update their registered events" not a mandatory field?

Jambrose777:
> I don't understand - why is "Deadline for competitors to update their registered events" not a mandatory field?
Leaving the field blank states this line:
"We encourage everyone to register for the events they want to compete in via your online registration, however you may add events to your registration up until the event has started."

Jambrose777:
This also prevents Delegates from deleting registrations on the edit registration page.

Jambrose777:
I'd like to reemphasize this as high priority for April.
I currently cannot edit my own registration's events due to this bug.
As a Delegate the `/edit/registrations` clicking edit and changing an event and click save causes the page to error out.
Clicking on the register button as if I was a competitor causes the page to error out.
<img width="1494" alt="image" src="https://user-images.githubusercontent.com/11577241/161170979-9f8150d8-c67a-4cd1-b48c-7ee3444fe6a0.png">

Jambrose777:
Its come to my attention that many competitors are unable to edit their events/registration for BuccaneerBonanza2022. They get an error when trying to do so. Might be related to this.

dunkOnIT:
Closing, as it looks like the relevant PR is merged. @Jambrose777 please reopen if you are not comfortable with the solution.

PR #6819, opened 2022-04-16
Add check that deadline date is present

kr-matthews:
Addresses #6635 - the steps described no longer produce an error for me. I'm not too familiar with Ruby or this part of the code though, so I could easily have overlooked something.

Solution based on @gregorbg's investigation [here](https://worldcubeassociation.slack.com/archives/C024WGWFUHY/p1644598698102479).


danieljames-dj (approved):
LGTM

PR #6829, opened 2022-04-28
Various fixes to dues estimates

viroulep:
This relates to #6552.

I noticed an error 500 on the competition page, and pulling the string made me fix a couple of issues:

### Fix a server error when requesting a conversion

This may actually be my bad; I naively thought that `something.updated_at` would refer to some shared object and the db, but it's definitely not the cases for `Money`'s rates, so my suggestion to use a job was actually a very bad one.
I basically reverted the code to update the rates as part of the request.

### Fix the dues not being computed on first load

I was frustrated that the dues would not be computed when loading the form, and I noticed the change handlers were not registered *after* the page was loaded so I wrapped them in a `$(function() { ... })`.

### Fix multiple onChange needlessly triggered

I think the `UpdateDues` was called thrice on page load, because each onChange handler was triggered when registered.
I kept the manual trigger on the last one since it's the one that matters.

I also fix a couple of syntax stuff, and removed the unneeded `compLimit`/`tempLimit` since we can get the information by querying the field.

PS: using the "diff settings" -> "ignore whitespace changes" should make the diff less disturbing.


WcaOnRails/app/views/competitions/_competition_form.html.erb
```diff
@@ -206,22 +206,25 @@
...
-          var tempLimit = null;
-
-          <% countryBand = CountryBand.where(iso2: @competition.serializable_hash[:country_iso2]).map(&:number)[0] %>
-          <% if !countryBand.nil? && countryBand > 0 %>
-            bandFeeUSD = <%= CountryBand::BANDS[countryBand][:value] %>;
-          <% end %>
 
+          // Note: unfortunately we cannot trigger a change on the currency
+          // code update, because we don't have access to the currency
+          // "subunit_to_unit" from the js world.
```
viroulep, line 213: I placed this comment because I wondered why we were not updating dues estimate on currency change, and then later understood we would need the `subunit_to_unit` for every currencies.

gregorbg:
Merci Philippe :) I am looping in @notrichardpeng because he worked on the initial implementation.

notrichardpeng (approved):
Thanks for fixing some of the issues I couldn't get around, like dues not being computed on first load (which I'm curious how wrapping the code in `$(function() { ... })` helped, maybe `$(function)` gets run upon load?), and also making the code more concise overall! I think your changes should work well, as long as there are no problems when you tried running it locally.

viroulep:
Thanks for the review!

`$(function() { ... })` is a shorthand for `$( document ).ready(function() { ... })`, which means that it's gonna be executed only when the DOM is ready, not just when this bit of js is executed.
Additionally to this, we also use some rails-specific onLoad functions like [here](https://github.com/thewca/worldcubeassociation.org/blob/master/WcaOnRails/app/assets/javascripts/competitions.js#L1).

Yes it does work locally :)

PR #6859, opened 2022-05-02
Remove an unused endpoint

viroulep:
Came across that old endpoint, it was used when the events page was still a standalone page and not a tab.


gregorbg (approved):
Thank you for your vigilance :)

PR #6860, opened 2022-05-02
Implement a warning for registration payment info

viroulep:
I already forgot twice to link a Stripe account to a competition and I'd feel really bad should it happen a third one, so I suggested something like that to the WCAT who agreed; I'll ask Daniel to come take a look at the wording.


Jambrose777 (approved):
Approved, but should wait to merge until WCAT approves wording :)

PR #6863, opened 2022-05-02
Have fun with Ruby 3

gregorbg:


gregorbg:
Merging this in an effort to clean our GH Issues list. All of the stuck dependencies are really becoming problematic.
I'm available to react if the website goes down.

PR #6864, opened 2022-05-02
Update RuboCop

gregorbg:
For most of the newer checks, I pretty much just tried to balance "impact of the change" vs usefulness.

Some cases where I disabled the checks actually seem reasonable, but having a machine do the conversion is too risky and doing it manually is too much work. But I am open for other opinions and debate regarding some of the cops that I disabled.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -79,6 +79,12 @@ Metrics/ClassLength:
 
 Metrics/CyclomaticComplexity:
   Max: 23
+  Exclude:
+    - 'WcaOnRails/app/controllers/competitions_controller.rb'
+    - 'WcaOnRails/app/controllers/registrations_controller.rb'
+    - 'WcaOnRails/app/jobs/sync_mailing_lists_job.rb'
+    - 'WcaOnRails/app/models/competition.rb'
+    - 'WcaOnRails/lib/results_validators/*.rb'
 
 Metrics/ParameterLists:
   Max: 5
@@ -86,6 +92,12 @@ Metrics/ParameterLists:
 
 Metrics/PerceivedComplexity:
   Max: 25
+  Exclude:
+    - 'WcaOnRails/app/controllers/competitions_controller.rb'
+    - 'WcaOnRails/app/controllers/registrations_controller.rb'
+    - 'WcaOnRails/app/jobs/sync_mailing_lists_job.rb'
+    - 'WcaOnRails/app/models/competition.rb'
+    - 'WcaOnRails/lib/results_validators/*.rb'
 
 Layout/LineLength:
   Max: 245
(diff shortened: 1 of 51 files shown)


WcaOnRails/app/controllers/competitions_controller.rb
```diff
@@ -532,16 +532,16 @@ def my_competitions
     competition_ids.concat(current_user.trainee_delegated_competitions.pluck(:competition_id))
     registrations = current_user.registrations.includes(:competition).accepted.reject { |r| r.competition.results_posted? }
     registrations.concat(current_user.registrations.includes(:competition).pending.select { |r| r.competition.upcoming? })
-    @registered_for_by_competition_id = Hash[registrations.uniq.map do |r|
+    @registered_for_by_competition_id = registrations.uniq.to_h do |r|
       [r.competition.id, r]
-    end]
+    end
```
viroulep, line 537: Ah TIL that `to_h` takes a block!

.rubocop.yml
```diff
@@ -212,3 +224,327 @@ Style/HashTransformKeys:
 
 Style/HashTransformValues:
   Enabled: true
+
```
viroulep, line 227: Generic comment on the file, rather than on that specific line: I'm actually not too sure what was the rationale for including or not a rule in there.
Since there are a lot of rules I'd expect to include them only if we change the default, but I don't know for sure if we've done that with the past updates.

Did you include all the rules that were added/available or did you pick just those for which we needed to change the default configuration?
(and do you have an opinion on if we should include everything or not?)

WcaOnRails/lib/city_comma_region_validator.rb
```diff
@@ -4,6 +4,8 @@ class CityCommaRegionValidator < CountryCityValidator
   attr_reader :valid_regions
 
   def initialize(type_of_region:, valid_regions:)
+    super
```
viroulep, line 7: I was bugged by the CI error and checked this out locally, turns out you need to do:
```suggestion
    super()
```
But I don't know why there is an argument passed to super otherwise...

viroulep (commented):
Thanks for bumping rubocop Gregor!

I've left a couple of comments (well mostly just one general question :p).
Do you want me to take a look at some of the complexity issues you excluded?
IMO we'd better exclude some specific methods from the validations rather than a whole bunch of files, but I definitely understand you didn't go ahead to investigate everything manually :)

WcaOnRails/lib/city_comma_region_validator.rb
gregorbg: Thanks for catching the CI error so quickly!

.rubocop.yml
gregorbg: I included all of the new ones that were available, even if it meant pasting their default config.
This has two reasons:
(a) If a rule is not explicitly included, RuboCop reminds you after every update that "Hey, these cool new rules exist and could be added to your config" by printing some message to stdout. That message in turn messes with overcommit.
(b) I feel that explicitly listing stuff gives us a better overview about what features exist in the first place, without having to dig through the documentation. Especially since a lot of the checks have clear and concise names anyways.

gregorbg:
Thank you for your feedback @viroulep! As to the complexity, if you feel like you want to investigate it I won't keep you :D

Right now I took this laissez-faire approach of just disabling certain checks because of two reasons:
(a) The RuboCop version we used until now is so horribly old that it starts blocking other infrastructure version upgrades, and I want to get this released ASAP in a minmal-invasive way.
(b) From a "pedagogical" standpoint I'd like other, less experiences members to look at the RuboCop checks that are currently marked `false` or that have `Exclude`s, as a chance to learn about our codebase and about Ruby.

gregorbg:
Thank you for your feedback @viroulep! As to the complexity, if you feel like you want to investigate it I won't keep you :D

Right now I took this laissez-faire approach of just disabling certain checks because of two reasons:
(a) The RuboCop version we used until now is so horribly old that it starts blocking other infrastructure version upgrades, and I want to get this released ASAP in a minmal-invasive way.
(b) From a "pedagogical" standpoint I'd like other, less experiences members to look at the RuboCop checks that are currently marked `false` or that have `Exclude`s, as a chance to learn about our codebase and about Ruby.

PR #6867, opened 2022-05-03
Fixes #6761 - Sends automatic email to WEAT every month on changes in…

### LINKED ISSUE #6761: Send email to WEAT on member changes in team/committee towards the month end
It will be good if an automatic email is sent to WEAT on changes in team/committee of that month, so that it will be easy for compiling the monthly digest.

In future, maybe we can include delegates as well (when the new table structure for the delegates is ready), so that 80% of monthly digest work of WEAT will be automated.

danieljames-dj:
… teams

### DIFF
--- a/WcaOnRails/app/mailers/wca_monthly_digest_mailer.rb
+++ b/WcaOnRails/app/mailers/wca_monthly_digest_mailer.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+class WcaMonthlyDigestMailer < ApplicationMailer
+  include MailersHelper
+
+  def send_weat_digest_content
+    mail(
+      to: ["board@worldcubeassociation.org", "assistants@worldcubeassociation.org"],
+      reply_to: ["board@worldcubeassociation.org", "assistants@worldcubeassociation.org"],
+      subject: "WCA Monthly Digest Draft",
+    )
+  end
+end
--- a/WcaOnRails/app/mailers/wca_monthly_digest_mailer.rb
+++ b/WcaOnRails/app/mailers/wca_monthly_digest_mailer.rb
@@ -5,8 +5,8 @@ class WcaMonthlyDigestMailer < ApplicationMailer
 
   def send_weat_digest_content
     mail(
-      to: ["board@worldcubeassociation.org", "assistants@worldcubeassociation.org"],
-      reply_to: ["board@worldcubeassociation.org", "assistants@worldcubeassociation.org"],
+      to: ["assistants@worldcubeassociation.org"],
+      reply_to: ["assistants@worldcubeassociation.org"],
       subject: "WCA Monthly Digest Draft",
     )
   end
--- a/WcaOnRails/app/models/team.rb
+++ b/WcaOnRails/app/models/team.rb
@@ -220,6 +220,21 @@ class Team < ApplicationRecord
     super(options)
   end
 
+  def self.changes_in_all_teams
+    team_changes = []
+    all_teams = Team.unscoped.all
+    all_teams.each do |team|
+      current_team_changes = team.changes_in_team
+      if !current_team_changes.empty?
+        team_changes.push(current_team_changes)
+      end
+    end
+    if team_changes.empty?
+      team_changes.push("There are no changes to show.")
+    end
+    team_changes.join("\n")
+  end
+
   # rubocop:disable Metrics/CyclomaticComplexity
   # rubocop:disable Metrics/PerceivedComplexity
   def changes_in_team
--- a/WcaOnRails/app/models/team.rb
+++ b/WcaOnRails/app/models/team.rb
@@ -322,32 +322,32 @@ class Team < ApplicationRecord
       end
     end
 
-    changes_of_last_month = ""
+    changes_of_last_month = []
     if leader_changes.count + promoted_senior_members.count + new_senior_members.count + new_members.count + demoted_senior_members.count + no_more_senior_members.count + no_more_members.count > 0
-      changes_of_last_month = "Changes in #{self.name}\n"
+      changes_of_last_month.push("Changes in #{self.name}")
       if leader_changes.count > 0
-        changes_of_last_month += "\nLeaders\n#{leader_changes.join("\n")}\n"
+        changes_of_last_month.push("\nLeaders\n#{leader_changes.join("\n")}")
       end
       if promoted_senior_members.count > 0
-        changes_of_last_month += "\nPromoted Senior Members\n#{promoted_senior_members.join("\n")}\n"
+        changes_of_last_month.push("\nPromoted Senior Members\n#{promoted_senior_members.join("\n")}")
       end
       if new_senior_members.count > 0
-        changes_of_last_month += "\nNew Senior Members\n#{new_senior_members.join("\n")}\n"
+        changes_of_last_month.push("\nNew Senior Members\n#{new_senior_members.join("\n")}")
       end
       if new_members.count > 0
-        changes_of_last_month += "\nNew Members\n#{new_members.join("\n")}\n"
+        changes_of_last_month.push("\nNew Members\n#{new_members.join("\n")}")
       end
       if demoted_senior_members.count > 0
-        changes_of_last_month += "\nDemotions from Senior Member to Member\n#{demoted_senior_members.join("\n")}\n"
+        changes_of_last_month.push("\nDemotions from Senior Member to Member\n#{demoted_senior_members.join("\n")}")
       end
       if no_more_senior_members.count > 0
-        changes_of_last_month += "\nResigned/Demoted Senior Members\n#{no_more_senior_members.join("\n")}\n"
+        changes_of_last_month.push("\nResigned/Demoted Senior Members\n#{no_more_senior_members.join("\n")}")
       end
       if no_more_members.count > 0
-        changes_of_last_month += "\nResigned/Demoted Members\n#{no_more_members.join("\n")}\n"
+        changes_of_last_month.push("\nResigned/Demoted Members\n#{no_more_members.join("\n")}")
       end
     end
-    changes_of_last_month
+    changes_of_last_month.join("\n")
   end
   # rubocop:enable Metrics/CyclomaticComplexity
   # rubocop:enable Metrics/PerceivedComplexity
--- a/WcaOnRails/app/models/team.rb
+++ b/WcaOnRails/app/models/team.rb
@@ -227,7 +227,8 @@ class Team < ApplicationRecord
     sorted_users = []
     team_member_changes = {}
 
-    leader_changes = []
+    leader_appointments = []
+    no_more_leaders = []
     promoted_senior_members = []
     new_senior_members = []
     new_members = []
@@ -281,22 +282,22 @@ class Team < ApplicationRecord
       name = user.name
       if new_leader || ex_leader || new_and_ex_leader
         if new_leader
-          leader_changes.append("#{name} has been appointed as the new Leader.")
+          leader_appointments.append("#{name} has been appointed as the new Leader.")
         elsif ex_leader
           if new_member
-            leader_changes.append("#{name} is no longer the Leader, but will continue as member.")
+            no_more_leaders.append("#{name} is no longer the Leader, but will continue as member.")
           elsif new_senior_member
-            leader_changes.append("#{name} is no longer the Leader, but will continue as Senior member.")
+            no_more_leaders.append("#{name} is no longer the Leader, but will continue as Senior member.")
           else
-            leader_changes.append("#{name} is no longer the Leader and no longer a member.")
+            no_more_leaders.append("#{name} is no longer the Leader and no longer a member.")
           end
         else
           if new_member
-            leader_changes.append("#{name} was the Leader for few days and is continuing as member.")
+            no_more_leaders.append("#{name} was the Leader for few days and is continuing as member.")
           elsif new_senior_member
-            leader_changes.append("#{name} was the Leader for few days and is continuing as Senior member.")
+            no_more_leaders.append("#{name} was the Leader for few days and is continuing as Senior member.")
           else
-            leader_changes.append("#{name} was the Leader for few days and is no longer a member.")
+            no_more_leaders.append("#{name} was the Leader for few days and is no longer a member.")
           end
         end
       else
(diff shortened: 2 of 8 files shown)


WcaOnRails/app/views/wca_monthly_digest_mailer/send_weat_digest_content.html.erb
```diff
@@ -0,0 +1,44 @@
+<p>Dear WCA Executive Assistants Team,</p>
+
+<p>Here's the draft for this month digest (Please verify whether all the changes were informed to the WCA Board):</p>
+
+<% @teams = Team.unscoped.all %>
+<% @duration_start_date = Time.now.beginning_of_month - 12.month %>
+
+<div><b>Changes in Teams/Committees/Councils</b></div>
+
+<% @teams.each do |team| %>
```
gregorbg, line 10: I am not too wild about doing data computations (determining which team members have been added/removed) within the template itself. It distracts from the actual template text too much imo.

If possible, move the data computations into the handler/controller and pass them to this mail template as variables. See our monthly WST chore email as an example for how to pass variables to an erb template.

gregorbg, line 10: I am not too wild about doing data computations (determining which team members have been added/removed) within the template itself. It distracts from the actual template text too much imo.

If possible, move the data computations into the handler/controller and pass them to this mail template as variables. See our monthly WST chore email as an example for how to pass variables to an erb template.

gregorbg: If you want to go the really, really clean software architecture approach, extend the `team.rb` model by methods similar to `members_added_after(date)` and `members_removed_before(date)` instead of computing these in-place. That's really only the optional icing on the cake, though :D

WcaOnRails/spec/mailers/previews/wca_monthly_digest_mailer_preview.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+# Preview all emails at http://localhost:3000/rails/mailers/wca_monthly_digest_mailer
+class WcaMonthlyDigestMailerPreview < ActionMailer::Preview
+  def send_weat_digest_content
+    mail = nil
+    ActiveRecord::Base.transaction do
+      mail = WcaMonthlyDigestMailer.send_weat_digest_content
+      raise ActiveRecord::Rollback
```
gregorbg, line 9: Why are you actively raising a rollback here? Was this c/p'ed from somehwere, and if yes, are you fully aware of why you c/p'ed it?

gregorbg, line 9: Why are you actively raising a rollback here? Was this c/p'ed from somehwere, and if yes, are you fully aware of why you c/p'ed it?

gregorbg (changes requested):
Thanks for tackling this, although I can see a connection between your WST and WEAT duties here :wink: 

A few remarks about the architecture.

WcaOnRails/app/helpers/test_script.rb
```diff
@@ -0,0 +1,14 @@
...
+team_member_1 = FactoryBot.create :team_member, user_id: user_1.id, team_id: wrt_team.id, start_date: Time.now-10.days
+user_2 = FactoryBot.create :user
+team_member_2 = FactoryBot.create :team_member, user_id: user_2.id, team_id: wrt_team.id, start_date: Time.now-4.days
+user_3 = FactoryBot.create :user
+team_member_3 = FactoryBot.create :team_member, user_id: user_3.id, team_id: wrt_team.id, start_date: Time.now-3.months, end_date: Time.now-11.days
+user_4 = FactoryBot.create :user
+team_member_4 = FactoryBot.create :team_member, user_id: user_4.id, team_id: wrt_team.id, start_date: Time.now-5.months, end_date: Time.now-16.days
+puts(team_member_1, team_member_2, team_member_3, team_member_4)
+puts('Number of team members:', wrt_team.team_members.count)
+puts('Object of first team member:', wrt_team.team_members.first)
```
gregorbg, line 14: This code yields an unintuitive output when run with `bin/rails r path/to/test_script.rb`.

The `team_member_*` are all being printed as (apparently) valid `TeamMember` instances and `team_members.count` is printed as `4`. But `team_members.first` yields `nil` and the printed output stays blank.

My guess is that this happens because we're only mocking stuff with `FactoryBot` and these are not real `ApplicationRecord` models, so somewhere something goes wrong in the deep abyss of the Ruby Interpreter. But that's just a vague suspicion. Do you have any ideas @jonatanklosko?

jonatanklosko: Note that `team_members.count` always issues a count query to the db, so it makes sense that the result is accurate.

My first guess would be that the `team_members` association is already loaded on creation and it's an empty list (perhaps it gets loaded in a validation), so `first` is indeed nil. If that's the case then we can call `wrt_team.team_members.reload` after creating the team members, to make sure we have the up-to-date list :)

gregorbg: Ah, that makes a lot of sense! I guess I was stuck too much with the "FactoryBot only does mocks" perspective that I didn't realise an actual DB call might be the issue here. Thanks for clearing that up 🙃 

@danieljames-dj Changing the last line to use a `reload` call like Jonatan suggested fixes the issue!

jonatanklosko: > FactoryBot only does mocks

It actually doesn't mock at all, it is just a convenience for creating actual records :)

danieljames-dj: Thank you @jonatanklosko and @gregorbg :-)

danieljames-dj:
A small heads up - the latest commit is still in WIP. I have to stop working now and I didn't want to leave my changes in my laptop alone, so added the commit.

WcaOnRails/app/models/team.rb
```diff
@@ -197,4 +197,127 @@ def acronym
   def name
     I18n.t("about.structure.#{friendly_id}.name")
   end
+
+  # rubocop:disable Metrics/CyclomaticComplexity
+  # rubocop:disable Metrics/PerceivedComplexity
+  def get_changes_of_last_month?
+    changes_of_last_month = ""
+    duration_start_date = Time.now.beginning_of_month - 36.month # Change to 1.month
```
gregorbg, line 205: Local testing code left in PR

WcaOnRails/app/models/team.rb
```diff
@@ -197,4 +197,127 @@ def acronym
   def name
     I18n.t("about.structure.#{friendly_id}.name")
   end
+
+  # rubocop:disable Metrics/CyclomaticComplexity
+  # rubocop:disable Metrics/PerceivedComplexity
+  def get_changes_of_last_month?
```
gregorbg, line 203: Ruby convention: Method names only end in a question mark if they return a boolean. This method does not, so it should not end in `?`

Ruby convention: No `get_` or `set_` prefix on the method names.

WcaOnRails/spec/mailers/previews/wca_monthly_digest_mailer_preview.rb
danieljames-dj: Yes I copied and when I read about that, realized it's not needed. Removed it. :-)

WcaOnRails/app/views/wca_monthly_digest_mailer/send_weat_digest_content.html.erb
danieljames-dj: I've added the logic in team.rb.

WcaOnRails/app/models/team.rb
danieljames-dj: Removed.

WcaOnRails/app/models/team.rb
danieljames-dj: Changed.

WcaOnRails/app/models/team.rb
```diff
@@ -219,4 +219,136 @@ def serializable_hash(options = nil)
...
+        elsif ex_senior_member || new_and_ex_senior_member
+          no_more_senior_members.append(name)
+        end
+        if (ex_member || new_and_ex_member) && !new_senior_member && !new_member
+          no_more_members.append(name)
+        end
+      end
+    end
+
+    changes_of_last_month = ""
```
gregorbg, line 325: Can you build this as a list instead of a string, and then `.join('\n')` at the very end?

WcaOnRails/spec/mailers/previews/wca_monthly_digest_mailer_preview.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+# Preview all emails at http://localhost:3000/rails/mailers/wca_monthly_digest_mailer
+class WcaMonthlyDigestMailerPreview < ActionMailer::Preview
+  def send_weat_digest_content
+    mail = nil
+    ActiveRecord::Base.transaction do
+      mail = WcaMonthlyDigestMailer.send_weat_digest_content
+    end
+    mail
```
gregorbg, line 10: Ruby has implicit returns. You should be able to do something like

```ruby
def send_weat_digest_content
    WcaMonthlyDigestMailer.send_weat_digest_content
  end
end
```

as per https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/DatabaseStatements.html#method-i-transaction, where it say `[...] and returns the result of the block`.

On a more high-level sidenote, why are you even wrapping this into a transaction? You're not writing anything.

WcaOnRails/spec/models/team_spec.rb
```diff
@@ -0,0 +1,75 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Team do
```
gregorbg, line 5: Note for the entire test file: Using the `@variable` notation in test DSL is completely inconsistent with the rest of our codebase. Use `let(:variable) { block_to_initialize }` and `let!(:variable) { ... }` instead, as per the rest of our `specs` folder. Pick any random test to see examples of how this is done. (Note: `let` is only executed if you actually use the variable in a test scenario, whereas `let!` is _always_ executed)

This may also explain your test failure: `@variable` is a Ruby instance variable that is bound to god knows which instance in this test scenario because specs are set up using `rspec` which does a lot of reflection shenanigans under the hood.

WcaOnRails/spec/models/team_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+  before(:all) do
+    # Names were generated from FactoryBot.create :user
+    @names = ["Larry Ullrich", "Jeffry Kiehn", "Colby Nader", "Lakia Cremin", "Shanae Price", "Ka Hermiston", "Milton Pfeffer", "Jolie Cartwright", "Alana Monahan", "Elden Cruickshank"]
```
gregorbg, line 8: Not necessary. Just create N users (where N is any number you require for your tests) by invoking something like
```
let!(:dummy_users) { FactoryBot.create_list :user, N }
```

WcaOnRails/spec/models/team_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+  it "Added 2 new members" do
+    FactoryBot.create :team_member, user_id: @users[6].id, team_id: @wrt_team.id, start_date: Time.now - 4.days
+    FactoryBot.create :team_member, user_id: @users[7].id, team_id: @wrt_team.id, start_date: Time.now - 4.days
+
+    expected_output = [
+      "Changes in WCA Results Team",
+      "",
+      "New Members",
+      "Jolie Cartwright",
+      "Milton Pfeffer",
```
gregorbg, line 35: Refer to `users[n].name` here instead of expecting one exact name.

gregorbg (changes requested):
Some remarks mostly about the code style in your tests. I strongly reckon that these also explain the mysterious failures that you reported about on Slack

WcaOnRails/spec/models/team_spec.rb
danieljames-dj: Okay got it, I've changed to let variables.

WcaOnRails/spec/models/team_spec.rb
danieljames-dj: I'm actually giving the list sorted to WEAT, because they are supposed to show the members in alphabetically sorted. So, is it fine if I hard code names? Or is there any other solution?

WcaOnRails/spec/models/team_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+    new_leader.update_columns(end_date: Time.now - 10.days, updated_at: Time.now - 10.days)
+    FactoryBot.create :team_member, user_id: @users[0].id, team_id: @wrt_team.id, start_date: Time.now - 10.days, team_senior_member: true
+    FactoryBot.create :team_member, user_id: new_leader.user.id, team_id: @wrt_team.id, start_date: Time.now - 10.days, team_leader: true
+
+    expected_output = [
+      "Changes in WCA Results Team",
+      "",
+      "Leaders",
+      "Jeffry Kiehn has been appointed as the new Leader.",
+      "Larry Ullrich is no longer the Leader, but will continue as Senior member.",
```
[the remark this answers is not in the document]
danieljames-dj: Done.

WcaOnRails/spec/models/team_spec.rb
danieljames-dj: Done.

WcaOnRails/app/views/wca_monthly_digest_mailer/send_weat_digest_content.html.erb
```diff
@@ -0,0 +1,11 @@
+<p>Dear WCA Executive Assistants Team,</p>
+
+<p>Here's the draft for this month digest (Please verify whether all the changes were informed to the WCA Board):</p>
+
+<% @teams = Team.unscoped.all %>
+
+<div><b>Changes in Teams/Committees/Councils</b></div>
+
+<% @teams.each do |team| %>
+  <% team.changes_in_team %>
```
gregorbg, line 10: Seems like nitpick but is actually super important: You have to use `<%=` instead of `<%` to actually _print_ the result of the Ruby code into the mail

gregorbg: On that note, can you add some backup code that writes "there were no changes" if the return string is empty?

WcaOnRails/app/mailers/wca_monthly_digest_mailer.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class WcaMonthlyDigestMailer < ApplicationMailer
+  include MailersHelper
+
+  def send_weat_digest_content
+    mail(
+      to: ["board@worldcubeassociation.org", "assistants@worldcubeassociation.org"],
+      reply_to: ["board@worldcubeassociation.org", "assistants@worldcubeassociation.org"],
```
gregorbg, line 9: Does the Board want these emails? Shouldn't the purpose of this digest be to take work off them and let WEAT do the work instead?

gregorbg (approved):
Superb refactor on the tests, now they look clean and well-readable :heart_eyes: 
Just one nitpick on the mail, and then we're ready to deploy!

WcaOnRails/app/mailers/wca_monthly_digest_mailer.rb
danieljames-dj: You are right, it's not needed. I've removed it.

WcaOnRails/app/views/wca_monthly_digest_mailer/send_weat_digest_content.html.erb
danieljames-dj: Done, thanks for this.

gregorbg (approved):

PR #6886, opened 2022-05-04
Update regulations translation index

mirandapedro:
After this is merged, it would be awesome if we could get a redeploy so the changes go live right away. Thanks!


gregorbg:
Merged and live at https://www.worldcubeassociation.org/regulations/translations/

mirandapedro:
Thank you very much, Gregor!!

PR #6890, opened 2022-05-05
Fix active storage migration issue

viroulep:
I'm not sure this should be merged, but here is the issue: I tried updating my local repo because of the recent ruby/rails/everything version updates, and when running the latest migration I ran into this failure:

```
== 20220223163447 CreateActiveStorageVariantRecords: migrating ================
-- create_table(:active_storage_variant_records)
rake aborted!
StandardError: An error has occurred, all later migrations canceled:

Column `blob_id` on table `active_storage_variant_records` does not match column `id` on `active_storage_blobs`, which has type `bigint`. To resolve this issue, change the type of the `blob_id` column on `active_storage_variant_records` to be :bigint. (For example `t.bigint :blob_id`).
```

I know it's an automatically generated migration, and it seems related to [this](https://github.com/rails/rails/issues/44767), and it's also worth noting the source code for that migration has changed quite a bit in [the main branch](https://github.com/rails/rails/blob/main/activestorage/db/update_migrate/20191206030411_create_active_storage_variant_records.rb).

The issue I encountered can be fixed either by setting the version to 6.1, or by using the latest version of the migration.

The question is: is it just me?

I tried deploying the latest master to staging (motivating by my will to rebase #6360), but I haven't managed just yet to run chef, so I can't confirm or deny that this would be happening here.

@gregorbg did you manage to run these migrations on prod?
I tried to re-import the developer dump but it looks like these migrations are not in there yet since I had to run them again.
If that's just me it's likely something wrong with my db structure, I get this kind of diff when running the migrations locally:
```
-  `id` bigint(20) NOT NULL AUTO_INCREMENT,
+  `id` bigint NOT NULL AUTO_INCREMENT,
```

### DIFF
--- a/WcaOnRails/db/migrate/20220223163447_create_active_storage_variant_records.active_storage.rb
+++ b/WcaOnRails/db/migrate/20220223163447_create_active_storage_variant_records.active_storage.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 # This migration comes from active_storage (originally 20191206030411)
-class CreateActiveStorageVariantRecords < ActiveRecord::Migration[6.0]
+class CreateActiveStorageVariantRecords < ActiveRecord::Migration[6.1]
   def change
     create_table :active_storage_variant_records do |t|
       t.belongs_to :blob, null: false, index: false


gregorbg:
Hey Philippe, I managed to run the migration on prod, by locally pointing my Rails gem to the `6-1-stable` branch where the problem has been fixed, then running the Rails migration, and pasting the resulting `CREATE TABLE` DDL into phpMyAdmin on prod 😅 It's not part of the developer dump yet because the latter only runs every 3 days, and the most recent dump is from before the migration.

Does changing the index number from `6.0` to `6.1` actually fix the problem?! if so, that might be worth noting in the Rails issue that you linked. When I looked at the description there, the `6-1-stable` branch was the only viable solution that was described.

viroulep:
Ah, that's some decent level of dirtiness :p

> Does changing the index number from 6.0 to 6.1 actually fix the problem?!

Yes, at least locally.
I'm more than happy to test this on staging to see if it would work similarly.

gregorbg:
> Ah, that's some decent level of dirtiness :p

It does get the job done while no official bugfix has been released yet. Although to be honest I'm astonished such a blatant bug (especially seeing how Rails courteously refuses to even boot unless migrations are run) made it into a release without immediate remedy anyways. But I digress...

> I'm more than happy to test this on staging to see if it would work similarly.

By all means, feel free! But be warned that staging is quite far behind in terms of the huge infrastructure changes that I executed over the last 48 hours. If you run into troubles there, you can just hit me up on Slack 😄

viroulep:
> It does get the job done while no official bugfix has been released yet.

Absolutely :)

> By all means, feel free! But be warned that staging is quite far behind in terms of the huge infrastructure changes that I executed over the last 48 hours.

Thanks for getting it up to date and running!
I could confirm the issue does happen on staging, and that fixing the migration version to 6.1 did fix the issue :)

gregorbg:
Wow, fantastic! Then I shall merge this to help our other devs. Might also be worth mentioning your solution on the original issue thread :) I'll leave that up to you to earn the credit.

PR #6891, opened 2022-05-06
Only export registrations for visible competitions

viroulep:
Currently using the website with the developer dump may crash because it dumps registrations which have `competition_id` pointing to competition not present in the database (eg: by visiting the "my competitions" page).
It may be because of two things:
  - they are "not visible" competition
  - the competition may have been deleted, without the registrations being deleted.

For the first point this PR should fix it.
For the second point, someone with access to prod should run `Registration.where.not(competition_id: Competition.all.map(&:id)).size`, and if it's not 0 we should probably delete those registrations.
It should not happen because we have [this](https://github.com/thewca/worldcubeassociation.org/blob/00424633af2be306913c2631d9aa8fe2bb9d1f5d/WcaOnRails/app/models/competition.rb#L10), but who knows...

### DIFF
--- a/WcaOnRails/lib/database_dumper.rb
+++ b/WcaOnRails/lib/database_dumper.rb
@@ -563,7 +563,7 @@ module DatabaseDumper
     }.freeze,
     "registration_payments" => :skip_all_rows,
     "registrations" => {
-      where_clause: "",
+      where_clause: JOIN_WHERE_VISIBLE_COMP,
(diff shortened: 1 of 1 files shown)


Jambrose777 (approved):
Awesome!! I believe this explains a lot to why my local development is killed everytime I access the Competitions page.

Also I checked on Prod and all competition_ids in registrations table exist in the competitions table.

PR #6905, opened 2022-05-14
Migrate ReactJS integration to use the (semi-)official reactjs/react-rails gem

gregorbg:
cf. https://github.com/reactjs/react-rails

The one large advantage is that the entire React integration is being handled by a well-maintained library (from the React devs themselves!) instead of our hand-built solution.
- We don't have to worry about the extra steps like `registerComponent` and the likes anymore.
- We don't have to create packs for every new component anymore, because these are now loaded via some UJS magic :tm:

I tried taking a minimum invasive approach in general. The new library mandates that all components live under the same folder, so in some cases I had to resolve naming conflicts.
- All of the non-React code is kept under `lib` now. This folder name is completely arbitrary (the best that I being a non-frontend dev could come up with) and suggestions and/or debate are welcome!
- The import paths heavily changed. Some `export default` also had to be added so the React UJS loader could find it
- I have deleted the (now) overhead Webpacker packs wherever possible. `edit-schedule` and `edit-events` use some `renderRoot` shenanigans which I believe I veeeery basically understand but I'm definitely not proficient enough to fix/change/improve.
- The additional Babel plugins in `babel.config.ts` were suggested by the bootstrapping script of `react-rails`, cf. https://github.com/reactjs/react-rails#3-now-run-the-installers. I trusted that script since I don't understand enough Babel to ascertain whether these actually make sense/are necessary for our stack. By all means feel free to debate.

With this changeset, the general workflow for creating new ReactJS components literally becomes:
1. Write your component and store all necessary files in the `components` folder.
    1. [[ Pay attention that you export the component you want to use (as in, JS `export` statements) ]]
2. Use your component in Ruby view templates with the `react_component` helper.

The largest part of this changeset are `import` lines that changed. In any case, I'd appreciate a review by some frontend-savvy people, especially regarding the folder structure with `components` and `lib`.


jonatanklosko (approved):
I haven't used react-rails, but looks like a good step to me!

> edit-schedule and edit-events use some `renderRoot` shenanigans

In proper React the data structures kept in state shouldn't be mutated, so that React can detect when the state is updated with new data structures. It would then automatically re-render whatever it needs based on data comparisons. IIRC some of the components we have don't follow this rule, mutate the data and then call `rootRender` to forcefully re-render the whole component tree (since otherwise React wouldn't detect the change). We should definitely fix it in the long term!

WcaOnRails/app/webpacker/packs/auto_numeric.js
```diff
@@ -1 +1 @@
-import '../javascript/auto-numeric';
+import '../lib/auto-numeric';
```
kr-matthews, line 1: Do we need this file?

gregorbg: From my understanding, these one-liners are "entry points" generated by Webpack, so that the old, not-so-fancy part of Javascript that lives under `WcaOnRails/assets/javascripts` can access stuff that Webpack has compiled.

gregorbg: That being said, for this PR I just followed the mantra of "never touch a running system". Meaning I didn't contemplate on why these files were there, I just assumed they are there for a reason and the folder rename is "collateral damage" from my React folder changes.

PR #6909, opened 2022-05-16
Enable Automatic Security Updates 

### LINKED ISSUE #4537: Server has security updates pending
As can be seen when SSHing:

    > ssh cubing@worldcubeassociation.org
    Welcome to Ubuntu 14.04.5 LTS (GNU/Linux 3.13.0-116-generic x86_64)
    
     * Documentation:  https://help.ubuntu.com/
    
      System information as of Mon Aug 26 05:35:12 UTC 2019
    
      System load:  0.66               Processes:           130
      Usage of /:   54.7% of 31.36GB   Users logged in:     0
      Memory usage: 37%                IP address for eth0: 172.31.33.143
      Swap usage:   0%
    
      Graph this data and manage this system at:
        https://landscape.canonical.com/
    
      Get cloud support with Ubuntu Advantage Cloud Guest:
        http://www.ubuntu.com/business/services/cloud
    
    170 packages can be updated.
    130 updates are security updates.
    
    New release '16.04.6 LTS' available.
    Run 'do-release-upgrade' to upgrade to it.
    
    
    *** System restart required ***

I believe we recently started a new server, so we should have a process for updating. Perhaps we should also consider unattended upgrades?

FinnIckler:
Fixes #4537 by installing the unattended-upgrades package

https://help.ubuntu.com/community/AutomaticSecurityUpdates

@gregorbg Do you think there is any packages we should blacklist that it should not update?


gregorbg (commented):
Nice initiative, thanks! But is it really enough to just install the package? In the Ubuntu Readme that you linked it says

> To enable it, do:
> sudo dpkg-reconfigure --priority=low unattended-upgrades

So that would have to be incorporated into the Chef recipes as well. If you read further, the explanation states that...
> (it's an interactive dialog) which will create /etc/apt/apt.conf.d/20auto-upgrades with the following contents:

```
APT::Periodic::Update-Package-Lists "1";
APT::Periodic::Unattended-Upgrade "1";
```

EDIT: Not relevant anymore, see further comments below.

~I suggest creating a template file with exactly those contents and then install it into the preferred location. Have a look at the Chef `default` recipe to see several examples of how templates are installed and used in Chef :)~

gregorbg:
I don't have any strong feelings about pinning specific packages. Our Ruby version is pinned by `rbenv` and NodeJS is pinned by the fact that the `apt` source repository we're using is limited to the same SemVer major release anyways.

But I am concerned about someday, somehow, silently installing an upgrade that inadvertently breaks the whole website because some CLI param has been dropped or deprecated :(
When weighing the risks of not getting security upgrades vs. potentially installing version upgrades that break our toolchain, it's a really tough decision to make... What do you think?

gregorbg:
Nevermind, it seems that the official Chef `apt` cookbook has support for `unattended-upgrades`, so you can use that instead of my other suggestions with hand-made templates above.

Cf https://supermarket.chef.io/cookbooks/apt

FinnIckler:
I had the same thought process as you about accidentally breaking things, which is why I recommend that we just pull security updates by setting the origins_patterns to something like ["origin=Ubuntu,archive=trusty-security"]. I'll change it in a bit

gregorbg:
Cool! Have you triedout your suggested role JSON changes, for example by running the new updated version on staging? It should generate something similar to `/etc/apt/apt.conf.d/20auto-upgrades` as well as a potential cron at `/etc/cron.daily/apt`.

FinnIckler:
I ran `sudo chef-solo --legacy-mode --config "/etc/chef/solo.rb" -o 'role[wca]'` on staging and it successfully turned it on 
and created/configured the `10periodic`, `20auto-upgrades` and `50unattended-upgrades` files

gregorbg (approved):
Merge away good sir! 🚢

Issue #4106, opened 2019-05-01
Remove "Missing Averages" page


lgarron:
Now that we have big cube BLD means, https://www.worldcubeassociation.org/results/misc/missing_averages/ is obsolete.
(It also covers "2-means", but I don't think those are worth keeping around on the official site.)

AlbertoPdRF:
I'm adding this issue to https://github.com/thewca/worldcubeassociation.org/milestone/2 to make sure we don't forget to **not** port this :stuck_out_tongue:.

jfly:
I think to remove this, we'd want to do the following:

  - [ ] Remove `index.php` from https://github.com/jfly/worldcubeassociation.org/blob/fd5a534f7b118cb192225d911aae4fcd98122969/webroot/results/.gitignore#L24
  - [ ] Remove `webroot/results/misc/missing_averages/update7205.php`: https://github.com/jfly/worldcubeassociation.org/blob/master/webroot/results/misc/missing_averages/update7205.php
  - [ ] Remove all references to `git grep update7205` (try `git grep update7205`) specific to `missing_averages`. (Don't remove the stuff for `evolution`!)
  - [ ] Remove all references to `missing_averages` (try `git grep missing_averages`)

After this, I think the `webroot/results/misc/missing_averages/index.php` file will still exist in on our prod server (when deploying, we don't remove untracked files), so we'll need to either ssh to the server and manually rm the file, or spin up a brand new server.

denthebro:
Pinging this, because this still hasn't been fixed

Jambrose777:
This will be fixed in https://github.com/thewca/statistics.

We would still need to deleted all together.

campos20:
Fixed #6912. I took care of @jfly's list, except for the last two. git grep list only 2 lines and they look harmless

https://github.com/thewca/worldcubeassociation.org/blob/b1f0180616b6367cd0e95ab36ef6578b4679077b/webroot/results/includes/_parameters.php#L138-L139

PR #6912, opened 2022-05-17
WIP remove old statistics fixes #6910

### LINKED ISSUE #6910: Remove old statistics section
We have a new [statistics project](https://statistics.worldcubeassociation.org/), meaning we can safely remove [statistics.php](https://www.worldcubeassociation.org/results/statistics.php) page.

Delete old php portions, keep 6 degrees for now.

Delete old statistics parts (and delete 6 degrees anyways)

campos20:


webroot/results/includes/_database.php
```diff
@@ -189,11 +189,6 @@ function getAllIDs ( $rows ) {
 }
 
 function getAllEventIds                    () { return getAllIDs( getAllEvents()                    ); }
-function getAllRoundIds                    () { return getAllIDs( getAllRounds()                    ); }
-function getAllCompetitionIds              () { return getAllIDs( getAllCompetitions()              ); }
-function getAllUsedCountriesIds            () { return getAllIDs( getAllUsedCountries()             ); }
-function getAllUsedCountriesCompetitionIds () { return getAllIDs( getAllUsedCountriesCompetitions() ); }
-function getAllUsedContinentIds            () { return getAllIDs( getAllUsedContinents()            ); }
```
gregorbg, line 196: Why are you specifically deleting these but not `getAllEventIds`?
I get that some of the methods here may be unused but we're not maintaining the PHP code anymore, so it's not worth making such fine-grained changes.

I'd rather leave these methods in an unused state, instead of accidentally deleting something you (or your IDE) _think_ is unused but actually gets invoked in an obscure, dynamic way in some legacy results admin code (which then could break)

campos20: `getAllEventIds` is/was used here https://github.com/thewca/worldcubeassociation.org/blob/bdc4d8cb8da360e9d6e669cd48b94e447cb29c02/webroot/results/admin/migrate/migration9.php#L81, that's why I kept it. Anyways, I put everything back after your review =]

gregorbg (approved):
Hooray for throwing more legacy PHP out the window!!

campos20:
Waiting tests in staging before merge

campos20:
Before

image

scripts/cronned_results_scripts.sh
```diff
@@ -6,9 +6,6 @@ cd "$(dirname "$0")"/..
 # export url paramter.
 time (cd webroot/results/admin/; REQUEST_URI="export=" php export_public.php)
 
-# Update statistics page
-time (cd webroot/results/; php statistics.php update)
-
 # Update Evolution of Records
 # We need to set SERVER_NAME to avoid generating the "This is only a copy of
 # the WCA results system used for testing stuff" message.
```
gregorbg, line 11: Shouldn't the entire lower part of this script also be deleted?
As far as I'm concerned, Evolution of Records and Missing Averages have been ported over as well and don't need to be recomputed.

gregorbg: As such, the corresponding files under `webroot/results/misc` could also be deleted. I don't think we need the `evolution` and `missing_averages` folders there anymore either.

campos20:
Before, this was only fixing #6910 

Now, this is also fixing #4106

gregorbg:
I think this is good to go, what do you say? :D

campos20:
image

PR #6928, opened 2022-05-22
Prettier for javascript code

danieljames-dj:
Since we are in process of moving to React frontend, thought `prettier` will be a good tool for developers to write the code clean.

Here are some advantages of prettier:
- We can make the code pretty very easily if we have the prettier plugin in the code editor (in VSCode, we can prettify a JS code by just saving the file or applying some shortcut).
- We already do some kind of formatting checks in `eslint`, but if we are using `prettier`, the formatting rules will be handled by `prettier` and other code quality rules will be handled by `eslint` like before.
- Formatting rules will be flexible for us (by editing `.prettierrc`).
- Currently there is `.editorconfig` which I think is for Rails and prettierrc can be considered as the tool for JS. In VSCode, if we try to prettify, it will try to take rules from .prettierrc or .editorconfig. When .prettierrc is not there, it will take from .editorconfig and it will convert single quotes to double quotes but eslint will throw error if JS has double quotes. If prettierrc has separate rules, it will stay different and will be easy for future.

More details can be read here: https://prettier.io/docs/en/why-prettier.html

Do you guys think this tool will be a good addition to our website?


gregorbg:
Hey Daniel,

thanks for the initiative! I appreciate that you care about code style in a distributed team, as this is a rather "trivial" issue which is often overlooked. I have two questions regarding this suggestion, namely:
1. What is the difference between `prettier` and the [eslint-airbnb code style](https://github.com/airbnb/javascript) which we're already using?
    1. In particular, is there a risk that the may have conflicts, where one demands a semicolon at the end of the line and the other one demands no semicolon?
2. Please note that the CI currently fails because `prettier` is not happy with the style in a lot of files. If you submit a PR, please make sure that all CI checks proceed and format the files accordingly.

danieljames-dj:
Hi @gregorbg, thanks for your reply. The main difference I saw between prettier and eslint-airbnb was that prettier can act as a formatter and we can format the code from the editor itself. But after typing this statement, I just checked whether there is any way to format the code with eslint and it was easily possible. In my mind, only way to execute eslint was through command line to check if there is any problem and we will have to fix it, didn't knew the formatter can fix it itself.

And yeah, lots of files failing as well, like you pointed. Changing all the files won't be so worthy I guess.

So, in effect looks like we won't achieve anything new by integrating this, so I'll close this PR. Sorry for it. 😬

PR #6935, opened 2022-05-28
Update dockerfile to use Node v16 LTS

dmint789:
Our dockerfile doesn't build a valid image anymore. This PR updates the dockerfile to use Node 16 LTS and fixes the problem.

### DIFF
--- a/WcaOnRails/Dockerfile
+++ b/WcaOnRails/Dockerfile
@@ -6,9 +6,10 @@ WORKDIR /app
 
 RUN apt-get update && apt-get install -y curl gnupg
 
-# Install Node 14 LTS
+# Install Node 16 LTS
 # From: https://github.com/nodesource/distributions
-RUN curl -fsSL https://deb.nodesource.com/setup_14.x | bash -
+RUN curl -fsSL https://deb.nodesource.com/setup_16.x | bash -
+RUN apt-get install -y nodejs
 
 # Add PPA needed to install yarn.
 # From: https://yarnpkg.com/en/docs/install#debian-stable


dmint789:
Last time I made this fix, @gregorbg asked to implicitly use version 14 to avoid issues with some dependencies needing a specific version. I decided to stick to lts this time, as I believe it's more future-proof, seeing as this is the second time our dockerfile needed to be updated in just a year.

WcaOnRails/Dockerfile
```diff
@@ -6,9 +6,10 @@ WORKDIR /app
 
 RUN apt-get update && apt-get install -y curl gnupg
 
-# Install Node 14 LTS
+# Install Node 16 LTS
 # From: https://github.com/nodesource/distributions
-RUN curl -fsSL https://deb.nodesource.com/setup_14.x | bash -
+RUN curl -fsSL https://deb.nodesource.com/setup_lts.x | bash -
```
gregorbg, line 11: Please change to `setup_16`

gregorbg (changes requested):
The correct process would still be to pin this to "whatever is used in our current Chef infrastructure provisioning" as per https://github.com/thewca/worldcubeassociation.org/blob/master/chef/site-cookbooks/wca/recipes/default.rb#L8.

I can appreciate the logic that an LTS release "should work", but I want to keep our versions in sync as good as possible. Believe me, there have been times where implicitly causing a newer version of Node than we expected _has_ caused problems in the past.
Also, changing a Dockerfile twice in a couple of months is waaay below average for usual Docker workflows 😄

gregorbg (approved):

dmint789:
Wow, that was fast @gregorbg haha. Thanks for approving quickly.

gregorbg:
I was coincidentally browsing GitHub anyways when your new commit came in :D
Feel free to merge as soon as the checks pass.

PR #6939, opened 2022-06-01
React rewrite of Regional Organizations Overview page

notrichardpeng:
Still trying figure out the React equivalent of each `<%= %>` expression


gregorbg:
Wow, that was quick! Can you please post a screenshot of your local version, so that I don't have to check out this branch locally and boot my dev server just to admire your fast progress? 🤩 

Also, if you have any concrete questions or you feel stuck migrating some part of the existing HTML, just reach out on Slack!

WcaOnRails/app/webpacker/components/RegionalOrganizations/index.js
```diff
@@ -0,0 +1,43 @@
...
+
+function HasOrganizations({
+  orgs,
+}) {
+  if (orgs) {
+    return (
+      <>
+        <p>{I18n.t('regional_organizations.content')}</p>
+        <div className="organizations-list">
+          <DisplayOrganizations orgs={orgs} />
```
campos20, line 20: You can avoid writing a function for that component. Also, a key is needed for react to properfly handle multiple objects
`{orgs.map((org) => <p key={org.name}>{org.name}</p>)}`

campos20 (changes requested):

WcaOnRails/app/webpacker/components/RegionalOrganizations/index.js
notrichardpeng: Thanks for letting me know. Should I keep the function if I need to write multiple lines? Or can the expression you gave also be expanded into multiple lines?

notrichardpeng:
> Wow, that was quick! Can you please post a screenshot of your local version, so that I don't have to check out this branch locally and boot my dev server just to admire your fast progress? 🤩
> 
> Also, if you have any concrete questions or you feel stuck migrating some part of the existing HTML, just reach out on Slack!

image
This is the progress so far, not too fast haha

gregorbg:
Well it's certainly something! Any progress towards React is good progress, so you can be proud of that :)

I've also started a comprehensive Q&A which you might find helpful: https://docs.google.com/document/d/1X7veopxmMFKJtXT7WFpcn0w6MqW68fBybSlvBf25bYY/edit#heading=h.mo92jzd8wu47

notrichardpeng:
Now the remaining concerns are:
1. Adjusting the size of the RO logos and whether or not they can be passed in from ruby
2. Iterating through the list of requirements in the locale as the original code does, rather than listing all of them explicitly

Current progress 😃:
image
image

gregorbg:
Cool, it's taking shape more and more! Do you want explanations for how to handle the images or do you prefer trying out and looking for a solution by yourself?

notrichardpeng:
I think this should work properly... we might need to deploy this on staging first to make sure, because on local dev there are no images to display to see the page in action.

gregorbg:
Staging doesn't have images uploaded either. Please just download random cat images to your computer and use them to test.
You can go through the hassle of uploading them to Rails or you can just temporarily(!) hard-code the `/home/notrichardpeng/path/to/image` URL in React to see if it works

WcaOnRails/app/views/regional_organizations/index.html.erb
```diff
@@ -1,43 +1,15 @@
...
-                <%= ro.name %>
-              </div>
-            </div>
-          <% end %>
-        </div>
-      <% end %>
-    </div>
+    <% img_paths.append("") %>
   <% end %>
+<% end %>
```
gregorbg, line 8: There's absolutely no need to compute this within the view.
In `regional_organizations_controller`, you can paste this exact code without all of the `<%` tags and provide it to the view by prepending `@` to the `img_paths` variable.

gregorbg (changes requested):

WcaOnRails/app/views/regional_organizations/index.html.erb
notrichardpeng: :hushed: Oh I should have done that, guess it's time to refresh my Rails knowledge

notrichardpeng:
Thanks for providing the specifics on testing the images. Turns out uploading a bylaw and random PNG as a board member doesn't work (could this also be a local issue?), but hardcoding the http link to every logo does.

Uploading random PNGs via the website:
image


Hardcoding remote links as image sources:
image

notrichardpeng:
Update: PNG upload worked! Turns out I had to install ImageMagick locally beforehand

image

Seems pretty good to go, maybe except for one other thing different from the original code, which is that I listed all the  regional organization requirements out with 6 explicit lines of `I18n.t()` while the original code did it with an iteration.

WcaOnRails/app/controllers/regional_organizations_controller.rb
```diff
@@ -10,6 +10,14 @@ def admin
 
   def index
     @acknowledged_regional_organizations = RegionalOrganization.currently_acknowledged.order(country: :asc)
+    @img_paths = []
+    @acknowledged_regional_organizations.each do |ro|
```
gregorbg, line 14: Ruby has strong support for functional operations. You can do
```ruby
@img_paths = @acknowledged_regional_organizations.map do |ro|
  # ...stuff...
end
```

gregorbg: That being said, in terms of a 100% clean implementation I think it would be reasonable to consider adjusting the `serializable_hash` for the `RegionalOrganization` model to include its own `thumbnail_url` (or an empty string) as part of the serialized data.

Passing around a second array that has the exact same indices as the original data collection seems a bit overkill...

gregorbg (approved):
Cool! I'm very happy with the progress on this, and in terms of feature parity I believe it's good to go!

There's one remark however about code structure and model design. I think it's a great opportunity to get some more Ruby/Rails practice but it's arguably a little out of scope of this PR so I leave it to you to decide whether you want to take it the extra mile :)

WcaOnRails/app/controllers/regional_organizations_controller.rb
notrichardpeng: I tried rewriting `serializable_hash`, but `polymorphic_url` is not an available function in models. Perhaps I can find a way to make it available or not even need it?

gregorbg:
Hrm... Controllers inherit from a base class which in turn internally includes modules (think of it as packages) that provide `polymorphic_url`. You're right that in a model you'd probably have to import that specifc module yourself.

After some more digging, `ro.logo.variant(....).processed.url` might do the trick without even needing to invoke `polymorphic_url` at all. See if that works, otherwise just leave as-is.

WcaOnRails/app/controllers/regional_organizations_controller.rb
gregorbg: Hrm... Controllers inherit from a base class which in turn internally includes modules (think of it as packages) that provide `polymorphic_url`. You're right that in a model you'd probably have to import that specifc module yourself.

After some more digging, `ro.logo.variant(....).processed.url` seems like it could do the trick without even needing to invoke polymorphic_url at all. See if that works, otherwise just leave as-is.

notrichardpeng:
@campos20 Does my React code structure seem fine? I'm a little concerened about having a function `HasOrganizations` just for the sake of the `if` statement.

WcaOnRails/app/controllers/regional_organizations_controller.rb
gregorbg: Kudos for finding the proper `url_helper` methods! However, I have a bad feeling about using this in production.

Essentially, right now you found a way to make it work under the exact assumption that files are uploaded to the local storage (which is what happens in local testing). In real prod use, these assumptions will break because we actually store RO stuff on an Amazon S3 bucket. Especially the `only_path: true` part will mess things up there.

The reason why most URL methods do not work in models is because they rely on the _convention_ (i.e. unspoken silent assumption) that they're being called in controllers. Certain module imports and config values are only set in controllers, not in model classes. However, I feel like there has to be some way of "properly" including an asset URL in a model's serial hash.

Summoning @jonatanklosko @viroulep for help :pray:

gregorbg: There's https://guides.rubyonrails.org/active_storage_overview.html#serving-files which recommends using `url_for`. Yet again, that's a helper only available in views and controllers, not in models :/

jonatanklosko: From what I understand `rails_representstion_url` generates an internal URL, but that endpoint redirects the resource, which could be S3 or anywhere else. Based on some search the current approach looks reasonable, see [this discussion](https://github.com/rails/rails/issues/32500#issuecomment-380004250). I think we can use `rails_representation_path` instead of `only_path`.

As for using route helpers, we rarely want to do it in models, but calling `Rails.application.routes.url_helpers` is alright.

notrichardpeng: Thank you for the input! `rails_representation_url` is working, but a host url needs to be set. For dev environment, this is simply done with `Rails.application.routes.default_url_options[:host] = 'localhost:3000'`. I wonder what I should put in place of `localhost:3000` for production, `s3.us-west-2.amazonaws.com/www.worldcubeassociation.org`?

jonatanklosko: @notrichardpeng what is the generated path? It seems to me that this URL should always be internal (so in production it would be `https://worldcubeassociation.org/*` and that redirects to S3). If so, using `rails_representation_path` should do the job. I may totally be wrong on this though :)

notrichardpeng: @jonatanklosko This is the link of the image if it's what you mean by generated path: `http://localhost:3000/rails/active_storage/disk/.../random.png`. I went to our production site and copied the image address and it is `https://www.worldcubeassociation.org/rails/active_storage/representations/redirect/.../XXX.png`. Not sure if the small difference is negligible

jonatanklosko: And if you go to `https://www.worldcubeassociation.org/rails/active_storage/representations/redirect/.../XXX.png` it redirects to the image on S3?

notrichardpeng: yes! `https://s3.us-west-2.amazonaws.com/www.worldcubeassociation.org/variants/...`

jonatanklosko: Cool, so we can just use `rails_representation_path`, this way it's going to be `/rails/active_storage/representations/redirect/.../XXX.png` and should work both on localhost and in prod without additional configuration.

WcaOnRails/app/models/regional_organization.rb
```diff
@@ -30,4 +30,13 @@ def start_date_must_be_earlier_than_end_date
   def is_pending?
     start_date.nil?
   end
+
+  def serializable_hash(options = nil)
+    {
+      name: name,
+      website: website,
+      country: country,
+      logo: logo.attached? ? Rails.application.routes.url_helpers.rails_representation_path(logo.variant(resize: "500x300").processed, host: EnvVars.ROOT_URL) : '',
```
jonatanklosko, line 39: We want a relative URL, so there's no need to pass the host, no?

notrichardpeng: oh, the fucntion requires host to be set, otherwise it wants us to use `only_path`

notrichardpeng: but I think this is fine, because in production `ROOT_URL` is `worldcubeassociation.org` right

jonatanklosko: Interesting, `_path` is essentially `_url` with only path.

But either way, I've just realised it's an API response, so it's better to return an absolute path. So I think the combination we want is `rails_representation_url` and with explicit `host:`. We already do the same here:

https://github.com/thewca/worldcubeassociation.org/blob/fe6cf26b90f2d78dda44b318b765e2096767555c/WcaOnRails/app/models/person.rb#L273

notrichardpeng: :O aha

jonatanklosko: If we find ourselves using this more, we can do [this](https://andycroll.com/ruby/url-helpers-outside-views-controllers/), which looks pretty elegant. Alternatively we could set global `default_url_options` as described [here](https://github.com/jbranchaud/til/blob/master/rails/set-default-url-options-for-entire-application.md), we currently do it for mailers specifically.

notrichardpeng: I actually tried the latter option on my [previous commit](https://github.com/thewca/worldcubeassociation.org/pull/6939/commits/4fd7def5ad7f68c97493aba3e3bf4163a4a17e46), but I got rid of it, afraid that it would affect code elsewhere. Should I bring it back?

jonatanklosko: Sounds good to me, but it should be more complete, as in:

https://github.com/thewca/worldcubeassociation.org/blob/fe6cf26b90f2d78dda44b318b765e2096767555c/WcaOnRails/config/environments/production.rb#L83-L88

(for each environment file respectively)

With that we most likely no longer need to set `config.action_mailer.default_url_options`, best to check an email preview to verify.

notrichardpeng:
(Broken into many commits because RuboCop was somehow giving "Unexpected Error")

WcaOnRails/app/models/regional_organization.rb
notrichardpeng: I implemented the changes, it works, and the mailer previews are showing up fine, except for a few mailers that are already throwing exceptions beforehand `undefined_method 'preferred_locale' for "dummy_link"`. :)

WcaOnRails/config/environments/development.rb
```diff
@@ -2,6 +2,17 @@
 
 require "active_support/core_ext/integer/time"
 
+root_url = EnvVars.ROOT_URL
+unless root_url.present?
+  root_url = "http://localhost:3000"
+end
+root_url = URI.parse(root_url)
```
jonatanklosko, line 9:
```suggestion
root_url = URI.parse(EnvVars.ROOT_URL)
```

I believe we can simplify now as `ROOT_URL` has a default :)

https://github.com/thewca/worldcubeassociation.org/blob/fe6cf26b90f2d78dda44b318b765e2096767555c/WcaOnRails/env_vars.rb#L22-L37

WcaOnRails/config/environments/production.rb
```diff
@@ -2,6 +2,13 @@
 
 require "active_support/core_ext/integer/time"
 
+root_url = URI.parse(EnvVars.ROOT_URL)
+Rails.application.routes.default_url_options = {
+  protocol: root_url.scheme,
+  host: root_url.host,
+  port: root_url.port,
+}
+
```
jonatanklosko, line 11: We should mirror it for `environments/test.rb` as well :D

WcaOnRails/app/models/regional_organization.rb
jonatanklosko: Just to clarify, with the previews I meant specifically that the URLs get the expected hostname.

> Are the config code best placed at where they are right now?

I think so!

WcaOnRails/config/environments/production.rb
notrichardpeng: yep!

WcaOnRails/config/environments/development.rb
```diff
@@ -2,6 +2,13 @@
 
 require "active_support/core_ext/integer/time"
 
+root_url = URI.parse(EnvVars.ROOT_URL)
```
jonatanklosko, line 5: Just for the reference

```suggestion
# See https://github.com/rails/rails/issues/29992#issuecomment-761892658
root_url = URI.parse(EnvVars.ROOT_URL)
```

WcaOnRails/app/webpacker/components/RegionalOrganizations/index.js
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  return <p>{I18n.t('regional_organizations.empty')}</p>;
+}
+
+function ROOverview({
+  orgs,
+}) {
+  const title = I18n.t('regional_organizations.title');
+  document.title = `${title} | World Cube Association`;
```
jonatanklosko, line 56: It's best to wrap this kind of direct page manipulation in `useEffect`, so it doesn't run on every re-render (i.e. whenever the component function is called). In this case this would be

```js
useEffect(() => {
  const title = I18n.t('regional_organizations.title');
  document.title = `${title} | World Cube Association`;
}, []);
```

That said, I'd keep the title management on the Rails side using `provide(:title, t('regional_organizations.title'))` for simplicity. If we migrate off of that completely, we will handle it uniformly on the client.

WcaOnRails/app/models/regional_organization.rb
```diff
@@ -30,4 +30,13 @@ def start_date_must_be_earlier_than_end_date
   def is_pending?
     start_date.nil?
   end
+
+  def serializable_hash(options = nil)
+    {
+      name: name,
+      website: website,
+      country: country,
+      logo: logo.attached? ? Rails.application.routes.url_helpers.rails_representation_url(logo.variant(resize: "500x300").processed) : '',
```
jonatanklosko, line 39: Semantically it's better to use `nil` for absent fields :)

```suggestion
      logo: logo.attached? ? Rails.application.routes.url_helpers.rails_representation_url(logo.variant(resize: "500x300").processed) : nil,
```

jonatanklosko (approved):
Last two tiny comments from me!

@gregorbg let us know if the URL stuff makes sense now :)

WcaOnRails/app/webpacker/components/RegionalOrganizations/index.js
notrichardpeng: Got it. Thanks so much for all the help so far!

gregorbg:
Yeah, the entire URL mess makes a lot more sense now, thanks for your thorough explanations and review @jonatanklosko 🤗

jonatanklosko:
To share the configuration I think we could create `config/environments/base.rb` and then require it in each specific configuration with `require_relative "base"`.

jonatanklosko:
Cool, `config/application.rb` sounds good then :)

WcaOnRails/app/models/regional_organization.rb
notrichardpeng: > Just to clarify, with the previews I meant specifically that the URLs get the expected hostname.

does that mean having the mailer preview at `localhost:3000/rails/mailers/...` on dev?

jonatanklosko: The preview is always going to be there. What I'd check is trigger any email that has a link (reset password or competaition registration?) and check if the URL has the expected hostname.

notrichardpeng: 👌 I checked "new registration" email that has a link to "login form", which correctly leads to `localhost:3000/users/sign_in`. Looking good

notrichardpeng:
If there are no further concerns, I will merge this?

gregorbg:
I am a bit hesitant because I want to test out the config changes on staging, but we need staging for a PR by Philippe right now.
You can consider this PR done and accepted, but please hold off the merge for now.

I'll get back to you about follow-up tasks ASAP.

notrichardpeng:
Sounds good 👌

PR #6967, opened 2022-06-02
Update translation index

mirandapedro:
It would be awesome if we could get a redeploy after this is merged as well!


gregorbg:
Merged and deployed! https://www.worldcubeassociation.org/regulations/translations/

PR #6261, opened 2021-05-29
Expose all competitions attended for person in API

saranshgrover:
Closes #4864

With this PR an API req like `http://worldcubeassociation.org/api/v0/persons/2014GROV01?competitions=true` will expose all competitions for the person.

I was also trying to expose it in the `/users` controller, but could not figure out how to. [Here's](https://github.com/thewca/worldcubeassociation.org/blob/master/WcaOnRails/app/controllers/api/v0/api_controller.rb#L119) where I was trying to add it - simply showing accepted registrations was not matching up with competition counts (it was undercounting) but I couldn't figure out why. Any help with this is appreciated, but if not this PR should be good since at least its exposed somewhere.

PR #6980, opened 2022-06-08
Expose a user's ongoing competitions

### LINKED ISSUE #6970: View user's ongoing competitions from WCA API
https://github.com/thewca/worldcubeassociation.org/blob/6e253ed58528223c789897d24909b1fbafba69fd/WcaOnRails/app/controllers/api/v0/api_controller.rb#L124

Right now we can query upcoming competitions but not their ongoing ones.

This would be super useful for applications ([such as mine](https://competition-groups.surge.sh/)) which show a user's groups for an upcoming competitions and it'd be nice to show the competition happening the same day.

I can likely send in a fix. This seems like adding something to the competition model and adding a query param?

coder13:
This calculates if a competition is ongoing by looking at if it has 1. started and 2. probably not over.
It then exposes this as a query param when fetching `/users/{userid}`

This directly fixes #6970 but I'd like to also, in a later PR, expose a route `/api/v0/users/{userid}/competitions` to fetch all of a user's competitions.

Alternatively, I could modify the `/api/v0/competitions` to add a query param to search for competitions with registrations containing a certain userId.

### DIFF
--- a/WcaOnRails/app/controllers/api/v0/api_controller.rb
+++ b/WcaOnRails/app/controllers/api/v0/api_controller.rb
@@ -122,18 +122,10 @@ class Api::V0::ApiController < ApplicationController
     if user
       json = { user: user }
       if params[:upcoming_competitions]
-        json[:upcoming_competitions] = user.registrations
-                                           .accepted
-                                           .includes(competition: [:delegates, :organizers, :events])
-                                           .map(&:competition)
-                                           .select(&:upcoming?)
+        json[:upcoming_competitions] = user.accepted_competitions.select(&:upcoming?)
       end
       if params[:ongoing_competitions]
-        json[:ongoing_competitions] = user.registrations
-                                          .accepted
-                                          .includes(competition: [:delegates, :organizers, :events])
-                                          .map(&:competition)
-                                          .select(&:ongoing?)
+        json[:ongoing_competitions] = user.accepted_competitions.select(&:in_progress?)
       end
       render status: :ok, json: json
     else
--- a/WcaOnRails/app/controllers/api/v0/api_controller.rb
+++ b/WcaOnRails/app/controllers/api/v0/api_controller.rb
@@ -128,6 +128,13 @@ class Api::V0::ApiController < ApplicationController
                                            .map(&:competition)
                                            .select(&:upcoming?)
       end
+      if params[:ongoing_competitions]
+        json[:ongoing_competitions] = user.registrations
+                                          .accepted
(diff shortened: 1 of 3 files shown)


WcaOnRails/app/controllers/api/v0/api_controller.rb
```diff
@@ -128,6 +128,13 @@ def show_user(user)
                                            .map(&:competition)
                                            .select(&:upcoming?)
       end
+      if params[:ongoing_competitions]
+        json[:ongoing_competitions] = user.registrations
+                                          .accepted
+                                          .includes(competition: [:delegates, :organizers, :events])
+                                          .map(&:competition)
+                                          .select(&:ongoing?)
```
gregorbg, line 136: The entire notion of querying a user's competitions is identical to `upcoming_competitions` above.

```ruby
user.registrations
.accepted
.includes(competition: [:delegates, :organizers, :events])
.map(&:competition)
```

Please _at least_ move that part out of the two if-blocks and only apply the `select` filter based on which query param is present.
Ideally, consider moving that entire block into the `user.rb` model as a kind of `accepted_competitions` method. (although that's arguably a bit out of scope of this PR, so I leave it to you to decide)

gregorbg (changes requested):
Thanks for the quick implementation!
One nitpick about code reuse.

Also, please do not hesitate to pick up #6261 if you're interested, Saransh doesn't seem to be available to finish it right now.

And for the love of god please don't introduce a user ID filtering query param to the `/competitions` endpoint 😂

coder13:
> And for the love of god please don't introduce a user ID filtering query param to the /competitions endpoint joy

haha, ok. Is this a performance thing?

Is it because you query for all competitions -> for all registrations where registrations contains user Id?

Just spitballing but we could even do a `/api/v0/competitions/mine` to mirror how it is in terms of the pages.

gregorbg:
> > And for the love of god please don't introduce a user ID filtering query param to the /competitions endpoint joy
> 
> haha, ok. Is this a performance thing?
> 
> Is it because you query for all competitions -> for all registrations where registrations contains user Id?
> 
> Just spitballing but we could even do a `/api/v0/competitions/mine` to mirror how it is in terms of the pages.

It is mostly for avoiding feature duplication. I generally don't want to have the exact same information be available in two different ways, as that tends to confuse people who use the API externally.
We also currently don't have a really fleshed-out design spec, so I'd rather not take each and every idea about our API and implement them straight away.

gregorbg (approved):
Thanks for making the adjustments!

coder13:
Thanks for merging!

PR #6981, opened 2022-06-08
Expose person's competitions in API

coder13:
Fixes #6261

### DIFF
--- a/WcaOnRails/app/controllers/api/v0/persons_controller.rb
+++ b/WcaOnRails/app/controllers/api/v0/persons_controller.rb
@@ -1,4 +1,4 @@
-  # frozen_string_literal: true
+# frozen_string_literal: true
 
 class Api::V0::PersonsController < Api::V0::ApiController
   def index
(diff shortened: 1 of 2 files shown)


WcaOnRails/app/controllers/api/v0/persons_controller.rb
```diff
@@ -1,4 +1,4 @@
-# frozen_string_literal: true
+  # frozen_string_literal: true
```
gregorbg, line 1: Why are you adding a space here? Seems like some local editor config that is doing shenanigans on auto-format. Also makes me wonder why our Ruby code linter doesn't catch this, it's usually very very whiny about stuff like this.

gregorbg (commented):
Potential choke in your local code editor config, auto-adding spaces where it shouldn't.
Otherwise, this is good to go. Thanks for the swift and concise implementation!

gregorbg:
Tests fail because of some caching errors with our dependency bundles. Will fix this ASAP, you will then have to rebase on top of master.

gregorbg:
Update: The reason for the test fails has been fixed, please include https://github.com/thewca/worldcubeassociation.org/commit/40d4b02fca6c35d995e34d151ac97cd760849ec6 to make sure that the tests don't fail

gregorbg (approved):

PR #6989, opened 2022-06-18
Redirect to new sor page

campos20:


campos20:
Merging after we get a green since this is trivial

Issue #6621, opened 2022-01-17
Competitors now able to delete their registration when accepted


EdHollingdale:
When you enable competitors being able to edit their events it seems they are also able to delete their registrations after they have been accepted. This is undesirable for now, since there are further actions that need to be taken for an accepted registration (e.g. processing a refund if it is due, and accepting the next competitor on the waiting list). While competitors being able to edit their events is great, it would be good if the ability to change their status was not an option for them. Obviously this can be re-enabled when there is automated waiting list handling!

AlexKatyukov:
IMO, there should be two checkboxes: one for editing events and second for deleting the registrations. Sometimes deleting is okay, but sometimes are headache.

EdHollingdale:
I know this issue has a high priority label, but it is getting really frustrating. When someone deleted registration there are extra steps (refunds, processing waiting list), and people starting the process themselves rather and having a free spot in the competition until someone notices is very frustrating.

gregorbg:
Fixed in #6991

PR #6991, opened 2022-06-19
Restrict competitors from self-deleting registration after acceptance

notrichardpeng:
#6621 Adding a separate field for organizers to decide whether competitors can delete their own registration after having their registration accepted.


WcaOnRails/config/locales/en.yml
```diff
@@ -305,6 +305,7 @@ en:
         on_the_spot_registration: "On the spot registrations"
         on_the_spot_entry_fee_lowest_denomination: "On the spot base registration fee"
         allow_registration_edits: "Registration edits"
+        allow_registration_self_delete_after_acceptance: "Registration self-deletion"
```
Jambrose777, line 308: "Registration Cancellations"

Jambrose777 (approved):
Overall looks very good and clean! I added one comment for something that could be a bit better to read, but it is no big deal and even the original works for its intentions IMO.

WcaOnRails/config/locales/en.yml
notrichardpeng: Yep that is indeed a better wording! I will probably make the 'C' lower case since "Registration edits" has the second word lowercase too

WcaOnRails/app/models/user.rb
```diff
@@ -731,6 +731,12 @@ def can_edit_registration?(registration)
     can_manage_competition?(registration.competition) || (registration.user_id == self.id && editable_by_user)
   end
 
+  def can_delete_registration?(registration)
+    # A registration can only be deleted by a user after it has been accepted if the organizers allow
+    can_delete_by_user = (!registration.accepted? || registration.competition.registration_delete_after_acceptance_allowed?)
+    registration.user_id == self.id && can_delete_by_user
```
gregorbg, line 737: ⚠️ It is imperative that we always allow organisers/admins to delete any registration, regardless of how the competition is configured!

```suggestion
    can_manage_competition?(registration.competition) || (registration.user_id == self.id && can_delete_by_user)
```

gregorbg (changes requested):
Thank you very much for the quick and concise implementation! 😄 

You missed one very important point about organisers always being able to delete registrations, see my inline code comment. Otherwise good to go!

WcaOnRails/app/models/user.rb
notrichardpeng: Your outline helped a lot with the speed!

For this though, admins are still able to delete registration from the admin registration list, and the new variable does not interfere with that. It only interferes with users deleting registration from the register form.

gregorbg: Okay, makes sense in terms of functionality. I'd still argue for having a consistent codebase where the concrete implementation of methods does not change depending on which frontend views these methods are accessed from.

Unifying it would also mean you could (probably?) redirect the admin check to this new deletion function as well.

notrichardpeng: I see, that makes sense.

gregorbg (approved):

PR #7001, opened 2022-06-27
Fix CAD: Make script more resilient against temporary person data

gregorbg:
Following a recent error message during regular WRT workflow CAD runs.


gregorbg:
@thewca/wrt I have opted for a quite straightforward "just skip over the person if it's not found" approach. It seemed reasonable to me because these kinds of temporary persons won't have any meaningful rankings to include in the tables in the first place.

One possible alternative would be to adjust `compute_concise_results` to make sure these temporary persons don't end up in the `Concise*Result` tables in the first place. Please advise how to proceed :)

Issue #2051, opened 2017-10-18
Integrate qualification times into the registration process


jgouly:
You can currently register in events for a competition that has pre-qualification limits.
There could be some information displayed during registration to show that there is qualification per event. So you know to either not sign up, or you need to beat the qualification in an earlier competition.

Mollerz:
I don't think it should fully prevent you from registering for the event, you might have a competition between registering and the time of the competition that allows you to qualify for it. Perhaps give organisers an option to press a button that removes all those who haven't got results with the designated qualifying times? Maybe with a check...

jgouly:
Yup, that's why I was suggesting to show the qualification information!

cubewhiz:
It's not a "must have," but something like this would be very helpful for big competitions like Nats or Worlds, so that we don't have to manually edit hundreds of registrations.

cubewhiz:
This year we were able to edit the registrations for Nationals through SQL query, but many competitors (and their parents) were quite angry because they were removed from the events (and they didn't quite understand the requirements). I think it is much better to prevent them from registering for these events in the first place than it is to remove them from the events later. We would be grateful if this could be implemented by the end of 2018 so it is available for Nats 2019.

Mollerz:
I think preventing people from registering for not making the qualification is a bad idea. In many cases these qualifications are for major championships which are announced quite far in advanced. It is very likely that someone will compete in another local competition in the mean time and possibly get some of the qualifications. In which case they will need to email the organisers and ask to be added to events.

I think if you select events with qualification in the registration procedure it should warn you if there are qualification times that you have not reached. Perhaps under the registration requirements it puts a red box that shows all qualification times, and if you are logged in it only shows the ones you have not made yet.

ToastyKen:
@Mollerz Fyi, we've actually used the system we're recommending, and it worked better. We're not just posing a hypothetical.

For CubingUSA Nationals 2017, we rolled our own registration system, and it worked like this:

1) You could not register for events you didn't qualify in.

2) You needed to qualify in at least one event before registering at all.

3) If you were already registered, and you later qualified for an event, you could just log back in and add it (and pay the additional event fee). No email to organizers needed.

The only issue was that it took a few days for us to pick up WCA results, so people had to wait a few days to add events they just qualified for. This led to a little confusion but no significant problems, as they were able to register in a few days.

[...]

PR #7016, opened 2022-06-29
Bug fixes to Qualification.js.

timreyn:
-Set a default 'type' after resultType is newly set, so that fields below appear.
-Show the qualification deadline field for any 'type', not just ones with time inputs.
-Delete a stray Console.log.

Part of #2051.


gregorbg:
Pretty straight-forward, but nonetheless thanks for addressing these so promptly!

PR #7019, opened 2022-06-30
Add option to require qualification for registered events.

timreyn:
Adds an option to the competition form to require that competitors be qualified for events that they are registering for.

This does not prevent registering for events with a "Top N" qualification, because it's not yet known what the qualifying standard will be. In theory, we could do "Top N of people currently registered", but that would lead to weird effects if someone qualified cancelled their registration.

Part of #2051.


timreyn:
Here's what this looks like when you are not qualified for an event:

image

WcaOnRails/app/views/registrations/_register_form.html.erb
```diff
@@ -1,9 +1,10 @@
 <% disabled = !current_user.can_edit_registration?(@registration) %>
 <% url = @registration.new_record? ? competition_registrations_path(@competition)
                                    : registration_path(@registration) %>
+<% ineligible_events = @competition.competition_events.select{|e| !e.can_register?(current_user) }.map{|e| e.event } %>
```
gregorbg, line 4: Nit: `.map(&:event)`

gregorbg, line 4: More nit: The lambda parameter in `select` should be called `ce` to avoid confusing `competition_event` and `event`.

WcaOnRails/app/views/registrations/edit.html.erb
```diff
@@ -13,10 +13,13 @@
     <% end %>
     <h4><%= @registration.name %></h4>
 
+    <% ineligible_events = @competition.competition_events.select{|e| !e.can_register?(@registration.user) }.map{|e| e.event } %>
```
gregorbg, line 16: Same two nits as above. Why is this code duplicated anyways?

WcaOnRails/app/views/shared/_associated_events_picker.html.erb
```diff
@@ -32,16 +34,21 @@ Note: form_builder.object is the object having associated events.
...
         <% end %>
         <%= label_tag "#{events_association_name}_#{event.id}" do %>
-          <%= f.check_box "_destroy", { checked: selected_events.include?(event), id: "#{events_association_name}_#{event.id}", disabled: disabled }, "0", "1" %>
-          <%= cubing_icon(event.id, data: { toggle: "tooltip", placement: "top" }, title: event.name) %>
+          <%= f.check_box "_destroy", { checked: selected_events.include?(event) && !e_disabled, id: "#{events_association_name}_#{event.id}", disabled: disabled || e_disabled }, "0", "1" %>
+          <% tooltip = e_disabled ? disabled_tooltip : event.name %>
+          <%= cubing_icon(event.id, data: { toggle: "tooltip", placement: "top" }, title: tooltip) %>
+        <% end %>
+        <% if e_disabled %>
+          <%= f.hidden_field "_destroy", { value: "1" } %>
```
gregorbg, line 51: I had to pause and ponder for several seconds why you're introducing a `hidden_field` for `_destroy` in addition to the checkbox above. Then I realised that the if condition here and the disabled condition are mutually exclusive. On that note, wouldn't it make more sense to do something along the lines of
```ruby
<% if e_disabled %>
  <%= f.hidden_field "_destroy", { value: "1" } %>
<% else %>
  <%= f.check_box "_destroy", ....... %>
<% end %>
```

WcaOnRails/config/locales/en.yml
```diff
@@ -343,6 +343,7 @@ en:
         event_restrictions: "I would like to restrict competitors from registering for specific combinations of events"
         event_restrictions_reason: "Reason and details for event restrictions"
         main_event_id: "Main event"
+        allow_registration_without_qualification: "Events with Qualification"
```
gregorbg, line 346: The text here doesn't seem to fit the description. Where is this used?
(I'd expect something more along the lines of "Registering without automated checking of qualification" or "Registering without being qualified is allowed")

WcaOnRails/db/migrate/20220627195217_add_allow_registration_without_qualification_to_competitions.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddAllowRegistrationWithoutQualificationToCompetitions < ActiveRecord::Migration[5.2]
```
gregorbg, line 3: Your `bin/rails` runner seems to still generate migrations on version 5.2.
When did you run `bundle install` for the last time? 😅

WcaOnRails/db/migrate/20220627195217_add_allow_registration_without_qualification_to_competitions.rb
```diff
@@ -0,0 +1,7 @@
...
+  def change
+    add_column :Competitions, :allow_registration_without_qualification, :boolean, default: true
```
gregorbg, line 5: What's the rationale behind using `true` as default?

WcaOnRails/lib/qualification.rb
```diff
@@ -35,11 +35,36 @@ def self.load(json)
...
       end
-      p out
       out
     end
   end
 
+  def can_register?(user, event_id)
+    if user.person.nil?
+      return false
+    end
```
gregorbg, line 45: Ruby syntactic sugar: `return false if user.person.nil?`

gregorbg: (Yes, as a one-liner.)

WcaOnRails/lib/qualification.rb
```diff
@@ -35,11 +35,36 @@ def self.load(json)
...
+        qualifying_results = qualifying_results.average_succeeded
+      end
+    when "attemptResult"
+      case self.result_type
+      when "single"
+        qualifying_results = qualifying_results.single_better_than(self.level)
+      when "average"
+        qualifying_results = qualifying_results.average_better_than(self.level)
+      end
+    end
```
gregorbg, line 64: Not a fan of self-overriding variables. Call the first occurence `before_deadline_results` or something?

WcaOnRails/lib/qualification.rb
```diff
@@ -35,11 +35,36 @@ def self.load(json)
       rescue ArgumentError
         nil
       end
-      p out
       out
     end
   end
 
+  def can_register?(user, event_id)
```
gregorbg, line 42: Debating with myself whether this should be split up into `qualifying_results_for(user, event_id)` here and the actual `can_register?` over in `competition_event`. I have no strong arguments other than gut feeling. What is your opinion?

WcaOnRails/app/views/shared/_associated_events_picker.html.erb
```diff
@@ -32,16 +34,21 @@ Note: form_builder.object is the object having associated events.
...
           <%= f.hidden_field :competition_event_id %>
         <% else %>
           <%= f.hidden_field :event_id %>
         <% end %>
         <%= label_tag "#{events_association_name}_#{event.id}" do %>
-          <%= f.check_box "_destroy", { checked: selected_events.include?(event), id: "#{events_association_name}_#{event.id}", disabled: disabled }, "0", "1" %>
-          <%= cubing_icon(event.id, data: { toggle: "tooltip", placement: "top" }, title: event.name) %>
+          <%= f.check_box "_destroy", { checked: selected_events.include?(event) && !e_disabled, id: "#{events_association_name}_#{event.id}", disabled: disabled || e_disabled }, "0", "1" %>
+          <% tooltip = e_disabled ? disabled_tooltip : event.name %>
+          <%= cubing_icon(event.id, data: { toggle: "tooltip", placement: "top" }, title: tooltip) %>
```
gregorbg, line 48: Can we have the icon appear in some gentle tone of red (along the lines of `#FF7F7F`) when you're not qualified? Just to distinguish between "It's not selected because you haven't clicked it yet" and "It's not selected because you _cannot_ click it"

gregorbg (changes requested):
Thanks, very succinct implementation in general! Cosmetic remarks here and there, though.

WcaOnRails/lib/qualification.rb
timreyn: The tricky thing here is "ranking" qualifications. If we did that, the code in `qualification.rb` would either need to ignore "ranking", or be more complicated to consider registered competitors (and then have `CompetitionEvent.can_register?` ignore it). I went with this naming scheme to accurately represent what the code actually checks: whether, based on the qualification standard, the competitor is allowed to register for the event.

For what it's worth, the code to remove unqualified competitors won't use this; it'll do one big SQL query to generate the psych sheet as of the deadline, to avoid doing N SQL queries for N competitors.

WcaOnRails/lib/qualification.rb
timreyn: Thank you!

WcaOnRails/lib/qualification.rb
timreyn: Done.

WcaOnRails/app/models/result.rb
```diff
@@ -38,8 +39,15 @@ def unique_result_per_round
 
   scope :final, -> { where(roundTypeId: RoundType.final_rounds.map(&:id)) }
   scope :succeeded, -> { where("best > 0") }
+  scope :average_succeeded, -> { where("average > 0") }
   scope :podium, -> { final.succeeded.where(pos: [1..3]) }
   scope :winners, -> { final.succeeded.where(pos: 1).joins(:event).order("Events.rank") }
+  scope :no_later_than, lambda { |date|
+    joins(:competition).where("start_date <= ?", date)
```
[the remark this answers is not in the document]
timreyn: I'm not sure that applies, because that's about solves in a multi-day round, rather than a multi-day competition.

The really correct way to do this would be based on the competition schedule. That sounds like a more complicated join, though, and I'm not that enthusiastic about trying to write that join.

Maybe this is a question for WCAT? I'll start a thread with them. WCRP does not specify:

> 5.1 The competitor limit can be on either a “first come first serve” basis or based upon
qualification results or rankings in the WCA world rankings of a previously announced
calendar date.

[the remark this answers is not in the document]
timreyn: I changed this to `:before` and used `end_date` rather than `start_date`, and strictly-less-than.

gregorbg:
Thanks for the changes, code is looking good in its current state but there are a few interesting test fails.
Also, what about the Ranking qualification SQL check that you mentioned in one of the comments? Do you plan on adding that as part of this PR?

timreyn:
Looking into the test failures (but I'm a little short on time this weekend due to Euros! Probably the next round of changes will come early next week).

Re the SQL query -- this is coming in my next PR, which adds a button to remove non-qualified competitors from events (see https://github.com/thewca/worldcubeassociation.org/compare/master...timreyn:worldcubeassociation.org:qualification_clear?expand=1 if you're interested, but it's not ready for an actual review yet).

WcaOnRails/app/models/registration.rb
```diff
@@ -231,6 +231,16 @@ def self.wcif_json_schema
     end
   end
 
+  validate :cannot_register_for_unqualified_events
+  private def cannot_register_for_unqualified_events
+    if competition && competition.allow_registration_without_qualification
+      return
+    end
+    if registration_competition_events.reject(&:marked_for_destruction?).select { |event| event.competition_event.nil? || !event.competition_event.can_register?(user) }.any?
```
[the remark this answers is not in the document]
gregorbg: In contrast to #7050 (which has a very strong motivation because the WCIF PATCH endpoint is even exposed to the OAuth public!!), deleting events for which competitors have already registered can sometimes happen. So your code should be able to deal with it anyways.

But interesting how `registration_competition_events` does not seem to be updated when overall events are deleted... Maybe this will actually be implicitly fixed by #7050, who knows :thinking: 

In terms of this change, what about `!event.competition_event&.can:register?(user)`? The `&` is a nil-safe operator in Ruby, and if the event doesn't exist it breaks out of the call chain and returns nil which then gets negated to `true` so it ultimately gets rejected.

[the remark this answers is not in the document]
timreyn: Ah thank you!

timreyn:
Friendly ping @gregorbg -- I think this should be good to go, pending another rebase!

WcaOnRails/app/views/shared/_associated_events_picker.html.erb
gregorbg: Fine by me!

gregorbg:
Thanks, sorry this has slipped my attention. Going through the remaining open conversations now :)

WcaOnRails/app/views/registrations/edit.html.erb
gregorbg: That does make sense, but it shouldn't keep you from refactoring the common parts of the code into some helper method that can be invoked on both separate pages :)

WcaOnRails/db/migrate/20220627195217_add_allow_registration_without_qualification_to_competitions.rb
gregorbg: I think you should be able to do this as part of the current migration. Something along the lines of
```ruby
def change
  # N.B. This has the default set to false
  add_column ...foo... default: false
  
  # Except all of the competitions that are currently open
  Competition.where("registration_close >= :close", close: DateTime.now).update_all(allow_registration_without_qualification: true)
end
```

timreyn: Done!

WcaOnRails/app/views/registrations/edit.html.erb
timreyn: Ah, got it :) Added a method to Competition.

WcaOnRails/app/views/competitions/_competition_form.html.erb
```diff
@@ -306,6 +306,11 @@
         <%= f.input :extra_registration_requirements, input_html: { class: disable_form ? "" : "markdown-editor" } %>
 
         <hr>
+        <% if @competition.uses_qualification? %>
+          <%= f.input :qualification_results_reason %>
```
gregorbg, line 310: Sorry I realised this only now but you're adding a field about `qualification_results_reason` which seems unrelated (and is also included further down in the same form!)

Perhaps a mistake from merging, please fix.

WcaOnRails/db/structure.sql
```diff
@@ -746,7 +747,7 @@ CREATE TABLE `competition_events` (
   `competition_id` varchar(191) CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci NOT NULL,
   `event_id` varchar(191) CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci NOT NULL,
   `fee_lowest_denomination` int(11) NOT NULL DEFAULT '0',
-  `qualification` text CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci,
```
gregorbg, line 749: Unrelated change

WcaOnRails/app/views/competitions/_competition_form.html.erb
```diff
@@ -306,6 +306,11 @@
         <%= f.input :extra_registration_requirements, input_html: { class: disable_form ? "" : "markdown-editor" } %>
 
         <hr>
+        <% if @competition.uses_qualification? %>
```
gregorbg, line 309: This is tricky because it uses a method on `competition` while filling out a form that modifies that same competition. I'd rather have the Yes/No dropdown be controlled by jQuery, much like `qualification_results_reason`.

(Perhaps you can even use the same jQuery `.on('change'` body for both? Cf. `assets/javascripts/competitions.js:30`

gregorbg (changes requested):
Noticed some quirks while trying the feature out locally before merging 🙃

gregorbg:
Kindly ping @timreyn. What's the status on this issue?

timreyn:
Sorry, got busy with local competition stuff. I have a long weekend this weekend and should be able to resolve the remaining issues!

gregorbg:
Cool! No rush, just wanted to check because the WST planning meeting for September is coming up.
Thanks for the quick update! :blush:

WcaOnRails/db/structure.sql
timreyn: Argh, every time I edit the db schema unrelated changes creep in...

Fixed.

WcaOnRails/app/views/competitions/_competition_form.html.erb
timreyn: Sounds good, done.

WcaOnRails/app/views/competitions/_competition_form.html.erb
timreyn: Added to the edit below.

timreyn (commented):
Looks like some extraneous files under webpacker/packs/generated_locales got added. Manually reverting these.

gregorbg:
> Looks like some extraneous files under webpacker/packs/generated_locales got added. Manually reverting these.

Yes and No :sweat_smile: Between now and the time this PR was originally opened, we changed our JS i18n infrastructure. It used to write stuff to the `generated_locales` folder which we now renamed to `rails_translations` as part of the update process.
Arguably it was a bit unclever to un-ignore the old `generated_locales` folder because now older PRs like yours are trying to commit them :see_no_evil: Sorry!

gregorbg (changes requested):
Code is all good now, but I hate to point out that `structure.sql` is pretty messed up again :see_no_evil:

Sorry, we have to endure this as long as there's still PHP portions of the code relying on our DB structure and DSL... Very very high on WST's radar to get rid of this!

timreyn:
D'oh! Okay, should be fixed again.

gregorbg (approved):

Issue #7032, opened 2022-07-02
Result validators catch possible duplicate persons insufficiently


tnorr:
**Is your feature request related to a problem? Please describe.**
If a returning competitor registers with a new account or without an account and their name doesn't exactly match their previous profile, a duplicate profile is easily created. During a WRT investigation, we've noticed this happens a lot if:
- Competitor registers using a middle name that they didn't have in their previous profile or vice versa
- There's a minor difference in the names. This may stem from for example using a slightly different Chinese character, or having a typo (can happen with especially on-site registrations).

**Describe the solution you'd like**
A new result validator (similar to SAME_PERSON_NAME_WARNING we have currently) that catches matching birthdates and similar names. It should create a warning in these two cases:
- All of dob, first name and last name match to an existing person
- Dob matches and Levenshtein distance is 1 (2 can be considered too) between the newcomer's and an existing person's names

**Describe alternatives you've considered**
Creating a sanity check to cover the same issue. A result validator is more practical since it allows delegates to catch the possible duplicates and investigate already before posting the results.

**Additional context**
I created this issue to more easily give context to my draft PR.

danieljames-dj:
I would like to propose to replace SAME_PERSON_NAME_WARNING with following three warnings:

- SAME_NAME_SAME_DOB_WARNING
- SAME_NAME_DIFFERENT_DOB_WARNING
- SIMILAR_NAME_SAME_DOB_WARNING

### SAME_NAME_SAME_DOB_WARNING
Two persons will have exactly same name and exactly same DOB.

### SAME_NAME_DIFFERENT_DOB_WARNING
Two persons will have exactly same name but different DOB.

### SAME_DOB_SIMILAR_NAME_WARNING
Two persons will have exactly same DOB and similar name.

Algorithm to check if names are similar:
- If name is having two words: (eg: Lorem Ipsum)
    - Check if there is a name which starts with first name and ends with last name and there is something in middle (eg: Lorem Dolor Ipsum)
- If name is having more than two words: (eg: Lorem Dolor Ipsum)
    - Check if there is a name which has just first name and last name of the original person (eg: Lorem Ipsum)

This is just one algorithm, maybe we can add more algorithms to check similar names in future.

PR #7033, opened 2022-07-02
Adding a person validator for matching dob, first name and last name

tnorr:
Only implementing the first validator described in #7032 .

Still work in progress. There might be mistakes since I've not coded in Ruby before, so any feedback would be great.

The implementation is similar to the same person name warning, but I'm not sure how much computing the first and last name for every person slows the execution.

All searching, mapping and grouping is done with a pair of first name, last name and dob. This means the information about the full name of the person triggering the warning is not preserved and has to be searched separately. Line 89 is basically saying "what is the full name of the newcomer that has the same first and last name than the newcomer who triggered the warning". If there happened to be two newcomers with the exact same first name, last name and birthdate (extremely unlikely), the warning would be triggered twice, but only with the first newcomer's name. This is something that would probably need a more robust solution.


gregorbg:
Hey Timo, Thanks for your first contribution! I am terribly busy right now so I'm not sure when I will have the time to squeeze in a proper code review for this PR. I'll try my best to submit some thoughts by the end of this week.
In the meantime, if any of the other WST members know this area of the code or have some general remarks about Ruby, feel free to add your thoughts as you please!

tnorr:
Hi @gregorbg 

I'll still go through the errors that failed the check. I created #7039 to add simpler validators, and it would probably be best to review that first.

PR #7035, opened 2022-07-06
Feature/cache results

campos20:
Minimum impact in fixing rankings


campos20:
Early PR open to discuss. Each event should cache up to about 700kb (years excluded). I think we could also dismiss the 1000 results for good.

campos20:
We can also dismiss the hash need and cache it in a table similar to

key, rows (json), cache_time

and retrieve it. Cache would be done in the database.

campos20:
Should we display the date when results were cached?

WcaOnRails/db/migrate/20220706232200_create_cached_results.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class CreateCachedResults < ActiveRecord::Migration[6.1]
+  def change
+    create_table(:cached_results, id: false) do |t|
+      t.string :key_params
+      t.json :payload
+      t.timestamps
+    end
+    execute "alter table cached_results add primary key (key_params);"
```
gregorbg, line 10: No need for this, see my other comment above.

WcaOnRails/db/migrate/20220706232200_create_cached_results.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class CreateCachedResults < ActiveRecord::Migration[6.1]
+  def change
+    create_table(:cached_results, id: false) do |t|
+      t.string :key_params
```
gregorbg, line 6: `t.string :key_params, primary_key: true`

WcaOnRails/app/controllers/results_controller.rb
```diff
@@ -141,14 +128,23 @@ def rankings
...
+
     else
       flash[:danger] = t(".unknown_show")
       redirect_to rankings_path
     end
+
+    @rows = ActiveRecord::Base.connection.exec_query(@query)
+
+    # For safety, we delete possible duplicated (does active redord have on duplicated update or ignore?)
+    safe_duplate_query = "delete from cached_results where key_params = '#{cached_key}'"
```
gregorbg, line 143: Typo: `duplate` --> `duplicate`.
Also, `CachedResult.delete_by(key_params: cached_key)`

gregorbg (commented):
Thanks, but why are you still disallowing `year`? I remember our long video call, but that discussion was because `year` broke the new table structure that you had created.
Now, you're caching the **old** existing table structure, which can easily support `year`. Why drop it?

Also, wonderful idea about just clearing the cache with CAD. I like! :D

There's a part in `views/results/rankings.html.erb` which GitHub doesn't let me comment on. It has a block with `@skip_robot_indexing` that can be entirely deleted now.

gregorbg:
On second thought, I am a bit concerned whether this will solve our problem of bots killing the server.
If Chinese crawlers decide to try all Results permutations at once, it will still kill/overload the DB during initial computation. The fact that we cache stuff _after_ the initial load doesn't change the problem of initial load. Thoughts?

campos20:
> On second thought, I am a bit concerned whether this will solve our problem of bots killing the server. If Chinese crawlers decide to try all Results permutations at once, it will still kill/overload the DB during initial computation. The fact that we cache stuff _after_ the initial load doesn't change the problem of initial load. Thoughts?

We can put AWS lambdas cron to run accessing #GET...?params..., one after another, just after CAD (or scheduled independently each 12h so we don't impact developing code). If we put in CAD, we will have to make sure the project has AWS SDK and that the associated IAM role has trigger lambda permissions.

docker-compose-local-db.yml
```diff
@@ -0,0 +1,20 @@
+version: "3"
```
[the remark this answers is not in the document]
campos20: I'll open, but I have already a local WCA database setup and I'd like to run this with docker. Setting up a second WCA db is a waste of effort.

WcaOnRails/lib/auxiliary_data_computation.rb
```diff
@@ -113,4 +115,9 @@ def self.delete_php_cache
     cache_files = Dir.glob(Rails.root.join("../webroot/results/generated/cache/*.cache"))
     FileUtils.rm cache_files
   end
+
+  def self.delete_cached_results
+    # By deleting existing results, they can be cached again on the first execution
+    ActiveRecord::Base.connection.execute "delete from cached_results where 1=1"
```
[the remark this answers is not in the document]
campos20: Fixed

WcaOnRails/db/migrate/20220706232200_create_cached_results.rb
campos20: Fixed.

WcaOnRails/app/controllers/results_controller.rb
campos20: Fixed

docker-compose-local-db.yml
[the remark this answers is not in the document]
campos20: Robot indexing handled

WcaOnRails/db/migrate/20220706232200_create_cached_results.rb
campos20: Fixed

gregorbg:
Pre-caching the results after CAD is run is too much of a rabbit hole because of AWS SDK etc.

I think if we remove the 1000 option and stick with 100 for now, it will be fine. Can you change the code such that the 1000 Results option is _always_ greyed out with a gentle "WST is working on a more permanent solution" kind of tooltip?

campos20:
> Pre-caching the results after CAD is run is too much of a rabbit hole because of AWS SDK etc.
> 
> I think if we remove the 1000 option and stick with 100 for now, it will be fine. Can you change the code such that the 1000 Results option is _always_ greyed out with a gentle "WST is working on a more permanent solution" kind of tooltip?

Done

image

gregorbg:
Please be more specific about "Years will cause us issues". Again, same discussion as above. Why are you so adamant about years even in the old inefficient table structure?

campos20:
> Please be more specific about "Years will cause us issues". Again, same discussion as above. Why are you so adamant about years even in the old inefficient table structure?

2 Type
[single, average]

43 Years
[all years, 1982,2003-2022 only, 1982,2003-2022 until]

3 Gender
[All, male, female]

3 Show
[100 Person, 100 Results, By Region] 1000 is currently disabled

17 Events
215 regions

Total
2828970 keys

Each with about 45kb. Total up to 121gb
Each year adds to new keys ("only" and "until"), 131580 new keys, or up to 5gb.

gregorbg (approved):
Thanks, this is an explanation that I can live with 😄

PR #7039, opened 2022-07-07
Fixes #6782 - Add validators for obvious person name issues

### LINKED ISSUE #6782: Add additional result validators for obvious person name issues
Currently, the [persons validators](https://github.com/thewca/worldcubeassociation.org/blob/997e519371cf3758c3d72fca768edfbbd37733f3/WcaOnRails/lib/results_validators/persons_validator.rb) do not include checks for obvious person name issues, for instance:

- lowercase first name
- lowercase last name
- single letter middle name(s) without abbreviation period
- single letter first name (with or without abbreviation period)

Adding such checks would be of great help for WRT. The checks should be implemented as warnings (not errors).

tnorr:
Add a warning for 
- lowercase names
- single letter middle names / abbreviations without a period
- single letter first and last names (with or without a period)

Update unit tests to include the new cases.

### DIFF
--- a/WcaOnRails/lib/results_validators/persons_validator.rb
+++ b/WcaOnRails/lib/results_validators/persons_validator.rb
@@ -115,20 +115,20 @@ module ResultsValidators
             duplicate_newcomer_names << p.name
           end
           # Look for obvious person name issues
-          splitted_name = p.name.split
-          if splitted_name.any? { |n| n.downcase == n }
+          splt_name = p.name.split
+          if splt_name.any? { |n| n.downcase == n }
             @warnings << ValidationWarning.new(:persons, competition_id,
                                                LOWERCASE_NAME_WARNING,
                                                name: p.name)
           end
-          if splitted_name.length > 2
-            if splitted_name[0, splitted_name.length-2].any? { |n| n.length == 1 }
+          if splt_name.length > 2
+            if splt_name[0, splt_name.length-1].any? { |n| n.length == 1 }
               @warnings << ValidationWarning.new(:persons, competition_id,
                                                  MISSING_ABBREVIATION_PERIOD_WARNING,
                                                  name: p.name)
             end
           end
-          if p.name[0, 2].include?(" ") || p.name[0, 2].include?(".") || p.name.reverse[0, 2].include?(" ") || p.name.reverse[0, 2].include?(".")
+          if [' ', '.'].include?(p.name[1]) || p.name[-2] == (" ") || p.name[-1] == "." && p.name[-3] == " " 
             @errors << ValidationError.new(:persons, competition_id,
                                            SINGLE_LETTER_FIRST_OR_LAST_NAME_ERROR,
                                            name: p.name)
--- a/WcaOnRails/lib/results_validators/persons_validator.rb
+++ b/WcaOnRails/lib/results_validators/persons_validator.rb
@@ -114,21 +114,29 @@ module ResultsValidators
           if without_wca_id.select { |p2| p2.name == p.name }.length > 1 && !duplicate_newcomer_names.include?(p.name)
             duplicate_newcomer_names << p.name
           end
-          # Look for obvious person name issues
-          splt_name = p.name.split
-          if splt_name.any? { |n| n.downcase == n }
+          # Look for obvious person name issues (in roman-readable part)
+          if p.name.include? " ("
+            roman_readable = p.name[0, p.name.index('(')-1]
+          else
+            roman_readable = p.name
+          end
+          split_name = roman_readable.split
+          if split_name.any? { |n| n.downcase == n }
             @warnings << ValidationWarning.new(:persons, competition_id,
                                                LOWERCASE_NAME_WARNING,
                                                name: p.name)
           end
-          if splt_name.length > 2
-            if splt_name[0, splt_name.length-1].any? { |n| n.length == 1 }
+          if split_name.length > 2
+            if split_name[1, split_name.length-2].any? { |n| n.length == 1 }
               @warnings << ValidationWarning.new(:persons, competition_id,
                                                  MISSING_ABBREVIATION_PERIOD_WARNING,
                                                  name: p.name)
             end
           end
-          if [' ', '.'].include?(p.name[1]) || ((p.name[-2] == " ") && !['I', 'V'].include?(p.name[-1])) || ((p.name[-1] == ".") && (p.name[-3] == " "))
+          non_word_after_first_letter = [' ', '.'].include?(p.name[1])
+          space_before_last_letter = (p.name[-2] == " ") && !['I', 'V'].include?(p.name[-1]) # Numerical suffixes I and V have to be excluded to pass unit tests
+          abbreviated_last_name = (p.name[-1] == ".") && (p.name[-3] == " ")
+          if non_word_after_first_letter || space_before_last_letter || abbreviated_last_name
             @warnings << ValidationWarning.new(:persons, competition_id,
                                                SINGLE_LETTER_FIRST_OR_LAST_NAME_WARNING,
                                                name: p.name)
--- a/WcaOnRails/lib/results_validators/persons_validator.rb
+++ b/WcaOnRails/lib/results_validators/persons_validator.rb
@@ -128,7 +128,7 @@ module ResultsValidators
                                                  name: p.name)
             end
           end
-          if [' ', '.'].include?(p.name[1]) || p.name[-2] == (" ") || p.name[-1] == "." && p.name[-3] == " " 
+          if [' ', '.'].include?(p.name[1]) || (p.name[-2] == (" ")) || ((p.name[-1] == ".") && (p.name[-3] == " "))
             @errors << ValidationError.new(:persons, competition_id,
                                            SINGLE_LETTER_FIRST_OR_LAST_NAME_ERROR,
                                            name: p.name)
--- a/WcaOnRails/lib/results_validators/persons_validator.rb
+++ b/WcaOnRails/lib/results_validators/persons_validator.rb
@@ -128,7 +128,7 @@ module ResultsValidators
                                                  name: p.name)
             end
           end
-          if [' ', '.'].include?(p.name[1]) || ((name[-2] == " ") && !['I', 'V'].include?(name[-1])) || ((p.name[-1] == ".") && (p.name[-3] == " "))
+          if [' ', '.'].include?(p.name[1]) || ((p.name[-2] == " ") && !['I', 'V'].include?(p.name[-1])) || ((p.name[-1] == ".") && (p.name[-3] == " "))
             @warning << ValidationWarning.new(:persons, competition_id,
                                               SINGLE_LETTER_FIRST_OR_LAST_NAME_WARNING,
                                               name: p.name)
--- a/WcaOnRails/lib/results_validators/persons_validator.rb
+++ b/WcaOnRails/lib/results_validators/persons_validator.rb
@@ -117,13 +117,13 @@ module ResultsValidators
           end
           # Look for obvious person name issues
           splitted_name = p.name.split
-          if splitted_name.any?( |n| n.downcase == n)
+          if splitted_name.any? { |n| n.downcase == n}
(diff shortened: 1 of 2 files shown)


WcaOnRails/lib/results_validators/persons_validator.rb
```diff
@@ -111,6 +114,25 @@ def validate(competition_ids: [], model: Result, results: nil)
           if without_wca_id.select { |p2| p2.name == p.name }.length > 1 && !duplicate_newcomer_names.include?(p.name)
             duplicate_newcomer_names << p.name
           end
+          # Look for obvious person name issues
+          splitted_name = p.name.split
```
coder13, line 118: Nit on spelling. I would probably calling this `splt_name`.

"Splitted" is [an archaic past tense spelling](https://www.merriam-webster.com/dictionary/splitted) and doesn't read well.

WcaOnRails/lib/results_validators/persons_validator.rb
```diff
@@ -111,6 +114,25 @@ def validate(competition_ids: [], model: Result, results: nil)
...
+                                               name: p.name)
+          end
+          if splitted_name.length > 2
+            if splitted_name[0, splitted_name.length-2].any? { |n| n.length == 1 }
+              @warnings << ValidationWarning.new(:persons, competition_id,
+                                                 MISSING_ABBREVIATION_PERIOD_WARNING,
+                                                 name: p.name)
+            end
+          end
+          if p.name[0, 2].include?(" ") || p.name[0, 2].include?(".") || p.name.reverse[0, 2].include?(" ") || p.name.reverse[0, 2].include?(".")
```
coder13, line 131: This seems convoluted. Couldn't you just take the split name and check if any of them are length 1?

tnorr: This should only apply to first and last names, since abbreviations are allowed for middle names. It should also catch abbreviations with a period, which have length 2.

WcaOnRails/lib/results_validators/persons_validator.rb
```diff
@@ -22,6 +22,9 @@ class PersonsValidator < GenericValidator
     MULTIPLE_NEWCOMERS_WITH_SAME_NAME_WARNING = "There are multiple new competitors with the exact same name: %{name}. Please ensure that all results are correct for these competitors " \
                                                 "and that all results are correctly seperated by their corresponding id."
     WRONG_PARENTHESIS_TYPE_ERROR = "The parenthesis character used in '%{name}' is an irregular character, please replace it with a regular parenthesis '(' or ')' and with appropriate spacing."
+    LOWERCASE_NAME_WARNING = "'%{name}' has a lowercase name, please ensure the correct spelling."
+    MISSING_ABBREVIATION_PERIOD_WARNING = "'%{name}' is missing an abbreviation period from a single letter middle name, please ensure the correct spelling."
```
gregorbg, line 26: This warning message says "middle name", but below you're actually also checking for the first name (the array index runs from 0 inclusively)

WcaOnRails/lib/results_validators/persons_validator.rb
```diff
@@ -111,6 +114,25 @@ def validate(competition_ids: [], model: Result, results: nil)
           if without_wca_id.select { |p2| p2.name == p.name }.length > 1 && !duplicate_newcomer_names.include?(p.name)
             duplicate_newcomer_names << p.name
           end
+          # Look for obvious person name issues
+          splt_name = p.name.split
```
gregorbg, line 118: Typo: `splt` --> `split` (throughout the entire set of checks, because the variable needs to be renamed everywhere)

WcaOnRails/lib/results_validators/persons_validator.rb
```diff
@@ -111,6 +114,25 @@ def validate(competition_ids: [], model: Result, results: nil)
...
+                                               name: p.name)
+          end
+          if splt_name.length > 2
+            if splt_name[0, splt_name.length-1].any? { |n| n.length == 1 }
+              @warnings << ValidationWarning.new(:persons, competition_id,
+                                                 MISSING_ABBREVIATION_PERIOD_WARNING,
+                                                 name: p.name)
+            end
+          end
+          if [' ', '.'].include?(p.name[1]) || ((p.name[-2] == " ") && !['I', 'V'].include?(p.name[-1])) || ((p.name[-1] == ".") && (p.name[-3] == " "))
```
gregorbg, line 131: This `if`-check might be clever and efficient, but it looks really confusing and incredibly complex to the naked eye.

What about splitting the different `||` parts into separate variables with descriptive names first, and then on a separate line combine these?

```ruby
non_word_after_first_letter = [' ', '.'].include?(p.name[1])
descriptive_foo = (p.name[-2] == " ") && !['I', 'V'].include?(p.name[-1])
descriptive_bar = (p.name[-1] == ".") && (p.name[-3] == " ")

if non_word_after_first_letter || descriptive_foo || descriptive_bar
  ... do stuff ...
end
```

gregorbg: Also, unrelated to my previous comment: Probably worth including a comment on why `I` and `V` are specifically excluded. I have a hunch, but not every person looking at this code might know.

WcaOnRails/spec/lib/results_validators/persons_validator_spec.rb
```diff
@@ -154,6 +157,18 @@
...
+                                          eventId: "333oh")
+        res_lowercase.person.update(name: "First Middle last")
+        res_missing_period = FactoryBot.create(:inbox_result,
+                                               competition: competition1,
+                                               eventId: "333oh")
+        res_missing_period.person.update(name: "First A B. Last")
+        res_single_letter = FactoryBot.create(:inbox_result,
+                                              competition: competition1,
+                                              eventId: "333oh")
+        res_single_letter.person.update(name: "A. B. Last")
```
gregorbg, line 171: What about test cases with foreign scripts like Japanese or Cyrillic?

In particular, your "downcase" verification might (or might not, I legitimately don't know which is why you should definitely add test cases to make sure!) catch stuff like `Yu Nakajima (中島悠)` where the `(中島悠)` is also `(中島悠)` in "lowercase".

gregorbg (changes requested):
One typo and some remarks regarding code style.
Also, I am curious to see how your verifications handle foreign alphabets.

WcaOnRails/lib/results_validators/persons_validator.rb
tnorr: Fixed!

WcaOnRails/lib/results_validators/persons_validator.rb
tnorr: Fixed!

WcaOnRails/lib/results_validators/persons_validator.rb
tnorr: I split the expressions into variables, it's definitely a lot neater this way.

I initially didn't exclude I and V, but it turned out I had to in order to pass unit tests (fake names sometimes had numerical suffixes, which created unexpected warnings).

WcaOnRails/spec/lib/results_validators/persons_validator_spec.rb
tnorr: This is a good catch, I confirmed that at least Japanese and Chinese characters are the same in downcase, so they would have always triggered the warning. I excluded local names from all checks. Now the validators also catch single letter last names followed by a local name. A lowercase check for Cyrillic local names would still be practical, but it's a bit more tedious to implement if we'd want to exclude other character sets.

WcaOnRails/spec/lib/results_validators/persons_validator_spec.rb
```diff
@@ -154,6 +157,22 @@
...
+                                           eventId: "333oh")
+        res_lowercase1.person.update(name: "First Middle last")
+        res_lowercase2 = FactoryBot.create(:inbox_result,
+                                           competition: competition1,
+                                           eventId: "333oh")
+        res_lowercase2.person.update(name: "İlis хocavənd V")
+        res_missing_period = FactoryBot.create(:inbox_result,
+                                               competition: competition1,
+                                               eventId: "333oh")
+        res_missing_period.person.update(name: "First A B. Last (最初)")
```
gregorbg, line 171: lol, that Japanese part actually means "first" (in a row of elements or in a sequence).
If you want to test a typical Japanese generic name, pick "Yamada Taro (山田太郎)".

gregorbg (commented):
Totally fine now code-wise, just one nitpick about "We need this rule because otherwise some test breaks" comment.

WcaOnRails/spec/lib/results_validators/persons_validator_spec.rb
tnorr: Yeah, my intention was to just add something representing a local name without putting much thought. It might be a good idea to write a bit more sensible test names.

campos20:
Side note: we have a project that runs every month and check for obvious irregularities like the ones you mentioned. Queries used to be public and they can be found here

https://github.com/thewca/db-sanity-check/blob/d4a96677023acee17033399b8def452137dade14/queries.txt

Maybe it's worth it looking at the first 8 queries (N1-N8)

tnorr:
> Side note: we have a project that runs every month and check for obvious irregularities like the ones you mentioned. Queries used to be public and they can be found here
> 
> https://github.com/thewca/db-sanity-check/blob/d4a96677023acee17033399b8def452137dade14/queries.txt
> 
> Maybe it's worth it looking at the first 8 queries (N1-N8)

I'm well aware of the sanity checks. However, these do not completely serve the same purpose as having a warning for this issue.

Every incorrect name overlooked by the delegate means increased workload, for both the delegate and WRT. During posting results, we as WRT catch incorrectly written names a lot. Usually we fix these ourselves, but still need to confirm with the delegate. In more severe cases, name issues lead to postponing the result posting, which affects a number of competitors. If an incorrect name is completely overlooked and triggers a sanity check, it will still require the same amount of additional work and also means the data integrity has been compromised for up to a month. 

Having the warning would ensure that incorrectly written names don't go unnoticed by delegates, saving unnecessary work (which is surely what @SAuroux also had in mind with #6782 )

gregorbg:
Thanks for providing the additional explanation and resolving the merge conflict. My two remarks from the last review still remain unresolved, though.

Do you need any additional input from WST? Let's get this done together ASAP :)

WcaOnRails/lib/results_validators/persons_validator.rb
```diff
@@ -111,6 +114,33 @@ def validate(competition_ids: [], model: Result, results: nil)
...
+          end
+          if split_name.length > 2
+            if split_name[1, split_name.length-2].any? { |n| n.length == 1 }
+              @warnings << ValidationWarning.new(:persons, competition_id,
+                                                 MISSING_ABBREVIATION_PERIOD_WARNING,
+                                                 name: p.name)
+            end
+          end
+          non_word_after_first_letter = [' ', '.'].include?(p.name[1])
+          space_before_last_letter = (p.name[-2] == " ") && !['I', 'V'].include?(p.name[-1]) # Numerical suffixes I and V have to be excluded to pass unit tests
```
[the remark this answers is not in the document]
tnorr: Changed this comment!

WcaOnRails/spec/lib/results_validators/persons_validator_spec.rb
```diff
@@ -160,19 +160,19 @@
...
+        res_lowercase1.person.update(name: "Yamada taro (山田太郎)")
         res_lowercase2 = FactoryBot.create(:inbox_result,
                                            competition: competition1,
                                            eventId: "333oh")
         res_lowercase2.person.update(name: "İlis хocavənd V")
         res_missing_period = FactoryBot.create(:inbox_result,
                                                competition: competition1,
                                                eventId: "333oh")
-        res_missing_period.person.update(name: "First A B. Last (最初)")
+        res_missing_period.person.update(name: "Missing A Period")
```
gregorbg, line 171: LOL

gregorbg (approved):
Thank you for the fixes and the wonderful easteregg :)

Issue #7041, opened 2022-07-11
"Application Instructions" in "Regional Organizations" page is displayed in HTML


BelbioK:
**Describe the bug**
The link to the form is displayed in HTML.

**To Reproduce**
1. Go to [https://www.worldcubeassociation.org/organizations](https://www.worldcubeassociation.org/organizations)
2. Scroll down to the bottom of the page
3. See error

**Expected behavior**
You should see the link to the form

**Screenshots**
![image](https://user-images.githubusercontent.com/109048458/178169159-d302b151-9173-47e0-8694-168f59c6164f.png)

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser Opera GX and Firefox

dunkOnIT:
Thanks for raising this! @gregorbg am I correct that this should be resolved once @notrichardpeng PR with the React Regional Orgs is merged? 

Or if that has already happened, I would assume we need to fix this :P

gregorbg:
It has actually been merged, and this is a follow-up bug from the React migration. We need to fix this.

notrichardpeng:
oops this wasn't spotted, I will try to get this fixed soon

gregorbg:
Fixed per #7042

PR #7042, opened 2022-07-11
Fix RO apply link

notrichardpeng:
#7041 

In React, HTML written in locales break because JSX expressions automatically add double quotes to the translated sentences. This is fixed by moving the apply link to ruby views, but makes the link also appear in the case where there are no regional organizations, which on a second thought, makes sense too.

Case I'm talking about:

image

### DIFF
--- a/WcaOnRails/config/i18n-tasks.yml.erb
+++ b/WcaOnRails/config/i18n-tasks.yml.erb
@@ -154,6 +154,10 @@ ignore:
 <%
   require './lib/custom_wca_i18n_scanner.rb'
   I18n::Tasks.add_scanner 'CustomWcaI18nScanner', only: %w(*.erb)
+
+  I18n::Tasks.add_scanner 'I18n::Tasks::Scanners::PatternMapper',
+    only: %w(*.js),
+    patterns: [['<I18nHTMLTranslate\si18nKey=%{key}\s/>', '%{key}']]
 %>
 # <%#= I18n::Tasks.add_scanner 'I18n::Tasks::Scanners::PatternMapper',
 #        only: %w(*.html.haml *.html.slim),
--- a/WcaOnRails/config/i18n-tasks.yml.erb
+++ b/WcaOnRails/config/i18n-tasks.yml.erb
@@ -157,7 +157,7 @@ ignore:
 
   I18n::Tasks.add_scanner 'I18n::Tasks::Scanners::PatternMapper',
     only: %w(*.js),
-    patterns: [['<I18nHTMLTranslate\si18nKey=%{key}\s/>', '%{key}']]
+    patterns: [['I18nHTMLTranslate\si18nKey=%{key}', '%{key}']]
 %>
 # <%#= I18n::Tasks.add_scanner 'I18n::Tasks::Scanners::PatternMapper',
 #        only: %w(*.html.haml *.html.slim),
--- a/WcaOnRails/config/i18n-tasks.yml.erb
+++ b/WcaOnRails/config/i18n-tasks.yml.erb
@@ -157,7 +157,7 @@ ignore:
 
   I18n::Tasks.add_scanner 'I18n::Tasks::Scanners::PatternMapper',
     only: %w(*.js),
-    patterns: [['I18nHTMLTranslate\si18nKey=%{key}', '%{key}']]
+    patterns: [['I18nHTMLTranslate\s+i18nKey=%{key}', '%{key}']]
 %>
 # <%#= I18n::Tasks.add_scanner 'I18n::Tasks::Scanners::PatternMapper',
 #        only: %w(*.html.haml *.html.slim),
--- a/WcaOnRails/app/views/regional_organizations/index.html.erb
+++ b/WcaOnRails/app/views/regional_organizations/index.html.erb
@@ -4,4 +4,5 @@
   <%= react_component("RegionalOrganizations/index", {
     orgs: @acknowledged_regional_organizations,
   }) %>
+  <%= t('regional_organizations.application_instructions.description_html') %>
 </div>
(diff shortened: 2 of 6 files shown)


gregorbg (changes requested):
I am not happy about backporting stuff to Ruby just because of minor nuisances like this.

Also, please consider: This is possible because the link text coincidentally appears at the bottom of the page, so you can extract it quite easily. What if link texts appear in the middle of the page instead? What about dynamic I18N texts?

Please research a "proper" React sokution for this kind of problem.

notrichardpeng:
Great second point. After a while I found `dangerouslySetInnerHTML`, but I should include [DOMpurify](https://github.com/cure53/DOMPurify) for security purposes.

gregorbg:
Thanks, in terms of your changes it's all good now!

However, I am still not happy about using `dangerouslySetInnerHTML` all the time. It will happen more and more as we start migrating more components to React, and it's definitely not good practice. I am open for ideas and debate how to improve this.

coder13:
We could use markdown to react libraries and assert that the locales text is written in valid markdown

gregorbg:
That would require us to either
1. Change our existing Ruby templates, which use the same i18n YAML
2. Maintain some keys in plain HTML template and some in Markdown, confusing both translators and developers.

I think the main issue is that Ruby and React have two different opinions about how to handle inline HTML in translations.

gregorbg:
Separate YAML files are not an option because it makes life hard for non-tech-savvy translators.
What about creating a React component like `TranslationString` that can be used as 
```jsx
<SomeComponent foo="bar">
  <h3>Heading</h3>
  <TranslationString key="some.translation.yaml.key"/>
</SomeComponent>
```

It can then handle all of the `dompurify` and innerHTML stuff by itself and other parts of the code can reuse it as well.

notrichardpeng:
Ohh okay, seems good. So at last we can trust `dangerouslySetInnerHTML` + `dompurify`?

coder13:
I think it's fine to use `dangerouslySetInnerHTML` as long as you know what you're doing. That's why it's called `dangerouslySetInnerHTML`. We should make sure that our content is sanitized well as it goes into `dangerouslySetInnerHTML`.

gregorbg:
+1 Caleb. I think this underlines the importance of creating a custom component, that can only be passed an i18n key and not actual HTML. In doing so, we don't encourage devs to `dangerouslySetInnerHTML` all the time and we also have implicit control over that part of sanitizing because the data only comes from one "source of truth" -- namely the translation YAMLs

notrichardpeng:
Sounds like a viable solution. I will go ahead and implement this then?

gregorbg:
Please do :)

notrichardpeng:
I changed it to a more generic solution to avoid getting the "I18n key not used" test fail, but maybe this is actually just better because it can also be used for things other than translations?

gregorbg:
At the current point in time, I am specifically opposed to allowing generic content here. We should limit it to i18n keys only.

Preventing the test fail is a matter of configuring our "find unused i18n keys" library in a way that detects these usages. See https://github.com/glebm/i18n-tasks/blob/main/templates/config/i18n-tasks.yml#L142 for reference.

gregorbg (approved):

gregorbg:
Code is fine, now you just have to update the scanner for `i18n-tasks` to make sure it detects the usages.

notrichardpeng:
I finally have more free time now. I looked at how to add the scanner again and this might take a while as I have no experience with regex :P ! Also `i18n-tasks` points to `custom_wca_i18n_scanner.rb`, are scanners for React also there? Want to make sure because I only see stuff for Ruby.

gregorbg:
`CustomWcaI18NScanner` is its own thing that is not related to this feature. Just ignore it and pretend it doesn't even exist.

Essentially, `i18n-tasks` is doing a very stupid walk over all source files, looking for strings along the form of `t('some.key')` and if it finds such a string, it considers `'some.key'` as "used" and all is good.

You need to make the library understand that `<YourCoolNewReactComponent key='some.key' />` is equal to 't('some.key')' in terms of considering the i18n key as "used". To do so, you can use the pre-configured `PatternMapper` that comes delivered with the library, you just have to tell it which patterns to look out for.

The basic syntax is like this:
```ruby
I18n::Tasks.add_scanner 'I18n::Tasks::Scanners::PatternMapper',
    only: %w(*.html.haml *.html.slim),
    patterns: [['= title\b', '.page_title']] %>
```

where `only` is a simple glob pattern over the files you want to consider in the first place (probably something like `*.jsx` in this case) and `patterns` is an array of replacement rules where each replacement rule in itself is a two-element-array of the shape `[how_it_looks_in_our_actual_source_code, what_key_you_should_extract_from_it]`. One `PatternMapper` can hold arbitrarily many patterns, but in the above example it holds just one replacement pattern which makes the double array braces look weird.

Does that help?

notrichardpeng:
Thanks Gregor that helped a lot! The tests are working locally and I just pushed the changes.

WcaOnRails/config/i18n-tasks.yml.erb
```diff
@@ -154,6 +154,10 @@ ignore:
 <%
   require './lib/custom_wca_i18n_scanner.rb'
   I18n::Tasks.add_scanner 'CustomWcaI18nScanner', only: %w(*.erb)
+
+  I18n::Tasks.add_scanner 'I18n::Tasks::Scanners::PatternMapper',
+    only: %w(*.js),
+    patterns: [['I18nHTMLTranslate\si18nKey=%{key}', '%{key}']]
```
gregorbg, line 160: Please change the pattern to use something like `\s+` (where `+` is RegExp for "one or more") such that line breaks and tab characters aren't ignored by this pattern match.

gregorbg (changes requested):
Cool! Small but important nit about the pattern you came up with. Everything else good to go!

gregorbg (approved):

PR #7050, opened 2022-07-15
Run Competition validators on WCIF PATCH API call

gregorbg:
See title. Previously, it was possible to create invalid competitions (as in, the Rails validations don't check out) even through our own website. An example is `HelloQ8Speedcubing`, where an event was edited using our own Events Editor to include a qualification limit, while qualification limits were never applied for in the first place.

This happened because the WCIF patch API internally uses `set_wcif!` which in its current state only operates on entities affiliated with a competition, but never on the competition row itself. In particular, the `update_column` call at the end never triggered validations as per https://guides.rubyonrails.org/active_record_validations.html#skipping-validations.

My suggested solution is to trigger a (seemingly) redundant `save!` within the transaction itself, so that validations are run and a rollback is triggered when necessary (i.e. when the competition is invalid after applying the patches). This will also update the `updated_at` timestamp, so the previous manual call is not necessary anymore.


gregorbg:
A note about the updated test: That spec is using a mock competition with events `["222", "444"]`. Our test factory assumes `222` is the main event, and before this change the test was trying to remove `222`, i.e. the main event. But not even Board members are allowed to remove the main event, as this would leave the competition in an incosistent state. They _are_ allowed to _change_ the main event and then potentially delete 222 as the old main event, but that's covered by a different test altogether.

gregorbg:
CC @timreyn This kind of inconsistent behavior has been discovered now that your qualifications feature is out in the wild.

The original `patch_wcif` method was not created by you, so it's not your fault. Plus I accept much of the blame for not spotting the superfluous `puts` statement during parsing, as I originally provided the code reviews.

Just wanted to FYI you nonetheless so that you can keep stuff like this in mind for future contributions :)

timreyn:
Thanks Gregor! And sorry for the extraneous `puts`, I'll try to make sure to remove my debugging in the future before merging :)

Just to make sure I understand -- the issue is that you can create a competition, add qualification times, and it's now invalid? An issue is that we don't actually show the qualification reason box unless the competition already has qualification times, so the first time you set qualification times, the competition is invalid until you go back to the competition form and add a reason. So I think you also need to drop this `if` https://github.com/thewca/worldcubeassociation.org/blob/master/WcaOnRails/app/views/competitions/_competition_form.html.erb#L309, otherwise it would be impossible to set qualifications on a new competition.

timreyn:
I suppose one option is to move the validation for qualification times to a warning, rather than a validation error (https://github.com/thewca/worldcubeassociation.org/blob/591dec77f6bb7e9e6ba36f18d4c42f8b8e226dfd/WcaOnRails/app/models/competition.rb#L386). Then the organizer and WCAT will see the alert but it won't prevent saving. WDYT?

gregorbg:
I do think we should have a "strong" validation, i.e. an error, that allows qualifications only if they have been permitted by WCAT. Maybe we should take a step back and re-introduce the separate "I want to use qualifications" check box that was coupled to the bigger "reasons" text field via jQuery?

That way, you can click the checkbox to enable basically setting the qualifications in the "EditEvents" form, but you can save time by leaving the text field blank and fill it out later before submitting to WCAT.

gregorbg:
I have now reverted the qualification limit configuration to an explicit checkbox. I understand the ambition to make this "convenient" for the user by only showing the textbox if necessary, but this has already caused confusion and a minor incident. We have Competition Requirement Policies which require a certain degree of formality and bureaucratic strictness, and as such I feel strongly in favor of adding back the checkbox.

As part of the aforementioned incident, I have also changed the button so that qualifications cannot be added / changed after the competition has been announced. @timreyn Please feel free to add your thoughts :)

gregorbg:
Also CC @coder13 because this PR introduces a minute but very important change to the `disabled` property of the Qualifications Button in the form that you're refactoring.

gregorbg:
Penny for your thoughts @timreyn :) I have to

WcaOnRails/app/models/competition.rb
```diff
@@ -195,6 +195,8 @@ class Competition < ApplicationRecord
            allow_nil: true,
            with_model_currency: :currency_code
   validates :early_puzzle_submission_reason, presence: true, if: :early_puzzle_submission?
+  # cannot validate `qualification_results IN [true, false]` because we historically have competitions
```
timreyn, line 198: Is it worth having a nullable boolean, versus just defaulting to false for old competitions?

timreyn (approved):
Thanks! Just one comment, looks good otherwise.

WcaOnRails/app/models/competition.rb
gregorbg: Neat thought that I also had while working on this fix. But changing the column default of an already existing column feels out of the scope of this PR, especially given how urgent it is.

PR #7052, opened 2022-07-16
Fixes #4535 team cache issue

### LINKED ISSUE #4535: Teams are cached for too long
While debugging tests for #4534 we found that `Team.wrc` was cached across tests. In theory, this could have repercussions in production (although the impact is limited in practice, because team members don't change often).

danieljames-dj:
(Sorry for the 4th active PR from me. But thought of giving this as I'm in a blocker for other three PRs)

This PR is the fix as suggested by @jonatanklosko for issue #4535 in [the comment](https://github.com/thewca/worldcubeassociation.org/issues/4535#issuecomment-1102487741).

### DIFF
--- a/WcaOnRails/app/models/team.rb
+++ b/WcaOnRails/app/models/team.rb
@@ -75,7 +75,15 @@ class Team < ApplicationRecord
 
   # Code duplication from Cachable concern, as we index by friendly_id and not by id :(
   def self.c_all_by_friendly_id
-    @@teams_by_friendly_id ||= all.with_hidden.index_by(&:friendly_id)
+    @@teams_by_friendly_id ||= nil
+    @@teams_by_friendly_id_timestamp ||= nil
+
+    if @@teams_by_friendly_id.nil? || @@teams_by_friendly_id_timestamp < 15.minutes.ago
+      @@teams_by_friendly_id = all.with_hidden.index_by(&:friendly_id)
+      @@teams_by_friendly_id_timestamp = DateTime.now
+    end
+
+    @@teams_by_friendly_id
   end
 
   def self.c_find_by_friendly_id!(friendly_id)
--- a/WcaOnRails/app/models/team.rb
+++ b/WcaOnRails/app/models/team.rb
@@ -78,7 +78,7 @@ class Team < ApplicationRecord
     @@teams_by_friendly_id ||= nil
     @@teams_by_friendly_id_timestamp ||= nil
 
-    if @@teams_by_friendly_id.nil? || @@teams_by_friendly_id_timestamp < 15.minutes.ago
+    if @@teams_by_friendly_id_timestamp.nil? || @@teams_by_friendly_id_timestamp < 15.minutes.ago
       @@teams_by_friendly_id = all.with_hidden.index_by(&:friendly_id)
       @@teams_by_friendly_id_timestamp = DateTime.now
     end


WcaOnRails/app/models/team.rb
```diff
@@ -75,7 +75,14 @@ def self.all_officers
 
   # Code duplication from Cachable concern, as we index by friendly_id and not by id :(
   def self.c_all_by_friendly_id
-    @@teams_by_friendly_id ||= all.with_hidden.index_by(&:friendly_id)
+    @@teams_by_friendly_id ||= nil
```
gregorbg, line 78: Why are you setting `teams_by_friendly_id` explicitly to `nil`, but not `teams_by_friendly_id_timestamp`?
Please change it for either both or none. I think unknown variables in Ruby are set to `nil` anyways, so my preference would be to drop this line altogether. But please check about the implicit `nil` beforehand.

gregorbg (changes requested):
Thanks for implementing the suggested fix! Just a nitpick about code consistency.

WcaOnRails/app/models/team.rb
jonatanklosko: I'm pretty sure we need to initialize a class variable before accessing it. For `teams_by_friendly_id_timestamp`teams_by_friendly_id_timestamp it's not the case, because it's guarded by the `@@teams_by_friendly_id.nil?` check. That said, initializing both explicitly sounds good too :)

danieljames-dj: I've initialized `teams_by_friendly_id_timestamp` as well to nil.

WcaOnRails/app/models/team.rb
```diff
@@ -75,7 +75,15 @@ def self.all_officers
...
+    @@teams_by_friendly_id_timestamp ||= nil
```
gregorbg, line 79: Initializing to `nil` won't work because then you check `nil < 15.minutes.ago`.

The only reason why this doesn't explode right now is that `teams_by_friendly_id.nil?` is checked first "by accident". For a clean implementation, please use `Time.at(0)` (UNIX epoch) or a similar past value like `1.year.ago` as default.

gregorbg (changes requested):

WcaOnRails/app/models/team.rb
jonatanklosko: I don't really agree, the check is not by accident, it's very deliberate. Initializing with nil is natural in this case, since we don't really have a value for it. If relying on the other check feels like to much, we can also add and explicit `teams_by_friendly_id_timestamp.nil?`. I'd say it's not necessary, but that's subjective I guess :)

gregorbg: But still, a part of the if-check is an invalid `<` that definitely fails if someone were to swap the operands. Are you really convinced this isn't an issue?

jonatanklosko: I think chaining conditions, in the sense that the first one "guards" the next one, is common in a number of languages. In most cases it concerns the same attribute as in `user.age && user age >= 21`, but that's also not swappable.

Maybe it would be easier to follow if we did `if @@teams_by_friendly_id_timestamp.nil? or @@teams_by_friendly_id_timestamp < 15.minutes`, but I'm fine with either :) I just wouldn't initialize the timestamp with an arbitrary value, because I consider that more confusing.

gregorbg: Yes, but the notable difference is that in your `user.age && user.age >= 21` sample you're checking a variable's existence (or non-nil-ness) before accessing _that exact variable_. I don't have an issue with these kinds of guard clauses at all, as you say yourself they're a common pattern in many languages.

The point why I'm still debating here is because we're checking two (technically!) unrelated variables where one guards the other just by the way the code two lines further down happens to be designed. This is a very fragile setup in my opinion.

Anyways, @danieljames-dj You have a pretty good idea of our two standpoints and arguments. It's your feature so please make a call on how you would like to proceed.

gregorbg: Something along those lines! Except I'd reckon `||` instead of `or` will do.

jonatanklosko: Ah totally, `or` was just me being too Elixir-ish :D

danieljames-dj: I've changed to `if @@teams_by_friendly_id_timestamp.nil? or @@teams_by_friendly_id_timestamp < 15.minutes` since that's the middle ground for all of us. :-)

gregorbg (approved):

PR #7055, opened 2022-07-20
Add registration eligibility to competitions on the competition list (Fixes #7036)

### LINKED ISSUE #7036: Add registration eligibility to competitions on the competition list
From "Fwd: [WCA Website] General Competition Comment by Joshua Wilbourn on 25 Apr 2022 at 12:40":

> Hello! I just has a small suggestion for the WCA Competition list website. If an icon can be added that shows registration is either open, closed, or upcoming before needing to click on it, that would be awesome! I tend to scour the website on occasion so that I can find a competition in my area only sometimes to be disappointed to find that it's already closed. The addition of a quick visual identifier would be a great benefit to potential competitors. Thanks!

From "Fwd: [WCA Website] General Competition Comment by Srisaran on 04 Jul 2022 at 18:24":
>  Kindly show registration FULL or OPEN information. Even if close date has passed alert. 

From these we can infer a few statuses:
- Announced (registration not yet open)
- Open (competitors may now register)
- Full (competitor limit reached, waiting list now in effect)
- Closed (competition is no longer taking registrations)

The task for this issue would be to hook into the existing state management for competitions (likely a superset of what I've described above) and make that represented on the [list of competitions ](https://www.worldcubeassociation.org/competitions) and on the competition page itself.

simonkellly:
This PR aims to add icons showing the current status of a competition's registration on the competitions list, in response to issue #7036.

In it's current state it does not currently integrate with the translations, but I thought it would be worth making a draft to see if this is the correct direction to head in
image

### DIFF
--- a/WcaOnRails/app/helpers/competitions_helper.rb
+++ b/WcaOnRails/app/helpers/competitions_helper.rb
@@ -268,4 +268,33 @@ module CompetitionsHelper
   def cached_pdf_name(competition, colors)
     "#{pdf_name(competition)}_#{competition.updated_at.iso8601}_#{colors}"
   end
+
+  def registration_status_icon(competition)
+    icon = ""
+    title = ""
+    icon_class = ""
+
+    if competition.registration_not_yet_opened?
+      icon = "clock"
+      title = "Registraion opens in #{(competition.start_date - Date.today).to_i} days"
+      icon_class = "blue"
+    elsif competition.registration_past?
+      icon = "times circle"
+      title = "Registration closed"
+      icon_class = "red"
+    elsif competition.registration_full?
+      icon = "exclamation circle"
+      title = "Registration full, waiting list open"
+      icon_class = "orange"
+    else
+      icon = "check circle"
+      title = "Registration open"
+      icon_class = "green"
+    end
+
+    ui_icon(icon,
+            title: title,
+            class: icon_class,
+            data: { toggle: "tooltip" })
+  end
 end
--- a/WcaOnRails/app/helpers/competitions_helper.rb
+++ b/WcaOnRails/app/helpers/competitions_helper.rb
@@ -276,18 +276,18 @@ module CompetitionsHelper
 
     if competition.registration_not_yet_opened?
       icon = "clock"
-      title = I18n.t('competitions.index.tooltips.registration.opens_in', days: t('common.days', count: (Date.today - competition.end_date).to_i))
+      title = I18n.t('competitions.index.tooltips.registration.opens_in', duration: distance_of_time_in_words_to_now(competition.registration_open))
       icon_class = "blue"
     elsif competition.registration_past?
-      icon = "times circle"
+      icon = "user times"
       title = I18n.t('competitions.index.tooltips.registration.closed')
       icon_class = "red"
     elsif competition.registration_full?
-      icon = "exclamation circle"
+      icon = "user"
       title = I18n.t('competitions.index.tooltips.registration.full')
       icon_class = "orange"
     else
-      icon = "check circle"
+      icon = "user plus"
       title = I18n.t('competitions.index.tooltips.registration.open')
       icon_class = "green"
     end
--- a/WcaOnRails/app/helpers/competitions_helper.rb
+++ b/WcaOnRails/app/helpers/competitions_helper.rb
@@ -283,7 +283,7 @@ module CompetitionsHelper
       title = I18n.t('competitions.index.tooltips.registration.closed', days: t('common.days', count: (competition.start_date - Date.today).to_i))
       icon_class = "red"
     elsif competition.registration_full?
-      icon = "user"
+      icon = "user clock"
       title = I18n.t('competitions.index.tooltips.registration.full')
       icon_class = "orange"
     else
--- a/WcaOnRails/app/helpers/competitions_helper.rb
+++ b/WcaOnRails/app/helpers/competitions_helper.rb
@@ -276,19 +276,19 @@ module CompetitionsHelper
 
     if competition.registration_not_yet_opened?
       icon = "clock"
-      title = "Registraion opens in #{(competition.start_date - Date.today).to_i} days"
+      title = I18n.t('competitions.index.tooltips.registration.opens_in', days: t('common.days', count: (Date.today - competition.end_date).to_i))
       icon_class = "blue"
     elsif competition.registration_past?
       icon = "times circle"
-      title = "Registration closed"
+      title = I18n.t('competitions.index.tooltips.registration.closed')
       icon_class = "red"
     elsif competition.registration_full?
       icon = "exclamation circle"
-      title = "Registration full, waiting list open"
+      title = I18n.t('competitions.index.tooltips.registration.full')
       icon_class = "orange"
     else
       icon = "check circle"
-      title = "Registration open"
+      title = I18n.t('competitions.index.tooltips.registration.open')
       icon_class = "green"
     end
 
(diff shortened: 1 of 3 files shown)


FinnIckler:
Hi Simon,
Thanks for picking this up! Looks good to me, but can you add I18n strings to the tooltips so these can be translated?
You can find an example [here](https://github.com/thewca/worldcubeassociation.org/blob/40d4b02fca6c35d995e34d151ac97cd760849ec6/WcaOnRails/app/controllers/users_controller.rb#L65)

simonkellly:
I wasn't sure when initially making this PR if additional steps apart from just adding the english localised strings were needed to support translations, but it appears not so hopefully the above commit should satisfy that

WcaOnRails/config/locales/en.yml
```diff
@@ -1459,6 +1459,12 @@ en:
         search: "Name or city"
         #context: the "recent" button
         recent: "Last %{count} days"
+        #context: the registration status icon
+        registration:
+          opens_in: "Registration will open in %{days}"
+          closed: "Registration has closed"
+          open: "Registraion is open!"
```
FinnIckler, line 1466: There is a typo for Registration here.

WcaOnRails/config/locales/en.yml
```diff
@@ -1459,6 +1459,12 @@ en:
...
+          full: "Competitor limit reached, the waiting list is now in effect"
```
FinnIckler, line 1467: @Jambrose777 Do we have already have some standardized wording for "the waiting list is now in effect" from the Competition Announcement Team?

dunkOnIT:
This is looking great, thanks so much for the quick work on it! 

Two thoughts from me on the design side, which may honestly be completely useless: 
1. **Icons** - I'm really not sure about this, but could the current icons be mistaken for other things? Green tick could look like a competition is completed, red cross could look like it is cancelled. I also don't have suggestions off-hand of what to use instead, so please take this with a large helping of salt.
2. **Hourglass** - I wonder if these icons could just replace the hourglass that gives a tooltip for time until the competition starts? I've personally never used it, and these seem like much more valuable information than time until the comp starts (which can be easily figured out from looking at the date).

simonkellly:
Potentially something like this may work (with or without the hourglass icon)

image

dunkOnIT:
Yeah that looks absolutely perfect! I think I prefer it without hourglass, but will let other WST members weigh in on that, as I wasn't around for the design decision.

simonkellly:
I just pushed the icon changes (without touching the hourglass) so I could have a copy of them in case more changes are needed. 

As a side note, the tests appear to fail on 

``rspec` './spec/requests/media_spec.rb[1:7:1:3:1]' # media DELETE #destroy it should behave like only WCT when signed in as a WCT member can perform action``
``Randomized with seed 47063``

Which I don't think I touched, so I'm not too sure whats going on there

gregorbg:
Hey Simon, thank you for your contribution! It's great to see you were able to get going and set everything up yourself :)
I like the icons and I think we can do without the hourglasses.

Will take a look at your code in a few minutes, just wanted to mention that the test failure you quoted is one of our "unstable" tests. We have 4 or 5 of those in our entire 1000+ test suite, and we haven't yet invested the time to find out why they sometimes fail. Ignore that failure, I'll just re-run the suite here on GitHub until that hiccup calms down

WcaOnRails/app/views/competitions/_index_table.html.erb
```diff
@@ -9,6 +9,7 @@
     <% li_classes << "cancelled" if competition.cancelled? %>
     <li class="<%= li_classes.join(' ') %>">
       <span class="date">
+        <%= registration_status_icon(competition) %>
```
gregorbg, line 12: Could you include this as part of the bigger "if-cascade" below? For example, showing the registration icon doesn't make a lot of sense when the competition is over. So I think you can just replace the last `else` hourglass (the one with the `competitions.index.tooltips.hourglass.starts_in` tooltip) with your new icon. What do you think about that?

WcaOnRails/config/locales/en.yml
gregorbg: Not Jacob but I do think "waiting list in effect" is not a commonly used term. What about "new registrations will go to the waiting list"?

gregorbg (commented):
Thanks, great first PR! Just a minor remark about the placement of the new icon.

WcaOnRails/app/views/competitions/_index_table.html.erb
simonkellly: That makes sense, I have done that in the latest commit

saranshgrover:
Quick UX lens to this - Using a label with icon (see https://semantic-ui.com/elements/label.html) will be much easier for people to decipher what the icons mean and not just have in on a tooltip. i.e have text (Open, Closed, Waitlist and Not Open) along with the icons.

gregorbg:
Using labels opens a can of worms because in different languages the translations can have (wildly) different string lengths. So when introducing labels we'd also have to think about the layout of the table a lot more. I think the icons are reasonably self-explanatory right now (and don't forget they still have tooltips if needs be) and we can consider labels when we redo the competitions list in React.

WcaOnRails/app/helpers/competitions_helper.rb
```diff
@@ -268,4 +268,33 @@ def pdf_name(competition)
...
+    if competition.registration_not_yet_opened?
+      icon = "clock"
+      title = I18n.t('competitions.index.tooltips.registration.opens_in', duration: distance_of_time_in_words_to_now(competition.registration_open))
+      icon_class = "blue"
+    elsif competition.registration_past?
+      icon = "user times"
+      title = I18n.t('competitions.index.tooltips.registration.closed')
+      icon_class = "red"
+    elsif competition.registration_full?
+      icon = "user"
```
gregorbg, line 286: What about using `user clock` here? I'm a bit at odds with the fact that both green and red have a "user plus symbol" icon but orange doesn't.

WcaOnRails/config/locales/en.yml
```diff
@@ -1459,6 +1459,12 @@ en:
         search: "Name or city"
         #context: the "recent" button
         recent: "Last %{count} days"
+        #context: the registration status icon
+        registration:
+          opens_in: "Registration will open in %{duration}"
+          closed: "Registration has closed"
+          open: "Registration is open!"
```
gregorbg, line 1466: What about `Registration is open until %{date}`? I'm debating whether this could be too "risky" in terms of users thinking they can wait a few more days while the registration is _technically_ still open, only to find out that the competitor limit has been reached.

WcaOnRails/config/locales/en.yml
```diff
@@ -1459,6 +1459,12 @@ en:
         search: "Name or city"
         #context: the "recent" button
         recent: "Last %{count} days"
+        #context: the registration status icon
+        registration:
+          opens_in: "Registration will open in %{duration}"
+          closed: "Registration has closed"
```
gregorbg, line 1465: Change to `Registration has closed. Competition starts in %{duration}`?
Tips the hat to the old hourglass which had a "competition starts in %duration" tooltip.

gregorbg (changes requested):
The failing test complains about an old hourglass i18n string that is now unused.
Please remove it, and also consider/discuss the suggestions I made on the other new tooltips.

WcaOnRails/config/locales/en.yml
simonkellly: I would tend to agree with not having the days written in the tooltip. The large variability in how fast competitions fill up  make the effective closing date not just the registration closing time

gregorbg: Good point. We can come back to this discussion later once the current version has been deployed and we have some feedback from the community.

gregorbg (approved):
LGTM :ship:

PR #7056, opened 2022-07-20
Update de translation

AnnikaStein:
This picks up all items listed as "Needs attention" at https://www.worldcubeassociation.org/translations/status for locale de

- Added 41 missing items
- Removed 2 unused items
- Updated 3 outdated items

Additional changes not listed above on the status page that were found after Euros:

- Fix typos (type selector; merchandise box on homepage, ending with a full stop; other minor things).
- Fix grammar (winner sentence on competition homepage). Was introduced to the German translation in 2020 and I believe these keys only play a role for the highlights section. Therefore it should be safe to change the grammatical case to „Dativ“ (oblique case) (was: „Nominativ“ (nominative case) before) because these translated keys are not used elsewhere (correct me if I’m wrong).
- Improve consistency (using „Einzelergebnis“ instead of „Einzelversuch“ / „Einzelzeit“).

All changes have been implemented with Internationalize, taking the most recent en.yml to compare against.

This PR is from 2014STEI03 (it's my second translation from en to de), and parts of it have been discussed already privately. Thanks to @LauraHolzhauer (2016HOLZ01) for pointing out the issue with the winner sentence!

### DIFF
--- a/WcaOnRails/config/locales/de.yml
+++ b/WcaOnRails/config/locales/de.yml
@@ -259,6 +259,49 @@ de:
     333mbf: '10:00.00 Minuten pro Würfel, bis zu 60:00.00'
     #original_hash: da39a3e
     333mbo: ''
+  qualification:
+    #original_hash: a2b25bf
+    type_label: Art der Qualifikation
+    #original_hash: 65c848f
+    result_type: Ergebnisformat
+    #original_hash: 29ef642
+    for_event: 'Qualifikation für %{event}'
+    type:
+      #original_hash: f5c3f76
+      none: Ohne Qualifikation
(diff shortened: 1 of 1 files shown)


gregorbg (approved):

gregorbg:
Thank you for the thorough submission!

For future translations, I can add you to the "Verified Translators" list for German. This will allow you to
1. Submit translation updates easily via https://www.worldcubeassociation.org/translations/status
2. Receive email notifications whenever new strings that need to be translated are introduced.

Would you be interested in becoming part of the translators group?

AnnikaStein:
Sure, that would be a good idea! Thank you.

PR #7060, opened 2022-07-24
Add "Championship Type" and "Exempt from WCA Dues" columns to WFC export

### LINKED ISSUE #7053: Add championship_type to WFC export
WCA Dues policy exempts World and Continental Championships from paying Dues - currently this requires the WFC member issuing Dues to notice this and manually override the Dues amount. If the championship_type column was joined on using competition_id in the WFC export we would be able to automate this in the Dues calculation spreadsheets. 

WST: probably good to mark as "good first issue" or similar, as someone can leverage the updates from #6497

JonEsparaz:
- Added the "Championship Type" column to the WFC export (closes #7053)
- Figured it would be more useful to export a boolean indicating if a competition is exempt from WCA Dues, so I added an "Exempt from WCA Dues" column too

### DIFF
--- a/WcaOnRails/app/models/competition.rb
+++ b/WcaOnRails/app/models/competition.rb
@@ -1719,10 +1719,15 @@ class Competition < ApplicationRecord
     cal
   end
 
-  def exempt_from_wca_dues?
-    world_or_continental_championship = (championships.map(&:championship_type) & Championship::MAJOR_CHAMPIONSHIP_TYPES).any?
-    multi_country_fmc_competition = events.length == 1 && events[0].fewest_moves? && Country::FICTIVE_IDS.include?(countryId)
+  def world_or_continental_championship?
+    championships.map(&:championship_type).any? { |ct| Championship::MAJOR_CHAMPIONSHIP_TYPES.include?(ct) }
+  end
 
-    return world_or_continental_championship || multi_country_fmc_competition
+  def multi_country_fmc_competition?
+    events.length == 1 && events[0].fewest_moves? && Country::FICTIVE_IDS.include?(countryId)
+  end
+
+  def exempt_from_wca_dues?
+    world_or_continental_championship? || multi_country_fmc_competition?
   end
 end
--- a/WcaOnRails/app/models/competition.rb
+++ b/WcaOnRails/app/models/competition.rb
@@ -1718,4 +1718,11 @@ class Competition < ApplicationRecord
     cal.publish
     cal
   end
+
+  def exempt_from_wca_dues?
+    world_or_continental_championship = (championships.map(&:championship_type) & Championship::MAJOR_CHAMPIONSHIP_TYPES).any?
+    multi_country_fmc_competition = events.length == 1 && events[0].fewest_moves? && Country::FICTIVE_IDS.include?(countryId)
+
+    return world_or_continental_championship || multi_country_fmc_competition
+  end
 end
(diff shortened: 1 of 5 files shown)


WcaOnRails/app/models/competition.rb
```diff
@@ -1718,4 +1718,11 @@ def to_ics
     cal.publish
     cal
   end
+
+  def exempt_from_wca_dues?
+    world_or_continental_championship = (championships.map(&:championship_type) & Championship::MAJOR_CHAMPIONSHIP_TYPES).any?
+    multi_country_fmc_competition = events.length == 1 && events[0].fewest_moves? && Country::FICTIVE_IDS.include?(countryId)
```
JonEsparaz, line 1724: (Rails novice).

Accessing these fields here _feels_ weird to me. Perhaps these should be methods on the `Championship` and `Country` models instead?

WcaOnRails/app/models/competition.rb
```diff
@@ -1718,4 +1718,11 @@ def to_ics
     cal.publish
     cal
   end
+
+  def exempt_from_wca_dues?
+    world_or_continental_championship = (championships.map(&:championship_type) & Championship::MAJOR_CHAMPIONSHIP_TYPES).any?
```
gregorbg, line 1723: Mathematically elegant approach, but not the easiest to read.
What about something along the lines of `championships.map(&:championship_type).any? { |ct| Championship::MAJOR_CHAMPIONSHIP_TYPES.include?(ct) }`?

WcaOnRails/app/models/competition.rb
gregorbg: Could you please go into detail what feels weird here?

You're working in the context of the `Competition` model. There's absolutely nothing wrong with having a competition compute whether it is a multi localtion FMC comp.
It might be worth extracting that heuristic into its own separate function, but then it would still stay a part of `competition.rb`.

WcaOnRails/spec/models/competition_spec.rb
```diff
@@ -1098,4 +1098,49 @@ def change_and_check_activities(new_start_date, new_end_date)
     expect(competition).not_to be_valid
     expect(competition.errors.messages[:organizer_ids].first).to match "Need a country"
   end
+
+  describe "exempt_from_wca_dues" do
```
gregorbg, line 1102: We don't do underscores in our spec names usually. These are supposed to be more human-readable. in this case it could be `describe "is exempt from Dues" do`

gregorbg (commented):
Thank you for the contribution, I am especially amazed by the large amount of test coverage. Some open comments regarding specific code style.

WcaOnRails/app/models/competition.rb
JonEsparaz: I wasn't sure if `Championship::MAJOR_CHAMPIONSHIP_TYPES` and `Country::FICTIVE_IDS` are "private" fields per se.

After more thought, I agree that computing at this level (with the changes you suggested) makes sense.

WcaOnRails/app/models/competition.rb
JonEsparaz: Good call. Updated!

WcaOnRails/spec/models/competition_spec.rb
JonEsparaz: Updated. Thank you!

gregorbg (approved):
Thanks for the quick fixes, LGTM!
Fingers crossed the tests pass :D

Issue #3810, opened 2019-01-12
Waiting list public visibility


c-goodyear:
Waiting lists are not displayed publicly for obvious reasons, so organizers and Delegates are contacted to ask how big a waiting list is and what position said competitor is on the waiting list.

An extra field to be included in registration requirements when the limit is reached could display this information. Something like "The registration limit for this competition has been reached. The current waiting list has XX people on it."
During the creation of the competition page the Delegate should be given the choice to select either 'Waiting list is counted by paid' and 'Waiting list is counted by total including unpaid' since some competitions require payment to order the waiting list. This requires some level of implementation of elements of #3451 but doesn't require the automatic registrations in #328 

Additionally, there could be a column appended to the right side of the table on https://www.worldcubeassociation.org/competitions/mine page displaying waiting list position. This would stop members having to contact organizers/Delegates while maintaining the privacy of waiting list order.

Mollerz:
https://github.com/thewca/worldcubeassociation.org/issues/2413 Also covers having a separate waiting list section.

c-goodyear:
For starters, it would lead to more congestion of the information pages, and more confusion especially for first timers registering. It also doesn't serve much of a purpose, the registrations are public because those people will (should) be attending the competition.
It also stops people not cancelling (if they can no longer attend) on purpose if they don't want the waiting list #1 to get a spot.
IMO it's just as irrelevant as wanting removed/deleted registrations to also be publicly visible.

viroulep:
FYI these information are actually "public": except from who have paid what and when: pending and deleted registrations are in the public export of the database; so if you setup the website locally (or just get the developer dump, or just go on staging which may lag behind) then you'll be able to log in as a competition manager and see when everyone submitted their registration (it's not directly the waiting list, but it's reasonable to assume people paid pretty much immediately after registering).

SAuroux:
It should be added that many competition use payment system other than stripe. In such cases the WCA website cannot give accurate information about the current state of a waiting list.

wcajaepark:
[...]

PR #7079, opened 2022-07-31
Display more clearly when registration is full

DanielEgdal:
Per this comment ( https://github.com/thewca/worldcubeassociation.org/issues/3810#issuecomment-1200439737) I tried coming up with a solution

### DIFF
--- a/WcaOnRails/app/models/competition.rb
+++ b/WcaOnRails/app/models/competition.rb
@@ -396,7 +396,7 @@ class Competition < ApplicationRecord
         warnings[:results] = I18n.t('competitions.messages.results_not_posted')
       end
 
-      if self.registration_full? && registration_opened?
+      if self.registration_full? && self.registration_opened?
         warnings[:waiting_list] = I18n.t('registrations.registration_full')
       end
 
--- a/WcaOnRails/app/models/competition.rb
+++ b/WcaOnRails/app/models/competition.rb
@@ -395,6 +395,11 @@ class Competition < ApplicationRecord
       if self.results.any? && !self.results_posted?
         warnings[:results] = I18n.t('competitions.messages.results_not_posted')
       end
+
+      if self.registration_full? && registration_opened?
+        warnings[:waiting_list] = I18n.t('registrations.registration_full')
+      end
+
     else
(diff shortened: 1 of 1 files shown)


WcaOnRails/app/models/competition.rb
```diff
@@ -395,6 +395,11 @@ def warnings_for(user)
       if self.results.any? && !self.results_posted?
         warnings[:results] = I18n.t('competitions.messages.results_not_posted')
       end
+
+      if self.registration_full? && registration_opened?
```
gregorbg, line 399: Nit: `self.registration_opened?`

gregorbg (commented):
Thanks for your contribution! Do you have a screenshot of how your new warning message looks like?

DanielEgdal:
I don't have the website running locally, but the warning will look the same as other warnings displayed on the front page of competitions. Like this one: 
image

gregorbg:
Okay, I'll accept that as a monkey-patch screenshot :D

gregorbg (approved):

DanielEgdal:
Hmm, I just checked this on the website and saw there is an error. I don't quite know how to fix it to be displayed correctly, like it is on the registration page
image
registration page:
image

gregorbg:
Ah, my bad, I forgot to check the actual translation value during review.
You need what's called an "interpolation", or in other words to tell the translation engine which values to insert into the `%{gaps}` in pre-translated strings. To do so, you have to invoke the `I18n.t()` method like this:

```ruby
I18n.t('registrations.registration_full', competitor_limit: self.competitor_limit)
```

Where the key name `competitor_limit` (the first occurence before the colon) stems from the fact that the original string in the translations file is

```yaml
"The competitor limit for this competition is %{competitor_limit} and it has already been reached. If you register, your registration will be placed on a waiting list. It may not be approved unless a spot frees up."
```

KerrieJarman:
The %(competitor_limit) message is still showing - can it be either fixed or rolled back ?

gregorbg:
Hotfixed, might take a minute or two to deploy.

DanielEgdal:
Thanks :)

PR #7120, opened 2022-08-04
Add a per-user registrations notifications setting

### LINKED ISSUE #4673: Notify me when someone registers checkbox errors when less than 28 days before the competition
I (and other organizers) can't change 'Notify me by email when someone registers for this competition' checkbox in competition less that in 28 days.
image

viroulep:
Fixes #4673.


WcaOnRails/app/models/concerns/registrations_notifications.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+require 'active_support/concern'
+
+module RegistrationsNotifications
+  extend ActiveSupport::Concern
+
+  included do
+    # NOTE: it turns out user can be nil in tests.
```
gregorbg, line 9: Wouldn't it make more sense to fix the tests then, instead of being "sloppy" with a nil-safe operator here? How much more additional work would changing the tests entail?

WcaOnRails/app/models/competition.rb
```diff
@@ -1117,7 +1119,11 @@ def has_date_errors?
   end
 
   # The competition must be at least 28 days in advance in order to confirm it. Admins are able to modify the competition despite being less than 28 days in advance.
-  validate :start_date_must_be_28_days_in_advance, if: :confirmed_or_visible?
+  # We only run this validation if we're actually changing the start_date or
+  # confirming the competition, to not prevent organizers/delegates from
+  # updating competition-specific setttings, such as the receive notifications checkbox.
+  validate :start_date_must_be_28_days_in_advance,
+           if: -> { confirmed_or_visible? && (will_save_change_to_start_date? || will_save_change_to_confirmed_at?) }
```
gregorbg, line 1126: Can we move this out into its own method, along the lines of `changes_critical_dates?`?

gregorbg (commented):
Fine by me, just some cosmetic remarks.

WcaOnRails/app/models/concerns/registrations_notifications.rb
viroulep: That's a very good point!

The failing tests in itself are fairly weird because they intentionally introduce an invalid record to make sure save a competition cleans it, but I don't see how one would end up in the db in the first place :s

Anyway I've adjusted the concern to assume user is not nil (which is reasonable in normal website usage), and updated the two failing tests to skip callbacks.

WcaOnRails/app/models/competition.rb
viroulep: To be honest I don't quite see the advantage of that *if that method isn't reused* by some other part of the code; it also feels quite common to me to use a lambda in this circumstance.

If you feel strongly about this I can definitely change that part!

WcaOnRails/app/models/concerns/registrations_notifications.rb
gregorbg: Cool, thanks! 😄

WcaOnRails/app/models/competition.rb
gregorbg: It's not necessarily about code reuse. I do feel rather strongly about this topic simply because I don't want too much to happen within one (or two) lines of code.

The primary purpose of the code sections with the validations should be (imo) to declare _that validations exist_. The specifics of _how_ to validate things, and _when_, is a separate concern from a high-level point of view. This is especially true when the validation has some higher semantic meaning within our business logic (as is the case here: Does the user change critical details of the competition?). Extracting it does not only help with reusing the method (to which I agree that it likely won't happen in this specifc case) but also to give a concise name to the lambda by virtue of declaring it as a full-fledged method, so that other contributors can figure out what's going on at a glance and without reading several lines of comments.

TL;DR: The `if ->` lambda you're constructing here has a meaning, and I believe we should express that meaning by moving it into a function with a name.

(Sidenote: I know that the strict consequence of this would be a huge restructure of the entire `competition.rb` file, but hey, you gotta start somewhere, right? 😸)

viroulep: Change done and PR rebased!

I would agree to outline the function if it wasn't a one-liner or if it was re-used somewhere else, but for one-liners a lambda looks perfectly fine to me (the `start_date_must_be_28_days_in_advance` is actually a very good example of what should be in its own method IMO!).

> but also to give a concise name to the lambda by virtue of declaring it as a full-fledged method, so that other contributors can figure out what's going on at a glance and without reading several lines of comments.

By this logic, what's the point of commenting at all? Comments are meant to be read: the code should be clear on what it does, but why/how this or that needs to be done seems perfectly appropriate for a comment.

> (Sidenote: I know that the strict consequence of this would be a huge restructure of the entire competition.rb file, but hey, you gotta start somewhere, right? smile_cat)

A strict consequence of this would be a huge restructure of the competition.rb file, pretty much all our models, the [rails guide about validations](https://guides.rubyonrails.org/active_record_validations.html#using-a-proc-with-if-and-unless) which explicitly recommends lambda for one-liners, and pretty much all code bases which follow this guide :smiley:

gregorbg: Thanks! I appreciate that this is a subjective discussion with different opinions, and I most certainly don't want to overthrow the Rails guide single-handedly. However, the complexity of what a "one-liner" (quote from that guide) actually means is difficult to define.

> By this logic, what's the point of commenting at all? Comments are meant to be read: the code should be clear on what it does, but why/how this or that needs to be done seems perfectly appropriate for a comment.

I absolutely agree. Which is why in this case, with clear and concise method names, the first half of the comment seems a bit redundant. The idea should be that comments are there to explain stuff that is too complex to "explain" in the code (by virtue of giving good names to variables / functions in the first place) directly.

Also, you opted for the rather simple choice of calling the extracted method `should_validate_start_date?`. That's technically not wrong, but only because right now it is exclusively used as an if-check in the Rails model validation. My intention behind the original request was that this method seems so useful that it could potentially be used in other places in the future, along the lines of `is_chaging_critical_data?`. That way, having it be a standalone method makes a lot more sense and I agree that the current naming scheme feels a bit overkill.

Anyways, I don't want to hold up this feature by more philosophy deliberations. Merging ASAP :D

gregorbg (approved):

WcaOnRails/app/models/competition.rb
viroulep: Oh about the naming: your initial suggestion was regarding *dates*, which is why I didn't keep it in the first place; I did try to find a more describing name but failed to find something real nice (`is_changing_critical_data?` isn't quite good either because what is "critical" is very dependent on the context); maybe something like `is_confirming_competition?` could have been a good candidate?

viroulep:
Thanks @gregorbg !

WcaOnRails/app/models/competition.rb
gregorbg: Yes, that sounds like a good option! Or maybe `is_changing_announcement_dates?`?

gregorbg:
Just as a sidenote: I had to provide a hotfix at https://github.com/thewca/worldcubeassociation.org/commit/5545ffe0c0d65b36bb95de9505f2c28f415fb2f6 because otherwise, the Ruby bindings detected the `tinyint` as `integer` for some reason, and the "Edit preferences" dialog with the form was offering me a number spinner to choose between 0 and 1 on prod :sweat_smile:

viroulep:
What ? :o

I honestly don't understand: obviously I did test this locally, but even now when checking out b88a7b0297e95d9e2ad0bda190a8039e7c9bcc53 locally (with the appropriate migration set up) it does work as expected.
I'm gonna put this on the difference in the mysql version (because this is pretty much the only thing that I don't have in common with the prod setup), but I don't get how it could be happening since - as far as rails is concerned - this *is* a boolean.

Anyway I'm not sure it's worth investigating this behavior, but thanks for the quick troubleshooting!

gregorbg:
I was also astonished (precisely because I did assume you tried it locally) but to be honest the only thing I can say here (regarding prod) is "I saw it with my own eyes!" :see_no_evil:

PR #7122, opened 2022-08-04
Fix a spring loader error

viroulep:
When developing locally we use spring (by default I think), which prevents reloading the whole app when only a single file is changed.
Unfortunately as soon as I changed a ruby file and refreshed a page I ran into the following error:
>superclass mismatch for class RankingCondition

It actually makes coding in the ruby part a real pain; I actually don't get how no one else ran into this or complained about this, so maybe this is something wrong with my setup?

The fix is very easy though, we just need to put each class in its own file and the loader does its job properly.


gregorbg:
We've encountered this error very sporadically, and were never able to reproduce.
The occurences were so rare that we didn't bother digging any deeper. Just out of curiosity, how did you come up with the fix?

In terms of an actual code review: Would you mind putting the subclasses into separate folders, similar to the `results_validators`?

gregorbg:
Also, CC @timreyn Just so you're aware this happened 😄

timreyn:
Thanks @viroulep !!

WcaOnRails/lib/advancement_conditions/advancement_condition.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+module AdvancementConditions
+  class AdvancementCondition
+    include ActiveModel::Validations
+
+    attr_accessor :level
+    validates :level, numericality: { only_integer: true }
+
+    @@advancement_conditions = [AttemptResultCondition, PercentCondition, RankingCondition].freeze
```
viroulep, line 10: This is a change compared to the initial file, where `AdvancementCondition.subclasses` was used, but when lazy loading is used by spring the file containing the subclasses are not loaded at this point.
In any case it's more explicit to state exactly which advancement conditions we want available.

gregorbg (approved):
LGTM, waiting to see whether Tim has any thoughts. Otherwise I'll proceed to merge this in the next 48 hours.

viroulep:
> We've encountered this error very sporadically, and were never able to reproduce.
The occurences were so rare that we didn't bother digging any deeper. Just out of curiosity, how did you come up with the fix?

Oh wow I can actually trigger this every time (which is why it was so annoying): just visit a page (eg: a "show competition" page), change something in the relevant controller, and it will trigger.

When I took a look at the ranking condition file it was fairly obvious what could go wrong: IIRC the spring loader relies a lot on the fact that classes matches the filename, so it was likely the issue.

> In terms of an actual code review: Would you mind putting the subclasses into separate folders, similar to the results_validators?

I can, but then I need to put them in a module matching the folder name (I did move everything in a separate folder, but I also move the base class as it would be weird if it wasn't in the same module as its inherited classes).

gregorbg:
> I can, but then I need to put them in a module matching the folder name (I did move everything in a separate folder, but I also move the base class as it would be weird if it wasn't in the same module as its inherited classes).

I don't see what's wrong with having these code files live in their own module...?

viroulep:
Absolutely nothing, I'm just stating the necessary changes for them to be moved to a subfolder ^^

PR #330, opened 2016-01-14
Removed wca_table_for from my_comps and changed the list, as per the …

pedrosino:
…discussion with delegates.

PR #7126, opened 2022-08-06
Fixes #7068: Past Competitions obstructed by floating header bar

### LINKED ISSUE #7068: Under "My competitions" some of the Past Competitions are obstructed
**Describe the bug**

The  headers  `Bookmarked competitions` table stays in place when `Past competitions` is expanded, thus covering up some of the rows in the `Past competitions` table.

**To Reproduce**
1. Bookmark any upcoming competition
2. Go to `/competitions/mine` (as long as you are a user with past competitions)
3.  Click `Past competitions` to expand that table

**Expected behavior**
I should be able to see all my past competitions and the header for `Bookmarked competitions` should move down the page along with the rest of the elements.

**Screenshots**
image

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser: Brave
 - Version: 1.41 (Chromium version 103)

**Smartphone (please complete the following information):**
 - Device: Pixel 4xl
 - OS: Android 13
 - Browser: Chrome
 - Version: 103
 
**Additional context**
Add any other context (for instance the language you're using the site in) about the problem here.

revers3ntropy:
Just removed the unnecessary CSS class from the table head which triggered some JQuery to try to keep it in the right place which wasn't working properly for whatever reason.


gregorbg:
Thanks! The prize question is: Is that CSS class actually necessary? I understand that it caused the bug, but I'd like to think that it was there for some other reason.

Can you check in the git blame who added this in the first place? The original commit message might have some indication as to what's going on. I'm not comfortable merging this PR "just like that", knowing that we might break some other part of the results tables that we didn't intend to change.

revers3ntropy:
So it was changed from 'floatThread' to 'floatThead' in a commit with the message 'Fixed Jeremy's comments' in 2016 (https://github.com/thewca/worldcubeassociation.org/commit/0642420d2f44d6cc4a5cb49499cc83856073326c), where I believe floatThread it a typo as it doesn't appear anywhere else in the repo.

floatTHead makes the header float when the table is scrolled, but I the table doesn't have a maximum height, so there is never any scrolling happening anyway. I imagine it was added to all tables at some point in case they overflowed on the y maybe?

gregorbg:
Thanks for the detective work! If you look at the commit hash you mentioned here on GitHub, you can figure out that it was merged as part of PR #330. Looking at that PR, there's one commit where (barring the typo) `floatThead` was _explicitly_ added in https://github.com/thewca/worldcubeassociation.org/pull/330/commits/e0ae973d0f93c0e349fe11290a5eab7ab32cf17b.

I cannot help but to keep thinking this must have some purpose. Can you play around with the dev tools in any browser of your choice to simulate different screen sizes and landscape mode? If you tried that and you're still fully confident that this can be removed, I'll proceed with the merge :v:

revers3ntropy:
Ok, I've tried it for mobile and desktop in Chrome, Firefox and even a really old version of Safari and it all seems to be working.
I only managed to test the case with no bookmarked comps in Chrome, as it keeps on throwing this error when I view a competition (??):
'superclass mismatch for class RankingCondition'
Which seems to be unrelated though as it also appears when using master.

revers3ntropy:
(which means I can't un-bookmark or bookmark comps as far as I can see)

gregorbg:
The `RankingCondition` bug has been fixed, please pull from master and rebase your feature on top.

revers3ntropy:
Great thanks will do

gregorbg:
I'll just merge this and if anybody complains we call roll back the changes.

revers3ntropy:
Thanks. Also now tested in Chrome & Firefox with 0 and 1 bookmarked events and everything seems fine...

PR #7131, opened 2022-08-08
Bump i18n-js library to v4

gregorbg:
A lot has changed with this new major release. Most conveniently, we can have the lib "watch itself" and we don't need the additional rake task anymore. But the watcher only runs in development mode so the deploy script has to be changed as well.

Also, we don't need to provide a pack per locale anymore. I18n-js can import its own JSON files as needed.


gregorbg:
@viroulep I think you created a lot of the initial "wiring" behind the JS-Webpacker-React architecture so I'm kindly asking for your review :) I'd like to get this merged before the end of the month (when the next batch of Dependabot updates comes in) so I'd appreciate if you could take a moment to look this over. If you don't have the chance, I'll probably merge this around August 30.

viroulep:
Will take a look!

WcaOnRails/config/i18n.yml
```diff
@@ -25,21 +25,20 @@
...
 translations:
-<% I18n.available_locales.each do |l| %>
-  - file: "app/webpacker/packs/generated_locales/<%= l %>.js"
-    only: [<%= export_keys.map { |k|  "\"#{l}.#{k}\"" }.join(",") %>]
-    prefix: |
-      // This file is generated automatically by the rake middleware!
-      // DO NOT EDIT IT MANUALLY!
-      import I18n from '../../lib/i18n';
-<% end %>
+  # Using the `:locale` template in the filename here automagically makes i18n-js iterate over the available languages.
```
viroulep, line 32: It does look far better than the trick we had to use before!

WcaOnRails/config/initializers/i18n_js.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+# copied from https://github.com/fnando/i18n-js/blob/main/MIGRATING_FROM_V3_TO_V4.md
+Rails.application.config.after_initialize do
+  require "i18n-js/listen"
+  # This will only run in development.
+  I18nJS.listen
```
viroulep, line 7: Although the migration guide does mention to put that file here, I believe it would actually be executed in production (which the migration guide also strongly suggest not to do)!

Since we export the locales for production and tests, I'd suggest to either:
  - make sure this runs only in development (see the suggestion below)
  - move this part to `config/environments/development.rb`

```suggestion
  I18nJS.listen if Rails.env.development?
```

Since I'm not sure if I18nJS actually expects the configuration to be fully initialized, I believe my first suggestion should be the best/safest route!

viroulep (changes requested):
Thanks a lot @gregorbg!

I went through the migrating guide: as far as can understand it, and based on what we need, it all looks pretty good to me!

I did not check that the locales actually still work in js, but I obviously assumed that you tested this locally :)

I've not approved this right away because I just have a concern about when one of the initializers is ran, but this should be fairly straightforward to check/fix :)

WcaOnRails/config/initializers/i18n_js.rb
gregorbg: Yes, I agree. Although based on what the migration guide says, I'd expect them to run a similar "is Rails running in development mode?" check within the `listen` method that I'm calling here. But either way, your suggestion is the super safe route :)

gregorbg:
Thanks for the review! Yes, I did test that translations get displayed locally. Hopefully it will also work with the different webpack configuration on prod 🙃

WcaOnRails/config/initializers/i18n_js.rb
viroulep: You're right actually, they [do check for that](https://github.com/fnando/i18n-js/blob/1e04f527eb683b7200db1fcd3cdbcdbba645575b/lib/i18n-js/listen.rb#L12)!
Weird to be doing that when the lib is not rails-specific but alright ^^

viroulep (approved):
Cool, the PR LGTM, it might be worth trying it out on staging first just in case :)

gregorbg:
Staging works! Proceeding with the merge.

PR #6377, opened 2021-07-20
Add more options other than "spectators can attend for free"

notrichardpeng:
Addresses #6321 

Right now the idea is to have a selection menu pop up in the competition form when the entry fee for guests is set to 0.

But in Javascript I'm breaking the design pattern currently, because `.on('change')` to the element `name=competition[guests_entry_fee_lowest_denomination]` can only be detected when the page is loaded up initially. So, I had to use the other input field of the whole currency input div, that didn't have a `name` attribute before I added it myself. However I couldn't figure out how to add the `competition[]` bracket around `guests_entry_fee_lowest_denomination` around its name to keep things consistent.

PR #7140, opened 2022-08-13
Optionally enforce maximum number of guests per competitor on registrations

### LINKED ISSUE #7059: Add a "Maximum Number of Guests Allowed Per Competitor" field on competition websites
When organizers want to impose a guest limit, they currently write it as a sentence in the _Extra registration requirements_ and/or the _Information_ section. As a result, this information is easy to miss by registrants who don't fully read all the information about a competition. Registrants are able to enter a number of guests beyond the limit and still get their registration accepted, giving them the false idea that they can bring the number of guests they indicated. This can be very frustrating for them, especially if they find out about the guests limit after they have already paid for travel and lodging, or worse, right at the beginning of the competition.

We should prevent registrations from getting submitted if the number of guests is over the guest limit. To do that, I believe we need a field on the *Organizer View* page for organizers to enter a non-negative integer, then use that as the upper bound for the number of guests field on the *Registration Details* page.

**Additional context**
Although many regions are running competitions just as they did before the pandemic, some regions are still enforcing COVID-related policies, including mask mandates and guests limits. This feature would make running competitions with guest limits smoother as the organizers don't have to deal with as many angry emails and in-person conversations from competitors who don't fully read the competition information.

JonEsparaz:
Allows organizers to optionally enforce a per-competitor guest limit on registrations. As mentioned in #7059 this is useful for areas still taking precautions against COVID-19 (or if they return to other areas). Outside of pandemic times, this will still be useful for competitions with limited venue space and/or championships that need to limit traffic from non-competitors.

First commit just adds `guests_per_registration_limit` column and renames `free_guest_entry_status` to `guest_entry_status`.

Second commit has all the logic changes:
- If you select "Spectators are only permitted as companions of competitors", there will be a field that to (optionally) specific a maximum number of guests per registration.
- If the above field is used, we validate that the value is an integer >= 1 and <= 100. We also validate that the competition asks about guests if there is a restriction on the number of guests.
- The registration requirements will display the limit if applicable:

image

- Registrations with too many guests are blocked:

image

<details>
  <summary>Pre-refactor description</summary>
  Changes (the outermost bullets roughly align with the changes by commit):

  - Add three columns: `guests_per_registration_limit_enabled`, `guests_per_registration_limit`, `guests_per_registration_limit_reason`, which mirror the `competitor_limit*` columns. I intentionally made `guests_per_registration_limit_enabled` separate from `guests_enabled` since you can ask about guests without enforcing a guest limit and I felt that determining if the guest limit is "enabled" via a null-check on `guests_per_registration_limit` would be confusing for users and make the code more difficult to understand.
  - Make the following validations on the guest limit:
    - If you are enforcing a guest limit, you must also ask about guests.
    - The guest limit must be an integer.
    - The guest limit must  be >= 1 and <= 100 (I picked 100 somewhat arbitrarily, but I doubt any cubers have that many friends).
    - A reason for the guest limit must be provided to the WCAT.
  - Add a `guests_per_registration_limit_enabled` checkbox, `guests_per_registration_limit` input and `guests_per_registration_limit_reason` input to the competition form.
  - Add  `Competitors may bring at most N guest` text to registration requirements and validate registrations, _blocking anything that exceeds an enabled guest limit_.
  - Add guest limit info in the WCAT email.
  - Tests, plus fix handle when `competition` can be `nil` within `Registration`. 
</details>

Closes #7059.


gregorbg:
Thanks for your Pull Request! It's great to see that you're making more complex contributions now :)

However, in this specific case I'm not too wild about adding yet another three fields to the competitions table as if it wasn't already crammed 🙈 Do you have a basic idea on how widely used this feature would be? Covid will be over eventually...

Also, this seems related to https://github.com/thewca/worldcubeassociation.org/pull/6377. For example, it wouldn't make sense to set the Free Guest status to "Anyone can attend" but then limit the number of guests per competitor.

I hope you don't feel overwhelmed by this, but could you consider re-working your ideas and the code from that other PR by combining them into a sort of a "guest management" feature? Essentially, change the current "guest policy if entry is free" logic to be independent of the the cost, and always pick a policy. Then optionally also specify a limit if "only guests of competitors can attend" is picked.
(That would also alleviate my concern about adding more fields to the DB, because you could probably combine and reuse a few of the existing fields)

JonEsparaz:
Thanks for the suggestions @gregorbg!

> Do you have a basic idea on how widely used this feature would be? Covid will be over eventually...

I think it would be used more than a feature like qualification requirements (although, admittedly less useful).

Long-term (when the pandemic is over), this will still be useful for areas with a high demand for competitions like Ontario (my area) and probably others like California. There's pressure to increase competitor limits to meet demand. For example, when we previously would've had 100 competitors in a room with a 250 person capacity (physical capacity, not pandemic-induced limited capacity), we'll push it to 125, but that requires having at most one guest per competitor (125 * 2 = 250).

> I hope you don't feel overwhelmed by this, but could you consider re-working your ideas and the code from that other PR by combining them into a sort of a "guest management" feature? Essentially, change the current "guest policy if entry is free" logic to be independent of the cost, and always pick a policy. Then optionally also specify a limit if the "only guests of competitors can attend" policy is picked.
> (That would also alleviate my concern about adding more fields to the DB, because you could probably combine and reuse a few of the existing fields)

Definitely, can do! I won't get to this until the weekend, so I'll run the implementation by you first:
- Add `3: limited` to the `free_guest_entry_status` enum, which corresponds to "Each competitor can bring a maximum of N guests"
- Just add two fields: `guests_per_registration_limit` and `guests_per_registration_limit_reason` (we don't need to add the boolean field)
  - If the WCAT doesn't need written explanation, we can drop `guests_per_registration_limit_reason` too

gregorbg:
> For example, when we previously would've had 100 competitors in a room with a 250 person capacity (physical capacity, not pandemic-induced limited capacity), we'll push it to 125, but that requires having at most one guest per competitor (125 * 2 = 250).

That makes a lot of sense, I have a better idea now of how useful this feature will be. Thanks!

> I'll run the implementation by you first

The direction you're heading is great. A few more thoughts to refine your ideas:
- Rename the existing columns and attributes (NB a migration will be required to rename the columns!) to drop the `free_` part. In theory, it should be possible to charge a guest entry fee _and_ specify that either "Anyone can attend" or "Only attend as companions of competitors".
- Is a separate option like `3: limited` really necessary? Wouldn't it suffice to alter the competition form such that if `2: restricted` is picked, you can also optionally choose a limit per competitor? See `assets/javascripts/competitions.js` to figure out how you can conditionally make inputs appear/disappear based on form states.
- Could you reach out to WCAT via mail (with WST in CC) and clarify whether they would need a written explanation?

gregorbg:
Also, CC @notrichardpeng since you worked on the previous feature.
If you have ideas about the suggested implementation in this thread, feel free to bring them up 😄

JonEsparaz:
Thanks @gregorbg! I just reached out to the WCAT. I'll look to get your feedback implemented in the next couple days.

JonEsparaz:
I haven't forgotten about this, but I'm back to university this week... Once I hear back from the WCAT on their functional requirements, I'll find a couple hours to wrap this up.

(Also, I can see there's more pressing PRs and changes coming to the registration system, so no worries if this falls lower on the priority queue.)

gregorbg:
Hey @JonEsparaz! What's the progress on this issue? There are a couple of merge conflicts to be resolved at this point, but I think we can proceed based on what WCAT has answered so far.

JonEsparaz:
Sorry, @gregorbg. Been busy with school and other cubing responsibilities. Next opportunity I'll have to wrap this up is week of November 7.

gregorbg:
Okay, thanks for letting me know. And good luck with whatever is keeping you busy! :smile:

WcaOnRails/app/assets/javascripts/competitions.js
```diff
@@ -52,6 +52,10 @@ onPage('competitions#edit, competitions#update, competitions#admin_edit, competi
     $('.adjacent-competitions .wca-adjacent-competitions-hide-events').show();
   });
 
+  $('select[name="competition[guest_entry_status]"]').on('change', function() {
+    $('.competition_guests_per_registration_limit').toggle(this.value === 'restricted');
```
JonEsparaz, line 56: I don't think there's a way to access the enum from here, but please let me know if I'm missing something @gregorbg.

JonEsparaz:
@gregorbg should be good to go! Thanks for your patience with this PR. Please let me know if you have any feedback.

WcaOnRails/app/assets/javascripts/competitions.js
gregorbg: I think in theory you could move the entire thing over to the `html.erb` world and inject a Ruby `<%= %>` template there. But it's really not necessary since we're looking to migrate the entire form to React soon anyways.

WcaOnRails/app/models/competition.rb
```diff
@@ -79,7 +79,7 @@ class Competition < ApplicationRecord
   scope :confirmed, -> { where.not(confirmed_at: nil) }
   scope :not_confirmed, -> { where(confirmed_at: nil) }
 
-  enum free_guest_entry_status: {
+  enum guest_entry_status: {
     unclear: 0,
     anyone: 1,
```
gregorbg, line 84: Rename this to `free` now instead?

WcaOnRails/app/models/registration.rb
```diff
@@ -25,13 +25,23 @@ class Registration < ApplicationRecord
...
       errors.add(:registration_competition_events, I18n.t('registrations.errors.cannot_be_deleted_and_accepted'))
     end
   end
 
+  def guest_limit
+    competition.guests_per_registration_limit
+  end
+
+  def check_guest_limit?
+    !!competition&.guests_per_registration_limit_enabled?
```
gregorbg, line 42: Two observations:
1. `!!` double negation is a bastard child of JS. Please use Ruby idioms like `my_value.present?` instead.
2. You assume `competition` might be nil by virtue of the nil-safe `&.` operator. In the `guest_limit` method above, you do not make that same assumption. Why?

WcaOnRails/spec/features/competition_management_spec.rb
```diff
@@ -118,15 +118,15 @@
     end
 
     scenario "select anyone free guest entry status" do
-      competition = FactoryBot.create(:competition, :with_delegate, id: "OldId2016", free_guest_entry_status: 1)
+      competition = FactoryBot.create(:competition, :with_delegate, id: "OldId2016", guest_entry_status: 1)
```
gregorbg, line 121: Access the enum `Competition.guest_entry_statuses['restricted']` like the way you do in the factory method?

WcaOnRails/spec/features/competition_management_spec.rb
```diff
@@ -118,15 +118,15 @@
...
       visit competition_path(competition)
       find('div', id: 'show_registration_requirements').click_link('here')
 
       expect(page).to have_text("Any spectator can attend for free.")
     end
 
     scenario "select restricted free guest entry status" do
-      competition = FactoryBot.create(:competition, :with_delegate, id: "OldId2016", free_guest_entry_status: 2)
+      competition = FactoryBot.create(:competition, :with_delegate, id: "OldId2016", guest_entry_status: 2)
```
gregorbg, line 129: See above.

WcaOnRails/spec/models/competition_spec.rb
```diff
@@ -75,11 +75,11 @@
     competition = FactoryBot.create :competition
 
     # According to the property's enum definition, 1 means anyone and 2 means restricted
-    competition.free_guest_entry_status = 1
+    competition.guest_entry_status = 1
```
gregorbg, line 78: See above.

WcaOnRails/spec/models/competition_spec.rb
```diff
@@ -75,11 +75,11 @@
...
     expect(competition.all_guests_allowed?).to be true
     expect(competition.some_guests_allowed?).to be false
 
-    competition.free_guest_entry_status = 2
+    competition.guest_entry_status = 2
```
gregorbg, line 82: See above.

gregorbg (changes requested):
Feature-wise we're fine now! A little bit of cosmetics here and there, plus I'm afraid I have to ask you for another rebase because of merge conflicts with our migration file :see_no_evil:

WcaOnRails/app/models/registration.rb
JonEsparaz: 1. Will do.
2. This was for a test case where the competition DNE. `guest_limit` is accessed after the `check_guest_limit?` conditional passes. The more I think about it, the test case is nonsensical (why would we have a registration for a competition that DNE?). Will revise.

WcaOnRails/spec/features/competition_management_spec.rb
JonEsparaz: Was trying to minimize the changeset for `free_guest_entry_status` -> `guest_entry_status`. Will do in a separate commit!

WcaOnRails/app/models/competition.rb
JonEsparaz: Done.

WcaOnRails/spec/models/competition_spec.rb
JonEsparaz: Thanks. All four changed.

WcaOnRails/app/models/registration.rb
JonEsparaz: Re 2: I removed my (nonsensical) tests, but a check on `competition` is still needed to satisfy these existing tests:

https://github.com/thewca/worldcubeassociation.org/blob/947b0a65e1b576472c10d2c5bf38a73a281a01bb/WcaOnRails/spec/models/registration_spec.rb#L12-L22

Will use `competition.present?`

WcaOnRails/app/views/competitions/_registration_requirements.html.erb
```diff
@@ -119,13 +119,18 @@
   <br>
 <% else %>
   <% if @competition.all_guests_allowed? %>
-    <%= t('competitions.competition_info.guests_free.anyone') %>
+    <%= t('competitions.competition_info.guests_free.free') %>
```
JonEsparaz, line 122: The "free free" in `guests_free.free` is a little awkward. I can change to something else if you have a good suggestion @gregorbg (trying to keep the number of line changes as low as possible for now).

JonEsparaz:
Comments addressed and rebase complete. Thanks for the review, @gregorbg!

WcaOnRails/app/views/competitions/_registration_requirements.html.erb
gregorbg: It's only a translation string index/key. No need to make this matter even more complicated :)

gregorbg (approved):
LGTM :ship:

PR #7161, opened 2022-08-22
[Migrate React] Static pages: Disclaimer

revers3ntropy:
Migrated the Disclaimer page to use React


WcaOnRails/app/views/static_pages/disclaimer.html.erb
```diff
@@ -3,11 +3,5 @@
 <div class="container">
   <h1><%= yield(:title) %></h1>
```
gregorbg, line 4: This `h1` can also be moved into React. If you need to access the title, you can pass it to the React props via the second argument (the one that's currently an empty object.)

WcaOnRails/app/views/static_pages/disclaimer.html.erb
```diff
@@ -3,11 +3,5 @@
 <div class="container">
```
gregorbg, line 3: The `react_component` call automatically creates a `div` that houses the to-be-rendered component. Its HTML properties can be controlled by passing a second additional parameter to the method like so:
```ruby
react_component('MyComponent', {
  'someProp': true,
  'otherProp': 123
}, {
  id: 'html-id',
  class: 'container'
})
```

WcaOnRails/app/webpacker/components/StaticPages/Disclaimer/index.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+
+import { List } from 'semantic-ui-react';
+
+function Disclaimer() {
+  return (
+    <List as="ul">
+      <List.Item as="li">
```
gregorbg, line 8: What happens if you omit the `as=ul` and `as=li`?

WcaOnRails/app/webpacker/components/StaticPages/Disclaimer/index.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
```
gregorbg, line 1: Comment about the filename (not sure where to put this, so I'm arbitrarily picking line 1)

If it's only a single component that has no nested components or shared logic, you don't have to create an entire new folder. `.../components/StaticPages/Disclaimer.js` will do.

WcaOnRails/app/webpacker/components/StaticPages/Disclaimer/index.js
```diff
@@ -0,0 +1,33 @@
...
+        Unless otherwise noted, sponsorship and partnership agreements are arranged with
+        individual competition organizers and not with the WCA.
+      </List.Item>
+      <List.Item as="li">
+        Since the WCA itself does not organize competitions, the WCA is not responsible for
+        damage or injury incurred at WCA Competitions.
+      </List.Item>
+      <List.Item as="li">
+        It is not permitted to use the WCA&apos;s non-profit status determination letter
+        without explicit permission from the WCA Board of Directors.
```
gregorbg, line 27: Please add a code comment that we're deliberately _not_ handling these strings as part of the usual localization / i18n process, as we need the exact English wording for legal reasons.

gregorbg (changes requested):
Great first attempt! With a little fine-tuning, we can merge this soon :)

WcaOnRails/app/views/static_pages/disclaimer.html.erb
gregorbg: Although arguably, you could also hard-code "Disclaimer" in this case, because we're going to keep this page in English anyways and not localize it for now (see also my other comment below)

WcaOnRails/app/views/static_pages/disclaimer.html.erb
revers3ntropy: Ah so no need for this div then

WcaOnRails/app/webpacker/components/StaticPages/Disclaimer/index.js
revers3ntropy: You lose the bullet points, I just followed the docs for created a bullet point list...

WcaOnRails/app/webpacker/components/StaticPages/Disclaimer/index.js
revers3ntropy: Ok yes, so I was just thinking that some of the static pages would have more than one file, so for the sake of sameness put all of the pages in their own folder but that's fine.

WcaOnRails/app/webpacker/components/StaticPages/Disclaimer/index.js
revers3ntropy: Ah ok sounds good

gregorbg (approved):

PR #7162, opened 2022-08-23
Switching HTTPS Termination to the AWS Load Balancer

FinnIckler:
This change needs a redeployment of the Server!

Changes done:
- Removing the acme.sh cron jobs to renew the certificate
- Removing the Certificate Check from the server controller
- Changing the base.rb Chef files to not create the https nginx config
- Changing the nginx config to server traffic on port 80
- Changing the servers.sh to
  - SSH into the already existing server via elastic ip not hostname (as the hostname will be attached to the load balancer)
  - Doing the health check using the load balancer internal health check
  - Register the targets with the load balancer in the passthetorch phase
  - Deregistering the targets in the reap phase
 - Changing bootstrap.sh to not sync the certificates

This also includes infrastructure changes
- Creating a Load Balancer with a *.worldcubeassociation.org certificate 
- Creating a target group for staging and production

While I have tested the code/infrastructure changes on staging, I have not made the necessary DNS changes for production.
So my recommendation is, after this is reviewed, we start a new staging and then prod server with the changes, but we do not reap the server. Instead, we perform a DNS fallover to the loadbalancer after checking everything works as intended to achieve a no downtime migration.

PS: Sorry for all the commits, I had to rebase a commit out of there


gregorbg:
The commit history is a hot mess. Please fix this before proceeding with the review.

FinnIckler:
I am just going to close and reopen this one

PR #4799, opened 2019-11-05
Change registration user validations to run on creation only

jonatanklosko:
Context: if a competitor registers and then gets banned, the registration is left in a permanent invalid state. One of the consequences is that saving WCIF fails due to the validation errors. Thinking about it, I believe it's better to run the user-related validations only when creating registrations.

If someone has a better solution in mind (specifically when it comes to WCIF update), please let me know!

PR #7165, opened 2022-08-26
Upgrade to Rails 7

gregorbg:
~Three~ Two notable changes:
- ~`starburst` is incompatible and the project looks abandoned. I have commented out the lines for now so that we can choose to re-enable later.~
- `devise_two_factor` had a breaking change which is pretty significant: They changed their entire encryption backend from handmade procedures to the general `ActiveStorage` encryption introduced in Rails 7. Subsequently, we have to enable encrypted attributes as well. Migration path follows their migration guide at https://github.com/tinfoil/devise-two-factor/blob/main/UPGRADING.md
- We were still running on Rails 5 default settings. I decided to switch to 7.0, which most notably includes a default `presence` validation on all `belongs_to` associations. In other words, you have to explicitly specify when a `belongs_to` is NOT required anymore. As such, I was able to drop a lot of validators by virtue of the fact that they're now "baked in" to the `belongs_to`. The error messages change from `can't be blank` to `must exist`.

Other than that, just your usual Rails upgrade migrations and config changes.


WcaOnRails/app/models/registration.rb
```diff
@@ -8,9 +8,9 @@ class Registration < ApplicationRecord
   scope :with_payments, -> { joins(:registration_payments).distinct }
 
   belongs_to :competition
-  belongs_to :user
-  belongs_to :accepted_user, foreign_key: "accepted_by", class_name: "User"
-  belongs_to :deleted_user, foreign_key: "deleted_by", class_name: "User"
+  belongs_to :user, optional: true # A user may be deleted later. We only enforce validation directly on creation further down below.
```
viroulep, line 11: Can it? :o

As far as I remember we've always had a policy to never destroy users (but instead anonymize them if needed).
I wouldn't be surprised if some parts of the code just assume the user will always be there.

gregorbg: We have a test which suggests that it is possible to have a reg without user:
https://github.com/thewca/worldcubeassociation.org/blob/master/WcaOnRails/spec/models/registration_spec.rb#L24

viroulep: Wow that is weird :s

From the dev dump:
```
irb(main):001:0> Registration.where(user_id: nil).size
   (349.1ms)  SELECT COUNT(*) FROM `registrations` WHERE `registrations`.`user_id` IS NULL
=> 0 
```

Just pointing out a single use of user which would just break the whole registrations page if it wasn't there: https://github.com/thewca/worldcubeassociation.org/blob/master/WcaOnRails/app/views/registrations/edit_registrations.html.erb#L109 (there are probably others).

I'd argue we definitely want it there, but it was introduced by #4799, which makes me think there is more than we can see (and that running the same query on prod would not return 0 :p).
We should probably check what happens when a user is banned (which is hopefully not too often), but I'd bet a reasonable amount of money that something would break ^^

viroulep:
Just a general comment as I see the PR history is kept "as-is" quite often: I'd definitely recommend PR authors to squash their commits into meaningful ones when it's ready to merge: it makes the history easier to read, and make git-bisecting easier because you have less chances to run into a commit which doesn't represent something "not finished" (and therefore not meaningful when bisecting).

gregorbg:
> Just a general comment as I see the PR history is kept "as-is" quite often: I'd definitely recommend PR authors to squash their commits into meaningful ones when it's ready to merge: it makes the history easier to read, and make git-bisecting easier because you have less chances to run into a commit which doesn't represent something "not finished" (and therefore not meaningful when bisecting).

Thanks for the suggestion! We actually squash upon merging when there's too many "as-is" commits.

WcaOnRails/app/models/registration.rb
gregorbg: I'll be the first in line to remove that weird "verify existence only on create" check.
For the sake of this PR, I only considered the "status quo" and I didn't bother digging into the historic details of this design decision.

gregorbg: With warm regards from our production machine:
image

PR #7167, opened 2022-08-30
Migrated logo page to React.

revers3ntropy:
- Created React component
- Added required translations to config
- Replaced /logo page content with react_component call


WcaOnRails/app/webpacker/components/StaticPages/Logo.js
```diff
@@ -0,0 +1,25 @@
...
+function Logo({ title }) {
+  return (
+    <>
+      <h1>{title}</h1>
+      {I18n.t('logo.paragraphs').map((paragraph, index) => paragraph && (
+        <p
+          // eslint-disable-next-line react/no-array-index-key
+          key={`logo-paragraphs-${index}`}
+          // eslint-disable-next-line react/no-danger
+          dangerouslySetInnerHTML={{ __html: sanitize(paragraph) }}
```
gregorbg, line 18: We have a component called `I18nHTMLTranslate` which is made precisely to avoid manual inner HTML overrides like this. Something along the lines of
```jsx
{I18n.t('logo.paragraphs').map((paragraph, index) => paragraph && (
    <I18nHTMLTranslate
      // eslint-disable-next-line react/no-array-index-key
      key={`logo-paragraphs-${index}`}
      i18nKey={'logo.paragraphs-${index+1}'}
    />
))}
```

should do the trick.

gregorbg (changes requested):

WcaOnRails/app/webpacker/components/StaticPages/Logo.js
revers3ntropy: Ahh ok thanks, I didn't realise you could index in the i18n key like that.

WcaOnRails/app/webpacker/components/StaticPages/Logo.js
```diff
@@ -0,0 +1,25 @@
...
+ */
+function Logo({ title }) {
+  return (
+    <div className="wca-logo-information">
+      <h1>{title}</h1>
+      {I18n.t('logo.paragraphs').map((_, index) => (
+        <I18nHTMLTranslate
+          // eslint-disable-next-line react/no-array-index-key
+          key={`logo-paragraphs-${index}`}
+          i18nKey={`logo.paragraphs.${index}`}
```
gregorbg, line 18: In our translation YML, we index paragraphs starting with `1`. I am 99.9999999999% sure that the `.map((_, index))` invocation starts counting `index` at `0`. Have you tried this out locally to make sure that you're not getting any off-by-one errors?

WcaOnRails/app/webpacker/stylesheets/static_pages/logo.scss
```diff
@@ -0,0 +1,7 @@
+.wca-logo-information {
+  div {
+    /* Apply same styles as to 'p' tag, as tag is not allowed in context */
+    line-height: 1.4285em;
+    margin: 0 0 1em;
+  }
+}
```
gregorbg, line 7: I'm not a big friend of "magic numbers" anywhere in the code, particularly not in CSS.

What about wrapping the `div` returned by `I18nHtmlTranslate` inside a `p` tag to get this styling "for free"? Something along the lines of
```jsx
<p>
  <I18nHtmlTranslate ...foo... />
</p>
```

gregorbg (changes requested):

WcaOnRails/app/webpacker/stylesheets/static_pages/logo.scss
revers3ntropy: I did try this, but it throws an error in the console at runtime (which I assume is worse) because `div`s are not allowed inside `p` tags. The alternative I think, which would make more sense in general, would be putting the contents of the `I18nHtmlTranslate` in a `span` rather than a `div` as you would expect to use it in a `p` most of the time.

WcaOnRails/app/webpacker/components/StaticPages/Logo.js
revers3ntropy: Yes I was surprised by this too! As the first element in the array is `empty` (NOT `undefined` or `null`), `map` starts at the first element in the array, not from 0. From the docs:
image
Therefore, as the index of the first element is 1, it starts from 1.


This code:
image
Logs this:
image

WcaOnRails/app/webpacker/stylesheets/static_pages/logo.scss
gregorbg: Yes, having a span totally makes sense! Would you mind submitting a separate PR that
1. Changes the `I18nHtmlTranslate` component to use `<span>` instead of `<div>`
2. Adds a `// eslint-disable-next-line react/no-danger` comment in that file to prevent those annoying CI warnings?

WcaOnRails/app/webpacker/components/StaticPages/Logo.js
gregorbg: Did some digging and I found it! The library we're using to export Ruby on Rails translations to the JS world is internally doing some stuff with `lodash`. In particular, it uses `_.set` to write the flat key hierarchy to the global `translations` object that is being accessed when you use `I18n.t`.

Try the following code to see what's going on:
```js
// skip this if you already imported lodash
import _ from 'lodash';

var translationsStore = {};
_.set(translationStore, 'foo.bar.1', 'an awesome translated paragraph!');
_.set(translationStore, 'foo.bar.2', 'another awesome translated paragraph!');

console.log(translationsStore.foo.bar);
```

The problem is the Lodash `set` method that interprets our keys as array indices. I don't see how this could be cleverly prevented without changing all of our original `yml` translations to start indexing at zero (which we definitely *don't* want to do right now because it's too much work for this small PR!)

Realistically, I see two options:
1. Change the code to something like `I18n.t('logo.paragraphs').filter(Boolean).map(...foo....)`
2. Hard-code the four paragraphs like in `RegionalOrganizations/index.js`

I tend towards option (2) but I'm open for arguments and debate!

WcaOnRails/app/webpacker/stylesheets/static_pages/logo.scss
revers3ntropy: Yep I'll get on that

WcaOnRails/app/webpacker/components/StaticPages/Logo.js
revers3ntropy: Thanks for the research!

I feel like (2) makes it very annoying to change the contents of that page, although as I assume it hasn't changed much since the logo's creation in 2005 that may not be an issue :P

I also like (1) because it feels almost like something you should do regardless when using 'dirty' arrays like this, to make it more resilient to bad yml, say empty lines or other unexpected 'empty' elements of the array, especially while indexing is still from 1.

I did think of some other, less good options:

3. Change the indexing in that one place in the yml to start form 0, as this is the only place it is used.
4. Create a custom translate wrapper which will filter out empty values from arrays.

I also wonder why indexing was every from 1 in the first place? What are the chances that it is due to something with Rails or the Rails i18n library that might break if changed?

gregorbg: Indeed (2) is a bit annoying to change / update later on. The reason why we started indexing the YAML at 1 instead of 0 is unclear. I was unable to find Rails documentation that clearly tells you to do so.

Of course, we could just as well use a YML llist. But even changing only the current `logo.paragraphs` key (while there are many more keys that use a similar notion of enumerated "paragraphs", they just haven't been ported over to React yet) would entail changing it in all existing translation YML files (over 20 methinks) which is a lot of tedious work. If you want to go down that path, you should opt for a native array structure like so:

```yaml
logo:
  paragraphs:
    - "some paragraph"
    - "some other paragraph"
    - etc...
  some_other_key:
    foo:
    bar:
```

and see whether that works.

Defining a helper to access paragraphs also sounds like a good idea, see `webpacker/lib/i18n.js` for the current configuration. Monkey-patching a `I18n.paragraphs('some.key')` method should be feasible in JS.

Of course, you can also create the monkey patch _and_ update the YML, but I won't blame you if you don't want to go through the repetitive labor of updating 20+ files :sweat_smile:

WcaOnRails/app/webpacker/components/StaticPages/Logo.js
gregorbg: Marking this as Resolved as per the other PR that I just merged.
Don't forget to rebase this PR and delete the (now redundant) CSS file!

WcaOnRails/app/webpacker/stylesheets/static_pages/logo.scss
gregorbg: Awesome, thanks a lot!

WcaOnRails/app/webpacker/components/StaticPages/Logo.js
revers3ntropy: Ah that is very true I completely forgot you have to update in every locale file...

I like the idea of adding something to the `I18n` global object actually, I'll create a seperate PR for that and you can see if you like that.

WcaOnRails/app/webpacker/components/StaticPages/Logo.js
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * @returns {JSX.Element}
+ * @constructor
+ */
+function Logo({ title }) {
+  return (
+    <div className="wca-logo-information">
+      <h1>{title}</h1>
+      {Object.keys(I18n.t('logo.paragraphs')).map((index) => (
```
coder13, line 14: nit: I would rename `index` to `key`. `index` implies a numerical index value which does not appear to be happening here.

coder13, line 14: nit: I would probably rename index` here to `key`. If it were `index`, I would expect a numerical index value.

revers3ntropy: Yeah, fair enough, I agree.
The only thing is that it's not black and white, because it's the keys of an array, so I think they might be both in this case? Things get messy with JS haha

revers3ntropy:
I think this is ready to merge now

gregorbg (approved):
I am (positively!) surprised that the current approach does not trigger "unused 18n keys" warnings. After all, you're never _explicitly_ using `logo.paragraphs.1`.

Any hunch why this works here but (apparently?) broke things over on the Regional Organizations page?

revers3ntropy:
Yeah I didn't think about that actually!

My only idea is that it cleverly ignores lists, because the old implementation with Rails also didn't explicitly use each index.
I wonder if this can be a workaround for now for `I18n.tArray`, add the comment saying that the path in the yml is used, but not each index of the array? I'll give it a go...

PR #7195, opened 2022-09-01
Created 'I18n.tArray'

revers3ntropy:
Added method to the global I18n object.

Extracted default locale ('en') into constant in i18n, and added some jsdoc type hints.

Converted the regional_organizations/index.js to use the new function over hard coding the index.


WcaOnRails/app/webpacker/lib/i18n.js
```diff
@@ -2,20 +2,62 @@ import { I18n } from 'i18n-js';
...
 // load the actual locale as determined by app/views/layouts/application.html.erb
 window.I18n.locale = window.wca.currentLocale;
+window.I18n.tArray = tArray;
+
+/**
+ * Type hints for tArray too.
+ * @type {I18n & {
+ *  tArray: (scope: string | string[], options?: TranslateOptions) => string | string[]
+ * }}
+ */
```
gregorbg, line 58: I don't think this is the way to properly handle type hints, but I'm open to opinions from other members

WcaOnRails/app/webpacker/lib/i18n.js
```diff
@@ -2,20 +2,62 @@ import { I18n } from 'i18n-js';
...
+ *     scope?: string | string[],
+ *     defaults?: Record<string, any>[],
+ *     missingBehavior?: "message" | "guess" | "error" | string
+ * } & Record<string, *>} TranslateOptions
+ */
+
+/**
+ * @param {I18n} i18n
+ * @param {string} locale
+ */
```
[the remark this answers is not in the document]
revers3ntropy: I'm happy to remove them, personally I don't see it as messy or not-meaningful but that's just my opinion.

I only added them to give better type hinting, as we are not using TypeScript.

gregorbg:
Aaaah, our wonderfully annyoing "is every key used?" test...

This problem will persist every time you use `tArray` from now on. There are two options:
1. Use comments to make the library understand which keys are actually being used. `// i18n-tasks-use t('foo.bar.baz')` makes the library check off `foo.bar.baz` as "alright, it's being used. Works kinda similar to suppressing eslint warnings. But it would require you to hard-code every single string (meaning you would have to know beforehand how many paragraphs there are), which kinda defeats the purpose of the dynamic nature of `tArray`.
2. Teach our custom file scanner in `lib/custom_wca_i18n_scanner.rb` to work with `tArray`. This won't work so smoothly either because the return value of that scan method should be a determinate, definitive list of keys that are used, not a general "something with foo.bar.N where N is some number".

Both approaches are kinda sub-optimal :/

revers3ntropy:
Ah that is very annoying...

Option 2 is especially annoying because I assume that it would fail (unknown keys) if you use all indexes to some arbitrary number; although it is obviously much more flexible.

I'll have a think about this and for now I will just use `.filter(Boolean)` for the /logo page, which should do for now I think.

WcaOnRails/config/i18n-tasks.yml.erb
```diff
@@ -157,7 +157,7 @@ ignore:
 
   I18n::Tasks.add_scanner 'I18n::Tasks::Scanners::PatternMapper',
     only: %w(*.js),
-    patterns: [['I18nHTMLTranslate\s+i18nKey=%{key}', '%{key}']]
+    patterns: [['I18nHTMLTranslate\s+i18nKey=%{key}', '%{key}', 'tArray(%{key})']]
```
gregorbg, line 160: I think you have mis-understood some of the pattern mapping used in `i18n-keys`.

The `patterns` array is of the shape `[[a,b],[a,b],[a,b]]` where `a` = "something that appears in our source code" and `b` = "the key that this library should extrapolate from `a`". This can be useful because normally, when checking for unused keys the library only considers occurences like `I18n.t(some_key)`. When we do `<I18nHTML key=some_key>`, we need to teach it that `some_key` should be extrapolated from that React tag.

Having an entry in the array in the shape of `[a,b,c]` like you're trying here won't work, I'm afraid.

gregorbg: Or, to summarize: If `a` appears in our source code, pretend it's actually `b`.

gregorbg:
I see you've been trying to work around the unused keys issue, but I'm afraid it won't work. I've added a code comment with some explanation.

Looking at https://github.com/glebm/i18n-tasks/issues/456 it seems like the library _does_ support 0-based array indices. So a possible solution would be to re-index **all** the "paragraph" translations in our YML files with 0, including all of the existing translations. That would be a lot of manual work, but it would save us the need of `filter`ing like your custom method does in the first place, and it would also be consistent in terms of YAML notation.

CC @jonatanklosko If you have any background knowledge on why the original English YAMLs were indexed "manually" based on `1` rather than ``0`, please shout 😃

WcaOnRails/config/i18n-tasks.yml.erb
gregorbg: See also https://github.com/thewca/worldcubeassociation.org/pull/6939/files#diff-c00fd3fabe65500332a033bff4b9f9805b48889b2e8eca1764fa3c09a6e4586fL35-L39

The GitHub syntax highlighting is broken and marks the entire paragraph as comment (despite the fact that only the first line actually is a comment) but you can see that we used to be doing `I18n.t(foo.bar.#{index})` using string interpolation and "somehow" ™️ the library was able to figure out the (un)used keys.

jonatanklosko:
@gregorbg I don't recall any specific reason back then, but I don't think lists would work in Internationalize, so I would keep the explicit keys, regardless whether it's 0 or 1-based indexing.

gregorbg:
Thanks for the quick reply! As you can see from our entire discussion above, lists are causing us quite a headache in terms of porting frontend pages over to React / JS while determining (un)used keys. Is there any chance you could update `Internationalize` to support lists?

Or if you have some other smart idea of dealing with this problem, I'm also all ears for any alternative solution!

jonatanklosko:
Are lists going to help with determining unused keys though? The keys would still be generated dynamically. I think we currently explicitly ignore list-like entries that we iterate over:

https://github.com/thewca/worldcubeassociation.org/blob/8a1ed14583e0c210cb176c771a37f00efd4b7486/WcaOnRails/config/i18n-tasks.yml.erb#L109-L111

or is there a bigger issue I'm missing?

gregorbg:
Ah, no you're not missing anything. Thanks for pointing me in the right direction!

@revers3ntropy Suddenly it makes a lot of sense why "some" enumerated lists fail during CI and others don't :sweat_smile: We simply ignore some of them and forgot to explicitly ignore others. I think you can proceed with the original `tArray` implementation and just add the relevant "parent key" to the ignore section that Jonatan pointed us at :)

revers3ntropy:
Ah great OK! Yes, my messing around didn't turn out very useful haha

I'll add that to the yml then thanks!

WcaOnRails/config/i18n-tasks.yml.erb
revers3ntropy: That makes sense, I was hoping that would work but I can see why not now :P

Thanks!

gregorbg: Please revert the change then :)

revers3ntropy: Whoops sure thing!

gregorbg (approved):

PR #7196, opened 2022-09-02
Minor performance fixes to reduce server load during registration peaks

gregorbg:
- Disable registration status icons by default, add an option to show them if desired
- Load event models during registration via `CompetitionEvent` cache
- Cache Records page (Bonus: Gender filter)

### DIFF
--- a/WcaOnRails/app/models/registration_competition_event.rb
+++ b/WcaOnRails/app/models/registration_competition_event.rb
@@ -7,4 +7,5 @@ class RegistrationCompetitionEvent < ApplicationRecord
   validates :competition_event, presence: true
 
   has_one :event, through: :competition_event
+  delegate :event, to: :competition_event, allow_nil: true
 end
--- a/WcaOnRails/app/views/results/_results_selector.html.erb
+++ b/WcaOnRails/app/views/results/_results_selector.html.erb
@@ -86,15 +86,16 @@
   </div>
 </div>
 
-<% if show_rankings_options %>
-  <div class="form-group">
-    <%= label_tag(t("results.selector_elements.gender_selector.gender")) %>
-    <div id="gender" class="btn-group">
-      <%= link_to t("results.selector_elements.gender_selector.gender_all"), request.params.merge(gender: "All"), class: "btn btn-info" + (@gender == "All" ? " active" : "") %>
-      <%= link_to t("results.selector_elements.gender_selector.male"), request.params.merge(gender: "Male"), class: "btn btn-info" + (@gender == "Male" ? " active" : "") %>
-      <%= link_to t("results.selector_elements.gender_selector.female"), request.params.merge(gender: "Female"), class: "btn btn-info" + (@gender == "Female" ? " active" : "") %>
-    </div>
+<div class="form-group">
+  <%= label_tag(t("results.selector_elements.gender_selector.gender")) %>
+  <div id="gender" class="btn-group">
+    <%= link_to t("results.selector_elements.gender_selector.gender_all"), request.params.merge(gender: "All"), class: "btn btn-info" + (@gender == "All" ? " active" : "") %>
+    <%= link_to t("results.selector_elements.gender_selector.male"), request.params.merge(gender: "Male"), class: "btn btn-info" + (@gender == "Male" ? " active" : "") %>
+    <%= link_to t("results.selector_elements.gender_selector.female"), request.params.merge(gender: "Female"), class: "btn btn-info" + (@gender == "Female" ? " active" : "") %>
   </div>
+</div>
+
+<% if show_rankings_options %>
   <div class="form-group">
     <%= label_tag(t("results.selector_elements.show_selector.show")) %>
     <div id="show" class="btn-group">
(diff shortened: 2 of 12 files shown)


WcaOnRails/app/models/registration_competition_event.rb
```diff
@@ -7,4 +7,10 @@ class RegistrationCompetitionEvent < ApplicationRecord
   validates :competition_event, presence: true
 
   has_one :event, through: :competition_event
+
+  def event
+    # This may look silly, but actually CompetitionEvent is using a cache
+    # that would normally be circumvented by the default has_one_through association
+    competition_event&.event
```
viroulep, line 14: Naive question: have you tried `delegate :event, to: :competition_event` instead of implementing the method yourself?
I actually don't know what happens if `competition_event` is nil but maybe it "just works".

gregorbg: Thanks for the idea! Turns out there is `allow_nil: true` for precisely such a case.

WcaOnRails/app/views/results/_results_selector.html.erb
```diff
@@ -86,15 +86,16 @@
...
+<div class="form-group">
+  <%= label_tag(t("results.selector_elements.gender_selector.gender")) %>
+  <div id="gender" class="btn-group">
+    <%= link_to t("results.selector_elements.gender_selector.gender_all"), request.params.merge(gender: "All"), class: "btn btn-info" + (@gender == "All" ? " active" : "") %>
+    <%= link_to t("results.selector_elements.gender_selector.male"), request.params.merge(gender: "Male"), class: "btn btn-info" + (@gender == "Male" ? " active" : "") %>
+    <%= link_to t("results.selector_elements.gender_selector.female"), request.params.merge(gender: "Female"), class: "btn btn-info" + (@gender == "Female" ? " active" : "") %>
   </div>
+</div>
+
+<% if show_rankings_options %>
```
gregorbg, line 98: This change looks odd in the GitHub diff. What effectively happens is that I moved the gender selectbox out of the `if show_rankings_options` block, so that it is always displayed by default. The entire inner block with nested divs is unchanged.

WcaOnRails/app/models/registration_competition_event.rb
viroulep: Awesome!

WcaOnRails/app/views/results/_results_selector.html.erb
viroulep: FTR: it looks great if you check the "hide whitespace change" option ;)

WcaOnRails/app/models/registration_competition_event.rb
```diff
@@ -6,5 +6,5 @@ class RegistrationCompetitionEvent < ApplicationRecord
   belongs_to :competition_event
   validates :competition_event, presence: true
 
-  has_one :event, through: :competition_event
+  delegate :event, to: :competition_event, allow_nil: true
```
viroulep, line 9: Ah I resolved my comment a bit too soon: while this is definitely correct for a basic usage, it's worth noting that removing the `has_one` may prevent us from doing `joins` on `event` in some places of the code.
If we don't do it it's fine, but it's worth keeping in mind!

gregorbg: Good point, updated!

danieljames-dj (approved):
LGTM (I checked each lines and doesn't look like there are any errors or something missed out. I didn't understand the logic completely though.)

FinnIckler:
It looks good to me too, but I am wondering how you pin pointed these things to be a significant load during registration

gregorbg:
> I am wondering how you pin pointed these things to be a significant load during registration

Whether they are really significant remains to be seen. They are better than nothing, which is good enough at this point.

The starting point was the NewRelic dashboard. I limited the reports to a time window of 30 minutes before and after the crash, and looked at the SQL logs. I made two observations:
1. Even with the huge run on registrations, over half (!!) of our SQL load was caused by something New Relic identifed as "raw SQL query". From experience, I know that only very very few places in our codebase use hand-written select queries, which almost immediately pointed me in the direction of the Records page. (Also worth noting that the NewRelic report detects and marks all SQL queries that are passed through the Rails bindings as such. So seeing a raw SQL select in NewRelic is really unusual.)
2. With (1) out of the way, I saw there was still a comparatively high load of `SELECT` queries as opposed to actually `INSERT`ing registrations. So I booted a local server and looked at all the SQL queries logged by Rails while pretending to register for a random comp from the dev export.
    1. It immediately struck me that there was a lot of `SELECT * FROM Events` going on. This is unusual because `Events` is one of our most "static" tables and we cache it as a `Cachable` concern. Luckily, Rails provides stack traces so I found `RegistrationCompetitionEvent` and optimised it to actually _use_ the cache we have.
    2. Not immediately related to the registration process for one specific comp, but there was still a lot of `SELECT` going on and the stack trace pointed to the icons. So I made those optional instead of loading them by default. Most competitors who want to register probably open our website around the time registration opens and go along "Homepage -> Competitions -> [the competition they want to register for]". This loads *all* current registration counts as part of the "Competitions" page which is ridiculously unnecessary when they want to register for one specifc comp anyways (instead of casually browsing). When you take the time to see which competition still has free spots to attend, you probably also have the time to click one extra checkbox :P

PR #7198, opened 2022-09-03
Add colliding reg dates table to competition form

gregorbg:
To help Delegates with making conscious decisions about (not) overloading our website.
I also took the liberty of refactoring small parts of the AJAX code from the "nearby competitions" table to make it more reusable.

**Nothing is enforced at this point!** The table is just a "nice to know" kind of bulletin. It is still perfectly possible to choose a colliding date, if so desired.

**Code review**: I am a bit unhappy about the super-long `colliding_registration_start_competitions` and `minutes_until_other_registration_starts` names. If you have suggestions for more concise names, please shout.

## Reasoning behind the 3 hours
From WST perspective, ~45 minutes would suffice, but that's probably not enough room to pick a new free spot from.
If the table only shows the surrounding 45 minutes, it would be a lot of trial and error until you hit a date that has no collisions. By making the window larger than it technically needs to be, we allow Delegates to spot adjacent "gaps" in the registration time frame easier.

Does that make sense? Are there any other arguments in favor of a longer/shorter time window?

2022-09-03T14:08:27,202371831+09:00


WcaOnRails/app/models/competition.rb
```diff
@@ -1190,6 +1201,24 @@ def distance_adjacent_to?(c, distance_km)
...
+  def minutes_until_other_registration_starts(c)
+    if !c.has_registration_start_date? || !self.has_registration_start_date?
+      return false
+    end
+    seconds_until = (c.registration_open - self.registration_open).to_i
+    if seconds_until < 0
+      seconds_until = (self.registration_open - c.registration_open).to_i * -1
+    end
+    seconds_until / 60
+  end
```
gregorbg, line 1220: This might look confusing, but it follows the same logic as `days_until_competition` above: https://github.com/thewca/worldcubeassociation.org/pull/7198/files#diff-3148368ff9102d87290f301299773c8e08ebbedccadec8ec1e93b9c1c35fa470R1174

Jambrose777:
The time frame makes sense (maybe the red being <30 minutes between reg start times). I as a Delegate might look to choose an hour difference from what I originally selected, but otherwise just choose a different day. The reason is that I would want to have a reg opening between 6-9 local time. So I think expanding the time from 3 hours would not be advantageous to many Delegates (as they'd either select within the 3 hours or just go to a different day)

Issue #6361, opened 2021-07-02
Add React-Query (or similar) in react components


saranshgrover:
https://react-query.tanstack.com/

We currently have self-created 2 hooks for loading and saving data, `useLoadedData` and `useSaveData`. While both work well, by migrating to react-query, we would be able to easily an exhaustive API as well as caching while still maintaining almost the same functionality.

This will be blocked until #5300 and #6360 are merged, and I'd like more feedback - specifically with members who want to work on the React end about this change.

viroulep:
Using a well tested, used, and maintained library for our queries is definitely something we should look into!
I've never looked into them, is there any other alternatives out there, and did you have a good experience using any of them?

Do you know how caching is implemented?
I've taken a look at the [`queryCache`](https://github.com/tannerlinsley/react-query/blob/master/src/core/queryCache.ts) implementation and it looks to me that it's done in-memory (but I'm very very unsure all the logic regarding its storage is here!).
It's worth keeping in mind that, at the moment, our js "frontend" is quite unusual compared to regular react applications: small components are mounted on "regular" html pages, and we don't have components with long-lasting lives (if you change pages you basically reload everything).
The only example I have in mind that would last longer than a single page is the competition results component: if my understanding of the implementation is right caching would benefit a user switching back and forth from different events on the competition's results page, but upon refreshing the page the cache would be gone.
Writing this I notice [this experimental feature](https://react-query.tanstack.com/plugins/persistQueryClient) that allows to persist the cache to local storage, which would better suit our setup.

Anyway, just some stuff to keep in mind when we do decide we want to cache some queries.

> This will be blocked until #5300 and #6360 are merged

If you want to try the integration of the library I definitely think you could do so on the [post widget](https://github.com/thewca/worldcubeassociation.org/blob/master/WcaOnRails/app/webpacker/javascript/posts_widget/index.js)!
There is a simple load so I'd expect the diff to switch from our hook to the library to be fairly minimal.
And if it is small it should definitely be easy to switch other components too :)

dunkOnIT:
@FinnIckler is this still relevant?

gregorbg:
We've certainly added it in a lot of places but we haven't fully made the switch. This would be a very nice project for someone who wants to take a "stroll" through our React codebase. Excellent second (maybe even first, if you know the basics of React) issue for frontend-minded people.

FinnIckler:
[...]

PR #7208, opened 2022-09-10
Searchable incidents

viroulep:
This is a placeholder for implementing searchable incidents, and it's mainly meant for @kr-matthews.

I have *not* had the time to do the incident index controller method, nor the pagination, nor exploiting it through a dummy react component, nor the additional search fields (ie: tags, competition ids).

It's actually pretty much mergeable as-is (if there was an issue "make incidents searchable in omnisearch"), so I figured I would give you a rough idea of how a very basic implementation would look.

I'll complete that PR with the additional features you need asap ;)

I've also taken notes as I was implementing it about what was going on in my mind, hopefully it may be useful:
```
## Implementing searchable incidents

At this point I now that for most models it's just a matter of implementing a static "search" method, and if we have this quite naturally my first approach would be to integrate it into our omnisearch.
ATM my plan is to:
  - implement a basic search method that use a `q` parameter to search into the public content
  - serialize with an "incident" class, and deal with it in omnisearch

If that works well, I'll go ahead with implementing an index page that rely on "search" + pagination, then add some seach parameters (competitions, tags).

While implementing the search I figured out which attributes I wanted to search in (title, public_summary), and chained some `LIKE %something%` based on the query's parts (it's actually done like this for posts for instance).
When implementing the search, I noticed the search api relies on the default serializable_hash for incidents, which is not suitable as it contains private fields.
I went ahead and implemented a default set of options for serializable_hash.

While implementing the index search "view" page I noticed the use of present? then empty? is a bit weird as it could just be a "if any?"; "else", but I resisted the temptation to rewrite the whole page.
I implemented the search in our homepage omnisearch, then realized that on the "search results" page it's a different omni search: given basically no one uses this one and it's bound to disappear, I did not bother touching this.

When implementing tag search I considered two approaches:
1:
> incidents = incidents.where(incident_tags: IncidentTag.where(tag: tags))
2:
> incidents = incidents.where(id: IncidentTag.where(tag: tags).pluck(:incident_id))

2 unfortunately fires two sql queries, and 1 just works as expected so I kept this one.
I took the same approach for competitions.


```

### DIFF
--- a/WcaOnRails/app/models/incident.rb
+++ b/WcaOnRails/app/models/incident.rb
@@ -38,4 +38,35 @@ class Incident < ApplicationRecord
       errors.add(:digest_sent_at, "can't be set if incident is not resolved.")
     end
   end
+
+  def url
+    Rails.application.routes.url_helpers.incident_url(self)
+  end
+
+  DEFAULT_SERIALIZE_OPTIONS = {
+    only: ["id", "title", "public_summary"],
+    methods: ["url"],
+  }.freeze
+
+  def serializable_hash(options = nil)
+    json = super(DEFAULT_SERIALIZE_OPTIONS.merge(options || {}))
+    json.merge!(
+      class: self.class.to_s.downcase,
+    )
+  end
+
+  def self.search(query, params: {})
+    incidents = Incident
+    query&.split&.each do |part|
+      like_query = %w(public_summary title).map { |col| "#{col} LIKE :part" }.join(" OR ")
+      incidents = incidents.where(like_query, part: "%#{part}%")
+    end
+    if params[:tags]
+      incidents = incidents.where(incident_tags: IncidentTag.where(tag: params[:tags].split(",")))
+    end
+    if params[:competitions]
+      incidents = incidents.where(incident_competitions: IncidentCompetition.where(competition_id: params[:competitions].split(",")))
+    end
+    incidents.order(created_at: :desc)
+  end
 end
--- a/WcaOnRails/app/webpacker/components/SearchWidget/IncidentItem.js
+++ b/WcaOnRails/app/webpacker/components/SearchWidget/IncidentItem.js
@@ -1,0 +1,16 @@
+import React from 'react';
+import I18n from '../../lib/i18n';
+
+function IncidentItem({
+  item,
+}) {
+  return (
+    <div className="omnisearch-item-incident">
+      {I18n.t('incidents_log.incident')}
+      {' '}
+      {item.title}
+    </div>
+  );
+}
+
+export default IncidentItem;
--- a/WcaOnRails/app/controllers/api/v0/api_controller.rb
+++ b/WcaOnRails/app/controllers/api/v0/api_controller.rb
@@ -115,11 +115,15 @@ class Api::V0::ApiController < ApplicationController
     search(Regulation)
   end
 
+  def incidents_search
+    search(Incident)
+  end
+
   def omni_search
     # We intentionally exclude Post, as our autocomplete ui isn't very useful with
     # them yet.
     params[:persons_table] = true
-    search(Competition, User, Regulation)
+    search(Competition, User, Regulation, Incident)
   end
 
   def show_user(user)
--- a/WcaOnRails/app/controllers/search_results_controller.rb
+++ b/WcaOnRails/app/controllers/search_results_controller.rb
@@ -11,6 +11,7 @@ class SearchResultsController < ApplicationController
       @persons = User.search(@omni_query, params: { persons_table: true }).page(params[:people_page]).per(SEARCH_RESULT_LIMIT)
       @posts = Post.search(@omni_query).page(params[:posts_page]).per(SEARCH_RESULT_LIMIT)
       @regulations = Kaminari.paginate_array(Regulation.search(@omni_query)).page(params[:regulations_page]).per(SEARCH_RESULT_LIMIT)
+      @incidents = Incident.search(@omni_query).page(params[:incidents_page]).per(SEARCH_RESULT_LIMIT)
     end
   end
 end
--- a/WcaOnRails/app/views/search_results/index.html.erb
+++ b/WcaOnRails/app/views/search_results/index.html.erb
@@ -95,6 +95,21 @@
       <%= paginate @regulations, param_name: :regulations_page, params: { anchor: "regulations-and-guidelines" } %>
     <% end %>
 
+    <% if @incidents.present? %>
+      <h2><%= anchorable t('.incidents') %></h2>
+      <% @incidents.each do |incident| %>
+        <div>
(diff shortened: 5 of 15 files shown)


viroulep:
I just pushed the tags/competitions filtering with a dummy incidents index components that takes the tags/competitions from the url params, it looks something like this:
filter_tags

It should be enough to give you a good idea of how it can work server side, how you can paginate stuff, and how you can handle interactions between react states and url params!

@gregorbg for pagination-in-header to work I had to implement this: https://github.com/thewca/worldcubeassociation.org/pull/7208/commits/63c48bf429bd45e97bd5b20d79729aa08d856a36
What do you think about merging this part in master?
If I'm not mistaken we would have to do something like this even if we were using an external library like react-query (keeping in mind #6361).

kr-matthews:
Thanks, this is helpful. It looks pretty similar to what I had been trying (though cleaner and more efficient, such as with the header data), so I'm not sure why I couldn't get it to work :(.

I'll start incorporating the relevant changes to my PR - assuming only the single commit mentioned above is going to be merged into master? Or will I need to rebase on/merge all the changes here?
- The existing incidents log doesn't have the ability to search by competition (just by tag and by title via a string), and I'm not sure how useful that would be, so I'm not sure I'll do that.
- The omni search wasn't part of the scope of my PR but I think it makes sense to include it.

viroulep:
> The existing incidents log doesn't have the ability to search by competition (just by tag and by title via a string), and I'm not sure how useful that would be, so I'm not sure I'll do that.

Oh I was convinced I made it possible to filter based on a competition id xD
Yeah don't bother doing the UI for it then!

> I'll start incorporating the relevant changes to my PR - assuming only the single commit mentioned above is going to be merged into master? Or will I need to rebase on/merge all the changes here?

> The omni search wasn't part of the scope of my PR but I think it makes sense to include it.

Addressing both comment at once: I actually think the first two commits of this PR should land in master, they are fairly standalone and useful independently of your PR.
If you/Gregor are fine with it I can remove the last commit from this PR to make it mergeable (and if you need the changes they will always be available in 4575c95a585a0c1028965ec4221d7f31fa1df10c, with the other commits showing up in [the history](https://github.com/thewca/worldcubeassociation.org/commits/4575c95a585a0c1028965ec4221d7f31fa1df10c), regardless of the state of that PR).

kr-matthews:
> Addressing both comment at once: I actually think the first two commits of this PR should land in master, they are fairly standalone and useful independently of your PR. If you/Gregor are fine with it I can remove the last commit from this PR to make it mergeable (and if you need the changes they will always be available in [4575c95](https://github.com/thewca/worldcubeassociation.org/commit/4575c95a585a0c1028965ec4221d7f31fa1df10c), with the other commits showing up in [the history](https://github.com/thewca/worldcubeassociation.org/commits/4575c95a585a0c1028965ec4221d7f31fa1df10c), regardless of the state of that PR).

Ok, that's fine with me, I just wasn't sure because it has several conflicts with my in-progress PR and thought it might be easier for me to make changes myself rather than deal with merge conflicts.

It would be good if the last commit were merged into master so that I can build on that and use the header attributes in my pagination.

gregorbg:
I don't mind merging parts of this PR (or the entire thing) but I'd like to avoid cherry-picking stuff manually.
If you need parts of this to continue working, please submit the relevant commits as a separate PR (or tell me if this entire thing should be merged)

viroulep:
> It would be good if the last commit were merged into master so that I can build on that and use the header attributes in my pagination.

I don't think it can land on master, it basically messes up the incident log index for regular users, but you don't actually need it to land on master to use it (see below).

> Ok, that's fine with me, I just wasn't sure because it has several conflicts with my in-progress PR and thought it might be easier for me to make changes myself rather than deal with merge conflicts.

I just took a look because I thought I had done stuff in the first two commits totally independently to your PR, and the one conflict I got was because of the serializable_hash stuff that we both implemented, as well as the search method being added right next to it: it should be fairly easy to fix as you probably want to keep your implementation of serializable_hash.

Here is what I can suggest:
  - I remove 4575c95a585a0c1028965ec4221d7f31fa1df10c from that PR
  - This gets reviewed/merged
  - You squash your commits in #6971, otherwise rebasing over anything that has a conflict is a pain
  - You rebase your changes fixing the conflict with serializable_hash
  - Depending on what you need for the header pagination and stuff, you either cherry-pick 4575c95a585a0c1028965ec4221d7f31fa1df10c (which will most likely conflict in a bunch of place), or simply take what you need from the diff manually.

The main reason I'm considering to get this merged first is that this is easier to review/merge as a standalone patch rather than a part of a much bigger change :)
If that sounds good to you I'll proceed with the first bullet point!

viroulep:
Done, removed the draft prefix, I think it's ready for review :)

WcaOnRails/app/models/incident.rb
```diff
@@ -38,4 +38,35 @@ def digest_sent_at_consistent
...
+
+  DEFAULT_SERIALIZE_OPTIONS = {
+    only: ["id", "title", "public_summary"],
+    methods: ["url"],
+  }.freeze
+
+  def serializable_hash(options = nil)
+    json = super(DEFAULT_SERIALIZE_OPTIONS.merge(options || {}))
+    json.merge!(
+      class: self.class.to_s.downcase,
```
kr-matthews, line 54: Is this needed for the omnisearch, or anything in particular?

viroulep: Yes absolutely: omnisearch gets a bunch of different objects to display in the same select (users, competitions, regs, incidents, etc), this field let's the client now about which objects it's manipulating (for instance it's used [here](https://github.com/thewca/worldcubeassociation.org/pull/7208/files#diff-320e33e2efdf553ae279a78415ac32a4f30169fcf21f17a7af5b922d73526c5dR22) in this diff).

In retrospect we should probably have named the field `classString` or something to avoid the conflict with ruby's `class` method, but it's definitely too late now as it's propagated everywhere in public api endpoints :(

WcaOnRails/app/webpacker/components/SearchWidget/IncidentItem.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+function IncidentItem({
+  item,
+}) {
+  return (
+    <div className="omnisearch-item-incident">
+      Incident:
```
gregorbg, line 8: i18n?

viroulep: Ah right (the title of incidents can't be localized though)!

This gave me the opportunity to test the "new" way locales are automatically reloaded and it simply works great :)

gregorbg (approved):

viroulep:
Thanks a lot for the review, for the record I don't have anything else to add :)

PR #7211, opened 2022-09-11
[Migrate React] Static Pages: About

revers3ntropy:
Migrate the /about page to React


WcaOnRails/app/views/static_pages/about.html.erb
```diff
@@ -1,101 +1,7 @@
...
-    </em>
-    <%= t('about.tax_deductible_note') %>
-  </p>
-
-  <h2><%= t('about.other_help_title') %></h2>
-  <%= t('about.other_help_content_html') %>
-
-</div>
+<%= react_component('StaticPages/About', {
+  currentBoardMembers: Team.board.current_members.includes(:user).map(&:user).sort_by(&:name)
```
kr-matthews, line 4: More of a general question, rather than for this specific PR, but I'm a little unclear for React components on when we should pass in user/competition/incident/etc. data as params, and when we should fetch it from an API endpoint.

WcaOnRails/app/webpacker/components/StaticPages/About.js
```diff
@@ -0,0 +1,170 @@
...
+    <React.StrictMode>
+      <div className="jumbotron">
+        <p className="lead">
+          {I18n.t('about.donation_banner.content')}
+        </p>
+        <hr className="my-4" />
+
+        <a className="btn btn-primary btn-lg" href={DONATE_GUIDESTAR_URL} role="button">
+          {I18n.t('about.donation_banner.donate_credit')}
+        </a>
```
kr-matthews, line 25: Should we use semantic-ui `Button` for these three?

WcaOnRails/app/views/static_pages/about.html.erb
revers3ntropy: Yes I was wondering exactly the same thing actually!

I only made it take the data as a prop as that option is much easier to implement, but I do think that this case would be a perfect opportunity to create a new API endpoint.
I'd be happy to make the endpoint though if others agree?

gregorbg: I've added my two cents at https://docs.google.com/document/d/1X7veopxmMFKJtXT7WFpcn0w6MqW68fBybSlvBf25bYY/edit#heading=h.w4hztc87n5so

gregorbg: In this specific case, I don't see a real value to creating an endpoint because the Board members never change once the page has been loaded.

WcaOnRails/app/webpacker/components/StaticPages/About.js
gregorbg: Yes please!

WcaOnRails/app/views/static_pages/about.html.erb
viroulep: (I can't access the doc, sorry if my opinion is already reflected there)
IMO basically components should be written as if they were getting their data from "somewhere" not caring about where that somewhere is (except for the component actually doing the data fetch obviously :)); it will make it easier to re-use them if/when the data provider change.

As for static pages, I think providing the data directly in the react component render helper is more helpful: if I'm not mistaken it would also let us enable server side rendering, which is always something good to consider to help search engines.

WcaOnRails/app/webpacker/components/StaticPages/About.js
revers3ntropy: Yep, that's fine

WcaOnRails/app/views/static_pages/about.html.erb
revers3ntropy: The only thing I'd like to point out is that this component is only used once, and the same data is being passed to the component every time. I can't see a case where you would pass any other data to the component, which means that surely the data should be entirely handled within the component? I feel like in cases like this where, wherever you use it, the same data is required, the component should be responsible for getting that data, rather than the parent component (or Rails) passing the same data down every time.

Having said that, using an API endpoint is harder to implement, and I am completely happy to leave it as is.

gregorbg: I don't see an advantage in having the page fetch its own data if that data never changes after the initial fetch.
Do you have a concrete reason for your feeling that the component should be responsible for getting that data?

viroulep: > The only thing I'd like to point out is that this component is only used once, and the same data is being passed to the component every time. I can't see a case where you would pass any other data to the component, which means that surely the data should be entirely handled within the component?

For the record I was commenting on the general approach for creating a component, in this specific case it's clear that with data changing once every year or so it's not quite efficient to make an API call for it, and that passing them through the rails rendering helper should be the way to go.

revers3ntropy:
I have also slightly updated the UI by adding some icons, I'm not sure what the rules are about design choices like this. 
image

WcaOnRails/app/views/static_pages/about.html.erb
revers3ntropy: I don't doubt that it would be much more performant to pass this very small amount of data as a prop, however I don't feel that the UX would be impacted at all regardless of what approach is used.

I was more thinking purely from a design choice perspective, and what parts of the page should be responsible for what data. It just feels neater if the component can be placed into any page without that page having to worry about giving data to the component, data which is always the same regardless of where the component is (that is as opposed to a button, say, which could be rendered with any text in it).

gregorbg:
No rules other than "act responsibly". In this case, it makes perfect sense and is a nice addition to the overall layout! Thanks :)

WcaOnRails/app/webpacker/components/StaticPages/About.js
```diff
@@ -0,0 +1,172 @@
...
+const DONATE_GUIDESTAR_URL = 'https://www.nfggive.com/guidestar/82-3825954';
+
+/**
+ * @param {{ name: string, country_name: string, id: number, [k: *]: * }[]} currentBoardMembers
+ * @returns {JSX.Element}
+ * @constructor
+ */
+function About({ currentBoardMembers }) {
+  return (
+    <React.StrictMode>
```
gregorbg, line 16: What does this wrapper tag do? What benefits does it have?

gregorbg (approved):

PR #7216, opened 2022-09-12
Add migration task for rehashing asset links

gregorbg:
Rails changed the `ActiveSupport` key hashing algorithm to SHA256 in Rails 7. This breaks all of our asset links because they contain encoded data that is checked against a stored hash before you're being forwarded to Amazon. If `hash(encoded_url_data) != stored_hash`, you're getting a 404.

### DIFF
--- a/WcaOnRails/lib/tasks/rehash_competition_tab_links.rake
+++ b/WcaOnRails/lib/tasks/rehash_competition_tab_links.rake
@@ -1,0 +1,71 @@
+# frozen_string_literal: true
+
+namespace :competition_tabs do
+  desc "Rehash links in tabs with SHA256 digest"
+  task rehash_active_record_links: [:environment] do
+    key_generator = ActiveSupport::CachingKeyGenerator.new(
+      ActiveSupport::KeyGenerator.new(Rails.application.secret_key_base,
+                                      iterations: 1000,
+                                      hash_digest_class: OpenSSL::Digest::SHA1),
+    )
+
+    # It's just called like that. Copied from the Rails source code.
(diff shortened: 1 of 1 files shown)


WcaOnRails/lib/tasks/rehash_competition_tab_links.rake
```diff
@@ -0,0 +1,71 @@
+# frozen_string_literal: true
+
+namespace :competition_tabs do
+  desc "Rehash links in tabs with SHA256 digest"
+  task rehash_active_record_links: [:environment] do
+    key_generator = ActiveSupport::CachingKeyGenerator.new(
+      ActiveSupport::KeyGenerator.new(Rails.application.secret_key_base,
+                                      iterations: 1000,
+                                      hash_digest_class: OpenSSL::Digest::SHA1)
```
gregorbg, line 9: This is emulating the old Rails 6 behavior

timreyn (approved):
Thank you, Gregor!

Goosly:
@gregorbg  When could this be merged?
This is quite important, for example we are using images to guide competitors to where the venue and parkings spots are.
Thanks

gregorbg:
I'll greenlight this based on the apparent urgency and also the review provided by Tim.
Needs to be manually run anyways, so I can closely monitor if anything goes wrong.

gregorbg:
I've ran the complete script on production. Images in competition tabs and the "General" page should be restored.

@Goosly https://www.worldcubeassociation.org/competitions/GanshorenSaturdayOpen2022#20599-parking is looking good on my end, can you confirm that the image is correct?

PR #7221, opened 2022-09-13
Optimise results cache by using unified cache keys

gregorbg:
Also add a header line about some features being disabled.

### DIFF
--- a/WcaOnRails/app/controllers/results_controller.rb
+++ b/WcaOnRails/app/controllers/results_controller.rb
@@ -31,7 +31,7 @@ class ResultsController < ApplicationController
     @is_results = splitted_show_param[1] == "results"
     limit_condition = "LIMIT #{@show}"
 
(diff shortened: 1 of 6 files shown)


campos20:
Small note: we should run CAD once this is deployed so we don't (possibly) double the amount of cached data

gregorbg:
> Are we sure that we want translators to translate (maybe) temporary messages?

Yes. If some of the languages are not translated before the keys are removed again, then that's okay. The more outreach we can have via localized messages, the better.

I've also had friends who are not fluent at English ask me what happened to the rankings. Translated messages will help them understand.

FinnIckler:
The code looks good to me, how does this interact with the caching that Alexandre implemented? Is it just a second layer or something completely different

gregorbg:
It's unifying the cache that Alexandre created, and an older HTML fragment cache that has been there for I don't know how long.
Funnily, the two have been using different cache keys, leading to inconsistent results (like Tymon appearing in the UK rankings because the cache key for the DB rows was loading the correct stuff but the cache key for the view was injecting it into the wrong fragment)

PR #7224, opened 2022-09-14
Implement a posting results "check-in"

viroulep:
The WRT has a bunch of new members, and the risk of two (or more) members posting results at the same time has increased.
The current workflow is to send a "posting" email in the results thread, but it's definitely error prone and may or may not reflect what's being done on the website depending on the mail clients/internet.

The idea behind this PR is to introduce some kind of "posting results" index, which would list the competition pending posting from a WRT members, let WRT member declare they are posting one or more competitions (aka "taking the  lock"), let other members visiting the page know that someone is posting (and who).

As far as the WRT is concerned I think this contains just what we need, but I'll let other members chime in if needed.

Screenshot of a list of competitions needing posting:
initial

This is the view when taking the lock on one competitions:
locked

It sends a mail in the results email thread of that competition:
mail

This is the same view but from another WRT member:
someone_posting

Clicking "post the competition" on the competition admin page clear the lock for that competition, and it looks like this:
unlocked

It's worth noting that the person posting one competition can add more competitions to those currently being posted (since we can post multiple competitions at the same time), however the form would be totally disabled if another member is posting at least one competition.

If nothing more urgent requires it, do you think it would be possible to deploy these changes to staging so that the WRT can take a look at how it looks "in real time"?

### DIFF
--- a/WcaOnRails/app/controllers/admin/results_controller.rb
+++ b/WcaOnRails/app/controllers/admin/results_controller.rb
@@ -4,6 +4,61 @@ module Admin
   class ResultsController < AdminController
     # NOTE: authentication is performed by admin controller
 
+    def posting_index
+      respond_to do |format|
+        format.html do
+          render :posting_index
+        end
+        format.json do
+          @pending_competitions = Competition.pending_posting
+          user_attributes = {
+            only: ["id", "name"],
+            methods: [],
+            include: [],
+          }
+          render json: {
+            current_user: current_user.as_json(user_attributes),
+            competitions: @pending_competitions.as_json(
+              only: ["id", "name"],
+              methods: ["city", "country_iso2"],
+              include: { posting_user: user_attributes },
+            ),
+          }
+        end
+      end
+    end
+
+    def start_posting
+      # The purpose of this action is to allow an admin to "lock" and start
+      # posting competitions. This is an informative lock that relies on
+      # WRT members common sense though, nothing should actually prevent
+      # posting from happening.
+      # If posting is possible, the idea is to:
+      #   - get all pending competitions
+      #   - filter out those who are already being posted (by the same member)
+      #   - intersect with the competitions in the parameters
+      #   - either lock them and reply ok, or there is none to lock and reply
+      #   it was a no-op.
+      @other_posting = Competition.where.not(posting_user: [nil, current_user]).any?
+      if @other_posting
+        return render json: { error: "Someone is already posting!" }
+      end
+      @updated_competitions = Competition.pending_posting.where(posting_user: nil).where(id: params[:competition_ids])
+      if @updated_competitions.empty?
+        return render json: { error: "No competitions to lock." }
+      end
+
+      json = { error: "Something went wrong." }
+      if @updated_competitions.update(posting_user: current_user)
+        @updated_competitions.each do |c|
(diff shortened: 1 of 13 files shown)


viroulep:
Almost forgot, it's accessible from a "posting dashboard" entry in the admin panel:
menu

Jambrose777:
Awesome work!

One quick question: If I decide that I cannot post results after checking the box next to a competition, would I be able to uncheck the box to release the lock?

Is the following possible?
For releasing the lock:
Change it from the button on the dashboard "Post Results" to Competition Page > Edit > Admin View > "Post the Results" Button? Clicking the actual post results button would release the lock for that competition.

viroulep:
> One quick question: If I decide that I cannot post results after checking the box next to a competition, would I be able to uncheck the box to release the lock?

No, since checking-in/locking sends a mail I figured it would be weird having multiple mails; it wouldn't be too much trouble to allow the user to unlock it though (and it's worth noting that regardless of the state displayed on the page, you can always do any admin task you want, including uploading any result).

> Is the following possible?
For releasing the lock:
Change it from the button on the dashboard "Post Results" to Competition Page > Edit > Admin View > "Post the Results" Button? Clicking the actual post results button would release the lock for that competition.

I think the behavior you want is already implemented: the "post the results" button on this new page link to the php posting page for that competition!
The lock for that competition is indeed release when clicking the "post results" button on the competition admin page (see [here](https://github.com/thewca/worldcubeassociation.org/pull/7224/files#diff-27c86a93def2d6c9ce49ba5296920e61a710a2259983f78d3875985869991308R264)).

gregorbg:
What's the current status on this? Are there any open "procedural" questions that need to be clarified among WRT?
Or is this just pending a code review from us?

gregorbg:
Whoops, forgot to mention but: This is now active on staging.

FinnIckler:
Was this played with on staging? I don't remember any discussion

gregorbg:
> Was this played with on staging? I don't remember any discussion

It was briefly active but with all the recent re-deploys it definitely isn't on staging anymore.

gregorbg:
Potentially dumb question, but: With the PHP migration done, do you even need this feature anymore?

dunkOnIT:
@viroulep just following up if this is still needed?

viroulep:
Ha nice coincidence, I just pushed a rebased version of this PR!

Yes I think it would be a nice addition, especially given @danieljames-dj has some improvements idea for it!
Removing php did let us post in parallel, but we still "synchronize" through a slack channel to make sure a competition is posted by a single WRT member.
We also still don't have a nice page displaying all the competitions which need to be posted (I personally just keep unposted competitions in my mail inbox and archive the thread when it's done).
I believe this PR provides a very basic but efficient dashboard to know what's being done/need to be done :)

viroulep:
The screenshots are still accurate, I just changed the "Post the results" button to "Import the results" to match the route/page name.

danieljames-dj (approved):
LGTM

Thanks a lot @viroulep for working on this. I ran this code in my local and it looks good.

This will definitely help me to extend to other WRT stuffs.

viroulep:
Thanks for the fast review Daniel!

gregorbg:
Merged and deployed :D

PR #7225, opened 2022-09-15
Fix use of a removed attribute in rails 7

viroulep:
Just had to run a load of the database dump and ran into this, it looks like the attribute is gone in rails 7!


gregorbg (approved):
Indeed this was deprecated in Rails 6.1, nice catch!
https://github.com/rails/rails/blob/6-1-stable/activerecord/lib/active_record/connection_handling.rb#L309

viroulep:
Thanks for the quick merge/review!

PR #7229, opened 2022-09-17
Sends friendly reminder to submit delegate report.

### LINKED ISSUE #6936: Send reminder to delegates ~24 hours before report / results deadline
Hi all! First Github issue, so I hope I got everything right.

Just having missed my report deadline (mainly because I got sick during most of the last week) got me thinking about the fact that no deadline warnings are issued to delegates when the deadline is approaching. I know it is the delegate´s responsibility to get it in on time regardless, but all delegates are just volunteers trying their best to serve their local communities, so a friendly reminder in case they just forgot will surely be beneficial to everyone? It should make our lives a bit easier, and could also reduce the workload for assistants and seniors. 

I propose a 24 hour reminder, and I think that it could be implemented for results submissions at the same time (even though I feel like getting reports in on time might be a bigger issue).

tank-bohr:
Fixes #6936

### DIFF
--- a/WcaOnRails/app/jobs/submit_report_nag_job.rb
+++ b/WcaOnRails/app/jobs/submit_report_nag_job.rb
@@ -11,11 +11,19 @@ class SubmitReportNagJob < SingletonApplicationJob
     return unless competition.end_date
     return if competition.delegate_report.reminder_sent_at
 
-    deadline = 7.days.since(competition.end_date)
-    1.day.before(deadline).today?
+    # Send a reminder 6 days after the end of the competition
+    6.days.since(competition.end_date).today?
   end
 
   def perform
+    competitions =
+      Competition
+      .visible
+      .not_cancelled
+      .includes(:delegate_report)
+      .where("start_date >= ?", DelegateReport::REPORTS_ENABLED_DATE) # Don't send nag emails for very old competitions without reports.
+      .where(delegate_reports: { posted_at: nil })
+
     competitions.each do |competition|
       if reminder_needed?(competition)
         send_reminder(competition)
@@ -35,14 +43,5 @@ class SubmitReportNagJob < SingletonApplicationJob
   def send_reminder(competition)
     competition.delegate_report.update(reminder_sent_at: Time.now)
     CompetitionsMailer.submit_report_reminder(competition).deliver_now
-  end
-
-  def competitions
-    Competition
-      .visible
-      .not_cancelled
-      .includes(:delegate_report)
-      .where("start_date >= ?", DelegateReport::REPORTS_ENABLED_DATE) # Don't send nag emails for very old competitions without reports.
-      .where(delegate_reports: { posted_at: nil })
   end
 end
--- a/WcaOnRails/app/jobs/submit_report_nag_job.rb
+++ b/WcaOnRails/app/jobs/submit_report_nag_job.rb
@@ -3,20 +3,46 @@
 class SubmitReportNagJob < SingletonApplicationJob
   queue_as :default
 
-  def nag_needed(competition)
+  def nag_needed?(competition)
     (competition.delegate_report.nag_sent_at || competition.end_date) <= 8.days.ago
   end
 
+  def reminder_needed?(competition)
+    return unless competition.end_date
+    return if competition.delegate_report.reminder_sent_at
+
+    deadline = 7.days.since(competition.end_date)
+    1.day.before(deadline).today?
+  end
+
   def perform
+    competitions.each do |competition|
+      if reminder_needed?(competition)
+        send_reminder(competition)
+      elsif nag_needed?(competition)
+        send_nag(competition)
+      end
+    end
+  end
+
+  private
+
+  def send_nag(competition)
+    competition.delegate_report.update(nag_sent_at: Time.now)
+    CompetitionsMailer.submit_report_nag(competition).deliver_now
+  end
+
+  def send_reminder(competition)
+    competition.delegate_report.update(reminder_sent_at: Time.now)
+    CompetitionsMailer.submit_report_reminder(competition).deliver_now
+  end
+
+  def competitions
     Competition
       .visible
       .not_cancelled
       .includes(:delegate_report)
       .where("start_date >= ?", DelegateReport::REPORTS_ENABLED_DATE) # Don't send nag emails for very old competitions without reports.
       .where(delegate_reports: { posted_at: nil })
-      .select { |c| nag_needed(c) }.each do |competition|
-        competition.delegate_report.update_attribute(:nag_sent_at, Time.now)
-        CompetitionsMailer.submit_report_nag(competition).deliver_now
-      end
   end
 end
--- a/WcaOnRails/app/jobs/submit_report_nag_job.rb
+++ b/WcaOnRails/app/jobs/submit_report_nag_job.rb
@@ -8,11 +8,7 @@ class SubmitReportNagJob < SingletonApplicationJob
   end
 
   def reminder_needed?(competition)
-    return unless competition.end_date
-    return if competition.delegate_report.reminder_sent_at
-
-    # Send a reminder 6 days after the end of the competition
-    6.days.since(competition.end_date).today?
+    competition.delegate_report.reminder_sent_at.blank? && 6.days.since(competition.end_date).today?
   end
 
   def perform
--- a/WcaOnRails/app/mailers/competitions_mailer.rb
+++ b/WcaOnRails/app/mailers/competitions_mailer.rb
@@ -129,6 +129,17 @@ class CompetitionsMailer < ApplicationMailer
     )
   end
 
+  def submit_report_reminder(competition)
+    @competition = competition
+    mail(
+      from: Team.weat.email,
+      to: competition.all_delegates.pluck(:email),
+      cc: ["assistants@worldcubeassociation.org"] + delegates_to_senior_delegates_email(competition.all_delegates),
+      reply_to: delegates_to_senior_delegates_email(competition.all_delegates),
+      subject: "Friendly reminder to submit #{competition.name} Delegate Report",
+    )
+  end
+
   def results_submitted(competition, results_submission, submitter_user)
     @competition = competition
     @results_submission = results_submission
--- a/WcaOnRails/app/mailers/competitions_mailer.rb
+++ b/WcaOnRails/app/mailers/competitions_mailer.rb
@@ -135,7 +135,7 @@ class CompetitionsMailer < ApplicationMailer
       from: Team.weat.email,
       to: competition.all_delegates.pluck(:email),
       cc: ["assistants@worldcubeassociation.org"] + delegates_to_senior_delegates_email(competition.all_delegates),
-      reply_to: delegates_to_senior_delegates_email(competition.all_delegates),
+      reply_to: competition.all_delegates.pluck(:email),
       subject: "Friendly reminder to submit #{competition.name} Delegate Report",
     )
   end
--- a/WcaOnRails/app/mailers/competitions_mailer.rb
+++ b/WcaOnRails/app/mailers/competitions_mailer.rb
@@ -135,7 +135,7 @@ class CompetitionsMailer < ApplicationMailer
       from: Team.weat.email,
       to: competition.all_delegates.pluck(:email),
       cc: ["assistants@worldcubeassociation.org"] + delegates_to_senior_delegates_email(competition.all_delegates),
-      reply_to: competition.all_delegates.pluck(:email),
+      reply_to: delegates_to_senior_delegates_email(competition.all_delegates),
       subject: "Friendly reminder to submit #{competition.name} Delegate Report",
     )
   end
--- a/WcaOnRails/config/application.rb
+++ b/WcaOnRails/config/application.rb
@@ -78,8 +78,6 @@ module WcaOnRails
     config.autoload_paths << Rails.root.join('lib')
     config.eager_load_paths << Rails.root.join('lib')
 
-    config.action_mailer.preview_path = "#{Rails.root}/lib/mailer_previews"
-
     # Set global default_url_options, see https://github.com/rails/rails/issues/29992#issuecomment-761892658
     root_url = URI.parse(EnvVars.ROOT_URL)
     routes.default_url_options = {
--- a/WcaOnRails/config/application.rb
+++ b/WcaOnRails/config/application.rb
@@ -78,6 +78,8 @@ module WcaOnRails
     config.autoload_paths << Rails.root.join('lib')
     config.eager_load_paths << Rails.root.join('lib')
 
+    config.action_mailer.preview_path = "#{Rails.root}/lib/mailer_previews"
+
     # Set global default_url_options, see https://github.com/rails/rails/issues/29992#issuecomment-761892658
     root_url = URI.parse(EnvVars.ROOT_URL)
     routes.default_url_options = {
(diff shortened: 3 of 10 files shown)


tank-bohr:
@gregorbg Could you please take a look :pray:

viroulep:
@gregorbg I can take a look if you want.

gregorbg:
@viroulep By all means feel free!

But actually the Board wants to have a say in the concrete wording of the reminder email. So we should hold off merging this until they give their OK,

WcaOnRails/app/jobs/submit_report_nag_job.rb
```diff
@@ -3,20 +3,46 @@
...
+  def reminder_needed?(competition)
+    return unless competition.end_date
+    return if competition.delegate_report.reminder_sent_at
+
+    deadline = 7.days.since(competition.end_date)
+    1.day.before(deadline).today?
+  end
+
   def perform
+    competitions.each do |competition|
```
viroulep, line 19: This is not a a comment on the code itself, but rather an alternative approach we could have used if we had like 10 things to check like this.
You could create a array looking like this:
```ruby
updates = [
  { cond: "reminder_needed?", action: "submit_report_reminder", field: "reminder_sent_at"},
  { cond: "nag_needed?", action: "submit_report_nag", field: "nag_sent_at" },
]
```
And then used it in something like that to DRY it up:
```ruby
competitions.each do |c|
  updates.each do |up|
    if c.send(up[:cond])
      CompetitionsMailer.send(up[:action], c).deliver_now
      competition.delegate_report.update_attribute(up[:field], Time.now)
    end
  end
end
```

It's mostly an FYI, your code is perfectly fine as-is :)

WcaOnRails/app/mailers/competitions_mailer.rb
```diff
@@ -129,6 +129,17 @@ def submit_report_nag(competition)
     )
   end
 
+  def submit_report_reminder(competition)
+    @competition = competition
+    mail(
+      from: Team.weat.email,
+      to: competition.all_delegates.pluck(:email),
+      cc: ["assistants@worldcubeassociation.org"] + delegates_to_senior_delegates_email(competition.all_delegates),
+      reply_to: delegates_to_senior_delegates_email(competition.all_delegates),
```
viroulep, line 138: I think you just want the delegates email here: for the "nag" email we want the include the Senior in the reply-all because the delegates are actually late to submit the report, and it may or may not have some consequences.
```suggestion
      reply_to: competition.all_delegates.pluck(:email),
```

WcaOnRails/config/application.rb
```diff
@@ -78,6 +78,8 @@ class Application < Rails::Application
     config.autoload_paths << Rails.root.join('lib')
     config.eager_load_paths << Rails.root.join('lib')
 
+    config.action_mailer.preview_path = "#{Rails.root}/lib/mailer_previews"
```
viroulep, line 81: I'm actually out of my knowledge area here: could you explain a bit why this is necessary and what it does?
(ok I've reached the end of the PR and now understand: keeping this comment here to remember that this probably needs to go)

WcaOnRails/lib/mailer_previews/competitions_mailer_preview.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class MailerPreviews::CompetitionsMailerPreview < ActionMailer::Preview
```
viroulep, line 3: Why do you need this class and what does it do?

I'm not sure if it does something relevant for us (and if it does it only does so on 1 competition?).

WcaOnRails/spec/jobs/submit_report_nag_job_spec.rb
```diff
@@ -27,4 +27,25 @@
       SubmitReportNagJob.perform_now
     end.to change { ActionMailer::Base.deliveries.length }.by(0)
   end
+
+  it "schedules report reminder email" do
+    _unscheduled_competition = FactoryBot.create :competition, starts: nil
```
viroulep, line 32: Nit: on this variable and others, could you please remove the leading `_`?
I know the original code has them, but I have no idea why this file has a totally different naming convention from the rest of the code :')

If we can avoid introduce more of these that would be awesome.

viroulep (changes requested):
Thanks a lot for the pull request!

I've made a bunch of comments requesting some changes, but I want to emphasize that the proposed diff is already very good and that the high level approach to fix the issue looks good to me!

WcaOnRails/app/jobs/submit_report_nag_job.rb
```diff
@@ -3,20 +3,46 @@
...
+  def nag_needed?(competition)
     (competition.delegate_report.nag_sent_at || competition.end_date) <= 8.days.ago
   end
 
+  def reminder_needed?(competition)
+    return unless competition.end_date
+    return if competition.delegate_report.reminder_sent_at
+
+    deadline = 7.days.since(competition.end_date)
+    1.day.before(deadline).today?
```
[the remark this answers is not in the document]
tank-bohr: Maybe my rubocop is too aggressive. it enforces the [guard clause](https://docs.rubocop.org/rubocop/1.36/cops_style.html#styleguardclause) rule.

`nil` is a perfect falsey value. Ruby encourages you not to write something you don't have to write.

WcaOnRails/app/jobs/submit_report_nag_job.rb
```diff
@@ -3,20 +3,46 @@
...
+    competition.delegate_report.update(nag_sent_at: Time.now)
+    CompetitionsMailer.submit_report_nag(competition).deliver_now
+  end
+
+  def send_reminder(competition)
+    competition.delegate_report.update(reminder_sent_at: Time.now)
+    CompetitionsMailer.submit_report_reminder(competition).deliver_now
+  end
+
+  def competitions
```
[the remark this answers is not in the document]
tank-bohr: The aim of extracting functions is not only code reuse. We do it also for giving the piece of code a name. We help others to read this code in the future.

WcaOnRails/app/jobs/submit_report_nag_job.rb
tank-bohr: You suggest generalizing the code by introducing the abstraction (mini data DSL). But any abstractions have a cost. It's harder to read, harder to maintain. It's even harder to extend if you didn't predict the further evolution of the code :)
Yes, we remove code duplication. But at what cost? I believe that it will bring more harm than use.

WcaOnRails/config/application.rb
tank-bohr: This configuration tells rails where to lookup [mail preview]((https://api.rubyonrails.org/classes/ActionMailer/Base.html#class-ActionMailer::Base-label-Previewing+emails) ) classes. For automatic mail preview generation.

WcaOnRails/lib/mailer_previews/competitions_mailer_preview.rb
tank-bohr: It's a standard way to preview emails in the modern rails version. Because it's built-in. I needed that just for a quick check if I broke something in the mail markup or not.

WcaOnRails/spec/jobs/submit_report_nag_job_spec.rb
tank-bohr: the underscore is used to explicitly mark this variable as unused. We tell the compiler not to raise the corresponding warning this way. I can still get rid of the variable at all. But I find it useful TBH.

tank-bohr:
@viroulep Thank you very much for your review. I tried to explain the reasons behind my changes. If they are still not convincing for you, I will change the code according to the comments obediently :)

WcaOnRails/app/jobs/submit_report_nag_job.rb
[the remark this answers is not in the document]
viroulep: > nil is a perfect falsey value. Ruby encourages you not to write something you don't have to write.

Yeah but a function which can return different types just sounds silly to me, regardless of the language ^^

> Maybe my rubocop is too aggressive. it enforces the [guard clause](https://docs.rubocop.org/rubocop/1.36/cops_style.html#styleguardclause) rule.

Wow ok, nevermind then.

> This is exactly what was asked in the original ticket. I find it very cool

Well 24h before the deadline is simply 6 days after the competition so I think there is no need to write this in a convoluted way, but this is definitely subjective so your way is fine too.

WcaOnRails/app/jobs/submit_report_nag_job.rb
viroulep: Which is why I wrote "if we had like 10 things to check", not 2 ^^

Arguably it's harder to maintain 10 similar parts of code than maintaining the "template" used for these 10 parts.

WcaOnRails/config/application.rb
viroulep: Oh that does sound awesome, it means we could get rid of mailcatcher at some point, right?

Maybe not in this PR, but how do you feel about improving our documentation by instructing the reader how to preview/what to open to preview emails (taking the reminder emails as an example for instance)?
If my understanding is correct and if you feel really motivated, how do you feel about migrating a couple of our emails to this system so that we can simply say goodbye to mailcatcher?

WcaOnRails/lib/mailer_previews/competitions_mailer_preview.rb
viroulep: Thanks, I'll resolve this since there is the other thread!

WcaOnRails/spec/jobs/submit_report_nag_job_spec.rb
viroulep: Thanks I forgot about the use of `_` in that case.

In general I would get rid of an unused variable if I could (eg: by putting a comment before the bunch of unused variables if needed), but I don't feel too strongly about that.

tank-bohr:
@viroulep, Could you please take a look? I believe I managed to fix all the comments.

viroulep (approved):
Awesome, thanks for your work!

I can't deploy code but it LGTM :)

gregorbg:
Hey, as stated above the Board wants a say in the decision about the email contents.
I will tr to move this forward but please understand that the process might be inherently slow.

gregorbg:
There is a draft that should be accessible read-only under https://docs.google.com/document/d/1xtsg-mIVE9Gc2b4Vyor9cGEGDAh2bv4CeKxdfFWBTf4/edit?usp=sharing

Please change the wording of your email in this PR accordingly.

WcaOnRails/app/jobs/submit_report_nag_job.rb
```diff
@@ -3,20 +3,45 @@
...
+  def nag_needed?(competition)
     (competition.delegate_report.nag_sent_at || competition.end_date) <= 8.days.ago
   end
 
+  def reminder_needed?(competition)
+    return unless competition.end_date
+    return if competition.delegate_report.reminder_sent_at
+
+    # Send a reminder 6 days after the end of the competition
+    6.days.since(competition.end_date).today?
```
gregorbg, line 15: Is there any way that the logic behind this check could be "unified" with the `nag_needed?` above?

In particular, the nag check is very "dense", while this reminder check is spread over three (non-comment) lines. I definitely prefer the thorough line-by-line check but in general I think it would be great for code maintainability to have these two methods look (almost) the same.

gregorbg (commented):
General comment about code structure, but generally I agree with Philippe's LGTM

WcaOnRails/app/jobs/submit_report_nag_job.rb
tank-bohr: Fixed :wrench:

tank-bohr:
> There is a draft that should be accessible read-only under https://docs.google.com/document/d/1xtsg-mIVE9Gc2b4Vyor9cGEGDAh2bv4CeKxdfFWBTf4/edit?usp=sharing
> 
> Please change the wording of your email in this PR accordingly.

Done! :sunglasses:

WcaOnRails/app/mailers/competitions_mailer.rb
```diff
@@ -129,6 +129,17 @@ def submit_report_nag(competition)
     )
   end
 
+  def submit_report_reminder(competition)
+    @competition = competition
+    mail(
+      from: Team.weat.email,
+      to: competition.all_delegates.pluck(:email),
+      cc: ["assistants@worldcubeassociation.org"] + delegates_to_senior_delegates_email(competition.all_delegates),
+      reply_to: competition.all_delegates.pluck(:email),
```
gregorbg, line 138: Nitpick: This has `reply_to` set to all Delegates of the competition, but the body text from the Google Doc template says "please reply to this email to update and explain this to your Senior Delegate". So I think it would make sense to change this to the senior Delegate emails, similar to the existing "nag" mail.

gregorbg (changes requested):
One nitpick about the mail headers, otherwise all good to go.

WcaOnRails/app/mailers/competitions_mailer.rb
tank-bohr: Fixed :wrench: 
It's funny that it was the first version. I changed it after [this](https://github.com/thewca/worldcubeassociation.org/pull/7229#discussion_r975349408) comment :smile_cat:

gregorbg: Good point, but that conversation actually happened before the Board provided us with a concrete draft.

gregorbg (approved):

gregorbg:
@tank-bohr Code changes are fine. Please look at the output from the failing tests to fix the naming convention of one of your classes.

tank-bohr:
Should be good now

gregorbg:
Thanks a lot! I see that your fix for the last error was to just delete the `mailer_preview` part altogether; was that intended?

tank-bohr:
> Thanks a lot! I see that your fix for the last error was to delete the `mailer_preview` part; was that intended?

Yes, It was intended. I removed the file and moved all the code to the corresponding mailer preview class. He will be better there with the rest of his friends :smile_cat: I just missed this file initially, because it is not in the exact default place. So now it's good.

gregorbg:
> moved all the code to the corresponding mailer preview class

In the Diff view, I cannot see the Mailer Preview anymore, and it seems to me you deleted the feature altogether. Can you please clarify whether this was intended?

tank-bohr:
> > moved all the code to the corresponding mailer preview class
> 
> In the Diff view, I cannot see the Mailer Preview anymore, and it seems to me you deleted the feature altogether. Can you please clarify whether this was intended?

I removed the file but didn't remove the feature. The feature was already [there](https://github.com/thewca/worldcubeassociation.org/blob/2adf5b098d4ab9328ac51b27562f8e37d8cbf5a8/WcaOnRails/spec/mailers/previews/competitions_mailer_preview.rb#L3-L4) for a long time. So I should put it there and not create a new class for that. But I wasn't able to find this class due to
- it's not located on the default path (default path is `test/mailers/previews/`)
- there is no configuration put into the custom path

So I decided that there is no such preview mailer class and created a new one. I missed the fact that since the testing framework is changed to the rspec, the default path of the tests is `spec/`. And default path for mailer previews becomes `spec/mailers/previews/`. It's not obvious. Convention over configuration could be tricky sometimes. :man_shrugging: 

At the end of the day, I figured everything out and put this mail preview [where I should have put it at the very beginning](https://github.com/tank-bohr/worldcubeassociation.org/blob/92899ac494aaff292dec943b4be40c73d968618e/WcaOnRails/spec/mailers/previews/competitions_mailer_preview.rb#L82-L85).

gregorbg:
Okay gotcha, thanks for the thorough explanation. Shipping now! :ship: 



> It's not obvious. Convention over configuration could be tricky sometimes.

Preaching to the choir here :smile:

PR #7235, opened 2022-09-20
Remove edit-result link from the cached table

viroulep:
The results table "per-person" is cached, but the "is admin" part isn't in the cache key, and therefore if an admin view the results first the cached version has the edit link for everyone.

It doesn't mean they can actually edit a result, but it's definitely a bit confusing!

As far as I know we don't actually use this link in the WRT anyway: the proper fix would have been to adjust the cache key but we can simply remove the link :)

### DIFF
--- a/WcaOnRails/app/helpers/competitions_helper.rb
+++ b/WcaOnRails/app/helpers/competitions_helper.rb
@@ -324,4 +324,9 @@ module CompetitionsHelper
 
     CompetitionSeries.new(competitions: [form_competition, competition])
   end
+
+  def result_cache_key(competition, view, is_admin: false)
+    results_updated_at = competition.results.order('updated_at desc').limit(1).pluck(:updated_at).first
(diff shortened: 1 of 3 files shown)


gregorbg:
Thanks for figuring this one out! Seeing that you're a newly appointed WRT member, can I take this as a confirmation that WRT is fine with this change?

viroulep:
Sorry for being unclear: it's just my limited experience with the WRT so far making me think this proposal is suitable; I've asked for more feedback from other WRT members that this is an acceptable solution, I'll post a confirmation here when I'm confident enough it's the consensus in the WRT :)

viroulep:
Cool, I'll try to find a bit of time to update that PR!

> Are you sure the result_cache_key is only used in that one place? If yes, it might be worth moving that method from the competition model over to a helper...

Yes, and yes also it should definitely have been in some kind of helper in the first place!

viroulep:
Updated with only one version of the fix!

gregorbg (approved):

PR #7269, opened 2022-10-02
Remove Zeitwerk migration remnants

gregorbg:


gregorbg:
Merging as part of the monthly dependency update cycle

PR #7277, opened 2022-10-08
[Migrate React] Static Pages: Teams Committees

revers3ntropy:


revers3ntropy:
Interestingly, this fails on my machine on this test:

```
Capybara.using_wait_time(15) do
    # The 3D secure challenge is in a modal in an iframe in an iframe
    within_frame(page.find("body>div>iframe")) do
      within_frame(page.find("#challengeFrame")) do
        within_frame(page.find("iframe[name='acsFrame']")) do
          find(:button, id: button_id).click
        end
      end
    end
    expect(page).to have_text(message)
  end
```

Changing `15` to `60` makes it pass, but apparently works fine in CI, so I'm not sure if this is a problem?

gregorbg:
This is the test that clicks through a Stripe payment process. Navigating the iFrames has been a pain in the a** for a long time, as long as CI passes on the second or third attempt it's fine.

revers3ntropy:
Ah ok, thanks

gregorbg:
Cool! Can I haz screenshots? :smile:

revers3ntropy:
image
image
image
image
image


How's that? I definitely changed up the design a little but I think it is fairly reasonable...

gregorbg:
> How's that? I definitely changed up the design a little but I think it is fairly reasonable...

I love it! Two minor remarks:
1. I'm not too wild about the popup inside a popup when copying the mail address. See if there is some other design / mechanism that you can use. (Completely random thought: Have the mail icon "roll open" on mouseover, i.e. just enlarge it so that the mail address can fit next to the envelope icon whenever you move the mouse there)
2. The black frame around the profile picture might be interpreted as "this person passed away" depending on the culture. Could you either consider another color (there's no law or WCA Motion mandating these frames to be black) or another design instead of the frame?

WcaOnRails/app/webpacker/components/StaticPages/TeamsCommittees.js
```diff
@@ -0,0 +1,82 @@
...
+
+function TeamsCommittees({ officers = [], teams = [], officerTitles = [] }) {
+  return (
+    <>
+      <h1>{I18n.t('about.structure.teams_committees_councils')}</h1>
+      <p>
+        {I18n.t('about.structure.committees')}
+      </p>
+
+      <h3 id="officers">{I18n.t('about.structure.officers.name')}</h3>
```
kr-matthews, line 15: What's this id used for?

WcaOnRails/app/webpacker/components/StaticPages/TeamsCommittees.js
```diff
@@ -0,0 +1,82 @@
...
+        ))}
+      </div>
+
+      {teams.map((team) => (
+        <div className="team" key={team.id}>
+          <h3>
+            <span className="name">{team.name}</span>
+            {team.acronym && team.acronym !== 'BOARD' && (
+              <span className="acronym">
+                {team.acronym ? `(${team.acronym})` : ''}
```
kr-matthews, line 36: Won't we already know that `team.acronym` is truthy inside the span?

WcaOnRails/app/webpacker/components/UserAvatar.js
```diff
@@ -0,0 +1,41 @@
...
+
+  const image = (
+    <div
+      className={`user-avatar-image ${avatarClass}`}
+      style={{ backgroundImage: `url(${url})` }}
+      title={title}
+    />
+  );
+
+  if (!url && !doNotShowFullImageOnHover) {
```
kr-matthews, line 26: Since only the negation of `doNotShowFullImageOnHover` seems to be used, I think it would be clearer to use `showFullImageOnHover` instead.

Also, should this be `||`, or is `&&` correct?

WcaOnRails/app/webpacker/components/UserBadge.js
```diff
@@ -0,0 +1,102 @@
...
+    positions.push(I18n.t('about.structure.treasurer.name'));
+  }
+
+  return positions;
+}
+
+function UserBadge({
+  user, subtexts = [], background = '', badgeClasses = '', senior = false, leader = false,
+}) {
+  let classes = `user-badge ${badgeClasses}`;
```
coder13, line 49: bringing on the [classnames](https://www.npmjs.com/package/classnames) package should improve the readability of this code if this class isn't already included.

WcaOnRails/app/webpacker/components/UserBadge.js
```diff
@@ -0,0 +1,102 @@
...
+  let subtext = (
+    <div className="subtext">
+      {subtexts.map((t, i) => (
+        <div key={`${user.wca_id}-${t}-subtext-${i.toString()}`}>{t}</div>
+      ))}
+    </div>
+  );
+
+  if (!subtexts.length) {
+    subtext = null;
```
coder13, line 67: I don't like this code making the variable null after initializing it. A more standard code pattern would be something like 
```
const subtext = subtexts.length ? <div>...</div> : null;
```

WcaOnRails/app/webpacker/components/UserBadge.js
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * @param {{ teams: { friendly_id: string, leader: boolean }[]}} user
+ * @param {{ friendly_id: string, name: string }[]} officerTitles
+ * @returns {string[]}
+ */
+export function subtextForOfficer(user, officerTitles) {
+  const positions = user.teams
+    .map((team) => {
+      const title = officerTitles.find((t) => t.friendly_id === team.friendly_id);
+      if (title) return title.name;
```
coder13, line 34: nit: Once again, ternary operators can clean this code up.
Or even just a

```js
return title ? title.name : ''
```

Or if this repository supports a slightly more advance js feature:

```js
return title?.name || '';
```

WcaOnRails/app/webpacker/components/StaticPages/TeamsCommittees.js
revers3ntropy: Nothing as far as I can tell! Must have slipped in from Rails, my bad.

WcaOnRails/app/webpacker/components/StaticPages/TeamsCommittees.js
revers3ntropy: Yep, not sure why I've done that :P

WcaOnRails/app/webpacker/components/UserAvatar.js
revers3ntropy: As this prop is never actually used, I think I will remove it for now, and then once it becomes needed someone can add it back, especially as it is so simple to implement.

(And yes I think it would be ||)

WcaOnRails/app/webpacker/components/UserBadge.js
revers3ntropy: Ah that sounds good, I hadn't heard of that

WcaOnRails/app/webpacker/components/UserBadge.js
revers3ntropy: Yep totally agree

WcaOnRails/app/webpacker/components/UserBadge.js
revers3ntropy: Yep fair enough; unfortunately optional chaining does not seem to be supported...

revers3ntropy:
I've got the email address to expand on hover now and only use one popup (absolute pain, ended up using css transitions after trying to do it with Semantic UI).

I only copied the colours around the different roles from the current UI. Personally I feel that all the colours in the user badges detract from the design and just make it feel more cluttered, but I can see why you might want to highlight certain people.
I'm especially happy to treat the board members as normal team members in the UI, as I think if colours are used it should be to highlight members in teams not an entire team.

gregorbg (approved):
Ruby code is fine, and I'll gladly defer to Caleb for his JS expertise :smile:

gregorbg:
Please rebase to resolve the conflicts, and then we're good to merge!

gregorbg:
It still says "This branch cannot be rebased due to conflicts". If you can hit the button on your end, please go ahead.
Otherwise, please make sure to rebase cleanly.

revers3ntropy:
Yep I should be able to, I'll just also tweak the background colours...

Thanks so much for the review, I know you're very busy at the moment.

DanielEgdal:
Quick note, it seems like many avatars are displayed poorly because the regular user defined thumbnail isn't used, but rather some custom crop/minimisation.

revers3ntropy:
I can't see what you mean... Screenshots maybe?

DanielEgdal:
Look at the difference in zoom and centering of WRC. Of course some people don't have the same picture anymore, but focus on the ones that do (Antonio, Han, Oliver, Patrick as examples)
Staging:
image
WCA now
image

revers3ntropy:
Ah yes I do see what you mean.
It appears that the thumbnail should be `..._thumb.jpeg`, thanks for noticing that!

EdHollingdale:
Not sure if this was a desired design change, but it looks like the colours on badges on individual profiles have gone. Personally speaking, the orange on grey is pretty hard to read. 
image

gregorbg:
@revers3ntropy I may have been a bit premature with merging your recent CSS fixes PR. Please investigate the issue described by Edward because it still persists even after your recent round of fixes (and is definitely not intended!)

Randomno:
Senior members are not being ordered correctly (i.e. they should be listed after the team Leader)
image

revers3ntropy:
Interesting... This isn't happening for me on prod 😅
What is your environment?

Randomno:
I'm using the website on Firefox. Works on Edge.

revers3ntropy:
Great thanks, I'll look into it :)

PR #7279, opened 2022-10-10
Add new person validators

tnorr:
Added
- UPPERCASE_NAME_WARNING (raised if a name part is all caps)
- PERIOD_WARNING (raised if a name part contains a period and has a length of at least 3)
- SINGLE_NAME_WARNING (raised if a competitor has a single-piece name)

Updated unit tests to account for the new warnings. Improved the wording of some existing warning texts.

PR #7297, opened 2022-10-11
Add new person validators

tnorr:
Added

- UPPERCASE_NAME_WARNING (raised if a name part is all caps)
- LETTER_AFTER_PERIOD_WARNING (raised if a space is missing after a period)
- SINGLE_NAME_WARNING (raised in case of a single-piece name)

Updated unit tests to account for the new warnings.

Fixed the issues with #7279

### DIFF
--- a/WcaOnRails/lib/results_validators/persons_validator.rb
+++ b/WcaOnRails/lib/results_validators/persons_validator.rb
@@ -22,7 +22,7 @@ module ResultsValidators
     MULTIPLE_NEWCOMERS_WITH_SAME_NAME_WARNING = "There are multiple new competitors with the exact same name: %{name}. Please ensure that all results are correct for these competitors " \
                                                 "and that all results are correctly seperated by their corresponding id."
     WRONG_PARENTHESIS_TYPE_ERROR = "The parenthesis character used in '%{name}' is an irregular character, please replace it with a regular parenthesis '(' or ')' and with appropriate spacing."
-    UPPERCASE_NAME_WARNING = "'%{name}' has an all caps name. Please confirm that this is indeed the correct spelling or fix the name."
+    UPPERCASE_NAME_WARNING = "'%{name}' has successive uppercase letters in their name. Please confirm that this is indeed the correct spelling or fix the name."
     LOWERCASE_NAME_WARNING = "'%{name}' has a lowercase name. Please confirm that this is indeed the correct spelling or fix the name."
     MISSING_PERIOD_WARNING = "'%{name}' has a single letter middle name without an abbreviation period. Please confirm that this is indeed the correct spelling or fix the name."
     LETTER_AFTER_PERIOD_WARNING = "'%{name}' is missing a space after a period in their name. Please confirm that this is indeed the correct spelling or fix the name."
@@ -129,7 +129,7 @@ module ResultsValidators
                                                LOWERCASE_NAME_WARNING,
                                                name: p.name)
           end
-          if split_name.any? { |n| n.upcase == n && n.length > 2 && n != 'DVM' }
+          if split_name.any? { |n| n =~ /[[:upper:]]{2}/ && n.length > 2 }
             @warnings << ValidationWarning.new(:persons, competition_id,
                                                UPPERCASE_NAME_WARNING,
                                                name: p.name)
--- a/WcaOnRails/lib/results_validators/persons_validator.rb
+++ b/WcaOnRails/lib/results_validators/persons_validator.rb
@@ -22,9 +22,6 @@ module ResultsValidators
     MULTIPLE_NEWCOMERS_WITH_SAME_NAME_WARNING = "There are multiple new competitors with the exact same name: %{name}. Please ensure that all results are correct for these competitors " \
                                                 "and that all results are correctly seperated by their corresponding id."
     WRONG_PARENTHESIS_TYPE_ERROR = "The parenthesis character used in '%{name}' is an irregular character, please replace it with a regular parenthesis '(' or ')' and with appropriate spacing."
-    LOWERCASE_NAME_WARNING = "'%{name}' has a lowercase name, please ensure the correct spelling."
-    MISSING_ABBREVIATION_PERIOD_WARNING = "'%{name}' is missing an abbreviation period from a single letter middle name, please ensure the correct spelling."
-    SINGLE_LETTER_FIRST_OR_LAST_NAME_WARNING = "'%{name}' has a single letter as first or last name, please fix the name."
     UPPERCASE_NAME_WARNING = "'%{name}' has an all caps name. Please confirm that this is indeed the correct spelling or fix the name."
     LOWERCASE_NAME_WARNING = "'%{name}' has a lowercase name. Please confirm that this is indeed the correct spelling or fix the name."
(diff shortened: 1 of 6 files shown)


tnorr:
Note: some test names contained DVM (doctor of veterinary medicine) as a suffix. I excluded this from the uppercase warning to prevent unwanted warnings in unit tests.

WcaOnRails/lib/results_validators/persons_validator.rb
```diff
@@ -126,13 +129,27 @@ def validate(competition_ids: [], model: Result, results: nil)
                                                LOWERCASE_NAME_WARNING,
                                                name: p.name)
           end
-          if split_name.length > 2
+          if split_name.any? { |n| n.upcase == n && n.length > 2 && n != 'DVM' }
```
gregorbg, line 132: I understand why you excluded this on behalf of the tests. But the goal should not be "let's not make the tests fail" but rather "let's make some code that works exactly the way WRT wants". So the question is: Does WRT accept suffixes like `DVM` in a competitors name? If not, then it should trigger the warning.

If you need to fix the tests (so that our CI passes) then please fix the part where the names are being _generated_.

WcaOnRails/lib/results_validators/persons_validator.rb
```diff
@@ -126,13 +129,27 @@ def validate(competition_ids: [], model: Result, results: nil)
...
+                                               name: p.name)
+          elsif split_name.length > 2
             if split_name[1, split_name.length-2].any? { |n| n.length == 1 }
               @warnings << ValidationWarning.new(:persons, competition_id,
-                                                 MISSING_ABBREVIATION_PERIOD_WARNING,
+                                                 MISSING_PERIOD_WARNING,
                                                  name: p.name)
             end
           end
+          if split_name.any? { |n| n.chop.include? '.' }
```
gregorbg, line 148: Just curious: What exactly does `.chop` do?

gregorbg (commented):
Just one remark about the DVM, otherwise looking good to go!

WcaOnRails/lib/results_validators/persons_validator.rb
tnorr: It removes the last character of a string

WcaOnRails/lib/results_validators/persons_validator.rb
tnorr: I confirmed that we don't allow title suffixes, and now fixed the name factories instead!

tnorr:
After a WRT discussion, I modified UPPERCASE_NAME_WARNING to cover all instances of 2 or more successive uppercase letters (except when the name length is only 2). Ready for final review!

WcaOnRails/spec/factories/persons.rb
```diff
@@ -9,7 +9,7 @@
       id
     end
     subId { 1 }
-    name { Faker::Name.name }
+    name { Faker::Name.name.gsub(" DVM", "") } # DVM removed to prevent unwanted warnings (we don't allow titles as suffixes)
```
gregorbg, line 12: See if you can merge this into a helper method to use in both factories.

gregorbg (changes requested):
Generally fine by me, but I'm not too wild on `gsub`ing `DVM` in several places. See if you can move it to a distinct helper method (something like creating your own `Faker::Name.name_without_title`) and reference that instead.

WcaOnRails/spec/factories/persons.rb
viroulep: I actually have an alternate - arguably cleaner if it works - suggestion: Faker generates the names based on a locale file, if I'm not mistaken we use `:en` and therefore I think we could add this:
```yml
en:
  faker:
      prefix: []
      suffix: []
```
into a `faker.en.yml` in [this folder](https://github.com/thewca/worldcubeassociation.org/tree/master/WcaOnRails/config/locales), similarly to how we override defaults for devise/carrierwave.
With this config I believe Faker should no longer generate names with either prefixes or suffixes (which I think is the default we want, but maybe I'm missing something).

gregorbg: I love this! That's the kind of solution I was hoping for, but didn't know it existed. Thanks!
Timo, please proceed like Philippe suggested.

tnorr: I added the yml file in config/locales: doesn't work correctly. I'm lost on where faker is actually configured; is there another location where it could be possible to add the commands to override the defaults?

viroulep: I just tried it out locally and it does work, however I missed a nesting level (`name:`) in my original suggestion, sorry for that.
Upon trying it out I also noticed it would generate names like ` John Doe` or `John Doe ` (with an extra space at the beginning or at the end), which is far from ideal.
However a still as simple solution is to override how the name is generated:
```yml
en:
  faker:
    name:
      name:
        - "#{first_name} #{last_name}"
```
The double `name` is intended, see the original faker configuration [here](https://github.com/faker-ruby/faker/blob/main/lib/locales/en/name.yml).

tnorr: Ah, glancing at the original configuration helps a ton! Removing all suffixes wouldn't be ideal for validators since we allow at least Jr., Sr. and numerals. I cut down the suffix list to those, and removed prefixes:

`en:
  faker:
    name:
      suffix: [ Jr., Sr., I, II, III, IV, V ]
      name:
        - "#{first_name} #{last_name} #{suffix}"
        - "#{first_name} #{last_name}"
        - "#{first_name} #{last_name}"
        - "#{first_name} #{last_name}"
        - "#{first_name} #{last_name}"
      name_with_middle:
        - "#{first_name} #{last_name} #{last_name} #{suffix}"
        - "#{first_name} #{last_name} #{last_name}"
        - "#{first_name} #{last_name} #{last_name}"
        - "#{first_name} #{last_name} #{last_name}"
        - "#{first_name} #{last_name} #{last_name}"
`

I'll commit if you agree with this!

viroulep: That sounds like a great solution :)

tnorr: I am still getting the unused key exception, what file name have you been using in your testing?

gregorbg: Not Philippe but basically our code checks think that this is a general translation file and then worry because the content of the file is not actually used anywhere. This is a reasonable check to have in general, but I guess in this case we should configure it so that your file is excluded from the checks.

gregorbg: Cf. `WcaOnRails/config/i18n-tasks.yml.erb`

gregorbg: There is an `ignore` section towards the bottom half of the file that you can leverage.

viroulep: > I am still getting the unused key exception, what file name have you been using in your testing?

Ah I didn't actually run the tests, I just made sure the name generation did work; Gregor is right, you can simply add the faker keys to the one ignored by our [i18n tasks](https://github.com/thewca/worldcubeassociation.org/blob/master/WcaOnRails/config/i18n-tasks.yml.erb#L142).

tnorr: I added the keys to ignore_unused, I believe everything's in order now

gregorbg (approved):

PR #7301, opened 2022-10-16
Add base registration fee and more registration data to WCIF

DanielEgdal:
This allows for accessing more data through the API to organisers. I personally plan to use some of these fields to set up automatic registration acceptance through the API.


gregorbg:
Please create a matching PR at https://github.com/thewca/wcif/pulls describing your changes. I think we should debate over there whether or not adding this information should be included in the first place.

gregorbg (changes requested):
Formally blocking this until we have reached a consensus on whether or not to publicly expose this information in the first place.

gregorbg:
Closing due to inactivity

Issue #7299, opened 2022-10-15
Translation (DE) does not use uppercase on Competitions page


AnnikaStein:
**Describe the bug**
The keys `show_cancelled` as well as `show_registration_status` did not pick up the translation with uppercase notation, despite it being written correctly in the locales file (see here: https://github.com/thewca/worldcubeassociation.org/blob/master/WcaOnRails/config/locales/de.yml#L2984-L2987). This had been merged a month ago, and another part of that PR is correctly visualized, so I think the build has correctly used the most recent .yml file, it's just these two keys not acting as they should.

**To Reproduce**
1. Go to https://www.worldcubeassociation.org/competitions
2. Switch language to "Deutsch" (= German)

**Expected behavior**
Translation should maintain lower-/uppercase exactly as written in locales file

**Screenshots**
<img width="1653" alt="Bildschirmfoto 2022-10-15 um 22 15 05" src="https://user-images.githubusercontent.com/53974095/196006885-01876ba8-2a76-4fbb-9c89-a00d87651c14.png">

**Desktop (please complete the following information):**
 - OS: MacOS 12.3.1
 - Browser: Chrome
 - Version: 106.0.5249.119 (Official Build) (arm64)

**Smartphone (please complete the following information):**
 - Device: iPhone 12 mini
 - OS: iOS 14.2.1
 - Browser: safari

**Additional context**
- Happens only for the German translation, which is (besides Letzebuergesch maybe) the only language I know for which the uppercase does indeed matter in this particular context.
- I could exclude that the newest YAML file might not have been used in the latest website build, and base this statement on the observation that a new translation which was added in the same PR successfully made it to e.g. the user edit page (the `registration_notifications_enabled` key was not there before, and the correct translation appears in my personal profile next to the checkbox).
- I checked a couple of other languages, the translations as shown on the website seemed to match what translators wrote into the respective YAMLs.
- Signing in/out has no effect on the problem.
- Problem also occurs when testing locally (using the docker container setup).
- I tested putting the translation into quotation marks (single or double), it did not change the situation. If I prepend the "Zeige" with something, also "Zeige" will not be shown with capital Z. From this I conclude there is something that automatically renders every character after the first one with lowercase letters.
[...]

PR #7304, opened 2022-10-20
Use content_or_options in label_tag to bypass humanize method [competitions page / translation]

AnnikaStein:
Summarizing the original problem described in https://github.com/thewca/worldcubeassociation.org/issues/7299: When passing only one parameter inside a `label_tag`, this automatically calls the `humanize` method. For the German translation of two keys at the competitions page, this leads to unwanted behavior: we would like to have several words capitalized, not only the first one at the beginning of the translated key.

This PR now explicitly instructs the `label_tag` to write the content exactly as specified in the locales/*.yml file and bypasses the humanize method. From what I can tell, this should not cause trouble for non-German translations, because we usually expect the text to appear exactly as translated with the i18n keys and don't expect `humanize` to take care of underscores or capital letters. As this seems to be the first time this needs to be introduced for this repository (the search only leads to the issue quoted above), I made it explicit in the fix by writing keyword arguments instead of positional arguments only.

Result in local docker build:
Bildschirmfoto 2022-10-21 um 00 13 56

Fixes https://github.com/thewca/worldcubeassociation.org/issues/7299

### DIFF
--- a/WcaOnRails/app/views/competitions/_index_form.html.erb
+++ b/WcaOnRails/app/views/competitions/_index_form.html.erb
@@ -114,14 +114,14 @@
 </div>
 
 <div id="registration-status" class="form-group registration-status-selector">
-  <%= check_box_tag(:show_registration_status, params[:show_registration_status], params[:show_registration_status] == "on") %>
-  <%= label_tag(name = t('competitions.index.show_registration_status'),
+  <%= check_box_tag("#{:show_registration_status}-checkbox", params[:show_registration_status], params[:show_registration_status] == "on") %>
+  <%= label_tag(name = "#{:show_registration_status}-checkbox",
                 content_or_options = t('competitions.index.show_registration_status')) %>
 </div>
 
 <div id="cancelled" class="form-group cancel-selector">
-  <%= check_box_tag(:show_cancelled, params[:show_cancelled], params[:show_cancelled] == "on") %>
-  <%= label_tag(name = t('competitions.index.show_cancelled'),
+  <%= check_box_tag("#{:show_cancelled}-checkbox", params[:show_cancelled], params[:show_cancelled] == "on") %>
+  <%= label_tag(name = "#{:show_cancelled}-checkbox",
                 content_or_options = t('competitions.index.show_cancelled')) %>
 </div>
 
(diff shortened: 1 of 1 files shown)


gregorbg:
Thank you for the incredible thorough investigation! Actually, it seems that `label_tag` in Rails is supposed to be used like so:
```ruby
label_tag(html_id, display_text)
```
such that the first argument does not need to be a translated string from our `t('...')` method. In context, the labels control form elements like so:

```ruby
<% Event.official.each do |event| %>
<span class="event-checkbox">
  <%= label_tag "checkbox-#{event.id}", "Awesome human readable event: #{event} with CapiTAliZatioN!!1!111!" %>
  <%= check_box_tag "checkbox-#{event.id}" %>
</span>
<% end %>
```

So you should change the code to have the first parameter be a computer-readable ID. You can invent anything really.

gregorbg (changes requested):
See my longer explanation post for the requested changes.

AnnikaStein:
Thank you @gregorbg for your feedback! Indeed it sounds good to make a distinction between human-readable parts and the actual identifier.
Now, I currently can't access the device with which I'm usually coding for some more days 🏥, would you mind co-authoring this PR? It looks like it's already quite close to what you suggested, just the mapping between the identifiers needs to be added.

gregorbg:
I don't mind if this stays open for a couple more days.

AnnikaStein:
Finally got to work on this again. The two checkboxes now use the computer-readable IDs as requested to map them to their `label_tags`.

The HTML renders as

```
<div id="registration-status" class="form-group registration-status-selector">
  <input type="checkbox" name="show_registration_status-checkbox" id="show_registration_status-checkbox">
  <label for="show_registration_status-checkbox">Zeige Anmeldestatus-Symbol</label>
</div>
<div id="cancelled" class="form-group cancel-selector">
  <input type="checkbox" name="show_cancelled-checkbox" id="show_cancelled-checkbox">
  <label for="show_cancelled-checkbox">Zeige abgesagte Wettbewerbe</label>
</div>
```

So each label knows for which checkbox it is used for.

@gregorbg Did I understand the requested changes correctly? 🙂

gregorbg:
Yup, all fine. Although the extra `#{:foo}-checkbox` Interpolation probably isn't necessary. The important part is not that it ends in `-checkbox`, but rather that it makes a distinction between the ID (as first argument) and the human label string (as second argument) at all.

Feel free to remove the `-checkbox` Interpolation if you want to, but I won't block the merge at this point.

AnnikaStein:
Ok, thanks! Just used the same pattern that you gave in the example, as I found it quite instructive and a nice cross-check that this mapping works as expected. Like, despite it being computer-readable, there's still the additional string around that makes it clear for us, so I, as an absolute beginner, came to prefer this extra hint. Otherwise I think I understood the explanation (i.e. that `-checkbox` is not the important part). :+1:

Issue #7073, opened 2022-07-27
Allow Organizers and Delegates to register for competitions before they open


simonkellly:
**Is your feature request related to a problem? Please describe.**
Usually when organizing competitions, we have had to change the registration date to the current time before the competition is announced to allow for the organizers and delegates to register for the competition and avoid having to register themselves when the normal competition registration period opens. I can also envision this being especially annoying if the competition has already been announced and another organizer/delegate is to be added

**Describe the solution you'd like**
Remove the checks to stop someone registering early if they are an organizer / delegate

**Describe alternatives you've considered**
Keep using the current solution of changing the registration date.

gregorbg:
This is not teriibly hard to do in terms of code, but there's a long and heated discussion about the "moral" aspect of this issue. As such, we won't immediately implement it nor accept any PRs.

Right now, the WCA needs to define what is and isn't okay for this procedure. Then we can potentially follow up with the code.

EdHollingdale:
Currently many people already do this.  Before the competition is announced adjust the registration open date to be some date in the past, sign up (leaving registrations for organisers and delegates pending on the waiting list), then adjusting the registration opening to be some date in the future before the competition is announced. Once registration does open the organisers and delegates are the first to be processed since they registered earliest. This is explicitly allowed by 6.2.1.2 of the competition requirements policy.

simonkellly:
The request was mainly for before it is announced, and potentially once announced but registration is yet to open

SAuroux:
@gregorbg As Ed already mentioned briefly, the moral behind this is already solved as this is regulated in WCA's [Competition Requirements Policy](https://www.worldcubeassociation.org/documents/policies/external/Competition%20Requirements.pdf).

So in fact this would be a very helpful Software change to eliminate the annoying back and forth changing of the registration start just to allow Organizers and Delegates to register.

simonkellly:
If this is something that is okay to be implemented, I would be happy to write the PR.

dunkOnIT:
I would advocate for only implementing this so that 6.2.1.2 is provided for in the code, but at this stage not add the functionality for  delegate/organiser to be added _after_ a competition opens with any priority, as that seems like a more complex issue - both in terms of the principle and the ease of it. 

@gregorbg would you have any opposition to that? Limiting the scope in that way would seem above board, and I'd be eager given that we have a willing contributor.

Jambrose777:
[...]

PR #7311, opened 2022-10-25
Allow Organisers and Delegates to register for competitions before they open

simonkellly:
This lets listed competition staff register before the normal registration period opens, solving issue #7073.


simonkellly:
I am not 100% sure what to do in terms of tests for this (I probably should have opened this as a draft)

WcaOnRails/app/controllers/registrations_controller.rb
```diff
@@ -604,7 +604,7 @@ def refund_payment
 
   def create
     @competition = competition_from_params
-    if !@competition.registration_opened?
+    if !@competition.registration_opened? && !@competition.user_can_pre_register?(current_user)
```
gregorbg, line 607: Ruby nit: `if !foo` should become `unless foo`. To be fair, this was already wrong before your commit, but now that you're touching this line might as well clean it up :upside_down_face: 

`unless @competition.registration_opened? || @competition.user_can_pre_register?(current_user)`

WcaOnRails/app/views/registrations/register.html.erb
```diff
@@ -32,7 +32,8 @@
     <% end %>
   <% else %>
     <% user_may_register = current_user.cannot_register_for_competition_reasons.length == 0 %>
-    <% if @registration.show_details? %>
+    <% current_user_can_pre_register = @competition.user_can_pre_register?(current_user) %>
+    <% if @registration.show_details? || current_user_can_pre_register %>
```
gregorbg, line 36: Why do you need this additional `OR` check here?
If it is absolutely necessary, can you perhaps merge it into the `show_details?` method? Or will that have too many side effects?

WcaOnRails/app/views/registrations/register.html.erb
```diff
@@ -82,7 +83,7 @@
         </div>
       <% end %>
 
-      <% if @registration.show_details? %>
+      <% if @registration.show_details? || current_user_can_pre_register%>
```
gregorbg, line 86: See above about the additonal `OR` check.

gregorbg (changes requested):
Thanks for the contribution! What is troubling you about the tests in particular?

WcaOnRails/app/controllers/registrations_controller.rb
simonkellly: Nifty, I have made this change

WcaOnRails/app/views/registrations/register.html.erb
simonkellly: Yeah, the show_details method is only really used for this purpose anyway so I have gone ahead and merged them.

simonkellly:
I was not sure if this is the kind of thing that tests should be written for or not

gregorbg:
We most certainly want tests for this, to make sure that your "Average Joe" competitor cannot pre-register. Please consider the following test cases:

1. A Delegate pre-registers. This test should expect to work / return true.
2. A random competitor pre-registers. This test should expect to fail / return false.
3. A Delegate registers during the normal, public registration phase. This should work.

Also see if you can find existing tests, I'd bet there are some that check existing behavior mostly for Average Joe competitors.

simonkellly:
I have added tests which cover the logic for a delegate pre registering as well as for the not already handled behaviour for organisers and registered speedcubers. I don't have much experience in writing tests so I ended up copying much of the logic from the other tests so there is a strong chance I missed something.

gregorbg (approved):
The tests are fine! Thanks for figuring that one out so easily by yourself :D

One nitpick: When you expect any text field or flash warning to contain localized text, you should refer to `I18n.t("foo.bar.key")` directly in case we decide to change the content of the text in the source locale at some point.
But I understand that this behavior might be copy/pasted from existing tests, so I won't blame you for that. Otherwise LGTM!

PR #7358, opened 2022-11-01
Properly link images in /about

revers3ntropy:
Thought I could cheat and say they are just at `/assets/...` but apparently that only works in dev. This should fix it but apparently it working in dev doesn't mean it will in production...

Now passes the urls from the helper `asset_url` to the `About` React component.


WcaOnRails/app/webpacker/components/StaticPages/About.js
```diff
@@ -51,7 +53,11 @@ function About({ currentBoardMembers }) {
           <I18nHTMLTranslate i18nKey="about.who_we_are_content_html" />
         </div>
         <div className="col-md-4">
-          <img className="about-page-image" src="/assets/about/bulgarian-open.jpg" alt="" />
```
viroulep, line 54: If that image is going to be used only in the react parts of the website, wouldn't it work to move them to `app/webpacker/images` (or even in the local component folder actually) and import them directly in `About.js`?
I think `import bulgarianOpen from './bulgarian-open.jpg'` should work if it's in the same folder, but maybe I'm missing something.

gregorbg:
I agree with Philippe that we should "properly" bundle these assets instead of referencing image links across pipelines.

revers3ntropy:
Seems that the jpg images in the repo were empty? Not sure what was going on but it seems to be working correctly at least in dev now.

gregorbg:
> Seems that the jpg images in the repo were empty? Not sure what was going on but it seems to be working correctly at least in dev now.

At least for me, I have two perfectly valid JPG files in my local repo. Where did you take the 1MB big files from that you're adding in this PR?

revers3ntropy:
Huh yeah, seems something odd happened when I moved the images from their original folder but I copied the images from a previous commit and it seems to work fine now...

gregorbg (approved):

PR #7359, opened 2022-11-03
Hide registration edit fields if external website is used

gregorbg:
Allowing competitors to edit the registration doesn't make sense if an external system is used.
This commit just hides the fields altogether, which relies on our database default. (i.e. sets them to `false`)

The selection caused confusion to a Japanese Delegate because they use their own third party registration system over here. So they were not sure what to select, and I figured hiding the selection altogether would make the most sense.

### DIFF
--- a/WcaOnRails/app/views/competitions/_competition_form.html.erb
+++ b/WcaOnRails/app/views/competitions/_competition_form.html.erb
@@ -354,8 +354,10 @@
         <%= f.input :on_the_spot_registration, collection: [ :true, :false ] %>
         <%= f.input :on_the_spot_entry_fee_lowest_denomination, as: :money_amount, currency: @competition.currency_code, currency_selector: "#competition_currency_code" %>
 
-        <%= f.input :allow_registration_edits, collection: [ :true, :false ] %>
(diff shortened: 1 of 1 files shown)


dunkOnIT:
A bit out of my lane here, so feel free to ignore. But would this be vulnerable to an industrious Dev Tools user changing the "false" flag to "true" and editing their registration anyway?

Seems like a pretty edge case, but if there would be a way to not render it all that would seem like the more elegant solution.

That said, the present problem is much more inconvenient than the potential edge case, so I'm in favour of this change overall.

gregorbg:
There is no danger here because this is a form filled out by Delegates which is then submitted to (and checked thoroughly by!) WCAT. The end user never even reaches this page.
Once confirmed, the information is stored in our database which is definitely reasonably safe.

Even assuming they did manage to inject something: Changing this to `true` does nothing when you use an external registration system. To be blunt, there is no registration on the WCA website to edit. At least not until a few days before the competition, at which point the registration edit deadline will kick in anyways.

viroulep (approved):
LGTM :)

PR #7360, opened 2022-11-03
Typescript

revers3ntropy:
This PR allows TypeScript to be incrementally adopted into the React portion of the site.

I think this may be slightly controversial, but I really do think adding Typescript will make everything easier in the long term. I'm not suggesting that all JS be migrated to TS or even that all new code is written in TS, just that using TS becomes an option. TS will especially help as the React portion grows, by giving tooling and type checking, helping to catch common and hard-to-debug errors in JS, for example by making sure that you pass the correct arguments to functions, even if you aren't using an IDE.

One issue is that the Webpack compile time is significantly increased, even using the `ForkTsCheckerWebpackPlugin` plugin, which alone takes nearly 30s to run (as measured by `SpeedMeasurePlugin`), which increases compile times in dev from around 15s to 45s for me.


gregorbg:
We cannot afford increasing compilation times so significantly. Staging already fails at re-compiling our assets while running, you have to hard-reboot the server every time a new branch or feature has to be deployed.

Production is not much better with compilation times ~3 minutes. It's simply not feasible at this point. I agree that having TypeScript will be nice and isn't even that controversial, but we need a "proper" JS toolchain and unified asset handling before going any further.

revers3ntropy:
Yep I understand. I'll keep the branch for future reference.

PR #7364, opened 2022-11-05
Implement correct handling for Stripe API sub-hundred 'special case' currencies

### LINKED ISSUE #7363: 500 Error During Stripe Payments
We've had multiple reports of payments through Stripe failing to load/execute effectively. See User reports quoted at the end of this Issue.

One of the more promising user reports notes a 500 error returned by the `process-payment-intent` file - probably good to start investigating here.

image

From a [forum thread](https://forum.worldcubeassociation.org/t/trying-to-register-but-running-into-stripe-error/27472/2):

> When I try to register to a competition, Stripe payment not working only a spinner is rotating endlessly.
> I have got an email from the organizer, because others have the same issue.
> Looking deeper into issue, I think the problem is not on Stripe side, because the WCA server returns 500 error in the background, see the screenshot from developer console:

From "Budapest ReOpen 2022 Stripe Payment Issue" email thread

> We tried to use Stripe payments for Budapest ReOpen 2022, but the system crashed and failed to process even a single payment. Competitors were perfectly able to complete their WCA registrations (so the part where you pick events, add guests, etc.), but the screen kept loading forever after they added their payment information and wanted to send the registration fees.
> 
> As a competitor I have experienced similar issues during registering for Luxembourg Open 2022 a couple of months ago, so the problem doesn’t seem to be one-off case, rather a constantly occurring issue. I am quite convinced the issue happens because of the huge load of competitors trying to use Stripe in the same time (we had about 120 registrations in the first two minutes I believe). As Stripe is probably prepared to deal with 120 customers and the WCA registrations went fine as well, I suspect the connection between WCA and Stripe could be at fault. However this is only my non-programmer mind trying to find logic in the randomness, so don’t take it too seriously. Last time I faced the same issue for Luxembourg Open, it basically sorted itself out and even the organizers were surprised by my hackerman skills: I kept the page open and pressed the pay button in every five minutes until it finally went through.
> 
> So as you could see, the Stripe payment method on WCA is quite unreliable, and could crash more often I could accept as an organizer or a competitor. I’d also like to point out that many competitions set the order of the reg and waiting list based on the timestamp of the received payments, which is obviously difficult and potentially unfair if there’s an issue with Stripe. This is made even worse if you take into consideration the huge amount of people trying to register to basically every competition (at least in Europe), and know that registering just a couple of seconds late is enough to miss a competition.

gregorbg:
Fixes #7363 

See also https://stripe.com/docs/currencies#special-cases. As part of this PR, I have opened up the error reporting in the payments handler to catch all Stripe API-related errors and not just card errors, in the hopes that we can detect/diagnose problems in the future more easily.

### DIFF
--- a/WcaOnRails/app/controllers/registrations_controller.rb
+++ b/WcaOnRails/app/controllers/registrations_controller.rb
@@ -476,10 +476,14 @@ class RegistrationsController < ApplicationController
           render json: { error: { message: t("registrations.payment_form.alerts.amount_too_low") } }
           return
         end
+
+        currency_iso = registration.outstanding_entry_fees.currency.iso_code
+        stripe_amount = StripeCharge.amount_to_stripe(amount, currency_iso)
+
         payment_intent_args = {
           payment_method: params[:payment_method_id],
-          amount: amount,
-          currency: registration.outstanding_entry_fees.currency.iso_code,
+          amount: stripe_amount,
+          currency: currency_iso,
           confirmation_method: "manual",
           confirm: true,
           receipt_email: user.email,
@@ -582,10 +586,13 @@ class RegistrationsController < ApplicationController
       return
     end
 
+    currency_iso = registration.competition.currency_code
+    stripe_amount = StripeCharge.amount_to_stripe(refund_amount, currency_iso)
+
     refund = Stripe::Refund.create(
       {
         charge: payment.stripe_charge_id,
-        amount: refund_amount,
+        amount: stripe_amount,
       },
       stripe_account: registration.competition.connected_stripe_account_id,
     )
--- a/WcaOnRails/app/controllers/registrations_controller.rb
+++ b/WcaOnRails/app/controllers/registrations_controller.rb
@@ -515,7 +515,7 @@ class RegistrationsController < ApplicationController
           stripe_account: registration.competition.connected_stripe_account_id,
         )
       end
-    rescue Stripe::CardError => e
+    rescue Stripe::StripeError => e
       # Log and display error to client
       stripe_charge.update!(status: "failure", error: e.message)
       render json: { error: { message: e.message } }
--- a/WcaOnRails/app/models/stripe_charge.rb
+++ b/WcaOnRails/app/models/stripe_charge.rb
@@ -7,4 +7,29 @@ class StripeCharge < ApplicationRecord
     success: "success",
     failure: "failure",
   }
+
+  # as per https://stripe.com/docs/currencies#special-cases
+  ZERO_DECIMAL_CURRENCIES = %w[HUF TWD UGX].freeze
+
+  # Stripe has a small handful of fancy snowflake currencies
+  # that need to be submitted as multiples of 100. The details are documented
+  # at https://stripe.com/docs/currencies#special-cases
+  def self.amount_to_stripe(amount_lowest_denomination, iso_currency)
+    if ZERO_DECIMAL_CURRENCIES.include?(iso_currency)
+      amount_times_hundred = amount_lowest_denomination * 100
+
+      # Stripe API rejects payments that are below the hundreds sub-unit.
+      # In practice this should never happen because the inflation on those currencies
+      # makes it absolutely impractical for Delegates to charge 0.45 HUF for example.
+      # If this error is actually ever thrown, talk to the Delegates of the competition in question.
+      if amount_times_hundred % 100 != 0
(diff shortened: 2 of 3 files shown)


gregorbg:
@viroulep Git blame tells me that you have worked a lot on the Stripe integration before. Would you mind reviewing this PR based on the issue mentioned above, please?

Also, while I was going through the Stripe documentation to figure out what's going on, I found it very scary that we can ask competitors for their credit card details in the Stripe JS script and then use the `payment_method` ID in our backend to charge virtually anything we want without the user giving their explicit consent. This is dangerous for currencies like `MGA` (Malagasy Ariary) that have `subunit_to_unit` specified as `5` in RubyMoney but Stripe API lists them as zero subunit currency. As far as I can tell, there is no way to test what Stripe actually charges when sending `amount: 5000, currency: MGA` through the API in our playground with the `4242 4242 4242 4242` Visa. Did you previously put any thought into how to improve this situation?

viroulep:
I can but I probably will only be able to do this some time around Friday, hopefully that's not an issue :s

viroulep (approved):
These changes look great to me!

I definitely do not recall anything like that for HUF, it's so serious it must mean that either the behavior changed on Stripe side recently, or we simply never used that currency until recently :o

> As far as I can tell, there is no way to test what Stripe actually charges when sending amount: 5000, currency: MGA through the API in our playground with the 4242 4242 4242 4242 Visa. Did you previously put any thought into how to improve this situation?

I wouldn't bet currencies with "non 100" subunit just work currently.

Regarding the testing question: it's a bit blurry in my head but I think we do have a mean to see what Stripe actually charges, as I remember having access to a Stripe dashboard for the tests.
For tests we use a dev Stripe account id, to which the WST should have access to (checkout the doc with the WST accounts details, IIRC the details are in there); it means you can create a charge on your local copy of the website, then go to the account's dashboard and see exactly what Stripe does, and hopefully the actually amount charge is part of the details!

Regarding `MGA`: since I'm not sure about how we actually deal with such currency, I'd say take a quick look in the aforementioned dashboard and if it looks weird just mark the currency as unsupported until the support is actually implemented?

gregorbg:
Thanks a lot for your kind feedback :)

> For tests we use a dev Stripe account id, to which the WST should have access to (checkout the doc with the WST accounts details, IIRC the details are in there); it means you can create a charge on your local copy of the website, then go to the account's dashboard and see exactly what Stripe does, and hopefully the actually amount charge is part of the details!

That makes sense, and I guess the individual organisers can also use the dashboards of whatever account they have linked to the competition in "real world" scenarios. However, I still feel uncomfortable asking the user's credit card details in a way that (very technically!!) allows us to do whatever the hell we want with them. Like, we could charge them thousands of USD and if neither the organiser nor the competitor really cares, they will only notice it days or even weeks later. This is extremely bad IMO.

My preferred solution would be to implement the Stripe payment workflow in a way that we already pre-populate the "Please enter your credit card" window with an amount and currency, so that from Stripe's side they can display "Please enter your credit card to pay 1234 XYZ currency". That way, the user themselves can stop the process and alert us if we convert the currencies the wrong way.

As a matter of fact, Stripe itself [recommends](https://stripe.com/docs/payments/payment-element/migration) migrating to the `Payment` element, as opposed to the `Card` Element that we're using right now. Do you remember any reasoning behind using `Card` for the WCA payment workflow?

> Regarding MGA: since I'm not sure about how we actually deal with such currency, I'd say take a quick look in the aforementioned dashboard and if it looks weird just mark the currency as unsupported until the support is actually implemented?

That is a very smart and surprisingly simple solution, not sure why I hadn't thought of that myself :O

viroulep:
> As a matter of fact, Stripe itself [recommends](https://stripe.com/docs/payments/payment-element/migration) migrating to the Payment element, as opposed to the Card Element that we're using right now. Do you remember any reasoning behind using Card for the WCA payment workflow?

I can't think of anything besides "it was not recommended at the time of writing the code"!
I mean when I implemented the update in the Stripe component we were using there was one main concern: that the WCA server would actually make the charge, so that it's easy to know when something happens.
It lead me to using the `PaymentIntent` system to be able to have a payment intent created on stripe side, while still being able to create the actual charge from the WCA  server; apart from that I simply followed the guide on using the "Elements" js API to create the UI.

Even though it could have happened, I doubt I would have missed it if the documentation at the time was just like the documentation today: there is no doubt using the `Payment` element looks far better!

gregorbg:
Then we shall put that very high on our priority list. Thank you for the historical context! 😄

PR #7371, opened 2022-11-08
fix: restore checkbox functionality for `/competitions`

### LINKED ISSUE #7369: Checkboxes on the /competitions page don't work
**Describe the bug**
The list of competitions on `https://www.worldcubeassociation.org/competitions` doesn't update based on the checkboxes 'Show registration status icon' and 'Show cancelled competitions'.

**To Reproduce**
1. Go to `https://www.worldcubeassociation.org/competitions`.
2. Click on 'Show registration status icon' or 'Show cancelled competitions'.
3. See the bug - the list of competitions doesn't change.

**Expected behavior**
The list of competitions should be updated based on the checkboxes.

**Screenshots**
image

Doesn't work on several devices with several locales on the website.

AnnikaStein:
Apparently, it's not safe to invent an arbitrary name for the checkboxes & labels, as was introduced in https://github.com/thewca/worldcubeassociation.org/pull/7304 - likely because the original name (as is) is used in a controller (?). Sorry for that.

Removed the expression around the names in quotation marks.

Fixes #7369 

Screenshots:
Bildschirmfoto 2022-11-08 um 15 10 09
Bildschirmfoto 2022-11-08 um 15 09 49

@Nevseros thanks Ekaterina for reporting the issue originally mentioned by [Sehoon Kim (김세훈)](https://forum.worldcubeassociation.org/u/Sehoon_Kim)
[(2022KIMS02)](https://www.worldcubeassociation.org/persons/2022KIMS02)

### DIFF
--- a/WcaOnRails/app/views/competitions/_index_form.html.erb
+++ b/WcaOnRails/app/views/competitions/_index_form.html.erb
@@ -114,14 +114,14 @@
 </div>
 
 <div id="registration-status" class="form-group registration-status-selector">
-  <%= check_box_tag("#{:show_registration_status}-checkbox", params[:show_registration_status], params[:show_registration_status] == "on") %>
-  <%= label_tag(name = "#{:show_registration_status}-checkbox",
+  <%= check_box_tag(:show_registration_status, params[:show_registration_status], params[:show_registration_status] == "on") %>
+  <%= label_tag(name = :show_registration_status,
                 content_or_options = t('competitions.index.show_registration_status')) %>
 </div>
 
 <div id="cancelled" class="form-group cancel-selector">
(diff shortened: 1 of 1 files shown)


gregorbg:
For context: I should have caught this during my review on your initial PR, sorry for missing it. The checkbox is part of an HTML form that's being encoded into the URL. The controller at https://github.com/thewca/worldcubeassociation.org/blob/master/WcaOnRails/app/controllers/competitions_controller.rb#L106 expects the parameter to be named in a certain way, so either you fix the ID that the checkbox sends or you change the exoected ID that the controller receives. Your current solution is fine.

gregorbg (approved):

PR #7372, opened 2022-11-08
Fixed order of team members

revers3ntropy:
Apparently slight differences in the way V8 and SpiderMonkey have implemented `Array.prototype.sort` meant that in Firefox the members of teams on `/teams-committees` were being shown out of order. This fixes the ambiguity.

Also changed `p` to `div` to fix console error.


gregorbg (approved):
Oh, interesting quirk! Thanks for figuring that one out.

PR #7376, opened 2022-11-11
More validations for advancement conditions

### LINKED ISSUE #5587: Result validations for advancement conditions
We (WCA Results Team) were fixing the inconsistencies in the existing advancement conditions. To avoid many inconsistencies in future, it would be better if there is result validation for advancement conditions while submitting results.

According to the discussion within WRT, the following cases can be marked as error:
- actual advanced number > planned advanced number
- actual advanced number or planned advanced number > 75%
- cases where advancement_condition's type is `attemptResult` and it is wrongly applied

Also, it can be marked as a warning if both the following condition are true:
- (actual advanced number / planned advanced number) <= 0.8
- planned advanced number - actual advanced number >= 3

EDIT: Added the third case for error.

viroulep:
Fixes #5587.

This implements a couple more errors and warning regarding the advancement conditions declared in the events/round data, here is a WIP list:
  - [x] error when the announced advancement condition would let more than 75% qualify.
  - [x] error when more people qualified than what was initially planned.
  - [x] error when using an attempt result qualification and people over that result qualify
  - [x] warning when an unusual number of competitors is missing compared to the expected number of qualified competitors.
  - [x] tests

This is just a draft for now, I hope to finish this soon!

### DIFF
--- a/WcaOnRails/lib/results_validators/advancement_conditions_validator.rb
+++ b/WcaOnRails/lib/results_validators/advancement_conditions_validator.rb
@@ -2,12 +2,22 @@
 
 module ResultsValidators
   class AdvancementConditionsValidator < GenericValidator
-    REGULATION_9M_ERROR = "Event %{event_id} has more than four rounds, which is not permitted as per Regulation 9m."
+    # PV: There used to be an error for regulation 9M, but now all results
+    # must belong to a round, and rails validations make sure the total
+    # number of rounds is <= 4, and that each round number is unique.
     REGULATION_9M1_ERROR = "Round %{round_id} has 99 or fewer competitors but has more than two subsequent rounds, which is not permitted as per Regulation 9m1."
     REGULATION_9M2_ERROR = "Round %{round_id} has 15 or fewer competitors but has more than one subsequent round, which is not permitted as per Regulation 9m2."
     REGULATION_9M3_ERROR = "Round %{round_id} has 7 or fewer competitors but has at least one subsequent round, which is not permitted as per Regulation 9m3."
     REGULATION_9P1_ERROR = "Round %{round_id}: Fewer than 25%% of competitors were eliminated, which is not permitted as per Regulation 9p1."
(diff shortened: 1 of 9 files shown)


viroulep:
I've finally finished this pull request, feel free to take a look :)

gregorbg (approved):

WcaOnRails/lib/results_validators/advancement_conditions_validator.rb
```diff
@@ -85,16 +112,43 @@ def validate(competition_ids: [], model: Result, results: nil)
...
+                  end
+                  if theoritical_number_of_people > max_advancing
+                    @errors << ValidationError.new(:rounds, competition_id,
+                                                   ROUND_9P1_ERROR,
+                                                   round_id: round_id,
+                                                   condition: condition.to_s(previous_round))
+                  end
+                  # This comes from https://github.com/thewca/worldcubeassociation.org/issues/5587
+                  if theoritical_number_of_people - number_of_people_in_round >= 3 &&
+                     (number_of_people_in_round / theoritical_number_of_people) <= 0.8
```
timreyn, line 141: I noticed this warning while submitting CubingUSANationals2023:

> Round 333-2: according to the events data, at most 400 could have proceed, but only 395 competed in the round. Please leave a comment about that (or fix the events data if you applied a different advancement condition).
> Round 555-2: according to the events data, at most 100 could have proceed, but only 96 competed in the round. Please leave a comment about that (or fix the events data if you applied a different advancement condition).
> Round pyram-2: according to the events data, at most 100 could have proceed, but only 96 competed in the round. Please leave a comment about that (or fix the events data if you applied a different advancement condition).
> Round skewb-2: according to the events data, at most 100 could have proceed, but only 96 competed in the round. Please leave a comment about that (or fix the events data if you applied a different advancement condition).

I think this isn't actually doing the second check because of integer division -- the left side of the equation is always 0 if number_of_people_in_round < theoretical_number_of_people.

Issue #7143, opened 2022-08-15
Dues calculation no longer takes into account the minimum due amount


Jambrose777:
**Describe the bug**
The dues estimation on WCA when creating a competition no longer takes into account the minimum due amount (for US it is currently $1). This causes the estimation to be wrong: 

<img width="553" alt="image" src="https://user-images.githubusercontent.com/11577241/184677603-e6de55d4-1b26-4487-b3ba-8c67cacdfcf0.png">

gregorbg:
For context: When you say "no longer", can you confirm that it was correct at some point in the past? Just wondering because we haven't really touched that part of the code in a while.

Jambrose777:
I am 90% certain it was correct when it was implemented, but not certain.

notrichardpeng:
Hi Jacob, I'm trying to replicate this locally but on my end it took the minimum amount into account:

![image](https://user-images.githubusercontent.com/58278719/186979920-5b42a8e0-c608-4948-8fbe-b2a380ec4481.png)

I imagine the comp is GA Cubers Turkey Gobblers 2022? Sorry if this was creepy but I logged into your dummy account on dev to see where exactly the issue is :P. But here it appeared fine.

Jambrose777:
Just looked at it. It is correct in production now.

I did repro this by doing the following:
Competitions > New competition
Fill in country as United States
Fill in competitor limit (100)
Fill in base registration fee ($5)
Notice it does not account for the minimum.

If I had to guess its not taking into account the country when first creating the forum.

notrichardpeng:
I see now. Yeah after hitting "create competition" it becomes correct. I think adding a callback for when country changes can fix this.

notrichardpeng:
Actually it wouldn't be that easy; getting the country band needs the backend function `CountryBand.where`, so I don't think it can update dynamically with frontend changes. Maybe the fix here is to just add a note saying something along the lines of "hit Update Competition after changing the country" or "Refresh the page after indicating the country to make estimation accurate"?

gregorbg:
I think this should be possible without saving. Note that fields like the "Country" dropdown or the "Currency code" selector can all be dynamically read and evaluated through jQuery.

[...]

PR #7377, opened 2022-11-11
Restructuring Dues Estimation

notrichardpeng:
#7143

Looks much cleaner now! Avoided some bugs and I think it's also fully dynamic (changing anything that impacts dues calculation on the page would reflect on the display). Only tested its calculations several times, not extensively.

### DIFF
--- a/WcaOnRails/app/controllers/competitions_controller.rb
+++ b/WcaOnRails/app/controllers/competitions_controller.rb
@@ -393,7 +393,7 @@ class CompetitionsController < ApplicationController
   end
 
   def calculate_dues
-    country_iso2 = Country.find(params[:country_id]).iso2
+    country_iso2 = Country.find(params[:country_id])&.iso2
     country_band = CountryBand.find_by(iso2: country_iso2)&.number
 
     input_money_us_dollars = Money.new(params[:entry_fee_cents].to_i, params[:currency_code]).exchange_to("USD").amount
--- a/WcaOnRails/app/controllers/competitions_controller.rb
+++ b/WcaOnRails/app/controllers/competitions_controller.rb
@@ -393,7 +393,7 @@ class CompetitionsController < ApplicationController
   end
 
   def calculate_dues
-    country_iso2 = Country.find(params[:country_id])&.iso2
+    country_iso2 = Country.find_by(id: params[:country_id])&.iso2
     country_band = CountryBand.find_by(iso2: country_iso2)&.number
 
     input_money_us_dollars = Money.new(params[:entry_fee_cents].to_i, params[:currency_code]).exchange_to("USD").amount
--- a/WcaOnRails/app/controllers/competitions_controller.rb
+++ b/WcaOnRails/app/controllers/competitions_controller.rb
@@ -404,19 +404,21 @@ class CompetitionsController < ApplicationController
 
     # times 100 because later currency conversions require lowest currency subunit, which is cents for USD
     price_per_competitor_us_cents = [registration_fee_dues_us_dollars, country_band_dues_us_dollars].compact.max * 100
-    estimated_dues = "" # return empty string if no competitor limit
 
     if params[:competitor_limit_enabled]
       estimated_dues_us_cents = price_per_competitor_us_cents * params[:competitor_limit].to_i
       estimated_dues = Money.new(estimated_dues_us_cents, "USD").exchange_to(params[:currency_code]).format
+
+      render json: {
+        dues_value: estimated_dues,
+      }
+    else
+      price_per_competitor = Money.new(price_per_competitor_us_cents, "USD").exchange_to(params[:currency_code]).format
+
+      render json: {
+        dues_value: price_per_competitor,
+      }
     end
-
-    price_per_competitor = Money.new(price_per_competitor_us_cents, "USD").exchange_to(params[:currency_code]).format
-
-    render json: {
-      per_competitor: price_per_competitor,
-      dues: estimated_dues,
-    }
   end
 
   def show
--- a/WcaOnRails/app/controllers/competitions_controller.rb
+++ b/WcaOnRails/app/controllers/competitions_controller.rb
@@ -396,7 +396,9 @@ class CompetitionsController < ApplicationController
     country_iso2 = Country.find_by(id: params[:country_id])&.iso2
     country_band = CountryBand.find_by(iso2: country_iso2)&.number
 
+    update_exchange_rates_if_needed
     input_money_us_dollars = Money.new(params[:entry_fee_cents].to_i, params[:currency_code]).exchange_to("USD").amount
+
     registration_fee_dues_us_dollars = input_money_us_dollars * CountryBand::PERCENT_REGISTRATION_FEE_USED_FOR_DUE_AMOUNT
     country_band_dues_us_dollars = CountryBand::BANDS[country_band][:value] if country_band.present? && country_band > 0
 
--- a/WcaOnRails/app/controllers/competitions_controller.rb
+++ b/WcaOnRails/app/controllers/competitions_controller.rb
@@ -397,7 +397,7 @@ class CompetitionsController < ApplicationController
     country_band = CountryBand.find_by(iso2: country_iso2)&.number
 
     input_money_us_dollars = Money.new(params[:entry_fee_cents].to_i, params[:currency_code]).exchange_to("USD").amount
-    registration_fee_dues_us_dollars = input_money_us_dollars * 0.15
+    registration_fee_dues_us_dollars = input_money_us_dollars * CountryBand::PERCENT_REGISTRATION_FEE_USED_FOR_DUE_AMOUNT
     country_band_dues_us_dollars = CountryBand::BANDS[country_band][:value] if country_band.present? && country_band > 0
 
     # times 100 because later currency conversions require lowest currency subunit, which is cents for USD
--- a/WcaOnRails/app/controllers/competitions_controller.rb
+++ b/WcaOnRails/app/controllers/competitions_controller.rb
@@ -393,30 +393,23 @@ class CompetitionsController < ApplicationController
   end
 
   def calculate_dues
-    country_iso2 = Country.where(id: params[:country_id]).pluck(:iso2)[0]
-    country_band = CountryBand.where(iso2: country_iso2).map(&:number)[0]
+    country_iso2 = Country.find(params[:country_id]).iso2
+    country_band = CountryBand.find_by(iso2: country_iso2)&.number
 
-    update_currency_rates_if_needed
-    currency_subunit = Money::Currency.new(params[:currency_code]).subunit_to_unit
-    money_value = params[:entry_fee].to_i / currency_subunit
-    money_value_usd = Money.new(money_value, params[:currency_code]).exchange_to("USD").cents
-
-    base_fee_usd = money_value_usd * 0.15
-    band_fee_usd = 0
-    if country_band.present? && country_band > 0
-      band_fee_usd = CountryBand::BANDS[country_band][:value]
-    end
+    input_money_us_dollars = Money.new(params[:entry_fee_cents].to_i, params[:currency_code]).exchange_to("USD").amount
+    registration_fee_dues_us_dollars = input_money_us_dollars * 0.15
+    country_band_dues_us_dollars = CountryBand::BANDS[country_band][:value] if country_band.present? && country_band > 0
 
     # times 100 because later currency conversions require lowest currency subunit, which is cents for USD
-    price_per_competitor_usd = [base_fee_usd, band_fee_usd].max * 100
+    price_per_competitor_us_cents = [registration_fee_dues_us_dollars, country_band_dues_us_dollars].compact.max * 100
     estimated_dues = "" # return empty string if no competitor limit
 
     if params[:competitor_limit_enabled]
-      estimated_dues_usd = price_per_competitor_usd * params[:competitor_limit].to_i
-      estimated_dues = Money.new(estimated_dues_usd, "USD").exchange_to(params[:currency_code]).format
+      estimated_dues_us_cents = price_per_competitor_us_cents * params[:competitor_limit].to_i
+      estimated_dues = Money.new(estimated_dues_us_cents, "USD").exchange_to(params[:currency_code]).format
     end
 
-    price_per_competitor = Money.new(price_per_competitor_usd, "USD").exchange_to(params[:currency_code]).format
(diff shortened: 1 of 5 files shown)


WcaOnRails/app/controllers/competitions_controller.rb
```diff
@@ -582,7 +605,7 @@ def for_senior
     params[:commit] == "Confirm"
   end
 
-  private def update_rates_if_needed
+  private def update_currency_rates_if_needed
```
gregorbg, line 608: `update_exchange_rates_if_needed`

WcaOnRails/app/views/competitions/_competition_form.html.erb
```diff
@@ -261,87 +261,51 @@
...
-            var compLimit = $('select[name="competition[competitor_limit_enabled]"]').val() === "true"
-              ? $('input[name="competition[competitor_limit]"]').val() : null;
-            window.wca.cancelPendingAjaxAndAjax('currency_convert', {
-              url: '<%= currency_convert_path %>',
+
+            var compLimitEnabled = $('select[name="competition[competitor_limit_enabled]"]').val();
+            var compLimit = compLimitEnabled === "true" ? $('input[name="competition[competitor_limit]"]').val() : null;
+            var countryId = $('select[name="competition[countryId]"]').val()
+            var currencyCode = $('select[name="competition[currency_code]"]').val()
+            var entryFee = $('input[name="competition[base_entry_fee_lowest_denomination]"]').val();
```
gregorbg, line 270: Please take a look at how the big code block towards the top of the file handles serialization in the `handleAdjacentCompetitionAjax` method.

In particular, you can encode the entire form via `var obj = $("#competition-form").serializeJSON();` and then only access the parts of the encoded object that you need (again, see above for reference). Saves a ton of `$(selector).val()` calls.

WcaOnRails/app/views/competitions/_competition_form.html.erb
```diff
@@ -261,87 +261,51 @@
...
-                value: compLimit === null ? pricePerCompetitor : pricePerCompetitor * compLimit,
-                from_currency: "USD",
-                to_currency: "<%= @competition.currency_code %>",
+                competitor_limit_enabled: compLimitEnabled,
+                competitor_limit: compLimit,
+                country_id: countryId,
+                currency_code: currencyCode,
+                entry_fee: entryFee,
               },
               success: function(data) {
```
gregorbg, line 281: I'm not too wild about the entire approach of dynamically computing frontend strings to inject here.
I know this page will be migrated to React soon, but in the meantime your approach will stay around for a few weeks so I'd like to suggest the following alternative:
1. Create a small HTML snippet that is only responsible for showing the Dues estimate textbox in this folder. Be sure to name it with an underscore in front, otherwise the entire page layout will be injected.
2. Have that template do essentially what you're doing now, but as a "real" Rails template and not some "glue and stick stuff together with JS".
3. Have the currency endpoint return that fragment with `render partial: 'your_new_snippet'`
4. Inject the fully rendered HTML fragment into the DOM as-is.

It will change error handling a bit, in that Rails is responsible for showing error strings if anything goes wrong, and then the actual jQuery `error` function below should resort to making the background of the text turn red or something really rudimentary, because that function only concerns real technical errors (like 500 responses) that should only ever happen in 1 / 1000 cases anyways.

See the current `series_eligible_competitions` and `nearby_competitions` controller methods (and corresponding view fragments) for reference. Your view fragment here would be significantly less complex, though.

Taking this approach will also save us from the "i18n within JS that needs comments to be detected by our test suite checks" pain.

WcaOnRails/app/controllers/competitions_controller.rb
```diff
@@ -392,12 +392,35 @@ def time_until_competition
...
-    update_rates_if_needed
-    converted = Money.new(params[:value], params[:from_currency]).exchange_to(params[:to_currency])
+  def calculate_dues
+    country_iso2 = Country.where(id: params[:country_id]).pluck(:iso2)[0]
+    country_band = CountryBand.where(iso2: country_iso2).map(&:number)[0]
+
+    update_currency_rates_if_needed
+    currency_subunit = Money::Currency.new(params[:currency_code]).subunit_to_unit
+    money_value = params[:entry_fee].to_i / currency_subunit
+    money_value_usd = Money.new(money_value, params[:currency_code]).exchange_to("USD").cents
```
gregorbg, line 402: There is `Money.from_amount(human_entered_amount, currency_iso)` which completely saves you from reading the currency and parsing the subunit above.

WcaOnRails/app/controllers/competitions_controller.rb
```diff
@@ -392,12 +392,35 @@ def time_until_competition
     }
   end
 
-  def currency_convert
-    update_rates_if_needed
-    converted = Money.new(params[:value], params[:from_currency]).exchange_to(params[:to_currency])
+  def calculate_dues
+    country_iso2 = Country.where(id: params[:country_id]).pluck(:iso2)[0]
```
gregorbg, line 396: `Country.find(params[:country_id]).iso2`

WcaOnRails/app/controllers/competitions_controller.rb
```diff
@@ -392,12 +392,35 @@ def time_until_competition
...
+    country_band = CountryBand.where(iso2: country_iso2).map(&:number)[0]
```
gregorbg, line 397: `CountryBand.find_by(iso2: country_iso2)&.number`

gregorbg: Note the `&` as nil-safe operator if no matching country band is found.

WcaOnRails/app/controllers/competitions_controller.rb
```diff
@@ -392,12 +392,35 @@ def time_until_competition
...
+
+    update_currency_rates_if_needed
+    currency_subunit = Money::Currency.new(params[:currency_code]).subunit_to_unit
+    money_value = params[:entry_fee].to_i / currency_subunit
+    money_value_usd = Money.new(money_value, params[:currency_code]).exchange_to("USD").cents
+
+    base_fee_usd = money_value_usd * 0.15
```
gregorbg, line 404: Magic number is bad. Where does it come from? What policy is it based upon? (Those are rhetorical questions).
Please move this to a constant in `CountryBands` (or another model class that you deem more conveninent) and document it thoroughly!

WcaOnRails/app/controllers/competitions_controller.rb
```diff
@@ -392,12 +392,35 @@ def time_until_competition
...
+    band_fee_usd = 0
+    if country_band.present? && country_band > 0
+      band_fee_usd = CountryBand::BANDS[country_band][:value]
```
gregorbg, line 407: `band_fee_usd = foo_bar if country_band.present? && country_band > 0`

WcaOnRails/app/controllers/competitions_controller.rb
```diff
@@ -392,12 +392,35 @@ def time_until_competition
...
+    country_iso2 = Country.where(id: params[:country_id]).pluck(:iso2)[0]
+    country_band = CountryBand.where(iso2: country_iso2).map(&:number)[0]
+
+    update_currency_rates_if_needed
+    currency_subunit = Money::Currency.new(params[:currency_code]).subunit_to_unit
+    money_value = params[:entry_fee].to_i / currency_subunit
+    money_value_usd = Money.new(money_value, params[:currency_code]).exchange_to("USD").cents
+
+    base_fee_usd = money_value_usd * 0.15
+    band_fee_usd = 0
```
gregorbg, line 405: Variable names are a bit tricky here. I had to pay _really_ close attention to realise that one is ba*nd* fee and the other is ba*se* fee. both four letters, both with the exact same suffix. Can you come up with something more discernible?

WcaOnRails/app/controllers/competitions_controller.rb
```diff
@@ -392,12 +392,35 @@ def time_until_competition
...
+    if country_band.present? && country_band > 0
+      band_fee_usd = CountryBand::BANDS[country_band][:value]
+    end
+
+    # times 100 because later currency conversions require lowest currency subunit, which is cents for USD
+    price_per_competitor_usd = [base_fee_usd, band_fee_usd].max * 100
```
gregorbg, line 411: Using my "inline if" suggestion from above, you can change this to `[...].compact.max` where the `compact` is Ruby's name for "filter out nil values".

WcaOnRails/app/controllers/competitions_controller.rb
```diff
@@ -392,12 +392,35 @@ def time_until_competition
...
+    estimated_dues = "" # return empty string if no competitor limit
```
gregorbg, line 412: In line with my longer comment about HTML fragments below: Here you could do an if-else case to decide which HTML fragment to render in the first place. That way, you don't throw around unneeded variables.

WcaOnRails/app/controllers/competitions_controller.rb
```diff
@@ -392,12 +392,35 @@ def time_until_competition
...
+    money_value = params[:entry_fee].to_i / currency_subunit
+    money_value_usd = Money.new(money_value, params[:currency_code]).exchange_to("USD").cents
+
+    base_fee_usd = money_value_usd * 0.15
+    band_fee_usd = 0
+    if country_band.present? && country_band > 0
+      band_fee_usd = CountryBand::BANDS[country_band][:value]
+    end
+
+    # times 100 because later currency conversions require lowest currency subunit, which is cents for USD
```
gregorbg, line 410: General remark about money variables (didn't want to bother marking the entire method body): Please pick variable names that make clear whether you're currently handling a `Money` object-oriented model or an integer representing cents.

For example. the below calculation result should be called something like `price_per_competitor_us_cents`. Similar remarks apply for other variables throughout this method body.

gregorbg (changes requested):
Thanks, this is exactly the kind of rewrite that I was aiming for! Please don't be intimidated by the number of comments.

For the Ruby parts, please take my remarks into consideration directly.

For the JS parts, and especially the jQuery AJAXing, I am a bit torn myself. I know that this will be replaced by React soon, so it seems a bit futile to invest time to make it prettier. But on the other hand, I want you to develop an eye for clean code and practices that are consistent with the rest of our codebase (wherever possible). So please see through my remarks (especially the one lengthy monologue about injecting HTML fragments) and try them. If you can't make sense of my explanations, reach out on Slack 😄

WcaOnRails/app/controllers/competitions_controller.rb
notrichardpeng: I don't see how this function would help...we still need to divide `params[:entry_fee]` by the subunit to unit ratio because, in USD for example, there is always two zeroes attached behind every amount entered (because we read it from HTML). If the user entered 20 then `params[:entry_fee]` would be 2000 and `from_amount(2000, "USD")` would consider the input to be 2000 dollars and not 20 dollars. Maybe `Money` has a function like `from_cents` that would make everything easier?  No luck finding it for now...

notrichardpeng: Guess what, the constructor takes in the fractional unit, so Money.new(entered_amount, currency).exchange_to("USD") is all we need! lol

WcaOnRails/app/controllers/competitions_controller.rb
notrichardpeng: added to `CountryBands` for convenience, but I'm thinking if this is bad practice because the constant doesn't really have much to do with countries or bands.

gregorbg: It's "less bad" practice than having the number `0.15` randomly appear as integer literal in your code

WcaOnRails/app/views/competitions/_competition_form.html.erb
```diff
@@ -261,87 +261,52 @@
...
+              url: '<%= calculate_dues_path %>',
               data: {
-                value: compLimit === null ? pricePerCompetitor : pricePerCompetitor * compLimit,
-                from_currency: "USD",
-                to_currency: "<%= @competition.currency_code %>",
+                competitor_limit_enabled: compLimitEnabled,
+                competitor_limit: compLimit,
+                country_id: countryId,
+                currency_code: currencyCode,
+                entry_fee_cents: entryFee,
```
gregorbg, line 281: I think you can inline all of these variables like so:
```js
data: {
  competitor_limit_enabled: obj['competition[competitor_limit_enabled]']
  etc...
}
```

WcaOnRails/app/controllers/competitions_controller.rb
```diff
@@ -392,12 +392,30 @@ def time_until_competition
...
+    country_band_dues_us_dollars = CountryBand::BANDS[country_band][:value] if country_band.present? && country_band > 0
+
+    # times 100 because later currency conversions require lowest currency subunit, which is cents for USD
+    price_per_competitor_us_cents = [registration_fee_dues_us_dollars, country_band_dues_us_dollars].compact.max * 100
+    estimated_dues = "" # return empty string if no competitor limit
+
+    if params[:competitor_limit_enabled]
+      estimated_dues_us_cents = price_per_competitor_us_cents * params[:competitor_limit].to_i
+      estimated_dues = Money.new(estimated_dues_us_cents, "USD").exchange_to(params[:currency_code]).format
+    end
```
gregorbg, line 412: Suggestion to avoid empty variables:
```ruby
if params[:competitor_limit_enabled]
  estimated_dues_us_cents = foo
  estimated_dues = bar
  
  return render json: {
    stuff
  } 
else
  price_per_competitor = foo
  
  return render json: {
    stuff
  }
end
```

WcaOnRails/app/views/competitions/_competition_form.html.erb
gregorbg: Closing this because the current approach is better suited for the upcoming React migration

gregorbg (changes requested):

WcaOnRails/app/views/competitions/_competition_form.html.erb
notrichardpeng: `compLimitEnabled`, `compLimit`, and `currencyCode` are used later in the HTML injection. Is it better to keep those variables or reference them again through `obj[...]`?

gregorbg: That's very debatable and mostly depends on taste. I'm fine with the current solution as it hopefully won't stick around for long.

gregorbg (approved):

PR #7379, opened 2022-11-13
Implement asynchronous Stripe `PaymentElement` workflow

gregorbg:
I have taken a shot at implementing the `PaymentElement` workflow described by Stripe at https://stripe.com/docs/payments/quickstart. In short, we create one PI, then let Stripe "do its thing", and then only once _everything_ has been dealt with on Stripe's side, the user is redirected to a URL that we can choose.

- The central, relevant entity is called `PaymentIntent`. It encapsulates our desire for the user to pay XY currency.
- This `PaymentIntent` (PI) is passed on to the frontend via a so-called `client_secret`. Stripe docs are a bit vague on how "secret" this actually is (you can retrieve it easily as long as you know the Stripe `acct_*` account under which the original PR was created) but I encrypted it just to be sure.
- Stripe does its own thing with the PI, including 3DS and SCA and other finance security regulations.
- Only at the very end we are redirected to a `redirect_url` that we can freely specify upon passing off the PI to Stripe's frontend.
- All events that happen "in the background" while the PI is at Stripe are relayed to our server via one central Webhook: https://stripe.com/docs/webhooks/quickstart

This is an improvement over the current situation in the following ways:
- We are not limited to one payment method any more.
- We are not responsible for managing the payment method ourselves (includes detecting SCA requirements and stuff)
- The payment logic is fully asynchronous, potentially allowing us to create our own abstraction along the lines of "Registration PI" (where `PaymentIntent` is really just a very convenient name and has nothing to do with the Stripe model) that we can pass off to a third party (along with a redirect_url) and let them do their thing. This will allow us to let ROs implement their own payment services (like recently seen with WC2023 Korea "Toss Payments") and hook them back into the main website registration.

In order to make this change possible, the following high-level changes have been made: (along with the minor but substantial groundwork in #7740!)
- Introduce a `stripe_payment_intents` model and table. Here, we basically record every time that a user has clicked the "Pay now" button. (This can also be extended to a more abstract `registration_payment_intents` table in the future, as per my remark above)
- Addn an endpoint under `/registration/stripe-webhook` to receive Stripe webhook events. Right now, it only cares about the completion of a PI, but this can be easily extended in the future by adding more `case` branches to the `when` statement. **Important: ALL Webhook events are being logged in their own, separate journal table. So if anything ever goes wrong, we have full paranoid logs.**
- `PaymentIntent`s now have their own "receipts" for confirmation, analogous to the receipts introduced to `registration_payments` in #7740. This will allow us to track why our code decided to mark an Intent as complete.

## Closing remarks
How should we handle asynchronous payment methods like SEPA bank transfers? These things can take up to 48 hours, should we consider the registration "paid" when the user has completed the Stripe form or only when Stripe reports that the money has been fully transferred?
--> The current state of the PR mimics the behavior of the current "credit card only" implementation: We only register a `RegistrationPayment` on our side once Stripe sends the "PI completed" event. From experience with the current system, this happens more or less instantly with credit cards (barring completion of SCA authentication), but with the new system and other payment providers this _could_ be significantly later than the initial payment UI has been completed. Are we happy with this?

There are also a small handful of `TODO` remarks in the frontend JS section that I would like feedback on. If you have ideas how to improve the readability of my JS code or make it more consistent in general, don't hesitate to criticise - I don't have a lot of experience in that arena.


WcaOnRails/app/views/registrations/_payment_form.html.erb
```diff
@@ -1,128 +1,125 @@
...
-    var cardElement = elements.create('card');
-    cardElement.mount('#card-element');
-    var cardholderName = document.getElementById('cardholder-name');
-    var $cardButton = $('#card-button');
-
-    var toggleSaving = function(saving) {
-      $cardButton.attr("disabled", saving);
-      $cardButton.toggleClass("saving", saving);
-    }
+    var $stripeButton = $('#stripe-button');
```
FinnIckler, line 36: Do we want to continue using jQuery for this? If we want to get rid of this I think we should be using document.querySelector instead

WcaOnRails/app/views/registrations/_payment_form.html.erb
```diff
@@ -1,128 +1,125 @@
...
-        });
-      } else {
-        // We got a successful payment, redirect to the registration page with a
-        // nice flash message.
-        window.location.replace('<%= competition_payment_success_path(@competition) %>');
-      }
+    var elements;
+
+    function toggleSaving(saving) {
+      $paymentButton.attr("disabled", saving);
```
FinnIckler, line 42: same here, this can be done with vanilla js. paymentButton.setAttribute

WcaOnRails/app/views/registrations/_payment_form.html.erb
```diff
@@ -1,128 +1,125 @@
...
+      $paymentButton.toggleClass("saving", saving);
     }
 
-    $cardButton.on('click', function(ev) {
-      ev.preventDefault();
-      toggleSaving(true);
+    // We need to ask the user to explicitly start the checkout process, because they may or may not enter
+    // a donation amount that needs to be included in the "amount" computation we send over to Stripe.
+    // TODO if anybody has smarter ideas how to handle the donation amount, feel free to change this!
+    $stripeButton.on('click', function(e) {
```
FinnIckler, line 49: `.addEventListener( (e) => stripeButton.classList.toggle("hidden") ...`

WcaOnRails/app/views/registrations/_payment_form.html.erb
```diff
@@ -1,128 +1,125 @@
...
+      });
+
+      var $paymentElement = $('#group-payment-element')
+      $paymentElement.removeClass("has-error");
+
+      if (error) {
+        var $paymentError = $('#payment-error');
+
+        if (error.type === 'card_error' || error.type === 'validation_error') {
+          $paymentElement.addClass("has-error");
```
FinnIckler, line 113: `paymentElement.classList.add("has-error")`

FinnIckler (commented):
I added some vanilla js options to not introduce more jquery

gregorbg:
I have decided to remove the test about SCA altogether. Besides the fact that validating nodes inside an iframe inside an iframe inside an iframe is ridiculous, the execution of SCA is simply not our responsibility anymore.

In the previous implementation, it was important to have this test because we (our server) had to handle triggering and validating SCA flows by itself. In the new implementation, the responsibility of handling _any_ third party authentication (whether it be credit card SCA or some SEPA mandate stuff or third party Klarna invoicing) is up to Stripe.

Reinstating these tests in the current framework would effectively boil down to "checking that Stripe correctly invokes our return URL" which is silly because we don't have to test Stripe's correctness.

WcaOnRails/app/controllers/registrations_controller.rb
```diff
@@ -439,163 +441,246 @@ def register
...
+      stored_intent = stored_transaction.stripe_payment_intent
+      stored_intent.update_status_and_charges(stripe_intent, audit_event)
+    else
+      logger.info "Unhandled Stripe event type: #{event.type}"
+    end
+
+    head :ok
+  end
+
   def payment_success
```
dunkOnIT, line 521: Should we use a different name - eg, "payment_completion"? The name could imply that the payment itself was successful - when in fact this functionn seems to render all possible payment state (

dunkOnIT:
In order to test this PR locally, we will need to do the following:
* Clone Gregor's PR (recommended using Github cli: `gh pr checkout 7379`)
* Pick any competition
* Ensure competition has entrance fee (Edit -> Admin view -> Base registration fee)
* Connect Stripe for testing: NO STRIPE LOGIN NEEDED (Edit -> Setup payments -> "Connect Stripe" -> "Skip this form")

Testing data available here: https://stripe.com/docs/testing

WcaOnRails/app/controllers/registrations_controller.rb
```diff
@@ -439,163 +441,246 @@ def register
...
+      # Invalid payload
+      logger.warn "Stripe webhook error while parsing basic request. #{e.message}"
+      return head :bad_request
+    end
+    # Check if webhook signing is configured.
+    if EnvVars.STRIPE_WEBHOOK_SECRET.present?
+      # Retrieve the event by verifying the signature using the raw body and secret.
+      signature = request.env['HTTP_STRIPE_SIGNATURE']
+      begin
+        event = Stripe::Webhook.construct_event(
```
Rubiks-boy, line 463: nice!

WcaOnRails/spec/requests/registrations_spec.rb
```diff
@@ -580,133 +584,256 @@
...
+          # mimic the response that Stripe sends to our return_url after completing the checkout UI
+          get registration_payment_success_path(registration.id), params: {
+            payment_intent: payment_intent.stripe_id,
+            payment_intent_client_secret: payment_intent.client_secret,
+          }
+
           expect(registration.reload.outstanding_entry_fees.cents).to eq(-donation_lowest_denomination)
           expect(registration.paid_entry_fees.cents).to eq payment_amount
-          charge = Stripe::Charge.retrieve(registration.registration_payments.first.receipt.stripe_id, stripe_account: competition.connected_stripe_account_id)
+          charge = registration.registration_payments.first.receipt.retrieve_stripe
```
Rubiks-boy, line 610: nice! these are some really good tests in here.

WcaOnRails/spec/requests/registrations_spec.rb
```diff
@@ -580,133 +584,256 @@
...
+          payment_intent = registration.reload.stripe_payment_intents.first
+          expect(payment_intent).to_not be_nil
+          # Intent should not be confirmed at this stage, because we have never received a receipt charge from Stripe yet
+          expect(payment_intent.confirmed_at).to be_nil
+
+          expect {
+            # mimic the user clicking through the interface
+            Stripe::PaymentIntent.confirm(
+              payment_intent.stripe_id,
+              { payment_method: 'pm_card_visa_chargeDeclined' },
```
Rubiks-boy, line 818: Would it be possible to add tests for other payment methods too?

WcaOnRails/spec/features/sca_payments_spec.rb
```diff
@@ -1,64 +0,0 @@
...
-      fill_confirm_and_expect(card, "test-source-authorize-3ds", "Your payment was successful")
-    end
-
-    it "user fails to complete the 3D secure challenge", js: true do
-      fill_confirm_and_expect(card, "test-source-fail-3ds", "We are unable to authenticate your payment method")
-    end
-  end
-end
-
-def fill_confirm_and_expect(card, button_id, message)
```
Rubiks-boy, line 33: Can we add tests for this back? Some possible ideas:

1. Bringing back a test where we type card details
2. Checking for some basic events like whether a `ready` event is emitted.
3. Making sure the Element is rendered with a payment method we expect

WcaOnRails/app/models/registration.rb
```diff
@@ -18,6 +18,7 @@ class Registration < ApplicationRecord
   has_many :events, through: :competition_events
   has_many :assignments, dependent: :delete_all
   has_many :wcif_extensions, as: :extendable, dependent: :delete_all
+  has_many :stripe_payment_intents, as: :holder, dependent: :delete_all
```
Rubiks-boy, line 21: Having `registration_payments`, `stripe_payment_intents`, and `stripe_transactions` seems like similar ideas. What do each of these do? Is there a way we could remove one/derive it instead?

Rubiks-boy (changes requested):
Woohoo! I'm really excited about all of this :) This PR is a certified chonker, but it'll bring a lot of cool features to the site!

I left a bunch of comments - I hope that that's ok. I'm still super new to the codebase, so please call me out on things that don't make sense or seem way off!

WcaOnRails/app/models/registration.rb
Rubiks-boy: ✅ Discussed on Google Meet: Separation of `registration_payments` from `stripe_payment_intents` allows for going the multi-processor route later. `stripe_transactions` serves as additional logging.

WcaOnRails/app/views/registrations/_payment_form.html.erb
```diff
@@ -1,130 +1,145 @@
...
+    async function initializeStripeElements(data) {
+      $stripeCheckoutButton.hide();
+      $stripeElementsDiv.show();
+
+      const { clientSecret } = data;
+      const appearance = { theme: 'flat' };
+
+      elements = stripe.elements({ appearance, clientSecret });
+
+      const paymentElement = elements.create("payment");
```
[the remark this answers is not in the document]
Rubiks-boy: ✅ Discussed on Google Meet: Out of scope for this

WcaOnRails/app/controllers/registrations_controller.rb
```diff
@@ -439,163 +441,246 @@ def register
...
+    # because our CI containers are not exposed to the public. So we need this little hack :/
+    enable_automatic_pm = !Rails.env.test?
 
-  private def error_to_s(error)
-    error.inspect + "\n" + error.backtrace.join("\n")
+    payment_intent_args = {
+      amount: stripe_amount,
+      currency: currency_iso,
+      # recommended as per https://stripe.com/docs/payments/payment-element/migration
+      automatic_payment_methods: { enabled: enable_automatic_pm },
```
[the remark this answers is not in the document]
Rubiks-boy: ✅ Discussed on Google Meet: In this case, it makes sense to do it all in 1 go. There's going to be extensive manual testing day-of to make sure it's safe.

WcaOnRails/app/controllers/registrations_controller.rb
```diff
@@ -439,163 +441,246 @@ def register
...
+    @competition = registration.competition
+
+    # Provided by Stripe upon redirect when the "PaymentElement" workflow is completed
+    intent_id = params[:payment_intent]
+    intent_secret = params[:payment_intent_client_secret]
+
+    stored_transaction = StripeTransaction.find_by(stripe_id: intent_id)
+    stored_intent = stored_transaction.stripe_payment_intent
+
+    unless stored_intent.client_secret == intent_secret
```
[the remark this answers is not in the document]
Rubiks-boy: ✅ From Google Meet: It shouldn't ever happen. Just being extra safe

WcaOnRails/app/controllers/registrations_controller.rb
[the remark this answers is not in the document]
Rubiks-boy: ✅ Discussed on google meet: No changes. We should keep this endpoint. The flow uses the redirect/confirmation page to log the payment once it's successful (and webhooks are just a fallback). It retrieves server-side using our secret key, and only uses the client secret as a sanity check.

WcaOnRails/app/models/stripe_transaction.rb
```diff
@@ -2,10 +2,19 @@
...
+    processing: "processing",
+    requires_capture: "requires_capture",
+    canceled: "canceled",
+    succeeded: "succeeded",
+    pending: "pending",
+    failed: "failed",
+    legacy_unknown: "unknown",
+    legacy_payment_intent_registered: "payment_intent_registered",
+    legacy_success: "success",
+    legacy_failure: "failure",
```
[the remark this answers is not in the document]
Rubiks-boy: ✅ Ignore this, it's just an enum copied from the stripe docs.

WcaOnRails/db/migrate/20230315170143_create_stripe_webhook_events.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class CreateStripeWebhookEvents < ActiveRecord::Migration[7.0]
+  def change
+    create_table :stripe_webhook_events do |t|
```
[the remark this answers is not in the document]
Rubiks-boy: ✅ See https://github.com/thewca/worldcubeassociation.org/pull/7379#discussion_r1194535694

WcaOnRails/app/models/stripe_payment_intent.rb
```diff
@@ -0,0 +1,76 @@
...
+  belongs_to :canceled_by, polymorphic: true, optional: true
+
+  scope :pending, -> { where(confirmed_at: nil, canceled_at: nil) }
+  scope :started, -> { joins(:stripe_transaction).where.not(stripe_transaction: { status: 'requires_payment_method' }) }
+  scope :processing, -> { started.merge(pending) }
+
+  delegate :stripe_id, :status, :money_amount, to: :stripe_transaction
+
+  # Stripe secrets are case-sensitive. Make sure that this information is not lost during encryption.
+  encrypts :client_secret, downcase: false
```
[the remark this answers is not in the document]
Rubiks-boy: ✅ See https://github.com/thewca/worldcubeassociation.org/pull/7379#discussion_r1194507539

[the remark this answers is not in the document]
Rubiks-boy: (I would vote for removing this, at least eventually. It seems like this is adding unnecessary complexity, given the extra row in the DB)

WcaOnRails/app/views/registrations/_payment_form.html.erb
```diff
@@ -1,130 +1,145 @@
...
+      button.attr("disabled", saving);
+      button.toggleClass("saving", saving);
     }
 
-    $cardButton.on('click', function(ev) {
-      ev.preventDefault();
-      toggleSaving(true);
+    // We need to ask the user to explicitly start the checkout process, because they may or may not enter
+    // a donation amount that needs to be included in the "amount" computation we send over to Stripe.
+    // TODO if anybody has smarter ideas how to handle the donation amount, feel free to change this!
```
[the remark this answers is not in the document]
simonkellly: Currently, it seems like if someone enters that they want to give a donation, clicks the pay with stripe button, and then changes to not have a donation / reduces the amount, then it still charges them the originally entered amount

WcaOnRails/app/controllers/registrations_controller.rb
```diff
@@ -439,163 +441,246 @@ def register
...
+      # Customer did not complete the payment
+      # For example, 3DSecure could still be pending.
+      flash[:warning] = t("registrations.payment_form.errors.payment_pending")
+    when 'requires_payment_method'
+      # Payment failed. If a payment fails, it is "reset" by Stripe,
+      # so from our end it looks like it never even started (i.e. the customer didn't choose a payment method yet)
+      flash[:error] = t("registrations.payment_form.errors.payment_reset")
+    when 'processing'
+      # The payment can be pending, for example bank transfers can take multiple days to be fulfilled.
+      flash[:warning] = t("registrations.payment_form.payment_processing")
```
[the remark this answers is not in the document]
gregorbg: It is not a success state because we do not consider the competition registration as complete in this state.

WcaOnRails/app/models/stripe_payment_intent.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
+
+class StripePaymentIntent < ApplicationRecord
+  belongs_to :holder, polymorphic: true
+  belongs_to :stripe_transaction
+  belongs_to :user
```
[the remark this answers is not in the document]
gregorbg: Registration `belongs_to` User = the user who initiated the registration
PaymentIntent `belongs_to` User = the user who initiated the payment. These two need not be the same person, because payments can (in theory!) be initiated by a different person, or even be initiated for something that is not a registration at all.

No performance tradeoffs or anything.

WcaOnRails/app/controllers/registrations_controller.rb
```diff
@@ -439,163 +441,246 @@ def register
...
+    end
+
+    # Payment Intent lifecycle as per https://stripe.com/docs/payments/intents#intent-statuses
+    case stored_transaction.status
+    when 'succeeded'
+      flash[:success] = t("registrations.payment_form.payment_successful")
+    when 'requires_action'
+      # Customer did not complete the payment
+      # For example, 3DSecure could still be pending.
+      flash[:warning] = t("registrations.payment_form.errors.payment_pending")
```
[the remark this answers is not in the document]
gregorbg: The `flash` displays a small pop-up like snackbar at the top of the screen. You can see it in action when you sign in and you get a "signed in successfully" -- that's implemented as a `flash`.

The interpretation of this specific case was actually c/p'ed from the previous implementation; GitHub marks it as diff because the lines around it changed too much.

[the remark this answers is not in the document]
gregorbg: I don't know when this would realistically ever come up, but better to have it and never use it than to be sorry we deleted it

WcaOnRails/app/controllers/registrations_controller.rb
gregorbg: Thanks although this is literally me copying Stripe documentation suggestions

WcaOnRails/spec/requests/registrations_spec.rb
gregorbg: Absolutely!

WcaOnRails/spec/features/sca_payments_spec.rb
gregorbg: Not for SCA specifically. Our testing framework is a _nightmare_ when it comes to iframes.
The reason why we had this test previously is that we were responsible for detecting and forwarding SCA events to Stripe manually. Now, Stripe does it for us so testing it in the way you described would be like testing Stripe's SDK is correct... Which I hope we shouldn't have to test at all.

WcaOnRails/app/views/registrations/_payment_form.html.erb
```diff
@@ -1,130 +1,145 @@
...
+      toggleSaving(true, $stripeCheckoutButton);
+      $intentAjaxError.hide();
+
+      loadPaymentIntent();
+    });
+
+    // Fetches a payment intent and captures the client secret
+    function loadPaymentIntent() {
       var registration_fees_to_pay = <%= @registration.outstanding_entry_fees.cents %>;
       var donation = parseInt($('#payment_donation').val() || 0);
```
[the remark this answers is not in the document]
gregorbg: The donation input field is handled through a `monetize` 3rd-party library that is guaranteed to be 100% compatible with our Ruby currency backend in terms of denomination. So it's sufficient to perform a simple addition on our servers first and _then_ do one single conversion to weird Stripe currencies before sending the PI off.

WcaOnRails/app/views/registrations/_payment_form.html.erb
```diff
@@ -1,130 +1,145 @@
...
+
       var amount = registration_fees_to_pay+donation;
+
       if (isNaN(amount)) {
         alert('<%= t("registrations.payment_form.alerts.not_a_number") %>');
       } else if (amount < registration_fees_to_pay) {
```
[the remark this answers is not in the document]
gregorbg: There is server-side validation already!
https://github.com/thewca/worldcubeassociation.org/pull/7379/files#diff-2d34ca02cb1e0bf6c7b8e232e34b00642c84a4222616a1fcc4c8f777da02bb44R599-R605

WcaOnRails/spec/requests/registrations_spec.rb
gregorbg: ...or not. Stripe does not provide convenient PaymentMethod ID shorthands for payment methods other than credit cards.
Manually creating such PaymentMethod objects is possible but very cumbersome because it requires pre-configuring the PI and also adding stuff like `mandate_data` for processing SEPA payment methods etc...

WcaOnRails/spec/features/sca_payments_spec.rb
Rubiks-boy: Missed that this was an SCA-specific file; I was thinking more about generic tests to make sure that the WCA site's integration is correct.

For instance, making sure we get the Payment Element rendering is an easy way to make sure that all the code that power the integration works. The goal wouldn't be to test Stripe's code or any particular edge cases that Stripe already handles. Instead, the goal would be an end-to-end test for the WCA site to create a payment intent, load the page, and render the Payment Element successfully.

WcaOnRails/app/views/registrations/_payment_form.html.erb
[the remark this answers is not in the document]
gregorbg: Good catch! This made me rethink the frontend logic and I have decided to go with the deferred PI approach after all.
Most importantly, pre-loading the PI doesn't have any kind of influence on the money that's being charged, setting `amount` to a high number in JS does nothing if the PI from our AJAX backend contains the correct (small) number in the end.

I shall also add some integration tests to cover the frontend logic better.

WcaOnRails/app/controllers/registrations_controller.rb
gregorbg: Done :)

WcaOnRails/spec/features/sca_payments_spec.rb
gregorbg: From the [Stripe documentation](https://stripe.com/docs/automated-testing):
> Frontend interfaces, like [Stripe Checkout](https://stripe.com/docs/payments/checkout) or the [Payment Element](https://stripe.com/docs/payments/payment-element), have security measures in place that prevent automated testing

This indeed makes it a pain in the butt to test, especially since the frontend is using iframes within iframes within iframes. I shall leave it at one simple setup test, that checks whether something renders.

Even creating a payment intent isn't necessarily a given anymore, seeing that we're using deferred PaymentIntents now. Checking those is really hard especially because of redirect URLs (our testing env isn't reachable from the outside world, and setting redirects to `if_required` goes against the "we don't want you to do automated frontend testing" policy of Stripe above)

PR #7392, opened 2022-11-21
Implement WCRP 2022-11 changes

gregorbg:
cf. https://github.com/thewca/wca-documents/pull/316


gregorbg:
Open question: Now that the `danger` radius for the competition map has been reduced drastically (100km to 10km), should we adjust the `warning` radius accordingly?

The latter is not binding (or even relevant) in the sense of the WCRP, but it feels silly to make the user aware of all competitions within 250km when suddenly only 10km matter. Thoughts?

c-goodyear:
I think this warning flag is usually best for flagging nearby competitions in terms of when they cross state / country borders because internally to those regions communication is normally high enough that people know what competitions are planned.

I think it then boils down to how much across a border becomes relevant to flag for organisers / Delegates in terms of them selecting dates etc. No strong personal opinion from me here on an actual number 🤷

c-goodyear:
Ready to merge now that WCRP updates have been approved by the Board. @gregorbg please merge this with the WCRP update when you have the time.

c-goodyear (approved):
LG! Thanks for doing this so quickly :-D

gregorbg:
@c-goodyear Both PRs have been merged and the updated km distance values are deployed to prod :tada:

PR #7393, opened 2022-11-21
PHP Docker Environment 

FinnIckler:
This PR enables local development of the PHP Portion of the Website using Docker by adding a php and nginx service to Docker compose. For this I followed the default.rb chef recipe, but here is a list of differences between the two environments:
- The templates previously generated by Chef (fcgi.conf, my.cnf, nginx configuration files) are now in the ./docker folder. There are some minor differences, because in Docker the services are not technically running on the same machine, but otherwise I kept the configuration the same.
- There are three new configuration files, a my.cnf, a php-fpm.cnf  and a php.ini which were previously configured using chef recipes. These also have minor differences because of the nature of docker. 
- The dockerized nginx is not build from source, as the auth-request-module is included by default in the image

I did two changes in the PHP codebase:
- Fixed an error in the error handling of the DBConn class
- removed all references to mysql_connect and forwarded them to the "new" mysqli driver because I couldn't install a plugin that has been deprecated for 9 years. 

What this PR does not do:
- Remove Vagrant, I felt like it would be a bit unclear if I also had all references to vagrant removed in this already complex PR. I will make a separate PR for this.
- Enable us to start the Website with Docker in prod. There are still other services, mainly supervisor and email sending that need to be ported from chef provisioning to Docker for it to work


docker-compose.yml
```diff
@@ -38,12 +40,49 @@ services:
...
     volumes:
       - wca_db_volume:/var/lib/mysql
+      - ./docker/my.cnf:/etc/my.cnf
     # Suppress unneeded messages via https://stackoverflow.com/a/55706057/2558618
     cap_add:
       - SYS_NICE
+    networks:
+      - wca-main
+
+  nginx:
```
gregorbg, line 50: Change it so the user has to specify a flag in order to boot nginx and php

gregorbg (changes requested):

docker-compose.yml
FinnIckler: I added an overwrite file for this

docker-compose.php.yml
```diff
@@ -0,0 +1,32 @@
...
+      - ./webroot:/webroot
+      - ./secrets/uploads:/secrets/uploads # so we can pretend to upload stuff
+      networks:
+        - wca-main
+        environment:
+          - MYSQL_HOST=wca_db
+          - MYSQL_USER=root
+          - MYSQL_PASSWORD=
+          - MYSQL_DB=wca_development
+
```
gregorbg, line 32: Trailing whitespace :upside_down_face:

FinnIckler: removed!

gregorbg (approved):
:ship:

PR #7395, opened 2022-11-23
Unify trainee_delegates and delegates association tables

### LINKED ISSUE #6892: Update Delegate Status at upcoming competitions after promotions
When updating A Delegate status from Trainee Delegate to Junior Delegate, for that Delegate's upcoming Competitions they are still listed under trainee Delegates.

Whenever the update is made, their upcoming competitions should be updated to update which Delegate role they are for the competition.

### LINKED ISSUE #7324: Trainee Delegate promotions - simplifying changing competition websites
I recently became a juniot delegate and it wasn't ideal to send so many emails emails on competition email threads so that WCAT could change me from being listed as a trainee delegate to a delegate on competitor websites. I'm sure WCAT and everyone else on the email threads didn't enjoy all the inbox clutter and that WCAT would appreciate being able to skip out on having to go through that process 😛 

It would be great to make this shift automatic after trainees become juniors. Maybe the website could check at view time instead of assigning a value?

gregorbg:
Partially reverts #5249
Fixes #7324
Fixes #6892

We actually had already added `trainee_delegate` to the `delegate_status` enum in the previous implementation. Only the associations with competitions were stored in a separate table, which caused some pains (see issue above)

If we ever decide that we want to formalise more Delegate roles (or even implement currently formalised roles like `regional_delegate` that are not currently present in the code) it would make sense to have one unified table as well. Introducing copy-pasta association tables for every kind of Delegate does not seem future-proof.

I get that this case might have been an exception because trainees are not staff, but actually it's quite easy to implement `competition.trainee_delegates` and `competition.all_other_not_trainee_delegates` even with a unified table. To that effect, I have introduced the terminology of `staff_delegates` which refer to "everything except trainee".

We may re-think this approach when we have grown to a size and scale that we introduce trainee team members. But for now, I think this cleanup will adequately solve our current problems. Let's see if I can keep it below 3 additional fixups that I have to push to compensate for failing legacy tests :smile:


gregorbg:
Worth noting that we keep the distinct `Delegates` and `Trainee Delegates` fields in the Create Competition form. The only difference is that these will be stored in the same table in the background now, which means that whenever somebody gets promoted from Trainee to Junior, they automagically "switch" to the Staff Delegate field.

danieljames-dj (commented):
Thanks Greg for working on this. Everything looks good to me.
Shouldn't [this line](https://github.com/thewca/worldcubeassociation.org/blob/ac458d70430c3dd576bc955ae0fa59f6e80cb711/WcaOnRails/app/models/user.rb#L1281) be removed?

gregorbg:
Thanks for the review :smile: 

> Shouldn't [this line](https://github.com/thewca/worldcubeassociation.org/blob/ac458d70430c3dd576bc955ae0fa59f6e80cb711/WcaOnRails/app/models/user.rb#L1281) be removed?

Yes, good catch! I had to look in quite a few places to find all usages of the old model approach, and I was already expecting to miss a few here and there. Good catch!

For future reference, stuff like this is a perfectly fine reason to set reviews to "Request Changes". No need to be shy :upside_down_face:

PR #7448, opened 2022-12-01
Fix DB seeds for use with Docker Ruby image

### LINKED ISSUE #7403: Error running db:migrate on a fresh docker container of the WST website
## Steps to reproduce:

1. Follow setup guide defined in the [README](https://github.com/thewca/worldcubeassociation.org/blob/master/README.md) // Open repo in Github Codespace
2. Run `docker compose up`, wait for the container to be ready for connections
3. Attempt to connect to localhost endpoint
4. Greeted with a screen requesting that `bin/rails db:migrate RAILS_ENV=development` be run
5. `cd` to `WcaOnRails` in a `docker exec` terminal, run the command from (4)
6. You may encounter a compatbility issue here. Adding 7.0 to the end of the file's class declaration seems to help, ie: `class RenameDrupalIndicesToBeUnique < ActiveRecord::Migration[7.0]`
7. Repeat Step 5 if error in (6) encountered
8. Encounter error: `Mysql2::Error: Table 'wca_development.cache' doesn't exist`

I've followed a variant of this process and ended up at Step 8 on 2 different machines and in a codespace. Looking for anyone who has successfully set up the WCA website recently who can report on whether they encountered this issue, or anyone follow the README steps in a fresh docker container without issue.

gregorbg:
(Kinda) fixes #7403 

For some :tm: reason (didn't want to dig any deeper) the classloader inside the Ruby Docker env wants fully qualified constants in the `lib` folder. This is a bit odd because on a "real" Ruby env on my dev machine it works fine, and in both cases the invocation happens through Rake, so normally they _should_ behave the same. Oh well...

I also switched the DB warmup to use our seeds, because loading the Dev Dump just takes soooo long these days.
People who want to use the dev data can SSH into the container and trigger the loading process via the usual `db:load:development` task themselves.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -23,7 +23,7 @@ services:
       done &&
       echo "MySQL is ready!" &&
       if ! [[ "`mysqlshow --user=root --host=wca_db wca_development`" =~ "Tables" ]] ;
-        then echo "Populating development db, this will take a while" && bin/rake db:load:development ;
+        then echo "Populating development db, this will take a while" && bin/rake db:reset ;
       fi &&
       bin/rails server -b 0.0.0.0'
     depends_on:
--- a/WcaOnRails/db/seeds/development/competitions.seeds.rb
+++ b/WcaOnRails/db/seeds/development/competitions.seeds.rb
@@ -78,7 +78,7 @@ after "development:users" do
           total_number_of_rounds: round_types.length,
           time_limit: event.can_change_time_limit? ? TimeLimit.new : nil,
           cutoff: nil,
-          advancement_condition: is_final ? nil : RankingCondition.new(16),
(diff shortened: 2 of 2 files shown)


FinnIckler:
LGTM, it should also not clash with my PHP Docker PR.

docker-compose.yml
```diff
@@ -23,7 +23,7 @@ services:
       done &&
       echo "MySQL is ready!" &&
       if ! [[ "`mysqlshow --user=root --host=wca_db wca_development`" =~ "Tables" ]] ;
-        then echo "Populating development db, this will take a while" && bin/rake db:load:development ;
+        then echo "Populating development db, this will take a while" && bin/rake db:reset ;
```
FinnIckler, line 26: I think this is a good idea!

FinnIckler (approved):

PR #7452, opened 2022-12-01
Add Caleb's tool to list of Software tools

Goosly:
- I didn't find any tests, presumably there are none :-)
- Needs translations in other languages, I don't know how that works in this repo.


coder13:
Respectfully, I don't feel confident enough about where delegatedashboard is just yet to put it up on the main website. I do thank you for the consideration though.

I do support competitiongroups.com though. There should be a basic set of tests, it's written in typescript to add more reliability, I use `Intl` in places to try to keep it consistent across the world, there's very few words that need to be translated, I feel it does it's job well.

gregorbg:
> * I didn't find any tests, presumably there are none :-)

No tests indeed. Currently Not necessary for static pages like this.

> * Needs translations in other languages, I don't know how that works in this repo.

Translations work by detecting changes to the `en.yml` and then sending automated messages to the translators mailing list whenever such changes are deployed.

gregorbg (changes requested):
Based on Caleb's comment above I am not comfortable merging this as-is, as I only want to list software tools that are currently maintained / active to appear on the main website. Please change your PR accordingly or submit a new one.

Goosly:
Several delegates are already using Delegate Dashboard so I'm sure it qualifies as maintained / active :-)
Well anyway. Updated PR.

coder13:
> Several delegates are already using Delegate Dashboard so I'm sure it qualifies as maintained / active :-)

I'm on a short coding break, but there's a number of bugs in it and I'm happy to be working with the limited number of delegates that I am working with.

Eventually I'll be happy to include it.

gregorbg (approved):

PR #7453, opened 2022-12-02
Increase Unicorn timeout

jonatanklosko:
Saving WCIF for large competitions may take longer than 30s, ideally we would optimise the request itself, but for now we need to make sure the server doesn't close the connection too soon.

### DIFF
--- a/WcaOnRails/config/unicorn.rb
+++ b/WcaOnRails/config/unicorn.rb
@@ -25,7 +25,7 @@ end
 listen "/tmp/unicorn.wca.sock"
 pid "#{dir}/pids/unicorn.pid"
 
-timeout 30
+timeout 60
 
 preload_app true
 


gregorbg:
Thanks for the quick fix! I think it makes sense to increase the timeouts as a temporary measure.
However, I would like to keep the issue open for brainstorming and  investigation on how to optimise the underlying requests themselves.

gregorbg (approved):

viroulep:
(also I've said this privately but I'll repeat this here: thank you very much for the super fast workaround :))

gregorbg:
It has been deployed and the server-status page also points to the correct commit.

I'll restart Unicorn manually just to make sure, hold on...

FinnIckler:
Unicorn wasn't restarted when Gregor deployed 10 hours ago (because there is still some issue with chef not marking supervisor_worker as executable), but I fixed that and the deployment ran through just a couple minutes ago

gregorbg:
Yeah, Unicorn has been restarted now. I also manually checked that...
```
~/worldcubeassociation.org @production> cat WcaOnRails/config/unicorn.rb | grep timeout
timeout 60
```
yields the expected result.

viroulep:
Alright thank you so much!
I managed to patch the WCIF and it took about 50s, definitely close :p

Now that most assignments are in, a subsequent patching took about 10s :)

FinnIckler:
Screenshot 2022-12-02 at 14 26 13
Pretty sure that where those attempts :P

PR #7454, opened 2022-12-02
Bulk-insert assignments

### LINKED ISSUE #7303: Patch WCIF is timing out
Please refer to "Patch WCIF is timing out" email thread for more details.

From Manu's original email:

> Sam is trying to save the WCIF of his competition KewbzUKChampionship2022 from AGE. Basically what AGE does is take a csv with assignments (see attachment) and convert it to the WCIF format.
> 
> Unfortunately Sam is getting a timeout after ~ 32 to 33 seconds. I get the same behavior on staging. For smaller competitions (like 100 competitors) this usually takes a few seconds, but this one (450 competitors) seems to time out.

The workaround he found:

> I tried some stuff on the side of AGE and apparently uploading the competitors assignments in chunks did the trick. I'm still interested in figuring out why it did work without my hacky workaround :) But I understand that this is not a priority for you now.

Finn's observations after investigating:

> I've never researched how wcif is implemented, but in
> https://github.com/thewca/worldcubeassociation.org/blob/master/WcaOnRails/app/controllers/api/v0/competitions_controller.rb
> There is no timeout specified in update wcif, so the timeout is either a whole server kind of thing or somewhere else

viroulep:
Fixes #7303.

The way we import assignments right now has 2 issues:
  - we fire one SQL query for the object validation
  - we fire one SQL query for updating/inserting each of them (and therefore we fire thousands of queries for a large competition!)

First of all it's worth noting that we do eager-load properly everything needed, which means just changing a bunch of assignments, or patching a WCIF with no change to them does not imply any performance issues.
However since we assign the `schedule_activity_id` rather than the `schedule_activity` itself, rails actually has to load the schedule activity for the `belongs_to` association.
I fixed that by building the assignment with the schedule activity object, which fixes the first point.

The second point would have been tricky to handle with rails 5, but thankfully bulk insert have been added with rails 6: I've changed a bit the way we would import assignments to first build an array of built object, and then I delete existing assignment before bulk-inserting the new one.

I think inserting or updating existing assignments would be possible through `upsert_all`, but I would have to track all assignments to delete.
I can probably investigate how hard/awkward this would be, but it seemed like a very straightforward approach to just  delete_all + insert_all.
Overall this takes exactly two queries.

I'm not sure how relevant this is on a development environment, but I've dropped the overall query time from 8s to 3s on my computer: we should definitely try this on staging to see how it looks.

### DIFF
--- a/WcaOnRails/app/models/competition.rb
+++ b/WcaOnRails/app/models/competition.rb
@@ -1729,37 +1729,50 @@ class Competition < ApplicationRecord
       :registration_competition_events,
     ]
     competition_activities = all_activities
+    new_assignments = []
+    removed_assignments = []
     wcif_persons.each do |wcif_person|
+      local_assignments = []
       registration = registrations.find { |reg| reg.user_id == wcif_person["wcaUserId"] }
+      next unless registration
       # NOTE: person doesn't necessarily have corresponding registration (e.g. registratinless organizer/delegate).
-      if registration && wcif_person["roles"]
+      if wcif_person["roles"]
         roles = wcif_person["roles"] - ["delegate", "trainee-delegate", "organizer"] # These three are added on the fly.
         registration.update!(roles: roles)
       end
-      if registration && wcif_person["assignments"]
-        registration.assignments = wcif_person["assignments"].map do |assignment_wcif|
-          existing_assignment = registration.assignments.find do |assignment|
-            assignment.wcif_equal?(assignment_wcif)
+      if wcif_person["assignments"]
+        wcif_person["assignments"].each do |assignment_wcif|
+          schedule_activity = competition_activities.find do |competition_activity|
+            competition_activity.wcif_id == assignment_wcif["activityId"]
           end
-          if existing_assignment
-            existing_assignment
+          unless schedule_activity
+            raise WcaExceptions::BadApiParameter.new("Cannot create assignment for non-existent activity with id #{assignment_wcif["activityId"]}")
+          end
+          assignment = registration.assignments.find do |a|
+            a.wcif_equal?(assignment_wcif)
+          end
+          # We need to be very careful about how we build the assignment:
+          # providing just the registration_id or the schedule_activity_id would
+          # actually trigger a select for each validation.
+          assignment ||= registration.assignments.build(
+            schedule_activity: schedule_activity,
+          )
+          assignment.assign_attributes(
+            station_number: assignment_wcif["stationNumber"],
+            assignment_code: assignment_wcif["assignmentCode"],
+          )
+          if assignment.valid?
+            local_assignments << assignment
           else
-            schedule_activity = competition_activities.find do |competition_activity|
-              competition_activity.wcif_id == assignment_wcif["activityId"]
-            end
-            unless schedule_activity
-              raise WcaExceptions::BadApiParameter.new("Cannot create assignment for non-existent activity with id #{assignment_wcif["activityId"]}")
-            end
-            registration.assignments.build(
-              schedule_activity_id: schedule_activity.id,
-              station_number: assignment_wcif["stationNumber"],
-              assignment_code: assignment_wcif["assignmentCode"],
-            )
+            raise WcaExceptions::BadApiParameter.new("Invalid assignment: #{a.errors.map(&:full_message)} for #{assignment_wcif}")
           end
         end
-        registration.save!
       end
+      new_assignments.concat(local_assignments.map(&:attributes))
+      removed_assignments.concat(registration.assignments.ids - local_assignments.map(&:id))
     end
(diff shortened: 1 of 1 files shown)


WcaOnRails/app/models/competition.rb
```diff
@@ -1737,29 +1738,33 @@ def update_persons_wcif!(wcif_persons, current_user)
...
+          # We need to be very careful about how we build the assignment:
+          # providing just the registration_id or the schedule_activity_id would
+          # actually trigger a select for each validation.
+          a = registration.assignments.build(
+            schedule_activity: schedule_activity,
+            station_number: assignment_wcif["stationNumber"],
+            assignment_code: assignment_wcif["assignmentCode"],
+          )
+          if a.valid?
+            new_assignments << a.attributes
```
gregorbg, line 1757: Do you really have to explicitly append the attributes? I would expect Rails to be "smart" enough to handle actual model objects in the `insert_all` list

WcaOnRails/app/models/competition.rb
```diff
@@ -1737,29 +1738,33 @@ def update_persons_wcif!(wcif_persons, current_user)
...
-              station_number: assignment_wcif["stationNumber"],
-              assignment_code: assignment_wcif["assignmentCode"],
-            )
+            raise WcaExceptions::BadApiParameter.new("Invalid assignment: #{a.errors.map(&:full_message)} for #{assignment_wcif}")
           end
         end
-        registration.save!
       end
     end
+    Assignment.where(registration_id: registrations.ids).delete_all
```
gregorbg, line 1764: I am not too wild about flat out deleting all existing assignments. This leads to conflicts when two different WCIF tools write assignments and sync them back to the website, where the first can create an assignment that is later deleted and re-created with different values by a second tool

gregorbg (commented):
Great catch on the `insert` VS `insert_all` subtleties! I have two remarks that don't necessarily _have_ to be changed, but that I would still like to exchange opinions on. So I've set my CR to a neutral "Comment only" :smile_cat:

WcaOnRails/app/models/competition.rb
viroulep: According to the [documentation](https://apidock.com/rails/v6.0.0/ActiveRecord/Persistence/ClassMethods/insert_all) it does expect attributes; I didn't check before that PR but now I can confirm you can't just pass objects :s

WcaOnRails/app/models/competition.rb
viroulep: I'm not too wild either, but rather because it creates unnecessary read/write to the db; as far as the WCIF and external tools are concerned there is absolutely no changes: the WCIF values (`activityId`/`stationNumber`/`assignmentCode`) stay the same, and therefore external tools would see the "same" assignment.
What does change is our internal database id associated with that assignment.

I'll try to see if I can come up with something better.

viroulep: I just pushed an implementation which tries to re-use existing assignment, but to be honest it doesn't change much in practice: either with the current implementation (in master) or the one I'm suggesting here we rely on the whole WCIF attributes to figure out if assignments are equal.
It's the only solution because assignments don't have ids, but it means that if we change one single attribute (eg: the station number) we don't find the existing assignment and we end up deleting the old one and creating a new one.

I've still pushed the new version because it avoids deleting everything if nothing changed.

viroulep:
Kind ping to see if there is anything left for me to do: Oceanic Championship are next Thursday and it would be awesome to have a fix merged by then, just in case we do hit the threshold again.

WcaOnRails/app/models/competition.rb
```diff
@@ -1729,37 +1729,50 @@ def update_persons_wcif!(wcif_persons, current_user)
...
+      if wcif_person["roles"]
         roles = wcif_person["roles"] - ["delegate", "trainee-delegate", "organizer"] # These three are added on the fly.
         registration.update!(roles: roles)
       end
-      if registration && wcif_person["assignments"]
-        registration.assignments = wcif_person["assignments"].map do |assignment_wcif|
-          existing_assignment = registration.assignments.find do |assignment|
-            assignment.wcif_equal?(assignment_wcif)
+      if wcif_person["assignments"]
+        wcif_person["assignments"].map do |assignment_wcif|
```
gregorbg, line 1744: Nitpick: You're not using the result of the `map` invocation, so `my_array.each do |assignment_wcif|` should do.

gregorbg (approved):
`upsert_all` seems really cool! I'm totally fine with the new logic, thanks for coming up with a concise alternative solution after my first review :)

WcaOnRails/app/models/competition.rb
viroulep: Fixed!

viroulep:
Awesome thanks!

PR #7456, opened 2022-12-04
Bump Shakapacker to v6.5.4 in Bundler and Yarn

gregorbg:
Because they need to be bumped together and Dependabot doesn't understand that concept.


gregorbg:
Supersedes #7444 
Supersedes #7437

gregorbg:
Treating by the same rules as a Dependabot update.

Issue #7307, opened 2022-10-23
Redefine Registration Full Text to appear when the number of registered competitors exceeds the limit


Jambrose777:
**Describe the bug**
Currently, we display this message whenever the number of accepted registrations >= the competitor limit:
<img width="1268" alt="image" src="https://user-images.githubusercontent.com/11577241/197404956-af6f2cd3-e84a-4408-bf08-2f5f9fabe61e.png">

However, whenever we remove a registration (say someone cancels) and inquire the first on the waiting list about the open spot. During this time the warning disappears (even when there are 70+ people on the waiting list). This causes some confusion as the competition is currently full, but we are working with the waiting list registrants.

**Expected behavior**
The warning should continue to appear if the number of accepted competitors **+ the number of paid registrations on the waiting list** >= the competitor limit. The warning can also be modified to say "may be filled" in this case?

**Additional context**
relevant code: models/competition.rb:342
`def registration_full?
    self.competitor_limit_enabled? && self.registrations.accepted.count >= self.competitor_limit
  end`

navaneethkp36:
hi @Jambrose777 I would like to work on this issue.

dunkOnIT:
Thanks for the interest @navaneethkp36 ! I have assigned you - please shout here if you have any questions or run into any issues.

navaneethkp36:
hi @dunkOnIT I saw this same logic to check if registration is full being used in 3 places:

1. in `/competitions/competition_name/registrations/add`, a warning message - `Note: The competitor limit has been reached and there are no more spots available` is shown with same logic.
2. registration status icon in `/competitions` when `Show registration status icon` checkbox is ticked. A user with clock icon is shown with warning text - `Competitor limit reached, new registrations will go to the waiting list` with same logic.
3. Alert message in `competitions/competition_name/register` with same logic.

The model function `registration_full?` in `models/competition.rb` is used at all these places. Changing the function logic will make the logic change in all the places mentioned above.

So is it fine to make changes at all places or should I create a new function and change the logic for only the registration warning mentioned in this issue? Please advise.

Thanks

dunkOnIT:
Hi @navaneethkp36, 

If I'm understanding you correctly (I mostly deal with project management, not code!) changed the logic of `registration_full?` to "number of accepted competitors + the number of paid registrations on the waiting list >= the competitor limit" will cause the warning to change in .../registrations/add, /competitions, and .../competition_name/register?

[...]

PR #7458, opened 2022-12-04
Redefine Registration Full Text to appear when the number of registered competitors exceeds the limit

navaneethkp36:
issue - https://github.com/thewca/worldcubeassociation.org/issues/7307

### DIFF
--- a/WcaOnRails/app/models/registration.rb
+++ b/WcaOnRails/app/models/registration.rb
@@ -219,6 +219,12 @@ class Registration < ApplicationRecord
     }
   end
 
+  class << self
+    def accepted_and_paid_pending_count
+      accepted.count + pending.with_payments.count
+    end
+  end
+
   # Only run the validations when creating the registration as we don't want user changes
   # to invalidate all the corresponding registrations (e.g. if the user gets banned).
   # Instead the validations should be placed such that they ensure that a user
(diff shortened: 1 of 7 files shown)


WcaOnRails/app/models/registration.rb
```diff
@@ -208,6 +208,12 @@ def self.wcif_json_schema
     }
   end
 
+  class << self
```
gregorbg, line 211: Why is this `class << self` extension necessary here?

gregorbg (changes requested):
Thanks for your contribution! Very concise implementation, I am just startled by the non-trivial extension in the `registration` model.

WcaOnRails/app/models/registration.rb
navaneethkp36: hi @gregorbg I was just following the rubocop convention - (**_Use `class << self` to define a class method. (convention:Style/ClassMethodsDefinitions)_**). 

`def self.accepted_and_paid_pending_count` will work too. But this will throw a `rubocop` warning. I can use this convention if the `rubocop` suggested convention seems confusing. Should I make the change?

navaneethkp36: Changed to old syntax in https://github.com/thewca/worldcubeassociation.org/pull/7458/commits/bac8dd0831b07a2018841e40ad1e57263e75bd24

gregorbg (approved):

PR #7459, opened 2022-12-07
[Migrate React] Create competition form

simonkellly:
In progress migration of the [create competition](https://www.worldcubeassociation.org/competitions/new) page


gregorbg:
For the record: You have the [restructured Dues estimation PR](https://github.com/thewca/worldcubeassociation.org/pull/7377) available on `master` now :upside_down_face: 

It removed a lot of jQuery pains and instead provides an endpoint to return pure JSON that contains all relevant information.

WcaOnRails/app/controllers/competitions_controller.rb
```diff
@@ -365,6 +365,34 @@ def clone_competition
     render :new
   end
 
+  def comeptition_form_nearby_json(c)
```
kr-matthews, line 368: 'competition' typo

WcaOnRails/app/webpacker/components/CompetitionForm/index.js
```diff
@@ -0,0 +1,3 @@
+import CompetitionForm from './CompetitionForm';
+
+export default CompetitionForm;
```
kr-matthews, line 3: I'm curious why this file is needed?

WcaOnRails/app/webpacker/components/CompetitionForm/CompetitionForm.js
```diff
@@ -0,0 +1,153 @@
...
+  const longData = useFormInputState('longitude_degrees', competition);
+
+  const startDateData = useFormInputState('start_date', competition);
+  const endDateData = useFormInputState('end_date', competition);
+
+  return (
+    <>
+      <Form>
+        {competition.persisted && adminView ? <AdminView competition={competition} /> : null}
+        {competition.persisted && !adminView ? (
```
kr-matthews, line 111: `a ? b : null` -> `a && b`

WcaOnRails/app/webpacker/components/CompetitionForm/CompsTable.js
```diff
@@ -0,0 +1,157 @@
...
+      data-container="body"
+      title="This competition is not confirmed yet."
+    />
+  );
+}
+
+function CompsTableContent({ nearby, action }) {
+  const [showEvents, setShowEvents] = useState(false);
+
+  const headerRow = () => (
```
kr-matthews, line 24: I think something like this could just be a constant instead of a function, unless I'm missing something. Possibly wrapped in a `useMemo` dependent on `showEvents`.

WcaOnRails/app/webpacker/components/CompetitionForm/FormInputs.js
```diff
@@ -0,0 +1,107 @@
...
+import React, { useState, useCallback } from 'react';
+
+import {
+  Checkbox,
+  Form,
+  Input,
+} from 'semantic-ui-react';
+import I18n from '../../lib/i18n';
+
+// Modified from '../../lib/hooks/useInputState';
```
kr-matthews, line 14: I haven't checked the original hook, but what's the difference? Any chance the other hook could be refactored instead of making a new one? If not, should it be moved to the hooks folder too?

WcaOnRails/app/webpacker/components/CompetitionForm/NearbyComps.js
```diff
@@ -0,0 +1,45 @@
...
+    params.append(`competition[${startDateData.attribute}]`, startDateData.value);
+    params.append(`competition[${endDateData.attribute}]`, endDateData.value);
+
+    fetchJsonOrError(`${competitionNearbyJsonUrl}?${params.toString()}`).then(({ data }) => {
+      setNearby(data);
+      setLoading(false);
+    });
+  }, [latData.value, longData.value, startDateData.value, endDateData.value]);
+
+  const label = I18n.t('competitions.adjacent_competitions.label', { days: 5, kms: 10 });
```
kr-matthews, line 31: Probably should extract the 5 and 10 as constants somewhere. I assume 5 days and 10 kms shows up elsewhere in the code, and ideally they would all have a common source of truth.

WcaOnRails/app/webpacker/components/CompetitionForm/SeriesComps.js
```diff
@@ -0,0 +1,50 @@
+/* eslint-disable react/no-danger */
```
kr-matthews, line 1: If a rule is only being violated once or twice, I'm guessing it would be better practice to only disable individual lines: `eslint-disable-next-line`?

WcaOnRails/app/webpacker/components/CompetitionForm/VenueMap.js
```diff
@@ -0,0 +1,130 @@
...
+function roundToMicrodegrees(toRound) {
+  const val = toRound || 0;
+  // To prevent are you sure? from firing even when nothing has changed,
+  // explicitly round coordinates to an integer number of microdegrees.
+  return Math.trunc(parseFloat(val) * 1e6) / 1e6;
+}
+
+function DraggableMarker({ latData, longData }) {
+  const position = { lat: latData.value, lng: longData.value };
+  const markerRef = useRef(null);
```
kr-matthews, line 26: I'm not too familiar with `useRef`, could you explain why it's needed here?

WcaOnRails/app/webpacker/components/CompetitionForm/VenueMap.js
```diff
@@ -0,0 +1,130 @@
...
+    map.on('geosearch/showlocation', (e) => {
+      if (!e.location) return;
+      latData.setValue(roundToMicrodegrees(e.location.y));
+      longData.setValue(roundToMicrodegrees(e.location.x));
+    });
+
+    return () => {
+      map.removeControl(searchControl);
+    };
+  }, []);
```
kr-matthews, line 72: Is this intentionally omitting dependencies (ie `map`)? Generally that's a bad sign.

WcaOnRails/app/webpacker/components/CompetitionForm/CompsTable.js
coder13: Would love to see this as a component just outside of the parent component. It does take a prop "showEvents" so it can't exactly be a constant.

`useMemo` feels like a bit overkill for this simple design.

WcaOnRails/app/webpacker/components/CompetitionForm/CompsTable.js
```diff
@@ -0,0 +1,157 @@
...
+          {I18n.t('competitions.adjacent_competitions.events')}
+        </Table.HeaderCell>
+      ) : null}
+    </Table.Row>
+  );
+
+  const compRow = (comp) => (
+    <Table.Row key={comp.name} warning={!comp.danger} error={comp.danger}>
+      <Table.Cell name="name" width={3}>
+        <span dangerouslySetInnerHTML={{ __html: comp.nameLink }} />
```
coder13, line 58: Why is this `dangerouslySetInnerHTML` used? Can it not be an `a` tag?

simonkellly: comp.nameLink is passed as html through comeptitionform_nearby_json in WcaOnRails/app/controllers/competitions_controller.rb

WcaOnRails/app/views/competitions/_competition_form_new.html.erb
```diff
@@ -0,0 +1,26 @@
+<% is_actually_confirmed = @competition.persisted? ? Competition.find(@competition.id).confirmed? : false %>
+<% mail_to_wcat = mail_to Team.wcat.email, t("competitions.competition_form.wcat"), target: "_blank" %>
+<%= react_component("CompetitionForm", {
+    competition: {
```
coder13, line 4: is there a reason you can't do `competition: @competition`?

WcaOnRails/app/webpacker/components/CompetitionForm/CompsTable.js
simonkellly: I think I will end up making this a bit more reusable anyway as there was more variation than I had initially planned for in each of the places where its going to be used, so I will reevaluate this later on

WcaOnRails/app/webpacker/components/CompetitionForm/SeriesComps.js
simonkellly: Yeah, ill write a todo comment to remind myself to go through all of these once the PR is almost done, as I have just left them disabled for ease of development while I work

WcaOnRails/app/webpacker/components/CompetitionForm/VenueMap.js
simonkellly: Im not 100% certain, as I was followng the react-leaflet docs, but I think it's to do with how its handling storing the state of the marker and using it in the dragend callback

WcaOnRails/app/views/competitions/_competition_form_new.html.erb
simonkellly: I was going to write a specific serializable_hash for it, but as Comptitions already had one and I didn't want to break anything (yet) I did it this way. I plan to move it either to the serializable_hash method or a seperate one in the competitions controller once all the properties are filled out

WcaOnRails/app/webpacker/components/CompetitionForm/VenueMap.js
coder13: You would use `useRef` whenever you need to reference the element. Either in an (antipattern) way to trigger a function on it or access data in a backwards way.

The more standard practice would be to retrieve this data from an event object.

from looking at source code, `onDragend` should pass in an "event" prop. Can you look at it to see if you can access the position from there?

simonkellly:
Just a brief progress update on what is left at this point

- [ ] Need logic for submitting / editing the comp in the backend
- [ ] Need to move serialization logic from the form_new page to the controller
- [ ] Make the map provider the same as the original form and check for zooming after searching
- [ ] DateTimes are handled in kind of a bad way
- [ ] Bottom action buttons don't yet exist
- [ ] There is no "Are You Sure" when leaving with changes
- [ ] Error/warning handling on fields is missing
- [ ] No global constants for some days and distances in tables
- [ ] Need to add another admin only all comps in past 365 days table
- [ ] Need to update the unit tests to work with the new UI
- [ ] Need to remove the old code once everything is working

WcaOnRails/app/webpacker/components/CompetitionForm/CompsTable.js
```diff
@@ -0,0 +1,168 @@
...
+function CompsTableContent({ comps, action }) {
+  const [showEvents, setShowEvents] = useState(false);
+
+  return (
+    <Table structured>
+      <Table.Header>
+        <CompsTableHeaderRow showEvents={showEvents} />
+      </Table.Header>
+      <Table.Body>
+        {comps.reverse().map((comp) => (
```
kr-matthews, line 110: `reverse` reverses the array in place, which could lead to unintended consequences here or in the parent component. Probably best to make a copy? `[...comps].reverse().map(...)`

WcaOnRails/app/webpacker/components/CompetitionForm/DuesEstimate.js
```diff
@@ -0,0 +1,53 @@
...
+          } else {
+            text = `${I18n.t('competitions.competition_form.dues_estimate.per_competitor', {
+              estimate: json.dues_value,
+            })} (${currency})`;
+          }
+          setDuesText(text);
+        }));
+  });
+
+  return !duesText ? null : (
```
kr-matthews, line 46: `duesText && ( ...`? or `!!duesText && ( ...`, not sure if returning an empty string is different from returning null/false.

WcaOnRails/app/webpacker/components/CompetitionForm/MarkdownEditor.js
```diff
@@ -0,0 +1,132 @@
+import React, { useMemo } from 'react';
```
coder13, line 1: Some of these components can very likely be moved up a directory level to increase reusability.

WcaOnRails/app/webpacker/components/CompetitionForm/ChampionshipInput.js
```diff
@@ -0,0 +1,116 @@
...
+        <Form.Button icon color="red" onClick={remove} disabled={disabled}>
+          <Icon inverted name="close" />
+        </Form.Button>
+      </Form.Group>
+    </Form.Field>
+  );
+}
+
+function getAvailibleId(data) {
+  const ids = data.map((championship) => championship.id);
```
coder13, line 52: I would just do `Math.max(...data.map((championship) => championship.id));`
This is safe *just in case* there's a gap in the IDs and they are not all strictly consecutive.

WcaOnRails/app/webpacker/components/CompetitionForm/NearbyComps.js
```diff
@@ -0,0 +1,55 @@
+/* eslint-disable react/no-danger */
```
coder13, line 1: Is this line needed?

simonkellly:
Note to self: need to integrate changes in #7896

gregorbg:
Note: I am working on the backend in two separate phases. Reading the new format is done, but writing back (i.e. clicking the "update" or "create" button at the bottom) does not currently work.

dunkOnIT (commented):
nvm Gregor is doing this :P

gregorbg:
There is a (somewhat) functioning backend now, but it's missing extensive tests. The only guarantee that I can give right now is "works on my machine". Still missing:
- Adjusting the existing tests to work with the new UI
- Fine-tuning for the Series element (in particular, including the current comp that is being edited in the series when clicking the "Add to Series" button)
- AutoNumeric in React still behaves unintuitively and has glitches
- (Maybe?? 365 day admin table for nearby comps)

@simonkellly How available are you to work on these tasks? Which ones do you feel comfortable with and where do you need more help?

gregorbg:
Sidenote: The horrendous line count in the diff is because of moving a lot of i18n keys across ALL locale files around. You can subtract ~10.000 lines from the "Add" count.

simonkellly:
I can definitly do 365 day admin table, and give a decent stab at the tests

I will have a look into what is happening with autonumeric but as im not sure what is happening no promises here, and same with the series element.

I will likely have the most availibilty this week towards the weekend (thursday evening through monday)

gregorbg:
Note: I am running into one of the weirdest bugs I have potentially ever witnessed, cf. https://github.com/glebm/i18n-tasks/issues/537

Short version: If the file `foo.html.erb` (an abridged version of the old `_competition_form.html.erb` is not present, the check for unused keys crashes with **completely unrelated** error messages (unrelated i18n keys in unrelated files). I have tried everything I can think of, including the exclusion of all new JS code from the "unused keys" scans to no avail.

The only way that I can fix this error after 6+ hours of debugging is by keeping **some parts** of the old competition form in our code, unused and orphaned.

simonkellly:
Here are a couple of things I have found while testing it out on staging:

- [ ] Competitor limit doesn't show further options until reselected
- [ ] No error is shown if start/end date is missing, just silently fails
- [ ] User search throws error when selecting people `Uncaught TypeError: e is not a function at setSelectedValue (WcaSearch.jsx:36:22)`
- [ ] Fixing an error and submitting doesnt remove error text even when successful

I will try get through a couple of them this evening

gregorbg:
I fixed three out of four, the remaining error is because there is no validation for start and end date in Rails.

The form errors that are being displayed rely on Rails `validates :some_column` validation error messages. In order for an error message to be shown, we need the validation to fail. But "simply" adding a new validation might break (a lot of!!) existing assumptions and test cases.

The sanest route (in terms of our existing data) would be to investigate which validation triggers in the old form, and then find out why it doesn't trigger in `def form_errors` in `competition.rb` for the new form.

gregorbg:
Also, if you don't fill in any value, the JSONSchema check fails. This is currently not "properly" handled by the frontend.

simonkellly:
> The sanest route (in terms of our existing data) would be to investigate which validation triggers in the old form, and then find out why it doesn't trigger in def form_errors in competition.rb for the new form.

So I had a look at leaving out the start date in the old form, and it just brings you to an error page when you try to save with the rest of the form valid except the start/end date. Same thing happens when you remove the date from an existing comp

gregorbg:
Hm. This sounds like a less trivial fix then.

Anyways, the current implementation is just expecting form errors (the ``form_errors` object that we custom-build in the backend) and if it's not that, it throws the error. That's not good practice in production anyways. What about changing the store from `errors` and `setErrors` to `formErrors` (and `setFormErrors`, respectively) and then displaying non-form errors in the same error box on top?

simonkellly:
By non form errors, do you mean errors such as `ArgumentError (comparison of ActiveSupport::TimeWithZone with nil failed):` which is being thrown in this case?

Also as an aside, adding `return unless start_date.present?` to the start of the registration_dates_must_be_valid validation makes creating a comp without dates work fine from my initial testing

simonkellly:
To note, errors which are thrown during competition confirmation are not handled / shown in the UI

gregorbg:
Merging on my authority. If something goes wrong, I'll be available to handle it.

simonkellly:
🎉

PR #7476, opened 2022-12-16
Update friendly report reminder template

gregorbg:
Based on feedback from WEAT after the initial implementation went live.

Template file is available for reference at https://docs.google.com/document/d/1xtsg-mIVE9Gc2b4Vyor9cGEGDAh2bv4CeKxdfFWBTf4/edit

### DIFF
--- a/WcaOnRails/app/mailers/competitions_mailer.rb
+++ b/WcaOnRails/app/mailers/competitions_mailer.rb
@@ -134,7 +134,6 @@ class CompetitionsMailer < ApplicationMailer
     mail(
       from: Team.weat.email,
       to: competition.delegates.pluck(:email),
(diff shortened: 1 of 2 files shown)


danieljames-dj (approved):
LGTM. I'm assuming the change is that - the 'reminder' email will go only to delegate and not to WEAT & Senior like the 'delayed report' email.

gregorbg:
Correct. Thanks for reviewing!

peri-ld:
Hi, sorry to reopen this, but I just realised the tense is incorrect "must have been submitted..." instead of "must be..." and it would be good to avoid confusion that the report is late, which it isn't yet at this point. Would it be possible to update the wording again to what is currently in the template?

Issue #7284, opened 2022-10-11
Get rid of Webpack rails-erb-loader


dunkOnIT:
Mostly relates to data fetching from the Ruby backend. The replacement should be a (global?) store that holds data like events, formats, genders, currencies, etc. that are currently recignised by our code.

- Everything under `webpacker/lib/wca-data`
    - `rails-env.js.erb` feels very silly to export to JS, and we currently only need it for **one** single hack to avoid map geocoding calls during testing. Figure out a better way to handle this.
- `webpacker/lib/requests/routes.erb` (hardcode? load routing table from backend dynamically?)
- `webpacker/lib/leaflet-wca/wca-provider.js.erb` (why is this even an ERB in the first place?)

gregorbg:
Worth noting that I am open to the possibility that we just use the `rails-erb-loader` gem completely wrong and that there might be a configuration option to speed things up??

I really don't know at this point and I'd prefer not having `.erb` files under `app/webpacker` at all, but in the end what I want is a reasonably fast Webpacker compilation that doesn't fry our staging server. I don't feel very strongly about how this is achieved.

viroulep:
> `webpacker/lib/leaflet-wca/wca-provider.js.erb` (why is this even an ERB in the first place?)

My guess is that I intended to *not* hardcode the route in the first place, even though it is hardcoded right now; it can definitely just be changed to actually use the generated routes!

For the record if I had to do this my plan would be pretty much the same as Caleb's; I'd most likely take a look at how `i18n-js` works under the hood, because it pretty much does something similar to what we'd like: generate a set of static files based on some ruby code, and it supports watching for file changes and just appropriately hooks into the rails system.

gregorbg:
Funnily enough, `rails-erb-loader` does exactly this, but using a "pull" instead of a "push" approach. I mean, literally all that happens now is that the plugin boots Rails, converts the ERB to "proper" JS files, and passes them to Webpack for the actual compilation process.

The (very!) relevant problem here is that it boots Rails _once per file_, so right now every time we run `assets:precompile`, 12 Rails instances are spun up in parallel: https://github.com/usabilityhub/rails-erb-loader/issues/73

gregorbg:
One "temporary" solution would be to bundle all of the constants under the `wca-data` folder into **one** erb file so that we keep the number of spawning Rails processes as low as possible.

Also, could this be a good use case for deliberately booting the temporary "ERB compiler" Rails behind Spring in prod?

PR #7477, opened 2022-12-16
Bundle all Ruby constant exports to Webpacker in one file

gregorbg:
...so that staging doesn't die trying to spin up 12 servers in parallel.

Temporary solution to the symptoms discussed in #7284. Not beautiful, but necessary.


FinnIckler:
Sounds like a good solution for now, I mean our goal is to have 0 .js.erb files in the end right?

gregorbg:
Yes, that would be the end goal. The current PR is a step along the way.

jonatanklosko (approved):
Looks like a good workaround!

FWIW I think it's actually better to have a single ~100 loc file with the "bridge" stuff, rather than a bunch of ~10 loc files :)

gregorbg:
Thanks! :smile:

PR #7478, opened 2022-12-16
Bump Ruby to version 3.1.3

gregorbg:


gregorbg:
Ruby 3.1.3 has been succesfully installed on all production servers.
Now treating this the same way we would treat a normal dependency update, given the fact that it's a patch release anyways.

PR #7480, opened 2022-12-17
changed bullet to be disabled by default

dunkOnIT:
Set `disable_bullet=true` in `.env.development`, which will make bullet optimiser opt-in only, and not confuse noobs like me


gregorbg:
On second thought, how about changing it to `ENABLE_BULLET=false` by default? This would be less convoluted, but it will require changing the code in `env_vars.rb` as well as all occurences (should only be one anyways=

dunkOnIT:
I love it when you think twice

gregorbg:
Gives you a nice opportunity to practice on some Rails code :smile:

FinnIckler:
can this be merged?

gregorbg:
Technically yes, but I'd prefer to update this to a state where the vars are called `BULLET_ENABLED`.
Bullet is an inspection tool that should be disabled by default and only _enabled_ consciously when you want to investigate some performance issues or perform benchmarks.

dunkOnIT:
Happy to make the requested changes, just a bit snowed under at the moment!

WcaOnRails/config/environments/development.rb
```diff
@@ -72,7 +72,7 @@
   config.active_storage.service = :local
 
   config.after_initialize do
-    Bullet.enable = !EnvVars.DISABLE_BULLET?
+    Bullet.enable = !EnvVars.ENABLE_BULLET?
```
gregorbg, line 75: NB You have to flip the negation here!

gregorbg (changes requested):
Somehow the local changes to your structure.sql made it into the diff. Please clear them out before merging.

dunkOnIT:
Removed `structure.sql` diff and flipped negation in `development.rb` - whoops :sweat:

dunkOnIT:
The failing test is confusing me:
> `SuperConfig::MissingEnvironmentVariable: ENABLE_BULLET is not defined.`

Looks to me like this is being run in the test environment (`RACK_ENV: test`). `ENABLE_BULLET` isn't defined there, but `DISABLE_BULLET` never was either. I see why it should be defined - `env_vars.rb:12` makes it mandatorry unless the environment is `production`. 

I believe the fix is to define `ENABLE_BULLET` somewhere in `test.rb` - but I'd like to understand why this wasn't causing an error before?

gregorbg:
You're close. The configuration values are loaded from the `.env.*` files in `WcaOnRails`. So the error is correct insofar as `.env.test` still contains the old `DISABLE_BULLET` and SuperConfig is rightfully complaining that `ENABLE_BULLET` is not defined in that file :upside_down_face:

dunkOnIT:
Ah thanks! Thought I had grepped for `DISABLE_BULLET` in the project folder, but clearly not (:

gregorbg (approved):

PR #7489, opened 2022-12-21
Make the check of similar persons case insensitive Fixes #7486

### LINKED ISSUE #7486: Make the check of similar persons case insensitive
When results are uploaded by delegate, few validations are ran, and out of that one validator is 'Persons validator' which checks many things regarding the Persons data like name, DOB, etc. In 'persons validator', one of the validator is [SAME_PERSON_NAME_WARNING](https://github.com/thewca/worldcubeassociation.org/blob/f4c8353b0266579772d959ae895c2f158a5e62bd/WcaOnRails/lib/results_validators/persons_validator.rb#L69) which currently checks if the database has a person with same name. But the check is case-sensitive. If the check is case-insensitive, it will be able to catch some more cases. So, this check has to be changed to case-insensitive check.

AxelFl:


WcaOnRails/lib/results_validators/persons_validator.rb
```diff
@@ -65,17 +65,17 @@ def validate(competition_ids: [], model: Result, results: nil)
...
-          existing_person_in_db_by_name.each do |name, persons|
-            @warnings << ValidationWarning.new(:persons, competition_id,
-                                               SAME_PERSON_NAME_WARNING,
-                                               name: name,
-                                               wca_ids: persons.map(&:wca_id).join(", "))
-          end
-        end
         duplicate_newcomer_names = []
         without_wca_id.each do |p|
+          existing_person = Person.where("LOWER(name) = ?", p.name.downcase).group_by(&:name)
```
viroulep, line 70: Thanks a lot for your PR!
In my opinion something needs to be changed here: there used to be one query for all newcomers, which were then grouped by name.
Now there is one query per newcomer, and if I'm understanding this right there is no need for the group_by since the name is known.

Can I suggest to try avoiding doing N queries and keep only one, for instance by doing something like this:
```Person.where("LOWER(name) IN (?)", without_wca_id.map(&:name).map(&:downcase)).group_by(&:name)```?
I believe it retains the same behavior while ignoring the case.

Comment (AxelFl) [reply WcaOnRails/lib/results_validators/persons_validator.rb]: Yup, worked great. Have done basically 0 work with SQL before so I did not think about IN existing.

File: WcaOnRails/lib/results_validators/persons_validator.rb [hunk 2]
```diff
@@ -65,8 +65,8 @@ def validate(competition_ids: [], model: Result, results: nil)
         end
 
         without_wca_id, with_wca_id = persons_by_id.values.partition { |p| p.wca_id.empty? }
-        if without_wca_id.any?
-          existing_person_in_db_by_name = Person.where(name: without_wca_id.map(&:name)).group_by(&:name)
+        existing_person_in_db_by_name = Person.where("LOWER(name) IN (?)", without_wca_id.map(&:name).map(&:downcase)).group_by(&:name)
+        if existing_person_in_db_by_name.any?
```
gregorbg, line 69: Not sure whether an `if any?` check is necessary here. The next line does `.each` which simply won't loop if the result array is empty.

gregorbg (changes requested):
Logic itself looks good, please add one or two test cases.
Preferrably, one test case that was already caught by the old logic, and one test case that would not have been caught by the old logic but is caught by the new logic.

viroulep (approved):
Looks great to me, thanks for PR and the fixes :)

PR #7497, opened 2022-12-27
Exclude Trainee Delegates from WFC Export

### LINKED ISSUE #7493: Exclude Trainee Delegates from WFC Export
With the recent change in structure of the Delegates table (#7395), Trainee Delegates are now included in the Delegates column of the WFC export, which interferes with the processes the WFC has for addressing Dues invoices. It would be preferable to exclude trainee delegates from this column - I assume by doing a join and a filter somewhere in the query. 

See #7144 for an example of recent changes to the WFC export. cc @danieljames-dj since he suspected this might happen.

JonEsparaz:
Fixes #7493. Back at it with the WFC export ;)

@danieljames-dj's [suggestion](https://github.com/thewca/worldcubeassociation.org/issues/7493#issuecomment-1365623663) didn't quite work:
```
     NoMethodError:
       undefined method `reject' for :delegates:Symbol

                           .includes(:delegates.reject{|delegate| !delegate.trainee_delegate}, :championships, :organizers, :events)
                                               ^^^^^^^
```

Thus, I filtered in the view instead of the controller (we do other operations like sorting at this level too). If there's a more elegant way to do this during the query, I'd be grateful for the pointers :)

### DIFF
--- a/WcaOnRails/app/views/wfc/competition_export.csv.erb
+++ b/WcaOnRails/app/views/wfc/competition_export.csv.erb
@@ -11,7 +11,7 @@
   <%= CSV.generate_line([
     c.id, c.name, c.country.iso2, c.continent.id,
     c.start_date, c.end_date, c.announced_at, c.results_posted_at,
-    competition_url(c.id), c.num_competitors, c.delegates.map(&:name).sort.join(","),
+    competition_url(c.id), c.num_competitors, c.delegates.reject(&:trainee_delegate?).map(&:name).sort.join(","),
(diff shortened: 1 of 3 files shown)


danieljames-dj:
Thanks, @JonEsparaz for working on this. I've gone through the changes and it looks good to me. But I'm not so good at rails, so will wait for somebody more experienced for the review. And sorry about my wrong suggestion. 😅

gregorbg (approved):
Thanks for the submission and the additional tests!

PR #7500, opened 2022-12-29
Fix Series associations in form preloading

gregorbg:
This concerns a bug reported by Japanese Delegate Ryutaro Miyazaki via email.
Series with more than two competitions would break depending on the order that you load them in.

Bonus: This also automates assigning an ID and nickname to a newly created (not yet persisted) Series.


gregorbg:
Will probably treat this as emergency fix, i.e. merge without review.
Especially seeing that the current Series implementation is one gigantic hack and will probably be reconsidered when porting the form to React.

PR #7521, opened 2023-01-02
Allow WEAT to edit user profiles

gregorbg:
Per Board request via email


viroulep (approved):
Not even sure this needs to be really reviewed but LGTM!

PR #7526, opened 2023-01-05
Allow non-admins to close registration when the competitor limit is reached.

### LINKED ISSUE #6818: Waiting list limit
**The problem:**
As an organizer right now, sometimes you have a competition where you know the number of spots is low and you will very likely have a large amount of people who will try to register. And if the competition have registrations fees you will need to reimburse everybody who tried to register and was on the waiting list when the registrations closes. The problem with that system is that it gives hope to people that have absolutely no chance to have a spot (for example the competition I'm organizing at the moment is having 51 people on waiting list and the 50 spots are already filled) and also that for every competitor that registers after something like the 10th on waiting list the organizer is just going to pay stripe fees to reimburse them for no reason. This is a serious problem when the waiting list gets way too long because the fees can become so big that it will eventually become bigger than the income - due.

**My solution:**
I think a solution for that problem that could be not to hard to implement would be to add a setting into the competition administration tab to have a maximum size of the waiting list. That way, organizers can avoid people paying to be on waiting list when they have absolutely no chance to get in and the competitor car easily see that it is full because they  will no longer be able to register and see a message like "the waiting list is full, you can't register for the moment".

**Other ideas:**
As the problem is complex there is probably other things that could be done to solve it but I propose this solution because I think it's doable without changing the registration process at all. Other ideas could be to let organizer edit the registration period at any time, or have a button to close it instantly. But I like the waiting list size limit better as it's something automatic and easier to understand and use for both sides.

DanielEgdal:
Giving refunds can be a big expense to organisers, if a waiting list gets too large before WCAT can action the closure. This PR should fix that issue.

Fixes #6818

### DIFF
--- a/WcaOnRails/spec/controllers/competitions_controller_spec.rb
+++ b/WcaOnRails/spec/controllers/competitions_controller_spec.rb
@@ -901,18 +901,18 @@ RSpec.describe CompetitionsController do
       before :each do
         sign_in orga
       end
-      
+
       it "can close registration with full limit" do
-        comp_with_full_reg = FactoryBot.create(:competition, :registration_open, competitor_limit:1)
+        comp_with_full_reg = FactoryBot.create(:competition, :registration_open, competitor_limit: 1)
         comp_with_full_reg.organizers << orga
-        FactoryBot.create_list(:registration, 1, :accepted, :newcomer, competition: comp_with_full_reg) 
+        FactoryBot.create_list(:registration, 1, :accepted, :newcomer, competition: comp_with_full_reg)
         patch :orga_close_reg_when_full_limit, params: { id: comp_with_full_reg }
         expect(response).to redirect_to edit_competition_path(comp_with_full_reg)
         expect(comp_with_full_reg.reload.registration_past?).to eq true
       end
 
       it "cannot close registration non full limit" do
-        comp_without_full_reg = FactoryBot.create(:competition, :registration_open, competitor_limit:100)
+        comp_without_full_reg = FactoryBot.create(:competition, :registration_open, competitor_limit: 100)
         comp_without_full_reg.organizers << orga
         FactoryBot.create_list(:registration, 2, :pending, :newcomer, competition: comp_without_full_reg)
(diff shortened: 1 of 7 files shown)


FinnIckler:
Considering no tests failed, should we add a test for this?

viroulep (changes requested):
For the record this would fix #6818 I think.

Thanks for the PR @DanielEgdal, I'm really looking forward to having this feature as it happens sometimes for us as well and it would be really awesome to cut down refund fees!

The implementation itself could look fine to me, but I believe it opens a behavior the WCAT would not be fine with: it would let organizers/delegates re-open/re-close registrations as they want when the registrations list is full.

There might be a way around this by letting them change the `registration_close` param only if registrations are actually still opened, but I think there is an even simpler solution: don't let organizers/delegates change this date, just provide a button to "close" the registrations with an action that set the `registration_close` param to `Time.now`.

Wether you want to give a shot to the alternate implementation or not, I think the PR should be changed a bit so that changing the registration closing date is not possible when registrations are already close.

DanielEgdal:
Allow should be good now! 
image

3 working test cases added as well.

WcaOnRails/spec/controllers/competitions_controller_spec.rb
```diff
@@ -895,6 +895,46 @@
...
 
+  describe 'POST #orga_close_reg_when_full_limit' do
+    context 'organiser trying to close registration via button' do
+      let(:orga) { FactoryBot.create(:user) }
+      before :each do
+        sign_in orga
+      end
+
+      it "can close registration with full limit" do
+        comp_with_full_reg = FactoryBot.create(:competition, :registration_open, competitor_limit_enabled: true, competitor_limit: 1, competitor_limit_reason: "we have a tiny venue")
```
viroulep, line 906: Thanks for the good laugh xD

viroulep (approved):
Thanks for changing the implementation, it looks good to me!

Before this is merged I believe the commits should be squash into a single commit with a meaningful message (but I believe whoever merge this from WST can do this upon merging using github's UI).

When this goes live I think it's worth to inform WCA Staff that this button exists now :)

FinnIckler:
Thanks daniel!

PR #7163, opened 2022-08-23
Add support for Series competitions

gregorbg:
Fixes #6060

The current implementation in the Edit Competitions form is a little bit hacky, because it's not trivial to edit `1:N` associations from the `1` side. If Rails has some cool shenanigans that I don't know about, please shout! I have tried to document the quirks in the model updates as good as possible.

Workflow: For the "Create Competition" form there is now a second "Nearby Competitions" table that displays all competitions that would in principle qualify as Series Competitions per WCRP. They each show a button to link them to the competition that is currently being created. If the link partner competition is already part of a Series, that Series is inserted and used instead of creating a new one.

The information is displayed in the following places:
- The registration requirements on a competition overview page
- The competitor registration screen
- The "Edit Registrations" table
- The detail screen for editing an individual registration

The following validations are in place:
- For competitions, it is checked that all "sibling competitions" (via the Series) are within 100km and 33 days as per WCRP.
- For a registration, it is checked that only one Series registration is set to status `accepted` at a time. **In particular, this means that it is not possible to accept more than one registration by accident.**

# TODO

- [x] Write test
- [ ] Collect feedback
- [x] Performance optimisations for the `registrations` table

# Screenshots

## Choice in the Competition Edit form
2022-08-24T01:56:44,542685017+09:00
2022-08-24T01:57:42,033638781+09:00

## Details once a Series has been selected in the Competition Edit form
2022-08-24T01:58:39,500011152+09:00

## Hints and warnings for competitors
2022-08-24T01:59:15,753814867+09:00
2022-08-24T02:03:59,067233167+09:00

## Hints, warnings and errors for organisers
2022-08-24T02:07:03,008446174+09:00
2022-08-24T02:07:36,085355079+09:00
2022-08-24T02:10:50,924563828+09:00

PR #7534, opened 2023-01-14
Make order of competition form more intuitive

DanielEgdal:
Per request of some WCAT members, I have made the form order more intuitive. As well, I have added more clear display of the competitor limit as it was in the past according to Wilson.

Additionally when reordering, I noticed some bug (not caused but this PR) that I think was introduced from https://github.com/thewca/worldcubeassociation.org/pull/7163. The first table (from competitions/:id/edit/admin) which has the label "Nearby competitions (within 5 days and 10 km)" actually has content for (NEARBY_DAYS_WARNING, NEARBY_DISTANCE_KM_WARNING), which is 180 days and 250km currently. I couldn't figure out where the variables are actually be called to some function, just something that seemed rather hacky that I couldn't update without breaking a lot of other stuff.

### DIFF
--- a/WcaOnRails/app/views/competitions/_competition_form.html.erb
+++ b/WcaOnRails/app/views/competitions/_competition_form.html.erb
@@ -160,17 +160,6 @@
           <%= render "nearby_competitions" %>
         </div>
 
-        <%= f.input :information, input_html: { class: disable_form ? "" : "markdown-editor" } %>
-        <%= f.input :staff_delegate_ids, as: :user_ids, only_staff_delegates: true %>
-        <%= f.input :trainee_delegate_ids, as: :user_ids, only_trainee_delegates: true %>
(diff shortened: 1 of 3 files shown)


FinnIckler:
Can you look at the failing tests? Are they related to your changes? Especially the js ones

DanielEgdal:
I fixed one error, but I can't see why the current cases are failing. When trying to run stuff locally, the one at 63 works fine and the 113 fails because of not having a chrome driver. So I don't quite understand what would be the cause for the server.

FinnIckler:
I reran the tests and it works now. Reviewing...

FinnIckler (approved):
Really like we finally are putting the competitor limit somewhere more visible, I always struggle looking through the registration requirements.
LGTM

PR #7535, opened 2023-01-14
Password reset email rewording

danieljames-dj:
Current password reset email keeps on going into spam folder and for some it's not even reaching at least spam folder. We were not able to figure out the reason for the same. The email is marked by Gmail and the most probable reason for it might be the content. I've tried changing the content of password reset email, making it more 'best practice of password reset email' so that we can expect this won't go to spam anymore.


viroulep (commented):
I can't comment on the actual content of the email, but as far as the code goes it looks like a clean implementation to me!

FinnIckler:
Am I reading this correctly that before the text of the email came from Devise directly?

danieljames-dj:
Thanks @viroulep.

@FinnIckler Yes, I found the same wordings of our email [here](https://github.com/tigrish/devise-i18n/blob/master/rails/locales/en.yml).

FinnIckler (approved):

PR #7536, opened 2023-01-16
Remove .lvimrc

dunkOnIT:
This is a trivial config file for LunarVim I believe - doesn't need to be in the main repo, lvim users should have their own config.


FinnIckler (approved):
Can't believe this has been here for 7 years..

PR #7551, opened 2023-01-21
Dynamically select calendar text color based on background

### LINKED ISSUE #7276: Make schedule text readable on light color rooms
For CubingUSANationals2023, we'd like to have one stage color be yellow. But because the calendar view uses white text, it's pretty unreadable on a yellow background.

Either the "chroma" gem (https://github.com/jfairbank/chroma#other-methods) or "chroma.ts" in typescript (https://github.com/NaridaL/chroma.ts#colortextcolor-color-src) should distinguish between light and dark colors, and use black or white text accordingly.

### LINKED ISSUE #7549: Allow color customization of text in room schedules
image
Currently if a white background is used for a room on a competition website schedule then it becomes very difficult to read the text. It would be great to have some text customization options (even if the options are just white and black)

JonEsparaz:
Closes #7549 and closes #7276.

Converts background HEX color code to RGB, then computes overall intensity (how "light" or "dark" the color is) to determine if we should use white or black for the text color. 

Note: as per https://stackoverflow.com/a/3943023, there are _some_ cases where this _could_ break W3C rules. I'm an advocate for web accessibility, but I think competition organizers are capable of selecting background colors that provide sufficient contrast against `#ffffff` and `#000000`. If this is really a concern and we're ok with a bulkier solution, we can find a library that handles the color intensity computation.

White background:

image

Yellow background:

image

Cream background (color used in #7549):

image

### DIFF
--- a/WcaOnRails/app/webpacker/lib/show-schedule.js
+++ b/WcaOnRails/app/webpacker/lib/show-schedule.js
@@ -17,7 +17,7 @@ const HEX_BASE = 16;
  *
  * @param {string} hexColor HEX color code to convert to RGB
  *
- * @returns {Array<int>} RBG values, defaults to `[0, 0, 0]` if `hexColor` cannot be parsed
+ * @returns {Array<number>} RBG values, defaults to `[0, 0, 0]` if `hexColor` cannot be parsed
  */
 const hexToRgb = (hexColor) => {
   if (/#[0-9A-Fa-f]{6}/.test(hexColor)) {
--- a/WcaOnRails/app/webpacker/lib/show-schedule.js
+++ b/WcaOnRails/app/webpacker/lib/show-schedule.js
@@ -16,6 +16,10 @@ const HEX_CHANNEL_REGEX = /^#(?<r>[0-9A-Fa-f]{2})(?<g>[0-9A-Fa-f]{2})(?<b>[0-9A-
 /**
  * Convert a HEX color code to RGB values.
  *
+ * @example
+ * // returns [255, 255, 255]
+ * getTextColor('#ffffff');
+ *
  * @param {string} hexColor HEX color code to convert to RGB
  *
  * @returns {Array<number>} RBG values, defaults to `[0, 0, 0]` if `hexColor` cannot be parsed
@@ -40,6 +44,10 @@ const BLACK = '#000000';
 /**
  * Compute appropriate text color (black or white) based on how "light" or "dark"
  * the background color of a calendar item is.
+ *
+ * @example
+ * // returns #000000 (black given white background color)
+ * getTextColor('#ffffff');
  *
  * @param {string} backgroundColor Calendar item's background color (in HEX)
  *
--- a/WcaOnRails/app/webpacker/lib/show-schedule.js
+++ b/WcaOnRails/app/webpacker/lib/show-schedule.js
@@ -11,6 +11,7 @@ window.wca.registerVenueData = (id, venueData) => {
 };
 
 const HEX_BASE = 16;
+const HEX_CHANNEL_REGEX = /^#(?<r>[0-9A-Fa-f]{2})(?<g>[0-9A-Fa-f]{2})(?<b>[0-9A-Fa-f]{2})$/;
 
 /**
  * Convert a HEX color code to RGB values.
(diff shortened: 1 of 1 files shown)


WcaOnRails/app/webpacker/lib/show-schedule.js
```diff
@@ -10,6 +10,44 @@ window.wca.registerVenueData = (id, venueData) => {
...
+
+/**
+ * Convert a HEX color code to RGB values.
+ *
+ * @param {string} hexColor HEX color code to convert to RGB
+ *
+ * @returns {Array<number>} RBG values, defaults to `[0, 0, 0]` if `hexColor` cannot be parsed
+ */
+const hexToRgb = (hexColor) => {
+  if (/#[0-9A-Fa-f]{6}/.test(hexColor)) {
```
FinnIckler, line 23: I have no preference, but if you use the regex hexcolor.match(/.{1,2}/g) you can get all three parts of the string directly see https://convertingcolors.com/blog/article/convert_hex_to_rgb_with_javascript.html

WcaOnRails/app/webpacker/lib/show-schedule.js
```diff
@@ -10,6 +10,44 @@ window.wca.registerVenueData = (id, venueData) => {
...
+/**
+ * Compute appropriate text color (black or white) based on how "light" or "dark"
+ * the background color of a calendar item is.
+ *
+ * @param {string} backgroundColor Calendar item's background color (in HEX)
+ *
+ * @returns {string} white for "dark" backgrounds, black for "light" backgrounds
+ */
+const getTextColor = (backgroundColor) => {
+  const [red, green, blue] = hexToRgb(backgroundColor);
```
FinnIckler, line 46: can you add a toLowerCase() to backgroundColor because it is not clear for the user (if they want to use this function in any other code) if the Color is supposed to be using lower or uppercase letters (I just did it wrong testing it)

FinnIckler (changes requested):

WcaOnRails/app/webpacker/lib/show-schedule.js
JonEsparaz: Sure, I'll add `.toLowerCase()` to makes things clear, but I'm curious if you have an example of the function breaking?

The regex in `hexToRgb` handles lowercase characters and `parseInt` interprets uppercase and lowercase the same way:

```js
> const HEX_BASE = 16;
undefined
> const hexToRgb = (hexColor) => {
...   if (/#[0-9A-Fa-f]{6}/.test(hexColor)) {
.....     return [
.....       parseInt(hexColor.slice(1, 3), HEX_BASE),
.....       parseInt(hexColor.slice(3, 5), HEX_BASE),
.....       parseInt(hexColor.slice(5, 7), HEX_BASE),
.....     ];
.....   }
...
...   return [0, 0, 0];
... };
undefined
> hexToRgb("#Fa8bCd");
[ 250, 139, 205 ]
> hexToRgb("#FA8CCD");
[ 250, 140, 205 ]
> hexToRgb("#fa8ccd");
[ 250, 140, 205 ]
> parseInt("FC", HEX_BASE);
252
> parseInt("fc", HEX_BASE);
252
> parseInt("Fc", HEX_BASE);
252
>
```

WcaOnRails/app/webpacker/lib/show-schedule.js
JonEsparaz: I'd like to keep the regex more specific. Would you be ok with `hexColor.match(/^#([0-9A-Fa-f]{2})([0-9A-Fa-f]{2})([0-9A-Fa-f]{2})$/)`, then I can use the capture groups?

WcaOnRails/app/webpacker/lib/show-schedule.js
FinnIckler: Ah sorry, I tested it like this in the console
```
const colors = ["#FFFFFF", "#AAAAAA", "#99AD32"]
for (color in colors){
console.log(getTextColor(color))
}
```
and made the error that color is not the value but the array index... 
So it's actually correct, sorry for that

WcaOnRails/app/webpacker/lib/show-schedule.js
FinnIckler: yeah sure

PR #7554, opened 2023-01-22
Add QiYi Display to default Delegate Report Text

EdHollingdale:
QiYi Displays are being increasingly used. This adds it as a default option to save time for Delegates when completing the report.


FinnIckler (approved):
LGTM

PR #7556, opened 2023-01-22
Add lib-yaml-dev dependecy

FinnIckler:
With the recent upgrade to Version 5 of Psych we need lib-yaml as a dependency as it doesn't come with it's own bindings anymore. I added it both to chef and to Docker setups, additionally I added a reference to installing it in the readme if you are running the ruby portion of the website only (we already had some questions about an error in slack)

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,9 +52,10 @@ This repository contains all of the code that runs on [worldcubeassociation.org]
   Feel free to take a look at our [chef recipe](https://github.com/thewca/worldcubeassociation.org/blob/master/chef/site-cookbooks/wca/recipes/default.rb#L6-L23)
   for the accurate versions we use and how we install them.
   Please note that other versions may work, but it is not guaranteed.
-  3. `bundle install && bin/yarn`
-  4. `bin/rake db:load:development` - Download and import the [developer's database export](https://github.com/thewca/worldcubeassociation.org/wiki/Developer-database-export). Depending on your computer it may take a long time. Alternatively you can run `bin/rake db:reset` which will create the database and seed it with random data (it's way faster, but less representative of our website content).
-  5. `bin/rails server` - Run rails. The server will be accessible at localhost:3000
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -52,9 +52,10 @@ This repository contains all of the code that runs on [worldcubeassociation.org]
   Feel free to take a look at our [chef recipe](https://github.com/thewca/worldcubeassociation.org/blob/master/chef/site-cookbooks/wca/recipes/default.rb#L6-L23)
   for the accurate versions we use and how we install them.
   Please note that other versions may work, but it is not guaranteed.
-  3. `bundle install && bin/yarn`
-  4. `bin/rake db:load:development` - Download and import the [developer's database export](https://github.com/thewca/worldcubeassociation.org/wiki/Developer-database-export). Depending on your computer it may take a long time. Alternatively you can run `bin/rake db:reset` which will create the database and seed it with random data (it's way faster, but less representative of our website content).
-  5. `bin/rails server` - Run rails. The server will be accessible at localhost:3000
+  3. Install `libyaml` using your package manager of choice, eg. [Mac](https://formulae.brew.sh/formula/libyaml) or [Ubuntu](https://launchpad.net/ubuntu/+source/libyaml)
```
gregorbg, line 55: In the recipe+Dockerfile below, you install `libyaml-dev`. But here in the Readme, you recommend `libyaml`. I think that these are two different packages.

gregorbg (commented):
Cool! Minor comment about the Readme.

Also, it feels weird that we have to tell our users to install stuff for a transitive dependency that is included with Ruby. Perhaps we are missing something? Would it be worth posting an issue against `psych` to ask them for support?

README.md
FinnIckler: So libyaml-dev are only header files, which are required for psych, in brew you can't actually install the header files only, which is why you install the whole package. For Ubuntu users I link the whole package and the user can choose to install 
the whole package or just the header files.

gregorbg (approved):

PR #7560, opened 2023-01-29
Add check-in and multi cube submission as custom calendar activities

kr-matthews:
The validation logic assumes the activity code stars with `other` and has a single `-`, so I used `other-checkin` and `other-multi`.
Also changed the registration text to include "On-site".

### DIFF
--- a/WcaOnRails/app/models/schedule_activity.rb
+++ b/WcaOnRails/app/models/schedule_activity.rb
@@ -3,7 +3,7 @@
 class ScheduleActivity < ApplicationRecord
   # See https://docs.google.com/document/d/1hnzAZizTH0XyGkSYe-PxFL5xpKVWl_cvSdTzlT_kAs8/edit#heading=h.14uuu58hnua
   VALID_ACTIVITY_CODE_BASE = (Event.official.map(&:id) + %w(other)).freeze
-  VALID_OTHER_ACTIVITY_CODE = %w(registration breakfast lunch dinner awards unofficial misc tutorial).freeze
+  VALID_OTHER_ACTIVITY_CODE = %w(registration checkin breakfast lunch dinner awards unofficial misc tutorial).freeze
   belongs_to :holder, polymorphic: true
   has_many :child_activities, class_name: "ScheduleActivity", as: :holder, dependent: :destroy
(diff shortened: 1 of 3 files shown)


FinnIckler:
Does renaming affect any localization strings?

kr-matthews:
> Does renaming affect any localization strings?

I couldn't find any localization for any of the 'other-' calendar activities, so as far as I can tell, no. But I haven't worked with localization stuff before so I'm not 100% sure.

FinnIckler:
mh.. I have the feeling we need a basic guide on what should and what shouldn't be localised.

WcaOnRails/app/models/schedule_activity.rb
```diff
@@ -60,7 +60,7 @@ def localized_name(rounds_by_wcif_id = {})
     parts = ScheduleActivity.parse_activity_code(activity_code)
     if parts[:event_id] == "other"
       # TODO/NOTE: should we fix the name for event with predefined activity codes? (ie: those below but 'misc' and 'unofficial')
-      # VALID_OTHER_ACTIVITY_CODE = %w(registration breakfast lunch dinner awards unofficial misc).freeze
+      # VALID_OTHER_ACTIVITY_CODE = %w(registration checkin multi breakfast lunch dinner awards unofficial misc).freeze
```
FinnIckler, line 63: Anyone now why this is commented out and what this todo means in this context?

FinnIckler (commented):
Technically it's fine, would love to know though if we are planning to localise these OTHER_ACTIVITY_CODE and what that comment means

WcaOnRails/app/models/schedule_activity.rb
viroulep: I think the idea behind it was to provide a consistent translation for the most common "other" activities (eg: registrations, awards, lunch, etc); basically I didn't go through with it because I was unsure who we wanted this to appear and I think because it implied locking the "other" activity name input for some activities and it was not a priority in creating the schedule tool.

As for why this line is commented out I think it was just so that there was a reference to the current "other" activity code we know about ^^

FinnIckler (approved):
LGTM

PR #4682, opened 2019-10-05
Prevent refund for old registrations

viroulep:
Fixes #3224.
This change addresses a potential financial loss if an organizer or WCA staff decides to go crazy.
I initially prevented this at a controller level, but I realized some WCA staff may change the competition's date and overrun a check doing roughly `current_date > competition_date + 21.days`.
So instead I created a job that clears the `connected_stripe_account_id` after an arbitrary amount of days (I set it to 21).

This way there is no way for anyone to issue refund, except by going directly to the connected Stripe account dashboard.

cc @Mollerz, @ChrisWrightWCA, @EdHollingdale

PR #7562, opened 2023-01-31
Allow competition admins to disconnect Stripe

DanielEgdal:
It happens rather often that an organiser has set up an incorrectly configured Stripe account, or that the Stripe account gets blocked for receiving payments. 
As such, WCAT gets request to remove Stripe accounts for the organiser to set up a new one, but we cannot currently do this and the WRT has to manually go in the DB and clear it. This PR will make that workflow a lot easier.


FinnIckler:
Hey Daniel,
Thanks for looking into it. If I am reading this correctly, couldn't this cause a state where the stripe dashboard and the WCA website are out of sync? 
This is of course an issue with our approach how we link and unlink competitions through the stripe dashboard. Would be great to make sure that the account has been unlinked before removing it from the WCA page. Maybe there is a Stripe API we can use?

DanielEgdal:
Hi Finn, I have never worked with Stripe from the admin view as I never handle Stripe finances. 
I just know that this adds a button for the current workflow done by WCAT/WRT.
What issues could be caused in the case the organiser neglects to disconnect their account before WCAT is requested to press this button?

FinnIckler:
Nothing particularly bad happens (I hope it would just raise some errors), but I'd like to have a single source of truth, especially when it comes to finance related data. 
Stripe issues (especially payment issues) are mostly found out only when there is already money in transit, which makes the situation even more important.
There is a bunch of undefined behaviour that I am not sure about what would happen if the state of stripe and WCA does not match, like what would happen if someone tries to refund a competitor. 
Of course we currently already have this issue when stripe accounts are unlinked and then have to be manually changed on the WCA page, but at least the stripe side will always be done first. 
  
If it's not possible to automate this, there should at least be a message that this should only be done after the account is unlinked on stripe.

DanielEgdal:
I think @viroulep has previously worked with Stripe sections of the website (e.g. https://github.com/thewca/worldcubeassociation.org/pull/4682) and can maybe give some feedback to your questions? This PR does the same with setting the current Stripe id to nil, but of course no new refunds are performed afterwards in those cases.

The text in [payment_setup.html.erb](https://github.com/thewca/worldcubeassociation.org/pull/7562/files#diff-3baf23985a0ef72578a98fa6bb750ce822464a69a8ad218c7bf9a32598056726) is saying that the Stripe dashboard disconnect must be done first, like you mention at last.

gregorbg:
We need more thorough investigation and design decision regarding what "disconnecting Stripe" actually means. I do not have the capacity to investigate at the moment, if anyone else is interested feel free.

EdHollingdale:
> couldn't this cause a state where the stripe dashboard and the WCA website are out of sync? 

My current understanding is that this already happens. If you go to your stripe dashboard and remove permissions for the WCA site it is disconnected but there is no way for the WCA website to know that. As far as the WCA database is concerned you still have stripe connected for that comp, when in reality once you have disconnected it on the stripe dashboard it shouldn't work. 

We should maybe ping someone on WRT since they usually action disconnecting the stripe for a specific comp, but my understanding is that this is just to remove it from the WCA site so that another stripe can be connected. It also gives you the option to change a stripe account on a specific competition without removing permissions for _all_ competitions by disconnecting the WCA site completely (what you can do in the stripe dashboard).

DanielEgdal:
@thewca/wrt as Ed suggested.

Jambrose777:
> > couldn't this cause a state where the stripe dashboard and the WCA website are out of sync?
> 
> My current understanding is that this already happens. If you go to your stripe dashboard and remove permissions for the WCA site it is disconnected but there is no way for the WCA website to know that. As far as the WCA database is concerned you still have stripe connected for that comp, when in reality once you have disconnected it on the stripe dashboard it shouldn't work.
> 
> We should maybe ping someone on WRT since they usually action disconnecting the stripe for a specific comp, but my understanding is that this is just to remove it from the WCA site so that another stripe can be connected. It also gives you the option to change a stripe account on a specific competition without removing permissions for _all_ competitions by disconnecting the WCA site completely (what you can do in the stripe dashboard).

This is correct

WcaOnRails/app/views/competitions/payment_setup.html.erb
```diff
@@ -28,8 +28,16 @@
   <% else %>
     <p>
       This competition is set up to accept payments through Stripe.
-      If you ever want to disconnect your account, you can do so from your Stripe account dashboard
+      If you want to disconnect your account, you will first do so from your Stripe account dashboard
```
gregorbg, line 31: After reading the discussion under the PR: Is this really technically necessary? Or do we just _prefer_ if the competitor first disconnects their Stripe Dashboard?

In any case, this phrasing of "you will first do..." sounds a bit harsh (to my non-native ears). What about "you should first do so from your Stripe dashboard"?

gregorbg (changes requested):
Code is okay, some comments about the i18n text and phrasing though.

WcaOnRails/app/views/competitions/payment_setup.html.erb
FinnIckler: The order doesn't matter, but both should be done so a new account can be associated.

DanielEgdal: I updated it based on your phrasing.

DanielEgdal:
Requested changes should be done now.

gregorbg (approved):

PR #7563, opened 2023-01-31
Prepend competition series fields with "Series"

### LINKED ISSUE #7559: Unclear error message when naming a new competition series
When the name of a series does not meet the naming requirements, you get an error message. However the message text is unclear.

While creating the series Cubing with the Cats, I forgot to add the year to the end of the series name, and so received this warning:

image

In the image you can see that the name of the competition is fine, but the error message would lead you to believe otherwise. It's not clear from the error message that it is the **series** name that needs amending.

Changing the error message to specify the series name would make things clearer:

**Series Name** must end with a year and must contain only alphanumeric characters, dashes(-), ampersands(&), periods(.), colons(:), apostrophes('), and spaces( )

JonEsparaz:
Changes "ID" to "Series ID" and "Name" to "Series name". Closes #7559 (for English only).

Example with incorrect competition _series_ name:

series name

I don't think there's a great way to unit test this. (I could be wrong - please let me know!)

### DIFF
--- a/WcaOnRails/config/locales/en.yml
+++ b/WcaOnRails/config/locales/en.yml
@@ -350,8 +350,8 @@ en:
       championship:
         championship_type: "Championship type"
       competition_series:
-        wcif_id: "ID"
-        name: "Name"
+        wcif_id: "Series ID"
+        name: "Series name"
         short_name: "Series nickname"
         competition_ids: "All competitions"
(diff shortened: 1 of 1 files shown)


FinnIckler (approved):
LGTM, I hope the translators get notified when something like this changes?

Issue #7587, opened 2023-02-01
"Export to calendar" not specifying time zone


timreyn:
When I download the ics calendar export from a competition in my time zone (America/New_York, currently UTC-5), the ics file does not specify a timezone. The `DTSTART` / `DTEND` fields are in UTC, but do not specify that the time zone is UTC, so they're interpreted as local time.

e.g. for NorthamptonWinter2023A, this looks like

```
BEGIN:VEVENT
DTSTAMP:20230201T122034Z
UID:d8a70ad1-9aee-47ee-a2d0-c7734e03bdb6
DTSTART:20230225T124500
DTEND:20230225T130000
SUMMARY:Registration
END:VEVENT
```

When I import this to Google Calendar, it shows up at 12:45 PM in my local time zone (should be 7:45 AM).

FinnIckler:
Do you know if that problem has been always there of if was introduced in https://github.com/thewca/worldcubeassociation.org/pull/7490 ?

timreyn:
It looks like the export of individual events was introduced in #7490, so this issue wouldn't have appeared before then. (This was reported by a competitor; I've never used the export myself, but that's just my understanding based on reading that PR).

FinnIckler:
For implementation: https://github.com/icalendar/icalendar#timezones
@AxelFl you want to add this?

AxelFl:
Yeah sure, I think I only created test competitions and always used UTC.

PR #7590, opened 2023-02-01
Add timezone info for events in calendar export

AxelFl:
Fixes issue #7587 where calendar events are created as UTC but read by most calendar programs as the local timezone.


AxelFl:
No problem, just a very dumb and european mistake.

PR #7592, opened 2023-02-02
Allow Banned Competitors to register for competitions after their ban expires

### LINKED ISSUE #7591: Banned Competitors cannot register to competitions that occur after their ban expires
It was recently brought to the WDC's attention that banned competitors are unable to register for any competitions while they are banned, even competitions past the expiration date of their ban. With many competitions having registration open and filling months before actual competitions, some competitors are at risk of a "prolonged" ban as they may be unable to compete for several months even after their ban has expired. 

We would like this resolved by restricting banned competitors from registering for competitions scheduled to take place during their ban period while leaving the banned competitors free to register for any competitions outside of their ban period.

FinnIckler:
Fixes #7591. 
I decided to remove the banned? check from `cannot_register_for_competition_reasons` as it's also used in the notification helper line 55 where it doesn't make sense. Instead I implemented a check directly in `user_can_register_for_competition`.
This ruby code looks pretty ugly, so I would love to get some suggestions on what is better!

### DIFF
--- a/WcaOnRails/app/models/registration.rb
+++ b/WcaOnRails/app/models/registration.rb
@@ -226,8 +226,8 @@ class Registration < ApplicationRecord
   private def user_can_register_for_competition
     if user&.cannot_register_for_competition_reasons.present?
       errors.add(:user_id, user.cannot_register_for_competition_reasons.to_sentence)
-    elsif user.banned?
-      ban_end = user.current_team_members.select(:name == "Banned Competitors").first.end_date
+    elsif user&.banned?
+      ban_end = user.current_team_members.select(:team == Team.banned).first.end_date
       if !ban_end.present? || competition.start_date < ban_end
         errors.add(:user_id, I18n.t('registrations.errors.banned_html').html_safe)
       end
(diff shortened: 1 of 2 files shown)


simonkellly:
Would cannot_be_undeleted_when_banned in registration.rb need to be altered also?

FinnIckler:
Mh... Wouldn't a situation for this to be relevant look something like
- competitor registered for a competition
- gets banned and deleted from registration
- the ban ends before the competition starts so he reregisters

Could this happen?

FinnIckler:
I also just noticed that this breaks for competitors that are indefinitely banned, so I will update the code

FinnIckler:
pushed some fixes

WcaOnRails/app/models/registration.rb
```diff
@@ -226,6 +226,11 @@ def self.wcif_json_schema
   private def user_can_register_for_competition
     if user&.cannot_register_for_competition_reasons.present?
       errors.add(:user_id, user.cannot_register_for_competition_reasons.to_sentence)
+    elsif user.banned?
+      ban_end = user.current_team_members.select(:name == "Banned Competitors").first.end_date
```
viroulep, line 230: I think using the team name here is quite risky, I'd rather use a more reliable way like:
```suggestion
      ban_end = user.current_team_members.select { |m| m.team == Team.banned }.first.end_date
```
I didn't actually run the code so I may have messed up something, but hopefully you get my point :)

FinnIckler: thanks, implemented

FinnIckler:
Tests run through now

viroulep (approved):
LGTM, but please squash the commit before merging ;)

PR #7595, opened 2023-02-03
Update the API help page to clarify usage

danielpunkass:
Update the API help page to clarify the need for client_id and client_secret parameters, and that the username should typically be a user's email. I assume the software_team/wca username and password are more useful for members of the WCA admin team who are more likely to already know how to use the API. The aim of this PR is to make the API more accessible to members of the public who are interesting in exploring its capabilities.

### DIFF
--- a/WcaOnRails/app/views/api/v0/api/help.html.erb
+++ b/WcaOnRails/app/views/api/v0/api/help.html.erb
@@ -5,26 +5,24 @@
 
   <h2>OAuth</h2>
   <p>
-    Go to <%= link_to oauth_applications_path, oauth_applications_path %> to
-    manage your OAuth applications.
+    <%= t('api.manage_oauth_html', oauth_link: link_to(oauth_applications_path, oauth_applications_path)) %>
   </p>
   <p>
-    Here is an example, using curl, of obtaining an access token using an Application ID, Client Secret, your WCA email, and your password:
+    <%= t('api.token_request_intro') %>
   </p>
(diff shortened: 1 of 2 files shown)


FinnIckler:
While we at it, can you wrap these with an i18n call so we can localise this?

danielpunkass:
I am new to ERB and JavaScript localization, so I hope I put things in the right place and used the right technique.

FinnIckler:
Thanks, it does look a bit weird localizing constants like: 'APP_ID',  'APP_SECRET', 'YOUR_EMAIL' and 'YOUR_PASSWORD', but maybe this is just my bias always coding in english. Thoughts?

danielpunkass:
> Thanks, it does look a bit weird localizing constants like: 'APP_ID', 'APP_SECRET', 'YOUR_EMAIL' and 'YOUR_PASSWORD', but maybe this is just my bias always coding in english. Thoughts?

The reason I chose to localize them is because they are not constants, per se, but placeholders meant to instruct the reader what should be provided there in lieu of the placeholder. So in French for example it might localize `YOUR_EMAIL` to `VOTRE_EMAIL` to make it more obvious to a French developer what value is expected.

FinnIckler:
I think you are right for 'YOUR_EMAIL' and 'YOUR_PASSWORD', but APP_ID and APP_SECRET might be mistranslated by community members that are not developers. I Just looked at some translations done for https://www.worldcubeassociation.org/oauth/applications/ where application id and secret are already translated and the German version is wrong (which I should do a PR for...). If these are translated and they don't match it could add new confusion.

danielpunkass:
I defer to you on that! I updated the PR with no localization for those terms.

FinnIckler (approved):
LGTM, Thanks!

Issue #7599, opened 2023-02-04
Cannot create competition for San Juan, Puerto Rico in the United States


dunkOnIT:
Source email thread: "Create Competition Page: Puerto Rico not recognized"

> I'm trying to create a competition for San Juan, Puerto Rico and the website will not allow me to do so. Puerto Rico is a territory, so not a state or country, however it is part of the U.S. Can we please fix this so this competition can be created please?

I've replicated this on my end - selecting United States as the country (there is no Puerto Rico), entering "San Juan" or "San Juan, Puerto Rico" as the city both trigger the following error: "City name Puerto Rico is not a valid state", which appears at the top of the screen, or the following error directly below the "City name" field:

![image](https://user-images.githubusercontent.com/52967253/216751481-acd9b47c-a74f-4295-9ad7-0078932964fb.png)

I think the desired solution would be that "San Juan, Puerto Rico" is recognised as a valid City, State combination (and optionally we call it a "City, State/Territory" combination) - is there a list of states that we can edit? Or is this provided by a library?

PR #7600, opened 2023-02-04
Update US city validator to include US territories

simonkellly:
This should resolve issue #7599 with the city name being in the format "San Juan, Puerto Rico"


FinnIckler:
looks good, I guess the type_of_region: "state" is not accurate anymore, but I am not sure what happens if you rename this?

simonkellly:
It looks like it was only used in the error text for the validation.

I had a look on some e-commerce sites (TheCubicle and Amazon US) and they both just used state, so I left it as is

FinnIckler (approved):
Ok then, LGTM!

Issue #7597, opened 2023-02-03
Add a separate tab for cancelled competitions under "My Competitions"


dunkOnIT:
Source email thread: "Canceled Competitions"

> I was recently looking at my "My Competitions" page on my profile and "Past Competitions" included all of my cancelled ones from covid. I had one competition, Puget Sound Valentine 2019, that was cancelled prior to Covid in which a WCAT member just unconfirmed the competition and therefore it is not showing in my "Past Competitions" list. However, I was curious if those cancelled competitions during covid would remain on that page indefinitely or if they eventually would/could be removed since they did not happen?

PR #7602, opened 2023-02-04
Dont show cancelled comps in my past comps table

DanielEgdal:
From the original email, this solves https://github.com/thewca/worldcubeassociation.org/issues/7597

### DIFF
--- a/WcaOnRails/app/controllers/competitions_controller.rb
+++ b/WcaOnRails/app/controllers/competitions_controller.rb
@@ -587,6 +587,9 @@ class CompetitionsController < ApplicationController
     if current_user.person
       competition_ids.concat(current_user.person.competitions.pluck(:competitionId))
     end
+    # An organiser might still have duties to perform for a cancelled competition until the date of the competition has passed. 
+    # For example, mailing all competitors about the cancellation. 
+    # In general ensuring ease of access until it is certain that they won't need to frequently visit the page anymore.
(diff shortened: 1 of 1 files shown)


WcaOnRails/app/controllers/competitions_controller.rb
```diff
@@ -588,7 +588,7 @@ def my_competitions
       competition_ids.concat(current_user.person.competitions.pluck(:competitionId))
     end
     competitions = Competition.includes(:delegate_report, :delegates)
-                              .where(id: competition_ids.uniq)
+                              .where(id: competition_ids.uniq).where("cancelled_at is null or end_date >= curdate()")
```
FinnIckler, line 591: why do we need the `end_date >= curdate()` check here?

DanielEgdal: The logic is that an organiser might still have duties to perform for a competition until the date of the competition has passed. For example, mailing all competitors about the cancellation. In general ensuring ease of access until it is certain that they won't need to frequently visit the page anymore.

FinnIckler: ah ok, can you write a comment about that?

DanielEgdal: Done

FinnIckler (approved):
LGTM!

PR #7603, opened 2023-02-04
Add option for organisers to force competitors to make a comment.

DanielEgdal:
Particularly a useful feature for FM simultaneous competitions, as competitors forgetting to enter a location is an annoying hunt-down process.


WcaOnRails/app/models/registration.rb
```diff
@@ -253,6 +253,16 @@ def self.wcif_json_schema
     end
   end
 
+  validate :forcing_competitors_to_add_comment
+  private def forcing_competitors_to_add_comment
+    if !competition.force_comment_in_registration
+      return
+    end
+    if comments.length == 0
```
kr-matthews, line 261: Might be worth trimming/stripping the comment to make sure it's not just whitespaces? `comments.strip.length == 0`

DanielEgdal: Good idea, done.

WcaOnRails/config/locales/en.yml
```diff
@@ -338,6 +339,7 @@ en:
         refund_policy_limit_date: "Limit date for refunds"
         waiting_list_deadline_date: "Deadline for accepting competitors from the waiting list"
         event_change_deadline_date: "Deadline for competitors to update their registered events"
+        # force_comment_in_registration: "Force competitors to "
```
kr-matthews, line 342: This can be deleted?

kr-matthews:
I'm not very familiar/comfortable with Ruby, but this looks good to me and it seems to work as expected locally.

WcaOnRails/config/locales/en.yml
DanielEgdal: Good catch, done

WcaOnRails/app/models/registration.rb
```diff
@@ -263,6 +263,16 @@ def self.wcif_json_schema
     end
   end
 
+  validate :forcing_competitors_to_add_comment
+  private def forcing_competitors_to_add_comment
+    if !competition || !competition.force_comment_in_registration
```
FinnIckler, line 268: I see you took this from `cannot_register_for_unqualified_events`, but I think the ruby-way of doing it would be `competition&.force_comment_in_registration.blank?` 
The &. operator is a 'safe access' Operator and returns nil if competition is nil. I am also fairly new to ruby, but that's how it's done in a lot of other code I saw

FinnIckler (changes requested):

WcaOnRails/app/models/registration.rb
DanielEgdal: done, still works locally after that change

WcaOnRails/app/models/registration.rb
```diff
@@ -263,6 +263,16 @@ def self.wcif_json_schema
     end
   end
 
+  validate :forcing_competitors_to_add_comment
+  private def forcing_competitors_to_add_comment
+    if competition&.force_comment_in_registration.blank?
+      return
+    end
+    if comments.strip.empty?
```
FinnIckler, line 271: now that I am looking at this with a fresher morning brain, I don't think you need both ifs anymore.
 ```
if competition&.force_comment_in_registration.present? && comments.strip.empty?
      errors.add(:user_id, I18n.t('registrations.errors.cannot_register_without_comment'))
    end
```

FinnIckler (changes requested):
Thanks, now can you add a test for a competition like this?

WcaOnRails/app/models/registration.rb
DanielEgdal: Works too!

DanielEgdal:
Test cases added and working locally

FinnIckler (approved):
LGTM! I'll wait till tomorrow to merge as I'll need to run the migrations too

PR #7606, opened 2023-02-06
Sort tied results by earliest date achieved first

### LINKED ISSUE #7605: Sort tied results by earliest date achieved first
Source email thread: "Tie results sorting"

WRC and WRT are both happy with the proposed change. 

> Ordering by date of achievement is indeed less arbitrary than ordering by name, in particular as you could always argue that achieving a certain results earlier deserves slightly more merit than "just" tying the same result later.

**The change:**
* Currently, tied results are sorted alphabetically. We want this to be changed so that whoever achieved the received on the earlier date would be displayed first. 
* For identical results achieved on the same day, defaulting back to alphabetical sorting seems reasonable.

_Note: Ensure that MBLD sorting prioritises points, then time, then lower number of cubes unsolved, THEN date achieved._

simonkellly:
Fixes #7605

### DIFF
--- a/WcaOnRails/app/controllers/results_controller.rb
+++ b/WcaOnRails/app/controllers/results_controller.rb
@@ -42,7 +42,7 @@ class ResultsController < ApplicationController
         FROM (
           SELECT
             result.*,
-            result.#{value} value
+            result.#{value} value,
             RANK() over (order by result.#{value}) rnk
           FROM (
             SELECT MIN(valueAndId) valueAndId
@@ -73,8 +73,8 @@ class ResultsController < ApplicationController
           FROM (
             SELECT
               result.*,
-              average value
-            RANK() over (order by average) rnk
+              average value,
+              RANK() over (order by average) rnk
             FROM Results result
             #{@gender_condition.present? ? "JOIN Persons persons ON result.personId = persons.id and persons.subId = 1" : ""}
             JOIN Competitions competition on competition.id = competitionId
--- a/WcaOnRails/app/controllers/results_controller.rb
+++ b/WcaOnRails/app/controllers/results_controller.rb
@@ -31,6 +31,27 @@ class ResultsController < ApplicationController
     @is_results = splitted_show_param[1] == "results"
     expanded_limit_condition = "LIMIT #{@show * 2}"
 
+    # Required fields gotten from LightResult
+    result_select = "
+      result.id,
+      result.value1,
+      result.value2,
+      result.value3,
+      result.value4,
+      result.value5,
+      result.best,
+      result.average,
+      result.personName,
+      result.pos,
+      result.personId,
+      result.regionalSingleRecord,
+      result.regionalAverageRecord,
+      result.countryId,
+      result.formatId,
+      result.roundTypeId,
+      result.eventId
+    "
+
     @cache_params = ['rankings', params[:event_id], params[:region], params[:years], params[:show], params[:gender], params[:type]]
 
     if @is_persons
@@ -41,7 +62,7 @@ class ResultsController < ApplicationController
           rnk
         FROM (
           SELECT
-            result.*,
+            #{result_select},
             result.#{value} value,
             RANK() over (order by result.#{value}) rnk
           FROM (
@@ -72,7 +93,7 @@ class ResultsController < ApplicationController
             rnk
           FROM (
             SELECT
-              result.*,
+              #{result_select},
               average value,
               RANK() over (order by average) rnk
             FROM Results result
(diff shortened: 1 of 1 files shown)


DanielEgdal:
Potentially, it also makes sense to change the way limits are done? For example, if several people are tied for 100th, I think all should be displayed instead of just one.

simonkellly:
I think if its ordered properly (as in first achieved is higher in the list), then it's okay to cut off the list once it hits 100, even if they have the same result (as they are not in the top 100 persons). Not sure how WRT / others feel about this though.

FinnIckler:
Do we want to ask WRT first? There is already an email thread going

simonkellly:
Sure, I will email now

FinnIckler:
Seems to be the consensus of WRT that it should be top 100 ranks

simonkellly:
I haven't tested this commit locally yet, as I am currently updating my local db

simonkellly:
Tested it and it is working as expected

FinnIckler:
Doing some performance tests

FinnIckler:
I am getting an `Mysql2::Error: Duplicate column name 'id'` error on `/results/rankings/333/single?show=100+results` when I cloned this. Is it just me?

simonkellly:
It worked on my pc and the tests passed so I'm not sure about what that error is caused by. I did some brief performance tests and it was very similar for simpler queries, and about 50% slower with all filters turned on (not exact numbers).

simonkellly:
I have reproduced that error 😎

WcaOnRails/app/controllers/results_controller.rb
```diff
@@ -29,49 +29,82 @@ def rankings
     @show = splitted_show_param[0].to_i
     @is_persons = splitted_show_param[1] == "persons"
     @is_results = splitted_show_param[1] == "results"
-    limit_condition = "LIMIT #{@show}"
+    expanded_limit_condition = "LIMIT #{@show * 2}"
+
+    # Required fields gotten from LightResult
+    result_select = "
```
FinnIckler, line 35: I generally like that we aren't using result.* anymore, do you know what happens with the duplicate values that are generated here?

simonkellly: Duplicate values?

simonkellly:
BTW this is still not working, but I wasn't able to properly test it while writing the code. I also want to remove some of the code duplication within the different queries, as it makes it really annoying to edit

WcaOnRails/app/controllers/results_controller.rb
```diff
@@ -29,49 +29,82 @@ def rankings
     @show = splitted_show_param[0].to_i
     @is_persons = splitted_show_param[1] == "persons"
     @is_results = splitted_show_param[1] == "results"
-    limit_condition = "LIMIT #{@show}"
+    expanded_limit_condition = "LIMIT #{@show * 2}"
```
FinnIckler, line 32: I guess this guarantees that everyone will be shown (except for weird edge cases where more than 2n people are tied?)

WcaOnRails/app/controllers/results_controller.rb
```diff
@@ -89,42 +122,55 @@ def rankings
...
+              #{@years_condition_result}
+              #{@gender_condition}
+            GROUP BY result.countryId
+          ) record
+          JOIN Results result ON result.#{value} = recordValue AND result.countryId = recordCountryId
+          JOIN Competitions competition on competition.id = competitionId
           #{@gender_condition.present? ? "JOIN Persons persons ON result.personId = persons.id and persons.subId = 1" : ""}
           WHERE 1
             #{@event_condition}
-            #{@years_condition_result}
```
FinnIckler, line 116: is this just formatting or did you really replace `years_condition_result` with` years_condition_competition` ? If so, why?

FinnIckler:
yes, generally a great idea rewriting this query...

WcaOnRails/app/controllers/results_controller.rb
simonkellly: This is just the formatting of the pr, the current code on master uses both in the same way

WcaOnRails/app/controllers/results_controller.rb
simonkellly: This value can likely be a lot smaller (maybe the limit + 25 or somthing), as it's highly unlikely to run into a case where even 25 people are tied for 100th

DanielEgdal: Fewest moves is probably going to be over 25 tied near the 100th spot at one point

simonkellly:
Converting this to a draft as I will probably leave this for a while to work on my other pr, as this probably requires more than just simple changes especially when considering performance or how we might optimise how results work in the relativly near future

gregorbg:
Closing this PR because after abandoning PHP, this is no longer a simple "fix" on top of the old tables.
We should rethink how we store results fundamentally. A dedicated project + proposal will follow.

PR #7608, opened 2023-02-06
Removing Black Background for Seniors on Delegate Page

ando527:
changed the background colour from black to light grey to match similar styling on tables elsewhere on the site. This also required a font colour change. I also increased the readability of the text by removing the bold font weight and replacing it with a slightly larger font.


WcaOnRails/app/assets/stylesheets/delegates.scss
```diff
@@ -1,7 +1,8 @@
 tr.senior-delegate {
-  background-color: black;
-  color: white;
-  font-weight: bold;
+  background-color: #e9e9e9;
+  color: black;
+  //font-weight: bold;
```
FinnIckler, line 4: why did you comment this out instead of removing it?

ando527: Sorry, bad habit 🤣, Ill fix that now

FinnIckler (approved):
Definitely looks better, especially when the board is the senior delegate, Thanks!

dunkOnIT:
Way more readable, thanks a lot @ando527 ! Just one small note - if you could tag the issue in your PR description that would be awesome :D

Issue #6968, opened 2022-06-02
Combined Third Rounds will have the wrong name


Jambrose777:
**Describe the bug**
A combined third round (roundType = g) will have the name 'Third Round' instead of 'Semi Final' that non combined third rounds will have.

**Expected behavior**
These should show up as 'Semi Final' Instead to be consistent.

**Additional context**
There is no such case yet, but is a small easy fix that just needs a migration.

alexanderc360:
#7610

alexanderc360:
Hey Jacob, this should be resolved. I submitted pull request #7610

PR #7610, opened 2023-02-07
switched type g from 'Third Round' to 'Semi Final'

alexanderc360:
If a third round out of four has a cut-off, the round would show up as "Third Round" instead of "Semi Final". This has been fixed to keep consistency with non-cut-off rounds

### DIFF
--- a/WcaOnRails/config/locales/en.yml
+++ b/WcaOnRails/config/locales/en.yml
@@ -133,8 +133,8 @@ en:
       name: "Final"
       cellName: "Final"
     "g":
-      name: "Third round"
-      cellName: "Third round"
+      name: "Semi Final"
+      cellName: "Semi Final"
     "h":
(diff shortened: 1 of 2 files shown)


dunkOnIT:
Closes #6968

FinnIckler:
Jacob mentions that this needs a migration, which I guess is so it's consistent with earlier Competitions.

Jambrose777:
> Jacob mentions that this needs a migration, which I guess is so it's consistent with earlier Competitions.

This should be fine, (migration meaning change in DB structure, so as long as the DB shows semi final in the roundTypes Table, then it should be fine). There weren't any past competitions the last I checked.

FinnIckler:
cool, then I give you the honors to approve

Jambrose777 (approved):

PR #7611, opened 2023-02-08
Add WCIF extensions for people.

### LINKED ISSUE #7492: Add WCIF extensions to Persons
For CubingUSA Nationals scheduling, I'd like to be able to put a WCIF extension on a Person. There are a few person-specific properties I'd like to be able to set, such as:

- which staff team they are on
- preferences for jobs they can do
- any times that they are unavailable for staff jobs

Pending #7491, this can probably be attached to Registration objects.

timreyn:
This is added to the Person object in the WCIF output, and stored as a child of the Registration object in the database.

Fixes #7492.

### DIFF
--- a/WcaOnRails/app/models/competition.rb
+++ b/WcaOnRails/app/models/competition.rb
@@ -1761,6 +1761,7 @@ class Competition < ApplicationRecord
         updated_at: DateTime,
         is_competing: false,
       )
+      WcifExtension.update_wcif_extensions!(registration, wcif_person["extensions"]) if wcif_person["extensions"]
       # NOTE: person doesn't necessarily have corresponding registration (e.g. registratinless organizer/delegate).
       if wcif_person["roles"]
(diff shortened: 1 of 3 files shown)


timreyn:
> Why are you storing it in the registrations object? One cool way to use this extension is to specify preferences of non-competing staff, which won't have a registration. So where would the extension be stored in that case?

See #7601, which proposes adding a Registration object for non-competing staff.

The higher-level reason is that Registration is the object that represents a {person, competition}, which is why I propose adding the WCIF extension there in this PR, and why I propose using (abusing?) it for non-competing staff in #7601. I'm definitely open to alternate design proposals though!

gregorbg:
I must admit that I (foolishly) wrote my last comment _before_ checking the details in #7601. Now with the changes over there and the `is_competing` boolean, your implementation makes sense.

gregorbg (approved):

PR #7617, opened 2023-02-10
Split every localization into its own pack that can be dynamically loaded

FinnIckler:
I recently noticed that we are loading an 1.6MB js file that contains all the localizations on each page. 1.6MB is huge in web terms and while it will be cached  by the browser on subsequent loads, this change will still reduce a good amount of traffic.
This brings the average translation pack to less than 50kb.

### DIFF
--- a/WcaOnRails/.eslintrc.json
+++ b/WcaOnRails/.eslintrc.json
@@ -13,7 +13,7 @@
         "SharedArrayBuffer": "readonly",
         "_": true
     },
-    "parser": "babel-eslint",
+    "parser": "@babel/eslint-parser",
     "parserOptions": {
         "ecmaFeatures": {
             "jsx": true
--- a/WcaOnRails/.eslintrc.json
+++ b/WcaOnRails/.eslintrc.json
@@ -13,6 +13,7 @@
         "SharedArrayBuffer": "readonly",
         "_": true
     },
+    "parser": "babel-eslint",
     "parserOptions": {
         "ecmaFeatures": {
             "jsx": true
--- a/WcaOnRails/.eslintrc.json
+++ b/WcaOnRails/.eslintrc.json
@@ -18,7 +18,8 @@
             "jsx": true
         },
         "ecmaVersion": 2018,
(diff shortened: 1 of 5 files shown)


FinnIckler:
Not sure how to fix the errors. This runs fine locally and produces the correct effect..

FinnIckler:
I had to update our eslint parser to @babel/eslint-parser as the default one we were using didn't support dynamic imports

FinnIckler:
seems like there is still a difference in how we are loading translations in tests as the tests can't find the rails_translations folder (which is usually generated before running the website). Will look into this further tomorrow

FinnIckler:
ok. so we either need to 
1) not load translations at all for test (have an if in the load translation method or something) or
2) to the yaml -> json conversion for translations also for tests.

FinnIckler:
I added a .catch for the test environment as a possible solution

gregorbg:
I played around with this a bunch and I've come to realize that it's _not_ an inherent test environment problem. CI tests were totally fine using the exported JSON translations before this PR, and they continue to be fine after this PR.

The actual problem is a racing condition: Your new approach is inherently async. The translation data is added *asynchronously* after the import finished. This means that in theory it's possible for the page to load, the import to trigger, but then **before** the import completes, some frontend part is already requesting translations which are not (yet) present.

I don't have a smart solution for this in mind.

dunkOnIT:
@FinnIckler is this PR still relevant? Or would the implementation of translations change based on the React migration?

FinnIckler:
I would still like to have a solution to this, but we probably will work on this part of a bigger rewrite on how we bundle js

dunkOnIT:
Can I suggest we convert it to a draft then? Just to make it easier to see what the "active" PR's are

FinnIckler:
we can also just close it, instead you can make an issue about "improve javascript bundling" which relies on removing sprockets and link this PR as one of the things we want to implement (along with other things like fixing the way we include semantic)

dunkOnIT:
Closed in favour of #8346

PR #7619, opened 2023-02-11
API route to return countries

coder13:
I want to do some calculations involving people's continental records and don't want to import the database.

Added:
```
GET /api/v0/countries

[
{
  "id": "Afghanistan",
  "name": "Afghanistan",
  "continentId": "_Asia",
  "iso2": "AF"
},
{
  "id": "Albania",
  "name": "Albania",
  "continentId": "_Europe",
  "iso2": "AL"
},  
...
]

```

### DIFF
--- a/WcaOnRails/app/controllers/api/v0/api_controller.rb
+++ b/WcaOnRails/app/controllers/api/v0/api_controller.rb
@@ -209,4 +209,8 @@ class Api::V0::ApiController < ApplicationController
 
     @current_api_user = User.find_by_id(doorkeeper_token&.resource_owner_id)
   end
+
+  def countries
+    render json: Country.all
+  end
 end
--- a/WcaOnRails/config/routes.rb
+++ b/WcaOnRails/config/routes.rb
@@ -259,6 +259,7 @@ Rails.application.routes.draw do
       get '/persons/:wca_id/results' => "persons#results", as: :person_results
       get '/persons/:wca_id/competitions' => "persons#competitions", as: :person_competitions
       get '/geocoding/search' => 'geocoding#get_location_from_query', as: :geocoding_search
+      get '/countries' => 'api#countries'
       resources :competitions, only: [:index, :show] do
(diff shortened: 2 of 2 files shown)


FinnIckler:
looks self explanatory. I wonder if we should start putting new api routes on /v1/ ? 
Maybe we have missed that point already though

simonkellly:
I'm not hugely familier with REST versioning, but is v0 supposed to be like an unfinalised version, then with v{n>0} you try to avoid breaking changes as much as possible?

FinnIckler:
Yes, that is true. But we already have a good amount of tools built upon our v0 APIs which why it probably makes sense to start finalizing it?

coder13:
I would not move forward with v1 routes until we do a couple of meetings to layout all of the information we want to lay out.

I have some thoughts and suggestions that revolve around creating a separate project specifically for serving this data via graphql.

Graphql is very very powerful and would be the ideal tool to use in this case given the way that our data works.

FinnIckler:
You are right. We will tackle this when the architecture is ready to have multiple services running.

FinnIckler (approved):
LGTM

gregorbg:
This PR is perfectly fine, but please note for the future that I'm not too wild about "I don't want to import the database" quick fixes. Keep a balance between individual devs' convenience and the greater WST/WCA vision for our software development.

dunkOnIT:
@gregorbg could you elaborate on how this might not fit into our vision for software development? Is it not a priority to be adding API endpoints / not keen on adding more endpoints / cautious about adding endpoints without an overall architecture for how we want the API to function / something else?

gregorbg:
It is not a priority to add API endpoints that have not been planned as part of a team effort. Exposing data to the public almost guarantees that at some point, someone will find some inconsistency in the data, and then open a GitHub issue, or complain via email, etc... And we certainly don't have the capacity nor the momentum to handle this kind of support.

PR #7622, opened 2023-02-12
Fix trailing space before period on About page

JonEsparaz:
Blink and you'll miss it 👀 

Before:

image

After:

image

I used https://prettier.io/playground/ to safely format these few lines.


FinnIckler (approved):
LGTM

PR #7625, opened 2023-02-13
Expose results in a bulk API route

coder13:
I may want to do bulk analysis on results from a number of competitions without importing the wca database. This API endpoint *should not* be too burdensome on the server.

### DIFF
--- a/WcaOnRails/app/controllers/api/v0/api_controller.rb
+++ b/WcaOnRails/app/controllers/api/v0/api_controller.rb
@@ -213,4 +213,45 @@ class Api::V0::ApiController < ApplicationController
   def countries
     render json: Country.all
   end
+
+  def results
+    unless params[:competition_ids] || (params[:start_date] && params[:end_date])
+      render status: :bad_request, json: { error: "No competition_ids or start/end dates specified" }
+      return
+    end
+
+    if params[:start_date] && params[:end_date]
+      
+      competitions = Competition.where(["start_date >= ? AND start_date <= ?", params[:start_date], params[:end_date]])
+      competition_ids = competitions.pluck(:id)        
+      puts competition_ids
+    end
+
+    if params[:competition_ids]
+      # We don't want to return results for competitions that are not visible to the public.
+      competition_ids = params[:competition_ids].split(",")
+    end
+
+    if competition_ids.empty?
+      render status: :not_found, json: { results: [] }
+      return
+    end
+
+    if competition_ids.length > 50 && !(params[:event_id] || params[:personId])
+      render status: :bad_request, json: { error: "Too many results requested" }
+      return
+    end    
+    
+    results = Result.where(competitionId: competition_ids)
+
+    if params[:event_id]
+      results = results.where(eventId: params[:event_id])
+    end
+
+    if params[:personId]
+      results = results.where(personId: params[:personId])
+    end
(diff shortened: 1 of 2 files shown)


FinnIckler:
Two ideas come to mind here:
- Should we not design this API Route so it can be used for a future rework of the Competition page
- If that is too far in the future should we add authentication to this API?

coder13:
> Should we not design this API Route so it can be used for a future rework of the Competition page

A fast efficient competition page would pull the wcif. This would immediately provide the bulk of the data needed for a snappy single page app experience. This endpoint already loads very quickly as the data is cached. There also already exists `GET /api/v0/competitions/{competitionId}/results` which returns similar data but for a single competition.

This api was designed to do a bulk fetch.

> If that is too far in the future should we add authentication to this API?
At this moment I don't think that authentication is required.

We could rate limit the endpoint or I could try to apply a layer of caching for it if we'd like.

FinnIckler:
ah yeah I forgot about `/api/v0/competitions/{competitionId}/results` and of course wcif would also work. I guess nothing really is hindering us to already use these right now for the React rework (except still not having finalized the api I guess).

Honestly the best solution would be to hook this up to our read replica which has not been set up for years, it can be between a few minutes and a few hours out of date, but that wouldn't matter for this right?
I don't expect you to set it up in this PR though, as one needs AWS access to test it properly. I'll see how easy it is tomorrow and maybe we can finally start using it.

FinnIckler:
I am done implementing the Read Replica integration and just need to get the environment variable into our bootstrap process. Will test on staging tomorrow and then we should be able to proceed with this.

FinnIckler:
My change works on staging with successful reads from the Read Replica! Would love to have some Rails guru to have a look at it though...

gregorbg:
> This API endpoint _should not_ be too burdensome on the server.

I've heard this argument oh so many times, not counting the occasions where I myself have fallen into this trap.

I am not ready to release public results APIs yet. The maintenance effort is just too high. If we release this, people are bound to found out sooner or later and then the feature requests will come pouring in. We don't have the capacity to do any bugfixing, let alone fulfill those feature requests.

This is just too much of a risk as per "Oh if you set this esoteric combination of query strings it spits out results that have not yet been published" or some crazy shenanigans like that. Please set this PR to Draft state and resort to the [results export](https://www.worldcubeassociation.org/results/misc/export.html) for the time being

coder13:
> > This API endpoint _should not_ be too burdensome on the server.
> 
> I've heard this argument oh so many times, not counting the occasions where I myself have fallen into this trap.
> 
> I am not ready to release public results APIs yet. The maintenance effort is just too high. If we release this, people are bound to found out sooner or later and then the feature requests will come pouring in. We don't have the capacity to do any bugfixing, let alone fulfill those feature requests.
> 
> This is just too much of a risk as per "Oh if you set this esoteric combination of query strings it spits out results that have not yet been published" or some crazy shenanigans like that. Please set this PR to Draft state and resort to the [results export](https://www.worldcubeassociation.org/results/misc/export.html) for the time being

Thanks for your wisdom Gregor.

sanmaydotsr:
How to get api of a particular compitation, pls answer

sanmaydotsr:
Sorry not api but ids

PR #7627, opened 2023-02-14
Introduce the Read replica for ActiveRecord Models

FinnIckler:
This allows Active Record Models to run queries on the read replica instead of the primary Database. 
I added the /countries API endpoint as an example of its usage. We should start slowly introducing more parts in the code where the replica is used.


FinnIckler:
testing on staging...

FinnIckler:
Successfully deployed it to staging
https://staging.worldcubeassociation.org/api/v0/countries works correctly

FinnIckler:
After discussing with Gregor we will also be retiring our old self implemented cachable concern to just read from the Read Replica by default

FinnIckler:
remaking this pr

PR #7525, opened 2023-01-04
Add a reminder on the sign up page to give the details of the competitor

Nevseros:
Closes #7488.

The title is descriptive itself.

PR #7629, opened 2023-02-15
Make new competitor welcome key HTML safe

JonEsparaz:
On https://www.worldcubeassociation.org/users/sign_up:

image

---

Fixed by changing `wca.devise.welcome` to `wca.devise.welcome_html`:

image

^ HTML is parsed now, but looks a little odd in my opinion - not sure if we should add two line breaks or have no line break at all.

---

HTML was added in #7525.


FinnIckler:
yeah, I think two or 0 line breaks are the way to go

JonEsparaz:
Added two line breaks:

image

image

FinnIckler (approved):
LGTM

PR #7634, opened 2023-02-17
Migrate Analytics to GA4

gregorbg:
As per https://support.google.com/analytics/answer/9304153?sjid=4362531488141620959-EU#add-tag&zippy=%2Cadd-the-google-tag-directly-to-your-web-pages


gregorbg:
Should ideally be merged at the beginning of the month to avoid skewing numbers between the old and new property.

PR #7636, opened 2023-02-18
Retry some of the frequently failing tests

simonkellly:
This adds retrying the test to some of our most frequently failing tests. This has been already done in similar parts of the codebase and might (not certain about this as its very hard to reproduce the failures locally) relieve some of the test failures we see in the short term.

This spec will likely end up being changed around a lot once the competition form is finished getting migrated which may lead to a more proper fix.


FinnIckler:
I hope the tests not failing on this PR is foreshadowing :D

gregorbg (approved):
Very elegant way to spare developer nerves :smile: 
Ideally, we should find out _why_ these tests fail but for the time being, your idea works.

PR #7641, opened 2023-02-19
Add Read Replica for Active Record Models

FinnIckler:
This PR introduces the Read Replica to use for our ActiveRecord Models. 
There are two ways of using it:
- Extending the `ReadonlyRecord` class makes all reads go through the read replica
- Using `ActiveRecord::Base.connected_to(role: :read_replica) do` to use the read replica for the whole block

There is some limitations of what model can be switched to the ReadonlyRecord, because of cross database joins not working. This is why the tests were failing before. This does not affect `ActiveRecord::Base.connected_to(role: :read_replica) do` because it will run the whole query from the replica DB.

I also upgraded the read replica to a t3.medium to accommodate our increased use


gregorbg (changes requested):
More nitpick: Please add the read replica in the phpMyAdmin config (Chef line 341, GH doesn't let me comment in-line because it's not part of the diff...) so that we can access the slave server manually should it ever become necessary (I hope not...)

PR #7642, opened 2023-02-20
[PHP WRT #4/5] Migrate public results export to Rails

gregorbg:
Basically (ab)using a custom-designed table sanitizer to replicate the developer export, but the columns replicate the existing results export instead.

Remarks:
- PHP code is completely removed :tada:
- Developer dump gets a tiny public page as well


FinnIckler:
nice! Now that we are changing it, would it make sense to upload the export to S3 to save on some traffic to the website?

gregorbg:
Essentially yes, but this is "only" the result export which weighs ~150MB. Not in the scope of this PR to establish and move to S3.

FinnIckler:
ah ok. The other export is already done using rails?

gregorbg:
> The other export is already done using rails?

Yes indeedy! As written in my initial description, this port basically abuses the existing export infrastructure for the "big" developer dump.

WcaOnRails/app/views/database/_public_results_readme.md.erb
```diff
@@ -1,7 +1,9 @@
+<% long_date ||= DateTime.now %>
```
FinnIckler, line 1: I read that you should Time.current instead of (Date)Time.now ([here](https://discuss.rubyonrails.org/t/time-now-vs-time-current-vs-datetime-now/75183))

WcaOnRails/lib/database_dumper.rb
```diff
@@ -1,10 +1,21 @@
 # frozen_string_literal: true
 
 module DatabaseDumper
-  JOIN_WHERE_VISIBLE_COMP = "JOIN Competitions ON Competitions.id=competition_id WHERE showAtAll=1"
-  DUMP_TIMESTAMP_NAME = "developer_dump_exported_at"
-  VISIBLE_ACTIVITY_IDS = "SELECT A.id FROM schedule_activities AS A JOIN venue_rooms ON (venue_rooms.id = holder_id AND holder_type='VenueRoom') JOIN competition_venues ON competition_venues.id = competition_venue_id #{JOIN_WHERE_VISIBLE_COMP}"
-                         .freeze
+  WHERE_VISIBLE_COMP = "WHERE Competitions.showAtAll=1"
```
FinnIckler, line 4: why are we freezing only some of the strings here? They are all constants right?

FinnIckler (commented):
Just two minor things

danieljames-dj:
Hi @gregorbg, happy to know that the PHP migration project is started 😍
Couple of queries:
1. Is there any software limitation because of which we won't be able to trigger the export manually? From WRT perspective, it's a required feature because when a user requests for anonymizing their profile, one of the steps that WRT is following is to remove it from public export and we are doing it by creating a new public export and hence  the data of the user won't be available in the new export.
2. I'm assuming we will be having many more PRs coming related to PHP migration, maybe can we have a common prefix in the title (something like [PHP Migration]) which might be useful for us in searching if needed?

gregorbg:
> 1. Is there any software limitation because of which we won't be able to trigger the export manually?

No. I have added a panel to manually trigger the dev export and public export :D

WcaOnRails/lib/database_dumper.rb
gregorbg: Because constants that only contain string values, like
```ruby
  MY_CONST = "some_value"
```
are automatically frozen. Only the constants that have interpolation, like
```ruby
  MY_CONST = "some_value"
  MY_OTHER_CONST = "#{MY_CONST} yet again"
```
need manual freezing.

WcaOnRails/app/views/database/_public_results_readme.md.erb
gregorbg: I'm more confused than convinced by this article. We use a wild mix of `Time.now` and `DateTime.now` in our codebase, and both seem to work fine. In this case, the "worst" thing that could happen is that the timezone of some timestamp in the results export is off. Absolutely not critical and not worth investing more time into IMHO

WcaOnRails/lib/database_dumper.rb
FinnIckler: Cool, Thanks!

WcaOnRails/app/views/database/_public_results_readme.md.erb
FinnIckler: I got no hard feelings one way or the other, I did use `DateTime.current` in some of my code though if that is ok

WcaOnRails/app/controllers/admin_controller.rb
```diff
@@ -157,14 +157,26 @@ def person_data
...
   def compute_auxiliary_data
-    @reason_not_to_run = ComputeAuxiliaryData.reason_not_to_run
   end
 
   def do_compute_auxiliary_data
     ComputeAuxiliaryData.perform_later unless ComputeAuxiliaryData.in_progress?
     redirect_to admin_compute_auxiliary_data_path
   end
 
+  def generate_exports
```
FinnIckler, line 167: Why is there an empty method here?

gregorbg: Because Rails mandates that for every route in `config/routes.rb`,  there must be a corresponding method in the corresponding controller. Basically, this method loads a page that is so plain and basic that it doesn't need any computations.

WcaOnRails/app/controllers/database_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+  RESULTS_README = "README.md"
+  RESULTS_METADATA = "metadata.json"
+
+  RESULTS_SQL_PERMALINK = "#{RESULTS_SQL}.zip".freeze
+  RESULTS_TSV_PERMALINK = "WCA_export.tsv.zip"
+
+  DEVELOPER_SQL = "wca-developer-database-dump.sql"
+  DEVELOPER_SQL_PERMALINK = "wca-developer-database-dump.zip"
+
+  RESULTS_README_TEMPLATE = 'database/public_results_readme'
```
danieljames-dj, line 19: This line is in single quotes while the lines before it in double quotes. Was it intentional?

WcaOnRails/app/controllers/database_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+  EXPORT_PUBLIC_FOLDER = Rails.root.join('public', 'export')
+
+  RESULTS_EXPORT_FOLDER = EXPORT_PUBLIC_FOLDER.join('results')
+  DEVELOPER_EXPORT_FOLDER = EXPORT_PUBLIC_FOLDER.join('developer')
+
+  RESULTS_SQL = "WCA_export.sql"
+  RESULTS_README = "README.md"
+  RESULTS_METADATA = "metadata.json"
+
+  RESULTS_SQL_PERMALINK = "#{RESULTS_SQL}.zip".freeze
```
danieljames-dj, line 13: Just searched what `freeze` is, and it is used to ensure object cannot be modified. I'm bit curious why it's applicable for this file alone? 😅

WcaOnRails/app/controllers/database_controller.rb
danieljames-dj, line 19: How about ordering it separately as follows?

```
EXPORT_PUBLIC_FOLDER = Rails.root.join('public', 'export')

RESULTS_EXPORT_FOLDER = EXPORT_PUBLIC_FOLDER.join('results')
RESULTS_EXPORT_FILENAME = "WCA_export"
RESULTS_EXPORT_SQL = "#{RESULTS_EXPORT_FILENAME}.sql"
RESULTS_EXPORT_README = "README.md"
RESULTS_EXPORT_METADATA = "metadata.json"
RESULTS_EXPORT_SQL_PERMALINK = "#{RESULTS_EXPORT_SQL}.zip".freeze
RESULTS_EXPORT_TSV_PERMALINK = "#{RESULTS_EXPORT_FILENAME}.tsv.zip"

DEVELOPER_EXPORT_FOLDER = EXPORT_PUBLIC_FOLDER.join('developer')
DEVELOPER_EXPORT_FILENAME = "wca-developer-database-dump"
DEVELOPER_EXPORT_SQL = "#{DEVELOPER_EXPORT_FILENAME}.sql"
DEVELOPER_EXPORT_SQL_PERMALINK = "#{DEVELOPER_EXPORT_FILENAME}.zip"
```

WcaOnRails/app/views/admin/_delayed_job_runner.html.erb
```diff
@@ -0,0 +1,28 @@
...
+<h2><%= "Controls for #{job.name}" %></h2>
+
+<%= yield %>
+
+<% if job.respond_to?(:reason_not_to_run) && job.reason_not_to_run %>
+  <%= alert :warning, job.reason_not_to_run %>
+<% else %>
+  <% if job.in_progress? %>
+    <%= alert :info, "The job is running. Thanks for checking =)" %>
+  <% elsif job.is_a? TimedApplicationJob %>
```
danieljames-dj, line 13: Any reason why only CAD is TimedApplicationJob and exports (dev & results) are not TimedApplicationJob?

WcaOnRails/lib/database_dumper.rb
danieljames-dj, line 4: Nit: `showAtAll = 1` (to match with other queries in the file)

WcaOnRails/app/jobs/dump_developer_database.rb
```diff
@@ -3,12 +3,17 @@
...
-    if last_developer_db_dump.not_after?(3.days.ago)
-      Rake::Task["db:dump:development"].invoke
-      last_developer_db_dump.touch :date
+    if force_export || last_developer_db_dump.not_after?(3.days.ago)
+      running_on_dev_dump = Timestamp.exists?(name: DatabaseDumper::DEV_TIMESTAMP_NAME)
+
+      unless running_on_dev_dump
+        Rake::Task["db:dump:development"].invoke
+        last_developer_db_dump.touch :date
+        Rake::Task["db:dump:public_results"].reenable
```
danieljames-dj, line 15: reenable is called for public_results in this file & `dump_public_results_database.rb`. But invoke was to the file it belongs to in both. I don't know the logic behind it, if it's intentional, please mark this as resolved.

danieljames-dj (approved):
I was able to understand around 70% of the code. Maybe it will be nice if some senior too do a quick review.

WcaOnRails/app/controllers/database_controller.rb
gregorbg: Nope, Ruby doesn't enforce anything regarding quotes. I wonder why RuboCop didn't complain though...

WcaOnRails/app/controllers/database_controller.rb
gregorbg: It is applicable for strings in any file based on the `# frozen_string_literal: true` line at the top of the file.
This string that you marked is an exception because it contains interpolation (basically another string inserted into the string with `#{}`" and the `frozen_string_literal` doesn't apply because then it's not a **literal** anymore.
So I have to explicitly freeze it while all other "straight-forward-strings" in the file are implicitly frozen.

WcaOnRails/app/views/admin/_delayed_job_runner.html.erb
gregorbg: Because until now we didn't need it. `Timed` in this case does not mean that the Job is timed on a schedule, but rather that we're measuring the time how long the job takes. And I'm not sure if we need that for these exports. What's your opinion?

WcaOnRails/app/jobs/dump_developer_database.rb
gregorbg: Whoops, indeed a copy/paste mistake :see_no_evil: Thanks for catching!

WcaOnRails/app/views/admin/_delayed_job_runner.html.erb
danieljames-dj: It was not there earlier as well, so as per our PHP migration project, definitely it's not needed.

But if it's just changing the class name or too less effort (maybe 3-4 lines of codes max), and if it doesn't add any extra cost, it will be nice to have I guess.

Currently, all WRT members knows that the CAD takes around 4-5 minutes to run. So, if at some point we see it is taking more time like 8 minutes, we will know something is wrong. Similarly, if the time of these two exports are available, eventually WRT members will know the time taken for those scripts and maybe inform WST if the time is increasing consistently, or if the time has became unusually large.

viroulep (commented):
Here are some comments :)

WcaOnRails/app/jobs/dump_developer_database.rb
```diff
@@ -3,12 +3,18 @@
 class DumpDeveloperDatabase < SingletonApplicationJob
   queue_as :default
 
-  def perform
+  def perform(force_export: false)
     # Create developer database dump every 3 days.
     last_developer_db_dump = Timestamp.find_or_create_by(name: 'developer_db_dump')
-    if last_developer_db_dump.not_after?(3.days.ago)
-      Rake::Task["db:dump:development"].invoke
```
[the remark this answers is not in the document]
gregorbg: I came to the same conclusion that Rake tasks probably shouldn't be invoked like this when I realised that you have to intentionally `reenable` them when calling them twice from the same process. I guess it only worked to build the dev dump through this routine on a three day cron schedule like this because of some supervisor shenanigans that spin up new processes(??)
Since the server is running in the same PID, you have to hijack Rake. With that being said, I'm not wild about refactoring the entire way we generate exports when the focus right now is to ditch PHP

PR #7649, opened 2023-02-22
Improve signup page "previously competed" selection

### LINKED ISSUE #421: Sign up page polish
Some suggestions from @StachuDotNet:

- The visual separation of "I have competed in a comp" and "I haven't..." is bad right now
I kinda had to guess that that was a thing to click
image
  - needs a border or background or something
- Also maybe explain /why/ I should sign up on the sign up page

havijw:
Fixes #421.


dunkOnIT:
Thanks a lot! Comment from one of the devs re the tests below: 

> I cloned their fork and ran the tests and a bunch of them in the spec he mentioned failed. Looked like it was just the tests hadn't been updated to use the new UI

I do see that no tests have been updated in the PR, but the conversation is outside my expertise so I'll tag in @simonkellly for any other clarity/input needed :D

simonkellly:
Just to clarify, I think when I did clone the repo (which was about 3 days ago) it was before you had pushed your current changes to the spec file, and the last time it had been moified, so while that test with the DOB did fail, so did many othes at that point.

This is no longer the case as you have fixed those test cases so what I said is no longer relavent anymore I think.

havijw:
Thanks, that's what I figured. I don't know what caused the check to fail - I think I had the same thing fail earlier today but thought I'd fixed the issue. Everything passes when I run it locally, so I'm not sure how to replicate the problem.

simonkellly:
For the DOB thing, it seems to be caused by this https://github.com/thewca/worldcubeassociation.org/blob/8f498d4f319f555fea8a294f8016761bae55b443/WcaOnRails/app/models/user.rb#L338

where it assigns the Person's DOB to the user account even if they only entered a WCA ID. This should probably be changed to only do so if they correctly validated the DOB

havijw:
> This should probably be changed to only do so if they correctly validated the DOB

This makes sense as an explanation to me. Should I try to fix this?

What I've done improves on what currently exists (clearing the fields at least makes it less obvious that the data is there), but I can look into it if it should be addressed before merging.

dunkOnIT:
Hey @havijw! Looks like we've managed to open a can of worms with this one :D 

Is this an accurate summary of things as they stand? 
- When trying to claim a WCA ID, the details of the holder of the WCA ID are revealed to whoever is trying to claim it - this poses a risk that (a) someone's data can be extracted from the system, and/or (b) a bad actor could claim WCA ID's which don't belong to them. 
- Your changes haven't introduced this behaviour, but you have come across it while investigating failed tests (arguably, the tests shouldn't have been passing in the first place?)

If that's accurate, sounds like something we should look to fix. Would you be either record a brief video demonstrating the concerning behaviour, or jump on a call to demonstrate/discuss it? Whichever is easiest for you.

dunkOnIT:
Ok with Gregor's PR merged, @havijw could you remove the JS fixes you'd made to obscure sensitive data, and re-run test suite? 

Once that's done we should be good to review and merge. What a rollercoaster! :D

havijw:
Just made the changes! I can't run the test suite here, but everything passed for me locally.

Glad to be focused on the changes this issue was actually about!

dunkOnIT (approved):
Tests passing, changes look reasonable, [deployed to staging](http://localhost:3000/users/sign_up) and it works there

dunkOnIT (approved):
Tests passing after rerun, changes look reasonable, [deployed to staging](http://localhost:3000/users/sign_up) and it works there

dunkOnIT:
@gregorbg happy if I merge and deploy?

dunkOnIT:
@gregorbg just pinging in case this got buried in notifs - let me know if there's cause for caution here

PR #7651, opened 2023-02-25
Add Conditional GET Support

FinnIckler:
Conditional GETs are a feature of the HTTP specification that provide a way for web servers to tell browsers that the response to a GET request hasn't changed since the last request and can be safely pulled from the browser cache. 
Rails supports these out of the box, read about it [here](https://guides.rubyonrails.org/caching_with_rails.html#conditional-get-support).

This PR adds Conditional GET checks for Persons and Competitions, which is going to save traffic for instances where the user revisits these pages without them having changed. The user and the Locale is included in the etag so wrong content is not served when a user changes permissions, logs into a different account or changes locale.

We will get even more benefits from this when we get I18n out of Ruby


FinnIckler:
superseeded by #8100 and some other stuff

PR #7705, opened 2023-03-01
Update Ruby to version 3.2.1

gregorbg:


gregorbg:
Standard dependency update, proceeding like it was sent by Dependabot.

PR #7708, opened 2023-03-01
Update Shakapacker to version 6.5.6

gregorbg:
Supersedes https://github.com/thewca/worldcubeassociation.org/pull/7692


gregorbg:
Treating as a Dependabot update.

PR #7716, opened 2023-03-02
Display estimated dues per competitor when no competitor limit is specified

### LINKED ISSUE #7648: WCA Dues estimate for competitions without competitor limit
**Describe the bug**
Create a competition without a competitor limit (e.g., a multi location FMC competition), then you'll see something like: `The estimated WCA Dues for [missing "%{limit}" value] competitors are $0.00 (CAD)`

**Expected behavior**
We should probably display something like "WCA Dues cannot be estimated without a competitor limit".

**Screenshots**

image

**Additional context**
N/A

dunkOnIT:
Fixes #7648.

A few notes: 
- This issue was checking the truthyness of a string which we expected to be a bool, in 2 different areas of the code.
- Both of these arose from passing variables between Ruby code and JS code. Do we have access to ways of specifying expected types, through type hinting or otherwise? 
- We may also want to reconsider the use of shorthands like `if (variable) {}`, expecting that `variable` is a bool - this is an easy way to have something always go down the "true" path of an if. 
- I'm moving around the code like a ballerina dude, this is so sick

Just going to use this PR to store notes as I figure out what's happening


dunkOnIT:
I swear bro I started this thing as a draft

dunkOnIT:
#7648 defines the expected behaviour as saying we can't estimate the dues. Per the existing code, I believe the intention is to provide an estimated dues amount per competitor instead of for the overall competition, which makes a lot of sense. 

`config/locales/en.yml` currently has the following strings defined:

>       dues_estimate:
>         calculated: "The estimated WCA Dues for %{limit} competitors are %{estimate}"
>         per_competitor: "The estimated WCA Dues per competitor are %{estimate}"
>         ajax_error: "There is an error with Dues estimation, possibly because there are no available exchange rates for the selected currency."

We should be getting the per_competitor string, but instead" we're getting the `calculated` string. Why?

dunkOnIT:
From `views/competitions/_competition_form.html.erb`:

```
                if (compLimitEnabled) {
                  <% # i18n-tasks-use t('competitions.competition_form.dues_estimate.calculated') %>
                  $('#dues_estimate').html("<b>" + I18n.t('competitions.competition_form.dues_estimate.calculated', {limit: compLimit, estimate: data.dues_value}) + " (" + currencyCode + ")</b>");
                }
                else{
                  <% # i18n-tasks-use t('competitions.competition_form.dues_estimate.per_competitor') %>
                  $('#dues_estimate').html("<b>" + I18n.t('competitions.competition_form.dues_estimate.per_competitor', {estimate: data.dues_value}) + " (" + currencyCode + ")</b>");
                }
              },

```

This would suggest that compLimitEnabled is coming back `true`, when `false` is expected. Can I print that to the console to confirm?

dunkOnIT:
compLimitEnabled was a string. I'm coercing it to be a bool when it was assigned, as that is the intuitive variable type. `compLimit` was using a `=== "true"` comparison to account for the fact that it was a string, but as we've seen, accounting for this on a per-usage basis causes errors, which is why I'd rather fix it at the source. (And have updated the `compLimit` assignment accordingly).

Dues per competitor are still being estimated as $0.00 - investigating.

WcaOnRails/app/controllers/competitions_controller.rb
```diff
@@ -410,7 +410,10 @@ def calculate_dues
     # times 100 because later currency conversions require lowest currency subunit, which is cents for USD
     price_per_competitor_us_cents = [registration_fee_dues_us_dollars, country_band_dues_us_dollars].compact.max * 100
 
-    if params[:competitor_limit_enabled]
+    # This is quite a hacky fix to getting a boolean as a string from the view, however I'm not sure how to/if we can
+    # specify or coerce types for params we receive other option would be to manually override the hash value, or define
+    # a variable based on the param at the start of the function
+    if params[:competitor_limit_enabled] == true || params[:competitor_limit_enabled] == "true"
```
gregorbg, line 416: We have a hilariously complicated way to detect this, which seems like a joke but is actually used all throughout our codebase:
```ruby
if ActiveRecord::Type::Boolean.new.cast(params[:competitor_limit_enabled])
  ...do something
end
```

WcaOnRails/app/views/competitions/_competition_form.html.erb
```diff
@@ -252,8 +252,8 @@
           function UpdateDues(){
 
             var obj = $("#competition-form").serializeJSON();
-            var compLimitEnabled = obj['competition[competitor_limit_enabled]']
-            var compLimit = compLimitEnabled === "true" ? obj['competition[competitor_limit]'] : null
+            var compLimitEnabled = ("true" === obj['competition[competitor_limit_enabled]']) // Coercing this variable to be a boolean, as one expects.
```
gregorbg, line 255: Comment seems a bit unnecessary. Also, why is this specific change needed?

gregorbg (changes requested):

WcaOnRails/app/views/competitions/_competition_form.html.erb
dunkOnIT: In general I'm a huge fan of over-commenting, especially on a shared codebase. I can edit it to make it more clear if needed - but essentially `obj['competition[competitor_limit_enabled]']` returns a string value of `"true"`, instead of a bool `true` as one would intuitively expect from a variable named compLimitEnabled.

The issue is if someone writes `if (compLimitEnabled) {...`, it will always return true. Thus, I want to force compLimitEnabled to be a boolean when it is assigned, so that it behaves as expected throughout the code if anyone makes changes to this file later on.

gregorbg (approved):
JS change isn't _strictly_ necessary, but I think it should fix the problem by casting the `"false"` sstring to a proper boolean on the Rails side of things.

dunkOnIT:
Wait why isn't the JS fix necessary? If it isn't put in place, then the `if` statement on line 272 always returns `true` and `competitions.competition_form.dues_estimate.per_competitor` never gets executed, which was the root cause of the issue.

The Rails fix just stops an error in the calculation of the dues amount once the per_competitor estimate is called.

gregorbg:
I am still reasonably confident that the _Ruby_ part is what makes the difference here, despite your explanation. Look at the diff:

# Before
```js
var compLimitEnabled = obj['competition[competitor_limit_enabled]']
var compLimit = compLimitEnabled === "true" ? obj['competition[competitor_limit]'] : null
```

The first line, `compLimitEnabled` is conceptually supposed to be a boolean, but factually encoded as a string. So in the next line, it is compared with the string value `=== "true"` to convert it to a "proper" boolean (the result of the `===` comparison) and the proper boolean is used to evaluate a ternary expression.

# After
```js
// Coerce compLimitEnabled to a bool, as it is a string obj['competition[competitor_limit_enabled]'] returns a string
var compLimitEnabled = ("true" === obj['competition[competitor_limit_enabled]'])
var compLimit = compLimitEnabled ? obj['competition[competitor_limit]'] : null
```

The first line (barring the comment) reads the _exact same_ index from the `obj` hash as the old version, and then immediately converts it to a boolean via the `exact same` comparison as the old version. That "proper" boolean is then used in the ternary expression.

In conclusion, I think you have done nothing more than moving the `=== "true"` comparison (which was already there before as per the GitHub diff view!!) to the assignment rather than doing it "on the spot" in the ternary check. But my observation is that the overall JS code hasn't changed a single bit. (Please do correct me if I'm overlooking something super duper obvious here!)

PR #7717, opened 2023-03-03
Move rubocop config to the rails folder

simonkellly:
This should allow some IDEs to detect it when only opening the rails portions of the codebase without having to manually copy over the files

### DIFF
--- a/.overcommit.yml
+++ b/.overcommit.yml
@@ -53,6 +53,7 @@ PreCommit:
 
   RuboCop:
     enabled: true
+    command: ['bundle', 'exec', 'rubocop', '--config', './WcaOnRails/.rubocop.yml']
     on_warn: fail # Treat all warnings as failures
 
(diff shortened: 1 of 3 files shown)


gregorbg:
Don't over-complicate the overcommit workflow by introducing an additional command.
Instead, move `overcommit.yml` and all dependent files to `WcaOnRails`. In particular, this will allow you to get rid of an [ugly hack](https://github.com/thewca/worldcubeassociation.org/blob/master/.github/workflows/ruby.yml#L50-L51) in our GitHub CI workflow

simonkellly:
It looks like overcommit references .overcommit.yml directly when it installs the hooks so it may not be so simple to just move the files

gregorbg:
I see, probably the `overcommit` gem makes some implicit assumptions about the config file always living in the git root.

gregorbg (approved):

PR #7718, opened 2023-03-03
Update anonymize_person.html.erb

ohexter:
Change to align website instructions with current WRT anonymization process.


Jambrose777 (approved):
LGTM

PR #7726, opened 2023-03-09
[PHP WRT #2/5] Create Newcomers

gregorbg:
Now includes the ability to filter for a single competition, which brings us a big step closer to parallel posting. Now we only need to make sure that we avoid scenarios like "WRT member 1 clicks check button" --> "WRT member 2 clicks check button" --> "member 1 submits" --> "member 2 submits". (Is this even a problem though? The second submission would just fire a lot of "empty" `UPDATE` queries because the `WHERE` doesn't match anything...)

A word about the 20 person limit: It's gone in principle. BUT I now understand why it was there in the first place: The name similarity comparison is _sinfully_ expensive. It takes ~7 seconds per newcomer. There is no caching because every newcomer is a new name that we haven't seen before. So we have to compare him to all existing persons aaaaaall over again. What should we do about this?

Also, the forms code is ugly as hell. I'm hacking together a nested object by using keys like `foo][bar` which, when inserted, "accidentally" yields `object[   foo][bar   ]` aka `object[foo][bar]`. Rails has helpers for nested forms, but they make implicit assumptions about the underlying nested models being `ActiveRecord` objects. But we don't have a database binding so we can "only" use the weaker `ActiveModel`, which doesn't work with the nested helpers (as far as I'm concerned). I am open for suggestions.


WcaOnRails/lib/finish_unfinished_persons.rb
```diff
@@ -0,0 +1,178 @@
...
+  end
+
+  def self.compute_similar_persons(result, n = 5)
+    res_roman_name = self.extract_roman_name(result.person_name)
+
+    only_probas = []
+    persons_with_probas = []
+
+    # pre-cache probabilities, so that we avoid doing string computations on _every_ comparison
+    Person.all.each do |p|
```
viroulep, line 109: > A word about the 20 person limit: It's gone in principle. BUT I now understand why it was there in the first place: The name similarity comparison is sinfully expensive. It takes ~7 seconds per newcomer. There is no caching because every newcomer is a new name that we haven't seen before. So we have to compare him to all existing persons aaaaaall over again. What should we do about this?

Answering this here to have a thread somehow linked to the place where the "issue" is.
If I'm reading this correctly we do "for each newcomer iterate over all persons", since the number of persons is much much higher than the number of newcomers, why don't you do it the other way around: "for each person iterate over all newcomers"?

I know the number of iterations stays the same but we wouldn't have to fire an expensive query multiple times, and the newcomers data would already be "there".

According to the comment the expensive part is the `sort_by`: isn't there a way to find an arbitrary threshold to filter out candidates?
It would avoid having to filter at all *after* computing the probas since we could simply not insert them, and hopefully have a small enough list of candidates.
By the way do you have any idea how many `sorting_candidates` there usually are for each iterations?

I'm not exactly sure what is the "range" of value that `name_similarity` can take so maybe my suggestion to find an arbitrary threshold doesn't make sense :s

gregorbg: > If I'm reading this correctly we do "for each newcomer iterate over all persons", since the number of persons is much much higher than the number of newcomers, why don't you do it the other way around: "for each person iterate over all newcomers"?
>
> I know the number of iterations stays the same but we wouldn't have to fire an expensive query multiple times, and the newcomers data would already be "there".

Fortunately, Rails is clever enough to cache the `Persons.all` invocation. Thus, I'm not sure whether changing the order of the loops changes anything. But it's certainly worth a try!

> According to the comment the expensive part is the `sort_by`: isn't there a way to find an arbitrary threshold to filter out candidates? It would avoid having to filter at all _after_ computing the probas since we could simply not insert them, and hopefully have a small enough list of candidates. By the way do you have any idea how many `sorting_candidates` there usually are for each iterations?

I'm not sure if I get your idea correctly. In my mind, establishing a threshold value doesn't change anything: Even if we have a threshold value of X percent, we still have to compare _every_ newcomer against _every_ existing person to potentially check whether the threshold is exceeded or not. So no matter which X we pick, we still have to do all of the comparisons at least once.

With that in mind, I've ported over the existing logic from PHP because it is already decently optimised: We calculate all similarities, then sort them in descending order, and then take the `2 * n` highest probas (where `n` is the number of similar persons we want to retrieve in the end) to filter for. Persons with a similarity value lower than the `2*n`-highest value are not even considered.

> I'm not exactly sure what is the "range" of value that `name_similarity` can take so maybe my suggestion to find an arbitrary threshold doesn't make sense :s

It's a percentage from 0 to 1.

viroulep: > I'm not sure if I get your idea correctly. In my mind, establishing a threshold value doesn't change anything: Even if we have a threshold value of X percent, we still have to compare every newcomer against every existing person to potentially check whether the threshold is exceeded or not. So no matter which X we pick, we still have to do all of the comparisons at least once.

Sorry if I wasn't clear: from the comment in the code below I understood that the `sorting_candidates.sort_by` below was very expensive, and based on your reply it must not be the case because there are like 10 elements in there.

However my idea can still be applied: if you take a look at the `only_probas` array it should contain as many elements as the number of persons right?
Then we have to sort that, which definitely sounds expensive.
However if you filter out right away, let's say with only people with a proba >0.9, `only_probas` and `persons_with_probas` now only contain the most relevant candidates, and hopefully sorting that should be significantly faster.

Of course this means you can't guarantee to always offer `n` people as a comparison (for instance if there is no match over 0.9), but I think you would avoid having to sort a large array for every newcomer.

Do you understand better what I meant now?

Of course if all these sorting/filtering is insignificant compared to the cost of computing the matching proba for everyone, then I think it's a minor detail that we can definitely ignore (I'd definitely update the performance comment to put it next to the really expensive section though!).

gregorbg: Ah, gotcha. Actually, we can do that, as long as WRT is confident enough that >0.9 catches all "relevant" cases. From a software perspective, I can set any threshold that you want. But then at the same time, you (as in, the WRT) take the responsibility for when the checks miss a duplicate just because it is baaaaarely under the (arbitrary) threshold.

With that being said, even with a threshold in place, we still have to compare over _every_ person in the entire DB to figure out whether that threshold is exceeded or not. So I'm not sure how much we're gaining.

Regarding the comment, actually `sort_by` is the most expensive part by far _if we computed the similarities directly in the sort_by lambda` (which is the straight-forward thing to do, basically reducing the entire method body to a one-liner). But I want to explain why we're taking this much more convoluted approach, which is far more complex than one `Persons.sort_by {}` invocation. And that's why I placed the comment next to the `sort_by`, because it is expensive and I'm trying to make a point why I'm avoiding it.

In the meantime, I found out that caching the `Person`s in a Ruby `@persons_cache` variable does save ~30% time, despite ActiveRecord claiming that it already caches.

viroulep (changes requested):
I noticed two crashes when using the page, these suggestions fixed them!

WcaOnRails/app/controllers/admin_controller.rb
```diff
@@ -224,6 +225,61 @@ def do_anonymize_person
...
+    render :finish_unfinished_persons
+  end
+
+  def do_complete_persons
+    # memoize all WCA IDs, especially useful if we have several identical semi-IDs in the same batch
+    # (siblings with the same last name competing as newcomers at the same competition etc.)
+    wca_id_index = Person.pluck(:wca_id)
+
+    ActiveRecord::Base.transaction do
+      params[:person_completions].each do |person_key, procedure|
```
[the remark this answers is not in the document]
gregorbg: Resolved by the fact that the competition ID (if any, explicit `nil` otherwise) is now included in the params.

WcaOnRails/app/controllers/admin_controller.rb
```diff
@@ -224,6 +225,61 @@ def do_anonymize_person
...
+          raise "Invalid action: #{action}" unless action == "merge"
+
+          new_person = Person.find(merge_id)
+
+          FinishUnfinishedPersons.adapt_results(nil, old_name, old_country, new_person.wca_id, new_person.name, new_person.countryId)
+        end
+      end
+    end
+
+    redirect_to :admin_finish_persons
```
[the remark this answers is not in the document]
gregorbg: Resolved when I changed the pagination behaviour.

WcaOnRails/lib/finish_unfinished_persons.rb
```diff
@@ -0,0 +1,211 @@
...
+
+      semi_id = nil
+      cleared_id = false
+
+      until cleared_id || letters_to_shift > WCA_QUARTER_ID_LENGTH
+        quarter_id = last_name[...(WCA_QUARTER_ID_LENGTH - letters_to_shift)] + padded_rest_of_name[...letters_to_shift]
+        semi_id = competition_year.to_s + quarter_id
+
+        unless available_id_spots.key?(semi_id)
+          # TODO: Merge with anonymization code
```
FinnIckler, line 59: Should this TODO be done before merge?

WcaOnRails/lib/finish_unfinished_persons.rb
```diff
@@ -0,0 +1,211 @@
...
+                                computed_semi_id: semi_id,
+                                similar_persons: similar_persons,
+                              })
+    end
+
+    unfinished_persons
+  end
+
+  def self.extract_roman_name(person_name)
+    name_matches = person_name.match(/(.*)\((.*)\)$/)
```
FinnIckler, line 103: Would be nice to have a comment here that we store People with non Roman name in the form `Roman Name (non roman name)` because the regex looks confusing otherwise

FinnIckler (commented):
One TODO left in code (intentionally?) and I would like one comment on a regex that took me a visit to regex101.com to understand. Looks good otherwise

PR #7728, opened 2023-03-09
[PHP WRT #5/5] Move Results posting process to Rails

gregorbg:
It turns out that basically the entire page can be reduced to "copy stuff over from InboxResults" -> "create newcomers" (cf. https://github.com/thewca/worldcubeassociation.org/pull/7726) -> "drop remaining entries from InboxPersons". DONE.

A lot of the "convenience links" to the competition admin pages are not needed anymore, because the new script now lives within the admin page of each competition. The UI in general is much more sleek, and looks _very_ unspectacular right now.

The table with Results/Scrambles overview at the end is not (yet) red/green. But I changed it from "Y/N" to "Yes/Nope." so that the strings have different lengths and the human eye should notice. But I'm very open to help for frontend reworks in general.

This PR probably needs to be merged in last (or mostly last) because it currently still links to some other old PHP panels. I have developed every PHP->Rails migration independtly, so on this branch the "Create Newcomers" script is still in PHP. Will need to update once everything else is merged.


WcaOnRails/app/controllers/admin_controller.rb
```diff
@@ -69,11 +70,100 @@ def clear_results_submission
...
+    ActiveRecord::Base.transaction do
+      @competition.inbox_results
+                  .joins("LEFT JOIN InboxPersons ON InboxPersons.id = InboxResults.personId AND InboxPersons.competitionId = InboxResults.competitionId")
+                  .select("InboxResults.*, InboxPersons.wcaId AS personWcaId, InboxPersons.countryId AS personCountryIso2")
+                  .each do |inbox_res|
+        person_id = inbox_res.personWcaId.presence || inbox_res.personId
+        person_country = Country.find_by_iso2(inbox_res.personCountryIso2)
+
+        # TODO: This runs a lot of validations and is slow(ish, aka. bearable but could be faster).
+        #   Is it safe to just let Results Team insert rows without Rails validations?
```
viroulep, line 105: If I'm not mistaken I think it is: both `InboxResult` and `Result` have similar validations for average, round format and so on, so whatever data is in Inbox, it should be pass the "regular" result validations!

WcaOnRails/config/routes.rb
```diff
@@ -78,7 +78,12 @@
     get '/admin/check-existing-results' => "admin#check_results", as: :admin_check_existing_results
     post '/admin/upload-json' => "admin#create_results", as: :admin_upload_results
     post '/admin/clear-submission' => "admin#clear_results_submission", as: :clear_results_submission
+    get '/admin/post-results' => 'admin#post_results', as: :admin_post_results
```
viroulep, line 81: If the routes changes something in the db, it definitely shouldn't be a get: most of the routes here should be `post` or `delete` I think.

You would still be able to use a simple `link_to` to link to them though, by passing `method: :delete`/`method: :post`.

WcaOnRails/app/views/admin/post_results.html.erb
```diff
@@ -0,0 +1,129 @@
...
+    <%= alert :info do %>
+      This competition has no data yet. <%= link_to "Upload something", competition_admin_upload_results_edit_path(@competition.id) %> to get started.
+    <% end %>
+  <% else %>
+    <% if @existing_data[:inbox_result] > 0 || @existing_data[:inbox_person] > 0 %>
+      <%= alert :info do %>
+        <p>This competition is in the process of having data uploaded. If necessary, you can:</p>
+        <ul>
+          <% if @existing_data[:inbox_result] > 0 %>
+            <li>clear <%= link_to "InboxResults", competition_admin_delete_inbox_results_path(@competition.id) %></li>
```
viroulep, line 15: Would you mind adding a confirmation for links which lead to destructive actions please?
I've definitely misclicked a lot of time and been saved by such confirmation :p

As far as the implementation is concerned I believe it's a matter of simply adding `data: { confirm: "some string" }` to add in the `link_to` :)

WcaOnRails/config/routes.rb
gregorbg: Cool, done!

WcaOnRails/app/views/admin/post_results.html.erb
gregorbg: Makes sense, I've added some straight forward confirmation messages.

WcaOnRails/app/controllers/admin_controller.rb
gregorbg: From a WST perspective, we just have to change two lines. But I want to be very clear here that I'm leaving this decision entirely to WRT, which also means that you would take the responsibility for any inconsistencies that stem from skipping the validations in this import step.

Personally (not WRT) I'd say double security net at the price of 15% more runtime is better than accidentally introducing unchecked/invalid Result rows into our DB.

WcaOnRails/app/views/admin/_import_inbox_step.erb
```diff
@@ -0,0 +1,58 @@
...
+    <%# So we just create one on the fly and "run" the checks without actually checking anything. %>
+    <% results_validator = ResultsValidators::CompetitionsResultsValidator.new %>
+    <% results_validator.validate [@competition.id] %>
+
+    <%= render "results_submission/results_preview_panel", results_validator: results_validator, max_height: "500px", expand_all: true, unindent_solves: true %>
+
+    <p><%= link_to "Finish importing", competition_admin_import_inbox_results_path(@competition), class: 'inbox-trigger', data: { jquery_method: :post } %>
+      results</p>
+  <% elsif render_step == :inbox_person %>
+    <%# TODO: Redo link pending other PR %>
```
viroulep, line 36: I'm leaving a comment here to make sure we don't forget about that TODO ;)

WcaOnRails/app/views/admin/_import_inbox_step.erb
```diff
@@ -0,0 +1,58 @@
...
+      <li>
+        <b style="color:<%= @existing_data[:newcomer_result] == 0 ? "#0B0" : "#F00" %>"><%= @existing_data[:newcomer_result] %></b>
+        newcomers in the Results table are still pending.
+      </li>
+    </ul>
+
+    <% if @existing_data[:newcomer_result] == 0 %>
+      <p>You
+        can <%= link_to "delete", competition_admin_delete_inbox_data_path(@competition.id, model: :inbox_person), class: 'inbox-trigger', data: { jquery_method: :delete, jquery_confirm: delete_inbox_confirm_message(:inbox_person) } %>
+        the inbox data, as all rows seem to have been fully imported.</p>
```
viroulep, line 53:
```suggestion
      <p>You
        can <%= link_to "delete the inbox persons", competition_admin_delete_inbox_data_path(@competition.id, model: :inbox_person), class: 'inbox-trigger btn btn-primary', data: { jquery_method: :delete, jquery_confirm: delete_inbox_confirm_message(:inbox_person) } %>
        , as all rows seem to have been fully imported.</p>
    <% else %>
      <%# TODO: Redo link pending other PR %>
      <p>Please <%= link_to "assign WCA IDs to newcomers", '/results/admin/persons_finish_unfinished.php?check=+Check+now+', target: '_blank', class: 'btn btn-primary' %>
      to proceed to the next step.</p>
```

The idea here is to show either the "assign WCA IDs" (which is required to proceed to the next step), or the "delete inbox persons" (which is required to proceed to the next step, and if we're offering that action I don't think we need to show the "assign" action anymore).

I also turned them into button so that it catches the eye more easily.

WcaOnRails/app/views/admin/_import_inbox_step.erb
```diff
@@ -0,0 +1,58 @@
...
+    <% results_validator.validate [@competition.id] %>
+
+    <%= render "results_submission/results_preview_panel", results_validator: results_validator, max_height: "500px", expand_all: true, unindent_solves: true %>
+
+    <p><%= link_to "Finish importing", competition_admin_import_inbox_results_path(@competition), class: 'inbox-trigger', data: { jquery_method: :post } %>
+      results</p>
+  <% elsif render_step == :inbox_person %>
+    <%# TODO: Redo link pending other PR %>
+    <p><%= link_to "Assign WCA IDs", '/results/admin/persons_finish_unfinished.php?check=+Check+now+', target: '_blank' %>
+      to newcomers</p>
```
viroulep, line 38:
```suggestion
    <p><%= link_to "Finish importing results", competition_admin_import_inbox_results_path(@competition), class: 'inbox-trigger btn btn-primary', data: { jquery_method: :post } %></p>
  <% elsif render_step == :inbox_person %>
```
This turns the link into a button to make it more catchy, and removes the "assign WCA IDs" link which has been moved below.

gregorbg:
I've incorporated your feedback about buttons, along with some more consistent wording. The final "Post Results" button is also a `btn-primary` now. Thanks for the suggestions! :smile:

PR #7729, opened 2023-03-11
Stripe hotifx: Add conversion back from Stripe denomination to RubyMoney

gregorbg:
We forgot/overlooked this when #7364 was merged. This adds back a reverse operation, so that 'special currencies' (see other PR for what that means) are correctly converted _back_ when inserting them into the Rails/Ruby world.

### DIFF
--- a/WcaOnRails/app/models/stripe_charge.rb
+++ b/WcaOnRails/app/models/stripe_charge.rb
@@ -36,4 +36,23 @@ class StripeCharge < ApplicationRecord
     # Stripe API will be happy as-is.
     amount_lowest_denomination
   end
+
+  def self.amount_to_ruby(amount_stripe_denomination, iso_currency)
+    # For the specifics of these currencies, see the comments in `amount_to_stripe`
+    if ZERO_DECIMAL_CURRENCIES.include?(iso_currency)
+      amount_div_hundred = amount_stripe_denomination.to_f / 100
+
+      # We're losing precision after dividing it down to the "smaller" denomination.
+      # Normally, this should not happen as the Stripe API docs specify that sub-hundreds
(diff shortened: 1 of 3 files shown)


danieljames-dj (approved):
LGTM

WcaOnRails/app/models/stripe_charge.rb
```diff
@@ -36,4 +36,23 @@ def self.amount_to_stripe(amount_lowest_denomination, iso_currency)
...
+  def self.amount_to_ruby(amount_stripe_denomination, iso_currency)
+    # For the specifics of these currencies, see the comments in `amount_to_stripe`
+    if ZERO_DECIMAL_CURRENCIES.include?(iso_currency)
+      amount_div_hundred = amount_stripe_denomination.to_f / 100
+
+      # We're losing precision after dividing it down to the "smaller" denomination.
+      # Normally, this should not happen as the Stripe API docs specify that sub-hundreds
+      # on the special currencies are not accepted and thus should never be returned by the API.
+      if amount_div_hundred.truncate != amount_div_hundred
+        raise "Trying to receive an amount of #{amount_stripe_denomination} #{iso_currency}, which is more precise than what the Stripe API returns for sub-hundred currencies"
```
danieljames-dj, line 49: Maybe Nit: Is it actually "Stripe API returns" or is it "Stripe API accepts" like above?

gregorbg: We're trying to convert an amount of money that was returned by Stripe to process further in our Ruby world.
In the other method above, we care about conversion when sending _to_ Stripe. Here, we care about receiving _from_ Stripe.

gregorbg: So it is an object returned by the Stripe API.

Issue #7209, opened 2022-09-11
Make Donation Field Clearer on Register page


EdHollingdale:
We often have people paying exactly double the registration fee, as they type the registration fee into the optional donation box (we then have to refund them the donation portion, losing stripe fees). 

I think this is due to the layout of the payment section: it has paid, remaining amount to be paid, so many people will type the amount they want to pay below that (i.e. the remaining amount to be paid) and then fill out their payment details. Hence they get charged double.

I feel it would be better to have something like a checkbox "I would like to make an additional optional donation that will go only to the organization team", and only after selecting that box does the Donation Amount field become visible. 

![image](https://user-images.githubusercontent.com/45908872/189510677-b84400f1-234f-4b20-83e6-f2be82850ab7.png)

This could be a good task to do in combination with #7077

dunkOnIT:
Needs attention as this currently causes frustration/issues for users when running competitions with donations enabled.

Other (possibly interim) options to consider to avoid double-payment:
* prominently displaying the total a user will pay
* when a donation is entered, prompt for confirmation that the user wants to make a donation

Basically, we need to break the "flow" of someone who is just trying to register as fast as possible, and not being attentive to what they're doing.

coder13:
> Basically, we need to break the "flow" of someone who is just trying to register as fast as possible, and not being attentive to what they're doing.

This could be addressed if we make it not part of the flow. Perhaps something after the fact "You have paid for your registration fee", however if you'd like to donate extra money, you can.

dunkOnIT:
https://github.com/thewca/worldcubeassociation.org/blob/master/WcaOnRails/app/views/registrations/_payment_form.html.erb#L5-L7

Should be a simple matter of adding one `<%= f.input :make_donation, as: :boolean ...etc... %>` and two lines of jQuery to toggle the visibility of the existing input based on the new checkbox that you've added. `WcaOnRails/app/assets/javascript/competition.js` has a LOT of samples on how to do that.
If you're feeling fancy, you can even add some CSS to animate the fade-in / fade-out of the donation input spinner.

PR #7731, opened 2023-03-13
Only show registration donation field when requested

simonkellly:
This adds a checkbox, which when checked will show the donation field which will otherwise be hidden, as talked about in #7209.


notrichardpeng:
I think for the hint missing key warning it's better to put an empty key in the locale rather than filling it in the code. Otherwise LGTM!

gregorbg (approved):
There's no clear/consistent guideline on when the i18n string should be empty.
In this particular case, we're not even using a "proper" Rails model and just stuff the values into impromptu symbolized keys, so it's not worth the extra effort.

Plus I'm reasonably confident that this won't need an extra hint, so no need to bother the translators with a blank/dummy "There is new stuff to translate!" email.

PR #7739, opened 2023-03-15
Update _organization_default.md

sarahstrong314:
Grammar fixes:
* "Closest living Delegate" without the hyphen suggested that the interpretation is the "closest Delegate who is alive" as opposed to the "the Delegate who is located the closest".
* A few sentences of the same section were written in present tense although the rest of the document was in past tense.
* One instance of "i.e." should have been "e.g." since Groupifier is not the only way to generate groups. 
* Three instances of "or" or "and" were used to begin an independent clause without a preceding comma.


coder13:
> "Closest living Delegate" without the hyphen suggested that the interpretation is the "closest Delegate who is alive" as opposed to the "the Delegate who is located the closest".

Completely off topic but I find it funny that I was thinking about this just week but only mentioned it to very few people haha.

gregorbg (approved):
Thanks for your keen eye! Although I'm sad to see this zombie joke go...

PR #7740, opened 2023-03-15
Restructure Stripe journal: Record all transactions

gregorbg:
Stripe has an onthology of payment objects, namely `PaymentIntent`, `Charge` and `Refund` (among others, but these are the ones relevant to us.)

Until now, we used to create a PaymentIntent and then handle the charge manually. We only log the Stripe `Charge` ID in the registration payment but the code is highly inconsistent with the rest. Sometimes, IDs for `PaymentIntent` and `Charge` are written into the same Stripe journal row, overriding each other upon update call. If a refund is made, that refund possibly conflicts with the entries in the Stripe logs because we don't journal it at all... Apart from writing it into a column in `registration_payments` that is thought to be for `Charge` IDs.

This PR aims to simplify working with the Stripe integration by:
- Renaming `StripeCharge` to `StripeTransaction`, which is supposed to be a general name for payment-related events going on with Stripe.
- Logging _everything_, no matter whether it's a `PaymentIntent`, a `Charge` or a `Refund`. Everything we send/communicate with the Stripe API, we also very precisely insert it into our journal.
- Referring the `RegistrationPayment` to the corresponding Stripe transaction not through an overloaded `stripe_charge_id` column (which half of the time isn't even containing proper `Charge` IDs) but rather through a polymorphic `receipt` association. So a receipt for a `RegistrationPayment` could really be anything (makes integrating other payment gateways much easier in the future!) and we just set the value to the corresponding `StripeTransaction` from our journal upon payment success.

Note that this PR primarily renames stuff. There is no substantial change in the payment logic (that comes in a second PR after this one has been merged) and the only non-renaming changes are a bit of backwards compatibility with old charges that are already in the system that may not have had the same journaling as the one introduced in this PR.

### DIFF
--- a/WcaOnRails/app/controllers/registrations_controller.rb
+++ b/WcaOnRails/app/controllers/registrations_controller.rb
@@ -497,7 +497,7 @@ class RegistrationsController < ApplicationController
         # Log the payment attempt
         stripe_transaction = StripeTransaction.create!(
           parameters: payment_intent_args,
-          api_type: "PaymentIntent",
+          api_type: "payment_intent",
           stripe_id: nil,
           status: "unknown",
           account_id: account_id,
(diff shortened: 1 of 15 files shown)


FinnIckler (approved):
LGTM! Looking forward to the second PR

PR #7769, opened 2023-03-21
Added competition tooltip for case where registration is deleted

### LINKED ISSUE #7656: Bookmarked Competition Shows Message for on the Waitlist When Registration is Deleted
Pretty much what the title is, when you have a competition bookmarked and then your registration gets deleted from said competition, the message that you get when you hover over the competition is the same as when you're on the waitlist, which doesn't make sense in this case. 

I have bookmarked MBLD in a Madison Meeting Room and got my registration deleted and in the screenshot below you can see the incorrect message. 
Screenshot (399)

Pretty obscure and low priority but just something I noticed.

dunkOnIT:
Fixes #7656. A third case for deleted registration was necessary - made this a secondary `if` statement for better readability.

### DIFF
--- a/WcaOnRails/app/helpers/competitions_helper.rb
+++ b/WcaOnRails/app/helpers/competitions_helper.rb
@@ -7,7 +7,14 @@ module CompetitionsHelper
     if competition.cancelled?
       messages << t('competitions.messages.cancelled')
     elsif registration
-      messages << (registration.accepted? ? t('competitions.messages.tooltip_registered') : t('competitions.messages.tooltip_waiting_list'))
+      # Display a tooltip on a user's bookmarked competition based on their registration status.
+      if registration.accepted?
+        messages << t('competitions.messages.tooltip_registered')
+      elsif registration.deleted?
+        messages << t('competitions.messages.tooltip_deleted')
+      else # If not delted or accepted, assume user is on the waiting list
+        messages << t('competitions.messages.tooltip_waiting_list')
+      end
     end
     visible = competition.showAtAll?
     messages << if competition.confirmed?
--- a/WcaOnRails/config/locales/en.yml
+++ b/WcaOnRails/config/locales/en.yml
@@ -1362,6 +1362,7 @@ en:
       reg_opens_too_early: "Registration is set to start in less than the required 48 hours."
       tooltip_registered: "You are registered."
       tooltip_waiting_list: "You are currently on the waiting list."
+      tooltip_deleted: "Unfortunately your registration has been deleted. Please check your email for details, or contact the organiser."
       confirmed_visible: "This competition is confirmed and visible"
       confirmed_not_visible: "This competition is confirmed but not visible"
       not_confirmed_visible: "This competition is not confirmed but visible"


WcaOnRails/app/helpers/competitions_helper.rb
```diff
@@ -7,7 +7,14 @@ def competition_message_for_user(competition, user)
     if competition.cancelled?
       messages << t('competitions.messages.cancelled')
     elsif registration
-      messages << (registration.accepted? ? t('competitions.messages.tooltip_registered') : t('competitions.messages.tooltip_waiting_list'))
+      # Display a tooltip on a user's bookmarked competition based on their registration status.
+      if registration.accepted?
+        messages << t('competitions.messages.tooltip_registered')
+      elsif registration.deleted?
+        messages << t('competitions.messages.tooltip_deleted')
+      else # If not delted or accepted, assume user is on the waiting list
```
kr-matthews, line 15:
```suggestion
      else # If not deleted or accepted, assume user is on the waiting list
```

WcaOnRails/app/helpers/competitions_helper.rb
```diff
@@ -7,7 +7,14 @@ def competition_message_for_user(competition, user)
...
+        messages << t('competitions.messages.tooltip_waiting_list')
+      end
     end
     visible = competition.showAtAll?
     messages << if competition.confirmed?
```
kr-matthews, line 20: This next step has the `messages <<` factored out to the front. Consistency is good, so I'd suggest doing the same thing on line 9: `messages << if competition.cancelled? ...`. It also makes it clear that this method only adds 2 messages to the array, which is not obvious at a glance with the current code.

WcaOnRails/config/locales/en.yml
```diff
@@ -1362,6 +1362,7 @@ en:
       reg_opens_too_early: "Registration is set to start in less than the required 48 hours."
       tooltip_registered: "You are registered."
       tooltip_waiting_list: "You are currently on the waiting list."
+      tooltip_deleted: "Unfortunately your registration has been deleted. Please check your email for details, or contact the organiser."
```
kr-matthews, line 1365:
```suggestion
      tooltip_deleted: "Unfortunately your registration has been deleted. Please check your email for details, or contact the organiser(s)."
```

WcaOnRails/app/helpers/competitions_helper.rb
dunkOnIT: Thanks Kevin! I've done a slightly more aggressive refactor of the function, that I think achieves that consistency and (at least for me) makes it more readable/followable overall.

WcaOnRails/app/helpers/competitions_helper.rb
```diff
@@ -2,26 +2,39 @@
...
-      else # If not delted or accepted, assume user is on the waiting list
-        messages << t('competitions.messages.tooltip_waiting_list')
-      end
+    # Generates a list of messages, which will be combined and displayed in a tooltip to the user in their bookmarked
+    # competitions list when they hover over a competition.
+    # Message indicates the state of the competition, and the state of the user's registration.
+
+    messages = [] # Messages to be combined
+
+    if competition.cancelled? # If the competition is cancelled, just return the cancellation message to the user.
```
kr-matthews, line 11: Nice, I like the changes in general, but I think it might be a little bit over-commented in places. This comment, for example, is literally restating the code, and isn't really significantly clearer. Plus if this gets changed in the future, both the code and the comment needing updating, and it's possible the comment will be missed, leading to confusion down the line.
Same with the comment above the `join`.

WcaOnRails/app/helpers/competitions_helper.rb
```diff
@@ -2,26 +2,39 @@
...
+      return t('competitions.messages.cancelled')
+    end
+
+    # Registration state messages
```
kr-matthews, line 15: A good rule of thumb is that if you need a comment to describe the next bunch of code, it's often a good idea to extract that code into a helper function and give it a descriptive name. If the name is well-chosen, it eliminates the need for the comment - and makes the original function easier to understand at a glance because it's shorter and calls a well-named helper function.

I'm not saying it's necessary to make such changes in this case, but if the rest of the function below here just called a helper `addRegistrationStatusMessageIfRegistered` then called another helper `addCompetitionVisibilityMessage` then returned `messages.join()`, it would be super easy to understand.

Although I'm not very familiar with Ruby, so someone correct me if that's an un-Ruby way of doing things.

WcaOnRails/app/helpers/competitions_helper.rb
dunkOnIT: Great points, especially about needing to change two things instead of 1, and using things like variable names to communicate without the need for comments

I've opted to leave the comment next to `competition.cancelled?`, to explain the logic/intention of returning immediately there. It might just be my baby brain, but it feels like it makes it easier to understand :P

WcaOnRails/app/helpers/competitions_helper.rb
dunkOnIT: Very interesting idea, thanks! I did what you suggested in terms of helper functions, mostly as a learning exercise. Would be curious on your feedback! And very happy if your feedback is "this is worse, revert it" - I'm not convinced of it myself.

If I was in Python I would have defined the helper functions as inline functions, to minimise clutter in the rest of the file. Ruby doesn't allow this, so I've defined them as separate functions. Curious what your thoughts are on this - I think it makes the `competition_message_for_user` function much easier to understand, but makes `competitions_helper.rb` as a whole look more cluttered? Perhaps we could define these as private methods to help alleviate that?

kr-matthews: I like it this way, with the helper functions. (I'm not familiar with Python.) Only thing I would change is putting the primary function, `get_competition_status_message`, above the helpers. So the most important thing is at the top, and if you want more details you can continue reading down. (Unless Ruby would complain about using something before it's defined, in which case ignore this.) If these helpers aren't intended to be used outside then making them private makes sense to me.

WcaOnRails/config/locales/en.yml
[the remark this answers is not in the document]
dunkOnIT: Fixed

WcaOnRails/app/helpers/competitions_helper.rb
dunkOnIT: Thanks! Functions need to be defined before they're called in Ruby, and I agree it looks gross. Luckily `private` lets us kill two birds with one stone here - we can make stuff private _and_ get them out the way: 
* I've moved the helper functions into a `private` section at the bottom of the file, which is the Ruby convention anyway.
* Also moved some helpers elsewhere in the file with inline `private` definitions, which isn't great Ruby convention

Aside: In Python, I could do the following (I think, I haven't played with nested function definitions much):

```python
def main_method:
  messages = []
  messages.append(helper_function_1())
  messages.append(helper_function_2())

  def helper_function_1():
    # Do some stuff
    return

  def helper_function_2():
    # Do some stuff
    return

  return messages

**Comment by dunkOnIT** [comment]
Currently blocked by #7782

**Comment by kr-matthews** [comment]
The changes look good to me, but it would be best if someone more familiar with rails gave it an approval.

> Do we need to display the "This competition is confirmed and visible" message? It's assigned in Line 32.
> 
> I feel like its useless for competitors, and for organisers we only need need to display a message if a comp _isn't_ confirmed or visible.
> 
> Most of the time I feel like the message creates redundant clutter for normal users.

I don't have any strong opinions on this, but your reasoning makes sense.

File: WcaOnRails/app/helpers/competitions_helper.rb [hunk 6]
```diff
@@ -2,20 +2,20 @@
...
-    registration = competition.registrations.find_by_user_id(user.id)
-    if competition.cancelled?
-      messages << t('competitions.messages.cancelled')
-    elsif registration
-      messages << (registration.accepted? ? t('competitions.messages.tooltip_registered') : t('competitions.messages.tooltip_waiting_list'))
+    # Generates a list of messages, which will be combined and displayed in a tooltip to the user in their bookmarked
+    # competitions list when they hover over a competition.
+    # Message indicates the state of the competition, and the state of the user's registration.
+
+    messages_to_join = []
```
gregorbg, line 9: Super nitpick, not a blocker but more a guidance: Don't name variables according to what will be done with them later. `messages` is totally fine, and calling it `messages_to_join` is considered bad style because it "forbids" you to do anything else with that variable except joining.

gregorbg (approved):

PR #7777, opened 2023-03-23
Change sql notice

dunkOnIT:
Updates the notice regarding disabled filters on [records](https://www.worldcubeassociation.org/results/records) and [rankings](https://www.worldcubeassociation.org/results/rankings/333/single) pages.

* Wording needs to be confirmed by someone like @gregorbg 
* I opted not to remove the translation string for the previous announcement - would rather remove it all together when the announcement is removed entirely. I've added a comment remind to remove both translation strings when (when!) the announcements are removed.


dunkOnIT:
I'm going to cry. Somehow I don't understand git at all. I was SO careful to create a fresh branch with a clean `git status`. How did all of my commits from another branch come with?

dunkOnIT:
Closing in favour of #7778

viroulep:
I'm not sure how you created the branch but you probably didn't specify explicitly the starting point, and therefore it took the current commit as the starting point, and you were probably not on `master`.
I think both git switch and the "old way" git checkout take a starting point that you can force; using `git switch -c somename upstream/master` should likely be helpful for you in the future ;) (assuming you have a remote `upstream` pointing to the WCA repo!)

git status only shows the status with regard to the last commit, so having a "clean" git status really doesn't guarantee anything for a new branch's starting point (but it does tell you on which branch you are :p)!

dunkOnIT:
Thanks Philippe! You guessed my issue exactly :D I didn't create my new branch from `master`. Learned some good lessons though, as we do xD will be using `git diff --name-only master` going forward. As well as forcing the upstream branch, that's a great feature I didn't know about!

PR #7778, opened 2023-03-23
Edited results announcement and added new translation string

dunkOnIT:
(Update to #7777)

Updates the notice regarding disabled filters on [records](https://www.worldcubeassociation.org/results/records) and [rankings](https://www.worldcubeassociation.org/results/rankings/333/single) pages.

* Wording needs review
* I opted not to remove the translation string for the previous announcement - would rather remove it all together when the announcement is removed entirely. I've added a comment remind to remove both translation strings when (when!) the announcements are removed.

### DIFF
--- a/WcaOnRails/app/views/results/rankings.html.erb
+++ b/WcaOnRails/app/views/results/rankings.html.erb
@@ -8,8 +8,6 @@
 <div class="container">
   <h1><%= yield(:title) %></h1>
   <p><%= t("results.last_updated_html", timestamp: wca_local_time(ranking_timestamp)) %></p>
-
-  <%# Note: When removing this announcement and its translation string, also remove results.heading_disabled_functions from i18n files. %>
   <p><i><%= t('results.filters_fixes_underway') %></i></p>
 
   <div id="results-selector" class="results-select form-inline">
--- a/WcaOnRails/app/views/results/rankings.html.erb
+++ b/WcaOnRails/app/views/results/rankings.html.erb
@@ -8,7 +8,9 @@
 <div class="container">
   <h1><%= yield(:title) %></h1>
   <p><%= t("results.last_updated_html", timestamp: wca_local_time(ranking_timestamp)) %></p>
-  <p><i><%= t('results.heading_disabled_functions') %></i></p>
+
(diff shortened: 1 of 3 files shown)


dunkOnIT:
Removed the unused translation string in `en.yml` due to a failing test. Two interesting notes:
* The test only seems to check the `en.yml` file for unused keys - not other translations
* I wasn't able to replicate the test's results when running `rails-tasks unused` on my local machine. Not sure if I was doing something wrong, or if there's a replication issue here.

Either way, fairly minor points.

gregorbg:
> * The test only seems to check the `en.yml` file for unused keys - not other translations

This is completely intended as `en.yml` is the only file we (the WST) control. Other files are contributed by volunteer translators from the community, and there is absolutely no use for automated tests that check whether the volunteers have found time yet to add the most recent translation strings.

WcaOnRails/app/views/results/rankings.html.erb
```diff
@@ -8,7 +8,9 @@
 <div class="container">
   <h1><%= yield(:title) %></h1>
   <p><%= t("results.last_updated_html", timestamp: wca_local_time(ranking_timestamp)) %></p>
-  <p><i><%= t('results.heading_disabled_functions') %></i></p>
+
+  <%# Note: When removing this announcement and its translation string, also remove results.heading_disabled_functions from i18n files. %>
```
gregorbg, line 12: This `Note:` comment is now redundant.

dunkOnIT: Fixed

WcaOnRails/app/views/results/records.html.erb
```diff
@@ -8,7 +8,9 @@
 <div class="container">
   <h1><%= yield(:title) %></h1>
   <p><%= t("results.last_updated_html", timestamp: wca_local_time(record_timestamp)) %></p>
-  <p><i><%= t('results.heading_disabled_functions') %></i></p>
+
+  <%# Note: When removing this announcement and its translation string, also remove results.heading_disabled_functions from i18n files. %>
```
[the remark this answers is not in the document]
dunkOnIT: Fixed

gregorbg (approved):

dunkOnIT:
Ok cool thanks! That makes a lot of sense :D

PR #7780, opened 2023-03-23
Update Shakapacker to version 6.6.0

gregorbg:


gregorbg:
Merging as a normal Dependabot PR (which it essentially is, see mentions, except that Dependabot is too stupid to link the two dependency updates into one)

PR #7785, opened 2023-03-25
Fix validations for non-competing registrations

### LINKED ISSUE #7775: Disable force comments validation for non-competing registrations
Some competitions have organizers/delegates that don't compete (have no registration), however on WCIF patch we try to create one with `is_competing = false`:

https://github.com/thewca/worldcubeassociation.org/blob/0518f057433bd0c5bc318f4516e20ea5d583759e/WcaOnRails/app/models/competition.rb#L1765-L1771

This breaks on this validation:

https://github.com/thewca/worldcubeassociation.org/blob/0518f057433bd0c5bc318f4516e20ea5d583759e/WcaOnRails/app/models/registration.rb#L278-L283

It would make sense to disable that validation for non-competing registrations, which we already do here:

https://github.com/thewca/worldcubeassociation.org/blob/0518f057433bd0c5bc318f4516e20ea5d583759e/WcaOnRails/app/models/registration.rb#L251-L256

---

On a more general note: whenever there is an organizer/delegate that doesn’t compete, we create a registration for them the first time there is a WCIF patch; which shouldn’t necessarily break anything (other than this issue), but it is a bit weird.

gregorbg:
Fixes #7775
Also fixes the issue with AnkaraSummer2023 that reached WST via mail.

Rough draft because I'm at a conference and low on battery. If tests fail, please someone pick this up and follow the line of thought to fix the tests.

### DIFF
--- a/WcaOnRails/app/models/registration.rb
+++ b/WcaOnRails/app/models/registration.rb
@@ -235,7 +235,7 @@ class Registration < ApplicationRecord
   # change doesn't lead to an invalid state.
   validate :user_can_register_for_competition, on: :create
   private def user_can_register_for_competition
-    cannot_register_reasons = user&.cannot_register_for_competition_reasons(competition)
+    cannot_register_reasons = user&.cannot_register_for_competition_reasons(competition, is_competing: self.is_competing?)
     if cannot_register_reasons.present?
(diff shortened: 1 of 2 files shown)


FinnIckler (approved):
LGTM!

PR #7792, opened 2023-03-27
Refactor ResultsValidators to bundle data loading

gregorbg:
## Problem situation
With the perspective PHP refactor in #7643 we are facing the possibility of running `ResultsValidator`s on **A LOT** of competitions at once. This is problematic because right now, all validators are loading data pretty much "as they please". This leads to a high amount of potentially inefficient SQL queries.

## Suggested solution
Every `ResultsValidator` "announces" which data it needs by virtue of an `.includes` specification on the `Competitions` table. When the validator is run, all of this data is bundled into one central (newly introduced!) `ValidatorData` object. This object can also be shared across multiple validators when they are bundled via `CompetitionsResultsValidator`.

## Benefits
- The individual validators don't have to worry about the several different "entry points" to validation (pass competition ID plus model VS. pass `Resultable` rows) anymore. `GenericValidator` handles the entry point and bundles the different variants into `ValidatorData` which is then passed down.
- The individual validators also don't have to worry about `reset_state` and similar anymore.
- The combined validator can now efficiently batch Competitions if required.

## Remarks
- There are additional abstractions on `Persons` and `InboxPersons` now, because there are special cases depending on whether you're actually checking the real results or not. Both models now have a `ref_id` (which is the actual WCA ID in case of `Person`s and the DB increment ID in case of `InboxPerson`s)
- One small piece of logic in the `PersonsValidator` changed: Both "types" of persons now have a `wca_person` shorthand which yields the actual `Person` model (in case of `Person`, just returns `self` and in case of `InboxPerson` tries to load via `wca_id` foreign key if present). This avoids an additional SQL query just for one validator.
- The batching in `CompetitonsResultsValidator` is not yet used, because in the current codebase we only use it for validating single competitions. The aforementioned PR will use the feature after a rebase.

### DIFF
--- a/WcaOnRails/lib/results_validators/advancement_conditions_validator.rb
+++ b/WcaOnRails/lib/results_validators/advancement_conditions_validator.rb
@@ -29,7 +29,7 @@ module ResultsValidators
       false
     end
 
-    protected def competition_associations
+    def competition_associations
       {
         rounds: [],
       }
--- a/WcaOnRails/lib/results_validators/advancement_conditions_validator.rb
+++ b/WcaOnRails/lib/results_validators/advancement_conditions_validator.rb
@@ -35,7 +35,7 @@ module ResultsValidators
       }
     end
 
-    protected def run_validation(validator_data)
+    def run_validation(validator_data)
       ordered_round_type_ids = RoundType.order(:rank).all.map(&:id)
 
       validator_data.each do |competition_data|
--- a/WcaOnRails/lib/results_validators/advancement_conditions_validator.rb
+++ b/WcaOnRails/lib/results_validators/advancement_conditions_validator.rb
@@ -29,50 +29,49 @@ module ResultsValidators
       false
     end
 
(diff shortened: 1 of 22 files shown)


WcaOnRails/lib/results_validators/advancement_conditions_validator.rb
```diff
@@ -29,50 +29,49 @@ def self.has_automated_fix?
...
 
-      competitions_start_dates = Competition.where(id: results_by_competition_id.keys).select(:id, :start_date).to_h do |c|
-        [c.id, c.start_date]
-      end
+      validator_data.each do |competition_data|
+        competition = competition_data.competition
+        comp_start_date = competition.start_date
 
-      results_by_competition_id.each do |competition_id, results_for_comp|
-        competition = Competition.includes(:rounds).find(competition_id)
```
gregorbg, line 44: `.includes(:rounds)` moved to the new `competition_associations` method that you can see on the right

PR #7826, opened 2023-04-11
Check if announced is null before ordering

campos20:
After clicking order by announcement, we are currently getting this error

image

This PR aims to fix it

The root cause of it was actually this line accessing a method on a nil string

https://github.com/thewca/worldcubeassociation.org/blob/63808dcf3487399d669f4f6a4bf1b8383cbe241e/WcaOnRails/app/views/competitions/_index_table.html.erb#L21

### DIFF
--- a/WcaOnRails/app/models/competition.rb
+++ b/WcaOnRails/app/models/competition.rb
@@ -72,7 +72,7 @@ class Competition < ApplicationRecord
       ).group(:id)
   }
   scope :order_by_date, -> { order(:start_date, :end_date) }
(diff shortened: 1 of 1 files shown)


danieljames-dj (approved):

gregorbg:
I am not a big fan of filtering out a state that shouldn't even be possible in the first place. WCAT has been contacted to advise on the viability of a more bullet-proof solution (i.e. disallow patching competitions into a state where `showAtAll` is true but `announced_at` is `nil`).

In any case, thank you for the fix! Merging to avoid getting us even more swamped with emails.

PR #7830, opened 2023-04-13
Make the password reset link clickable.

herpiko:
Currently this is how the email message appears on my inbox. No clickable link. I've to select the particular text then copy and paste it into a browser. This journey is not feasible in mobile phone.

Screenshot 2023-04-13 at 21 56 35

### DIFF
--- a/WcaOnRails/app/views/devise/mailer/reset_password_instructions.html.erb
+++ b/WcaOnRails/app/views/devise/mailer/reset_password_instructions.html.erb
@@ -11,7 +11,7 @@
 </p>
 
 <p>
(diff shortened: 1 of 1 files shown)


danieljames-dj (approved):
LGTM, thanks @herpiko for fixing this.

gregorbg (changes requested):
Rails has a `link_to` helper that can be used here.

Also, this feels like a fix designed for very specific email clients. Mail is very fragile in terms of being classified as spam easily (we have personal experience with that...) How can we be sure that changing this to an a-href link does not conversely affect our "credibility" with the Google Mail client for example?

gregorbg:
Closing due to inactivity

PR #7845, opened 2023-04-19
Fixes the bug that dues estimate throwing error in announced comps

danieljames-dj:


gregorbg (changes requested):
Why is this a proper fix? Can you describe why you chose this jQuery galore over a simple plugin?

Also, I have a gut feeling that `serializeJSON()` may be used in other places as well, so deleting it altogether may be a bit too much

danieljames-dj:
The issue was that `serializeJSON()` was not giving values of disabled fields (https://stackoverflow.com/questions/15958671). So, instead of fetching all the variables from the form, I thought of just fetching the required one.

Also, while removing `serializeJSON`, I confirmed that it is not used anywhere else in the code.

danieljames-dj:
@notrichardpeng Okay, I didn't knew `serializeJSON` was to reduce number of `.val()` calls. But I feel `serializeJSON` can be a bit more expensive as it reads all the values of the form while `.val()` reads only the required one. I also agree the difference is very very negligible.

Anyway I've replaced most of the `.val()` with rails code (except two places where I couldn't find an alternative).

Reason why I don't want `serializeJSON` back is to reduce jquery code in our codebase which we are anyway planning to get rid of completely in future.

notrichardpeng:
The problem with rails code is that it doesn't dynamically update. If one of inputs in the fields is changed the dues wouldn't update accordingly, unless if the user refreshes the page. So we should either use `serializeJSON` or a bunch of `.val()`. Which one is better might still be up for debate, perhaps @gregorbg can provide more opinion on his side. Maybe it's better to take care of this decision later. Right now what we're trying to fix is adding a condition for when `comp_limit` is disabled so we don't get an error for dues estimation, and a `.val()` fix for it seems to be good.

danieljames-dj:
Oh yes, you are right @notrichardpeng. So, now there are three options:
1. Use `.val()`.
2. Bring back serializeJSON.
3. Use rails.

Among 1 and 2, I prefer 1 because `.val()` is actually not so expensive, and `serializeJSON` is like fetching every element (maybe internally through `val()`, not sure) and convert to JSON.

3 is the best solution. I created a function inside `competition.rb` which will update the value of dues estimate, and it will be rendered in the `competition_form.html.erb`. But I'm not able to figure out how to call the function to update dues estimate, whenever there is a change in competition_form.

@gregorbg can you please give your input on the same?

gregorbg:
You should go with option 1 and use `val` for the time being. The old `serializeJSON` is not expensive per se, but it provides way too much information for way too little benefit (we only use 3 to 5 fields out of the 20+ fields that it serializes, so even though the serialization is cheap in terms of code execution, it's a waste and completely overkill in terms of software design)

Option 3 is flawed. The Rails competition model is part of our **backend**, which means that it represents the competition **as is currently stored in our database**. If the user makes some changes to the frontend input field specifying the competitor limit, the backend couldn't care less. Injecting values through `<%= %>` template notation or through a dedicated method in `competition.rb` both rely on what is currently saved in the DB. This would in theory even work for announced competitions that are "locked" for the average user (except Admins and WCAT) but it would require having separate logic for calculation of dues for unannounced competitions (when organisers and Delegates can dynamically update frontend form fields without saving the edits in the database) and announced competitions (which are generally consistent with our database). TL;DR Use jQuery to fetch these values no matter what.

WcaOnRails/app/views/competitions/_competition_form.html.erb
```diff
@@ -250,21 +248,19 @@
...
         <script>
           function UpdateDues(){
-
-            var obj = $("#competition-form").serializeJSON();
-            // Coerce compLimitEnabled to a bool, as it is a string obj['competition[competitor_limit_enabled]'] returns a string
-            var compLimitEnabled = ("true" === obj['competition[competitor_limit_enabled]'])
-            var compLimit = compLimitEnabled ? obj['competition[competitor_limit]'] : null
-            var currencyCode = obj['competition[currency_code]']
-
+            var compLimitEnabled = $("#competition_competitor_limit_enabled").val();
```
gregorbg, line 251: You are missing the previous comparison to `"true"` to make sure that it's coerced into a boolean on the JS side. If you just use `.val()` there is a risk that this `var compLimitEnabled` contains a **string** that reads `"true"` or `"false"` which is dangerous because `if ("false") console.log("foo") else console.log("bar")` logs `foo`. Non-empty strings are truthy in JS, even if the string reads "false".

For reference, the previous code was
```js
// Coerce compLimitEnabled to a bool, as it is a string obj['competition[competitor_limit_enabled]'] returns a string
var compLimitEnabled = ("true" === obj['competition[competitor_limit_enabled]'])
```

gregorbg (changes requested):
One small but important nitpick about coercing variable types in JS

WcaOnRails/app/views/competitions/_competition_form.html.erb
danieljames-dj: Done, thanks for pointing it out, @gregorbg.

gregorbg (approved):

PR #7851, opened 2023-04-26
Registration error fixes

dunkOnIT:
**Disclaimer: I'm not sure this is worth spending time on. It feels fairly complex relative to the benefit it brings, and our time is likely better spent on just implementing more fundamental reg fixes. If someone sees an easy way to merge this with high confidence, though, it might be worth it.** 

This PR does 2 things:
1. Adds a basic text notice telling users not to refresh after registration is submitted
2. Adds error handling for a Duplicate Record Exists error which continues the registration flow if a record already exists.

I'm very uncertain on the implementation of change (2). A few notes:
- Error handling was chosen instead of checking if the record already existed in the DB, to prevent multiple calls to the DB under periods of already high load.
- I've opted to allow registration emails to be sent again in the case of form resubmission. This is a crude solution, but I figured the trade off of more emails vs 
- I have no idea how to implement tests that check whether this fix actually works, or if it introduces other errors.

### DIFF
--- a/WcaOnRails/app/controllers/registrations_controller.rb
+++ b/WcaOnRails/app/controllers/registrations_controller.rb
@@ -645,39 +645,51 @@ class RegistrationsController < ApplicationController
       return
     end
     @registration = @competition.registrations.build(registration_params.merge(user_id: current_user.id))
-    if @registration.save
+    begin
+      if @registration.save
+        saved_registration_actions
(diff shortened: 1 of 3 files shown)


dunkOnIT:
Closing in favour of server improvements and imminent (hopefully) new Registration Service

PR #7896, opened 2023-05-10
Competition form improvement

DanielEgdal:
This should clear up some confusion we see from a lot of a delegate, and cut down on the amount explanation wcat has to do for some competitions.

### DIFF
--- a/WcaOnRails/app/controllers/competitions_controller.rb
+++ b/WcaOnRails/app/controllers/competitions_controller.rb
@@ -4,9 +4,6 @@ class CompetitionsController < ApplicationController
   include ApplicationHelper
 
   PAST_COMPETITIONS_DAYS = 90
-  MAX_CELL_NAME_LENGTH = 32
-  MAX_ID_LENGTH = 32
-  VALID_NAME_RE = /\A([-&.:' [:alnum:]]+) (\d{4})\z/ # Duplicate of that in models/competition.rb
   CHECK_SCHEDULE_ASSOCIATIONS = {
     competition_events: [:rounds],
     competition_venues: {
@@ -554,20 +551,19 @@ class CompetitionsController < ApplicationController
       end
     elsif @competition.update(comp_params_minus_id)
       # The two extra updates here are done to automatically compute the cellname and ID for competitions with a short name.
-      m = VALID_NAME_RE.match(@competition.name) # These three lines are from models/competition.rb
+      m = Competition::VALID_NAME_RE.match(@competition.name) # These three lines are from models/competition.rb
       name_without_year = m[1]
       year = m[2]
       year_space = " " + year
       if @actually_using_organizer_view && @competition.name.length <= 32
-        @competition.update(cellName: name_without_year.truncate(MAX_CELL_NAME_LENGTH - year_space.length) + year_space)
+        @competition.update(cellName: name_without_year.truncate(Competition::MAX_CELL_NAME_LENGTH - year_space.length) + year_space)
       end
 
       if @actually_using_organizer_view && @competition.name.length <= 32
         safe_name_without_year = ActiveSupport::Inflector.transliterate(name_without_year).gsub(/[^a-z0-9]+/i, '')
-        inferred_id = safe_name_without_year[0...(MAX_ID_LENGTH - year.length)] + year
+        inferred_id = safe_name_without_year[0...(Competition::MAX_ID_LENGTH - year.length)] + year
         @competition.update(id: inferred_id)
       elsif new_id && !@competition.update(id: new_id)
-        # 2023 remark, The code inside this elsif doesnt seem to actually do anything, but the magic is performed during the elsif
         # Changing the competition id breaks all our associations, and our view
         # code was not written to handle this. Rather than trying to update our view
         # code, just revert the attempted id change. The user will have to deal with
--- a/WcaOnRails/app/controllers/competitions_controller.rb
+++ b/WcaOnRails/app/controllers/competitions_controller.rb
@@ -550,7 +550,7 @@ class CompetitionsController < ApplicationController
       end
     elsif @competition.update(comp_params_minus_id)
       # Automatically compute the cellName and ID for competitions with a short name.
-      if @competition_organizer_view && @competition.name.length <= Competition::MAX_CELL_NAME_LENGTH
+      if !@competition.confirmed? && @competition_organizer_view && @competition.name.length <= Competition::MAX_CELL_NAME_LENGTH
         old_competition_id = @competition.id
         @competition.create_id_and_cell_name(force_override: true)
 
--- a/WcaOnRails/app/controllers/competitions_controller.rb
+++ b/WcaOnRails/app/controllers/competitions_controller.rb
@@ -557,9 +557,7 @@ class CompetitionsController < ApplicationController
       year_space = " " + year
       if @actually_using_organizer_view && @competition.name.length <= Competition::MAX_CELL_NAME_LENGTH
         @competition.update(cellName: name_without_year.truncate(Competition::MAX_CELL_NAME_LENGTH - year_space.length) + year_space)
-      end
-
-      if @actually_using_organizer_view && @competition.name.length <= 32
+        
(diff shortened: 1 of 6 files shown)


gregorbg:
Can you please describe what you're actually trying to achieve? In particular, please be more specific than "improve the form".

I get a good idea of what you're trying to do just by staring at the code, but depending on your exact use case I can recommend some better alternatives to you.

DanielEgdal:
Sure:

1. Give WCAT/Delegates a warning if the delegate didn't provide a deadline for the waiting list.
2. Rename the 'nickname' of a competition to 'shortname', since people often think you need to come up with something smart / witty. Additionally clarify the field on the form.
3. From the organizer view: Block the ability to edit the ID/shortname in the case the competition name is shorter than 32 chars, as this is what WCAT enforces anyway.
3.1. Automatically perform updates to the shortname/ID on the organiser view, as the organiser cannot edit it.
3.2. Still allow edits on the admin view of the form.
3.3. Include/Alter test cases for the functionality.

WcaOnRails/config/locales/en.yml
```diff
@@ -296,7 +296,7 @@ en:
         showAtAll: "Make competition visible to the public"
         name: "Name"
         name_reason: "The reason for the name"
-        cellName: "Competition nickname"
+        cellName: "Competition shortname"
```
gregorbg, line 299: Spelling: "short name" (confirmed by native English speaker)

WcaOnRails/app/views/competitions/_competition_form.html.erb
```diff
@@ -28,14 +28,17 @@
       <% disable_form = !can_edit_competition %>
 
       <fieldset <%= disable_form && "disabled" %>>
+        <% edit_cellName_id = !(@competition_admin_view || (@competition.name.length > 32)) %>
```
gregorbg, line 31: Are you sure the second pair of brackets is needed?
Also, please reference `Competitions::MAX_CELL_NAME_LENGTH` instead of magically inserting the literal `32` out of nowhere.

WcaOnRails/app/controllers/competitions_controller.rb
```diff
@@ -532,6 +535,7 @@ def update
     @competition = competition_from_params(includes: CHECK_SCHEDULE_ASSOCIATIONS)
     @competition_admin_view = params.key?(:competition_admin_view) && current_user.can_admin_competitions?
     @competition_organizer_view = !@competition_admin_view
+    @actually_using_organizer_view = !params.key?(:competition_admin_view) # If the edit is done through this view
```
gregorbg, line 538: Why is this relevant? Can't you use `@competition_organizer_view` directly above?
From what I can tell just by looking at the code itself, this creates an edge case only when somebody who isn't an admin uses the admin view, which shouldn't even be possible in the first place.

WcaOnRails/app/controllers/competitions_controller.rb
```diff
@@ -549,7 +553,21 @@ def update
...
+      if @actually_using_organizer_view && @competition.name.length <= 32
+        @competition.update(cellName: name_without_year.truncate(MAX_CELL_NAME_LENGTH - year_space.length) + year_space)
+      end
+
+      if @actually_using_organizer_view && @competition.name.length <= 32
+        safe_name_without_year = ActiveSupport::Inflector.transliterate(name_without_year).gsub(/[^a-z0-9]+/i, '')
+        inferred_id = safe_name_without_year[0...(MAX_ID_LENGTH - year.length)] + year
+        @competition.update(id: inferred_id)
+      elsif new_id && !@competition.update(id: new_id)
+        # 2023 remark, The code inside this elsif doesnt seem to actually do anything, but the magic is performed during the elsif
```
gregorbg, line 570: Yes, this code does something very relevant, which is related to the fact that our competition table has no `AUTO INCREMENT` id column. Please remove this comment.

WcaOnRails/app/controllers/competitions_controller.rb
```diff
@@ -549,7 +553,21 @@ def update
...
+      year = m[2]
+      year_space = " " + year
+      if @actually_using_organizer_view && @competition.name.length <= 32
+        @competition.update(cellName: name_without_year.truncate(MAX_CELL_NAME_LENGTH - year_space.length) + year_space)
+      end
+
+      if @actually_using_organizer_view && @competition.name.length <= 32
+        safe_name_without_year = ActiveSupport::Inflector.transliterate(name_without_year).gsub(/[^a-z0-9]+/i, '')
+        inferred_id = safe_name_without_year[0...(MAX_ID_LENGTH - year.length)] + year
+        @competition.update(id: inferred_id)
```
gregorbg, line 568: There is already a section of code in `competition.rb`, method `create_id_and_cell_name`, which is automatically run on inserts and updates. Why is it not satisfactory to use/adapt that part of the code?

gregorbg (changes requested):
Code duplication in a lot of places, please re-use variables and constants as much as (reasonably) possible.

WcaOnRails/config/locales/en.yml
```diff
@@ -521,7 +521,7 @@ en:
       #context: for a competition
       competition:
         name: "The full name of the competition including the year at the end (e.g., Danish Open 2016). Be sure to capitalize. The name must contain only alphanumeric characters, dashes(-), ampersands(&), periods(.), colons(:), apostrophes('), and spaces( )."
-        cellName: "A short name for displaying. If the name of the competition is already short, you can re-use the name (e.g., Europe 2006)."
+        cellName: "A short name for displaying. If the name of the competition is under 32 characters this should not be edited and should be the same as the normal name."
```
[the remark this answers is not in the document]
DanielEgdal: I don't see the difference in below/under here. I wrote 'below' now (and did the other change)

WcaOnRails/app/views/competitions/_competition_form.html.erb
DanielEgdal: About the first question: Yes? If I do `edit_cellName_id = !@competition_admin_view || (@competition.name.length > 32)` then the name/nickname is not possible to edit from the organiser view, despite the name being over 32 chars.
(second half done)

WcaOnRails/app/controllers/competitions_controller.rb
DanielEgdal: Using `@competition_organizer_view` would result in the short name / id not automatically updating when a competition admin updates their competition name from the organiser view.

WcaOnRails/app/controllers/competitions_controller.rb
DanielEgdal: I tried modifying the function you are referencing at first, however when I asked on the Slack I was told I could not get any pass to the function whether or not I am on the admin view page. Thus I moved it into this file. I agree this is annoying how the function is written basically two times across files, but I do not know how I can avoid that (given the subsequent updates depend on `@competition_admin_view`) .

kr-matthews:
The delegate crash course mentions the "competition nickname" so that should be updated at the same time.

WcaOnRails/app/controllers/competitions_controller.rb
gregorbg: I don't think so: If a competition admin updates their own competition from the organizer view, then...
- `params.key?(:competition_admin_view)` is `false` because they are using the **organizer** view
- `current_user.can_admin_competitions?` is `true` because they are a competition admin
- `@competition_admin_view` evaluates to `false && true` which is clearly `false`
- `@competition_organizer_view` evaluates to `!false` which is `true`.

So even in the scenario that you described, the `@competition_organizer_view` variable should be `true`, and the update trigger should run. Otherwise, I have a flaw in my logic (which you're free to point out) or it's a bug in the code

DanielEgdal: Everything still works after your change, so I think I just misunderstood the original code those months back (your logic makes sense to me).

DanielEgdal:
All concerns here should be fixed now, thanks for the help Gregor. Hopefully merging should be possible now.

From Kevin's remark, I know WQAC will need to finish a revision on the crash course, so that can be fixed at that time.

gregorbg (approved):

PR #7897, opened 2023-05-12
Added newcomer verification in delegate report template

danieljames-dj:
This PR is as per the discussion between WRT & WQAC.


Roco033:
LGTM, hope this leads to less fixes after the competition.

EdHollingdale:
I don't think adding more detail is the direction we should be going with reports. If WRT want more Delegates to make sure there is an appropriate process for checking newcomer information I don't think having a section on it in reports is the most effective way.

danieljames-dj:
@EdHollingdale The reason why we are doing this is, WRT is receiving lots of requests to fix DOBs. Also, we find that many delegates are not validating DOB during registration. This change will allow WRT to understand how many competitions actually does the DOB validation during registration and can help resolve more data integrity problems which is one of the responsibility of WRT in the motion.

DanielEgdal:
I'd think the section should should be called something like 'registration and check in', since the idea from the thread was to know more about registration as a whole, as opposed to just topics related to newcomers (though that is one of the more significant parts). Other than that, the content itself should be fine (as it is exactly the same as on the thread. :P)

danieljames-dj:
Thanks @DanielEgdal. I've changed the section name as you suggested.

DanielEgdal:
Idk if I'm doing something wrong, but the PR looks like the exact same as when I looked yesterday (still says newcomer verification)

danieljames-dj:
Oops sorry @DanielEgdal, I somehow forgot to stage the changes before commit while doing it in rush. I've made the change now.

DanielEgdal (approved):

Roco033 (approved):

PR #7898, opened 2023-05-12
Mask Result upload PII on staging systems

gregorbg:
see title.

### DIFF
--- a/WcaOnRails/app/models/upload_json.rb
+++ b/WcaOnRails/app/models/upload_json.rb
@@ -42,6 +42,8 @@ class UploadJson
       persons_to_import = []
       parsed_json["persons"].each do |p|
         new_person_attributes = p.merge(competitionId: competition_id)
+        # mask uploaded DOB on staging to avoid accidentally importing PII
+        new_person_attributes["dob"] = "1954-12-04" if Rails.env.production? && !EnvVars::WCA_LIVE_SITE?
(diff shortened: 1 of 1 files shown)


danieljames-dj (approved):
LGTM

WcaOnRails/app/models/upload_json.rb
```diff
@@ -42,6 +42,8 @@ def import_to_inbox
       persons_to_import = []
       parsed_json["persons"].each do |p|
         new_person_attributes = p.merge(competitionId: competition_id)
+        # mask uploaded DOB on staging to avoid accidentally importing PII
+        new_person_attributes["dob"] = "1954-12-04" if Rails.env.production? && !EnvVars::WCA_LIVE_SITE?
```
danieljames-dj, line 46: I've a minor doubt - is staging running on production mode? I'm confused whether this will return true for normal prod website as well?

gregorbg: Staging is running using the Rails `production` configuration. This is normal and considered best practice by Rails themselves.

That's the reason why we have the additional `EnvVars::WCA_LIVE_SITE?` check, which (as the name suggests) only returns `true` for the real, public website. Note that this variable was introduced back when Cubecomps was still a thing, so it has nothing to do with our modern "WCA Live".

Issue #7875, opened 2023-05-07
Developer database export does not update


jonatanklosko:
At the moment (May 7th) the latest `wca-developer-database-dump.zip` is from April 11th, but the job seems to run as usually:

```
mysql> select * from timestamps where name = 'developer_db_dump';
+-------------------+---------------------+
| name              | date                |
+-------------------+---------------------+
| developer_db_dump | 2023-05-06 04:04:31 |
+-------------------+---------------------+
1 row in set (0.00 sec)
```

So likely there's an error during export, but it passes through silently.

OldManLink:
Can confirm. Steps to reproduce: 
* save attached script `wca_dump_date.sh` :
```
#!/bin/bash
DATABASE_EXPORT_URL="https://www.worldcubeassociation.org/wst/wca-developer-database-dump.zip"
ZIP_FILENAME=wca-developer-database-dump.zip
FILENAME=wca-developer-database-dump.sql
VERSION_FILENAME="wca-developer-database-dump.ver"
cd /tmp
wget --quiet $DATABASE_EXPORT_URL
unzip $ZIP_FILENAME
tail -1 $FILENAME > $VERSION_FILENAME
rm $FILENAME
rm $ZIP_FILENAME
cat $VERSION_FILENAME
```
* execute script
* expected result `-- Dump completed on 2023-05-07 04:04:31`
* actual result when executed just now:
````
09:33 $ ./wca_dump_date.sh 
Archive:  wca-developer-database-dump.zip
  inflating: wca-developer-database-dump.sql  
-- Dump completed on 2023-04-11 20:16:20
✔ ~/Downloads/temp/Cube 
09:34 $
````

gregorbg:
The export times out because our database has grown too large. The (terribly, terribly old!) PHP cron scripts are not tied into our usual error reporting, so nobody noticed. There is https://github.com/thewca/worldcubeassociation.org/pull/7642 which will replace the public export cron script by a proper Rails background job instead. Unfortunately, we won't fix this issue on the PHP side in the meantime, so please wait until the aforementioned PR is merged.

OldManLink:
Thanks for the explanation, @gregorbg. Looking at the PR I can see why it's taken a while to get it approved! Fingers crossed that it can be merged soon.

jonatanklosko:
@gregorbg #7642 is about the public export, right? It's actually the dev db dump that's out of date. It is handled by this background job:

[...]

PR #7899, opened 2023-05-13
Move dumping functions to common DbDumpHelper

gregorbg:
Because invoking Rake tasks from within the Rails application code ain't cool.

Essentially, everything that happened within the Rake task body is now moved to a separate module function. The Rake task directly calls that module's function, so literally nothing changed there.

The hourly cronjobs can also call the module under `libs`, instead of having to rely on Rake.


gregorbg:
Also, potential remedy for https://github.com/thewca/worldcubeassociation.org/issues/7875 (at least to a degree where we will get error messages again)

viroulep:
Thanks!

Issue #7850, opened 2023-04-23
Migrating Competitions Overview Page to React


notrichardpeng:
https://www.worldcubeassociation.org/competitions

PR #1206, opened 2017-01-30
Treat a competition as over if today is the last day of that competition.

jfly:
This fixes #737.

PR #7902, opened 2023-05-14
React Rewrite of Competition Overview Page

notrichardpeng:
#7850

Currently rewriting `_index_table.html.erb`, will see if it's feasible to also rewrite bigger pieces of code like `_index_competitions_list.html.erb` and `index_form.html.erb` without breaking this PR up.


WcaOnRails/yarn.lock
kr-matthews, line 1: Are these changes intentional?

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionTable.js
```diff
@@ -0,0 +1,41 @@
...
+  return (
+    <ul className="list-group">
+      <li className="list-group-item">
+        <strong>
+          {`${title} (${competitions.length})`}
+        </strong>
+      </li>
+      {competitions.map((comp, index) => (
+        <>
+          {(index > 0 && comp.year !== competitions[index - 1].year && !sortByAnnouncement) && <li className="list-group-item break">{comp.year}</li>}
```
kr-matthews, line 17: Extracting this condition to a helper function `isFirstCompOfAYear(competitions, index)` would make it easier to understand.

WcaOnRails/yarn.lock
notrichardpeng: no!

WcaOnRails/app/models/competition.rb
```diff
@@ -1908,6 +1908,7 @@ def serializable_hash(options = nil)
       resultsPosted: results_posted?,
       inProgress: in_progress?,
       dateRange: ApplicationController.helpers.wca_date_range(start_date, end_date),
+      dateCount: (is_probably_over? ? (Date.today - end_date) : (start_date - Date.today)).to_i, # Comp starts in X days / Comp ended X days ago. Calculating here because too nasty in ReactJS.
```
coder13, line 1911: Does this function / data also get used outside of your specific use-case (this PR)?
I'd actually rather we calculate this on the client. since this value is pretty specific.

notrichardpeng: I wasn't able to find any other uses of this specific value, but yeah doing it on the client is better for the server anyways so I'll try that.

notrichardpeng:
I believe the functionality of `_index_table.html.erb` is now fully implemented in `CompetitionTable.js`, feel free to provide feedback.

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionTable.js
```diff
@@ -0,0 +1,110 @@
+import React from 'react';
+import ReactMarkdown from 'react-markdown';
+
+import I18n from '../../lib/i18n';
+
+function calculateDate(comp, mode) {
```
kr-matthews, line 6: A better name might be `calculateDayDifference` or `getDayDifference` or something, since this isn't returning a date.

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionTable.js
```diff
@@ -0,0 +1,110 @@
...
+  const dateToday = new Date();
+  const startDate = new Date(comp.start_date);
+  const endDate = new Date(comp.end_date);
+
+  if (mode === 'future') {
+    const msDifference = startDate.getTime() - dateToday.getTime();
+    const dayDifference = Math.ceil(msDifference / (1000 * 3600 * 24));
```
kr-matthews, line 13: Could extract the denominator as a `const msInADay = ...`, here and below.

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionTable.js
```diff
@@ -0,0 +1,110 @@
...
+  if (mode === 'past') {
+    const msDifference = dateToday.getTime() - endDate.getTime();
+    const dayDifference = Math.floor(msDifference / (1000 * 3600 * 24));
+    return dayDifference;
+  }
+
+  return -1;
+}
+
+function renderYearHeader(competitions, index, sortByAnnouncement) {
```
kr-matthews, line 25: How come this is a regular function rather than a component `YearHeader({ competitions, index, ... })` called via `<YearHeader competitions={...} ... />`? When functions are components, react is able to optimize re-renders and stuff like that (as far as I understand).

WcaOnRails/app/views/competitions/_index_competitions_list.html.erb
```diff
@@ -7,30 +7,55 @@
     <% if @past_selected %>
       <div class="col-md-12" id="past-comps">
         <% table_title = year_is_a_number?(params[:year]) ? t('competitions.index.titles.past', year: params[:year]) : t('competitions.index.titles.past_all') %>
-        <%= render 'index_table', competitions: competitions, title: table_title %>
```
kr-matthews, line 10: Doesn't the old `index_table` need deleting?

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionTable.js
```diff
@@ -0,0 +1,110 @@
...
+    return <i className="icon hourglass start" data-toggle="tooltip" data-original-title={I18n.t('competitions.index.tooltips.hourglass.announced_on', { announcement_date: comp.announcedDate })} />;
+  }
+  if (showRegistrationStatus) {
+    return renderRegistrationStatus(comp);
+  }
+
+  return <i className="icon hourglass end" data-toggle="tooltip" data-original-title={I18n.t('competitions.index.tooltips.hourglass.starts_in', { days: I18n.t('common.days', { count: calculateDate(comp, 'future') }) })} />;
+}
+
+function CompetitionTable({
```
kr-matthews, line 68: I think it's standard to have the main component/function at the top, for quick access to see what this component is about and so that function dependencies point downward.

WcaOnRails/config/locales/vi.yml
kr-matthews, line 1: Are these `locales` files meant to have been changed?

notrichardpeng: They need to be there for the icon tooltip to work fully. Some locales without `days.one` couldn't display the fact that competitions are a day away.

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionTable.js
notrichardpeng: good point

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionTable.js
notrichardpeng: The real header itself is just one line of HTML. This is more of a function to check whether a header should be displayed at this point of the big `competitions.map()` loop (thus `competitions` and `index` are passed in). I struggled to find a good way to conditionally render the header during the loop so I used this function. Perhaps I should rename it. Would making it a component make it easier to be conditionally rendered?

WcaOnRails/app/views/competitions/_index_competitions_list.html.erb
notrichardpeng: yes!

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionTable.js
notrichardpeng: I can't pass the pre-commit hooks with this because ESLint isn't happy with it... `no-use-before-define`. Should I try circumventing this?

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionTable.js
kr-matthews: Generally, I guess there's not much of a practical difference between a react component and a regular function in this case, since the 'component' is so small/simple and there are no hooks or anything around ([stack overflow reference](https://stackoverflow.com/questions/71658569/react-whats-the-difference-between-a-function-returning-jsx-and-function-compo)). Having a function be a component when it's returning JSX seems like a good rule of thumb though.

I would just do something like
```
{shouldShowYearHeader(competitions, index, sortByAnnouncement) && <li className="list-group-item break">{competitions[index].year}</li>}
```
with `shouldShowYearHeader` defined as a boolean helper function. But I think that's more of a preference thing so I wouldn't worry about it too much. If we're keeping it as it currently is (which I think is fine) then `renderYearHeaderIfNecessary` or something to that effect would be a better name, since this function isn't actually always rendering the year header.

kr-matthews (changes requested):
It's probably just my bad laptop, but it takes forever to load the list view of all competitions (but map and admin views are relatively quick). Is it a noticeably different speed compared to the original page on your machine?
I'm seeing `Warning: Each child in a list should have a unique "key" prop.`, double-check every `map` which returns JSX has a key for each child (and don't use the index if possible).

kr-matthews:
I also have this header bug, which shows up when running your branch but not when viewing the real site. Not sure if it's related to anything you did.
image

kr-matthews:
There are different types of icons and I think you missed at least one (the sand timer with only the top half full). Screenshot from real site:
image

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionTable.js
notrichardpeng: I think the boolean helper function approach looks cleaner so I'm going ahead with that. It's also better practice it seems. Thanks for the suggestion.

notrichardpeng:
Having issues with shakapacker compiling so I can't even see what my code renders now... I should probably spend time setting up docker first

notrichardpeng:
Ready for another review! For the speed issue map view is also faster than the list view on my end, and the list does take a few seconds to load. I wonder if it's just because of webpacker being slow or are there efficiency issues elsewhere?

kr-matthews (approved):
Yeah the speed thing apparently might be an unrelated issue with staging/local.
LGTM but I'd feel better if someone else also approved before merging.

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionTable.js
```diff
@@ -0,0 +1,108 @@
+import React from 'react';
+import ReactMarkdown from 'react-markdown';
+
+import I18n from '../../lib/i18n';
+
+function calculateDayDifference(comp, mode) {
```
coder13, line 6: Would prefer this move to `utils/`

notrichardpeng: so generalizing this function and moving it into one of the existing `utils/` js files?

notrichardpeng:
Got it, I'll check out the document, thanks!

gregorbg:
Adding to Duncan's comment, I would love to see a full rewrite that includes the "filtering and sorting box" at the top of the page. In its current form, it's very ugly passing down a ton of parameters to URL query params, parsing them, and then rendering the "same but different" table in different modes.

This all seems quite awkward and I think it could be simplified _a lot_ if it is managed through React `useState` components instead. I do realise however that this is a hefty amount of work and so I am opening discussion rather than giving binding instructions.

The main point why I'm mentioning is this: In its current form, it might not "make sense" why you should fetch competition data from an extra, separate API route (other than "because our conventions say so"). As soon as you introduce the filters as a React-native component though, it suddenly becomes very evident why you _must_ load them dynamically through a separate frontend route: Because the data of _which_ competitions to load will change so frequently and dynamically that you have to be able to adapt and adjust.

notrichardpeng:
I think this is a good opportunity to spend time learning more about React and APIs, so I'll try to follow through with adding the filtering menu. I'll reach out when I encounter obstacles 👀

notrichardpeng:
At this point I believe all of the functionalities of the competitions overview page has been implemented in this PR, would be great if someone would like to test it out or review the code! The only big things I can think of that is not yet included is the admin view (could/should it be done in this PR too?) and rewriting some tests if necessary. There's also small things like the Semantic UI `Loader` not displaying at all and not yet deleting all the old Rails views code.

Key things that need feedback on: 
- Whether or not the big `CompetitionFilters` file should be broken into smaller chunks and ways of doing that
- Are data filtered in an efficient manner? Some data filtering is done in the front end (events, show cancelled comps) and others are done in the back end by the API (delegates, search queries), is this appropriate?
- Whether `react-query` is properly used
- Whether good React conventions are followed

If there are other concerns please let me know, thanks!

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionTable.js
```diff
@@ -0,0 +1,190 @@
...
+}
+
+function CompetitionTable({
+  competitionData,
+  title,
+  showRegistrationStatus,
+  showCancelled,
+  selectedEvents,
+  loading,
+  loaded,
```
kr-matthews, line 126: I haven't looked at where this component is being called yet, but components should make sense in isolation and I'm a bit confused why there's `loading` and `loaded` - and why the main data `competitionData` seems to be presented regardless of the `load...` values.

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionTable.js
```diff
@@ -0,0 +1,190 @@
...
+      content={tooltipInfo}
+      position="top center"
+      size="tiny"
+    />
+  );
+}
+
+// Currently, the venue attribute of a competition object can be written as markdown,
+// and using third party libraries like react-markdown to parse it requires to much work
+function VenueMarkdown({ venueText }) {
```
kr-matthews, line 93: Function name should make it clear that it's not real markdown. And it's generic, not venue-related, so no need to mention 'venue' anywhere.
```suggestion
function PseudoMarkdown({ text }) {
```

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionTable.js
```diff
@@ -0,0 +1,190 @@
...
+      trigger={<Icon className={iconClass} />}
+      content={tooltipInfo}
+      position="top center"
+      size="tiny"
+    />
+  );
+}
+
+// Currently, the venue attribute of a competition object can be written as markdown,
+// and using third party libraries like react-markdown to parse it requires to much work
```
kr-matthews, line 92:
```suggestion
// and using third party libraries like react-markdown to parse it requires too much work
```

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionTable.js
```diff
@@ -0,0 +1,190 @@
...
+    <Popup
+      trigger={<Icon className="user plus green" />}
+      content={I18n.t('competitions.index.tooltips.registration.open')}
+      position="top center"
+      size="tiny"
+    />
+  );
+}
+
+function DateIcon({ comp, showRegistrationStatus, sortByAnnouncement }) {
```
kr-matthews, line 56:
```suggestion
function StatusIcon({ comp, showRegistrationStatus, sortByAnnouncement }) {
```

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionTable.js
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+  return (
+    <a href={venueText.slice(openParenIndex + 1, closeParenIndex)} target="_blank" rel="noreferrer">
+      <p>{venueText.slice(openBracketIndex + 1, closeBracketIndex)}</p>
+    </a>
+  );
+}
+
+function LoadedDisplay({ numCompetitions }) {
```
kr-matthews, line 111:
```suggestion
function EndOfCompListMessage({ numCompetitions }) {
```
or something would be better - it's usually best to describe what the component is doing rather than when/where it's used.

When I first saw this being called in `CompetitionTable` I thought it must be displaying data (because of the name, and since it's only called when `loaded` is true), which was confusing.

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionFilters.js
```diff
@@ -0,0 +1,662 @@
...
+          }
+        </div>
+      </Container>
+
+      {!competitionsIsFetching && hasUnloadedCompetitions && displayMode === 'list' && <div ref={ref} name="page-bottom" />}
+    </Container>
+  );
+}
+
+export default CompetitionFilter;
```
kr-matthews, line 662: Default export should match the file name (file is plural, function is singular).

WcaOnRails/app/models/competition.rb
```diff
@@ -910,6 +910,17 @@ def can_edit_registration_fees?
...
 
+  def registration_status
+    if registration_not_yet_opened?
+      return "not_yet_opened"
+    elsif registration_past?
+      return "past"
+    elsif registration_full?
+      return "full"
+    end
+    "open"
```
danieljames-dj, line 921: If this is moved to `else`, won't you be able to avoid the `return` in the previous if & elsif? (not 100% sure, please check it by running the code)

WcaOnRails/app/models/competition.rb
```diff
@@ -1561,6 +1572,15 @@ def self.search(query, params: {}, managed_by_user: nil)
       competitions = Competition.visible
     end
 
+    if params[:continent].present?
+      continent = Continent.find(params[:continent])
```
danieljames-dj, line 1576: Maybe `params.require(:continent)`?

WcaOnRails/app/models/competition.rb
```diff
@@ -1561,6 +1572,15 @@ def self.search(query, params: {}, managed_by_user: nil)
...
+      if !continent
```
danieljames-dj, line 1577: `if !x` can always be replaced with `unless x`.
That being said, you can also make it single line like:
`raise WcaExceptions::BadApiParameter.new("Invalid continent: '#{params[:continent]}'") unless continent.nil?`.

Also, if you do `Continent.find!` instead of `Continent.find`, then it will automatically raise exception and you don't have to do it explicitly.

WcaOnRails/app/models/competition.rb
```diff
@@ -1569,6 +1589,15 @@ def self.search(query, params: {}, managed_by_user: nil)
       competitions = competitions.where(countryId: country.id)
     end
 
+    if params[:delegate].present?
+      delegate = User.find_by(wca_id: params[:delegate])
```
danieljames-dj, line 1593: Same comments as in continent.

WcaOnRails/app/models/competition.rb
```diff
@@ -1585,6 +1614,14 @@ def self.search(query, params: {}, managed_by_user: nil)
       competitions = competitions.where("end_date <= ?", end_date)
     end
 
+    if params[:ongoing_and_future].present?
+      target_date = Date.safe_parse(params[:ongoing_and_future])
```
danieljames-dj, line 1618: Same comments as in continent.

WcaOnRails/app/models/competition.rb
```diff
@@ -1957,6 +1994,21 @@ def serializable_hash(options = nil)
     # we want to change the existing behavior of our API which returns a string.
     json.merge!(
       class: self.class.to_s.downcase,
+      displayName: display_name(short: true),
+      countryName: country ? country.name : nil,
```
danieljames-dj, line 1998: Nit: `country&.name`.

Same for other ternary operators in this block.

WcaOnRails/app/webpacker/lib/requests/routes.js.erb
```diff
@@ -34,6 +34,8 @@ export const competitionReportUrl = (id) => `<%= CGI.unescape(Rails.application.
 
 export const competitionApiUrl = (id) => `<%= CGI.unescape(Rails.application.routes.url_helpers.api_v0_competition_path("${id}"))%>`;
 
+export const competitionsApiUrl = `<%= CGI.unescape(Rails.application.routes.url_helpers.api_v0_competitions_path)%>`;
```
danieljames-dj, line 37: Maybe you can initialize this inside `apiV0Urls` which is there at bottom of this file.

WcaOnRails/app/webpacker/lib/requests/routes.js.erb
```diff
@@ -88,6 +90,8 @@ export const adminImportResultsUrl = (competitionId) => `<%= CGI.unescape(Rails.
 
 export const adminCheckUploadedResults = (competitionId) => `<%= CGI.unescape(Rails.application.routes.url_helpers.competition_admin_upload_results_edit_path("${competitionId}"))%>`;
 
+export const delegatesApiUrl = `<%= CGI.unescape(Rails.application.routes.url_helpers.api_v0_delegates_path)%>`;
```
danieljames-dj, line 93: Is this used anywhere? I'm not sure if this is one of the URL that I've removed.

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionMap.js
```diff
@@ -0,0 +1,47 @@
...
+
+import { userTileProvider } from '../../lib/leaflet-wca/providers';
+import { redMarker, blueMarker } from '../../lib/leaflet-wca/markers';
+import ResizeMapIFrame from '../../lib/utils/leaflet-iframe';
+import 'leaflet/dist/leaflet.css';
+
+function CompetitionMap({
+  competitionData,
+  selectedEvents,
+  showCancelled,
```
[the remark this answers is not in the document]
coder13: +1 I think this is a good code pattern for a long lasting codebase.

WcaOnRails/app/webpacker/lib/utils/competition-table.js
```diff
@@ -0,0 +1,19 @@
+export default function calculateDayDifference(startDateString, endDateString, mode) {
+  const dateToday = new Date();
+  const startDate = new Date(startDateString);
+  const endDate = new Date(endDateString);
+  const msInADay = 1000 * 3600 * 24;
+
+  if (mode === 'future') {
+    const msDifference = startDate.getTime() - dateToday.getTime();
+    const dayDifference = Math.ceil(msDifference / msInADay);
```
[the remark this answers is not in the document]
coder13: If this code works on it's own, I wouldn't recommend an extra dependency here. 
If we need to start doing timezone related stuff, I can see how we can benefit greatly from someone else doing the work however.

Recognizing that `startDateString` is likely a data with format 'yyyy-mm-dd', I already anticipate we'll get weird timezone stuff.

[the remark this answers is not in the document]
notrichardpeng: The parameters are indeed in the format 'yyyy-mm-dd' as they are used for `comp.start_date` and `comp.end_date`, but currently I don't think there are timezone issues with competition dates as of now...

Is `luxon` already part of our dependencies though? I heard there were mentions of it already being used elsewhere.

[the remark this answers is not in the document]
coder13: The common date issues is when you call `new Date()` on a particular date, the default time gets set to 00:00.00 in UTC, say you're in the pacific timezone, that date will be converted to 2pm in the evening the day before. When you calculate the days till, it'll be 1 less than expected.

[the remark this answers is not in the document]
notrichardpeng: Ah yes this does happen. If I remember correctly the official competitions page is also off by a little bit when I checked if it was consistent with my local server so I thought it behaved fine and everything was supposed to be in UTC. I might look into `luxon` then.

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionTable.js
notrichardpeng: Yep you're right the problem was the function calls. I turned things into components and this worked.

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionTable.js
notrichardpeng: True, I will say this is still specific to only parsing link markdown `[]()`. It probably makes more sense to also move this to `lib/utils`.

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionTable.js
notrichardpeng: helpful suggestions, didn't see them at first so I can no longer commit them here but I added them now.

WcaOnRails/app/models/competition.rb
notrichardpeng: Yep you're right, that would work and be more concise.

WcaOnRails/app/models/competition.rb
notrichardpeng: The existing code for other parameters follow a similar syntax of checking `if !x` and then `raise`, should I just simplify all of them or keep them as is and follow the convention myself?

WcaOnRails/app/models/competition.rb
notrichardpeng: Hm is this necessary even after ensuring `params[:continent].present?`

WcaOnRails/app/webpacker/lib/requests/routes.js.erb
notrichardpeng: Yes, this is used for loading the list of delegates for the delegate selection filtering in the overview page.

WcaOnRails/app/webpacker/lib/requests/routes.js.erb
notrichardpeng: Oh my, my bad, my local branch has not fetched from main for so long that that part is entirely not there. I see what you mean. Yeah it would make sense to put that there. But why aren't all the api urls in that object? Is that something to be done?

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionFilters.js
```diff
@@ -0,0 +1,662 @@
...
+              placeholder={I18n.t('competitions.index.tooltips.search')}
+              onChange={(event, data) => editSearchQuery(data.value)}
+            />
+          </Form.Field>
+        </Form.Group>
+
+        <Form.Group>
+          <Form.Field width={8}>
+            <label htmlFor="delegate">{I18n.t('layouts.navigation.delegate')}</label>
+            <Dropdown
```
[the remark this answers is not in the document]
notrichardpeng: I was entirely unaware of the component `<WcaSearch />`, that would have been really helpful. But the dropdown does look clean and it enables search! Here's a demo. However loading up all the delegates in the initialization of the page costs a bit computationally. It might be bias but I like this style of dropdown better than the original search, please let me know what your thoughts are. Perhaps I should post this in Slack to gather more opinion. But if there's a clear advantage/purpose to sticking with `<WcaSearch />` please do not feel bad to turn this idea down.

https://github.com/thewca/worldcubeassociation.org/assets/58278719/c695aace-21d7-4881-bcfd-acb3d52c63e4

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionFilters.js
```diff
@@ -0,0 +1,662 @@
...
+
+  const editPastComps = (comps, year) => {
+    setPastComps((prevPastComps) => ({
+      ...prevPastComps,
+      [year]: comps,
+    }));
+  };
+
+  const {
+    data: competitionsData,
```
kr-matthews, line 158: How does infinite query work - `data` has a `pages` property, which is an array of competition data, and fetching the next page pushes to the end of `data.pages`?

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionFilters.js
```diff
@@ -0,0 +1,662 @@
...
+              style={{ textAlign: 'center' }}
+              options={[{ key: 'None', text: I18n.t('competitions.index.no_delegates'), value: 'None' }, ...(delegatesInfo?.filter((item) => item.name !== 'WCA Board').map((delegate) => (
+                {
+                  key: delegate.id,
+                  text: `${delegate.name} (${delegate.wca_id})`,
+                  value: delegate.wca_id,
+                  image: { avatar: true, src: delegate.avatar?.thumb_url },
+                }
+              )) || [])]}
+              onChange={(event, data) => editSelectedDelegate(data.value)}
```
kr-matthews, line 400:
```suggestion
              onChange={(_, data) => editSelectedDelegate(data.value)}
```

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionFilters.js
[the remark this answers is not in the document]
danieljames-dj: I too like the Dropdown design more. But having two different components in the codebase doing the same thing doesn't look so good because it will be like reinventing the same wheel.

Is the UI design the only difference? If yes, is it possible to change the UI of WCA Search to the way you did in Dropdown? WcaSearch is also using Dropdown internally, so I guess implementing the design in WcaSearch would be possible. You can post this in Slack as well to gather more opinion.

WcaOnRails/app/webpacker/lib/utils/competition-table.js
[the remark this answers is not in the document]
kr-matthews: > The common date issues is when you call new Date() on a particular date, the default time gets set to 00:00.00 in UTC

Doesn't it get set to midnight in your local time zone?

I'm pretty sure Gregor was already using luxon in the codebase somewhere (venue/schedule editor?) so I'd recommend it.

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionFilters.js
```diff
@@ -0,0 +1,662 @@
...
+      id="custom"
+      value="custom"
+      onClick={() => editTimeOrder('custom')}
+      active={timeOrder === 'custom'}
+    >
+      <span className="caption">{I18n.t('competitions.index.custom')}</span>
+    </Button>
+  );
+
+  return (
```
kr-matthews, line 322: So this component is called `CompetitionFilters`, but it renders both the filters and the resulting data, which is confusing.

You can refactor out the filters and the actual data into separate components. And then the filters themselves can also be refactored into separate components per filter. All that will drastically simplify the return body here. It might seem like a lot of prop-drilling will happen (deep chains of components that are just passing props along without using them) but I don't think that will be the case because of the `filterState` object I suggested (switching to a reducer) and the single `reduceFilter` function which replaces all the `edit...` functions.

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionFilters.js
```diff
@@ -0,0 +1,662 @@
...
+    } else {
+      setCompetitionApiKey((prevKey) => ({ ...prevKey, sort_by: order, year: '' }));
+    }
+
+    setTimeOrder(order);
+  };
+
+  const [inProgressComps, setInProgressComps] = useState([]);
+  const [notInProgressFutureComps, setNotInProgressFutureComps] = useState([]);
+  const [recentComps, setRecentComps] = useState([]);
```
kr-matthews, line 144: Are all these states necessary? Can't they be derived from the `competitionsData` object? Less state is usually better.

kr-matthews (changes requested):

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionFilters.js
```diff
@@ -0,0 +1,662 @@
...
+function CompetitionFilters() {
+  const [competitionApiKey, setCompetitionApiKey] = useState({
+    sort_by: 'present', year: '', region: '', delegate: '', search: '',
+  });
+  const [selectedEvents, setSelectedEvents] = useState([]);
+  const [selectedRegion, setSelectedRegion] = useState();
+  const [selectedDelegate, setSelectedDelegate] = useState();
+  const [pastSelectedYear, setPastSelectedYear] = useState('all_years');
+  const [customStartDate, setCustomStartDate] = useState();
+  const [customEndDate, setCustomEndDate] = useState();
```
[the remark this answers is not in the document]
notrichardpeng: Due to being so new to React I wasn't even aware `useReducer` was a thing... it is indeed so much better than using this many states. What's even better is that the state generated by `useReducer` can be used as the query key straight away.

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionFilters.js
```diff
@@ -0,0 +1,662 @@
...
+    }));
+  };
+
+  const {
+    data: competitionsData,
+    fetchNextPage: competitionsFetchNextPage,
+    isFetching: competitionsIsFetching,
+    hasNextPage: hasMoreCompsToLoad,
+  } = useInfiniteQuery({
+    queryKey: ['competitions', competitionApiKey],
```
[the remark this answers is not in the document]
notrichardpeng: Based on testing I think it is smart enough to figure out what to do when an object is passed in and caches the data correctly. The official [documentation](https://tanstack.com/query/v4/docs/react/guides/query-keys) also shows that objects are supported as query keys.

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionFilters.js
notrichardpeng: Sounds correct

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionFilters.js
```diff
@@ -0,0 +1,662 @@
...
+    queryFn: ({ pageParam = 1 }) => {
+      const dateNow = new Date(Date.now());
```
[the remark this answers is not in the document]
notrichardpeng: I think both give the same result in UTC. I will get rid of `Date.now()` tho for consistency.

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionFilters.js
```diff
@@ -0,0 +1,662 @@
...
+        searchParams.append('delegate', selectedDelegate);
+      }
+      if (searchQuery) {
+        searchParams.append('q', searchQuery);
+      }
+
+      return fetchJsonOrError(`${competitionsApiUrl}?${searchParams.toString()}`);
+    },
+    getNextPageParam: (lastPage, allPages) => {
+      // Assuming the last page has less than the max number of competitions fetched per query
```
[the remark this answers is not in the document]
notrichardpeng: The latter

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionFilters.js
```diff
@@ -0,0 +1,662 @@
...
+      }
+
+      return allPages.length + 1;
+    },
+  });
+
+  useEffect(() => {
+    const flatData = competitionsData?.pages
+      .map((page) => page.data)
+      .flatMap((comp) => comp);
```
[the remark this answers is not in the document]
notrichardpeng: `page.data` is an array of 25 competition objects, but yeah `.flat()` would be more concise.

WcaOnRails/app/webpacker/lib/requests/routes.js.erb
danieljames-dj: Actually I created that object recently because it is easy to understand the purpose from the path. Changing all existing APIs under this is really difficult. So, maybe we can add the new ones here and eventually change the old ones when we change something in them. Also, in long term, I don't know whether we even have this `routes.js.erb`, maybe we will find some other routing way.

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionFilters.js
notrichardpeng: They're probably not necessary. I initially only created `inProgressComps` and `notInProgressFutureComps` just to separate the two and render the two tables for the "present" tab, but it snowballed into having one for every tab. This can most likely be replaced by some more filtering logic in `CompetitionTable.js`.

notrichardpeng: Ah the other thing is I thought switching tabs would look smoother if the comps were stored somewhere and pulled up. After testing how it is with all the `CompetitionTable`s sharing the same state for comps one thing I noticed is that the comps for the previous tab would still be displayed before the comps for the new tab are fetched. This means that the user would not see an empty list with "Loading..." when switching to a new tab, but rather the previous comps lagging behind and then suddenly replaced by the new comps. I wonder if there's some way to circumvent this while keeping the code efficient.

notrichardpeng: > This means that the user would not see an empty list with "Loading..." when switching to a new tab

Ok I think this can be easily solved by temporarily setting the shared state to an empty array whenever `timeOrder` is changed by the reducer.

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionFilters.js
```diff
@@ -0,0 +1,591 @@
...
+  const editSelectedEvents = (eventId) => {
+    setSelectedEvents((prevSelectedEvents) => (
+      prevSelectedEvents.includes(eventId)
+        ? prevSelectedEvents.filter((id) => id !== eventId)
+        : [...prevSelectedEvents, eventId]
+    ));
+  };
+
+  const filterReducer = (state, action) => {
+    const newState = { ...state, ...action };
```
kr-matthews, line 97: Typically the reducer `action` is an object with a `type` property along with whatever data is required to perform that action. Such as `{ type: 'update-region', region: [...] }`, `{ type: 'search', text: [...] }`, or `{ type: 'reset' }`. I'm not sure it's necessarily wrong to just pass in the partial object and merge that into the existing object though.

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionFilters.js
```diff
@@ -0,0 +1,591 @@
...
+
+function CompetitionFilters() {
+  const [selectedEvents, setSelectedEvents] = useState([]);
+  const [shouldShowRegStatus, setShouldShowRegStatus] = useState(false);
+  const [shouldShowCancelled, setShouldShowCancelled] = useState(false);
+  const [displayMode, setDisplayMode] = useState('list');
+  const [competitionData, setCompetitionData] = useState([]);
+  const [competitionQueryKey, setCompetitionQueryKey] = useState({
+    timeOrder: 'present', region: 'all_regions', delegate: '', search: '', time: '',
+  });
```
kr-matthews, line 86: If you have two states where one can be calculated from the other, then that one should (typically) not be a state.

This `useState` can be replaced with something like
```
const competitionQueryKey = calculateQueryKey(filterState)
```
or maybe
```
const competitionQueryKey = useMemo(() => calculateQueryKey(filterState), [filterState])
```

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionFilters.js
```diff
@@ -0,0 +1,591 @@
...
+    setSelectedEvents((prevSelectedEvents) => (
+      prevSelectedEvents.includes(eventId)
+        ? prevSelectedEvents.filter((id) => id !== eventId)
+        : [...prevSelectedEvents, eventId]
+    ));
+  };
+
+  const filterReducer = (state, action) => {
+    const newState = { ...state, ...action };
+    setCompetitionQueryKey(calculateQueryKey(newState));
```
kr-matthews, line 98: This won't be necessary once `competitionQueryKey` is calculated directly.

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionFilters.js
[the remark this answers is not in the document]
kr-matthews: So the `lastPage` param here is the most recent page fetched, not the actual last page of data? I see that the infinite query documentation uses `lastPage`, but `previousPage` seems more logical to me, especially since our infinite scroll always has a last page.

I don't know if it's just me, but if a comment says that it's assuming something, I assume the code may potentially break or behave unexpectedly if the assumption fails. There's no reason that the back-end would send less than a full page unless it's the end of the results right?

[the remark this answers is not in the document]
kr-matthews: I'd say it's clearer to frame this as "if we get a full page of results back, then there may be another page" rather than "if we don't get a full page back then we assume there are no more pages".

Minor difference, but it might help whoever is modifying this code a few years from now.

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionFilters.js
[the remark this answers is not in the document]
kr-matthews: Given that it's called `filterState` it might make sense to include _all_ these filter-based states in it (that is, selected events, should show cancelled, etc.). The other benefit of doing that is that right now the reducer vs individual states are split between those which are handled by the back-end vs the front-end which means that if one of those gets handled differently later then the reducer would have to be modified. If everything's in the reducer from the start then nothing would have to change.

That last point is a bit hypothetical though and it's not always a good use of time to prepare for things that might not happen.

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionFilters.js
```diff
@@ -0,0 +1,591 @@
...
+    delegate: filterState?.delegate,
+    search: filterState?.search,
+    time: timeKey,
+  };
+};
+
+function CompetitionFilters() {
+  const [selectedEvents, setSelectedEvents] = useState([]);
+  const [shouldShowRegStatus, setShouldShowRegStatus] = useState(false);
+  const [shouldShowCancelled, setShouldShowCancelled] = useState(false);
```
kr-matthews, line 81: To differentiate the fact that should _show_ registration status is about showing additional information vs should _include_ cancelled is about the filter logic.
```suggestion
  const [shouldIncludeCancelled, setShouldIncludeCancelled] = useState(false);
```

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionFilters.js
```diff
@@ -0,0 +1,591 @@
...
+              style={{ textAlign: 'center' }}
+              options={[{ key: 'None', text: I18n.t('competitions.index.no_delegates'), value: '' }, ...(delegatesInfo?.filter((item) => item.name !== 'WCA Board').map((delegate) => (
+                {
+                  key: delegate.id,
+                  text: `${delegate.name} (${delegate.wca_id})`,
+                  value: delegate.wca_id,
+                  image: { avatar: true, src: delegate.avatar?.thumb_url },
+                }
+              )) || [])]}
+              onChange={(event, data) => dispatchFilter({ delegate: data.value })}
```
kr-matthews, line 327:
```suggestion
              onChange={(_, data) => dispatchFilter({ delegate: data.value })}
```

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionFilters.js
kr-matthews: I haven't looked into this much since my last comment, but I think the query has a method to invalidate its data, which could be called whenever the filters update. I'm not sure that's a good solution, but it's something to consider if it works.

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionFilters.js
notrichardpeng: I looked up examples of `useReducer` and was going to do that, but I was unsure if that would be redundant.

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionFilters.js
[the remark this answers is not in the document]
notrichardpeng: Now that I'm no longer using the entire `filterState` object as the query key I agree this should be done. It would previously have been a problem since changing the selected events would generate a new query key whereas filtering by events is not done by the API. 

I talked with Gregor about filtering in the front-end vs the back-end and I think eventually we will move everything to the back-end (not necessarily in this PR, and maybe except showing cancelled competitions?).

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionFilters.js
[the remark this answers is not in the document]
notrichardpeng: Sounds good, I see where you're coming from

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionFilters.js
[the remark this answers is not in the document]
notrichardpeng: Should `filterReducer`, `filterInitialState`, and `calculateQueryKey` be separated into another file?

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionFilters.js
kr-matthews: Nice, looking a lot better now. The return statement is still quite long (over 200 lines) so breaking out a separate component for each filter UI element might be worthwhile (up to you).

WcaOnRails/app/webpacker/components/CompetitionOverview/useDelegatesData.js
```diff
@@ -0,0 +1,41 @@
+import { useEffect, useState } from 'react';
+import { useInfiniteQuery } from '@tanstack/react-query';
+
+import { delegatesApiUrl, WCA_API_PAGINATION } from '../../lib/requests/routes.js.erb';
+import { fetchJsonOrError } from '../../lib/requests/fetchWithAuthenticityToken';
+
+const useDelegatesData = () => {
+  const [delegatesData, setDelegatesData] = useState([]);
```
kr-matthews, line 8: Would it work to just define
```
const delegatesData = rawDelegatesData?.pages?.flatMap((page) => page.data)
```
(possibly with `useMemo`), or is the effect & state necessary?

WcaOnRails/app/webpacker/components/CompetitionOverview/useDelegatesData.js
```diff
@@ -0,0 +1,41 @@
...
+
+  const {
+    data: rawDelegatesData,
+    fetchNextPage: delegateFetchNextPage,
+    hasNextPage: delegateHasNextPage,
```
kr-matthews, line 13: Since this is isolated inside the `useDelegatesData` hook, it's kind of implicit that everything is a delegate, and there are no other queries to potential mix up, so you could just use the original names `data`, `fetchNextPage`, and `hasNextPage` to (slightly) shorten some code.

WcaOnRails/app/webpacker/components/CompetitionOverview/ManageFilterState.js
kr-matthews, line 1: This file isn't exporting a component so shouldn't be capitalized. I think `filterUtils.js` or something would be a reasonable name.

WcaOnRails/app/webpacker/components/CompetitionOverview/QueryHelper.js
```diff
@@ -0,0 +1,69 @@
...
+  return {
+    timeOrder: filterState?.timeOrder,
+    region: filterState?.region,
+    delegate: filterState?.delegate,
+    search: filterState?.search,
+    time: timeKey,
+  };
+}
+
+export function createSearchParams(filterState, pageParam) {
```
kr-matthews, line 20: Suggestion: you could get rid of all the `filterState.`s by adding `const { region, delegate, search, [...etc] } = filterState` at the top of this function. It's just up to personal preference though.
ie the first `if` below becomes `if (region && region !== 'all_regions')`

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionFilters.js
notrichardpeng: Yep, was going to, cleaned up now!

WcaOnRails/app/webpacker/components/CompetitionOverview/useDelegatesData.js
notrichardpeng: The state is not necessary! The effect needs to stay to make it keep fetching as long as new data is coming in (unless there's some other way to do a while loop here?). Also now I realize what you meant with the `flatMap` thing haha.

notrichardpeng: This means it also isn't necessary to do `useState` for competitionData, hm.

notrichardpeng: And now without the `useState` but with a `const` switching between tabs look noticably smoother too. Experiencing firsthand not to overuse states 💀

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionFilters.js
```diff
@@ -109,71 +95,13 @@ function CompetitionFilters() {
...
-      }
-      if (filterState.delegate) {
-        searchParams.append('delegate', filterState.delegate);
-      }
-      if (filterState.search) {
-        searchParams.append('q', filterState.search);
-      }
-
+      const searchParams = createSearchParams(filterState, pageParam);
       return fetchJsonOrError(`${competitionsApiUrl}?${searchParams.toString()}`);
```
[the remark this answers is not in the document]
notrichardpeng: You're right, thought it was necessary to explicitly convert `URLSearchParams` objects to string

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionFilters.js
```diff
@@ -56,50 +59,33 @@ const filterInitialState = {
   region: 'all_regions',
   delegate: '',
   search: '',
+  selectedEvents: [],
+  shouldShowRegStatus: false,
```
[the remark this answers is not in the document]
notrichardpeng: Having to pass in four more arguments into `CompetitionFilters` (the two states and setStates) is kind of painful, but this makes sense.

WcaOnRails/app/webpacker/lib/utils/competition-table.js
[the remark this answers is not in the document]
notrichardpeng: switched everything to luxon!

WcaOnRails/app/webpacker/lib/requests/routes.js.erb
notrichardpeng: done. not sure if calling these URLs 'list' is appropriate

WcaOnRails/app/models/competition.rb
notrichardpeng: I don't see a clean way of doing it this way if we want to keep the exceptions for each parameter..

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionView.js
```diff
@@ -0,0 +1,107 @@
...
+      competitionsFetchNextPage();
+    }
+  }, [bottomInView, competitionsFetchNextPage]);
+
+  useEffect(() => {
+    if (hasMoreCompsToLoad && displayMode === 'map' && competitionData?.length < MAP_DISPLAY_LIMIT) {
+      competitionsFetchNextPage();
+    }
+  }, [rawCompetitionData, displayMode, hasMoreCompsToLoad, competitionData,
+    competitionsFetchNextPage]);
```
kr-matthews, line 58: Since this only applies to the map, it would make sense to put it in the `CompetitionMap` component.

Also, `rawCompetitionData` isn't used.

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionView.js
```diff
@@ -0,0 +1,107 @@
...
+    },
+  });
+  const competitionData = rawCompetitionData?.pages.flatMap((page) => page.data);
+
+  const { ref: bottomRef, inView: bottomInView } = useInView();
+  useEffect(() => {
+    if (bottomInView) {
+      competitionsFetchNextPage();
+    }
+  }, [bottomInView, competitionsFetchNextPage]);
```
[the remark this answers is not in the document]
kr-matthews: As a follow to my comment on the other effect below, I think it would make sense to put this effect (and the ref) in the `CompetitionList` component.

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionView.js
```diff
@@ -0,0 +1,107 @@
...
+        filterState={filterState}
+        dispatchFilter={dispatchFilter}
+        displayMode={displayMode}
+        setDisplayMode={setDisplayMode}
+        shouldShowRegStatus={shouldShowRegStatus}
+        setShouldShowRegStatus={setShouldShowRegStatus}
+      />
+
+      <Container id="search-results" className="row competitions-list">
+        <div id="competitions-list">
```
kr-matthews, line 73: Should this div be _inside_ the same parens as `CompetitionList`? Or does it need to be rendered even when the display mode is map.

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionView.js
```diff
@@ -0,0 +1,107 @@
...
+                filterState={filterState}
+                shouldShowRegStatus={shouldShowRegStatus}
+                isLoading={competitionsIsFetching}
+                hasMoreCompsToLoad={hasMoreCompsToLoad}
+              />
+            )
+          }
+        </div>
+        {/* Old JS code does a lot of things to id=comeptitions-map, to be included? */}
+        <div name="competitions-map">
```
kr-matthews, line 88: Likewise.

kr-matthews (commented):
Nice, this is looking really good. I still need to take another pass through `CompetitionList`, `CompetitionList`, and `CompetitionTable` next time.

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionFilters.js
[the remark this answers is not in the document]
kr-matthews: Yeah, separating them out feels _technically_ more 'correct', but that's not always the best way to go. :shrug:

WcaOnRails/app/webpacker/lib/utils/competition-table.js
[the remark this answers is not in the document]
notrichardpeng: But the previous /lib/utils files are spaced with dash 🤔

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionView.js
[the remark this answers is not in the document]
notrichardpeng: since the ref will not be rendered (so bottomInView will always be false) if there are no more comps to load we won't need to check that

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionView.js
notrichardpeng: turns out the official page only renders the div if it's selected

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionTable.js
```diff
@@ -0,0 +1,178 @@
...
+            </Loader>
+          </List.Item>
+        )
+      }
+      {
+        !hasMoreCompsToLoad
+        && !isLoading
+        && !isRenderedAboveAnotherTable
+        && <EndOfCompListMessage numCompetitions={competitions?.length} />
+      }
```
kr-matthews, line 81: It seems to me like there should be a single footer or end-of-list component that's responsible for showing the message, the bottom-ref div, or nothing. (It could also include the loading spinner maybe?) Rather than having this component (sometimes) show the message, and the parent (sometimes) show the ref.

Does this message need to be in the `List`, or could it be a separate component outside (after) the `List`?

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionTable.js
```diff
@@ -0,0 +1,178 @@
...
+                {`, ${comp.cityName}`}
+              </div>
+              <div className="venue-link">
+                <PseudoLinkMarkdown text={comp.venue} />
+              </div>
+            </span>
+          </List.Item>
+        </React.Fragment>
+      ))}
+      {/* Could not figure out why Semantic UI's animated loader icon doesn't show */}
```
kr-matthews, line 65: Is this still an issue? Maybe it will work, or be easier to debug, given all the recent refactoring?

notrichardpeng: Yeah there should be an animated circle rather than just the "loading..." text. I suspect it's something strange with CSS but just couldn't pinpoint the issue after having tried for a while.

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionList.js
```diff
@@ -0,0 +1,119 @@
+import React, { useEffect } from 'react';
+import { useInView } from 'react-intersection-observer';
+
+import I18n from '../../lib/i18n';
+import { competitionConstants } from '../../lib/wca-data.js.erb';
+
+// Should CompetitionTable be renamed to CompetitionSubList or something?
+import CompetitionTable from './CompetitionTable';
```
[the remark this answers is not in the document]
notrichardpeng: Sounds good, haven't thought about the name of the parent folder when naming these.

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionTable.js
```diff
@@ -0,0 +1,178 @@
...
+function CompetitionTable({
+  competitionData,
+  title,
+  shouldShowRegStatus,
+  shouldIncludeCancelled,
+  selectedEvents,
+  isLoading,
+  hasMoreCompsToLoad,
+  isSortedByAnnouncement = false,
+  isRenderedAboveAnotherTable = false,
```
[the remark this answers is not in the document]
notrichardpeng: Agree with the naming. I don't feel that it's necessary to pass the component in as a child when it can be defined in the same file, and perhaps move the conditions into the function for that component.

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionTable.js
notrichardpeng: They don't have to be in the list, the font and size would just look more alike and the spacing is more even. 

A footer does sound like a clean way to organize these things though.

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionTable.js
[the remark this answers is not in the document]
notrichardpeng: Resolved because of footer component.

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionView.js
```diff
@@ -0,0 +1,90 @@
...
+              competitionData={competitionData}
+              filterState={filterState}
+              shouldShowRegStatus={shouldShowRegStatus}
+              isLoading={competitionsIsFetching}
+              fetchMoreCompetitions={competitionsFetchNextPage}
+              hasMoreCompsToLoad={hasMoreCompsToLoad}
+            />
+          )
+        }
+        {/* Old JS code does a lot of things to id=comeptitions-map, to be included? */}
```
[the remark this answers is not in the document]
notrichardpeng: I will keep the Old JS code because admin view still uses it, but otherwise so far I don't see a point of keeping it.

WcaOnRails/app/webpacker/components/CompetitionOverview/CompetitionTable.js
notrichardpeng: Turns out I just needed to import semantic UI's css. Thought that was already done!

WcaOnRails/app/webpacker/lib/utils/competition-table.js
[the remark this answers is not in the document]
notrichardpeng: I'm not sure if we should switch to using UTC or continue with using local timezones in luxon

One reason is our rails codebase can't deal with timezones. Attributes like `comp.inProgress` and `comp.isProbablyOver` are calculated based on UTC, so there's a consistency issue. In a scenario where UTC crosses midnight but when a comp's end date is still today in the local timezone, the current PR would say that they're not in progress, because they were categorized using these rails attributes, whereas the status icon calculated with luxon shows that they 'Ended 0 days ago'. To address this maybe we should determine inProgress using luxon entirely without any ruby, or switch to UTC and keep it consistent with ruby.

The other reason is using UTC to estimate when a comp is over seems more accurate than using everyone's own timezones(?), since UTC gives a better standardized sense of time, and competitions take place all over the world. 

I also found a whole discussion about `comp.is_probably_over?` here, might give some more insights: https://github.com/thewca/worldcubeassociation.org/pull/1206/files#r98485399

WcaOnRails/app/webpacker/components/CompetitionsOverview/CompetitionsFilters.js
```diff
@@ -0,0 +1,394 @@
...
+      {
+        displayMode === 'list'
+        && (
+          <div id="registration-status" className="registration-status-selector">
+            <Form.Checkbox
+              label={I18n.t('competitions.index.show_registration_status')}
+              name="show_registration_status"
+              id="show_registration_status"
+              onChange={() => setShouldShowRegStatus(!shouldShowRegStatus)}
+            />
```
kr-matthews, line 371: If this is on and I switch to map view then back to list view, it's off. Keeping it on probably makes the most sense. I can't see why it's being reset though - might be fixed by explicitly passing `checked={shouldShowRegStatus}`?

WcaOnRails/app/webpacker/components/CompetitionsOverview/MapView.js
kr-matthews, line 1: Doesn't necessarily need to be done in this PR (or done at all), but either some kind of loading indicator, or a 'no competitions found' message when applicable, might be good? I selected all events in the filter and couldn't tell if there weren't any or they just weren't found yet. That might be tricky and/or misleading since you can zoom/pan around the map and things get hidden?

kr-matthews (commented):
I think this is pretty much ready to go - it would be best if someone else could provide a review now.

WcaOnRails/app/webpacker/components/CompetitionsOverview/CompetitionsFilters.js
notrichardpeng: This actually makes the display out of sync with the actually value of `shouldShowRegStatus`. But yes it's easily fixed by your suggestion

gregorbg (approved):

PR #7909, opened 2023-05-16
Add administrative notes to registrations

### LINKED ISSUE #7628: Organizers' comment textbox on registrations
We often add administrative notes on registrations in the same comment textbox that the competitor can fill out. For example, if a new competitor didn't include their last name on their registration, we might write "Note: Emailed to ask about last name".

Currently, we (obviously) don't put sensitive information in this textbox (since the competitor can see it and in some cases could edit it), but it is likely confusing for a competitor to see an administrative note in the comment textbox.

Add a second textbox for organizers' comments.

Data permissions: only competition managers can view/edit
UI: same as existing comment textbox

Could build/use an external tool, but we'd have to sync with WCIF (or keep pulling a CSV) and the data would be in two places.

**Additional context**
N/A

Rubiks-boy:
Fixes #7628

Adds the ability to add administrative notes when editing a competitor's registration:

Screenshot 2023-05-15 at 10 59 53 PM

Once >=1 registration has admin notes, they show up on `/edit/registrations`:

| No registrations have admin notes | >=1 has admin notes |
| --- | --- |
| Screenshot 2023-05-15 at 11 00 47 PM | Screenshot 2023-05-15 at 11 00 29 PM |

This also adds `administrative_notes` to the non-public WCIF.

**Help requested:** For whatever reason, when I try to run the DB migration, I keep getting really large diffs in the generated structure.sql. It'll replace things like `int(11)` with `int`. I'm running things locally using `docker-compose up`. Has anyone run into this type of thing before?

### DIFF
--- a/WcaOnRails/config/locales/en.yml
+++ b/WcaOnRails/config/locales/en.yml
@@ -639,7 +639,7 @@ en:
         comments: ""
         guests: ""
         status: ""
-        administrative_notes: ""
+        administrative_notes: "Additional notes kept as part of managing this registration. These can't be viewed by the competitor."
       vote:
         comment: ""
       competition_medium:
--- a/WcaOnRails/config/locales/en.yml
+++ b/WcaOnRails/config/locales/en.yml
@@ -362,6 +362,7 @@ en:
         comments: "Comments"
         paid_entry_fees: "Paid"
         status: "Status"
+        administrative_notes: "Administrative notes"
         # We can remove the following 3 keys once https://github.com/thewca/worldcubeassociation.org/issues/275 is done! (spec will complain anyway)
         name: "Name"
         birthday: "Birthday"
@@ -638,6 +639,7 @@ en:
         comments: ""
         guests: ""
         status: ""
+        administrative_notes: ""
       vote:
         comment: ""
       competition_medium:
--- a/WcaOnRails/app/controllers/registrations_controller.rb
+++ b/WcaOnRails/app/controllers/registrations_controller.rb
@@ -786,6 +786,7 @@ class RegistrationsController < ApplicationController
         :deleted_at,
         :accepted_by,
         :deleted_by,
+        :administrative_notes,
       ]
       params[:registration].merge! case params[:registration][:status]
                                    when "accepted"
--- a/WcaOnRails/app/models/competition.rb
+++ b/WcaOnRails/app/models/competition.rb
@@ -240,6 +240,10 @@ class Competition < ApplicationRecord
     persisted_events_id.length
   end
 
+  def has_administrative_notes?
+    registrations.any? { |registration| !registration.administrative_notes.blank? }
+  end
+
   NEARBY_DISTANCE_KM_WARNING = 250
   NEARBY_DISTANCE_KM_DANGER = 10
   NEARBY_DISTANCE_KM_INFO = 100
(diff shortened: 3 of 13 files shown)


gregorbg:
Thank you for your contribution! I will try to get a review in by the end of the week.

In the meantime:
> This also adds administrative_notes to the non-public WCIF.

Changes to the WCIF scheme need to be accompanied by a corresponding PR in https://github.com/thewca/wcif

> For whatever reason, when I try to run the DB migration, I keep getting really large diffs in the generated structure.sql. It'll replace things like int(11) with int. I'm running things locally using docker-compose up. Has anyone run into this type of thing before?

This is a long-standing problem due to the fact that the historical version of `structure.sql` was created with a very different DB server version than what we're running in the Docker container nowadays. Unfortunately, our current solution is to require all devs to manually revert / check-in only the actually relevant lines.

The good news however is that we're literally two weeks away from purging that problem forever. So please bear with it in the meantime :see_no_evil:

Rubiks-boy:
Sounds good! I opened up the PR for the WCIF.

For now, I also checked in the relevant lines from `structure.sql`, slightly modifying the character set to match the existing `comments` row. That being said I'm happy to hold off a couple of weeks to merge and then rebase off those changes if that makes more sense!

WcaOnRails/config/locales/en.yml
```diff
@@ -638,6 +639,7 @@ en:
         comments: ""
         guests: ""
         status: ""
+        administrative_notes: ""
```
gregorbg, line 642: It might be worth actually adding a hint here, to point out the difference between "Comments" and "Administrative notes".

gregorbg (commented):
Looking good overall, thanks for the clean contribution! One remark about the general idea: You're not allowing Delegates to edit administrative comments (as per the tests) but are you introducing any similar restrictions for _viewing_ the notes in the first place? How safe is the assumption that "whoever has access to the edit registrations panel should also have access to at least view administrative notes"?

Also, I vaguely remember that competitors can view their own "Edit registration" dashboard, but I'm not really sure. Might be worth checking, to review whether they should see their own administrative notes.

Rubiks-boy:
> How safe is the assumption that "whoever has access to the edit registrations panel should also have access to at least view administrative notes"?

This is a really good point on viewing the notes! Based on [this line](https://github.com/thewca/worldcubeassociation.org/blob/master/WcaOnRails/app/controllers/registrations_controller.rb#L20), it looks like we block anybody who isn't a competition manager from viewing this page, so we should be safe.

To double-check, I did some manual testing:
* When a competitor originally registers or modifies their own registration in `/register`, we don't show the admin notes field.
* When signed in as a delegate for the competition, the delegate can view/modify admin notes from the `/registrations/{number}/edit` page.
* When visiting the same `/registrations/{number}/edit` page while signed in as that competitor, it redirects you to the WCA home page with the error `You are not allowed to manage competition`.

(I also added a couple more tests in line with trying to maintain this)

gregorbg (approved):
Absolutely cool with the code changes themselves, wonderfully clean submission :sunglasses:

Tests also seem to check out and your reasoning makes sense. I'll rebase on top of the latest master to resolve `structure.sql` changes and then ship! :ship:

FinnIckler:
I'll make sure we add it as a feature to wca-registrations

Issue #7904, opened 2023-05-14
Annoying overlay when using the webpack-dev-server


justin808:
> This statement is correct, i.e. our frontend code compiles without error, but the warning is displayed as a transparent overlay when running webpacker dev server. It is also constantly printed to our logs. Is there a more elegant way than shutting webpacker up via config? Why is it trying to even import react-dom/client anyways when we are clearly referencing "react-dom": "^16.14.0" in our package.json?

Per https://github.com/reactjs/react-rails/pull/1269#issuecomment-1546960928

Please update the webpack config per https://github.com/shakacode/react_on_rails/blob/master/docs/guides/rails-webpacker-react-integration-options.md?plain=1#L59-L86

PR #7913, opened 2023-05-19
Suppress React 18 import warning

gregorbg:
Yes, this is the officially recommended way as per https://github.com/reactjs/react-rails\#getting-warning-for-cant-resolve-react-domclient-in-react--18

Fixes https://github.com/thewca/worldcubeassociation.org/issues/7904


gregorbg:
Merging as dependency update follow-up to https://github.com/thewca/worldcubeassociation.org/pull/7892

PR #7919, opened 2023-05-21
Fix for inbox results not working

danieljames-dj:


gregorbg:
Mh, your solution runs a full check in a place where the check results are not needed. I have take a different approach at trying to fix this :smile:

PR #7920, opened 2023-05-21
Remove intermediate clear inbox step

danieljames-dj:


WcaOnRails/app/controllers/admin_controller.rb
```diff
@@ -481,7 +481,7 @@ def do_complete_persons
 
     ActiveRecord::Base.transaction do
       params[:person_completions].each do |person_key, procedure|
-        next if [:competition_ids, :continue_batch].include? person_key.to_sym
+        next if [:competition_ids, :continue_batch, :clear_inbox_persons_after_finish].include? person_key.to_sym
```
danieljames-dj, line 484: I really don't know what is happening here, I did this because `continue_batch` did this. 😅

gregorbg: What happens is that the `each` loop above expects every entry in the array to be a "do something with inbox persons" action. The few exceptions mentioned here in the `next if` array are exceptions that should not be processed.

WcaOnRails/app/views/admin/complete_persons.html.erb
```diff
@@ -114,7 +114,8 @@
       <% end %>
       </tbody>
     <% end %>
-    <%= f.input :continue_batch, as: :boolean, label: "Continue with the next batch of newcomers if there are more pending entries", hint: "If you are redirected to the 'Create Newcomers' entry point, it means that there are no more pending entries.", input_html: { checked: competition_ids.present? } %>
+    <%= f.input :continue_batch, as: :boolean, label: "Continue with the next batch of newcomers if there are more pending entries", input_html: { checked: competition_ids.present? } %>
+    <%= f.input :clear_inbox_persons_after_finish, as: :boolean, label: "Clear InboxPersons if there are no more pending entries", hint: "If you are redirected to the 'Create Newcomers' entry point, it means that there are no more pending entries.", input_html: { checked: competition_ids.present? } %>
```
gregorbg, line 118: You moved the hint from `continue_batch` to `clear_inbox_persons_after_finish`. This could lead to confusion regarding what the hint is actually referring to. Also, every `f.input` needs an explicit mention of `hint` unless it's not in the i18n YAMLs (and for results posting, we don't want that!) so you have to add something like `hint: false` when you don't want any hint

gregorbg:
Closing as per WRT's (and particularly, Sébastien's) objection.

PR #7924, opened 2023-05-22
Move database environment from DATABASE_URL to DATABASE_HOST

FinnIckler:
Now that we are using the database.yml to define the read replica we should stop using the DATABASE_URL environment variable to set the database so we don't configure them in two places


chef/site-cookbooks/wca/recipes/default.rb
```diff
@@ -424,7 +423,7 @@
   execute "bundle exec rake db:setup" do
     cwd rails_root
     environment({
-                  "DATABASE_URL" => db_url,
+                  "DATABASE_HOST" => db["host"],
```
gregorbg, line 426: The bundle exec scripts should be able to read this from the Rails config YAML

chef/site-cookbooks/wca/recipes/default.rb
```diff
@@ -438,7 +437,7 @@
     cwd rails_root
     user username
     environment({
-                  "DATABASE_URL" => db_url,
+                  "DATABASE_HOST" => db["host"],
```
gregorbg, line 440: same, see above.

chef/site-cookbooks/wca/templates/wca.screenrc.erb
```diff
@@ -5,7 +5,7 @@ defscrollback 50000
 chdir <%= @rails_root %>
 screen -t run
 <% if @rails_env == "development" %>
-stuff "RACK_ENV=development DATABASE_URL=<%= @db_url %> bundle exec rails server\012"
+stuff "RACK_ENV=development DATABASE_HOST=<%= @db_host %> bundle exec rails server\012"
```
gregorbg, line 8: perhaps not necessary, see above

gregorbg (approved):
some redundancies in the Chef file (but we don't really care about Chef setup anymore anyways, so feel free to ignore)

FinnIckler:
Works on staging. Merging

PR #7925, opened 2023-05-22
Fetch data from the read replica for often used expensive routes

FinnIckler:
This PR adds the read replica as the data source for the following routes:
- /competition/:compid/psychsheet
- /competitions/mine
- /api/v0/search (which is used for search bar)

These are currently 3 of the most computationally intensive and most used routes. 

Other routes we could consider
- all /api/v0 routes [these are not used by a lot of people, but can be quite expensive]
- routes used by user profile [we currently have no replica lag, but could be confusing for people if they get a notification that results are uploaded, but the profile isn't updated]


gregorbg:
Note for reviewers: Under the "Files changed" review/diff tab, go to Settings > Hide whitespace. Then you will see that the >100 line diff is really just a 6 line diff where a lot of indentation changes (and therefor, the lines themselves _technically_ change)

FinnIckler:
I am pretty sure the errors will go away when #7924 is merged. Will investigate more tomorrow

gregorbg (approved):

gregorbg:
Note to future WST about why the truncation strategy is necessary: https://stackoverflow.com/questions/73460447/multiple-databases-replica-not-working-on-test-environment-with-rails-6

(in short, the `:transaction` strategy wraps every test into a transaction block, which conflicts with the notion of switching databases in the middle of a test, i.e. in the middle of a transaction)

PR #7929, opened 2023-05-24
Make `nil` be the default value for regional markers

viroulep:
When implementing #6360 it seems I did found both occurrences in the db and picked `""` as the default, it turns out having `nil` as the default sounds better for the WRT so this PR intends to fix the current implementation.
This fixes the validation and form to make sure `nil` is the default value, and that `""` is not allowed anymore.
I'll update all existing `""` regional markers before this get merged so that all results in the db are in a "valid" state (as far as rails is concerned) using these queries:
```sql
update `Results` set `regionalSingleRecord`=null WHERE `regionalSingleRecord` ="";
update `Results` set `regionalAverageRecord`=null WHERE `regionalAverageRecord` ="";
```

### DIFF
--- a/WcaOnRails/app/models/result.rb
+++ b/WcaOnRails/app/models/result.rb
@@ -18,11 +18,12 @@ class Result < ApplicationRecord
   # we also need sure to query the correct competition as well through a custom scope.
   belongs_to :inbox_person, ->(res) { where(competitionId: res.competitionId) }, primary_key: :id, foreign_key: :personId, optional: true
 
-  # NOTE: both nil and "" exist in the database, we may consider cleaning that up.
-  MARKERS = [nil, "", "NR", "ER", "WR", "AfR", "AsR", "NAR", "OcR", "SAR"].freeze
+  MARKERS = [nil, "NR", "ER", "WR", "AfR", "AsR", "NAR", "OcR", "SAR"].freeze
 
   validates_inclusion_of :regionalSingleRecord, in: MARKERS
   validates_inclusion_of :regionalAverageRecord, in: MARKERS
+  alias_attribute :regional_single_record, :regionalSingleRecord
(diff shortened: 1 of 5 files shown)


WcaOnRails/app/models/result.rb
```diff
@@ -64,11 +63,11 @@ def country_iso2
...
 
   def regional_single_record
-    regionalSingleRecord || ""
+    regionalSingleRecord
   end
 
   def regional_average_record
-    regionalAverageRecord || ""
+    regionalAverageRecord
   end
```
gregorbg, line 71: Could this be implemented by virtue of Rails `alias_attribute`?

gregorbg (changes requested):
Formally giving a blocking review until the question about your update query from the original post is resolved.

WcaOnRails/app/models/result.rb
viroulep: > At the risk of overlooking something completely obvious here: Isn't the update query you suggested extremely dangerous? It overrides the columns if either one of the regional record markers is blank, which would in theory override a perfectly valid single `WR` just if the average WR regional record marker is set to an empty string...

Let's say it was a test to make sure you were paying attention and that I did not actually suggest to run this :sweat_smile: 

I've updated the original post with two separate queries which should be safe.

viroulep:
Thanks for the quick and efficient review @gregorbg!

I've updated the PR to address your comments and to also include a fix for the record class (which now matches how ruby's one is implemented).

gregorbg (approved):

gregorbg:
As far as I can tell, this PR is ready to be merged. When do you intend to run the database query?

viroulep:
I think we should simply synchronize so that I run them when you merge this, let's figure this out on slack ;)

gregorbg:
Executed the queries on your behalf :wink:

PR #7932, opened 2023-05-27
Update competitions_controller.rb

coder13:
We shouldn't need to do a DB call before checking cache.

Note:
edited directly from browser, untested.


dunkOnIT:
Tested locally and it worked fine, ie no change in functionality when hitting the pubic competition endpoint.

gregorbg (approved):

PR #7954, opened 2023-06-01
Fixed dues value fetching bug

danieljames-dj:
There was an issue because of which `$("#base_entry_fee_lowest_denomination_input_field").val()` was returning the value as given in the input, ie, $300.00, and hence `to_i` returned 0. This change will remove the currency and decimal and make it in the lowest denomination 30000.

### DIFF
--- a/WcaOnRails/app/views/competitions/_competition_form.html.erb
+++ b/WcaOnRails/app/views/competitions/_competition_form.html.erb
@@ -252,7 +252,7 @@
             var compLimit = compLimitEnabled ? $("#competition_competitor_limit").val() : null;
             var currencyCode = $("#competition_currency_code").val();
             var countryId = $("#competition_countryId").val();
-            var entryFee = $("#base_entry_fee_lowest_denomination_input_field").val();
+            var entryFee = $("#competition_base_entry_fee_lowest_denomination").val();
             window.wca.cancelPendingAjaxAndAjax('calculate_dues', {
               url: '<%= calculate_dues_path %>',
               data: {


gregorbg (changes requested):
Good catch! Unfortunately, money and currencies in the frontend are quite a mess. The basic problem is that most human currencies are expected to be floating point numbers ($12.34 as "twelve US dollars and thirty-four US cents") but computers _suck_ at storing floating point numbers precisely. So the solution is to store all money amounts as integers in the lowest denomination (aka. always store cents, for example the above would be stored as `USD 1234` which is correct if you interpret it as US *cents*)

In frontend, the problem is that humans intuitively read USD 1234 as "one thousand two hundred thirty-four dollars" and freak out. So what happens in the frontend is that we have *two* input fields. One "nice-looking" field for humans, and one *hidden* field for storing the internal, "lowest denomination" currency.

Regarding your PR: `val()`'ing the _human_ field and then stripping away non-digits doesn't work for currencies that don't use sub-units, like the Japanese Yen for example. Calling string replacement methods on monetary amounts is _never_ a good idea in general.

The clean solution is to call `val` on the underlying background hidden field in the form and send that value (which is guaranteed to be correct integer-only) to the backend: `#competition_base_entry_fee_lowest_denomination` holds that value in the case of the competition form. So changing the jQuery selector should be enough

(SIde note: Yes it's weird that nobody noticed this before.)

danieljames-dj:
Thanks Greg for detailed comment. I didn't think about `competition_base_entry_fee_lowest_denomination`, I think I missed it during initial implementation as well, my bad. Anyway, I've made the change as you suggested.

gregorbg (approved):

PR #7977, opened 2023-06-03
Fix developer db import

coder13:
The location of the developer export changed but was not reflected in the import command.


gregorbg:
Why was this suddenly closed right after it was openend?
Also, there is a redirect in place so the old link _should_ still work in theory

PR #5417, opened 2020-05-05
Delegate Table Restructuring

danieljames-dj:
This is the draft PR to discuss the Delegate table restructuring. The following are the new proposed table structure. Please go through it and give your suggestions.

### Delegates Table (name: `delegates`)

| Column name | Column type | Details |
|---|---|---|
| id | bigint(20) | Primary key |
| user_id | bigint(20) | Used as a foreign key to reference `users` table |
| status | varchar(191) | This will represent the status of the delegate - like 'trainee', 'junior', etc |
| region_id | bigint(20) | Used as a foreign key to reference `regions` table |
| country_iso2 | varchar(191) | Used to represent the country |
| location | varchar(191) | The location where the delegate will be delegating |
| start_date | date | Start date of this delegating status |
| end_date | date | End date of this delegating status |
| created_at | datetime | Date & time at which the row is created |
| updated_at | datetime | Date & time at which the row is last edited |

### Regions Table (name: `regions`)

| Column name | Column type | Details |
|---|---|---|
| id | bigint(20) | Primary key |
| name | varchar(191) | Name of the region |
| friendly_id | varchar(191) | Friendly ID to easy access |
| parent_region | bigint(20) | Used as a foreign key to reference `regions` table |
| is_active | tinyint(1) | Whether the region is still active |

The migration files to create the tables are available in the commit, please check that as well.

(@moralsh was supposed to send this PR, but I'm doing this as he is busy right now)
cc: @moralsh @Jambrose777 
Related issue: Issue #4222

PR #7979, opened 2023-06-04
Migrate delegate page to react

danieljames-dj:


WcaOnRails/app/webpacker/components/Delegates/index.jsx
```diff
@@ -0,0 +1,112 @@
...
+      .filter((user) => user.delegate_status === 'senior_delegate')
+      .sort((user1, user2) => {
+        if (user1.region < user2.region) {
+          return -1;
+        } if (user1.region > user2.region) {
+          return 1;
+        }
+        return 0;
+      }));
+  }, []);
```
kr-matthews, line 31: This effect is missing dependencies. If it's really just meant to run once initially then you could just pass this function to the `useState` for `seniorDelegates` to be the initial value.
If it is meant to change whenever `delegates` changes then similarly to my comment below, just having a const which calculates this directly, possibly wrapped in a `useMemo`, would be cleaner.

WcaOnRails/app/webpacker/components/Delegates/index.jsx
```diff
@@ -0,0 +1,112 @@
...
+          } if (user1.region > user2.region) {
+            return 1;
+          } if (user1.name < user2.name) {
+            return -1;
+          } if (user1.name > user2.name) {
+            return 1;
+          }
+          return 0;
+        })]));
+  }, [seniorDelegates]);
```
kr-matthews, line 49: Since this is directly calculated from the `seniorDelegates` state, it doesn't need to be stored in its own state, it could just be a const:
```
const delegatesUnderSeniorDelegates = seniorDelegates.map(...)
```
If this computation is expensive then wrapping it in `useMemo` would prevent it recalculating every render.
```
const delegatesUnderSeniorDelegates = useMemo(seniorDelegates.map(...), [seniorDelegates])
```

WcaOnRails/app/webpacker/components/Delegates/index.jsx
```diff
@@ -0,0 +1,112 @@
...
+
+  const dasherize = (string) => string.replace(/_/g, '-');
```
kr-matthews, line 51: I think it's nice to put functions like this (which don't depend on any component state) outside the component, to avoid them being redefined (as a fresh reference) every render.

WcaOnRails/app/webpacker/components/Delegates/index.jsx
```diff
@@ -0,0 +1,112 @@
...
+        <thead>
+          <tr>
+            <th />
+            <th>{I18n.t('delegates_page.table.name')}</th>
+            <th>{I18n.t('delegates_page.table.role')}</th>
+            <th>{I18n.t('delegates_page.table.region')}</th>
+          </tr>
+        </thead>
+        <tbody>
+          {!!delegatesUnderSeniorDelegates[index]
```
kr-matthews, line 74: Could this ever be false?

WcaOnRails/app/webpacker/components/Delegates/index.jsx
danieljames-dj: Okay, done.

WcaOnRails/app/webpacker/components/Delegates/index.jsx
danieljames-dj: Done.

WcaOnRails/app/webpacker/components/Delegates/index.jsx
danieljames-dj: Done.

WcaOnRails/app/webpacker/components/Delegates/index.jsx
danieljames-dj: Yes in initial load, but with above changes, it is no more required. :-)

danieljames-dj (commented):
Thanks @kr-matthews for your review. :-)

WcaOnRails/app/webpacker/components/Delegates/index.jsx
```diff
@@ -0,0 +1,106 @@
...
+    .sort((user1, user2) => {
+      if (user1.region < user2.region) {
+        return -1;
+      } if (user1.region > user2.region) {
+        return 1;
+      }
+      return 0;
+    });
+
+  const [seniorDelegates] = React.useState(getSeniorDelegates());
```
kr-matthews, line 27: As written, the initial state `getSeniorDelegates()` for this state will be recalculated every render, even though it's only used once (the initial render). If you pass in the literal function `getSeniorDelegates` instead of the result `getSeniorDelegates()`, it will only calculate it the first render. [See the docs](https://legacy.reactjs.org/docs/hooks-reference.html#lazy-initial-state).

Having said that, if you're not using a setter then you don't need a state. This could be refactored to
```
const seniorDelegates = delegates
    .filter((user) => user.delegate_status === 'senior_delegate')
    .sort((user1, user2) => {
      if (user1.region < user2.region) {
        return -1;
      } if (user1.region > user2.region) {
        return 1;
      }
      return 0;
    });
```
or
```
const seniorDelegates = useMemo(() => delegates
    .filter((user) => user.delegate_status === 'senior_delegate')
    .sort((user1, user2) => {
      if (user1.region < user2.region) {
        return -1;
      } if (user1.region > user2.region) {
        return 1;
      }
      return 0;
    }), [delegates]);
```

danieljames-dj: Done, thank you.

WcaOnRails/app/webpacker/components/Delegates/index.jsx
```diff
@@ -0,0 +1,104 @@
...
+      if (user1.region < user2.region) {
+        return -1;
+      } if (user1.region > user2.region) {
+        return 1;
+      }
+      return 0;
+    }), [delegates]);
+
+  // TO_VERIFY: I assume there are no cases where there are no delegates without
+  // senior delegate unless they are senior delegate
```
kr-matthews, line 28: I assume this still needs to be resolved before the PR is ready to be merged?

danieljames-dj: Hi Kevin, Sorry for late reply. Actually yes. As far as I know, there are no cases right now where there are delegates without senior delegates (unless they themself is a senior delegate). So, it should be safe to assume this, but for a double verification, would like to get a yes from anybody from Board or our Leader (Gregor).

WcaOnRails/app/webpacker/components/Delegates/index.jsx
```diff
@@ -0,0 +1,104 @@
...
+}) {
+  const seniorDelegates = React.useMemo(() => delegates
+    .filter((user) => user.delegate_status === 'senior_delegate')
+    .sort((user1, user2) => {
+      if (user1.region < user2.region) {
+        return -1;
+      } if (user1.region > user2.region) {
+        return 1;
+      }
+      return 0;
```
gregorbg, line 24: Have you considered `.sort((user1, user2) => user1.region.localeCompare(user2.region))`?

WcaOnRails/app/webpacker/components/Delegates/index.jsx
```diff
@@ -0,0 +1,104 @@
...
+          return 1;
+        }
+        return 0;
+      })];
+    return (
+      <div
+        className="table-responsive"
+        key={`region-${seniorDelegate.id}`}
+      >
+        <table className="table delegates-table">
```
gregorbg, line 50: Semantic UI React has dedicated tables: https://react.semantic-ui.com/collections/table/

WcaOnRails/app/webpacker/components/Delegates/index.jsx
gregorbg: It is worth noting that the old Ruby implementation had a `@delegates_without_senior_delegates` variable. So it does seem like this edge case can occur.

Why is it so important to explicitly assume that _Senior Delegates_ are the top of the hierarchy? I'd like to suggest some approach where you iterate over the list of delegates and group them by whoever seems to be the "senior-most Delegate"? In essence, just interpret `senior_delegate_id` as a hierarchy that you can traverse upwards

gregorbg (changes requested):
Suggestions for improval inline :)

WcaOnRails/app/webpacker/components/Delegates/index.jsx
danieljames-dj: Actually, I feel earlier there were cases where delegates existed without senior delegates. But now, there are nobody like that. Even though I'm 99% sure that the case doesn't exist now, I would like to get a confirmation about the same. :-)

The next PR coming immediately after merging this will be introducing of regions table. Once that comes, every delegates will be mapped to the region instead of senior delegate and this issue won't be there anymore.

This PR was basically to implement the same thing that was there earlier, and I feel adding @delegates_without_senior_delegates won't be so good because we don't have any use-case for that right now and once the next PR comes, the code has to be removed.

What do you think regarding the same?

WcaOnRails/app/webpacker/components/Delegates/index.jsx
danieljames-dj: Thank you, migrated to semantic-ui table.

danieljames-dj (commented):
Thanks for reviewing, @gregorbg.

WcaOnRails/app/webpacker/components/Delegates/index.jsx
danieljames-dj: I think I didn't give much context regarding regions table. It's the same regions table mentioned in [this PR](https://github.com/thewca/worldcubeassociation.org/pull/5417). I was initially thinking to do both delegates table and regions table together, but there are plenty of cases to handle in that case and is completely out of my scope. So I thought doing it step by step will be better. The next PR will have just the regions table implementation (it won't be dead table for sure, I'll be connecting the delegates to regions instead of senior delegates in the same PR). Some PRs after that will be some UI improvements to make the delegates table implementation easy.  And finally, I'll be bringing the delegates table PR.

WcaOnRails/app/webpacker/components/Delegates/index.jsx
```diff
@@ -0,0 +1,97 @@
...
+  const seniorDelegates = React.useMemo(() => delegates
+    .filter((user) => user.delegate_status === 'senior_delegate')
+    .sort((user1, user2) => user1.region.localeCompare(user2.region)), [delegates]);
+
+  // TO_VERIFY: I assume there are no cases where there are no delegates without
+  // senior delegate unless they are senior delegate
+
+  return seniorDelegates.map((seniorDelegate) => {
+    const delegatesUnderSenior = [seniorDelegate, ...delegates
+      .filter((user) => user.senior_delegate_id === seniorDelegate.id && user.delegate_status !== 'trainee_delegate')
```
gregorbg, line 25: Why are we excluding Trainee Delegates?

One common nitpick that we sometimes get with the current Delegate page is that Trainees are not marked as such. We could use this rewrite as a motivation to include them, either by adding them in a separate section of the table below all Staff Delegates or by adding them inline with the other Delegates but using a different color.

WcaOnRails/app/webpacker/components/UserBadge.js
```diff
@@ -50,8 +50,10 @@ function UserBadge({
   badgeClasses = '',
   senior = false,
   leader = false,
+  hideBorder = false,
+  leftAlign = false,
 }) {
-  const classes = classnames('user-badge', badgeClasses, { senior, leader });
+  const classes = classnames('user-badge', badgeClasses, { senior, leader, leftAlign });
```
gregorbg, line 56: Our naming convention for CSS classes is `left-align`

WcaOnRails/app/webpacker/stylesheets/user_badge.scss
```diff
@@ -22,14 +22,19 @@
     }
   }
 
+  &.leftAlign {
```
gregorbg, line 25: `left-align` as per my other comment

gregorbg (commented):
I played around with the refactor on my local machine and it looks good! Thanks for taking the initiative. Some remarks about individual sections of code are listed below

WcaOnRails/app/webpacker/components/UserBadge.js
danieljames-dj: Done, thanks for pointing this.

WcaOnRails/app/webpacker/stylesheets/user_badge.scss
danieljames-dj: Done, again thank you

WcaOnRails/app/webpacker/components/Delegates/index.jsx
danieljames-dj: The main reason why trainee delegate is not added is because it is not there currently. 😅
And yes, adding it as part of this migration is possible, but I'm more excited to have the delegates table come into picture fast. If that happens, it will be more easier to show trainee delegates as well. I can add the trainee delegates in this PR, but I might have to change it completely once the next few PRs for this project comes.

Is it fine if I don't add it in this PR? I can promise that the next few PRs of delegate table project will come soon after merging this. In those PRs, trainee delegates will also come into the page.

WcaOnRails/app/webpacker/components/Delegates/index.jsx
```diff
@@ -0,0 +1,97 @@
...
+      .sort((user1, user2) => {
+        if (user1.region < user2.region) {
+          return -1;
+        } if (user1.region > user2.region) {
+          return 1;
+        } if (user1.name < user2.name) {
```
coder13, line 31: nit: use `.localeCompare` for name comparisons.

danieljames-dj: Done, thanks Cailyn.

WcaOnRails/app/webpacker/components/Delegates/index.jsx
gregorbg: Makes sense to me!

WcaOnRails/app/webpacker/components/Delegates/index.jsx
gregorbg: Agreed as per your suggestions above

WcaOnRails/app/webpacker/components/Delegates/index.jsx
```diff
@@ -0,0 +1,91 @@
...
+                    display: 'flex',
+                    alignItems: 'center',
+                  }}
+                  >
+                    <Button
+                      href={`mailto:${delegate.email}`}
+                      className="ui icon button"
+                    >
+                      <i className="envelope icon" />
+                    </Button>
```
gregorbg, line 61: Semantic UI has cool shorthands for buttons with icons: https://react.semantic-ui.com/shorthand-props
```suggestion
                    <Button
                      href={`mailto:${delegate.email}`}
                      icon="envelope"
                    />
```

WcaOnRails/app/webpacker/components/Delegates/index.jsx
```diff
@@ -0,0 +1,91 @@
...
+                    {isEditVisible && (
+                    <Button
+                      href={`users/${delegate.id}/edit`}
+                      className="ui icon button"
+                    >
+                      <i className="edit icon" />
+                    </Button>
```
gregorbg, line 68: Should be indented, and use a "shorthand icon" as above.

WcaOnRails/app/webpacker/components/Delegates/index.jsx
```diff
@@ -0,0 +1,91 @@
...
+
+export default function Delegates({
+  delegates,
+  isEditVisible,
+}) {
+  const seniorDelegates = React.useMemo(() => delegates
+    .filter((user) => user.delegate_status === 'senior_delegate')
+    .sort((user1, user2) => user1.region.localeCompare(user2.region)), [delegates]);
+
+  // TO_VERIFY: I assume there are no cases where there are no delegates without
```
gregorbg, line 20: Resolve this comment from "TO VERIFY" into something like "NOTE: This UI currently assumes that..."

gregorbg (changes requested):
Design decisions are okay, some minor changes because I found out cool stuff about Sematic UI :smile:

WcaOnRails/app/webpacker/components/Delegates/index.jsx
danieljames-dj: Done.

WcaOnRails/app/webpacker/components/Delegates/index.jsx
danieljames-dj: Thanks Greg.

WcaOnRails/app/webpacker/components/Delegates/index.jsx
danieljames-dj: Thank you.

WcaOnRails/app/webpacker/components/Delegates/index.jsx
danieljames-dj: Looks neat, thanks for this. Changed.

WcaOnRails/app/webpacker/components/Delegates/index.jsx
danieljames-dj: Done.

danieljames-dj (commented):
Thanks for the review @gregorbg. I've done the changes as per your comments.

WcaOnRails/app/webpacker/components/Delegates/index.jsx
```diff
@@ -0,0 +1,87 @@
...
+                    alignItems: 'center',
+                  }}
+                  >
+                    <Button
+                      href={`mailto:${delegate.email}`}
+                      icon="envelope"
+                    />
+                    {isEditVisible && (
+                      <Button
+                        href={`mailto:${delegate.email}`}
```
gregorbg, line 62: You accidentally did a mistake while copy-pasting (or did I make the mistake in my code review suggestion? :fearful:)
The `href` here should *not* be mailto

gregorbg (changes requested):
copy/paste fail, hope I can catch you while you're still active :P

WcaOnRails/app/webpacker/components/Delegates/index.jsx
danieljames-dj: Oops I messed it up while copy-pasting 🙊 Sorry for it. Fixed it now.

gregorbg (approved):

PR #7982, opened 2023-06-05
Change DB password to empty

danieljames-dj:
We used to had empty string as our DB password in our test environment for so long but got changed recently [here](https://github.com/thewca/worldcubeassociation.org/pull/7924/files#diff-3a2e07e774ec7158ae277fe968a714d988fd15c2351be6d4fa2f92b125bafaf2R17).

Even in our documentations, we have said to have empty password for root user. This change is actually failing tests in local, and might trouble new developers to figure out the issue. Hence changing back to empty string - which is how it was earlier.


gregorbg:
You're changing the test environment, which is not relevant for developing. The test password has always been set to `root`, this has been unchanged since at least 2021.

In Finn's diff that you mentioned, you can cleary see it was changed from a previous `mysql2://root:root@127.0.0.1:3306/wca_test` password: https://github.com/thewca/worldcubeassociation.org/commit/755d73308879ad1eb425063cc14ebc5d6dc86fa9#diff-ce0dee93a528f6c7648f4cd671a3ec7be6a80f976c88f3aa1c322b7a80828fbaL13

FinnIckler:
Yeah you can see that the tests are failing immediately because the test environment _expects_ the password to be root.

danieljames-dj:
Okay, I'm still wondering why my bin/rspec threw database connection error in my local from the commit of PR #7924. It was working fine till the commit before that. 🤔

FinnIckler:
probably because you are running it against the dev database and not the test one?

danieljames-dj:
I'll explain my case. I've SQL in my system and to sign into MySQL, I am using empty password. So, let's say whatever database I select - let it be wca_development (dev database) or wca_test (test database), both is supposed to have same password because the password is for MySQL and not the database. So, when I run bin/rspec, it's using wca_test, but still it will expect empty password. I'm not sure how I can set different passwords for dev and test when both databases are part of same SQL.

Sorry if I'm missing something here, correct me if I'm wrong. 🙂

FinnIckler:
In that case, please maintain your own config.

gregorbg:
> please maintain your own config.

You can easily create a `.env.test.local` file under `WcaOnRails` that overrides any settings in `.env.test` for your local machine, without affecting git.

danieljames-dj:
Okay, will do that. Thank you.

PR #7983, opened 2023-06-05
Clean up SQL schema: Change column definitions

gregorbg:
- Drop the `rails_persons` view and directly change `Persons` to act like the old `rails_persons`
- Change `(day, month, year)` tuples into proper SQL `date` type columns

NOT inclued b/c out of scope:
Changing naming conventions from camelCase to snake_case. This will follow in a later PR.


FinnIckler (approved):
LGTM

PR #7984, opened 2023-06-06
Only record Stripe charges that are actually successful

gregorbg:
Because we don't care whether people had a typo in their CVC...


dunkOnIT:
My reading of the effect this will have: 
* `holder.record_payment` will only be called if a new charge has a "successful" state

This makes sense as `record_payment` creates recording a payment for a registration, and we don't want that to happen for unsuccessful transactions.

dunkOnIT (approved):
LGTM but I won't be hurt if you want to wait for a second opinion :P

PR #7985, opened 2023-06-06
Fix error response for Results Admin API

gregorbg:
as per problem reported by WRT via email.
The `errors` object was refactored in Rails 6.1 and we migrated _almost_ every place except this one :sweat_smile: https://code.lulalala.com/2020/0531-1013.html


FinnIckler (approved):
LGTM!

PR #7986, opened 2023-06-07
Add devise-jwt and jwt auth route to Monolith

FinnIckler:
This PR is a reference implementation for adding jwt to our devise workflow. This does not have to be merged for quite a while, but I will use it to implement our auth structure in the wca-registration repo.

This PR adds:
- a JWT creation secret (let's hope we will store this in vault by the time this is merged)
- a JWT payload for users that contain the wca_id and user_id (we might need to add more in the future and I will update this PR as necessary)
- a JWT auth route that the frontend can call to get a token (I really don't know how the CORS setting work on here right now but this might be accessible from other websites currently which we don't want)

The flow looks like this:
- The wca.org frontend serves the bundle for the micro service frontend
- The frontend contains code hitting the JWT auth route to get a token on requests
- The JWT token is sent to the microservice where it can be validated and used

Corresponding PR in wca-registration https://github.com/thewca/wca-registration/pull/66


WcaOnRails/.env.test
```diff
@@ -16,3 +16,4 @@ ACTIVERECORD_PRIMARY_KEY=abcdefghijklmnopqrstuvwxyz1234567890
 ACTIVERECORD_DETERMINISTIC_KEY=abcdefghijklmnopqrstuvwxyz1234567890
 ACTIVERECORD_KEY_DERIVATION_SALT=abcdefghijklmnopqrstuvwxyz1234567890
 DATABASE_PASSWORD=root
+DEVISE_JWT_SECRET_KEY=1ee0f7e7439ae477bb423e672e118704e1ed43eeeb4ac796c7a9ca3e0e7a431097e1eb976ce827cc94e7728e75cb2e91a0f3c7d8e6b1a3f7e9d5
```
gregorbg, line 19: see above.

WcaOnRails/app/controllers/api/v0/api_controller.rb
```diff
@@ -30,6 +30,14 @@ def auth_results
     render json: { status: "ok" }
   end
 
+  def jwt
+    if !current_user
```
gregorbg, line 34: `unless`

WcaOnRails/config/initializers/devise.rb
```diff
@@ -237,6 +237,14 @@
   # The default HTTP method used to sign out a resource. Default is :delete.
   config.sign_out_via = :delete
 
+  config.jwt do |jwt|
+    jwt.secret = ENV.fetch("DEVISE_JWT_SECRET_KEY")
```
gregorbg, line 241: Please configure this in EnvVars via the existing `SuperConfig` integration

WcaOnRails/app/models/jwt_denylist.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class JwtDenylist < ApplicationRecord
+  include Devise::JWT::RevocationStrategies::Denylist
+
+  self.table_name = 'jwt_denylist'
```
gregorbg, line 6: Please consider another name for the table so it adheres to Rails conventions

gregorbg (changes requested):
Minor changes as per Google Meet

FinnIckler:
Don't think it makes any sense to reuse this branch as it uses the old database structure. Will open a new one

PR #7988, opened 2023-06-09
Registration import issue of country fixed

danieljames-dj:
WST received an email where the import was failing. The reason for failing was they were sending the country name in country column, but we were considering it as country ID and hence we were not getting country object for Country since we were using c_find.


gregorbg (changes requested):
This change is definitely rejected. It has been our standard to import CSVs by country ID for *years* and a lot of competitions all around the world have systems that are calibrated to use the IDs correctly. If a CSV upload for one specific competition fails, please ask them to change their CSV file.

danieljames-dj:
Oops, okay. I misunderstood it. I'll reply accordingly. Thanks Gregor.

PR #7989, opened 2023-06-09
Changed cutoff to greater than instead of greater than or equal to

danieljames-dj:
This bug was informed to WST via email

### DIFF
--- a/WcaOnRails/app/models/round.rb
+++ b/WcaOnRails/app/models/round.rb
@@ -142,7 +142,7 @@ class Round < ApplicationRecord
   end
 
   def cutoff_is_greater_than_time_limit?
-    cutoff && time_limit != TimeLimit::UNDEF_TL && time_limit.cumulative_round_ids.empty? ? cutoff.attempt_result >= time_limit.centiseconds : false
+    cutoff && time_limit != TimeLimit::UNDEF_TL && time_limit.cumulative_round_ids.empty? ? cutoff.attempt_result > time_limit.centiseconds : false
   end
 
   # cutoffs are too fast if they are less than 5 seconds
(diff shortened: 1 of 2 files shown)


gregorbg:
What is the stance of the WCA Regulations on this matter? Should this change be green-lit by WRT?

danieljames-dj:
I've gone through the regulation and there is no statement that prevents delegates from having same cutoff and time limit. But I'm tagging @thewca/wrc-team here to get a confirmation.

@thewca/wrc-team could you please let us know if it is fine to change the condition from "time limit should be greater than or equal to cutoff" to "time limit should be greater than cutoff", because technically the delegates can have an event with same time limit and cutoff (or in other words, they can say only those who solves within the time limit in first two solves get to do all solves).

CarterKoala:
Let someone else respond too (hasn't been discussed with the team) but I don't think this should be changed. I don't really see what the point of this is - there is nothing in the regulations stating that the cutoff has to be less than the time limit, and I think that it would be reasonable to have them be the same.

danieljames-dj:
Hi @CarterKoala I'm extremely sorry for wrong wording. I've corrected the wording now. And it is correct, currently it's not possible to put a time limit equal to cutoff.

CarterKoala:
Thanks for the clarification! In that case the change looks good :)

danieljames-dj:
Thanks a lot, @CarterKoala. I hope your approval can be considered as WRC approval.

gregorbg (approved):

DougCube:
> there is no statement that prevents delegates from having same cutoff and time limit

Minor nitpick: I think you mean "organizers," not "delegates."
Btw, I was the one that originally proposed this change.

PR #7991, opened 2023-06-09
Change the qualification text to match implementation.

timreyn:
The website implements qualification times as "before X", the current text says "the deadline is X", which is ambiguous and most people interpret that as "no later than X".


gregorbg (commented):
Would it be worth specifying as "must be set **strictly** before"?

timreyn:
I think "before" is clear enough -- in my opinion, "strictly" adds harshness, but doesn't add any extra clarity.

gregorbg (approved):

gregorbg:
Okay, you're a native English speaker so I happily defer to your intuition about the exact phrasing :D

PR #8001, opened 2023-06-21
Include non-competing staff in the public WCIF export

timreyn:
This will allow tools like competitiongroups.com to show non-competing staff members' assignments.

Depends on thewca/wcif#15.

### DIFF
--- a/WcaOnRails/app/models/competition.rb
+++ b/WcaOnRails/app/models/competition.rb
@@ -1771,7 +1771,8 @@ class Competition < ApplicationRecord
     wcif_persons.each do |wcif_person|
       local_assignments = []
       registration = registrations.find { |reg| reg.user_id == wcif_person["wcaUserId"] }
-      # If no registration is found, assume that this is a non-competing staff member being added.
+      # If no registration is found, and the Registration is marked as non-competing, add this person as a non-competing staff member.
+      next unless registration || !wcif_person["registration"] || wcif_person["registration"]["is_competing"] != false
       registration ||= registrations.create(
         competition: self,
         user_id: wcif_person["wcaUserId"],
--- a/WcaOnRails/app/models/competition.rb
+++ b/WcaOnRails/app/models/competition.rb
@@ -1645,7 +1645,7 @@ class Competition < ApplicationRecord
                                 .includes(includes_associations)
                                 .to_enum
                                 .with_index(1)
-                                .select { |r, registrant_id| authorized || r.accepted? || !r.is_competing? }
+                                .select { |r, registrant_id| authorized || r.wcif_status == "accepted" }
                                 .map do |r, registrant_id|
                                   managers.delete(r.user)
                                   r.user.to_wcif(self, r, registrant_id, authorized: authorized)
--- a/WcaOnRails/app/models/competition.rb
+++ b/WcaOnRails/app/models/competition.rb
@@ -1638,6 +1638,9 @@ class Competition < ApplicationRecord
         person: [:ranksSingle, :ranksAverage],
       } },
     ]
+    # NOTE: we're including non-competing registrations so that they can have job
+    # assignments as well. These registrations don't have accepted?, but they
+    # should appear in the WCIF.
     persons_wcif = registrations.order(:id)
                                 .includes(includes_associations)
                                 .to_enum
@@ -1772,14 +1775,17 @@ class Competition < ApplicationRecord
       local_assignments = []
       registration = registrations.find { |reg| reg.user_id == wcif_person["wcaUserId"] }
       # If no registration is found, and the Registration is marked as non-competing, add this person as a non-competing staff member.
-      next unless registration || !wcif_person["registration"] || wcif_person["registration"]["is_competing"] != false
-      registration ||= registrations.create(
-        competition: self,
-        user_id: wcif_person["wcaUserId"],
-        created_at: DateTime,
-        updated_at: DateTime,
-        is_competing: false,
-      )
+      adding_non_competing = wcif_person["registration"].present? && wcif_person["registration"]["isCompeting"] == false
+      if adding_non_competing
+        registration ||= registrations.create(
+          competition: self,
+          user_id: wcif_person["wcaUserId"],
+          created_at: DateTime.now,
+          updated_at: DateTime.now,
+          is_competing: false,
+        )
+      end
+      next unless registration.present?
       WcifExtension.update_wcif_extensions!(registration, wcif_person["extensions"]) if wcif_person["extensions"]
       # NOTE: person doesn't necessarily have corresponding registration (e.g. registratinless organizer/delegate).
       if wcif_person["roles"]
--- a/WcaOnRails/app/models/competition.rb
+++ b/WcaOnRails/app/models/competition.rb
@@ -1642,7 +1642,7 @@ class Competition < ApplicationRecord
                                 .includes(includes_associations)
                                 .to_enum
                                 .with_index(1)
-                                .select { |r, registrant_id| authorized || r.accepted? }
+                                .select { |r, registrant_id| authorized || r.accepted? || !r.is_competing? }
                                 .map do |r, registrant_id|
                                   managers.delete(r.user)
                                   r.user.to_wcif(self, r, registrant_id, authorized: authorized)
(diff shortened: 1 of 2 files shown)


gregorbg (changes requested):
Formally blocking this implementation until the design discussion in the accompanying WCIF PR has reached a conclusion

timreyn:
Per the discussion on thewca/wcif#15, I switched this to just use "accepted" rather than a new status.

I did not add a "roles: ['staff']", since it feels wrong to include that only for non-competing staff. I think that if this is desired, it should be through a more systematic marking of which competitors are staff.

Another downside is that this includes non-human accounts like CubingUSA Treasurer. As far as I know there is nothing in the database to indicate which accounts are humans, but I filed #8011 to track this.

gregorbg (commented):
Why are non-competing staff registrations not marked as `accepted?` anyways? Shouldn't they be accepted in the competitor waitlist anyways, making this a zero line code change?

timreyn:
@gregorbg : There's a few reasons I'd prefer not tracking `accepted` for staff currently:

1. If in the future we have a formal UI for registering to be staff, then I think it would make sense to have a parallel accepted / waitlist / denied structure for staff, which would include the possibility of someone being accepted to compete but denied to staff. There are lots of considerations involved in that, such as whether that should be a separate `staff` table, versus a single entry for both staff and volunteers, or if not then making sure all of the possible state transitions are handled reasonably.
2. In the meantime, marking staff as `accepted?` (the same field as competitors) seems like it's likely that many different checks would need to be updated. e.g. in `competition.rb`, `registration_full?` and `registration_full_message` both reference the competitor limit. I'm worried that this would need to touch a lot of things, and that we'd miss something.
3. I think that, without a formal UI for managing staff on the WCA site, tracking staff as being accepted / waitlisted / denied on the WCA site (and doing all the work involved in 2) is overkill. Any competition that wants to push non-competing staff to the WCA site will probably be tracking the staff list (and accepted / waitlisted / denied) separately e.g. in Google Sheets.

Thoughts?

(holding off on adding isCompeting until the WCIF change is approved + merged)

gregorbg:
Point number 2 is especially convincing. It is worth noting though that we are building the new registration microservice and UI with exactly the type of considerations from (1) in mind. I will approve the WCIF schema change very soon, feel free to start working on the implementation.

timreyn:
Great! Looks like you merged the WCIF PR. I added the isCompeting field as discussed over there, and I also updated `update_persons_wcif` to only add registrations that have this field set.

Good to hear this is the kind of thing that the new registration flow is hoping to support! I also think the schema questions in (1) are worth thinking through as part of that process.

timreyn:
(the failures look unrelated to my change)

gregorbg:
> (the failures look unrelated to my change)

Yup, sorry! Restarted the CI, this is an old choke that we hope to fix once we're off `structure.sql` (which is possible now that PHP is dead, we just need to find time to actually do the change)

WcaOnRails/app/models/competition.rb
```diff
@@ -1771,7 +1771,8 @@ def update_persons_wcif!(wcif_persons, current_user)
     wcif_persons.each do |wcif_person|
       local_assignments = []
       registration = registrations.find { |reg| reg.user_id == wcif_person["wcaUserId"] }
-      # If no registration is found, assume that this is a non-competing staff member being added.
+      # If no registration is found, and the Registration is marked as non-competing, add this person as a non-competing staff member.
+      next unless registration || !wcif_person["registration"] || wcif_person["registration"]["is_competing"] != false
```
gregorbg, line 1775: Rails has a nifty shorthand to make this more readable: `registration.present?` works even on `nil` values and essentially returns whether a value is nil or not (plus some other very arcane ActiveRecord edge cases which are not relevant here anyways)

gregorbg, line 1775: First part of the OR is obvious.

Third part should use `wcif_person["registration"]["isCompeting"]` in camel case instead of `is_competing` snake case. Also, at this point in time the WCIF has already been schema-checked (this is actually the first thing that happens when hitting the PATCH endpoint) so I believe it is fair to assume that `isCompeting` is always a boolean anyways. So does explicitly checking `!= false` make sense?!

In a wider sense, what about reworking this into `next if !registration.present? && wcif_person["registration"] && !wcif_person["registration"]["isCompeting"]`? I find it much easier to understand than the implicitly negated form baked into the `unless`, although that might just be personal preference. Totally open to debate and to being convinced of the current `unless` version.

WcaOnRails/app/models/competition.rb
```diff
@@ -1642,7 +1642,7 @@ def persons_wcif(authorized: false)
                                 .includes(includes_associations)
                                 .to_enum
                                 .with_index(1)
-                                .select { |r, registrant_id| authorized || r.accepted? }
+                                .select { |r, registrant_id| authorized || r.accepted? || !r.is_competing? }
```
gregorbg, line 1645: Might be worth adding a comment here why this extra `is_competing` check is necessary, for people who don't know that we (currently) don't mark non-competing staff as "accepted" in the UI.

WcaOnRails/app/models/registration.rb
```diff
@@ -204,13 +204,14 @@ def to_wcif(authorized: false)
     {
       "wcaRegistrationId" => id,
       "eventIds" => events.map(&:id).sort,
-      "status" => if accepted?
+      "status" => if accepted? || !is_competing?
```
gregorbg, line 207: Might be worth a comment, see above. On second thought, might even be worth a refactor into its own method in `Registration`, along the lines of `attending?`

gregorbg (changes requested):
The only definitive change that we need is to use `isCompeting` camel case when reading from the WCIF. The rest of my remarks are opinions about code style which are open to debate.

WcaOnRails/app/models/competition.rb
gregorbg: Hold on, now I confused myself with that suggestion about reworking the `next` condition. What are we even trying to achieve here? Sorry :see_no_evil: :woozy_face:

timreyn: Oops, fixing to isCompeting.

What are we trying to achieve? The idea was to go back to the old behavior (see base side of https://github.com/thewca/worldcubeassociation.org/pull/7601/files) unless the app PATCHing explicitly says isCompeting: false. This means all the implicitly-added organizer entities won't show up as non-competing registrations anymore.

I was trying to explicitly handle a missing isCompeting differently than a set-to-false isCompeting (does the schema validation fail if isCompeting is missing?) My assumption was that missing isCompeting meant the app writer didn't consider that field, and set-to-false means they considered it and said "yes this person is not competing", but both nil and false would be treated as false. Thoughts?

Agree with reversing the negatives.

timreyn (commented):
(responding to one comment now before I make changes + test)

WcaOnRails/app/models/competition.rb
gregorbg: The idea makes sense. So what about a two-step approach?
```ruby
# ...from before, nothing new:
registration = registrations.find { |reg| reg.user_id == wcif_person["wcaUserId"] }

adding_non_competing = wcif_person["registration"].present? && wcif_person["registration"]["is_competing"] == false
registration ||= ...whatever... if adding_non_competing

# similar to the old status quo from the PR you referenced, but with a tiny bit more readable `.present?` now
next unless registration.present?
```

WcaOnRails/app/models/competition.rb
timreyn: Done.

WcaOnRails/app/models/competition.rb
timreyn: OK, done! Rubocop doesn't like `registration ||= ... if adding_non_competing`, so I went with a normal if statement to appease it.

timreyn: Done.

WcaOnRails/app/models/registration.rb
timreyn: Moved into its own method.

timreyn (commented):
Thanks Gregor!

WcaOnRails/app/models/competition.rb
gregorbg: Now you have an extra method below but you don't use it up here...
What about `.select { |r, registrant_id| authorized || r.wcif_status == "accepted" }`?

timreyn: Sure, done.

gregorbg (approved):

PR #8005, opened 2023-06-22
Changes to the Recognition of Regions

gregorbg:
As per Board of Directors request via email. The WST is not making any political statements about China or Chinese diplomacy. We are acting on a binding instruction of our supervising Board authority.

Please direct all relevant discussions and opinions towards the public or internal mail threads. Discussion comments under this PR that go beyond the technical implementation will be disregarded.


WcaOnRails/config/locales/ja.yml
```diff
@@ -40,6 +40,8 @@ ja:
     SZ: エスワティニ
     #original_hash: 4d206a8
     TT: トリニダード・トバゴ
+    #original_hash: fefc818
+    TW: 中華台北
```
hanwu85, line 44:
```suggestion
    TW: チャイニーズタイペイ
```
This is how Japanese use in sports broadcast.

gregorbg:
I don't want to force anybody to take the burden of approving this PR. Just considering it Board-approved for the sake of deployment.

PR #8007, opened 2023-06-24
Fixes #2969 made delegate notification to deliver_later

### LINKED ISSUE #2969: Rewrite DelegateStatusChangeMailer.notify_board_and_wqac_of_delegate_status_change to not depend on `deliver_now` semantics
When @AlbertoPdRF and I worked on the DelegateStatusChangeMailer, we didn't consider the fact that `deliver_later` semantics means that all of the fancy `.delegate_status_before_last_save` attributes and the like that we were using would not work.

As a quick fix, I've changed our code to deliver immediately, rather than in a background job. That should fix the issue, but we should really rewrite our mailer to be safe to run in the background.

(As a side note: I'm not in love with how ActiveMailer's background job semantics are to put the entire mailer function call in the background, I think it would be better for just the SMTP action itself to be what happens in the background. I don't know if it's possible to configure ActiveMailer to behave that way, though.)

(As another side note: It would be good if we could change our email mailer specs to behave more like how `deliver_later` behaves in production, by actually looking up fresh rows from the database. That way, we would have at least caught this issue when writing our tests instead of in production.)

(As yet another side note: shame on @AlbertoPdRF and I for not actually testing this out on our staging server. We would have caught this there as well if we had just tried it out.)

danieljames-dj:


WcaOnRails/app/controllers/users_controller.rb
```diff
@@ -170,8 +170,16 @@ def update
...
     if @user.update(user_params)
       if @user.saved_change_to_delegate_status
-        # TODO: See https://github.com/thewca/worldcubeassociation.org/issues/2969.
-        DelegateStatusChangeMailer.notify_board_and_assistants_of_delegate_status_change(@user, current_user).deliver_now
+        if @user.delegate_status
+          if @user.delegate_status == "senior_delegate"
+            @user_senior_delegate = @user
+          else
+            @user_senior_delegate = @user.senior_delegate
+          end
```
gregorbg, line 178: This might be worth refactoring into a convenience method under `user`:

```ruby
def senior_or_self
  return nil unless self.delegate_status.present?
  
  self.delegate_status == "senior_delegate" ? self : self.senior_delegate
end
```

WcaOnRails/app/controllers/users_controller.rb
```diff
@@ -170,8 +170,16 @@ def update
...
+        else
+          @user_senior_delegate = User.find_by_id!(@user.senior_delegate_id_before_last_save)
```
gregorbg, line 180: This _should_ be the same (haven't tested, but pretty sure about the standard ActiveRecord APIs) as `User.find!` (where any model's `find` method takes the model's primary key as search criterion by default)

gregorbg (changes requested):
Cool! Just a few comments about code style, otherwise good to go!

WcaOnRails/app/controllers/users_controller.rb
danieljames-dj: Done.

WcaOnRails/app/controllers/users_controller.rb
danieljames-dj: Done.

danieljames-dj (commented):
Thanks @gregorbg for the review. I've made the changes as you suggested.

gregorbg (approved):

Issue #7853, opened 2023-04-28
Adjust advancement text for top X to "up to top X"


coder13:
**Is your feature request related to a problem? Please describe.**
Problem: parents tend to get confused reading the WCA website and reading "Top 90" as being fixed and not changing.
When a significant number of no shows happen, the top 90 can change to something smaller such as top 80. This will confuse parents.

**Describe the solution you'd like**
Adjust text around the advancement of Top X to "up to top x"

dunkOnIT:
Personally I think this makes a lot of sense. Great first issue for someone looking to just get to grips with navigating the codebase.

PR #8008, opened 2023-06-26
Change "Top X" to "Up to Top X" In Events Tab #7853

Carter-Thomas:


coder13:
Can you provide a screenshot? It'd be nice if screenshots can be provided for FE tickets.

coder13:
I think we can also benefit from having some explanation text explaining the caveat for the advancement

image

E.g. something like "a minimum of 25% of participants must be eliminated going into the next round."

@gregorbg thoughts?

Carter-Thomas:
Sure!

Proposed: 
Screenshot (447)

Current:
Screenshot (448)

WcaOnRails/app/controllers/registrations_controller.rb
```diff
@@ -221,7 +221,8 @@ def do_add
     registration_row[:wca_id]&.upcase!
     registration_row[:email]&.downcase!
     person_details = {
-      name: registration_row[:name],
+      first_name, last_name = registration_row[:name].split(' ', 2),
```
coder13, line 224: Is this supposed to be in this PR?

WcaOnRails/config/locales/en.yml
coder13, line 1: Is this supposed to be in this PR?

WcaOnRails/app/controllers/registrations_controller.rb
Carter-Thomas: No, it's getting randomly added for whatever reason

gregorbg: Then just rebase on master and make sure that it's not included in this PR :)

WcaOnRails/config/locales/en.yml
gregorbg: I agree that the changes to the mail subjects should be part of a separate PR

gregorbg (changes requested):
Thank you for your contribution! Please clean up the unrelated changes in the git diff

gregorbg:
Was it your intention to close this PR?

PR #8016, opened 2023-06-29
Added filter because sorting competition events in import results

danieljames-dj:


gregorbg (changes requested):
Instead of silently filtering, you should add explicit warnings if a JSON contains result data for an event that doesn't exist on the WCA website

danieljames-dj:
Hi @gregorbg, copy-pasting my reply in Slack here so that @thewca/wrt can also join the discussion:

In case where there are no participants in an event, how we usually do is, the event should be selected in manage events, but with 0 rounds, so that in future we can know that the competition had that event with 0 participants. The event won't show up in events section because the events tab show only events with at least 1 round.

gregorbg:
Quote from Sébastien on the relevant email thread:

> "However, [...] what you state on Github is false. We have simply removed non-held events for good so far and should just continue to do so in the future."

This is in line with my suggestion about adding an explicit warning. We might even make that warning an error, so that Delegates are blocked from uploading results for non-held events altogether.

danieljames-dj:
Thanks @gregorbg and sorry for the confusion. I've created a Github issue to show error in case the Delegate uploads results without removing events: https://github.com/thewca/worldcubeassociation.org/issues/8050.

On that note, I'm closing this PR.

PR #8058, opened 2023-07-01
Upgrade Shakapacker to v7

gregorbg:
As per https://github.com/shakacode/shakapacker/blob/master/docs/v7_upgrade.md


gregorbg:
Treating as (glorified) Dependabot update

PR #8064, opened 2023-07-02
Optimise WCIF endpoints

jonatanklosko:
Roughly 2-3x improvement for GET and ~5-8x for PATCH, as from my local tests (tested with a couple larger comps and no results).

The PATCH improvement concerns competitions with `allow_registration_without_qualification = false`, which is both competitions without qualification and competitions with qualification where enforcing is enabled.

### DIFF
--- a/WcaOnRails/app/models/competition.rb
+++ b/WcaOnRails/app/models/competition.rb
@@ -1637,6 +1637,7 @@ class Competition < ApplicationRecord
       { user: {
         person: [:ranksSingle, :ranksAverage],
       } },
+      :wcif_extensions,
     ]
     persons_wcif = registrations.order(:id)
                                 .includes(includes_associations)
@@ -1734,7 +1735,7 @@ class Competition < ApplicationRecord
 
     # Create missing events.
     wcif_events.each do |wcif_event|
-      event_found = competition_events.find_by_event_id(wcif_event["id"])
+      event_found = competition_events.find { |ce| ce.event_id == wcif_event["id"] }
       event_to_be_added = wcif_event["rounds"]
       if !event_found && event_to_be_added
         unless current_user.can_add_and_remove_events?(self)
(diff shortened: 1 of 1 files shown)


WcaOnRails/app/models/competition.rb
```diff
@@ -1637,6 +1637,7 @@ def persons_wcif(authorized: false)
       { user: {
         person: [:ranksSingle, :ranksAverage],
       } },
+      :wcif_extensions,
```
jonatanklosko, line 1640: Without preloading we emitted a single additional query per registration.

WcaOnRails/app/models/competition.rb
```diff
@@ -1734,7 +1735,7 @@ def set_wcif_events!(wcif_events, current_user)
 
     # Create missing events.
     wcif_events.each do |wcif_event|
-      event_found = competition_events.find_by_event_id(wcif_event["id"])
+      event_found = competition_events.find { |ce| ce.event_id == wcif_event["id"] }
```
jonatanklosko, line 1738: We call `self.competition_events` above which loads the whole list into memory. By using `#find` we can do lookup in that list, whereas `find_by_event_id` would **always** make a query.

Same below.

WcaOnRails/app/models/competition.rb
```diff
@@ -1783,7 +1784,9 @@ def update_persons_wcif!(wcif_persons, current_user)
       # NOTE: person doesn't necessarily have corresponding registration (e.g. registratinless organizer/delegate).
       if wcif_person["roles"]
         roles = wcif_person["roles"] - ["delegate", "trainee-delegate", "organizer"] # These three are added on the fly.
-        registration.update!(roles: roles)
+        # The additional roles are only for WCIF purposes and we don't validate them,
+        # so we can safely skip validations by using update_attribute
+        registration.update_attribute(:roles, roles)
```
jonatanklosko, line 1789: Validating a registration involves a couple additional queries. We could avoid some of them via preload, but if the competition has qualifications we make a query checking against competitor's prior results, which we can't preload reasonably. In this case we only want to update the roles and can skip the validations altogether.

coder13:
Great stuff! I'm in no place to review the improvements but they seem to make sense to me!

jonatanklosko:
The CI fails with unrelated tests, same as on main, so I assume it's expected.

timreyn:
Thanks so much, @jonatanklosko !!

gregorbg (approved):
Thanks for a couple of short and concise but important optimizations!

Issue #7838, opened 2023-04-18
Make the subject for the email more distinct for competitors being accepted or deleted from a competition


Sixstringcal:
Currently, the only difference in the subject is the last word.  On my phone, that cuts off.  It's probably best if we could change the email to something else that makes it more obvious that their registration has been deleted.

Another proposal would be that if the registration period has ended, a different email is sent out explaining that they did not make it off the waiting list if their registration is being deleted from the waiting list.  I currently email competitors that information, but to avoid multiple emails being sent, possibly spamming their inbox, and probably not even being read anyway, I keep them on the wait list (I've seen other people do this as well, I know I'm in the minority, but as far as I can tell, it's not an incorrect procedure).  That way they get the email I want them to see.

 Another alternative would be to allow organizers to make their own email for it to send/allow emails to not be sent after the deadline to be accepted off the waiting list.

It's probably more work than it seems to be worth, but I figured it might be a good idea.

dunkOnIT:
I'd suggest going with changing the title of the email - something like "Registration Accepted: {Competition Name}" or "Registration Deleted: {Competition Name}" would be a lot clearer in an email inbox. 

Could also consider using the term "rejected" instead of "deleted", as "deleted" implies user action (to me at least).

PR #8074, opened 2023-07-05
More Distinct Email Headers (#7838)

Carter-Thomas:
Changed wording of email headers


gregorbg:
You need to adjust the tests that verify the mail subject. They still expect the old subject line in the test cases.

dunkOnIT:
- [ ] Staff would need to be notified of change
- [ ] We need to figure out how this would effect spam filtering - are there any services we can use to "verify" these things?

Carter-Thomas:
This is probably a newbie question but how does spam filtering work in this repo? What files would I have to change in order to get this to work?

dunkOnIT:
Thanks for that @Carter-Thomas! I am going to email the WCA staff to solicit feedback on the email header changes, and once we have those this should be ready to merge.

WcaOnRails/config/locales/en.yml
```diff
@@ -1179,10 +1179,10 @@ en:
       #context: for a new registration
       new:
         #context: "comp_name" contains the competition's name
-        mail_subject: "Your registration for %{comp_name} is submitted and pending approval"
+        mail_subject: "Registration Pending: %{comp}"
```
Goosly, line 1182: I think this won't work since the variable is %{comp_name}, not %{comp} ?

Carter-Thomas: I believe that would be correct. @dunkOnIT

dunkOnIT: Correct! Will change this when feedback concludes on Monday (I expect the tests will not pass without it, but I'll leave this conversation unresolved as well)

dunkOnIT:
@Carter-Thomas looks like this is on the master branch in your fork, and you've just merged an unrelated change. Would you be able to create a separate branch for any other changes you're working on? 

(Usually I'd suggest moving this PR to a new branch, but given that we expect to move it soon it isn't necessary - but you're welcome to do that as well if you'd like)

Shout if you need any help there!

Carter-Thomas:
Oops, completely my mistake, will deal with this soon!

Carter-Thomas:
Ok I undid the commit related to the other issue in the master branch, so everything should be the exact same as thewca/master with the exception of the edited en.yml file. Let me know if this isn’t the case, I’m not too used to working with several branches yet.

dunkOnIT:
Looks good, thanks for handling that!

WcaOnRails/config/locales/en.yml
```diff
@@ -1168,7 +1168,7 @@ en:
       updated: "Updated registration"
       failed: "Could not update registration"
       deleted: "Successfully deleted your registration for %{comp}"
-      registered: "Successfully registered!"
+      registered: "Registration submitted - make sure to follow payment instructions to avoid disappointment!"
```
gregorbg, line 1171: What if a competition does not have payment instructions?

WcaOnRails/spec/features/register_for_competition_spec.rb
```diff
@@ -16,7 +16,7 @@
       visit competition_register_path(competition)
       check "registration_competition_events_333"
       click_button "Register!"
-      expect(page).to have_text("Successfully registered!")
+      expect(page).to have_text("Registration submitted - make sure to follow payment instructions to avoid disappointment!")
```
gregorbg, line 19: see above.

gregorbg (changes requested):

WcaOnRails/config/locales/en.yml
```diff
@@ -1168,7 +1168,7 @@ en:
       updated: "Updated registration"
       failed: "Could not update registration"
       deleted: "Successfully deleted your registration for %{comp}"
-      registered: "Registration submitted - make sure to follow payment instructions to avoid disappointment!"
+      registered: "Registration submitted - make check for any payment instructions to avoid disappointment!"
```
gregorbg, line 1171: Typo / grammar mistake

gregorbg (changes requested):

WcaOnRails/config/locales/en.yml
gregorbg: I also think it makes sense to explicitly mention Registration Requirements here, because they are such an important and "official" part of how the WCA sanctions competitions

gregorbg: I've made a separate in-code suggestion above

WcaOnRails/config/locales/en.yml
dunkOnIT: Suggest we keep it as "payment instructions" as it is correct 99%+ of the time, and I think we save organizers more work by drawing attention to payment requirements for the average competition than if they need to follow up for other registration requirements for outlier competitions

dunkOnIT: Can go back to the staff for feedback though

dunkOnIT: @gregorbg wording has been updated based on staff feedback in "FEEDBACK REQUEST: Competition Notification Email Changes"

gregorbg (approved):

PR #8079, opened 2023-07-06
Allow updating the competitor limit within the WCIF

Rubiks-boy:
This PR adds support for updating a competition's `competitor_limit` by passing a patch request into the WCIF. It also includes automated tests.

Note that this only supports updating the competitor limit from one number to another. Because having a competitor limit requires adding a `competitor_limit_reason` that is not part of the WCIF, this does not support adding or removing a competitor limit (ex. passing in `competitorLimit: null`).

### Motivation

As part of a [competition schedule generator](https://competition-scheduler.netlify.app/) I'm working on, I want to be able to update the competitor limit.

(I acknowledge the `competitor_limit_reason` stuff is semi-clunky; I'm happy to send a separate POST to `/competitions/<competition_id>` and close the PR if that is preferable)

### DIFF
--- a/WcaOnRails/app/models/competition.rb
+++ b/WcaOnRails/app/models/competition.rb
@@ -1671,6 +1671,7 @@ class Competition < ApplicationRecord
       set_wcif_schedule!(wcif["schedule"], current_user) if wcif["schedule"]
       update_persons_wcif!(wcif["persons"], current_user) if wcif["persons"]
       WcifExtension.update_wcif_extensions!(self, wcif["extensions"]) if wcif["extensions"]
+      set_wcif_competitor_limit!(wcif["competitorLimit"], current_user) if wcif["competitorLimit"]
 
       # Trigger validations on the competition itself, and throw an error to rollback if necessary.
       # Context: It is possible to patch a WCIF containing events/schedule/persons that are valid by themselves,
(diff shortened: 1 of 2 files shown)


WcaOnRails/app/models/competition.rb
```diff
@@ -1681,6 +1682,22 @@ def set_wcif!(wcif, current_user)
     end
   end
 
+  def set_wcif_competitor_limit!(wcif_competitor_limit, current_user)
+    unless current_user.can_admin_competitions? || !confirmed?
```
coder13, line 1686: Only thing I was worried about when I saw this PR. I appreciate the checks. I think it should be fine to change any field as long as it's the right time still.

gregorbg, line 1686: I think you should separate these two checks to provide clear error messages.

if not admin: "Cannot edit the competition."
if confirmed: "Cannot edit the competitor limit because the competition has been confirmed by WCAT."

Rubiks-boy: Thanks! I updated the error message to `"Cannot edit the competitor limit because the competition has been confirmed by WCAT."`. However, I originally intended this as a single check: if it's confirmed, and the person isn't an admin, raise the error. This would let the WCAT change the competitor limit after a comp is confirmed. Does this seem reasonable?

As a follow-up, I got a big confused about user permissions: do I need to perform a separate check here? If so, is `admin` the correct check, or should I be doing `can_manage_competitions` or something else?

WcaOnRails/app/models/competition.rb
```diff
@@ -1681,6 +1682,22 @@ def set_wcif!(wcif, current_user)
     end
   end
 
+  def set_wcif_competitor_limit!(wcif_competitor_limit, current_user)
+    if confirmed? && !current_user.can_admin_competitions?
```
gregorbg, line 1686: `can_admin_competition` is the right permission here :)
Using something like `can_manage_competition?(self)` will include competition Delegates and their Senior delegates, whereas the former only includes global results admins (WST and WRT) as well as WCAT

gregorbg (commented):
Some more thoughts :)

WcaOnRails/spec/models/competition_wcif_spec.rb
```diff
@@ -18,6 +18,8 @@
       event_ids: %w(333 444 333fm 333mbf),
       with_schedule: true,
       competitor_limit: 50,
+      registration_open: "2013-12-01",
+      registration_close: "2013-12-31",
```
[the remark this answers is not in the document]
Rubiks-boy: Without this, I was getting a validation error, since the `registration_open`/`registration_close` are [54 weeks ago](https://github.com/thewca/worldcubeassociation.org/blob/master/WcaOnRails/spec/factories/competitions.rb#L99-L100), whereas the `start_date` earlier in this file is 2014.

gregorbg (approved):

PR #8080, opened 2023-07-06
Clean up and unify docker-compose setup

gregorbg:
Because we might want to add many more cool services in the future.
And because the way we waited for MySQL to be ready was hacky.


FinnIckler:
Can you also give every service a `container_name` attribute, just so the output in the terminal is better?

FinnIckler:
Love the healthcheck, I should do the same in https://github.com/thewca/wca-registration/blob/63ac10f52a5a48469715863af33a2a1a44b152bd/docker-compose.dev.yml#L68

FinnIckler (approved):
LGTM

PR #8081, opened 2023-07-07
Moved Disclaimer and 404 to react router

danieljames-dj:


gregorbg:
I am not sure whether introducing React Router at this point in time is a good idea. We will probably eventually end up using it some time in the future, but the codebase doesn't feel ready yet.

One of the reasons why we are in the current mess is that people introduced new technologies before fully commiting to them and migrating all old technologies.
- Jeremy thought Rails was cool but introduced it without fully migrating away from PHP. This took us 15 years to fix,
- Philippe(?) thought React was cool but introduced it without fully migrating away from the Sprockets asset pipeline. This is one of the major reasons why we now have two frontend stacks
- Somebody (legitimately don't know who) thought introducing Semantic UI would be cool without migrating the existing React pages away from Bootstrap. So even _within_ React, we use different frontent technologies.

Due to the reasons outlined above, I oppose introducing another new technology before the old one is completely migrated away. If you have any strong arguments in favor of React Router then please let me know, it's not very likely that you will be able to change my mind due to the complex nature and sheer scale of our migration projects, but I still want to be open for debate.

danieljames-dj:
Hi @gregorbg, thanks for your reply. I think I've a very strong point supporting react-router. You have mentioned "the current mess is that people introduced new technologies before fully commiting to them and migrating all old technologies". Currently, we are in process of migrating erb code to react and introducing react-router is for that. The current migration process will always leave a `.html.erb` file in the views folder - ie, the erb is not "fully" getting migrated. If we have a react-router, we can migrate erb to react completely and delete the erb file. Deleting the erb file in the process of migration will reduce the number of erb files, and hence will motivate people to finish off the migration. Maybe the number of erb files will be a good metric in that case to know what is the progress of react migration.

Another point which I would like to say is, we have to implement react-router some day if we want to split frontend. If we do at the end for all components together, it can be really difficult. But if we are doing part by part, it will definitely be easy.

gregorbg:
I'm with Finn on this one. We need to think about our entire WCA architecture before we can choose a technology that best suits this architecture.

Also, metrics about "being able to delete the .erb file" are never a good idea to motivate change in my opinion.

PR #8083, opened 2023-07-07
Fix delegate mailer using ActiveRecord::Dirty attribute on a delay

gregorbg:
Fixes a problem reported by WEAT via email


danieljames-dj (approved):
LGTM, thanks @gregorbg for fixing this.

EDIT: Just now noticed failed test case, you might probably know it already that `user_controller_spec` changes were missing.

PR #8084, opened 2023-07-07
Fix failure during docker compose up

viroulep:
I just ran a `docker compose up` on a fresh repo and ran into a failure: I believe it's linked to #7929.

Interestingly there was also some changes to `Gemfile.lock`: the new diff looks good (as in: it's ordered alphabetically), but I don't how it ended up the way it is now on master.


gregorbg:
The Gemfile quirk was caused by Finn and me being sloppy during a manual rebase of the PR that introduced Redis. It is also being fixed in https://github.com/thewca/worldcubeassociation.org/pull/8080 :innocent:

gregorbg (approved):

viroulep:
Updated :)

viroulep:
Thanks for the fast review/merge!

PR #8087, opened 2023-07-08
Switch from Supervisor/DelayedJob to Sidekiq for job processing

gregorbg:
One step closer to moving away from Chef. Allows us much more flexibility in terms of job monitoring, middleware, and scheduling. Also lets us run Ruby cron jobs within Ruby and rely on OS crontab less.


dunkOnIT:
- [ ] Figure out what the status is
- [ ] PR Review
- [ ] Discuss with Gregor/make changes as necessary

FinnIckler:
TODO:
- [ ] Separate REDIS_URL for caching and sidekick
- [ ] Create RDS to S3 Backup job instead 
- [ ] Delete Delayed Jobs Tables
- [ ] CompletedJobs Table <- Do we still need this? @jfly any comments? Was this ever useful?
- [ ] isRunning? as method for Jobstatistics class?
- [ ] Move Middle Ware to ActiveJob

jfly:
I haven't checked the latest code, but we `CompletedJobs` table just as a debugging tool (to be able to answer questions like "how many emails have we sent?").

FinnIckler:
Don't think anything has changed in that regard. The thing is that this table is by far the biggest table in our database and we are getting metrics like "how many emails have we sent" quite easily out of the ses dashboards. So it doesn't really feels necessary anymore?

jfly:
That sounds reasonable to me. If y'all aren't using it, then it's probably OK to let it go.

FinnIckler:
I have moved the RDS Backup Job to native S3 Snapshots, I will schedule them weekly once this is merged

FinnIckler (approved):
Successfully tested on staging

dunkOnIT:
Awesome work guys, really exciting :D

Issue #7820, opened 2023-04-06
Update text in yellow box on "edit profile" page to email WRT, not delegates


KitClement:
When editing a profile that has a WCA ID assigned, competitors are presented with this message in a yellow box:

"You cannot change your name, birthdate, gender, or country. Reason: you have a WCA ID assigned. Contact your [Delegate](https://www.worldcubeassociation.org/delegates) if you need to change any of these."

However, delegates do not have access to change this personal info. In fact, the current regulations suggest reaching the results team through the contact form:

"[2c2](https://www.worldcubeassociation.org/regulations/#2c2)) ... If a competitor wishes to change or update their personal information between competitions (e.g. a change of nationality, name, or gender), they must [contact the WCA Results Team](https://www.worldcubeassociation.org/contact/website)."

I feel like it would be best if these two statements agreed with each other on the website.

paarthchhabra:
Hi, I would like to fix this

danieljames-dj:
Hi @KitClement, WRT had a quick internal discussion regarding it. So, there are cases where the competitor cannot contact the WRT directly - like if there is language barrier. We feel it's better to change it to something like "Contact the WCA Results Team either directly or through their local Delegate".

But this is contradicting a bit to [2c2](https://www.worldcubeassociation.org/regulations/#2c2). @thewca/wrc-team may I know if it's possible to update the regulation statement of 2c2 to "Contact the WCA Results Team either directly or through their local Delegate"?

Nanush7:
Hi all,

We will discuss this and get back to you ASAP. But any change won't be done until the next amendment cycle.

Nanush7:
Hello again. We are going to implement this change to Regulation 2c2 in the next amendment cycle. ;)

danieljames-dj:
Thanks a lot @Nanush7. So, we can change it to "Contact the WCA Results Team either directly or through their local Delegate" in the edit profile page as well.

@paarthchhabra hope you are still interested to fix this issue in code, I've assigned the issue to you. Thanks for your interest again.

dunkOnIT:
Hi @paarthchhabra are you still interested in working on this? Let me know if I can do anything to help!

maxidragon:
@dunkOnIT it's already done, see PR #8090

dunkOnIT:
Ah nice thank you!

PR #8088, opened 2023-07-08
Reg2c2 editinfo update #7820

paarthchhabra:
Used google translate to change the current message to "Contact the WCA Results Team/your delegate" from "your delegate" in accordance with Reg 2c2


gregorbg:
Thank you for your contribution! We are not quite ready to trust Google Translate yet, so the flow for updating locale strings is as follows:
- You submit a PR that changes only the English string
- We merge and deploy the change
- Our code detects that the English locale (so-called "base locale") has changed and sends emails to verified translators in our community
- They submit translations at their own pace

paarthchhabra:
> 

Got it, will open a new PR that changes only the English string

gregorbg:
Superseded by https://github.com/thewca/worldcubeassociation.org/pull/8090

PR #8089, opened 2023-07-08
Switch to Ruby DSL for DB schema dumping

gregorbg:
**FINALLY** getting rid of those annoying "structure.sql" problems in the git diff due to different local hardware versions of MySQL. Eveything is abstracted away behind Ruby now.

The `DatabaseDumper` had to be changed because it was hard-coded to use `.sql` schema dumps. Now relying on the same mechanics that drive the Rails CLI instead :)


FinnIckler:
Column names have to be changed after https://github.com/thewca/worldcubeassociation.org/pull/8077 is merged right? Or will we merge this first and then change it in 8077?

gregorbg:
This PR does not change any column names, so in fact it is completely independent from #8077. They can be merged either way around.

viroulep:
Really glad to see this PR! Thanks a lot!

FinnIckler:
If I want to test this, I still use the same rake tasks as before right?

gregorbg:
Yup!

WcaOnRails/config/database.yml
```diff
@@ -27,6 +36,13 @@ development:
     <<: *default
     database: wca_development
     replica: true
+  developer_dump:
+    <<: *default
+    database: wca_dump_developer
```
FinnIckler, line 41: why do these two not have database_tasks: false, but the other dump dbs all have it?

FinnIckler (approved):
Looks good, just one question in the database.yml

FinnIckler:
Also: does this require any special deployment?

danieljames-dj (approved):
LGTM

WcaOnRails/config/database.yml
gregorbg: Because in local development, it makes sense to be able to operate on the dump databases (for example, if you try to develop a new feature for the dev dump). On test and production environments, dropping, resetting, altering or otherwise migrating the temporary dump databases does not make sense.

WcaOnRails/db/results_dump_schema.rb
```diff
@@ -0,0 +1,153 @@
...
+    t.string "sorry_message", limit: 172, default: "", null: false, collation: "utf8mb3_general_ci"
+  end
+
+  create_table "Scrambles", id: false, charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
+    t.integer "scrambleId", default: 0, null: false, unsigned: true
+    t.string "competitionId", limit: 32, null: false
+    t.string "eventId", limit: 6, null: false
+    t.string "roundTypeId", limit: 1, null: false
+    t.string "groupId", limit: 3, null: false
+    t.boolean "isExtra", null: false
```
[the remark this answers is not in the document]
gregorbg: No. The `boolean` type is a type within the Rails ORM. Internally, in MySQL, this type is implemented as `TINYINT(1)` and Rails is smart enough to interpret this as a boolean internally. But on the raw data layer, nothing changes and no migration is necessary.

PR #8095, opened 2023-07-14
Include model information in the search cache key

gregorbg:
(sidenote regarding the use of model classes as cache keys: Rails `ActiveRecord` classes respond to `to_param`, which simply yields the class name as string: `Competition.to_param --> "Competition"`. This matches the specification for cache keys at https://guides.rubyonrails.org/caching_with_rails.html#cache-keys)


FinnIckler (approved):
Thanks for catching that!

PR #8100, opened 2023-07-19
Add Conditional GET to public wcif endpoint

FinnIckler:
We have seen some crawling on public wcif endpoints kill our servers over the last ~2 weeks and this would be one version to remedy this: 
Additionally to caching the endpoint for repeated request from multiple users, we can add a conditional get request to cache multiple requests for the same user, even if they are more than 5 minutes old.

Other solutions we could think about: 
- using a json serializer that caches automatically on last_modified like https://github.com/codenoble/cache-crispies instead of checking manually
- blocking certain user-agents from this route
- redirecting routes to use api.worldcubeassociation.org if they are not (so cloudfront can do the caching)
- precomputing wcif for every competition and updating on patch
- ???

### DIFF
--- a/WcaOnRails/app/controllers/api/v0/competitions_controller.rb
+++ b/WcaOnRails/app/controllers/api/v0/competitions_controller.rb
@@ -109,11 +109,12 @@ class Api::V0::CompetitionsController < Api::V0::ApiController
     id = params[:competition_id] || params[:id]
     cache_key = "wcif/#{id}"
     expires_in 5.minutes, public: true
-    render json: Rails.cache.fetch(cache_key, expires_in: 5.minutes) {
-      competition = competition_from_params
-      fresh_when(:etag => competition, :last_modfied => competition.updated_at, :public => true)
-      competition.to_wcif
-    }
+    if stale?(last_modified: competition.updated_at, public: true)
+      render json: Rails.cache.fetch(cache_key, expires_in: 5.minutes) {
+        competition = competition_from_params
+        competition.to_wcif
+      }
+    end
   end
 
   def update_wcif
--- a/WcaOnRails/app/controllers/api/v0/competitions_controller.rb
+++ b/WcaOnRails/app/controllers/api/v0/competitions_controller.rb
@@ -108,10 +108,10 @@ class Api::V0::CompetitionsController < Api::V0::ApiController
   def show_wcif_public
     id = params[:competition_id] || params[:id]
     cache_key = "wcif/#{id}"
+    competition = competition_from_params
     expires_in 5.minutes, public: true
     if stale?(last_modified: competition.updated_at, public: true)
       render json: Rails.cache.fetch(cache_key, expires_in: 5.minutes) {
-        competition = competition_from_params
         competition.to_wcif
       }
     end
--- a/WcaOnRails/app/controllers/api/v0/competitions_controller.rb
+++ b/WcaOnRails/app/controllers/api/v0/competitions_controller.rb
@@ -110,7 +110,7 @@ class Api::V0::CompetitionsController < Api::V0::ApiController
     cache_key = "wcif/#{id}"
     competition = competition_from_params
     expires_in 5.minutes, public: true
-    if stale?(last_modified: competition.updated_at, public: true)
+    if stale?(competition, public: true)
       render json: Rails.cache.fetch(cache_key, expires_in: 5.minutes) {
         competition.to_wcif
       }
--- a/WcaOnRails/app/controllers/api/v0/competitions_controller.rb
+++ b/WcaOnRails/app/controllers/api/v0/competitions_controller.rb
@@ -107,11 +107,11 @@ class Api::V0::CompetitionsController < Api::V0::ApiController
 
   def show_wcif_public
     id = params[:competition_id] || params[:id]
-
     cache_key = "wcif/#{id}"
     expires_in 5.minutes, public: true
     render json: Rails.cache.fetch(cache_key, expires_in: 5.minutes) {
       competition = competition_from_params
+      fresh_when(:etag => competition, :last_modfied => competition.updated_at, :public => true)
       competition.to_wcif
     }
   end


WcaOnRails/app/controllers/api/v0/competitions_controller.rb
```diff
@@ -107,13 +107,14 @@ def show_wcif
...
     id = params[:competition_id] || params[:id]
-
     cache_key = "wcif/#{id}"
+    competition = competition_from_params
     expires_in 5.minutes, public: true
-    render json: Rails.cache.fetch(cache_key, expires_in: 5.minutes) {
-      competition = competition_from_params
-      competition.to_wcif
-    }
+    if stale?(last_modified: competition.updated_at, public: true)
```
gregorbg, line 113: https://apidock.com/rails/ActionController/ConditionalGet/stale%3F specifies that you can simply pass the competition model itself and Rails is smart enough to figure out that it has an `updated_at` flag that can be used. Also gives you `etag` for free.

gregorbg: Also, I don't really understand the code logic behind this. You use an `if` check to figure out whether you should return something. But what if that `if` check yields `false`? What does the controller return in that case?

gregorbg: Reading the documentation, it seems as though Rails sends a `304` response if the resource is considered stale. Are we confident that our existing API users can handle this kind of response?

gregorbg (commented):
Seems like a good approach that _basically_ is identical to pre-computing WCIFs and only returning new ones when something new has been written to the DB. Just a few comments about the implications of using this approach in production :)

dunkOnIT:
I'm stroooongly in favour of precomputing WCIF/doing cache warming. I've written up some more detailed thoughts in [this doc](https://docs.google.com/document/d/1rxbauGkW5bNOIJEJ7CWecaEC2ZNcrb3hMduHJAIHAqo/edit) (it's just over a page, mostly bullet points :P), including some basic stabs at implementation - but obviously open to correction there.

I think conditional GETs do definitely have a place in our code, but:
* They may not address the vulnerability to requests that the crawling has revealed, if they only check the staleness relative to the user making the request? (My understanding here might be wrong - does `is_stale?` work on a per-user basis?)
* I agree with the concern re whether our current API consumers handle 304's - it may require API version advancement to implement this?
* If the caches for the resources being requested aren't warmed, our servers will still die

gregorbg:
Nice write-up, but completely blows the scope of this PR out of proportion. A major rewrite like the one you indicated will probably not be happening before Worlds.

> They may not address the vulnerability to requests that the crawling has revealed, if they only check the staleness relative to the user making the request? (My understanding here might be wrong - does is_stale? work on a per-user basis?)

Your understanding here is wrong. The `is_stale?` implementation works based on the `updated_at` timestamp of the model, which is a global property for that row in the database, irrespective of the user.

gregorbg:
Say there is a new competition that wants to use WCIF. Can the following scenario occur?
1. Some PATCH is written to stipulate an update to the `updated_at` column.
2. Groupifier queries the WCIF --> should get rendered, because the competition has a new `updated_at` timestamp
3. Without any changes to the competition, WCA Live wants to load the WCIF to import into scoretaking. It immediately gets 304'ed because the `updated_at` hasn't changed, so WCA Live doesn't get to see the WCIF despite the fact that it never queried it before

gregorbg:
So if a client does not send along the whole etag and modified-since shenanigans, we always send it the whole WCIF by default anyways?

FinnIckler:
yes, but every client that follows the http standard should do that

gregorbg:
But I still don't get where the code would get those ETags from. Your average pseudocode looks like this:

```
require_once 'fancy_rest_library`

def get_wcif(competition_id)
  let url = "worldcubeassociation.org/wcif/${competition_id}"
  let response = fancy_rest.get(url)

  return response.body.json
end
```

where in that (literal!) two-liner is the etag being set?

FinnIckler:
This is done by the fancy_rest_library under the hood. It will sent either a `If-None-Match: "etag"` or a `If-Modified-Since:  date` with the request. 
See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag#caching_of_unchanged_resources

gregorbg:
Okay, that means our only worry is to make Groupifier, WCA Live et al. aware that the API can now return 304 responses

FinnIckler:
Honestly I am 99% sure that the libraries they are using already now how to handle a 304. They all use javascripts fetch implementation, which knows how to do it. I am using fetch for wca-registration and I am getting 304s most of the time (when I hit api.worldcubeassocation.org or test-registration.worldcubeassociation.org which are both behind cloudfront which returns 304s)

FinnIckler:
Exactly, every library that implements Section [RFC 9111 on HTTP Caching](https://www.rfc-editor.org/rfc/rfc9111.txt) will respect the etag/not-modifed/cache-control whatever headers. I mean cURL won't if you don't tell it too, but I am pretty sure that this will cover a lot of libraries/standard library implementations (I say this knowing Ruby's implementation of [Net::HTTP](https://ruby-doc.org/stdlib-2.7.0/libdoc/net/http/rdoc/Net/HTTP.html#class-Net::HTTP-label-Setting+Headers) does not do this out of the box, but come on, it doesn't even do https without telling it to). Fetch definitely does so we have all of javascript covered.

gregorbg (approved):
Technically, the changes are fine now. But here's another twist (sorry!): I remember some weeks ago we deliberately changed it so that the competition will only be fetched if there is no cache hit, i.e. we _deliberately_ made a change to move `competition_from_params` inside the Rails cache block. Now you're undoing that change, which means you add an efficiency improvement at the cost of sacrificing another. Does the efficiency gain cancel out or are we confident that this is actually worth the change?

EDIT: https://github.com/thewca/worldcubeassociation.org/pull/7932

FinnIckler:
I can profile the route, but the performance improvements we got by moving this into the cache seem to be negligible.

FinnIckler:
There seems to be some kind of edge case where updating group assignments does not update the competition's updated_at, which causes some people to get their updated group assignment, but others not (if they hit a 304). So I reverted this manually on prod because it is causing some major headaches at us nats.

gregorbg:
The group assignments are part of the `Assignment` model in our code, which is linked to a competition through the `registrations` object. So indeed, it is possible that during PATCHing, the competition itself (in the strict sense of the DB `Competition` model) never changes, and only the associated `assignments` table changes.

In that case, the conditional GET code is correct in assuming that the competition model itself did not change. To fix this (instead of just hard-reverting the commit) we can either
- change the `stale?` check to consider _all_ associated models of a competition (which seems like a clean but overly complicated solution, because there are a _lot_ of associated models)
- OR change the WCIF PATCH code to always set a new `updated_at` timestamp (which is a bit hacky but very cheap and easy solution)

gregorbg:
Suggested fix at https://github.com/thewca/worldcubeassociation.org/pull/8126

FinnIckler:
I would also go for the 'cheap' and easy solution for now and then we evaluate any update calls where this also could happen (I think we are covering most, if not all of it with the wcif patch)

PR #8103, opened 2023-07-21
Fix translations status page

jonatanklosko:
We generally assume the translation YAML is maps and not lists, even for enumerations. Currently there is a list and this crashes the translations page and Internationalize doesn't handle it either.


FinnIckler (approved):
Thanks and TIL about who we do paragraphs. But: can you fork the repo next time?

jonatanklosko:
> But: can you fork the repo next time?

Sure, though as long as you prefix branches and delete after merging, this doesn't really matter IMO :)

PR #8108, opened 2023-07-21
Removed sorted by average option in MBLD psychsheet

### LINKED ISSUE #7981: Psych sheet for MBLD should only have the sorted by single option
Currently, when you go to the psych sheet of a competition, it shows options for sorting by single and by average, even if the event in question is MBLD. There should only be the option for sorting by single for this event.

maxidragon:
Fixes #7981 

Before:
image

After:
image


WcaOnRails/app/views/registrations/psych_sheet_event.html.erb
```diff
@@ -5,7 +5,10 @@
     <%= cubing_icon @event.id %> <%= @event.name %> Psych Sheet
   </h2>
   <p>
-  Sorted by: <%= @psych_sheet.sort_by %>, sort by <a href="<%= competition_psych_sheet_event_path(@competition, @event.id) %>?sort_by=<%= @psych_sheet.sort_by_second %>"><%= @psych_sheet.sort_by_second %></a>
+  Sorted by: <%= @psych_sheet.sort_by %>
+  <% if @event.id != "333mbf" %>
```
gregorbg, line 9: We have `event.multiple_blindfolded?` which avoids the magic string `"333mbf"` in the view code

WcaOnRails/app/views/registrations/psych_sheet_event.html.erb
```diff
@@ -5,7 +5,10 @@
...
+  , sort by <a href="<%= competition_psych_sheet_event_path(@competition, @event.id) %>?sort_by=<%= @psych_sheet.sort_by_second %>"><%= @psych_sheet.sort_by_second %></a>
```
gregorbg, line 10: I understand that you're copying something that has been there before, but touching this part of the code could be a nice incentive to clean up this ugly hack.

Rails has a `link_to` helper which generates the `a href` tags for you and which can also support query parameters: https://apidock.com/rails/ActionView/Helpers/UrlHelper/link_to

gregorbg (commented):
Logic is totally fine, just some suggestions for code cleanup :)

WcaOnRails/app/views/registrations/psych_sheet_event.html.erb
```diff
@@ -5,7 +5,7 @@
     <%= cubing_icon @event.id %> <%= @event.name %> Psych Sheet
   </h2>
   <p>
-  Sorted by: <%= @psych_sheet.sort_by %>, sort by <a href="<%= competition_psych_sheet_event_path(@competition, @event.id) %>?sort_by=<%= @psych_sheet.sort_by_second %>"><%= @psych_sheet.sort_by_second %></a>
+    Sorted by: <%= @psych_sheet.sort_by %><% if !@event.multiple_blindfolded? %>, sort by <%= link_to @psych_sheet.sort_by_second, competition_psych_sheet_event_path(@competition, @event.id, sort_by: @psych_sheet.sort_by_second) %><% end %>
```
gregorbg, line 8: `if !abc` in Rails can be written as `unless abc`

gregorbg: Chances are that our linter will even complain about this during CI checks :innocent:

gregorbg (changes requested):

gregorbg (approved):
Thanks for the lightning-fast fixes! :zap:

gregorbg (approved):

PR #8111, opened 2023-07-24
Fix psychsheet rankings for tied averages

### LINKED ISSUE #7738: Incorrect Psych Sheet Rankings for Tied Averages/Means
I believe this has been a bug ever since psych sheets were introduced on the WCA site, but I didn't see anything about it when I did a search here. My bad if this is a duplicate.

**Describe the bug**

Whenever there is a tied average or mean on a psych sheet, the ranking for the competitors are the same regardless of their singles.

**To Reproduce**

Go to any psych sheet page with tied averages/means but different singles. Example: https://www.worldcubeassociation.org/competitions/CubingUSANationals2023/registrations/psych-sheet/333fm.

**Expected behavior**

For tied averages/means, the single should be a tiebreaker when determining rankings.

**Screenshots**

This is the top of the psych sheet for FMC at CubingUSA Nationals 2023.

FMC Psych Sheet

The rankings shown are { 1, 2, 2, 2, 5, 5, 5, 5, 9 } but it should actually be {1, 2, 3, 4, 5, 6, 6, 8, 9} (note the duplicate 6) since only the 6th and 7th competitors are tied for _both_ mean and single.

**Additional Information**

The rankings are particularly important for the example above since only the top 100 competitors will qualify to compete at the competition. Having multiple rankings near the qualification cutoff can falsely imply that all competitors with a ranking less than or equal to the cutoff will qualify when they actually won't due to other competitors with the same ranking having better singles.

maxidragon:
Fixes #7738 

After changes psychsheet looks like this:
image

### DIFF
--- a/WcaOnRails/app/models/competition.rb
+++ b/WcaOnRails/app/models/competition.rb
@@ -1474,16 +1474,14 @@ class Competition < ApplicationRecord
       prev_sorted_registration = nil
       sorted_registrations = []
       registrations.each_with_index do |registration, i|
-        if sort_by == 'single'
-          rank = registration.single_rank
-        else
-          rank = registration.average_rank
-        end
-        if rank
-          tied_previous = registration.average_rank == prev_sorted_registration&.registration&.average_rank && registration.single_rank == prev_sorted_registration&.registration&.single_rank
-          pos = tied_previous ? prev_sorted_registration.pos : i + 1
-        else
-          # Hasn't competed in this event yet.
+        rank = sort_by == 'single' ? registration.single_rank : registration.average_rank
+        average_tied_previous = registration.average_rank == prev_sorted_registration&.registration&.average_rank
+        single_tied_previous = registration.single_rank == prev_sorted_registration&.registration&.single_rank
+        # Change position to previous if both single and average are tied with previous registration
+        tied_previous = single_tied_previous && average_tied_previous
+        pos = tied_previous ? prev_sorted_registration.pos : i + 1
+        # Hasn't competed in this event yet.
+        unless rank
           tied_previous = nil
           pos = nil
         end
--- a/WcaOnRails/app/models/competition.rb
+++ b/WcaOnRails/app/models/competition.rb
@@ -1475,13 +1475,14 @@ class Competition < ApplicationRecord
       sorted_registrations = []
       registrations.each_with_index do |registration, i|
         rank = sort_by == 'single' ? registration.single_rank : registration.average_rank
-        average_tied_previous = registration.average_rank == prev_sorted_registration&.registration&.average_rank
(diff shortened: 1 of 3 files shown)


WcaOnRails/spec/controllers/registrations_controller_spec.rb
```diff
@@ -647,30 +647,40 @@
 
     it "sorts 444 by single, and average, and handles ties" do
       registration1 = FactoryBot.create(:registration, :accepted, competition: competition, events: [Event.find("444")])
-      FactoryBot.create :ranks_average, rank: 10, best: 4242, eventId: "444", personId: registration1.personId
-      FactoryBot.create :ranks_single, rank: 20, best: 2000, eventId: "444", personId: registration1.personId
+      FactoryBot.create :ranks_average, rank: 1, best: 2000, eventId: "444", personId: registration1.personId
+      FactoryBot.create :ranks_single, rank: 1, best: 1500, eventId: "444", personId: registration1.personId
 
       registration2 = FactoryBot.create(:registration, :accepted, competition: competition, events: [Event.find("444")])
+      registration2.person.update!(name: "A")
```
gregorbg, line 654: Why is it important that you have the name set to `A`? As far as I can tell, the names in the resulting psych sheet are not being tested.

Also, if you _really_ need the name to specifically be A,
1. Think whether there is _really_, _really_ a good reason for this
2. If the answer to (1) is Yes, mock a user called `A` using `user_a = FactoryBot.create(:user, :wca_id, name: 'A')` and attach them to the registration like `registration2 = FactoryBot.create(:registration, :accepted, user: user_a, ...etc...)`

gregorbg (changes requested):
Questions mostly about your new test cases, as well as some clean code nitpicks in the actual implementation

WcaOnRails/spec/controllers/registrations_controller_spec.rb
maxidragon: @gregorbg I did that, because psychsheet is sorted by name (when results are tied), so if I didn't do that, the code might not pass the tests - because the tests check whether the registrations are in the right order, and in the case of the same results, they are sorted by name.

WcaOnRails/app/models/competition.rb
```diff
@@ -1476,13 +1476,11 @@ def psych_sheet_event(event, sort_by)
       registrations.each_with_index do |registration, i|
         if sort_by == 'single'
           rank = registration.single_rank
-          prev_rank = prev_sorted_registration&.registration&.single_rank
         else
           rank = registration.average_rank
-          prev_rank = prev_sorted_registration&.registration&.average_rank
         end
         if rank
```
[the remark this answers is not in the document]
maxidragon: Its existence is needed to verify that a person has a result in that event

gregorbg (approved):

PR #8119, opened 2023-07-28
Updating Palestine's (CN) Translation

hanwu85:
another country's translation was mitakenly input here.


gregorbg:
Thanks for catching that edge case!

PR #8009, opened 2023-06-26
Replaced senior_delegate_id with region_id and introduced Regions table

danieljames-dj:

PR #8120, opened 2023-07-29
Change column name region to location

danieljames-dj:
Once the [regions PR](https://github.com/thewca/worldcubeassociation.org/pull/8009) is live, we might get confused between `region` and `region_id`. While `region_id` stands for the region of regions table, `region` stands for actually "location". Hence, renaming `region` to `location`.

(I initially thought of combining with other PR, but I didn't want to make the other PR more complex, hence did this separately)


dunkOnIT:
@danieljames-dj is this still relevant given that #8120 is now closed?

danieljames-dj:
Hi @dunkOnIT, yes it is. This is just an improvement to that column name of users table. This will also be really help in implementation of regions with new model.

dunkOnIT:
Ok cool! What is required before merging? Just reviews?

danieljames-dj:
Yes :-)

dunkOnIT:
This is an interesting one, as if we miss something in review it could cause a spectacular failure in production :P Do you have any thoughts on how we mitigate that risk / how confident are you that you caught all usages of `region` and changed them to `location`? 

Review steps:
- [ ] Make sure that all appropriate instances of `region` have been changed to location 
- [ ] Test a core set of critical functions - can we make this an automated process somehow? (_It would alleviate a lot of anxiety/uncertainty around core changes like this. Perhaps our test suite already covers them?_)

danieljames-dj:
Hi Duncan, I've made sure all instances of region has been changed to location. Sorry that I didn't understand what you mean by critical functions. Is it actions like 'view delegates page', 'view user edit page', etc where we use the region?

danieljames-dj:
I'll explain little more about why I actually need this PR to work on the delegates restructuring. I've already [created a PR](https://github.com/thewca/worldcubeassociation.org/pull/8300) to create the groups and roles table with implementation of one group. For the implementation of delegates, I'm planning to do in different phases because doing it in a single PR is really difficult for me.

In the first phase, I'll be creating all the regions in groups table, and map all the delegates under corresponding region without changing anything else. For mapping delegates under a region, I have to introduce region_id in users table, which will show the region they belong to. At that time, if both region and region_id exists in users table and if there is no relation between them, then it will cause confusion. Also, in the final delegates_metadata table, the current region column will be named under location column. So, renaming it now will really help in coding more easily.

dunkOnIT:
Hey Daniel Thanks so much for the detail there. It sounds like a really good change. 

We should be able to get this through pretty quickly - I'll grep and double-check that "region" has been changed to "location" when appropriate, and then we'll deploy to staging and ensure that core functionality is working as expected.

The "critical functions" thing is really outside the scope of this PR, more just me thinking about our testing in general. The thinking is basically - we have a set of critical functions that the WCA website provides to its users - competition creation, registration, results posting, etc. In theory, identifying those critical functions and making sure that they continue to function would help us push these changes with a lot of confidence. Many of these cases are already covered in our test suite, but I'm not exactly sure how comprehensive they are.

dunkOnIT (approved):
Approved after (partial) manual review, and confirmation with Gregor and Finn.

Deployment note: Migration must be run manually when deploying.

danieljames-dj:
Thanks @dunkOnIT, can you please merge this PR and run the migration in server? I can't do this from my side as I don't have access to run migration in server.

dunkOnIT:
Sure thing - I'll probably do it tomorrow morning with Finn, as I haven't run migrations on prod before.

PR #8125, opened 2023-07-29
Competition series API route

### LINKED ISSUE #8115: API route to expose competition series data
Speedcubing Canada would like to develop landing pages for a competition series on our website that displays information about all the competitions in a series (so we can provide a single link).

API route `GET /competition_series/:id` that returns at least:
- the competition series name
- an array of IDs of competitions in the series

(Likely just use the `to_wcif` method from `CompetitionSeries`, then add an API test for the route.) 

N/A

**Additional context**
I'm happy to take this on if there are no objections from WST.

JonEsparaz:
Closes #8115.


WcaOnRails/app/controllers/api/v0/api_controller.rb
```diff
@@ -216,4 +216,13 @@ def export_public
   def countries
     render json: Country.all
   end
+
+  def competition_series
+    competition_series = CompetitionSeries.find_by_wcif_id(params[:id])
```
gregorbg, line 221: You can also raise the `WcaException::NotFound` if the series was actually not found. That will also avoid the null-safe `&` call below.

WcaOnRails/app/controllers/api/v0/api_controller.rb
```diff
@@ -216,4 +216,13 @@ def export_public
...
+    if competition_series&.all_competitions_public?
```
gregorbg, line 222: Cool that you thought about this. I wonder whether it might be a better approach to just filter out the competitions that are not public, and then perhaps throw some error (like pretend the series was not found) if the filtered list is empty

gregorbg (changes requested):
General approach in code is fine, I just want to discuss how to handle non-public competitions in a series

1. Block the entire series (pretend it doesn't exist) if one of the competitions isn't announced yet
2. Block only the competitions that are not announced and normally return the rest

I don't have any strong preference right now, are there use-cases for external third-party tools that favor the first or the second approach?

JonEsparaz:
Thanks for the review @gregorbg! 

> I don't have any strong preference right now, are there use-cases for external third-party tools that favor the first or the second approach?

I'm going to go with approach 2. I like your suggestion since it covers the following scenario:
- Competition A 2023 and Competition B 2023 are in Example Series 2023, both are announced
- Later, Competition C 2023 is added to Example Series 2023 due to demand
- If we 404 when some of the competitions are not yet announced, the endpoint would 404 for Example Series 2023 during the period of time when Competition C is added to the series, but not announced by WCAT

---

Also, the frontend and public full WCIF will reveal the names/IDs of competitions in a series that are not yet announced. The WCIF fix would come naturally with this PR, however, I would like to make any frontend changes in a separate PR.

WcaOnRails/app/controllers/api/v0/api_controller.rb
JonEsparaz: Changed to a `.blank? ||` condition. Not sure if `.present?`, `.nil?`, or `.empty?` (or something else) would be more appropriate - please let me know.

WcaOnRails/app/controllers/api/v0/api_controller.rb
JonEsparaz: Addressed [below](https://github.com/thewca/worldcubeassociation.org/pull/8125#issuecomment-1657184645) and in following commits. Thanks for the suggestion.

WcaOnRails/app/controllers/api/v0/api_controller.rb
gregorbg: I think `if !competition_series.present?` is the cleanest approach here

WcaOnRails/app/models/competition_series.rb
```diff
@@ -129,4 +129,8 @@ def self.wcif_to_attributes(wcif)
     # reset them so that upon the next read they will be fetched based on what's just been written.
     @competition_ids = nil
   end
+
+  def public_competitions
+    self.competitions.filter(&:showAtAll?)
```
gregorbg, line 134: Are you sure this uses the ORM query interface directly? I _think_ (I am genuinely not sure) `filter` first loads all competitions into memory and then throws away the hidden ones. With an approach like `self.competitions.where(showAtAll: false)` we could make sure that the Rails ActiveRecord ORM only loads those competitions into memory that are public in the first place.

gregorbg (changes requested):
Approach is clean now, some nitpicks about the ActiveRecord API :)

WcaOnRails/app/models/competition_series.rb
JonEsparaz: Ah good point. Thanks, Gregor!

PR #8126, opened 2023-07-30
Always touch updated_at timestamp for WCIF PATCH calls

gregorbg:
See discussion under https://github.com/thewca/worldcubeassociation.org/pull/8100#issuecomment-1656936460


FinnIckler (approved):
LGTM!

PR #8127, opened 2023-07-30
Add non-trivial password to Staging

gregorbg:
With the exponential growth of cubing, we unfortunately also experience an exponential growth of idiots who abuse the fact that the password `wca` grants you unlimited access on staging.

We have seen unfortunate and despicable incidents in the past couple of weeks which force us to enact better security measures. The first "quick fix" that I could come up with during the Worlds preparation trouble was to introduce a non-trivial password. We might consider other additional approaches when things settle down after US Nats and Worlds, but I think this will do for now.

The import code is configured to still use `wca` as a password for local imports of the dev dump.


gregorbg:
Merging without a review due to security concerns

PR #8128, opened 2023-07-30
Fix regression of allowlist in Carrierwave upgrade

gregorbg:
Because they renamed it a looooong time ago.
 Bye-bye HEIC :D


danieljames-dj (approved):
LGTM

PR #8130, opened 2023-07-30
README updated the order

danieljames-dj:
Currently, when a new contributor comes into the application, when they go through the setup, they feel like they have to do all the installing. This PR makes it bit more clear that they have two ways to proceed, and will redirected to each way right after cloning.

The purpose of this PR is to make the onboarding of a new contributor a bit smoother (if they are using docker).

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,23 @@ This repository contains all of the code that runs on [worldcubeassociation.org]
   git clone https://github.com/thewca/worldcubeassociation.org
   cd worldcubeassociation.org
   ```
+
+We can run the application in two ways:
+- [Run using Docker](#run-using-docker)
+- [Run directly with Ruby](#run-directly-with-ruby-lightweight-also-only-runs-the-rails-portions-of-the-site)
+
+### Run using Docker (strongly recommended)
+
+- Install [Docker](https://docs.docker.com/get-docker/) (remember to complete the [Linux post-install steps](https://docs.docker.com/engine/install/linux-postinstall/) if you're on Linux)
+- Install docker-compose. The best way to get an up-to-date version is to get it from their [releases page](https://github.com/docker/compose/releases)
+- Navigate into the repository's main directory (`worldcubeassociation.org`)
+- To start the server at `http://localhost:3000`, run `docker-compose up` and to bring it down, `docker-compose down` (or just press ctrl + c in the same terminal)
+- If you want to run the php part of the website as well, run `docker compose -f docker-compose.yml -f docker-compose.php.yml up`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -9,6 +9,23 @@ This repository contains all of the code that runs on [worldcubeassociation.org]
...
+- Install docker-compose. The best way to get an up-to-date version is to get it from their [releases page](https://github.com/docker/compose/releases)
+- Navigate into the repository's main directory (`worldcubeassociation.org`)
+- To start the server at `http://localhost:3000`, run `docker-compose up` and to bring it down, `docker-compose down` (or just press ctrl + c in the same terminal)
+- If you want to run the php part of the website as well, run `docker compose -f docker-compose.yml -f docker-compose.php.yml up`
+- To run tests, run `docker-compose exec wca_on_rails bash -c "RAILS_ENV=test bin/rake db:reset && RAILS_ENV=test bin/rake assets:precompile && bin/rspec"`
+- If you're using Visual Studio Code to develop, you can [attach it to the Docker container](https://code.visualstudio.com/docs/remote/containers) so that your extensions can take advantage of the Ruby environment and so the terminal runs from inside the container
+
+### Run directly with Ruby (lightweight, also only runs the Rails portions of the site)
```
gregorbg, line 27: The "only Rails portion" part is obsolete now that we don't have PHP anymore. There is only Rails and running Rails runs the entire website.

README.md
```diff
@@ -9,6 +9,23 @@ This repository contains all of the code that runs on [worldcubeassociation.org]
   git clone https://github.com/thewca/worldcubeassociation.org
   cd worldcubeassociation.org
   ```
+
+We can run the application in two ways:
+- [Run using Docker](#run-using-docker)
```
gregorbg, line 14: Add a note that Docker is strongly recommended, and that choosing a manual installation happens "at your own risk" and we may not be able to assist with detailed troubleshooting

gregorbg (changes requested):
Cool idea! Just some remarks for clarifications

README.md
danieljames-dj: Oh yes, so is line 23 also not needed anymore? Looks like `docker-compose.php.yml` is still there.

README.md
danieljames-dj: Okay done. Added the 'risk' sentence to rails steps.

danieljames-dj: In docker, should we also add a note that while committing, they have to use `--no-verify` to skip the precommit? Because the precommit requires rails to be installed.

gregorbg: No, our precommit hook has been unmaintained for ages. I'd rather recommend against installing it into local environments (we use it mainly for CI...)

gregorbg (approved):

PR #8131, opened 2023-07-30
Hide warning "registration open in less than required 48 hours" on announced competitions

### LINKED ISSUE #7487: Registration is set to start in less than the required 48 hours showing on websites of announced competition
**Describe the bug**
When you go to announced competition where registration is supposed to start soon, you will see the warning. 

**To Reproduce**
1. Go to https://www.worldcubeassociation.org/competitions/CyprusWinterCup2023 . Since today is 21st December and the registration starts on 22nd, it's there. It's there both when I am signed it (I am a junior delegate) and when I am not.

**Expected behavior**
I would expect the yellow box about registration starting in less than required 48 hours to not be there. When it's like this, it may confuse less experienced competitors (What do they mean by required? What am I doing wrong? etc.).

**Smartphone:**
 - Device: Samsung Galaxy S22
 - OS: Android 13
 - Browser: Chrome 108

maxidragon:
Fixes #7487


gregorbg:
At a quick glance, it seems like you are removing the warning _completely_, while the goal here should be to remove it only for _announced_ competitions. Are you sure this works as intended? Can you add test cases?

maxidragon:
@gregorbg Okey, I'll add test cases, but I didn't remove it completely, when competition is not announced warning is displayed.

gregorbg:
Hm, with some time having passed I changed my mind on this PR. Instead of deleting the warning for announced competitions, I think you should keep all warnings as-is but only show them when the user opening the competition page is actually entitled to see these warnings.

Approach: The current `reg_warnings` is only ever called in `warnings_for` in the Competition model. This `warnings_for` receives the current user as a parameter, so you know who is trying to access the warnings. Look at `user.rb` model to figure out a relevant permission that tells you "is this user somehow organizing/moderating/administrating this competition?" and use that to decide whether you should add the warnings or not.

gregorbg (changes requested):
(see my other comment above for change request)

gregorbg (approved):

PR #8133, opened 2023-07-31
Declutter multiple room schedule

### LINKED ISSUE #7841: Declutter the schedule view when organizers want to utilize "multiple rooms" with the same schedule
With most group software (as far as I'm aware) different groups are seen by different "rooms" in the same venue.  Unfortunately, as it is, even if the events line up perfectly, it shows the same thing multiple times on the table view (once for each "room").  It does the same with the calendar view, but at least that is less visually cluttered.

I think the better UI would be something where instead of "2x2x2 Cube First round" appearing on both the blue stage, then again for the red stage (in this example, the rooms are red stage and blue stage) would be for it to just say "2x2x2 Cube First round" on the red and blue stage.  I think that this would make a less confusing, and less cluttered UI.

maxidragon:
Closes #7841 

Schedule for the competition with multiple rooms now looks like this:

image


gregorbg (changes requested):
The idea is good, but with the execution you need to be _very_ careful.

Right now, you are using the title as the only exclusive property for grouping. This can be mis-leading when activities with the same name happen at different times. Edge-cases like this need to be considered thoroughly, before we can "just" merge this as-is.

dunkOnIT:
@maxidragon would you still like to work on this? Let me know if you need any help with brainstorming around edge cases.

maxidragon:
@dunkOnIT Yes, I'll work on that this week

maxidragon:
Now it looks like this ;)

image

coder13:
I'll just say this looks fantastic!

gregorbg (approved):

PR #8135, opened 2023-08-01
Added queue status for delayed jobs

danieljames-dj:
Recently, WRT faced a case where they started CAD, and took around 15 minutes to complete. But the UI said it started just 4 minutes ago even though the button was clicked 15 minutes ago. From the code, I understood that the job stays in queue initially for sometime, and for some reasons the queue was huge during that time.

This PR is to show the queue status if the job is queued. This will be helpful for WRT to get an idea if the job actually started, or is it still in queue.

### DIFF
--- a/WcaOnRails/app/jobs/timed_application_job.rb
+++ b/WcaOnRails/app/jobs/timed_application_job.rb
@@ -3,6 +3,10 @@
 module TimedApplicationJob
   extend ActiveSupport::Concern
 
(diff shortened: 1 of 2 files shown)


gregorbg:
We're adding this exact feature as part of the migration to Sidekiq in #8087 
The plan is to merge that other Sidekiq PR some time next week, so I think there's no need to merge this feature now.

PR #8158, opened 2023-08-05
Moved calculateDues to helper and added dues to WFC export

danieljames-dj:
If the WFC export from panel has the calculated dues as well, it will be really helpful. Currently, it returns 0 for some currencies because the currency is not available with RubyMoney. But the available data might be helpful data for WFC a bit.

### DIFF
--- a/WcaOnRails/app/controllers/competitions_controller.rb
+++ b/WcaOnRails/app/controllers/competitions_controller.rb
@@ -411,32 +411,16 @@ class CompetitionsController < ApplicationController
   end
 
   def calculate_dues
-    country_iso2 = Country.find_by(id: params[:country_id])&.iso2
-    country_band = CountryBand.find_by(iso2: country_iso2)&.number
-
-    update_exchange_rates_if_needed
-    input_money_us_dollars = Money.new(params[:entry_fee_cents].to_i, params[:currency_code]).exchange_to("USD").amount
-
-    registration_fee_dues_us_dollars = input_money_us_dollars * CountryBand::PERCENT_REGISTRATION_FEE_USED_FOR_DUE_AMOUNT
-    country_band_dues_us_dollars = CountryBand::BANDS[country_band][:value] if country_band.present? && country_band > 0
+    competition = Competition.find(params[:competition_id])
+    dues_per_competitor_in_usd = ApplicationController.helpers.dues_per_competitor_in_usd(competition)
 
     # times 100 because later currency conversions require lowest currency subunit, which is cents for USD
-    price_per_competitor_us_cents = [registration_fee_dues_us_dollars, country_band_dues_us_dollars].compact.max * 100
-
-    if ActiveRecord::Type::Boolean.new.cast(params[:competitor_limit_enabled])
-      estimated_dues_us_cents = price_per_competitor_us_cents * params[:competitor_limit].to_i
-      estimated_dues = Money.new(estimated_dues_us_cents, "USD").exchange_to(params[:currency_code]).format
-
-      render json: {
-        dues_value: estimated_dues,
-      }
-    else
-      price_per_competitor = Money.new(price_per_competitor_us_cents, "USD").exchange_to(params[:currency_code]).format
-
-      render json: {
-        dues_value: price_per_competitor,
-      }
-    end
+    price_per_competitor_us_cents = dues_per_competitor_in_usd * 100
+    multiplier = competition.competitor_limit_enabled? ? competition.competitor_limit : 1
+    total_dues = price_per_competitor_us_cents * multiplier
+    render json: {
+      dues_value: Money.new(total_dues, "USD").exchange_to(competition.currency_code).format,
+    }
   end
 
   def show
@@ -636,12 +620,6 @@ class CompetitionsController < ApplicationController
 
   private def confirming?
     params[:commit] == "Confirm"
-  end
-
-  private def update_exchange_rates_if_needed
-    if !Money.default_bank.rates_updated_at || Money.default_bank.rates_updated_at < 1.day.ago
-      Money.default_bank.update_rates
-    end
   end
 
   private def competition_params
--- a/WcaOnRails/app/controllers/competitions_controller.rb
+++ b/WcaOnRails/app/controllers/competitions_controller.rb
@@ -411,16 +411,15 @@ class CompetitionsController < ApplicationController
   end
 
   def calculate_dues
-    competition = Competition.find(params[:competition_id])
     country_iso2 = Country.find_by(id: params[:country_id])&.iso2
     dues_per_competitor_in_usd = DuesCalculator.dues_per_competitor_in_usd(country_iso2, params[:base_entry_fee_lowest_denomination], params[:currency_code])
 
     # times 100 because later currency conversions require lowest currency subunit, which is cents for USD
     price_per_competitor_us_cents = dues_per_competitor_in_usd * 100
-    multiplier = competition.competitor_limit_enabled? ? competition.competitor_limit : 1
+    multiplier = params[:competitor_limit_enabled] ? params[:competitor_limit].to_i : 1
     total_dues = price_per_competitor_us_cents * multiplier
     render json: {
-      dues_value: Money.new(total_dues, "USD").exchange_to(competition.currency_code).format,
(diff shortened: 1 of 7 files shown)


gregorbg (changes requested):
Two remarks:

1- If you put the code somewhere, put it under `lib` instead of putting it in a `Helper`. The latter are Rails support structures for **Views**, and you might have realised yourself that calling a Helper outside of views via `SomeController.helpers.my_cool_function` is quite hacky. `lib` code can be accessed in Views and in Controllers anywhere.
2- Your approach of replacing the current AJAX with only the competition ID does not work. If you load the competition from the DB based on the ID, you will get outdated information. The reason why we bundle all of these individual bits of information (currency code, competitor limit, etc) into the AJAX request is because they might change on the spot before the changes are persisted to the database. So the backend cannot know about changes to your base fee unless you send them in the AJAX itself.

danieljames-dj:
Thanks @gregorbg for your comment. I've made changes to address both the cases that you mentioned.

WcaOnRails/app/models/competition.rb
```diff
@@ -1965,4 +1965,10 @@ def find_round_for(event_id, round_type_id, format_id = nil)
         (format_id.nil? || format_id == r.format_id)
     end
   end
+
+  def dues_per_competitor_in_usd
+    DuesCalculator.dues_per_competitor_in_usd(self.country_iso2, self.base_entry_fee_lowest_denomination, self.currency_code)
+  rescue StandardError
```
gregorbg, line 1971: Why are you catching an error here? When could an error potentially be thrown?

WcaOnRails/app/views/competitions/_competition_form.html.erb
```diff
@@ -251,6 +251,7 @@
 
         <script>
           function UpdateDues(){
+            var competitionId = $("#competition_id").val();
```
gregorbg, line 254: Where is this competition ID being used? I think it is not necessary.

WcaOnRails/lib/dues_calculator.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module DuesCalculator
+  def self.update_exchange_rates_if_needed
+    if !Money.default_bank.rates_updated_at || Money.default_bank.rates_updated_at < 1.day.ago
+      Money.default_bank.update_rates
+    end
+  end
+
+  def self.dues_per_competitor_in_usd(country_iso2, base_entry_fee_lowest_denomination, currency_code)
```
gregorbg, line 10: You are passing in the `currency_code` in order to be able to convert to USD internally.
Would it make sense to convert back to the local `currency_code` currency before returning?

gregorbg: In other words, are there use cases in the code where it makes sense to have the Dues always be provided in USD?

WcaOnRails/lib/dues_calculator.rb
```diff
@@ -0,0 +1,21 @@
...
+    country_band = CountryBand.find_by(iso2: country_iso2)&.number
+
+    DuesCalculator.update_exchange_rates_if_needed
+    input_money_us_dollars = Money.new(base_entry_fee_lowest_denomination.to_i, currency_code).exchange_to("USD").amount
```
gregorbg, line 14: Why are you converting this to an integer?
- If it already is an integer, then `to_i` is not necessary.
- If it already is a money object, then creating a `Money.new` is not necessary.

WcaOnRails/app/controllers/competitions_controller.rb
```diff
@@ -412,31 +412,15 @@ def time_until_competition
...
-        dues_value: estimated_dues,
-      }
-    else
-      price_per_competitor = Money.new(price_per_competitor_us_cents, "USD").exchange_to(params[:currency_code]).format
-
-      render json: {
-        dues_value: price_per_competitor,
-      }
-    end
+    price_per_competitor_us_cents = dues_per_competitor_in_usd * 100
```
gregorbg, line 418: Rather than doing arbitrary multiplications with a "magic" 100 number, you could pass around the Ruby `Money` object instead.

gregorbg (changes requested):

WcaOnRails/app/controllers/competitions_controller.rb
danieljames-dj: But don't we need to convert to cent? I'm not sure exactly why, but we used to multiply by 100 earlier too. Is there any way to do with Ruby Money here?

WcaOnRails/app/controllers/competitions_controller.rb
```diff
@@ -412,31 +412,15 @@ def time_until_competition
...
+    multiplier = params[:competitor_limit_enabled] ? params[:competitor_limit].to_i : 1
```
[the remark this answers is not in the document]
danieljames-dj: Done, thank you.

WcaOnRails/app/controllers/competitions_controller.rb
```diff
@@ -412,31 +412,15 @@ def time_until_competition
...
+    total_dues = price_per_competitor_us_cents * multiplier
+    render json: {
+      dues_value: Money.new(total_dues, "USD").exchange_to(params[:currency_code]).format,
```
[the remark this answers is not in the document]
danieljames-dj: Done.

WcaOnRails/app/models/competition.rb
danieljames-dj: The case where the currency is not available in Ruby money.

WcaOnRails/app/views/competitions/_competition_form.html.erb
danieljames-dj: Haa yes it's not used, I removed the usage in some previous commit. Thanks for pointing it, removed.

WcaOnRails/lib/dues_calculator.rb
danieljames-dj: Yes there is, For WFC export, we want the dues in USD.

WcaOnRails/lib/dues_calculator.rb
danieljames-dj: It can be string when it comes from competition_controller, hence added `to_i`.

WcaOnRails/app/models/competition.rb
gregorbg: In that case, think about more clever ways to handle this edge case. Investigate whether there is a `currency_available?` method in the Money gem that we use.

Catching errors to return a default value is bad practice, and catching generic `StandardError` is very bad practice :see_no_evil:

WcaOnRails/app/controllers/competitions_controller.rb
gregorbg: You can have the `DuesCalculator.dues_per_competitor_in_usd` method return a RubyMoney object. That has the additional advantage of making it obvious that the return value is USD. You can rename the method to `DuesCalculator.dues_per_competitor` and the return object itself knows (and can report to the user if necessary) that it is a money object representing USD.

Since Ruby is a dynamic language, you can do shenanigans like:
```ruby
  ruby_money = Money.new(200, "USD") # 2 USD because 200 means "US cents"
  competitor_limit = 50 # read from request body or whatever

  estimated_total_dues = ruby_money * competitor_limit # yes, you can multiply a RubyMoney object with an integer

  puts estimated_total_dues.format # expected output: 100 USD 
```

WcaOnRails/lib/dues_calculator.rb
gregorbg: Okay, then just return them flat in USD and let other consumers of this method handle conversions the way they want.

WcaOnRails/lib/dues_calculator.rb
gregorbg: Then `competition_controller` should be responsible for converting it to an integer before passing it to the dues calculator.

gregorbg (changes requested):

danieljames-dj (commented):
Thanks @gregorbg for the review again. I'm expecting some more changes in rails, as for some lines I think there is a better way, but didn't spend much time to figure out what was the better way. 🙈

gregorbg (approved):

PR #8165, opened 2023-08-09
Media table migration to react component

MibettyMeanie:
React component to render media table in multimedia page

Potential Bugs:

1. Submission date sorting might not be working properly
2. Page might not load properly in Safari


WcaOnRails/app/webpacker/components/Media/MediaTable.jsx
```diff
@@ -0,0 +1,119 @@
...
+        month: "short",
+        day: "numeric",
+    });
+    const date1 = new Date(fromDate);
+    const date2 = new Date(toDate);
+    const ans = format.formatRange(date1, date2);
+    return ans;
+}
+
+function exampleReducer(state, action) {
```
coder13, line 20: It would be good to have proper naming of variables in this situation.

WcaOnRails/app/webpacker/components/Media/MediaTable.jsx
coder13, line 1: It looks like 4 width indentation was used when the standard so far has been to use 2 width indentation.

WcaOnRails/app/webpacker/components/Media/MediaTable.jsx
```diff
@@ -0,0 +1,119 @@
+import _ from 'lodash'
+import React, { useState, useEffect } from 'react';
+import { Table } from "semantic-ui-react";
+import { countries } from '../../lib/wca-data.js.erb';
+import { competitionUrl } from '../../lib/requests/routes.js.erb';
+import moment from 'moment';
+
+function dateRangeBetween(fromDate, toDate) {
```
coder13, line 8: Would you want to create a file in `utils/` for formatting date strings and start putting code like this there?

WcaOnRails/app/webpacker/components/Media/MediaTable.jsx
```diff
@@ -0,0 +1,119 @@
...
+    return ans;
+}
+
+function exampleReducer(state, action) {
+    switch (action.type) {
+        case 'CHANGE_SORT':
+            if (state.column === action.column) {
+                return {
+                    ...state,
+                    data: state.data.slice().reverse(),
```
coder13, line 26: Unless this will be manipulated further (adding / removing items from this list), this can very likely be switched to a derived prop and we just compute it's state in the `MediaTable` component wrapping it with `useMemo`

dunkOnIT:
Hi @MibettyMeanie! Thanks for the PR on this. The main change I think we need to make is to move to serving the Media data via an API, rather than as a Rails object (`@media`) in the View. 

I've done my best to explain this [here](https://docs.worldcubeassociation.org/guides/passing_data_react_rails.html) - please let me know if any of that doesn't make sense, I'd be happy to help further!

Specifically, I think we'd need to make the following changes:
- [ ] Have `get_media`return a JSON of the data it fetches
- [ ] Create a /media API endpoint (it doesn't look like we have one at the moment) which returns the JSON that `get_media` creates
  - [ ] This endpoint should accept the parameters that `get_media` expects (`year`, `region`, `status`). I'd suggest that we pass these parameters as URL parameters
- [ ] Update your React code to fetch this data via the API call, rather than expecting to receive it immediately from the view (once it has the data, no change should be needed - the JSON format from the API can match pretty closely the one you get from the view at the moment)

Let me know if any of that doesn't make sense, or if there's anything I can do to help! I am _literally_ here to help :P

MibettyMeanie:
Thanks a lot for the review @coder13 and @dunkOnIT. I'll work on the changes you both suggested and will push it soon.

WcaOnRails/app/webpacker/components/Media/MediaTable.jsx
MibettyMeanie: I have moved this to libs/helper as its already created used

WcaOnRails/app/webpacker/components/Media/MediaTable.jsx
```diff
@@ -0,0 +1,171 @@
...
+
+import { competitionUrl } from '../../lib/requests/routes.js.erb';
+import { countries } from '../../lib/wca-data.js.erb';
+import { dateRangeBetween } from '../../lib/helpers/media-table';
+import useLoadedData from '../../lib/hooks/useLoadedData';
+import useSaveAction from '../../lib/hooks/useSaveAction';
+
+export default function MediaTable({ isValidate }) {
+  const handleUpdateSuccess = () => {
+    setModalOpen(true);
```
coder13, line 14: Does this line of code work? Trying to access a variable that doesn't get defined until line 28?

WcaOnRails/app/webpacker/components/Media/MediaTable.jsx
MibettyMeanie: FIxed

MibettyMeanie:
> Hi @MibettyMeanie! Thanks for the PR on this. The main change I think we need to make is to move to serving the Media data via an API, rather than as a Rails object (`@media`) in the View.
> 
> I've done my best to explain this [here](https://docs.worldcubeassociation.org/guides/passing_data_react_rails.html) - please let me know if any of that doesn't make sense, I'd be happy to help further!
> 
> Specifically, I think we'd need to make the following changes:
> 
> * [ ]  Have `get_media`return a JSON of the data it fetches
> * [ ]  Create a /media API endpoint (it doesn't look like we have one at the moment) which returns the JSON that `get_media` creates
>   
>   * [ ]  This endpoint should accept the parameters that `get_media` expects (`year`, `region`, `status`). I'd suggest that we pass these parameters as URL parameters
> * [ ]  Update your React code to fetch this data via the API call, rather than expecting to receive it immediately from the view (once it has the data, no change should be needed - the JSON format from the API can match pretty closely the one you get from the view at the moment)
> 
> Let me know if any of that doesn't make sense, or if there's anything I can do to help! I am _literally_ here to help :P

Hey @dunkOnIT, I have added API referring to other controllers and it is now only for fetching and updating, the old controller has ruby API's which are used in the edit view (I haven't changed it in this PR). Also @danieljames-dj informed me that WCT leader has asked him not to delete rejected media from table rather than keep it intact with the status as rejected, so I have implemented like that.

WcaOnRails/app/webpacker/components/Media/MediaTable.jsx
```diff
@@ -0,0 +1,257 @@
...
+      key: 'all',
+      text: 'All Years',
+      value: 'All Years',
+    },
+    ...years.map((year) => ({
+      key: year,
+      text: year,
+      value: year,
+    })),
+  ];
```
kr-matthews, line 39: These could be defined outside the component (otherwise they're recomputed each re-render).

WcaOnRails/app/webpacker/components/Media/MediaTable.jsx
```diff
@@ -0,0 +1,257 @@
...
+import useLoadedData from '../../lib/hooks/useLoadedData';
+import useSaveAction from '../../lib/hooks/useSaveAction';
+import {
+  DropdownMenu,
+  DropdownItem,
+  DropdownHeader,
+  DropdownDivider,
+  Dropdown,
+} from 'semantic-ui-react'
+export default function MediaTable({ isValidate }) {
```
kr-matthews, line 17:
```suggestion
} from 'semantic-ui-react'

export default function MediaTable({ isValidate }) {
```

WcaOnRails/app/webpacker/components/Media/MediaTable.jsx
```diff
@@ -0,0 +1,257 @@
...
+    } else {
+      setMediaCombined([...mediaCombined].reverse());
+      setSortDirection(sortDirection === 'ascending' ? 'descending' : 'ascending');
+    }
+  };
+  const handleSelectYear = (event, data) => {
+    setSelectedYear(data.value);
+  };
+  const handleSelectRegion = (event, data) => {
+    console.log(data)
```
kr-matthews, line 110: Can this be removed?

WcaOnRails/app/webpacker/components/Media/MediaTable.jsx
```diff
@@ -0,0 +1,257 @@
...
+  const [mediaCombined, setMediaCombined] = useState([])
+  const type = isValidate ? 'pending' : 'accepted'
+  const { loading, error, data, sync } = useLoadedData(
+    apiV0Urls.media.list(type, selectedYear, selectedRegion)
+  );
+  const [modalOpen, setModalOpen] = useState(false);
+
+  useEffect(() => {
+    if (data) {
+      setMediaCombined(data?.map((medium, idx) => {
```
kr-matthews, line 61: The current implementation isn't necessarily bad, but generally avoiding side effects and having fewer states is good. I'd suggest making `mediaCombined` state-less:
```
const mediaCombined = data?.map((medium, idx) => { ... })
```
and then having a separate (state-less, memoized) const for the sorted data:
```
const sortedMediaCombined = useMemo(
  () => [...mediaCombined]?.sort(...), // probably doesn't work exactly as written in the null case
  [mediaCombined, sortColumn, sortDirection],
]
```
and then using
```
{sortedMediaCombined?.map((media_row) => (
```
in the jsx return.

Thoughts?

WcaOnRails/app/webpacker/components/Media/MediaTable.jsx
```diff
@@ -0,0 +1,257 @@
...
+          startDate: medium.competition.start_date,
+          endDate: medium.competition.end_date,
+          mediaId: medium.id,
+          competetionDate: dateRangeBetween(medium.competition.start_date, medium.competition.end_date),
+          competetionUrl: competitionUrl(medium.competition.id)
+        };
+      }))
+    }
+  }, [data]);
+  const [modalParams, setModalParams] = React.useState({
```
kr-matthews, line 84:
```suggestion
  const [modalParams, setModalParams] = useState({
```

WcaOnRails/app/webpacker/components/Media/MediaTable.jsx
```diff
@@ -0,0 +1,257 @@
...
+  const handleSort = (clickedColumn) => () => {
+    if (sortColumn !== clickedColumn) {
+      setSortColumn(clickedColumn);
+      setSortDirection('ascending');
+      setMediaCombined([...mediaCombined].sort((a, b) => a[clickedColumn] - b[clickedColumn]));
+    } else {
+      setMediaCombined([...mediaCombined].reverse());
+      setSortDirection(sortDirection === 'ascending' ? 'descending' : 'ascending');
+    }
+  };
```
kr-matthews, line 105: This could be encapsulated via `useReducer`:
```
const [{ sortColumn, sortDirection }, handleSort] = useReducer(sortReducer, { sortColumn: null, sortDirection: null })
```
(with a definition of `sortReducer` outside the component).

WcaOnRails/app/webpacker/components/Media/MediaTable.jsx
```diff
@@ -0,0 +1,257 @@
...
+              onClick={handleSort('link')}>{I18n.t('activerecord.attributes.competition_medium.uri')}</Table.HeaderCell>
+            {isValidate && <Table.HeaderCell></Table.HeaderCell>}
+          </Table.Row>
+        </Table.Header>
+        <Table.Body>
+
+          {mediaCombined?.map((media_row) => (
+            <Table.Row>
+              <Table.Cell>
+                {moment.utc(media_row.timestampSubmitted).format('MMMM DD, YYYY HH:mm UTC')}
```
kr-matthews, line 202: I think we prefer to use luxon over moment? Someone correct me If I'm wrong.

kr-matthews (changes requested):

WcaOnRails/app/webpacker/components/Media/MediaTable.jsx
MibettyMeanie: Fixed

WcaOnRails/app/webpacker/components/Media/MediaTable.jsx
MibettyMeanie: Fixed

WcaOnRails/app/webpacker/components/Media/MediaTable.jsx
MibettyMeanie: Fixed

WcaOnRails/app/webpacker/components/Media/MediaTable.jsx
MibettyMeanie: Done

WcaOnRails/app/webpacker/components/Media/MediaTable.jsx
MibettyMeanie: Done

WcaOnRails/app/webpacker/components/Media/MediaTable.jsx
MibettyMeanie: Done

WcaOnRails/app/webpacker/components/Media/MediaTable.jsx
MibettyMeanie: Done

app/webpacker/components/Media/MediaTable.jsx
```diff
@@ -0,0 +1,254 @@
...
+    text: 'All Years',
+    value: 'All Years',
+  },
+  ...years.map((year) => ({
+    key: year,
+    text: year,
+    value: year,
+  })),
+];
+handleSort = (clickedColumn) => () => {
```
coder13, line 39: This variable should be declared with `const`

app/webpacker/components/Media/MediaTable.jsx
```diff
@@ -0,0 +1,254 @@
...
+import useSaveAction from '../../lib/hooks/useSaveAction';
+import { DateTime } from 'luxon';
+import {
+  DropdownMenu,
+  DropdownItem,
+  DropdownHeader,
+  DropdownDivider,
+  Dropdown,
+} from 'semantic-ui-react'
+const countryOptions = _.map(countries.byIso2, (country) => ({
```
coder13, line 17: super nit: would be good to see some newlines in between these variable declarations.

app/webpacker/components/Media/MediaTable.jsx
```diff
@@ -0,0 +1,254 @@
...
+  const { save, saving } = useSaveAction();
+  const [mediaCombined, setMediaCombined] = useState([])
+  const type = isValidate ? 'pending' : 'accepted'
+  const { loading, error, data, sync } = useLoadedData(
+    apiV0Urls.media.list(type, selectedYear, selectedRegion)
+  );
+  const [modalOpen, setModalOpen] = useState(false);
+
+  useEffect(() => {
+    if (data) {
```
coder13, line 69: nit: I'd use a `useMemo` to transform the data rather than a useEffect unless you also need to modify this data in `mediaCombined` after initial fetching.

app/webpacker/components/Media/MediaTable.jsx
```diff
@@ -0,0 +1,254 @@
...
+    open: false,
+    mediaId: null,
+    status: null,
+  });
+  const confimMedia = () => {
+    save(apiV0Urls.media.update(modalParams.mediaId), { status: modalParams.status }, () => {
+      sync(); handleUpdateSuccess(); setModalParams({ ...modalParams, open: false })
+    }, { method: 'PATCH' });
+  };
+  const [{ sortColumn, sortDirection }, handleSort] = useReducer(sortReducer, { sortColumn: null, sortDirection: null })
```
kr-matthews, line 103: Is `sortReducer` defined anywhere? I can't find it,

app/webpacker/components/Media/MediaTable.jsx
kr-matthews: This has the same name as the dispatch in the `userReducer` below. Also, you can't access these setters outside of the component (plus some of them don't exist). I don't think the reducer has been set up properly (but the intention seems right). I'm happy to help on slack if you want.

app/webpacker/components/Media/MediaTable.jsx
```diff
@@ -0,0 +1,254 @@
...
+          type: medium.type,
+          text: medium.text,
+          uri: medium.uri,
+          country_iso2: iso2,
+          country_name: country ? country.name : '',
+          city: medium.competition.city,
+          startDate: medium.competition.start_date,
+          endDate: medium.competition.end_date,
+          mediaId: medium.id,
+          competetionDate: dateRangeBetween(medium.competition.start_date, medium.competition.end_date),
```
kr-matthews, line 87: "competetion" should be "competition"; six occurrences overall.

app/webpacker/components/Media/MediaTable.jsx
kr-matthews: That is along the lines of what I suggested [here](https://github.com/thewca/worldcubeassociation.org/pull/8165#discussion_r1495037652); I didn't include a `useMemo` wrapper, but including one would be good.

PR #8166, opened 2023-08-10
Rename "Citizenship" to "Representing" and "Country" to "Region"

gregorbg:
As per WSOT request leading up to the World Championships


viroulep:
Doesn't this introduce an inconsistency between the Regulations and the website?

@thewca/wrc-team the Regulations explicitly talk about "countries", should the other occurrences of "country" be renamed to "region"?

gregorbg:
> Doesn't this introduce an inconsistency between the Regulations and the website?

Yes. Right now for the upcoming competition, the website is much more important though.

gregorbg:
Silly test failing. Merging as-is due to high priority before Worlds

Issue #8122, opened 2023-07-29
Model for Delegates & Memberships


danieljames-dj:
This GitHub issue is to discuss the model for Delegates & Memberships (memberships is same as existing team_members table, but with little optimizations and maybe a slightly better name).

### Delegates Table (name: `delegates`)

| Column name | Column type | Details |
|---|---|---|
| id | bigint(20) | Primary key |
| user_id | bigint(20) | Used as a foreign key to reference `users` table |
| status | varchar(191) | This will represent the status of the delegate - like 'trainee', 'junior', etc |
| region_id | bigint(20) | Used as a foreign key to reference `regions` table |
| start_date | date | Start date of this delegating status |
| end_date | date | End date of this delegating status |
| created_at | datetime | Date & time at which the row is created |
| updated_at | datetime | Date & time at which the row is last edited |
| country_iso2 | varchar(191) | Used to represent the country |
| location | varchar(191) | The location where the delegate will be delegating |
| metadata | varchar(191) | This will be a stringified JSON which will have some optional metadata like in_probation, probation_start_date, probation_end_date, etc |

The above structure is same as the structure discussed [here](https://github.com/thewca/worldcubeassociation.org/pull/5417), but I've added one more column `metadata` which currently have three optional data like in_probation, probation_start_date, probation_end_date, etc, but we can have much more in future like pending_dues, number_of_competitions_delegated, etc.

### Memberships Table (name: `memberships`)

| Column name | Column type | Details |
|---|---|---|
| id | bigint(20) | Primary key |
| user_id | bigint(20) | Used as a foreign key to reference `users` table |
| status | varchar(191) | This will represent the status of the membership - like 'member', 'senior_member', 'leader', 'vice_leader', 'chair', 'vice_chair', 'secretary', etc |
| team_id | bigint(20) | Used as a foreign key to reference `teams` table |
| start_date | date | Start date of this membership status |
| end_date | date | End date of this membership status |
| created_at | datetime | Date & time at which the row is created |
| updated_at | datetime | Date & time at which the row is last edited |
| metadata | varchar(191) | This will be a stringified JSON which will have some optional metadata like translation_locale, main_role, etc |

The above structure is similar to current structure, but we don't have `team_leader` and `team_senior_member` columns, instead we have `status` column.

The metadata columns of both tables looks bit odd because they don't have a proper type, and JSON is not a type of relational DB. We might have to discuss how to approach the `metadata` column. There are many options:
[...]

PR #8209, opened 2023-08-21
Fixes #4328 - Delegates UI revamp

### LINKED ISSUE #4328: Shortcuts on Delegates Page
The delegates page is very long (which is good to see). A byproduct of this is that it requires quite a lot of scrolling to find the delegates of a certain region.

It would be great if there were shortcut links/table of contents that allow you to jump to the start of a particular region.

danieljames-dj:
Here is the demo:

Version 3:

https://github.com/thewca/worldcubeassociation.org/assets/9512698/83744a90-b4d6-4245-af28-5573e7d4514c

Version 2:

https://github.com/thewca/worldcubeassociation.org/assets/9512698/6ec63b77-272d-44ce-a8cc-647a8d69d836

Version 1:

https://github.com/thewca/worldcubeassociation.org/assets/9512698/0a2d19fd-62fe-4d7d-9c85-e99035820b73

Open to discussion on UI improvements

### DIFF
--- a/WcaOnRails/app/webpacker/components/Delegates/index.jsx
+++ b/WcaOnRails/app/webpacker/components/Delegates/index.jsx
@@ -1,10 +1,15 @@
-/* eslint-disable jsx-a11y/control-has-associated-label */
 import React from 'react';
 
 import {
-  Button, Dropdown, Segment, Table, Grid, Menu,
+  Button,
+  Dropdown,
+  Grid,
+  Header,
+  Label,
+  Menu,
+  Segment,
+  Table,
 } from 'semantic-ui-react';
-import { useMediaQuery } from 'react-responsive';
 import I18n from '../../lib/i18n';
 import UserBadge from '../UserBadge';
 
@@ -82,7 +87,6 @@ export default function Delegates({
   delegates,
   isEditVisible,
 }) {
-  const isMobile = useMediaQuery({ query: '(max-width: 760px)' });
   const seniorDelegates = React.useMemo(() => delegates
     .filter((user) => user.delegate_status === 'senior_delegate')
     .sort((user1, user2) => (user1.location || '').localeCompare(user2.location || '')), [delegates]);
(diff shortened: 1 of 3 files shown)


EdHollingdale:
One thought: it might make more sense to combine Delegates/Junior Delegates together. That way if someone is looking for all the delegates in an area (e.g. "Australia (Victoria)") they will all be next to each other instead of spread over two different tables?

If we implement regional delegates in the website how would that look? Would Delegates/Juniors be able to appear 'under' their regional delegate?

danieljames-dj:
Hi @EdHollingdale, combining Delegates/Junior Delegates looks like a good idea, I'll implement that, thanks for the idea.

Regarding regional delegates, as you might already know it cannot be implemented along with this. That can be done only after the [model discussion](https://github.com/thewca/worldcubeassociation.org/issues/8122) is completed.

Also, currently the delegates are mapped under a senior delegate. Under the new structure, every delegates including the senior delegate and regional delegate will be mapped under a region. So, under each region, we can show all the sub-regions and the regional delegate and other delegates of that sub-region.

danieljames-dj:
I've updated the UI as per the suggestion from @EdHollingdale. I've also changed the side bar to a dropdown menu which looks a bit better I guess.

dunkOnIT:
I strongly prefer the sidebar in Version 1, both from a UX (it's 1 click to any region as opposed to 2 + scrolling), and UI (I think it looks better) perspective.

For displaying role, I think there are a few options: 
- just keep the role column (we have a lot space in the design)
- specify if someone is a Junior/Trainee in brackets next to their name
- group the delegates in a region by role and denote it with a heading and possibly a slightly different background shade or similar

EdHollingdale:
I agree with Duncan that the sidebar is preferable. Specifying Delegate vs Junior Delegate in brackets next to the name could be a good solution. But delegates should be grouped by sub-region (e.g. country/state) not by role (junior or full delegate).

dunkOnIT:
We could also consider 2 levels of sorting - first sub-region then role (ie order roles within the same region)

danieljames-dj:
The main drawback of sidebar is that it won't be responsive and won't look good in mobile devices.

dunkOnIT:
Perhaps we change to a dropdown view on certain resolutions then? Although if's not as elegant as just having one responsive design, I think the UX benefits on each platform are significant.

danieljames-dj:
Yes I've just done the same thing as you suggested in your previous comment @dunkOnIT :-)

I was about to comment after updating the screen recording. Will upload the screen recording soon.

danieljames-dj:
Hi all, I've updated the screen recording of version 3 which includes good side of both version 1 and version 2. Please go through it and give further suggestions on the UI.

PS: The react code is not clean and localizations are not done for few strings. So, it might be better to hold reviewing till the UI is finalized.

dunkOnIT:
Looks awesome, thanks! 
* **Dropdown:** What if we remove "Region" and just have the name of the Region and the dropdown icon - looks cleaner, and I think it will be intuitive
* Scroll headers and selector with the page: Literally no idea how hard this is to do, but it would be great if the dropdown/sidebar selector could scroll with the user, so that it is always visible. It would also be cool if the table header could always be visible ("frozen" in Excel terminology) but again, not sure how hard this is

danieljames-dj:
Thanks @dunkOnIT. Both suggestions looks good. I've executed the first one. I couldn't do the second one as I think it requires some css changes, which I'm not sure will be able to do at this point (when I checked, I see that we need to do 'position:fixed', but alone with that some other changes need to be made). Maybe I'll leave that for now and later once merged, I'll create a github issue so that somebody else can take it?

dunkOnIT:
Sounds good to me! Will probably make a nice  "good first issue" for a frontend dev

gregorbg (changes requested):
Please do not implement the responsive aspect via a separate package. The 760px come across as a "magic constant" that can really clutter our code in the long run.

Semantic UI has a different approach about Responsive Design: https://react.semantic-ui.com/layouts/responsive (there is a "Show source" button in the bottom-left corner)

danieljames-dj:
> Please do not implement the responsive aspect via a separate package. The 760px come across as a "magic constant" that can really clutter our code in the long run.
> 
> Semantic UI has a different approach about Responsive Design: https://react.semantic-ui.com/layouts/responsive (there is a "Show source" button in the bottom-left corner)

But in our current use-case, we need to change the component if the device is different - like dropdown should come in mobile version. But in the example that you sent, I guess they are using grids to make it responsive. So don't we need some library to achieve our use case easily?

dunkOnIT:
I believe we should change it based on the screen resolution, not based on the device? But IANA a frontend dev :P

gregorbg:
Please hold off with the different libraries for now. I feel like there _should_ be a clever responsive solution for this, and I'll be happy to help investigating a little later on. Please hold the line :D

gregorbg:
I've made a suggestion how to implement the design using only SemUI responsive components, without the need for an external library. Please see if this looks good :)

danieljames-dj:
Wow this looks really amazing. 👏👏👏👏👏 

Thanks a lot @gregorbg.

gregorbg (approved):

PR #8230, opened 2023-08-22
Recompute skipped indexes after Persons column renames

gregorbg:
Follow-up to #7983


gregorbg:
For future reference: The columns were renamed in the aforementioned PR, but the indices on those columns were not recomputed correctly. This migration forces MySQL to recompute the indices, but it does not change the schema (except for the fact that Rails auto-names the second index differently these days).

The fact that the schema seems unchanged is intentional!

PR #6374, opened 2021-07-15
Added Swedish Translation

notrichardpeng:
Shoutout to Viktor Zenk for his amazing work! 🎉🎉

PR #8232, opened 2023-08-23
Add catalan translation

dunkOnIT:
Based on #6374. Still waiting on a reply from the translation submitter to add their user_id to `/controllers/translations_controller.rb`.

Please let me know if anything else is required here - this is a best guess attempt based on the Wiki docs and the linked PR :D


dunkOnIT:
AAAAAH I didn't branch off main. Hold for stupidity.

PR #8234, opened 2023-08-24
Add Catalan translation

dunkOnIT:
Based on https://github.com/thewca/worldcubeassociation.org/pull/6374. Still waiting on a reply from the translation submitter to add their user_id to /controllers/translations_controller.rb.

Please let me know if anything else is required here - this is a best guess attempt based on the Wiki docs and the linked PR :D


WcaOnRails/app/assets/javascripts/fullcalendar/locales/ca.js
```diff
@@ -0,0 +1 @@
+//= require fc_locales/cs.js
```
gregorbg, line 1: `cs` is Czech. You should import `ca`.

WcaOnRails/app/assets/javascripts/locales/ca.js
```diff
@@ -0,0 +1 @@
+//= require moment/cs.js
```
gregorbg, line 1: `ca`, not `cs`. See my other comment.

gregorbg (changes requested):
You also need to teach our website overall that "Catalan exists" by adding an entry to `WcaOnRails/config/locales.rb`.

Plus I think the error output from the CI tries to tell you that the YAML syntax is malformed somewhere around line 850.

WcaOnRails/app/assets/javascripts/fullcalendar/locales/ca.js
dunkOnIT: Silly mistakes - thanks!

dunkOnIT:
YAML fixed, all changes made. Only thing I'm no sure about if the flag_id for Catalan - but it doesn't really look like this is used anywhere in the code (which is for the best imo)

gregorbg:
The correct **country** locale (as opposed to **language** locale) is `es-ct` (or `es-CT` in some circumstances)

gregorbg:
Thanks for the YML fix, please correct the country code as well and then this can go live

dunkOnIT:
I've updated the flag_id to `es-ct` - not clear on whether the key in `locales.rb` should also be updated to `es-ct` (and placed under `es`. Based on your description I think so, but I'm not clear how a relationship is created between the locale and the translation if they don't share names/keys, so have left it as-is for no.

gregorbg:
Most parts in our code, including the keys in `locale.rb`, reference the **language**. That's the code `ca`.

The flag is an exception because it references the **region** which may or may not have a different ISO code.

As stated above, we generally care about languages so `ca` is fine unless explicitly required otherwise.

dunkOnIT:
Ok thanks, that makes sense and I've added a clarifying note in our translations documentation as well.

dunkOnIT:
Test failures may be from the new edit events tests merged into main (the failure looked familiar to the tests I looked through), so merging master and then re-pushing

PR #8238, opened 2023-08-28
Bump shakapacker to v7.0.3

gregorbg:
Manual bump because Dependabot cannot figure out that both need to be updated together across dependency ecosystems (across Bundler AND Yarn)


gregorbg:
Treating as dependency update. Merging.

PR #8240, opened 2023-08-28
Record remote timestamp in our DB when processing backed-off Stripe webhooks after server failure

gregorbg:
This aims to hotfix a problem with the async Stripe integration.

## Problem

The registrations table uses the `created_at` timestamp of `registration_payment` objects for ordering "who paid first" in the organizer view. This `created_at` timestamp is created by Rails internally when the object is first instantiated in the database. It normally happens automagically "under the hood".

When the reg system crashes under high load, we cannot receive Stripe Webhooks. Stripe realizes that we send an error response code and waits some amount of time before it tries sending again. We then recover and receove the webhook event after the backoff time. Then, when creating the `registration_payment` object, Rails records the `created_at` timestamp from after the recovery.

## Solution

Neither the PaymentIntent nor the Charge object in the Stripe API contain any `completed_at` timestamp or similar information. (This is a bit unfortunate)

However, every Webhook event carries a timestamp of when it was _originally created_ at Stripe. This timestamp stays the same even if Stripe re-tries sending with some exponential backoff. This PR introduces an additional column for the `created_at` according to Stripe and uses it when writing async `registration_payment`s.


WcaOnRails/app/controllers/registrations_controller.rb
```diff
@@ -508,23 +508,29 @@ def stripe_webhook
 
       stored_intent = stored_transaction.stripe_payment_intent
 
-      stored_intent.update_status_and_charges(stripe_intent, audit_event) do |charge_transaction|
+      stored_intent.update_status_and_charges(stripe_intent, audit_event, audit_event.created_at_remote) do |charge_transaction|
```
FinnIckler, line 511: Why do you need to give the method `audit_event.created_at_remote` if it already has audit_event, can't you just use action.source.created_at_remote in the update_status_and_charge method?

FinnIckler (approved):
Just a question why we explicitly have the third variable if the field is already in the second

WcaOnRails/app/controllers/registrations_controller.rb
gregorbg: Because in general, it cannot be determined that the source timestamp always comes from `action_source`. For example, if the servers don't crash and the user is successfully redirected to the `redirect_url` in Stripe, then the `action_source` is set to be the `User` object of the user who paid. That user does not have a `created_at_remote` timestamp. So we cannot make any generalised assumptions here

PR #8242, opened 2023-08-30
Fix etag not updating on wcif patch when only associations have changed

### LINKED ISSUE #8160: Wcif etag doesn't update
**Describe the bug**
When patching a wcif, the etag doesn't update. This is only noticed when actively benefiting from caching a wcif either locally, with cloudfront, or both, the wcif can get significantly out of sync when dealing with multiple versions with the same etag.

**To Reproduce**
1. Go to a competition on competitiongroups.com
2. Notice what the etag is
3. Make updates to the competition
4. Wait 5 minutes and see that the etag doesn't update. One should expect to see that fresh content gets fetched when hard reloading but then stale content is still loaded on soft reloads.

**Expected behavior**
A different etag should be generated whenever a wcif changes.

FinnIckler:
While we fixed setting the last modified in #8126, there is still the issue that the browser does not update the wcif correctly in some cases. If two wcifs have the same etag (because associations are not considered when computing the etag), the browser will still use their cached version even if the last-modfied header is updated. We tested this and documented this in #8160. 
This PR closes #8160 by adding assignments to the competition associations, I am not 100% if this is enough or it needs even deeper nesting (like `:schedule_activity`). Would love some input on this @gregorbg 

Another solution would be to set the etag computation to
```
if stale?(etag: [competition, competition.updated_at], last_modified: competition.updated_at, public: true)
      render json: Rails.cache.fetch(cache_key, expires_in: 5.minutes) {
        competition.to_wcif
      }
    end
```
This kinda feels silly, but TIL that Ruby does not use the updated_at in its hash function for computing these, maybe it's not in there because it doesn't end up in the wcif? Just giving rails the competition object is definitely a bit too automagic for my taste

### DIFF
--- a/WcaOnRails/app/controllers/api/v0/competitions_controller.rb
+++ b/WcaOnRails/app/controllers/api/v0/competitions_controller.rb
@@ -108,7 +108,7 @@ class Api::V0::CompetitionsController < Api::V0::ApiController
   def show_wcif_public
     id = params[:competition_id] || params[:id]
     cache_key = "wcif/#{id}"
-    competition = competition_from_params(associations: { registrations: [:assignments]})
+    competition = competition_from_params(associations: { registrations: [:assignments] })
     expires_in 5.minutes, public: true
     if stale?(competition, public: true)
       render json: Rails.cache.fetch(cache_key, expires_in: 5.minutes) {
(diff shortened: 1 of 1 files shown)


gregorbg:
I don't fancy the "include associations" approach. It requires us to take note of whenever the WCIF schema changes, and then potentially update the etag here. I much prefer the approach of including `updated_at` in the etag, as per the code snippet in your original post.

gregorbg:
Relevant: https://guides.rubyonrails.org/caching_with_rails.html#strong-v-s-weak-etags

FinnIckler:
I also read about strong etags and was thinking about including them in my original suggestion, the issue is that rails gives like no concrete result on what they mean with semantic vs byte wise equivalence here

PR #8284, opened 2023-09-05
Update node installation in Dockerfile

FinnIckler:
Node has deprecated its installation scripts in favour of installing via packages. See https://github.com/nodesource/distributions#debian-and-ubuntu-based-distributions.

### DIFF
--- a/WcaOnRails/Dockerfile
+++ b/WcaOnRails/Dockerfile
@@ -4,11 +4,20 @@ EXPOSE 3000
 ENV DEBIAN_FRONTEND noninteractive
 WORKDIR /app
 
-RUN apt-get update && apt-get install -y curl gnupg
-
 # Add PPA needed to install nodejs.
-# From: https://github.com/nodesource/distributions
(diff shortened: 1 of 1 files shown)


gregorbg (changes requested):
The libffi change is tricky. The problem is not libffi8 itself. The ffi Ruby gem is well capable of compiling against libffi8. The actual issue is that in our Rails dev setup, we mount the gems as a shared volume, which means that we persist gems across Ruby versions. So, at some point prior to using the Ruby 3.2.2 image, libffi7 was still there and the ffi gem's native extensions were built against libffi7. Now whenever we upgrade our Ruby container, the gems are reused from the previous installation, where of course libffi7 gems cannot run under libffi8 unless they're specifically compiled against it.
The actual solution here is to remove/clear your shared Docker volume that holds the Ruby gems and let Ruby install everything from scratch. Then libffi8 will work.

FinnIckler:
ok, deleting the volumes and rerunning worked for me, so this is only about the deprecated node installation now

gregorbg (approved):

PR #8286, opened 2023-09-05
Remove /relations

FinnIckler:
Remove this feature because
- We don't even link to it anymore
- Less than 20 people use it per week
- It requires an entire Table that has to be re-computed on CAD
- As a feature, it should exist on statistics
- It removes a sprocket asset (slick.min) and a use wca-autocomplete

Do I have to run a migration to remove the table?


gregorbg:
lib/compute_linkings.sql can also go.

FinnIckler:
Funny thing: This files was required in random places, like the city_validator_spec or the url_validator_spec. I am pretty sure that was a mistake!
Also I am keeping the translations in the file?

FinnIckler:
removed all mention of them!

gregorbg:
You also need
- A migration to remove the table
- Change the lib/database_dumper.rb file and remove the few lines pertaining to the now-dropped table

FinnIckler:
with the new schema.rb I don't need to add my migration to a list of ran migrations anymore right?

gregorbg:
> with the new schema.rb I don't need to add my migration to a list of ran migrations anymore right?

Well, kinda. There's no explicit list anymore, but the top of the file details which is the _latest_ migration that has been run. You likely need to bump that timestamp.

The only exception is when you're working on some migration-relevant PR for so long, that an even newer migration gets merged to the main branch before you finish work on your PR. In that case, the other newer PR will bump the date in the header of the file and you don't need to worry about it.

PR #8288, opened 2023-09-06
Have registrations export generate wcif_status instead of single-letter

dunkOnIT:
Still need to test locally - but I think if we just return the WCIF status in the endpoint we improve the CSV export considerably.

# Manual tests:
- [x] Exporting registrations in different states returns the WCIF states in the CSV
- [ ] Importing registrations with different WCIF states adds only accepted registrations (existing behaviour)
- [ ] Importing registrations with "a" states adds them as registrations (backwards compatibility)

### DIFF
--- a/WcaOnRails/app/views/registrations/export.csv.erb
+++ b/WcaOnRails/app/views/registrations/export.csv.erb
@@ -3,7 +3,7 @@
 <%= CSV.generate_line(headers).html_safe -%>
 <% @registrations.each do |registration| %>
 <%= CSV.generate_line([
-  registration.wcif_status
+  registration.wcif_status,
   registration.name,
   registration.country.id,
   registration.wca_id,
--- a/WcaOnRails/app/views/registrations/export.csv.erb
+++ b/WcaOnRails/app/views/registrations/export.csv.erb
@@ -3,7 +3,7 @@
 <%= CSV.generate_line(headers).html_safe -%>
 <% @registrations.each do |registration| %>
 <%= CSV.generate_line([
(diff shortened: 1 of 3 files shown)


WcaOnRails/app/views/registrations/export.csv.erb
```diff
@@ -3,7 +3,7 @@
 <%= CSV.generate_line(headers).html_safe -%>
 <% @registrations.each do |registration| %>
 <%= CSV.generate_line([
-  registration.pending? ? "p" : "a",
+  registration.wcif_status
```
gregorbg, line 6: You forgot a comma at the end of the line

gregorbg (changes requested):
Syntax changes requested.

Also consider that this may have broad implications for people who currently rely on the CSV export. Might be worth digging through some logs to see how many times the registrations CSV route was actually hit, and sending an announcement email.

gregorbg (approved):
Change fine per se, but please hold off with merging until we figured out how broad the implications of this change actually are (i.e. how many people rely on this CSV export)

dunkOnIT:
The current plan is to create backwards compatibility for CSV _imports_ by supporting either the letter or the status - that's the current blocker for this PR. I'll also ask around for anyone using the export function, though

gregorbg:
Kindly pinging this PR

dunkOnIT:
Pretty low priority for me compared to the Registration Service work, but I may pick this up as part of the WCIF/Registration Service integration I'll be starting soon

dunkOnIT:
Closing (at last!) in favour of #11165

PR #8293, opened 2023-09-08
Dump schema_migrations in dev_dump

gregorbg:
Because it's not implicitly imported through 'structure.sql' anymore


FinnIckler:
For everyone who loaded their dev dumps in between now and a couple weeks ago: this is why the migrations show up incorrectly

FinnIckler (approved):

Issue #7917, opened 2023-05-20
Remove 100% Volunteer References


EdHollingdale:
Home page: Support the WCA section says "As a 100% volunteer-run nonprofit", I propose saying "as a volunteer led nonprofit"

About page: Donation box at the top says "100% volunteer-run", propose saying "volunteer led", Who we are section says "entirely run by volunteers", propose saying "predominantly run by volunteers".

Speedcubing History page: says "the WCA remains a fully volunteer-run organization", propose saying "the WCA remains a fully volunteer-led organization"

There may be more references that I missed.

PR #8296, opened 2023-09-10
Change 100% volunteer references

dunkOnIT:
Addresses #7917

### DIFF
--- a/WcaOnRails/app/views/static_pages/speedcubing_history.html.erb
+++ b/WcaOnRails/app/views/static_pages/speedcubing_history.html.erb
@@ -58,7 +58,7 @@
   </p>
 
   <p>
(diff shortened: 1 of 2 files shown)


gregorbg:
This should be approved by the WCA Board.

dunkOnIT:
Cool - my suggestion is to include it for approval in the next Report? Shout if you want me to raise it with them sooner.

gregorbg:
Shout! No need to glue this into the report, just send them a separate email.
Increases our chance of getting an earlier reply.

dunkOnIT:
Shout acknowledged :P Just confirming you're aware it has already been approved by WCT in "Wording of "100% Volunteer" References"?

gregorbg:
Thanks for the reminder, but actually WCT may not be able to evaluate legal consequences this may or may not have for our non-profit status.

I may be hugely overcomplicating things here, but better be safe than sorry.

KerrieJarman (approved):
Confirmed with other Board members, all good

PR #8300, opened 2023-09-10
Created groups and roles tables, added delegate_probations to groups

danieljames-dj:
This is the first implementation of the new models of groups and roles. To start with, I have selected 'delegate_probations' as the code is going to be relatively easier.


WcaOnRails/app/controllers/delegates_controller.rb
```diff
@@ -7,4 +7,29 @@ class DelegatesController < ApplicationController
   def stats
     @delegates = User.delegates.includes(:actually_delegated_competitions)
   end
+
+  def probations
+    @probation_roles = Role.where(group_id: Group.where(group_type: "delegate_probation"))
+    @probation_users = {}
+    @probation_roles.each { |probation_role|
+      @probation_users[probation_role.user_id] = User.find_by_id(probation_role.user_id)
+    }
```
gregorbg, line 16:
```suggestion
    @probation_users = @probation_roles.index_with { |probation_role|
      User.find_by_id(probation_role.user_id)
    }
```

WcaOnRails/app/controllers/delegates_controller.rb
```diff
@@ -7,4 +7,29 @@ class DelegatesController < ApplicationController
...
+
+  def start_probation
+    wca_id = params[:wcaId]
+    user = User.find_by_wca_id!(wca_id)
+    role = Role.new(
+      user_id: user.id,
+      group_id: Group.find_by!(name: "Delegate Probation").id,
+      start_date: Date.today,
+    )
+    role.save!
```
gregorbg, line 27:
```suggestion
    role = Role.create!(
      user_id: user.id,
      group_id: Group.find_by!(name: "Delegate Probation").id,
      start_date: Date.today,
    )
```

gregorbg: Saves you the additional `save`

WcaOnRails/app/controllers/delegates_controller.rb
```diff
@@ -7,4 +7,29 @@ class DelegatesController < ApplicationController
...
+  end
+
+  def end_probation
+    probation_role_id = params[:probationRoleId]
+    role = Role.find_by_id(probation_role_id)
+    role.update(end_date: Date.today)
```
gregorbg, line 33:
```suggestion
    role.update!(end_date: Date.today)
```

gregorbg: The exclamation mark is important because it makes the `update` method throw an error in case the validations fail. The version without `!` simply returns `false`, and since you're not checking the return value it would fail silently

WcaOnRails/app/models/role.rb
```diff
@@ -0,0 +1,4 @@
+# frozen_string_literal: true
+
+class Role < ApplicationRecord
+end
```
gregorbg, line 4:
```suggestion
class Role < ApplicationRecord
  belongs_to :user
end
```

gregorbg: That way, you can load `my_role = Role.find(123)` and then use `my_role.user` directly. Rails is smart enough to figure out the link as long as the database columns follow certain naming conventions (i.e. using `belongs_to :user` only works if there is a DB column `user_id`. If there wasn't, you would have to tell Rails explicitly how the reference column is called)

WcaOnRails/app/views/delegates/probations.html.erb
```diff
@@ -0,0 +1,4 @@
+<%= react_component("DelegateProbations", {
+    probationRoles: @probation_roles,
+    probationUsers: @probation_users,
```
gregorbg, line 3: In the long run, React components should be responsible for actively fetching data from Rails, instead of Rails passing the data.

Rails should really only pass a minimum amount of information to React. For example, the React component for showing a competition's scrambles receives _only_ the competition ID and then during runtime on client-side starts a request to the Rails backend to fetch scrambles based on the competition ID. It should not receive the scrambles directly during render.

Similarly, your React component should fetch the probation Delegates from the backend (probably using our custom data loading hooks). I'll accept the PR in its current state if a rewrite is too much work, but I still thought I'd let you know :)

WcaOnRails/config/routes.rb
```diff
@@ -167,6 +167,9 @@
   post 'upload/image', to: 'upload#image'
 
   get 'admin/delegates' => 'delegates#stats', as: :delegates_stats
+  get 'admin/delegate_probations' => 'delegates#probations', as: :delegate_probations
+  post 'admin/start_probation' => 'delegates#start_probation', as: :start_probation
```
gregorbg, line 171:
```suggestion
  post 'admin/start_delegate_probation' => 'delegates#start_probation', as: :start_delegate_probation
```

WcaOnRails/config/routes.rb
```diff
@@ -167,6 +167,9 @@
...
+  post 'admin/end_probation' => 'delegates#end_probation', as: :end_probation
```
gregorbg, line 172: `end_delegate_probation` similar to my other comment above

WcaOnRails/db/migrate/20230910114732_add_probations_to_roles.rb
```diff
@@ -0,0 +1,15 @@
...
+class AddProbationsToRoles < ActiveRecord::Migration[7.0]
+  def change
+    Team.probation.team_members.each do |team_member|
+      role = Role.new(
+        user_id: team_member.user_id,
+        group_id: Group.find_by!(name: "Delegate Probation").id,
+        start_date: team_member.start_date,
+        end_date: team_member.end_date,
+      )
+      role.save!
```
gregorbg, line 12:
```suggestion
      role = Role.create!(
        user_id: team_member.user_id,
        group_id: Group.find_by!(name: "Delegate Probation").id,
        start_date: team_member.start_date,
        end_date: team_member.end_date,
      )
```

WcaOnRails/db/schema.rb
```diff
@@ -1120,6 +1145,8 @@
 
   add_foreign_key "active_storage_attachments", "active_storage_blobs", column: "blob_id"
   add_foreign_key "active_storage_variant_records", "active_storage_blobs", column: "blob_id"
+  add_foreign_key "groups", "groups", column: "parent_group_id"
+  add_foreign_key "roles", "groups"
```
gregorbg, line 1149: It's a bit curious that the foreign key for roles -> users is not listed here. Can you run `bin/rails db:schema:dump:primary` again to make sure that this is really intended by Rails? (and also make sure that the foreign key is actually present in your local DB and you're not on some work-in-progress state where the FK isn't there yet)?

gregorbg (changes requested):
Thanks! Makes a lot of sense based on your initial suggestion. Mostly Rails nitpicks, otherwise all good :)

WcaOnRails/app/webpacker/components/DelegateProbations/index.jsx
```diff
@@ -0,0 +1,101 @@
...
+export default function DelegateProbations({ probationRoles, probationUsers }) {
+  const [wcaId, setWcaId] = React.useState('');
+
+  function startProbation() {
+    fetchWithAuthenticityToken(startProbationUrl, {
+      method: 'POST',
+      body: JSON.stringify({ wcaId }),
+      headers: { 'Content-Type': 'application/json' },
+    }).then(async () => {
+      window.location.reload();
```
coder13, line 16: This likely goes along with Gregor's mention of how component should `get` data but it'll be real nice when that's implemented, this should just refetch a specific request as opposed to making a full page refresh (poor UX)

WcaOnRails/app/controllers/delegates_controller.rb
danieljames-dj: Done.

WcaOnRails/app/controllers/delegates_controller.rb
danieljames-dj: Done.

WcaOnRails/app/controllers/delegates_controller.rb
danieljames-dj: Done.

WcaOnRails/app/models/role.rb
danieljames-dj: Cool, thanks. Done.

WcaOnRails/config/routes.rb
danieljames-dj: Done.

WcaOnRails/config/routes.rb
danieljames-dj: Done.

WcaOnRails/db/migrate/20230910114732_add_probations_to_roles.rb
danieljames-dj: Done.

WcaOnRails/db/schema.rb
danieljames-dj: I tried running, but it didn't come. When I checked my local sql, even there the foreign key is not mentioned. Can it be because I did something wrong in migration file?

danieljames-dj (commented):
Thanks for the review @gregorbg and @coder13. I've fixed the suggestions on non-react part except the foreign key which I couldn't find the reason. @gregorbg can you please let me know if my migration is correct?

Regarding the rails changes, I missed to inform that I didn't pay much attention to react code, even now I haven't fixed the react part as I had only less time to spend on this PR today. I'll make the react changes soon.

WcaOnRails/app/models/group.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class Group < ApplicationRecord
+  enum group_type: {
```
gregorbg, line 4: Super nit-pick, but I thought I'd mention it: The Rails 6 way of doing enums is:

```ruby
enum my_column, {
  foo: "foo",
  bar: "bar",
}
```

Note the **comma** in `enum :my_column, {}` as opposed to `enum my_column: {}`. In our code, we use the latter notation a lot because it was created before Rails 6. But now at least with new features we might as well stick to the new notation.

WcaOnRails/db/migrate/20230910105700_create_roles.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class CreateRoles < ActiveRecord::Migration[7.0]
+  def change
+    create_table :roles do |t|
+      t.integer :user_id, null: false, foreign_key: { to_table: :users }
```
gregorbg, line 6: What about just `t.references :user`?

WcaOnRails/db/schema.rb
gregorbg: Maybe it's because in your migration file, you're listing `t.references` for most foreign keys but `t.integer :user_id` for users. Also see my other comment directly in the migration file

WcaOnRails/db/migrate/20230910105700_create_roles.rb
danieljames-dj: Worked with `t.references :user, type: :integer, null: false, foreign_key: { to_table: :users }` :-) Thanks Greg.

WcaOnRails/db/schema.rb
gregorbg: Yeah, sometimes there are legacy problems like the one you described. My guess is that Rails used to use `integer` as a default value at some point in time, and then switched to `bigint` later. Now tables that have been created long ago still use `integer` ID columns, and the reference type has to be changed accordingly.

WcaOnRails/app/controllers/delegates_controller.rb
gregorbg:
```suggestion
    user_ids = @probation_roles.pluck(:user_id)
    @probation_users = User.find(user_ids).index_by(&:id)
```

danieljames-dj: Done, thank you.

WcaOnRails/app/controllers/delegates_controller.rb
```diff
@@ -7,4 +7,56 @@ class DelegatesController < ApplicationController
...
+    respond_to do |format|
+      format.json do
+        @probation_roles = Role.where(group_id: UserGroup.where(group_type: "delegate_probation"))
+        @probation_users = {}
+        @probation_roles.each { |probation_role|
+          @probation_users[probation_role.user_id] = User.find_by_id(probation_role.user_id)
+        }
+        render json: {
+          probationUsers: @probation_users.as_json,
+          probationRoles: @probation_roles.as_json,
```
gregorbg, line 27: The separate `as_json` calls are redundant

WcaOnRails/app/controllers/delegates_controller.rb
```diff
@@ -7,4 +7,56 @@ class DelegatesController < ApplicationController
...
+        }
+      end
+    end
+  end
```
gregorbg, line 31: This complete method is duplicated with `def probations` above. One of them should go, it definitely shouldn't be necessary to compute the same data in two different places,

gregorbg: Looking at the rest of the code, the part in `def probations` seems to be the one that is unnecessary.
Use the short, concise syntax from there and put it under `delegate_probation_data` instead.

WcaOnRails/app/webpacker/components/DelegateProbations/index.jsx
```diff
@@ -0,0 +1,95 @@
...
+
+export default function DelegateProbations() {
+  const [wcaId, setWcaId] = React.useState('');
+  const {
+    data, loading, error, sync,
+  } = useLoadedData(delegateProbationDataUrl);
+  const { save, saving } = useSaveAction();
+
+  if (loading || saving) return 'Loading...'; // No i18n because this page is used only by WCA Staff.
+  if (error) throw error;
```
gregorbg, line 66: Don't throw in React components. Semantic UI has `Message` components that have special usage examples dedicated to displaying errors.

WcaOnRails/app/webpacker/components/DelegateProbations/index.jsx
```diff
@@ -0,0 +1,95 @@
...
+            </Table.Cell>
+          </Table.Row>
+        ))}
+      </Table.Body>
+    </Table>
+  );
+}
+
+export default function DelegateProbations() {
+  const [wcaId, setWcaId] = React.useState('');
```
gregorbg, line 59: We have a self-written `useInputState` hook somewhere under `lib`. It will allow you to use `setWcaId` directly in the `<Input>` component below.

WcaOnRails/app/webpacker/components/DelegateProbations/index.jsx
```diff
@@ -0,0 +1,95 @@
...
+
+  if (loading || saving) return 'Loading...'; // No i18n because this page is used only by WCA Staff.
+  if (error) throw error;
+
+  const { probationRoles, probationUsers } = data;
+
+  return (
+    <>
+      <h1>Delegate Probations</h1>
+      <Input value={wcaId} onChange={(e) => setWcaId(e.target.value)} placeholder="Enter WCA ID" />
```
gregorbg, line 73: As written above, you can pass `setWcaId` directly to onChange (like so: `<Input onChange={setWcaId} />` if you create it using `useInputState`.

WcaOnRails/app/webpacker/components/DelegateProbations/index.jsx
```diff
@@ -0,0 +1,95 @@
...
+      <ProbationListTable
+        roleList={probationRoles.filter((probationRole) => probationRole.end_date === null)}
+        userMap={probationUsers}
+        isActive
+        save={save}
+        sync={sync}
+      />
+      <h2>Past Probations</h2>
+      <ProbationListTable
+        roleList={probationRoles.filter((probationRole) => probationRole.end_date !== null)}
```
gregorbg, line 89: This should explicitly filter for `end_date < now`. There may be probation entries where for example the WDC or the Board decides that the Delegate will be on probation for 3 months, when they can already set the probation end to be a future date.

gregorbg (changes requested):
Excellent use of the load/save hooks! :smile: 

Still some nitpicking, as usual.

WcaOnRails/app/webpacker/components/DelegateProbations/index.jsx
danieljames-dj: I found Errored component which internally uses Message component. So I've changed to that now.

WcaOnRails/app/webpacker/components/DelegateProbations/index.jsx
danieljames-dj: I actually didn't consider that case, and had option to only 'end probation'. I've changed it now to allow to add end date and hence the mentioned case is covered now.

WcaOnRails/app/controllers/delegates_controller.rb
```diff
@@ -7,4 +7,50 @@ class DelegatesController < ApplicationController
...
   end
+
+  def delegate_probation_data
+    respond_to do |format|
+      format.json do
+        @probation_roles = Role.where(group_id: UserGroup.where(group_type: "delegate_probation"))
+        @probation_users = {}
+        @probation_roles.each { |probation_role|
+          @probation_users[probation_role.user_id] = User.find_by_id(probation_role.user_id)
+        }
```
gregorbg, line 18: This uses the "ugly" `each`-approach. You should use the more efficient version from the now-deleted `def probations` method (see our previous GitHub conversations)

gregorbg (changes requested):
You accidentally deleted the wrong part of the code in `delegates_controller`. Otherwise, we're good to go! :smile:

WcaOnRails/app/controllers/delegates_controller.rb
danieljames-dj: Oops yes sorry, I've changed it now. Hope it's good now.

gregorbg (approved):

PR #8303, opened 2023-09-11
Fix schema migration loading from temporary working dir in dev_dump

gregorbg:
See comment in code to figure out why this is necessary. In the long run, we should perhaps get rid of temp directories anyways...


FinnIckler (approved):
Thanks for the explanation this morning. LGTM!

PR #8304, opened 2023-09-11
Remove key syncing to machine from GitHub

FinnIckler:
Instead of relying on GitHub Keys to provide access to our Machines we will be using AWS.
If you like to use SSH:
- Make sure your aws cli is up to date (minimum Version 2.12)
- run `aws ec2-instance-connect ssh --instance-id i-0e94e30c5f1912525 --os-user cubing` if you want to connect to prod
- You can get the instance-id of other instances by running `aws ec2 describe-instances --filters 'Name=tag:Name,Values=INSTANCE_NAME' --output text --query 'Reservations[*].Instances[*].InstanceId'` or get them from the console

If you like to use the browser
- If it's important to get all the environment variables for the user cubing: Click on the Instance, press Connect -> Instance Connect and choose the user cubing
- If not: Click on the Instance, press Connect -> Session Manager -> Connect

I will provide new Instructions when we use ECS, which allows to connect to the docker services directly


FinnIckler:
Should I also remove the cubing password from the secrets in this PR?

gregorbg:
In that case, you can also remove the part of the Chef `default.rb` recipe that sets the cubing user password (and the stuff around it)

FinnIckler:
Not sure what will happen if we just set no password, we don't want people to be able to ssh into the instance without it.

gregorbg:
Good argument, but then you shouldn't remove the cubing password from Knife either

FinnIckler:
I vote for just disabling password login on the ssh agent on the machine

gregorbg:
Where does that leave us in terms of cubing user and password?

FinnIckler:
That would leave us with removing it. 
As far as I can see you can just do this like in wca.json
```
{
  (...)
    "openssh": {
      "server": {
        "password_authentication": "no"
      }
    },
  (...)
}
```

gregorbg:
Go for it!

gregorbg (approved):

PR #8306, opened 2023-09-12
switched dev and results export timing

dunkOnIT:
Dev export should run every day, while results export should run every 3 days. These schedles were switched in schedule.yml


gregorbg:
What makes you think the schedules were changed? I am quite confident that dev every 3 days and results every morning is correct.

dunkOnIT:
Oops sorry you're right, Finn and I got confused between them. Our bad!

PR #8307, opened 2023-09-12
Fix dockerfile being run on single line

FinnIckler:
Not sure why this worked on my machine, but we should not run all of these on a single line as it causes apt-get to error out.
I've also moved the NODE_MAJOR Version to ARG


dunkOnIT (approved):
works on my machine

PR #8309, opened 2023-09-12
Move phpmyadmin database auth to RDS IAM

FinnIckler:
This gets rid of the last secret handled by chef, which is why I deleted the databag.
The flow is still not fully tested, but I have set up the needed infrastructure (IAM policy and role) and ran these commands on the staging DB:
```
CREATE USER phpmyadmin IDENTIFIED WITH AWSAuthenticationPlugin AS 'RDS';
ALTER USER 'phpmyadmin'@'%' REQUIRE SSL;
GRANT ALL PRIVILEGES ON cubing.* TO 'phpmyadmin'@'%';

```


FinnIckler:
Current status: 
I can log into the phpmyadmin user on the cli by generating a token like this
```
RDSHOST="staging-worldcubeassociation-dot-org.comp2du1hpno.us-west-2.rds.amazonaws.com"
TOKEN="$(aws rds generate-db-auth-token --hostname staging-worldcubeassociation-dot-org.comp2du1hpno.us-west-2.rds.amazonaws.com --port 3306 --region us-west-2 --username phpmyadmin )"

mysql --host=$RDSHOST --port=3306 --ssl-ca=/etc/phpmyadmin/rds-combined-ca-bundle.pem --enable-cleartext-plugin --user=phpmyadmin --password=$TOKEN
```
But phpmyadmin seems to not send the password correctly, I think this might be because we are using phpmyadmin version 4.95 on staging which doesn't have this fix https://github.com/phpmyadmin/phpmyadmin/issues/16451 ? I will update and try again

FinnIckler:
Ok, I found out that staging still uses Ubuntu 20.04 where that is the latest version, so we need to either install by ppa or update to Ubuntu 22.04 like we did with prod. Don't have time to this today though.

gregorbg:
In the conversation of the issue that you linked, there's also a piece of information that...

> if you have already assumed an IAM role with permission to connect to a database user
before sending the auth token request, the response will come back with a length of about 700-800 chracters.

Could we use this to our advantage? This should guarantee <1000 chars as a "hotfix"

FinnIckler:
I just read that a second ago :D, let me try it out...

FinnIckler:
yeah not sure what AWS Support means, but it doesn't seem to work
```
root@staging:/tmp# aws sts get-caller-identity
{
    "UserId": "(...):i-08ccefec1ec6fe1ce",
    "Account": "(...)",
    "Arn": "arn:aws:sts::(...):assumed-role/wca_role-stg/i-08ccefec1ec6fe1ce"
}
root@staging:/tmp# TOKEN="$(aws rds generate-db-auth-token --hostname staging-worldcubeassociation-dot-org.comp2du1hpno.us-west-2.rds.amazonaws.com --port 3306 --region us-west-2 --username phpmyadmin )"
root@staging:/tmp# echo $TOKEN | wc
      1       1    1781
```
You can see in the aws sts get-caller-identity that I have assumed the role, but the token is still 1781 characters long

gregorbg:
Does the role `wca_role-stg` have the correct permissions set in the AWS Console?

FinnIckler:
yes, otherwise it would not be able to get generate the auth token in the first place

FinnIckler:
```
import boto3

region_name = 'us-west-2'

# RDS instance details
db_hostname = 'staging-worldcubeassociation-dot-org.comp2du1hpno.us-west-2.rds.amazonaws.com'
db_port = 3306
db_username = 'phpmyadmin'

# Create an RDS client using the assumed role credentials
rds_client = boto3.client(
    'rds',
    region_name=region_name,
)

# Generate the DB authentication token
token_response = rds_client.generate_db_auth_token(
    DBHostname=db_hostname,
    Port=db_port,
    DBUsername=db_username,
    Region=region_name,
)

print(f"Generated DB authentication has the length: {len(token_response)}")
```
This also creates a token with length 1780 on the machine

gregorbg:
> yes, otherwise it would not be able to generate the auth token in the first place

You are arguing that there is one permission and if you don't have it, you cannot even access the token in the first place. From the quoted reply of AWS support, it seems like there is/are other permission(s) where if you don't have it, you can still access the token but the length differs. Maybe there are more IAM shenanigans implied here?

gregorbg:
For example, there could be a difference for the permissions of the EC2 instance IAM user and the `phpmyadmin` user. Both should have independent permissions, can you make sure both are configured correctly? (Just guessing completely randomly and taking shots in the dark here...)

FinnIckler:
There are two ways of accessing AWS Resources, IAM Users and IAM Roles. An IAM User (usually a person) can assume a role to get more permissions for specific purposes. What AWS Support is referring to if an IAM User assumes a role that has the permissions to connect to the db, instead of the permissions being directly attached to the user. 
In our case, we have an EC2 instance (not a person), which automatically assumes the role defined in its instance profile (see [note](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html)).

There is no IAM User for the EC2 instance and there is no IAM User for phpmyadmin. Instead there is a trust relationship defined through an IAM Policy that is attached to the wca_stg role it looks like this
```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "rds-db:connect"
            ],
            "Resource": [
                "arn:aws:rds-db:us-west-2:(...):dbuser:db-(...)/phpmyadmin"
            ]
        }
    ]
}
```

gregorbg:
Okay, in that case just go with the semi-official PMA 5.2.2 image.

FinnIckler:
I don't want to create an IAM User which needs static credentials that then assumes the role, because that would introduce two more secrets which kinda defeats the point....

gregorbg:
Yup, absolutely. Don't introduce additional complexities! We know that a fix for our problem in the mainline releases is on the horizon, so we will dare to experiment a little bit with the not-quite-official PMA Docker images for now :smile:

FinnIckler:
Tested on Phpmyadmin version 5.2.2-dev+20230911.bd8b15ebb4

FinnIckler:
Before merging this we will need to update phpmyadmin on production. I followed the "upgrade manually" section on this guide https://doc.owncloud.com/server/next/admin_manual/installation/manual_installation/upgrade_install_phpmyadmin.html but used the https://files.phpmyadmin.net/snapshots/phpMyAdmin-5.2+snapshot-all-languages.zip link instead

FinnIckler:
Now it works. You can test the flow on staging.

gregorbg (approved):

PR #8315, opened 2023-09-15
Add file credentials to Vault

gregorbg:
Adds the ability to read stuff from Vault into a file instead of into a Ruby string. Useful for GCE application credentials.


FinnIckler (approved):
LGTM!

PR #8321, opened 2023-09-18
Migrate Timestamps table into ServerSettings table

gregorbg:
Most of `Timestamps` is now present in `CronjobStatistics`. As a replacement, we offer a more general table because we want to stop relying on git just "randomly existing" and instead properly hash the file contents


FinnIckler (approved):
LGTM!

PR #8322, opened 2023-09-19
Push cronjobs to non-concurrent queue via Sidekiq capsule

gregorbg:
Matches how there was only one single worker in the DelayedJob Supervisor setup.
Hope this solves some of the problems WRT has been facing.


FinnIckler (approved):
Looks correct according to https://github.com/sidekiq/sidekiq/blob/main/docs/7.0-Upgrade.md#capsules

PR #8325, opened 2023-09-20
Port avatar to react

campos20:
This was opened so another person can continue it


dunkOnIT:
Thanks Alexandre!

gregorbg:
I'll be handling the merge conflicts and building a backend for the React components

infra/wca_on_rails/staging/s3.tf
```diff
@@ -10,3 +10,9 @@ resource "aws_s3_bucket" "avatars" {
     "Name" = "wca-avatar"
   }
 }
+resource "aws_s3_bucket" "avatars_private" {
```
FinnIckler, line 13: you are declaring the bucket twice, if you want staging and production to have access to the bucket you need to create it in the shared folder, declare it as an output (and in variables.tf under the shared module) and use that output in the staging and production files. I can do it if you want!

gregorbg: I would actually like to take you up on this offer. TIA :smile:

FinnIckler: I'll do it now

FinnIckler:
Before I run terraform apply, should staging have access to the private bucket? If yes, do you have any safeguards in place that users can't somehow replace their avatars there?

gregorbg:
> Before I run terraform apply, should staging have access to the private bucket? If yes, do you have any safeguards in place that users can't somehow replace their avatars there?

There is no "smart" safeguard in place, but the `AVATARS_PUBLIC_STORAGE` and `AVATARS_PRIVATE_STORAGE` variables can control which storage backends (as defined in `config/storage.yml`) the server even tries to read from.

Can you configure the infrastructure in such a way that (a) staging gets its own private bucket and (b) the access from staging to the public bucket is read-only?

FinnIckler:
done!

FinnIckler:
I can run terraform whenever

gregorbg:
@thewca-bot deploy staging

gregorbg:
This has grown too big over time to demand a review from anybody.
Merging this on my authority.

maxidragon:
Default avatars seem to be broken. See example https://www.worldcubeassociation.org/persons/2023ROGA01

FinnIckler:
already a fix open at https://github.com/thewca/worldcubeassociation.org/pull/10083

PR #8326, opened 2023-09-20
Removed hungarian translation

dunkOnIT:
Has he done it, could this really be a one-commit wonder? 

(Removes Hungarian translation as requested by Bence and approved by Gregor)


gregorbg:
This would be a one-commit wonder if the task was to "blast the Hungarian translation into oblivion".

In this case, as we're fairly confident the translation will return soon :tm:, removing the mention from `locales.rb` and deleting the actual YML file would have been enough. You can keep the third-party translations under `assets/javascript` and you can also change the verified translator ID from 378 (Niki) to 368 (Bence) (yes it's a funny coincidence that their user IDs are so close)

gregorbg (approved):

PR #8330, opened 2023-09-22
Fix test choke due to race condition in EditEventsSpec

gregorbg:
The fact that we had previously set `wait_for_completion` to `false` meant that the test terminated faster than the WCIF `PATCH` in the background. When the test terminates, it clears its data from the test DB, and then the still-ongoing WCIF update in the background was trying to reload an entity that doesn't exist.

By using `accept_modal` in Capybara, we explicitly expect that a "some error occured during the PATCH" modal pops up, and only continue the test when the server in hte background is done.


gregorbg:
Merging as critical fix

PR #8331, opened 2023-09-22
Reinstate dummy version of Hungarian locale

gregorbg:
We were requested to remove the locale for a complete makeover, but this messed with users and mailers who had their preferred_locale set to 'hu'. So we just keep a dummy version to not make the code crash (and to save us from manually removing preferred_locale in the database for users that had it set to HU) until the new version is submitted


FinnIckler (approved):
let's try it out

PR #8333, opened 2023-09-23
Clean up unused code from EditEvents

gregorbg:
While playing around with Webpacker, I realized that the `rootRender` in EditEvents was no longer needed.

I also made one minor change to an `useState` invocation that looks more like it was supposed to be `useEffect` CC @coder13


WcaOnRails/app/webpacker/components/EditEvents/index.js
```diff
@@ -33,7 +38,7 @@ function EditEvents() {
     return null;
   }, [unsavedChanges]);
 
-  useState(() => {
+  useEffect(() => {
```
coder13, line 41: Ah yes! This is indeed supposed to be a useEffect

coder13 (approved):
These changes seem good!

PR #8335, opened 2023-09-24
Posting dashboard fix

danieljames-dj:
This PR does following changes in posting dashboard:

- Sorts competition based on time results are submitted at.
- No longer sends email notifying that results are in posting process.
- In import page, shows who is currently posting results if somebody is posting.
- Changed results email text to include posting dashboard.
- Added submission timestamp to posting dashboard (not so human readable format, will fix it to proper one sometime later).
- Changed text of buttons in posting dashboard.


WcaOnRails/app/controllers/admin/results_controller.rb
```diff
@@ -10,7 +10,7 @@ def posting_index
           render :posting_index
         end
         format.json do
-          @pending_competitions = Competition.pending_posting
+          @pending_competitions = Competition.pending_posting.sort_by { |comp| comp.results_submitted_at }
```
viroulep, line 13: I think we can let mysql do the sorting here and put `order(announced_at: :desc)` either here or right in the `pending_posting` scope.

WcaOnRails/app/webpacker/components/PostingCompetitions/index.js
```diff
@@ -92,10 +92,14 @@ function PostingCompetitionsIndex({
                 {' '}
                 <CountryFlag iso2={c.country_iso2} />
               </Header.Subheader>
+              <Header.Subheader>
+                {`Submission Timestamp: ${c.results_submitted_at}`}
+                {/* TODO: Convert the above timestamp to a readable way */}
+              </Header.Subheader>
```
viroulep, line 98: If the competitions are shown with the "oldest" pending first, do we even need to display the timestamp?

viroulep (commented):
Thanks for the quick fixes Daniel!
I left a couple of comments but definitely nothing major :)

WcaOnRails/app/webpacker/components/PostingCompetitions/index.js
danieljames-dj: Yes according to @SAuroux. I too think it will be good so that we can easily know from dashboard when is the oldest pending results.

WcaOnRails/app/controllers/admin/results_controller.rb
danieljames-dj: Thanks, done. (Hope you meant `.order(results_submitted_at: :asc)`)

danieljames-dj (commented):
Thanks @viroulep for reviewing.

WcaOnRails/app/webpacker/components/PostingCompetitions/index.js
SAuroux: @viroulep @danieljames-dj My thought here was that this often influences in what sets we are posting results (when we are not able to post all at once). Maybe instead of displaying raw results_submitted_at we can rather just show "Submitted N hours ago"?

danieljames-dj: Yes it's possible, done. 👍

WcaOnRails/app/controllers/admin/results_controller.rb
viroulep: I did, thanks for fixing my mind :)

viroulep (approved):
LGTM :)

I'd suggest squashing the two commits now that review is done, but it's definitely not strictly necessary ;)

PR #8336, opened 2023-09-24
Replace unofficial logo variants with official versions

nsilvestri:
In a number of places across the WCA website, an unofficial version of the WCA logo with a black border was used. This is not an official logo according to the [WCA Logo Usage Policy](https://www.worldcubeassociation.org/documents/policies/external/Logo%20Usage.pdf). 

I'm not sure where the outlined variant logo came from, as it has never been an official version of the logo, [at least as far back as 2012](https://web.archive.org/web/20130102145513/http://www.worldcubeassociation.org/sites/default/files/WCAlogo_notext.svg). Unfortunately this variant has spread across the internet and into numerous other resources, which now use unofficial versions of the logo. Fixing the version of the logo used by the WCA website itself should at least help this matter.

This PR replaces every incorrect version of the logo that I could locate with the official WCA logo.


FinnIckler:
Looks like the outlines logo was added in 2015 to improve readability: 
https://github.com/thewca/worldcubeassociation.org/commit/db92432cab1739bb500110a8730384ebeb40bec4
Are we sure that we are not decreasing readability with this change?

nsilvestri:
For reference, here's what it looks like with the replaced logo:

image

While I will concede that it does reduce readability and contrast, ultimately the purpose of the logo is to be recognizable, and not necessarily readable. I think it is still amply recognizable the WCA logo in this case.

A secondary purpose of the logo is as a canonical reference to the WCA's logo. The logo modification from 2015 has mistakenly been used as the logo for the WCA both internally on the WCA website and externally on places like the WCA's Wikipedia page, by people who "Open image in new tab"'d the logo. This has caused it to be used in contexts where the border is not necessary. For this reason I think it is important that whatever the final solution ends up being, the actual logo file used in that part of the website should be the official WCA logo.

If having a border is a must-have, there are some workarounds that can alleviate the issue somewhat without forcing us to create a new logo variant. Below is an experiment using CSS to place the logo on a black circle. There are still some non-functional modifications that need to be made to the logo, like making the white areas in some parts actually blank, but I think it may be able to be leveraged for the best of both worlds.

image

```css
element {
  border: 2px solid black;
  background-color: black;
  border-radius: 50%;
}
```

WMT will discuss this more fully internally and come back with a resolution of some kind.

lgarron:
The logo with the black borders was approved by the Board but never fully adopted. I will strongly advocate for it, as there are strong contrast issues with the "C" in WCA in a lot of media.

I would also suggest going with even thicker strokes:

Revised (Thicker)
Revised (Thicker)

lgarron:
Also, in case anyone is curious, some other prototyped alternatives with better contrast:

WCA Logo Color Letters
WCA Logo Dark C

lgarron:
Also, note that the policy covers usage of particular versions of the logo, but doesn't prevent the WCA from using other variants ourselves.

For another example, we've used a variety of favicons that are not the full logo. (In particular, the main website has never used the full logo as a favicon.)

gregorbg:
What's the current status on this PR?

nsilvestri:
There are some other logo-related changes that are slowly being made, and it put this PR on hold somewhat. I will close this and re-open if/when there is something actionable.

nsilvestri:
#9216 is moving forward with removing the border, but for more stringent legal reasons regarding our trademark on the WCA logo and not just internal WCA policy.

PR #8314, opened 2023-09-15
Replace Chef with ECS Deployment

FinnIckler:
After testing the docker-compose landscape, we have decided it makes more sense to go straight to ECS, as updating the images by ourselves is much more of a hassle
TODO:
- [x] Create the correct infra files for our ECS setup
- [x] test on staging
- [x] make a PR to this rep which builds images and triggers a deploy
- [x] Create New Load Balancer and reference it for the services
- [x] Create a mixed instance fleet to better support prod on a bigger instance

PR #8338, opened 2023-09-25
Add workflows for Building and Deploying Docker Images

FinnIckler:
In Preparation for #8314 I added two new workflows.
On Release:
- Builds new Images and marks them as latest for staging and prod
- Pushing an image triggers the deployment pipeline for prod
- Sidekiq image is updated and deployed
- Staging image is updated and deployed

On PR Comment:
- Allows deploying PRs to staging by writing '@thewca-bot deploy staging' 
- Replaces Main and Sidekiq Staging images if triggered

We could think of another way of deploying staging images, but I think this could work well. This almost makes the deploy repository redundant, depending on if we want to add a separate workflow for updating documents.

### DIFF
--- a/.github/workflows/deploy-on-release.yml
+++ b/.github/workflows/deploy-on-release.yml
@@ -1,0 +1,52 @@
+name: Deploy Image on GitHub Release
+on:
+  release:
+    types: [created]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v3
+        with:
+          aws-access-key-id: ${{ secrets.CI_CD_AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.CI_CD_AWS_SECRET_ACCESS_KEY }}
+          aws-region: ${{ secrets.AWS_DEFAULT_REGION }}
+      - name: Login to Amazon ECR
+        id: login-ecr
+        uses: aws-actions/amazon-ecr-login@v1
+      - name: Build and Push Sidekiq image
+        uses: docker/build-push-action@v4
+        with:
+          context: ./WcaOnRails
+          push: true
+          file: Dockerfile.sidekiq
+          tags: |
+            ${{ steps.login-ecr.outputs.registry }}/wca-on-rails:sidekiq-production
+            ${{ steps.login-ecr.outputs.registry }}/wca-on-rails:sidekiq-staging
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+        # this will trigger the deployment pipeline for prod
+      - name: Build and push Prod Image
+        uses: docker/build-push-action@v4
+        with:
+          context: ./WcaOnRails
+          push: true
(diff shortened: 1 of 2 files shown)


gregorbg:
> This almost makes the deploy repository redundant, depending on if we want to add a separate workflow for updating documents.

I think the documents repository is self-sufficient and they don't manage their deployments over `thewca/deploy`. But obviously they still need to maintain a way to update their documents independent of the main code, so you probably need a PR over there to adjust the commands to the new infrastructure.

Then the only remaining GH Workflow in the deploy repository is the one for Regulations. WRC definitely needs it, and the reasons why it lives in the deploy repository are:

1. It already had a lot of Secrets installed so it was more convenient to configure there, rather than moving it to the regulations repository.
2. We were not sure whether having a workflow that can be manually triggered in a public repository could be dangerous. External people defintely don't have permission to trigger workflows, but (I think) they do have permissions to view workflow logs. Could that be problematic in any way if we moved the Regulations workflow to the (public!) Regulations repository?

.github/workflows/deploy-on-release.yml
```diff
@@ -0,0 +1,52 @@
...
+          tags: |
+            ${{ steps.login-ecr.outputs.registry }}/wca-on-rails:latest
+            ${{ steps.login-ecr.outputs.registry }}/wca-on-rails:staging
+            ${{ steps.login-ecr.outputs.registry }}/wca-on-rails:${{ github.event.release.tag_name }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+        # Replace the old sidekiq image with the new one
+      - name: Deploy Sidekiq
+        run: |
+          aws ecs update-service --cluster wca-on-rails --service wca-on-rails-auxiliary-services --force-new-deployment
```
gregorbg, line 48: Here it seems like you're deploying the _production_ Sidekiq but...

.github/workflows/deploy-on-release.yml
```diff
@@ -0,0 +1,52 @@
...
+        # There is no pipeline for staging so we manually force to update the image
+      - name: Deploy staging
+        run: |
+          aws ecs update-service --cluster wca-on-rails --service wca-on-rails-staging --force-new-deployment
```
gregorbg, line 52: ...here in turn you're deploying the _staging_ image. What about production?

.github/workflows/staging-deployment-from-prs.yml
```diff
@@ -0,0 +1,69 @@
...
+        uses: docker/build-push-action@v4
+        with:
+          context: ./WcaOnRails
+          push: true
+          file: Dockerfile
+          tags: |
+            ${{ steps.login-ecr.outputs.registry }}/wca-on-rails:staging
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+      - name: Deploy staging if triggered
```
gregorbg, line 66: You've built Sidekiq but it seems like you're never deploying it.

FinnIckler: On staging there is only one service, so forcing `wca-on-rails-staging ` to update will deploy both the rails and the sidekiq image

.github/workflows/deploy-on-release.yml
FinnIckler: Production is deployed automatically on Docker image push through a blue-green deployment Pipleine

Issue #7724, opened 2023-03-07
Add WCA Live as a tab in the main menu on the website


dunkOnIT:
This is a piece of feedback that was pretty common on the survey - I think the current "Results -> WCA Live" flow is not intuitive/obvious to people. I definitely see merit in making it easier for people to quickly access WCA Live.

jfly:
For the record, this is what https://github.com/thewca/worldcubeassociation.org/issues/4642 is talking about. It's perspective is a little more egalitarian, though (ideally we doesn't assume anywhere that WCA Live is the platform being used)

dunkOnIT:
Thanks for the context there! It seems like there are two use cases here:
1. Easily navigating to WCA Live 
2. Navigating to the WCA Live page for your competition

In general, given how widely used it is, I don't necessarily have an issue with making WCA Live the default results platform on the homepage (Item 1).

#4642 covers the second case - there, too, I like the idea of easily setting up a WCA Live page as a default option, but definitely agree that it should be possible for organisers to go with a custom option.

PR #8339, opened 2023-09-25
Add WCA Live to Nav Bar (#7724)

Carter-Thomas:
Removed WCA Live from "Results" dropdown to its own separate header.

Here is a screenshot of the corrected Navbar with this commit: 

Screenshot 2023-09-25 174726

### DIFF
--- a/WcaOnRails/app/views/layouts/_navigation.html.erb
+++ b/WcaOnRails/app/views/layouts/_navigation.html.erb
@@ -55,8 +55,6 @@
           <span class="caret"></span>
         </a>
         <ul class="dropdown-menu" role="menu">
-          <li><a href="https://live.worldcubeassociation.org"><%= ui_icon('sort numeric down') %>WCA Live</a></li>
-          <li class="divider"></li>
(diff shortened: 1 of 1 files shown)


dunkOnIT:
I strongly support this change, but more discussion needed - particularly Gregor's input.

FinnIckler:
I am also fine with it. We could have it in both places?

dunkOnIT:
I've asked for feedback on the staff discord

dunkOnIT:
All positive feedback so far

dunkOnIT (approved):
Code looks good, strong support/no opposition from members of the staff Discord, and code looks good. I propose we merge @gregorbg

Carter-Thomas:
Some issues have been presented as seen in the unofficial staff server, I believe I forgot to add a span class which is causing the tab to remain unminimized when the screen gets smaller. Should I make another pr for this?

Here’s an example: 

8E8CE052-3FCA-42CC-8AA8-39C23CFA1C3A

dunkOnIT:
Please do Carter!

Issue #8340, opened 2023-09-26
Dynamically update competitor count in WCA about page


dunkOnIT:
See #8316 for discussion. 

Change the competitor count in the about page to take the current number of competitors, rounded down. So, if there are 258 951 competitor ID's , we would report "over 250 000 competitors".

Issue #8316, opened 2023-09-17
Update About the WCA Page Statistics


StanFromIreland:
The following sections should be updated on the [WCA about page](https://www.worldcubeassociation.org/about):
**Who we are**
...over 100,000 people...
**What's next for the WCA?**
...over 100,000 unique competitors across 140 different regions...

To:

**Who we are**
...over 200,000 people...
**What's next for the WCA?**
...over 200,000 unique competitors across 158 different regions...

### SQL Queries for verification:

_Total Number of Competitions_
`
mysql> SELECT COUNT(DISTINCT id) as TOTAL_IDS FROM Persons
`
|TOTAL_IDS            |
|--------------------|
|205792                 |

_Total Number of Countries Where there is at least one competitor_
`
mysql> SELECT COUNT(DISTINCT countryId) as COUNTRIES FROM Persons
`
|COUNTRIES            |
|--------------------|
|158            |

dunkOnIT:
Fair point - PR's welcome

Carter-Thomas:
Would this require changing the about page (currently static) to a dynamic page? Or could we just periodically change the numbers, as changes like this probably don't come too often?

gregorbg:
To be honest, having this number be dynamically generated based on the actual count in our DB sounds pretty cool. And it's also a nice and easy change for people looking to become more familiar with Rails

Carter-Thomas:
I made a commit for the static option, anyone else can feel free to make the dynamically updating page. Imo, having the about page have the exact member count may not look as professional and may kind of feel out of place. This could be done for countries in theory, but it's still quite a rare occurrence for a new country to gain its first competitor.

dunkOnIT:
I tend to agree that having a non-round number would look unprofessional. Perhaps we could round the live number down to the nearer 10,000 though - that way instead of "259 765 competitors", we would have "over 250 000 competitors". I'll create a new issue for this change, and propose we merge Carter's in the meantime.

dunkOnIT:
@Carter-Thomas you can make a PR for your static option and reference this issue - I've created a new one for the dynamic count

PR #8342, opened 2023-09-26
Update About Page Member and Country Counts

Carter-Thomas:
Country and competitor counts and the about page are now updated to reflect current counts.


FinnIckler:
I thought we would handle this dynamically as mentioned in https://github.com/thewca/worldcubeassociation.org/issues/8340

Carter-Thomas:
Check #8316

FinnIckler:
ah okay. Could have mentioned it in the PR :D

FinnIckler (approved):

PR #8350, opened 2023-09-29
Fixes Span Issues That New WCA Live Tab Had

Carter-Thomas:
Changed the new WCA Live header to minimize and maximize correctly according to screen resolution. Related to #8339.


WcaOnRails/config/locales/en.yml
danieljames-dj, line 1: Did these changes (other than wca_live) come in while rebasing? If so, can you please remove them?

danieljames-dj (changes requested):

WcaOnRails/config/locales/en.yml
Carter-Thomas: Everything in the commits involving this branch are up to date with the current repository if that's what your asking. The merge was just syncing the branch so that everything was up to date. Other than that, everything involved is related to the WCA Live issues

dunkOnIT: I agree with Dan that we shouldn't be seeing that changes in en.yml reflected as diffs - I think this might be a result of the fact that your local master branch has an open PR against it?

gregorbg: Just +1'ing the notion here that these en.yml changes should _not_ be included in this PR diff.
The fact that they are means that the branch which you're making a PR from is not up to date with the upstream branch that you're pointing the PR at.

Carter-Thomas: ugh I don't know why that specific change keeps getting included, I'll make a new branch and close this pr.

PR #8352, opened 2023-09-29
Throttle Sidekiq mailer threads

gregorbg:
Because otherwise they eat 5 by default, and cronjobs have 1 (linear execution), which is 5+1=6 and that's more than the Rails default for DB pooling :sweat_smile:


FinnIckler (approved):
LGTM!

PR #8400, opened 2023-10-02
Fix against changed numeric key export in i18n-js

gregorbg:
Our translation YML (the source _Ruby_ object) looks like this:

```yml
avatar_guidelines:
  '1': "All avatars should be pictures of real persons."
  '2': "The person shown must be your current self."
  '3': "If there is more than one person on the avatar, it should be clear who of them is you."
  '4': "Your head, but not necessarily the full face, should be visible."
  '5': "The avatar should not include overlay texts or unnatural filters, including borders."
  '6': "The avatar must be submitted in correct orientation."
  '7': "The avatar must not contain offensive content (e.g. offensive gestures)."
```

This is supposed to represent an array of objects, but the numbering is using stringified keys, mostly because of historic reasons. This clashes with changes introduced in https://github.com/fnando/i18n/pull/74, which leads to the JS export creating an _Object_.

Note that this is technically correct and the i18n library is not to blame here. Our broken legacy processes are.
However, we do rely on certain assumptions of translations being arrays when reading them in React during runtime. So we need to patch this behavior on our side.

NOTE: Of course, the "proper" patch would be to convert all instances in all of our translated YML files to "real" YML arrays. But this would be a ton of work and most notably it would prompt all of our translators to "please submit a new translation" when in reality, there is nothing to do.

### DIFF
--- a/WcaOnRails/app/webpacker/lib/i18n.js
+++ b/WcaOnRails/app/webpacker/lib/i18n.js
@@ -20,10 +20,10 @@ function tArray(scope, options) {
     const maybeNumericKeys = resKeys.map(Number);
     maybeNumericKeys.sort();
 
-    // Our i18n export library changed behavior in a minor version bump (sigh…) to maintain numeric keys as JS objects
-    // even when the keys clearly indicate index ordering, implying an array.
-    // We need to circumvent this behavior because we rely on external tools for our translators which require the YML
-    // to contain string keys (instead of changing the YML itself to a "proper" array)
+    // Our i18n export library changed behavior in a minor version bump (sigh…) to maintain numeric
+    // keys as JS objects even when the keys clearly indicate index ordering, implying an array.
+    // We need to circumvent this behavior because we rely on external tools for our translators,
+    // which require the YML to contain string keys (instead "proper" YML arrays)
     const isPseudoArray = maybeNumericKeys.every((key, idx) => key === (idx + 1));
 
     if (isPseudoArray) {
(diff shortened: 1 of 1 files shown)


dunkOnIT:
If this is a data formatting issue, is there not a possibility of using a script of even soliciting volunteers? If we can revert the package update and take some pain now to avoid adding more tech debt my vote would go towards that.

gregorbg:
In order to figure out the full scope of this problem, we first need to make sure _why_ exactly we choose this ugly "stringified keys" approach -- it most likely not only requires a script, but also lots of historical digging. So the choice here is between introducing work through technical debt (which is abstracted away behind a nice convenient method that is useful for exactly these kinds of scenarios) or introducing work through historical research. It's work in either case

FinnIckler:
I know that @jonatanklosko relies on this behavior in internationalize, maybe he knows why we use it and what we would need to do to replace it?

dunkOnIT:
Ah ok makes sense! If we don't have an understanding of what the historic reasoning for this is, I agree with the approach here (pending Jonatan's comments). I'm probably not the best person to review the code, but happy to do some local testing - would just need to know where this bug showed up (I'm guessing in lots of our {language}.yml text not displaying?

jonatanklosko:
Correct, Internationalize doesn't support arrays, and I'm pretty sure the same goes for translation status page. Using arrays, while being for a cleaner YAML, introduces much complexity.

FinnIckler (approved):
LGTM!

PR #8403, opened 2023-10-03
Improve ranking links in user profile

### LINKED ISSUE #8094: Add links to NR/CR/WR rankings on profile page
Instead of having a link to world ranking on the single/average columns of the personal records table, I think it makes more sense to have links on the NR/CR/WR numbers instead. For example, clicking on the single/CR column of the 3x3 row in 2012PARK03's profile should link to https://www.worldcubeassociation.org/results/rankings/333/single?region=_North%20America.

If the user does show up in the ranking after clicking on the link, I think it would also be nice to highlight the user's row and scroll to it to put it in view.

jacklee1792:
Closes #8094.

Links to the rankings page are put on the NR/CR/WR numbers instead, and redirect to their respective region-based rankings (national, continental, world). When one of these links is clicked, the "focus" parameter highlights the user's row in the rankings if applicable and scrolls it to the center of the viewport.

Example after clicking an "NR" ranking number on the user profile:

image

### DIFF
--- a/WcaOnRails/app/assets/stylesheets/results.scss
+++ b/WcaOnRails/app/assets/stylesheets/results.scss
@@ -49,4 +49,8 @@
   .cubing-icon {
     vertical-align: baseline;
   }
+
+  .highlight-focus {
+    background-color: $highlight-row-color;
+  }
 }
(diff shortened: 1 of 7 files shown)


jacklee1792:
@gregorbg I want to mention that this seems to have broken all filtering by region, and links on profiles seem to be cached (?), clicking on certain ranking numbers shows highlights for the wrong people. Might be worth reverting this for now

gregorbg:
Thanks for letting us know so fast! I have reverted the change on `master`.

Issue #8399, opened 2023-10-02
Migrate edit persons form to React with a modification


danieljames-dj:
Currently [edit persons page](https://www.worldcubeassociation.org/admin/edit_person) ([link to code](https://github.com/thewca/worldcubeassociation.org/blob/master/WcaOnRails/app/views/admin/edit_person.html.erb)) is written in `html.erb` and we have plans of improvements to the form. As a first step, it will be good if we have the page migrated to React.

Modification required:
- "Incorrect WCA ID claim count" field is no longer needed, instead when we click 'fix' or 'update', the claim count should be set to 0.

FinnIckler:
I actually already migrated the page, but got so wrapped up in the rest of the work I didn't make a PR. Do you want me to push it and you can take it over and finish it with the changes?

danieljames-dj:
Sure I can do that. Thanks Finn.

PR #8406, opened 2023-10-03
Migrate Edit Person To React

FinnIckler:
Halfway fixing #8399, pushing so Daniel can work on it. 

Work still needing to be done:
- Make sure the data is loaded correctly in UserSearch onSelect (line 27) [I am getting an error about wca cronjobs and read only data so I can't test it right now]
- Make the changes Daniel mentioned in #8399

UPDATE: Attaching the screenshot
Screenshot 2023-11-18 at 12 14 42


FinnIckler:
It is also important to now that I am replacing the old jquery based search with a UserSearch component based on Jeremy's OmniSearch.

danieljames-dj:
I'll continue with this after merging of https://github.com/thewca/worldcubeassociation.org/pull/7459, so that it will be easy for implementing the search component. Currently, the component is not expecting exactly as expected.

WcaOnRails/app/webpacker/components/Admin/EditPerson.js
```diff
@@ -0,0 +1,92 @@
...
+  const [name, setName] = useState('');
+  const [representing, setRepresenting] = useState('');
+  const [gender, setGender] = useState('');
+  const [birthdate, setBirthdate] = useState('');
+  const [incorrectClaimCount, setIncorrectClaimCount] = useState(0);
+  const [personFound, setPersonFound] = useState(false);
+  return (
+    <>
+      {/* eslint-disable-next-line react/no-danger */}
+      <div dangerouslySetInnerHTML={{ __html: description }} />
```
coder13, line 32: This should never be necessary. Even though this can't be exploited, the description text can just be put in the jsx.

danieljames-dj:
I've undraft-ed this PR as this is now ready for review.

kr-matthews (changes requested):
Reminder that screenshots are always helpful.

WcaOnRails/app/webpacker/components/Admin/EditPerson.js
danieljames-dj: The description is actually having HTML stuffs like p tag. That's why using this. Please let me know if this can be changed without causing warning.

WcaOnRails/app/webpacker/components/Admin/EditPerson.js
```diff
@@ -0,0 +1,92 @@
...
+      <UserSearch onSelect={(el) => {
+        const userId = el[0].id;
+        fetchJsonOrError(usersApiUrl(userId)).then(({ data }) => {
+          setPersonFound(true);
```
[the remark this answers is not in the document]
danieljames-dj: The search was optimized later in another PR, so  now we don't have to use this component anymore.

danieljames-dj (commented):
Done with the changes suggested by @kr-matthews. And sorry for missing screenshot again, I've added it now.

WcaOnRails/app/webpacker/components/Admin/EditPerson.js
kr-matthews:
```
return (
  <>
    <div>
      <p>The description.</p>
      <p>More text.</p>
    </div>
    ...
```

WcaOnRails/app/controllers/admin_controller.rb
```diff
@@ -291,43 +291,54 @@ def edit_person
...
-    if @person
-      person_params = params.require(:person).permit(:name, :countryId, :gender, :dob, :incorrect_wca_id_claim_count)
+    wca_id = params.require(:person).permit(:wcaId)[:wcaId]
+    person = Person.current.find_by(wca_id: wca_id)
+    success_message = nil
+    warning_message = nil
+    error_message = nil
+    if person
+      person_params = params.require(:person).permit(:name, :gender, :dob)
+      person_params[:countryId] = Country.find_by_iso2(params[:person][:representing]).id
```
gregorbg, line 301: You are very inconsistent about how you're reading params.
Especially handling it like `params.require(:person).permit(:wcaId)[:wcaId]` is horrible. Please find a consistent way to `params.require(:person)` **once**, and then read from that `require`d hash

WcaOnRails/app/controllers/admin_controller.rb
```diff
@@ -291,43 +291,54 @@ def edit_person
...
-      @person = Person.new
-      flash.now[:danger] = "No person has been chosen."
+      error_message = "No person has been chosen."
     end
-    render :edit_person
+    render json: {
+      success_message: success_message,
+      warning_message: warning_message,
+      error_message: error_message,
+    }
```
gregorbg, line 341: I am not a huge fan of handling different "success states" in one return statement.

If there is an error message, it should be returned with a separate "this was not okay" response, that also contains an error code.

If there is no error, there should be just **one** message, that is the success message, and perhaps an optional (nullable) `hint` or similar

WcaOnRails/app/controllers/api/v0/api_controller.rb
```diff
@@ -106,7 +106,7 @@ def search(*models)
       end
     end
 
-    render status: :ok, json: { result: result }
+    render status: :ok, json: { result: result }, option_name: current_user&.can_admin_results? ? :can_admin_results : nil
```
gregorbg, line 109: What does this `option_name` do?

WcaOnRails/app/models/person.rb
```diff
@@ -280,6 +280,11 @@ def serializable_hash(options = nil)
       id: self.wca_id,
     )
 
+    if options[:option_name] == :can_admin_results
+      json[:dob] = dob.to_s
+      json[:incorrect_wca_id_claim_count] = incorrect_wca_id_claim_count
+    end
```
gregorbg, line 286: This is not the correct way to handle serialization for specific controllers.
The controller calling the serialization method should add the *specific names* of the parameters it requires in the `only` or `methods` keys similar to `DEFAULT_SERIALIZE_OPTIONS` above.

gregorbg: See `api/v0/competitions_controller#show` for inspiration

WcaOnRails/app/views/admin/edit_person.html.erb
```diff
@@ -4,39 +4,9 @@
...
-        </button>
-
-        <button type="submit" name="method" value="destroy" class="btn btn-danger pull-right" data-confirm="Are you sure that you want to remove that competitor?">
-          <%= ui_icon("trash") %> Destroy
-        </button>
-      </div>
-    <% end %>
+    <%= react_component("Panel/Wrt/EditPerson", {
+      countryList: Country.all,
+      genderList: [:m, :f, :o],
```
gregorbg, line 9: You don't have to pass this gender list statically. It is available in Webpack-world as an import from `wca-data.js`.

gregorbg (changes requested):

WcaOnRails/app/controllers/admin_controller.rb
danieljames-dj: Okay, I cleaned it a bit. Hope it's fine now.

WcaOnRails/app/webpacker/components/Panel/Wrt/EditPerson.jsx
```diff
@@ -0,0 +1,183 @@
...
+
+const genderOptions = _.map(genders.byId, (gender) => ({
+  key: gender.id,
+  text: gender.name,
+  value: gender.id,
+}));
+
+const countryChangeWarning = `The change you made may have affected national and continental records, be sure to run <a href=${adminCheckRecordsUrl}>check_regional_record_markers</a>.`;
+
+function EditPerson({ countryList }) {
```
gregorbg, line 39: Same as for genders: You don't need to pass this as a prop, it's available in `wca-data`.

WcaOnRails/app/webpacker/components/Panel/Wrt/EditPerson.jsx
```diff
@@ -0,0 +1,183 @@
...
+  const [person, setPerson] = React.useState();
+  const [editedUserDetails, setEditedUserDetails] = React.useState();
+  const [originalUserDetails, setOriginalUserDetails] = React.useState();
+  const [incorrectClaimCount, setIncorrectClaimCount] = useState(0);
+  const { save, saving } = useSaveAction();
+  const [response, setResponse] = useState();
```
gregorbg, line 45: You're inconsistent with `React.useState` VS simple `useState`+import.

WcaOnRails/app/webpacker/components/Panel/Wrt/EditPerson.jsx
```diff
@@ -0,0 +1,183 @@
...
+  It will modify the Persons table accordingly and the Results table if the person's name is different.
+  This should be used to fix mistakes in the database.
+  </p>
+
+  <p>
+  Choose 'Update' if the person's name or country has been changed.
+  It will add a new entry in the Persons table and make it the current information for that person (subId=1)
+  but it will not modify the Results table so previous results keep the old name.
+  </p>
+`;
```
gregorbg, line 27: Why are you encoding this to be a string and then dangerously setting inner HTML below?
This should be directly in the HTML source returned by the React component (unless I'm overseeing a *very* good reason to have it be separate)

WcaOnRails/app/webpacker/components/Panel/Wrt/EditPerson.jsx
```diff
@@ -0,0 +1,183 @@
...
+
+const dateFormat = 'YYYY-MM-DD';
+
+const genderOptions = _.map(genders.byId, (gender) => ({
+  key: gender.id,
+  text: gender.name,
+  value: gender.id,
+}));
+
+const countryChangeWarning = `The change you made may have affected national and continental records, be sure to run <a href=${adminCheckRecordsUrl}>check_regional_record_markers</a>.`;
```
gregorbg, line 37: Same as for `const description` above. Why are you representing HTML source as a string?!

WcaOnRails/app/webpacker/components/Panel/Wrt/EditPerson.jsx
```diff
@@ -0,0 +1,183 @@
...
+      <div dangerouslySetInnerHTML={{ __html: description }} />
+      {response != null && (
+        <Message
+          success={response.success}
+          error={!response.success}
+          content={response.message}
+        >
+          <Message.Content>
+            {/* eslint-disable-next-line react/no-danger */}
+            <div dangerouslySetInnerHTML={{ __html: response.message }} />
```
gregorbg, line 105: Why not just `<Message.Content>{response.message}</Message.Content>`?

WcaOnRails/config/locales/en.yml
```diff
@@ -469,11 +469,6 @@ en:
           member: "Member"
           senior_member: "Senior Member"
           leader: "Leader"
-    person:
-      gender:
-        m: "Male"
-        f: "Female"
-        o: "Other"
```
gregorbg, line 476: Why are you dropping these without replacement?

WcaOnRails/app/controllers/admin_controller.rb
```diff
@@ -291,43 +291,57 @@ def edit_person
...
+      else
+        render status: :unprocessable_entity, json: { error: "Error while updating #{person.name}." }
+      end
+    when "destroy"
+      if person.results.any?
+        render status: :unprocessable_entity, json: { error: "#{person.name} has results, can't destroy them." }
+      elsif person.user.present?
+        render status: :unprocessable_entity, json: { error: "#{person.wca_id} is linked to a user, can't destroy them." }
+      else
+        name = person.name
```
gregorbg, line 335: There is already a `name` variable above.

WcaOnRails/app/controllers/admin_controller.rb
```diff
@@ -291,43 +291,57 @@ def edit_person
...
+    gender = person_params.require(:gender)
+    dob = person_params.require(:dob)
+    country_id = Country.find_by_iso2(representing).id
+
+    edit_params = {
+      name: name,
+      countryId: country_id,
+      gender: gender,
+      dob: dob,
+      incorrect_wca_id_claim_count: 0,
```
gregorbg, line 314: Why are you always assuming this to be zero?

gregorbg: And why is it not part of the request params?

WcaOnRails/app/controllers/admin_controller.rb
```diff
@@ -291,43 +291,57 @@ def edit_person
...
+    }
+    case params[:method]
```
gregorbg, line 316: I am not a big fan of having a switch-case for which action to take.
The design paradigm behind REST APIs should be to make a different route (possibly with a different HTTP verb) for different actions. This case is a little bit special because `fix` and `update` are two very similar actions, so I would potentially agree to push them into one common `PATCH` route.
But `"destroy"` should definitely be a separate HTTP `DELETE` route, and `"reset-claim-count"` should be a separate HTTP `PUT` with own, individual endpoints.

gregorbg (changes requested):

WcaOnRails/config/locales/en.yml
danieljames-dj: This is not used anywhere now because we changed to the gender picker from wca-data.

WcaOnRails/app/controllers/admin_controller.rb
danieljames-dj: So, the flow is like, whenever we update some person's data, we have to reset the incorrect WCA ID claim count as well. This is because whenever WRT receives a request to reset claim ID, we usually get a edit request as well connected with it because the claim ID count exceeded because they tried many times and then figured out that problem was that the data in database is wrong. So, reseting to 0 while editing the person will be a good thing for WRT.

danieljames-dj (commented):
I've addressed couple of comments which doesn't (probably) require code change from my side. I'll address the remaining along with a commit later.

WcaOnRails/app/webpacker/components/Panel/Wrt/EditPerson.jsx
danieljames-dj: Okay, sorry that @coder13 and @kr-matthews has already mentioned this in https://github.com/thewca/worldcubeassociation.org/pull/8406#discussion_r1398233362, but I forgot to address it further.

danieljames-dj: I've also changed the text of description to avoid duplication of explanation in two places.

WcaOnRails/app/webpacker/components/Panel/Wrt/EditPerson.jsx
danieljames-dj: This is a different case, where this string will be assigned to warning message if there is a change in country. There are other warning messages, hence added this as a string.

WcaOnRails/app/webpacker/components/Panel/Wrt/EditPerson.jsx
danieljames-dj: Sorry for inconsistencies. I thought `React.useState` was more standard. I've changed everything in this file to `useState`.

WcaOnRails/app/webpacker/components/Panel/Wrt/EditPerson.jsx
danieljames-dj: Because the message can contain hyperlink (like the check records link).

danieljames-dj (commented):
I've moved the functions inside admin_controller to separate controller because admin_controller is already having plenty of functions and splitting the `update_person` will make it more bigger.

I've moved it under `api/v0` under a new `wrt` scope. I created a new scope because there are many more WRT-only APIs that are going to come in future.

WcaOnRails/app/webpacker/components/Panel/Wrt/EditPerson.jsx
gregorbg: Then make the message return some error code and the raw link URL separately, and load the i18n'd version here (which will also allow you to inject the link by using `I18nTranslate` React component).

I have very strong reservations about allowing more `dangerouslySetInnerHTML` in our codebase.

WcaOnRails/config/routes.rb
```diff
@@ -346,6 +345,12 @@
         get '/group-type/:group_type' => 'roles#index_for_group_type', as: :index_for_group_type
       end
       resources :user_groups, only: [:index, :create, :update]
+      namespace :wrt do
+        resources :persons, only: [:update, :destroy]
+        scope 'persons' do
```
gregorbg, line 350: You can directly enter the namespace by introducing a `do` block in the `resources` call.

gregorbg (changes requested):

danieljames-dj (commented):
Just realized I forgot to reply and my replies were in draft.

gregorbg (approved):

gregorbg (approved):

PR #8407, opened 2023-10-03
Add Get Permissions and Me route

FinnIckler:
For wca-registrations we will need some new user related routes. This PR adds 2 public and one internal route.

Public Routes:
- Returns all information about the current user so we can display it in the frontend
- Returns the current users permissions so we can display the correct elements in the frontend

Private Routes:
- Returns the permissions of any user so we can validate if a user is allowed to perform an action on a microservice

Private Routes are secured using [Vault's Identity tokens ](https://developer.hashicorp.com/vault/api-docs/secret/identity/tokens#generate-a-signed-id-token), this way only services that are connected to vault and have the correct permissions to generate tokens can use the internal routes.

See an example on how you use it on the other side [here](https://github.com/thewca/wca-registration/pull/247). 

I have not added these routes to the routes.rb yet, as where we will mount them still needs to be discussed. I have cautiously added a /api/internal/v1 route for now.

Should I add tests for this?

### DIFF
--- a/WcaOnRails/app/controllers/api/internal/v1/api_controller.rb
+++ b/WcaOnRails/app/controllers/api/internal/v1/api_controller.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+class Api::Internal::V1::ApiController < ApplicationController
+  prepend_before_action :validate_token
+
+  def validate_token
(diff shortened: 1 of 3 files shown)


FinnIckler:
oops this somehow is opened against my fork. redoing...

PR #8408, opened 2023-10-03
Added Get Permissions and Me route

FinnIckler:
For wca-registrations we will need some new user related routes. This PR adds 2 public and one internal route.

Public Routes:
- Returns all information about the current user so we can display it in the frontend
- Returns the current users permissions so we can display the correct elements in the frontend

Private Routes:
- Returns the permissions of any user so we can validate if a user is allowed to perform an action on a microservice

Private Routes are secured using [Vault's Identity tokens ](https://developer.hashicorp.com/vault/api-docs/secret/identity/tokens#generate-a-signed-id-token), this way only services that are connected to vault and have the correct permissions to generate tokens can use the internal routes.

See an example on how you use it on the other side [here](https://github.com/thewca/wca-registration/pull/247). 

I have not added these routes to the routes.rb yet, as where we will mount them still needs to be discussed. I have cautiously added a /api/internal/v1 route for now.

Should I add tests for this?


WcaOnRails/app/controllers/api/internal/v1/api_controller.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class Api::Internal::V1::ApiController < ApplicationController
+  prepend_before_action :validate_token
+
+  def validate_token
+    service_token = request.headers["X-WCA-Service-Token"]
```
gregorbg, line 7: Consider moving the header name into a constant somewhere

WcaOnRails/app/controllers/api/internal/v1/api_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+    unless service_token.present?
+      return render json: { error: "Missing Authentication" }, status: :forbidden
+    end
```
gregorbg, line 10: `return ... unless ...` fits on one line in Ruby :wink:

WcaOnRails/app/controllers/api/internal/v1/api_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+    response = Vault.with_retries(Vault::HTTPConnectionError) do
+      Vault.logical.write("identity/oidc/introspect", data: { token: service_token })
+    end
+    unless response.data.present? && response.data[:active]
+      render json: { error: "Authentication Expired or Token Invalid" }, status: :forbidden
+    end
```
gregorbg, line 16: See above about shortening this to one line

WcaOnRails/app/controllers/api/internal/v1/permissions_controller.rb
```diff
@@ -0,0 +1,18 @@
+class Api::Internal::V1::PermissionsController < Api::Internal::V1::ApiController
+  def index
+    user_id = params.require(:user_id)
+    user = User.find(user_id)
+    render json: {
+      can_attend_competitions: {
+        scope: user.cannot_register_for_competition_reasons.empty? ? "*" : [],
+        until: user.banned? ? user.current_team_members.select(:team == Team.banned).first.end_date : nil
```
gregorbg, line 8: First, you probably want `current_team_members.select(team: Team.banned)`. Also, `select` on Rails ORM collections are just an alias to `where`, which in turn is closer to the actual SQL syntax.

Second, I feel like there _should_ be a shorthand for the "until when is this competitor banned?" information... But of course I don't know all 1000+ lines of the User class by heart so I might be wrong.

WcaOnRails/app/controllers/api/v0/users_controller.rb
```diff
@@ -0,0 +1,32 @@
+class Api::V0::PersonsController < Api::V0::ApiController
+  def me
```
gregorbg, line 2: Don't we already have a `/me` route?!

gregorbg (changes requested):

gregorbg:
> Should I add tests for this?

Of course! :stuck_out_tongue:

WcaOnRails/app/controllers/api/internal/v1/api_controller.rb
FinnIckler: I guess it should be the microservice_helper that I made in the other PR?

WcaOnRails/app/controllers/api/internal/v1/permissions_controller.rb
FinnIckler: I took this from the banned_at_date? function in user.rb
```
def banned_at_date?(date)
    if banned?
      ban_end = current_team_members.select(:team == Team.banned).first.end_date
      !ban_end.present? || date < ban_end
    else
      false
    end
  end
```

FinnIckler: I could make this a new function in the model though if we use it twice now

WcaOnRails/app/controllers/api/v0/users_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+            until: user.banned? ? user.current_team_members.select(:team == Team.banned).first.end_date : nil
+          },
+          can_organize_competitions: {
+            scope: current_user.can_create_competitions? ? "*" : [],
+          },
+          can_administer_competitions: {
+            scope: current_user.can_admin_competitions? ? "*" : (current_user.delegated_competitions + current_user.organized_competitions).pluck("id"),
+          },
+        }
+      end
```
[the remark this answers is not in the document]
FinnIckler: moved to the user model

WcaOnRails/app/controllers/api/internal/v1/permissions_controller.rb
FinnIckler: made a ban_end function in the model

WcaOnRails/app/controllers/api/internal/v1/api_controller.rb
gregorbg: Sounds like a good place for it!

WcaOnRails/app/controllers/api/v0/users_controller.rb
gregorbg: Yup, if we can recycle the same piece of code in two different places then go for it!

WcaOnRails/app/models/user.rb
```diff
@@ -632,15 +632,33 @@ def banned?
     current_teams.include?(Team.banned)
   end
 
+  def ban_end
+    current_team_members.select(:team == Team.banned).first.end_date
```
gregorbg, line 636: I understand that you copied it from another place, but then again the code that you copied was just terrible. Perhaps it's not your fault, but touching this part of the code gives us a chance to fix it.

- Using `orm_association.select(some_predicate)` (like is the case here) loads all ORM associations from SQL into memory and _then_ iterates them using the predicate to evaluate which items yield true.
- Using `orm_association.where(field_name: field_value)` is a bit more restrictive because it only allows you to write SQL-esque queries (and not a generic expression that might evaluate to a boolean) but it actually also renders the Ruby code invocation as SQL query and is therefor _much_ more efficient.

gregorbg: Something along the lines of `current_team_members.where(team: Team.banned).first` should do.

gregorbg: I think that ActiveRecord even has an ORM interface for selecting the first entry using SQL `LIMIT 1` directly via `current_team_members.first(team: Team.banned)`, but I am not 100% sure though

WcaOnRails/app/models/user.rb
```diff
@@ -632,15 +632,33 @@ def banned?
...
+    {
+      can_attend_competitions: {
+        scope: cannot_register_for_competition_reasons.empty? ? "*" : [],
+        until: banned? ? ban_end : nil,
+      },
+      can_organize_competitions: {
+        scope: can_create_competitions? ? "*" : [],
+      },
+      can_administer_competitions: {
+        scope: can_admin_competitions? ? "*" : (delegated_competitions + organized_competitions).pluck("id"),
```
gregorbg, line 657: `pluck` is an ORM call, where we generally try to use symbols: `.pluck(:id)`

WcaOnRails/app/models/user.rb
FinnIckler: I mean I did write the original code too so I guess it is kinda my fault :P

WcaOnRails/app/controllers/api/v0/users_controller.rb
FinnIckler: That me action currently has `before_action :doorkeeper_authorize!, only: [:me]` though, so it might overcomplicate things if I merge the two?

WcaOnRails/app/models/user.rb
FinnIckler: https://apidock.com/rails/ActiveRecord/FinderMethods/first does first does only take the limit as an argument

FinnIckler:
I added a bunch of tests, this is the first time adding tests for me, so there are still things I need to figure out like how do add PRs to a user so I can correctly test the me route and how to set banned date and delegated/organized competitions

FinnIckler:
The tests shouldn't pass I think? I am pretty sure a lot of assumptions I did do not work like that...
Do I need to tell rspec about the new test file?

WcaOnRails/app/controllers/api/internal/v1/api_controller.rb
FinnIckler: We still haven't decided if we should put it in lib or helpers though! Considering it contains routes which might be used in views or helpers...

WcaOnRails/app/controllers/api/v0/users_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  def show_user_by_wca_id
+    user = User.find_by_wca_id(params[:wca_id])
+    show_user(user)
+  end
+
+  def permissions
+    if current_user
+      if stale?(current_user)
```
gregorbg, line 27: Not sure if this `stale` call catches all relevant information for permissions... For example a `team_member` association to the "banned" team might change, but the `User` model itself didn't. Then I think this will return outdated responses.

WcaOnRails/app/models/user.rb
```diff
@@ -632,15 +632,33 @@ def banned?
...
     else
       false
     end
   end
 
+  def permissions
+    {
+      can_attend_competitions: {
+        scope: cannot_register_for_competition_reasons.empty? ? "*" : [],
+        until: banned? ? ban_end : nil,
```
gregorbg, line 651: Wouldn't it make more sense to straight-out discard this key if the user is banned?

gregorbg: Also, what is the overlap between `cannot_register_for_competition_reasons` and `banned?`? Are we confident that the first returns something non-empty exactly if (genau dann, wenn) the other is not empty?

Even if we are confident about that, why are we using both checks disjointly?

gregorbg (commented):
You do not need to tell RSpec about new files. As long as they are under the `specs` folder and ending in `_spec.rb` it will pick the file up automatically (we have no central directory of "which tests to run". Every CI run pretty much just does a glob.)

WcaOnRails/app/models/user.rb
FinnIckler: `cannot_register_for_competition_reasons` is not empty if a user is banned OR if they have not set up their DOB/name/gender/country. I agree that setting until only if banned makes sense, can I do that directly in the has somehow?

gregorbg:
Ah! Your tests don't get run because you (mistakenly) called the file `foo.spec.rb`. The convention dictates `foo_spec.rb`. The `spec` is not a file suffix but rather a part of the filename itself.

FinnIckler:
any idea why I had to remove the brackets for `sign_in FactoryBot.create :user, :wrt_member` while everywhere else in our test code it works using `sign_in { FactoryBot.create :user, :wrt_member }`?

gregorbg:
> any idea why I had to remove the brackets for `sign_in FactoryBot.create :user, :wrt_member` while everywhere else in our test code it works using `sign_in { FactoryBot.create :user, :wrt_member }`?

I did some digging and actually the perception "everywhere else in our test code" is not quite true. There are _plenty_ of cases where sign_in is used without braces.

Now please don't ask me how or why, but in the file `spec/support/sign_in_macros.rb` Jeremy specifically defined that only **controller** and **request** tests need the curly braces. Oddly enough, all they do is evaluate the braces and pass them on to the brace-less `sign_in` function of our test driver -- my guess would be it's syntactic sugar for lazy evaluation or something. But it's undocumented so who knows :shrug:

FinnIckler:
While the tests pass now (yay!). This should not be merged until tested (best would be somehow with the other PRs) on staging against wca-registrations

gregorbg:
> This should not be merged until tested

If you anticipate that testing is not scheduled to happen for another few days, please set this PR to Draft.

WcaOnRails/config/routes.rb
```diff
@@ -294,8 +294,11 @@
       get '/search/users' => 'api#users_search'
       get '/search/regulations' => 'api#regulations_search'
       get '/search/incidents' => 'api#incidents_search'
-      get '/users/:id' => 'api#show_user_by_id', constraints: { id: /\d+/ }
-      get '/users/:wca_id' => 'api#show_user_by_wca_id'
+      get '/users' => 'users#show_users_by_id'
+      get '/users/me' => 'users#show_me'
```
coder13, line 298: How is this route different from `api/v0/me`?

WcaOnRails/app/controllers/api/internal/v1/api_controller.rb
coder13, line 1: In this situation, it seems like the file path doesn't match the url path? I feel it's in our (and external devs) best interest to have the file path be as similar to the url path as possible.

WcaOnRails/config/routes.rb
FinnIckler: api/v0/me is just available for oauth users

WcaOnRails/app/controllers/api/internal/v1/api_controller.rb
FinnIckler: not sure what you mean, the path is also /api/internal/v1/{route} I just didn't add it to the routes.rb yet

coder13: > I just didn't add it to the routes.rb yet

That makes sense.

WcaOnRails/app/controllers/api/internal/v1/api_controller.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+class Api::Internal::V1::ApiController < ApplicationController
+  prepend_before_action :validate_token
+
+  def validate_token
+    service_token = request.headers[Microservices::Auth::MICROSERVICE_AUTH_HEADER]
+    unless service_token.present?
+      return render json: { error: "Missing Authentication" }, status: :forbidden
+    end
```
gregorbg, line 10:
```suggestion
    return render json: { error: "Missing Authentication" }, status: :forbidden unless service_token.present?
```

WcaOnRails/app/controllers/api/internal/v1/api_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+    # The Vault CLI can't parse the response from identity/oidc/introspect so
+    # we need to request it instead see https://github.com/hashicorp/vault/issues/9080
+
+    vault_token_data = Vault.auth_token.lookup_self.data
+    # Renew our token if it has expired or is close to expiring
+    if vault_token_data[:ttl] < 300
+      Vault.auth_token.renew_self
+    end
```
gregorbg, line 18:
```suggestion
    Vault.auth_token.renew_self if vault_token_data[:ttl] < 300
```

WcaOnRails/app/controllers/api/internal/v1/api_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+    response = client.post do |req|
+      req.url '/v1/identity/oidc/introspect'
+      req.headers['X-Vault-Token'] = vault_token_data[:id]
+      req.body = { token: service_token }.to_json
+    end
+    if response.success?
+      result = JSON.parse(response.body)
+      unless result["active"]
+        render json: { error: "Authentication Expired or Token Invalid" }, status: :forbidden
+      end
```
gregorbg, line 32:
```suggestion
      render json: { error: "Authentication Expired or Token Invalid" }, status: :forbidden unless result["active"]
```

gregorbg (changes requested):

FinnIckler:
test failures seem not related

WcaOnRails/app/controllers/api/v0/users_controller.rb
FinnIckler: I tested it out in on staging and it correctly showed it after a refresh (actually two refreshs?)

gregorbg (approved):

PR #8409, opened 2023-10-04
Add more info to the competition API

FinnIckler:
To Render the WCA-Registrations Frontend we need some more info as part of the competition API. I also added a `stale?` check because this route will be used much more heavily on the Frontend and we don't want to recompute it if not needed.

I chose to keep it under it's current route as this change is compatible with the current response, but I can add a new path if it makes more sense.

### DIFF
--- a/WcaOnRails/app/controllers/api/v0/competitions_controller.rb
+++ b/WcaOnRails/app/controllers/api/v0/competitions_controller.rb
@@ -18,7 +18,8 @@ class Api::V0::CompetitionsController < Api::V0::ApiController
   end
 
   COMPETITION_INFO_SERIALIZE_OPTIONS = {
-    only: %w[extra_registration_requirements enable_donations refund_policy_limit_date event_change_deadline_date waiting_list_deadline_date on_the_spot_registration on_the_spot_entry_fee_lowest_denomination qualification_results event_restrictions base_entry_fee_lowest_denomination currency_code allow_registration_edits allow_registration_self_delete_after_acceptance allow_registration_without_qualification refund_policy_percent use_wca_registration guests_per_registration_limit venue contact force_comment_in_registration use_wca_registration external_registration_page guests_entry_fee_lowest_denomination guest_entry_status information],
+    only: %w[extra_registration_requirements enable_donations refund_policy_limit_date event_change_deadline_date waiting_list_deadline_date on_the_spot_registration on_the_spot_entry_fee_lowest_denomination qualification_results
+             event_restrictions base_entry_fee_lowest_denomination currency_code allow_registration_edits allow_registration_self_delete_after_acceptance allow_registration_without_qualification refund_policy_percent use_wca_registration guests_per_registration_limit venue contact force_comment_in_registration use_wca_registration external_registration_page guests_entry_fee_lowest_denomination guest_entry_status information],
     methods: %w[registration_opened? main_event_id number_of_bookmarks using_stripe_payments? uses_qualification? uses_cutoff? events_with_rounds schedule_wcif],
     include: %w[tabs],
   }.freeze
--- a/WcaOnRails/app/controllers/api/v0/competitions_controller.rb
+++ b/WcaOnRails/app/controllers/api/v0/competitions_controller.rb
@@ -17,9 +17,16 @@ class Api::V0::CompetitionsController < Api::V0::ApiController
     paginate json: competitions
   end
 
+  COMPETITION_INFO_SERIALIZE_OPTIONS = {
+    only: %w[id name website start_date registration_open registration_close announced_at cancelled_at end_date competitor_limit extra_registration_requirements enable_donations refund_policy_limit_date event_change_deadline_date waiting_list_deadline_date on_the_spot_registration on_the_spot_entry_fee_lowest_denomination qualification_results event_restrictions base_entry_fee_lowest_denomination currency_code allow_registration_edits allow_registration_self_delete_after_acceptance allow_registration_without_qualification refund_policy_percent use_wca_registration guests_per_registration_limit venue contact force_comment_in_registration use_wca_registration external_registration_page guests_entry_fee_lowest_denomination guest_entry_status information],
+    methods: %w[url website short_name city venue_address venue_details latitude_degrees longitude_degrees country_iso2 event_ids registration_opened? main_event_id number_of_bookmarks using_stripe_payments? uses_qualification? uses_cutoff? events_with_rounds schedule_wcif],
+    include: %w[delegates organizers tabs],
+  }.freeze
+
   def show
     competition = competition_from_params
-    render json: competition
+    json = competition.serializable_hash(COMPETITION_INFO_SERIALIZE_OPTIONS)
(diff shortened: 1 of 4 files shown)


WcaOnRails/app/models/competition.rb
```diff
@@ -391,6 +396,16 @@ def main_event_id=(event_id)
     super(event_id.blank? ? nil : event_id)
   end
 
+  def events_with_rounds
+    includes_associations = [
+      { rounds: [:competition_event] },
+    ]
+    competition_events
+      .includes(includes_associations)
```
gregorbg, line 404: If you only need the `includes_associations` variable once, you might as well pass it directly:

```ruby
competition_events
  .includes(rounds: [:competition_event])
```

WcaOnRails/app/controllers/api/v0/competitions_controller.rb
```diff
@@ -17,9 +17,18 @@ def index
     paginate json: competitions
   end
 
+  COMPETITION_INFO_SERIALIZE_OPTIONS = {
+    only: %w[id name website start_date registration_open registration_close announced_at cancelled_at end_date competitor_limit extra_registration_requirements enable_donations refund_policy_limit_date event_change_deadline_date waiting_list_deadline_date on_the_spot_registration on_the_spot_entry_fee_lowest_denomination qualification_results event_restrictions base_entry_fee_lowest_denomination currency_code allow_registration_edits allow_registration_self_delete_after_acceptance allow_registration_without_qualification refund_policy_percent use_wca_registration guests_per_registration_limit venue contact force_comment_in_registration use_wca_registration external_registration_page guests_entry_fee_lowest_denomination guest_entry_status information],
+    methods: %w[url website short_name city venue_address venue_details latitude_degrees longitude_degrees country_iso2 event_ids registration_opened? main_event_id number_of_bookmarks using_stripe_payments? uses_qualification? uses_cutoff? events_with_rounds schedule_wcif],
+    include: %w[delegates organizers tabs],
+  }.freeze
```
gregorbg, line 24: This slightly reminds me of https://github.com/thewca/wcif/pull/17.
What information is there in this big array that you need, but that you cannot currently get through WCIF?

gregorbg:
> I chose to keep it under it's current route as this change is compatible with the current response, but I can add a new path if it makes more sense.

No, as long as it's backwards-compatible a new route is not necessary

FinnIckler:
```
only: ["extra_registration_requirements", "enable_donations", "refund_policy_limit_date", "event_change_deadline_date", "waiting_list_deadline_date", "on_the_spot_registration", "on_the_spot_entry_fee_lowest_denomination", "qualification_results", "event_restrictions", "base_entry_fee_lowest_denomination", "currency_code", "allow_registration_edits", "allow_registration_self_delete_after_acceptance", "allow_registration_without_qualification", "refund_policy_percent", "use_wca_registration", "guests_per_registration_limit", "venue", "contact", "force_comment_in_registration", "use_wca_registration", "external_registration_page", "guests_entry_fee_lowest_denomination", "guest_entry_status", "information"]
methods: ["registration_opened?", "main_event_id", "number_of_bookmarks", "using_stripe_payments?", "uses_qualification?", "uses_cutoff?", "events_with_rounds", "schedule_wcif"]
include: ["tabs"]
```
These are all the fields I added, most of them are not available in wcif

WcaOnRails/app/controllers/api/v0/competitions_controller.rb
FinnIckler: I changed it to only have the new fields that I need, but as mentioned above, most of them are not in the wcif

WcaOnRails/app/models/competition.rb
FinnIckler: done

WcaOnRails/app/models/competition.rb
```diff
@@ -391,6 +395,13 @@ def main_event_id=(event_id)
     super(event_id.blank? ? nil : event_id)
   end
 
+  def events_with_rounds
+    competition_events
+      .includes({ rounds: [:competition_event] })
```
gregorbg, line 400: Nitpick: You don't even need the outer `{}` braces. The `.includes` invocation supports kwargs.

gregorbg:
> These are all the fields I added, most of them are not available in wcif

The `schedule_wcif` most definitely is available in WCIF :innocent:
Actually, that one is the only item that I have certain reservations against: It is huge and it is redundant (regarding the fact that it is available in another endpoint).

Is there any chance your new frontend can fetch the WCIF and the "internal competition format" separately and combine the two as needed?

FinnIckler:
That is no problem, that would also mean I don't need events_with_rounds.

FinnIckler:
I just tested it in the ruby console and DEFAULT_SERIALIZE_OPTIONS.merge(COMPETITION_INFO_SERIALIZE_OPTIONS) is just COMPETITION_INFO_SERIALIZE_OPTIONS, so the code in serializable_hash actually never worked

FinnIckler:
looking at the documentation for merge it makes sense:

> If no block is specified, the value for entries with duplicate [keys](https://apidock.com/ruby/Hash/keys) will be that of other_hash. Otherwise the value for [each](https://apidock.com/ruby/Hash/each) duplicate [key](https://apidock.com/ruby/Hash/key) is determined by calling the block with the key, its value in hsh and its value in other_hash.
So maybe adding a block to the merge function instead will be the day

I added a block to see if that is the way

gregorbg:
I solved it using the https://github.com/danielsdeleo/deep_merge third-party library. For context:
- `merge(a, b)` simply takes the value from b.
- `deep_merge(a, b)` tries to recursively merge, but crucially _only works for hashes_. Merging arrays inside hashes upon colliding keys is not supported by default :(
    - Providing a block works because it "teaches" Rails how to merge arrays. But it may break when we ever pass nested _hashes_ in the `include` part.
    - The library dynamically determines what to do. Yay!

FinnIckler:
and tests pass! We should be able to merge it then, which is going to enable me to change the code on wca-registrations accordingly

WcaOnRails/app/models/competition.rb
```diff
@@ -379,6 +379,10 @@ def registration_full?
     competitor_limit_enabled? && registrations.accepted_and_paid_pending_count >= competitor_limit
   end
 
+  def number_of_bookmarks
+    bookmarked_users.length
```
gregorbg, line 383: Nitpick: This should be `bookmarked_users.count` instead of `.length`. The `bookmarked_users` field is a Rails ORM association mapping (more or less) directly to SQL, but `length` is a general **Ruby** method. Doing it this way loads all users via `SELECT *` into a Ruby array and then counts the length of the array. Using `.count` does a SQL `SELECT COUNT(*)` directly.

gregorbg (changes requested):

WcaOnRails/app/models/competition.rb
FinnIckler: done!

gregorbg (approved):

PR #8410, opened 2023-10-04
Implement Payment Routes 

FinnIckler:
For WCA-Registration we need to be able to call `finish_payment` and `refund_payment` from the Frontend, so I created a separate  controller for it.
There are two more changes needed for WCA-Registration to interact with Payments:
- When the stripe webhook is called the monolith needs to update wca-registration which it does via an internal route. I created a job for this. Did I do it correctly?
- After a user registers, wca-registrations needs to ask the monolith to create a payment intent, this is again done via an internal route. 

Again, I haven't added anything to routes yet


WcaOnRails/Gemfile
```diff
@@ -45,6 +45,7 @@ gem 'octokit'
 gem 'stripe'
 gem 'oauth2'
 gem 'openssl'
+gem 'httparty'
```
gregorbg, line 48: I'm quite sure we should already have something for POST requests in our zoo of dependencies.

Also, why is `Gemfile.lock` not part of the diff?

WcaOnRails/env_config.rb
```diff
@@ -22,6 +22,7 @@
     mandatory :VAULT_APPLICATION, :string
     mandatory :VAULT_AWS_REGION, :string
     mandatory :INSTANCE_ROLE, :string
+    mandatory :WCA_REGISTRATION_URL, :string
```
gregorbg, line 25: Where is the value for this variable being populated? --> Chef
How is it being defined in non-production environments? --> Second optional declaration below (and possible changes to our development env files)

WcaOnRails/app/jobs/update_wca_registration.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+#
+require_relative '../helpers/microservices_helper'
+class UpdateWCARegistration < ApplicationJob
+  def perform(status)
+    token = get_wca_token
+    response = httparty.post(update_payment_status_path, headers: { 'X-WCA-Service-Token' => token, "Content-Type" => "application/json" }, body: { payment_status: status }.to_json)
+    unless response.ok?
+      throw Error "Updating wca-registration failed with error #{response.status} #{response.body}"
```
gregorbg, line 9: Shouldn't it be `raise` in Ruby?!

WcaOnRails/app/jobs/update_wca_registration.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+#
+require_relative '../helpers/microservices_helper'
+class UpdateWCARegistration < ApplicationJob
+  def perform(status)
+    token = get_wca_token
```
gregorbg, line 6: If you only ever use this once, you can just as well inline it

gregorbg (changes requested):

WcaOnRails/app/controllers/api/internal/v1/payment_controller.rb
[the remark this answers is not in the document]
FinnIckler: yes, because that old code will still be used if people pay via the old registration service! I am not routing to the new code as mentioned because I don't know which route would 'make sense'

WcaOnRails/app/models/stripe_payment_intent.rb
```diff
@@ -83,5 +83,6 @@ def update_status_and_charges(api_intent, action_source, source_datetime = DateT
         )
       end
     end
+    UpdateWCARegistration.perform(api_intent.status)
```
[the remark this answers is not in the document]
FinnIckler: That's what I started with, but I wanted something that retries and keeps logs of failed attempts of updating the registration service, I guess we could also let the stripe update fail so stripe will try again instead?

WcaOnRails/app/jobs/update_wca_registration.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+#
+require_relative '../helpers/microservices_helper'
```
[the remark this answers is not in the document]
FinnIckler: mh... because this code contains routes, it might  be used both by views and controllers, what to do then?

WcaOnRails/app/controllers/api/internal/v1/api_controller.rb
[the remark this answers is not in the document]
FinnIckler: it is copy pasta from there!

WcaOnRails/Gemfile
FinnIckler: because I didn't actually install it locally :P. I will check the other dependencies, but I don't think we do

WcaOnRails/app/helpers/microservices_helper.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  def get_wca_token
+    # Uses Vault ID Tokens: see https://developer.hashicorp.com/vault/docs/secrets/identity/identity-token
+    Vault.with_retries(Vault::HTTPConnectionError) do
+      data = Vault.logical.read("identity/oidc/token/#{EnvConfig.VAULT_APPLICATION}")
+      if data.present?
+        data.data[:data][:token]
+      else
+        raise Error "Can't get a Vault token"
+      end
```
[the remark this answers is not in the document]
FinnIckler: done

WcaOnRails/app/jobs/update_wca_registration.rb
FinnIckler: [throw](https://apidock.com/ruby/v2_5_5/Kernel/throw) actually works too, but I think just for legacy reasons which why my ide didn't complain

WcaOnRails/app/jobs/update_wca_registration.rb
FinnIckler: we will definitely use get_wca_token again for communicating with user microservices!

WcaOnRails/app/models/stripe_payment_intent.rb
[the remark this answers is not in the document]
FinnIckler: I will still move it outside of the model

WcaOnRails/app/controllers/payment_controller.rb
```diff
@@ -0,0 +1,129 @@
...
+  def payment_refund
+    competition_id, user_id = params["attendee_id"].split("-")
+    competition = Competition.find(competition_id)
+
+    unless competition.using_stripe_payments?
+      flash[:danger] = "You cannot emit refund for this competition anymore. Please use your Stripe dashboard to do so."
+      return redirect_to edit_registration_path(competition_id, user_id)
+    end
+
+    payment = RegistrationPayment.find(params[:payment_id])
```
FinnIckler, line 76: Any idea how we can find the payment without using RegistrationPayment?

WcaOnRails/Gemfile
gregorbg: If you want this gem, you definitely have to install it and check in the changes to `Gemfile.lock`.

gregorbg: Also, there is [`faraday`](https://github.com/lostisland/faraday) already in our Gemfile, which I think does what you want.

WcaOnRails/app/jobs/update_wca_registration.rb
gregorbg: My bad, the comment sounds ambiguous. What I meant is: If you only use **this variable** once, you might as well inline it.

WcaOnRails/app/jobs/update_wca_registration.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+require_relative '../helpers/microservices_helper'
+class UpdateWCARegistration < ApplicationJob
+  def perform(status)
+    token = get_wca_token
+    response = httparty.post(update_payment_status_path, headers: { 'X-WCA-Service-Token' => token, "Content-Type" => "application/json" }, body: { payment_status: status }.to_json)
```
gregorbg, line 7:
```suggestion
    response = httparty.post(update_payment_status_path, headers: { 'X-WCA-Service-Token' => get_wca_token, "Content-Type" => "application/json" }, body: { payment_status: status }.to_json)
```

WcaOnRails/app/jobs/update_wca_registration.rb
```diff
@@ -0,0 +1,12 @@
...
+    unless response.ok?
+      raise Error "Updating wca-registration failed with error #{response.status} #{response.body}"
+    end
```
gregorbg, line 10: `raise ... unless ...`

gregorbg (changes requested):

WcaOnRails/app/controllers/api/internal/v1/payment_controller.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+class Api::Internal::V1::PaymentController < Api::Internal::V1::ApiController
+  def init
+    competition_id, user_id = params["attendee_id"].split("-")
```
gregorbg, line 5: This kind of parsing logic could move into a dedicated `Attendee` monolith model.

WcaOnRails/app/controllers/api/internal/v1/payment_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    amount = params["amount"].to_i
```
gregorbg, line 7: Passing around `amount`s is tricky because of https://stripe.com/docs/currencies#special-cases that sometimes don't comply with the ISO standards that our RubyMoney gem is trying hard to use.

We have a dedicated `StripeTransaction.amount_to_stripe` and `.amount_to_ruby` helper duo which works as long as we can safely assume that all Ruby amounts come from the RubyMoney gem. Now that they come from the external world, this assumption might break.

Not necessarily requiring a specific change here, just trying to point out to be _very_ careful and explicitly mark amounts as "Ruby amount" or "Stripe amount" at least through a comment.

WcaOnRails/app/controllers/api/internal/v1/payment_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+    user = User.find(user_id)
+    account_id = competition.connected_stripe_account_id
+
+    registration_metadata = {
+      competition: competition.name,
+      registration_url: edit_registration_path(competition_id, user_id),
+    }
+
+    currency_iso = params["currency_code"]
+    stripe_amount = StripeTransaction.amount_to_stripe(amount, currency_iso)
```
gregorbg, line 19: see comment above.

WcaOnRails/app/controllers/api/internal/v1/payment_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+      stripe_account: account_id,
+    )
+
+    # Log the payment attempt. We register the payment intent ID to find it later after checkout completed.
+    stripe_transaction = StripeTransaction.create_from_api(intent, payment_intent_args, account_id)
+
+    # memoize the payment intent in our DB because payments are handled asynchronously
+    # so we need to be able to retrieve this later at any time, even when our server crashes in the meantime…
+    StripePaymentIntent.create!(
+      holder: { type: "attendee", value: params["attendee_id"]},
```
gregorbg, line 51: This won't work. Rails polymorphic foreign keys expect an `ActiveRecord` model to be passed as `holder`.

WcaOnRails/app/controllers/payment_controller.rb
```diff
@@ -0,0 +1,103 @@
...
+
+    unless stripe_intent.present?
+      return redirect_to competition_register_path(competition, "not_found")
+    end
+
+    stored_intent.update_status_and_charges(stripe_intent, current_user) do |_|
+      begin
+        update_registration_payment(stripe_intent.id, stripe_intent.status)
+      rescue Error
+        return redirect_to competition_register_path(competition_id, "Couldn't reach Registration Service")
```
gregorbg, line 35: You use stringified error messages like `"not_found"` above but then suddenly a humanized English phrase here. Please be consistent in the error handling.

WcaOnRails/app/controllers/payment_controller.rb
```diff
@@ -0,0 +1,103 @@
...
+    stripe_intent = stored_intent.retrieve_intent
+
+    unless stripe_intent.present?
+      return redirect_to competition_register_path(competition, "not_found")
+    end
+
+    stored_intent.update_status_and_charges(stripe_intent, current_user) do |_|
+      begin
+        update_registration_payment(stripe_intent.id, stripe_intent.status)
+      rescue Error
```
gregorbg, line 34: rescuing each and any error is bad practice.

WcaOnRails/app/controllers/payment_controller.rb
gregorbg: (discussed on a phone call this morning)

WcaOnRails/app/controllers/registrations_controller.rb
```diff
@@ -534,7 +534,14 @@ def stripe_webhook
     else
       logger.info "Unhandled Stripe event type: #{event.type}"
     end
-
+    if stored_intent.holder.type == "attendee"
```
gregorbg, line 537: This won't work. Calling `holder` immediately prompts ActiveRecord to try and parse the columns `holder_id` and `holder_type` into a model called `attendee` which doesn't exist. An error would be thrown even before checking `type` is possible.

gregorbg (changes requested):
The more I think about it, the more I get the feeling that we should introduce a rudimentary `Attendee` model in the monolith codebase, whose only purpose is to store information about the payment requests made by the external microservice.

You don't even have to call it `Attendee` because to be fair, it won't store any actual attendee information. Maybe calling it `AttendeePaymentRequest` or something makes more sense, and then you can store when you recieved the request as well as some other details that may be important (at least for logging). That way, you will also have an ActiveRecord model to attach to `StripePaymentIntent` as `holder`.

The general idea is to use this new model as a "blueprint" for when we want to fully decouple our payments service.

WcaOnRails/app/controllers/api/internal/v1/payment_controller.rb
FinnIckler: This doesn't necessarily come from the "external world" as it is an internal request, I will make sure to use Ruby Money there, but is there maybe a way we can guarantee that this fulfills some kind of API contract?

gregorbg: The only thing close to a contract that we can rely on is ISO 4217. It's the one that defines the currency codes, like `USD` and `EUR`, but actually it also defines sub-units: https://en.wikipedia.org/wiki/ISO_4217#Minor_unit_fractions

WcaOnRails/app/controllers/api/internal/v1/payment_controller.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
+
+class Api::Internal::V1::PaymentController < Api::Internal::V1::ApiController
+  def init
+    return json: { error: "Missing fields" } unless params["attendee_id"].present? && params["amount"].present?
```
gregorbg, line 5: ActiveView will do this for you if you use it like so;

```ruby
attendee_id = params.require(:attendee_id)
iso_amount = params.require(:amount)
```

WcaOnRails/app/controllers/api/internal/v1/payment_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+
+    holder = AttendeePaymentRequest.new(attendee_id: params["attendee_id"])
```
gregorbg, line 7: `Model.new(...)` does create a Ruby object instance but does not actually save it to the database

WcaOnRails/app/controllers/api/internal/v1/payment_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+    competition_id, user_id = holder.competition_and_user_id
+    amount = params["amount"].to_i
+
+    competition = Competition.find(competition_id)
+    user = User.find(user_id)
```
gregorbg, line 12: How do we report errors if either isn't found? `Model.find` just returns `nil`, whereas `Model.find!` throws an error

WcaOnRails/app/controllers/api/internal/v1/payment_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+    # so we need to be able to retrieve this later at any time, even when our server crashes in the meantime…
+    StripePaymentIntent.create!(
+      holder: holder,
+      stripe_transaction: stripe_transaction,
+      client_secret: intent.client_secret,
+      user: user,
+    )
+
+    holder.payment_id = intent.id
+    holder.save
```
gregorbg, line 60: see above: `save` vs. `save!`

WcaOnRails/app/controllers/payment_controller.rb
```diff
@@ -0,0 +1,96 @@
+# frozen_string_literal: true
+
+
+class PaymentController < ApplicationController
+  def payment_config
+    return json: { error: "Please Log in" }, status: :unauthorized unless current_user.present?
+    return json: { error: "Missing fields" } unless params["attendee_id"].present?
```
gregorbg, line 7: see above about `params.require`.

WcaOnRails/app/controllers/payment_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+      rescue Error
+        return redirect_to competition_register_path(competition_id, "registration_down")
+      end
+    end
+
+    redirect_to competition_register_path(competition_id, stored_transaction.status)
+  end
+
+  def payment_refund
+    return json: { error: "Missing fields" } unless params["attendee_id"].present?
```
gregorbg, line 52: see above about `params.require`

WcaOnRails/app/controllers/payment_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+      refund_args,
+      stripe_account: account_id,
+    )
+
+    refund_receipt = StripeTransaction.create_from_api(refund, refund_args, account_id)
+    refund_receipt.update!(parent_transaction: payment_request.receipt) if payment_request.receipt.present?
+
+    begin
+      update_registration_payment(refund_receipt.intent.id, "refund")
+    rescue Error
```
gregorbg, line 90: see above about not rescuing any generic `Error`.

WcaOnRails/app/controllers/registrations_controller.rb
gregorbg, line 537: This will likely not work even now, because `.holder` will already return the AttendeeRequest object and that in turn does **not** have a `type` field. Perhaps something like `stored_intent.holder.is_a? AttendeePaymentRequest`?

WcaOnRails/app/controllers/registrations_controller.rb
```diff
@@ -534,7 +534,14 @@ def stripe_webhook
...
+      begin
+        update_registration_payment(stored_intent.id, stored_intent.status)
+      rescue Error
```
gregorbg, line 540: see above about rescuing errors

WcaOnRails/app/models/attendee_payment_request.rb
```diff
@@ -0,0 +1,6 @@
+class AttendeePaymentRequest < ApplicationRecord
+  belongs_to :receipt, polymorphic: true, optional: true
```
gregorbg, line 2: You only use `belongs_to` if you yourself are storing the foreign ID. This is not the case here because the model you're introducing does not have a `receipt_id`.

WcaOnRails/db/migrate/20231010152057_create_attendee_payment_requests.rb
```diff
@@ -0,0 +1,10 @@
+class CreateAttendeePaymentRequests < ActiveRecord::Migration[7.0]
+  def change
+    create_table :attendee_payment_requests do |t|
+      t.string :attendee_id
+      t.string :payment_id
```
gregorbg, line 5: What is this ID referencing? Is it a Stripe API ID? If yes, we should rather point to our own Stripe log tables which can then be used to retrieve an ID

WcaOnRails/lib/microservices/registrations.rb
```diff
@@ -0,0 +1,24 @@
...
+    "https://#{EnvConfig.ROOT_URL}/competitions/#{competition_id}/#{user_id}/edit&stripe_error=#{stripe_error}"
+  end
+
+  def update_payment_status_path
+    "https://#{EnvConfig.WCA_REGISTRATION_URL}/api/internal/v1/update_payment"
+  end
+
+  def update_registration_payment(id, status)
+    token = get_wca_token
+    response = Faraday.post(update_payment_status_path, headers: { MICROSERVICE_AUTH_HEADER => token, "Content-Type" => "application/json" }, body: { payment_id: id, payment_status: status }.to_json)
```
gregorbg, line 19:
```suggestion
    response = Faraday.post(update_payment_status_path, headers: { MICROSERVICE_AUTH_HEADER => get_wca_token, "Content-Type" => "application/json" }, body: { payment_id: id, payment_status: status }.to_json)
```

WcaOnRails/lib/microservices/registrations.rb
```diff
@@ -0,0 +1,24 @@
...
+    unless response.ok?
+      raise Error "Updating wca-registration failed with error #{response.status} #{response.body}"
+    end
```
gregorbg, line 22:
```suggestion
    raise "Updating wca-registration failed with error #{response.status} #{response.body}" unless response.ok?
```

gregorbg (changes requested):
The thing with Stripe IDs is a bit tricky. The answer to your question is: "It depends". When you're using a Stripe API model, like the one in the controllers that is retrieved by `Stripe::PaymentIntent.create`, then `intent.id` will be a Stripe alphanumeric ID. If you're working with one of the log entries from our `stripe_payment_intents` table, then `our_db_intent.id` will be an integer ID and the actual Stripe "external" ID is stored as `stripe_id`.

WcaOnRails/app/controllers/api/internal/v1/payment_controller.rb
FinnIckler: I did,  `render json: { error: "Registration not found" }, status: :not_found unless competition.present? && user.present?`
because this way we don't need to rescue the error. Not sure what is more 'rails-y'

WcaOnRails/app/controllers/api/internal/v1/payment_controller.rb
FinnIckler: what could go wrong here?

WcaOnRails/app/models/attendee_payment_request.rb
FinnIckler: Then I will add one ;)

WcaOnRails/lib/microservices/registrations.rb
FinnIckler: changed to be more Faraday-y

WcaOnRails/db/migrate/20231010152057_create_attendee_payment_requests.rb
FinnIckler: it's an intent.id, so I think yes?

WcaOnRails/env_config.rb
FinnIckler: Have we talked about how to handle wca-registrations being an external microservice in local development?

gregorbg: Oh heavens no :scream:

WcaOnRails/app/controllers/api/internal/v1/payment_controller.rb
gregorbg: - Validations
- Timeouts
- Intranet network issues
- Locking races
- ...

FinnIckler:
One open question for me is how we prevent users getting the payment config for other payments by 'guessing' the incremental numerical payment_id. One idea I had that we just check if the Transaction Holder is the competitor, which would at least ensure that, while you can still try and guess ids, you only get returned info if you are allowed to

gregorbg:
> One open question for me is how we prevent users getting the payment config for other payments by 'guessing' the incremental numerical payment_id. One idea I had that we just check if the Transaction Holder is the competitor, which would at least ensure that, while you can still try and guess ids, you only get returned info if you are allowed to

Yes, that's the "cheapest" version to prevent guessing payment info. Another would be to assign UUIDs upon ticket generation and use those across services instead of the numeric ID

WcaOnRails/app/controllers/api/internal/v1/payment_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+    # Log the payment attempt. We register the payment intent ID to find it later after checkout completed.
+    stripe_transaction = StripeTransaction.create_from_api(intent, payment_intent_args, account_id)
+
+    # memoize the payment intent in our DB because payments are handled asynchronously
+    # so we need to be able to retrieve this later at any time, even when our server crashes in the meantime…
+    StripePaymentIntent.create!(
+      holder: holder,
+      stripe_transaction: stripe_transaction,
+      client_secret: intent.client_secret,
+      user: user,
```
gregorbg, line 59: I'd prefer to use the Devise `current_user` here. In the long run, it is conceivable that users (for example: parents) are allowed to pay for other users (for example: their kids attending a competition).

In that case, the user derived from the attendee ID would be different from the user ID actually making the payment.

In the case of the Registration microservice, the `holder` will always be an AttendeePaymentRequest which we can use to reconstruct the user whose reg was being paid. By contrast, the `user` column in this DB model should be the user who _made_ the payment. (Right now, this will always be the same person anyways but it doesn't _have_ to be the same person by design)

WcaOnRails/app/controllers/payment_controller.rb
```diff
@@ -0,0 +1,102 @@
...
+
+class PaymentController < ApplicationController
+  def payment_config
+    payment_id = params.require(:payment_id)
+    competition_id = params.require(:competition_id)
+
+    competition = Competition.find(competition_id)
+    stripe_transaction = StripeTransaction.find(payment_id)
+    secret = stripe_transaction.stripe_payment_intent.client_secret
+    render json: { stripe_publishable_key: AppSecrets.STRIPE_PUBLISHABLE_KEY,
```
gregorbg, line 11: I presume the Reg microservice needs this to render the Stripe UI. Why can't you store this in Vault and have the Reg service access it directly from there?

WcaOnRails/app/controllers/payment_controller.rb
```diff
@@ -0,0 +1,102 @@
...
+                   connected_account_id: competition.connected_stripe_account_id,
```
gregorbg, line 12: Also, assuming that this request is trackable in Chrome or Mozilla Dev Tools "Network" tab: Is it safe to pass the `connected_account_id` around? Is that information visible in our current monolith implementation?

WcaOnRails/app/controllers/payment_controller.rb
```diff
@@ -0,0 +1,102 @@
...
+                   client_secret: secret }
+  end
+
+  def payment_finish
+    attendee_id = params.require(:attendee_id)
+    payment_request = AttendeePaymentRequest.find_by(attendee_id: attendee_id)
+    return redirect_to Microservices::Registrations.competition_register_path(competition, "not_found") unless payment_request.present?
+    competition_id, = payment_request.competition_and_user_id
```
gregorbg, line 20: This is very tricky code design (and terribly unintuitive by the designers of Ruby IMO!) because one single comma after the variable name makes all the difference.

To avoid confusion, I'd prefer if you created a `competition_id` method in AttendeePaymentRequest and used that here.

gregorbg: Looking at the next line below, maybe even make that `def competition` directly in the APR

WcaOnRails/app/controllers/payment_controller.rb
```diff
@@ -0,0 +1,102 @@
...
+
+    stored_transaction = StripeTransaction.find_by(stripe_id: intent_id)
+    stored_intent = stored_transaction.stripe_payment_intent
+
+    return redirect_to Microservices::Registrations.competition_register_path(competition, "secret_invalid") unless stored_intent.client_secret == intent_secret
+
+    # No need to create a new intent here. We can just query the stored intent from Stripe directly.
+    stripe_intent = stored_intent.retrieve_intent
+
+    return redirect_to Microservices::Registrations.competition_register_path(competition, "not_found") unless stripe_intent.present?
```
gregorbg, line 35: This should be a different `not_found` compared to the other "APR by attendee ID not found" above

WcaOnRails/app/controllers/payment_controller.rb
```diff
@@ -0,0 +1,102 @@
...
+
+    stored_intent.update_status_and_charges(stripe_intent, current_user) do |charge|
+      ruby_money = charge.money_amount
+      Microservices::Registrations.update_registration_payment(attendee_id, charge.id, ruby_money.cents, ruby_money.currency.iso_code, stripe_intent.status)
+      return redirect_to Microservices::Registrations.competition_register_path(competition_id, "registration_down")
```
gregorbg, line 40: Why is it called `"registration_down"`? What does that mean? Makes me think of "White House Down"...

WcaOnRails/app/controllers/payment_controller.rb
```diff
@@ -0,0 +1,102 @@
...
+    end
+
+    redirect_to Microservices::Registrations.competition_register_path(competition_id, stored_transaction.status)
+  end
+
+  def available_refunds
+    attendee_id = params.require(:attendee_id)
+    payment_request = AttendeePaymentRequest.find_by(attendee_id: attendee_id)
+    transactions = StripePaymentIntent.where(holder: payment_request)
+    render json: { charges: transactions.map { |t| { payment_id: t.stripe_transaction.id, amount: t.parameters["amount"] } } }, status: :ok
```
gregorbg, line 50: This line is too dense, please extract the mapping into one or two lines of separate code

WcaOnRails/app/controllers/payment_controller.rb
```diff
@@ -0,0 +1,102 @@
...
+    payment_id = params.require(:payment_id)
+    attendee_id = params.require(:attendee_id)
+
+    payment_request = AttendeePaymentRequest.find_by(attendee_id: attendee_id)
+    competition_id, = payment_request.competition_and_user_id
+    competition = Competition.find(competition_id)
+
+    unless competition.using_stripe_payments?
+      return render json: { error: "no_stripe" }
+    end
```
gregorbg, line 63:
```suggestion
    return render json: { error: "no_stripe" } unless competition.using_stripe_payments?
```

WcaOnRails/app/controllers/payment_controller.rb
```diff
@@ -0,0 +1,102 @@
...
+
+    charge = StripeTransaction.find(payment_id)
+
+    refund_amount_param = params.require(:refund_amount)
+    refund_amount = refund_amount_param.to_i
+
+    if refund_amount < 0
+      return render json: { error: "refund_zero" }
+    end
```
gregorbg, line 72:
```suggestion
    return render json: { error: "refund_zero" } if refund_amount < 0
```

WcaOnRails/app/controllers/payment_controller.rb
```diff
@@ -0,0 +1,102 @@
...
+      )
+
+    refund_receipt = StripeTransaction.create_from_api(refund, refund_args, account_id)
+    refund_receipt.update!(parent_transaction: charge)
+
+    begin
+      update_registration_payment(attendee_id, refund_receipt.id, refund_amount, currency_iso, "refund")
+    rescue Faraday::Error => e
+      puts e.message
+      puts e.backtrace
```
gregorbg, line 96: Leftover console debugging :upside_down_face:

WcaOnRails/app/controllers/payment_controller.rb
```diff
@@ -0,0 +1,102 @@
...
+    render json: { charges: transactions.map { |t| { payment_id: t.stripe_transaction.id, amount: t.parameters["amount"] } } }, status: :ok
+  end
+
+  def payment_refund
+    payment_id = params.require(:payment_id)
+    attendee_id = params.require(:attendee_id)
+
+    payment_request = AttendeePaymentRequest.find_by(attendee_id: attendee_id)
+    competition_id, = payment_request.competition_and_user_id
+    competition = Competition.find(competition_id)
```
gregorbg, line 59: see other comment above.

WcaOnRails/app/controllers/payment_controller.rb
```diff
@@ -0,0 +1,102 @@
...
+
+    unless competition.using_stripe_payments?
+      return render json: { error: "no_stripe" }
+    end
+
+    charge = StripeTransaction.find(payment_id)
```
gregorbg, line 65: error handling if the `StripeTransaction` record was not found? (Which can simply happen if somebody passes an invalid `payment_id`)

WcaOnRails/app/controllers/registrations_controller.rb
```diff
@@ -525,6 +525,15 @@ def stripe_webhook
           #   in our DB _after_ the backed-off event has been processed. This can lead to a wrong registration order :(
           stored_payment.update!(created_at: audit_event.created_at_remote)
         end
+        if stored_intent.holder.is_a? AttendeePaymentRequest
```
gregorbg, line 528:
```suggestion
        elsif stored_intent.holder.is_a? AttendeePaymentRequest
```

WcaOnRails/app/controllers/registrations_controller.rb
```diff
@@ -525,6 +525,15 @@ def stripe_webhook
...
+          ruby_money = charge_transaction.money_amount
+          begin
+            update_registration_payment(stripe_intent.holder.attendee_id, stored_intent.id, ruby_money.cents, ruby_money.currency.iso_code, stored_intent.status)
+          rescue Faraday::Error
+            logger.error "Couldn't update Microservice"
```
gregorbg, line 533: This error message is not very helpful :/

WcaOnRails/app/controllers/registrations_controller.rb
```diff
@@ -534,7 +543,6 @@ def stripe_webhook
     else
       logger.info "Unhandled Stripe event type: #{event.type}"
     end
-
```
gregorbg, line 537: Why are you removing a blank line that helps readability?

WcaOnRails/config/routes.rb
```diff
@@ -98,6 +98,11 @@
     get '/admin/results/:round_id/new' => 'admin/results#new', as: :new_result
   end
 
+  get 'payment/config' => 'payment#payment_config'
+  get 'payment/finish' => 'payment#payment_finish'
+  get 'payment/refunds' => 'payment#available_refunds'
+  get 'payment/refund' => 'payment#payment_refund'
```
gregorbg, line 104: Can you scope this? cf. https://guides.rubyonrails.org/routing.html#controller-namespaces-and-routing

WcaOnRails/spec/models/attendee_payment_request_spec.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe AttendeePaymentRequest, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
gregorbg, line 6: ...

gregorbg (changes requested):

WcaOnRails/app/controllers/payment_controller.rb
FinnIckler: The stripe microservice needs this to render the stripe UI

WcaOnRails/app/controllers/payment_controller.rb
FinnIckler: I was following the stripe example on their github https://github.com/stripe-samples/accept-a-payment/blob/main/payment-element/client/react-cra/src/App.js
where they also have a config route to read the key. Regardless if the registration service or the payment service saves them in vault, we still need to use the config route to fetch it

WcaOnRails/app/controllers/api/internal/v1/payment_controller.rb
FinnIckler: current_user doesn't exist here, because this is the registration service calling the payment service

WcaOnRails/app/controllers/payment_controller.rb
FinnIckler: not sure how that happened, but the rescue block got eaten

WcaOnRails/app/controllers/api/internal/v1/payment_controller.rb
FinnIckler: There is no api user either, currently the 'user' is the registration service. But I do have a current_user variable in the registration service that I can just pass as a parameter!

gregorbg (approved):

PR #8412, opened 2023-10-06
Changed link to probation and removed probation from teams

danieljames-dj:
This PR will change the link to probation under senior delegate panel & WFC panel. It won't remove the corresponding rows from team_members. After deploying of this PR, I'll manually remove the rows from team_members for probation after verifying that nobody used the old probation edit page in between the migration.

### DIFF
--- a/WcaOnRails/app/models/user.rb
+++ b/WcaOnRails/app/models/user.rb
@@ -684,9 +684,7 @@ class User < ApplicationRecord
     can_manage_teams? ||
       team_leader?(team) ||
       # The leader of the WDC can edit the banned competitors list
-      (team == Team.banned && team_leader?(Team.wdc)) ||
-      # Senior Delegates and WFC Leader and Senior Members can edit Delegates on probation
-      (team == Team.probation && (senior_delegate? || team_leader?(Team.wfc) || team_senior_member?(Team.wfc)))
+      (team == Team.banned && team_leader?(Team.wdc))
   end
 
   def can_view_banned_competitors?
@@ -1248,5 +1246,9 @@ class User < ApplicationRecord
   def senior_or_self
     return nil unless self.delegate_status.present?
     self.delegate_status == "senior_delegate" ? self : self.senior_delegate
+  end
+
+  def is_delegate_in_probation
+    Role.where(user_id: self.id).where("end_date is null or end_date >= curdate()").present?
   end
 end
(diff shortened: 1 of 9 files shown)


WcaOnRails/db/migrate/20231006072155_remove_probation_from_teams.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class RemoveProbationFromTeams < ActiveRecord::Migration[7.0]
+  def change
+    Team.c_find_by_friendly_id!('probation').destroy
```
gregorbg, line 5: Be careful with `destroy` invocations. In your PR description, you wrote that this PR won't remove any rows from `team_members`, but actually it is possible that Rails "magically" (i.e. by virtue of the fact that _we defined_ in our code that there is an association between Team and TeamMember) realizes "oh, the parent team doesn't exist anymore, so all other rows can be cleaned up".

WcaOnRails/app/models/user.rb
```diff
@@ -684,9 +684,7 @@ def can_edit_team?(team)
     can_manage_teams? ||
       team_leader?(team) ||
       # The leader of the WDC can edit the banned competitors list
-      (team == Team.banned && team_leader?(Team.wdc)) ||
-      # Senior Delegates and WFC Leader and Senior Members can edit Delegates on probation
-      (team == Team.probation && (senior_delegate? || team_leader?(Team.wfc) || team_senior_member?(Team.wfc)))
```
gregorbg, line 689: Why is this check no longer necessary? The general statement from the comment directly above (`Senior Delegates and WFC Leader and Senior Members can edit Delegates on probation`) is still valid, even in your new `roles` system. So why is this check disappearing?

gregorbg (changes requested):
You are removing a lot of tests. Is there any way you can come up with equivalent tests that aim to check roughly the same as the old tests, just using your new models and UIs?

WcaOnRails/db/migrate/20231006072155_remove_probation_from_teams.rb
danieljames-dj: I think you are right. I just now checked my database and couldn't find the entry in `team_members` table. In that case, can I do this as well manually instead of going through migration?

gregorbg: Yes, absolutely!

danieljames-dj (commented):
Added the test cases too.

gregorbg (approved):

PR #8413, opened 2023-10-08
MIgrate 'Edit Schedule' page to SemanticUI + React w/ Dispatch

gregorbg:
Bye bye, dear `rootRender()` :wave:
Bye bye, dear `react-bootstrap` :wave:

### (Note: This PR is missing some finishing touches, see below.)

Acts as a (more or less) drop-in replacement for the "Edit Schedule" part. The top-level structure is shamelessly stolen from the "Edit Events" component, and we might consider unifying the "unsaved changes alert" logic.

FullCalendar was migrated to the official React integration, and I've also added timezone support through the Luxon plugin. Native JS `Date` support is really good these days, but unfortunately FullCalendar doesn't recognize this fact and refuses to support timezones unless you provide him with a third-party Date-and-Time plugin. I chose Luxon over MomentJS because the latter seems to be in maintenance-only mode and active development has ceased.

Leaflet also uses the React component now, albeit in an outdated version. If possible, I would like to merge https://github.com/thewca/worldcubeassociation.org/pull/7459 first (which adds accessible Map components in a _much_ cleaner way than here) and then rebase this to reuse the new map integration.

## Things that are still missing

- [x] Tests need to be adjusted to work with the CSS selectors in the new UI
- [x] The UI itself needs some polishing, in some grid rows SemUI refuses to center the buttons and I don't understand why or how to fix it.
- [x] The dropdown to choose a venue room can maybe(?) be made more user friendly
- [ ] The Leaflet map integration is just hastily patched together, see remark above.
- [x] Keyboard support is missing, it's a bit unfortunate but didn't break our neck during EditEvents either.
- [x] Saying "No" to the confirmation messages (especially Deleting Activity from the drag and drop red message rectangle) prompts a mysterious error


WcaOnRails/app/webpacker/components/EditSchedule/index.js
```diff
@@ -0,0 +1,183 @@
...
+
+import { useSaveWcifAction } from '../../lib/utils/wcif';
+import { changesSaved } from './store/actions';
+import wcifScheduleReducer from './store/reducer';
+import Store, { useDispatch, useStore } from '../../lib/providers/StoreProvider';
+import ConfirmProvider from '../../lib/providers/ConfirmProvider';
+import EditVenues from './EditVenues';
+import EditActivities from './EditActivities';
+
+function EditScheduleNew({
```
coder13, line 25: At what point will "new" be removed from this component name?

gregorbg: Until somebody smart like yourself realizes that I accidentally left it in there :innocent:

WcaOnRails/app/webpacker/components/EditSchedule/EditActivities/EditActivityModal.js
```diff
@@ -0,0 +1,110 @@
...
+  'other-tutorial': 'Tutorial for new competitors',
+  'other-multi': 'Cube submission for 3x3x3 Multi-Blind',
+  'other-breakfast': 'Breakfast',
+  'other-lunch': 'Lunch',
+  'other-dinner': 'Dinner',
+  'other-awards': 'Awards',
+  'other-misc': 'Other',
+};
+
+const otherActivityCodeOptions = Object.keys(commonActivityCodes).map((activityCode) => ({
```
kr-matthews, line 23: `Object.entries` gives the key and value, if you want to avoid `commonActivityCodes[activityCode]` for the text.

WcaOnRails/app/webpacker/components/EditSchedule/EditActivities/EditActivityModal.js
```diff
@@ -0,0 +1,110 @@
...
+
+    setActivityCode(evt, data);
+  };
+
+  // We have to assign state in this awkward way because of an opinionated conflict
+  //   between FullCalendar and SemanticUI. See comment at the beginning of index.js for details.
+  useEffect(() => {
+    setActivityCode(activity?.activityCode);
+    setActivityName(activity?.name);
+  }, [activity, setActivityCode, setActivityName]);
```
kr-matthews, line 56: You should be able to omit `setActivityCode` and `setActivityName` from the dependency array; React knows they don't change.

WcaOnRails/app/webpacker/components/EditSchedule/EditActivities/index.js
```diff
@@ -0,0 +1,443 @@
...
+
+  // This part is ugly because Semantic-UI and Fullcalendar disagree
+  //   about how modals should be handled.
+  // According to Semantic-UI, modals are "always there" in the DOM,
+  //   just that their "isOpen" state is false most of the time.
+  //   So they simply don't show but they are already part of the DOM tree.
+  // The (click-)event-based model of Fullcalendar however dictates that
+  //   we can only "instantiate" a modal once the user actually clicks somewhere on the calendar.
+  //   So we pre-fill the modal with empty state and set it accordingly on every event click.
+  // If somebody has a better idea how to handle this, please shout.
```
kr-matthews, line 72: I had a similar issue once. At first we tried only including the modal when it was open:
```
...
  {showActivityModal && <EditActivityModal
                                          // showModal={showActivityModal} // <- don't need this anymore
                                          activity={modalActivity}
                                          startLuxon={modalLuxonStart}
                                          endLuxon={modalLuxonEnd}
                                          dateLocale={calendarLocale}
                                          onModalClose={onActivityModalClose}
                                        />}
...
```
This worked well for having the props/data behave as expected, but it ruined the open/close animation on the modal so ultimately we switched to more-or-less what you're current doing.

WcaOnRails/app/webpacker/components/EditSchedule/EditActivities/EditActivityModal.js
```diff
@@ -0,0 +1,110 @@
...
+};
+
+const otherActivityCodeOptions = Object.keys(commonActivityCodes).map((activityCode) => ({
+  key: activityCode,
+  text: commonActivityCodes[activityCode],
+  value: activityCode,
+}));
+
+function EditActivityModal({
+  showModal,
```
kr-matthews, line 30: I always think this is a function to show the modal. I'd recommend `isModalOpen`. Maybe it's just me though.

WcaOnRails/app/webpacker/components/EditSchedule/store/reducer.js
```diff
@@ -0,0 +1,216 @@
...
+    },
+  }),
+};
+
+export default function rootReducer(state, action) {
+  const reducer = reducers[action.type];
+  if (reducer) {
+    return reducer(state, action);
+  }
+  return state;
```
kr-matthews, line 215: Do we want to log some kind of error or warning here about reducer case not found?

WcaOnRails/app/webpacker/components/EditSchedule/store/reducer.js
```diff
@@ -0,0 +1,216 @@
...
+              ? room.activities.map((activity) => (
+                changeActivityTimezone(
+                  activity,
+                  venue.timezone,
+                  payload.newProperty,
+                )
+              ))
+              : room.activities
+          ),
+        })),
```
kr-matthews, line 136: It would be ideal if objects/arrays which don't get any actual updates could keep the same reference, rather than creating a new identical object/array. That way if they're ever used in a dependency array it won't unexpectedly trigger the callback/effect.
In this case that would mean moving `payload.propertyKey === 'timezone'` up outside of the rooms map.
I don't think any other reducers here have such an issue (unless we let users try and delete ids which don't exist).

kr-matthews (commented):
I reviewed some of the files, will come back for the rest later.

WcaOnRails/app/webpacker/components/EditSchedule/store/reducer.js
coder13: That'd be helpful for debugging. Ideally there shouldn't be errors when this is tested and pushed to prod.

WcaOnRails/app/webpacker/lib/utils/edit-schedule.js
```diff
@@ -9,65 +11,33 @@ export function toDegrees(coord) {
...
-  currentElementsIds.venue = maxId(venues);
-  currentElementsIds.room = maxId(rooms);
-  currentElementsIds.activity = maxId(activities);
-}
-
-export function newVenueId() {
-  currentElementsIds.venue += 1;
-  return currentElementsIds.venue;
-}
+const maxId = (objects) => _.max(objects.map((wcifObj) => wcifObj.id)) || 0;
```
kr-matthews, line 23:
```suggestion
const maxIdOrZero = (objects) => _.max(objects.map((wcifObj) => wcifObj.id)) || 0;
```

WcaOnRails/app/webpacker/lib/show-schedule.js
```diff
@@ -1,5 +1,6 @@
 import _ from 'lodash';
 import { events } from './wca-data.js.erb';
+import { getTextColor } from './utils/calendar';
```
kr-matthews, line 3: Don't think this is used?

kr-matthews (commented):
5 more files to finish reviewing next time I come back!
I think it's standard to use `.jsx` for files which have React components.

WcaOnRails/app/webpacker/components/EditSchedule/EditActivities/index.js
```diff
@@ -0,0 +1,443 @@
...
+    setModalLuxonStart(null);
+    setModalLuxonEnd(null);
+  };
+
+  return (
+    <>
+      <Container textAlign="center">
+        <Dropdown fluid placeholder="Venue" clearable selection options={venueOptions} value={selectedRoomId} onChange={setSelectedRoomId} />
+      </Container>
+      {!!selectedRoomId && (
```
kr-matthews, line 283: I don't think the `!!` is needed; if it's falsey then it won't render, and if it's truthy then the fragment will display.

WcaOnRails/app/webpacker/components/EditSchedule/EditActivities/EditActivityModal.js
gregorbg: Surprisingly enough, this is ESLint which insists that I have to put them in there:

```
ESLint: React Hook useEffect has missing dependencies: 'setActivityCode' and 'setActivityName'. Either include them or remove the dependency array.(react-hooks/exhaustive-deps)
```

WcaOnRails/app/webpacker/components/EditSchedule/EditActivities/index.js
gregorbg: Okay, thanks for hte input! In terms of code changes, I will consider this comment a no-op

WcaOnRails/app/webpacker/components/EditSchedule/store/reducer.js
gregorbg: To be honest, I blatantly stole this from `editEvents`, which doesn't have dedicated error handling either.
Rather than introducing different approaches here and there, I'd prefer having a unified approach with react-redux (or similar) in a separate PR.

WcaOnRails/app/webpacker/components/EditSchedule/store/reducer.js
```diff
@@ -0,0 +1,216 @@
...
+            },
+          ],
+        }) : room)),
+      })),
+    },
+  }),
+
+  [EditActivity]: (state, { payload }) => ({
+    ...state,
+    wcifSchedule: {
```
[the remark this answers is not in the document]
gregorbg: Yup, but the nuances are more subtle. Sometimes I do a `find` to get the reference to the wcifEvent, sometimes I do `map` to change something in-place... On one occasion there's even a `flatMap` so I don't currently see a smart way to unify this.

WcaOnRails/app/webpacker/lib/utils/edit-schedule.js
gregorbg: Changed. I didn't commit your suggestion because of follow-up changes that would only be propagated through my IDE, not through GitHub Review suggested changes.

WcaOnRails/app/webpacker/lib/show-schedule.js
gregorbg: It actually is used: The tab that statically shows a competition's schedule in the public view currently relies on this file.
We can only get rid of it once that other schedule has also been migrated (in a follow-up PR)

WcaOnRails/app/webpacker/components/EditSchedule/index.js
```diff
@@ -0,0 +1,183 @@
...
+
+  const dispatch = useDispatch();
+
+  const [openAccordion, setOpenAccordion] = useState(-1);
+
+  const unsavedChanges = useMemo(() => (
+    !_.isEqual(wcifSchedule, initialWcifSchedule)
+  ), [wcifSchedule, initialWcifSchedule]);
+
+  const onUnload = useCallback((e) => {
```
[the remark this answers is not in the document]
gregorbg: Yes, but then I'd rather rename it everywhere all at once. Similar approaches exist in the `EditEvents` and the soon-to-be-merged `EditCompetition` forms, and we can contemplate a unified approach with a reasonable name for all of them at once.

WcaOnRails/app/webpacker/components/EditSchedule/index.js
```diff
@@ -0,0 +1,183 @@
...
+
+  const save = useCallback(() => {
+    saveWcif(
+      competitionId,
+      { schedule: wcifSchedule },
+      () => dispatch(changesSaved()),
+    );
+  }, [competitionId, dispatch, saveWcif, wcifSchedule]);
+
+  const renderUnsavedChangesAlert = () => (
```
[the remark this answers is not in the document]
gregorbg: Same as above: Unify all usages at once rather than having diverging code.

WcaOnRails/app/webpacker/components/EditSchedule/store/reducer.js
gregorbg: Ah, now I got it. Thanks for your explanation, fixed!

PR #8414, opened 2023-10-09
Bump Fomantic stylesheets to v2.9.3

gregorbg:
Based on the theming overrides from https://github.com/fomantic/Fomantic-UI/commit/0ea6511eddc4203fa25329782ec45ab54f7d2f4d

There is not much to review here, because this is really just third-party minified code being replaced with other third-party minified code. The only notable change is that the bump to Fomantic 2.9 dropped support for non-WOFF files. So we have a couple less assets to ship.

(Sidenote: We should figure out a more consistent way to deploy our own Fomantic theme once we have more microservices)


gregorbg:
Treating as a (glorified) dependency update

PR #8416, opened 2023-10-11
Allow WRT to choose between replica and main database when generating Tokens

FinnIckler:
I added a select to choose between the DBs and a simple vanilla JS function to update the display. 
React seemed to be a bit overkill for this, but some kind of document onready wrap for `document.getElementById("dbs").addEventListener("change", handleDataBaseOptions)` might be needed. 
I tested this locally on the browser, but it might act differently, so please test on staging first before deploying.

### DIFF
--- a/WcaOnRails/app/controllers/admin_controller.rb
+++ b/WcaOnRails/app/controllers/admin_controller.rb
@@ -367,9 +367,14 @@ class AdminController < ApplicationController
     role_credentials = Aws::InstanceProfileCredentials.new
     token_generator = Aws::RDS::AuthTokenGenerator.new credentials: role_credentials
 
-    @token = token_generator.auth_token({
+    @main_token = token_generator.auth_token({
                                           region: EnvConfig.DATABASE_AWS_REGION,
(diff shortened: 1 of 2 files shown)


dunkOnIT:
Just checking in on the status of this PR? Sebastien asked me to keep him updated - let me know if I can help with staging testing or anything!

FinnIckler:
From a code perspective this looks good, if it works on staging I am all for merging it

gregorbg:
This is such a trivial change that I'm almost tempted to ship it as is. Finn, please give a formal code review about the source code changes if you say that they "look good". In the meantime, I will test on staging and report back.

gregorbg:
I've added a small convenience feature that pre-selects the correct database in the PMA dropdown when clicking the "Database" redirect button on the page

gregorbg (approved):
Finn and I agree that it looks cool. He cannot review because technically it's his PR.

dunkOnIT:
Thoughts on making read-replica the default token generated? I'll make the PR if we agree - my case in favour of it is that read-replica should be the default db accessed, and people should opt-in to the live database only if they need write access

PR #8417, opened 2023-10-11
Create a way to enable the New Registration System for competitions

FinnIckler:
Not 100% sure if this is the way to go, but I added
- A new DB Column to Competition (migration pending) that determines if the new Registration System is enabled
- A new /competitions/v2 namespace that renders the new Frontend
- A new Environment variable for the URL of the CDN where the files for the frontend are stored
- An option in the Competition Navigation to enable it, this might just make sense now when testing

If we think this a good approach I will add the migration


WcaOnRails/app/models/competition.rb
```diff
@@ -660,6 +660,14 @@ def trainee_delegate_ids
     @trainee_delegate_ids || trainee_delegates.map(&:id).join(",")
   end
 
+  def enable_v2_registrations
```
gregorbg, line 663: By Ruby convention, methods that modify an existing object in-place should end with an exclamation mark

gregorbg, line 663:
```suggestion
  def enable_v2_registrations!
```

gregorbg: Usages of this method in the controller need to be updated accordingly

WcaOnRails/app/controllers/competitions_v2_controller.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class CompetitionsV2Controller < ApplicationController
+  def show
+    competition_id = params[:id]
+    redirect_to competition_list_path unless competition_id.present?
```
gregorbg, line 6:
```suggestion
    competition_id = params.require(:id)
```

gregorbg: This will do appropriate error handling for you if the param is not present.

gregorbg (changes requested):
Cosmetics

WcaOnRails/app/models/competition.rb
FinnIckler: I thought ! means it returns an error

FinnIckler: like find! vs. find

FinnIckler:
so I need to add something to the dev export?

gregorbg (changes requested):
The diff contains the schema.rb definition for jwt_denylist. Please rebase on master to get rid of this unrelated change

gregorbg (approved):

PR #8422, opened 2023-10-15
Replace WCA ID input with user search

danieljames-dj:
This is basically a rough code to do the basic working implementation. Can someone please suggest if there is a better way to handle single input in OmniSearch?

### DIFF
--- a/WcaOnRails/app/webpacker/components/DelegateProbations/index.jsx
+++ b/WcaOnRails/app/webpacker/components/DelegateProbations/index.jsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Button, Input, Table } from 'semantic-ui-react';
+import { Button, Table } from 'semantic-ui-react';
 import DatePicker from 'react-datepicker';
 import UserBadge from '../UserBadge';
 import useLoadedData from '../../lib/hooks/useLoadedData';
@@ -62,7 +62,7 @@ function ProbationListTable({
 }
 
 export default function DelegateProbations() {
-  const [wcaId, setWcaId] = useInputState('');
+  const [userId, setUserId] = React.useState();
   const {
     data, loading, error, sync,
   } = useLoadedData(delegateProbationDataUrl);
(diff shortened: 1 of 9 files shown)


gregorbg (changes requested):
Please wait for https://github.com/thewca/worldcubeassociation.org/pull/7459, which will simplify working with the Semantic search bars from `OmnisearchDropdown`.

Your suggested implementation unfortunately has a few flaws (most notably, you're relying on inline CSS for components that should be shared across a whole wide range of different layouts, which is very impractical!) so I'll just block this until the aforementioned PR is merged.

danieljames-dj:
Okay, will wait for that PR to be merged.

danieljames-dj:
Marking this PR as blocked, awaiting for https://github.com/thewca/worldcubeassociation.org/pull/7459 to merge.

danieljames-dj:
@gregorbg Since Board has requested for this feature in probation page (mainly because it blocks the case where there is delegate without WCA ID), I've brought the corresponding code from PR #7459 to this PR. Could you please review again if this is fine?

WcaOnRails/app/webpacker/components/DelegateProbations/index.jsx
```diff
@@ -75,9 +76,18 @@ export default function DelegateProbations() {
...
+        value={userId}
+        onChange={setUserId}
+        multiple={false}
+        delegateOnly
+      />
       <Button
-        onClick={() => save(startDelegateProbationUrl, { wcaId }, sync, { method: 'POST' })}
+        onClick={() => {
+          save(startDelegateProbationUrl, { userId }, sync, { method: 'POST' });
+          setUserId(null);
```
gregorbg, line 88: The `setUserId` action should be part of the success action in the `save` invocation (the part that currently forwards to `sync`)

gregorbg (changes requested):
Tiny change about the success hook after saving, otherwise fine!

WcaOnRails/app/webpacker/components/DelegateProbations/index.jsx
danieljames-dj: Okay, done.

gregorbg (approved):

Issue #8349, opened 2023-09-28
Contact form for Results Team


danieljames-dj:
This feature request have three subtasks:

1. Modify [WCA contacts form](https://www.worldcubeassociation.org/contact/website) to add 'Contact Results Team' and should send email to 'results@worldcubeassociation.org' instead of 'contacts@worldcubeassociation.org'.
2. Replace all `mailto:results@` with the contact form (with 'Contact Results Team' pre-selected - the pre-selection can be done by probably adding some query params in URL).
3. Add description in the form when 'Contact Results Team' is selected and the description has to be fetched from 'app/views/contact/_website_wrt_description.md' which can be set as an empty file initially. (Later WRT will be adding the required description in that file)

Reason for the feature request: Currently, WRT is receiving lots of emails directly from users as the website is redirecting them to the email. Many of them doesn't have clear information on the requirement, so we are expecting this form to make the work easier for both the users and WRT.

If anybody is willing to work on any of the above three subtasks, feel free to ping in this thread. You don't have to take up all three subtasks together, in case you want to do only one of them, you can do that and leave remaining for others.

danieljames-dj:
I've done the first subtask [here](https://github.com/thewca/worldcubeassociation.org/pull/8425/files). The second and third subtasks are open to pickup by anyone.

pranav-027:
Hello @danieljames-dj worked on 2nd subtask here is the MR: https://github.com/thewca/worldcubeassociation.org/pull/10174
I could only find some locale YAML files with results team mail. Is there any other place that I might be missing?

danieljames-dj:
I think you covered all cases @pranav-027, some locale files might already be updated by the translators.

danieljames-dj:
Now this is completely done except 3. But I think 3 is not needed now as the contact form is pretty clear currently. Thanks to everyone who worked on this and helped in review.

PR #8425, opened 2023-10-17
Contact form for results team

danieljames-dj:
First subtask of https://github.com/thewca/worldcubeassociation.org/issues/8349.

### DIFF
--- a/WcaOnRails/app/models/website_contact.rb
+++ b/WcaOnRails/app/models/website_contact.rb
@@ -24,6 +24,8 @@ class WebsiteContact < ContactForm
       if competition.present?
         return ValidateEmail.valid?(competition.contact) ? competition.contact : competition.managers.map(&:email)
       end
+    elsif inquiry == "results_team"
+      return Team.wrt.email
     end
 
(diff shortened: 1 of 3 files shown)


WcaOnRails/app/models/website_contact.rb
```diff
@@ -24,6 +24,8 @@ def to_email
       if competition.present?
         return ValidateEmail.valid?(competition.contact) ? competition.contact : competition.managers.map(&:email)
       end
+    elsif inquiry == "results_team"
+      return "results@worldcubeassociation.org"
```
gregorbg, line 28: Please try to use proper data bindings wherever possible: `Team.wrt.email`

(I know that `contact@` is hardcoded below, but that's because it's a general contact address and not WCT's _actual_ email)

gregorbg (changes requested):

WcaOnRails/app/models/website_contact.rb
danieljames-dj: Okay, done.

gregorbg (approved):

PR #8426, opened 2023-10-17
Enable parallel results posting

danieljames-dj:
WRT would like to try out parallel results posting as it is actually possible now. If it is creating trouble for us, maybe we will bring it back later. But we hope it won't cause any trouble to us. Also, in long future, we should be able to post results in parallel.


gregorbg:
Am I correct in assuming that this is green-lit by WRT?

SAuroux:
@gregorbg Yes, we discussed this in the team and decided to give this a try.

Issue #5118, opened 2020-01-30
Implement Trainee Delegates


cubewhiz:
Following the passage of the Motions, we will need to crate a new Delegate role for Trainee Delegates.

They should have the following rights on the website:
- to be listed as a Delegate for WCA Competitions (only if another non-Trainee Delegate is also listed)
- to read and participate in Delegate reports
- to participate in writing Delegate reports (but not be able to submit)

While we wait for a restructured Delegates table, I wonder if it makes sense to create this as a "team" for now. I do not think Trainee Delegates should be listed on the website as they are not WCA Staff (or if they are, it should not be integrated in the table of Delegates).

maconard:
I'd just like to present a user interface idea for implementing this:

In the Organizer View edit screen of a competition where delegates/organizers are selected, we could put a similar box between WCA Delegates and Organizers called Trainee Delegates that is grayed out until at least one Delegate has been set in the WCA Delegates selector. 

Then trainees aren't mixed in the results with regular delegates, and it's fairly clear that trainees are being added second to work with the primary delegate(s). This would also need one catch to not allow a lone WCA Delegate to be de-selected while the trainee box is filled.

![trainee_selector](https://user-images.githubusercontent.com/31703304/73977132-bdffc780-48f7-11ea-8755-5e84a101b643.png)
(image is easier to read if you click it)

This could also be hidden until a checkbox "Allow adding Trainee Delegates to this competition" is selected, if people think it's too much space in general (as most comps may not have any trainees for some time).

cubewhiz:
I like that. I don't think it needs to be hidden on the Organizer page.

viroulep:
A few comments and hopefully useful hints:
  - It's not clear what would be best between adding a team and adding an extra status in our delegate_status enum. Adding an element to the enum would require to adjust our `staff?` method which consider staff anyone who is any kind of delegates. But besides that, adding trainee as a delegate status pretty much avoid any other change regarding access right.
  - if we go the new status way, on the WCA Delegates page we can filter out trainees
  - I guess the UI Michael described makes sense, on the server side the changes needed would be to validate the inclusion of the delegates field in all but trainee status, and for the field they would need to be trainee only. We may need to adjust the javascript and our user API to be able to filter users based on trainee delegate status.

KitClement:
[...]

PR #8430, opened 2023-10-20
Moving delegates page completely to react and bringing admin mode button

danieljames-dj:


WcaOnRails/app/webpacker/components/Delegates/index.jsx
```diff
@@ -87,84 +97,121 @@ function DelegatesOfRegion({ activeSeniorDelegate, delegates, isAdminMode }) {
...
+          <Header>{I18n.t('delegates_page.regions')}</Header>
+          <Menu vertical>
+            {seniorDelegates.map((seniorDelegate) => (
+              <Menu.Item
+                key={`region-${seniorDelegate.id}`}
+                name={seniorLocationToRegion(seniorDelegate.location || '')}
+                active={activeSeniorDelegate === seniorDelegate}
+                onClick={() => {
+                  setActiveSeniorDelegate(seniorDelegate);
                 }}
```
kr-matthews, line 158:
```suggestion
                onClick={() => setActiveSeniorDelegate(seniorDelegate)}
```

WcaOnRails/app/webpacker/components/Delegates/index.jsx
```diff
@@ -87,84 +97,121 @@ function DelegatesOfRegion({ activeSeniorDelegate, delegates, isAdminMode }) {
...
-  const seniorDelegates = React.useMemo(() => delegates
+export default function Delegates() {
+  const { data, loading, error } = useLoadedData(delegatesStaticPageDataUrl);
+
+  const { delegates, canViewDelegateMatters } = data || {};
+  const seniorDelegates = React.useMemo(() => !!delegates && delegates
     .filter((user) => user.delegate_status === 'senior_delegate')
     .sort((user1, user2) => (user1.location || '').localeCompare(user2.location || '')), [delegates]);
 
   const [activeSeniorDelegate, setActiveSeniorDelegate] = React.useState(seniorDelegates?.[0]);
```
kr-matthews, line 108: Since `seniorDelegates` will be falsey on initial render now, and to avoid duplicating this logic from the use effect.
```suggestion
  const [activeSeniorDelegate, setActiveSeniorDelegate] = React.useState();
```

WcaOnRails/app/webpacker/components/Delegates/index.jsx
kr-matthews, line 1: Some indentation is off, can you run prettier.

WcaOnRails/app/webpacker/components/Delegates/index.jsx
```diff
@@ -87,84 +97,121 @@ function DelegatesOfRegion({ activeSeniorDelegate, delegates, isAdminMode }) {
...
+  const [adminMode, setAdminMode] = React.useState(false);
+
+  React.useEffect(() => {
+    setTitle(document, I18n.t('delegates_page.title'));
+  }, []);
```
kr-matthews, line 113: Should the react components be the ones responsible for setting the document title and such?

WcaOnRails/app/webpacker/components/Delegates/index.jsx
```diff
@@ -87,84 +97,121 @@ function DelegatesOfRegion({ activeSeniorDelegate, delegates, isAdminMode }) {
...
+                  value={activeSeniorDelegate.id}
+                  onChange={(event, { value }) => {
+                    setActiveSeniorDelegate(
+                      seniorDelegates.find((seniorDelegate) => seniorDelegate.id === value),
+                    );
+                  }}
+                />
+              </Grid.Row>
+              {/* TODO: Fix Senior Delegate ribbon CSS for tablet and mobile view,
+            and enable this component for all devices */}
```
kr-matthews, line 191: Is this going to be done in this PR? If not, is there an issue for it?

kr-matthews (changes requested):

WcaOnRails/app/webpacker/components/Delegates/index.jsx
danieljames-dj: Done.

WcaOnRails/app/webpacker/components/Delegates/index.jsx
danieljames-dj: Won't be done in this PR, as it requires some CSS changes and it might take some time for me.  Maybe somebody else can take it or I'll do it in long future if nobody take it up.

danieljames-dj (commented):
Thanks @kr-matthews for the review.

WcaOnRails/app/views/application/react_component.erb
danieljames-dj, line 1: @gregorbg The purpose of this file is to allow delete the `.html.erb` if it contains only rendering the `react_component` and instead use this as given in the `routes.rb`. I hope this is fine.

WcaOnRails/app/webpacker/components/Delegates/index.jsx
kr-matthews: Sounds good, I just want to make sure there's an issue on GitHub so that the TODO doesn't get forgotten.

WcaOnRails/app/views/application/react_component.erb
gregorbg: I'm afraid this is too generic. We can think about such approaches when we have decided on a routing application and other general layout design decisions. For now, please use the "classic" approach even if it feels a bit silly to create lots of specific `.html.erb` files that only serve one React component.

WcaOnRails/config/routes.rb
```diff
@@ -185,7 +185,8 @@
   get 'contact' => 'static_pages#contact'
   get 'documents' => 'static_pages#documents'
   get 'education' => 'static_pages#education'
-  get 'delegates' => 'static_pages#delegates'
+  get 'delegates' => 'application#react_component', defaults: { component_name: "Delegates", title: 'delegates_page.title' }
```
gregorbg, line 188: Use a specific controller action please. (Also see my other comment above).
This `defaults` and `locals` approach feels way to hacky for something that we haven't even fully decided on yet.

WcaOnRails/app/webpacker/lib/requests/routes.js.erb
```diff
@@ -89,3 +89,5 @@ export const delegateProbationDataUrl = `<%= CGI.unescape(Rails.application.rout
 export const startDelegateProbationUrl = `<%= CGI.unescape(Rails.application.routes.url_helpers.start_delegate_probation_path) %>`;
 
 export const endDelegateProbationUrl = `<%= CGI.unescape(Rails.application.routes.url_helpers.end_delegate_probation_path) %>`;
+
+export const delegatesStaticPageDataUrl = `<%= CGI.unescape(Rails.application.routes.url_helpers.delegates_data_path(format: "json")) %>`;
```
gregorbg, line 93: Is the `format: "json"` part strictly necessary? I don't recall ever using this for other routes...

gregorbg (changes requested):

WcaOnRails/app/views/application/react_component.erb
danieljames-dj: Okay got it, removed it.

WcaOnRails/config/routes.rb
danieljames-dj: Done.

WcaOnRails/app/webpacker/lib/requests/routes.js.erb
danieljames-dj: The reason was that the path had same URL as that of delegates page, and hence to distinguish format "json" was needed. I've now created a new scope under routes named `page_data` and added `get 'delegates'` there to get the data of delegates page. This helped to remove `format "json"`.

I could have actually named the function name in `static_controller` to something else like `delegate_data` to fix it, but I feel a new scope named `page_data` will be good to get the page data and maybe can be extended to fetching page data of other pages as well in future.

WcaOnRails/app/webpacker/components/Delegates/index.jsx
danieljames-dj: Done, https://github.com/thewca/worldcubeassociation.org/issues/8435.

WcaOnRails/app/controllers/static_pages_controller.rb
```diff
@@ -6,8 +6,12 @@ class StaticPagesController < ApplicationController
   def home
   end
 
-  def delegates
+  def delegates_data
     @delegates = User.where.not(delegate_status: nil)
```
gregorbg, line 10: It is not necessary to prefix variables with `@` if there is no view template that references these (global) variables. Using `delegates = ...` should suffice

WcaOnRails/config/routes.rb
```diff
@@ -200,6 +200,10 @@
   get 'wca-workbook-assistant' => 'static_pages#wca_workbook_assistant'
   get 'wca-workbook-assistant-versions' => 'static_pages#wca_workbook_assistant_versions'
 
+  scope 'page_data' do
```
gregorbg, line 203: Is there an exsting "delegates" scope that this can be moved under?
`page_data` is a bit too generic for my taste. If no existing scope really fits, what about `roles_data` or similar?

WcaOnRails/app/webpacker/components/Delegates/index.jsx
```diff
@@ -77,94 +86,135 @@ function DelegatesOfRegion({ activeSeniorDelegate, delegates, isAdminMode }) {
...
-  delegates,
-  isEditVisible,
-}) {
-  const seniorDelegates = React.useMemo(() => delegates
-    .filter((user) => user.delegate_status === 'senior_delegate')
-    .sort((user1, user2) => (user1.location || '').localeCompare(user2.location || '')), [delegates]);
+export default function Delegates() {
+  const { data, loading, error } = useLoadedData(delegatesStaticPageDataUrl);
+
+  const { delegates, canViewDelegateMatters } = data || {};
```
gregorbg, line 100: You should be able to destructure this directly in the `useLoadedData` return value above

WcaOnRails/app/webpacker/components/Delegates/index.jsx
```diff
@@ -77,94 +86,135 @@ function DelegatesOfRegion({ activeSeniorDelegate, delegates, isAdminMode }) {
...
+  const seniorDelegates = React.useMemo(
+    () => !!delegates
+      && delegates
+        .filter((user) => user.delegate_status === 'senior_delegate')
+        .sort((user1, user2) => (user1.location || '').localeCompare(user2.location || '')),
+    [delegates],
```
gregorbg, line 106: Something like `if (!delegates) return [];` looks more readable. This is too dense.

gregorbg (changes requested):

WcaOnRails/app/controllers/static_pages_controller.rb
danieljames-dj: Done.

WcaOnRails/app/webpacker/components/Delegates/index.jsx
danieljames-dj: I tried that, but was getting "Cannot read properties of null" error. Even if I give default value `{}` to data, it's not accepting as `useLoadedData` is sending some value.

WcaOnRails/app/webpacker/components/Delegates/index.jsx
danieljames-dj: Done.

WcaOnRails/config/routes.rb
danieljames-dj: I thought about that, but it's actually returning two data: `delegates` and `canViewDelegateMatters`. So I thought either I should go with some generic name which can accept both or I should go with two different APIs instead of getting both into one. Is it fine if I use two different APIs to get both the data?

danieljames-dj: After thinking again, I think having two different APIs will be better - one to fetch delegate retails, and another to fetch roles of current user. Can I go ahead with two different APIs?

danieljames-dj:
(Rebased to avoid conflicts with master)

gregorbg:
Needs rebase after merging https://github.com/thewca/worldcubeassociation.org/pull/8464

danieljames-dj:
Rebased. Also, I'm adding the doubt that I asked you in Slack so that you won't miss it in case you see this comment before my slack message:

I made the `page_data#delegates` because it had not just one data, but two: `delegates` data and `canViewDelegateMatters` data. If you are fine with me creating `api/v0/my_roles`, I can also create `api/v0/delegates` which will return the delegates data and hence use the two APIs instead of single API, what do you think is the best approach?

gregorbg:
> what do you think is the best approach?

In the context of this specific PR, I don't understand yet why a `my_roles` API is necessary. You should focus on the APIs that are strictly necessary for this PR and create them as appropriate.

danieljames-dj:
> > what do you think is the best approach?
> 
> In the context of this specific PR, I don't understand yet why a `my_roles` API is necessary. You should focus on the APIs that are strictly necessary for this PR and create them as appropriate.

Okay, in that case will go with the current approach. I guess no change is required from my end.

danieljames-dj:
Again got merge conflict after last few commits, resolved it.

WcaOnRails/app/controllers/static_pages_controller.rb
```diff
@@ -13,8 +13,12 @@ class StaticPagesController < ApplicationController
   def home
   end
 
-  def delegates
```
gregorbg, line 16: You are replacing this `delegates` method by `delegates_data`.
But I think you also need an (at least empty!) `def delegates` method to redirect the controller to the Rails template that injects the React component(?)

WcaOnRails/app/webpacker/components/Delegates/index.jsx
```diff
@@ -34,35 +43,35 @@ function DelegatesOfRegion({ activeSeniorDelegate, delegates, isAdminMode }) {
...
       <Table.Body>
         {sortedDelegates([
           activeSeniorDelegate,
-          ...delegates.filter((user) => (
-            user.senior_delegate_id === activeSeniorDelegate.id
-              && user.delegate_status !== 'trainee_delegate'
-          )),
+          ...delegates.filter(
+            (user) => user.senior_delegate_id === activeSeniorDelegate.id
+              && (user.delegate_status !== 'trainee_delegate' || isAdminMode),
```
gregorbg, line 63: Am I reading it correctly that you are only allowing admins to edit Trainee Delegates?
If yes, how does that make sense with respect to WCA Motions?

WcaOnRails/app/webpacker/components/Delegates/index.jsx
```diff
@@ -77,94 +86,136 @@ function DelegatesOfRegion({ activeSeniorDelegate, delegates, isAdminMode }) {
...
+  const { data, loading, error } = useLoadedData(delegatesStaticPageDataUrl);
+
+  const { delegates, canViewDelegateMatters } = data || {};
+  const seniorDelegates = React.useMemo(() => {
+    if (delegates) {
+      return delegates
+        .filter((user) => user.delegate_status === 'senior_delegate')
+        .sort((user1, user2) => (user1.location || '').localeCompare(user2.location || ''));
+    }
+    return null;
```
gregorbg, line 107:
```suggestion
    return delegates
        ?.filter((user) => user.delegate_status === 'senior_delegate')
        ?.sort((user1, user2) => (user1.location || '').localeCompare(user2.location || ''));
```

WcaOnRails/app/webpacker/components/Delegates/index.jsx
```diff
@@ -77,94 +86,136 @@ function DelegatesOfRegion({ activeSeniorDelegate, delegates, isAdminMode }) {
...
+  }, [delegates]);
+
+  const [activeSeniorDelegate, setActiveSeniorDelegate] = React.useState();
+  const [adminMode, setAdminMode] = React.useState(false);
 
-  const [activeSeniorDelegate, setActiveSeniorDelegate] = React.useState(seniorDelegates?.[0]);
+  React.useEffect(() => {
+    setActiveSeniorDelegate(seniorDelegates?.[0]);
+  }, [seniorDelegates]);
```
gregorbg, line 115: Why are you assuming that there are multiple Senior Delegates, and then "hooking in" a single Senior Delegate? I am asking because this smells like a bad way to use the `useEffect` hook.

gregorbg (changes requested):

WcaOnRails/app/controllers/static_pages_controller.rb
danieljames-dj: I think no, because it is automatically taking <controller_name>/<method_name>.html.erb (here `static_pages/delegates.html.erb`). I have used empty functions in `panel_controller` in past as I had to restrict the access to that path based on permissions.

WcaOnRails/app/webpacker/components/Delegates/index.jsx
danieljames-dj: Actually nope. What I meant was only admins can view trainee delegates. Public users won't be able to see trainee delegates in delegates page because that was what decided by Board (https://github.com/thewca/worldcubeassociation.org/issues/5118).

Note: By admin, I meant the people who can access the delegate admin mode, which include delegates as well. Just pointing it not to get confused with the user.admin of rails, which is WST seniors.

WcaOnRails/app/webpacker/components/Delegates/index.jsx
danieljames-dj: The `seniorDelegates` array has the list of all senior delegates (from all regions). This useEffect is making the first senior delegate as the 'selected' senior delegate. Later, when user switches between regions, it will change again. Also, this is temporary because we will soon migrate to the user_groups API to fetch the regions and roles API to fetch the delegates. Once the APIs are ready, we will be able to migrate it to a better way.

WcaOnRails/app/controllers/static_pages_controller.rb
gregorbg: Cool I never knew that Rails was smart enough to route through these defaults. Thanks for letting me know!

WcaOnRails/app/webpacker/components/Delegates/index.jsx
gregorbg: Okay, then I'll accept it as a "temporary thing" :upside_down_face:

WcaOnRails/app/webpacker/components/Delegates/index.jsx
gregorbg: Okay, makes sense now!

PR #8431, opened 2023-10-25
Allow waiting list deadline during the competition

viroulep:
Based on previous WCAT decisions to allow this, and based on Wilson's input on Sorg'Open 2023 report, it's pretty clear that being unable to set the WL deadline during the competition is a software restriction, and not something forbidden by the WCRP.

This PR:
  - fixes the competition's validation to enable setting the deadline during the competition
  - fixes the test to check it cannot be set *after* the competition
  - fixes the test to check it *can* be set during the competition

Of course someone from WCAT should come here/contact WST to confirm I'm not making things up and that they are fine with these changes :)


gregorbg:
Confirmed by Wilson Alvis of WCAT via email.

viroulep:
Thanks @gregorbg !

PR #5198, opened 2020-02-21
Add .well-known/change-password url

UnsolvedCypher:
Closes #3991

PR #8433, opened 2023-10-28
Host the Apple Pay domain association file

Rubiks-boy:
This PR hosts the Apple Pay domain association file (downloaded from https://stripe.com/files/apple-pay/apple-developer-merchantid-domain-association) at the following URL: `https://www.worldcubeassociation.org/.well-known/apple-developer-merchantid-domain-association`

This will allow regional organizations to collect registration fees through Apple Pay after [verifying the WCA domain in their Stripe Dashboard](https://stripe.com/docs/payments/payment-methods/pmd-registration?platform=dashboard#verify-your-domain-with-apple-pay).

I manually tested the change by running the docker container and visiting http://localhost:3000/.well-known/apple-developer-merchantid-domain-association.


FinnIckler:
hey Adam,
this won't work on prod, because we are blocking all URLs starting with . in the nginx config see https://github.com/thewca/worldcubeassociation.org/pull/5198#issuecomment-1772742705

Rubiks-boy:
@FinnIckler good catch, thanks!

I followed the approach in [this stack overflow post](https://stackoverflow.com/questions/34259548/how-to-disallow-access-to-all-dot-directories-except-well-known). Do you have any advice on how to test the change in an env more similar to prod?

FinnIckler:
There is an nginx running on port 8000 in the `docker-compose.php.yml` which you can use by running `docker compose -f docker-compose.yml -f docker-compose.php.yml up` (wrote this from memory but I think that's how it should work)

gregorbg:
Cool stuff, thanks for the contribution!

I'm a bit hesitant to merge as-is, because using Apple Pay could cause us to "appear more on the radar" of Stripe, and our legal foundations of using Stripe (for what they sometimes perceive to be "gambling") are fragile at best.

How urgent is this feature for you? Could you wait until early next year when we will likely onboard other payment gateways before making a big change like this?

Rubiks-boy:
yep! no rush on it, we can wait

gregorbg:
Bump after eons! It seems that per the documentation that you linked above:

> This will allow regional organizations to collect registration fees through Apple Pay after [verifying the WCA domain in their Stripe Dashboard](https://stripe.com/docs/payments/payment-methods/pmd-registration?platform=dashboard#verify-your-domain-with-apple-pay).

...that things have changed on Stripe's side and hosting this file on our end is no longer necessary as a prerequisite for enabling Apple Pay. Instead, my understanding of the (now rewritten) Stripe documentation is that individual accounts (that hook into our WCA Connect through OAuth) are repsonsible for individually registering the WCA homepage as a "verified URL" that collects funds, and allow Apple Pay on their end. For us, the embedded Payment Element will then just happily oblige and display whatever the connected account has configured.

@Rubiks-boy Can you confirm this is correct, and by extension that this PR can be closed without further action from us?

Rubiks-boy:
It looks like more needs to be done - @coder13 and I followed these updated instructions for our RO, and even though Apple Pay was now shown as being set up, Apple Pay isn’t showing in the Payment Element.

I’m not sure if it’s related to hosting this file or something else with the Elements integration.

Rubiks-boy:
I had tried it on production against a real competition website using my iPhone, and Apple Pay wasn’t showing. I have cards registered with Apple Pay, and I’ve used Apple Pay on it before with other websites.

I’m not sure - this was on my phone, so I didn’t have easy access to the dev console.

gregorbg:
> I’ll give it a shot on my Macbook tonight with Safari and see what outputs in the console.

Kindly ping @Rubiks-boy :upside_down_face: Apple Pay is notoriously hard to test/debug locally, because they're incredibly strict with non-HTTPS connections. So I'm sorry that I have to rely on you so much since I don't own any Apple devices :see_no_evil:

gregorbg:
Kindly ping about Macbook updates @Rubiks-boy

Rubiks-boy:
Hi Gregor, sorry for the really long delay - I finally had the chance to pull up the dev tools on my laptop.

* Nothing appears in the console logs
* I dug through some of the analytics Stripe was sending, and it seemed like it was trying to check Apple Pay's availability but that it's getting marked as not available by Apple.
  * To confirm this, I ran `await window.ApplePaySession.canMakePaymentsWithActiveCard("merchant.www.worldcubeassociation.org.acct_<omitted>.stripe")` in my console (this is the merchant identifier sent as part of those analytics I was digging through), and it returns that Apple Pay isn't available.
  * For comparison, running `await window.ApplePaySession.canMakePaymentsWithActiveCard("merchant.docs.stripe.com.stripe")` from the console of [Stripe's docs](https://docs.stripe.com/payments/payment-element) returns true (Apple Pay shows up as a payment method on that site)
* My best guess is that getting it to work would require following [this doc](https://docs.stripe.com/payments/payment-methods/pmd-registration?dashboard-or-api=dashboard#register-your-domain-while-using-connect) – specifically the section on Connect.
  * For example, as part of the onboarding process when a new connected account is linked to the WCA, you'd send that `/payment_method_domains` request with the connected account ID to set up Apple Pay.

Side note: for testing different browsers, would highly recommend trying out something like [Browserstack](https://www.browserstack.com/) - I think they even support testing things like Apple Pay!

PR #8463, opened 2023-11-01
Add devise-jwt and jwt auth route to Monolith

FinnIckler:
Superseeds #7986


WcaOnRails/.env.development
```diff
@@ -20,3 +20,4 @@ RECAPTCHA_PRIVATE_KEY=
 DATABASE_PASSWORD=
 MAILCATCHER_SMTP_HOST=localhost
 CRONJOB_POLLING_SECONDS=0
+JWT_KEY=jwt-test-secret
```
gregorbg, line 23: Maybe something akin to `jwt-development-secret` would suit this file better?

WcaOnRails/app/controllers/api/v0/users_controller.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class Api::V0::UsersController < Api::V0::ApiController
+  def token
+    if current_user
+      render json: { status: "ok" }
+    else
+      render status: :unauthorized, json: { error: "Please log in" }
```
gregorbg, line 8: Do we have the "Please log in" (or a similar, equivalent phrasing) somewhere in our i18n file?
I feel like some translation should exist already.

gregorbg (changes requested):

WcaOnRails/app/controllers/api/v0/users_controller.rb
FinnIckler: we do not have one, I added one though, because I can then use it in the other PRs as well

gregorbg (approved):

gregorbg:
Changes are fine, according to GitHub there are still merge conflicts to be resolved.

FinnIckler:
I doesn't show it on my nd

gregorbg:
How is that even possible...? :sweat_smile: 

image

gregorbg:
I'll just squash it in

PR #8464, opened 2023-11-01
[React Migration] Migrate WFC Panel to React

danieljames-dj:
Desktop view:
Screenshot 2023-11-03 at 15 15 32

Mobile view:
Screenshot 2023-11-05 at 07 46 51


WcaOnRails/app/webpacker/lib/hooks/useRequest.jsx
```diff
@@ -0,0 +1,49 @@
...
+  React.useEffect(() => {
+    if (!loading) {
+      if (data && successTrigger) {
+        successTrigger();
+      }
+      if (error && errorTrigger) {
+        errorTrigger();
+      }
+    }
+  }, [loading, data, error, successTrigger, errorTrigger]);
```
kr-matthews, line 18: I think the `if (!loading)` wrapping can be removed. The only time that `data` and `error` might potentially change is when `loading` changes from `true` to `false`. But even so, maybe keeping it in for clarity is best.

One problem with this is that if someone else uses this hook and passes in non-memoized functions then the effect will trigger every re-render. Could the triggers just be put in the `wcaRequest` below?

WcaOnRails/app/webpacker/lib/hooks/useRequest.jsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
+import { fetchWithAuthenticityToken } from '../requests/fetchWithAuthenticityToken';
+
+export default function useRequest(successTrigger = null, errorTrigger = null) {
+  const [loading, setLoading] = React.useState(false);
+  const [data, setData] = React.useState(null);
+  const [error, setError] = React.useState(null);
```
kr-matthews, line 7: Since these three are very dependent (when one updates, usually most/all of the others also update in sync) a reducer would make a lot of sense. Then instead of
```
      setLoading(true);
      setData(null);
      setError(null);
```
you could just `dispatch({ action: 'reset' } )`, etc.

WcaOnRails/app/webpacker/lib/hooks/useRequest.jsx
kr-matthews, line 1: Don't we have existing custom hooks which do data fetching?

gregorbg: Yes but they all assume JSON. Daniel introduced this hook because he wants to fetch non-JSON data, but I think it's possibly not necessary at all actually :sweat_smile: (bigger comment pending)

gregorbg (changes requested):
The way I see it, you only use the new `useRequest` hook to load a CSV (or TSV, doesn't matter) from the server and then _immediately_ present it to the user for download.

If this assumption is correct, then your entire new hook is unnecessary (sorry! :see_no_evil:). You just redirect the user to the same link that you're pointing your React component to. Instead of loading the CSV data into React and then creating a hidden `a href`, the user will be redirected to the backend which directly replies with the CSV data and sends headers that prompt the user's browser to display a download dialog".

Let the backend do its job and don't force everything to be React always at all cost ;) And on a sidenote, only create hidden `a href` elements (or any hidden DOM element, for that matter) if it's really, _really_ necessary, you know what you're doing and you have considered all alternatives.

WcaOnRails/app/webpacker/components/Panel/Wfc/index.jsx
```diff
@@ -0,0 +1,97 @@
...
+import Errored from '../../Requests/Errored';
+import Loading from '../../Requests/Loading';
+import DelegateProbations from '../../DelegateProbations';
+import DuesExport from './DuesExport';
+
+const sections = [
+  {
+    id: 'dues-export',
+    name: 'Dues Export',
+    Component: DuesExport,
```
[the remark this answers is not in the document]
danieljames-dj: I don't know if there is any convention, but I did it just out of intuition. But thinking again, I think it is not necessary to capitalize key's first letter, so I'm changing it to match other keys.

WcaOnRails/app/webpacker/components/Panel/Wfc/index.jsx
```diff
@@ -0,0 +1,97 @@
...
+  if (loading) return <Loading />;
+  if (error) return <Errored />;
+
+  function menuClickHandler(value) {
+    if (value === 'country-bands') {
+      window.open(countryBandsUrl, '_blank');
+      return;
+    }
+    setSelectedMenu(value);
+  }
```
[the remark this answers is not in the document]
danieljames-dj: Even though currently country band opens in new tab, plan is to migrate that as well to react in later stage, so that it will come within the panel itself. Also, the panel doesn't have any home page or anything, it will just have three items in menu. So I guess it's fine to have it as menu thinking that the react migration of country_bands will come soon?

And ya I've posted the screenshot in description.

WcaOnRails/app/webpacker/lib/hooks/useRequest.jsx
danieljames-dj: Ignoring this as this file is removed now. :-)

WcaOnRails/app/webpacker/lib/hooks/useRequest.jsx
danieljames-dj: I'm feeling myself stupid after reading that comment 🤦  I'm wondering why I didn't notice this. I guess this could have been done in old `html.erb` as well. Since it was not there, this thing didn't even go through my mind.

danieljames-dj (commented):
Thanks for the review @kr-matthews and @gregorbg. I'm really sorry for the straightforward approach of downloading file. I've also added `PanelTemplate` so that we don't have to do much work to build the panels for others.

WcaOnRails/app/webpacker/components/Panel/Wfc/index.jsx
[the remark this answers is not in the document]
kr-matthews: Thanks for the screenshots, very helpful!

I have 2 issues with the country bands being a button in disguise in the menu (aka section/tab list). I think it should just be a separate button somewhere (underneath the menu?).

1. It's unintuitive as a user. All the identical-looking elements in a group should behave identically.
2. It complicates the logic of the menu and makes it harder to understand the code. And if the country band is going to later become a tab anyway, then this logic will just need to be pulled out later - might as well keep it simple from the start.

Point 2 is only partially relevant since you already have the changes in there, but keeping the code simpler in the meantime still applies, as well as point 1.

On the other hand, if it's removed then in some cases there will only be one tab in the menu, which might be odd. But we could just hide the menu in that case.

But I'd be curious to hear what other people think.

WcaOnRails/app/webpacker/components/Panel/Wfc/index.jsx
```diff
@@ -0,0 +1,45 @@
...
+    component: DelegateProbations,
+    forAtleastSeniorMember: true,
+  },
+];
+
+export default function Wfc() {
+  const { data, loading, error } = useLoadedData(panelWfcPageDataUrl);
+
+  if (loading) return <Loading />;
+  if (error) return <Errored />;
```
kr-matthews, line 36: Would it be worth assuming that they aren't a senior member if this errors, rather than showing the error component? If I'm a senior and this errors, then I can't do anything, but if it assumed I wasn't instead then I could still access country bands and other public information. (With some error message somewhere to indicate why they can't see the other stuff.)

WcaOnRails/app/webpacker/components/Panel/Wfc/index.jsx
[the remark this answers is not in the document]
danieljames-dj: Makes sense to me. But I don't know having separate link below menu looks good or not. Anyway, I've added an 'open in new tab' icon and updated the description. Can you please check if the current one is fine as it is no longer 'identical elements'?

WcaOnRails/app/webpacker/components/Panel/Wfc/index.jsx
danieljames-dj: I'm not sure whether you understood it correctly, so the use-case here is that the WFC panel will be accessible to all WFC members irrespective of whether they are senior members or not. Only difference is, if they are senior member, they will get an additional option 'modify delegate probation'.

kr-matthews: I think I understand. I'm suggesting that if there's an error, then show `PanelTemplate` and assume `data.isAtleastSeniorMember` is false (plus show some error message _somewhere_), rather than omitting the component and only showing `Errored`. This way, if there's an error, the user can still access the standard options that all WFC members see.

WcaOnRails/app/webpacker/components/Panel/Wfc/index.jsx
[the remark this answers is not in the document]
kr-matthews: Yeah that mostly resolves my concerns, thanks.

WcaOnRails/app/webpacker/components/Panel/Wfc/index.jsx
danieljames-dj: Oh okay, got it. Thanks for clarifying, sorry for not getting it initially. This looks like a good idea, I implemented this.

WcaOnRails/app/webpacker/components/Panel/Wfc/index.jsx
```diff
@@ -0,0 +1,55 @@
...
+    );
+  }
+
+  return (
+    <PanelTemplate
+      heading="WFC Panel"
+      sections={sections
+        .filter((section) => !(section.forAtleastSeniorMember) || data.isAtleastSeniorMember)}
+    />
+  );
```
kr-matthews, line 54: No worries about misunderstanding.

To save some duplication:
```suggestion
  return (
      <>
        {error && <Errored />}
        <PanelTemplate
          heading="WFC Panel"
          sections={sections
            .filter((section) => !section.forAtleastSeniorMember || (data?.isAtleastSeniorMember == true))}
        />
      </>
    );
```

danieljames-dj: That's interesting, I've committed your suggestion, thank you.

gregorbg (approved):

PR #8483, opened 2023-11-02
Update wca-states.json

Nanush7:
Bump version and move Chinese Taipei up in order to preserve the alphabetical order [here](https://www.worldcubeassociation.org/regulations/countries/) (as reported by Felipe Rojas). I am assuming that the file does not need to be ordered by ID, which may be incorrect.

The version_hash field must be changed before merging, but I guess we need to merge the relevant PR in the wca-regulations repository first.

### DIFF
--- a/WcaOnRails/config/wca-states.json
+++ b/WcaOnRails/config/wca-states.json
@@ -295,6 +295,14 @@
         },
         {
           "info": "",
+          "name": "Chinese Taipei",
+          "continent_id": "_Asia",
+          "class": "state",
+          "iso2": "TW",
+          "id": "Taiwan"
+        },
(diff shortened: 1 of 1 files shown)


gregorbg:
Has the relevant PR been merged in wca-regulations?
If No, is there a timeline?

Nanush7:
The changes will be available in the official branch once we have Board's approval, which may happen at any point during this month (hopefully in a week or two).

Nanush7:
@gregorbg Bumping. I'm not sure about the version_hash value.

In case you need it, the commit hash for the official-2024-01-01 tag is `e3964c4`.

PR #8489, opened 2023-11-03
Added regions to users table

danieljames-dj:
This PR adds new column `region_id` to users table. Now, while changing delegate status, instead of selecting the senior delegate, we can select region. There are lots of place where senior_delegate_id is used, so even if we select region, we still compute senior_delegate_id and stores it in the users table. This will be temporary, because after this migration, will replace the senior_delegate_ids with regions.

### DIFF
--- a/WcaOnRails/app/controllers/api/v0/roles_controller.rb
+++ b/WcaOnRails/app/controllers/api/v0/roles_controller.rb
@@ -22,7 +22,6 @@ class Api::V0::RolesController < Api::V0::ApiController
   def show
     user_id = params.require(:userId)
     is_active_role = ActiveRecord::Type::Boolean.new.cast(params.require(:isActiveRole))
-    senior_delegates = User.where(delegate_status: "senior_delegate")
 
     if is_active_role
       user = User.find(user_id)
@@ -37,7 +36,7 @@ class Api::V0::RolesController < Api::V0::ApiController
     else
       render json: {
         roleData: {},
-        seniorDelegates: senior_delegates,
+        regions: UserGroup.regions,
       }
     end
   end
--- a/WcaOnRails/app/controllers/api/v0/roles_controller.rb
+++ b/WcaOnRails/app/controllers/api/v0/roles_controller.rb
@@ -29,10 +29,10 @@ class Api::V0::RolesController < Api::V0::ApiController
       render json: {
         roleData: {
           delegateStatus: user.delegate_status,
-          seniorDelegateId: user.senior_delegate.id,
+          regionId: user.region_id,
           location: user.location,
         },
-        seniorDelegates: senior_delegates,
+        regions: UserGroup.regions,
       }
     else
       render json: {
@@ -45,11 +45,17 @@ class Api::V0::RolesController < Api::V0::ApiController
   def update
     user_id = params.require(:userId)
     delegate_status = params.require(:delegateStatus)
-    senior_delegate_id = params.require(:seniorDelegateId)
+    region_id = params.require(:regionId)
     location = params.require(:location)
 
     user = User.find(user_id)
-    user.update!(delegate_status: delegate_status, senior_delegate_id: senior_delegate_id, location: location)
+    if delegate_status == "senior_delegate"
+      senior_delegate_id = nil
+      User.where(region_id: region_id).where.not(id: user_id).update_all(senior_delegate_id: user_id)
+    else
+      senior_delegate_id = User.where(delegate_status: "senior_delegate", region_id: region_id).first.id
+    end
+    user.update!(delegate_status: delegate_status, senior_delegate_id: senior_delegate_id, region_id: region_id, location: location)
     send_role_change_notification(user)
 
     render json: {
@@ -61,7 +67,7 @@ class Api::V0::RolesController < Api::V0::ApiController
     user_id = params.require(:userId)
 
     user = User.find(user_id)
-    user.update!(delegate_status: '', senior_delegate_id: '', location: '')
+    user.update!(delegate_status: '', senior_delegate_id: '', region_id: '', location: '')
     send_role_change_notification(user)
 
     render json: {
--- a/WcaOnRails/app/webpacker/components/RolesTab/DelegateForm.jsx
+++ b/WcaOnRails/app/webpacker/components/RolesTab/DelegateForm.jsx
@@ -32,7 +32,7 @@ export default function DelegateForm({
         onChange={handleFormChange}
       />
       <Form.Dropdown
-        label="Region"
+        label={I18n.t('activerecord.attributes.user.region')}
         fluid
         selection
         name="regionId"
--- a/WcaOnRails/app/webpacker/components/RolesTab/DelegateForm.jsx
+++ b/WcaOnRails/app/webpacker/components/RolesTab/DelegateForm.jsx
@@ -7,11 +7,12 @@ import I18n from '../../lib/i18n';
 // i18n-tasks-use t('enums.user.delegate_status.trainee_delegate')
 // i18n-tasks-use t('enums.user.delegate_status.candidate_delegate')
 // i18n-tasks-use t('enums.user.delegate_status.delegate')
+// i18n-tasks-use t('enums.user.delegate_status.senior_delegate')
 
 export default function DelegateForm({
   formValues,
   updateFormProperty,
-  seniorDelegates,
+  regions,
   delegateStatusOptions,
 }) {
   const handleFormChange = (_, { name, value }) => updateFormProperty({ [name]: value });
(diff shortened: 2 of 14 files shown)


WcaOnRails/spec/factories/user_groups.rb
```diff
@@ -0,0 +1,60 @@
...
+      is_active { true }
+      is_hidden { false }
+    end
+
+    factory :africa_region do
+      name { "Africa" }
+      group_type { :delegate_regions }
+      is_active { true }
+      is_hidden { false }
+      parent_group_id { (FactoryBot.create :delegate_regions_user_group).id }
```
gregorbg, line 17: This parent ID is handled differently than all the other parent IDs below. Is there a reason for this?

WcaOnRails/db/migrate/20150601224750_add_region_to_user.rb
gregorbg, line 1: Why is this file being deleted?

WcaOnRails/db/migrate/20231103102744_add_region_to_user.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddRegionToUser < ActiveRecord::Migration[7.0]
+  def change
+    add_reference :users, :region, foreign_key: { to_table: :user_groups }
```
gregorbg, line 5: Can you change this so it is added after the `delegate_status` column?

WcaOnRails/db/migrate/20231103103433_add_user_groups.rb
```diff
@@ -0,0 +1,19 @@
...
+  def change
+    delegate_regions_user_group = UserGroup.create!(name: "Delegate Regions", group_type: :delegate_regions, is_active: true, is_hidden: false)
+    senior_delegates = User.where(delegate_status: "senior_delegate")
+    senior_delegates.each do |delegate|
+      region = delegate.location
+      if region.include?("(")
+        region = region[0, region.index("(")]
+      end
+      region = region.strip
+      user_group = UserGroup.create!(name: region, group_type: :delegate_regions, parent_group_id: delegate_regions_user_group.id, is_active: true, is_hidden: false)
```
gregorbg, line 13: What is the reasoning behind having a shared parent group for all Delegates?
Currently in the real-life WCA (not code), Senior Delegate is the highest rank. There is no "World Delegate" or similar, so I'm wondering why you chose to create a global Delegate user group.

WcaOnRails/db/migrate/20231103103433_add_user_groups.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class AddUserGroups < ActiveRecord::Migration[7.0]
+  def change
+    delegate_regions_user_group = UserGroup.create!(name: "Delegate Regions", group_type: :delegate_regions, is_active: true, is_hidden: false)
+    senior_delegates = User.where(delegate_status: "senior_delegate")
+    senior_delegates.each do |delegate|
```
gregorbg, line 7:
```suggestion
    User.find_each(delegate_status: "senior_delegate") do |delegate|
```

gregorbg: Context: `find_each` is a Rails ActiveRecord method that automatically iterates over rows in a DB, but limits the batch size to 1000.

Your previous approach loads _all_ rows into memory, and _then_ starts looping over them 1 by 1. In case of Senior Delegates, this is not a problem because we only have a very small number of Senior Delegates in our DB, so it won't cause memory problems. But I think in general it's important to know about `find_each` because it helps you when there's a lot of entries.

WcaOnRails/db/migrate/20231103103433_add_user_groups.rb
```diff
@@ -0,0 +1,19 @@
...
+      region = delegate.location
+      if region.include?("(")
+        region = region[0, region.index("(")]
+      end
+      region = region.strip
+      user_group = UserGroup.create!(name: region, group_type: :delegate_regions, parent_group_id: delegate_regions_user_group.id, is_active: true, is_hidden: false)
+      delegate.region_id = user_group.id
+      delegate.save!
```
gregorbg, line 15:
```suggestion
      delegate.update!(region_id: user_group_id)
```

WcaOnRails/app/webpacker/components/RolesTab/DelegateForm.jsx
```diff
@@ -30,18 +31,15 @@ export default function DelegateForm({
...
-            fluid
-            selection
-            name="seniorDelegateId"
-            value={formValues.seniorDelegateId || ''}
-            options={seniorDelegates}
-            onChange={handleFormChange}
-          />
-        )}
+      <Form.Dropdown
+        label="Region"
```
gregorbg, line 35: Properly localize, please.

WcaOnRails/app/controllers/api/v0/roles_controller.rb
```diff
@@ -29,10 +29,10 @@ def show
...
           delegateStatus: user.delegate_status,
-          seniorDelegateId: user.senior_delegate.id,
+          regionId: user.region_id,
           location: user.location,
         },
-        seniorDelegates: senior_delegates,
+        regions: UserGroup.regions,
       }
     else
       render json: {
```
gregorbg, line 38: Should this also be replaced by `regions`?

gregorbg (changes requested):
Two general remarks aside from my other comments:

1. With the introuction of the new column, is the `senior_delegate_id` column now redundant? If the answer is No, then why not?
2. The Google Workspace (formerly GSuite) mailing list sync can be updated (and hugely simplified!! :tada:) along with this change :)

WcaOnRails/app/controllers/api/v0/roles_controller.rb
danieljames-dj: Oh my bad, thanks for this. This was indeed a critical bug. Sorry for missing this out.

WcaOnRails/app/webpacker/components/RolesTab/DelegateForm.jsx
danieljames-dj: Okay done.

WcaOnRails/db/migrate/20150601224750_add_region_to_user.rb
danieljames-dj: While running the command to generate the new migration, it threw error saying that this file has to be deleted before adding the new migration as both had the same name, so did it.

WcaOnRails/db/migrate/20231103102744_add_region_to_user.rb
danieljames-dj: Done.

WcaOnRails/db/migrate/20231103103433_add_user_groups.rb
danieljames-dj: Okay got it, thanks for this. Changed.

WcaOnRails/db/migrate/20231103103433_add_user_groups.rb
danieljames-dj: I guess as per our discussion, the user_groups will have the following on highest level: Delegate probation, Delegate regions, Teams, Committees, Councils, Banned group, Translators, etc. We can have all Delegate regions at top level, but the topmost level might have lots of data. Will it be fine if we have a highest level of Delegate regions so that the highest level looks clean?

WcaOnRails/spec/factories/user_groups.rb
danieljames-dj: This was supposed to be the correct one. I've made it same in remaining as well. Thanks for pointing. Reason was that I initially (or rather copilot) wrote everything in the way like below, but it failed test case. So I tried changing like in this line.

Also, I'm using only :africa_region for test cases. I just added other regions so that it can be used somewhere in future if needed.

danieljames-dj (commented):
Thanks for the review @gregorbg. The `senior_delegate_id` column is actually not yet redundant. There are some cases and one of the use-case is Google Workspace that you pointed out. I was thinking not to make any changes to make `senior_delegate_id` redundant in this PR, instead after merging this, will make sure the migrations were successful and UI is working fine. If everything is okay, I'll create another PR in which I'll do the changes to make `senior_delegate_id` redundant and remove it from `users` table.

WcaOnRails/db/migrate/20150601224750_add_region_to_user.rb
gregorbg: Then the solution should be to give a different name to your new migration file.
Deleting an old file is simply not an option.

WcaOnRails/db/migrate/20231103103433_add_user_groups.rb
gregorbg: I'm not sure whether we're on the same page about the meaning of "groups on the highest level".

Your list is accurate and it reflects what we agreed on. So that's why we have the `group_type` column, right? I mean, you can just look at that column to figure out which groups are Delegate regions, and if we implement the Senior regions as the "highest" rank (which matches real-life WCA currently) then setting the senior region's parent_group to nil should be fine.

What do you mean by...?
> but the topmost level might have lots of data

danieljames-dj: Ha okay, sorry for making it confused. As you said, group_type was the one that decides where it belongs. So having the highest level regions with parent_id null is better. I've changed it now.

WcaOnRails/db/migrate/20150601224750_add_region_to_user.rb
danieljames-dj: Okay, I was in a thought that the migration file name should never be changed from the generated one. I've changed it now and looks like migration worked.

gregorbg (approved):

PR #8490, opened 2023-11-04
Switch API search to RO replica only during actual search

gregorbg:
The `ComputeAuxiliaryData.end_date` method does a `find_or_create` under the hood, which may result in a write operation. This is not possible if we are already in the read replica scope.


gregorbg:
Comment for code review: All I've done is nesting the read replica block further down (and add two blank lines for readability). This means that indentation changes in a lot of places and the diff looks weird at first.

In the top navbar of the review tab, there's a small gear icon where you can click "Hide Whitespace". That should make the changes much easier to grasp :)

danieljames-dj (approved):
LGTM, thanks @gregorbg.

PR #8493, opened 2023-11-06
Improved Search in React & added support for persons model

danieljames-dj:
The support for persons model is to search with WCA ID alone (irrespective of whether they have user account or not). This will help in migration of `EditPersons` to React.


WcaOnRails/app/webpacker/lib/hooks/useLoadedData.js
```diff
@@ -35,50 +33,4 @@ const useLoadedData = (url) => {
   };
 };
 
-export const useManyLoadedData = (ids, urlFn) => {
```
gregorbg, line 38: Why is this being deleted?

WcaOnRails/app/webpacker/components/DelegateProbations/index.jsx
```diff
@@ -62,7 +61,7 @@ function ProbationListTable({
 }
 
 export default function DelegateProbations() {
-  const [userId, setUserId] = React.useState();
+  const [user, setUser] = React.useState();
```
gregorbg, line 64: I am not a big fan of the search bar holding the entire `user` or `person` model. Holding the `userId` only (as was the case previously) makes it much easier to store data. Why did you decide to change this to hold the entire model instead?

gregorbg (changes requested):

WcaOnRails/app/webpacker/lib/hooks/useLoadedData.js
danieljames-dj: I deleted it because it's no longer used. It was introduced in https://github.com/thewca/worldcubeassociation.org/pull/8422.

It was used in `UserSearch`, it was used to call multiple APIs in parallel. In the new flow, that API call is not needed as per my reply to other comment, where I've explained the old flow and new flow.

WcaOnRails/app/models/person.rb
```diff
@@ -256,7 +256,7 @@ def gender_visible?
     %w(m f).include? gender
   end
 
-  def self.search(query)
+  def self.search(query, params = {})
```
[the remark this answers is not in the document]
gregorbg: Ah. In that case, you need to follow other existing models like `User` and `Competition`, which define the method as follows:

```suggestion
  def self.search(query, params: {})
```

WcaOnRails/app/webpacker/components/DelegateProbations/index.jsx
gregorbg: Okay, the flow makes sense, but you've got to check other places in the code that already use this search component (if there are any at all) and see whether they rely on ID or on the whole model.

WcaOnRails/app/models/person.rb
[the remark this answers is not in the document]
danieljames-dj: Okay done.

WcaOnRails/app/webpacker/components/DelegateProbations/index.jsx
danieljames-dj: Since this search component was introduced to React 1 week ago, I've checked through the commits in between. Also searched in the code. I confirm that there are no usage other than DelegateProbations/index.jsx.

PR #8499, opened 2023-11-10
Made WFC Panel accessible to admin team & changed filename of WFC export

danieljames-dj:


WcaOnRails/app/controllers/static_pages_controller.rb
```diff
@@ -5,7 +5,7 @@ class StaticPagesController < ApplicationController
 
   before_action :current_user_is_authorized_for_action!, only: [:panel_wfc]
   private def current_user_is_authorized_for_action!
```
gregorbg, line 7: This method name tells you nothing about the fact that this authorization is based on WFC.

gregorbg (changes requested):
I'm a bit reluctant about the complexity of the permission changes.
Usually, we go about this by introducing a `def can_do_something?` method in `user.rb` and referencing that "permission method" in a `redirect_to_root_unless_user` hook.

I think to avoid code duplication and to keep all of our permissions in one place, you should modify `user.rb` in this case as well and remove all new (direct!) WFC-related permission checks in the controller (of course, the controller still has to do the permission checks but it should do so by delegating them to the user, similar to how other permission checks are handled in this very controller already.)

WcaOnRails/app/controllers/static_pages_controller.rb
danieljames-dj: Okay changed.

danieljames-dj:
> I'm a bit reluctant about the complexity of the permission changes. Usually, we go about this by introducing a `def can_do_something?` method in `user.rb` and referencing that "permission method" in a `redirect_to_root_unless_user` hook.
> 
> I think to avoid code duplication and to keep all of our permissions in one place, you should modify `user.rb` in this case as well and remove all new (direct!) WFC-related permission checks in the controller (of course, the controller still has to do the permission checks but it should do so by delegating them to the user, similar to how other permission checks are handled in this very controller already.)

Done, I've added function in user.rb and used it to check permission.

WcaOnRails/app/models/user.rb
```diff
@@ -1251,4 +1251,8 @@ def senior_or_self
   def is_delegate_in_probation
     Role.where(user_id: self.id).where("end_date is null or end_date >= curdate()").present?
   end
+
+  def can_view_wfc_matters?
```
gregorbg, line 1255: There is already `user.can_admin_finances?` which I believe can be reused for this PR

gregorbg (changes requested):

gregorbg (approved):

PR #8500, opened 2023-11-11
UI changes in roles page

danieljames-dj:
Lists all the roles in roles page, and make the links editable only if the current_user is authorized for the same. Also removes some lines from users.js that are no longer needed.

Attached the screenshot for the same. The edit icon currently works only for delegates. For teams/committees roles, for now I've just added team edit page link to the team name. In future, once teams/committees is completely migrated to new roles structure, will expand the role form to accommodate those as well.

Screenshot 2023-11-11 at 11 25 40

### DIFF
--- a/WcaOnRails/app/models/user.rb
+++ b/WcaOnRails/app/models/user.rb
@@ -568,6 +568,10 @@ class User < ApplicationRecord
     self.current_team_members.select { |t| t.team_id == team.id }.count > 0
   end
 
+  def team_membership_details?(team)
+    self.current_team_members.find_by(team_id: team.id)
+  end
+
   def team_senior_member?(team)
     self.current_team_members.select { |t| t.team_id == team.id && t.team_senior_member }.count > 0
   end
@@ -1250,5 +1254,9 @@ class User < ApplicationRecord
 
   def is_delegate_in_probation
     Role.where(user_id: self.id).where("end_date is null or end_date >= curdate()").present?
+  end
+
+  def region
+    UserGroup.find_by_id(self.region_id)
   end
 end
--- a/WcaOnRails/app/models/user.rb
+++ b/WcaOnRails/app/models/user.rb
@@ -568,7 +568,7 @@ class User < ApplicationRecord
     self.current_team_members.select { |t| t.team_id == team.id }.count > 0
   end
 
-  def team_membership_details?(team)
+  def team_membership_details(team)
     self.current_team_members.find_by(team_id: team.id)
   end
 
--- a/WcaOnRails/app/webpacker/components/RolesTab/index.jsx
+++ b/WcaOnRails/app/webpacker/components/RolesTab/index.jsx
@@ -6,12 +6,7 @@ import {
   List,
 } from 'semantic-ui-react';
 import useLoadedData from '../../lib/hooks/useLoadedData';
-import useSaveAction from '../../lib/hooks/useSaveAction';
-import {
-  roleListUrl,
-  currentUserUrl,
-  teamUrl,
-} from '../../lib/requests/routes.js.erb';
+import { roleListUrl, teamUrl } from '../../lib/requests/routes.js.erb';
 import Errored from '../Requests/Errored';
 import Loading from '../Requests/Loading';
 import RoleForm from './RoleForm';
@@ -26,58 +21,57 @@ import I18n from '../../lib/i18n';
 // i18n-tasks-use t('enums.user.role_status.senior_member')
 // i18n-tasks-use t('enums.user.role_status.leader')
 
-export default function RolesTab({ userId }) {
-  const loggedInUserFetch = useLoadedData(currentUserUrl);
+export default function RolesTab({ userId, loggedInUserId }) {
   const roleListFetch = useLoadedData(roleListUrl(userId));
-  const { data, sync } = roleListFetch;
-  const { save, saving } = useSaveAction();
+  const loggedInUserRolesFetch = useLoadedData(roleListUrl(loggedInUserId));
 
   const [open, setOpen] = React.useState(false);
-  const [currentUserRoles, setCurrentUserRoles] = React.useState(null);
-  const [error, setError] = React.useState(null);
 
   const activeNonHiddenRoles = React.useMemo(
     () => {
-      if (data) {
-        return data.activeRoles.filter((role) => role.group.is_hidden === false);
+      if (roleListFetch.data) {
+        return roleListFetch.data.activeRoles.filter((role) => role.group.is_hidden === false);
       }
       return [];
     },
-    [data],
+    [roleListFetch.data],
   );
+
+  const loggedInUserRoles = React.useMemo(
+    () => {
+      if (loggedInUserRolesFetch.data) {
+        return loggedInUserRolesFetch.data.activeRoles.reduce((roleMap, role) => ({
+          ...roleMap,
+          [role.group.id]: role,
+        }), {});
+      }
+      return {};
+    },
+    [loggedInUserRolesFetch.data],
+  );
+
   const isDelegate = React.useMemo(
     () => {
-      if (data) {
-        return data.activeRoles.some((role) => role.group.group_type === 'delegate_regions');
+      if (roleListFetch.data) {
+        return roleListFetch.data.activeRoles.some(
+          (role) => role.group.group_type === 'delegate_regions',
+        );
       }
       return false;
     },
-    [data],
+    [roleListFetch.data],
   );
 
-  React.useEffect(() => {
-    if (!loggedInUserFetch.loading) {
-      save(roleListUrl(loggedInUserFetch.data.id), {}, (roles) => {
-        setCurrentUserRoles(roles.activeRoles.reduce((roleMap, role) => ({
-          ...roleMap,
-          [role.group.id]: role,
-        }), {}));
-      }, { method: 'GET', body: null }, setError);
-    }
-  }, [loggedInUserFetch.data, loggedInUserFetch.loading, save]);
-
   const canEditRole = (role) => role.group.group_type === 'delegate_regions' && (
-    !!currentUserRoles.admin || currentUserRoles[role.group.id].status === 'senior_delegate'
+    !!loggedInUserRoles.admin || loggedInUserRoles[role.group.id]?.status === 'senior_delegate'
   );
 
   const canEditTeamOfRole = (role) => role.group.group_type !== 'delegate_regions' && (
-    !!currentUserRoles.admin || currentUserRoles[role.group.name]?.status === 'leader'
+    !!loggedInUserRoles.admin || loggedInUserRoles[role.group.name]?.status === 'leader'
   );
 
-  if (
-    roleListFetch.loading || loggedInUserFetch.loading || saving || currentUserRoles === null
-  ) return <Loading />;
-  if (roleListFetch.error || loggedInUserFetch.error || error) return <Errored />;
+  if (roleListFetch.loading || loggedInUserRolesFetch.loading) return <Loading />;
+  if (roleListFetch.error || loggedInUserRolesFetch.error) return <Errored />;
 
   return (
     <>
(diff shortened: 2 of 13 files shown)


WcaOnRails/app/webpacker/components/RolesTab/index.jsx
```diff
@@ -6,77 +6,133 @@ import {
...
+// i18n-tasks-use t('enums.user.role_status.delegate')
+// i18n-tasks-use t('enums.user.role_status.senior_delegate')
+// i18n-tasks-use t('enums.user.role_status.member')
+// i18n-tasks-use t('enums.user.role_status.senior_member')
+// i18n-tasks-use t('enums.user.role_status.leader')
 
 export default function RolesTab({ userId }) {
+  const currentUser = useLoadedData(currentUserUrl);
   const {
     data, sync, loading, error,
```
kr-matthews, line 32: Since there are multiple fetches happening, it might be best to rename these to `userRoleData`, etc., or just name the whole const `userRoles` and access them via `roles.data`, etc.? Or maybe `profileRolesData`/`profileRoles` to try and avoid confusion between the user being edited and the current user? Currently this component a bit hard to understand based on the names of things.

WcaOnRails/app/webpacker/components/RolesTab/index.jsx
```diff
@@ -6,77 +6,133 @@ import {
...
+  React.useEffect(() => {
+    if (!currentUser.loading) {
+      save(roleListUrl(currentUser.data.id), {}, (roles) => {
+        setCurrentUserRoles(roles.activeRoles.reduce((roleMap, role) => ({
+          ...roleMap,
+          [role.group.id]: role,
+        }), {}));
+      }, { method: 'GET', body: null }, setCurrentUserFetchError);
+    }
+  }, [currentUser.data, currentUser.loading, save]);
```
kr-matthews, line 68: I'm confused why we're `save`ing things in a side effect here. I would expect the only saving to be done in `RoleForm` when the (current) user makes a change.

And why we can't get the current user roles via `useLoadedData` -- is it (just) because the url isn't known until `currentUser` finishes loading, or is there something else too?

WcaOnRails/app/webpacker/components/RolesTab/index.jsx
```diff
@@ -6,77 +6,133 @@ import {
...
+  const { save, saving } = useSaveAction();
+
+  const [open, setOpen] = React.useState(false);
+  const [currentUserRoles, setCurrentUserRoles] = React.useState(null);
+  const [currentUserFetchError, setCurrentUserFetchError] = React.useState(null);
+
+  const activeNonHiddenRoles = React.useMemo(
+    () => {
+      if (data) {
+        return data.activeRoles.filter((role) => role.group.is_hidden === false);
```
kr-matthews, line 43: I'm not familiar with hidden roles, though they sound familiar. How "hidden" are they meant to be? If they're actually secret and shouldn't be leaked, then the data endpoint should not be returning them at all right?

kr-matthews (commented):
Some of the indentation is off, could you run prettier.

WcaOnRails/app/webpacker/components/RolesTab/index.jsx
danieljames-dj: Hi @kr-matthews, we are not actually saving here, instead we are fetching the current logged in user's roles. For rendering this page, we will be needing two data: roles of the user whose page is edited, and roles of the logged-in user. But we won't be having loggedIn user's ID, that has to be fetched initially. Once we have that, we will fetch the loggedIn user's roles.

Do you have any better suggestion to handle this? I'm thinking to pass `userId` as props to this component. I didn't do that because as per one of the previous discussions in another PR, we shouldn't pass any props to components rendered from rails, unless it's something like query params. But can `userId` can be considered as an exception as the future router can probably have `loggedInUserId` as it will be very much used in the whole application? What do you think? (@gregorbg would like to get your opinion as well on this)

WcaOnRails/app/webpacker/components/RolesTab/index.jsx
danieljames-dj: Currently the data endpoint returns only delegate roles, and we are not returning any hidden roles.
I did this filter just for a double safety. In future, the endpoint will return hidden roles as well like probation, etc. I'll make sure not to send hidden roles from there once I implement that.

WcaOnRails/app/webpacker/components/RolesTab/index.jsx
danieljames-dj: I've renamed couple of them by appending `fetch` at end. Hope it's more clear now. 🙂

danieljames-dj (commented):
@kr-matthews I've actually used eslint for formatting and made sure all the files are formatted. Could you please let me know which part do you think is missing formatting so that I can check why eslint is not covering that?

WcaOnRails/app/webpacker/components/RolesTab/index.jsx
kr-matthews: I'm confused when you say that we're not saving; the inside of the `if` starts with `save(roleListUrl(...`.

I don't really know enough to comment on `userId` being an exception. But if you don't want to pass `userId` as a prop then you can still make this work cleanly. Generally, if you have `a`, and need to fetch `b` and `c`, but you need `b` to get `c`, then you can make a wrapper component. So instead of
```
function MyComponent({ a }) {
  const b = useLoadedData(...)
  
  // an effect here to fetch c once b is loaded

  if (...anything is loading...) return <Loading />
  if (...anything is errored...) return <Error />

  // actual component return, given b and c have loaded
}
```
you can do
```
function MyComponent({ a }) {
  const b = useLoadedData()

  if (b.isLoading) return <Loading />
  if (b.isError) return <Error />

  return <HelperComponent a={a} b={b} />
}

function HelperComponent({ a, b}) {
  const c = useLoadedData(...something with b...)

  if (c.isLoading) return <Loading />
  if (c.isError) return <Error />

  // actual component return, given b and c have loaded
}
```

WcaOnRails/app/webpacker/components/RolesTab/index.jsx
```diff
@@ -6,77 +6,136 @@ import {
...
+                    {!canEditTeamOfRole(role) && (
+                    <List.Header>
+                      {`${I18n.t(`enums.user.role_status.${role.status}`)}, ${role.group.name}`}
+                    </List.Header>
+                    )}
+                    {!!role.start_date && (
+                    <List.Description>
+                      {`Since ${role.start_date}`}
+                    </List.Description>
+                    )}
```
kr-matthews, line 114: The indentation seems off here, for example. The `List` should be indented more.

WcaOnRails/app/webpacker/components/RolesTab/index.jsx
```diff
@@ -6,77 +6,136 @@ import {
...
+  const activeNonHiddenRoles = React.useMemo(
+    () => {
+      if (data) {
+        return data.activeRoles.filter((role) => role.group.is_hidden === false);
+      }
+      return [];
+    },
+    [data],
+  );
+  const isDelegate = React.useMemo(
```
coder13, line 48: This doesn't need to be a useMemo.

If the return type of an expression is a primitive like a string, boolean, or number, then you don't need a useMemo.

WcaOnRails/app/webpacker/components/RolesTab/index.jsx
gregorbg: I apologize if the previous discussion about this topic caused a mis-understanding. The rule of thumb is **not** that passing props from Rails in the `react_component` is forbidden. It is only forbidden/impractical/discouraged to pass data to React components that changes / re-loads during the lifecycle of the component.

For example, say you're writing a React component to display a list of upcoming competitions, and that component has pagination. The pagination makes it so that the list of competitions (or to be precise, the current page of upcoming competitions) is determined by what the user clicks in their browser at runtime. In that case, you shouldn't pass an initial batch of competitions to the component as props, nor should you pass a complete list of _all_ competitions to the React component and let React "butcher" the pagination by artificially chunking a pre-set array. You should instead load the data because it makes sense to load the data.

In this case, the user ID of the user whose roles are being edited does never change, so it's completely and utterly fine to pass it as prop.

WcaOnRails/app/webpacker/components/RolesTab/index.jsx
```diff
@@ -6,77 +6,136 @@ import {
...
+    [data],
   );
-}
 
-export default function RolesTab({ userId }) {
-  const {
-    data, sync, loading, error,
-  } = useLoadedData(roleListUrl(userId));
+  React.useEffect(() => {
+    if (!loggedInUserFetch.loading) {
```
coder13, line 59: We shouldn't need this if check on loading. Remove this line and the loggedInUserFetch.loading and the code should work all the same.

`useEffect` only runs when the dependencies update. If those dependencies are just "save" and loggedInUserFetch.data, then this will just update when those change.

danieljames-dj: Okay thanks Cailyn. But after some optimizations, this useEffect has been removed.

WcaOnRails/app/webpacker/components/RolesTab/index.jsx
```diff
@@ -6,77 +6,136 @@ import {
...
-          sync={sync}
-        />
-      </>
-    )
+      <Modal
+        size="fullscreen"
+        onClose={() => {
+          setOpen(false);
+          sync();
+          window.location.reload(); // TODO: This is a hack to force the page to reload after
```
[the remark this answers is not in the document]
danieljames-dj: Actually not to see updates, we can see the updates without even reloading. But one other tab will be visible only if the user is delegate. I've made the comment more elaborate so that it is more easier to understand now.

WcaOnRails/app/webpacker/components/RolesTab/index.jsx
danieljames-dj: Oh okay, somehow eslint formatter didn't fix that. It didn't throw warning/error as well in both correct indentation and wrong indentation. I've anyway added more indentation now.

WcaOnRails/app/webpacker/components/RolesTab/index.jsx
danieljames-dj: Okay, got it. Thanks @gregorbg. And sorry for making it confused @kr-matthews. I didn't think of using another component, that would have been a better approach if userId was not passed as props. Since userId is passed as props now, I guess this issue is solved.

WcaOnRails/app/webpacker/components/RolesTab/index.jsx
kr-matthews: That's correct, using the `useMemo` will prevent it recalculating each re-render. But unless this array is _really_ big, or this component is being used _a lot_ of places at once, the performance is unlikely to be affected. At least, that's my understanding.

The other reason to memoize something is so that you can use it in a dependency array, but that's not applicable here because `true` is `true` and `false` is `false` (unlike `[]` which is not `[]`). Plus I don't think it's in a dependency array anyway.

WcaOnRails/app/models/user.rb
```diff
@@ -568,6 +568,10 @@ def team_member?(team)
     self.current_team_members.select { |t| t.team_id == team.id }.count > 0
   end
 
+  def team_membership_details?(team)
```
gregorbg, line 571: Nitpick: In Rails, methods that end with `?` should return a boolean by convention.
There is no equivalent convention (to my knowledge) for nullable methods, so calling this `team_membership_details` should be fine.

WcaOnRails/config/locales/en.yml
```diff
@@ -459,6 +459,14 @@ en:
         candidate_delegate: "Junior Delegate"
         delegate: "Delegate"
         senior_delegate: "Senior Delegate"
+      role_status:
```
gregorbg, line 462: Can we further group this by the type of role? Something like...

```yaml
role_status:
  delegate:
    trainee_delegate: "Trainee Delegate"
    candidate_delegate: "Junior Delegate"
    ...
  teams_committees:
    member: "Team Member"
    senior_member: "Senior Member"
    leader: "Leader"
  probation:
    ...
```

gregorbg: (note that the keys like `delegate`, `teams_committees`, etc. are provisional only for the sake of demonstration. The actual keys should match the group_type constants in the DB)

gregorbg (changes requested):

WcaOnRails/app/webpacker/components/RolesTab/index.jsx
danieljames-dj: Okay got it.

WcaOnRails/config/locales/en.yml
danieljames-dj: Okay done. For teams_committees, for now I added them under teams. They don't have an entry in group_type yet. The group_type in mind are `teams`, `committees`, `councils`.

WcaOnRails/app/models/user.rb
danieljames-dj: Done.

gregorbg (approved):

PR #8501, opened 2023-11-12
Migrate Board Panel to React

danieljames-dj:


WcaOnRails/app/views/panel/index.html.erb
```diff
@@ -93,16 +93,23 @@
     </ul>
   <% end %>
 
-  <% if current_user.can_admin_finances? %>
+  <% if current_user.financial_committee? || current_user.board_member? || current_user.admin? %>
     <h2>Other Panels</h2>
 
     <%# Other Panels will list the list of panels that have separate panel page - like WFC panel.%>
     <%# Currently, there is only WFC panel, but in the future, there might be more panels. %>
```
kr-matthews, line 100: Comment no longer accurate.

WcaOnRails/app/webpacker/lib/requests/routes.js.erb
```diff
@@ -96,10 +96,24 @@ export const endDelegateProbationUrl = `<%= CGI.unescape(Rails.application.route
 
 // dummyRoleId is used temporarily because delegates is yet to get a role id
```
kr-matthews, line 97: I think this comment needs to be moved down to line 110.

WcaOnRails/app/webpacker/components/Panel/Board/RegionManager.jsx
```diff
@@ -0,0 +1,172 @@
...
+        onClick={() => setOpen(true)}
+      />
+      )}
+      {userGroup.is_hidden && (
+      <List.Icon
+        name="eye slash"
+        link
+        onClick={() => setOpen(true)}
+      />
+      )}
```
kr-matthews, line 36:
```suggestion
        <List.Icon
          name={iconName}
          link
          onClick={() => setOpen(true)}
        />
```
with
```
const iconName = userGroup.is_hidden ? "eye slash" : "eye"
```

WcaOnRails/app/webpacker/components/Panel/Board/RegionManager.jsx
```diff
@@ -0,0 +1,172 @@
...
+  const [openNewSubregionModal, setOpenNewSubregionModal] = React.useState(false);
+  const [newRegion, setNewRegion] = React.useState(defaultRegion);
+
+  const closeModal = () => {
+    if (openNewRegionModal) {
+      setOpenNewRegionModal(false);
+    } else {
+      setOpenNewSubregionModal(false);
+    }
+  };
```
kr-matthews, line 71: Maybe
```suggestion
  const closeModal = () => {
    setOpenNewRegionModal(false);
    setOpenNewSubregionModal(false);
  };
```

WcaOnRails/app/webpacker/components/Panel/Board/RegionManager.jsx
```diff
@@ -0,0 +1,172 @@
...
+
+export default function RegionManager() {
+  const {
+    data, loading, error, sync,
+  } = useLoadedData(userGroupsUrl({
+    groupType: 'delegate_regions',
+  }));
+  const { save, saving } = useSaveAction();
+  const [openNewRegionModal, setOpenNewRegionModal] = React.useState(false);
+  const [openNewSubregionModal, setOpenNewSubregionModal] = React.useState(false);
```
kr-matthews, line 62: Since there's just a single modal reused between these, it might be simpler to track the type of modal
```
const [openModalType, setOpenModalType] = useState(null) // options: 'region' and 'subregion'
```
This reduces the number of states, and makes it clear that you can't have both modals open at once. But the current implementation is also fine, so feel free to ignore this.

WcaOnRails/app/webpacker/components/Panel/Board/RegionManager.jsx
```diff
@@ -0,0 +1,172 @@
...
+            />
+            <Form.Button
+              onClick={() => {
+                closeModal();
+                setNewRegion(defaultRegion);
+              }}
+            >
+              Cancel
+            </Form.Button>
+            <Form.Button
```
kr-matthews, line 157: Should this be disabled if the input is invalid (default parent region in particular)?

WcaOnRails/app/webpacker/components/Panel/Board/RegionManager.jsx
```diff
@@ -0,0 +1,172 @@
...
+              onClick={() => {
+                closeModal();
+                save(userGroupsUrl(), newRegion, () => sync(), { method: 'POST' });
```
kr-matthews, line 160: Is there any error handling for if `save` fails?

WcaOnRails/app/webpacker/components/Panel/Board/RegionManager.jsx
```diff
@@ -0,0 +1,172 @@
...
+                ))}
+              </List.List>
+            </List.Content>
+          </List.Item>
+        ))}
+      </List>
+      <ButtonGroup>
+        <Button onClick={() => setOpenNewRegionModal(true)}>Add new region</Button>
+        <Button onClick={() => setOpenNewSubregionModal(true)}>Add new subregion</Button>
+      </ButtonGroup>
```
kr-matthews, line 119: Maybe disabling these while `saving` would be good.

kr-matthews (changes requested):

WcaOnRails/app/views/panel/index.html.erb
danieljames-dj: Oh yes, thanks for pointing. Removed.

WcaOnRails/app/webpacker/components/Panel/Board/RegionManager.jsx
danieljames-dj: Good catch, done.

WcaOnRails/app/webpacker/components/Panel/Board/RegionManager.jsx
danieljames-dj: Done.

WcaOnRails/app/webpacker/components/Panel/Board/RegionManager.jsx
danieljames-dj: Thanks, but got overridden by previous comment :-)

WcaOnRails/app/webpacker/components/Panel/Board/RegionManager.jsx
danieljames-dj: Yes it will be good to disable in that case. I've done it now.

WcaOnRails/app/webpacker/components/Panel/Board/RegionManager.jsx
danieljames-dj: Ya it will be good to have, I've done it now.

WcaOnRails/app/webpacker/lib/requests/routes.js.erb
danieljames-dj: Oh yes, thanks for pointing.

danieljames-dj (commented):
Thanks @kr-matthews for reviewing.

danieljames-dj:
Converting this to draft because while implementing migration of WAC, I figured out that the roles listing API is bit complex and not scalable, so thought of cleaning it up first. I've raised the PR for that: https://github.com/thewca/worldcubeassociation.org/pull/8513. Will rebase this and undraft it once the other PR is merged.

danieljames-dj:
Rebased and marked it as 'ready to review'.

WcaOnRails/app/controllers/api/v0/roles_controller.rb
```diff
@@ -100,6 +116,12 @@ def index_for_group_type
     # Filter the list based on the permissions of the logged in user.
     roles = filter_roles_for_logged_in_user(roles)
 
+    # Filter the list based on the other parameters.
+    roles = filter_roles_for_parameters(
+      roles,
+      params.key?(:status) ? params.require(:status) : nil,
```
gregorbg, line 122: Your line implies that the `status` param is not necessarily required, so it doesn't make sense to use `params.require`. A simple `params[:status]` should do

WcaOnRails/app/controllers/api/v0/user_groups_controller.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+class Api::V0::UserGroupsController < Api::V0::ApiController
+  before_action :current_user_is_authorized_for_action!, only: [:update]
+  private def current_user_is_authorized_for_action!
+    unless current_user.board_member? || current_user.admin?
```
gregorbg, line 6: Use permissions method in the `user` class please

gregorbg: (further down below you're even introducing a `can_access_board_panel` method which exactly fits this check)

WcaOnRails/app/views/panel/index.html.erb
```diff
@@ -83,26 +83,22 @@
...
-
-    <ul>
-      <li>
-        <%= link_to "List of Senior Delegates", panel_seniors_path %>
-      </li>
-    </ul>
-  <% end %>
-
-  <% if current_user.can_admin_finances? %>
+  <% if current_user.financial_committee? || current_user.board_member? || current_user.admin? %>
```
gregorbg, line 86: Permissions method in `user` class, please. Does `current_user.can_admin_finances?` fit this use case?

WcaOnRails/app/views/panel/index.html.erb
```diff
@@ -83,26 +83,22 @@
...
 
-    <%# Other Panels will list the list of panels that have separate panel page - like WFC panel.%>
-    <%# Currently, there is only WFC panel, but in the future, there might be more panels. %>
+    <%# Other Panels will list the list of panels that have separate panel page.%>
 
     <ul>
-      <li>
-        <%= link_to "WFC panel", panel_wfc_path %>
-      </li>
+      <% if current_user.financial_committee? || current_user.admin? %>
```
gregorbg, line 92: permissions method in `user` class, please :)
Also check if it makes sense to use a similar permission to the other `if` statement on line 86 above (the only difference is that above, it also allows `board_member?` whereas here it only strictly allows WFC or admin) so that you can reuse code

WcaOnRails/app/views/panel/index.html.erb
```diff
@@ -83,26 +83,22 @@
...
+        <li>
+          <%= link_to "WFC panel", panel_wfc_path %>
+        </li>
+      <% end %>
+      <% if current_user.board_member? || current_user.admin? %>
```
gregorbg, line 97: permissions method in `user` class (pretty sure you can use the `can_access_board_panel` method you created above)

WcaOnRails/app/webpacker/components/Panel/Board/RegionManager.jsx
```diff
@@ -0,0 +1,170 @@
...
+      />
+      <Confirm
+        open={open}
+        content={`Are you sure you want to ${userGroup.is_active ? 'deactivate' : 'activate'} ${userGroup.name}?`}
+        onCancel={() => setOpen(false)}
+        onConfirm={() => {
+          setOpen(false);
+          save(
+            userGroupsUpdateUrl(userGroup.id),
+            { is_active: !userGroup.is_active, is_hidden: false },
```
gregorbg, line 37: Why are you assuming that a deactivated group is always un-hidden?
In other words, why do you hard-code `isHidden` to `false` when changing the "active-ness"?

WcaOnRails/app/webpacker/components/Panel/Board/RegionManager.jsx
```diff
@@ -0,0 +1,170 @@
...
+    const subRegionsMap = {};
+    data?.forEach((group) => {
+      if (group.parent_group_id) {
+        if (!subRegionsMap[group.parent_group_id]) {
+          subRegionsMap[group.parent_group_id] = [];
+        }
+        subRegionsMap[group.parent_group_id].push(group);
+      }
+    });
+    return subRegionsMap;
```
gregorbg, line 71:
```suggestion
    const subRegions = data?.filter((group) => group.parent_group_id);
    return Object.groupBy(subRegions, (group) => group.parent_group_id);
```

gregorbg: cf. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/groupBy

WcaOnRails/app/webpacker/components/Panel/Board/RegionManager.jsx
```diff
@@ -0,0 +1,170 @@
...
+            </Form.Button>
+            <Form.Button
+              disabled={
+                newRegion.name.length === 0
+                || (openModalType === 'newSubregion' && !newRegion.parent_group_id)
+              }
+              onClick={() => {
+                closeModal();
+                save(addUserGroupsUrl, newRegion, () => sync(), { method: 'POST' }, setSaveError);
+                sync();
```
gregorbg, line 159: Why are you calling `sync` in the `onSuccess` hook of the `save` method _and_ after the `save` invocation?

gregorbg: Seems like you are syncing twice

WcaOnRails/spec/controllers/panel_controller_spec.rb
```diff
@@ -13,11 +13,6 @@
       get :pending_claims_for_subordinate_delegates
       expect(response.status).to eq 200
     end
-
-    it "can't view the seniors list" do
```
gregorbg, line 17: Why are you dropping these tests without replacement?

WcaOnRails/spec/controllers/panel_controller_spec.rb
```diff
@@ -26,11 +21,6 @@
       sign_in board_member
     end
 
-    it "can view the seniors list" do
```
gregorbg, line 29: Dropping without replacement, see above.

gregorbg (changes requested):

WcaOnRails/app/views/panel/index.html.erb
danieljames-dj: I created `can_access_panel` so that I can use the same function when we have to add a panel section in "top navigation right dropdown" which will list all the panels accessible by the user and eventually getting rid of this page.

WcaOnRails/app/views/panel/index.html.erb
danieljames-dj: Done. Hope the change is fine.

WcaOnRails/app/views/panel/index.html.erb
danieljames-dj: Done.

WcaOnRails/app/webpacker/components/Panel/Board/RegionManager.jsx
danieljames-dj: Oops my bad, changed it 🙂

WcaOnRails/app/webpacker/components/Panel/Board/RegionManager.jsx
danieljames-dj: New learning again, thank you.

WcaOnRails/app/webpacker/components/Panel/Board/RegionManager.jsx
danieljames-dj: My bad, removed one of them. :-)

WcaOnRails/spec/controllers/panel_controller_spec.rb
danieljames-dj: Answered above.

WcaOnRails/spec/controllers/panel_controller_spec.rb
danieljames-dj: Actually the seniors panel (/panel/seniors) doesn't exist anymore. This was only accessible to board and now it's moved to Board panel. Sorry that I missed to remove it from routes.rb and panel_controller.rb. I've removed them now.

PR #8502, opened 2023-11-12
Switch to more robust way to find the tab from the DB in Competition Tabs controller

simonkellly:
Fixes issues mentioned in slack

### DIFF
--- a/WcaOnRails/app/controllers/competition_tabs_controller.rb
+++ b/WcaOnRails/app/controllers/competition_tabs_controller.rb
@@ -27,12 +27,12 @@ class CompetitionTabsController < ApplicationController
 
   def edit
     @competition = competition_from_params
-    @competition_tab = CompetitionTab.find(params[:id])
+    @competition_tab = CompetitionTab.find_by!(id: params[:id], competition: @competition)
   end
 
   def update
     @competition = competition_from_params
-    @competition_tab = CompetitionTab.find(params[:id])
+    @competition_tab = CompetitionTab.find_by!(id: params[:id], competition: @competition)
     if @competition_tab.update(competition_tab_params)
       flash[:success] = "Successfully updated #{@competition_tab.name} tab."
       redirect_to edit_competition_tab_path(@competition, @competition_tab)
@@ -43,12 +43,13 @@ class CompetitionTabsController < ApplicationController
 
   def destroy
     @competition = competition_from_params
-    CompetitionTab.find_by(id: params[:id])&.destroy
+    CompetitionTab.find_by(id: params[:id], competition: @competition)&.destroy
     redirect_to competition_tabs_path(@competition)
   end
 
   def reorder
-    competition_tab = CompetitionTab.find(params[:id])
+    competition = competition_from_params
+    competition_tab = CompetitionTab.find_by!(id: params[:id], competition: competition)
     competition_tab.reorder(params[:direction])
     render nothing: true
   end
--- a/WcaOnRails/app/controllers/competition_tabs_controller.rb
+++ b/WcaOnRails/app/controllers/competition_tabs_controller.rb
@@ -27,12 +27,12 @@ class CompetitionTabsController < ApplicationController
 
   def edit
     @competition = competition_from_params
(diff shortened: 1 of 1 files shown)


WcaOnRails/app/controllers/competition_tabs_controller.rb
```diff
@@ -43,12 +43,13 @@ def update
 
   def destroy
     @competition = competition_from_params
-    CompetitionTab.find_by(id: params[:id])&.destroy
+    CompetitionTab.find_by(id: params[:id], competition: @competition)&.destroy
```
dunkOnIT, line 46: Is there a reason we aren't using `find_by!` here?

dunkOnIT (changes requested):
My understanding of the change is we specify comp id as well as tab id so that it throws an error if the tab id doesn't belong to the comp id. 

If my understanding is correct, LGTM - just think we should change to `find_by!` on line 46.

WcaOnRails/app/controllers/competition_tabs_controller.rb
simonkellly: From what I looked at, having find_by! would throw an exception, and here we are already accounting for it being potentially null with the & so I thought not having it throw in this case was closer to the original implementation

dunkOnIT: Ah I see! In this case I'm partial to raising an exception as that's more consistent with our other implementations, and this is a case where an exception seems appropriate, but I can see the argument both ways.

simonkellly: Reasonable, changes made in latest commit

simonkellly:
I think the tests failures are unrelated to the changes

dunkOnIT (approved):
LGTM! Agree the test failures look unrelated, have re-run the job and will look to merge once they pass

WcaOnRails/app/controllers/competition_tabs_controller.rb
```diff
@@ -27,12 +27,12 @@ def create
 
   def edit
     @competition = competition_from_params
-    @competition_tab = CompetitionTab.find(params[:id])
+    @competition_tab = CompetitionTab.find_by!(id: params[:id], competition: @competition)
```
gregorbg, line 30:
```suggestion
    @competition_tab = @competition.tabs.find(params[:id])
```

gregorbg: `find` expects an ID as its only parameter. If you access the tabs through the `has_many :tabs` association in the `Competition` model, you can let Rails handle the DB loading via caches and stuff and don't have to care about manually retrieving the tab yourself.

WcaOnRails/app/controllers/competition_tabs_controller.rb
```diff
@@ -43,12 +43,13 @@ def update
 
   def destroy
     @competition = competition_from_params
-    CompetitionTab.find_by(id: params[:id])&.destroy
+    CompetitionTab.find_by!(id: params[:id], competition: @competition).destroy
```
gregorbg, line 46:
```suggestion
    @competition.tabs.find(params[:id]).destroy
```

WcaOnRails/app/controllers/competition_tabs_controller.rb
```diff
@@ -43,12 +43,13 @@ def update
...
     redirect_to competition_tabs_path(@competition)
   end
 
   def reorder
-    competition_tab = CompetitionTab.find(params[:id])
+    competition = competition_from_params
+    competition_tab = CompetitionTab.find_by!(id: params[:id], competition: competition)
```
gregorbg, line 52:
```suggestion
    competition_tab = competition.tabs.find(params[:id])
```

WcaOnRails/app/controllers/competition_tabs_controller.rb
simonkellly: Very nice way of handling it 😎

gregorbg (approved):

PR #8513, opened 2023-11-17
Roles listing API cleanup

danieljames-dj:


WcaOnRails/app/controllers/delegates_controller.rb
```diff
@@ -3,8 +3,8 @@
 class DelegatesController < ApplicationController
   before_action :authenticate_user!
   before_action -> { redirect_to_root_unless_user(:can_view_delegate_matters?) }, only: [:stats]
-  before_action -> { redirect_to root_url unless current_user_is_authorized? }, only: [:delegate_probation_data, :probations]
-  before_action :current_user_is_authorized_for_action!, only: [:delegate_probation_data, :start_delegate_probation, :end_delegate_probation]
+  before_action -> { redirect_to root_url unless current_user_is_authorized? }, only: [:probations]
```
gregorbg, line 6: Please clean this up similar to what you did in the WFC panel fixup PR. This includes:
- creating a permission method in the `user.rb` model, that checks the same as `current_user_is_authorized?` below.
    - This should include a check whether the user is an admin. Admins are always allowed to do anything.
    - See whether you can reuse the existing `can_view_delegate_matters` or other methods. Only rewrite permissions that don't exist yet.
- Change this invocation (the line that is commented here on GH) to `redirect_to_root_unless_user` and use the newly created method
- Remove the `private def` block below

WcaOnRails/app/models/role.rb
```diff
@@ -2,4 +2,24 @@
 
 class Role < ApplicationRecord
   belongs_to :user
+  belongs_to :group, class_name: "UserGroup"
+  belongs_to :metadata, polymorphic: true, optional: true
+
+  def group_type
```
gregorbg, line 8: Neat trick: Rails can do

`delegate :group_type, to: :group`

which you can use to completely get rid of this def

WcaOnRails/app/models/user.rb
```diff
@@ -1259,4 +1259,48 @@ def is_delegate_in_probation
   def region
     UserGroup.find_by_id(self.region_id)
   end
+
+  def can_manage_delegate_probation?
```
gregorbg, line 1263: Oh hey, this seems like it's exactly the kind of method that I wanted you to use in my other lengthy comment above.

WcaOnRails/app/models/user.rb
```diff
@@ -1259,4 +1259,48 @@ def is_delegate_in_probation
...
+    roles = []
+    self.current_teams.each do |team|
+      team_membership_details = self.team_membership_details(team)
+      if team_membership_details.leader?
+        status = 'leader'
+      elsif team_membership_details.senior_member?
+        status = 'senior_member'
+      else
+        status = 'member'
+      end
```
gregorbg, line 1288: Can you introduce constants for these?

WcaOnRails/app/controllers/api/v0/roles_controller.rb
```diff
@@ -8,53 +8,95 @@ class Api::V0::RolesController < Api::V0::ApiController
     end
   end
 
+  # Returns a list of roles based on the request parameters.
```
gregorbg, line 11: I am not a big fan of having one route that does many things based on which query params you pass in.
Is there any chance you can create separate GET endpoints for the separate "types" of data?

gregorbg (changes requested):

WcaOnRails/app/controllers/api/v0/roles_controller.rb
danieljames-dj: (I'll attend to the other messages soon, but thought of attending this immediately as I am not clear how to proceed)

Currently, `/api/v0/roles` is returning the list of roles based on parameters. One of the following three parameters must be there:
- userId
- groupId
- groupType

I'm guessing what you meant is to have separate routes for each of the above three: like `api/v0/user-id-roles`, `api/v0/group-id-roles` and `api/v0/group-type-roles`.

But I'm also confused whether by 'types' you meant `groupType` and hence we will have endpoints like: `api/v0/delegate-regions-roles`, `api/v0/teams-roles`, `api/v0/councils-roles`, etc.

I hope you meant the first one, can you please clarify this?

gregorbg: Yes, I meant the first one. When I spoke of types, I explicitly mentioned "types of data", **not** the types of groups.

My suggestion would be the following, although this is not strict and you are welcome to make any changes as you see fit
- GET `api/v0/roles/user/{userId}`
- GET `api/v0/roles/{groupId}`
- GET `api/vo/roles/type/{groupType}`

WcaOnRails/app/webpacker/lib/requests/routes.js.erb
```diff
@@ -88,17 +88,29 @@ export const adminImportResultsUrl = (competitionId) => `<%= CGI.unescape(Rails.
...
 
-export const delegateProbationDataUrl = `<%= CGI.unescape(Rails.application.routes.url_helpers.delegate_probation_data_path(format: "json")) %>`;
-
 export const startDelegateProbationUrl = `<%= CGI.unescape(Rails.application.routes.url_helpers.start_delegate_probation_path) %>`;
 
 export const endDelegateProbationUrl = `<%= CGI.unescape(Rails.application.routes.url_helpers.end_delegate_probation_path) %>`;
 
-// dummyRoleId is used temporarily because delegates is yet to get a role id
+export const rolesOfUser = (userId, params) => {
+  const searchParams = new URLSearchParams(params);
```
gregorbg, line 96: I am unhappy about this additional `URLSearchParams`. It introduces logic ("take a key-value dict of options and format them prettily into some generic information that can be appended to the URL") and this file should not contain logic. At least not in a stricter sense. It should only glue strings together, not compile params.
- Why is this `params` even necessary in the first place? In my previous review, I suggested splitting the route into three routes, which you have done. Why are you still keeping query params even after the refactor?
- The `params` parameter is currently not used at all, at least not for `rolesOfUser`.
- If it's really, _really_ necessary to have params and they cannot be "hidden" behind clever API design, then please provide an **explicit** list of query params that are hard-coded into the URL. See for example `roleDataUrl` below as a good example.

WcaOnRails/app/models/user.rb
gregorbg: Thanks!

WcaOnRails/app/models/role.rb
```diff
@@ -2,4 +2,28 @@
 
 class Role < ApplicationRecord
   belongs_to :user
+  belongs_to :group, class_name: "UserGroup"
+  belongs_to :metadata, polymorphic: true, optional: true
+
+  delegate :group_type, to: :group
+
+  enum :team_status, {
```
gregorbg, line 10: Does the `roles` table have a `team_status` column?
The `enum` construct is specifically made by **Rails** (not Ruby in general!) to enumerate string (or integer) values in a DB column. It feels like a waste to use this for regular class constants if there is no column restriction to check.

gregorbg: In case of doubt, just use `TEAM_STATUS_LEADER = "leader"` and similar as top-level class constants.

WcaOnRails/app/webpacker/lib/requests/routes.js.erb
danieljames-dj: I've removed it now. The reason why I had it was because I'll be needing them for future APIs like list of senior delegates, list of delegates, etc. Also, since there are many types of roles, I was not sure what all types of params are going to come. So I added the generic way. But as you said, the way done in `roleDataUrl` looks better, will follow that while implementing those APIs.

WcaOnRails/app/models/role.rb
danieljames-dj: Actually the team_status should belong to the `team_role_metadata` which is yet to come. For now, I've created top-level class constant in `user.rb` since this is currently used in that class.

gregorbg (approved):

PR #8514, opened 2023-11-20
Skip unstable 'Claim WCA ID' tests

gregorbg:
I have not yet been able to figure out _why_ these fail.

Instead of sinking more hours (grr...) into this, I'd rather just disable them temporarily to unblock several unrelated PRs.
If anyone can come up with an easy fix, be my guest.

The code coverage will probably drop by a few hundreths, that's okay and obviously authorized by me (as per submitting this PR)


FinnIckler:
Claim WCA ID when signed in as user without wca id can claim WCA ID should also be skipped

PR #8515, opened 2023-11-21
Miss of jwt_denylist in schema.rb

danieljames-dj:


FinnIckler:
I thought creating it through the migration was enough?

viroulep:
I got the same diff in `db/schema.rb` when running the migration: you did create a migration but you also have to commit the changes to the db schema if any.

PR #8516, opened 2023-11-21
Skip average computation for 333fm if needed

viroulep:
This prevents us (the WRT) from editing results for bo1/bo2 FM rounds; the code on WCA Live is still the same and I'm unsure how it even works (maybe it pads up to 3 results and therefore it's skipped?).


viroulep:
Thanks for the review/merge!

Did you deploy it? I can't seem to be able to create the result which motivated this PR, and the error message makes me think it's still the "old" code.

FinnIckler:
Yes it was deployed. probably needs a full rebuild because it's a js changes?

FinnIckler:
or maybe just a hard refresh on your side?

danieljames-dj:
We have tried hard refresh, didn't work.

viroulep:
I'm not sure what you mean by "full rebuild" since I have not followed the changes in deployment, but for this yes javascript assets need to be rebuilt.
I can also confirm that I did a refresh clearing the cache.

PR #8521, opened 2023-11-22
New columns in dues export for 2024 dues calculations stuff

danieljames-dj:
Added three new columns to dues export: Dues Payer Name, Dues Payer Email, Is Combined Invoice.

Currently WFC is having lots of Google sheets stuffs with lots of complicated formulas and automations. And we are almost stuck in making the flow better without help of WST.

How WFC decides whom to bill to: we have a ‘redirects’ spreadsheet where we have two sheets: one with list of countries and corresponding xero user, and another with list of organizers and corresponding xero user. Based on country or organizer, we will decide whom we have to invoice to. If both country or organizer doesn’t exist in the sheet, we will select the first delegate.

Also, from 2024 onwards, we are introducing ‘combined invoice’ which means to some trusted Regional Organizations, we will not generate invoice for every competition, instead will generate invoice once in a month with all the competitions of that month in single invoice.

Screenshots:

Screenshot 2023-11-23 at 07 20 17

Screenshot 2023-11-23 at 07 20 42

Screenshot 2023-11-23 at 07 22 13

Screenshot 2023-11-23 at 07 22 21

### DIFF
--- a/WcaOnRails/app/controllers/api/v0/wfc_dues_redirects_controller.rb
+++ b/WcaOnRails/app/controllers/api/v0/wfc_dues_redirects_controller.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+class Api::V0::WfcDuesRedirectsController < Api::V0::ApiController
+  before_action :current_user_can_admin_finances!, only: [:index, :create]
+  private def current_user_can_admin_finances!
+    unless current_user.can_admin_finances?
+      render json: {}, status: 401
+    end
+  end
+
+  def index
+    render json: WfcDuesRedirect.all
+  end
+
+  def create
+    redirect_type = params.require(:redirectType)
+    if redirect_type == WfcDuesRedirect.redirect_types[:country]
+      redirect_from_country_iso2 = params.require(:redirectFromCountryIso2)
+      redirect_from_country_id = Country.find_by_iso2(redirect_from_country_iso2).id
+    elsif redirect_type == WfcDuesRedirect.redirect_types[:organizer]
+      redirect_from_organizer_id = params.require(:redirectFromOrganizerId)
+    else
+      raise "Unknown redirect type: #{redirect_type}"
+    end
+    redirect_to_id = params.require(:redirectToId)
+    wfc_dues_redirect = WfcDuesRedirect.new(redirect_type: redirect_type, redirect_from_country_id: redirect_from_country_id, redirect_from_organizer_id: redirect_from_organizer_id, redirect_to_id: redirect_to_id)
+    if wfc_dues_redirect.save
+      render json: wfc_dues_redirect, status: :created
+    else
+      render json: wfc_dues_redirect.errors, status: :unprocessable_entity
+    end
+  end
+end
--- a/WcaOnRails/app/controllers/api/v0/wfc_dues_redirects_controller.rb
+++ b/WcaOnRails/app/controllers/api/v0/wfc_dues_redirects_controller.rb
@@ -14,16 +14,17 @@ class Api::V0::WfcDuesRedirectsController < Api::V0::ApiController
 
   def create
     redirect_type = params.require(:redirectType)
-    if redirect_type == WfcDuesRedirect.redirect_types[:country]
+    redirect_to_id = params.require(:redirectToId)
+    if redirect_type == WfcDuesRedirect.redirect_source_types[:Country]
       redirect_from_country_iso2 = params.require(:redirectFromCountryIso2)
-      redirect_from_country_id = Country.find_by_iso2(redirect_from_country_iso2).id
-    elsif redirect_type == WfcDuesRedirect.redirect_types[:organizer]
+      redirect_source = Country.find_by_iso2(redirect_from_country_iso2)
+    elsif redirect_type == WfcDuesRedirect.redirect_source_types[:User]
       redirect_from_organizer_id = params.require(:redirectFromOrganizerId)
+      redirect_source = User.find(redirect_from_organizer_id)
     else
       raise "Unknown redirect type: #{redirect_type}"
     end
-    redirect_to_id = params.require(:redirectToId)
-    wfc_dues_redirect = WfcDuesRedirect.new(redirect_type: redirect_type, redirect_from_country_id: redirect_from_country_id, redirect_from_organizer_id: redirect_from_organizer_id, redirect_to_id: redirect_to_id)
+    wfc_dues_redirect = WfcDuesRedirect.new(redirect_source: redirect_source, redirect_to_id: redirect_to_id)
     if wfc_dues_redirect.save
       render json: wfc_dues_redirect, status: :created
     else
--- a/WcaOnRails/app/models/competition.rb
+++ b/WcaOnRails/app/models/competition.rb
@@ -2011,4 +2011,28 @@ class Competition < ApplicationRecord
     dues = DuesCalculator.dues_per_competitor_in_usd(self.country_iso2, self.base_entry_fee_lowest_denomination.to_i, self.currency_code)
     dues.present? ? dues : 0
   end
+
+  def dues_payer
+    country_id = self.country.id
+    country_redirect = WfcDuesRedirect.find_by(redirect_from_country_id: country_id)
+    if country_redirect.present?
+      return WfcXeroUser.find_by(id: country_redirect.redirect_to_id)
+    end
+    organizers.each do |organizer|
+      organizer_redirect = WfcDuesRedirect.find_by(redirect_from_organizer_id: organizer.id)
+      if organizer_redirect.present?
+        return WfcXeroUser.find_by(id: organizer_redirect.redirect_to_id)
+      end
+    end
+    staff_delegates.min_by(&:name)
+  end
+
+  def dues_payer_is_combined_invoice?
+    dues_payer = self.dues_payer
+    if dues_payer.is_a?(WfcXeroUser)
+      dues_payer.is_combined_invoice
+    else
+      false
+    end
+  end
 end
--- a/WcaOnRails/app/models/competition.rb
+++ b/WcaOnRails/app/models/competition.rb
@@ -2013,28 +2013,28 @@ class Competition < ApplicationRecord
   end
 
   private def xero_dues_payer
-    if self.country.wfc_dues_redirect.present?
-      self.country.wfc_dues_redirect.redirect_to
-    elsif self.organizers.any? { |organizer| organizer.wfc_dues_redirect.present? }
-      self.organizers.find { |organizer| organizer.wfc_dues_redirect.present? }.wfc_dues_redirect.redirect_to
-    end
+    (
+      self.country&.wfc_dues_redirect&.redirect_to ||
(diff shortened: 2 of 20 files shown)


WcaOnRails/app/controllers/api/v0/wfc_dues_redirects_controller.rb
gregorbg, line 1: Why are you creating separate controllers for separate sections in the WFC panel?

I believe it would be reasonable to have one (more or less) big "WFC Panel controller".
As a bonus, this would save you from the code duplication checking permissions.

WcaOnRails/app/models/competition.rb
```diff
@@ -2011,4 +2011,28 @@ def dues_per_competitor_in_usd
     dues = DuesCalculator.dues_per_competitor_in_usd(self.country_iso2, self.base_entry_fee_lowest_denomination.to_i, self.currency_code)
     dues.present? ? dues : 0
   end
+
+  def dues_payer
```
gregorbg, line 2015: I am unhappy with the fact that this can either return a `WfcXeroUser` or a simple `User`.

Rails is a weakly typed language, which usually means that you have to pay attention to all the typing yourself (as there is no compiler that can do it for you). As a result, deliberately creating methods that can return multiple different types should be considered bad practice.

Is there any way you can redesign your approach? For example there can be a `xero_dues_payer` method that returns only `WfcXeroUser` or `nil`, and then for other checks you can verify whether that return value is present or not and act accordingly.

WcaOnRails/app/models/competition.rb
```diff
@@ -2011,4 +2011,28 @@ def dues_per_competitor_in_usd
...
+      if organizer_redirect.present?
+        return WfcXeroUser.find_by(id: organizer_redirect.redirect_to_id)
+      end
+    end
+    staff_delegates.min_by(&:name)
+  end
+
+  def dues_payer_is_combined_invoice?
+    dues_payer = self.dues_payer
+    if dues_payer.is_a?(WfcXeroUser)
```
gregorbg, line 2032: Please try to avoid polymorphic type checks, see my other lengthy comment above.

WcaOnRails/db/migrate/20231122153755_create_wfc_dues_redirects.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class CreateWfcDuesRedirects < ActiveRecord::Migration[7.0]
+  def change
+    create_table :wfc_dues_redirects do |t|
+      t.string :redirect_type, null: false
+      t.references :redirect_from_country, type: :string, foreign_key: { to_table: :Countries }
+      t.references :redirect_from_organizer, type: :integer, foreign_key: { to_table: :users }
```
gregorbg, line 8: Rails has a built-in way of handling polymorphic associations: https://guides.rubyonrails.org/association_basics.html#polymorphic-associations

You have to redesign your database model slightly, but it will give you a huge benefit in terms of usability and code convenience.

gregorbg:
```suggestion
      t.references :redirect_source, polymorphic: true
```

WcaOnRails/app/models/wfc_dues_redirect.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+class WfcDuesRedirect < ApplicationRecord
+  belongs_to :redirect_to, class_name: "WfcXeroUser", foreign_key: "redirect_to_id"
```
gregorbg, line 4: Based on my suggestion of using polymorphic types below (code comment in the migration file), you can add a relation here:

```suggestion
  belongs_to :redirect_to, class_name: "WfcXeroUser", foreign_key: "redirect_to_id"
  belongs_to :redirect_source, polymorphic: true
```

WcaOnRails/app/models/wfc_dues_redirect.rb
```diff
@@ -0,0 +1,31 @@
...
+    end
+    Country.find_by_id(redirect_from_country_id)
+  end
+
+  def redirect_from_organizer
+    if redirect_from_organizer_id.nil?
+      return nil
+    end
+    User.find(redirect_from_organizer_id)
+  end
```
gregorbg, line 23: This can be completely skipped based on my other comment in this file.

WcaOnRails/db/migrate/20231122153755_create_wfc_dues_redirects.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class CreateWfcDuesRedirects < ActiveRecord::Migration[7.0]
+  def change
+    create_table :wfc_dues_redirects do |t|
+      t.string :redirect_type, null: false
```
gregorbg, line 6: You can drop this because the polymorphic `references` below adds a `redirect_source_type` column under the hood

WcaOnRails/app/models/wfc_dues_redirect.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+class WfcDuesRedirect < ApplicationRecord
+  belongs_to :redirect_to, class_name: "WfcXeroUser", foreign_key: "redirect_to_id"
+
+  enum :redirect_type, {
```
gregorbg, line 6:
```suggestion
  enum :redirect_source_type, {
```

WcaOnRails/app/models/competition.rb
```diff
@@ -2011,4 +2011,28 @@ def dues_per_competitor_in_usd
     dues = DuesCalculator.dues_per_competitor_in_usd(self.country_iso2, self.base_entry_fee_lowest_denomination.to_i, self.currency_code)
     dues.present? ? dues : 0
   end
+
+  def dues_payer
+    country_id = self.country.id
+    country_redirect = WfcDuesRedirect.find_by(redirect_from_country_id: country_id)
+    if country_redirect.present?
+      return WfcXeroUser.find_by(id: country_redirect.redirect_to_id)
+    end
```
gregorbg, line 2020: Based on the assumption of polymorphic reference (see code comment in the migration) you need to change the `country.rb` model as follows:

```ruby
  has_one :wfc_dues_redirect, as: :redirect_source, optional: true # where :redirect_source has to match the `belongs_to` association in the WfcDuesRedirect model
```

and then you can simply do as follows...

```suggestion
  if self.country.wfc_dues_redirect.present?
    return self.country.wfc_dues_redirect.redirect_to
  end
```

WcaOnRails/app/models/competition.rb
```diff
@@ -2011,4 +2011,28 @@ def dues_per_competitor_in_usd
...
+    organizers.each do |organizer|
+      organizer_redirect = WfcDuesRedirect.find_by(redirect_from_organizer_id: organizer.id)
+      if organizer_redirect.present?
+        return WfcXeroUser.find_by(id: organizer_redirect.redirect_to_id)
+      end
```
gregorbg, line 2025: Similar to the other remark about country redirects, you can change the `user.rb` model and add a `has_one` relation there to use as follows:
Similar to the other remark about country redirects, you can change the `user.rb` model and add a `has_one` relation there to use as follows:

```suggestion
  if organizer.wfc_dues_redirect.present?
    return organizer.wfc_dues_redirect.redirect_to
  end
```

gregorbg (changes requested):
Your database model is overly complicated because you're missing a really cool Rails feature called "polymorphic associations". I have added detailed comments to explain it as best as I can, see if it helps your implementation!

WcaOnRails/app/models/competition.rb
danieljames-dj: I've done it in another way, does it make sense now?

WcaOnRails/app/models/competition.rb
danieljames-dj: Done

WcaOnRails/app/models/competition.rb
danieljames-dj: Done

WcaOnRails/app/models/competition.rb
danieljames-dj: Done

WcaOnRails/app/models/wfc_dues_redirect.rb
danieljames-dj: Done

WcaOnRails/db/migrate/20231122153755_create_wfc_dues_redirects.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class CreateWfcDuesRedirects < ActiveRecord::Migration[7.0]
+  def change
+    create_table :wfc_dues_redirects do |t|
+      t.string :redirect_source_id, polymorphic: true, null: false
+      t.string :redirect_source_type, polymorphic: true, null: false
```
gregorbg, line 7: If you manually hack this, the `polymorphic: true` option is not necessary. It is only specific to `t.references`, not to `t.string` or `t.integer`.

Have you tried `t.references :redirect_source, polymorphic: true, type: :string`?

WcaOnRails/app/models/competition.rb
```diff
@@ -2011,4 +2011,30 @@ def dues_per_competitor_in_usd
...
+    dues_payer&.email
+  end
+
+  def dues_payer_is_combined_invoice?
+    if xero_dues_payer.present?
+      xero_dues_payer.is_combined_invoice
+    else
+      false
+    end
+  end
```
gregorbg, line 2039:
```suggestion
  def dues_payer_is_combined_invoice?
    xero_dues_payer&.is_combined_invoice || false
  end
```

WcaOnRails/app/models/competition.rb
```diff
@@ -2011,4 +2011,30 @@ def dues_per_competitor_in_usd
...
+
+  def dues_payer_name
+    dues_payer = xero_dues_payer || staff_delegates.min_by(&:name)
+    dues_payer&.name
+  end
+
+  def dues_payer_email
+    dues_payer = xero_dues_payer || staff_delegates.min_by(&:name)
```
gregorbg, line 2029: This duplicates the `staff_delegates.min_by` from above.
Please consider refactoring this into a separate `private def delegate_dues_payer` email, which can then include a comment about why sorting by name makes sense (or why WFC decided that it makes sense)

WcaOnRails/app/controllers/api/v0/wfc_dues_redirects_controller.rb
gregorbg: I guess with the extended functionality it is fine to keep separate controllers. Putting them under one scope makes sense.

WcaOnRails/app/models/competition.rb
gregorbg: See my follow-up review

WcaOnRails/app/controllers/api/v0/wfc_dues_redirects_controller.rb
danieljames-dj: Done.

gregorbg (approved):

gregorbg:
Deployed and migrated.

PR #8522, opened 2023-11-23
Fix occasionally failing JobCheck tests

gregorbg:
I am not 100% sure what happened, but my theory is the following:
- The test order is randomized. So this failing test can either be run very early on or very late in our CI test suite
- The _seeds_ for tests are run at the very beginning of the test suite
- The `JobsCheck` goes for all jobs that have a timestamp older than 5 minutes. As we have quite a lot of tests, this can become a problem if the randomized test order pushes execution towards the end of the test suite, which takes over 5 minutes
- The JobsCheck finds a lot of jobs it isn't supposed to find

Solution: Definitely reset the queue timestamps before running anything inside this spec block


danieljames-dj (approved):
Wow nice

PR #8527, opened 2023-11-23
Refactor homepage SearchWidget to use refactored WcaSearchInput

gregorbg:
This was missed (intentionally??) in a recent refactor of `WcaSearchComponent`.

CC @danieljames-dj


danieljames-dj (approved):
LGTM

Yes I skipped this intentionally mainly because I was not sure if this is good enough to go into a public facing page.

Also couple of things I noticed after implementing was (which are not relevant to this PR anyway):
1. For `goToItemUrlOnClick === false` case, if an option is selected, there should be 'close' icon instead of 'search' icon in the right side.
2. This cannot be used inside `Form.Field` of semantic-ui, so if it's part of form, it can be difficult.

Maybe I'll create a github issue so that some volunteer can look into fixing these two?

gregorbg:
> Maybe I'll create a github issue so that some volunteer can look into fixing these two?

Point number 2 will be fixed "for free" as part of the Competition Form PR (where I add an integration for SemUI forms because we need it in the scope of that PR)

Point number 1 should be fine as an issue

PR #8529, opened 2023-11-24
Implementation of hash params in panels

danieljames-dj:


WcaOnRails/app/webpacker/lib/hooks/useHash.js
gregorbg, line 1: Would be worth adding a comment that this was taken from `react-use`

WcaOnRails/app/webpacker/lib/hooks/useHash.js
```diff
@@ -0,0 +1,37 @@
...
+  }
+  return null;
+}
+
+export default function useHash() {
+  const [hash, setHash] = React.useState(getHashFromBrowserUrl());
+
+  const onHashChange = React.useCallback(() => {
+    setHash(getHashFromBrowserUrl());
+  }, []);
```
gregorbg, line 19: What does this callback do? The dependencies are empty.

gregorbg (changes requested):

WcaOnRails/app/webpacker/lib/hooks/useHash.js
danieljames-dj: Done in the useHash.js.

WcaOnRails/app/webpacker/lib/hooks/useHash.js
danieljames-dj: It was already there in react-use. I guess this is to avoid the function getting recreated while re-rendering.

gregorbg: But memoizing doesn't make any sense when there is an empty dependency array.

@kr-matthews @coder13 We stole this hook from an open source library (didn't want to install several hundred KB and 7 transitive dependencies just for one hook...), can you figure out what this seemingly "empty" `useCallback` might be doing?

WcaOnRails/app/webpacker/components/Panel/PanelTemplate.jsx
```diff
@@ -3,21 +3,39 @@ import {
...
+    const selectedSectionIndex = sections.findIndex((section) => section.id === hash);
+    const selectedSection = sections[selectedSectionIndex] || sections[0];
+    if (selectedSectionIndex === -1) {
+      setHash(selectedSection.id);
+    }
+    if (selectedSection.component) {
+      return selectedSection.component;
+    }
+    window.location.href = selectedSection.link;
+    return () => null;
```
gregorbg, line 29: What does this return value do?

WcaOnRails/app/webpacker/components/Panel/PanelTemplate.jsx
```diff
@@ -3,21 +3,39 @@ import {
...
+  const SelectedComponent = React.useMemo(() => {
+    const selectedSectionIndex = sections.findIndex((section) => section.id === hash);
+    const selectedSection = sections[selectedSectionIndex] || sections[0];
+    if (selectedSectionIndex === -1) {
+      setHash(selectedSection.id);
+    }
+    if (selectedSection.component) {
+      return selectedSection.component;
+    }
+    window.location.href = selectedSection.link;
```
gregorbg, line 28: Why are you overriding the browser's location href? Shouldn't a simple `setHash` be enough?
(I thought that's the whole reason why we introduced the `useHash` hook in the first place, to avoid manually setting location hrefs?!)

WcaOnRails/app/webpacker/components/Panel/PanelTemplate.jsx
```diff
@@ -3,21 +3,39 @@ import {
...
 
-  function menuClickHandler(index) {
-    const section = sections[index];
+  const selectedMenu = React.useMemo(() => (hash ? sections.findIndex(
+    (section) => section.id === hash,
+  ) : 0), [hash, sections]);
+
+  if (selectedMenu === -1) {
+    setHash(sections[0].id);
+  }
```
gregorbg, line 17: Not a big fan of introducing a forced re-render by manually setting the hash here.

Can you just assume that "no hash" refers to the first available section and call it a day? You can start setting the hash as soon as the user actually clicks something, and otherwise just let React happily render the panel with a default selection without enforcing any hash.

WcaOnRails/app/webpacker/components/Panel/PanelTemplate.jsx
```diff
@@ -3,21 +3,39 @@ import {
...
+
+  const SelectedComponent = React.useMemo(() => {
+    const selectedSectionIndex = sections.findIndex((section) => section.id === hash);
```
gregorbg, line 20: How is this value ever different from `selectedMenu` above?

WcaOnRails/app/webpacker/components/Panel/PanelTemplate.jsx
```diff
@@ -3,21 +3,39 @@ import {
...
+    const selectedSection = sections[selectedSectionIndex] || sections[0];
+    if (selectedSectionIndex === -1) {
+      setHash(selectedSection.id);
+    }
```
gregorbg, line 24: See my remark about manually setting hash as a side-effect above.

gregorbg (changes requested):
From what I can tell, it seems like you have a lot of "duplicated logic". The code is not _exactly_ copy-paste, but the intention behind the code is identical.

At first, you search for a selected component based on hash, and assume some defaults if the section wasn't found (or none was given). Then, in the `SelectedComponent` memo block, you repeat the same with a very similar approach. Is there any chance this can be cleaned up into one unified approach?

WcaOnRails/app/webpacker/lib/hooks/useHash.js
kr-matthews: It ensures that `onHashChange` points to the same function each re-render rather than pointing to a new (identical) function. This allows it to be used in _other_ dependency arrays without triggering them on every re-render.

Moving the definition into the event listener effect below (since it's only used there) would remove the need for this callback.

WcaOnRails/app/webpacker/components/Panel/PanelTemplate.jsx
danieljames-dj: Sorry there were some duplications here, I've fixed them now.

WcaOnRails/app/webpacker/components/Panel/PanelTemplate.jsx
danieljames-dj: I feel that won't look clean because none of the components in panel page is a 'home' page. Everything is independent. So, I felt the URL should have the ID as hash. Reason is, `/panel/wfc` will open `/panel/wfc/dues-export` but if the URL doesn't say what it is for dues-redirect, but it does say for other components I feel it won't look good. Can you please consider/accept this as an exception?

WcaOnRails/app/webpacker/lib/hooks/useHash.js
gregorbg: I think keeping it outside is fine for the time being.

gregorbg (approved):

PR #8530, opened 2023-11-24
Removal of senior_delegate_id from users table

danieljames-dj:
I've removed many cases because that won't be needed once we migrate to roles completely.

### DIFF
--- a/WcaOnRails/app/models/user.rb
+++ b/WcaOnRails/app/models/user.rb
@@ -132,8 +132,6 @@ class User < ApplicationRecord
     delegate: "delegate",
     senior_delegate: "senior_delegate",
   }
-  has_many :subordinate_delegates, class_name: "User", foreign_key: "senior_delegate_id", inverse_of: :senior_delegate
-  belongs_to :senior_delegate, -> { where(delegate_status: "senior_delegate").order(:name) }, class_name: "User", optional: true, inverse_of: :subordinate_delegates
 
   validate :wca_id_is_unique_or_for_dummy_account
   def wca_id_is_unique_or_for_dummy_account
@@ -172,13 +170,6 @@ class User < ApplicationRecord
       errors.add(:dob, I18n.t('users.errors.dob_past'))
     elsif dob && dob >= 2.years.ago
       errors.add(:dob, I18n.t('users.errors.dob_recent'))
-    end
-  end
-
-  validate :cannot_demote_senior_delegate_with_subordinate_delegates
-  def cannot_demote_senior_delegate_with_subordinate_delegates
-    if delegate_status_was == "senior_delegate" && delegate_status != "senior_delegate" && !subordinate_delegates.empty?
-      errors.add(:delegate_status, I18n.t('users.errors.senior_has_delegate'))
     end
   end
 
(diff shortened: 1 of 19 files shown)


WcaOnRails/app/webpacker/components/Delegates/index.jsx
```diff
@@ -38,6 +38,12 @@ function sortedDelegates(delegates) {
 }
 
 function DelegatesOfRegion({ activeSeniorDelegate, delegates, isAdminMode }) {
+  const delegatesOfRegion = React.useMemo(() => sortedDelegates([
+    ...delegates.filter(
+      (user) => user.region_id === activeSeniorDelegate.region_id
+        && (user.delegate_status !== 'trainee_delegate' || isAdminMode),
+    ),
+  ]), [activeSeniorDelegate, delegates, isAdminMode]);
```
kr-matthews, line 46:
```suggestion
  const delegatesOfRegion = React.useMemo(() => sortedDelegates(
    delegates.filter(
      (user) => user.region_id === activeSeniorDelegate.region_id
        && (user.delegate_status !== 'trainee_delegate' || isAdminMode),
    ),
  ), [activeSeniorDelegate, delegates, isAdminMode]);
```
and renaming the helper to `sortDelegates` would be nice.

danieljames-dj: Done, thank you

danieljames-dj:
Marking this as draft as the tests are becoming increasingly complex. I'll resume this after merging of https://github.com/thewca/worldcubeassociation.org/pull/8532.

WcaOnRails/app/models/user.rb
```diff
@@ -175,13 +173,6 @@ def name_must_match_person_name
     end
   end
 
-  validate :cannot_demote_senior_delegate_with_subordinate_delegates
-  def cannot_demote_senior_delegate_with_subordinate_delegates
-    if delegate_status_was == "senior_delegate" && delegate_status != "senior_delegate" && !subordinate_delegates.empty?
-      errors.add(:delegate_status, I18n.t('users.errors.senior_has_delegate'))
-    end
-  end
-
```
gregorbg, line 184: Why are you dropping this validation? Even after this PR is merged, `delegate_status` field still exists, and `subordinate_delegates` also continues to exist (only the implementation has changed, but the accessor itself is still there). So why drop this?

WcaOnRails/spec/models/user_spec.rb
```diff
@@ -479,27 +485,6 @@
       expect(user_with_wca_id).to be_invalid_with_errors(unconfirmed_wca_id: ["cannot claim a WCA ID because you already have WCA ID #{user_with_wca_id.wca_id}"])
     end
 
-    context "when the delegate to handle WCA ID claim is demoted" do
```
gregorbg, line 482: Related to my other comment above about dropping a validation.

gregorbg (commented):
Implementation is fine, I'm just a little bit unhappy about losing a validation with no (apparent) reason

WcaOnRails/app/models/user.rb
danieljames-dj: Thing is, since all delegates were connected to a senior delegate, we cannot demote a senior delegate without mapping all the subordinate delegates to the new senior delegate. The process that is done by board to change the senior delegate was as follows:
1. Change status of new senior delegate as 'senior delegate'.
2. Go to each subordinate delegates, and change the senior delegate to the new senior delegate
3. Now old senior delegate doesn't have any subordinate delegates, so demote them.

But now since it's connected to region, demoting a senior delegate won't affect the subordinate delegates since they are not tightly connected to the senior delegate. So now even if there are subordinate delegates, they can be demoted and the new senior delegate can be added after that. To get the freedom of removing senior delegate and adding new senior delegate, we are removing this validation.

If we need to add an alternative validation, it will be to make sure demotion of senior delegate doesn't take place unless there is another senior delegate for the region. But I think it will be better to implement that once the roles migration is completed.

gregorbg: Okay, makes sense! Thanks for your thorough explanation.

gregorbg (approved):

Issue #283, opened 2015-12-31
Redirect back to /register page after logging in


jfly:
Urg... <https://github.com/plataformatec/devise/wiki/How-To:-Redirect-back-to-current-page-after-sign-in,-sign-out,-sign-up,-update> looks promising, but work. @timhabermaas, do you have any suggestions for how to give someone a sign in link that brings them back to the page they were just on?

timhabermaas:
Mhh, shouldn't this work out of the box with devise? https://github.com/plataformatec/devise/blob/280d12b81cbb15796188a6b6c00dc9658a3f6961/lib/devise/controllers/helpers.rb#L214 grabs the stored location if it exists and the location is being stored in the [`FailureApp`](https://github.com/plataformatec/devise/blob/dd5de829c944dd955e4b35074c60798c3ad4b947/lib/devise/failure_app.rb#L79). I'm not sure why this isn't working.

If you build authentication from scratch that's also pretty much how you'd do it yourself: Store a session with the current path if the user is about to be redirected to the sign in page.

jfly:
Yeah, the problem here is that I built a page that looks like this: 

![2015-12-30_1248x414_kaladin](https://cloud.githubusercontent.com/assets/277474/12060810/65c843b2-af2c-11e5-8ec6-30bfb298c03a.png "/register when not logged in")

See that "Log in" link? Before or after clicking that link, none of the code in `FailureApp` runs. I created a new route: `/register-require-sign-in` that *does* have a `before_action` that requires you to be logged in. You can see this [here](https://github.com/jfly/worldcubeassociation.org/commit/ce0ec3827eb80041d08d98a5c3e67a9aafc71f5e). Kind of hacky, but it works.

PR #8533, opened 2023-11-28
Redirect the user to the page they were on before login page

dunkOnIT:


gregorbg:
The test fails are chokes. Re-run and they will go away eventually.

gregorbg:
Also, if you enable this redirection feature globally, you can remove the "local" hack that we have for competition registrations: https://github.com/thewca/worldcubeassociation.org/issues/283

dunkOnIT:
This redirection feature is already global by default I believe (if "globally" means "not just for competition pages"?). But yeah that we be an awesome improvement :D

gregorbg:
I mean that in some parts of our website (namely, when trying to register for a competition) we already have a "user gets redirected to the page they were on before clicking 'Sign In' button" feature. Now you're (seemingly!) introducing the same feature to the rest of the codebase, so it is important to make sure they are unified in the way they handle the redirection.

dunkOnIT:
## How the old redirect worked
- **No redirect back to competition for users who had to sign up** as far as I can tell
- Users logging in were sent to `/competitions/{comp-id}/register-require-sign-in
- That goes to the `register_require_sign_in` method in registrations_controller, which is as follows: 

```ruby
  def register_require_sign_in
    redirect_to competition_register_path(@competition)
  end
```

- This works (ie it takes the user to the signup page and then redirects back to the competition), but **I don't understand how the user is routed to the sign_in page** - it seems to just redirect back to the competition_register_path (which is just `/competitions/{comp-id}/register

## Changes made to remove old redirect functionality: 
- Removed `register_require_sign_in` function in registrations_controller
- Removed `register-require-sign-in` route
- Altered `please_sign_in` function to link to normal register route
- Removed other mentions of `register_require_sign_in`

Will manually make these changes on staging and see how it works with comps from new/old systems

gregorbg:
> * Will need advice on the merge conflict - it looks like `:new` has been removed from the `redirect_unless_root_user` before_action in the registration_controller in the main branch - not sure why this is

The method `:new` was specified as "needs permissions" in an array in this controller, Even though the `:new` method was never defined there. In Rails 7.1, there was a change that makes Rails throw an error upon boot (and thus stops the website from coming up) if you try to mark routes as "authenticated" that don't exist.

TL;DR: Removing that method from the array totally makes sense.

dunkOnIT:
Ah thanks that does indeed make total sense! Have resolved the merge conflict by accepting `master`. 

This PR is ready for review when you have a moment.

gregorbg:
> Ah thanks that does indeed make total sense! Have resolved the merge conflict by accepting `master`.
> 
> This PR is ready for review when you have a moment.

I'll gladly review once the tests pass. Actually, that very "a method is listed in an array even though it doesn't exist (anymore)" error that's new in Rails 7.1 is screwing your CI suite over.
Seems like you left a `register_require_sign_in` somewhere.

dunkOnIT:
Noting that I've resolved this merge conflict:
```ruby
<<<<<<< feature/redirect-to-source-after-login
                except: [:create, :index, :psych_sheet, :psych_sheet_event, :register, :payment_completion, :load_payment_intent, :stripe_webhook, :stripe_denomination, :destroy, :update]
=======
                except: [:create, :index, :psych_sheet, :psych_sheet_event, :register, :register_require_sign_in, :payment_completion, :load_payment_intent, :stripe_webhook, :stripe_denomination, :destroy,
                         :update, :create_paypal_order, :capture_paypal_payment]
>>>>>>> master
```

By just accepting master

gregorbg (approved):

PR #8583, opened 2023-12-01
Port my-competitions to React

FinnIckler:
Ports my-competition Form to React

TODO: 
- implement registration status icons
- add new registration service registrations


WcaOnRails/app/webpacker/components/Competitions/index.jsx
```diff
@@ -0,0 +1,215 @@
...
+                  />
+                  <Popup
+                    content="Edit the Report"
+                    trigger={(
+                      <a href={competitionReportEditUrl(competition.id)}>
+                        <Icon name="edit" />
+                      </a>
+                    )}
+                  />
+                </>
```
kr-matthews, line 158: The two tables have a lot of repeated code, which is understandable. I think big chunks like this which are identical should be factored out into reusable components.

FinnIckler:
my-comps1
my-comps2
my-comps3

FinnIckler:
I am done with the monolith part. The only issue I have is this `[missing "en.layouts.navigation.my_results" translation] ` which I don't get, because that key definitely exists

gregorbg:
> I am done with the monolith part. The only issue I have is this `[missing "en.layouts.navigation.my_results" translation] ` which I don't get, because that key definitely exists

It's not only about whether the key _exists_ in the YML world, it's also a matter of whether it's being exported to the JS world. In an effort to keep bundle size low, we only export the I18N keys to the React frontend that we actually need.

Check `config/i18n.yml` to get a quick overview of what keys are being exported, and feel free to add your own. You can (and should!) use wildcards at the lowest nesting level that makes sense.

FinnIckler:
Ok. That fixed it. That should be it for the PR. I think including the wca-registration competitions should be a seperate PR, because it will include new hooks for doing requests with JWT tokens from the monolith to the microservices.

WcaOnRails/app/webpacker/components/Competitions/MyCompetitions/index.jsx
```diff
@@ -0,0 +1,310 @@
...
+  );
+}
+
+export default function MyCompetitions() {
+  const [isAccordionOpen, setIsAccordionOpen] = React.useState(false);
+  const [shouldShowRegistrationStatus, setShouldShowRegistrationStatus] = React.useState(false);
+
+  const { data: competitions, loading: competitionsLoading } = useLoadedData(myCompetitionsAPIUrl);
+  const { data: me, loading: meLoading } = useLoadedData(meAPIUrl);
+  const { data: permissions, loading: permissionsLoading } = useLoadedData(permissionsAPIUrl);
```
kr-matthews, line 260: Do we want/need any error handling here?
There's one instance where you use `competitions?.` but the rest are all `competitions.` which will crash.

kr-matthews (changes requested):

WcaOnRails/app/webpacker/components/Competitions/MyCompetitions/index.jsx
```diff
@@ -0,0 +1,310 @@
...
+          </Table.Row>
+        ))}
+      </TableBody>
+    </Table>
+  );
+}
+
+export default function MyCompetitions() {
+  const [isAccordionOpen, setIsAccordionOpen] = React.useState(false);
+  const [shouldShowRegistrationStatus, setShouldShowRegistrationStatus] = React.useState(false);
```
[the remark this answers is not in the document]
FinnIckler: yes and I agree, this was done before that discussion

FinnIckler:
Now I just need to fix the tests

WcaOnRails/app/webpacker/components/Competitions/MyCompetitions/index.jsx
```diff
@@ -0,0 +1,310 @@
...
+            {`${I18n.t('competitions.my_competitions.past_competitions')} (${competitions?.past_competitions?.length ?? 0})`}
+          </Accordion.Title>
+          <Accordion.Content active={isAccordionOpen}>
+            <PastCompetitionsTable
+              competitions={competitions?.past_competitions ?? []}
+              permissions={permissions ?? defaultPermissions}
+            />
+          </Accordion.Content>
+        </Accordion>
+        <a href={personUrl(me.user.wca_id)}>{I18n.t('layouts.navigation.my_results')}</a>
```
kr-matthews, line 290: What if `me` errors?

WcaOnRails/app/webpacker/components/Competitions/MyCompetitions/index.jsx
```diff
@@ -0,0 +1,310 @@
...
+import Loading from '../../Requests/Loading';
+import { countries } from '../../../lib/wca-data.js.erb';
+
+const defaultPermissions = {
+  can_attend_competitions: { scope: [] },
+  can_organize_competitions: { scope: [] },
+  can_administer_competitions: { scope: [] },
+};
+
+const dateRange = (startDate, endDate) => {
```
coder13, line 33: Can you move utils outside of this file into a (probably already established) utils file/directory

WcaOnRails/app/webpacker/components/Competitions/MyCompetitions/index.jsx
```diff
@@ -0,0 +1,310 @@
...
+          <Table.HeaderCell />
+          <Table.HeaderCell />
+          <Table.HeaderCell />
+        </Table.Row>
+      </TableHeader>
+
+      <TableBody>
+        {competitions.map((competition) => (
+          <Popup
+            position="top center"
```
coder13, line 154: We should really have lint errors turned on for react.

Don't forget to be adding `key` to the top most react elements in a map. Table.Row doesn't need it, popup should need it.

WcaOnRails/app/webpacker/components/Competitions/MyCompetitions/index.jsx
coder13: Seems like we need a generic "something went wrong, please refresh" page to show when errors pop up. Do we currently have such an Error Boundary for our react pages?

FinnIckler:
I'm redoing this

PR #8586, opened 2023-12-02
Speed up senior_delegate computation during transition to user_groups

gregorbg:
This does not change the underlying query, but
1. It now has an index for lightning-fast `SELECT`. Previously, when `senior_delegate` was computed via `senior_delegate_id`, it was able to use the primary key of `users` table, but with the new method it had no index anymore. So I added one.
2. ~It can cache the result. Rails associations like `belongs_to` or `has_many` can do caching under the hood, while a method like `def senior_delegate` is "just a method" that has a return value which is being re-computed every time it is called.~

### DIFF
--- a/WcaOnRails/app/models/user.rb
+++ b/WcaOnRails/app/models/user.rb
@@ -132,11 +132,8 @@ class User < ApplicationRecord
     delegate: "delegate",
     senior_delegate: "senior_delegate",
   }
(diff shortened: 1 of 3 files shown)


gregorbg:
It turns out we cannot use the Rails association hack listed under point (2), because that affects writing back to the DB as well. And we are not ready yet to abandon writing to this column.

gregorbg:
Test failure is a quirk. Merging anyways because this is a critical hotfix and it's really only adding an index anyways.

PR #8587, opened 2023-12-02
Upgrade to NodeJS v20 (LTS)

gregorbg:


gregorbg:
Works on Production, merging as "glorified dependency update" as soon as tests pass.

PR #8588, opened 2023-12-02
Upgrade to Yarn Berry (modern, maintained version of Yarn)

gregorbg:
Now that we have Node 20, we might as well use a version of our package manager that is not stuck in "only backporting critical bugfixes" mode.

`corepack` is technically still experimental, but most modern package managers (most notably `yarn`, `pnpm`) all recommend it


gregorbg:
Whoops, turns out we need to wait for Shakapacker 7.2.0 release to include Yarn v4 support.

FinnIckler (approved):
LGTM

PR #8589, opened 2023-12-03
Council Leaders UI to change leaders

danieljames-dj:
This creates a new 'Council Leaders' page in Board panel with which we can change the leadler of a council.

The rails code changes are mostly temporary, as we will have it much more simpler way once the migration to roles/groups is completed.

Attached the UI (ignore duplicate row)
Screenshot 2023-12-03 at 07 37 01
Screenshot 2023-12-03 at 07 37 24


WcaOnRails/app/webpacker/components/Panel/Board/CouncilLeaders.jsx
```diff
@@ -0,0 +1,74 @@
...
+    rolesOfGroupType(GROUP_TYPE.COUNCILS, COUNCILS_STATUS.LEADER),
+  );
+
+  const leaders = useMemo(() => {
+    const leaderMap = {};
+    councilLeadersFetch?.data?.forEach((leader) => {
+      leaderMap[leader.group.id] = leader;
+    });
+    return leaderMap;
+  }, [councilLeadersFetch.data]);
```
kr-matthews, line 23: Rather than create this mapping of group ids to leaders, could we just do a `find` when we need to get a leader from a group id? Something like this (may handle edge cases incorrectly):
```
fun getLeaderOf(council) {
  return councilLeadersFetch?.data?.find((leader) => leader.group.id = council.id
}
```
Or maybe that wouldn't be any better? I might be overthinking this one.

WcaOnRails/app/webpacker/components/Panel/Board/LeaderChangeForm.jsx
```diff
@@ -0,0 +1,110 @@
...
+  const [saving, setSaving] = useState(false);
+
+  const handleFormChange = (_, { name, value }) => setFormValues({ ...formValues, [name]: value });
+
+  const changeOldLeaderStatus = async () => new Promise((resolve) => {
+    if (!oldLeader) { // No old leader
+      resolve();
+      return;
+    }
+    if (formValues.oldLeaderStatus === oldLeaderStatusOptions[2]) { // Resign
```
kr-matthews, line 33: Relying on `oldLeaderStatusOptions[2]` to be the resign case seems prone to future bugs. An enum would be nice but I guess we don't really have those in js. I'd rather just use 'resign' here (maybe via a top-level `const RESIGN_KEY = 'resign'`?), since then the code is clearer and if the options are changed later on, there's no chance of another option accidentally carrying out the resign action.

WcaOnRails/app/webpacker/components/Panel/Board/LeaderChangeForm.jsx
```diff
@@ -0,0 +1,110 @@
...
+
+export default function LeaderChangeForm({
+  setEditLeader, syncData, group, oldLeader,
+}) {
+  const [formValues, setFormValues] = useState({
+    newLeader: null,
+    oldLeaderStatus: oldLeaderStatusOptions[0],
+  });
+  const { save } = useSaveAction();
+  const [saving, setSaving] = useState(false);
```
kr-matthews, line 24: The `useSaveAction` hook already provides `saving`, I don't think we need this state. There's the sync and the state reset after the save calls, but neither of those is `await`ed so I don't think it matters.

WcaOnRails/app/webpacker/components/Panel/Board/LeaderChangeForm.jsx
```diff
@@ -0,0 +1,110 @@
...
+        control={Form.Dropdown}
+        name="oldLeaderStatus"
+        value={formValues?.oldLeaderStatus}
+        onChange={handleFormChange}
+        options={oldLeaderStatusOptions.map((option) => ({
+          key: option,
+          text: oldLeaderStatusOptionsMap[option],
+          value: option,
+        }))}
+      />
```
kr-matthews, line 105: Probably makes sense to hide this if there is no old leader.

WcaOnRails/app/webpacker/components/Panel/Board/LeaderChangeForm.jsx
```diff
@@ -0,0 +1,110 @@
...
+      <Form.Button onClick={() => setEditLeader(null)}>Cancel</Form.Button>
+      <Form.Button type="submit">Save</Form.Button>
```
kr-matthews, line 107: Might be nice to have these side-by-side, and colour-coded (they both look grey in the screenshot).

WcaOnRails/app/webpacker/components/Panel/Board/CouncilLeaders.jsx
danieljames-dj: The reason why I created map was to reduce time complexity. In your case, if we need to get the leaders of n councils, the complexity will be O(n^2).

But the value of n will be maximum 50 I guess (currently if teams/committees are also considered, it will come to around only 12, expecting big increase in future, it can go up to only 50 I guess).

I've heard that in react or rather in frontend, we don't look much into complexity of smaller ones. I'm bit confused, what is your view in the tradeoffs between clean code and time complexity for frontend?

WcaOnRails/app/webpacker/components/Panel/Board/LeaderChangeForm.jsx
```diff
@@ -0,0 +1,110 @@
...
+import Loading from '../../Requests/Loading';
+import { COUNCILS_STATUS } from '../../../lib/helpers/groups-and-roles-constants';
+import { apiV0Urls } from '../../../lib/requests/routes.js.erb';
+
+const oldLeaderStatusOptions = ['senior_member', 'member', 'resign'];
+const oldLeaderStatusOptionsMap = {
+  senior_member: 'Senior Member',
+  member: 'Member',
+  resign: 'Resign',
+};
```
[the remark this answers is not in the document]
danieljames-dj: Actually I wanted the keys to be in the same order every time. If we use `Object.keys`, it can go in any order right?

WcaOnRails/app/webpacker/components/Panel/Board/LeaderChangeForm.jsx
danieljames-dj: Thing is, I have two actions to be made and one should happen after the other request is *completed*. If I use the `saving` of `useSaveAction`, it will have a very short time in between the requests where the `saving` will be false right?

WcaOnRails/app/webpacker/components/Panel/Board/LeaderChangeForm.jsx
danieljames-dj: Yes, but I couldn't find a way to do it without extra css. I'm pretty sure there is some way in semantic-ui for achieving it, but I couldn't find it yet. Also since this is not public facing, I thought of taking it up later in all places when I get a solution.

WcaOnRails/app/webpacker/components/Panel/Board/CouncilLeaders.jsx
kr-matthews: I was thinking about it from a clarity perspective, since I was a bit confused about what was going on when I first read this (partly because I wasn't familiar with the format of the underlying data). Since the object itself was only needed in order to get leaders of councils, it seemed clearer to have that as a function.

But looking at it again, I don't think there's a huge difference in clarity. Especially since we need to get the leader of every group for the table (I was originally only thinking about getting the leader for the editing modal). So I'd say it's fine to leave it alone and ignore my original comment.

(As far as efficiency goes, I don't think there will be any significant difference for an array/object with a few hundred items.)

WcaOnRails/app/webpacker/components/Panel/Board/LeaderChangeForm.jsx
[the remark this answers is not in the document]
danieljames-dj: That's a good idea, I've changed it.

WcaOnRails/app/webpacker/components/Panel/Board/LeaderChangeForm.jsx
danieljames-dj: I've done something similar, hope this is fine?

WcaOnRails/app/webpacker/components/Panel/Board/LeaderChangeForm.jsx
danieljames-dj: Done

kr-matthews: Nit: Don't need the `!!`.
Everything else (React-related) looks good.

danieljames-dj: Done, thanks.

WcaOnRails/app/controllers/api/v0/roles_controller.rb
```diff
@@ -41,6 +41,17 @@ class Api::V0::RolesController < Api::V0::ApiController
       end
     end
 
+    unless is_active.nil?
```
gregorbg, line 44: Nit: Rails has `if is_active.present?` which I think is a tad bit more readable because there's no negation baked into the `unless` condition

gregorbg: Also valid for the `unless status.nil?` --> `if status.present?` above

WcaOnRails/app/controllers/api/v0/roles_controller.rb
```diff
@@ -107,21 +118,67 @@ def index_for_group_type
...
     roles = Role.where(group_id: group_ids).to_a # to_a is for the same reason as in index_for_user.
 
-    # Temporary hack to support the old delegate structure, will be removed once all roles are
+    # Temporary hack to support the old system roles, will be removed once all roles are
     # migrated to the new system.
     if group_type == UserGroup.group_types[:delegate_regions]
       roles.concat(User.where.not(delegate_status: nil).map(&:delegate_role))
+    elsif group_type == UserGroup.group_types[:councils]
+      Team.all_councils.each do |council|
+        leader = TeamMember.find_by(team_id: council.id, team_leader: true, end_date: nil)
```
gregorbg, line 127: This _should_ be eqauivalent to (please double-check) `council.leader`. Or something equivalent, maybe it's called `council.current_leader` or similar. In any case, I'd be very surprised if `Team` model objects didn't have a convenience method for retrieving the leader.

WcaOnRails/app/controllers/api/v0/roles_controller.rb
```diff
@@ -107,21 +118,67 @@ def index_for_group_type
...
   end
 
+  def create
+    user_id = params.require(:userId)
+    group_id = params.require(:groupId)
+
+    if group_id.include?("_") # Temporary hack to support some old system roles, will be removed once all roles are
+      # migrated to the new system.
+      original_group_id = group_id.split("_").last
+      group_type = group_id.split("_").first
```
gregorbg, line 163: You're listing the line that reads `.last` first and the line that reads `.first` last. This is a bit confusing in terms of "code aesthetics" :)

WcaOnRails/app/controllers/api/v0/roles_controller.rb
```diff
@@ -145,36 +202,84 @@ def show
...
-    delegate_status = params.require(:delegateStatus)
-    region_id = params.require(:regionId)
-    location = params.require(:location)
+    id = params.require(:id)
 
-    user = User.find(user_id)
-    if delegate_status == "senior_delegate"
-      senior_delegate_id = nil
-      User.where(region_id: region_id).where.not(id: user_id).update_all(senior_delegate_id: user_id)
+    if id == "dummyRoleId" # dummyRoleId is temporarily reserved for delegate changes.
```
gregorbg, line 207: Even if it's only temporary, it might be worth extracting it into a constant.
If the frontend needs to know about this constant, you might even consider exposing it via `lib/wca-data.js.erb` to have a single source of truth and avoid typos in the string literal costing you 30 minutes of debugging :)

WcaOnRails/app/controllers/api/v0/roles_controller.rb
```diff
@@ -145,36 +202,84 @@ def show
...
+        User.where(region_id: region_id).where.not(id: user_id).update_all(senior_delegate_id: user_id)
+      else
+        senior_delegate_id = User.where(delegate_status: "senior_delegate", region_id: region_id).first.id
+      end
+      user.update!(delegate_status: delegate_status, senior_delegate_id: senior_delegate_id, region_id: region_id, location: location)
+      send_role_change_notification(user)
+
+      render json: {
+        success: true,
+      }
```
gregorbg, line 225: Isn't there another PR waiting and ready that drops `senior_delegate_id`? Guess that one should be merged first, and then you can rebase this to save yourself some `if`s

WcaOnRails/app/webpacker/lib/helpers/groups-and-roles-constants.js
```diff
@@ -0,0 +1,12 @@
...
+  DELEGATE_REGIONS: 'delegate_regions',
+  TEAMS_COMMITTEES: 'teams_committees',
+  COUNCILS: 'councils',
+};
+
+export const COUNCILS_STATUS = {
+  MEMBER: 'member',
+  SENIOR_MEMBER: 'senior_member',
+  LEADER: 'leader',
+};
```
gregorbg, line 12: If needed, you can pull this in from Rails. `UserGroup.types` (or is it `UserGroup.group_types`? Depends on how the Rails model `enum` is called) gives you a hash with keys and values. That can be conerted to json using `to_json` convenience method and then loaded in `wca-data.js.erb`.

Sidenote: The wca-data.js.erb file is big, but that's on purpose. Rendering these files (i.e. resolving the Ruby template calls and producing "pure" JS) is very costly and the computation cost inccreases dramatically per file, so we keep all Ruby->JS computations in one single file even if it's not clean design-wise.

WcaOnRails/app/webpacker/lib/requests/routes.js.erb
```diff
@@ -1,3 +1,9 @@
+function jsonToQueryString(json) {
+  return Object.keys(json)
+    .filter((key) => json[key] !== undefined && json[key] !== null)
+    .map((key) => `${encodeURIComponent(key)}=${encodeURIComponent(json[key])}`).join('&');
+}
+
```
gregorbg, line 6: I feel like vanilla JS `URLSearchParams` should have a solution for this. Perhaps the constructor supports directly passing a hash?

gregorbg (changes requested):

WcaOnRails/app/webpacker/lib/helpers/groups-and-roles-constants.js
danieljames-dj: Okay, done. I've also moved to `wca-data.js.erb`. Everything is good so far.

WcaOnRails/app/webpacker/lib/requests/routes.js.erb
danieljames-dj: Actually, a use-case that we want is, if the JSON is having any value as null, we shouldn't pass that value to the server. So, this code actually removes the keys which have value as null.

gregorbg: Then it's fine to remove the nulls "manually" and _then_ letting vanilla JS `URLSearchParams` do the rest.

The problem is that you're doing _everything_ manually, including encoding the param names and glueing together the `?` and `=` symbols. Let `URLSearchParams` handle as much as reasonably possible, and only add the functionality on top that really isn't there.

gregorbg (approved):

PR #8590, opened 2023-12-03
Prevent confirming competitions with closed registration

DanielEgdal:
WCAT recently had a Delegate submit a competition where registration was closed, which would have been a quite bad thing to slip through. To our surprise there was no indication of this problem, so after discussion we thought the website should just prevent such submissions to go through which this PR does.

Note, because of this validator several past test cases broke. I made sure that the existing once in the competition_controller_spec still work, though it seemed like not the best fix (I don't know if we should just set the default start date to be in the future? In that case I will need a little hint since the existing code for this seemed a bit hacky).


WcaOnRails/app/models/competition.rb
```diff
@@ -306,6 +306,17 @@ def advancement_condition_must_be_present_for_all_non_final_rounds
     end
   end
 
+  def should_registration_closing?
```
danieljames-dj, line 309: Can we have this private as it is used only in this file?

danieljames-dj (changes requested):

gregorbg:
Thanks for your submission, merging and deploying! :ship:

PR #8593, opened 2023-12-07
Cleanup of RolesTab and rename of teams to teams_committees

danieljames-dj:


WcaOnRails/app/webpacker/components/RolesTab/index.jsx
```diff
@@ -4,95 +4,78 @@ import {
...
-
-  if (roleListFetch.loading || loggedInUserRolesFetch.loading) return <Loading />;
-  if (roleListFetch.error || loggedInUserRolesFetch.error) return <Errored />;
+  if (roleListFetch.loading || roleLoading) return <Loading />;
+  if (roleListFetch.error || roleError) return <Errored />;
 
   return (
     <>
-      {activeNonHiddenRoles.length > 0
+      {roles.length > 0
```
kr-matthews, line 55:
```suggestion
      {roleListFetch.data?.length
```
and change the map just below to
```
            {roleListFetch.data.map((role => (
```
then you can get rid of the `roles` const definition.

Or, if you'd prefer to keep that definition, it doesn't need to be memoized and can just be
```
const roles = roleListFetch.data ?? []
```

danieljames-dj: Oh thanks for catching that, I changed it. I was memoizing just because there was filtering earlier and I forgot to remove it after moving the filtering to server side.

WcaOnRails/app/controllers/api/internal/v1/api_controller.rb
```diff
@@ -1,7 +1,8 @@
 # frozen_string_literal: true
 
 class Api::Internal::V1::ApiController < ApplicationController
-  prepend_before_action :validate_token
+  # DON'T APPROVE WITH THIS CHANGE
+  # prepend_before_action :validate_token
```
danieljames-dj, line 5: @FinnIckler @dunkOnIT: I like to use the permissions API for enabling or disabling edit icon in roles tab. But I can't use it directly because it requires some headers. To make it work, I commented this line. Pretty sure this is not the correct solution, could you please let me know what has to be done for me to use it in our website?

FinnIckler: just use the non internal route api/v0/users/me/permissions

danieljames-dj: Done, thank you :-)

WcaOnRails/app/webpacker/lib/helpers/user-groups-and-roles-constants.js
```diff
@@ -0,0 +1,19 @@
...
+  DELEGATE: 'delegate',
+  JUNIOR_DELEGATE: 'candidate_delegate',
+  TRAINEE_DELEGATE: 'trainee_delegate',
+};
+
+export const TEAMS_COMMITTEES_STATUS = {
+  MEMBER: 'member',
+  SENIOR_MEMBER: 'senior_member',
+  LEADER: 'leader',
+};
```
gregorbg, line 19: See my remark about exporting these enums from Rails via `wca-data.js.erb` in another PR.

WcaOnRails/app/webpacker/lib/hooks/useLoggedInUserPermissions.js
```diff
@@ -0,0 +1,31 @@
...
+            default:
+              return false;
+          }
+        },
+      };
+    }
+    return {};
+  }, [data]);
+
+  return [role, loading, error];
```
gregorbg, line 30: Can you return these as named variables, like the way `useLoadedData` does?

WcaOnRails/app/webpacker/lib/requests/routes.js.erb
```diff
@@ -1,3 +1,9 @@
+function jsonToQueryString(json) {
+  return Object.keys(json)
+    .filter((key) => json[key] !== undefined && json[key] !== null)
+    .map((key) => `${encodeURIComponent(key)}=${encodeURIComponent(json[key])}`).join('&');
+}
```
gregorbg, line 5: See my remark about vanilla JS in another PR

gregorbg (changes requested):

WcaOnRails/app/webpacker/lib/requests/routes.js.erb
danieljames-dj: (Copying my response in the other PR)
Actually, a use-case that we want is, if the JSON is having any value as null, we shouldn't pass that value to the server. So, this code actually removes the keys which have value as null.

WcaOnRails/app/webpacker/lib/helpers/user-groups-and-roles-constants.js
danieljames-dj: Done

WcaOnRails/app/webpacker/lib/hooks/useLoggedInUserPermissions.js
danieljames-dj: Done

WcaOnRails/app/webpacker/lib/hooks/useLoggedInUserPermissions.js
```diff
@@ -14,6 +15,19 @@ export default function useLoggedInUserPermissions() {
 
   const loggedInUserPermissions = React.useMemo(() => ({
     canViewDelegateAdminPage: () => Boolean(data?.can_view_delegate_admin_page.scope === '*'),
+    canEditRole: (_role) => {
```
gregorbg, line 18: What is the underscore doing here?

danieljames-dj: Oops, actually there was a `role` outside this function, but it got cleaned up during the course of rebasing 😅 Sorry for it, removed the `_` now.

gregorbg (approved):

PR #8612, opened 2023-12-12
Migrate React-Leaflet to v4 and React to v18

gregorbg:
Basically a glorified dependency update to React 18.

Also need to migrate React-Leaflet in the same PR because our old version only supported 16 and the newer versions only support 18. There is no middle ground that supports both 16 and 18.


gregorbg:

PR #8617, opened 2023-12-12
Add competitor info route

FinnIckler:
This route allows organizers to get the PII data for their competitors. This is needed for the new registration frontend

 Should we use this opportunity to create some kind of PII logging?

This is best merged after payment-routes because I am using the `EnvConfig.WCA_REGISTRATIONS_URL` environment variable from that PR.


gregorbg:
> Should we use this opportunity to create some kind of PII logging?

What do you mean? Logging the PII itself is a horrible, _horrible_ idea that could get us sued. Logging that somebody accessed PII kinda makes sense, but we're very restrictive about this anyways.

FinnIckler:
I'm obviously talking about logging who accessed PII data...

gregorbg (commented):
I am debating with myself whether the introduction of this route under our (semi-)public v0 API namespace is justified.

The key problem is that you're introducing a route that _looks_ generic from the outside (aka. "request any competitor, get their info as you requested") but instead the route has a "secret agenda" of servicing the Reg microservice. Competitions that don't use V2 registration cannot even make any use of this endpoint.

For external users calling this API (it is public in its current state!) it doesn't make a whole lot of sense when you should use this, and how. Have you considered moving it to the v1/internal namespace?

FinnIckler:
That is a good point, the issue is that /internal/v1 routes should always be protected by a Vault check in my opinion. This route should be called by the frontend though and is protected by a 'normal' monolith permissions check. I totally agree this route needs to also check if v2 is even activated which what I will add now.

FinnIckler:
Ok, I think a better solution is this:
- We just use an internal route and the Registration service fetches the PII data and attaches into to the registration objects if the user is authenticated

FinnIckler:
Changed it to an internal route

gregorbg (approved):

PR #8620, opened 2023-12-13
Add Preferred Events and seperate Personal Recods Route

FinnIckler:
This is needed for the Registration Frontend.

I moved the personal records back out of the `me` route because I started putting preferred events into it, but then I felt like it's getting bloated. I also moved the require_user! method out of the `competition_controller` into the `api_controller` because we now need it everywhere. 
Technically you can always get personal records using the `persons/:wca_id/` route, but I feel like we should keep that separate. Thoughts?

### DIFF
--- a/WcaOnRails/app/controllers/api/v0/users_controller.rb
+++ b/WcaOnRails/app/controllers/api/v0/users_controller.rb
@@ -3,7 +3,10 @@
 class Api::V0::UsersController < Api::V0::ApiController
   def show_me
     require_user!
-    show_user(current_user)
+    if stale?(current_user)
+      # Also include the users current prs so we can handle qualifications on the Frontend
+      show_user(current_user, show_rankings: true)
+    end
   end
 
   def show_user_by_id
@@ -32,13 +35,13 @@ class Api::V0::UsersController < Api::V0::ApiController
   def personal_records
     require_user!
     return render json: { single: [], average: [] } unless current_user.wca_id.present?
-    person = Person.includes(:user, :ranksSingle, :ranksAverage).find_by_wca_id!(current_user.wca_id)
+    person = Person.includes(:ranksSingle, :ranksAverage).find_by_wca_id!(current_user.wca_id)
     render json: { single: person.ranksSingle, average: person.ranksAverage }
   end
 
   def preferred_events
     require_user!
-    preferred_events = Rails.cache.fetch("#{current_user.id}-preferred", expires_in: 24.hours) do
+    preferred_events = Rails.cache.fetch("#{current_user.id}-preferred-events", expires_in: 24.hours) do
       current_user.preferred_events.pluck(:id)
     end
     render json: preferred_events
@@ -46,7 +49,7 @@ class Api::V0::UsersController < Api::V0::ApiController
 
   def bookmarked_competitions
     require_user!
-    bookmarked_competitions = Rails.cache.fetch("#{current_user.id}-bookmarked", expires_in: 60.minutes) do
+    bookmarked_competitions = Rails.cache.fetch("#{current_user.id}-competitions-bookmarked", expires_in: 60.minutes) do
       current_user.competitions_bookmarked.pluck(:competition_id)
     end
     render json: bookmarked_competitions
@@ -59,7 +62,7 @@ class Api::V0::UsersController < Api::V0::ApiController
 
   private
 
-    def show_user(user)
+    def show_user(user, show_rankings: false)
       if user
         json = { user: user }
         if params[:upcoming_competitions]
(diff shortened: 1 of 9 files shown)


dunkOnIT:
Personally I favour having more specific routes than bloated general ones. I'd say we add a dedicated `/results/me` or `/results/{wca-id}` route that only returns what is currently in the `rankings` key of `/users/me`

WcaOnRails/app/controllers/api/v0/users_controller.rb
```diff
@@ -2,14 +2,8 @@
...
-    if current_user
-      if stale?(current_user)
-        # Also include the users current prs so we can handle qualifications on the Frontend
-        show_user(current_user, show_rankings: true)
-      end
-    else
-      render status: :unauthorized, json: { error: I18n.t('api.login_message') }
-    end
+    require_user!
+    show_user(current_user)
```
gregorbg, line 6: Where did the `stale?` check go?

WcaOnRails/app/controllers/api/v0/users_controller.rb
```diff
@@ -29,37 +23,43 @@ def show_user_by_wca_id
...
+    require_user!
+    if stale?(current_user)
+      render json: current_user.permissions
     end
   end
 
+  def personal_records
+    require_user!
+    return render json: { single: [], average: [] } unless current_user.wca_id.present?
+    person = Person.includes(:user, :ranksSingle, :ranksAverage).find_by_wca_id!(current_user.wca_id)
```
gregorbg, line 35: Where does the `:user` in the includes come from? As far as I can tell, `ranksSingle` and `ranksAverage` should be enough to compute the return value below.

WcaOnRails/app/controllers/api/v0/users_controller.rb
```diff
@@ -29,37 +23,43 @@ def show_user_by_wca_id
...
+    render json: { single: person.ranksSingle, average: person.ranksAverage }
+  end
+
+  def preferred_events
+    require_user!
+    preferred_events = Rails.cache.fetch("#{current_user.id}-preferred", expires_in: 24.hours) do
```
gregorbg, line 41: Rails cache key doesn't say what is being preferred. Preferred... Delegates? Preferred... home cooking recipes?

WcaOnRails/app/controllers/api/v0/users_controller.rb
```diff
@@ -29,37 +23,43 @@ def show_user_by_wca_id
...
   def bookmarked_competitions
-    if current_user
-      bookmarked_competitions = Rails.cache.fetch("#{current_user.id}-bookmarked", expires_in: 60.minutes) do
-        current_user.competitions_bookmarked.pluck(:competition_id)
-      end
-      render json: bookmarked_competitions
-    else
-      render status: :unauthorized, json: { error: I18n.t('api.login_message') }
+    require_user!
+    bookmarked_competitions = Rails.cache.fetch("#{current_user.id}-bookmarked", expires_in: 60.minutes) do
```
gregorbg, line 49: Again, please specify _what_ is being bookmarked

WcaOnRails/app/controllers/api/v0/users_controller.rb
```diff
@@ -29,37 +23,43 @@ def show_user_by_wca_id
...
-      render status: :unauthorized, json: { error: I18n.t('api.login_message') }
-    end
+    require_user!
+    render json: { status: "ok" }
   end
 
   private
 
-    def show_user(user, show_rankings: false)
+    def show_user(user)
```
gregorbg, line 62: Please do not remove the `show_rankings` option to maintain backwards compatibility.

WcaOnRails/spec/controllers/api_users_controller_spec.rb
```diff
@@ -78,7 +78,6 @@
       expect(response.status).to eq 200
       json = JSON.parse(response.body)
       expect(json["user"]).to eq competed_user.as_json
-      expect(json.key?("rankings")).to eq true
```
gregorbg, line 81: This test case should not disappear. We might have external API users that rely on it.

gregorbg (changes requested):

WcaOnRails/app/controllers/api/v0/users_controller.rb
FinnIckler: lost in the copy paste

WcaOnRails/app/controllers/api/v0/users_controller.rb
FinnIckler: I added this like 2 mohts ago so I don't expect anyone to actually use it, but I'll keep it in

WcaOnRails/app/controllers/api/v0/users_controller.rb
FinnIckler: it comes from the way it's done in other places of the code where the ranksSingle and ranksAverage are queried, but I guess it's not needed here

gregorbg (approved):

PR #8623, opened 2023-12-15
Delegate page migration to roles model API

danieljames-dj:


WcaOnRails/app/webpacker/lib/hooks/useLoggedInUserPermissions.js
```diff
@@ -0,0 +1,18 @@
...
+import useLoadedData from './useLoadedData';
+import { permissionsUrl } from '../requests/routes.js.erb';
+
+export default function useLoggedInUserPermissions() {
+  const { data, loading, error } = useLoadedData(permissionsUrl);
+
+  const loggedInUserPermissions = React.useMemo(() => {
+    if (data) {
+      return {
+        canViewDelegateAdminPage: () => data.can_view_delegate_admin_page.scope === '*',
```
kr-matthews, line 11: `canViewDelegateAdminPage` is being defined as a function here, but you seem to be using it as a constant in `Delegates`?
```suggestion
        canViewDelegateAdminPage: data.can_view_delegate_admin_page.scope === '*',
```

danieljames-dj: Thanks for your comment @kr-matthews. At this point, I discovered that I didn't handle the case where the app is opened without logging in. I've fixed that issue now. Also, I added `canViewDelegateAdminPage` as function because that is how we will be having other functions in this hook (like in https://github.com/thewca/worldcubeassociation.org/pull/8593).

gregorbg (approved):

PR #8624, opened 2023-12-15
added event reg limit and comeptition series ids to competition api

dunkOnIT:
This adds two fields from the Competition model needed for registration validation to `/api/v0/comeptitions/{comp-id}`

Also reordered the fields in `api/competitions_controller#show` to take up more of the line before going to the next line

### DIFF
--- a/WcaOnRails/app/models/competition.rb
+++ b/WcaOnRails/app/models/competition.rb
@@ -1643,7 +1643,7 @@ class Competition < ApplicationRecord
   end
 
   def competition_series_ids
-    competition_series&.competition_ids
+    competition_series&.competition_ids || []
   end
 
   def persons_wcif(authorized: false)
--- a/WcaOnRails/app/models/competition.rb
+++ b/WcaOnRails/app/models/competition.rb
@@ -1642,6 +1642,10 @@ class Competition < ApplicationRecord
     competition_series&.to_wcif(authorized: authorized)
   end
 
+  def competition_series_ids
+    competition_series&.competition_ids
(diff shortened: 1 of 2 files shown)


WcaOnRails/app/models/competition.rb
```diff
@@ -1642,6 +1642,10 @@ def competition_series_wcif(authorized: false)
     competition_series&.to_wcif(authorized: authorized)
   end
 
+  def competition_series_ids
+    competition_series&.competition_ids
```
gregorbg, line 1646: Not a blocker, just a genuine question: Would it make sense to return something like `competition_series&.competition_ids || []` to make sure it's always an array?

FinnIckler: would basically save a null check on the other side

dunkOnIT: Oooh yes I like that a lot more, will make the change!

gregorbg: @dunkOnIT Kindly ping :)

FinnIckler: I pushed to it as Duncan is currently traveling and I was doing monolith work anyway

gregorbg (approved):

WcaOnRails/app/models/competition.rb
dunkOnIT: Thanks Finn! And sorry yeah. it's been a juggling act with the travelling :P

PR #8628, opened 2023-12-19
Allow POST request for show_users_by_id

FinnIckler:
This allows to query more than 500 users at once without going over the GET request size limit. This is an issue currently with competitions that have more than 500 registrations in the new registration service. Another solution would be to start batching the requests if that would be better?


gregorbg:
Not a fan of this workaround. Requesting 500 persons at once in a frontend-capable way hardly makes sense.
What about changing the frontend, so that the table showing the registrations is paginated in the first place?

PR #8631, opened 2023-12-19
Update link to results export in privacy page

AxelFl:


gregorbg:
Thanks for noticing and fixing!

Issue #8629, opened 2023-12-19
Competition ID transliterates differently depending on local


DanielEgdal:
When you update an unannounced ompetition which contains a non latin character (e.g., Å,Ø,Æ,Ö) in the name, the WCA website will change the ID based on which local the user has (with changes done in the organiser view) For example, "Dåstrup" will be converted to "Dastrup" when the local is English, and "Daastrup" when the local is Danish. 
This gives inconsistent behaviour on the WCA website. WCAT has already experienced one issue of this, but did not realise how this had occurred.

I assume the cause is from ActiveSupport::Inflector.transliterate in the models/competition file: https://github.com/thewca/worldcubeassociation.org/blob/master/WcaOnRails/app/models/competition.rb#L661

DanielEgdal:
[Screencast from 12-19-2023 06:20:11 PM.webm](https://github.com/thewca/worldcubeassociation.org/assets/28397408/efbd853b-8213-4a04-9509-feda4728a753)

PR #8632, opened 2023-12-19
Forced transliteration of Competition ID to use English

LarsJohanFolde:
Added a parameter to force transliteration of the Competition ID to use English, so it returns the correct output regardless of the locale the website is set to. See issue [#8629](https://github.com/thewca/worldcubeassociation.org/issues/8629)


gregorbg:
Thanks for the prompt fix!

PR #8635, opened 2023-12-20
Changed regional organization path for patching changes

dunkOnIT:
This allows the submission of the edit regional organization form - the previous path generated an error. Tested locally and it no longer generates the error.

Thanks @danieljames-dj for spotting the 1-word fix when I was massively overcomplicating things :P

### DIFF
--- a/WcaOnRails/app/views/regional_organizations/_regional_organization_form.html.erb
+++ b/WcaOnRails/app/views/regional_organizations/_regional_organization_form.html.erb
@@ -1,4 +1,4 @@
-<% url = @regional_organization.new_record? ? new_regional_organization_path : edit_regional_organization_path(@regional_organization) %>
+<% url = @regional_organization.new_record? ? new_regional_organization_path : regional_organization_path(@regional_organization) %>
 <%= simple_form_for @regional_organization, url: url, html: { class: 'form-horizontal' } do |f| %>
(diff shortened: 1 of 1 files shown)


danieljames-dj (approved):
LGTM

WcaOnRails/app/views/regional_organizations/_regional_organization_form.html.erb
```diff
@@ -1,4 +1,4 @@
-<% url = @regional_organization.new_record? ? new_regional_organization_path : edit_regional_organization_path(@regional_organization) %>
+<% url = @regional_organization.new_record? ? new_regional_organization_path : regional_organization_path(@regional_organization) %>
```
gregorbg, line 1: I think `new_regional_organization_path` also doesn't exist anymore after the Rails 7.2 migration (verify by logging in as admin and clicking "create new" in the "Manage Regional Organizations" panel)

Should probably be replaced by `regional_organizations_path`.

gregorbg (changes requested):

WcaOnRails/app/views/regional_organizations/_regional_organization_form.html.erb
dunkOnIT: Good catch, just pushed the update!

gregorbg (approved):

PR #8636, opened 2023-12-20
Rename roles to user_roles

danieljames-dj:


gregorbg (changes requested):
The `schema.rb` contains a lot of unrelated changes (council metadata table, senior_delegate_id is still present, ...)

This is normal when working on multiple feature branches at once that all have their own independent migrations. Please manually go through the diff and correct it, there is no smarter way as far as I'm aware.

danieljames-dj:
> The `schema.rb` contains a lot of unrelated changes (council metadata table, senior_delegate_id is still present, ...)
> 
> This is normal when working on multiple feature branches at once that all have their own independent migrations. Please manually go through the diff and correct it, there is no smarter way as far as I'm aware.

Oops sorry for that. I've fixed it now.

WcaOnRails/config/routes.rb
```diff
@@ -348,11 +348,11 @@
       end
       get '/records' => "api#records"
 
-      resources :roles, only: [:create, :show, :update, :destroy]
+      resources :user_roles, only: [:create, :show, :update, :destroy]
       scope 'roles' do
```
gregorbg, line 352: Shouldn't this be migrated to `scope user_roles` as well?

gregorbg: On that note: You can open the `do` block directly after the `resources`:

```ruby
resources :user_roles, only: [...] do
  get '/sub-route'
end
```

gregorbg (commented):
Almost there, just one question regarding routes

WcaOnRails/config/routes.rb
danieljames-dj: > Shouldn't this be migrated to `scope user_roles` as well?

Oh yes, thanks for pointing that. I've made the change. Glad that I didn't had to change anywhere other than that particular line.



> On that note: You can open the `do` block directly after the `resources`:
> 
> ```ruby
> resources :user_roles, only: [...] do
>   get '/sub-route'
> end
> ```

If I do that, the route will be like `user_roles/{id}/sub-route` and we don't have any `id` to give since the APIs are fetching a list of roles and not working on a single role.

This worked for `wrt/reset_claim_count` because there we had a `person_id` to work on.

gregorbg (approved):

PR #8639, opened 2023-12-20
Add query params to delegates page

### LINKED ISSUE #8622: WCA Delegates Page: Add query param when switching between regions such that on refresh, you see the same list you were seeing.

**Additional context**

maxidragon:
Closes #8622

### DIFF
--- a/WcaOnRails/app/webpacker/components/Delegates/index.jsx
+++ b/WcaOnRails/app/webpacker/components/Delegates/index.jsx
@@ -38,24 +38,14 @@ export default function Delegates() {
 
   const [hash, setHash] = useHash();
 
-  const SelectedComponent = React.useMemo(() => {
+  const activeRegion = React.useMemo(() => {
     const selectedRegionIndex = delegateRegions.findIndex((region) => region.id === +hash);
     if (selectedRegionIndex === -1) {
       setHash(delegateRegions[0]?.id);
-      return () => null;
+      return null;
     }
-    const selectedSection = delegateRegions[selectedRegionIndex];
-    if (selectedSection.component) {
-      return selectedSection.component;
-    }
-
-    return () => null;
+    return delegateRegions[selectedRegionIndex];
   }, [delegateRegions, hash, setHash]);
-
-  const activeRegion = React.useMemo(
-    () => delegateRegions.find((region) => region.id === +hash),
-    [delegateRegions, hash],
-  );
 
   const [adminMode, setAdminMode] = React.useState(false);
 
@@ -115,7 +105,6 @@ export default function Delegates() {
                   onChange={(__, { value }) => setHash(value)}
                 />
               </Grid.Row>
-              <Grid.Row><SelectedComponent /></Grid.Row>
               <DelegatesOfRegion
                 activeRegion={activeRegion}
                 isAdminMode={adminMode}
--- a/WcaOnRails/app/webpacker/components/Delegates/index.jsx
+++ b/WcaOnRails/app/webpacker/components/Delegates/index.jsx
@@ -39,9 +39,11 @@ export default function Delegates() {
   const [hash, setHash] = useHash();
 
   const activeRegion = React.useMemo(() => {
-    const selectedRegionIndex = delegateRegions.findIndex((region) => region.id === +hash);
+    const selectedRegionIndex = delegateRegions.findIndex(
+      (region) => region.metadata.friendly_id === hash,
+    );
     if (selectedRegionIndex === -1 && delegateRegions.length > 0) {
-      setHash(delegateRegions[0]?.id);
+      setHash(delegateRegions[0]?.metadata.friendly_id);
       return null;
     }
     return delegateRegions[selectedRegionIndex];
@@ -80,8 +82,8 @@ export default function Delegates() {
               <Menu.Item
                 key={region.id}
                 name={region.name}
-                active={region.id === hash}
-                onClick={() => setHash(region.id)}
+                active={region.metadata.friendly_id === hash}
+                onClick={() => setHash(region.metadata.friendly_id)}
               />
             ))}
           </Menu>
@@ -99,7 +101,7 @@ export default function Delegates() {
                   options={delegateRegions.map((region) => ({
                     key: region.id,
                     text: region.name,
-                    value: region.id,
+                    value: region.metadata.friendly_id,
                   }))}
                   value={hash}
                   onChange={(__, { value }) => setHash(value)}
--- a/WcaOnRails/app/webpacker/components/Delegates/index.jsx
+++ b/WcaOnRails/app/webpacker/components/Delegates/index.jsx
@@ -40,7 +40,7 @@ export default function Delegates() {
 
   const activeRegion = React.useMemo(() => {
     const selectedRegionIndex = delegateRegions.findIndex((region) => region.id === +hash);
-    if (selectedRegionIndex === -1) {
+    if (selectedRegionIndex === -1 && delegateRegions.length > 0) {
       setHash(delegateRegions[0]?.id);
       return null;
     }
--- a/WcaOnRails/app/webpacker/components/Delegates/index.jsx
+++ b/WcaOnRails/app/webpacker/components/Delegates/index.jsx
@@ -19,6 +19,7 @@ import Loading from '../Requests/Loading';
 import useLoggedInUserPermissions from '../../lib/hooks/useLoggedInUserPermissions';
 import { groupTypes } from '../../lib/wca-data.js.erb';
 import DelegatesOfRegion from './DelegatesOfRegion';
+import useHash from '../../lib/hooks/useHash';
 
 // let i18n-tasks know the key is used
 // i18n-tasks-use t('delegates_page.acknowledges')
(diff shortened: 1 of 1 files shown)


WcaOnRails/app/webpacker/components/Delegates/index.jsx
```diff
@@ -35,12 +36,28 @@ export default function Delegates() {
     [delegateGroups],
   );
 
-  const [activeRegion, setActiveRegion] = React.useState();
-  const [adminMode, setAdminMode] = React.useState(false);
+  const [hash, setHash] = useHash();
+
+  const SelectedComponent = React.useMemo(() => {
+    const selectedRegionIndex = delegateRegions.findIndex((region) => region.id === +hash);
```
danieljames-dj, line 42: Why do we have the '+' sign?

WcaOnRails/app/webpacker/components/Delegates/index.jsx
```diff
@@ -35,12 +36,28 @@ export default function Delegates() {
...
+    if (selectedRegionIndex === -1) {
+      setHash(delegateRegions[0]?.id);
```
danieljames-dj, line 44: Will this work if `delegateRegions` is empty? Like while the page is loading, `delegateRegions` will be empty right?

danieljames-dj (commented):
Thanks @maxidragon for working on this. I'm little busy so couldn't review it completely. I'll review it completely later, or in case Gregor reviews before me, this will definitely be good to go.

WcaOnRails/app/webpacker/components/Delegates/index.jsx
maxidragon: Because `+` is a short version of `parseInt`. I used region.id as hash and I need to parse it before checking if it equals to `region.id`.

WcaOnRails/app/webpacker/components/Delegates/index.jsx
maxidragon: If `delegateRegions` is empty, it will show the `Loading` component, I think it is expected behavior

WcaOnRails/app/webpacker/components/Delegates/index.jsx
```diff
@@ -35,12 +36,28 @@ export default function Delegates() {
...
+    }
+
+    return () => null;
+  }, [delegateRegions, hash, setHash]);
 
-  React.useEffect(() => {
-    setActiveRegion(delegateRegions?.[0]);
-  }, [delegateRegions]);
+  const activeRegion = React.useMemo(
+    () => delegateRegions.find((region) => region.id === +hash),
```
danieljames-dj, line 56: Instead of find, we can use the selectedRegionIndex from above right?

WcaOnRails/app/webpacker/components/Delegates/index.jsx
```diff
@@ -35,12 +36,28 @@ export default function Delegates() {
...
+  const [hash, setHash] = useHash();
+
+  const SelectedComponent = React.useMemo(() => {
+    const selectedRegionIndex = delegateRegions.findIndex((region) => region.id === +hash);
+    if (selectedRegionIndex === -1) {
+      setHash(delegateRegions[0]?.id);
+      return () => null;
+    }
+    const selectedSection = delegateRegions[selectedRegionIndex];
+    if (selectedSection.component) {
```
danieljames-dj, line 48: selectedSection is an item of delegateRegions and delegateRegions is fetched from server. But delegateRegions doesn’t have component key. Correct me if I’m wrong.

WcaOnRails/app/webpacker/components/Delegates/index.jsx
coder13: Does this do proper error handling @maxidragon?

WcaOnRails/app/webpacker/components/Delegates/index.jsx
```diff
@@ -94,12 +111,11 @@ export default function Delegates() {
...
                   }))}
-                  value={activeRegion.id}
-                  onChange={(__, { value }) => setActiveRegion(
-                    delegateRegions.find((region) => region.id === value),
-                  )}
+                  value={hash}
+                  onChange={(__, { value }) => setHash(value)}
                 />
               </Grid.Row>
+              <Grid.Row><SelectedComponent /></Grid.Row>
```
coder13, line 118: What does this do?

WcaOnRails/app/webpacker/components/Delegates/index.jsx
maxidragon: @coder13 yes, look at lines 43-45
```js
if (selectedRegionIndex === -1) {
      setHash(delegateRegions[0]?.id);
 }
```

WcaOnRails/app/webpacker/components/Delegates/index.jsx
maxidragon: If we want to use `selectedRegionIndex`, it has to be outside `SelectedComponent`. Did you mean that?

WcaOnRails/app/webpacker/components/Delegates/index.jsx
maxidragon: It renders a component responsible for setting the default value. I followed what Daniel did in #8529

danieljames-dj: The case in that PR is different. There, the component varies depending on the hash. But here, the component to be rendered is always the same - `DelegatesOfRegion`, which is already rendered. So, I guess this line is not needed. If you are also convinced, can you please remove this line and the initialization of `SelectedComponent`?

WcaOnRails/app/webpacker/components/Delegates/index.jsx
danieljames-dj: Yes, but according to the comment below, you won't be needing `SelectedComponent`. If that's the case, then this comment can be ignored.

WcaOnRails/app/webpacker/components/Delegates/index.jsx
```diff
@@ -35,12 +36,18 @@ export default function Delegates() {
...
   );
 
-  const [activeRegion, setActiveRegion] = React.useState();
-  const [adminMode, setAdminMode] = React.useState(false);
+  const [hash, setHash] = useHash();
+
+  const activeRegion = React.useMemo(() => {
+    const selectedRegionIndex = delegateRegions.findIndex((region) => region.id === +hash);
+    if (selectedRegionIndex === -1) {
+      setHash(delegateRegions[0]?.id);
```
danieljames-dj, line 44: I was about to approve the PR, and suddenly thought of this edge case. Can you please confirm if http://localhost:3000/delegates#europe is opening the Europe page?

Sorry that I couldn't try it myself because my local DB is bit messed up because of which I can't open anything that uses `roles` table as my local DB has `user_roles` instead of `roles`.

maxidragon: I used `region.id` as hash, so http://localhost:3000/delegates#3 is opening Europe page. Should I use region name instead of id?

danieljames-dj: Oops, just realized that there is no `friendly-id` field in metadata. So you can go ahead with ID itself. Will change once the `friendly_id` is added to metadata.

WcaOnRails/app/webpacker/components/Delegates/index.jsx
```diff
@@ -40,7 +40,7 @@ export default function Delegates() {
 
   const activeRegion = React.useMemo(() => {
     const selectedRegionIndex = delegateRegions.findIndex((region) => region.id === +hash);
-    if (selectedRegionIndex === -1) {
+    if (selectedRegionIndex === -1 && delegateRegions.length > 0) {
```
maxidragon, line 43: I realized that we need to check if `delegateRegions` is not empty; otherwise, `activeRegion` will always be the first element of the `delegateRegions` array.

maxidragon:
Everything works now ;)
[Screencast from 2023-12-23 14-50-32.webm](https://github.com/thewca/worldcubeassociation.org/assets/94986692/9fa154fc-92f8-4a5f-abd5-ee763de4615c)

danieljames-dj:
> Everything works now ;)
> 
> Video: [Screencast from 2023-12-23 14-50-32.webm](https://github.com/thewca/worldcubeassociation.org/assets/94986692/9fa154fc-92f8-4a5f-abd5-ee763de4615c)

Thanks @maxidragon. All changes looks good to me. Let's wait for a final review from @gregorbg as well.

danieljames-dj:
I've added the friendly ID to the regions in https://github.com/thewca/worldcubeassociation.org/pull/8645. If that gets merged before this PR, you should be able to use `region.metadata.friendly_id` instead of `region.id`. I'll update you in case the other PR gets merged earlier, if not it can be changed later as well.

maxidragon:
#8645 has already been merged, so I changed the `region.id` to `region.metadata.friendly_id `.

danieljames-dj (approved):
LGTM

PR #8642, opened 2023-12-23
Panel paths to a global variable for easy pathing

danieljames-dj:
If we want to route to the panel from any other part of the website, it's difficult currently. Example, if we want to link the edit icon of senior delegate in roles page to board panel's 'regions-manager' tab, only way is hardcoding but that doesn't look good. So instead, we can do something like `${panelRoute.board}/${PANEL_LIST.board.regionsManager}`.

There might be better way to handle this, if anybody get any idea, please pitch in.

### DIFF
--- a/WcaOnRails/app/webpacker/components/Panel/PanelList.js
+++ b/WcaOnRails/app/webpacker/components/Panel/PanelList.js
@@ -1,0 +1,23 @@
+const PANEL_LIST = {
+  board: {
+    seniorDelegatesList: 'senior-delegates-list',
+    councilLeaders: 'council-leaders',
+    regionsManager: 'regions-manager',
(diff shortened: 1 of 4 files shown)


WcaOnRails/app/webpacker/components/Panel/PanelList.js
```diff
@@ -0,0 +1,23 @@
+const PanelList = {
```
kr-matthews, line 1: I think this should be `panelList` (or maybe `PANEL_LIST`?).

danieljames-dj: `PANEL_LIST` looks better. Thank you Kevin.

gregorbg (approved):

PR #8644, opened 2023-12-27
Modify regions manager page to allow modify senior delegate

danieljames-dj:
Screenshot 2023-12-27 at 13 33 38

### DIFF
--- a/WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
+++ b/WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
@@ -297,7 +297,7 @@ class Api::V0::UserRolesController < Api::V0::ApiController
       end
     else
       group = UserGroup.find(group_id)
-      status = params.require(:status) if UserGroup.group_types_with_status_metadata.include?(group.group_type)
+      status = params.require(:status) if UserGroup.group_types_containing_status_metadata.include?(group.group_type)
       location = params.require(:location) if group.group_type == UserGroup.group_types[:delegate_regions]
       if status.present? && group.unique_status?(status)
         end_role_for_user_in_group_with_status(group, status)
--- a/WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
+++ b/WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
@@ -257,11 +257,28 @@ class Api::V0::UserRolesController < Api::V0::ApiController
     render json: roles
   end
 
+  private def end_delegate_role_for_user(user)
+    if user.delegate_status.present?
+      remove_pending_wca_id_claims(user)
+      user.update!(delegate_status: '', region_id: '', location: '')
+      send_role_change_notification(user)
+    end
+  end
+
+  private def end_role_for_user_in_group_with_status(group, status)
+    if group.group_type == UserGroup.group_types[:delegate_regions]
+      user = User.find_by(region_id: group.id, delegate_status: status)
+      if user.present?
+        end_delegate_role_for_user(user)
+      end
+    end
+  end
+
   def create
     user_id = params.require(:userId)
     group_id = params.require(:groupId)
 
-    if group_id.include?("_") # Temporary hack to support some old system roles, will be removed once all roles are
+    if group_id.is_a?(String) && group_id.include?("_") # Temporary hack to support some old system roles, will be removed once all roles are
       # migrated to the new system.
       group_type = group_id_of_old_system_to_group_type(group_id)
       original_group_id = group_id.split("_").last
@@ -279,7 +296,22 @@ class Api::V0::UserRolesController < Api::V0::ApiController
         render status: :unprocessable_entity, json: { error: "Invalid group type" }
       end
     else
-      render status: :unprocessable_entity, json: { error: "Invalid group id" }
+      group = UserGroup.find_by_id(group_id)
+      status = params.require(:status) if UserGroup.group_types_with_status_metadata.include?(group.group_type)
+      location = params.require(:location) if group.group_type == UserGroup.group_types[:delegate_regions]
+      if status.present? && group.unique_status?(status)
+        end_role_for_user_in_group_with_status(group, status)
+      end
+      if group.group_type == UserGroup.group_types[:delegate_regions]
+        user = User.find_by_id(user_id)
+        user.update!(delegate_status: status, region_id: group.id, location: location)
+        send_role_change_notification(user)
+        render json: {
+          success: true,
+        }
+      else
+        render status: :unprocessable_entity, json: { error: "Invalid group type" }
+      end
     end
   end
 
@@ -352,9 +384,7 @@ class Api::V0::UserRolesController < Api::V0::ApiController
       user_id = params.require(:userId)
 
       user = User.find(user_id)
-      remove_pending_wca_id_claims(user)
-      user.update!(delegate_status: '', region_id: '', location: '')
-      send_role_change_notification(user)
+      end_delegate_role_for_user(user)
 
       render json: {
         success: true,
--- a/WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
+++ b/WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
@@ -296,14 +296,14 @@ class Api::V0::UserRolesController < Api::V0::ApiController
         render status: :unprocessable_entity, json: { error: "Invalid group type" }
       end
     else
-      group = UserGroup.find_by_id(group_id)
+      group = UserGroup.find(group_id)
       status = params.require(:status) if UserGroup.group_types_with_status_metadata.include?(group.group_type)
       location = params.require(:location) if group.group_type == UserGroup.group_types[:delegate_regions]
       if status.present? && group.unique_status?(status)
         end_role_for_user_in_group_with_status(group, status)
       end
       if group.group_type == UserGroup.group_types[:delegate_regions]
-        user = User.find_by_id(user_id)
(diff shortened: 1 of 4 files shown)


WcaOnRails/app/webpacker/components/Panel/Board/RegionManager.jsx
```diff
@@ -45,50 +49,130 @@ function UserGroupVisibility({ userGroup, save, sync }) {
...
   const [saveError, setSaveError] = React.useState();
+  const [selectedGroup, setSelectedGroup] = React.useState();
+  const [newLeadDelegate, setNewLeadDelegate] = React.useState();
+  const [loading, setLoading] = React.useState(false);
+
+  React.useEffect(() => {
+    if (selectedGroup) {
+      setOpenModalType('newLeadDelegate');
+    }
+  }, [selectedGroup]);
```
kr-matthews, line 66: It's best to avoid side effects if possible. Just explicitly call `setOpenModalType('newLeadDelegate')` each time `setSelectedGroup` is set to a non-null value. Use a helper function if you want:
```
const selectedGroupAndShowModal = (group) -> {
  setSelectedGroup(group);
  setOpenModalType('newLeadDelegate');
}

File: WcaOnRails/app/webpacker/components/Panel/Board/RegionManager.jsx [hunk 2]
```diff
@@ -45,50 +49,130 @@ function UserGroupVisibility({ userGroup, save, sync }) {
...
 
-  const closeModal = () => setOpenModalType(null);
+  const closeModal = () => {
+    setOpenModalType(null);
+  };
```
kr-matthews, line 70: Any reason this was changed?

danieljames-dj: Oops did it accidentally. Thanks for pointing.

WcaOnRails/app/webpacker/components/Panel/Board/RegionManager.jsx
danieljames-dj: Makes sense, changed. Thank you Kevin.

WcaOnRails/app/webpacker/components/Panel/Board/RegionManager.jsx
```diff
@@ -156,6 +237,50 @@ export default function RegionManager() {
...
+              onClick={async () => {
+                closeModal();
+                setLoading(true);
+                await removeOldLeadDelegate(selectedGroup.lead_user);
+                await addNewLeadDelegate(newLeadDelegate, selectedGroup);
+                sync();
+                setSelectedGroup(null);
+                setNewLeadDelegate(null);
+                setLoading(false);
+              }}
```
gregorbg, line 277: Please no async/await in React components. Just do whatever needs to be done in the `onSuccess` hook of the `useSaveData` action

gregorbg (changes requested):

WcaOnRails/app/models/user_group.rb
```diff
@@ -28,8 +28,18 @@ def roles
     end
   end
 
+  # TODO: Once the roles migration is done, add a validation to make sure there is only one lead_user per group.
+  def lead_user
+    if self.group_type == UserGroup.group_types[:delegate_regions]
+      self.senior_delegate
+    else
+      nil
+    end
```
gregorbg, line 37:
```suggestion
    self.senior_delegate if self.group_type == UserGroup.group_types[:delegate_regions]
```

danieljames-dj: Done

WcaOnRails/app/webpacker/components/Panel/Board/RegionManager.jsx
danieljames-dj: Okay, removed them and added the logic in backend which looks more ideal.

WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -81,7 +81,7 @@ class Api::V0::UserRolesController < Api::V0::ApiController
 
   # Returns a list of roles by user which are not yet migrated to the new system.
   private def group_roles_not_yet_in_new_system(group_id)
-    group = UserGroup.find(group_id)
+    group = UserGroup.find_by_id(group_id)
```
gregorbg, line 84: `find_by_id` and `find` should be the same, as long as `id` is the primary key (which it should be in 99% of cases).
So why are you changing this?

WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -191,11 +191,28 @@ def index_for_group_type
...
+
   def create
     user_id = params.require(:userId)
     group_id = params.require(:groupId)
 
-    if group_id.include?("_") # Temporary hack to support some old system roles, will be removed once all roles are
+    if group_id.is_a?(String) && group_id.include?("_") # Temporary hack to support some old system roles, will be removed once all roles are
       # migrated to the new system.
       group_type = group_id.split("_").first
       original_group_id = group_id.split("_").last
```
gregorbg, line 218: Maybe this needs the "drop reverse drop last reverse reverse first backwards" hack from the other PR now?

WcaOnRails/app/models/user_group.rb
```diff
@@ -9,6 +9,14 @@ class UserGroup < ApplicationRecord
     translators: "translators",
   }
 
+  def self.group_types_with_status_metadata
```
gregorbg, line 12: If I'm not mistaken, this returns just an array of strings.
Why is it called `...with_status_metadata`?

WcaOnRails/app/models/user_group.rb
```diff
@@ -40,8 +48,25 @@ def roles
     end
   end
 
+  # TODO: Once the roles migration is done, add a validation to make sure there is only one lead_user per group.
+  def lead_user
```
gregorbg, line 52: I feel like I saw this in another PR before...

gregorbg (changes requested):

WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
danieljames-dj: Actually yes & no. Yes for consistency, hence I've added it now. No because this will be used only by group_type councils which doesn't have `_`. But maybe we will have it in future, so good to have.

WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
danieljames-dj: I actually faced an issue like find not there something. I don't remember the error exactly and don't know where it caused. So replaced `find` with `find_by_id`. Now I changed back and not getting the issue. Will check in deep if I face it again.

WcaOnRails/app/models/user_group.rb
danieljames-dj: These will return only those groups which have `status` metadata, and not the ones that doesn't have `status` metadata like `translators`, `delegate_probations`, etc.

WcaOnRails/app/models/user_group.rb
danieljames-dj: Oh is it? I don't think I've added it anywhere before? 🤔 Maybe you saw it in same PR in earlier review?

WcaOnRails/app/webpacker/components/Panel/Board/RegionManager.jsx
```diff
@@ -156,6 +205,60 @@ export default function RegionManager() {
...
+              closeModal();
+              setSelectedGroup(null);
+              setNewLeadDelegate(null);
+            }}
+            >
+              Cancel
+            </Form.Button>
+            <Form.Button
+              disabled={!newLeadDelegate}
+              onClick={async () => {
```
gregorbg, line 236: Still an `async` that is not necessary

gregorbg (changes requested):

WcaOnRails/app/webpacker/components/Panel/Board/RegionManager.jsx
gregorbg: Muuuuuuuch better now :heart_eyes:

WcaOnRails/app/models/user_group.rb
gregorbg: Ah, in that case I misunderstood. To me, the function name sounds like "return some group types and also return their status metadata". Instead, it should be like "return only those group types which can have status metadata".

Please spend a brief moment to see if you can come up with a more concise name, but don't waste more than 2 minutes on this. In case of doubt, it's fine the way it currently is :smile:

WcaOnRails/app/models/user_group.rb
gregorbg: Oh, it is really possible that I lost track of the different PRs while reviewing your work all at once :sweat_smile:

WcaOnRails/app/models/user_group.rb
danieljames-dj: Okay, in two minutes I got another name: `group_types_containing_status_metadata`. Pushed it, is this fine?

WcaOnRails/app/webpacker/components/Panel/Board/RegionManager.jsx
danieljames-dj: Oops, missed it. Removed now.

gregorbg (approved):

PR #8647, opened 2023-12-29
Added results/personal_records/:user_id API endpoint

dunkOnIT:
**Intention:** Expose personal records information which the Registration Service backend can use to verify qualifications. 

## Questions/Notes:
- Putting it under `/results`, as this should ultimately be provided by a Results Service
- Shares a lot of similarities with Finn's `/users/me/personal_records` route - not sure if this is of concern
- **Question 1**: I'm currently getting `ranksSingle` and `ranksAverage` and then transforming the data in the controller - should I instead have an `indexedRanksSingle` method on the model? (I don't want to introduce a method on the model that is only used in one place and returns similar data to another method - seems better to use the existing method and transform the data. )
- **Question 2:** Just noticed the methods use camelCase not snake_case - should we change that? 
- **Question 3**: Should we group PR's under single/average -> event_id, or event_id -> single/average? Do we care? (see below for example)
- **Question 4**: Does it matter that I'm removing fields that Finn exposes in /users/me/personal_records? (specifically: personId, id)

### WCIF Personal Bests are defined as an array of these JSON objects:
```json
[{
          "eventId": "222",
          "best": 311,
          "worldRanking": 2894,
          "continentalRanking": 726,
          "nationalRanking": 22,
          "type": "single"
}]
```

### The `/user/me/personal_records` PR iterates on this by grouping them as average or single, and adding some new fields (in **bold:** 
```json
{ "single": [
    {
      **"id": 1451696265,**
      "**personId": "2019HOBB02",**
      "eventId": "222",
      "best": 513,
      "worldRank": 52323,
      "continentRank": 442,
      "countryRank": 260
    },
]}
```

### This PR returns data in the following format, adding eventId as a key on the personal_record object, and removing `id` and `personId` 
```json
{ "single": {
    "222": {
      "best": 513,
      "worldRank": 52323,
      "continentRank": 442,
      "countryRank": 260
    },
}}
```

### To me, it would be more intuitive to group by event -> single/average, as follows: 
```json
{ "222": {
    "single": {
      "best": 513,
      "worldRank": 52323,
      "continentRank": 442,
      "countryRank": 260
    },
    "average": {
        ....
    }
}}
```


WcaOnRails/config/routes.rb
```diff
@@ -343,7 +346,10 @@
         get '/scrambles/:event_id' => 'competitions#event_scrambles', as: :event_scrambles
         patch '/wcif' => 'competitions#update_wcif', as: :update_wcif
       end
-      get '/records' => "api#records"
+
+      # namespace :results do
```
danieljames-dj, line 350: Was this accidentally added?

WcaOnRails/app/controllers/api/v0/results_controller.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+class Api::V0::ResultsController < Api::V0::ApiController
+  def personal_records
+    user = User.find_by_id(params[:user_id])
```
danieljames-dj, line 5: Maybe `params.require` if user_id is a mandatory param?

dunkOnIT:
Ok this gets WAY more complicated when trying to include the date on which a result was achieved, because `ranksSingle` and `ranksAverage` (where we are currently getting the PR's from) don't store them. 

## I see a few options for code changes, in order of my preference:
1. Store the `date_achieved` (or event just date_created?) on `ranksSingle` and `ranksAverage`. This would be hard to backfill, as there's no link from the `ranks` tables to the attempt date/competition date.
2. Store `latest_date_achieved` instead, and default all values to the date on which the field is added. Very easy to populate, and should work unless there are qualifications for competitions with registrations still open that have a date requirement before the date of the change (which I think is unlikely, but haven't verified)
3. Explore changing the data model to have PersonalBest somehow linked to the User and the Result - this could definitely work, but would represent a significant migration, perhaps more appropriate for the results rework? 

## Alternatives
1. Alternatively, we move the qualification logic entirely - and have a `GET /competition/{comp-id}/check_qualification/{user_id}` endpoint that uses. I really don't like this, because it generates an uncacheable REST call for _every single_ registration, which means that traffic traffic on the monolith scales with registration count, which sort of defeats what we're trying to do. 
2. Add this as a requirement for the Results rework, and don't support qualifications with date requirements on the new Registration System until the Results rework is done.

gregorbg:
> Ok this gets WAY more complicated when trying to include the date on which a result was achieved, because `ranksSingle` and `ranksAverage` (where we are currently getting the PR's from) don't store them.

For some high-level context: This fact, which you correctly identified, is one of our biggest pain points in the current Results tables from a software point of view. It not only affects qualifications but also rankings and regional records (NRs/CRs/WRs). It definitely has to be tackled as part of the Results redesign.

As such, I suggest simply skipping this feature in the new Reg service "for now" (the Monolith doesn't have automated qualification checks for a given deadline, either. It can only check whether qualification requirements are fulfilled _at the time of filling out the registration form_, which should be feasible in the new Reg service as well) and implementing it later when we have the proper data available.

dunkOnIT:
Ok great thanks! Will continue with the current behaviour then.

WcaOnRails/app/controllers/api/v0/results_controller.rb
dunkOnIT: Good point! Changed

app/controllers/api/v0/results_controller.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+class Api::V0::ResultsController < Api::V0::ApiController
+  def personal_records
+    user = User.find_by_id(params.require(:user_id))
```
danieljames-dj, line 5: Maybe `find` instead of `find_by_id` so that it throws exception if user ID is not available?

danieljames-dj: (Sorry for not pointing this in my initial review, I was not aware of this knowledge that time)

dunkOnIT: No worries! I have no idea why I was using `find_by_id` there :P

gregorbg (changes requested):
I have a strong bias towards WCIF. We should try to keep our APIs as consistent as we can.

If there's a strong reason why the Reg Service needs the elements grouped by Event ID, then the Reg Service as the one receiving the data should be the one to group the API output by event ID. The monolith / API should stay consistent with the /me/personal_records endpoint and just return a { single: ... average: ... } structure as-is. Consumers of this data can still trivially group it later if they need to.

app/controllers/api/v0/results_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+      single: create_simple_records_hash(person.ranksSingle),
+      average: create_simple_records_hash(person.ranksAverage),
+    }
+  end
+
+  private
+
+    def create_simple_records_hash(personal_records)
+      indexed_records = personal_records.index_by { |record| record['eventId'] }
+      indexed_records.transform_values { |record| record.attributes.except('id', 'personId', 'eventId') }
```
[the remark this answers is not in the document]
dunkOnIT: Changed the whole thing to just use WCIF, based on our convo ages ago re adhering to the WCIF spec and letting the client do the work

app/models/user.rb
```diff
@@ -1169,8 +1169,11 @@ def serializable_hash(options = nil)
     json
   end
 
+  def personal_records
```
gregorbg, line 1172: It's dangerous to call this `personal_records` and then actually return WCIF.
In the world of Ruby object models, this should actually return the `.compact.flatten`ed array. The consumer of the method should then map it to WCIF if necessary.

app/models/user.rb
```diff
@@ -1193,7 +1196,7 @@ def to_wcif(competition, registration = nil, registrant_id = nil, authorized: fa
       },
       "roles" => roles,
       "assignments" => registration&.assignments&.map(&:to_wcif) || [],
-      "personalBests" => person_pb.map(&:to_wcif),
+      "personalBests" => personal_records,
```
gregorbg, line 1199: Map to WCIF here as per my remark above.

gregorbg (changes requested):

app/models/user.rb
dunkOnIT: Changed

app/models/user.rb
dunkOnIT: Changed

app/controllers/api/v0/results_controller.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class Api::V0::ResultsController < Api::V0::ApiController
+  def personal_records
+    user = User.find(params.require(:user_id))
+
+    render json: user.personal_records
```
gregorbg, line 7: Thanks for the previous set of changes.
Now you've gotta pay attention that this call doesn't return WCIF automatically 🙃

gregorbg (changes requested):

app/controllers/api/v0/results_controller.rb
dunkOnIT: Fixed :melt:

app/controllers/api/v0/api_controller.rb
```diff
@@ -29,6 +29,11 @@ def auth_results
     render json: { status: "ok" }
   end
 
+  def user_qualification_data
+    user = User.find(params.require(:user_id))
+    render json: user.person.personal_records.map(&:to_wcif)
```
gregorbg, line 34: Attention! This code will currently crash for users that have never competed before.

Those users have never been assigned a WCA ID, so `user.person` will already return `nil`. You just happily proceed to call `.personal_records` on a potentially nil value, which will lead to crashes.

app/controllers/api/v0/api_controller.rb
```diff
@@ -29,6 +29,11 @@ def auth_results
     render json: { status: "ok" }
   end
 
+  def user_qualification_data
+    user = User.find(params.require(:user_id))
```
gregorbg, line 33:
```suggestion
    user = User.includes(person: [:ranksSingle, :ranksAverage]).find(params.require(:user_id))
```

gregorbg: Makes the DB query much more efficient methinks

gregorbg (changes requested):

dunkOnIT:
(Busy adding tests for this which I should have done in the first place, just had to fix some DB issues - hope to get it sorted soon!)

dunkOnIT:
Tests added

app/controllers/api/v0/persons_controller.rb
```diff
@@ -33,6 +33,12 @@ def competitions
     render json: person.competitions
   end
 
+  def personal_records
```
coder13, line 36: Shouldn't this be moved to a `users_controller`?

I would expect that routes for `persons` live in a `persons_controllers` and routes for `users` to live in a `users_controller`

app/controllers/api/v0/api_controller.rb
```diff
@@ -29,6 +29,49 @@ def auth_results
     render json: { status: "ok" }
   end
 
+  def user_qualification_data
```
coder13, line 32: Shouldn't this be moved to a `users_controller`?

I would expect that routes for `persons` live in a `persons_controllers` and routes for `users` to live in a `users_controller`

dunkOnIT: An early commit had it in a `results_controller`, we moved it here for reasons I can't quite recall. I'll check with Gregor on Friday.

dunkOnIT:
This PR is cleaned up + ready for review

app/controllers/api/v0/api_controller.rb
```diff
@@ -29,6 +29,52 @@ def auth_results
...
+        best: time,
+        on_or_before: cutoff_date,
+      }
+    end
+
+    render json: qualification_results
+  end
+
+  private def date_from_params
+    date = params.permit(:date)[:date]
```
dunkOnIT, line 73: Directly access the hash and handle the nil case

app/controllers/api/v0/api_controller.rb
```diff
@@ -29,6 +29,52 @@ def auth_results
     render json: { status: "ok" }
   end
 
+  def user_qualification_data
+    user = User.find(params.require(:user_id))
+    begin
```
dunkOnIT, line 34: Make this a `cutoff_date` function

app/controllers/api/v0/api_controller.rb
```diff
@@ -29,6 +29,52 @@ def auth_results
...
+      return render json: { error: 'Invalid date format. Please provide an iso8601 date string.' }, status: :bad_request
+    end
+
+    return render json: { error: 'You cannot request qualification data for a future date.' }, status: :bad_request if cutoff_date > Date.current
+    cutoff_date = cutoff_date.iso8601
+
+    qualification_results = []
+
+    # Compile singles
+    best_singles_by_cutoff = user.person&.best_singles_by(cutoff_date)
```
dunkOnIT, line 46: change this to `if.user.present?` and then don't worry about the safe accessing

app/controllers/api/v0/api_controller.rb
```diff
@@ -29,6 +29,52 @@ def auth_results
...
+    return render json: [] unless best_singles_by_cutoff.present? # If the user has no best singles, they have no qualification-relevant results
+    best_singles_by_cutoff.map do |event, time|
```
dunkOnIT, line 48: assign the return value of map to a `qualification_result_single` array and concat them at the end

app/controllers/api/v0/api_controller.rb
```diff
@@ -29,6 +29,52 @@ def auth_results
...
+  def user_qualification_data
+    user = User.find(params.require(:user_id))
+    begin
+      cutoff_date = date_from_params || Date.current
+    rescue Date::Error
+      return render json: { error: 'Invalid date format. Please provide an iso8601 date string.' }, status: :bad_request
+    end
+
+    return render json: { error: 'You cannot request qualification data for a future date.' }, status: :bad_request if cutoff_date > Date.current
+    cutoff_date = cutoff_date.iso8601
```
dunkOnIT, line 41: probably unnecessary - Rails may convert this

app/controllers/api/v0/api_controller.rb
```diff
@@ -29,6 +29,52 @@ def auth_results
...
+        type: 'single',
+        best: time,
+        on_or_before: cutoff_date,
+      }
+    end
+
+    # Compile averages
+    best_averages_by_cutoff = user.person&.best_averages_by(cutoff_date)
+    return render json: qualification_results unless best_averages_by_cutoff.present? # If the user has no best averages, we can just return their singles
+    best_averages_by_cutoff.map do |event, time|
```
dunkOnIT, line 60: have a function that takes event, type, time and cutoff as parameteres to return the hash I'm building here

app/models/result.rb
```diff
@@ -50,6 +50,7 @@ def unique_result_per_round
   scope :podium, -> { final.succeeded.where(pos: [1..3]) }
   scope :winners, -> { final.succeeded.where(pos: 1).joins(:event).order("Events.rank") }
   scope :before, ->(date) { joins(:competition).where("end_date < ?", date) }
+  scope :on_or_before, ->(date) { joins(:competition).where("end_date <= ?", date) }
```
danieljames-dj, line 53:
```
scope :on_or_before, ->(date) { joins(:competition).where(end_date: ..date) }
```

This should also probably work (@gregorbg taught me this)

dunkOnIT:
@thewca-bot deploy staging

app/controllers/api/v0/api_controller.rb
```diff
@@ -29,6 +29,47 @@ def auth_results
...
+  private def cutoff_date
+    if params[:date].present?
+      Date.safe_parse(params[:date])
+    else
+      Date.current
+    end
+  end
+
+  private def qualification_data(event, type, time, date)
+    raise ArgumentError.new("'type' may only contain the symbols `:single` or `:average`") unless [:single, :average].include?(type)
```
gregorbg, line 64: Not sure if you need this in a `private` method definition. This kind of error message is helpful in external, world-facing APIs but since this is by definition code that only we call ourselves, it feels a bit overkill.

You don't have to delete this line to get my approval, just noting it here.

gregorbg (approved):

app/controllers/api/v0/api_controller.rb
dunkOnIT: My thinking here is to rather fail loudly when someone gives an invalid input, than return `nil` or similar and then someone has to dig to figure out where the `nil` is coming from. Are there alternate approaches you'd suggest?

gregorbg: Use a strongly typed programming language.
I appreciate your thinking, but the ultimate consequence would be to bloat our entire codebase by at least factor 1.5 by introducing is_instance? checks everywhere.

dunkOnIT: > Use a strongly typed programming language.
TypeRuby :pray: 

> I appreciate your thinking, but the ultimate consequence would be to bloat our entire codebase by at least factor 1.5 by introducing is_instance? checks everywhere.

Not opposed to forgoing these types of checks, my arguments in favour of them would be: 
- I only see this being relevant/necessary in cases where we are defining some custom input values/symbols, which should be pretty rare.
- It's only a single line of code at the start of a function where it's needed, so the impact should hopefully be less than 1.5x our codebase - and it seems like the time saved on developers chasing obscure errors may be beneficial
- I assume that even in a strongly typed language you have to have some boilerplate somewhere that says "hey here are the explicitly-accepted input values for this custom argument"

PR #8648, opened 2023-12-29
Added /competitions/:comp_id/qualifications route

dunkOnIT:
Adds a route where competition qualification restrictions can be accessed directly, instead of needing to get them off the WCIF, for the Registration Service Backend to enforce qualification limits

### DIFF
--- a/app/controllers/api/v0/competitions_controller.rb
+++ b/app/controllers/api/v0/competitions_controller.rb
@@ -38,11 +38,14 @@ class Api::V0::CompetitionsController < Api::V0::ApiController
   end
 
   def qualifications
-    competition = competition_from_params
-    qualifications = competition.competition_events.each_with_object({}) do |event, hash|
-      next if event.qualification.nil?
-      hash[event.event_id] = event.qualification.to_wcif
-    end
+    competition = competition_from_params(associations: [:competition_events])
+
+    qualifications = competition.competition_events
+                                .where.not(qualification: nil)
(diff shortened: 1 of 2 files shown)


dunkOnIT:
**Question:** Should this be part of `/competitions/:comp_id`? I had opted not to add it there because that route is already returning a lot of data and needs to be reconsidered, so I don't want to make something else dependent on it.

WcaOnRails/app/controllers/api/v0/competitions_controller.rb
```diff
@@ -37,6 +37,15 @@ def show
     end
   end
 
+  def qualifications
+    competition = competition_from_params
+    qualifications = competition.competition_events.each_with_object({}) do |event, hash|
```
gregorbg, line 42: This `each_with_object` approach looks a bit ugly. Where does it come from?

coder13:
> **Question:** Should this be part of `/competitions/:comp_id`? I had opted not to add it there because that route is already returning a lot of data and needs to be reconsidered, so I don't want to make something else dependent on it.

The current `/competitions/:comp_id` endpoint is big and if you were to make it bigger, you'd be making other services that call this endpoint fetch more data than they expected, although the change is negligible. We should definitely think about a V1 API (graphql hint hint) but in my opinion, it should be fine to add this knowing we'd like to deprecate it at some point.

gregorbg (approved):

PR #8649, opened 2023-12-29
Bump Shakapacker to 7.2.0

gregorbg:


gregorbg:
Treating as a dependency update

PR #8652, opened 2023-12-29
Redirect downloads to TNoodle 1.2.0

gregorbg:


gregorbg:
Needs to be merged in conjunction with https://github.com/thewca/worldcubeassociation.org/pull/8651 and https://github.com/thewca/worldcubeassociation.org/pull/8650

PR #8654, opened 2023-12-29
Delete option in dues redirect

danieljames-dj:


WcaOnRails/app/webpacker/components/Panel/Wfc/DuesRedirect.jsx
```diff
@@ -115,6 +124,21 @@ export default function DuesRedirect() {
           </Form>
         </Modal.Content>
       </Modal>
+      <Confirm
+        open={toDeleteId}
+        content="Are you sure you want to delete?"
+        onCancel={() => setOpen(false)}
```
kr-matthews, line 130: Should this set the id to null?

WcaOnRails/app/webpacker/components/Panel/Wfc/DuesRedirect.jsx
```diff
@@ -115,6 +124,21 @@ export default function DuesRedirect() {
...
+        onConfirm={() => {
+          const duesRedirectId = toDeleteId;
+          setToDeleteId(null);
```
gregorbg, line 133: Setting the toDelete ID to null should only happen within the success callback.
This will also save you from copying the old value into a const.

WcaOnRails/app/controllers/api/v0/wfc/dues_redirects_controller.rb
```diff
@@ -31,4 +31,10 @@ def create
       render json: wfc_dues_redirect.errors, status: :unprocessable_entity
     end
   end
+
+  def destroy
+    wfc_dues_redirect = WfcDuesRedirect.find(params[:id])
```
gregorbg, line 36: Perhaps params.require will be safer?

WcaOnRails/app/webpacker/components/Panel/Wfc/DuesRedirect.jsx
danieljames-dj: Ya it was supposed to be `setToDeleteId(null)`, I missed it. Thanks.

WcaOnRails/app/webpacker/components/Panel/Wfc/DuesRedirect.jsx
danieljames-dj: Done

WcaOnRails/app/controllers/api/v0/wfc/dues_redirects_controller.rb
danieljames-dj: Oh yes, feeling bit embarrassing that I pointed out this in another PR very recently (in last 1 or 2 days), and I did that same mistake in my PR. 😅

gregorbg (approved):

PR #8657, opened 2023-12-30
Leader panel along with groups manager page

danieljames-dj:
Attached the screenshot of new Leader Panel. The Groups Manager page will currently only lists the members of the leader. I was also planning to add few more columns in which they can make actions like 'Promote', 'Demote', 'Resign'. But once I reached this stage, I felt like there are already too many changes that require review, so thought of moving the remaining part to another PR once this PR is merged.

Screenshot 2023-12-30 at 09 11 59

Future tasks related to this PR:
- Add action buttons in group manager page, so that the leader can do actions from here instead of team admin page.
- Migrate teams to user_groups
- Migrate team_members to user_roles
- Remove team admin page

### DIFF
--- a/WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
+++ b/WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class Api::V0::UserRolesController < Api::V0::ApiController
+  include EnumerableHelper
+
   before_action :current_user_is_authorized_for_action!, only: [:create, :update, :destroy]
   private def current_user_is_authorized_for_action!
     unless current_user.board_member? || current_user.senior_delegate?
--- a/WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
+++ b/WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
@@ -8,6 +8,40 @@ class Api::V0::UserRolesController < Api::V0::ApiController
     end
   end
 
+  private def status_sort_rank(status)
+    case status
+    when 'leader'
+      1
+    when 'senior_member'
+      2
+    when 'member'
+      3
+    else
+      4
+    end
+  end
+
+  # Sorts the list of roles based on the given list of sort keys and directions.
+  private def sorted_roles(roles, sort_param)
+    # The value of sort_param is inspired from https://specs.openstack.org/openstack/api-wg/guidelines/pagination_filter_sort.html.
+    sort_param ||= ''
+    sort_keys_and_directions = sort_param.split(',')
+    roles.sort_by { |role|
+      sort_keys_and_directions.map { |sort_key_and_direction|
+        sort_key = sort_key_and_direction.split(':').first
+        # FIXME: Utilize sort direction as well and reverse sort wherever necessary.
+        case sort_key
+        when 'startDate'
+          role[:start_date] # Can be changed to `role.start_date` once all roles are migrated to the new system.
+        when 'status'
+          status_sort_rank(role[:metadata][:status]) # Can be changed to `status_sort_rank(role.metadata.status)` once all roles are migrated to the new system.
+        when 'name'
+          role[:user][:name] # Can be changed to `role.user.name` once all roles are migrated to the new system.
+        end
+      }
+    }
+  end
+
   # Filters the list of roles based on the permissions of the current user.
   private def filter_roles_for_logged_in_user(roles)
     roles.select do |role|
@@ -137,6 +171,9 @@ class Api::V0::UserRolesController < Api::V0::ApiController
       status: params[:status],
     )
 
+    # Sort the roles.
+    roles = sorted_roles(roles, params[:sort])
+
     render json: roles
   end
 
@@ -150,6 +187,17 @@ class Api::V0::UserRolesController < Api::V0::ApiController
 
     # Filter the list based on the permissions of the logged in user.
     roles = filter_roles_for_logged_in_user(roles)
+
+    # Filter the list based on the other parameters.
+    roles = filter_roles_for_parameters(
+      roles: roles,
+      is_active: params.key?(:isActive) ? ActiveRecord::Type::Boolean.new.cast(params.require(:isActive)) : nil,
+      is_group_hidden: params.key?(:isGroupHidden) ? ActiveRecord::Type::Boolean.new.cast(params.require(:isGroupHidden)) : nil,
+      status: params[:status],
+    )
+
+    # Sort the roles.
+    roles = sorted_roles(roles, params[:sort])
 
     render json: roles
   end
@@ -199,6 +247,9 @@ class Api::V0::UserRolesController < Api::V0::ApiController
       status: status,
       is_active: is_active,
     )
+
+    # Sort the roles.
+    roles = sorted_roles(roles, params[:sort])
 
     render json: roles
   end
--- a/WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
+++ b/WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
@@ -20,7 +20,7 @@ class Api::V0::UserRolesController < Api::V0::ApiController
     # The value of sort_param is inspired from https://specs.openstack.org/openstack/api-wg/guidelines/pagination_filter_sort.html.
     sort_param ||= ''
     sort_keys_and_directions = sort_param.split(',')
-    roles.sort_by { |role|
+    roles.stable_sort_by { |role|
       sort_keys_and_directions.map { |sort_key_and_direction|
         sort_key = sort_key_and_direction.split(':').first
         # FIXME: Utilize sort direction as well and reverse sort wherever necessary.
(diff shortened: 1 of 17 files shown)


WcaOnRails/app/webpacker/components/Panel/Leader/GroupsManager.jsx
```diff
@@ -0,0 +1,74 @@
...
+import Errored from '../../Requests/Errored';
+import Loading from '../../Requests/Loading';
+
+function GroupTable({ groupId }) {
+  const { data: roles, loading, error } = useLoadedData(apiV0Urls.userRoles.listOfGroup(
+    groupId,
+    { isActive: true, isGroupHidden: false, status: 'leader' },
+  ));
+
+  if (loading || loading) return <Loading />;
```
kr-matthews, line 15: :thinking:

WcaOnRails/app/webpacker/components/Panel/Leader/GroupsManager.jsx
```diff
@@ -0,0 +1,74 @@
...
+      <Header as="h2">Active members</Header>
+      <Table celled>
+        <Table.Header>
+          <Table.Row>
+            <Table.HeaderCell>Name</Table.HeaderCell>
+            <Table.HeaderCell>Status</Table.HeaderCell>
+          </Table.Row>
+        </Table.Header>
+        <Table.Body>
+          {roles.map((role) => (
```
kr-matthews, line 29: Might be good to sort the array, maybe first by status and then alphabetically?

WcaOnRails/app/webpacker/components/Panel/Leader/GroupsManager.jsx
danieljames-dj: Oops 🤣 removed one now, thanks for noting.

WcaOnRails/app/webpacker/components/Panel/Leader/GroupsManager.jsx
danieljames-dj: Sorting looks like a good idea to sort. I feel giving priority to start_date over name would be also good. I've done the sorting now.

I also felt that sorting in server side would be better based on the sort parameter. I've done that now.

WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -8,6 +8,40 @@ class Api::V0::UserRolesController < Api::V0::ApiController
     end
   end
 
+  private def status_sort_rank(status)
```
gregorbg, line 11: What about `['leader', 'senior_member', 'member'].find_index(status) || 4`?

You could take this one step further and move the sorted array into a `STATUS_SORTING_ORDER` constant like so:
```ruby
STATUS_SORTING_ORDER.find_index(status) || STATUS_SORTING_ORDER.length # this ensures that the return value "grows" properly when the sorting order is adjusted later on
```

WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -8,6 +8,40 @@ class Api::V0::UserRolesController < Api::V0::ApiController
...
+      3
+    else
+      4
+    end
+  end
+
+  # Sorts the list of roles based on the given list of sort keys and directions.
+  private def sorted_roles(roles, sort_param)
+    # The value of sort_param is inspired from https://specs.openstack.org/openstack/api-wg/guidelines/pagination_filter_sort.html.
+    sort_param ||= ''
```
gregorbg, line 27:
```suggestion
  private def sorted_roles(roles, sort_param = '')
    # The value of sort_param is inspired from https://specs.openstack.org/openstack/api-wg/guidelines/pagination_filter_sort.html.
```

WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -8,6 +8,40 @@ class Api::V0::UserRolesController < Api::V0::ApiController
...
+    sort_keys_and_directions = sort_param.split(',')
+    roles.sort_by { |role|
+      sort_keys_and_directions.map { |sort_key_and_direction|
```
gregorbg, line 30: Are you sure that Ruby's `sort_by` is stable? I vaguely recall that it isn't, although I'm not 100% sure if I remember correctly.

WcaOnRails/app/models/team_member.rb
```diff
@@ -51,6 +55,34 @@ def cannot_ban_user_with_upcoming_comps
...
+      TEAM_STATUS_MEMBER
+    end
+  end
+
+  def role
+    {
+      start_date: start_date,
+      is_active: current_member?,
+      group: {
+        id: "teams_committees_" + team.id.to_s,
```
gregorbg, line 73: Isn't this prefix a bit redundant with the `group_type: UserGroup.group_types[:teams_committees]` below?

WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -81,12 +115,25 @@ class Api::V0::UserRolesController < Api::V0::ApiController
 
   # Returns a list of roles by user which are not yet migrated to the new system.
   private def group_roles_not_yet_in_new_system(group_id)
-    group = UserGroup.find(group_id)
     roles = []
+    if group_id.include?("_") # Temporary hack to support some old system roles, will be removed once all roles are
+      # migrated to the new system.
+      group_type = group_id.split("_").reverse.drop(1).reverse.join("_")
```
gregorbg, line 121: This "reverse drop reverse join" looks very unintuitive and is repeated a lot in other methods further down.
Can you extract it into a helper method along with a comment explaning what it actually does?

WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -187,7 +252,7 @@ def create
 
     if group_id.include?("_") # Temporary hack to support some old system roles, will be removed once all roles are
       # migrated to the new system.
-      group_type = group_id.split("_").first
+      group_type = group_id.split("_").reverse.drop(1).reverse.join("_")
```
gregorbg, line 255: Duplicate code, see above

WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -249,7 +314,7 @@ def update
       # all roles are migrated to the new system.
       group_id = params.require(:groupId)
       status = params.require(:status)
-      group_type = id.split("_").first
+      group_type = id.split("_").reverse.drop(1).reverse.join("_")
```
gregorbg, line 317: Duplicate code, see above

WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -286,7 +351,7 @@ def destroy
     elsif id.include?("_") # Temporary hack to support some old system roles, will be removed once
       # all roles are migrated to the new system.
       group_id = params.require(:groupId)
-      group_type = id.split("_").first
+      group_type = id.split("_").reverse.drop(1).reverse.join("_")
```
gregorbg, line 354: Duplicate code, see above

gregorbg (changes requested):

gregorbg:
...and merge conflict once again :sweat_smile:

danieljames-dj:
> ...and merge conflict once again 😅

Oops, rebased.

WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
gregorbg: Turns out, the sorting _is_ unstable even in the latest Ruby releases: https://ruby-doc.org/3.3.0/Array.html#method-i-sort

> When the [comparison callback] returns zero, the order for a and b is indeterminate, and may be unstable:

WcaOnRails/app/models/team_member.rb
```diff
@@ -51,6 +55,34 @@ def cannot_ban_user_with_upcoming_comps
...
+      TEAM_STATUS_MEMBER
+    end
+  end
+
+  def role
+    {
+      start_date: start_date,
+      is_active: current_member?,
+      group: {
+        id: UserGroup.group_types[:teams_committees] + "_" + team.id.to_s,
```
gregorbg, line 73: No, what I meant is: Why do you even _need_ this prefix, when the same information is already present in `group_type`?

gregorbg (changes requested):
Kindly pointing out that in addition to my new comment, there's two older code review comments left unsolved

WcaOnRails/app/models/team_member.rb
danieljames-dj: If I don't have the prefix, I won't be able to understand that this is not actual `role`. It might get confused between team ID and group ID at some places.

WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
danieljames-dj: Done

danieljames-dj (commented):
Oops just realized that I forgot to submit replies that I've written yesterday. I guess only sorting issue is left now. Looking into it.

gregorbg (approved):

PR #8689, opened 2024-01-03
Optionally override database cleaner suite setup for local testing

dunkOnIT:
If `SKIP_PRETEST_SETUP` is set to true in a `.env` file, skips DatabaseCleaner's pre-suite setup. This saves ~7s on test suite bootup. Combined with using spring in the Docker container, test setup time goes from ~14 seconds to <1s.

### DIFF
--- a/WcaOnRails/spec/support/database_cleaner.rb
+++ b/WcaOnRails/spec/support/database_cleaner.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 RSpec.configure do |config|
-  unless ENV['SKIP_PRETEST_SETUP'] == 'true'
+  unless EnvConfig.SKIP_PRETEST_SETUP?
     config.before(:suite) do
       DatabaseCleaner.clean_with :truncation
       TestDbManager.fill_tables
--- a/WcaOnRails/spec/support/database_cleaner.rb
+++ b/WcaOnRails/spec/support/database_cleaner.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 RSpec.configure do |config|
-  unless ENV['SKIP_PRETEST_SETUP']
+  unless ENV['SKIP_PRETEST_SETUP'] == 'true'
     config.before(:suite) do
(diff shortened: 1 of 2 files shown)


WcaOnRails/spec/support/database_cleaner.rb
```diff
@@ -1,9 +1,11 @@
 # frozen_string_literal: true
 
 RSpec.configure do |config|
-  config.before(:suite) do
-    DatabaseCleaner.clean_with :truncation
-    TestDbManager.fill_tables
+  unless ENV['SKIP_PRETEST_SETUP']
```
gregorbg, line 4: Accepting any value that is truthy (i.e. "smells" like a value that could be boolean `true`) is bad practice in general.
Try resorting to a method like `unless ENV['FOO_BAR'].present?` or `if ENV['FOO_BAR'].nil?` which have clearly defined behaviour.

gregorbg (changes requested):

WcaOnRails/spec/support/database_cleaner.rb
dunkOnIT: How about the latest push, which explicitly checks `== 'true'`? I agree with the concern re false positives, but am concerned that checking for presence leads to false positives where  `FOO_BAR=false` would resolve as true when checking `ENV['FOO_BAR'].present?`

dunkOnIT: This allows one to toggle true/false on SKIP_PRETEST_SETUP, which is nice if running the full suite or a set of tests which does rely on the pre-populated test data

FinnIckler: why not have it as an optional part of EnvConfig and then you can do `SKIP_PRETEST_SETUP?`

dunkOnIT: Oh yeah I like that a lot! Was trying to do it similar to `ENABLE_BULLET` but couldn't figure out how. Just done it in the latest push.

gregorbg (approved):

PR #8703, opened 2024-01-04
Trainee Delegates -> Staff Avatar Guidelines

### LINKED ISSUE #7819: Allow Trainee Delegates to see Staff Avatar Guidelines
Since Trainee Delegates are not staff members, the additional guidelines for staff members don't appear when they go to edit their avatar. But, these guidelines should still apply according to the Trainee Nomination process for Seniors and given they are listed on competition pages for competitors to find like any other delegate.

It would be good to fix this so that Trainees are able to see the guidelines they should follow.

I believe fixing this would also change WRT avatar approval by triggering their warning to apply the additional guidelines (in theory they do already apply the staff guidelines to Trainees but the warning is not triggered), correct me if I'm wrong!

Nanush7:
Closes #7819.

This allows Trainee Delegates to see the Staff avatar guidelines list and it also extends the WRT "warning" to Trainees.

I also updated the staff_avatar_guidelines.title locale to make it more precise (if you don't mind, I also updated the Spanish version, but I can revert that if you want).

### DIFF
--- a/WcaOnRails/app/views/admin/avatars/index.html.erb
+++ b/WcaOnRails/app/views/admin/avatars/index.html.erb
@@ -25,7 +25,7 @@
           <div class="panel-heading">
             <h3 class="panel-title">
               <%= link_to user.name, edit_user_path(user) %>
-              <% if user.staff? || user.trainee_delegate? %>
+              <% if user.staff_or_any_delegate? %>
                 (<strong>Staff Member or Trainee Delegate</strong> - see guidelines above)
               <% end %>
             </h3>
--- a/WcaOnRails/app/views/admin/avatars/index.html.erb
+++ b/WcaOnRails/app/views/admin/avatars/index.html.erb
@@ -25,8 +25,8 @@
           <div class="panel-heading">
             <h3 class="panel-title">
               <%= link_to user.name, edit_user_path(user) %>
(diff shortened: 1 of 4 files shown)


WcaOnRails/app/views/admin/avatars/index.html.erb
```diff
@@ -25,8 +25,8 @@
           <div class="panel-heading">
             <h3 class="panel-title">
               <%= link_to user.name, edit_user_path(user) %>
-              <% if user.staff? %>
-                (<strong>Staff Members</strong> - see guidelines above)
+              <% if user.staff? || user.trainee_delegate? %>
```
danieljames-dj, line 28: Maybe `user.staff_or_any_delegate?` instead?

WcaOnRails/app/views/users/edit.html.erb
```diff
@@ -199,7 +199,7 @@
                     <li><%= guideline %></li>
                   <% end %>
                 </ul>
-                <% if @user.staff? %>
+                <% if @user.staff? || @user.trainee_delegate? %>
```
danieljames-dj, line 202: (Same as above)
Maybe `user.staff_or_any_delegate?` instead?

gregorbg (approved):

Issue #8159, opened 2023-08-05
Have band details in database


danieljames-dj:
Currently, we have the fields like per_competitor_dues in each band and PERCENT_REGISTRATION_FEE_USED_FOR_DUE_AMOUNT hard-coded in https://github.com/thewca/worldcubeassociation.org/blob/master/WcaOnRails/app/models/country_band.rb.

It will be good if we have a table named something like `band_details` and have the above two band details there.

Note: PERCENT_REGISTRATION_FEE_USED_FOR_DUE_AMOUNT is currently 0.15 for all bands, but it can be different for different bands in future.

In the above mentioned file, if we have the values fetched from the table, it will be really easy for both WST & WFC for any future changes.

EdHollingdale:
Can we update prioritisation for this one? I think there are two sub tasks:
1. make it so that PERCENT_REGISTRATION_FEE_USED_FOR_DUE_AMOUNT is variable by band. As of 1 Jan 2024 with the updated policy this amount is variable (0% for Band 0, 5% for Bands 1 and 2, and 15% for Bands 3-5). 
2. make it so that per_competitor_dues and PERCENT_REGISTRATION_FEE_USED_FOR_DUE_AMOUNT are in the database rather than being hardcoded (lower priority, since we don't currently have plans to change them in the next two years at least)

Note, I have done a minor change to the Dues calculation disclaimer to warn that the calculation is incorrect currently (#8704) which will have to be reversed once 1. is complete (since that will fix the bug).

danieljames-dj:
I've fixed 1 here: https://github.com/thewca/worldcubeassociation.org/pull/8705. This will also predict dues with correct band in our WFC export, so that we won't have a warning there.

For 2, I would suggest having a table `band_details` with following columns:
- band_number
- per_competitor_dues
- competition_percent_dues
- start_date
- end_date

This way, we can store the historical change in dues values as well. @gregorbg could you please approve if the above columns will be fine for long term fix?

Since this is a low-priority change, I would suggest some volunteers to take it up once Gregor approves the above model.

PR #8704, opened 2024-01-04
Add note about Band 1 and 2 Dues estimate being incorrect to disclaimer

EdHollingdale:
Adds a note to the Dues Estimate calculation that it may be significantly overstated (up to 3x the amount) for Band 1 and 2 countries. 

Notes on a permanent fix related to #8159 

Please let me know if I've done this incorrectly - I'm not that familiar with how the website handles translations etc - if there's a better way to do this I'm happy for someone else to edit the disclaimer that way.

### DIFF
--- a/WcaOnRails/config/locales/en.yml
+++ b/WcaOnRails/config/locales/en.yml
@@ -545,7 +545,7 @@ en:
         end_date: ""
         generate_website: ""
         external_registration_page: "The page where the competitors have to register for the competition. Must be a valid http(s) url."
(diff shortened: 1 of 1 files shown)


danieljames-dj:
If https://github.com/thewca/worldcubeassociation.org/pull/8705 gets approved, then there won't be mistake in dues estimation. But leaving this open till the other one is reviewed, if significant change is required in that PR, then this can be merged I guess.

gregorbg:
I have just merged #8705 so I believe this can be closed. Feel free to re-open in case I missed something!

PR #8705, opened 2024-01-04
Dues percent change short-term fix

danieljames-dj:


WcaOnRails/app/models/country_band.rb
```diff
@@ -27,6 +27,16 @@ class CountryBand < ApplicationRecord
   # larger than the due amount per competitor calculated from the competition's country band.
   PERCENT_REGISTRATION_FEE_USED_FOR_DUE_AMOUNT = 0.15
 
+  def self.percent_registration_fee_used_for_due_amount(country_band)
```
gregorbg, line 30: A simple `return 0 if country_band.nil?` at the beginning of the method body can save you the additional if-checks

gregorbg (changes requested):

PR #8706, opened 2024-01-04
Changed probation APIs to roles APIs & cleanup old probation files

danieljames-dj:


WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -183,9 +183,9 @@ def index_for_group_type
 
   def create
     user_id = params.require(:userId)
-    group_id = params.require(:groupId)
+    group_id = params[:groupId] || UserGroup.find_by(group_type: params.require(:groupType)).id
 
-    if group_id.include?("_") # Temporary hack to support some old system roles, will be removed once all roles are
+    if group_id.is_a?(String) && group_id.include?("_") # Temporary hack to support some old system roles, will be removed once all roles are
```
gregorbg, line 188: Not a big fan of doing two separate things in one endpoint.

What about something like `def create_by_id` and `create_by_type` instead? I am open to other names if you can think about a better naming.

As a general rule of thumb: When you need to do type-checks in a weakly-typed programming language, there's at least 80% chance that your code design is flawed.

gregorbg (changes requested):
General question about the context of this PR: Why are you migrating the start probation part but not the end probation part?

This looks especially unintuitive in the tests because this PR deletes the "start probation" cases but keeps the "end probation" cases

WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
danieljames-dj: Basically, `create` method is to create a new role. The final body of `create` method (at the end of roles project) will be something like:

```
def create
  # initialization of all params like `user_id`, `group_id`, etc.
  # creation of metadata based on `group_type`
  role = UserRole.create!(
    ...
  )
  RoleChangeMailer.notify_role_start(role, current_user).deliver_later
  render json: {
    success: true,
  }
end
```

The current code inside `if` block is to create `team_member`, so that once migrated to `user_roles`, the backend can be silently changed to new way.

Also, I didn't change anything inside `if` block. The `group_id` is ideally integer, but for temporary cases, the group_id is string. So to handle those cases, it's converted into string.

danieljames-dj (commented):
The reason why I didn't migrate `end_probation` is because the `end_probation` code has to ideally go to `update` method of `user_roles_controller`. But the `update` method is still bit messy and need some cleanup before migration. So I thought will migrate `start_probation` alone for now, and will migrate `end_probation` once the cleanup is done.

gregorbg:
Hrm, I see your point. But then I don't see the value of holding back the end_probation change now.
Why not just wait until the cleanup is done and then migrate both at once? What benefit do we get from migrating only the start_probation part?

danieljames-dj:
> Hrm, I see your point. But then I don't see the value of holding back the end_probation change now. Why not just wait until the cleanup is done and then migrate both at once? What benefit do we get from migrating only the start_probation part?

Okay, I'm fine with holding this PR for a while. The benefit that I would get by merging this is, in one of my next upcoming PR where I migrate translators to roles & groups, if this PR is already merged I won't have to do much in backend for the API to create new translator. :-)

danieljames-dj:
After some thoughts, cleanup is dependent on many prior migrations and it'll be better if I move the end probation too now, so that this will be helpful for other PRs. Hence I've migrated the end_probation API as well to roles API.

WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
gregorbg: Okay, granted as an exception because this will be cleaned up after the transition is complete :)

gregorbg (approved):

PR #8708, opened 2024-01-05
Make me/personal_records wcif compliant

FinnIckler:
As per call


gregorbg (commented):
Strictly speaking, this is not WCIF compliant. Or at least, it is redundant:

The result of the `to_wcif` call yields an object which in itself contains a `type: "single"` (or `"average"`) entry. So the top-level keys `single: [...]` and `average: [...]` are unnecessary.

See how this happens in the API method that creates the actual WCIF:

```ruby
person_pb = [person&.ranksAverage, person&.ranksSingle].compact.flatten

{
  "personalBests" => person_pb.map(&:to_wcif),
}
```

gregorbg (approved):
Approving despite my previous comment, because the top-level keys were already there before and we should try to maintain backwards compatibility.

PR #8712, opened 2024-01-09
Bump Shakapacker to v7.2.1

gregorbg:
Because both need to be merged together, and Dependabot cannot figure that out


gregorbg:
Closing and reopening after recent GitHub outage to motivate CI

PR #8714, opened 2024-01-11
Allow reordering venues and venue rooms

kr-matthews:
~~Unfortunately I'm unable to run the repo locally, despite hours of trying to figure out why. So I haven't been able to properly test this. I will keep trying, but if someone else is able to do so in the meantime and see if it's working as expected, that would be helpful.~~


kr-matthews:
New "to front" buttons beside "remove" buttons, in venues and rooms (but not the 1st venue/room). For rooms, just icon buttons to save space (2 buttons don't fit on narrower screens, and I'll probably be putting another button or 2 in there in future PRs - if you hover it will show the original text, plus the venues have analogous buttons with the text visible).

image

image

Before:
image

After:
image

kr-matthews:
Realized the icon might imply moving up a single spot, rather than all the way to the front, so picked a new one:
image

WcaOnRails/app/webpacker/components/EditSchedule/EditVenues/VenuePanel.js
```diff
@@ -134,10 +146,13 @@ function VenuePanel({
         </Card.Header>
         <Card.Description>
           <Card.Group itemsPerRow={2}>
-            {venue.rooms.map((room) => (
+            {venue.rooms.map((room, index) => (
               <RoomPanel
                 key={room.id}
                 room={room}
+                sendToFront={
+                  index > 0 ? () => dispatch(reorderRoom(venue.id, index, 0)) : undefined
```
gregorbg, line 154: Why are you passing in the whole action (function) as a prop?
The `RoomPanel` component is able to access the dispatch on its own. I think passing the information which index this is should be enough, so that the panel can figure out the logic on its own.

WcaOnRails/app/webpacker/components/EditSchedule/EditVenues/index.js
```diff
@@ -33,11 +33,12 @@ function EditVenues({
 
       <Segment basic>
         <Card.Group centered itemsPerRow={2}>
-          {wcifSchedule.venues.map((venue) => (
+          {wcifSchedule.venues.map((venue, index) => (
             <VenuePanel
               key={venue.id}
               venue={venue}
               countryZones={countryZones}
+              sendToFront={index > 0 ? () => dispatch(reorderVenue(index, 0)) : undefined}
```
gregorbg, line 41: Again, I feel like `VenuePanel` can figure out the dispatch logic by itself and should only retrieve a minimum amount of information to know _if_ it should trigger/show that logic to the user or not.

gregorbg (changes requested):
Stylistic nitpick about passing around functions in props, otherwise wonderful feature!

WcaOnRails/app/webpacker/components/EditSchedule/EditVenues/VenuePanel.js
kr-matthews: My reasoning was that a room should only directly dispatch actions which affect only that room. This action affects other rooms too, so is a venue-level action.

I don't mind changing it though. That would require passing in the venue id too, in addition to the index (though I could rewrite the reducer to take the room id instead of the venue id, and have it look up which venue the room is in, but that complicates the reducer). What do you think?

kr-matthews:
I've made the suggested changes.

However, the re-ordering doesn't seem to be successfully saved - refreshing the page reverts to the original order. (This was an issue before my most recent changes.) But, if any of the venues/rooms _being_ reordered are _also_ renamed, then those new names do persist. It appears that the arrays being passed to `saveWcif` are in the correct (new) order. So as far as I can tell, the backend is getting the new data in the new order, but is ignoring that order. I don't know anything about the backend/database so will need some help here.

kr-matthews:
Closing due to incompatible approach with existing back-end.

PR #8715, opened 2024-01-11
All regions in Delegate page for admin mode

danieljames-dj:
This page can potentially be the replacement of Delegate admin page once more metadata are added to the delegates.

### DIFF
--- a/WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
+++ b/WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
@@ -180,12 +180,9 @@ class Api::V0::UserRolesController < Api::V0::ApiController
     roles = filter_roles_for_logged_in_user(roles)
 
     # Filter the list based on the other parameters.
-    status = params[:status]
-    is_active = params.key?(:isActive) ? ActiveRecord::Type::Boolean.new.cast(params.require(:isActive)) : nil
     roles = filter_roles_for_parameters(
-      roles: roles,
-      status: status,
-      is_active: is_active,
+      roles: roles, status: params[:status],
+      is_active: params.key?(:isActive) ? ActiveRecord::Type::Boolean.new.cast(params.require(:isActive)) : nil
     )
 
     render json: roles
--- a/WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
+++ b/WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
@@ -180,9 +180,12 @@ class Api::V0::UserRolesController < Api::V0::ApiController
     roles = filter_roles_for_logged_in_user(roles)
 
     # Filter the list based on the other parameters.
+    status = params[:status]
+    is_active = params.key?(:isActive) ? ActiveRecord::Type::Boolean.new.cast(params.require(:isActive)) : nil
     roles = filter_roles_for_parameters(
-      roles: roles, status: params[:status],
-      is_active: params.key?(:isActive) ? ActiveRecord::Type::Boolean.new.cast(params.require(:isActive)) : nil
+      roles: roles,
+      status: status,
+      is_active: is_active,
     )
 
     render json: roles
--- a/WcaOnRails/app/webpacker/components/Delegates/DelegatesOfRegion.jsx
+++ b/WcaOnRails/app/webpacker/components/Delegates/DelegatesOfRegion.jsx
@@ -6,11 +6,17 @@ import {
 import cn from 'classnames';
 import _ from 'lodash';
 import I18n from '../../lib/i18n';
-import { rolesOfGroup } from '../../lib/requests/routes.js.erb';
+import { rolesOfGroup, apiV0Urls } from '../../lib/requests/routes.js.erb';
+import { groupTypes } from '../../lib/wca-data.js.erb';
 import Errored from '../Requests/Errored';
 import Loading from '../Requests/Loading';
 import useLoadedData from '../../lib/hooks/useLoadedData';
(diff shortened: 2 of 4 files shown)


WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -176,7 +176,10 @@ def index_for_group_type
     roles = filter_roles_for_logged_in_user(roles)
 
     # Filter the list based on the other parameters.
-    roles = filter_roles_for_parameters(roles: roles, status: params[:status], is_active: params[:isActive])
+    roles = filter_roles_for_parameters(
+      roles: roles, status: params[:status],
+      is_active: params.key?(:isActive) ? ActiveRecord::Type::Boolean.new.cast(params.require(:isActive)) : nil
```
gregorbg, line 181: Duplicate from translators PR

WcaOnRails/app/webpacker/components/Delegates/DelegatesOfRegion.jsx
```diff
@@ -77,38 +109,26 @@ function DelegatesTable({ delegates, isAdminMode }) {
...
+      ? apiV0Urls.userRoles.listOfGroupType(groupTypes.delegate_regions)
+      : rolesOfGroup(activeRegion.id),
   );
 
   const seniorDelegate = useMemo(
-    () => delegates?.find((delegate) => delegate.metadata.status === 'senior_delegate'),
-    [delegates],
+    () => (isAllRegions
+      ? null
+      : delegates?.find((delegate) => delegate.metadata.status === 'senior_delegate')),
```
gregorbg, line 122: Does `!isAllRegions && delegates?.find(...)` work?

WcaOnRails/app/webpacker/components/Delegates/index.jsx
```diff
@@ -51,8 +52,23 @@ export default function Delegates() {
 
   const [adminMode, setAdminMode] = React.useState(false);
 
+  React.useEffect(() => {
```
gregorbg, line 55: Why do we need this effect to enforce admin mode?

gregorbg (changes requested):

WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
danieljames-dj: Yes, both the PRs needed this. Once any of them merges, I'll rebase the other one and probably gets rebased without any conflict as the changes are same.
I've made the same change I've done in the other PR here.

WcaOnRails/app/webpacker/components/Delegates/DelegatesOfRegion.jsx
danieljames-dj: But that will make it `false` instead of `null` right?

WcaOnRails/app/webpacker/components/Delegates/index.jsx
danieljames-dj: Because this is an admin-only feature, and hence if a person with admin access opens `/delegates#all`, if the admin mode is off, it might look bad.

gregorbg: Okay, makes sense from a design point of view. But then maybe `useEffect` is not the best way to achieve this, because it can uncontrollably overwrite some variable if you accidentally miss some side-effect.

What about the following instead?

```jsx
const [toggleAdmin, setToggleAdmin] = useState(false);

const adminMode = useMemo(() => toggleAdmin || (activeRegion === ALL_REGIONS && loggedInUserPermissions.canViewDelegateAdminPage), [toggleAdmin, activeRegion, loggedInUserPermissions]);
```

IMO that's much easier to read and easier to debug

WcaOnRails/app/webpacker/components/Delegates/DelegatesOfRegion.jsx
gregorbg: Yes, you'd have to adjust the code later down to check for `=== false` instead of `=== null`

WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
gregorbg: Other PR is merged :wink

WcaOnRails/app/webpacker/components/Delegates/index.jsx
kr-matthews: ^ the `useMemo` isn't really needed.

WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
danieljames-dj: Thanks, rebased and confirmed that this file disappeared from the change list :-)

WcaOnRails/app/webpacker/components/Delegates/DelegatesOfRegion.jsx
danieljames-dj: Okay, I've done another approach now - using useCallback. Hope this is fine?

WcaOnRails/app/webpacker/components/Delegates/index.jsx
danieljames-dj: Okay, this looks better. I've implemented it. Also removed `useMemo` and renamed `adminMode` to `isAdminMode` as Kevin suggested.

gregorbg (approved):
Much better now! :smile:

PR #8716, opened 2024-01-11
WST Panel with list of translators

danieljames-dj:


WcaOnRails/app/controllers/translations_controller.rb
```diff
@@ -87,6 +87,30 @@ def update
   }.freeze
   # rubocop:enable Style/NumericLiterals
 
+  def self.translators_to_roles
```
gregorbg, line 90: This should be feasible with `map` and `flat_map`:

```ruby
return VERIFIED_TRANSLATORS_BY_LOCALE.flat_map do |locale, user_ids|
  user_ids.map do |user_id|
    {
      ...role data...
    }
  end
end
```

WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -170,13 +170,21 @@ def index_for_group_type
...
 
     # Filter the list based on the permissions of the logged in user.
     roles = filter_roles_for_logged_in_user(roles)
 
     # Filter the list based on the other parameters.
-    roles = filter_roles_for_parameters(roles: roles, status: params[:status], is_active: params[:isActive])
+    roles = filter_roles_for_parameters(
+      roles: roles,
+      status: params[:status],
+      is_active: params.key?(:isActive) ? ActiveRecord::Type::Boolean.new.cast(params.require(:isActive)) : nil,
```
gregorbg, line 186: Consider outsourcing this into a separate line of code (or two)

gregorbg (changes requested):

WcaOnRails/app/controllers/translations_controller.rb
danieljames-dj: Done, this is nice.

WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
danieljames-dj: Done

gregorbg (approved):

PR #8717, opened 2024-01-11
Include `microservice registrations` in WCIF output

dunkOnIT:
This PR is in pretty rough state, was focused on getting tests to pass and didn't get to the point where I could refactor either the tests or the implementation code (such as it is). If stuff is gross/inefficient, it's because I'm trying to build up to passing tests before refactoring :D 

## General approach
- Create a custom implementation of persons_wcif which calls the registration microservice - this would be invoked when  `uses_v2_registrations` is true
- `registration_service_persons_wcif` would be generate a persons WCIF as follows:
  - call the GET `/registrations` for the comp
  - use the output to build person['registration'] hash for each person
  - use the user_ids to query Users for the rest of the persons WCIF data - re-using as much of the existing persons_wcif code as possible (most of those associations _should_ be possible to get from `Person` model instead of `Registration`?)

## Useful things this does
- Adds a Factory for the `/registrations` microservice endpoint, which returns a JSON that can be mocked using WebMock

## Where I got to
- Returning user_ids from the `/registrations` mock, and building the `registration` hash in the person WCIF

## What I got stuck on 
Trying to get User objects based on the list of user_ids - this would result in database locks that I haven't been able to figure out yet

### DIFF
--- a/WcaOnRails/.env.development
+++ b/WcaOnRails/.env.development
@@ -21,3 +21,4 @@ DATABASE_PASSWORD=
 MAILCATCHER_SMTP_HOST=localhost
 CRONJOB_POLLING_SECONDS=0
 JWT_KEY=jwt-development-secret
+WCA_REGISTRATIONS_URL=https://registration.worldcubeassociation.org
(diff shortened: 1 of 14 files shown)


FinnIckler:
Redoing this myself/Gregor

PR #8720, opened 2024-01-12
Migrate translators to roles

danieljames-dj:


WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -20,7 +20,7 @@ class Api::V0::UserRolesController < Api::V0::ApiController
         if group.group_type == UserGroup.group_types[:delegate_probation]
           current_user.can_manage_delegate_probation?
         else
-          false # Don't accept any other hidden groups.
+          group.group_type == UserGroup.group_types[:translators] # Don't accept any other hidden groups.
```
gregorbg, line 23: Feels like this line has a "hidden agenda". I would feel more comfortable with a switch-case statement on `group.group_type`.

WcaOnRails/app/jobs/sync_mailing_lists_job.rb
```diff
@@ -26,8 +26,8 @@ def perform
     GsuiteMailingLists.sync_group("results@worldcubeassociation.org", Team.wrt.current_members.includes(:user).map(&:user).map(&:email))
     GsuiteMailingLists.sync_group("software@worldcubeassociation.org", Team.wst.current_members.includes(:user).map(&:user).map(&:email))
     GsuiteMailingLists.sync_group("software-admin@worldcubeassociation.org", Team.wst_admin.current_members.includes(:user).map(&:user).map(&:email))
-    translators = User.where(id: TranslationsController::VERIFIED_TRANSLATORS_BY_LOCALE.values.flatten)
-    GsuiteMailingLists.sync_group("translators@worldcubeassociation.org", translators.map(&:email))
+    translator_users = UserGroup.translators.map(&:roles).flatten.map(&:user)
```
gregorbg, line 29: I feel like any given user group (same applies to Delegates, Team Members, etc...) should have an easy convenience method for getting the users inside that group.
Why is the `flatten` even necessary? And if there is a good reason why it is necessary, why aren't you just using `flat_map(&:roles)` directly?

WcaOnRails/db/migrate/20240112161627_create_translators_group.rb
```diff
@@ -0,0 +1,51 @@
...
+  # rubocop:enable Style/NumericLiterals
+  def change
+    user_group = UserGroup.create!(name: 'Translators', group_type: :translators, is_active: true, is_hidden: true)
+    CreateTranslatorsGroup::VERIFIED_TRANSLATORS_BY_LOCALE.flat_map do |locale, user_ids|
+      user_ids.map do |user_id|
+        metadata = GroupsMetadataTranslators.create!(locale: locale)
+        UserRole.create!(
+          user_id: user_id,
+          group_id: user_group.id,
+          start_date: '1900-01-01',
```
gregorbg, line 45: Hrm, I feel like there should be a more realistic start date. Can you do some (very basic! No need to dig super deep) `git blame` to find out when the feature was introduced? Or maybe look when the first `thewca-bot` PR was merged?

At the _very_ least, use the founding date of the WCA or something. It is highly improbable we had translators before the WCA even existed.

gregorbg (changes requested):

WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
danieljames-dj: Okay, this was a mistake by me. Hidden groups are not supposed to be sent to users who doesn't have permission. If I wanted to give it to people who have permission, I should have made the group non-hidden. I've also made it to switch block.

WcaOnRails/app/jobs/sync_mailing_lists_job.rb
danieljames-dj: Oh I didn't knew about `flat_map`. I changed it to `flat_map` now. I also created `users` method in `user_group`. The reason why flatten is there is because `UserGroup.translators` will give the list of user_groups which belongs to `group_type` translators. Currently there is only one group under this, but I thought still using `flatten` is better in long term.

WcaOnRails/db/migrate/20240112161627_create_translators_group.rb
danieljames-dj: First commit by `thewca-bot` is on 2017-01-30. I've given that for now.

WcaOnRails/app/models/user_group.rb
```diff
@@ -44,6 +44,10 @@ def roles
     end
   end
 
+  def users
```
gregorbg, line 47: For future reference, when the migration is complete: (Not relevant to merging this PR!)

I think this can be done even smarter in terms of SQL optimization. When we fully rely on the Groups/Roles model, there will be a `has_many :roles` statement for the Rails database binding.
Then, assuming that the `UserRole` model has a `belongs_to :user` attribute, you can use it here by stating `has_many :users, through: :roles`.

Doing it this way will help you tremendously to optimize DB performance later on.

WcaOnRails/app/jobs/sync_mailing_lists_job.rb
```diff
@@ -26,7 +26,7 @@ def perform
     GsuiteMailingLists.sync_group("results@worldcubeassociation.org", Team.wrt.current_members.includes(:user).map(&:user).map(&:email))
     GsuiteMailingLists.sync_group("software@worldcubeassociation.org", Team.wst.current_members.includes(:user).map(&:user).map(&:email))
     GsuiteMailingLists.sync_group("software-admin@worldcubeassociation.org", Team.wst_admin.current_members.includes(:user).map(&:user).map(&:email))
-    translator_users = UserGroup.translators.map(&:roles).flatten.map(&:user)
+    translator_users = UserGroup.translators.flat_map(&:users)
```
gregorbg, line 29: I am still not 100% happy because types are not obvious in Ruby. Calling `UserGroup.translators`, I would expect the result to be **ONE** group that is called "Translators" and which reprensents all translators.

I know that we have decided to implement translators for different languages through different sub-groups, but for somebody who wasn't involved in that decision, this variable name `UserGroup.Translators` might not at all be obvious.

What about calling this `UserGroup.translator_groups`?

gregorbg (changes requested):

WcaOnRails/app/jobs/sync_mailing_lists_job.rb
danieljames-dj: `UserGroup.translator_groups` looks like a good idea. I guess this is applicable to some other group type functions as well. Since those are not relevant to this PR, I'll change it there once I touch that code for some other PR.

gregorbg: Nitpick: Now you've changed the name below in the `UserGroup` class, but you haven't changed it here where you're calling it

gregorbg (changes requested):

WcaOnRails/app/jobs/sync_mailing_lists_job.rb
danieljames-dj: Sorry my bad. I've changed it now. Also rebased.

gregorbg (approved):

gregorbg:
Deployed and migrated.

PR #8721, opened 2024-01-12
Align competition schedule cubing icons

### LINKED ISSUE #6704: Centering of text and event icon on schedule looks weird
Screenshot_20220218-125740
See specifically clock final, but they're mostly all weird.

kcmyang:
### Description
- Closes #6704 
- Makes the "ALL" toggle button in the competition schedule events filter the same size as the cubing icons
- Vertically centres and aligns all the buttons in the filter to a grid

### Tested on
- Chromium
- Firefox
- Safari

### More details
- The icons in the events table were off-centre because the icon (in the `i::before`) was too large for the container (`<i>`). I set the width of the container to `auto` to address this.
  - The container height is still smaller than the icon height, but setting the height to `auto` introduces a tiny amount of extra vertical space in the table (more than what the left/right side text demands) so I opted to leave the height alone.
- The cubing icons in the events filter currently have a height of 25px, but again the containers were not large enough to hold the icons, causing them to be misaligned. I addressed this by:
  - setting an absolute height and width on the icon containers (5px "margin" on each side for total height/width of 35px)
  - unsetting the line height (I believe this is inherited from the icon library's css; it causes a bit of vertical shifting if not unset)
  - setting the margin to 0 to undo the 2px right margin that was inherited from somewhere
- To make the "ALL" toggle better match the icons, I made it the same size as the icons (25px square) with the same 5px margin (using actual css `margin`).
- This PR vertically centres the buttons in the events filter using a flexbox. I think this makes the intent more clear.
- The `.panel-body` no longer needs extra bottom padding because its contents do not overflow their container.

### Screenshots
Each image has production on the left and this PR on the right.

<details><summary>medium/large size</summary>

Screenshot 2024-01-12 at 5 33 29 PM
</details> 

<details><summary>sm size</summary>

Screenshot 2024-01-12 at 5 34 01 PM
</details> 

<details><summary>xs size</summary>

Screenshot 2024-01-12 at 5 34 40 PM
</details> 

<details><summary>sm size, with artificial * { outline: 1px solid blue; } for better visualisation</summary>

Screenshot 2024-01-12 at 5 36 02 PM
</details>


gregorbg:
Awesome, thanks for your contribution and massive props for your thorough investigation!

gregorbg (approved):

PR #8722, opened 2024-01-13
Delegate admin page migration to react

danieljames-dj:
Some points to be noted before reviewing:
- The main reason for this migration now is, it's bit difficult to modify `stats.html.erb` with delegate roles model, and hence this is a major blocker for starting the migration of delegates to roles.
- We are using `extraMetadata` to load additional metadata. This is because those metadata are costly to compute. This is temporary solution anyway because once we migrate delegates completely to roles, we will have the metadata column to store these metadata and we can update it every 24 hours via cronjob to make it updated.
- There are couple of known issues:
  - The link to 'history' is not center aligned. I couldn't fix it.
  - The table doesn't have sorting feature like in old stats table. If this comes as a requirement later, will take it up then.


WcaOnRails/app/models/user.rb
```diff
@@ -1289,6 +1289,22 @@ def delegate_role
...
+    {
+      end_date: nil,
+      is_active: true,
+      group: self.region,
+      user: self,
+      metadata: {
+        status: self.delegate_status,
+        location: self.location,
+        first_delegated: self.actually_delegated_competitions.min_by(&:start_date)&.start_date,
+        last_delegated: self.actually_delegated_competitions.max_by(&:start_date)&.start_date,
```
gregorbg, line 1302: Rails has `actually_delegated_competitions.minimum(:start_date)` and `.maximum(:start_date)` to compute these directly on SQL level

WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -149,7 +149,14 @@ def index_for_group_type
...
     if group_type == UserGroup.group_types[:delegate_regions]
-      roles.concat(User.where.not(delegate_status: nil).map(&:delegate_role))
+      # extra_metadata adds some extra metadata to the roles. These metadata computation is bit costly,
+      # so we only compute it when extra_metadata is true.
+      extra_metadata = params.key?(:extraMetadata) ? ActiveRecord::Type::Boolean.new.cast(params.require(:extraMetadata)) : nil
+      if extra_metadata
+        roles.concat(User.where.not(delegate_status: nil).map(&:delegate_role_with_extra_metadata))
+      else
+        roles.concat(User.where.not(delegate_status: nil).map(&:delegate_role))
+      end
```
gregorbg, line 159: Why are you doing `User.where.not(delegate_status: nil)` here, instead of using the available scope `User.delegates`?

gregorbg (changes requested):
If you use the `.includes(:actually_delegated_competitions)` trick (along with computing the min/max on SQL level already), the "extra metadata" shouldn't be very costly to compute.
You could then consider just leaving it as "one metadata" and kick out the "extra" quirk.

danieljames-dj (commented):
Thanks Gregor. But since we are calling it inside user.rb, can we use `.includes(:actually_delegated_competitions)`?

gregorbg:
> Thanks Gregor. But since we are calling it inside user.rb, can we use `.includes(:actually_delegated_competitions)`?

You don't have to use it inside user.rb.

In the controller, you can use it like `User.delegates.includes(...).map(&:metadata)` and within the `def metadata` method in `user.rb` you just assume that "whoever calls this method will be smart enough to call .includes(...) before calling me"

danieljames-dj:
> > Thanks Gregor. But since we are calling it inside user.rb, can we use `.includes(:actually_delegated_competitions)`?
> 
> You don't have to use it inside user.rb.
> 
> In the controller, you can use it like `User.delegates.includes(...).map(&:metadata)` and within the `def metadata` method in `user.rb` you just assume that "whoever calls this method will be smart enough to call .includes(...) before calling me"

Oh okay, done.

gregorbg (approved):

PR #8724, opened 2024-01-14
Cleanup of unused subtext_for_officer code

danieljames-dj:
The usage of the code was removed in https://github.com/thewca/worldcubeassociation.org/pull/7277. Found this while working on another PR.

### DIFF
--- a/WcaOnRails/app/helpers/static_pages_helper.rb
+++ b/WcaOnRails/app/helpers/static_pages_helper.rb
@@ -23,14 +23,6 @@ module StaticPagesHelper
     end
   end
 
-  def subtext_for_officer(user)
-    positions = user.current_teams.select { |team| Team.all_officers.include? team }.map(&:name)
-    if user.team_leader?(Team.wfc)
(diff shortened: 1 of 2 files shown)


gregorbg (commented):
I am not a big friend of deleting tests without replacement.

How is the functionality that was previously being provided by `subtext_for_officer` now provided in React? And how is that part being tested?

danieljames-dj:
> I am not a big friend of deleting tests without replacement.
> 
> How is the functionality that was previously being provided by `subtext_for_officer` now provided in React? And how is that part being tested?

The subtext is now re-written in React in UserBadge.js. This is not tested as well. I assume we currently don't have a way to test a function in React right?

gregorbg:
Hm, we don't have any direct React testing framework.
But we do some tests by fully rendering the page and just treating it as some HTML we can test, without considering whether that page was created through React or HTML ERB. I can however agree that this is not necessary for a small feature like this.

gregorbg (approved):

PR #8725, opened 2024-01-15
Migrate Badges to react

danieljames-dj:
Badges is a section which had a lot of dependencies on old system roles and hence was one of the blocker for migrating delegates/teams-committees/councils to roles. With this PR, we are more closer towards the migration.


WcaOnRails/app/webpacker/lib/requests/routes.js.erb
```diff
@@ -148,7 +152,7 @@ export const apiV0Urls = {
     resetClaimCount: (wcaId) => `<%= CGI.unescape(Rails.application.routes.url_helpers.api_v0_wrt_person_reset_claim_count_path("${wcaId}"))%>`,
   },
   userRoles: {
-    listOfUser: (userId, {isActive, isGroupHidden, status}) => `<%= CGI.unescape(Rails.application.routes.url_helpers.api_v0_index_for_user_path('${userId}')) %>?${jsonToQueryString({ isActive, isGroupHidden, status })}`,
+    listOfUser: (userId, sort, {isActive, isGroupHidden, status}) => `<%= CGI.unescape(Rails.application.routes.url_helpers.api_v0_index_for_user_path('${userId}')) %>?${jsonToQueryString({ sort, isActive, isGroupHidden, status })}`,
```
kr-matthews, line 155: Why not include `sort` in the existing object parameter? Then there wouldn't need to be clarifying comments every time this is called.

danieljames-dj: The object parameters are basically the parameters which are part of the role. Since sort is different, this is included separately. Another reason can be to match how `listOfGroup` is initialized.

WcaOnRails/app/webpacker/components/Panel/Leader/GroupsManager.jsx
```diff
@@ -44,6 +44,7 @@ function GroupTable({ groupId }) {
 export default function GroupsManager({ loggedInUserId }) {
   const { data: roles, loading, error } = useLoadedData(apiV0Urls.userRoles.listOfUser(
     loggedInUserId,
+    null, // Sort params
     { isActive: true, isGroupHidden: false, status: 'leader' },
   ));
```
kr-matthews, line 49: Then I would recommend either creating a new parameter object so that `{ sort: null }` can be passed in, or extracting the sort as a constant each time so that the const name replaces the need for a comment:
```suggestion
  const noSort = null;
  const { data: roles, loading, error } = useLoadedData(apiV0Urls.userRoles.listOfUser(
    loggedInUserId,
    noSort,
    { isActive: true, isGroupHidden: false, status: 'leader' },
  ));
```

danieljames-dj: Okay, in that case I think I'll have a sort here as well so that the list of groups come in alphabetical order. I've added sort now. :-)

WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -8,11 +8,23 @@ class Api::V0::UserRolesController < Api::V0::ApiController
     end
   end
 
-  # The order in which the roles should be sorted.
-  STATUS_SORTING_ORDER = ['leader', 'senior_member', 'member'].freeze
-
-  private def status_sort_rank(status)
-    STATUS_SORTING_ORDER.find_index(status) || STATUS_SORTING_ORDER.length
+  private def sort_rank(group_type, status)
```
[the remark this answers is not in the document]
danieljames-dj: Actually doing two sorts will not work because status of two different group types cannot be compared - example board's member should always be higher than team_committee's senior member.

But I was able to find another solution which should be clean. Now, I don't have 'rank', instead sorts based on multiple parameters: lead, eligible_voter, groupTypeRank, status and finally groupName to sort alphabetically if everything else is same.

WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -87,14 +101,17 @@ class Api::V0::UserRolesController < Api::V0::ApiController
...
 
-    if user.admin? || user.board_member?
+    if user.board_member?
       roles << {
         group: {
-          id: user.admin? ? 'admin' : 'board',
-          name: user.admin? ? 'Admin Group' : 'Board Group',
-          group_type: UserGroup.group_types[:teams_committees],
+          id: 'board',
+          name: 'WCA Board of Directors',
```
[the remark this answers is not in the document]
danieljames-dj: Actually here we are giving something that is in database. Since we have only english in database, I'm not changing here. Instead, now I have added the localization in Badges.jsx based on the group type. Hope that is fine.

WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -104,6 +121,51 @@ class Api::V0::UserRolesController < Api::V0::ApiController
...
+      elsif officer_team == Team.secretary
+        status = 'secretary'
+      elsif officer_team == Team.vice_chair
+        status = 'vice_chair'
+      end
+      if user.team_member?(officer_team)
+        roles << {
+          group: {
+            id: 'officers',
+            name: 'WCA Officers',
```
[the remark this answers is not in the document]
danieljames-dj: Addressed above

WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -104,6 +121,51 @@ class Api::V0::UserRolesController < Api::V0::ApiController
...
+          },
+        }
+      end
+    end
+
+    if Team.wfc.current_members.select(&:team_leader).map(&:user).include?(user)
+      roles << {
+        group: {
+          id: 'officers',
+          name: 'WCA Officers',
```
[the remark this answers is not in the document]
danieljames-dj: Addressed above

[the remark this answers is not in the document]
danieljames-dj: Addressed above

WcaOnRails/app/webpacker/lib/wca-data.js.erb
```diff
@@ -146,6 +146,8 @@ export const delegateRegionsStatus = {
   senior_delegate: 'senior_delegate',
 };
 
+export const userRolesSortParams = <%= Api::V0::UserRolesController::SORT_PARAMS.each_with_object({}) { |(key, _), new_hash| new_hash[key] = key }.to_json %>;
```
gregorbg, line 149: WAIT WHAT IN ALL HEAVENS NAME IS THIS??!?!???!?!!

gregorbg (changes requested):

WcaOnRails/app/webpacker/lib/wca-data.js.erb
danieljames-dj: These are the sortParams used to sort userRoles. Didn't get a better name 😅 Do you have any idea?

gregorbg: The name is fine. I am very, very confused by the "each_with_hash" hack. No idea what it does or why it's necessary, but I'm 99.97% sure it's overly complicated

danieljames-dj: Reason why it's necessary: I can use it in React instead of directly using the string. I'm trying to create following JSON:
```
{
  startDate: startDate,
  status: status,
  ...
}
```

Yes, I agree it's a bit complex. I've made it a bit more clean now. Hope it's good now.

gregorbg: There's an even simpler way:
```ruby
array_of_values.index_by { |v| doe_something_with_value(v) }
```

gregorbg: Why does it need to have the shape `{ key: key }`?
Can't you just access `SORT_PARAMS.keys` and call it a day?

danieljames-dj: But it cannot be access like a map right? Like we do `groupTypes.board`, etc.

gregorbg: Yes, you cannot access a list of keys like a map.
But if your goal is to access it like a map, and the value is the same like the key you're accessing, then you already know what the result will be based on the access you're trying.

gregorbg: If you really construct this like `const foo = { key: key }` then you already know that `foo.some_crazy_team` will definitely return `some_crazy_team`

danieljames-dj: Okay, so hard-coding the values directly as a string in the javascript code is a better approach?

gregorbg: Questioning why and where you need these values is a better approach.
Why does it have to be an object that has the same keys an values?

WcaOnRails/app/webpacker/components/Persons/Badges.jsx
```diff
@@ -0,0 +1,81 @@
...
+      url: teamsCommitteesPageUrl,
+    };
+  }
+  return {};
+}
+
+export default function Badges({ userId }) {
+  const { data } = useLoadedData(apiV0Urls.userRoles.listOfUser(
+    userId,
+    [sortParams.lead, sortParams.eligibleVoter, sortParams.groupTypeRank, sortParams.status, sortParams.groupName].join(','), // Sort params
```
gregorbg, line 51: Re the other discussion: If you access the keys here, you already know that the result of `sortParams.lead` will be `lead`.

So effectively, it is already hard-coded now because you "magically" know that the sort param `eligibleVoter` exists.
The only good reason to export this constant from Rails to JS is to let JS choose which sort params it should pick **without knowing what sort params exist**. But in the way you're currently using the code, all that's happening is an over-glorified version of "return a string that is called exactly like the variable name which I chose"

danieljames-dj: Okay got it, makes sense. What I was thinking is, if I do something like `sortParams.nonExistantParam`, it will show some error, but we don't get error in frontend via lint or anything, so it doesn't makes sense to have it. I've now removed it.

gregorbg: That is a very good thought and I appreciate the idea. However our infrastructure doesn't support it very well right now and because you tried so hard to squeeze it into a non-typed JS stack you ended up with a super-complicated implementation that took me 6 or so GitHub comments to figure out what's even going on.

gregorbg: Super-nice idea in theory, but please don't break a leg with implementing it, especially in a legacy tech stack that has lots of technical debt.

If your implementation contains `each_with_hash`, it's already a very high chance that something is not supported in the way you want. If you even have to pass lambda params `|(key, value), new_hash|` into the `each_with_hash`, then you know there's **definitely** an un-supported feature that you're trying too hard to figure out ;)

gregorbg (approved):

WcaOnRails/app/webpacker/components/Persons/Badges.jsx
danieljames-dj: Got it, thanks a lot for explaining. 🙂

PR #8729, opened 2024-01-15
Update to Bundler 2.5 for Ruby Gems

gregorbg:
This caught my attention while rolling out the Ruby 3.3.0 update.
Quite frankly, I'm surprised that Dpeendabot doesn't handle this.


gregorbg:

PR #8732, opened 2024-01-16
Hotfix for the delegate probation check issue

danieljames-dj:


danieljames-dj:
> Actually, does this hotfix make sure that only users who have the CURRENT role are considered? Reading the code, it looks like "any user who has ever been part of the group"

Oops, you are right. I've just added active users. Pushed the code since it's hotfix, while the test runs, I'll test the functionality in local.

danieljames-dj:
Tested in local, did one more minor fix. Should work now.

danieljames-dj:
@gregorbg Since I've made more changes after you approved, can you please take a look one more time?
After checking more, looks like this is not actually blocking WCAT from announcing the competition, so we have time for more changes if needed.

gregorbg (approved):

gregorbg:
Looking good now! Later down the line we asolutely need to improve the SQL performance but for the time being it will be fine

PR #8733, opened 2024-01-17
Added subregion field to Delegate change form

danieljames-dj:


WcaOnRails/app/webpacker/components/RolesTab/DelegateForm.jsx
```diff
@@ -7,10 +7,24 @@ export default function DelegateForm({
...
+    subregionsList = subRegions[formValues.regionId] || [];
+    if (subregionsList.length > 0) {
+      subregionsList.push({
+        name: 'None',
+        value: null,
+      });
+    }
+    return subregionsList;
+  }, [formValues.regionId, subRegions]);
+
```
gregorbg, line 27: Not a big fan of `let` in JS. How about the following flow?

```suggestion
  const subRegionsOfSelectedRegion = React.useMemo(() => {
    if (!formValues.regionId) return [];

    const subregionsList = subRegions[formValues.regionId] || [];

    if (subregionsList.length > 0) {
      return [...subregionsList, {
        name: 'None',
        value: null,
      }];
    }

    return subregionsList;
  }, [formValues.regionId, subRegions]);
```

WcaOnRails/app/webpacker/components/RolesTab/RoleForm.jsx
```diff
@@ -23,19 +25,58 @@ const delegateStatusOptions = ['trainee_delegate', 'candidate_delegate', 'delega
 
 export default function RoleForm({ userId, isActiveRole }) {
   const { data, loading, error } = useLoadedData(roleDataUrl(userId, isActiveRole));
+  const regionsFetching = useLoadedData(apiV0Urls.userGroups.list(groupTypes.delegate_regions));
```
gregorbg, line 28: Why are you not destructuring this into `data, loading, error` like you do with the role data directly above?

WcaOnRails/app/webpacker/components/RolesTab/RoleForm.jsx
```diff
@@ -23,19 +25,58 @@ const delegateStatusOptions = ['trainee_delegate', 'candidate_delegate', 'delega
...
+  const regions = React.useMemo(() => regionsFetching.data?.filter(
+    (group) => !group.parent_group_id,
+  ), [regionsFetching.data]);
+
+  const subRegions = React.useMemo(() => {
+    const subRegionsList = regionsFetching.data?.filter((group) => group.parent_group_id) || [];
+    return Object.groupBy(subRegionsList, (group) => group.parent_group_id);
+  }, [regionsFetching.data]);
+
+  React.useEffect(() => {
```
gregorbg, line 44: This is a very complex effect body. Why is it necessary and are you convinced that there is no way to achieve the outcome without `useEffect`?

WcaOnRails/app/webpacker/components/RolesTab/RoleForm.jsx
```diff
@@ -23,19 +25,58 @@ const delegateStatusOptions = ['trainee_delegate', 'candidate_delegate', 'delega
...
   React.useEffect(() => {
-    setFormValues({
-      delegateStatus: delegateStatusOptions[0],
-      location: '',
-      ...(data?.roleData || {}),
-    });
-  }, [data]);
+    if (formValues.regionId && formValues.subregionId) {
+      const subRegionList = subRegions[formValues.regionId] || [];
+      if (!subRegionList.find((subregion) => subregion.id === formValues.subregionId)) {
```
gregorbg, line 72: Can you extract the `find` statement inside the if into its own variable and give it a proper name?

gregorbg (changes requested):

WcaOnRails/app/webpacker/components/RolesTab/RoleForm.jsx
danieljames-dj: This is necessary and can be done only inside useEffect because roleData.regionId can either be region ID or subregion ID. I've added more comments in the code to get more clarity on it. I couldn't find a way to do it without `useEffect`.

WcaOnRails/app/webpacker/components/RolesTab/RoleForm.jsx
danieljames-dj: TBH, I was lazy to do it 😁 Did it now.

WcaOnRails/app/webpacker/components/RolesTab/RoleForm.jsx
```diff
@@ -23,19 +25,69 @@ const delegateStatusOptions = ['trainee_delegate', 'candidate_delegate', 'delega
...
+
   React.useEffect(() => {
-    setFormValues({
-      delegateStatus: delegateStatusOptions[0],
-      location: '',
-      ...(data?.roleData || {}),
-    });
-  }, [data]);
+    if (formValues.regionId && formValues.subregionId) {
+      const subRegionList = subRegions[formValues.regionId] || [];
```
kr-matthews, line 79: There's inconsistent capitalization of `subregion...` and `subRegion` in various places.

WcaOnRails/app/webpacker/components/RolesTab/RoleForm.jsx
kr-matthews: (It's unfortunate that 2 things which may not be the same are both called `regionId`.)

An effect does seem necessary given the existing setup, because we need to set the initial form values once the data has loaded, and also need to allow the user to update those values via the form. If we really didn't want an effect, it would be possible to remove it by creating a separate component which is only rendered once the data has loaded, and takes the data as a parameter. That is, instead of having this after the loading/error returns
```
  return (
    <Form...
```
do this instead
```
  return (
    <LoadedRoleForm roleData={data.roleData} />
  )
```
where all the `<Form...` has been moved inside this new component `LoadedRoleForm`, along with the form state `const [formValues, setFormValues] = React.useState(calculateInitialValues(roleData));` (note the new initial state, rather than just `{}`).

Is that worth doing? :shrug: There are other nearby components that don't do it.

danieljames-dj: I've changed the code a bit. I hope it's cleaner now. :-) Removed lots of complexities hopefully.

On a side note, if "GroupsManager with edit-features" (PR in progress) becomes successful and if the leaders like it, maybe we can bring the same for Senior Delegates as well, so that we don't have to think about any complexities like it.

gregorbg (approved):

PR #8735, opened 2024-01-17
Add OIDC compliant ouath using Doorkeeper

FinnIckler:
Configuration from https://github.com/doorkeeper-gem/doorkeeper-openid_connect

I'm testing this on our docker staging container


FinnIckler:
This PR is technically done, but we need to think about how we make sure that the OAuth Application persists on staging to authenticate to PMA

gregorbg:
> This PR is technically done, but we need to think about how we make sure that the OAuth Application persists on staging to authenticate to PMA

Use the `example-application-id` ;)

gregorbg (approved):

PR #8734, opened 2024-01-17
Renamed GroupsMetadataTranslators to RolesMetadataTranslators

danieljames-dj:

PR #8736, opened 2024-01-17
Added feature to end a translator role

danieljames-dj:


danieljames-dj:
Marking this PR as draft because this PR will fail test cases when [other PR](https://github.com/thewca/worldcubeassociation.org/pull/8734) is merged, but since there are conflicts this PR can still be merged and if both get approved & merged at same time, it might cause issue.

WcaOnRails/app/views/role_change_mailer/notify_role_end.erb
```diff
@@ -0,0 +1,10 @@
+<p>The role has been removed for <%= @role.user.name %> in <%= @group_type_name %> on <%= @role.end_date %>.</p>
```
gregorbg, line 1: You say "The role", but which role is it?
I feel like in addition to the metadata below, we should also include some "general role title" instead of just silently assuming everybody knows which role we're talking about

gregorbg (changes requested):

WcaOnRails/app/views/role_change_mailer/notify_role_end.erb
danieljames-dj: Current email is as below:
```
Subject: Role removed for Daniel in translators
Body:
The role has been removed for Daniel in translators on 20-01-2024.
The change has been executed by Gregor.
More details of the role were:
Locale: en
```

This already has enough details right? Should we add any more details?

gregorbg: Yup, okay!

gregorbg (approved):

gregorbg:
Can be rebased now :)

danieljames-dj:
> Can be rebased now :)

Thanks, will merge this. Actually I made this as draft as I thought of merging https://github.com/thewca/worldcubeassociation.org/pull/8734 first. But I think we need to discuss a bit on the metadata part via Slack. So I've marked that PR as draft and undrafted this PR.

PR #8744, opened 2024-01-20
Column Sorting for Comments On Registrations Page (#8743)

### LINKED ISSUE #8743: Column Sorting for Comments On Edit Registrations Page
Simple Feature Request from Michael Conard in the WCA Staff and Organizers Unofficial Discord Server. I think it's a one line fix that I'll open a PR for in a few minutes, but I can't necessarily check because I'm unable to get a working environment on my computer. Basically just adding the sort button for the comments column like it already does for birth date, WCA ID, etc.

Carter-Thomas:
Fixes #8743, I haven't tested this on my own environment yet so if someone would be willing to do that it'd be greatly appreciated.


kr-matthews (approved):
Works for me locally.

PR #8750, opened 2024-01-22
Show past roles in roles tab

danieljames-dj:


WcaOnRails/app/webpacker/components/RolesTab/index.jsx
```diff
@@ -27,105 +21,67 @@ import { groupTypes, delegateRegionsStatus } from '../../lib/wca-data.js.erb';
...
-
-  function getRoleDescription(role) {
-    let roleDescription = '';
-    if (role.metadata.status) {
-      roleDescription += `${I18n.t(`enums.user.role_status.${role.group.group_type}.${role.metadata.status}`)}, `;
-    }
-    roleDescription += role.group.name;
-    return roleDescription;
+  function hasAtLeastOneRole() {
+    return activeRoles?.length > 0 || pastRoles?.length > 0;
```
kr-matthews, line 55: Couldn't this be a const?

WcaOnRails/app/webpacker/components/RolesTab/index.jsx
```diff
@@ -27,105 +21,67 @@ import { groupTypes, delegateRegionsStatus } from '../../lib/wca-data.js.erb';
...
-                    )}
-                  </List.Content>
-                </List.Item>
-              ))}
-            </List>
+            {activeRoles?.length > 0 && (
+            <ActiveRoles
+              activeRoles={activeRoles}
+              setOpen={setOpen}
+            />
```
kr-matthews, line 70: Should be indented.

WcaOnRails/app/webpacker/components/RolesTab/index.jsx
```diff
@@ -27,105 +21,67 @@ import { groupTypes, delegateRegionsStatus } from '../../lib/wca-data.js.erb';
...
-  if (roleListFetch.loading || loading) return <Loading />;
-  if (roleListFetch.error) return <Errored />;
+  if (activeRolesLoading || pastRolesLoading) return <Loading />;
+  if (activeRolesError || pastRolesError) return <Errored />;
 
   return (
     <>
-      {roleListFetch.data?.length > 0
+      {hasAtLeastOneRole()
         ? (
```
kr-matthews, line 64: This ternary might be cleaner as
```
<>
  {activeRoles?.length > 0 && (
    ...
  )}
  {pastRoles?.length > 0 && (
    ...
  )}
  {hasNoRoles && (
    ...
  )}
</>
```
Note the use of `hasNoRoles` rather than `hasAtLeastOneRole`.

WcaOnRails/app/webpacker/components/RolesTab/index.jsx
danieljames-dj: Yes, that's better. Done, thank you.

WcaOnRails/app/webpacker/components/RolesTab/index.jsx
danieljames-dj: Looks good, done.

WcaOnRails/app/webpacker/components/RolesTab/index.jsx
danieljames-dj: Made it in single line as it is now small enough to fit within recommended number of characters of airbnb-eslint (I think 80 or 100)

WcaOnRails/app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -96,8 +96,8 @@ def self.status_sort_rank(role)
       # operator to access the parameters.
       (
         (status.present? && status != (is_actual_role ? role.metadata.status : role[:metadata][:status])) ||
-        (is_active.present? && is_active != (is_actual_role ? role.is_active? : role[:is_active])) ||
-        (is_group_hidden.present? && is_group_hidden != (is_actual_role ? role.group.is_hidden : role[:group][:is_hidden]))
+        (!is_active.nil? && is_active != (is_actual_role ? role.is_active? : role[:is_active])) ||
+        (!is_group_hidden.nil? && is_group_hidden != (is_actual_role ? role.group.is_hidden : role[:group][:is_hidden]))
```
gregorbg, line 100: Why are you changing this from `present?` to `!nil`?
And why are you changing it only for `is_active` and `is_group_hidden`, but not for `status` one line above?

danieljames-dj: I need to evaluate the expression to the right side of `&&` only if the corresponding parameter (status, is_active, is_group_hidden) is present. If status is present, then it's going to be a string so `status.present?` will be `true` always. But other two parameters are boolean, hence can be `false` as well. Unfortunately rails is returning `false` for `x.present?` if `x` is `false`.

gregorbg (approved):

PR #8753, opened 2024-01-23
Add Modal CSS overrides between SemUI/Bootstrap

gregorbg:
Removed this one a bit too prematurely :(


gregorbg:
Context: The CSS rule that is being added in this PR previously existed in our codebase. It was removed in https://github.com/thewca/worldcubeassociation.org/pull/8413/files#diff-c76c75c02227811a2f73a8d125563c231c36b18011a9e2771b4fe4e313413c8e because I thought there were no more remnants of bootstrap. I was wrong :(

FinnIckler (approved):

PR #8754, opened 2024-01-23
EditEvents: Rely on SemUI styles instead of custom CSS as much as possible

gregorbg:
Before:
image

After:
image

Context: We need a similar, shared "one box per event" panel for scrambles soon, and I don't want to rely on hard-coded `fontSize: 0.75em` and similar so much

### DIFF
--- a/WcaOnRails/app/webpacker/components/EditEvents/EventPanel/index.js
+++ b/WcaOnRails/app/webpacker/components/EditEvents/EventPanel/index.js
@@ -114,6 +114,7 @@ export default function EventPanel({
 
   return (
     <Card
+      style={{ minWidth: 'min-content' }}
       size="tiny"
       className={`event-panel event-${wcifEvent.id}`}
     >
--- a/WcaOnRails/app/webpacker/components/EditEvents/EventPanel/index.js
+++ b/WcaOnRails/app/webpacker/components/EditEvents/EventPanel/index.js
@@ -5,7 +5,6 @@ import {
   Card,
   Header,
   Icon,
-  Label,
   Segment,
 } from 'semantic-ui-react';
 import i18n from '../../../lib/i18n';
--- a/WcaOnRails/app/webpacker/components/EditEvents/EventPanel/index.js
+++ b/WcaOnRails/app/webpacker/components/EditEvents/EventPanel/index.js
@@ -115,11 +115,10 @@ export default function EventPanel({
   return (
     <Card
       size="tiny"
-      compact
       className={`event-panel event-${wcifEvent.id}`}
     >
       <Card.Content
-        // replicate the way SemUI Cards handle images (borderless) without an actual image
+        // replicate the way SemUI Cards handle images (borderless) without passing an actual image
         style={{ padding: 0 }}
       >
         <Segment basic tertiary>
--- a/WcaOnRails/app/webpacker/components/EditEvents/EventPanel/index.js
+++ b/WcaOnRails/app/webpacker/components/EditEvents/EventPanel/index.js
@@ -5,7 +5,8 @@ import {
   Card,
   Header,
   Icon,
-  Label, Segment,
+  Label,
+  Segment,
 } from 'semantic-ui-react';
 import i18n from '../../../lib/i18n';
 import { events } from '../../../lib/wca-data.js.erb';
@@ -86,7 +87,6 @@ export default function EventPanel({
             }
             onClick={handleRemoveEvent}
             negative
-            size="small"
           >
             Remove event
           </Button>
@@ -122,13 +122,13 @@ export default function EventPanel({
         style={{ padding: 0 }}
       >
         <Segment basic tertiary>
-          <Header as="span">
+          <Header as="span" size="tiny">
             <Icon className="cubing-icon" name={`event-${event.id}`} />
             <Header.Content>
               {event.name}
             </Header.Content>
           </Header>
-          <Button.Group floated="right">
+          <Button.Group floated="right" size="tiny">
             {renderRoundCountInputs()}
           </Button.Group>
         </Segment>
@@ -143,10 +143,9 @@ export default function EventPanel({
             />
           </Card.Content>
           <Card.Content>
-            <Label basic>
-              {i18n.t('competitions.events.qualification')}
-              :
-            </Label>
+            {i18n.t('competitions.events.qualification')}
+            :
+            {' '}
             {/* Qualifications cannot be edited after the competition has been announced. */}
             {/* Qualifications cannot be added if the box from the competition form is unchecked. */}
             <EditQualificationModal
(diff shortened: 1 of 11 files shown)


WcaOnRails/app/webpacker/components/EditEvents/EventPanel/index.js
```diff
@@ -104,86 +104,66 @@ export default function EventPanel({
...
-
-        </span>
-        <div>
-          {renderRoundCountInputs()}
-        </div>
-      </Header>
-
+        <Segment basic tertiary>
+          <Header as="span">
+            <Icon className="cubing-icon" name={`event-${event.id}`} />
```
kr-matthews, line 126: I'm seeing errors in the console that the `name` prop provided is invalid. I think it's meant to be one of the thousands of pre-defined `SemanticICONS` options.

kr-matthews, line 126: These icons appear 'clickable', but they shouldn't. They were already like this though.

WcaOnRails/app/webpacker/components/EditEvents/Modals/EditQualificationModal/index.js
```diff
@@ -126,7 +126,7 @@ export default function EditQualificationModal({
       hasUnsavedChanges={hasUnsavedChanges()}
       disabled={disabled}
       tooltip={disabledReason}
-      triggerButtonProps={{ name: 'qualification'}}
+      triggerButtonProps={{ name: 'qualification' }}
```
kr-matthews, line 129: This (the word on the left) looks like a clickable element now, which I don't like.
image

WcaOnRails/app/webpacker/components/EditEvents/EventPanel/index.js
```diff
@@ -104,86 +104,66 @@ export default function EventPanel({
...
+            <Header.Content>
+              {event.name}
+            </Header.Content>
+          </Header>
+          <Button.Group floated="right">
+            {renderRoundCountInputs()}
+          </Button.Group>
```
kr-matthews, line 133: The corners are inconsistently rounded:
image

kr-matthews:
My screen isn't particularly small but the "To Advance" column is already getting squished:

image

And we always have 3 columns no matter how wide the screen is, so on narrower screens things get bad/unusable:

image

The existing page responds better to different widths, and will use fewer columns on narrower screens.

I understand we probably don't expect people to use phones for this, and I don't think the table getting a bit squished is a big deal, but using a dynamic number of columns rather than always 3 would probably be worth doing.

WcaOnRails/app/webpacker/components/EditEvents/Modals/EditQualificationModal/index.js
gregorbg: Thanks, fixed!

WcaOnRails/app/webpacker/components/EditEvents/EventPanel/index.js
gregorbg: Indeed this was because cubing icons are not predefined in Semantic.
I'm a bit surprised that there's no better support for custom icon sets, but anyways I've circumvented the error by use of the `cn` helper function

gregorbg: This is a problem with our global stylesheets, which will get resolved once we get rid of our Sprockets assets one day

gregorbg:
I tried to make it as responsive as was possible within the constraints of the SemUI framework.
It's not 100% perfect but it's usable:
image

If anybody has any idea why the container grows without encompassing the table, feel free to shout.

kr-matthews:
I don't know why the container doesn't grow to include the table. As far as I can tell, cards are designed to always wrap their content? Unless you pass `fluid` which makes them the full width. Maybe using a different semantic ui component would be better?

If we're going to stick with cards, then I think setting the card width to be some fixed value (like 500px) would be preferable to the existing `itemsPerRow={3}` on the card group.

gregorbg:
> I don't know why the container doesn't grow to include the table. As far as I can tell, cards are designed to always wrap their content?

No, if they are in groups they are designed to take one "count per row"-th of space. In this case, we want three cards per row, so one third of space. I have reconfigured the card group to be stackable on mobile.

> Unless you pass `fluid` which makes them the full width. Maybe using a different semantic ui component would be better?

Passing `fluid` makes them take the full width available. Ironically, even on mobile screens, they do take the full width correctly -- the problem is that the table doesn't. It is simply too wide and there are myriads of articles online on why tables are a responsive design nightmare. It's not that the cards aren't wide enough (in the dev console CSS properties you can see they take 100% of the space) but rather that the tables take **too much** horizontal space.

So if anything, we should consider shrinking the table or replacing it with something more responsive like a grid...?

> If we're going to stick with cards, then I think setting the card width to be some fixed value (like 500px) would be preferable to the existing `itemsPerRow={3}` on the card group.

Fixed values that come out of thin air are almost never preferrable, except maybe in situations like setting the padding to be (fixed) zero.

kr-matthews:
> > I don't know why the container doesn't grow to include the table. As far as I can tell, cards are designed to always wrap their content?
> 
> No, if they are in groups they are designed to take one "count per row"-th of space. In this case, we want three cards per row, so one third of space. I have reconfigured the card group to be stackable on mobile.

Hm but if I remove the table (and the `itemsPerRow={3}`) and put a div with hard-coded width, the cards still ignore that width. So is the table really causing the issue?

image

> So if anything, we should consider shrinking the table or replacing it with something more responsive like a grid...?

I don't know if that would work but it seems reasonable to try.
 
> > If we're going to stick with cards, then I think setting the card width to be some fixed value (like 500px) would be preferable to the existing `itemsPerRow={3}` on the card group.
> 
> Fixed values that come out of thin air are almost never preferrable, except maybe in situations like setting the padding to be (fixed) zero.

I 100% agree - but isn't the 3 from `itemsPerRow={3}` coming out of equally-thin air? It's trading one fixed value for another, and makes it work well on any screen size. This is why I said preferable, rather than good :)

gregorbg:
With a directive like `width: max-content`, this is what happens on wider screens:
image

My suggested solution is to use `minWidth: min-content`, so that no matter the screen size the browser makes sure that the cards at the very least wrap around the content.

Widescreen computer:
image

Laptop: Note that this results in some unused space, because the min-width is now bigger than the SemUI-calculcated "one third of the screen", meaning it uses MORE than one third of the screen, effectively wasting some space but keeping the cards readable:
image

Mobile with stacked table:
image

Mobile without stacked table:
image

kr-matthews (approved):
I can get the Add Event button to move down at certain widths:
image

It's a bit annoying that editing an event can cause it to get wider and move to the next row.

But all the content is indeed contained inside each card, and they are always use-able at any screen width :+1:. So if we're ok with the above observations, then this looks good to me.

Issue #8634, opened 2023-12-20
Modal to select rounds for cumulative time limit gets cut-off


kr-matthews:
**Describe the bug**

If your competition has a lot of rounds and you try to create a shared cumulative time limit, the modal that pops up might not be big enough and there's no scrolling or anything.

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to a competition you can edit
2. Click on Edit then Manage Events in the left sidebar
3. Add a lot of events with a lot of rounds (not sure exactly how many, depends on screen size).
4. Edit the time limit for any particular event/round.
5. Select the cumulative option.
6. Click share with other rounds.
7. See the issue.

**Expected behavior**
Either all options are fit onto the screen, or a scroll bar appears, or something.

**Screenshots**

**Desktop (please complete the following information):**

PR #8755, opened 2024-01-23
Make Scroll Appear When Editing Cumulative Time Limit (#8634)

Carter-Thomas:
Haven't tested in my own environment yet, should work though. Makes a scrollbar appear if too many rounds are viewed at once.


gregorbg (changes requested):
Actually, our UI framework Semantic UI has a built-in solution for this: https://react.semantic-ui.com/modules/modal/#variations-scrolling-content

All you need to do is to enable it in `ButtonActivatedModal`.
Note: Due to some strict conventions, you also need to wrap the `Modal.Content` children inside a `Modal.Description` like so:

```jsx
<Modal.Content scrolling>
  <Modal.Description>
    {children}
  </Modal.Description>
</Modal.Content>
```

Carter-Thomas:
Should have fixed it, let me know if I messed something up when I updated my branch.

gregorbg:
It seems that you have copy/pasted the wrong file somewhere.
The diff is huge, and you're pasting seemingly unrelated changes...

PR #8766, opened 2024-01-24
Host Regulations and Documents Files on S3

FinnIckler:
This PR hosts all wca-document files and the egulations files on S3.


FinnIckler:
Changed the scope of this PR to just include the old regulations file, as the new ones are .erb files and can't be served from S3

FinnIckler:
I think this is done pending the tnoodle releases and deploy PRs in wca-regulations and wca-regulations-translations

FinnIckler:
I've now also done the tnoodle versions and made the other PRs. So this is ready for reviews

WcaOnRails/app/views/regulations/history.html.erb
gregorbg, line 1: This file has a lot of inconsistencies:
- There is `/history/files`
- there also is `/files/history`
- there are obvious typos like `/historyfiles/2013-05-10`

WcaOnRails/env_config.rb
```diff
@@ -30,7 +30,7 @@
     optional :SIDEKIQ_REDIS_URL, :string, ''
     optional :DISCOURSE_URL, :string, ''
     optional :STORAGE_AWS_BUCKET, :string, ''
-    optional :STORAGE_AWS_REGION, :string, ''
+    optional :STORAGE_AWS_REGION, :string, 'us-west-2'
```
gregorbg, line 33: Not a big fan of hardcoding this here.
Might cause a lot of scratching heads "Why am I automatically being thrown to `us-west-2` without even specifying anything?" down the line.
If you really need this default value, change it in the (checked-in!) `.env.development` file.

gregorbg (changes requested):

WcaOnRails/env_config.rb
FinnIckler: makes sense! Moved it

WcaOnRails/app/views/regulations/history.html.erb
FinnIckler: history/files is correct. fixed!

gregorbg (approved):

PR #8768, opened 2024-01-25
Removes downcasing of country_iso2 as it breaks posting dashboard

danieljames-dj:


gregorbg (changes requested):
Please be careful about changing the serialization format of commonly used data. There's no way of knowing whether other frontends break because they expect the data to be upper-cased.

Also, ISO 3166-1 explicitly specifies uppercase. Sending it as lowercase is wrong, to put it bluntly.

If there is one specific frontend which needs lowercase (for whatever reason, because that doesn't even adhere to the ISO standard) then please lowercase the value only in that frontend.

gregorbg (approved):
Nevermind I read the changelog in the wrong direction. Ignore my previous comment, I'm stupid :D

PR #8772, opened 2024-01-26
Giving Senior Delegates permission to subregions and overall permission cleanup

danieljames-dj:


WcaOnRails/app/models/user_group.rb
```diff
@@ -90,6 +91,10 @@ def lead_user
     end
   end
 
+  def child_groups
+    UserGroup.where(parent_group_id: self.id)
+  end
```
gregorbg, line 96: This looks like it can be implemented through a `has_many` relation.
See `app/models/stripe_transaction.rb` around line 30-ish for inspiration.

gregorbg: (This is important because it's not only shorter, but also much more efficient when Rails computes SQL queries)

WcaOnRails/app/models/user.rb
```diff
@@ -626,6 +626,20 @@ def banned_at_date?(date)
     end
   end
 
+  private def can_edit_any_groups?
+    admin? || board_member? || results_team?
+  end
+
+  private def groups_with_edit_access
```
gregorbg, line 633: The name implies that this is generic for any group. But in reality it only covers Delegate Regions.

In the `def permissions` method below, you also have cases for example for `can_edit_teams_committees`. In terms of code, these are groups as well, why not consider them?

gregorbg (changes requested):

WcaOnRails/app/models/user.rb
danieljames-dj: Currently, teams cannot be combined with this because teams doesn't have a groupId now. But once the migration of teams is completed, this function can be expanded to them as well. It can actually be expanded to other delegate group types that are already migrated like delegate_probations & translators, I'll do that later in another PR.

WcaOnRails/app/models/user_group.rb
danieljames-dj: Done, this is nice :-)

WcaOnRails/app/models/user_group.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class UserGroup < ApplicationRecord
+  has_many :child_groups, class_name: "UserGroup", foreign_key: "parent_group_id"
```
gregorbg, line 4: Nitpick 1: Move this down to the other associations like `metadata` and `parent_group`.
Nitpick 2: Use `inverse_of: :parent_group`, as per my advice to check out `stripe_transaction.rb`

gregorbg (changes requested):

WcaOnRails/app/models/user_group.rb
danieljames-dj: Done

gregorbg (approved):

PR #8779, opened 2024-01-28
GroupManager bug fix and test cleanup

danieljames-dj:


WcaOnRails/app/controllers/users_controller.rb
```diff
@@ -281,9 +281,6 @@ def acknowledge_cookies
 
   private def user_params
     params.require(:user).permit(current_user.editable_fields_of_user(user_to_edit).to_a).tap do |user_params|
-      if user_params.key?(:delegate_status) && !User.delegate_status_requires_senior_delegate(user_params[:delegate_status])
-        user_params["region_id"] = nil
-      end
```
danieljames-dj, line 286: Earlier, the senior_delegate_id was set to nil if a user didn't need senior delegate. But currently, this flow doesn't exist. We do change delegate roles through different API, so this code should be irrelevant now.

WcaOnRails/app/models/user.rb
```diff
@@ -416,17 +417,6 @@ def senior_delegate_must_be_senior_delegate
...
-  def self.delegate_status_requires_senior_delegate(delegate_status)
-    {
-      nil => false,
-      "" => false,
-      "trainee_delegate" => true,
-      "candidate_delegate" => true,
-      "delegate" => true,
-      "senior_delegate" => false,
-    }.fetch(delegate_status)
-  end
```
danieljames-dj, line 428: Not used anymore.

WcaOnRails/app/models/user.rb
```diff
@@ -532,12 +522,6 @@ def staff?
     staff_delegate? || member_of_any_official_team? || board_member? || officer?
   end
 
-  def staff_with_voting_rights?
-    # See "Member with Voting Rights" in:
-    #  https://www.worldcubeassociation.org/documents/motions/02.2019.1%20-%20Definitions.pdf
-    full_delegate? || senior_delegate? || senior_member_of_any_official_team? || leader_of_any_official_team? || board_member? || officer?
-  end
```
danieljames-dj, line 539: Not used anywhere, not sure if it was used earlier as well.

WcaOnRails/app/models/user.rb
```diff
@@ -579,7 +563,7 @@ def admin?
   end
 
   def any_kind_of_delegate?
-    delegate_status.present?
+    delegate_status.present? || active_roles.any? { |role| role.group_type == UserGroup.group_types[:delegate_regions] }
```
danieljames-dj, line 566: active_roles of user object returns only those roles which are already migrated to roles table. So, it will currently cover only regional_delegates. Once senior_delegates are migrated to roles, this function will work without any change. Hopefully next time function needs to be touched is in the PR where we remove delegate_status from users table.

WcaOnRails/app/webpacker/components/Panel/Leader/GroupsManager.jsx
```diff
@@ -54,7 +54,7 @@ export default function GroupsManager({ loggedInUserId }) {
   }, [roles]);
 
   if (!loading && roles.length === 0) return <p>You are not a leader of any groups.</p>;
-  if (loading || selectedGroupId === null) return <Loading />;
+  if (loading || selectedGroupId == null) return <Loading />;
```
danieljames-dj, line 57: Earlier, the condition was satisfied only when selectedGroupId was null, but undefined should also be handled, so changed `===` to `==`.

WcaOnRails/spec/factories/user_roles.rb
danieljames-dj, line 1: Cleaned up this file to avoid use of nested factory creation which I read as a bad practice somewhere.

WcaOnRails/spec/models/user_spec.rb
```diff
@@ -65,16 +65,7 @@
...
   end
 
-  it "allows demotion of a senior delegate with no subordinate delegates" do
-    senior_delegate = FactoryBot.create :senior_delegate
-
-    senior_delegate.delegate_status = ""
-    senior_delegate.region_id = nil
-    expect(senior_delegate.save).to eq true
-    expect(senior_delegate.reload.delegate_status).to eq nil
-  end
```
danieljames-dj, line 75: Now demotion of senior delegates doesn't depend on subordinate delegates because delegates are not connected to senior delegates, instead are connected to regions.

Couldn't think of any replacement test as well.

WcaOnRails/spec/models/user_spec.rb
```diff
@@ -65,16 +65,7 @@
...
-
-  it "requires senior delegate be a senior delegate" do
```
danieljames-dj, line 77: I think I forgot to rename this test earlier. This is actually replacement of the next test which is removed.

We don't need test for senior delegate to be a senior delegate because we anyway fetch senior delegates from roles with status senior_delegate. Also, currently it's in transition stage, so will have good tests for roles once the transition is completed.

WcaOnRails/spec/models/user_spec.rb
```diff
@@ -87,11 +78,6 @@
     expect(delegate).to be_valid
   end
 
-  it "requires senior delegate if delegate status allows it" do
-    delegate = FactoryBot.build :delegate, region_id: nil
-    expect(delegate).to be_invalid_with_errors(region_id: ["can't be blank"])
-  end
```
danieljames-dj, line 93: The test is not needed because senior_delegate is replaced with region and the test for that is the previous test.

WcaOnRails/spec/models/user_spec.rb
```diff
@@ -108,16 +94,6 @@
...
 
-  it "allows senior delegate if board member" do
-    board_member = FactoryBot.create :delegate, :board_member
-
-    senior_delegate = FactoryBot.create :senior_delegate
-
-    expect(board_member).to be_valid
-    board_member.region_id = senior_delegate.region_id
-    expect(board_member).to be_valid
-  end
```
danieljames-dj, line 119: The test is no longer needed because board & senior delegate are independent roles and don't have any connection once migrated to roles.

danieljames-dj (commented):
Since few testcases are removed and some changes looked confusing, I've left comments for major changes. These changes will also help in making the migration of Senior Delegates smoother.

WcaOnRails/app/webpacker/components/Panel/Leader/GroupsManager.jsx
gregorbg: Abusing the fact that "accidentally" null and undefined behave the same way when using `==` is not a good approach.
I would favor either
1. Using `!selectedGroupId`
2. Refactoring so that `selectedGroupId` only uses null OR undefined, but cannot become both.
(with a strong preference for option (2), but giving option (1) as a backup if the other is too hard)

danieljames-dj: Okay, will go with 2, I just checked and there is no case where `selectedGroupId` is `null`, so I changed it to `=== undefined`.
1st option is not possible because `selectedGroupId` can be 0 as well.

danieljames-dj:
> Mostly fine, just one nitpick which is added in my response to one of your explanation comments

Done

gregorbg:
Merge conflict :upside_down_face:

danieljames-dj:
> Merge conflict 🙃

Resolved, I didn't realize that I changed the same file in two different ways - one here and one in 'improvising mail' PR 😅

gregorbg (approved):

danieljames-dj:
> ...aaand there's a merge conflict again 😅

Resolved again :-)

PR #8781, opened 2024-01-28
Sync email for delegates fix

danieljames-dj:


gregorbg:
I am not a big fan of including child roles for every group in every case by default.

`def roles` should only be direct roles.
You can add a second `def child_roles` and then combine them at the place where the methods are being called, if needed.

danieljames-dj:
> I am not a big fan of including child roles for every group in every case by default.
> 
> `def roles` should only be direct roles. You can add a second `def child_roles` and then combine them at the place where the methods are being called, if needed.

Done

WcaOnRails/app/models/user_group.rb
```diff
@@ -60,10 +60,18 @@ def roles
     role_list
   end
 
+  def child_roles
+    child_groups.map(&:roles).flatten
```
gregorbg, line 64: `child_groups.flat_map(&:roles)` 😉

gregorbg (changes requested):

WcaOnRails/app/models/user_group.rb
danieljames-dj: Done

gregorbg (approved):

PR #8786, opened 2024-01-29
React Checkbox Component to Show Guidelines

Nanush7:
Needs to merge thewca/wca-regulations-compiler#28 and deploy to PyPI first.

The style of the component may be changed at the discretion of the WST (please feel free to do whatever you want with this file).

### DIFF
--- a/WcaOnRails/app/webpacker/components/Regulations/index.jsx
+++ b/WcaOnRails/app/webpacker/components/Regulations/index.jsx
@@ -1,0 +1,73 @@
+import React from 'react';
+
+class showGuidelines extends React.Component {
+    constructor(props) {
+        super(props);
+        if (window.location.hash.includes("+")) {
+            this.state = {
+                checkboxChecked: true
+            };
+            this.showGuidelinesSelected();
+        } else {
+            this.state = {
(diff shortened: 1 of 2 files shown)


FinnIckler:
I'm currently in the process of migrating the regulations off the machine, will this mean that guidelines won't be linked to anymore?

gregorbg (changes requested):
I am sorry but I cannot approve this at all.

The React component is written the way you would have done 5 years ago, and the approach is too jQuery-esque.

Also, the fact that you are manipulating the wrc-compiler tool to emit a `react_component` statement is horrendous because it relies on the fact that (a) The library that contains the `react_component` feature will be integrated into our codebase forever, which is definitely NOT true, and (b) that this component never gets renamed/moved. I definitely have to veto the prospect of changing the wrc-compiler (and releasing it to PyPI, which is a pain in the butt!) every time that our monolith code changes

gregorbg:
I truly appreciate your enthusiasm about contributing this yourself, and I think it's great that you figured out the intricacies of how React components are rendered. But the end result is not maintainable or scalable at all, so please consult WST next time you want to implement a feature like this.

PR #8789, opened 2024-01-30
Remove Building Regs and Docs on Machine

FinnIckler:
Now that these are hosted on S3, these are not needed (and actually cause some damage on the new files)


gregorbg:
This means you will need to touch https://github.com/thewca/deploy/blob/main/.github/workflows/deploy-production.yml and https://github.com/thewca/deploy/blob/main/.github/workflows/deploy-staging.yml in theory.

We're expected to change the whole deployment pipeline anyways, but in the meantime these scripts might not work for a short while because they use `rebuild_regs`

FinnIckler:
https://github.com/thewca/deploy/pull/5
removed them here

PR #8795, opened 2024-01-31
Host Exports on s3

FinnIckler:
I have tested the hosting and downloading part of this, but not the generation, which I will do after dinner


WcaOnRails/app/controllers/database_controller.rb
```diff
@@ -4,38 +4,41 @@ class DatabaseController < ApplicationController
...
+    redirect_to url, status: 301, allow_other_host: true
   end
 
-  def self.rel_download_path(base_folder, file_name)
-    file_path = base_folder.join file_name
-    relative_path = file_path.relative_path_from DbDumpHelper::EXPORT_PUBLIC_FOLDER.parent
+  def get_current_export(type)
+    export_timestamp = Time.new(DbDumpHelper::export_metadata["export_date"])
+
+    Rails.cache.fetch("database-export-#{export_timestamp}-#{type}", expires_in: 2.days) do
```
gregorbg, line 29: We generate the results export daily, so expiry of 2 days is definitely not enough.
You can use the timestamps from the cronjobs in the cache key, similar to how we do it for CAD tasks

WcaOnRails/app/controllers/database_controller.rb
```diff
@@ -4,38 +4,41 @@ class DatabaseController < ApplicationController
...
-    @rel_download_path = DatabaseController.rel_download_path DbDumpHelper::DEVELOPER_EXPORT_FOLDER, DbDumpHelper::DEVELOPER_EXPORT_SQL_PERMALINK
+  def tsv_permalink
+    url, _ = get_current_export("sql")
+    redirect_to url, status: 301, allow_other_host: true
   end
 
-  def self.rel_download_path(base_folder, file_name)
-    file_path = base_folder.join file_name
-    relative_path = file_path.relative_path_from DbDumpHelper::EXPORT_PUBLIC_FOLDER.parent
+  def get_current_export(type)
```
gregorbg, line 26: It's not clear which export this is referring to -- above it seems like `type` is getting values "sql" and "tsv", whereas I would expect "type of export" to be `developer` or `public_results`.

WcaOnRails/app/controllers/database_controller.rb
```diff
@@ -4,38 +4,41 @@ class DatabaseController < ApplicationController
...
+    export_timestamp = Time.new(DbDumpHelper::export_metadata["export_date"])
+
+    Rails.cache.fetch("database-export-#{export_timestamp}-#{type}", expires_in: 2.days) do
+      file_name = "#{DbDumpHelper::RESULTS_EXPORT_FOLDER}/WCA_export#{export_timestamp.strftime('%j')}_#{export_timestamp.strftime('%Y%m%dT%H%M%SZ')}.#{type}.zip"
+      bucket = Aws::S3::Resource.new(
+        region: EnvConfig.STORAGE_AWS_REGION,
+        credentials: Aws::InstanceProfileCredentials.new,
+        ).bucket(DbDumpHelper::BUCKET_NAME)
+      filesize_bytes = bucket.object(file_name).content_length
+      ["https://s3.#{EnvConfig.AWS_STORAGE_REGION}.amazonaws.com/#{DbDumpHelper::BUCKET_NAME}/#{file_name}", filesize_bytes]
```
gregorbg, line 36: Out of curiosity: Am I reading it correctly that amazon provides human-guessable permalinks to every public S3 bucket in the whole world? And does this imply that our `DbDumpHelper::BUCKET_NAME` has to be unique among the whole us-west-2 region?!

gregorbg (changes requested):

WcaOnRails/app/controllers/database_controller.rb
FinnIckler: yes, it even has to be globally unique (crazy right!)

WcaOnRails/app/controllers/database_controller.rb
FinnIckler: I'll change it to file_type

WcaOnRails/app/controllers/database_controller.rb
FinnIckler: that's good, that means I don't have to upload the metadata as well

WcaOnRails/app/controllers/database_controller.rb
gregorbg: In that case: What is `current` referring to?
As far as I can tell, this controller is for both developer and public_results. Which one is the "current" one w.r.t. to this method?

FinnIckler: this is just for the developer_export. There is no need to get the 'current export' for the developer export, because we always overwrite that one

gregorbg: In that case, please make it clear by giving the method a name like `def current_results_export(file_type)`

WcaOnRails/app/controllers/database_controller.rb
```diff
@@ -4,38 +4,41 @@ class DatabaseController < ApplicationController
...
-    @rel_download_path = DatabaseController.rel_download_path DbDumpHelper::DEVELOPER_EXPORT_FOLDER, DbDumpHelper::DEVELOPER_EXPORT_SQL_PERMALINK
+  def tsv_permalink
+    url, _ = get_current_results_export("sql")
+    redirect_to url, status: 301, allow_other_host: true
   end
 
-  def self.rel_download_path(base_folder, file_name)
-    file_path = base_folder.join file_name
-    relative_path = file_path.relative_path_from DbDumpHelper::EXPORT_PUBLIC_FOLDER.parent
+  def get_current_results_export(file_type)
```
gregorbg, line 26: super-nitpick: Rails methods usually don't include get or set. That's not a technical requirement though, only a convention.

WcaOnRails/app/controllers/database_controller.rb
FinnIckler: Sadly the cronjob timestamp is not the same as the metadata timestamp.
2024-01-31 00:00:06 UTC vs. 2024-01-31 00:17:37 UTC for the current export. I guess this is because the cronjob timestamp is the enqueue time and the other is the actual export time. Should I make the dump_results_db use the cronjob time stamp instead?

WcaOnRails/app/controllers/database_controller.rb
FinnIckler: removed the get

WcaOnRails/app/controllers/database_controller.rb
gregorbg: Hm, which cronjob timestamp are you using? They have a run_start and run_end timestamp, and one of the two should match the metadata very closely

FinnIckler:
Successfully tested uploading and serving the results dump, it does not allow me to make a developer dump though because I am using the developer dump.

gregorbg (approved):

PR #8796, opened 2024-01-31
Sanitize input in OmniSearch bar

gregorbg:
Fixes a mild vulnerability where you can inject HTML href-tags into the search results page


FinnIckler (approved):
LGTM

PR #8855, opened 2024-02-03
Remove container css from delegates page and panel

danieljames-dj:
I don't know the use of `container` CSS class, but removing it is making the UI slightly better. Maybe I was wrongly using the `container` class? No idea, looking forward for suggestions from CSS experts out there.

Posting the before and after screenshots of Delegates page and Panel page in Mac 14 inch Chrome browser:

Before
Screenshot 2024-02-03 at 21 01 13
Screenshot 2024-02-03 at 21 01 57

After
Screenshot 2024-02-03 at 21 02 18
Screenshot 2024-02-03 at 21 02 34

### DIFF
--- a/WcaOnRails/app/webpacker/components/Delegates/index.jsx
+++ b/WcaOnRails/app/webpacker/components/Delegates/index.jsx
@@ -82,7 +82,7 @@ export default function Delegates() {
   }
 
   return (
-    <div className="container">
+    <div style={{ margin: '16px' }}>
       <Header as="h1">{I18n.t('delegates_page.title')}</Header>
       <p>
         <I18nHTMLTranslate
@@ -101,10 +101,10 @@ export default function Delegates() {
           onChange={(__, { checked }) => setToggleAdmin(checked)}
         />
       )}
-      <Grid container>
+      <Grid>
         <Grid.Column only="computer" computer={4}>
           <Header>{I18n.t('delegates_page.regions')}</Header>
-          <Menu vertical>
+          <Menu vertical fluid>
             {delegateRegions.map((region) => (
               <Menu.Item
                 key={region.id}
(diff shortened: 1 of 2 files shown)


coder13 (approved):

WcaOnRails/app/webpacker/components/Delegates/index.jsx
```diff
@@ -101,10 +101,10 @@ export default function Delegates() {
           onChange={(__, { checked }) => setToggleAdmin(checked)}
         />
       )}
-      <Grid container>
+      <Grid>
         <Grid.Column only="computer" computer={4}>
           <Header>{I18n.t('delegates_page.regions')}</Header>
-          <Menu vertical>
+          <Menu vertical fluid>
```
coder13, line 107: What does this line do?

danieljames-dj: Without that, the UI will be as follows. `fluid` makes it to use the available width completely. According to the documentation in semantic UI site, the definition of fluid is "A vertical menu may take the size of its container".

Screenshot 2024-02-03 at 21 12 30

WcaOnRails/app/webpacker/components/Delegates/index.jsx
```diff
@@ -82,7 +82,7 @@ export default function Delegates() {
   }
 
   return (
-    <div className="container">
+    <div style={{ margin: '16px' }}>
```
gregorbg, line 85: Veto about including magic numbers in our CSS. We simply don't do that.
You can change the (old, pre-change) `<div className="container">` into an actual SemUI React `<Container>` and then play around with `<Container fluid>` or whatever else SemUI has to offer. But no hard-coded margins at any cost.

gregorbg (changes requested):

WcaOnRails/app/webpacker/components/Delegates/index.jsx
danieljames-dj: Thanks a lot @gregorbg, <Container fluid> did the magic. :-)

gregorbg (approved):

PR #8856, opened 2024-02-03
Translators metadata fix

danieljames-dj:
Initially, we planned to keep all translators under single group, but that had a problem like we won't get the list of locales available. So, we are now having one group for one locale.

Also, in the initial implementation, I accidentally named RolesMetadataTranslators as GroupsMetadataTranslators. Now we anyway need GroupsMetadataTranslators, so using that.


gregorbg:
Can you please add a description about what this PR tries to achieve and why it's necessary?

danieljames-dj:
> Can you please add a description about what this PR tries to achieve and why it's necessary?

Done

gregorbg (approved):

PR #8861, opened 2024-02-04
Add tentative competitor counts to report

kr-matthews:
Results must be _submitted_ before the report can be posted, but WRT may not have _posted_ the results yet, resulting in "Unknown, results are not posted yet." That text will still show up before the results/report have been submitted.

I did the first commit back in May 2022, but didn't know how to test it back then.

image

### DIFF
--- a/WcaOnRails/app/views/delegate_reports/_delegate_report.html.erb
+++ b/WcaOnRails/app/views/delegate_reports/_delegate_report.html.erb
@@ -33,6 +33,8 @@
 
     <% if competition.results_posted? %>
       <%= competition.competitors.count %>
+    <% elseif competition.results_submitted? %>
+      <%= competition.competitors.count %> (tentative)
     <% else %>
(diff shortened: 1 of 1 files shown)


gregorbg (approved):

EdHollingdale:
I'm not sure this is working @kr-matthews. See Cainta Cubing Mania 2024 report that lists the tentative competitors as 0.

kr-matthews:
> I'm not sure this is working @kr-matthews. See Cainta Cubing Mania 2024 report that lists the tentative competitors as 0.

It worked for the Stockholmsligan 358: Snabba Pengar - 2024 report.

The issue is likely that WRT were in the process of posting results when the Cainta Cubing Mania 2024 report was emailed out (had cleared the inbox but not posted the results yet). We have a potential fix for that edge case, and regardless we''ll make sure that if it somehow wants to print "0 (tentative)" then that will be replaced with the original "Unknown" message.

PR #8863, opened 2024-02-05
Add loading indicator for delegate selection 

notrichardpeng:
Added a loading spinner to the delegate selection box and modified the no results message for clarity

image

image

image

image

### DIFF
--- a/WcaOnRails/config/locales/en.yml
+++ b/WcaOnRails/config/locales/en.yml
@@ -1537,8 +1537,8 @@ en:
       past_from: "Past from %{year}"
       past_all: "Past from all years"
       no_delegates: "None"
-      delegates_loading: "Waiting to finish loading all delegates..."
-      no_delegates_found: "No delegate found."
+      delegates_loading: "Waiting to finish loading all Delegates..."
+      no_delegates_found: "No Delegates found."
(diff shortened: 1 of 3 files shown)


WcaOnRails/config/locales/en.yml
```diff
@@ -1537,6 +1537,8 @@ en:
       past_from: "Past from %{year}"
       past_all: "Past from all years"
       no_delegates: "None"
+      delegates_loading: "Waiting to finish loading all delegates..."
```
gregorbg, line 1540: Consistency: We upper-case terms like "Delegate"

WcaOnRails/config/locales/en.yml
```diff
@@ -1537,6 +1537,8 @@ en:
...
+      no_delegates_found: "No delegate found."
```
gregorbg, line 1541: Maybe add plural like so: "No Delegate(s) found."?

gregorbg (changes requested):

WcaOnRails/config/locales/en.yml
notrichardpeng: The default just says "No results found." so maybe "No Delegates found." is good?

gregorbg (approved):

PR #8865, opened 2024-02-05
Feature/stripe refactor

dunkOnIT:
PR 2/3, where the previous one introduced a PayPal proof-of-concept, and the next one will integrate both payment methods into an abstract payments handler (or something that generalizes away from a Stripe-specific integration)

- [x] Set up StripeAccount model
  - [x] add & link model
  - [x] change payments setup page
  - [x] change all connected_stripe_account_id references
- [x] gregor refactors
- [x] make sure I haven't broken anything
  - [ ] review existing tests for stripe
  - [ ] add any "missing" tests
  - [x] deploy to staging and test


WcaOnRails/config/locales/en.yml
```diff
@@ -1886,6 +1880,32 @@ en:
...
+      stripe: 'Stripe'
+    #context: Setting up and making payments
+    payment_setup:
+      account_connected: "You have successfully connected your %{provider} account and can now receive registration payments for this competition."
+      account_disconnected_success: "Sucessfully disconnected your %{provider} account."
+      account_disconnected_failure: "There was an error with disconnecting the %{provider} account."
+      account_not_connected: "There was a problem and we could not connect your %{provider} account."
+      disconnect_account: "Disconnect %{provider} Account"
+      confirm_account_disconnect: "Are you sure you want to disconnect the %{provider} account? This action cannot be undone."
+      supported_currency_warning: "Please make sure the currency you have selected is supported by your chosen payment integration! The supported currencies will depend on the country where the account account is based, you can check them below:"
```
hanwu85, line 1896: Found 2 accounts in the sentence.
```suggestion
      supported_currency_warning: "Please make sure the currency you have selected is supported by your chosen payment integration! The supported currencies will depend on the country where the account is based, you can check them below:"
```

WcaOnRails/app/controllers/competitions_controller.rb
```diff
@@ -345,19 +345,20 @@ def paypal_return
 
   def stripe_connect
     code = params[:code]
-    competition = Competition.find(params[:state])
-    unless current_user&.can_manage_competition?(competition)
+    @competition = Competition.find(params[:state])
```
gregorbg, line 348: The `@` is only necessary if there is an accompanying Rails view that needs to access this variable.

In this case, the method always ends with a redirect, which leads to another controller method. A view is never shown directly from this controller method, so the `@` is unnecessary.

WcaOnRails/app/jobs/clear_connected_payment_integrations.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class ClearConnectedPaymentIntegrations < WcaCronjob
+  DELAY_IN_DAYS = 21
+
+  def perform
```
gregorbg, line 6: I feel like there are more efficient solutions.
1. It is unnecessary to iterate over ALL competitions that ended more than 21 days ago. You will simply blow up the list of competitions more and more over time. Instead, you could just iterate over competitions that are more than X days old _and_ have connected accounts in the first place.
2. Can you have a disconnect that "fails silently" i.e. does nothing if the to-be-disconnected account type was never connected in the first place? Then you can simply iterate over all supported connections and "try to disconnect" them without having to run queries checking whether they're connected.
3. If you really must iterate over lots of competitions (see 1), then use `find_each` instead of `each`. This only works on Rails models and not in general Ruby, but it batches the underlying SQL query _much_ more efficiently.

Bonus idea if you want to feel super pro: What does the `disconnect` method actually do? If it just sets a flag on one table row or just issues a delete, you can re-implement all of this with one simple statement:

```ruby
CompetitionPaymentIntegration.includes(:competition).where(competition: { end_date: ...(DELAY_IN_DAYS.days.ago) }).update_all(is_active: false)
```

WcaOnRails/app/models/competition.rb
```diff
@@ -507,7 +507,7 @@ def warnings_for(user)
         warnings = championship_warnings.merge(warnings)
       end
 
-      if has_fees? && !connected_stripe_account_id
+      if has_fees? && !CompetitionPaymentIntegration.stripe_connected?(self) && !CompetitionPaymentIntegration.paypal_connected?(self)
```
gregorbg, line 510: Why not `!competiiton_payment_integrations.exists?`

WcaOnRails/config/routes.rb
```diff
@@ -138,6 +140,7 @@
 
   get 'competitions/:id/payment_setup' => 'competitions#payment_setup', as: :competitions_payment_setup
   get 'stripe-connect' => 'competitions#stripe_connect', as: :competitions_stripe_connect
+  get 'competitions/:id/paypal-return' => 'competitions#paypal_return', as: :competitions_paypal_return
```
gregorbg, line 143: Why was this return URL route not included in the previous PR?
Is this a merge whoopsie or a general whoopsie?

gregorbg:
General remark regarding code design: You're using a lot of `CompetitionPaymentIntegration.account_for(my_cool_competition, 'stripe')`. Why not `my_cool_competition.payment_integration_for('stripe')` or similar? Feels less verbose...

WcaOnRails/config/routes.rb
dunkOnIT: It was included in the previous PR under the `unless Rails.env.production?` clause - the actual whoopsie was that all the routes were duplicated

WcaOnRails/app/controllers/competitions_controller.rb
dunkOnIT: My understanding is that `@` defines an instance variable - so I like using @ by default so that if we call another method inside the controller we can access the "sane" instance variables (stuff like @competition, @registration, etc). 

The thinking is that even if we're not _currently_ calling a method that needs access to an instance variable, I'd rather it be easy to implement in future.

But if you'd rather we explicitly pass stuff as parameters only as the need arises, I'm happy to do that!

WcaOnRails/config/locales/en.yml
dunkOnIT: Good catch, thanks!

WcaOnRails/app/controllers/competitions_controller.rb
gregorbg: Your technical explanations are all correct. The Rails templates/views can see it _because_ it is an instance variable of the controller.

> But if you'd rather we explicitly pass stuff as parameters only as the need arises, I'm happy to do that!

Yes please! The argument that "somebody might need this at some point in the future" leads to pollution and cluttering of source code. If you follow this paradigm all the time, you will end up with a scenario where suddenly there's a superfuous instance variable and you don't know where it was defined.

dunkOnIT: Makes sense, will do!

WcaOnRails/app/jobs/clear_connected_payment_integrations.rb
dunkOnIT: Great input, thanks! I've updated the code to (I think) only look at comps older than 21 days with a competition payment integration - I think it's happening in 2 calls though, there might be a way to do it in 1. 

Then I'm still iterating through all results and applying `disconnect_all` - I think your super pro solution is more efficient, but hopefully we're dealing with few enough records by that point that efficiency isn't _that_ bad? I'm opting to stick with `disconnect_all` so that we only have to change code in one place if we alter how comps are disconnected in future

Shout if that all sounds ok!

WcaOnRails/app/models/competition.rb
dunkOnIT: Much better, thanks!

WcaOnRails/app/jobs/clear_connected_payment_integrations.rb
gregorbg: Yup, I believe your approach works. Now just replace `comps_to_disconnect.each` by `comps_to_disconnect.find_each` (see above for explanation) and I shall approve :D

gregorbg: Oh and also `.where(end_date: ...DELAY_IN_DAYS.days.ago)` :wink:

dunkOnIT:
> General remark regarding code design: You're using a lot of `CompetitionPaymentIntegration.account_for(my_cool_competition, 'stripe')`. Why not `my_cool_competition.payment_integration_for('stripe')` or similar? Feels less verbose...

Great point - I'm busy refactoring the CompetitionPaymentIntegration class methods to be Competition instance methods. Silly I didn't do it this way around first!

WcaOnRails/app/models/competition.rb
```diff
@@ -2405,6 +2405,28 @@ def set_form_data_series(form_data_series, current_user)
...
+    validate_integration_name!(integration_name)
+
+    payment_integrations = competition_payment_integrations.where(
+      connected_account_type: CompetitionPaymentIntegration::AVAILABLE_INTEGRATIONS[integration_name],
+    )
+
+    return nil unless payment_integrations.count > 0
+
+    # Take the first result as `where` always returns an array
+    payment_integrations.first.connected_account
```
gregorbg, line 2422:
```suggestion
    return competition_payment_integrations.first(
      connected_account_type: CompetitionPaymentIntegration::AVAILABLE_INTEGRATIONS[integration_name],
    )&.connected_account
```

gregorbg: Context: In Rails ActiveRecord, `first` can act the same way `where` does, taking certain search criteria using keywords. It's basically a drop-in "return only the first one found instead of a whole array" replacement for `where`.

Also, big fat no on the extra `unless count > 0` because that fires an extra SQL query.
Nice to know for later, but definitely not necessary here: `Competition.where(foo: 'bar').exists?` uses an optimised query to check for existence on raw SQL level, without evaluating the full count.

WcaOnRails/app/models/competition.rb
```diff
@@ -2405,6 +2405,28 @@ def set_form_data_series(form_data_series, current_user)
...
+  end
+
+  private def validate_integration_name!(integration_name)
```
gregorbg, line 2425: I would leave this specific method in the CPI model class and call it / reference it here.
```ruby
def payment_account_for(type)
  CompetitionPaymentIntegration.validate_integration_name!(type)
  ...do stuff...
end
```

WcaOnRails/app/jobs/clear_connected_payment_integrations.rb
dunkOnIT: - `find_each` - made the change, plus a query on Slack
- What change are you wanting in the `where` clause? the `end_date:` looks like you want me to specify a parameter hash, but that doesn't seem possible with a `<`?

WcaOnRails/app/models/competition.rb
dunkOnIT: Flagging this as another instance where I would expect no SQL query given that we've already assigned the result of a `where` clause to `payment_integrations`

dunkOnIT: How sure are you on that behaviour of `.first`? In the code, your suggestion is generating `TypeError: no implicit conversion of Integer into Hash`, and the [Rails docs](https://apidock.com/rails/v5.2.3/ActiveRecord/FinderMethods/first) don't list an example with that usage - they all seem to use .where.first

WcaOnRails/app/jobs/clear_connected_payment_integrations.rb
gregorbg: My point is precisely that it _is_ possible!

- `Competition.where(end_date: 21.days.ago)` loads all competitions where end_date is equal to the specified value
- `Competition.where(end_date: ...21.days.ago)` basically does the same thing (load comps where the `end_date` column is equal to a certain value) but actually in this case the value you're passing is a Range object `a...b`. We don't know the start of the range, so we specify an open-ended exclusive range `...b` which means "anything up to b"

dunkOnIT: Ooooh I thought that ellipsis was a rhetorical flourish. Makes a lot more sense! :D

dunkOnIT: Yep works nicely! Latest commit includes the change.

WcaOnRails/app/models/competition.rb
gregorbg: My bad, it's actually `CoolModel.find_by(column: 'amazing_value')`

gregorbg: Re query count: The CPI.where chaining does not fire any queries.

In your original version, the .count fires the first query, and the .first fires the second query. But the point is, you're not really using that count in a reasonable way: If there is nothing present, the `first` will yield nil and you can use that to react to the "there's nothing there" case

dunkOnIT: Cool, thanks! Updated it to find_by :+1:

WcaOnRails/app/controllers/api/internal/v1/payment_controller.rb
```diff
@@ -3,7 +3,7 @@
 class Api::Internal::V1::PaymentController < Api::Internal::V1::ApiController
   # We are using our own authentication method with vault
   protect_from_forgery except: [:init]
```
gregorbg, line 5: If you rename `init` to `init_stripe`, then you have to rename it here as well.

WcaOnRails/app/controllers/competitions_controller.rb
```diff
@@ -389,14 +390,15 @@ def disconnect_paypal
...
   def disconnect_stripe
-    comp = competition_from_params
-    if comp.connected_stripe_account_id
-      comp.update!(connected_stripe_account_id: nil)
-      flash[:success] = t('payments.payment_setup.account_disconnected_success', provider: t('payments.payment_providers.stripe'))
-    else
+    competition = competition_from_params
+    competition.disconnect_payment_integration(:stripe)
+
+    competition.stripe_connected?
```
gregorbg, line 396: I think this is missing an `if`

gregorbg (approved):

dunkOnIT:
@thewca-bot deploy staging

gregorbg (changes requested):
Important change (sorry I didn't catch this during my earlier review...):
If you rename a job, you have to check the cronjob configuration at `app/config/schedule.yml` and rename all job occurences there accordingly.

WcaOnRails/config/schedule.yml
```diff
@@ -1,5 +1,5 @@
 clear_stripe:
-  class: "ClearConnectedStripeAccount"
+  class: "ClearConnectedPaymentIntegrations"
```
dunkOnIT, line 2: @gregorbg I believe the change requested for `schedule.yml` is already in place - or am I missing something somewhere else in the code?

gregorbg: Oh, nevermind :see_no_evil:

gregorbg (approved):

dunkOnIT:
@thewca-bot deploy staging

gregorbg (changes requested):
In your most recent commit to this PR, you changed the `down` method and then probably tested it on your local DB. This test made Ruby change the schema version in `schema.rb` back to the pre-migration timestamp. During deployment, we need it to point to the latest post-migration timestamp.

Also, inconsistency about `save!` in `up` VS `save` (no exclamation) in `down`.

dunkOnIT:
Thanks! `save` inconsistency resolved, and migration run again to update the schema timestamp. 

From my side, I'm confident in the migration - so happy to merge this PR when you are (and when the tests pass, of course!)

PR #8867, opened 2024-02-05
Run Yarn dedupe to clean up lockfile

gregorbg:
This does not change any of our dependencies, it only makes the lockfile more compact.
See https://yarnpkg.com/cli/dedupe for details on what this is all about.


gregorbg:

PR #8868, opened 2024-02-05
Replace senior_delegate with senior_delegates and cleaned up senior_delegate usages

danieljames-dj:
An inch more closer to senior delegates migration. Next PR will most probably be to migrate all the senior delegates to roles table

Also need a thorough review because this change is to reduce number of bugs during senior delegate migration.

### DIFF
--- a/WcaOnRails/spec/factories/competitions.rb
+++ b/WcaOnRails/spec/factories/competitions.rb
@@ -319,7 +319,9 @@ FactoryBot.define do
 
     after(:create) do |competition|
       competition.delegates.each do |delegate|
-        FactoryBot.create(:senior_delegate, region_id: delegate.region_id)
+        unless delegate.region_id.nil? # There can be cases where the competition delegate is actually not a delegate (temporary delegate)
+          FactoryBot.create(:senior_delegate, region_id: delegate.region_id)
+        end
       end
     end
   end


WcaOnRails/spec/mailers/competitions_mailer_spec.rb
gregorbg, line 1: Why are the reloads in this file becoming necessary?

WcaOnRails/spec/mailers/role_change_mailer_spec.rb
gregorbg, line 1: In controlled test cases like this, you usually know who will be the senior delegate without calling `senior_delegates`, because you create the user with a certain Senior already in mind (for example, you're passing a Senior when using FactoryBot.create).

As such, you should manually insert all the Seniors' emails you're expecting, instead of letting the code compute them in a way that you know can never go wrong (because it imitates the actual mailer and thus the result can never be different)

gregorbg: If the Senior is really created randomly, then consider
(a) changing the test code to create the Senior in a controlled fashion or
(b) discarding my comment completely if option (a) is too difficult.

danieljames-dj: Okay, I did it with (a). Hope it's good now.

WcaOnRails/app/mailers/competitions_mailer.rb
```diff
@@ -13,7 +13,7 @@ def notify_wcat_of_confirmed_competition(confirmer, competition)
       mail(
         from: Team.wcat.email,
         to: Team.wcat.email,
-        cc: competition.delegates.flat_map { |d| [d.email, d.senior_delegate&.email] }.compact.uniq,
+        cc: competition.delegates.flat_map { |d| [d.email, d.senior_delegates.flat_map(&:email)] }.compact.uniq,
```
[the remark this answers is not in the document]
danieljames-dj: Done, this looks great.

WcaOnRails/spec/mailers/competitions_mailer_spec.rb
danieljames-dj: I've removed reloads from delegate and trainee. It's needed in senior because after creating senior, it's not used anywhere directly so I think it's doing some kind of 'lazy loading' stuff, which needs it to be reloaded to activate.

WcaOnRails/app/mailers/role_change_mailer.rb
gregorbg, line 1: For both occurences in this file: There is an edge case where the Board may be a Senior, which means the `to` recipients will contain duplicates.
Not sure whether this is a problem (probably isn't) but just to be safe you should consider a `.compact.uniq` call

WcaOnRails/app/models/user.rb
```diff
@@ -56,7 +56,7 @@ def self.eligible_voters
 
   def self.leader_senior_voters
     team_leaders = TeamMember.current.in_official_team.leader.map(&:user)
-    senior_delegates = User.where(delegate_status: "senior_delegate")
+    senior_delegates = UserGroup.delegate_region_groups.where(parent_group_id: nil).map(&:lead_user)
```
gregorbg, line 59: Might be worth moving `UserGroup.delegate_region_groups.where(parent_group_id: nil)` into its own `UserGroup.senior_delegate_region_groups` method.
If we _ever_ change the way senior regions are structured (for example, if in one year we decide to introduce a "World" Delegate level), you only have to touch that one method.

WcaOnRails/app/models/user.rb
```diff
@@ -612,14 +612,26 @@ def banned_at_date?(date)
...
-    if senior_delegate?
-      region = UserGroup.find(self.region_id)
-      [region.id] + region.all_child_groups.pluck(:id)
-    else
-      [] # FIXME: Consider groups of other groupTypes as well.
+    groups = []
+
+    senior_delegate_roles.map do |role|
+      region = UserRole.group(role)
+      groups += [region.id, region.all_child_groups.map(&:id)].flatten
```
gregorbg, line 629: `.uniq` perhaps?

WcaOnRails/app/models/user.rb
```diff
@@ -1284,6 +1298,7 @@ def delegate_role
       is_active: true,
       group: self.region,
       user: self,
+      is_lead: delegate_status == "senior_delegate",
```
gregorbg, line 1301: Do we have enumerated constants for Delegate status available somewhere?

WcaOnRails/spec/mailers/role_change_mailer_spec.rb
```diff
@@ -5,33 +5,43 @@
...
-    let(:role) { FactoryBot.create(:probation_role) }
+    let(:delegate_region_americas) { FactoryBot.create(:delegate_region_americas) }
+    let(:senior_delegate) { FactoryBot.create(:senior_delegate, region_id: delegate_region_americas.id) }
+    let(:delegate) { FactoryBot.create(:delegate, region_id: delegate_region_americas.id) }
+    let(:role) { FactoryBot.create(:probation_role, user: delegate) }
     let(:mail) { described_class.notify_role_start(role, user_who_made_the_change) }
 
     it 'renders the headers' do
-      expect(mail.to).to eq [user_who_made_the_change.email, Team.board.email, role.user.senior_delegate.email]
+      senior_delegate.reload
```
gregorbg, line 15: It is absolutely redundant to reload a model in the test when nothing happened yet.

When this test runs, it creates `senior_delegate` freshly through `let(:senior_delegate)` above. Reloading it immediately after it is created changes nothing. If it _does_ change something, your code design is off and we should discuss about how to improve it :)

WcaOnRails/spec/factories/competitions.rb
gregorbg, line 1: Do you have to copy/paste this `after(:create)` hook three times?
Can't you just paste it in the "overall" competition factory and trust that nothing happens if the list of Delegates is empty?

gregorbg (changes requested):

WcaOnRails/app/mailers/role_change_mailer.rb
danieljames-dj: Okay done. Added a testcase as well for this case. (Added `.compact.uniq` while calling `mail` method.

WcaOnRails/app/models/user.rb
danieljames-dj: Done

WcaOnRails/app/models/user.rb
danieljames-dj: Done

WcaOnRails/app/models/user.rb
danieljames-dj: Yes we do, added now. Sorry for missing it. :-)

WcaOnRails/spec/factories/competitions.rb
danieljames-dj: Sorry didn't understand what you meant. If I understand correctly, this is already added in competition factory right?

WcaOnRails/spec/mailers/role_change_mailer_spec.rb
danieljames-dj: When this is changed, the test case is failing. Not sure why, open to discuss what is the flaw in my design. :-)

WcaOnRails/spec/factories/competitions.rb
gregorbg: In the diff for this PR, I see that you are adding the same hook three times.
It is literally copy/paste of the following snippet:

```ruby
  competition.delegates.each do |delegate|
    FactoryBot.create(:senior_delegate, region_id: delegate.region_id)
  end
end
```

Why are you copying this exact same snippet three times?
In my mind, it should be possible to apply this after-create hook to _every_ competition, because the competitions which don't have any `:with_delegate` or similar traits will simply have an empty `competition.delegates` list so the `each` does nothing.

gregorbg: So in summary it won't hurt or cause damage to put this after-create hook in one central place where it always gets executed, instead of copy-pasting it to three different places

WcaOnRails/spec/mailers/role_change_mailer_spec.rb
gregorbg: Found the problem: You're creating the region first, then the Senior Delegate (above during the many `let` calls). This means that at the moment of creation, the region has no Senior and the call in `user.rb#senior_delegates` fails because `UserRole.group(role).senior_delegate` returns nil.

Reloading the Senior forces the group to realise it has a Senior now.

The proper fix is to instantiate your test variables in a way that they already know about each other during creation, i.e. create the Senior Delegate first, and then pass it to the Delegate region directly upon creation.

If you run into circular dependencies, you can resort to some googling (circular dependency of test data), try to look at how other factories in our code solve this problem, or ask ChatGPT.

gregorbg (changes requested):

WcaOnRails/spec/factories/competitions.rb
danieljames-dj: Oh yes that was way cleaner. Didn't think of that way 🙈  Did now

WcaOnRails/spec/mailers/role_change_mailer_spec.rb
gregorbg: You're overthinking. Previously, you had a `let` which created a delegate region without specifying a senior. The Senior simply did not exist when the Delegate region was created, so any code calling region.senior_delegate would receive `nil` as an answer

PR #8876, opened 2024-02-06
Add Github Action to reset Staging Database

FinnIckler:
As we can't run this command on the container, here's a github action that starts a task that resets the database.

I also fixed the link now that it is on S3


FinnIckler:
It's running...

gregorbg (approved):

PR #8877, opened 2024-02-06
Display validation errors when trying to confirm a competition

gregorbg:
Some validations are configured to only run for _confirmed_ competitions (god knows why...).

In other words, certain validation errors can only be displayed _upon confirming_ and we need to propagate them to the global model state in order to display them neatly to the user.


gregorbg:
Merging as a hotfix

PR #8878, opened 2024-02-06
Migrate Senior Delegate to roles

danieljames-dj:


WcaOnRails/db/migrate/20240206154319_migrate_seniors_to_roles.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class MigrateSeniorsToRoles < ActiveRecord::Migration[7.1]
+  def change
+    ActiveRecord::Base.transaction do
+      User.where(delegate_status: 'senior_delegate').each do |senior_delegate|
```
gregorbg, line 6: Nit: you can use `find_each` instead of `each` here.
`each` is a general Ruby language construct that works on any array/list. But in case of Rails database, it requires loading all results completely into memory first, before it can iterate over them.
`find_each` is a Rails-specific method for database queries, which loops over the entries using SQL `LIMIT` and thus is much more efficient because it can load the user rows step by step while you're iterating over them.

WcaOnRails/db/migrate/20240206154319_migrate_seniors_to_roles.rb
```diff
@@ -0,0 +1,21 @@
...
+        region = UserGroup.find(senior_delegate.region_id)
+        raise "Region not found for senior_delegate: #{senior_delegate.name}" unless region
+
+        metadata = RolesMetadataDelegateRegions.create!(status: RolesMetadataDelegateRegions.statuses[:senior_delegate])
+        UserRole.create!(
+          user_id: senior_delegate.id,
+          group_id: senior_delegate.region_id,
+          start_date: Date.today - 10.years, # For now, making the start date as 10 years before for everyone. Jacob will update the correct start date for each senior delegate after migration.
```
gregorbg, line 14: Can you make this a fixed date. like the founding date of the WCA?
This will make the migrated entries much easier to spot. Otherwise, you need to know when the migration was run to know which fake start_dates to look for.

WcaOnRails/db/migrate/20240206154319_migrate_seniors_to_roles.rb
```diff
@@ -0,0 +1,21 @@
...
+          metadata: metadata,
+        )
+        senior_delegate.update!(delegate_status: 'delegate')
```
gregorbg, line 17: Just for context: Why is this necessary, and are you confident that this won't "confuse" other pieces of the code (most notably old pieces that still make a distinction between `delegate` and `senior_delegate`)

WcaOnRails/spec/controllers/api_controller_spec.rb
```diff
@@ -227,20 +227,6 @@
       end
     end
 
-    context 'signed in as senior delegate' do
-      before :each do
-        api_sign_in_as(FactoryBot.create(:senior_delegate))
-      end
-
-      it 'has correct delegate_status' do
```
gregorbg, line 235: Would it make sense to replace this test by an equivalent check?

gregorbg (changes requested):

WcaOnRails/app/models/user.rb
```diff
@@ -48,10 +49,11 @@ class User < ApplicationRecord
   def self.eligible_voters
     team_leaders = TeamMember.current.in_official_team.leader.map(&:user)
     team_senior_members = TeamMember.current.in_official_team.senior_member.map(&:user)
-    eligible_delegates = User.where(delegate_status: %w(delegate senior_delegate))
+    eligible_delegates = User.where(delegate_status: %w(delegate))
```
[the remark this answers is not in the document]
gregorbg: That answers my question surprisingly well :)

WcaOnRails/db/migrate/20240206154319_migrate_seniors_to_roles.rb
gregorbg: Okay, makes sense!

WcaOnRails/spec/controllers/api_controller_spec.rb
gregorbg: Hm, Rails should be smart enough to figure this out and cancel the loop during serialization.
Will do some more reading and get back to you!

spec/factories/competitions.rb
```diff
@@ -355,7 +355,10 @@
     after(:create) do |competition|
       competition.delegates.each do |delegate|
         unless delegate.region_id.nil? # There can be cases where the competition delegate is actually not a delegate (temporary delegate)
-          FactoryBot.create(:senior_delegate, region_id: delegate.region_id)
+          if UserGroup.find(delegate.region_id).lead_user.nil? # Allowing to manually create senior delegate for the delegate if needed.
+            senior_delegate_role = FactoryBot.create(:senior_delegate_role)
+            delegate.update!(region_id: senior_delegate_role.group.id)
```
gregorbg, line 360: When is the value that's passed to this `update` ever different from the original `region_id`?

You're only entering this block if an "original" `region_id` is present (i.e. not nil). Only in that case are you trying to access the `UserGroup`, and only if that user group can be found you're creating a senior delegate for that exact user group.

The Senior Delegate you created will by definition always be the Senior Delegate for the original `region_id`, so I don't understand why this update is necessary.

gregorbg (changes requested):

app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -238,22 +238,19 @@ def index_for_group_type
...
+      role_to_end = group.lead_role
+      if role_to_end.present?
+        role_to_end.update!(end_date: Date.today)
+        RoleChangeMailer.notify_role_end(role_to_end, current_user).deliver_later
       end
     end
   end
 
+  # Whether the delegate status is already migrated to roles.
+  private def is_delegate_status_migrated?(status)
```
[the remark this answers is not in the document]
danieljames-dj: Changed, I knew it but forgot while writing this :-)

app/models/user.rb
```diff
@@ -20,6 +20,7 @@ class User < ApplicationRecord
   belongs_to :unconfirmed_person, -> { where(subId: 1) }, primary_key: "wca_id", foreign_key: "unconfirmed_wca_id", class_name: "Person", optional: true
   belongs_to :delegate_to_handle_wca_id_claim, -> { where.not(delegate_status: nil) }, foreign_key: "delegate_id_to_handle_wca_id_claim", class_name: "User", optional: true
   has_many :roles, class_name: "UserRole"
+  has_many :roles_metadata_delegate_regions, through: :roles, source: :metadata, source_type: "RolesMetadataDelegateRegions"
```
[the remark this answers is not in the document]
danieljames-dj: Oh it was needed for the complicated staff_delegates. Not used now, removed it.

app/models/user.rb
```diff
@@ -1084,7 +1085,7 @@ def self.search(query, params: {})
       search_by_email = ActiveRecord::Type::Boolean.new.cast(params[:email])
 
       if ActiveRecord::Type::Boolean.new.cast(params[:only_staff_delegates])
-        users = users.where(delegate_status: ["candidate_delegate", "delegate", "senior_delegate"])
+        users = User.staff_delegates
```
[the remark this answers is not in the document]

app/models/user.rb
```diff
@@ -1300,7 +1301,7 @@ def delegate_role
       is_active: true,
       group: self.region,
       user: self,
-      is_lead: delegate_status == RolesMetadataDelegateRegions.statuses[:senior_delegate],
+      is_lead: false,
```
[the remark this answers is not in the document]
danieljames-dj: I don't think so, because this should be false for all delegate roles who aren't senior/regional delegate. Even though we can do something like `!delegate_status_migrated`, if we migrate trainee delegate alone in the next phase (which we won't anyway), `is_lead` will become true for trainee delegates which is not correct. Currently since we already migrated all the lead delegates, I'm hard-coding the boolean.

db/migrate/20240206154319_migrate_seniors_to_roles.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class MigrateSeniorsToRoles < ActiveRecord::Migration[7.1]
+  def change
+    ActiveRecord::Base.transaction do
+      User.where(delegate_status: 'senior_delegate').find_each do |senior_delegate|
+        region = UserGroup.find(senior_delegate.region_id)
+        raise "Region not found for senior_delegate: #{senior_delegate.name}" unless region
```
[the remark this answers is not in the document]
danieljames-dj: Done

spec/factories/competitions.rb
danieljames-dj: Basically, if a delegate doesn't have a senior delegate, I'm creating a senior delegate here. But when I create a senior delegate using `FactoryBot.create(:senior_delegate_role)`, it's internally creating a group as well. And then I'm changing the real delegate's group to the new group. I've now changed it the reverse - like creating the senior delegate with the same group of the delegate. I don't remember which was the test affected, but if all tests pass then it's good to go. Otherwise I'll check on it.

app/models/user.rb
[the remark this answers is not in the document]
gregorbg: Okay, thanks for the explanation :)

gregorbg (approved):

Gemfile.lock
```diff
@@ -7,7 +7,7 @@ GIT
 
 GIT
   remote: https://github.com/glebm/i18n-tasks.git
-  revision: 56f378ac102b18c7de2cbcb69cdf2483b0d9f48a
+  revision: 0b4b483c82664f26c5696fb0f6aa1297356e4683
```
gregorbg, line 10: Whoopsie :stuck_out_tongue:

gregorbg (changes requested):

Gemfile.lock
danieljames-dj: Ouch, removed 😅

gregorbg (approved):

PR #8882, opened 2024-02-07
Fix renaming custom activities bug

kr-matthews:
When the issue was filed, renaming did nothing. After my recent PR to copy schedules and bulk-edit activities, renaming throws an error.

This PR 1) fixes that, 2) refactors some code to make it (in my opinion) clearer, and 3) fixes some state clean-up in the activity modal (1 commit per item).


gregorbg (approved):

gregorbg:
Thanks for finding a cleaner approach than I did :smile:

Issue #8880, opened 2024-02-06
Unable to easily remove delegates from cloned competition


2018RIDA01:
When cloning a competition, you are unable to remove delegates and organizers as easily as you once could (with the old style of competition page). Only after it has been cloned and created can you slowly remove delegates one by one (as described later). 

**To Reproduce**
1. On any existing competition, click "Clone".
2. After filling in all required details, click "Create Competition".
3. After competition has been made, go to "Edit -> Organizer View".
4. After scrolling down to the listed delegates section (this bug also happens with Trainee Delegates and Organizer section), click the "X" next to **one** of the delegates/organizers names. Hit "Save Changes" and refresh. After this, you can restart the process from step 3. 

**Expected/Desire Behavior**
It is strongly beneficial to be able to remove several delegates at once (as in the old system) compared to having to refresh/save every time. 

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser: Chrome
 - Version 121.0.6167.140 (latest)

gregorbg:
@2018RIDA01 Thanks for your clean issue description!

I haven't investigated by trying out exactly what you're describing, but just from reading it feels like this might be related to a patch in one of our data loading libraries. I just merged a fix which might have a positive effect on this issue (i.e. might potentially solve it), can you please check?

If this other fix didn't do anything for your issue, please let me know and I will investigate in more detail.

2018RIDA01:
@gregorbg yep everything seems to work now!

PR #8883, opened 2024-02-07
Use Tanstack-Query instead of useManyLoadedData custom hook

gregorbg:
Gets entirely rid of the `useManyLoadedData` hook. Might (potentially?) fix https://github.com/thewca/worldcubeassociation.org/issues/8880, not sure though.

In the long run, we should have the `WcaSearch` component be responsible for fetching its own internal data, and only pass the ID to the outside world. I might give this a shot as a follow-up PR if it's straight-forward to implement.

### DIFF
--- a/WcaOnRails/app/webpacker/components/CompetitionForm/index.js
+++ b/WcaOnRails/app/webpacker/components/CompetitionForm/index.js
@@ -11,6 +11,7 @@ import {
   Message, Sticky,
 } from 'semantic-ui-react';
 import _ from 'lodash';
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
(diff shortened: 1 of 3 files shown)


WcaOnRails/app/webpacker/components/CompetitionForm/index.js
```diff
@@ -318,7 +319,9 @@ export default function Wrapper({
       }}
     >
       <SectionProvider>
-        <CompetitionForm />
+        <QueryClientProvider client={new QueryClient()}>
```
coder13, line 322: One nit is to move the `new QueryClient()` into a const outside of Wrapper that way this object isn't being re-instantiated every time there's a re-render of Wrapper.

PR #8887, opened 2024-02-08
Use wkhtmltopdf-binary gem to install wkhtmltopdf

FinnIckler:
Now that our only use for wkhtmltopdf is to generate competition pdfs, we can use this gem to install it


WcaOnRails/config/initializers/wicked_pdf.rb
```diff
@@ -13,9 +13,8 @@
 WickedPdf.config = {
   # Path to the wkhtmltopdf executable: This usually isn't needed if using
   # one of the wkhtmltopdf-binary family of gems.
-  exe_path: '/usr/local/bin/wkhtmltopdf',
```
gregorbg, line 16: Super nitpick: Comment this out instead of deleting it to mimick how the configuration file looked before your change

gregorbg (approved):

PR #8891, opened 2024-02-08
Host TNoodle signing key through Vault

gregorbg:
One of the remnants of our ECS migration. Failing to load the TNoodle key makes the API return `false` instead of an actual key, and thus lets TNoodle think it's not signed even though it clearly is.

(Yes, TNoodle should handle it better when the API returns `false`, but I cannot be bothered to create a new release just for this purpose.)

### DIFF
--- a/WcaOnRails/app/controllers/api/v0/api_controller.rb
+++ b/WcaOnRails/app/controllers/api/v0/api_controller.rb
@@ -13,7 +13,6 @@ class Api::V0::ApiController < ApplicationController
   end
 
   DEFAULT_API_RESULT_LIMIT = 20
-  TNOODLE_PUBLIC_KEY_PATH = "#{Rails.root}/app/views/regulations/scrambles/tnoodle/TNoodle-WCA.pem".freeze
 
   def me
     render json: { me: current_api_user }, private_attributes: doorkeeper_token.scopes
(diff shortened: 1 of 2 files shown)


FinnIckler:
Doesn't a .pem key usually contain a private and a public key? Which I would think is why you call `rsa_key.public_key.to_der`. In that case I would call the Secret TNOODLE_KEY and not public key

gregorbg:
The contents of a `.pem` file are not clearly defined. In this case, it only contains the public key indeed.
As a matter of fact, we could just hard-add it to this monolith git repo if we wanted, because due to the nature of a **public** key the contents are completely harmless:

```
-----BEGIN PUBLIC KEY-----
MIICIjANBgkqhkiG9w0BAQEFAAOCAg8AMIICCgKCAgEAuhhenWFJfDKCZM6NI8jG
ZbNuBBMPL8l8yC6asrQ+FsRSvxaz4c8NAKI53ZmWeqB8k+ceYrLUXHSqHgwmKK2G
+4+Wn3bZLafvA7OvffJEETs6bAjg2Jris+3AsiZfpQgAaoKp4cyDUA/ff8Sd22nh
2gAnlhgsZNnbfA1dLXWet0VRJA7HaL+5n/59siuUb/kWX8h28L/tHKYKPJZ1kpcu
PHbAGWH1amdkVsxUvyzv1i0SXnSmbcw2ULCtwKjoQy4iGnaGKxBpkG1sIl6Y0lp4
vEFdnD26PbKpI1BZdWCFAyjdsF4Xh7S6Lr1w04By74895Jl6twBL2Y4XB/5aAIQX
Pnr/jVKrVbE8qHfDLhMth/577HoOEhafaKZs+Wq9BgJS22khmvYTatgcxpJDgUL4
d2D/J+APQuf2w7tKDbReYDILCNK4MRhnHYgpfPmQEkBp8LbxK6M0PYwVsHzZA4mG
5F3SQPJ7QPoDQk9YoMsiWTvnOdp4vFPmjEaSbUMtekJlYEYsoPP2/sBSwLeqy7Aa
FUQrZfLmO4L3L4un/W1cEX2epb3vgZW8mY+nO6tB+BODW3mU3NUsJftdwoI7dwKI
vqbWJuisg22sr8dPSd/ULp0L4jEnU4KKXBCOOEKXt5kmDbamQMF+Qo/ViLe80REW
Z8SLAdollxhxYT7G6b9I7QUCAwEAAQ==
-----END PUBLIC KEY-----
```

gregorbg:
Unfortunately this issue currently affects all TNoodle releases for all Delegates, so I'd rather have a quick-and-dirty solution than wasting more time pondering about what would be the most elegant way.

gregorbg:
Merging as a hotfix because the emails already keep coming in
(and the test fail is clearly an unrelated choke, so I'm not going to wait for another 15 minutes)

PR #8898, opened 2024-02-08
Accept empty strings from the competition form date pickers

gregorbg:
See code comment in the diff for further explanation


gregorbg:
Hotfix for another PR

PR #8907, opened 2024-02-12
Fix Competitor Info Route and add Test

FinnIckler:
Should have done this the first time!

### DIFF
--- a/WcaOnRails/app/controllers/api/internal/v1/users_controller.rb
+++ b/WcaOnRails/app/controllers/api/internal/v1/users_controller.rb
@@ -5,7 +5,7 @@ class Api::Internal::V1::UsersController < Api::Internal::V1::ApiController
   protect_from_forgery except: [:competitor_info]
   def competitor_info
     competitors = params.require(:ids)
-    users = User.find_all(competitors)
+    users = User.find(competitors)
     render json: users.to_json({
                                  only: %w[id wca_id name gender country_iso2 email dob],
                                })
--- a/WcaOnRails/app/controllers/api/internal/v1/users_controller.rb
+++ b/WcaOnRails/app/controllers/api/internal/v1/users_controller.rb
@@ -6,8 +6,12 @@ class Api::Internal::V1::UsersController < Api::Internal::V1::ApiController
   def competitor_info
     competitors = params.require(:ids)
     users = User.find(competitors)
-    render json: users.to_json({
-                                 only: %w[id wca_id name gender country_iso2 email dob],
-                               })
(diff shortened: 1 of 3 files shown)


WcaOnRails/app/models/user.rb
```diff
@@ -1116,7 +1116,7 @@ def url
     include: ["avatar", "teams"],
   }.freeze
 
-  def serializable_hash(options = nil)
+  def serializable_hash(options = nil, overwrite_default = false)
```
gregorbg: Sorry but veto.
We will not override the way Rails behaves -- fighting with Rails conventions is a bad idea because Rails always wins in the end.

I will submit another comment showing you how to achieve your desired behavior.

WcaOnRails/app/controllers/api/internal/v1/users_controller.rb
```diff
@@ -5,9 +5,13 @@ class Api::Internal::V1::UsersController < Api::Internal::V1::ApiController
...
-    render json: users.to_json({
-                                 only: %w[id wca_id name gender country_iso2 email dob],
-                               })
+    users = User.find(competitors)
+    data = users.map { |u|
+      u.serializable_hash({
+                            only: %w[id wca_id name gender country_iso2 email dob],
+                          },
+                          overwrite_default: true)
+    }
```
gregorbg, line 14: (If GitHub renders this comment at the top of the list, read my other comment first)

Please retain the old `to_json` invocation. The merge between the `DEFAULT_SERIALIZE_OPTIONS` and your custom options is **intentionally not** recursive. If you don't care about certain fields, you can skip them by providing an empty array.

gregorbg (changes requested):

WcaOnRails/app/controllers/api/internal/v1/users_controller.rb
```diff
@@ -5,9 +5,13 @@ class Api::Internal::V1::UsersController < Api::Internal::V1::ApiController
...
+    render json: data
```
gregorbg, line 15:
```suggestion
    render json: users.to_json({
                                 only: %w[id wca_id name gender country_iso2 email dob],
                                 methods: [],
                                 include: [],
                               })
```

WcaOnRails/app/controllers/api/internal/v1/users_controller.rb
FinnIckler: thanks! I was trying to find a solution like this!

gregorbg (approved):

PR #8918, opened 2024-02-16
Migrate sign up page to react

danieljames-dj:


WcaOnRails/app/webpacker/components/Auth/Signup/index.jsx
```diff
@@ -0,0 +1,181 @@
...
+import {
+  Button, Card, CardHeader, Container, Form, FormField, Message,
+} from 'semantic-ui-react';
+import ReCAPTCHA from 'react-google-recaptcha';
+import { GOOGLE_RECAPTCHA_SITE_KEY } from '../../../lib/wca-data.js.erb';
+import I18n from '../../../lib/i18n';
+
+export default function Signup() {
+  // initial is a minor hack to prevent fomantic CSS messing up the UI.
+  const [initial, setInitial] = useState(true);
```
kr-matthews, line 11: I'm in favour of naming things after what they are, rather then how they're used, so I'd recommend `hasClickedSignUpButton` instead of `initial`. (Note that this is the opposite boolean.)

WcaOnRails/app/webpacker/components/Auth/Signup/index.jsx
```diff
@@ -0,0 +1,181 @@
...
+    },
+    confirmPassword: {
+      error: null,
+      id: 'confirmPassword',
+      value: '',
+    },
+  });
+  const [captchaError, setCaptchaError] = useState(false);
+  const recaptchaRef = useRef();
+  const csrfToken = document.querySelector('meta[name=csrf-token]').content;
```
kr-matthews, line 31: This doesn't seem very React-like.

WcaOnRails/app/webpacker/components/Auth/Signup/index.jsx
```diff
@@ -0,0 +1,181 @@
...
+      [id]: {
+        ...formValues[id],
+        error,
+      },
+    });
+  };
+
+  const validation = {
+    email: () => {
+      const emailRegex = /^[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+@[a-zA-Z0-9-]+(?:\.[a-zA-Z0-9-]+)*$/;
```
coder13, line 45: Can we move constants like regexes into their own file?

coder13:
This PR is also making me think about the current security of the WCA website. A minimum of 3 characters for a password seems unsafe but there's also better models of security (such as passwordless) that we could consider so perhaps food for thought for later.

WcaOnRails/app/webpacker/components/Auth/Signup/index.jsx
danieljames-dj: Makes sense, changed.

WcaOnRails/app/webpacker/components/Auth/Signup/index.jsx
danieljames-dj: Ya that would be better. Done.

danieljames-dj:
> This PR is also making me think about the current security of the WCA website. A minimum of 3 characters for a password seems unsafe but there's also better models of security (such as passwordless) that we could consider so perhaps food for thought for later.

Agreed, we should do it sometime in future.

WcaOnRails/app/webpacker/components/Auth/Signup/index.jsx
danieljames-dj: Ya, but couldn't figure out an alternative.

danieljames-dj:
Closing this as I won't be able to work on fixing the UI unit tests.

PR #8919, opened 2024-02-17
Competitions overview search params in url

kr-matthews:
- Any non-default value is stored as a search param in the url. Any default value is omitted for simplicity. I mimicked the incidents log work that I did a few years ago.
- Bookmarking a url should work.
- There are some breaking changes with existing bookmarks for the old page. See the comment in `filterUtils`; most filter keys aren't affected.
- Fixed a console warning about size=10 not being valid.

There are a lot of params so there may be edge cases that I didn't think of; please play around with the page, ie change a bunch of state and then refresh the page.

Follow-up work:
- If we decide that the braking changes for existing bookmarks is a problem, then there would need to be a follow-up PR for that. The work would build on this PR, not replace anything, so I think it still makes sense to merge this regardless.
- Some kind of UI error indication (eg red border) on any input that has an invalid state (the user should only be able to get there by typing in a fake search param like `region=my_fake_region`).
- Add a button to reset filters to default values.

I'd be happy to work on the last 2 once this is merged.

### DIFF
--- a/WcaOnRails/app/webpacker/components/CompetitionsOverview/CompetitionsFilters.js
+++ b/WcaOnRails/app/webpacker/components/CompetitionsOverview/CompetitionsFilters.js
@@ -117,7 +117,7 @@ function EventSelector({ selectedEvents, dispatchFilter }) {
 
 function RegionSelector({ region, dispatchFilter }) {
   const regionsOptions = [
-    { key: 'all', text: I18n.t('common.all_regions'), value: 'all_regions' },
+    { key: 'all', text: I18n.t('common.all_regions'), value: 'all' },
     {
       key: 'continents_header', value: '', disabled: true, content: <Header content={I18n.t('common.continent')} size="small" style={{ textAlign: 'center' }} />,
     },
--- a/WcaOnRails/app/webpacker/components/CompetitionsOverview/CompetitionsFilters.js
+++ b/WcaOnRails/app/webpacker/components/CompetitionsOverview/CompetitionsFilters.js
@@ -49,7 +49,7 @@ function CompetitionsFilters({
 
       <Form.Group>
         <Form.Field width={8}>
-          <DelegateSelector delegate={filterState.delegate} dispatchFilter={dispatchFilter} />
+          <DelegateSelector delegateId={filterState.delegate} dispatchFilter={dispatchFilter} />
         </Form.Field>
       </Form.Group>
 
@@ -162,7 +162,7 @@ function SearchBar({ text, dispatchFilter }) {
   );
 }
 
-function DelegateSelector({ delegate, dispatchFilter }) {
+function DelegateSelector({ delegateId, dispatchFilter }) {
   const { delegatesLoading, delegatesData } = useDelegatesData();
 
   return (
@@ -187,7 +187,7 @@ function DelegateSelector({ delegate, dispatchFilter }) {
             image: { avatar: true, src: delegate.avatar?.thumb_url, style: { width: '28px', height: '28px' } },
           }
         )) || [])]}
-        value={delegate}
+        value={delegateId}
         onChange={(_, data) => dispatchFilter({ delegate: data.value })}
         noResultsMessage={delegatesLoading ? I18n.t('competitions.index.delegates_loading') : I18n.t('competitions.index.no_delegates_found')}
       />
--- a/WcaOnRails/app/webpacker/components/CompetitionsOverview/CompetitionsFilters.js
+++ b/WcaOnRails/app/webpacker/components/CompetitionsOverview/CompetitionsFilters.js
@@ -168,7 +168,7 @@ function DelegateSelector({ dispatchFilter }) {
     <>
       <div style={{ display: 'inline-block' }}>
         <label htmlFor="delegate">{I18n.t('layouts.navigation.delegate')}</label>
-        {delegatesLoading && <PulseLoader size="10" cssOverride={{ marginLeft: '5px' }} />}
+        {delegatesLoading && <PulseLoader size="10px" cssOverride={{ marginLeft: '5px' }} />}
       </div>
       <Dropdown
         name="delegate"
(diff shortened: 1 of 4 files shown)


WcaOnRails/app/webpacker/components/CompetitionsOverview/CompetitionsFilters.js
```diff
@@ -138,15 +138,15 @@ function RegionSelector({ dispatchFilter }) {
       <Dropdown
         search
         selection
-        defaultValue="all"
+        value={region}
```
gregorbg, line 141: Why are you replacing the `defaultValue` entirely?
Just looking at the code, it feels like we should keep the default value and also set `value={region}` so that it gets overridden if necessary/present.

WcaOnRails/app/webpacker/components/CompetitionsOverview/filterUtils.js
```diff
@@ -1,19 +1,105 @@
...
+  if (shouldIncludeCancelled) {
+    searchParams.set(INCLUDE_CANCELLED, INCLUDE_CANCELLED_TRUE);
+  } else {
+    searchParams.delete(INCLUDE_CANCELLED);
+  }
+
+  window.history.replaceState({}, '', `${window.location.pathname}?${searchParams}`);
 };
 
+const WCA_EVENT_IDS = Object.values(events.official).map((e) => e.id);
```
gregorbg, line 101: We also have `events.byId` which allows you to simply access `Object.keys(events.byId)` and save an awkward `map` :smile:

WcaOnRails/app/webpacker/components/CompetitionsOverview/filterUtils.js
```diff
@@ -1,19 +1,105 @@
...
+  searchParams.set(SELECTED_EVENTS, selectedEvents.join(','));
+  searchParams.delete(SELECTED_EVENTS, DEFAULT_EVENTS.join(','));
+  if (shouldIncludeCancelled) {
+    searchParams.set(INCLUDE_CANCELLED, INCLUDE_CANCELLED_TRUE);
+  } else {
+    searchParams.delete(INCLUDE_CANCELLED);
+  }
+
+  window.history.replaceState({}, '', `${window.location.pathname}?${searchParams}`);
 };
```
gregorbg, line 99: Please add some line breaks to this block and reorganize it a bit:
- First, all the "primitive" valuesthat follow a simple two-line set-delete pattern. Include a comment about what this pattern does.
- Then all the values that are wrapped inside an `if`.
- Lastly, why do you use `window.history.replaceState` instead of just overriding `window.location.search`?

WcaOnRails/app/webpacker/components/CompetitionsOverview/CompetitionsView.js
```diff
@@ -11,23 +11,42 @@ import { fetchJsonOrError } from '../../lib/requests/fetchWithAuthenticityToken'
...
+  );
+
+  const [filterState, dispatchFilter] = useReducer(
+    filterReducer,
+    searchParams,
+    createFilterState,
+  );
   const debouncedFilterState = useDebounce(filterState, DEBOUNCE_MS);
-  const [displayMode, setDisplayMode] = useState('list');
+  const [displayMode, setDisplayMode] = useState(() => getDisplayMode(searchParams));
```
gregorbg, line 38: Why are you passing this as a callback, instead of `useState(getDisplayMode(searchParams))`?

gregorbg (changes requested):

WcaOnRails/app/webpacker/components/CompetitionsOverview/CompetitionsFilters.js
kr-matthews: You're not meant to do that; default value is for uncontrolled components, value is for controlled components. Console warning:
> "value" is auto controlled. Specify either defaultValue or value, but not both.

WcaOnRails/app/webpacker/components/CompetitionsOverview/filterUtils.js
kr-matthews: This is shown as an added line, but it was just moved (unaltered) by me.
But that's great, I will change it.

WcaOnRails/app/webpacker/components/CompetitionsOverview/CompetitionsView.js
kr-matthews: No reason in particular. It saves the function `getDisplayMode` from evaluating, but that's negligible.

WcaOnRails/app/webpacker/components/CompetitionsOverview/filterUtils.js
kr-matthews: I use `window.history.replaceState` because I was following what was done in the incidents log. I don't remember why I did it this way there, might have just been the first thing I found that worked, or possibly your suggested way didn't work for some reason.

gregorbg: Okay, for now let's assume that you had a good reason to do it this way in the incidents log and trust your past self :)

WcaOnRails/app/webpacker/components/CompetitionsOverview/CompetitionsFilters.js
gregorbg: Oh, I wasn't aware of that. In this case, nevermind :smile:

WcaOnRails/app/webpacker/components/CompetitionsOverview/CompetitionsView.js
gregorbg: TIL https://react.dev/reference/react/useState#avoiding-recreating-the-initial-state is a thing, thanks :)

gregorbg (approved):

PR #8920, opened 2024-02-17
Add registrations from the new Registration Service to the my competitions page

FinnIckler:
This works. The only issue is the tooltip. Which is using `get_registration_status_message_if_registered`. Which for some reason is fetching the registration_status again, even though we already have it here. 

I can rewrite the `competition_message_for_user` to also take the registration to fix this. Thoughts?

### DIFF
--- a/WcaOnRails/app/controllers/competitions_controller.rb
+++ b/WcaOnRails/app/controllers/competitions_controller.rb
@@ -870,13 +870,12 @@ class CompetitionsController < ApplicationController
       competition_ids.concat(current_user.delegated_competitions.pluck(:competition_id))
       registrations = current_user.registrations.includes(:competition).accepted.reject { |r| r.competition.results_posted? }
       registrations.concat(current_user.registrations.includes(:competition).pending.select { |r| r.competition.upcoming? })
-      # Convert Registrations V2 to a format that the frontend can understand
-      registrations.concat(registrations_v2.map { |r| Microservices::Registrations.convert_registration(r['competition_id'], current_user.id, r['status']) })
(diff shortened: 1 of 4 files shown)


gregorbg:
> I can rewrite the `competition_message_for_user` to also take the registration to fix this. Thoughts?

Yup, but please be as careful as possible about changing the signature. My idea would be something along the lines of:

```ruby
def competition_message_for_user(competition, user, registration = nil)
  messages_to_join << get_registration_status_message_if_registered(competition, user, registration)
end

def get_registration_status_message_if_registered(competition, user, registration = nil)
  registration_status = registration || competition.registrations.find_by_user_id(user.id)
end
```

gregorbg (approved):
LGTM, just one merge conflict left :smile:

FinnIckler:
RuboCop doesn't like my OpenStruct usage, I thought it was the most elegant solution. I'll just use a struct instead then

gregorbg (changes requested):
You have two choices:
- Tell RuboCop to shut up about the OpenStruct
- Make your struct approach more database-friendly. Right now, you're using the correctly loaded `current_user` variable which contains a `User` model object, read its ID, pass that ID to the new Struct, and read the same User from the DB again using that ID. This is definitely too much. (Same argument applies to `competition_id` parameter)

FinnIckler:
In this case this is true. But if you look at the email PR, I just have the `user_id` and `competition_id` fields. 
I could check if they are already models and then read them if they are?

gregorbg (approved):

PR #8921, opened 2024-02-18
Add Internal Route to send Registration emails

FinnIckler:
So we can trigger emails from the registration service.
This depends on the convert method in the other PR


WcaOnRails/app/controllers/api/internal/v1/mailer_controller.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+class Api::Internal::V1::MailerController < Api::Internal::V1::ApiController
+  # We are using our own authentication method with vault
+  protect_from_forgery except: [:competitor_info]
```
gregorbg, line 5: Copy-paste whoopsie?

WcaOnRails/app/controllers/api/internal/v1/mailer_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+  def registration
+    registration_status = params.require(:status)
+    # Either create or update, we need this to send the correct emails if organizers move
```
gregorbg, line 8:
```suggestion
    # Either "create" or "update", we need this to send the correct emails if organizers move
```

WcaOnRails/app/controllers/api/internal/v1/mailer_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+    # Either "create" or "update", we need this to send the correct emails if organizers move
+    # a competitor back to pending
+    registration_action = params.require(:action)
+    registration_user = params.require(:user_id)
+    # So we know who to email on delete
+    requesting_user = params.require(:current_user)
+    registration_competition = params.require(:competition_id)
+    competition = Competition.find(registration_competition)
+    user = User.find(registration_user)
+    converted_registration = convert_registration(competition, user, registration_status)
```
gregorbg, line 17: You should use the fully qualified name here, presumably `Microservices::Registration.convert_registration`

gregorbg (changes requested):

gregorbg (approved):

PR #8923, opened 2024-02-18
Migrate claim WCA ID section of signup page

danieljames-dj:


WcaOnRails/app/webpacker/components/Auth/Signup/ClaimWcaId.jsx
```diff
@@ -0,0 +1,181 @@
...
+      queryFn: () => fetchJsonOrError(apiV0Urls.persons.likelyDelegates(unconfirmedWcaId)),
+      enabled: !!unconfirmedWcaId,
+    },
+  );
+  const { data: { can_claim: canClaim, reason } } = canClaimData || { data: {} };
+  const { data: { likely_delegates: likelyDelegates } } = likelyDelegatesData || { data: {} };
+  const loading = isCanClaimLoading || isLikelyDelegatesLoading;
+
+  useEffect(() => {
+    if (unconfirmedWcaId) {
```
coder13, line 54: I think this should be wrapped around a debounce?

danieljames-dj: That was a good catch. I didn't think of this case. I assumed this will cause useEffect to get called in every re-render as the value is getting changed. But to my surprise it's working fine - ie, it's getting changed only when the unconfirmedWcaId is changed. I think it's because `unconfirmedPerson` is a state and hence value of `unconfirmedWcaId` is same in re-render because it's an object of `unconfirmedPerson`. Hence, I assume there is no change needed as of now. Please let me know if I missed something.

WcaOnRails/spec/features/sign_up_spec.rb
gregorbg, line 1: Not happy at all about these tests being deleted.
If your React component behaves correctly, it should just serve as a drop-in replacement that doesn't affect the success of these tests. Why did you delete them all?

WcaOnRails/app/webpacker/components/Auth/Signup/ClaimWcaId.jsx
```diff
@@ -0,0 +1,181 @@
...
+            />
+          )}
+        </>
+      )}
+      <Form.Field
+        control={Input}
+        name="user[delegate_id_to_handle_wca_id_claim]"
+        style={{ display: 'none' }}
+        value={delegateToVerify?.id}
+      />
```
gregorbg, line 154: There is `Form.Input` directly as a component.

WcaOnRails/app/webpacker/components/QueryClientProviderWrapper/index.jsx
gregorbg, line 1: Seems like you are defining this file but not using it...?

gregorbg (changes requested):

WcaOnRails/spec/features/sign_up_spec.rb
danieljames-dj: Attaching a screenshot from testcase:

image

Unfortunately it's not rendering component rendered by react_component. I think I have experienced the same thing earlier as well.

Also, considering that we have a plan of decouple frontend and backend completely, it will be good to get the UI related test cases removed from the backend right?

gregorbg: Judging by the screenshot, you're probably not running the JS necessary to inject the React component.
Change the test case(s) from `it "tests something" do ... end` to `it "tests something", js: true do ... end` and it should work.

Also, yes in the long run splitting the test suite up will be good. But "splitting up" and "removing them without replacement" are two different stories... If we already had a separate frontend, and we had a place where you can _move_ these tests, it would be fine. As it stands right now, you're _removing_ them, which is not fine.

danieljames-dj: Did that, but no change. I also noticed that the React component is there at the bottom (if you look very carefully at the screenshot), but the height of the div is too small that the react component got overflowed. I'm trying to figure out a way to get more height for that div.

But is it fine if I add back the tests in the next PR (https://github.com/thewca/worldcubeassociation.org/pull/8918) because I'll anyway change the parent components there and at that time the UI has to be changed completely. I know it's not ideal because there will be few days where the test cases aren't there but can you please grand it as one-time exception?

WcaOnRails/app/webpacker/components/Auth/Signup/ClaimWcaId.jsx
```diff
@@ -0,0 +1,179 @@
...
+    // DatePicker inside semantic FormField works with proper CSS only if it's inside semantic Form.
+    // But currently it's inside rails form, hence to make the CSS work, <div className="ui form">
+    // is added instead of fragment.
+    <div className="ui form">
+      <Form.Field
+        control={WcaSearch}
+        label={I18n.t('common.user.wca_id')}
+        model="person"
+        multiple={false}
+        onChange={(_, { value }) => setUnconfirmedPerson(value)}
```
gregorbg, line 76: `useInputState` hook to allow `onChange={setUnconfirmedPerson}`

WcaOnRails/app/webpacker/components/Auth/Signup/ClaimWcaId.jsx
```diff
@@ -0,0 +1,179 @@
...
+                type="radio"
+              />
+            </FormGroup>
+          )}
+          {manualDelegateSelection && (
+            <Form.Field
+              control={WcaSearch}
+              model="user"
+              multiple={false}
+              onChange={(_, { value }) => setDelegateToVerify(value)}
```
gregorbg, line 141: `useInputState` hook to allow `onChange={setDelegateToVerify}`

gregorbg (changes requested):
React nitpicks :)

WcaOnRails/app/webpacker/components/Auth/Signup/ClaimWcaId.jsx
danieljames-dj: Done both

gregorbg:
The changes in this PR would normally be good to go, but I am extremely unhappy about removing so many important test cases.

Is there any chance you can (relatively) easily merge the changes from this commit over into the bigger "migrate sign-up page to React" commit and continue your work -- with the tests still in place and not removed -- over there?

PR #8927, opened 2024-02-19
Paypal payment form + payment logic integration

dunkOnIT:
_(Lots of commits because this is forked off the unmerged Stripe Refactor PR)_

This PR will cover 2 changes: 
- [x] Creating a separate paypal payment form, rendered based on which payment account is connected
- [x] Recognizing that a payment has taken place on our backend logic

Changes which could be included:
- [x] Support for different currency types (ie, `paypal_amount` function)
  - [ ] Support for currency conversion? ie, allowing fees to be charged in an unsupported currency, but then converting them to USD or something and charging with that
- [ ] Validation to stop organizer adding more than 1 account
- [x] Make sure that both payment forms can't be loaded at once

Actions:
- [x] get payments working again (lol?)
- [x] Send subtotal to Paypal (instead of registration fee)
- [x] Recognize payment after Paypal approves it
- [x] Make sure there is appropriate logging of paypal events
- [ ] Look at Stripe logic for exceptions etc
- [ ] Consider adding webhook support? (No, this should probably happen in a refactor?)
- [x] Paypal payment form
  - [x] Duplicate the existing form
  - [x] Stop errors from showing up
  - [x] Move subtotal
  - [x] Remove payment button
  - [x] Have donations still work
  - [x] Populate subtotal immediately upon load
- [ ] Finish TODOs in tests for paypal supported amounts

Questions:

Payment form improvements:
- [ ] Don't submit the form on Enter
- [ ] Does it even need to be a form for Paypal? Maybe we should remove the form components and display the data differently
- [ ] Uncheck the "Donation" tickbox on page reload


coder13:
What are the simplest set of features in this PR and could they be merged behind a feature flag?

That way we don't have a large PR.

I can also see the argument of: we could drop dead at any time so having some paypal code in the codebase but incomplete could be confusing.

dunkOnIT:
I think we're already pretty close to that core set of features - not sure if you have the context, but this PR is one in a chain of Stripe/Paypal-related PR's, and we already are using not-quite-feature-flags (`Rails.env.production` checks) to introduce Paypal functionality into the master branch ([example here](https://github.com/thewca/worldcubeassociation.org/blob/d457193fe8f0164389e8ec5162df32bb5b354f06/WcaOnRails/app/controllers/competitions_controller.rb#L320))

This PR currently looks much bigger than it is because it forks from #8865 - once that's merged into master, it'll look a lot more modest.

Still very open to feedback/suggestions, just seemed like you might not have some of the context!

dunkOnIT:
@gregorbg PR is now ready for review. A few issues/notes:
1. #8944 should be merged before this
2. This is a very basic implementation which doesn't cover any edge cases - didn't want to put any time into exception handling at this stage of the proof of concept
3. In a similar vein, tests could be refactored (setup logic could be split into functions, objects could be created by factories instead of API calls, webmock payloads could be defined in a separate file) but I figure significant changes will happen during the next PR, so don't want to unnecessarily refactor now
4. **Migrations** - originally, `20240222134242_paypal_transaction.rb` did not include the `transaction_type` column - I added this in a later migration, but during the cleanup I tried to have `transaction_type` be created by the original migration. This doesn't seem to have worked for me locally (its not showing up in dbeaver, and it isn't updating on schema.rb) - despite having run a db:reset (to drop/recreate the tables?) and having done db:rollback and db:migrate several times. Not sure what the issue is.

WcaOnRails/app/controllers/competitions_controller.rb
```diff
@@ -314,17 +314,18 @@ def payment_setup
 
     # Paypal setup URL
     # TODO: Don't generate this URL if there is already a connected paypal account?
-    @paypal_onboarding_url = PaypalInterface.generate_paypal_onboarding_link(@competition.id) unless Rails.env.production?
+    @paypal_onboarding_url = PaypalInterface.generate_paypal_onboarding_link(@competition.id) unless EnvConfig.WCA_LIVE_SITE?
   end
 
   def paypal_return
-    if Rails.env.production?
+    competition = competition_from_params
```
gregorbg, line 321: You're calling `competition_from_params` here and after the subsequent `if` check, which is unnecessary.

WcaOnRails/app/controllers/competitions_controller.rb
```diff
@@ -314,17 +314,18 @@ def payment_setup
...
+
+    if EnvConfig.WCA_LIVE_SITE?
       flash[:error] = 'PayPal is not yet available in production environments'
-      return redirect_to competitions_payment_setup_path(@competition)
+      return redirect_to competitions_payment_setup_path(competition)
     end
 
     competition = competition_from_params
```
gregorbg, line 328: Here's the duplicate call, see comment above.

WcaOnRails/app/controllers/competitions_controller.rb
```diff
@@ -314,17 +314,18 @@ def payment_setup
 
     # Paypal setup URL
     # TODO: Don't generate this URL if there is already a connected paypal account?
-    @paypal_onboarding_url = PaypalInterface.generate_paypal_onboarding_link(@competition.id) unless Rails.env.production?
+    @paypal_onboarding_url = PaypalInterface.generate_paypal_onboarding_link(@competition.id) unless EnvConfig.WCA_LIVE_SITE?
```
gregorbg, line 317: If your goal is to allwo tests in staging, you should perform _both_ checks to be sure. Something along the lines of `unless Rails.env.production? && EnvConfig.WCA_LIVE_SITE?`

WcaOnRails/app/controllers/competitions_controller.rb
```diff
@@ -314,17 +314,18 @@ def payment_setup
...
   end
 
   def paypal_return
-    if Rails.env.production?
+    competition = competition_from_params
+
+    if EnvConfig.WCA_LIVE_SITE?
```
gregorbg, line 323: Double-check, see other comment.

WcaOnRails/app/controllers/competitions_controller.rb
```diff
@@ -373,7 +374,7 @@ def stripe_connect
   end
 
   def disconnect_paypal
-    if Rails.env.production?
+    if EnvConfig.WCA_LIVE_SITE?
```
gregorbg, line 377: Double-check, see other comment.

WcaOnRails/app/controllers/registrations_controller.rb
```diff
@@ -828,19 +828,56 @@ def create
   end
 
   def create_paypal_order
-    return head :forbidden if Rails.env.production?
+    return head :forbidden if EnvConfig.WCA_LIVE_SITE?
```
gregorbg, line 831: Double-check, see other comment.

WcaOnRails/app/controllers/registrations_controller.rb
```diff
@@ -828,19 +828,56 @@ def create
...
 
     @registration = registration_from_params
-    render json: PaypalInterface.create_order(@registration)
+    render json: PaypalInterface.create_order(@registration, params[:total_charge], params[:currency_code])
```
gregorbg, line 834: I feel like the `currency_code` parameter might be redundant, as it can be safely derived from the `registration` object and it won't change even when a donation is being made.

(The `total_charge` parameter is justified because of donations, i.e. when a volunteer donation is being made the actual PayPal charge can differ from the amount affiliated with the registration. But for currencies, this is not the case.)

WcaOnRails/app/controllers/registrations_controller.rb
```diff
@@ -828,19 +828,56 @@ def create
...
   end
 
   def capture_paypal_payment
-    return head :forbidden if Rails.env.production?
+    return head :forbidden if EnvConfig.WCA_LIVE_SITE?
```
gregorbg, line 838: Double-check, see other comment.

WcaOnRails/app/models/registration.rb
```diff
@@ -335,6 +335,14 @@ def series_registration_info
                                    .join(" + ")
   end
 
+  # NOTE: This is really bad, only for a proof of concept. It assumes that there is only ONE PaypalTransaction and ONE PaypalCapture - not safe assumptions.
+  def get_capture_id
+    registration_payment = registration_payments.first
+    receipt = registration_payment.receipt
+    capture = receipt.paypal_captures.first
+    capture.capture_id
+  end
```
gregorbg, line 344: I would at least move this down one step to the `registration_payment` level, so that every pament has one capture ID. Refunds should always be issued based on payments, never based on registrations.

WcaOnRails/app/models/registration.rb
```diff
@@ -335,6 +335,14 @@ def series_registration_info
                                    .join(" + ")
   end
 
+  # NOTE: This is really bad, only for a proof of concept. It assumes that there is only ONE PaypalTransaction and ONE PaypalCapture - not safe assumptions.
+  def get_capture_id
```
gregorbg, line 339: In Ruby, convention dictates that method names are never prefixed with `get_` or `set_`.
This should just be called `def capture_id`

WcaOnRails/app/views/competitions/payment_setup.html.erb
```diff
@@ -12,7 +12,7 @@
         "https://support.stripe.com/questions/which-currencies-does-stripe-support#supportedcurrencies",
         target: '_blank'
       ) %></li>
-      <% unless Rails.env.production? %>
+      <% unless EnvConfig.WCA_LIVE_SITE? %>
```
gregorbg, line 15: Double-check, see other comment.

WcaOnRails/app/views/competitions/payment_setup.html.erb
```diff
@@ -54,7 +54,7 @@
 
   <% end %>
 
-  <% unless Rails.env.production? %>
+  <% unless EnvConfig.WCA_LIVE_SITE? %>
```
gregorbg, line 57: Double-check, see other comment.

WcaOnRails/app/views/registrations/_payment_form.html.erb
```diff
@@ -1,278 +1,3 @@
-<% add_to_packs("auto_numeric") %>
-<%= horizontal_simple_form_for :payment, url: "", html: { id: :form_payment } do |f| %>
-  <%= f.input :fees_paid, label: t('registrations.payment_form.labels.fees_paid'), hint: false do %>
-    <p class="form-control-static"><%= format_money(@registration.paid_entry_fees) %></p>
-  <% end %>
-  <%= f.input :fees_remaining, label: t('registrations.payment_form.labels.fees_remaining'), hint: false do %>
-    <p class="form-control-static"><%= format_money(@registration.outstanding_entry_fees) %></p>
-  <% end %>
+<%= render "stripe_payment_form" if @competition.stripe_connected? %>
+<%= render "paypal_payment_form" if @competition.paypal_connected? && !EnvConfig.WCA_LIVE_SITE? && !@competition.stripe_connected? %>
```
gregorbg, line 2: 1. Double-check, see other comment.
2. The additional ` && !@competition.stripe_connected` can be reprhased much more succintly by wrapping this rendering logic in a big if-elsif. `if @competition.stripe_connected render_stripe elsif @competition.paypal_connected? render_paypal`

WcaOnRails/app/views/registrations/edit.html.erb
```diff
@@ -55,7 +55,12 @@
           offering the refund payment button.
           %>
           <% if @competition.using_payment_integrations? && payment.amount_available_for_refund > 0 %>
-            <%= horizontal_simple_form_for :payment, url: registration_payment_refund_path(@registration, payment), html: { id: :form_refund } do |f| %>
+            <% if @competition.stripe_connected? %>
+              <% refund_url = registration_payment_refund_path(@registration, payment) %>
+            <% else %>
+              <% refund_url = paypal_payment_refund_path(@registration, payment) %>
```
gregorbg, line 61: Re one of my other comments above: You're passing in `payment` to the route helper here, which means that the controller not only knows the registration being refinded but also the exact `RegistrationPayment`. Meaning you can just access that one `receipt` directly. Take a look at the existing Stripe redirect controller method if you need breadcrumbs.

WcaOnRails/config/routes.rb
```diff
@@ -19,11 +19,12 @@
   end
 
   # Don't expose Paypal routes in production until we're reading to launch
-  unless Rails.env.production?
+  unless EnvConfig.WCA_LIVE_SITE?
```
gregorbg, line 22: Double-check, see other comment.

WcaOnRails/db/migrate/20240223032610_add_paypal_capture.rb
gregorbg, line 1: This table is completely useless if you aleady have `PaypalTransaction`. Why not store captures over there with `transaction_type='capture'`?

WcaOnRails/app/models/paypal_transaction.rb
```diff
@@ -0,0 +1,59 @@
...
+      "USD",
+    ],
+    cents_only: [ # Currencies that do not support decimals - should be passed as cents
+      "JPY",
+      "HUF",
+      "TWD",
+    ],
+  }.freeze
+
+  def self.get_paypal_amount(amount_in_cents, currency_code)
```
gregorbg, line 44: Naming convention: `def paypal_amount`. No `get_` in Ruby because... Ruby.

WcaOnRails/app/models/paypal_transaction.rb
```diff
@@ -0,0 +1,59 @@
...
+    if PAYPAL_CURRENCY_CATEGORIES[:decimal].include?(currency_code)
+      format("%.2f", amount_in_cents.to_i / 100.0)
+    else
+      amount_in_cents
+    end
+  end
+
+  def self.get_amount_in_cents(paypal_amount, currency_code)
```
gregorbg, line 52: `def amount_in_cents`, see other comment

WcaOnRails/lib/paypal_interface.rb
```diff
@@ -96,11 +140,13 @@ def self.capture_payment(competition, order_id)
 
       # Logs requests and responses.
       # By default, it only logs the request method and URL, and the request/response headers.
-      builder.response :logger
+      builder.response :logger, ::Logger.new($stdout), bodies: true if Rails.env.development?
     end
   end
 
   private_class_method def self.generate_access_token
+    return '' if EnvConfig.SKIP_PAYPAL_AUTH?
```
gregorbg, line 148: Heavens no. Big fat veto. What are you trying to accomplish here and why does `Rails.env.test?` not cut it?

gregorbg (changes requested):
There's a lot of comments about replacing `Rails.env.production?` with `EnvVars.WCA_LIVE_SITE?` and I'm making an argument that you should do both.

In order to avoid code duplication, you can consider simply adding a helper method to `PaypalInterface` that internally checks whether Paypal is availabe/enabled and then replace all those checks throughout the code by `unless PaypalInterface.paypal_enabled?` or such.

WcaOnRails/app/controllers/competitions_controller.rb
dunkOnIT: Changed to `paypal_enabled?`

WcaOnRails/app/controllers/competitions_controller.rb
dunkOnIT: Changed to `paypal_enabled?`

WcaOnRails/app/controllers/competitions_controller.rb
dunkOnIT: Changed to `paypal_enabled?`

WcaOnRails/app/controllers/registrations_controller.rb
dunkOnIT: Changed to `paypal_enabled?`

WcaOnRails/app/views/competitions/payment_setup.html.erb
dunkOnIT: Changed to `paypal_enabled?`

WcaOnRails/app/views/competitions/payment_setup.html.erb
dunkOnIT: Changed to `paypal_enabled?`

WcaOnRails/config/routes.rb
dunkOnIT: replaced with `paypal_enabled?`

app/views/registrations/_payment_form.html.erb
gregorbg, line 1: Missing an `<% end %>` to close the if-elsif

WcaOnRails/db/migrate/20240223032610_add_paypal_capture.rb
dunkOnIT: I'm not a huge fan of that design, for two reasons:
1. Naming (easily fixed) - the definition of "Transaction" is being stretched if we're storing refunds, orders and captures on it, resulting in a confusing model. Something like `PaypalInteraction` or `PaypalAPIObject` might be more appropriate if we wanted to do that.
2. Data modelling - a `capture` is always a child object of the `order` parent (represented by PaypalTransaction). It seems much more intuitive to represent this relationship accurately in the modelling?

WcaOnRails/app/models/paypal_transaction.rb
dunkOnIT: Didn't know that, thanks!

WcaOnRails/app/models/registration.rb
dunkOnIT: Only issue is that it's pretty long-winded to get hold of the capture_id then - but I don't think that's a huge issue as we're just at the proof of concept stage

```ruby
paypal_payment = registration.registration_payments.first.receipt
payment_capture_id = paypal_payment.get_capture_id
```

WcaOnRails/lib/paypal_interface.rb
dunkOnIT: The issue I was having is the competition controller trying to call the auth endpoint to generate the onboarding URL (line 319), while the auth endpoint had been mocked in a different test - so I would SKIP_PAYPAL_AUTH locally only, when I was only running a small subset of tests. 

Another viable option here would be to disable the stub 
- as part of test cleanup after it was created (best solution actually) 
- around the tests which call the Paypal onboarding URL

I'm happy to update the PR to disable the stub, or include that change in a subsequent PR if we just want to focus on getting this one merged for now.

app/controllers/registrations_controller.rb
```diff
@@ -822,19 +823,69 @@ def create
...
+    response = PaypalInterface.capture_payment(@competition, order_id)
+    if response['status'] == 'COMPLETED'
+
+      # TODO: Handle the case where there are multiple captures for a payment
+      # 1) Multiple installments
+      # 2) Some failed, some succeeded
+
+      amount_details = response['purchase_units'][0]['payments']['captures'][0]['amount']
+      currency_code = amount_details['currency_code']
+      amount = PaypalRecord.amount_in_cents(amount_details["value"], currency_code)
```
dunkOnIT, line 848: Nit: Double-quotes used here unnecessarily. We may want to change overcommit so that it picks this up? 

@gregorbg again happy to make a change now or in a subsequent PR, up to you

app/controllers/registrations_controller.rb
```diff
@@ -822,19 +823,69 @@ def create
...
+      record = PaypalRecord.find_by(record_id: response["id"])
+
+      # Create a Capture object and link it to the PaypalRecord
+      # NOTE: This assumes there is only ONE capture per order - not a valid long-term assumption
+      capture_from_response = response['purchase_units'][0]['payments']['captures'][0]
+      PaypalRecord.create(
+        record_id: capture_from_response['id'],
+        status: capture_from_response['status'],
+        payload: {}, # TODO: Refactor so that we can actually capture the payload? Perhaps this needs to be called in PaypalInterface?
```
dunkOnIT, line 857: This is a tremendous hack to avoid needing to rethink where this method gets called - again with the thinking of "finish the proof of concept". 

In general, the question of where all of the payments logic happens is an important one, because I don't think PaypalInterface is the right place for it - but to include payloads we either need to do it there, or pass in the payload as an argument.

(I haven't yet referenced the Stripe implementation for how its handling this, but I recall everything happened in the same function?)

app/controllers/registrations_controller.rb
```diff
@@ -822,19 +823,69 @@ def create
...
+      # 2) Some failed, some succeeded
+
+      amount_details = response['purchase_units'][0]['payments']['captures'][0]['amount']
+      currency_code = amount_details['currency_code']
+      amount = PaypalRecord.amount_in_cents(amount_details["value"], currency_code)
+      record = PaypalRecord.find_by(record_id: response["id"])
+
+      # Create a Capture object and link it to the PaypalRecord
+      # NOTE: This assumes there is only ONE capture per order - not a valid long-term assumption
+      capture_from_response = response['purchase_units'][0]['payments']['captures'][0]
```
dunkOnIT, line 853: Nit: Rename this to "capture" (but this whole implementation will probably be changed when add in support for multiple captures)

app/controllers/registrations_controller.rb
```diff
@@ -822,19 +823,69 @@ def create
...
+        @registration.user.id,
+      )
+    end
+
+    render json: response
+  end
+
+  def refund_paypal_payment
+    registration = Registration.find(params[:id])
+    paypal_order = RegistrationPayment.find(params[:payment_id]).receipt
```
dunkOnIT, line 878: Nit: We could pass payment.receipt directly in the function that calls the refund endpoint - I tried that briefly and got errors so I went back to what I knew worked.

Not sure what best practice would be in terms of which object to pass - but the fact that we use the id we're given to immediately call another object suggests that I'm passing in the wrong object here

app/controllers/registrations_controller.rb
```diff
@@ -822,19 +823,69 @@ def create
...
+    payment_capture_id = paypal_order.capture_id
+
+    refund = PaypalInterface.issue_refund(registration, payment_capture_id)
+
+    registration.record_refund(
+      refund.amount_in_cents.to_i,
```
dunkOnIT, line 884: Shouldn't this already be stored as an integer? That's what is done in the StripeTransaction table

dunkOnIT, line 884: Also, should I make the wording consistent with stripe_transaction - `amount_lowest_denomination`? I like `amount_in_cents` because it makes it very obvious* what is actually being stored ("denomination" is not a word I have an instant undertanding of, but it is obviously more technically correct)

*to me because currencies I'm familiar with use cents

app/models/paypal_record.rb
```diff
@@ -0,0 +1,66 @@
+# frozen_string_literal: true
+
+class PaypalRecord < ApplicationRecord
+  belongs_to :parent_record, class_name: "PaypalRecord", optional: true
+  has_many :child_records, class_name: "PaypalRecord", inverse_of: :parent_record, foreign_key: :parent_record_id
+
+  RECORD_TYPES = [
```
dunkOnIT, line 7: Should these be symbols? Seems correct in ruby-land, but they seem to come back as strings when returned fro the database

app/controllers/registrations_controller.rb
gregorbg: I am genuinely surprised that it doesn't pick this up already.

If we are to make this change, it should be a separate PR. That separate PR should then also rectify all other "mistakes" throughout the entire codebase (which are probably a lot!)

app/controllers/registrations_controller.rb
gregorbg: We "kinda" already have support for multiple captures (at least structurally). If you mentally map "Paypal Captures" to "Stripe charges" and you look at how the existing "update_status_and_charges" method in Stripe works, you will see that we can simply loop over the charges and create a `RegistrationPayment` for each one.

app/controllers/registrations_controller.rb
gregorbg: No, it does not. It only suggests that the frontend is agnostic about the payment service being used. And that's a good thing.

There is nothing to worry about here, this is perfectly fine polymorphic modelling at play.

app/controllers/registrations_controller.rb
gregorbg: What is the `PaypalInterface.issue_refund` method returning? That object being returned has an `amount_in_cents` method and you need to figure out whether that's already an int or not.

gregorbg: It's not necessarily technically correct. It's only what Philippe(?) happened to choose at the time.
Actually, we're trying to avoid the `_lowest_denomination` suffix in another WCIF discussion actively because it's too long and a mouthful.

The essential point here is not to find good financial terminology, but rather to distinguish between "the amount in a form that Paypal wants" and "the amount in a form that Ruby needs to write to our database". I think the conversion methods in Stripe are called "def amount_to_stripe(foo)" and "def amount_to_ruby(foo)". That's the point here.

app/models/paypal_record.rb
gregorbg: The database has to store them as strings because SQL does not have any notion of "symbols". They are strings because in MySQL-land, there are only strings and nothing else.

I think it makes more sense to store them as symbols in Ruby, because symbols generally imply more of a "this is a constant and unchangeable value" feeling, but then of course you have to pay attention during comparison because `"foo" == :foo` evaluates to `false`. You'd have to do `"foo".to_sym == :foo` or `"foo" == :foo.to_s` to yield the expected result.

app/controllers/registrations_controller.rb
dunkOnIT: 11000 instances, I'll raise the PR :+1:

app/controllers/registrations_controller.rb
dunkOnIT: That would work, but I don't like it from a design perspective - at least as I intended it, PaypalInterface should _only_ return data from the Paypal API - to start interacting with Models would be the domain of the Controller as I understand it.

So I'd advocate for a solution where PaypalInterface also passes the payload, either as a secondary return value (my preference) or by attaching it to the response object

app/controllers/registrations_controller.rb
dunkOnIT: Yeah this is just me doing a thin implementation for the proof of concept - had planned to add that as part of an upcoming edge cases/better implementation PR once we've fleshed out the rest of the modelling

app/models/paypal_record.rb
dunkOnIT: Cool, no change needed at present then

app/controllers/registrations_controller.rb
dunkOnIT: (should have made this comment on the create_table migration)

It's returning a `PaypalRecord` object of type `:refund` - the `amount_in_cents` is a property of PaypalRecord, currently being created in the db as a string, but I think it should be int instead?

dunkOnIT: Ok thanks for the insight! I've made the following changes:
- Changed method name from `amount_in_cents` to `ruby_amount`
- Kept name of database field the same (`amount_in_cents`) - both to reduce confusion between the method and the property, and to explicitly state what is being stored in the db

app/controllers/registrations_controller.rb
dunkOnIT: (Feel free to reopen conversation if you want me to include that functionality in this PR)

app/controllers/registrations_controller.rb
dunkOnIT: (Worth noting that currently at least PaypalInterface is creating our `charge` and `order` records - so maybe it makes sense to keep them all there for now, and then refactor in the cleanup?)

WcaOnRails/lib/paypal_interface.rb
dunkOnIT: (reasoning here is important so leaving you to Resolve the conversation)

app/controllers/registrations_controller.rb
gregorbg: Yes, in that case you should adjust your migration to be an integer. We definintely always want to store amounts as "what we think we're sending / what we want to send" in integer form.
If Paypal is dumb enough to actually require strings as amount (which pretty much any finance API guy in the whole wide world will tell you is a terrible idea, but eh... they're PayPal...) then we should include that string representation in the `payload` if at all necessary.

gregorbg (approved):

PR #8929, opened 2024-02-19
Gave report read access to WFC

danieljames-dj:


danieljames-dj:
Didn't receive any approval from the Board, but WFC internally decided it will be good if they get access because there are use-case where WFC need to refer to Delegate report while reviewing the funding applications (and many more use cases probably).

Also, WRT is planning to work on some permissions policy to make things clear like who all will have access to what all things, once that is up it will be more transparent.

But waiting for the policy to get ready will be too much time for WFC. Hence, can you please let me know if this needs to be approved from Board? If yes, I'll go ahead with the process to get the permission.

gregorbg (approved):

gregorbg:
Fine by me. Not like we're hiding state secrets :smile:

PR #8933, opened 2024-02-21
Remove Locales group

danieljames-dj:


WcaOnRails/config/application.rb
```diff
@@ -65,7 +64,7 @@ class Application < Rails::Application
     end
 
     # Setup available locales
-    I18n.available_locales = Locales::AVAILABLE.keys
```
danieljames-dj, line 68: @gregorbg Looks like we don't get the access to the database here. Do you have any suggestions on how to fix this?

danieljames-dj: Reason why I'm unhappy with `Locales::AVAILABLE` is just because it will be duplicate data. The same data will be there in UserGroups table as well.

In that case, can we generate `locales.rb` manually during build or through some cronjob to make it updated to the database? I really don't know how feasible it is, just exploring the ideas.

gregorbg: Hm... I haven't put any deep thought into this, but in my opinion it doesn't hurt to have a hard-coded set of locales and have some database entries which follow these hard-coded locales.

It has the benefit of easy validation -- we can check whether there are database entries which are not in the list. We can also easily add new entries (support a new locale on the website) without having to add an orphaned translator-less group to the DB.

danieljames-dj: Okay restored locales.rb and changed application.rb. In this PR, the migration is needed to add proper name to the group. When you get time, could you please check if any other changes also has to be reverted?

gregorbg:
I am still struggling to understand the purpose of this PR.
What's wrong with keeping the `Locales::AVAILABLE` everywhere as-is?

danieljames-dj:
> I am still struggling to understand the purpose of this PR. What's wrong with keeping the `Locales::AVAILABLE` everywhere as-is?

The purpose of this PR was to change the name of locales to the locale's name. But I was thinking to replace `Locales::AVAILABLE` just because I want to reduce the dependency of `Locales::AVAILABLE`. But ya we will always need `Locales::AVAILABLE` so having this is also fine. I've removed the other changes. Now there is only the migration change.

PR #8935, opened 2024-02-23
Create separate page for Board & Officers

danieljames-dj:
Screenshot:
Screenshot 2024-02-23 at 08 47 25


WcaOnRails/app/webpacker/components/OfficersAndBoard/index.jsx
```diff
@@ -0,0 +1,107 @@
...
+        accumulator[userIndex[userId]].status.push(currentValue.metadata.status);
+      } else {
+        userIndex[userId] = accumulator.length;
+        accumulator.push({
+          userData: currentValue.user,
+          status: [currentValue.metadata.status],
+        });
+      }
+      return accumulator;
+    }, []);
```
kr-matthews, line 50: I'm not sure if it's worth rewriting, but what if the accumulator were an object with a key per userId, rather then an array with a separate user-index lookup? Then apply `Object.values` to the result.

WcaOnRails/app/webpacker/components/OfficersAndBoard/index.jsx
```diff
@@ -0,0 +1,107 @@
...
+
+  if (boardLoading || officersLoading) return <Loading />;
+  if (boardError || officersError) return <Errored />;
+
+  return (
+    <Container fluid>
+      <Header as="h2">{I18n.t('page.officers_and_board.title')}</Header>
+      <Header as="h3">{I18n.t('user_groups.group_types.officers')}</Header>
+      <p>{I18n.t('page.officers_and_board.officers_description')}</p>
+      <>
```
kr-matthews, line 61: Is this fragment necessary? Same with the one around the board.

WcaOnRails/app/webpacker/components/OfficersAndBoard/index.jsx
```diff
@@ -0,0 +1,107 @@
...
+              icon
+              style={{
+                margin: '8px',
+              }}
+              onMouseEnter={() => setHoveringEmail(true)}
+              onMouseLeave={() => setHoveringEmail(false)}
+            >
+              <Icon name={hoveringEmail ? 'copy' : 'mail'} />
+              {hoveringEmail && (
+              <span>{boardEmail}</span>
```
kr-matthews, line 89:
```suggestion
                <span>{boardEmail}</span>
```

danieljames-dj: Done

WcaOnRails/app/webpacker/components/OfficersAndBoard/index.jsx
danieljames-dj: Looks like it doesn't. Removed it. Thank you.

WcaOnRails/app/webpacker/components/OfficersAndBoard/index.jsx
danieljames-dj: I've thought of that, but I wanted to maintain the sort order, which is why I did like this.

danieljames-dj: I've given a thought again on whether I can group and then sort on client-side. But sorting is not possible in client-side because we need to bring status-sorting logic to client side and it will cause code duplication.

kr-matthews: It seems like `_.groupBy` preserves the order, so could we use that here?

> [The order of grouped values is determined by the order they occur in collection.](https://lodash.com/docs/#groupBy)

danieljames-dj: Ya thanks for that, done.

WcaOnRails/app/webpacker/components/OfficersAndBoard/index.jsx
```diff
@@ -17,6 +18,10 @@ import UserBadge from '../UserBadge';
 // i18n-tasks-use t('user_roles.status.officers.vice_chair')
 // i18n-tasks-use t('user_roles.status.officers.treasurer')
 
+function arrayAfterRemovingDuplicates(array) {
+  return array?.filter((item, index) => array.indexOf(item) === index);
+}
```
gregorbg, line 23: Isn't this exactly `_.uniq(array)`?

gregorbg: https://lodash.com/docs/#uniq

gregorbg (changes requested):

WcaOnRails/app/webpacker/components/OfficersAndBoard/index.jsx
danieljames-dj: Oh I always forget about lodash. Done.

app/webpacker/components/OfficersAndBoard/index.jsx
```diff
@@ -0,0 +1,91 @@
...
+              onMouseEnter={() => setHoveringEmail(true)}
+              onMouseLeave={() => setHoveringEmail(false)}
+            >
+              <Icon name={hoveringEmail ? 'copy' : 'mail'} />
+              {hoveringEmail && (
+                <span>{boardEmail}</span>
+              )}
+            </Button>
+          )}
+        />
```
gregorbg, line 79: You seem to have mis-understood the use of a Popup. The children of the popup tag `<Popup>{children}</Popup>` are essentially the content of the "floating" container, that is being managed entirely by SemUI for you. Manually tracking `setHovering` states is not necessary and redundant.

gregorbg (changes requested):

gregorbg (approved):

PR #8938, opened 2024-02-24
Redirect to new registration import path if v2 is activated

FinnIckler:


WcaOnRails/app/controllers/registrations_controller.rb
```diff
@@ -116,6 +116,9 @@ def export
 
   def import
     @competition = competition_from_params
+    if @competition.uses_new_registration_service?
+      redirect_to Microservices.Registrations.registration_import_path(@competition.id)
```
gregorbg, line 120: Nit: Shouldn't this be `Microservices::Registrations`?

gregorbg (approved):

PR #8941, opened 2024-02-25
Fix Server Status page

FinnIckler:
Actually fixes the JSON Path this time (I tested with one bucket for documents/regulations which is why this happened).

For stripe: I perused the StripeTransaction Table and the Webhook table but I couldn't really find anything which we could use. Any idea?

### DIFF
--- a/WcaOnRails/app/controllers/server_status_controller.rb
+++ b/WcaOnRails/app/controllers/server_status_controller.rb
@@ -15,7 +15,6 @@ class ServerStatusController < ApplicationController
     [
       JobsCheck.new,
       RegulationsCheck.new,
-      StripeChargesCheck.new,
       MysqlSettingsCheck.new,
     ]
   end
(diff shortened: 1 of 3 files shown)


gregorbg:
> Any idea?

Actually, there's nothing here that we could really fix. The old status `unknown` comes from a time when we used a very old (now even deprecated) API to communicate with Stripe, which was synchronous and could have left a PaymentIntent in an orphaned state when the user aborted 3DSecure.
Nowadays we just create a PI and pass it on to Stripe directly, from which point on we don't have to worry about it anymore. So the check in its current state is definitely not needed anymore.

The only potentially dangerous scenario that I can come up with is when our completion callbacks, i.e. the return_url don't respond because a payment was initiated, the user was handed off to Stripe (which in turns redirects them to their bank to do authentication stuff) and while they're busy interacting with their bank's website our server crashes.

This issue would be noticeable by a high number of `requires_payment_method` status flags in the `StripeTransaction` table. However it is not as easy as checking `where count > 0` because even with healthy servers there will always be a non-zero amount of transactions where the user was handed off to Stripe, our PI is still stuck at `requires_payment_method` because the user is literally in the process of paying right now, and we're waiting for them to be redirected to us before writing another status flag to the DB.

So we need to determine some threshold when there are "too many" pending Stripe transactions. One way could be to look at registrations, count how many are pending (i.e. filled out the form but haven't been accepted by the organization team) and compare that with the number of pending Stripe transactions. But this only makes sense for registrations of competitions which even have Stripe connected in the first place (which complicates the query) and it still needs to set an arbitrary threshold of "how many percent of open registrations are allowed to have a pending transaction before we start worrying"?

TL;DR Just delete the existing Stripe check.

gregorbg (approved):

PR #8944, opened 2024-02-26
Add webmock for request stubbing during tests

dunkOnIT:
This will be used in #8927, but introducing it as a separate PR given that its out of scope for that PR. 
- [x] Explciitly enable net connect
- [x] Require webmock in spec_helper


gregorbg (changes requested):
Please add an evnrionment config (in the `.env` files) that allows you to enable/disable webmock by setting the corresponding env var to true/false (and have the spec config files pick up the env variable accordingly)

I also think that the default should be to disable webmock. and let developers choose locally whether they would like to enable it for local testing. The reasonable default should be to make live calls wherever and whenever possible

dunkOnIT:
Added env variables for the following:
* enabling/disabling HTTP connections
* enabling/disabling webmock stubs (not used in this PR, but will be incorporated into the Paypal PR)

gregorbg:
You famously missed an `end` in your latest ommit. The tests cannot even start to run because of syntax errors.

gregorbg (approved):

env_config.rb
```diff
@@ -44,10 +44,15 @@
 
     # Local-specific stuff
     optional :ENABLE_BULLET, :bool, false
-    optional :SKIP_PRETEST_SETUP, :bool, false
     optional :MAILCATCHER_SMTP_HOST, :string, ''
   end
 
+  if Rails.env.test?
+    optional :DISABLE_WEBMOCK, :bool, false
+    optional :ENABLE_NET_CONNECT_IN_TESTS, :bool, true
```
gregorbg, line 52: Here you declared it as `ENABLE` but below in the actual config you're using `DISABLE`

spec/rails_helper.rb
```diff
@@ -87,6 +87,12 @@
   config.include ApplicationHelper
 
   config.include ActiveJob::TestHelper
+
+  if EnvConfig.DISABLE_WEBMOCK?
+    WebMock.disable!
+  else
+    WebMock.allow_net_connect! unless EnvConfig.DISABLE_NET_CONNECT_IN_TESTS?
```
gregorbg, line 94: ENABLE vs DISABLE, see above.

gregorbg (changes requested):

gregorbg (approved):

PR #8975, opened 2024-03-01
Modified edit-person script to accept WCA ID via query param and search option in country picker

danieljames-dj:


WcaOnRails/app/webpacker/lib/hooks/useQueryParams.js
```diff
@@ -0,0 +1,30 @@
+import { useEffect, useState } from 'react';
+
+function getQueryParamsFromBrowserUrl() {
```
gregorbg, line 3: Can you use `Object.fromEntries` here?

WcaOnRails/app/webpacker/lib/hooks/useQueryParams.js
```diff
@@ -0,0 +1,30 @@
...
+    return accumulator;
+  }, {});
+}
+
+export default function useQueryParams() {
+  const [queryParams, setQueryParams] = useState(getQueryParamsFromBrowserUrl());
+
+  useEffect(() => {
+    const searchParams = new URLSearchParams(queryParams).toString();
+    if (window.location.search !== searchParams) {
```
gregorbg, line 17: `!==` checks for referential equality. Since you are creating a `new URLSearchParams` object directly above, I feel like this comparison always yields `true` because they can by defintion never be the same.

WcaOnRails/app/controllers/api/v0/persons_controller.rb
gregorbg, line 1: This approach looks like code duplication. I feel like we've dealt with this private attribute notion before, can you find a way to re-use the existing code?

gregorbg (changes requested):

danieljames-dj:
Added search option in country picker as selecting country is tough now

app/controllers/api/v0/persons_controller.rb
```diff
@@ -29,9 +33,9 @@ def competitions
     render json: person.competitions
   end
 
-  private def person_to_json(person)
+  private def person_to_json(person, private_attributes = [])
     {
-      person: person.serializable_hash(only: [:wca_id, :name, :url, :gender, :country_iso2, :delegate_status, :teams, :avatar]),
+      person: person.serializable_hash(only: [:wca_id, :name, :url, :gender, :country_iso2, :delegate_status, :teams, :avatar, private_attributes].flatten),
```
gregorbg, line 38: I think this should be `person.serializable_hash(only: [...], private_attributes: [...])` so that the existing code in `person.rb` can pick it up.

app/webpacker/lib/hooks/useQueryParams.js
```diff
@@ -0,0 +1,28 @@
...
+      window.location.search = searchParams;
+    }
+  }, [queryParams]);
+
+  function updateQueryParam(key, value) {
+    setQueryParams({
+      ...queryParams,
+      [key]: value,
+    });
+  }
```
gregorbg, line 25: Should this be wrapped in `useCallback`?

gregorbg (changes requested):

gregorbg (approved):

PR #8977, opened 2024-03-01
Competition Form fixes

gregorbg:
Some quickfixes for reports that came in via email. Specifically:

- Fix error where the number of guests per competitor restriction was not shown
- Don't allow adding a competition to a Series upon creation. This causes trouble when trying to load the items in the SearchBar, because it looks up values based on ID but the competition being created does not have an ID yet.
- Correctly report the error about the missing name field when adding to a Series


simonkellly:
LGTM code wise, with the caveat that I can't currently test it

PR #9000, opened 2024-03-01
Fix Webpack DevServer https option deprecation

gregorbg:
See https://webpack.js.org/configuration/dev-server/#devserverserver


gregorbg:
Merging as a prerequisite to the Webpack Dev Server update

PR #9010, opened 2024-03-03
Cleanup of WcaSearch model

danieljames-dj:
Currently for 'Delegate Search', it's considering only users table and hence regional delegates won't appear. Once this PR is merged, then the Delegate search can be done with two models: user & role. There are many other cases like SearchWidget where we need multi-model searching, so I guess this will be useful for that as well.


WcaOnRails/app/webpacker/components/CompetitionForm/Inputs/FormSearch.js
```diff
@@ -84,6 +85,7 @@ export function CompetitionSearch({
     <WcaSearch
       value={compOptions}
       onChange={onChangeIdOnly}
+      models={[SEARCH_MODELS.competition]} // TODO: Need to check if this is needed.
```
gregorbg, line 88: What do you mean by "if this is needed"?

WcaOnRails/app/controllers/api/v0/api_controller.rb
```diff
@@ -149,10 +159,16 @@ def incidents_search
...
 
   def omni_search
-    # We intentionally exclude Post, as our autocomplete ui isn't very useful with
-    # them yet.
-    params[:persons_table] = true
-    search(Competition, User, Regulation, Incident)
+    models = params[:models]
+    if models.present?
+      model_list = models.split(",")
+      search(*model_list.map { |model| Api::V0::ApiController.search_models[model] })
```
gregorbg, line 165: Too much happening on one line

WcaOnRails/app/controllers/api/v0/api_controller.rb
```diff
@@ -149,10 +159,16 @@ def incidents_search
...
+    else
```
gregorbg, line 166:
```suggestion
    model_names = params[:models]
    if model_names.present?
      model_list = model_names.split(",")
      models = model_list.map { |model_name| Api::V0::ApiController.search_models[model_name] }
      
      search(*models)
    else
```

WcaOnRails/app/controllers/api/v0/api_controller.rb
gregorbg: See my other suggested code change

WcaOnRails/app/webpacker/components/SearchWidget/index.js
```diff
@@ -1,22 +1,27 @@
...
   // purely a dummy for now...
   const [selectedValue, setSelectedValue] = useInputState([]);
 
   return (
-    <MultiSearchInput
-      selectedValue={selectedValue}
-      setSelectedValue={setSelectedValue}
+    <WcaSearch
+      value={selectedValue}
+      onChange={(_, { value }) => setSelectedValue(value)}
```
gregorbg, line 14: `useInputState` is our custom-designed hook whose only purpose is to prevent what you're doing.
`onChange={setSelectedValue}` will work fine.

gregorbg (changes requested):
I am not exactly happy about allowing this much flexibility in the frontend.

In the future this will probably make parsing responses very awkward because you have exactly zero guarantees about what format the returned JSON will have...

I understand the use-case for a combined "users+roles" search field, but have you suggested just introducing an extra "Delegates" search type which has its own backend endpoint that handles combining the two different models "behind the scenes"?

WcaOnRails/app/webpacker/components/CompetitionForm/Inputs/FormSearch.js
danieljames-dj: Actually wanted to check with you whether this is needed. Because I was trying to replace `model` with `models` and found this `WcaSearch which doesn't even have `model`. I don't think this feature currently works because there is no model. But I didn't test this as well, because I'm not aware of the exact process to create series competition (or in other words I didn't try to do it).

I assume this is needed, and removed the TODO comment. If you think this is fine, you can resolve it, else let me know what change needs to be made.

WcaOnRails/app/controllers/api/v0/api_controller.rb
danieljames-dj: Done.

WcaOnRails/app/controllers/api/v0/api_controller.rb
danieljames-dj: Done.

WcaOnRails/app/webpacker/components/SearchWidget/index.js
danieljames-dj: Oh didn't knew it 😅 Thanks, applied.

danieljames-dj (commented):
The main use-case for the combined models is the SearchWidget, where we need to search among all the models.

We can bring extra 'Delegates' search as well. But I don't know how easy it will be. I felt having multi-model support because there is currently at least one case where this can be used.

Once roles migration is completed, bringing 'Roles' search should be very easy to take care of Delegates search I guess.

gregorbg:
> The main use-case for the combined models is the SearchWidget, where we need to search among all the models.

That's why we have an omni-search backend, where we can take care of mixing stuff together and packaging it into a JSON that the frontend understands. We also get more fine-grained control over caching.

> We can bring extra 'Delegates' search as well. But I don't know how easy it will be. I felt having multi-model support because there is currently at least one case where this can be used.

"at least" one? For me this sounds like "only one", which is not a good argument to swap out the implementation in all other usages.

> Once roles migration is completed, bringing 'Roles' search should be very easy to take care of Delegates search I guess.

If that is the case, then you just delivered your own argument for reverting this change and having it be "one search bar, one model" (as it was before already). Temporarily adding multi-model support is not a good idea when the future perspective is that we can boil it back down to one model.

Please consider having a frontend "type=delegate" search that maps to a backend endpoint which internally (i.e. far away from the frontend) merges `Person` and `Roles` together as needed.

danieljames-dj:
I've changed to single model for WcaSearch. Now this PR can be considered as cleanup of WcaSearch model.

app/controllers/api/v0/api_controller.rb
```diff
@@ -149,10 +159,17 @@ def incidents_search
...
-    # We intentionally exclude Post, as our autocomplete ui isn't very useful with
-    # them yet.
-    params[:persons_table] = true
-    search(Competition, User, Regulation, Incident)
+    model_names = params[:models]
+    if model_names.present?
+      model_list = model_names.split(",")
+      models = model_list.map { |model_name| Api::V0::ApiController.search_models[model_name] }
+      search(*models)
+    else
```
gregorbg, line 167: I think this PR has been re-purposed and multiple models are not supported anymore.
So probably this controller handling can go away.

app/webpacker/components/SearchWidget/WcaSearch.jsx
```diff
@@ -16,21 +12,7 @@ export default function WcaSearch({
...
-      return `${personSearchApiUrl(query)}&${new URLSearchParams(params).toString()}`;
-    }
-
-    if (model === 'competition') {
-      return `${competitionSearchApiUrl(query)}&${new URLSearchParams(params).toString()}`;
-    }
-
-    throw new Error(`Invalid search type in WcaSearch component: ${model}`);
-  }, [params, model]);
+  const urlFn = useCallback((query) => apiV0Urls.search(query, [model], params), [model, params]);
```
gregorbg, line 15: Hm... I get that your new version looks cleaner, but it entirely discards the specialized search APIs.
Right now, these APIs do nothing more than forward to the general search method, but in the future we might leverage the advantage of having sspecialized endpoints for specialized caching or whatever.

Please consider adding back the previous version or make an argument in favor of the new version.

app/webpacker/components/SearchWidget/index.js
```diff
@@ -1,12 +1,19 @@
...
 
 function SearchWidget() {
   // purely a dummy for now...
   const [selectedValue, setSelectedValue] = useInputState([]);
+  const urlFn = useCallback((query) => apiV0Urls.search(query, [
+    SEARCH_MODELS.competition,
+    SEARCH_MODELS.person,
+    SEARCH_MODELS.regulation,
+    SEARCH_MODELS.incident,
+  ]), []);
```
gregorbg, line 16: Same argument about specialized endpoints as above.

app/webpacker/lib/wca-data.js.erb
```diff
@@ -188,3 +188,5 @@ export const delegateRegionsStatus = <%= RolesMetadataDelegateRegions.statuses.t
 // ----- PANEL LIST -----
 
 export const PANEL_LIST = <%= PanelController.panel_list.to_json.html_safe %>;
+
+export const SEARCH_MODELS = <%= Hash[Api::V0::ApiController.search_models.map { |key, value| [key.to_s, key.to_s] }].to_json.html_safe %>;
```
gregorbg, line 192: Why are you matching the key to the key?

gregorbg (changes requested):

app/controllers/api/v0/api_controller.rb
danieljames-dj: Done

app/webpacker/components/SearchWidget/WcaSearch.jsx
danieljames-dj: Okay added back the previous version.

app/webpacker/components/SearchWidget/index.js
danieljames-dj: Okay changed.

app/webpacker/lib/wca-data.js.erb
danieljames-dj: I wanted to use it as an object, but anyway this is removed now :-)

gregorbg (approved):

Issue #8948, opened 2024-02-28
Add registration date and time to csv export of registrations


AnnikaStein:
**Is your feature request related to a problem? Please describe.**
Usually, we need multiple people and hours to go through the list of people who just registered at the WCA page of a competition, to then copy&paste their names, and get their registration time up to the level of seconds. This is then put into another table that can be accessed by the organization team, where payment dates / times (e.g. from PayPal) are used as another column. Eventually, complete registrations are then sorted by whichever of the two came last, and people slowly get accepted from the waiting list. This takes much longer than it needs to take, triggering additional emails of people who don't know if or when the competitor list gets filled. If one could - with few clicks - get all competitors and their registration time as a column from the csv export, it would be a huge step forward - the faster we update the competitor list, the more transparent the process looks from the outside. Technically: no more typing by hand and using mouse-hovering to get the tooltip with precision up to seconds. Organization team members without access to regular workstations (tablets for example) are especially struggling to copy the content of the tooltip over into another tab and we sometimes even need someone to read information out loud to a second person, who is just typing for two hours straight.

**Describe the solution you'd like**
Add one column to the registrations csv export, i.e. `registration.created_at`. This can be formatted the way it is already in the tooltip, YYYY-MM-DD HH:MM:SS UTC, so no need to reinvent anything, we don't need the `registration_date_and_tooltip` function to split the date from the rest, that's something easily implemented in any tabular calculation software (custom date time format as shown above), where one anyway needs to adjust the UTC column to the time zone in which we record payments, or vice-versa. This part is something to be done by each team individually (just tested it with mock columns).

**Describe alternatives you've considered**
a) In the long run, a system that does the matching for us PayPal users (I guess its integration is already WIP).
b) Independent of the csv export function, come up with a hacky way to read whatever gets rendered by `title="<%= date_tooltip %>"` in the console for every row in the table (want to avoid, although parsing this is certainly possible).

**Additional context**
Would probably just edit https://github.com/thewca/worldcubeassociation.org/blob/master/WcaOnRails/app/views/registrations/export.csv.erb
and do something like
```
<% require 'csv' %>
<% headers = ["Status", "Name", "Country", "WCA ID", "Birth Date", "Gender" ] + @competition.events.map(&:id) + [ "Email", "Guests", "IP", "Registration Date Time (UTC)" ] %>
<%= CSV.generate_line(headers).html_safe -%>
<% @registrations.each do |registration| %>
<%= CSV.generate_line([
[...]

PR #9011, opened 2024-03-03
Add created_at to registration export csv

AnnikaStein:
See description in #8948 

This adds one column to the csv export of registrations to improve turnaround for organizers, who would otherwise refer to the tooltip.


AnnikaStein:
Just saw that there is a test case for this csv file. Will adapt the dummy registration over there then.

AnnikaStein:
The only remaining error is coming from JobsCheck which is long before the tests for the csv export are shown. I don't know if I just hit an unlucky moment (pending cronjobs not long ago), or if there is something from my side to change. Otherwise I think the test cases now successfully pick up the new column, with a dummy date/time injected (compared to this https://github.com/thewca/worldcubeassociation.org/actions/runs/8129384282/job/22216428869, no more such errors coming from registrations/export.csv.erb).
Looking forward to your feedback.

gregorbg (approved):

gregorbg:
Thanks for your submission! The test fail is an unrelated choke that we haven't been able to pinpoint for over half a year, so don't worry!

PR #9024, opened 2024-03-04
Replaced delegates API with method from UserGroup, and filtered out nil values from delegate_region_groups_senior_delegates method

danieljames-dj:


app/controllers/api/v0/api_controller.rb
```diff
@@ -156,7 +156,8 @@ def omni_search
   end
 
   def delegates
-    paginate json: User.staff_delegates
+    delegate_users = UserGroup.delegate_region_groups.map(&:active_users).flatten
```
gregorbg, line 159: .flat_map(&:active_users)

gregorbg (changes requested):

app/controllers/api/v0/api_controller.rb
danieljames-dj: Done

danieljames-dj:
The initial purpose of this PR was to remove User.staff_delegates. But the plan was changed later. Currently, this PR is doing two things:
1. Replace delegates API with method from UserGroup
2. Filtered out nil values from delegate_region_groups_senior_delegates method

gregorbg (approved):

PR #8976, opened 2024-03-01
GroupManager - feature to add new member

danieljames-dj:

PR #9029, opened 2024-03-05
Cleanup of user roles create API

danieljames-dj:
This also fixes the issue where a variable `group` is used before initialization. I accidentally removed it in https://github.com/thewca/worldcubeassociation.org/pull/8976.

### DIFF
--- a/app/controllers/api/v0/user_roles_controller.rb
+++ b/app/controllers/api/v0/user_roles_controller.rb
@@ -269,74 +269,77 @@ class Api::V0::UserRolesController < Api::V0::ApiController
     [RolesMetadataDelegateRegions.statuses[:regional_delegate], RolesMetadataDelegateRegions.statuses[:senior_delegate]].include?(status)
   end
 
-  def create
-    user_id = params.require(:userId)
-    group_id = params[:groupId] || UserGroup.find_by(group_type: params.require(:groupType)).id
+  private def team_role?(group_id)
+    group_id.is_a?(String) && group_id.include?("_") # Temporary hack to support some old system roles, will be removed once all roles are migrated to the new system.
+  end
 
-    if group_id.is_a?(String) && group_id.include?("_") # Temporary hack to support some old system roles, will be removed once all roles are
-      # migrated to the new system.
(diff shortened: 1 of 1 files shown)


app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -269,74 +269,77 @@ def index_for_group_type
...
-      RoleChangeMailer.notify_role_start(role, current_user).deliver_later
-      render json: {
+      return render json: {
         success: true,
       }
     end
+
+    if group.group_type == UserGroup.group_types[:delegate_regions]
+      metadata = RolesMetadataDelegateRegions.create!(status: status, location: location)
+    end
```
gregorbg, line 331: How does the variable `metadata` get declared for other group types?

app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -269,74 +269,77 @@ def index_for_group_type
...
+    ]
+    group = UserGroup.find(group_id)
+    status = params.require(:status) if UserGroup.group_types_containing_status_metadata.include?(group.group_type)
+    location = params[:location] if group.group_type == UserGroup.group_types[:delegate_regions]
+
+    return render status: :unprocessable_entity, json: { error: "Invalid group type" } unless create_supported_groups.include?(group.group_type)
+    return head :unauthorized unless current_user.has_permission?(:can_edit_groups, group_id)
+
+    if status.present? && group.unique_status?(status)
+      end_role_for_user_in_group_with_status(group, status)
```
gregorbg, line 316: Should the controller return here? Otherwise the flow of the method will continue to create a new user role below

gregorbg (changes requested):

app/controllers/api/v0/user_roles_controller.rb
gregorbg: This is not about the high-level logic which group needs metadata. This is about the low-level programming question where the variable is being defined.

In other words, if other groups don't need metadata it should be set to null somewhere.

gregorbg (approved):

PR #9031, opened 2024-03-05
Region page in Senior Delegate Panel

danieljames-dj:
Currently, one user can hold only one Delegate role. But once the role migration is completed, one user can be a Delegate of more than one subregion.

But the current way of 'creating new delegate' through the 'role form' won't allow multiple Delegate role. To achieve it in the same UI will make it more complicated.

To solve this, we now have a new page in Senior Delegate Panel named as 'Regions' which will list all the Delegates of their regions and subregions. From this UI, the Senior Delegate can create a new Delegate under respective subregion.

Attaching the screenshot:
Screenshot 2024-03-05 at 22 00 30

### DIFF
--- a/app/controllers/api/v0/user_groups_controller.rb
+++ b/app/controllers/api/v0/user_groups_controller.rb
@@ -21,6 +21,8 @@ class Api::V0::UserGroupsController < Api::V0::ApiController
   # Filters the list of groups based on given parameters.
   private def filter_groups_for_parameters(groups: [], is_active: nil, parent_group_id: nil)
     groups.reject do |group|
+      # Here, instead of foo.present? we are using !foo.nil? because foo.present? returns false if
+      # foo is a boolean false but we need to actually check if the boolean is present or not.
       (
         (!is_active.nil? && is_active != group.is_active) ||
         (!parent_group_id.nil? && group.parent_group_id != parent_group_id)
--- a/app/controllers/api/v0/user_groups_controller.rb
+++ b/app/controllers/api/v0/user_groups_controller.rb
@@ -23,7 +23,7 @@ class Api::V0::UserGroupsController < Api::V0::ApiController
     groups.reject do |group|
       (
         (!is_active.nil? && is_active != group.is_active) ||
-        (parent_group_id.present? && group.parent_group_id != parent_group_id)
+        (!parent_group_id.nil? && group.parent_group_id != parent_group_id)
       )
     end
   end
--- a/app/controllers/api/v0/user_groups_controller.rb
+++ b/app/controllers/api/v0/user_groups_controller.rb
@@ -19,10 +19,11 @@ class Api::V0::UserGroupsController < Api::V0::ApiController
   end
 
   # Filters the list of groups based on given parameters.
-  private def filter_groups_for_parameters(groups: [], is_active: nil)
+  private def filter_groups_for_parameters(groups: [], is_active: nil, parent_group_id: nil)
     groups.reject do |group|
       (
-        !is_active.nil? && is_active != group.is_active
+        (!is_active.nil? && is_active != group.is_active) ||
+        (parent_group_id.present? && group.parent_group_id != parent_group_id)
       )
     end
   end
@@ -53,6 +54,7 @@ class Api::V0::UserGroupsController < Api::V0::ApiController
     groups = filter_groups_for_parameters(
       groups: groups,
       is_active: params.key?(:isActive) ? ActiveRecord::Type::Boolean.new.cast(params.require(:isActive)) : nil,
+      parent_group_id: params.key?(:parentGroupId) ? params.require(:parentGroupId).to_i : nil,
     )
 
     # Sorts the list of groups by name.
--- a/app/controllers/api/v0/user_roles_controller.rb
+++ b/app/controllers/api/v0/user_roles_controller.rb
@@ -65,7 +65,7 @@ class Api::V0::UserRolesController < Api::V0::ApiController
   end
 
   # Filters the list of roles based on given parameters.
-  private def filter_roles_for_parameters(roles: [], status: nil, is_active: nil, is_group_hidden: nil)
+  private def filter_roles_for_parameters(roles: [], status: nil, is_active: nil, is_group_hidden: nil, group_type: nil, is_lead: nil)
     roles.reject do |role|
       is_actual_role = role.is_a?(UserRole) # See previous is_actual_role comment.
       # In future, the following lines will be replaced by the following:
@@ -79,7 +79,9 @@ class Api::V0::UserRolesController < Api::V0::ApiController
       (
         (status.present? && status != (is_actual_role ? role.metadata.status : role[:metadata][:status])) ||
         (!is_active.nil? && is_active != (is_actual_role ? role.is_active? : role[:is_active])) ||
-        (!is_group_hidden.nil? && is_group_hidden != (is_actual_role ? role.group.is_hidden : role[:group][:is_hidden]))
+        (!is_group_hidden.nil? && is_group_hidden != (is_actual_role ? role.group.is_hidden : role[:group][:is_hidden])) ||
+        (group_type.present? && group_type != UserRole.group_type(role)) ||
+        (!is_lead.nil? && is_lead != UserRole.is_lead?(role))
       )
     end
   end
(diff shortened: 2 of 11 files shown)


app/controllers/api/v0/user_groups_controller.rb
```diff
@@ -19,10 +19,11 @@ class Api::V0::UserGroupsController < Api::V0::ApiController
   end
 
   # Filters the list of groups based on given parameters.
-  private def filter_groups_for_parameters(groups: [], is_active: nil)
+  private def filter_groups_for_parameters(groups: [], is_active: nil, parent_group_id: nil)
     groups.reject do |group|
       (
-        !is_active.nil? && is_active != group.is_active
+        (!is_active.nil? && is_active != group.is_active) ||
+        (parent_group_id.present? && group.parent_group_id != parent_group_id)
```
gregorbg, line 26: Inconsistency between using `!foo.nil?` and `foo.present?` (which I believe both end up doing the same thing)

app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -79,7 +79,9 @@ def self.status_sort_rank(role)
       (
         (status.present? && status != (is_actual_role ? role.metadata.status : role[:metadata][:status])) ||
         (!is_active.nil? && is_active != (is_actual_role ? role.is_active? : role[:is_active])) ||
-        (!is_group_hidden.nil? && is_group_hidden != (is_actual_role ? role.group.is_hidden : role[:group][:is_hidden]))
+        (!is_group_hidden.nil? && is_group_hidden != (is_actual_role ? role.group.is_hidden : role[:group][:is_hidden])) ||
+        (group_type.present? && group_type != UserRole.group_type(role)) ||
+        (!is_lead.nil? && is_lead != UserRole.is_lead?(role))
```
gregorbg, line 84: `!nil?` vs `present?`, see above.

app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -329,6 +334,7 @@ def create
     if group.group_type == UserGroup.group_types[:delegate_regions] && !delegate_status_migrated?(status)
       # Creates deprecated role.
       user = User.find(user_id)
+      return render status: :unprocessable_entity, json: { error: "Already a delegate" } if user.any_kind_of_delegate?
```
gregorbg, line 337: Nitpick: Capital D in `"Already a Delegate"`

app/webpacker/components/Panel/SeniorDelegate/Regions/Subregion.jsx
```diff
@@ -0,0 +1,133 @@
...
+            <Form.Dropdown
+              label="Delegate Status"
+              fluid
+              selection
+              name="status"
+              value={formValues.status}
+              options={delegateStatusOptions.map((option) => ({
+                text: I18n.t(`enums.user.role_status.delegate_regions.${option}`),
+                value: option,
+              }))}
```
gregorbg, line 117: Maybe move the creation of these options outside of the component?

app/webpacker/components/Panel/SeniorDelegate/Regions/index.jsx
```diff
@@ -0,0 +1,54 @@
...
+  return (
+    <>
+      <Header>Regions</Header>
+      <div>
+        <Dropdown
+          options={seniorDelegateRoles.map((role, index) => ({
+            key: role.id,
+            text: role.group.name,
+            value: index,
+          }))}
```
gregorbg, line 46: Move this up into a `useMemo`?

gregorbg (changes requested):

app/webpacker/components/Panel/SeniorDelegate/Regions/Subregion.jsx
```diff
@@ -0,0 +1,134 @@
...
+    data: delegates, loading, error: delegatesFetchError, sync,
+  } = useLoadedData(apiV0Urls.userRoles.listOfGroup(
+    groupId,
+    'location,name',
+    {
+      isActive: true,
+      isLead: false,
+    },
+  ));
+  const [openModalType, setOpenModalType] = useState(null);
```
kr-matthews, line 34: Are there going to be other types of modals added to this component in the future?

app/webpacker/components/Panel/SeniorDelegate/Regions/index.jsx
```diff
@@ -0,0 +1,56 @@
...
+    if (seniorDelegateRoles?.length > 0) setSelectedGroupIndex(0);
+  }, [seniorDelegateRoles, setSelectedGroupIndex]);
+
+  const groupOptions = useMemo(() => seniorDelegateRoles.map((role, index) => ({
+    key: role.id,
+    text: role.group.name,
+    value: index,
+  })), [seniorDelegateRoles]);
+
+  if (!loading && seniorDelegateRoles.length === 0) return <p>You cannot manage any regions..</p>;
```
kr-matthews, line 39:
```suggestion
  if (!loading && seniorDelegateRoles.length === 0) return <p>You cannot manage any regions.</p>;
```

app/webpacker/components/Panel/SeniorDelegate/Regions/index.jsx
```diff
@@ -0,0 +1,56 @@
...
+      groupType: groupTypes.delegate_region,
+    },
+  ));
+  const [selectedGroupIndex, setSelectedGroupIndex] = useInputState();
+
+  useEffect(() => {
+    if (seniorDelegateRoles?.length > 0) setSelectedGroupIndex(0);
+  }, [seniorDelegateRoles, setSelectedGroupIndex]);
+
+  const groupOptions = useMemo(() => seniorDelegateRoles.map((role, index) => ({
```
kr-matthews, line 33: Won't this need
```suggestion
  const groupOptions = useMemo(() => seniorDelegateRoles?.map((role, index) => ({
```
?

app/webpacker/components/Panel/SeniorDelegate/Regions/index.jsx
```diff
@@ -0,0 +1,56 @@
...
+      isGroupHidden: false,
+      status: delegateRegionsStatus.senior_delegate,
+      groupType: groupTypes.delegate_region,
+    },
+  ));
+  const [selectedGroupIndex, setSelectedGroupIndex] = useInputState();
+
+  useEffect(() => {
+    if (seniorDelegateRoles?.length > 0) setSelectedGroupIndex(0);
+  }, [seniorDelegateRoles, setSelectedGroupIndex]);
```
kr-matthews, line 31: I think you could remove this effect (and the `selectedGroupIndex === undefined` loading check) and just initialize the index state to `0`.

app/webpacker/components/Panel/SeniorDelegate/Regions/index.jsx
kr-matthews: I don't think we need to memoize things as much as we've been doing lately (but I also think it doesn't really hurt to do it). React is designed to handle re-renders efficiently, and our components and arrays are pretty small.
(Of course, when I try to find a reference to link to, I find tons of people arguing each side, so who knows for sure.)

app/controllers/api/v0/user_groups_controller.rb
gregorbg: Ah, I remember we had a discussion about this before.
Please add a comment about this intended inconsistency (because of false vs nil) to make sure that future me doesn't complain about this in the next few code reviews :stuck_out_tongue:

danieljames-dj: Done

app/webpacker/components/Panel/SeniorDelegate/Regions/index.jsx
danieljames-dj: Yes. To be honest, it was there in my code but uncommitted. Added now.

app/webpacker/components/Panel/SeniorDelegate/Regions/index.jsx
```diff
@@ -0,0 +1,52 @@
...
+
+  const groupOptions = useMemo(() => seniorDelegateRoles?.map((role, index) => ({
+    key: role.id,
+    text: role.group.name,
+    value: index,
+  })), [seniorDelegateRoles]);
+
+  if (!loading && seniorDelegateRoles.length === 0) return <p>You cannot manage any regions.</p>;
+  if (loading) return <Loading />;
+  if (error) return <Errored />;
```
gregorbg, line 37: Same nitpick as before: Do the `if`s for loading and error first, so then afterwards you only have to check `if seniorDelegateRoles.length === 0`

gregorbg (changes requested):

app/webpacker/components/Panel/SeniorDelegate/Regions/index.jsx
danieljames-dj: Done

gregorbg (approved):

PR #9032, opened 2024-03-06
Add kill switch to allow organizers closing registration

gregorbg:
Tehcnically it's a bugfix because they were supposed to be allowed to close registration in the first place.

But now that the new form sets `edited_by_user` correctly (which the old form was supposed to do but didn't!) a validation was triggered that would normally have caused troubles already long, long time ago.


dunkOnIT (approved):
Changes tested locally

PR #9034, opened 2024-03-06
Update Yarn to version 4.1.1

gregorbg:
Basically just a dependency update


gregorbg:
Unfortunately. no. At least I couldn't find any. This may have to do with the fact that listing your package manager in `package.json` is part of the `corepack` feature set in NodeJS, which is still being considered "experimental" although everybody (NPM, PNPM, Yarn) uses it and openly recommends it these days. But Dependabot can still provide an excuse by saying "we don't want to support experimental features".

PR #9036, opened 2024-03-07
Cache Microservice Registrations in a separate monolith table

gregorbg:
Preparation for showing them in the WCIF. Decided to split it up because there's some Rails black magic happening under the hood which would have been too complex to review in one big PR.

We are essentially storing a copy of "which registrations exist" in a shadow table. The table does *not* contain any other details like status, event IDs, etc. because that would lead to data duplication across (micro)services and synchronization hell. It only contains foreign keys to competition and user. That's it.

Instead, we are reading the "relevant" bits of data directly from the microservice. To that effect, there are two caching hooks in place:
1. Every time a registration is read from the Microservice, it is also "upserted" (inserted or updated) in the database.
2. Every time a MicroserviceRegistration is read from the database, the models are "enriched" with microservice information

For realizing point (2) above, I had to tap into Rails' ActiveRecord query interface. There is a method called `def records` in the Rails interface which collects the records just before returning them to the user. I hook into this method with some Ruby reflection magic to be able to link the entities up with the pre-loaded data.

### DIFF
--- a/app/models/microservice_registration.rb
+++ b/app/models/microservice_registration.rb
@@ -1,0 +1,6 @@
+# frozen_string_literal: true
+
+class MicroserviceRegistration < ApplicationRecord
+  belongs_to :competition
+  belongs_to :user
+end
--- a/app/models/microservice_registration.rb
+++ b/app/models/microservice_registration.rb
@@ -7,12 +7,12 @@ class MicroserviceRegistration < ApplicationRecord
   delegate :name, :email, to: :user
 
   attr_accessor :ms_registration
-  attr_writer :status, :event_ids
+  attr_writer :competing_status, :event_ids
 
   def load_ms_model(ms_model)
     self.ms_registration = ms_model
 
-    self.status = ms_model['status']
+    self.competing_status = ms_model['competing_status']
 
     self.event_ids = ms_model['lanes']&.find do |lane|
       lane['lane_name'] == 'competing'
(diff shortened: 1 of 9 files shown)


app/models/microservice_registration.rb
```diff
@@ -0,0 +1,47 @@
...
+  def event_ids
+    self.read_ms_data :event_ids
+  end
+
+  def accepted?
+    self.status == "accepted"
+  end
+
+  def deleted?
+    self.status == "deleted"
```
FinnIckler, line 45: deleted = cancelled

app/models/microservice_registration.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  delegate :name, :email, to: :user
+
+  attr_accessor :ms_registration
+  attr_writer :status, :event_ids
+
+  def load_ms_model(ms_model)
+    self.ms_registration = ms_model
+
+    self.status = ms_model['status']
```
FinnIckler, line 15: status => competing_status

gregorbg:
@thewca-bot deploy staging

gregorbg:
@thewca-bot deploy staging

gregorbg:
@thewca-bot deploy staging

gregorbg:
@thewca-bot deploy staging

gregorbg:
@thewca-bot deploy staging

gregorbg:
@thewca-bot deploy staging

gregorbg:
Tested on staging, working just fine :blush:

FinnIckler (approved):

PR #9037, opened 2024-03-07
Group Manager admin page

danieljames-dj:
This page is same as the Group Manager page in Leaders panel, but this will be in Board panel with all the teams, committees and councils.

### DIFF
--- a/app/controllers/api/v0/user_groups_controller.rb
+++ b/app/controllers/api/v0/user_groups_controller.rb
@@ -47,7 +47,7 @@ class Api::V0::UserGroupsController < Api::V0::ApiController
     end
 
     # Temporary hack to support old system teams/committees.
-    if group_type == "teams_committees"
+    if group_type == UserGroup.group_types[:teams_committees]
       Team.all_official.each do |team_committee|
         groups << {
           id: group_type + "_" + team_committee.id.to_s,
--- a/app/controllers/api/v0/user_groups_controller.rb
+++ b/app/controllers/api/v0/user_groups_controller.rb
@@ -46,6 +46,19 @@ class Api::V0::UserGroupsController < Api::V0::ApiController
       end
     end
 
+    # Temporary hack to support old system teams/committees.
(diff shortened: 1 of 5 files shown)


app/controllers/api/v0/user_groups_controller.rb
```diff
@@ -46,6 +46,19 @@ def index
       end
     end
 
+    # Temporary hack to support old system teams/committees.
+    if group_type == "teams_committees"
```
gregorbg, line 50: Use the constant from `UserGroup`?

app/webpacker/components/Panel/Leader/GroupsManager.jsx
```diff
@@ -141,36 +141,36 @@ function GroupTable({ group }) {
...
+export default function GroupsManager({ loggedInUserId }) {
+  const { data: roles, loading, error } = useLoadedData(apiV0Urls.userRoles.listOfUser(
+    loggedInUserId,
+    'groupName', // Sort params
+    { isActive: true, isGroupHidden: false, status: 'leader' },
+  ));
+
+  if (!loading && roles.length === 0) return <p>You are not a leader of any groups.</p>;
+  if (loading) return <Loading />;
+  if (error) return <Errored />;
```
gregorbg, line 174: You should do the `loading` and `error` checks first, and then below these two simply do `if (roles.length === 0)`

gregorbg (changes requested):
Only small nitpicks :D

app/controllers/api/v0/user_groups_controller.rb
danieljames-dj: Done

app/webpacker/components/Panel/Leader/GroupsManager.jsx
danieljames-dj: Oops ya, done

gregorbg (approved):

PR #9038, opened 2024-03-07
Improve performance of delegate_role metadata computation

gregorbg:
Not a lot of comments because we are 100% certain the entire `delegate_role` method will go away 3 or 4 commits down the line in Daniel's project.

But the basic notion is: By pre-loading the `actually_delegated_competitions` Rails association, we make sure that all rows are already in memory. Because of that, we can safely treat them as array which is faster than triggering another `SELECT MIN()` or `SELECT()` max query.

As a result, instead of 1 global query + 3 (min, max, count) extra queries **per Delegate** we only have to issue one global query and that's it.


danieljames-dj (approved):
Thanks a lot Gregor, tested in local and it is working amazingly.

PR #9042, opened 2024-03-08
Remove some usages of delegate_status

danieljames-dj:


app/models/user.rb
```diff
@@ -560,8 +560,12 @@ def admin?
     Rails.env.production? && EnvConfig.WCA_LIVE_SITE? ? software_team_admin? : software_team?
   end
 
+  def delegate_roles
+    active_roles.filter { |role| UserRole.is_group_type?(role, UserGroup.group_types[:delegate_regions]) }
+  end
+
   def any_kind_of_delegate?
-    delegate_status.present? || active_roles.any? { |role| UserRole.is_group_type?(role, UserGroup.group_types[:delegate_regions]) }
+    active_roles.any? { |role| UserRole.is_group_type?(role, UserGroup.group_types[:delegate_regions]) }
```
gregorbg, line 568: It is my understanding that `active_roles` is basically just a filtered version of `has_many :roles, class_name: "UserRole"`, i.e. `active_roles` can only ever hold "real" roles from the new role system.

This would mean that with your proposed changes of this PR, some Delegates who don't have a "proper" role yet will return `false` here when they are supposed to return `true`.

In other words, are we confident that each and every single Delegate in our whole database already has a new `UserRole` mapping?

app/models/user.rb
```diff
@@ -947,10 +951,6 @@ def editable_fields_of_user(user)
     fields += editable_personal_preference_fields(user)
     fields += editable_competitor_info_fields(user)
     fields += editable_avatar_fields(user)
-    # Delegate Status Fields
-    if admin? || board_member? || senior_delegate?
-      fields += %i(delegate_status region_id location)
-    end
```
gregorbg, line 953: If you are removing them here, have you made sure that the actual corresponding form (that would have checked for the presence of these three values) is also gone?

app/models/user.rb
danieljames-dj: Currently `has_many :roles` is getting overriden by `def roles`. So it can hold `unreal` roles as well. So this will work as expected.

app/models/user.rb
danieljames-dj: Actually the three values are not used in the edit user form. They have been already moved to separate form and is handled differently.

spec/models/user_spec.rb
```diff
@@ -591,13 +591,6 @@
       expect(organizer.editable_fields_of_user(registration.user).to_a).to eq [:name]
     end
 
-    it "allows senior delegates to assign delegate status" do
-      user = FactoryBot.create :user
-      senior_delegate = (FactoryBot.create :senior_delegate_role).user
-      expect(senior_delegate.can_edit_user?(user)).to eq true
-      expect(senior_delegate.editable_fields_of_user(user).to_a).to include(:delegate_status, :region_id, :location)
-    end
-
```
danieljames-dj, line 600: @gregorbg This test-case is already there for user-roles update method: https://github.com/thewca/worldcubeassociation.org/blob/04d99b188ad39c00ff317afdb432efd89fe0b2ff/spec/controllers/api/v0/user_roles_controller_spec.rb#L53

I assume both are doing somewhat same thing. Please correct me if I'm wrong.

gregorbg: Nah, should be fine! :+1:

gregorbg (approved):

PR #9044, opened 2024-03-08
Implemented Delegate Search in Probations page

danieljames-dj:


app/webpacker/components/SearchWidget/MultiSearchInput.jsx
gregorbg, line 1: You're changing too much code in this file to cater to specific changes in the new search endpoint.
Please model your backend search so that it returns its data in a way that is compatible with what the frontend expects.

app/webpacker/components/SearchWidget/MultiSearchInput.jsx
```diff
@@ -20,31 +21,45 @@ const classToComponent = {
...
 
-function ItemFor({ item }) {
-  const Component = classToComponent[item.class];
-  return (
-    <div className="selected-item">
-      <Component item={item} />
-    </div>
-  );
+function ItemFor({ item, model }) {
+  switch (model || '') {
```
gregorbg, line 25: Veto for this entire switch block. `classToComponent` already does what you need. The component returned by that lookup can do whatever it wants with the data, like wrap it in a `UserItem` as you are doing here.

app/webpacker/components/SearchWidget/MultiSearchInput.jsx
```diff
@@ -93,18 +109,27 @@ export default function MultiSearchInput({
       // FUI's dropdown will automatically remove selected items from the
       // options left for selection.
       fetchJsonOrError(url(debouncedSearch))
-        .then(({ data }) => setResults(data.result.map(itemToOption)))
+        .then(({ data }) => {
+          switch (model || '') {
```
gregorbg, line 113: Veto for this switch block as well. The search results should be structured in a way that the existing iterator can access them without having to change frontend code.

app/webpacker/components/SearchWidget/MultiSearchInput.jsx
```diff
@@ -67,6 +82,7 @@ export default function MultiSearchInput({
   goToItemUrlOnClick = false,
   placeholder,
   removeNoResultsMessage,
+  model,
```
gregorbg, line 85: You shouldn't have to pass in the model separately.

app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -515,4 +521,23 @@ def destroy
...
     end
   end
+
+  def search
+    query = params.require(:query)
+    group_type = params.require(:groupType)
+    roles = roles_of_group_type(group_type)
+    roles = roles.select { |role| UserRole.is_active?(role) }
+
+    query.split.each do |part|
```
gregorbg, line 531: What does this `split` do? Which char is it using to split the input query and why is that necessary?

app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -515,4 +521,23 @@ def destroy
       ).deliver_later
     end
   end
+
+  def search
+    query = params.require(:query)
+    group_type = params.require(:groupType)
+    roles = roles_of_group_type(group_type)
+    roles = roles.select { |role| UserRole.is_active?(role) }
```
gregorbg, line 529: Maybe call this `active_roles`, then?

gregorbg (changes requested):
Cool to see this search endpoint appearing! Unhappy about the frontend changes, though :(

Please keep the diff in `MultiSearchInput` to an **absolute minimum** and instead have your search backend return the data in a way that is already understood by the frontend component in its current way.

app/webpacker/components/SearchWidget/MultiSearchInput.jsx
danieljames-dj: Okay changed.

app/webpacker/components/SearchWidget/MultiSearchInput.jsx
danieljames-dj: Removed

app/webpacker/components/SearchWidget/MultiSearchInput.jsx
danieljames-dj: Removed

gregorbg (approved):

PR #9046, opened 2024-03-09
Delegate Regions page actions - Promote, Demote and End Role

danieljames-dj:
The old UI still allows it, because the UI should be open till all the changes to subregion is completed.

### DIFF
--- a/app/mailers/role_change_mailer.rb
+++ b/app/mailers/role_change_mailer.rb
@@ -3,14 +3,15 @@
 class RoleChangeMailer < ApplicationMailer
   private def role_metadata(role)
     metadata = {}
+    group = UserRole.group(role)
 
     # Populate the metadata list.
-    case UserRole.group(role).group_type
+    case group.group_type
     when UserGroup.group_types[:delegate_regions]
-      metadata[:region_name] = UserRole.group(role).name
+      metadata[:region_name] = group.name
       metadata[:status] = I18n.t("enums.user.role_status.delegate_regions.#{UserRole.status(role)}")
     when UserGroup.group_types[:translators]
-      metadata[:locale] = UserRole.group(role).metadata.locale
+      metadata[:locale] = group.metadata.locale
     end
     metadata
   end
--- a/app/mailers/role_change_mailer.rb
+++ b/app/mailers/role_change_mailer.rb
@@ -5,12 +5,12 @@ class RoleChangeMailer < ApplicationMailer
     metadata = {}
 
     # Populate the metadata list.
-    case role.group.group_type
+    case UserRole.group(role).group_type
     when UserGroup.group_types[:delegate_regions]
-      metadata[:region_name] = role.group.name
-      metadata[:status] = I18n.t("enums.user.role_status.delegate_regions.#{role.metadata.status}")
+      metadata[:region_name] = UserRole.group(role).name
+      metadata[:status] = I18n.t("enums.user.role_status.delegate_regions.#{UserRole.status(role)}")
     when UserGroup.group_types[:translators]
-      metadata[:locale] = role.group.metadata.locale
+      metadata[:locale] = UserRole.group(role).metadata.locale
     end
     metadata
   end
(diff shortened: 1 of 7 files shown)


app/mailers/role_change_mailer.rb
```diff
@@ -5,12 +5,12 @@ class RoleChangeMailer < ApplicationMailer
...
+    case UserRole.group(role).group_type
     when UserGroup.group_types[:delegate_regions]
-      metadata[:region_name] = role.group.name
-      metadata[:status] = I18n.t("enums.user.role_status.delegate_regions.#{role.metadata.status}")
+      metadata[:region_name] = UserRole.group(role).name
+      metadata[:status] = I18n.t("enums.user.role_status.delegate_regions.#{UserRole.status(role)}")
     when UserGroup.group_types[:translators]
-      metadata[:locale] = role.group.metadata.locale
+      metadata[:locale] = UserRole.group(role).metadata.locale
     end
```
gregorbg, line 14: You should probably store `UserRole.group(role)` in its own variable, because it's used at three times here within six lines of code)

app/views/role_change_mailer/notify_role_end.erb
```diff
@@ -1,4 +1,4 @@
-<p>The role has been removed for <%= @role.user.name %> in <%= @group_type_name %> on <%= @role.end_date %>.</p>
+<p>The role has been removed for <%= UserRole.user(@role).name %> in <%= @group_type_name %> on <%= @role[:end_date] %>.</p>
```
gregorbg, line 1: You're changing this to `@role[:end_date]` which looks like a "backported legacy role". Will this notation work with actual new roles?

app/webpacker/components/Panel/SeniorDelegate/Regions/Subregion.jsx
```diff
@@ -81,13 +130,22 @@ export default function Subregion({ title, groupId }) {
...
           {delegates.map((delegate) => (
             <Table.Row key={delegate.id}>
               <Table.Cell>{delegate.user.name}</Table.Cell>
               <Table.Cell>{I18n.t(`enums.user.role_status.delegate_regions.${delegate.metadata.status}`)}</Table.Cell>
+              <Table.Cell>
+                {canPromote(delegate)
+                  && <Button onClick={() => promoteDelegateAction(delegate)}>Promote</Button>}
+                {canDemote(delegate)
+                  && <Button onClick={() => demoteDelegateAction(delegate)}>Demote</Button>}
+                {!isLead(delegate)
```
gregorbg, line 146: (sidenote, not blocking: In the medium-term future we should probably have a distinct "pass on the torch" action to transfer leadership. It seems cumbersome to first promote a new member to Leader and technically have two leaders for a short amount of time just so you can demote the ex-leader and end their role)

gregorbg (changes requested):

app/mailers/role_change_mailer.rb
danieljames-dj: Done

app/views/role_change_mailer/notify_role_end.erb
danieljames-dj: Yes, this works for actual new roles as well, I confirmed. Attaching the screenshot:

Screenshot 2024-03-09 at 16 25 28

app/webpacker/components/Panel/SeniorDelegate/Regions/Subregion.jsx
danieljames-dj: Actually we have that. The management of leaders (or senior delegates or regional delegates) cannot be done from this page or GroupsManager. For that, there will be a page in Board panel. For Senior/Regional Delegates, there is one already (Region Manager), the one for team/committee leaders is under construction, but council leaders is already there but code has to be changed a bit because it's using some async/await which got missed in the review and I came to know it's not good practice in a later PR where it was rejected.

app/webpacker/components/Panel/SeniorDelegate/Regions/Subregion.jsx
```diff
@@ -37,6 +55,7 @@ export default function Subregion({ title, groupId }) {
   const [newDelegateUser, setNewDelegateUser] = useState(null);
   const [formError, setFormError] = useState(null);
   const { save, saving } = useSaveAction();
+  const confirm = useConfirm();
```
gregorbg, line 58: Is there a `ConfirmProvider` available in the DOM tree of this component somewhere?

danieljames-dj: Yes, in PanelTemplate. Since this is very widely used in different panels, I've already added it there.

gregorbg (approved):

PR #9048, opened 2024-03-09
Replace staff? with a method in roles

danieljames-dj:
Since the check whether a person is staff is based on the roles, the core logic of checking whether a role is staff role has been moved to user roles model

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -455,11 +455,6 @@ class User < ApplicationRecord
     team_member?(Team.board)
   end
 
-  # Officers are defined in our Bylaws. Every Officer has a team on the website except for the WCA Treasurer, as it is the WFC Leader.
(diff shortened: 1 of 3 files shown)


gregorbg (commented):
Code changes are totally fine. How well is the test coverage for this `staff?` method in our existing test suite? I feel like this is a core method of the entire `user.rb` class and as such should be very well tested.

danieljames-dj:
> Code changes are totally fine. How well is the test coverage for this `staff?` method in our existing test suite? I feel like this is a core method of the entire `user.rb` class and as such should be very well tested.

Didn't find any unit tests for the `staff?`. I've added them now.

gregorbg:
Thanks!

gregorbg (approved):

PR #9049, opened 2024-03-09
Migrate officers to roles

danieljames-dj:


danieljames-dj:
Ready for review, but can be merged only after https://github.com/thewca/worldcubeassociation.org/pull/9048. Will fix the unit tests once that PR is merged.

gregorbg:
What's the plan for merging on these two PRs?
As I understand it, the other PR #9048 that you mentioned makes changes which expect Officers to be in the new roles format already. Now on this PR you're basically saying "Officers can only be migrated to Roles format once the other PR that expects them to already be in Roles format is merged". This confuses me.

danieljames-dj:
> What's the plan for merging on these two PRs? As I understand it, the other PR #9048 that you mentioned makes changes which expect Officers to be in the new roles format already. Now on this PR you're basically saying "Officers can only be migrated to Roles format once the other PR that expects them to already be in Roles format is merged". This confuses me.

The other PR doesn't expect officers to be in the new roles. It is using roles in staff? but they are not actual roles yet. This PR is dependent on that PR because that PR will remove the usages of `officer?` and this PR is still having it. Once that PR is merged, then this should be fine.

db/migrate/20240309075507_delete_officer_team.rb
gregorbg, line 1: I would like to keep the legacy data, and just let them silently stay in the table without being accessed.
That way, we will have an easier time rolling back when something goes wrong.

Is there any problem caused by just leaving these entries in the DB?

gregorbg (changes requested):

db/migrate/20240309075507_delete_officer_team.rb
danieljames-dj: Okay removed. No problem is caused by leaving the entries in the DB. I didn't change the timestamp in schema.rb as I assume it's not a good practice to edit the timestamp. Is that fine?

gregorbg: Quite the opposite: It is absolutely required that the timestamp in `schema.rb` matches **exactly** with the newest timestamp present on the migration files of your git branch. So if you delete a migration, you have to change at least the timestamp in schema.rb accordingly.

Think of it this way: The migration that you just deleted did not affect the structure of the table, only the data. What if you had an experimental migration there that deleted some actual columns or even dropped a whole table? And then we decided to not include that migration, so you remove the Rails migration file. In that case, you must also edit `schema.rb` to adjust and revert the "drop table" changes from your experimental migration.

TL;DR: Of course removing a proposed migration file implies changing `schema.rb`!

danieljames-dj: Okay, got it. Thanks, updated `schema.rb`.

gregorbg (approved):

PR #9054, opened 2024-03-09
Migrate treasurer to roles & remove officer methods from team.rb

danieljames-dj:


app/jobs/sync_mailing_lists_job.rb
```diff
@@ -31,6 +30,9 @@ def perform
     GsuiteMailingLists.sync_group("sports@worldcubeassociation.org", Team.wsot.current_members.includes(:user).map(&:user).map(&:email))
     GsuiteMailingLists.sync_group("archive@worldcubeassociation.org", Team.wat.current_members.includes(:user).map(&:user).map(&:email))
 
+    treasurers = UserGroup.officers[0].active_roles.filter { |role| role.metadata.status == 'treasurer' }
```
gregorbg, line 33: Do we have a constant somewhere that we can reference, instead of writing the raw string `'treasurer'` here?

gregorbg: Also, why `officers[0]`?

gregorbg: Also, after `active_roles` you can include `.includes(:user)` for more efficient loading ;)

gregorbg (changes requested):

app/jobs/sync_mailing_lists_job.rb
danieljames-dj: Yes there was a constant, I've added that. I added officers[0] because `UserGroup.officers` was returning list of groups. I've now changed it to consider every officer groups even though there is only one. I hope that is fine.

`.includes(:user)` won't work I guess, because `active_roles` is not yet `ActiveRecord`, it's still an array. Maybe after the migration is completed?

gregorbg: Ah, yes. In that case my brain was already thinking too far ahead. Sorry! :innocent:

db/migrate/20240309141549_add_treasurer_to_roles.rb
gregorbg, line 1: What about past treasurers? Can we migrate their `TeamMember` data as well?

gregorbg (changes requested):

db/migrate/20240309141549_add_treasurer_to_roles.rb
danieljames-dj: We never stored treasurers in our database :-)

gregorbg: Oh, nevermind in that case :smile:

gregorbg: But wait.. We definitely _do_ store past leaders. So the information about Treasurer role can be inferred

danieljames-dj: Ya we will be backfilling the data as part of WRT. It's not so clear since when WFC Leaders are considered as treasurers. So once we get that, we will execute the migration manually.

gregorbg (approved):

PR #9055, opened 2024-03-09
Use react-datepicker in Competition Form

gregorbg:
Replacing the native browser picker (which received rather negative feedback from both Delegates and WCAT) with the `react-datepicker` implementation that was already used in other places.


gregorbg:
@thewca-bot deploy staging

gregorbg:
@thewca-bot deploy staging

app/webpacker/components/CompetitionForm/FormSections/CompDates.js
```diff
@@ -21,12 +21,26 @@ function daysUntil(date) {
 }
 
 export default function CompDates() {
```
kr-matthews, line 23: Maybe out of the scope of this PR, but what about automatically setting the end date when the start date is initially set? In fact, it might make sense to also automatically offset the end date whenever the start date changes, to preserve the duration.
The latter would also prevent invalid date ranges (right now the end date must be after the start date, but you can later change the start date to after the end date).

app/webpacker/components/wca/UtcDatePicker.js
```diff
@@ -0,0 +1,74 @@
...
+  // eslint-disable-next-line implicit-arrow-linebreak
+  return DateTime.fromISO(isoString, { zone: 'UTC' })
+    // but the react-datepicker uses local TZ,
+    // so we have to make the date _think_ it's local without actually converting the time
+    .setZone('local', { keepLocalTime: true })
+    // and finally output as JS-compatible object
+    .toJSDate();
+};
+
+const useIsoDate = (isoString) => useMemo(() => loadAsPseudoLocal(isoString), [isoString]);
```
kr-matthews, line 20: Why can't we just directly use `loadAsPseudoLocal` everywhere? The performance won't be affected by recalculating, and there aren't any side effects to prevent.

kr-matthews (approved):

app/webpacker/components/CompetitionForm/FormSections/CompDates.js
gregorbg: That's a very reasonable suggestion but should be in a different PR

app/webpacker/components/wca/UtcDatePicker.js
gregorbg: The intention was to make sure that we don't re-calculate all four dates if only one date changes. But indeed I might have overlooked something, because my experience with render side effects in React is rather low.

gregorbg: I'll merge this for now and leave room for future optimization.

Issue #8665, opened 2024-01-01
Deprecate contact page and move contact/website page to contact page


danieljames-dj:
The contact page (https://www.worldcubeassociation.org/contact) doesn't have any useful data that is not available in teams-committees page or website contact form (https://www.worldcubeassociation.org/contact/website). So I guess it will be good to delete the contact page, and move the contact/website page to contact route.

Before making the change, we will be needing approval on following three things:

1. Whether the description of each team/committee should be shown anywhere else or combined with the ones in team-committee page.
2. Whether it's fine to remove /contact page considering only usage is linked from home page, which can be redirected to contact/website page.
3. Whether moving contact/website to contact will be good, which means if we go to <website_url>/contact, it will open the contact form directly.

dunkOnIT:
I agree with only having one contact page, and not duplicating information that is displayed about teams. The question then is how do people contact teams - I see two options:

## Option 1 - Contact Form
- have a "Contact a team" radio button `/contact/website`
- When "Contact a team" is selected, we give a dropdown that allows for different teams to be selected - when a team is selected, a description of the team (ideally the one used on the `teams-committees` page) is displayed 

## Option 2 - Teams Page
- Emails are already displayed there - we just leave it as-is and remove `/contact`? Only downside is that it's not necessarily intuitive to users that the contact information for teams would be found on Teams/Committees, not on Contact
- We could have a `Looking to contact a team? Find their email on the Teams page` starburst on `/contact` that would resolve that confusion?

PR #9057, opened 2024-03-10
Remove deprecated contact page

danieljames-dj:


config/locales/en.yml
gregorbg, line 1: Where did all of the removed text go?

app/views/contacts/index.html.erb
gregorbg, line 1: I feel like you should keep the translations in dot notation (i.e. `.inquiries.competition`) and move the i18n keys to fit what the Rails view expects.

Not 100% definitive on this, so feel free to argue.

gregorbg (commented):
Can you please give some context on this PR? Who initiated it, why was it initiated, are we truly okay with removing the old page, what kind of information is lost, etc.

app/views/contacts/index.html.erb
danieljames-dj: I didn't move any existing translations just to avoid extra work by translators. I've plans of migrating contacts page to React after few days. During that time, I'll anyway move the locales to `en.page.*`. So is it fine to not have a temporary transition for the locales?

config/locales/en.yml
danieljames-dj: They were part of the contact page that I deleted.

danieljames-dj (commented):
I actually felt a couple of months ago that the [contact information page](https://www.worldcubeassociation.org/contact) doesn't have any useful data and every data there is repetition of what is available in https://github.com/thewca/worldcubeassociation.org/issues/8665.

I think I forgot to check with you before removing, sorry for that.

We are not losing any data other than few descriptions. For those, I've anyway initiated an email thread with the Leaders few days ago to give a description that can go into the new teams-committees page.

The reason why I did this now is because this page displays the names and email of Board and team members, and hence it is creating one more page where I have to do the changes everytime during roles migration. So I thought removing it will be easy.

app/views/contacts/index.html.erb
gregorbg: Yup, makes sense!

gregorbg (approved):

PR #9066, opened 2024-03-11
Fix error thrown when trying to pay with Stripe using ISK currency

dunkOnIT:
Ran into this while testing for the StripeTransaction refactor - seems like special currency cases throw an error that wasn't caught by the tests (and I guess we haven't had a comp with a special currency using Stripe yet?), because `amount_to_stripe` returns a float which Stripe doesn't like.

You can repro the error by trying to register and pay for this competition: https://staging.worldcubeassociation.org/competitions/IcelandTestComp2024/register

The fix just ensures that the return value of stripe_amount is an integer, and the tests now explicitly check for integer as well.


gregorbg (approved):
Thanks for catching this! Seems like indeed we didn't have any ISK competition using Stripe yet (which is curious, but I'll accept it)

PR #9067, opened 2024-03-11
Migrated StripeTransaction to StripeRecord

dunkOnIT:
- Creates a migration which: 
  - Renames `stripe_transactions` table to `stripe_records`
  - Renames foreign keys accordingly
  - Changes all instances "StripeTransaction" in `registration_payments:receipt_type` to "StripeRecord" (this was the only polymorphic association I could find)

Big change so I'd rather raise it as a separate PR before doing the StripePaymentIntent -> PaymentIntent refactor


db/migrate/20240311053739_rename_stripe_transaction_to_stripe_record.rb
```diff
@@ -0,0 +1,27 @@
...
+    rename_column :stripe_webhook_events, :stripe_transaction_id, :stripe_record_id
+    rename_column :stripe_payment_intents, :stripe_transaction_id, :stripe_record_id
+
+    reversible do |direction|
+      direction.up do
+        execute <<-SQL
+          UPDATE registration_payments
+          SET receipt_type = 'StripeRecord'
+          WHERE receipt_type = 'StripeTransaction'
+        SQL
```
gregorbg, line 15: You can do this without manually writing SQL strings by executing `RegistrationPayment.update_all(receipt_type: 'StripeRecord')`

db/migrate/20240311053739_rename_stripe_transaction_to_stripe_record.rb
```diff
@@ -0,0 +1,27 @@
...
+      end
+
+      direction.down do
+        execute <<-SQL
+          UPDATE registration_payments
+          SET receipt_type = 'StripeTransaction'
+          WHERE receipt_type = 'StripeRecord'
+        SQL
```
gregorbg, line 23: String-less update, see other comment.

db/schema.rb
```diff
@@ -821,6 +821,19 @@
...
+    t.string "payload"
+    t.integer "amount_in_cents"
+    t.string "currency_code"
+    t.string "record_type"
+    t.bigint "parent_record_id"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["parent_record_id"], name: "index_paypal_records_on_parent_record_id"
+  end
+
```
gregorbg, line 836: Seems like this slipped in from some other changes?
Note that Rails `db:migrate` consists of a step that dumps the DB schema as is currently present in your local development setup. Rails has no way of telling whether a certain table was created as part of the feature branch/PR you're currently working on, or as one of the other three feature branches which also exist on your machine. That's why you need to be careful with `schema.rb`.

db/schema.rb
```diff
@@ -1059,7 +1072,7 @@
     t.datetime "updated_at", precision: nil, null: false
     t.bigint "parent_transaction_id"
     t.index ["parent_transaction_id"], name: "fk_rails_6ad225b020"
-    t.index ["status"], name: "index_stripe_transactions_on_status"
+    t.index ["status"], name: "index_stripe_records_on_status"
```
gregorbg, line 1075: Strictly speaking, you're not renaming this index as part of the migration. It would seem that you manually edited this into `schema.rb`, which is problematic because then the existing index in prod won't actually be renamed.

There is no `rename_index` method to my knowledge (Google may correct me on this one, though) and you probably have to run two separate `drop_index` (or was it `delete_index`? I can never remember sorry...) and `create_index` statements.

db/schema.rb
```diff
@@ -1068,10 +1081,10 @@
     t.string "account_id"
     t.datetime "created_at_remote", precision: nil, null: false
     t.boolean "handled", default: false, null: false
-    t.bigint "stripe_transaction_id"
+    t.bigint "stripe_record_id"
     t.datetime "created_at", precision: nil, null: false
     t.datetime "updated_at", precision: nil, null: false
-    t.index ["stripe_transaction_id"], name: "index_stripe_webhook_events_on_stripe_transaction_id"
+    t.index ["stripe_record_id"], name: "index_stripe_webhook_events_on_stripe_record_id"
```
gregorbg, line 1087: Index problem, see above.

lib/database_dumper.rb
```diff
@@ -856,7 +856,8 @@ def self.actions_to_column_sanitizers(columns_by_action)
         ),
       ),
     }.freeze,
-    "stripe_transactions" => :skip_all_rows,
+    "paypal_records" => :skip_all_rows,
```
gregorbg, line 859: Whoopsie?

db/schema.rb
```diff
@@ -1241,12 +1254,13 @@
   add_foreign_key "active_storage_attachments", "active_storage_blobs", column: "blob_id"
   add_foreign_key "active_storage_variant_records", "active_storage_blobs", column: "blob_id"
   add_foreign_key "oauth_openid_requests", "oauth_access_grants", column: "access_grant_id", on_delete: :cascade
+  add_foreign_key "paypal_records", "paypal_records", column: "parent_record_id"
```
gregorbg, line 1257: Whoopsie?

gregorbg (changes requested):
Some detail remarks, overall work looks fine!
Also, I want this tested on staging before final merge/deployment.

db/migrate/20240311053739_rename_stripe_transaction_to_stripe_record.rb
dunkOnIT: Ah of course! We don't need to have any `WHERE` logic because we only have `StripeTransaction`s at present

gregorbg: Actually, you can refine this approach: `RegistrationPayment.where(receipt_type: 'StripeTransaction').update_all(receipt_type: 'StripeRecord')`

gregorbg:
Also, you have merge conflicts :P

lib/database_dumper.rb
dunkOnIT: I think this is also resolved now?

db/schema.rb
gregorbg: TIL: The indexes are actually automatically migrated! :star_struck:

gregorbg (approved):

PR #9068, opened 2024-03-11
Add registration v2 internationalization strings

FinnIckler:
Will need to do a final check on if that is all we need. 
There's two ways we can include it in the v2 build:
a) Have the build pull the latest registrations from git and bundle it as part of the js build
b) Make sure that the translations file from webpacker always ends up with the same name and we just load it from the wca,org asset folder

If we don't want to massively blow up our js bundle, it's probably better to do b)


config/locales/en.yml
```diff
@@ -1658,6 +1658,79 @@ en:
...
+        spots_remaining: "; %{spots} spot(s) remaining"
+        waitlisted: "Waitlisted registrations"
+        comment: "Comment"
+        admin_note: "Admin Note"
+      tabs:
+        payment: "Payment"
+        more: "More"
+      fees:
+        none: "No Entry Fee"
+        guest_none: "Guests attend for free"
```
gregorbg, line 1678: Quite sure we already have those

config/locales/en.yml
```diff
@@ -1658,6 +1658,79 @@ en:
...
+      pdf:
+        #context: <0> indicates an html tag for internationalizing in React, see https://docs.worldcubeassociation.org/guides/translating_the_website.html#using-react
+        link: "As a <0>%{pdfLink}</0>"
+        download: "Download all of the competitions details."
+      list:
+        empty: "No matching Records found"
+        spots_remaining: "; %{spots} spot(s) remaining"
+        waitlisted: "Waitlisted registrations"
+        comment: "Comment"
+        admin_note: "Admin Note"
```
gregorbg, line 1672: Probably also have these

config/locales/en.yml
```diff
@@ -1658,6 +1658,79 @@ en:
             allow_registration_without:
               "true": "Competitors will be allowed to register for events they're not qualified for (I'll remove them later)"
               "false": "Competitors MUST be qualified to register for an event"
+        #context: new registration system
+    registration_v2:
+      pdf:
+        #context: <0> indicates an html tag for internationalizing in React, see https://docs.worldcubeassociation.org/guides/translating_the_website.html#using-react
+        link: "As a <0>%{pdfLink}</0>"
+        download: "Download all of the competitions details."
+      list:
```
gregorbg, line 1667: This mixes "enumerated values that describe what a column means" with "user-friendly help text to display below/above tables". They should go in separate branches.

config/locales/en.yml
```diff
@@ -1658,6 +1658,79 @@ en:
...
+        no_changes: "There are no changes"
+        email_send: "Send Email"
+        email_copy: "Copy Email"
+        move_pending: "Move to Pending"
+        move_waiting: "Move to Waiting List"
+        cancel: "Cancel Registration"
+        too_many: "Accepting all these registrations would go over the competitor limit by %{count}"
+      register:
+        error: "Registration failed with error %{error}"
+        early_registration: "Registration is not open yet, but you can still register as a competition organizer or delegate."
```
gregorbg, line 1696: Nitpick: `Delegate` (capital D)

config/locales/en.yml
```diff
@@ -1658,6 +1658,79 @@ en:
...
+        comment: "Additional comments to the organizers"
+        until: "You can update your registration until %{date}"
+        passed: "You can no longer update your registration"
+        editing_disabled: "Registration editing is disabled for this competition"
+        registration_status:
+          header: "Your Registration Status:"
+          accepted: "accepted"
+          pending: "pending"
+          waiting: "Waiting List"
+          cancelled: "cancelled"
```
gregorbg, line 1706: Inconsistent capitalization. Also, we probably already have those.

gregorbg (changes requested):
Some typos/nitpicks, and general remarks about reusing keys.

Stuff like translating `accepted` should go into a shared branch, and I am pretty sure we already have that information/translation somewhere.
As a general rule of thumb, a registration is not called "accepted" only because it lives in the V2 microservice. It would also be called "accepted" in legacy". So we should share those under a general "registration constants" key.

Stuff like localizing specific error codes, or the new "Taking a bit longer than usual, don't go away!" messages are very good and reasonable to keep under a "V2" parent key.

config/locales/en.yml
FinnIckler: There is 
```
competition_medium:
      status:
        accepted: "Accepted"
        pending: "Pending"
```
But what I want to do is just `t(competitions.registration_v2.register.registrationStatus.${registration.competing.registration_status})`, but I can't do that if half of the statusses are in one key and the other in the other. A solution would be to add those to `competition_medium.status,` but that would cause and unused key error

config/locales/en.yml
FinnIckler: There's "Payment Information"?

config/locales/en.yml
gregorbg: In that case, move the other new status flags to the same place where the old ones already are and add ignore rules to the `i18n-tasks.yml` config to suppress the unused keys

config/locales/en.yml
```diff
@@ -1691,6 +1693,67 @@ en:
...
+        #context: new registration system
+    registration_v2:
+      pdf:
+        #context: <0> indicates an html tag for internationalizing in React, see https://docs.worldcubeassociation.org/guides/translating_the_website.html#using-react
+        link: "As a <0>%{pdfLink}</0>"
+      info:
+        download: "Download all of the competitions details."
+      list:
+        empty: "No matching Records found"
+        spots_remaining: "; %{spots} spot(s) remaining"
```
gregorbg, line 1705: Not a big fan of a translation string beginning with a semi-colon.
Is there a more clever way to do this via string interpolation?

gregorbg (approved):

PR #9072, opened 2024-03-12
Clean up staging deployment workflow

gregorbg:
- Add `--fail` flag to MariaDB script curl to make sure we're not piping bogus into `bash`


gregorbg:
Merging because CI literally doesn't touch these files at all.

PR #9073, opened 2024-03-12
Refactor `StripePaymentIntent` to generic `PaymentIntent`

dunkOnIT:
### Status Notes
- I'm letting the PaymentRecord be the authority on what the appropriate wca_status should be for its corresponding payment provider status - it holds the mapping, and we have tests in place to make sure that the mapping is consistent with the list of wca_statuses defined in the PaymentIntent
- I've replaced the previous PaymentIntent scopes with `unpaid` and `started` - I believe these fulfill the same needs, but the logic of this will definitely need to be double-checked
- Note the wca_status definitions/purposes on `models/payment_intent.rb:24` - this is just my best guess at the moment

## Change backlog
- [ ] rename payment_intent.client_secret to something more generic
- [ ] rename/refactor delegated stripe_id property
- [x] Change `api_types` to record_types

## Steps
- [x] Add a polymorphic association to a :payment_record
- [x] Refactor stripe-specific method to be generic (use the same logic and just detect what the association is? Similar to what we do for CompetitionPaymentIntegration?)
- [x] Add status column to PaymentIntent
  - [x] Write migration
  - [x] Create a mapping between Stripe statuses and PaymentIntent statuses
  - [x] Add tests to make sure that (a) stripe mapping contains all PI statuses, and (b) all stripe statuses are mapped to wca_statuses
  - [x] Add a comprehensive set of tests that check for consistency in status advancements
  - [x] Update PaymentIntent status as part of update_status_and_charges
  - [x] Add a validation on the PaymentIntent model
  - [x] Add a validation on the StripeRecord model
- [x] Get tests passing
- [x] Write a migration for populating wca_status from stripe_status? 
- [x] Fix the hack in valid_status_combination
- [x] Figure out why payment form is showing `payment_service_error`
- [ ] (BLOCKED: NEED VPN CONNECTION) Deploy to staging & test
  - [ ] Making & refunding payments still works
  - [ ] Status migration worked as expected

##  Questions
- [x]  What is the user_id field for on StripePaymentIntent? The user who is paying? (I suggest a rename to make this explicit, unless its a foreign key I guess?)


dunkOnIT:
## Questions 
- [ ] Is my definition of `payment_intent.started?` appropriate given that it is used to determine whether or not to recycle a payment intent? (In light of our conversation on March 21st ~11:50 UTC)

app/models/payment_intent.rb
```diff
@@ -0,0 +1,140 @@
...
+  belongs_to :initiated_by, class_name: 'User' # For now only users can initiate payments - in future, this may become a polymorphic association
+  belongs_to :confirmation_source, polymorphic: true, optional: true
+  belongs_to :cancellation_source, polymorphic: true, optional: true
+
+  validate :valid_status_combination
+
+  scope :unpaid, -> { where.not(wca_status: 'succeeded') }
+  scope :started, -> { where.not(wca_status: ['created']) }
+
+  # TODO: Refactor this or move it into this class
```
dunkOnIT, line 15: I'm inclined to leave this for another PR - this one already has a LOT of changes

app/models/payment_intent.rb
```diff
@@ -0,0 +1,140 @@
...
+  enum wca_status: {
+    created: 'created', # A record has been created on the payment provider's system (no payment action yet initiated by the user)
+    pending: 'pending', # The user is now attempting to pay, but has not reached a completion state yet
+    partial: 'partial', # Some but not all funds have been paid (eg, if the user has selected an instalment-based payment option)
+    failed: 'failed', # The payment did not succeed for any reason (insufficient funds, user error) but the user can still try to pay
+    succeeded: 'succeeded', # Completion state - The full amount due is confirmed as being paid by the payment provider
+    canceled: 'canceled', # Completion state - the user has indicated that they will no longer attempt to complete payment
+  }
+
+  def confirmed?
```
dunkOnIT, line 33: Perhaps this should be `succeeded?` instead?

app/models/payment_intent.rb
```diff
@@ -0,0 +1,140 @@
...
+    !self.confirmed_at.nil?
+  end
+
+  def started?
```
dunkOnIT, line 37: My understanding is that `started` basically denotes that the user has taken some kind of action to advance the state of the PaymentIntent - ie, we haven't _just_ created it with the payment provider

app/views/registrations/_payment_intents.html.erb
```diff
@@ -1,7 +1,7 @@
 <div>
   There are pending payments for this registration. Please wait until they are confirmed before paying again.
 
-  <% @registration.stripe_payment_intents.pending.each do |payment_intent| %>
+  <% @registration.payment_intents.pending.each do |payment_intent| %>
```
dunkOnIT, line 4: Does the `pending` scope here cover enough possible statuses? (ie, created/partial/failed/canceled are all excluded from it, some of which should perhaps be included)

app/views/registrations/register.html.erb
```diff
@@ -77,7 +77,7 @@
             </h4>
           </div>
           <div class="panel-body">
-            <% if @registration.stripe_payment_intents.processing.any? %>
+            <% if @registration.payment_intents.unpaid.any? %>
```
dunkOnIT, line 80: See note about `unpaid` and whether it is an appropriate replacement for `processing`

dunkOnIT:
@thewca-bot deploy staging

dunkOnIT:
BLOCKER: 
- Need to test wca_status migration on staging
- Can't set up payments on staging because of lack of paypal keys in Vault
- Can't set paypal keys in vault because I can't access the URL in the vault gdoc

dunkOnIT:
@thewca-bot deploy staging

dunkOnIT:
ISSUE: 
- I can't remove the hackish substitute for stripe_record.payment_intent
- Changing the hack in stripe_record and running spec/models/stripe_record_spec, the status consistency tests fail.
- Running byebug in one of those status consistency tests that's failing, I get nil for stripe_record.payment_intent - will need Gregor's eyes on this probably. There must be some kind of subtle mistake I'm making.

app/views/registrations/_payment_intents.html.erb
gregorbg: No, I don't think so. This is only meant to show actual _pending_ intents. If an intent has failed, we will allow the user to restart it while loading the UI widget, instead of forcing them to stare at a list detailing that it failed.

app/views/registrations/register.html.erb
gregorbg: (same logic as in my other comment applies)

db/migrate/20240312122903_make_stripe_payment_intent_generic.rb
```diff
@@ -0,0 +1,47 @@
...
+    rename_table :stripe_payment_intents, :payment_intents
+    rename_column :stripe_records, :status, :stripe_status
+    rename_column :stripe_records, :api_type, :record_type
+    rename_column :paypal_records, :status, :paypal_status
+
+    reversible do |direction|
+      direction.up do
+        RegistrationPayment.update_all(receipt_type: 'StripeRecord')
+
+        PaymentIntent.update_all(payment_record_type: 'StripeRecord')
```
gregorbg, line 26: Haven't we already done this part in https://github.com/thewca/worldcubeassociation.org/pull/9067 ?!

db/migrate/20240312122903_make_stripe_payment_intent_generic.rb
```diff
@@ -0,0 +1,47 @@
...
+        RegistrationPayment.update_all(receipt_type: 'StripeTransaction')
+
+        PaymentIntent.find_each do |intent|
+          intent.assign_attributes(stripe_record_id: intent.payment_record_id)
+          intent.save(validate: false)
+        end
+      end
+    end
+
+    remove_column :payment_intents, :stripe_record_id, :integer
```
gregorbg, line 45: I think it is sufficient if you _rename_ this column (and drop the `t.integer` column creation above).

db/migrate/20240312122903_make_stripe_payment_intent_generic.rb
```diff
@@ -0,0 +1,47 @@
...
+
+        PaymentIntent.find_each do |intent|
+          intent.assign_attributes(payment_record_id: intent.stripe_record_id)
+          intent.set_wca_status
+          intent.save
+        end
+      end
+
+      direction.down do
+        RegistrationPayment.update_all(receipt_type: 'StripeTransaction')
```
gregorbg, line 36: Should already be done, see above

db/schema.rb
```diff
@@ -773,6 +774,16 @@
...
+  create_table "microservice_registrations", charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
+    t.string "competition_id"
+    t.integer "user_id"
+    t.text "roles"
+    t.boolean "is_competing", default: true, null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["competition_id", "user_id"], name: "index_microservice_registrations_on_competition_id_and_user_id", unique: true
+  end
+
```
gregorbg, line 786: Whoopsie?

db/migrate/20240312122903_make_stripe_payment_intent_generic.rb
```diff
@@ -0,0 +1,47 @@
...
+      t.rename :canceled_by_type, :cancellation_source_type
+
+      t.string :wca_status
+      t.string :payment_record_type
+      t.integer :payment_record_id
+    end
+
+    rename_table :stripe_payment_intents, :payment_intents
+    rename_column :stripe_records, :status, :stripe_status
+    rename_column :stripe_records, :api_type, :record_type
```
gregorbg, line 19: Why are you renaming this?

db/schema.rb
```diff
@@ -1254,11 +1266,10 @@
   add_foreign_key "active_storage_attachments", "active_storage_blobs", column: "blob_id"
   add_foreign_key "active_storage_variant_records", "active_storage_blobs", column: "blob_id"
   add_foreign_key "oauth_openid_requests", "oauth_access_grants", column: "access_grant_id", on_delete: :cascade
+  add_foreign_key "payment_intents", "users", column: "initiated_by_id"
   add_foreign_key "paypal_records", "paypal_records", column: "parent_record_id"
   add_foreign_key "sanity_check_exclusions", "sanity_checks"
   add_foreign_key "sanity_checks", "sanity_check_categories"
-  add_foreign_key "stripe_payment_intents", "stripe_records"
-  add_foreign_key "stripe_payment_intents", "users"
```
gregorbg, line 1261: At least this one foreign key should be maintained

gregorbg (changes requested):

app/models/payment_intent.rb
dunkOnIT: I'm happy with the naming of the db fields, I was thinking more along the lines of being consistent with the statuses defined in the wca_status enum

db/migrate/20240312122903_make_stripe_payment_intent_generic.rb
dunkOnIT: That's what I was doing initially, but it doesn't reorder them in the database - so you end up the record_id being the 3rd column, and record_type aaaaall the way at the end - which is (1) bad devex for developers who know there is a polymorphic association, and (2) doesn't make it obvious to developers who don't know there's a polymorphic association that one exists

Totally happy to rename if there's a way to reorder the SQL columns though

app/models/payment_intent.rb
```diff
@@ -0,0 +1,126 @@
...
+  belongs_to :holder, polymorphic: true
+  belongs_to :payment_record, polymorphic: true
+  belongs_to :initiated_by, class_name: 'User' # For now only users can initiate payments - in future, this may become a polymorphic association
+  belongs_to :confirmation_source, polymorphic: true, optional: true
+  belongs_to :cancellation_source, polymorphic: true, optional: true
+
+  validate :valid_status_combination
+
+  scope :unpaid, -> { where.not(wca_status: 'succeeded') }
+  scope :started, -> { where.not(wca_status: ['created']) }
```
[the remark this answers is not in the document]
dunkOnIT: I haven't the faintest clue

app/models/payment_intent.rb
```diff
@@ -0,0 +1,126 @@
...
+    self.wca_status != 'created'
+  end
+
+  def retrieve_intent
+    Stripe::PaymentIntent.retrieve(
+      self.stripe_id,
+      client_secret: client_secret,
+      stripe_account: self.find_account_id,
+    )
+  end
```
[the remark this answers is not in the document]
dunkOnIT: I would actually take the opposite approach, that this _shouldn't_ be made provider-specific until we actually implement Paypal in the PaymentIntent model. I'd rather change the other code that has clauses for Paypal (leaving in the `else` clause that raises a general error). 

And thus I should also add an if/else here as well, I guess.

app/models/payment_intent.rb
```diff
@@ -0,0 +1,126 @@
...
+
+  def update_status_and_charges(api_intent, action_source, source_datetime = DateTime.current, &block)
+    if payment_record_type == 'StripeRecord'
+      update_stripe_status_and_charges(api_intent, action_source, source_datetime, &block)
+    elsif payment_record_type == 'PaypalRecord'
+      raise 'Paypal is not enabled in production' if PaypalInterface.paypal_disabled?
```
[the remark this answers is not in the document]
dunkOnIT: This hints at my premature inclusion of paypal - see above comment

app/models/stripe_record.rb
```diff
@@ -36,28 +47,43 @@ class StripeRecord < ApplicationRecord
   # Also saves us from some pains because JSON columns are highly inconsistent among MySQL and MariaDB.
   serialize :parameters, coder: JSON
 
+  def determine_wca_status
+    WCA_TO_STRIPE_STATUS_MAP.each do |key, values|
+      return key if values.include?(stripe_status)
+    end
+
+    raise Exception("No associated wca_status for stripe_status: #{stripe_status} - our tests should prevent this from happening!")
```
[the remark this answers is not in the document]
dunkOnIT: Excellent point. That presents a whole world of complexity. The best option I can think of is, if this point is reached, store an `unknown` status on the PaymentIntent - which at some point we may hopefully pick up? 

(We could even add a note on the frontend - perhaps in the `/registrations/edit` view - to alert WST because of a status inconsistency? Because otherwise this could go undetected for months, if `unknown` doesn't cause any errors to be thrown anywhere else in the code)

app/views/registrations/_payment_intents.html.erb
dunkOnIT: Ah interesting - in this case, is our current status mapping for `pending` too broad? (wca_status: [stripe_statuses])
```ruby
pending: %w[pending requires_capture processing requires_confirmation requires_action],
```

app/views/registrations/register.html.erb
dunkOnIT: In which case, we're literally ONLY looking for the Stripe `processing` status? Presumably because for `processing` we want to show them that they have a payment busy processing, but in all other cases we just want to render the payment form? 

I'm a little confused why `processing` specifically is used here, when the general logic seems quite similar to the above case (where `pending` was used before)

db/migrate/20240312122903_make_stripe_payment_intent_generic.rb
dunkOnIT: Yes we absolutely have. Not sure how I ended up with records that hadn't been affected by it - change reverted.

db/migrate/20240312122903_make_stripe_payment_intent_generic.rb
dunkOnIT: Seems more intuitive to me to call it the type of the record as opposed to the type of API it came from (I see the argument for both though, not a hill I'll die on)

db/schema.rb
dunkOnIT: Ah this was probably because I pushed during some migration mayhem - should be resolved now?

db/schema.rb
dunkOnIT: I believe it is maintained by line 1275 - note the change in name from `user_id` to `iniatiated_by_id`

app/models/stripe_record.rb
```diff
@@ -36,28 +47,43 @@ class StripeRecord < ApplicationRecord
   # Also saves us from some pains because JSON columns are highly inconsistent among MySQL and MariaDB.
   serialize :parameters, coder: JSON
 
+  def determine_wca_status
+    WCA_TO_STRIPE_STATUS_MAP.each do |key, values|
+      return key if values.include?(stripe_status)
+    end
```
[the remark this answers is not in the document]
dunkOnIT: See updated code - I didn't see a way to return the key from within the find block

app/models/payment_intent.rb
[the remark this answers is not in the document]
dunkOnIT: See changes

app/models/payment_intent.rb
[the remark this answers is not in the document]
dunkOnIT: See changes

app/models/payment_intent.rb
gregorbg: Okay, but then how does `confirmed` match an enum'ed list of statuses that can contain `[created, pending, partial, failed, succeeded, canceled]`?

app/models/payment_intent.rb
```diff
@@ -0,0 +1,126 @@
...
+    succeeded: 'succeeded', # Completion state - The full amount due is confirmed as being paid by the payment provider
+    canceled: 'canceled', # Completion state - the user has indicated that they will no longer attempt to complete payment
+  }
+
+  def confirmed?
+    !self.confirmed_at.nil?
+  end
+
+  def started?
+    self.wca_status != 'created'
```
gregorbg, line 37: Fun fact about Rails enums: Because you defined `created` as an enum status, you can access `self.started?` for free. Rails adds the question-mark method as a convenience for you (same applies to `some_intent.failed?`, `some_intent.succeeded?` and similar)

app/models/payment_intent.rb
gregorbg: Also, just now realizing that this clashes with some assumptions about enums, see my other comment about Rails enum convenience methods

app/models/payment_intent.rb
```diff
@@ -0,0 +1,126 @@
...
+          confirmation_source: nil,
+          canceled_at: nil,
+          cancellation_source: nil,
+        )
+      end
+    end
+  end
+
+  # Sets the wca_status field based on the status field of the associated PaymentRecord
+  def set_wca_status
```
gregorbg, line 112: Generally no `set_` or `get_` in Ruby as per convention. Although I'm debating with myself what would be a better name here.
Also, since calling this method modifies the state of the object it is called on, you should end the method name with `!` (similar to how methods returning booleans should end with `?` according to convention)

app/models/stripe_record.rb
```diff
@@ -36,28 +47,42 @@ class StripeRecord < ApplicationRecord
   # Also saves us from some pains because JSON columns are highly inconsistent among MySQL and MariaDB.
   serialize :parameters, coder: JSON
 
+  def determine_wca_status
+    result = WCA_TO_STRIPE_STATUS_MAP.find { |key, values| values.include?(stripe_status) }
+    return result.first unless result.nil?
```
gregorbg, line 52: `result&.first || raise "Foo"` _should_ do the same thing, although I'm writing this from the top of my head so you should quickly test it

db/schema.rb
gregorbg: Ah, gotcha.

db/migrate/20240312122903_make_stripe_payment_intent_generic.rb
gregorbg: The reason I'm asking is because this collides with some of our other tables.

In `PaymentIntent` we use `record_type` to define "Is it Stripe or PayPal?" whereas now, inside the Stripe tables, there's _also_ a column called `record_type` which defines something else...

gregorbg: That's why I argued to stick with `api_type`, but also any other non-`record-type` name will do :)

db/migrate/20240312122903_make_stripe_payment_intent_generic.rb
```diff
@@ -0,0 +1,43 @@
...
+          intent.assign_attributes(payment_record_id: intent.stripe_record_id)
+          intent.set_wca_status
+          intent.save
+        end
+      end
+
+      direction.down do
+        PaymentIntent.find_each do |intent|
+          intent.assign_attributes(stripe_record_id: intent.payment_record_id)
+          intent.save(validate: false)
```
gregorbg, line 36: This is literally exactly what `intent.update_attribute(stripe_record_id: intent.payment_record_id)` was designed to do. (It implicitly saves and skips validations)

db/migrate/20240312122903_make_stripe_payment_intent_generic.rb
gregorbg: In line with the `status` -> `stripe_status` change above, we should either:
1. Change `status` to `api_status` XOR
2. Change `api_type` to `stripe_type`

db/migrate/20240312122903_make_stripe_payment_intent_generic.rb
gregorbg: In the statement that creates the column, add an `after: :existing_column_name` argument :wink: 

`t.string :cool_new_stuff, after: :existing_stuff_at_beginning_of_table`

db/migrate/20240312122903_make_stripe_payment_intent_generic.rb
```diff
@@ -0,0 +1,43 @@
...
+    change_table :stripe_payment_intents do |t|
+      t.rename :user_id, :initiated_by_id # Make the purpose of the user_id column more clear
+      t.rename :confirmed_by_id, :confirmation_source_id
+      t.rename :confirmed_by_type, :confirmation_source_type
+      t.rename :canceled_by_id, :cancellation_source_id
+      t.rename :canceled_by_type, :cancellation_source_type
+
+      t.string :wca_status
+      t.string :payment_record_type
+      t.integer :payment_record_id
```
gregorbg, line 14: Nitpick: This should be a b`t.bigint` instead of `integer`, but that won't matter anyways if you just use my renaming suggestion from the other comment

gregorbg (commented):
More comments :tada:

db/migrate/20240312122903_make_stripe_payment_intent_generic.rb
dunkOnIT: Call me crazy, I'm just calling it `type`. It will always be referenced (1) relative to its parent object (`payment_record.type`) or (2) in the scope of the model file (`self.type` or just `type`) so I think this is the least ambiguous way of doing it 

The only reason for `status` -> `stripe_status` is to differentiate it from `wca_status` in similar contexts

app/models/payment_intent.rb
dunkOnIT: Changed to `assign_wca_status!`

db/migrate/20240312122903_make_stripe_payment_intent_generic.rb
dunkOnIT: Noted!

db/migrate/20240312122903_make_stripe_payment_intent_generic.rb
dunkOnIT: Going with `stripe_record_type`

app/models/payment_intent.rb
dunkOnIT: Updated

dunkOnIT (commented):
Accidentally started a review while leaving comments

db/migrate/20240312122903_make_stripe_payment_intent_generic.rb
```diff
@@ -0,0 +1,40 @@
...
+    change_table :stripe_payment_intents do |t|
+      t.rename :user_id, :initiated_by_id # Make the purpose of the user_id column more clear
+      t.rename :confirmed_by_id, :confirmation_source_id
+      t.rename :confirmed_by_type, :confirmation_source_type
+      t.rename :canceled_by_id, :cancellation_source_id
+      t.rename :canceled_by_type, :cancellation_source_type
+      t.rename :stripe_record_id, :payment_record_id
+
+      t.string :wca_status
+      t.string :payment_record_type, after: :holder_id
```
gregorbg, line 14: The `payment_record_type` should definitely go after `payment_record_id`. These two belong together because they identify the polymorphic association `payment_record`.

db/migrate/20240312122903_make_stripe_payment_intent_generic.rb
```diff
@@ -0,0 +1,40 @@
...
+    rename_column :stripe_records, :status, :stripe_status
+    rename_column :stripe_records, :api_type, :stripe_record_type
+    rename_column :paypal_records, :status, :paypal_status
+
+    reversible do |direction|
+      direction.up do
+        PaymentIntent.update_all(payment_record_type: 'StripeRecord')
+
+        PaymentIntent.find_each do |intent|
+          intent.assign_attributes(payment_record_id: intent.payment_record_id)
```
gregorbg, line 27: This line doesn't make sense?!
It literally says "set the intent's `payment_record_id` to the intent's `payment_record_id`.

Possibly a remnant from an earlier state of the migration, now that we simply rename the column I think you can drop this altogether.

db/migrate/20240312122903_make_stripe_payment_intent_generic.rb
```diff
@@ -0,0 +1,40 @@
...
+          intent.assign_wca_status!
+          intent.save
+        end
+      end
+
+      direction.down do
+        PaymentIntent.find_each do |intent|
+          intent.update_attributes(payment_record_id: intent.payment_record_id)
```
gregorbg, line 35: Same as above, probably not needed at all.

db/schema.rb
```diff
@@ -845,9 +845,33 @@
...
+    t.string "confirmation_source_type"
+    t.bigint "confirmation_source_id"
+    t.datetime "canceled_at", precision: nil
+    t.string "cancellation_source_type"
+    t.bigint "cancellation_source_id"
+    t.string "wca_status"
+    t.index ["cancellation_source_type", "cancellation_source_id"], name: "index_stripe_payment_intents_on_canceled_by"
+    t.index ["confirmation_source_type", "confirmation_source_id"], name: "index_stripe_payment_intents_on_confirmed_by"
+    t.index ["holder_type", "holder_id"], name: "index_stripe_payment_intents_on_holder"
+    t.index ["initiated_by_id"], name: "fk_rails_2dbc373c0c"
```
gregorbg, line 868: Yikes, not a nice index name. How is this coming into being?

gregorbg: (I see it was there in the old diff as well. Feel free to ignore it if you cannot come up with a good and easy solution in 15 minutes)

db/schema.rb
```diff
@@ -1278,11 +1280,11 @@
   add_foreign_key "active_storage_attachments", "active_storage_blobs", column: "blob_id"
   add_foreign_key "active_storage_variant_records", "active_storage_blobs", column: "blob_id"
   add_foreign_key "oauth_openid_requests", "oauth_access_grants", column: "access_grant_id", on_delete: :cascade
+  add_foreign_key "payment_intents", "stripe_records", column: "payment_record_id"
```
gregorbg, line 1283: This foreign key should not exist anymore. `payment_record_id` could well also point to the `paypal_records` table.
The foreign key made sense when we had only Stripe and there was only one table to point to, but now with the polymorphic approach you have to deliberately delete the previously existing foreign_key in your migration.

gregorbg (changes requested):
Migration quirks

db/migrate/20240312122903_make_stripe_payment_intent_generic.rb
dunkOnIT: Doesn't the `type` field precede the `id` field on a polymorphic association? I used `competition_payment_integrations` and the `holder` field in `payment_intents` as a reference: 
image

image

db/migrate/20240312122903_make_stripe_payment_intent_generic.rb
dunkOnIT: Agreed!

db/migrate/20240312122903_make_stripe_payment_intent_generic.rb
dunkOnIT: Changed

db/schema.rb
dunkOnIT: Research suggests that Rails automatically creates these indexes when polymorphic associations are specified, but doesn't rename them automatically if a table name is changed (fair enough) - so I've added a `rename_index` line in the migration file

db/schema.rb
dunkOnIT: Apparently this foreign key would _not_ be automatically created, but I can't find where it was specified in any migration file.

In any case, I've removed it via the migration.

dunkOnIT:
Still need to make sure the migration works as expected on a populated db, just repopulating my local db because I'm too aggressive on `db:reset`

db/migrate/20240312122903_make_stripe_payment_intent_generic.rb
gregorbg: Okay, if that's the way Rails does things then fair enough. Then I assume `holder_id` is indeed the column that happens to be in front of `payment_record_id`?

app/models/payment_intent.rb
gregorbg: Is it though? In my most recent diff on GitHub, this still shows as `self.wca_status != 'created'`, but my point is you can just do `!self.created?`

app/models/stripe_record.rb
```diff
@@ -36,28 +49,33 @@ class StripeRecord < ApplicationRecord
...
+    case self.stripe_record_type
     when 'payment_intent'
       stripe_error = api_transaction.last_payment_error&.code
     when 'charge'
       stripe_error = api_transaction.failure_message
     end
 
-    self.update!(
-      status: api_transaction.status,
+    self.assign_attributes(
```
gregorbg, line 71: Be cautious here! Previously it was `update!` which triggers a DB write (and which is also intended to trigger a DB write here).
Your new `assign_attributes` does not trigger a DB write!

app/models/payment_intent.rb
```diff
@@ -0,0 +1,153 @@
...
+      else
+        self.update!(wca_status: payment_record.determine_wca_status)
+        save
+        payment_record.save
+      end
+    end
+  end
+
+  # Sets the wca_status field based on the status field of the associated PaymentRecord
+  def assign_wca_status
```
gregorbg, line 128: Where did the exclamation mark that I told you about in a previous review go?
`def assign_wca_status!`

lib/paypal_interface.rb
gregorbg, line 1: You are changing attribute names (i.e. column names) of `PaypalRecord` here even though that was never migrated and should be out of scope of this PR?!

app/models/payment_intent.rb
```diff
@@ -0,0 +1,153 @@
...
+        # The payment didn't need any additional actions and is completed!
+
+        # Record the success timestamp if not already done
+        unless self.succeeded?
+          self.assign_attributes(
+            confirmed_at: source_datetime,
+            confirmation_source: action_source,
+            wca_status: payment_record.determine_wca_status,
+          )
+          save
```
gregorbg, line 74: Calling `assign_attributes(foo: 123);save;` is literally the same as calling `update(foo: 123)`

app/models/payment_intent.rb
```diff
@@ -0,0 +1,153 @@
...
+          payment_record.save
```
gregorbg, line 75: Why?

app/models/payment_intent.rb
```diff
@@ -0,0 +1,153 @@
...
+        end
+      when 'canceled'
+        # Canceled by Stripe
+        self.update!(
+          canceled_at: source_datetime,
+          cancellation_source: action_source,
+          wca_status: payment_record.determine_wca_status,
+        )
+        save
+        payment_record.save
```
gregorbg, line 107: (Very probably) redundant
`update!` already fires the SQL call, and saving the associated `payment_record` seems unnecessary.

app/models/payment_intent.rb
```diff
@@ -0,0 +1,153 @@
...
+          canceled_at: nil,
+          cancellation_source: nil,
+          wca_status: payment_record.determine_wca_status,
+        )
+        save
+        payment_record.save
+      else
+        self.update!(wca_status: payment_record.determine_wca_status)
+        save
+        payment_record.save
```
gregorbg, line 122: See above.

gregorbg (changes requested):
Seems like you fell in love with `assign_attributes` a little bit too hard

lib/paypal_interface.rb
gregorbg: Nevermind I'm stupid.

db/migrate/20240312122903_make_stripe_payment_intent_generic.rb
dunkOnIT: Yep! You can see it in the 2nd screenshot, 3rd from the top

app/models/payment_intent.rb
dunkOnIT: Oof lol, fixed

app/models/payment_intent.rb
dunkOnIT: **Why am I calling `save` on `payment_record` if I'm already calling `save` on the `payment_intent`:** I _thought_ saving the PI would cause its related objects to be saved as well, but that's not the case - commenting out this line and running `bin/spring rspec spec/requests/registrations_controller.rb` causes the tests to fail because the updated status is not persisted to the db.

Not sure why - perhaps the polymorphic association? 

**Why save the payment_record here and not somewhere else:** We have to update the attributes of both the PI and payment_record before persisting it to the db, meaning either the PI or the payment_record will have a `save` not visible in their model code - it makes sense to me to have it happen in the model that's "pushing" the status change

dunkOnIT: (Even changing to `update!` doesn't solve it - I still have to save the payment_record separately)

app/models/payment_intent.rb
dunkOnIT: It is now actually updated :halo:

app/models/payment_intent.rb
dunkOnIT: See above

app/models/payment_intent.rb
dunkOnIT: Removed because we are no longer persisting the updated state to the db, and thus are no longer "perform[ing] some permanent or potentially dangerous change" (looked it up on [SO](https://stackoverflow.com/questions/7179016/what-is-the-purpose-of-and-at-the-end-of-method-names) to figure out if I should keep the `!` or not

dunkOnIT: Interesting tidbit from that thread btw: 
> One important note is that you should only have a bang method if you also have a corresponding non-bang method. The bang is used to distinguish the "more surprising" version of the method from the "less surprising" one. If you have only one method, then there's no need for a distinction, and you shouldn't name it with a bang. See Array#clear, for example. It clears the array. Clearing the array naturally mutates it. There's nothing surprising about that, the name already makes it clear, therefore: no bang. See [ruby-forum.com/topic/176830#773946](https://www.ruby-forum.com/topic/176830#773946)

app/models/payment_intent.rb
gregorbg: That last point is weird, `update!` most definitely fires an SQL call. Have you looked at https://github.com/thewca/worldcubeassociation.org/pull/9073#discussion_r1548026242?

gregorbg: Just staring at the code, I feel like reverting `def assign_status` to the old `def update_status` should definitely solve the problem (which is exactly what my other comment points to).

Have you specifically tried that? (Just double-checking because you did not reference my other comment)

gregorbg: If all else fails, you can also set `autosave: true` in the `belongs_to` association, but I would only consider this as a last resort.

app/models/stripe_record.rb
dunkOnIT: This was intentional - we can't update the status at this point, as we now have a validation on the confirmation/canceled_at as well. So we either have to
- also set confirmed/canceled at as part of this function
- delay the db write until confirmed/canceled_at are set in the calling function

I prefer the latter, as it doesn't feel appropriate/intuitive for a stripe_record method to be updating properties of the payment_intent object

app/models/stripe_record.rb
```diff
@@ -36,28 +48,33 @@ class StripeRecord < ApplicationRecord
...
+    case self.stripe_record_type
     when 'payment_intent'
       stripe_error = api_transaction.last_payment_error&.code
     when 'charge'
       stripe_error = api_transaction.failure_message
     end
 
-    self.update!(
-      status: api_transaction.status,
+    self.update(
```
gregorbg, line 70: Subtle but important: You're changing from `update!` to `update` (without the exclamation sign)
Was this on purpose?

db/migrate/20240312122903_make_stripe_payment_intent_generic.rb
```diff
@@ -0,0 +1,34 @@
...
+      t.rename :canceled_by_type, :cancellation_source_type
+      t.rename :stripe_record_id, :payment_record_id
+
+      t.string :wca_status
+      t.string :payment_record_type, after: :holder_id
+    end
+
+    rename_table :stripe_payment_intents, :payment_intents
+    # rename_index :payment_intents, 'index_stripe_payment_intents_on_holder', 'index_payment_intents_on_holder'
+    # remove_foreign_key :payment_intents, :stripe_records
```
gregorbg, line 19: Commented-out stuff... Why?
Either un-comment it or delete it altogether.

gregorbg (changes requested):

db/migrate/20240312122903_make_stripe_payment_intent_generic.rb
dunkOnIT: Fixed

PR #9074, opened 2024-03-12
Include V2 registrations in WCIF output

gregorbg:
This adds both reading and writing support for microservice registrations.

Note that writing support does not actually relay information to the microservice. It only stores information like the role strings in a temporary table, which is then also used to mock non-competing registrations.

Doing so is an intermediate solution until we have staff lanes in the microservice.

### DIFF
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class Assignment < ApplicationRecord
-  belongs_to :registration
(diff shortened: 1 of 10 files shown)


gregorbg:
@thewca-bot deploy staging

gregorbg:
@thewca-bot deploy staging

gregorbg:
@thewca-bot deploy staging

gregorbg:
@thewca-bot deploy staging

PR #9082, opened 2024-03-15
Prevent user being redirected to an API endpoint after login

dunkOnIT:
We could also add a check for the content type not being JSON, but the URL _seems_ like the most reliable method of determining whether a request is going to the API?


dunkOnIT:
@thewca-bot deploy-staging

dunkOnIT:
@thewca-bot deploy staging

danieljames-dj (approved):

gregorbg (commented):
Hm... Actually, the existing code already checks `!request.xhr?` which _should_ rule out these API calls. Not sure why Devise is storing them.

It's okay as a hotfix because it seems to affect users in production, but we definitely need to investigate this more.

PR #9090, opened 2024-03-19
WAC migration to user_groups

danieljames-dj:
This PR creates a new WAC group, but roles are yet to be migrated. The roles will be migrated in a later PR.

### DIFF
--- a/app/controllers/api/v0/user_roles_controller.rb
+++ b/app/controllers/api/v0/user_roles_controller.rb
@@ -93,7 +93,7 @@ class Api::V0::UserRolesController < Api::V0::ApiController
     roles = []
     group_type = group_id_of_old_system_to_group_type(group_id)
     original_group_id = group_id.split("_").last
-    if [UserGroup.group_types[:councils], UserGroup.group_types[:teams_committees]].include?(group_type)
+    if group_type == UserGroup.group_types[:teams_committees]
       TeamMember.where(team_id: original_group_id, end_date: nil).each do |team_member|
         roles << team_member.role
       end
@@ -187,26 +187,7 @@ class Api::V0::UserRolesController < Api::V0::ApiController
         roles.concat(User.delegates.map(&:delegate_role))
       end
     elsif group_type == UserGroup.group_types[:councils]
-      Team.all_councils.each do |council|
-        leader = council.leader
-        if leader.present?
-          roles << {
-            id: group_type + "_" + leader.id.to_s,
-            group: {
-              id: group_type + "_" + council.id.to_s,
-              name: council.name,
-              group_type: UserGroup.group_types[:councils],
-              is_hidden: false,
-              is_active: true,
-            },
-            user: leader.user,
-            metadata: {
-              status: 'leader',
-            },
-            is_active: true,
-          }
-        end
-      end
+      roles.concat(UserGroup.councils.flat_map(&:roles))
     elsif group_type == UserGroup.group_types[:board]
       roles.concat(UserGroup.board.flat_map(&:roles))
     end
@@ -291,6 +272,25 @@ class Api::V0::UserRolesController < Api::V0::ApiController
     end
   end
 
+  private def create_council_role(group, user_id, status)
+    team = group.team
+    return head :unauthorized unless current_user.has_permission?(:can_edit_groups, group.id)
+    if status == "leader"
+      old_leader = Team.find_by(id: team.id).leader
+      if old_leader.present?
+        old_leader.update!(end_date: Date.today)
+      end
+    end
+    already_existing_member = TeamMember.find_by(team_id: team.id, user_id: user_id, end_date: nil)
+    if already_existing_member.present?
+      already_existing_member.update!(end_date: Date.today)
+    end
+    TeamMember.create!(team_id: team.id, user_id: user_id, start_date: Date.today, team_leader: status == "leader", team_senior_member: status == "senior_member")
+    render json: {
+      success: true,
+    }
+  end
+
   def create
     user_id = params.require(:userId)
     group_id = params[:groupId] || UserGroup.find_by(group_type: params.require(:groupType)).id
@@ -317,6 +317,10 @@ class Api::V0::UserRolesController < Api::V0::ApiController
       location = params[:location]
     else
       location = nil
+    end
+
+    if group.group_type == UserGroup.group_types[:councils]
+      return create_council_role(group, user_id, status)
     end
 
     return render status: :unprocessable_entity, json: { error: "Invalid group type" } unless create_supported_groups.include?(group.group_type)
--- a/app/controllers/api/v0/user_roles_controller.rb
+++ b/app/controllers/api/v0/user_roles_controller.rb
@@ -276,14 +276,16 @@ class Api::V0::UserRolesController < Api::V0::ApiController
     team = group.team
     return head :unauthorized unless current_user.has_permission?(:can_edit_groups, group.id)
     if status == "leader"
+      # If the new role to be added is leader, we will be ending the leader role of already existing person.
       old_leader = Team.find_by(id: team.id).leader
       if old_leader.present?
         old_leader.update!(end_date: Date.today)
       end
     end
-    already_existing_member = TeamMember.find_by(team_id: team.id, user_id: user_id, end_date: nil)
-    if already_existing_member.present?
-      already_existing_member.update!(end_date: Date.today)
+    # If the person who is going to get the new role is already having a role, that role will be ended.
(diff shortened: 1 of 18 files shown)


gregorbg:
Is there any danger in merging and deploying this PR without the follow-up PR for the Roles?

app/models/user_role.rb
```diff
@@ -124,6 +124,18 @@ def self.is_eligible_voter?(role)
     end
   end
 
+  def self.discourse_user_group(role)
+    is_actual_role = role.is_a?(UserRole) # Eventually, all roles will be migrated to the new system, till then some roles will actually be hashes.
+    group_type = is_actual_role ? role.group.group_type : role[:group][:group_type]
+    group = is_actual_role ? role.group : role[:group]
```
gregorbg, line 130: Shouldn't this use the `UserGroup.group_type(foo)` helper methods?

db/seeds/groups_metadata_councils.seeds.rb
gregorbg, line 1: Why do you need an empty seeds file?

(The simple answer is probably because tests complain if `test_db_manager` defines a file that has no seeds attached, but that's not what I'm asking for. I'm asking why are you adding it to `test_db_manager` in the first place if you're not creating any data?)

db/migrate/20240318171344_migrate_councils_to_user_groups.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class MigrateCouncilsToUserGroups < ActiveRecord::Migration[7.1]
+  def change
```
gregorbg, line 4: The name of the migration implies you're migrating all councils. In reality, you're just migrating one specific council by a hard-coded ID.

This may "accidentally" work because we know as of 2024-03-25 there is only one council, but shouldn't there be some `UserGroup.all_councils` constant or similar, that you can loop over?

app/controllers/users_controller.rb
```diff
@@ -221,7 +225,8 @@ def sso_discourse
     all_groups = User.all_discourse_groups
 
     # Get the teams/councils/Delegate status for user
-    user_groups = current_user.current_teams.select(&:official_or_council?).map(&:friendly_id)
+    user_groups = current_user.current_teams.select(&:official?).map(&:friendly_id)
+    user_groups.concat(current_user.active_roles.map { |role| UserRole.discourse_user_group(role) }.compact)
```
gregorbg, line 229: `concat` changes the array that it's being called on under the hood, even though this is breaking Ruby's own naming conventions (it should be `foo.concat!` to make clear that `foo` is being modified in-place).

I would be happier with something like `user_groups +=` or `user_groups |=` instead.

gregorbg (changes requested):

app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -297,6 +280,25 @@ def index_for_group_type
...
+    team = group.team
+    return head :unauthorized unless current_user.has_permission?(:can_edit_groups, group.id)
+    if status == "leader"
+      old_leader = Team.find_by(id: team.id).leader
+      if old_leader.present?
+        old_leader.update!(end_date: Date.today)
+      end
+    end
+    already_existing_member = TeamMember.find_by(team_id: team.id, user_id: user_id, end_date: nil)
+    if already_existing_member.present?
```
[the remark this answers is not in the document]
danieljames-dj: Not really. I think the naming caused a confusion here.

`old_leader`: The person who was having the leader role earlier. Not the person whose role is going to be changed.
`already_existing_member`: This is the `TeamMember` row of the person whose role is going to be changed. Since for the new role (irrespective of whether a leader or not) we need to end the existing role, this is done.

I understand that `already_existing_member` name looks confusing. I've renamed it to `already_existing_row`, but couldn't find a better name. I've also added couple of comments to make it more clear.

app/controllers/users_controller.rb
danieljames-dj: Okay, changed to `+=`

app/models/team.rb
```diff
@@ -49,20 +49,6 @@ def self.all_official
     ]
   end
 
-  # Councils are recognized by Motion "20.2019.0":
-  #  https://documents.worldcubeassociation.org/documents/motions/20.2019.0%20-%20Councils.pdf
-  # Motions starting with "20.YYYY.N" define these councils:
-  #  https://www.worldcubeassociation.org/documents
```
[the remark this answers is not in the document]
danieljames-dj: Do we really need to place the motion link in our code? The main reason why I don't like to have it is because we don't have a permalink for any motions. For example, the link mentioned in the comment no longer exist. If we add the latest link, if the motion is modified again in next cycle, we again have to change it, which is not so easy to track.

app/models/user_role.rb
danieljames-dj: Oh yes, I think I copied this from somewhere and hence missed it. Changed it now, thanks.

db/migrate/20240318171344_migrate_councils_to_user_groups.rb
danieljames-dj: There was `Team.all_councils`, but that was removed in this PR because it's no longer used. Also, even inside `all_councils`, the list was hard-coded. So I thought not using that will be the easiest way.

Also, even if we bring back `all_councils` in this PR, we might remove it after few PRs. And if somebody is running pending migration, there might occur a case where `all_councils` method doesn't exist.

db/seeds/groups_metadata_councils.seeds.rb
danieljames-dj: (Same reason why I created `groups_metadata_board.seeds.rb`)
I'm actually creating the data, but in `user_groups`. I cannot create them separately because it's difficult to connect them. Also, I was not able to skip adding it in `test_db_manager`, because it will throw error saying I didn't add `groups_metadata_board` to the `test_db_manager`.

danieljames-dj (commented):
I hope I addressed all the comments.

db/migrate/20240318171344_migrate_councils_to_user_groups.rb
gregorbg: Hard-coding the list is okay if there's a good reason for it. And formal definitions in our Motions (which are outside of the realm of code) are a good reason for hard-coding stuff. Like "We have a Motion that defines these 3 councils currently exist". That's a bureaucratic definition that you just have to hard-code.

Your argument about running migrations much later also applies to your current approach of `Team.c_find_by_id!`. What if that method doesn't exist anymore in the feature because two months from now, your project is done and the `Team` entity won't exist in our code anymore?

gregorbg: (My point is: All migrations have inherent weaknesses.)

db/seeds/groups_metadata_councils.seeds.rb
gregorbg: Seeds have special macros to define the order in which they are executed. See the seeds for "eligible championship types" which uses an `after` macro and can thus rely on entites in the `Country` table already being present.

app/models/team.rb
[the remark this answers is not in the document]
gregorbg: I'm mostly talking about a reference to the Motions, where the notion of "Motion 20.2019.0 exists" is definitely permanent. Whether or not to include an actual URL is up for debate, but in my opinion it is "stable enough" to warrant inclusion. If you feel really bad about the URL, skip it. But at least include information about the Motion itself.

app/controllers/api/v0/user_roles_controller.rb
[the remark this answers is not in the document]
gregorbg: Okay, thanks for the additional explanation :) Makes total sense now!

db/migrate/20240318171344_migrate_councils_to_user_groups.rb
danieljames-dj: Okay makes sense. I've changed it to a list with just WAC in the array. Hope this is fine now, let me know if I need to re-introduce Team.all_councils method.

db/seeds/groups_metadata_councils.seeds.rb
danieljames-dj: Oh I didn't knew it. Thanks for that, changed it.
PS: I've changed it for board as well in this PR, hope this is fine.

app/models/team.rb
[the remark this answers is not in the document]
danieljames-dj: Okay, I've added it in user_group.rb now, because I couldn't find a better place. Please let me know if there is a better place. Also please let me know if the text has to be changed.

PR #9098, opened 2024-03-21
Fix error when trying to create a Continental Championship

### LINKED ISSUE #9097: Unable to add Championship Type to Competition
**Describe the bug**
Upon adding a championship type to an unannounced competition it not longer becomes possible to save the competition. I have tried in both the organizer and admin view

**To Reproduce**
1. Go to https://www.worldcubeassociation.org/competitions/OC2024/edit
2. Click on Add championship type "Oceania"
3. Click on "Save your changes" 

**Expected behavior**
Should be able to save the comp after adding a championship type

dunkOnIT:
Fixes #9097 

Validate Continental Championships on continent **name** not id (presumably we used to match on id because we were still using a Rails frontend - now it makes more sense to use name with a separate frontend/backend I think)

### DIFF
--- a/app/models/championship.rb
+++ b/app/models/championship.rb
@@ -62,7 +62,7 @@ class Championship < ApplicationRecord
   def self.grouped_championship_types
     {
       "planetary" => [CHAMPIONSHIP_TYPE_WORLD],
-      "continental" => Continent.all_sorted_by(I18n.locale, real: true).map(&:name),
+      "continental" => Continent.all_sorted_by(I18n.locale, real: true).map(&:id),
       "multi-national" => EligibleCountryIso2ForChampionship.championship_types,
       "national" => Country.all_sorted_by(I18n.locale, real: true).map(&:iso2),
     }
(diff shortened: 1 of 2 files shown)


dunkOnIT:
Ok the remaining test failure is a doozy. This PR uses continent name instead of id, but championship_podiums are still calculated using the id: 

`results.joins(:country, competition: [:championships]).where("championships.championship_type = Countries.continentId")`

Will wait for feedback on if this is the right approach before continuing. 

The solution I see here is to perhaps `join` the continent_name to the country table beforehand?

gregorbg:
To cut the chase short (although I'm sorry to say this) switching the backend to verify by name is not the right approach here.

IDs are specifically designed for the purpose of exchanging and **identifying** exchanged information. The clean solution here is to leave the backend completely as-is, roll back all of your changes, and make the frontend actually send the proper values in the AJAX requests.

dunkOnIT:
Ok this is _a_ solution that gets the frontend sending the right data.
- `grouped_championship types` is only used in the context of this code so I'm fine with changing its return value
- `generateChampionshipName` I'm brute-force removing the underscores from the string to generate the name of the continent. This might be a little hacky - the alternative I see is to explicitly define/export a mapping between continent.id and continent.name - if the data were more dynamic I'd be much more partial to this, but for continents it seems like there's a decent chance this doesn't change within the decade at least

gregorbg (approved):
Thanks!

PR #9103, opened 2024-03-23
Remove Deprecated Delegate edit page

danieljames-dj:
This PR is ready for review. But needs to be informed to Senior Delegate before merged. I'll undraft this PR once I inform Senior Delegate. (Only blocker for informing Senior Delegate is https://github.com/thewca/worldcubeassociation.org/pull/9095)


spec/controllers/api/v0/user_roles_controller_spec.rb
gregorbg, line 1: As always, I am very unhappy about tests being removed :(
Why did you decide to do this and why are these tests not necessary anymore?

gregorbg (changes requested):

spec/controllers/api/v0/user_roles_controller_spec.rb
danieljames-dj: Reason why these tests are removed is the corresponding code is also removed. There are no tests for the new flow of the actions because the action methods are currently bit complicated. Once the migration is completed, it will be very easy to add the test cases I believe.

danieljames-dj: PS: The 'new flow' I mentioned here is not added in this PR, it was already there. This PR is just removing some of the existing code which is no longer used.

gregorbg (approved):

gregorbg (approved):
FOrmally approved, feel free to merge this after March 31 at your own discretion

PR #9107, opened 2024-03-23
Added Contact & Navagation to 404.html, 422.html, and 500.html

jamescraft10:
I added a button to these pages and I added contact to 500.html.


FinnIckler:
These HTML files are not used. The error pages are [here](https://github.com/thewca/worldcubeassociation.org/blob/39d1799e1da0803d6b47b44623f11d8786e3f8d4/app/views/errors/500.html.erb#L4)

PR #9112, opened 2024-03-25
Bump Shakapacker to 7.2.3

gregorbg:
So we can have the webpack-dev-server 5.0 upgrades which is relevant for security fixes.

Basically this is a glorifed Dependabot PR, except that Shakapacker Ruby+JS needs to be updated at the same time and Dependabot cannot do this automatically.


gregorbg:
Merging without review because this is just a glorified Dependabot update

PR #9116, opened 2024-03-26
Migrate Board roles to user_roles

danieljames-dj:


gregorbg:
Code looks okay but you're getting actual test failures.
Please fix them before I submit a formal approval.

app/models/user.rb
```diff
@@ -1297,7 +1297,8 @@ def is_special_account?
       any_kind_of_delegate? ||
       !delegated_competitions.empty? ||
       !competitions_announced.empty? ||
-      !competitions_results_posted.empty?
+      !competitions_results_posted.empty? ||
+      board_member?
```
gregorbg, line 1301: Why is this change necessary? Feels like you're touching a method that should not be affected by your migration

app/webpacker/components/Panel/Board/BoardEditorPage/BoardEditor.jsx
```diff
@@ -0,0 +1,88 @@
...
+          ))}
+        </Table.Body>
+      </Table>
+      {openModal && (
+      <Modal
+        size="fullscreen"
+        open={openModal}
+        onClose={() => {
+          setOpenModal(false);
+          newBoardRole(null);
```
gregorbg, line 56: Shouldn't this be `setNewBoardRole`? In that case, you can extract it into its own function and use that new function here and also in the `onSubmit` success fn below.

db/migrate/20240326023855_create_board_roles.rb
```diff
@@ -0,0 +1,19 @@
...
+      user_id: team_member.user_id,
+      group_id: group.id,
+      start_date: team_member.start_date,
+      end_date: team_member.end_date,
+    )
+  end
+
+  def change
+    board_group = UserGroup.board_group
+    Team.c_find_by_friendly_id!('board').team_members.each do |member|
```
gregorbg, line 15: Shouldn't `Team.board` do the trick?

gregorbg (changes requested):

app/models/user.rb
danieljames-dj: Actually board member is also considered as special account according to unit test. Earlier, this was covered in `self.teams.any`, but now since board is out of teams, this is necessary..

app/webpacker/components/Panel/Board/BoardEditorPage/BoardEditor.jsx
danieljames-dj: Oh yes, done. Thanks.

db/migrate/20240326023855_create_board_roles.rb
danieljames-dj: Actually I was thinking to remove `self.board` from `team.rb` as it is no longer used, but I missed to remove it. I've removed it from `team.rb` now.

app/models/user.rb
gregorbg: Aha! Makes sense, thanks for the explanation.

gregorbg (approved):

PR #9118, opened 2024-03-26
Attempt to fix sync mail job failure

danieljames-dj:


gregorbg:
Ironically, this fails with the very error you're trying to fix

gregorbg (approved):

PR #9119, opened 2024-03-26
Stopped showing officers in teams edit page

danieljames-dj:


gregorbg:
I think the "cleaner" solution in terms of data would be to just remove the corresponding teams from the database.
If you do, the current implementation won't pick them up because they simply don't exist anymore.

What harm would there be in such a no-code solution?

danieljames-dj:
Oh yes, that is what we discussed over slack as well. I think I lost my mind. Closing this PR, I'll do it manually.

PR #9121, opened 2024-03-26
Migrate Teams & Committees to groups

danieljames-dj:


spec/jobs/sync_mailing_lists_job_spec.rb
gregorbg, line 1: You may have guessed this, but I am extremely unhappy about these tests getting removed.
Email ocmmunication is one of the central backbones of the whole WCA, and in the past when some emails went missing because of DKIM changes, all sorts of hell were breaking loose.
We definitely need these tests, so why did you delete them without replacement?

danieljames-dj: For delegate regions, I didn't test the sync of all delegate regions, instead created some dummy regions and tested with them. I thought of doing the same thing here - instead of creating all teams-committees manually, I thought just doing for around 5 should be enough.

app/models/team.rb
```diff
@@ -158,8 +158,17 @@ def name
...
+  end
+
   def group
-    council_if_any
+    council = council_if_any
+    if council.present?
+      council
+    else
+      team_if_any
+    end
```
gregorbg, line 171: This should simply be `council_if_any || team_if_any` as a one-liner

app/models/team.rb
```diff
@@ -158,8 +158,17 @@ def name
     UserGroup.find_by(metadata_id: GroupsMetadataCouncils.find_by(friendly_id: self.friendly_id)&.id, metadata_type: 'GroupsMetadataCouncils')
   end
 
+  private def team_if_any
+    UserGroup.find_by(metadata_id: GroupsMetadataTeamsCommittees.find_by(friendly_id: self.friendly_id)&.id, metadata_type: 'GroupsMetadataTeamsCommittees')
```
gregorbg, line 162: Can you lookup the metadata entity and then just have a Rails relation? Like `has_one` in the metadata... This query is definitely too complicated!

spec/jobs/sync_mailing_lists_job_spec.rb
gregorbg: I understand your logic and you have a point if this was a random, unlimited collection of data.

However, since we have a limited set of teams and there were already previously existing tests for those teams, I would prefer if you don't remove them.

Again, I would normally accept your argument for data collections that can have an arbitrary number of entries. However, the number of T/Cs in the WCA is not arbitrary and also very, very important, so in this case I'd rather keep the tests.

gregorbg (changes requested):

app/models/team.rb
danieljames-dj: Oh yes thanks.

spec/jobs/sync_mailing_lists_job_spec.rb
danieljames-dj: Okay, thanks for your detailed explanation. I've added them back.

app/models/team.rb
danieljames-dj: Oh I didn't think of that, did now. This is more easier. Did now. I might be able to simplify it further by creating a relation between team and groups_metadata_teams_committees through friendly_id, but didn't do because I anyway have to remove this after few PRs.

gregorbg (approved):

PR #9122, opened 2024-03-26
View history of roles in teams-committees-councils page

danieljames-dj:
This change is currently visible only to staffs, not to public.


gregorbg (approved):

app/models/groups_metadata_teams_committees.rb
gregorbg, line 1: Whoopsie?

danieljames-dj: Oops removed.

PR #9124, opened 2024-03-26
Move eligible_voters to user_roles_controller

danieljames-dj:


danieljames-dj:
Ready for review, but https://github.com/thewca/worldcubeassociation.org/pull/9121 needs to be merged before this so marking this PR as draft.

danieljames-dj:
The above mentioned PR is merged now, so ready for review.

app/controllers/admin_controller.rb
```diff
@@ -382,7 +382,7 @@ def do_override_regional_records
   end
 
   def all_voters
-    voters User.eligible_voters, "all-wca-voters"
+    voters Api::V0::UserRolesController.eligible_voters, "all-wca-voters"
```
gregorbg, line 385: I am not happy about calling controllers from other controllers.
Is there a reasonable place in `app/models` or in `lib` where you can store this code?

gregorbg (changes requested):

app/controllers/admin_controller.rb
danieljames-dj: Maybe I'll have it in `user` model itself for now. I was thinking to make it as an API for the react migration, but will do that later.

gregorbg (approved):

PR #9125, opened 2024-03-26
Migrate council to user_roles

danieljames-dj:


app/models/groups_metadata_teams_committees.rb
gregorbg, line 1: Whoopsie?

app/models/team.rb
```diff
@@ -154,18 +150,12 @@ def name
...
   end
 
-  private def council_if_any
-    UserGroup.find_by(metadata_id: GroupsMetadataCouncils.find_by(friendly_id: self.friendly_id)&.id, metadata_type: 'GroupsMetadataCouncils')
-  end
-
   def group
-    council_if_any
+    nil
   end
```
gregorbg, line 155: Not cool. Why do you need a method that has no other purpose except explicitly returning `nil`?!

gregorbg (changes requested):

app/models/groups_metadata_teams_committees.rb
danieljames-dj: Sorry again :-( Removed

app/models/team.rb
danieljames-dj: This will be replaced by the group for teams-committees.
In team_member.rb, `def role` will check if a group is available, and if it's not available, will create a group-hash.

I had two options:
1. Remove this method and remove the reference from `def role` in `team_member.rb`. Later add them back in teams-group migration (https://github.com/thewca/worldcubeassociation.org/pull/9121)
2. Make this method return nil, so that I don't have to touch `def role` in teams-group migration. In teams-group migration I'll add back the body.

I opted for 2 as the changes is slightly minimal.

danieljames-dj: Now since the other PR is merged, this should be fine now I guess.

db/migrate/20240326194731_migrate_wac_roles.rb
```diff
@@ -0,0 +1,29 @@
...
+      group_id: group.id,
+      start_date: team_member.start_date,
+      end_date: team_member.end_date,
+      metadata: metadata,
+    )
+  end
+
+  def change
+    UserGroup.councils.each do |council|
+      Team.find_by(friendly_id: council.metadata.friendly_id).team_members.each do |member|
```
gregorbg, line 24: Attention: `Team` has a "default scope" defined. It means that every query you do will have `WHERE hidden = 0` attached, **even when you do stuff like `Team.all`**.

In this case, it will be fine because we don't have a "hidden council". But when you migrate T/C memberships in the future, you probably also want to migrate WCT China and WST Admin, in which case you need to do `Team.with_hidden.find_by`

gregorbg (approved):

db/migrate/20240326194731_migrate_wac_roles.rb
danieljames-dj: Got it, thanks.

PR #9128, opened 2024-03-28
Use UtcDatePicker throughout Rails code

gregorbg:
See title. Hopefully gets rid of timezone offset errors, because while `react-datepicker` is really nice in terms of usability, it insists on _local_ JS dates. The common component should avoid the repeating `moment` idioms.

Bonus: Removes `moment` usages throughout the rest of the code as well.

### DIFF
--- a/app/webpacker/components/CompetitionsOverview/filterUtils.js
+++ b/app/webpacker/components/CompetitionsOverview/filterUtils.js
@@ -1,4 +1,3 @@
-import { DateTime } from 'luxon';
 import { events } from '../../lib/wca-data.js.erb';
 
 // note: inconsistencies with previous search params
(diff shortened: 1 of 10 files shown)


danieljames-dj (approved):

app/webpacker/components/CompetitionsOverview/filterUtils.js
```diff
@@ -89,13 +82,13 @@ export const updateSearchParams = (searchParams, filterState, displayMode) => {
 
   // for date values, format and add them if applicable, otherwise omit them
```
kr-matthews, line 83: comment needs updating since the code no longer formats

gregorbg: Well, it kinda does but just in a different place :smile: (The `UtcDatePicker` component internally does all the formatting)

Good catch!

PR #9131, opened 2024-03-28
Use roles in teams object of me API

danieljames-dj:


spec/controllers/api_controller_spec.rb
```diff
@@ -282,7 +282,7 @@
         expect(team['leader']).to eq false
         expect(team['friendly_id']).to eq 'wrc'
         expect(team['avatar']['thumb']['url']).to be_a String
-        expect(team['id']).to be_a Numeric
+        expect(team['id']).to be_a String
```
gregorbg, line 285: It _should_ still be numeric, right?
I mean, the primary key ID of a role object is still an integer in the end...

danieljames-dj: Since in this context the user is signing in as a team-member, the team's ID will be the ID of made-up group, which will be of something like `teams_committees_123`. But once the teams are migrated to groups, this test-case will start failing and that time will change it back to Numeric.

danieljames-dj: There is a small correction. team['id'] was team_member ID earlier, and now it's role ID. Since there are no role ID for teams-committees yet, this will be string. But once I raise the PR for roles migration of teams-committees, this will be changed back to Numeric.

app/models/user.rb
```diff
@@ -1162,7 +1175,7 @@ def serializable_hash(options = nil)
     # scope at the moment).
     json[:class] = self.class.to_s.downcase
     if include_teams
-      json[:teams] = current_team_members.includes(:team).reject(&:hidden?)
+      json[:teams] = deprecated_team_roles
```
gregorbg, line 1178: Does this change take into account the `hidden` filter that was present in the old implementation?

danieljames-dj: Actually no. I missed it, thanks for pointing. Did it now.

gregorbg (approved):

PR #9134, opened 2024-03-29
A few unit tests for user_groups_controller

danieljames-dj:
Added some unit tests for user_groups along with some code cleaning.
(Currently there is no 100% coverage for user_groups_controller, will add that eventually in some future PR(s))


app/models/groups_metadata_delegate_regions.rb
```diff
@@ -1,4 +1,9 @@
 # frozen_string_literal: true
 
 class GroupsMetadataDelegateRegions < ApplicationRecord
+  has_one :user_group, -> { where(metadata_type: 'GroupsMetadataDelegateRegions') }, class_name: 'UserGroup', foreign_key: 'metadata_id'
```
gregorbg, line 4: This is not how polymorphic associations work in Rails.
There should be a `belongs_to` in UserGroup, use the name of that association with the `as:` keyword here.

Further reading: https://guides.rubyonrails.org/association_basics.html#polymorphic-associations

app/models/user.rb
```diff
@@ -637,6 +637,10 @@ def banned_at_date?(date)
       end
     end
 
+    if can_manage_delegate_probation?
+      groups += UserGroup.delegate_probation.map(&:id)
```
gregorbg, line 641: I think `UserGroup.delegate_probation.ids` should work (if `delegate_probation` is a Rails scope and not a custom method returning an array)

db/seeds/user_groups.seeds.rb
```diff
@@ -1,6 +1,114 @@
...
+  UserGroup.create!(
+    name: 'South America',
+    group_type: :delegate_regions,
+    is_active: false,
+    is_hidden: false,
+    metadata: GroupsMetadataDelegateRegions.find_by!(friendly_id: 'south-america'),
+  )
+  UserGroup.create!(
+    name: 'Asia East',
+    parent_group_id: GroupsMetadataDelegateRegions.find_by!(friendly_id: 'asia').user_group.id,
```
gregorbg, line 56: `create` has a return value: It returns the entity you just created. So when calling the create statement for `asia` above, you can store that in a variable and use it here, instead of re-reading the thing you literally just wrote to the DB a few seconds ago

gregorbg: (same logic applies to all other child groups below as well)

gregorbg (changes requested):

app/models/groups_metadata_delegate_regions.rb
danieljames-dj: Oh that's shorter. Thanks. Changed in groups_metadata_teams_committees.rb as well in this PR, hope it's fine.

app/models/user.rb
danieljames-dj: Thanks, didn't knew about this. :-)

db/seeds/user_groups.seeds.rb
danieljames-dj: Done

gregorbg (approved):

PR #9184, opened 2024-04-01
Migrate the persons page to React

simonkellly:


simonkellly:
I think the best way to split this migration up is likely to do: 
- all of the relatively static tables in one pr
- big list of all results in another pr
- map in a 3rd PR (unless its trivial, but I have not investigated that yet)
- Final pr to turn on the changes for everyone

So this will be first one of those

simonkellly:
I have finished migrating the static parts (plus map).

What is left is the results for event table, and tying it together (as mentioned above) which will be a seperate pr.

There is some logic in the view files such as in the `_details.html.erb` file which handles the serialization for the react component. I think leaving this here is the best option for now, as it is likely parts of the data needed will be reused across different sections of the page, which will eventually all be rendered under one react component, but I am open to feedback on this!

simonkellly:
In case anyone wants to see different features, I have been testing with various different people namely:

2012BEAH01 - Lots of records, but still some stuff on the profile missing.
2012PATE01 - Has some PRs which were from a previous country
2015YAKH01 - Has multiple previous countries

gregorbg:
Can I haz screenshot plz? :D

(I am thrilled to see the progress but I can't boot my computer right now)

gregorbg:
Also, why did you decide to glue smaller React components into all the different legacy views (patchwork-style) instead of one top-level "If flag, then show THE react page (which may be incomplete and still missing some sections entirely) or else show legacy"?

I don't want to sound very strict or dictate your programming style, so if you feel comfortable with your current approach then okay. My thinking is just that by replacing bit by bit with smaller components (your current approach) limits you too much in terms of layout and design. It ties you to the exact layout of the legacy page.

My thinking is that having one global React component that is still "work-in-progress" (compared to the old legacy page) gives you more degrees of freedom, like a "blank canvas" to freely think about the best approach for arranging the individual sub-sections

simonkellly:
I didn't really think about having it in one big page with missing parts (sorry if I misunderstood what you meant in your message last week), but happy to do so.

In terms of the layout, its almost the exact same as before but with semantic ui. I don't have any ideas specifically about ways it could be made different, but its basically the same work to make those changes down the line vs now. If you have any suggestions in this department I would definitly like to hear them.

image

gregorbg:
> I didn't really think about having it in one big page with missing parts (sorry if I misunderstood what you meant in your message last week), but happy to do so.

I think the most important point is to find an approach that works well for you. Don't sink several hours into this refactor if it's inconvenient for you to work with and slows down your progress significantly. In my opinion, having one big page has the following advantages:
1. It more closely mirrors the "end result", that is when we fully replace the page it will be one central entry point to one (then fully migrated) React component.
2. It prevents you from having to pass small chunks of data around here and there. You can already start thinking about efficiently "bundling" the data, and avoid redundantly passing the same bits of data around to all the smaller components again and again
3. (Arguably the biggest advantage) It removes the restraints of the existing layout, giving you more degrees of freedom to play with the layout.

> In terms of the layout, its almost the exact same as before but with semantic ui. I don't have any ideas specifically about ways it could be made different, but its basically the same work to make those changes down the line vs now. If you have any suggestions in this department I would definitly like to hear them.

I don't have any specific directions or reference material that I would like you to implement. But there are *so* many small ideas that immediately come to my mind.
- "Medal Collection" and "Record Collection" can use icons and colors. Like rendering the counts actually in gold/sivler/bronze font color, or a "winner's podium" or "winner trophy" icon next to the championship podium.
- Records can be made a "collapsible" extension of the "Record collection" block
- Championship Podiums can be made a "collapsible" extension of the "Medal collection" block
- The Region/ WCAID/ Gender/ Competitions/ Solves statistics can be put into a vertical list next to the profile picture. Play around with https://react.semantic-ui.com/views/card/ just for the fun of it
- Do we want to more prominently place the map??

As previously stated, none of these are requirements or binding directives. They are really just random ideas and my point is to "free your mind" from the restraints of the previous layout (which I think is the easiest when you have one big work-in-progress React component that you can freely play around with). Are you fully happy with the page layout of the current page? Is there really nothing you had always wished would be laid out differently / more convenient? Now is the chance to change it!

app/webpacker/components/Persons/RegionalRecords.jsx
```diff
@@ -0,0 +1,111 @@
+import React from 'react';
+import {
+  Table, TableBody, TableCell, TableHeader, TableHeaderCell, TableRow,
+} from 'semantic-ui-react';
+import I18nHTMLTranslate from '../I18nHTMLTranslate';
+import EventIcon from '../wca/EventIcon';
+import { events } from '../../lib/wca-data.js.erb';
+import { AttemptItem } from './TableComponents';
+
+function groupByEvent(results) {
```
kr-matthews, line 10: lodash has a `groupBy` function; it's used in `OfficersAndBoard` for example.

app/webpacker/components/Persons/RegionalRecords.jsx
```diff
@@ -0,0 +1,111 @@
...
+      ))}
+    </>
+  );
+}
+
+const allEvents = events.official.map((event) => event.id);
+Object.entries(events.byId).forEach((entry) => {
+  if (allEvents.find((e) => e === entry[1].id)) return;
+  allEvents.push(entry[1].id);
+});
```
kr-matthews, line 62: It should be possible to do this with filters/maps, which I think would be better than modifying `const` arrays.

simonkellly:
image

I am looking at this kind of direction for the top of the page to replace the current image and single table row.

This is a _very_ rough draft but just wanted to put it here anyway (It is taking a while for me to get to grips with how semantic-ui handles grids and layouts etc.)

simonkellly:
image

FinnIckler:
I'm taking over the PR a little to prepare the page for the slate redesign. I took your components, but re-arranging them according to the slate design. Still need to figure out all the margins, but this is how it looks:
Screenshot 2025-01-01 at 18 13 55

FinnIckler:
image
Made the tables more basic to not have any double boxes

FinnIckler:
mobile: 
image
image

FinnIckler:
Added scroll stickiness for computer and tablet
image

FinnIckler:
image
Moved the medals and records over to statistics as well

FinnIckler:
With adding the highlighting of podiums I think we've reached feature parity

kr-matthews:
Is this ready for review?

gregorbg:
> Is this ready for review?

Frontend, yes. I'm still tinkering with the backend data binding a lot, and I have started glancing at the frontend. But don't let this keep you from reviewing!

Be warned that there are still a lot of "weird things" happening, like manually constructing throw-away lookup dictionaries or looping inside of a map inside of a loop inside of a group-by. I'd love to have your input on this, and if at any point you think that "Hm, if the backend delivered the data in this-and-that format instead, the code here could be a ton more efficient" please leave a comment and I will wave my magic backend wand to make it happen.

app/webpacker/components/Persons/Person/index.jsx
```diff
@@ -0,0 +1,177 @@
...
+  averageRanks,
+  singleRanks,
+  competitions,
+  medals,
+  records,
+  pbMarkers,
+  championshipPodiums,
+  canEditUser,
+}) {
+  const [highlight, setHighlight] = useState(-1);
```
kr-matthews, line 115: Not sure what this is just from looking at this file, is there a more descriptive name?

app/webpacker/components/Persons/Person/Details.jsx
```diff
@@ -0,0 +1,71 @@
...
+      {person.user?.avatar && (
+        <Card image={person.user.avatar.url} centered raised />
+      )}
+      <Header as="h2" textAlign="center">
+        <CountryFlag iso2={person.country_iso2} />
+        {' '}
+        {person.name}
+        {canEditUser && (
+          <a href={editPersonUrl(person.user.id)}>
+            {' '}
```
kr-matthews, line 42: The space should be outside the anchor.

app/webpacker/components/Persons/Person/PersonalRecords.jsx
```diff
@@ -0,0 +1,220 @@
...
+  Header, Icon, Popup, Table,
+} from 'semantic-ui-react';
+import I18nHTMLTranslate from '../../I18nHTMLTranslate';
+import I18n from '../../../lib/i18n';
+import { events } from '../../../lib/wca-data.js.erb';
+import EventIcon from '../../wca/EventIcon';
+import { rankingsPath } from '../../../lib/requests/routes.js.erb';
+import { formatAttemptResult } from '../../../lib/wca-live/attempts';
+
+function isOddRank(rank) {
```
kr-matthews, line 12: Maybe `isCounterintuitiveRank` or `rankNeedsExplanation` or something? Since odd means something else already for numbers.

app/webpacker/components/Persons/Badges.jsx
kr-matthews, line 1: It looks like the badges are the only piece of data we're actively fetching _within_ the profile? Which has the side effect of them potentially showing up a little late, causing the UI to jump. Can the badge data be provided directly along with all the other data?

kr-matthews (commented):
I haven't reviewed everything yet, nor run locally. Will continue later.

I think I mentioned this elsewhere recently, but I'm in favour of having the main/exported component at the top of the file, with the helper components below. It's the most important thing and should be first. I'm 99% sure this is fine and there are no 'that function is not defined' errors.

app/webpacker/components/wca/EventSelector.jsx
```diff
@@ -22,26 +22,29 @@ export function EventSelector({
   // Listing event as an argument here to indicate to developers that it's needed
   // eslint-disable-next-line no-unused-vars
   disabledText = (event) => {},
+  showLabels = false,
```
kr-matthews, line 25: Defaulting this to false means all existing uses no longer show the title and buttons.

app/webpacker/components/Persons/Person/TableComponents.js
kr-matthews, line 1: Will there be other table components? If not, the file name and component name should match.

kr-matthews: This seems like it could be reused in rankings and records?

app/webpacker/components/Persons/Person/Results.jsx
```diff
@@ -0,0 +1,131 @@
+import React, { useMemo, useState } from 'react';
+import { Divider, Table } from 'semantic-ui-react';
+import _ from 'lodash';
+import { events, roundTypes } from '../../../lib/wca-data.js.erb';
+import { EventSelector } from '../../wca/EventSelector';
+import { competitionUrl } from '../../../lib/requests/routes.js.erb';
+import I18n from '../../../lib/i18n';
+import { formatAttemptResult } from '../../../lib/wca-live/attempts';
+
+const colorForResult = (regionalRecord, pbMarker) => {
```
kr-matthews, line 10: I think the `let` could be removed and each `recordColor = ...` assignment could be replaced with `return ...`, if the order is reversed - i.e. regional record then pb marker then empty case.

The name should be `styleForResult`, or the function refactored to just return the string of the color.

PR #9185, opened 2024-04-01
Extract Form logic of "Create Competition" Form to its own folder/component

gregorbg:
This is the first step towards unifying all of the different "You have unsaved changes. Be sure to save them!"-style forms into a common React feature set.

Do not fear the diff line count! The biggest part of this PR is just moving code around, and copy/pasting code into their own files.
This makes a lot of JS `import` statements change, and some hooks are called _slightly_ differently.

The only "relevant" files in terms of a code review are under `components/wca/FormBuilder`. The general idea is that we expose the "current", i.e. "work in progress" object and the "committed to the store" version of the object via separate hooks.


gregorbg:
Right now, this is a plain refactor of the Competition Form, without adding any functionality.

The next step is to use this new `FormBuilder` for the existing `EditEvents` and `EditSchedule` components as well, which will be a follow-up PR.

app/webpacker/components/wca/FormBuilder/provider/FormObjectProvider.jsx
```diff
@@ -0,0 +1,110 @@
...
+  children,
+  initialObject,
+}) {
+  const initialState = useMemo(() => ({
+    object: initialObject,
+    initialObject,
+    errors: null,
+  }), [initialObject]);
+
+  const [formState, dispatch] = useReducer(formReducer, initialState);
```
kr-matthews, line 25: Might as well define
```
const createState = (initialObject) => ({
  object: initialObject,
  initialObject,
  errors: null,
})
```
_outside_ the component, then do
```
const [formState, dispatch] = useReducer(formReducer, initialState, createState);
```

gregorbg: Cool! I didn't know this was a thing, thanks for pointing it out :)

PR #9186, opened 2024-04-01
Add regional delegates to notication emails sent to WCAT

DanielEgdal:
WCAT is currently adding the regional delegates manually for each submission. This is a lot of work, so here is the code to do it automatically.

The current repo mixed with the docker file works quite poorly. But from the test I were able to do, this works.

### DIFF
--- a/app/mailers/competitions_mailer.rb
+++ b/app/mailers/competitions_mailer.rb
@@ -13,7 +13,7 @@ class CompetitionsMailer < ApplicationMailer
       mail(
         from: Team.wcat.email,
         to: Team.wcat.email,
-        cc: (competition.delegates.map(&:email) + delegates_to_senior_delegates_email(competition.delegates)).compact.uniq,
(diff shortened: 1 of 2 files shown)


app/mailers/competitions_mailer.rb
```diff
@@ -175,6 +175,10 @@ def registration_reminder(competition, user, registered_but_not_accepted)
     delegates.flat_map { |delegate| delegate.senior_delegates.map(&:email) }.uniq.compact
   end
 
+  private def delegates_to_regional_delegates_email(delegates)
+    delegates.flat_map { |delegate| delegate.region&.lead_user.email }.uniq.compact
```
danieljames-dj, line 179: `region` (of `delegate.region`) will soon be removed. User won't have direct relation with the region. But like `def senior_delegates` of `user.rb`, you can create a method for regional delegates as follows:
```
  def regional_delegates
    active_roles
      .select { |role| UserRole.is_group_type?(role, UserGroup.group_types[:delegate_regions]) }
      .map { |role| UserRole.group(role).lead_user }
  end
```

Then you can re-write this line as:
```
delegates.flat_map { |delegate| delegate.regional_delegates.map(&:email) }.uniq.compact
```

gregorbg (approved):

PR #9187, opened 2024-04-02
Added officer & board roles to db/seeds

danieljames-dj:


gregorbg:
This contains some interesting test fails, seems like statically seeding `user_roles` breaks quite a lot.

I am surprised to see that the old `user_seeds` contained seeding Board members, but for some reason did _not_ contain the legacy `team_members` table (which would have been necessary to store Board membership information) in the constant tables. So I am also confused about what's going on here.

How necessary / important is this change and what's the context?

danieljames-dj:
> This contains some interesting test fails, seems like statically seeding `user_roles` breaks quite a lot.
> 
> I am surprised to see that the old `user_seeds` contained seeding Board members, but for some reason did _not_ contain the legacy `team_members` table (which would have been necessary to store Board membership information) in the constant tables. So I am also confused about what's going on here.
> 
> How necessary / important is this change and what's the context?

Ya this looks very tough. I'll try to get it resolved.
This is done for 2 main reasons:
- When a developer runs `bin/rake db:reset`, they won't get proper roles & groups related to officers & board and might cause difficulty in their development. So this will create the required roles & groups.
- More unit tests are going to come to test roles fetching, searching, filtering, sorting, etc. Same for many test cases related to roles & groups, so having them loaded in seeds will make the unit tests very clean.

I'll get back to this PR soon.

danieljames-dj:
Closing this PR because it's going to be done in another way later.

PR #9188, opened 2024-04-02
Update non-lead delegates metadata

danieljames-dj:


app/jobs/delegates_metadata_sync_job.rb
```diff
@@ -0,0 +1,23 @@
...
+        total_delegated = user.actually_delegated_competitions.to_a.length
+        if role.metadata.first_delegated != first_delegated
+          role.metadata.update!(first_delegated: first_delegated)
+        end
+        if role.metadata.last_delegated != last_delegated
+          role.metadata.update!(last_delegated: last_delegated)
+        end
+        if role.metadata.total_delegated != total_delegated
+          role.metadata.update!(total_delegated: total_delegated)
+        end
```
gregorbg, line 19: You don't need to do these comparisons manually; Rails is clever enough to figure out whether an attribute changed.
The important difference is:
- Calling `my_model.update!(foo: 1, bar: 2, baz: 'abc')` **definitely** fires an SQL query.
- Setting `my_model.foo = 1; my_model.bar = 2; my_model.baz = 'abc'` and then calling `my_model.save!` only updates the attributes that actually changed.

Further reading (only if you're interested; this is **not** something you need to manually call/check in this PR. It's a reading to see what Rails ActiveRecord **already does** under the hood for you): https://api.rubyonrails.org/classes/ActiveModel/Dirty.html

gregorbg (changes requested):

app/jobs/delegates_metadata_sync_job.rb
```diff
@@ -0,0 +1,18 @@
...
+  def perform
+    UserGroup.delegate_regions.flat_map(&:roles).map do |role|
+      unless role.is_lead?
+        user = role.user
+        first_delegated = user.actually_delegated_competitions.to_a.minimum(:start_date)
+        last_delegated = user.actually_delegated_competitions.to_a.maximum(:start_date)
+        total_delegated = user.actually_delegated_competitions.to_a.length
+        role.metadata.first_delegated = first_delegated
+        role.metadata.last_delegated = last_delegated
+        role.metadata.total_delegated = total_delegated
```
gregorbg, line 13: Now you don't even need the variables anymore.
Instead of `foo = my_cool_method; role.metadata.first_delegated = foo;` why not just do `role.metadatafirst_delegated = my_cool_method`?

danieljames-dj: Oh yes, done.

gregorbg (approved):

PR #9190, opened 2024-04-02
Migrate remaining delegates to roles

danieljames-dj:
This is code complete, but need an end-to-end testing before merging. But should be ready for code-review.


spec/models/user_spec.rb
```diff
@@ -65,19 +65,6 @@
...
-    user = FactoryBot.create :user
-
-    delegate.region_id = user.region_id
-    expect(delegate).to be_invalid_with_errors(region_id: ["can't be blank"])
-
-    delegate.region_id = region_id
-    user.update(delegate_status: "delegate", region_id: region_id)
-    expect(delegate).to be_valid
-  end
-
```
danieljames-dj, line 80: This unit test is no longer required because the region_id column is no longer updated, and will be removed in another PR once this PR merges. In the intermediate period, the region_id is not updated even if a delegate role is added.

danieljames-dj:
Tested and looks good. Fixed unit tests too. :-)

app/models/roles_metadata_delegate_regions.rb
```diff
@@ -5,7 +5,7 @@ class RolesMetadataDelegateRegions < ApplicationRecord
     senior_delegate: "senior_delegate",
     regional_delegate: "regional_delegate",
     delegate: "delegate",
-    junior_delegate: "candidate_delegate",
+    candidate_delegate: "candidate_delegate",
```
gregorbg, line 8: Veto. The official terminology is "Junior Delegate", so you should change the actual stored values rather than changing the key

app/models/user.rb
```diff
@@ -240,7 +240,7 @@ def claim_wca_id_validations
         errors.add(:unconfirmed_wca_id, I18n.t('users.errors.already_have_id', wca_id: wca_id))
       end
 
-      if delegate_id_to_handle_wca_id_claim.present? && !delegate_to_handle_wca_id_claim
+      if delegate_id_to_handle_wca_id_claim.present? && !User.find_by(id: delegate_id_to_handle_wca_id_claim).any_kind_of_delegate?
```
gregorbg, line 243: Why are you changing this line? The `belongs_to` above should still work...

app/models/user.rb
```diff
@@ -561,15 +559,7 @@ def any_kind_of_delegate?
   end
 
   def trainee_delegate?
-    delegate_status == "trainee_delegate"
-  end
-
-  def staff_delegate?
```
gregorbg, line 567: Where did this method go?

app/models/user.rb
```diff
@@ -561,15 +559,7 @@ def any_kind_of_delegate?
...
-    any_kind_of_delegate? && !trainee_delegate?
-  end
-
-  def full_delegate?
```
gregorbg, line 571: Same here, where did it go?

app/models/user.rb
```diff
@@ -1433,12 +1413,33 @@ def roles
...
     roles.concat(team_roles)
 
     roles
   end
+
+  private def highest_delegate_role
+    highest_status_rank = nil
+    highest_role = nil
+    delegate_roles.each do |role|
+      status_rank = Api::V0::UserRolesController.status_sort_rank(role)
```
gregorbg, line 1425: Veto about accessing a controller from a model.
If really necessary, then the controller should access the model.

app/models/roles_metadata_delegate_regions.rb
gregorbg: (This also implies changes throughout the rest of the diff of the whole PR, that I will not point out one by one)

db/migrate/20240402051116_add_delegate_roles.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class AddDelegateRoles < ActiveRecord::Migration[7.1]
+  def change
+    User.where.not(delegate_status: nil).each do |user|
+      UserRole.create!(
+        user_id: user.id,
+        group_id: user.read_attribute(:region_id),
+        start_date: '2004-08-01',
```
gregorbg, line 9: I'm assuming you will manually backfill this once the migration is done?

app/models/roles_metadata_delegate_regions.rb
danieljames-dj: Thing is, the value in `users` table currently is `candidate_delegate` and is used in many places. Using the key `junior_delegate` here is not working because the value stored in database is `candidate_delegate`. My plan is to do the change all occurrences of candidate_delegate to junior_delegate in a later PR because it will involve much more files. Is that fine?

app/models/user.rb
danieljames-dj: One of the test-case which was failing because of this was "delegate_id_to_handle_wca_id_claim must be a delegate". The `delegate_to_handle_wca_id_claim` (the `belongs_to`) can be a non-delegate. The association-way of checking might be bit complicated, hence chose this way of checking if the delegate_id_to_handle_wca_id_claim is a delegate or not.

app/models/user.rb
danieljames-dj: This method is not used anywhere, hence removed it.

app/models/user.rb
danieljames-dj: This method is not used anywhere, hence removed it.

app/models/user.rb
danieljames-dj: Okay moved to UserRole model.

db/migrate/20240402051116_add_delegate_roles.rb
danieljames-dj: Yes correct.

app/models/user.rb
gregorbg: Are you sure? I feel like `competition.rb` has some "make sure that Delegates for this competition are actually Delegates" validation :thinking:

app/models/user.rb
```diff
@@ -1433,12 +1413,33 @@ def roles
...
+    highest_role = nil
+    delegate_roles.each do |role|
+      status_rank = UserRole.status_sort_rank(role)
+      if highest_status_rank.nil? || status_rank < highest_status_rank
+        highest_role = role
+        highest_status_rank = status_rank
+      end
+    end
+    highest_role
+  end
```
gregorbg, line 1432:
```suggestion
  private def highest_delegate_role
    delegate_roles.max_by { |role| UserRole.status_sort_rank(role)  }
  end
```

app/models/user.rb
```diff
@@ -1433,12 +1413,33 @@ def roles
...
+
+  def delegate_status
+    highest_delegate_role&.metadata&.status
+  end
+
+  def region_id
+    highest_delegate_role&.group&.id
```
gregorbg, line 1439: I feel like `role&.group_id` should be enough, is there an actual need to load the full `group` row into memory?

spec/controllers/api/v0/user_roles_controller_spec.rb
```diff
@@ -26,19 +26,7 @@
...
-          is_active: true,
-          group: user_senior_delegate_role.group,
-          user: user_whose_delegate_status_changes,
-          is_lead: false,
-          metadata: {
-            status: "candidate_delegate",
-            location: "Australia",
-          },
-          class: 'userrole',
-        }].to_json)
```
gregorbg, line 41: About this deletion: I see that you're removing the "manually glued together" roles in this PR. But shouldn't there be a replacement of a "clean" role object being sent instead?
I don't get why this `expect` gets deleted, from my perspective it should be _replaced_ with an equivalent "proper" role serialization. Can you please clarify?

spec/controllers/api_controller_spec.rb
```diff
@@ -178,15 +178,16 @@
   describe 'GET #delegates' do
     it 'includes emails and regions' do
       senior_delegate = FactoryBot.create :senior_delegate_role
-      delegate = FactoryBot.create :delegate, location: "SF bay area, USA", region_id: senior_delegate.group.id
+      delegate = FactoryBot.create :delegate_role, group_id: senior_delegate.group.id
+      delegate.metadata.update!(location: "SF bay area, USA")
```
gregorbg, line 182: I'm not super happy about this hard-wired update call in tests.
Would it be a lot of trouble to pass through the `location` to the `metadata` in the FactoryBot macro?

gregorbg (changes requested):
I wanna take a look at that `reload` quirk in the FactoryBot comment. In the meantime, here's some more minor code comments :wink:

app/models/roles_metadata_delegate_regions.rb
gregorbg: Okay, in that case you can just let the value be `"candidate_delegate"` but still keep the _key_ at `junior_delegate` (as it already has been before submitting this PR).

In other words: Why do you need to change the _key_ here? Rails enums feature is exactly the solution that should allow you to keep the _value_ in the database intact, but use a consistent _key_ in the code...

app/models/user.rb
gregorbg: In that case, you should probably rename the above `belongs_to` thingy to `user_to_handle_wca_id_claim` and change this call accordingly to `user_to_handle_wca_id_claim.any_kind_of_delegate?`. I'd like to keep manual "boring" `find_by` lookups (i.e. `find` that uses only `id` and nothing else) to an absolute minimum.

app/webpacker/components/Panel/SeniorDelegate/Regions/Subregion.jsx
```diff
@@ -29,7 +29,7 @@ const isLead = (role) => role.metadata.status === 'leader';
 
 const canPromote = (role) => (
   [
-    statusObjectOfGroupType(role.group.group_type).trainee_delegate,
+    statusObjectOfGroupType(role.group.group_type).candidate_delegate,
```
gregorbg, line 32: Are you sure you changed the correct line in this file?

app/models/user.rb
```diff
@@ -553,23 +551,15 @@ def admin?
   end
 
   def delegate_roles
-    active_roles.filter { |role| UserRole.is_group_type?(role, UserGroup.group_types[:delegate_regions]) }
+    active_roles(include_converted: false).filter { |role| UserRole.is_group_type?(role, UserGroup.group_types[:delegate_regions]) }
   end
 
   def any_kind_of_delegate?
-    active_roles.any? { |role| UserRole.is_group_type?(role, UserGroup.group_types[:delegate_regions]) }
+    !delegate_roles.empty
```
gregorbg, line 558: Attention: This must be `empty?` with a question mark at the end

gregorbg (changes requested):

app/webpacker/components/Panel/SeniorDelegate/Regions/Subregion.jsx
gregorbg: (Although this is irrelevant if we revert the "junior -> candidate" change

danieljames-dj: Oops my bad, reverted. Thanks.

app/models/user.rb
danieljames-dj: `delegate_roles.any?` looks better. Thanks for pointing `empty` vs `empty?`.

app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -434,29 +391,14 @@ def destroy
     return head :unauthorized unless current_user.has_permission?(:can_edit_groups, role.group.id)
     role.update!(end_date: Date.today)
     RoleChangeMailer.notify_role_end(role, current_user).deliver_later
+    if role.user.delegate_roles.count == 0
```
gregorbg, line 394: This should be almost identical to the `any_kind_of_delegate?` method in `user.rb`, right?!

app/models/user.rb
gregorbg: In that case, may I offer a compromise: Just keep the name as-is but still use the association here, despite the fact that the name might be confusing?

app/models/user.rb
gregorbg: Okay, gotcha

spec/controllers/api/v0/user_roles_controller_spec.rb
gregorbg: No, I think the way you outlined will be fine :) Thanks for the explanation!

app/controllers/api/v0/user_roles_controller.rb
danieljames-dj: Yes, thanks. Done.

app/models/user.rb
```diff
@@ -614,11 +604,7 @@ def banned_at_date?(date)
   end
 
   private def senior_delegate_roles
-    active_roles.select do |role|
-      UserRole.is_group_type?(role, UserGroup.group_types[:delegate_regions]) &&
-        UserRole.is_lead?(role) &&
-        UserRole.status(role) == RolesMetadataDelegateRegions.statuses[:senior_delegate]
-    end
+    delegate_roles.select { |role| role.metadata.status == RolesMetadataDelegateRegions.statuses[:senior_delegate] }
```
gregorbg, line 607: Here you dropped the `is_lead?` check from the previous implementation. Are you sure this is risk-free?

app/models/user.rb
```diff
@@ -1326,45 +1312,15 @@ def can_manage_delegate_probation?
...
 
   def regional_delegates
-    active_roles
-      .select { |role| UserRole.is_group_type?(role, UserGroup.group_types[:delegate_regions]) }
-      .map { |role| UserRole.group(role).lead_user }
-  end
-
-  def active_roles
-    roles.select { |role| UserRole.is_active?(role) }
+    delegate_roles.map { |role| UserRole.group(role).lead_user }
```
gregorbg, line 1319: Here you're inconsistent about using `UserRole.group` helper vs. actual `group` association.
I would prefer using the `UserRole.group` helper above, because that will give us easier access to find all places we can optimize at the end of this project.

app/models/user.rb
```diff
@@ -1415,10 +1371,9 @@ def can_access_panel?
   end
 
   def subordinate_delegates
-    roles
-      .filter { |role| UserRole.is_group_type?(role, UserGroup.group_types[:delegate_regions]) }
-      .filter { |role| UserRole.is_lead?(role) }
-      .flat_map { |role| UserRole.group(role).active_users + UserRole.group(role).active_users_of_all_child_groups }
+    delegate_roles
+      .filter { |role| role.is_lead? }
+      .flat_map { |role| role.group.active_users + role.group.active_users_of_all_child_groups }
```
gregorbg, line 1376: See my remark about `UserRole` helpers above.

config/locales/en.yml
```diff
@@ -392,7 +392,7 @@ en:
       status:
         delegate_regions:
           trainee_delegate: "Trainee Delegate"
-          candidate_delegate: "Junior Delegate"
+          junior_delegate: "Junior Delegate"
```
gregorbg, line 395: Just double-checking to make sure this value actually needs to be changed?
I mean, in the database it will still be `candidate_delegate`

gregorbg (changes requested):
More observations :innocent:

app/models/user.rb
danieljames-dj: Yes it is, because senior delegate is anyway a lead.

app/models/user.rb
danieljames-dj: Done

app/models/user.rb

gregorbg (approved):

config/locales/en.yml
danieljames-dj: Resolving this as we discussed this over slack.

app/models/roles_metadata_teams_committees.rb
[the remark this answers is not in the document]
danieljames-dj: Removed 🙈

gregorbg (approved):
Please only merge when you're available to oversee the deploy and react to issues that arise.

PR #9191, opened 2024-04-04
Competitions Overview - Search Param Sanitization

kr-matthews:
- Country/continent is now backward compatible.
- The date-picker won't crash when you manually insert an invalid date as a search param.
- Delegate param isn't touched, since it needs to be switched back to user ID (plus sanitizing it depends on first fetching the delegates data).
- Events are sanitized, but still don't have backward compatibility. @gregorbg said he could help with this (in a separate PR?).
- Search and shouldIncludeCancelled don't need sanitizing (right?).

Bonus: `PAST_YEARS_WITH_COMPETITIONS` includes the current year, so I renamed it.

### DIFF
--- a/app/webpacker/components/CompetitionsOverview/CompetitionsFilters.js
+++ b/app/webpacker/components/CompetitionsOverview/CompetitionsFilters.js
@@ -2,7 +2,6 @@ import React from 'react';
 import {
   Button, Icon, Form, Dropdown, Popup, List, Input, Header,
 } from 'semantic-ui-react';
-import { DateTime } from 'luxon';
 import PulseLoader from 'react-spinners/PulseLoader';
 
 import I18n from '../../lib/i18n';
@@ -12,13 +11,9 @@ import {
 
 import useDelegatesData from './useDelegatesData';
 import UtcDatePicker from '../wca/UtcDatePicker';
+import { YEARS_WITH_PAST_COMPETITIONS } from './filterUtils';
 
 const WCA_EVENT_IDS = Object.values(events.official).map((e) => e.id);
-const PAST_YEARS_WITH_COMPETITIONS = [];
-for (let { year } = DateTime.now(); year >= 2003; year -= 1) {
-  PAST_YEARS_WITH_COMPETITIONS.push(year);
-}
-PAST_YEARS_WITH_COMPETITIONS.push(1982);
 
 function CompetitionsFilters({
   filterState,
@@ -285,7 +280,7 @@ function PastCompYearSelector({ filterState, dispatchFilter }) {
           >
             {I18n.t('competitions.index.all_years')}
           </Dropdown.Item>
-          {PAST_YEARS_WITH_COMPETITIONS.map((year) => (
+          {YEARS_WITH_PAST_COMPETITIONS.map((year) => (
             <Dropdown.Item
               key={`past_select_${year}`}
               onClick={() => dispatchFilter({ timeOrder: 'past', selectedYear: year })}
--- a/app/webpacker/components/CompetitionsOverview/CompetitionsFilters.js
+++ b/app/webpacker/components/CompetitionsOverview/CompetitionsFilters.js
@@ -175,7 +175,9 @@ function DelegateSelector({ delegateId, dispatchFilter }) {
         search
         deburr
         selection
-        error={!delegatesLoading && delegateId && delegatesData.every(({ id }) => id !== delegateId)}
+        error={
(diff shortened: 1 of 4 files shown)


app/webpacker/components/CompetitionsOverview/CompetitionsFilters.js
```diff
@@ -285,7 +280,7 @@ function PastCompYearSelector({ filterState, dispatchFilter }) {
           >
             {I18n.t('competitions.index.all_years')}
           </Dropdown.Item>
-          {PAST_YEARS_WITH_COMPETITIONS.map((year) => (
+          {YEARS_WITH_PAST_COMPETITIONS.map((year) => (
```
danieljames-dj, line 283: Maybe `YEARS_LIST_OF_PAST_COMPETITIONS` or `YEARS_OF_PAST_COMPETITIONS`?

app/webpacker/components/CompetitionsOverview/filterUtils.js
```diff
@@ -1,4 +1,19 @@
...
+
+const WCA_EVENT_IDS = Object.keys(events.byId);
+
+const YEARS_WITH_PAST_COMPETITIONS = [];
+for (let { year } = DateTime.now(); year >= 2003; year -= 1) {
+  YEARS_WITH_PAST_COMPETITIONS.push(year);
+}
+YEARS_WITH_PAST_COMPETITIONS.push(1982);
+
+export { YEARS_WITH_PAST_COMPETITIONS };
```
danieljames-dj, line 16: I don't think @gregorbg will accept the generation of years list this way. This should have been missed by Gregor in the review of https://github.com/thewca/worldcubeassociation.org/pull/7902.

I think what Gregor suggests will be to export this in `wca-data.js.erb` in the following way:
```
export const YEARS_WITH_PAST_COMPETITIONS = <%= Competition.all.map(&:start_date).map(&:year).uniq.sort %>
```

gregorbg: Hehe, I've been reviewing too much of your code recently, it seems :stuck_out_tongue: 

As a matter of fact, I wouldn't call it "missed". More like "be aware that we have developers with different degrees of experience and different aspirations" :wink: 

But indeed, your suggestion looks more concise. Except for the fact that `Competition.all.map(...)` loads all competitions from our DB into memory via a `SELECT * FROM Competitions` call and then loops over them only for retrieving the start date. That's a complete waste of resources, and can be improved with `Competition.pluck(:start_date)` which translates to a `SELECT start_date FROM Competitions` SQL call

app/webpacker/components/CompetitionsOverview/CompetitionsFilters.js
gregorbg: Hm, that's why they say "naming stuff is hard" in software development :smile: 
I'm fine with either and I won't block this PR just because of some minor naming issue of a constant used to populate a dropdown. If it was a more central value, amybe I'd argue more, but honestly for a helper variable I am 50/50 on spending time to perfect it on one hand and just getting the task done on the other hand.

gregorbg (changes requested):
All of my remarks are follow-ups to Daniel's comments.

kr-matthews:
About delegate ids, if I add "&delegate=banana" to the url, then I get `ActiveRecord::RecordNotFound in CompetitionsController#index`. Not sure how that's happening, since we have `Number(searchParams.get(DELEGATE)) || DEFAULT_DELEGAT`.

Since we can't check the delegate id is valid until the data is loaded, I added error UI for once the data has loaded. But we could also use a side effect to reset the delegate id instead. That might be more user friendly.

image

gregorbg:
> About delegate ids, if I add "&delegate=banana" to the url, then I get `ActiveRecord::RecordNotFound in CompetitionsController#index`. Not sure how that's happening, since we have `Number(searchParams.get(DELEGATE)) || DEFAULT_DELEGAT`.

Fun fact, this is actually an error in the old code that already exists in the legacy competitions list (i.e. the one that is currently live on production): https://www.worldcubeassociation.org/competitions?delegate=banana breaks as of writing this comment. It has nothing to do with the changes introduced in this PR :innocent:

gregorbg (approved):
Happy to merge as-is, what else is there left to do?

kr-matthews:
> Happy to merge as-is, what else is there left to do?

I think that's it for this PR.
The only other things for this page are loading registration state separately, and the admin version of the page.

PR #9193, opened 2024-04-04
Fix registration mailer link

FinnIckler:


gregorbg (changes requested):
As discussed, this should probably just rename the controller file so that it matches what `routes.rb` expects. This is consistent with the rest of our codebase, which mostly uses plural for controllers anyways

gregorbg (approved):

PR #9194, opened 2024-04-05
Shuffle around PaymentIntent code to get rid of AttendeePaymentRequest

gregorbg:
We're storing `MicroserviceRegistration`s in the database now anyways (for WCIF and assignments compatibility) so we might as well use those as `holder`s of the Payment Tickets that the microservice is querying.

While working on this, I realised that an awful lot of code is duplicated and can probably be extracted to some common interface. This leads to a high line diff code, but I promise most of it is copy-paste.

Plus it will make it much easier to open up APIs to generic PayPal access later.


app/controllers/api/internal/v1/payment_controller.rb
```diff
@@ -3,65 +3,31 @@
...
-
-    payment_intent_args = {
-      amount: stripe_amount,
-      currency: currency_iso,
-      receipt_email: user.email,
-      description: "Registration payment for #{competition.name}",
-      metadata: registration_metadata,
-    }
+    competition = Competition.find(competition_id)
+    render json: { error: "Competition not found" }, status: :not_found unless competition_id.present?
```
dunkOnIT, line 15: Should this not go above Line 14? Given that we're trying to use competition_id in line 14

app/controllers/api/internal/v1/payment_controller.rb
```diff
@@ -3,65 +3,31 @@
...
-    enable_automatic_pm = !Rails.env.test?
+    ms_registration = competition.microservice_registrations
+                                 .includes(:competition, :user)
+                                 .find_by(user_id: registering_user_id)
 
-    # we cannot recycle an existing intent, so we create a new one which needs all possible PaymentMethods enabled.
-    # Required as per https://stripe.com/docs/payments/accept-a-payment-deferred?type=payment&client=html#create-intent
-    payment_intent_args[:automatic_payment_methods] = { enabled: enable_automatic_pm }
+    render json: { error: "Registration not found" }, status: :not_found unless ms_registration.present?
+    render json: { error: "Paying user not found" }, status: :not_found unless paying_user.present?
```
dunkOnIT, line 22: Move to below Line 12?

app/models/stripe_record.rb
```diff
@@ -76,11 +80,39 @@ def update_status(api_transaction)
...
+      Stripe::Charge.retrieve(self.stripe_id, stripe_account: self.account_id)
     when 'refund'
-      Stripe::Refund.retrieve(self.stripe_id, stripe_account: self.find_account_id)
+      Stripe::Refund.retrieve(self.stripe_id, stripe_account: self.account_id)
+    end
+  end
+
+  alias_method :retrieve_remote, :retrieve_stripe
+
+  def update_amount_remote(amount_iso, currency_iso)
```
dunkOnIT, line 93: Do we want to change this to `update_amount_stripe` for consistency with `retrieve_stripe`?

app/models/stripe_record.rb
```diff
@@ -76,11 +80,39 @@ def update_status(api_transaction)
...
+    if self.payment_intent?
```
dunkOnIT, line 94: - Should we be checking `payment_intent.present?` explicitly instead? 
- Stylistically, do we prefer to explicitly call `self`?

app/controllers/api/internal/v1/payment_controller.rb
dunkOnIT: Ah that makes!

dunkOnIT:
I raised a concern re making API calls from the model - did a bit of research, and I was just totally wrong about that being an issue :D 

I think there's an argument to be made for putting the API interactions in a `stripe_interface` file. I'm not sure how strong that argument is though, and there is a lot of convenience in keeping it in the model - so I'm on board with keeping it as-is.

gregorbg:
> I raised a concern re making API calls from the model - did a bit of research, and I was just totally wrong about that being an issue :D
> 
> I think there's an argument to be made for putting the API interactions in a `stripe_interface` file. I'm not sure how strong that argument is though, and there is a lot of convenience in keeping it in the model - so I'm on board with keeping it as-is.

Yup, we might move it to a separate interface but that is left for a follow-up PR. It's not a strict technical requirement here.

app/models/stripe_record.rb
gregorbg: I was debating with that but ideally, we want to move this method out of `stripe_account` and into `payment_intent`, so that _any_ record can update its remote regardless of whether it's Stripe or PayPal or whatnot.

app/models/stripe_record.rb
gregorbg: Re 1- No, because that's doing different things. In **this** file, there is an `enum` at the top which has a `payment_intent` entry. Declaring this enum also creates the method `payment_intent?` automagically, which is pretty much just a convenience alias for `self.stripe_record_type == 'payment_intent'`

gregorbg: Re 2- I do like it because it's more explicit and less "arcane" about where the methods come from. Especially useful in dynamic languages like Ruby where you cannot rely on typing.

dunkOnIT: The point is well-made because I misunderstood what `payment_intent?` was doing here lol.

dunkOnIT (approved):
LGTM, and tested locally

PR #9195, opened 2024-04-05
Add board roles to db/seeds

danieljames-dj:
This PR is kind of like subset of https://github.com/thewca/worldcubeassociation.org/pull/9187. This PR aims in debugging the issue of sign_in during unit tests with roles from db/seeds. One example is `bin/rspec spec/features/eligible_voters_spec.rb`.

### DIFF
--- a/app/models/user_role.rb
+++ b/app/models/user_role.rb
@@ -108,6 +108,9 @@ class UserRole < ApplicationRecord
   }.freeze
 
   def serializable_hash(options = nil)
+    puts("User ID: #{self.user_id}")
+    puts("User Object: #{User.find_by(id: self.user_id)}")
+    puts("User ID from user object#{user&.id}")
     json = super(DEFAULT_SERIALIZE_OPTIONS.merge(options || {}))
     json[:class] = self.class.to_s.downcase
     json
--- a/db/seeds/user_roles.seeds.rb
+++ b/db/seeds/user_roles.seeds.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+after :user_groups do
+  UserRole.create!(
+    user: FactoryBot.create(:user_with_wca_id),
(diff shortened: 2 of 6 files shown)


danieljames-dj:
@gregorbg I'm commenting the issue here instead of DM-ing in slack so that others can also chime-in if they know the solution.

Currently, then I create a user-role in db/seeds, and trying to access it in unit test, I'm able to access the role, but the corresponding user is not available. I wrote some logs in the code to debug and I was surprised to see following logs:

```
User ID: 1
User Object: 
User ID from user object
Role json: {"id":1,"start_date":"2014-12-08","end_date":null,"group":{"id":15,"name":"WCA Board of Directors","group_type":"board","parent_group_id":null,"is_active":true,"is_hidden":false,"metadata_id":1,"metadata_type":"GroupsMetadataBoard","created_at":"2024-04-10T10:43:01.156Z","updated_at":"2024-04-10T10:43:01.156Z","lead_user":null,"metadata":{"id":1,"email":"board@worldcubeassociation.org","created_at":"2024-04-10T10:43:00.984Z","updated_at":"2024-04-10T10:43:00.984Z"}},"class":"userrole"}
Board user ID: 
Board user name: 
```

This means that when the role was created, the user is also created with user ID 1, but then the user got deleted somehow. In the first line, I was printing the user_id of roles object which printed successfully. Second line is printing `User.find_by(id: user_id)`, and that is printing empty. (I tried `User.find(user_id)` as well. Then I printed role JSON and it didn't had user.

In local, when I run `bin/rspec spec/controllers/api_controller_spec.rb`, I'm able to reproduce this issue. But when I run one particular unit test context, it's not reproducing: `bin/rspec spec/controllers/api_controller_spec.rb:216`.

I am not understanding how the user got removed. Can someone please help in figuring out why this happened?

danieljames-dj:
Closing this PR because it's going to be done in another way later.

PR #9196, opened 2024-04-07
Harmonize Stripe SDK and Dashboard API versions

gregorbg:
Basically just an aftermath of dependency updates that has been overlooked for a long time.
Our Dashboard has been running on `2023-10-16` for over two months, so I expect that we'll be fine and the API will pass easily.


gregorbg:
Turns out, there are some non-trivial changes in our CI testing config. Will wait until https://github.com/thewca/worldcubeassociation.org/pull/9194 is merged.

PR #9197, opened 2024-04-07
Migrate contacts page to React

danieljames-dj:
This PR will help to customize "contact WRT" part of the contact form a lot.


danieljames-dj:
Need help in making the UI better. Currently there are plenty of space between radio button rows, can someone please help in suggesting a solution?

Screenshot 2024-04-07 at 9 10 41 PM

app/webpacker/components/ContactsPage/ContactForm.jsx
```diff
@@ -0,0 +1,119 @@
...
+  const [captchaError, setCaptchaError] = useState(false);
+  const recaptchaRef = useRef();
+  const isFormValid = useMemo(() => {
+    if (!formValues.contactType
+      || !formValues.userData.name
+      || !formValues.userData.email) {
+      return false;
+    }
+    return true;
+  }, [formValues]);
```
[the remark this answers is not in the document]
danieljames-dj: Done, thanks

danieljames-dj:
I managed to fix the UI by adding some CSS. So this is ready now.

Screenshot 2024-04-10 at 8 43 37 AM

app/webpacker/components/ContactsPage/ContactForm.jsx
```diff
@@ -0,0 +1,123 @@
...
+  const [captchaError, setCaptchaError] = useState(false);
+  const recaptchaRef = useRef();
+
+  const isFormValid = (
+    formValues.contactType && formValues.userData.name && formValues.userData.email
+  );
+  const SubForm = useMemo(() => {
+    if (!formValues.contactType) return null;
+    switch (formValues.contactType) {
+      case CONTACT_TYPES[0]: // competition
```
gregorbg, line 34: This should be managed better than "accidentally knowing" that index zero is `competition` (even with the code comment)

gregorbg: How does the panel index do it? I vaguely remember some cool hash/object indexing there that made the code much more readable

gregorbg (changes requested):

app/webpacker/components/ContactsPage/ContactForm.jsx
danieljames-dj: I've created an enum-kind of object in this class, will that be fine?

app/webpacker/lib/wca-data.js.erb
```diff
@@ -192,3 +194,14 @@ export const officersStatus = <%= RolesMetadataOfficers.statuses.to_json %>;
...
+// ----- CONTACT FORM -----
+export const CONTACT_TYPES = [
+  'competition',
+  'competitions_in_general',
+  'results_team',
+  'wca_id_or_profile',
+  'media',
+  'software',
+  'different',
+];
```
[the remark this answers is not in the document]
danieljames-dj: Since we don't store website_contact in DB, creating enum in model didn't work. I created enum kind of thing in react.

app/webpacker/components/ContactsPage/ContactForm.jsx
```diff
@@ -0,0 +1,123 @@
...
+            ...formValues,
+            [formValues.contactType]: subFormData,
+          })}
+        />
+      )}
+      <FormField>
+        <ReCAPTCHA
+          ref={recaptchaRef}
+          sitekey={RECAPTCHA_PUBLIC_KEY}
+          onChange={() => setCaptchaError(false)}
```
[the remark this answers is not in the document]
gregorbg: Thanks, so this is actually just poor naming on reCaptcha's part.

app/webpacker/components/ContactsPage/ContactForm.jsx
```diff
@@ -0,0 +1,135 @@
...
+  const initialFormValues = useMemo(() => ({
+    userData: {
+      name: userDetails?.user?.name || '',
+      email: userDetails?.user?.email || '',
+    },
+  }), [userDetails]);
+  const [formValues, setFormValues] = useInputState(initialFormValues);
+  const { save, saving } = useSaveAction();
+  const [captchaError, setCaptchaError] = useState(false);
+  const recaptchaRef = useRef();
```
gregorbg, line 41: I think you can achieve this without a `ref`. Just introduce a `useState` for the captcha value, and the `onChange` method below gets passed the new value as parameter: https://www.npmjs.com/package/react-google-recaptcha#usage. So you can just use the state setter as the `onChange` callback and access it wherever you need it without introducing a ref.

app/webpacker/components/ContactsPage/ContactForm.jsx
```diff
@@ -0,0 +1,135 @@
...
+}, {});
+
+export default function ContactForm({ userDetails }) {
+  const initialFormValues = useMemo(() => ({
+    userData: {
+      name: userDetails?.user?.name || '',
+      email: userDetails?.user?.email || '',
+    },
+  }), [userDetails]);
+  const [formValues, setFormValues] = useInputState(initialFormValues);
```
gregorbg, line 38: Hm, reading your other comments makes me think that you're stuffing too much into the `formValues` state here.
I believe you should have an independent state for "which toggle is selected" and then another state for the "sub form values", i.e. all relevant information from the form.

Another alternative would be to introduce a reducer for state management. I have a slight preference towards the first option, but ultimately I leave it up to you to decide which is better suited here.

app/webpacker/components/ContactsPage/ContactForm.jsx
```diff
@@ -0,0 +1,135 @@
...
+          />
+        )}
+      </FormField>
+      <Button
+        disabled={!isFormValid}
+        type="submit"
+        onClick={(e) => {
+          const captchaValue = recaptchaRef.current.getValue();
+          if (!captchaValue) {
+            setCaptchaError(true);
```
gregorbg, line 126: Instead of just setting this to a generic `true`, you can actually use a `onErrored` callback on the captcha component above.

app/webpacker/components/ContactsPage/ContactForm.jsx
```diff
@@ -0,0 +1,135 @@
...
+  'wca_id_or_profile',
+  'media',
+  'software',
+  'different',
+];
+
+const CONTACT_TYPES_MAP = CONTACT_TYPES.reduce((accumulator, contactType) => {
+  accumulator[contactType] = contactType;
+  return accumulator;
+}, {});
```
gregorbg, line 29:
```suggestion
const CONTACT_TYPES_MAP = _.keyBy(CONTACT_TYPES);
```

gregorbg: Normally, you would have to pass in a second argument, which is a callback that determines the key to use. But if you skip that argument, it will use "identity" function as a default: https://lodash.com/docs/4.17.15#keyBy

gregorbg (changes requested):

app/webpacker/components/ContactsPage/ContactForm.jsx
gregorbg: Yup

app/webpacker/components/ContactsPage/ContactForm.jsx
```diff
@@ -0,0 +1,123 @@
...
+                margin: '1.5px 0px',
+              }}
+              onChange={(_, { value }) => {
+                setFormValues({
+                  userData: formValues.userData,
+                  contactType: value,
+                  [value]: {
+                    competition: null,
+                    message: '',
+                  },
```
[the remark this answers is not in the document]
gregorbg: See my other remark about "overloading" the `formState` variable too much. Changing this according to my other comment will _probably_ solve this problem.

If it doesn't, please check in the clean code that behaves weirdly and I will have a look.

danieljames-dj (commented):
Should be good now, thanks for the review @gregorbg :-)

gregorbg (approved):

PR #9198, opened 2024-04-08
Some cleanups related to WST to help teams-committees migration

danieljames-dj:


danieljames-dj:
@gregorbg I tried rerunning CI multiple times but getting error in `spec/features/competition_management_spec.rb`. Any idea if it's known issue?

gregorbg:
I can see that the spec which is failing is "signed in as admin". What kind of admin is it, and could it be that this "admin" is generated based on WST?

app/models/user_group.rb
```diff
@@ -129,6 +129,10 @@ def self.council_group_wac
     UserGroup.find_by(metadata_id: GroupsMetadataCouncils.find_by(friendly_id: 'wac').id, metadata_type: 'GroupsMetadataCouncils')
   end
 
+  def self.teams_committees_group_wst
+    GroupsMetadataTeamsCommittees.find_by(friendly_id: 'wst').user_group
+  end
```
gregorbg, line 134: Where is this group (metadata) even coming from? I thought we had previously only migrated the empty groups, not the "TeamMember" roles themselves...

app/models/user.rb
```diff
@@ -498,8 +498,8 @@ def results_team?
     team_member?(Team.wrt)
   end
 
-  def software_team?
-    team_member?(Team.wst)
+  private def software_team?
```
gregorbg, line 501: Why are you changing this to private but not the WST Admin team below?

gregorbg (commented):
I don't understand the necessity or the context of this change at all, seeing that all other teams are still yet to be migrated (and I thought even WST memberships have not been migrated yet). Please explain.

Also, what about the WST Admin team?

app/models/user_group.rb
danieljames-dj: We had created the user_groups, but they are connected with converted team roles. So this will work here.

danieljames-dj (commented):
This PR mainly aims in converting the usages of `Team.wst` (there might be couple of related cleanups as well). This is one of the sub-task of teams-committees migration. The teams-committeees migration is very huge because of lots of unit-test changes. So I'm trying to change gradually so that it will be easy to debug if some problem occurs. One potential example is the unit test that is failing now.

danieljames-dj:
> I can see that the spec which is failing is "signed in as admin". What kind of admin is it, and could it be that this "admin" is generated based on WST?

I don't know actually. I think my changes are not related to that, but not 100% sure.

danieljames-dj:
Good news, this error got fixed after migration to roles: https://github.com/thewca/worldcubeassociation.org/pull/9342. I'll close this PR once that PR is merged.

PR #9205, opened 2024-04-08
Allow only organizers and delegates to issue refunds

DanielEgdal:
Too many users on the WCA website currently have access to mass-refund registrations for competitions. The motivation is to limit this to solely organizers and delegates of a competition.

I could not see a method equivalent to the one I created in `user.rb` which surprised me. Lmk if something like that exists.


gregorbg:
Thanks for your submission! There is `competition.managers` which is a list of organizers + delegates. So the whole flow can be simplified to `some_random_competition.managers.include?(some_user)`.

gregorbg:
With that being said, I'm not a fan of restricting features to **only** competition organizers and Delegates. We generally try to use these permission methods because they also give Board members and WST Admins access to everything (where "WST Admin" is a small subgroup of WST, similar to Senior Members.)

gregorbg:
Looking at the source code of `can_manage_competition?`, I can understand that you want to keep the "broader" public of WRC, WEC and Senior Delegates (which are all included in the aforementioned method) out of randomly refunding. But I'd argue that on top of the actual competition staff, at _least_ WST Admin should have privileges.

WRT probably should also keep their privileges, because they have master database access anyways and can already do pretty much anything they want (in theory). All Stripe credentials and account connections are stored in the DB so even if we restrict "convenient" UI access for them, they still could fetch all relevant credentials from our DB and make the Stripe API calls themselves. So it doesn't really make sense to lock them out, apart from the fact that you could argue locking them out from the convenience of having a nice UI reduces the "attack vector". But I'm not inclined to start distrusting people whom we give full read/write access to our DB in this way.

So perhaps `can_admin_competitions? || competition.managers.include?(some_user)` is a middle ground here?

EdHollingdale:
Hi Gregor - I'd in part prompted this change when I was speaking to Daniel more broadly about permissions. I think there needs to be a broader review of website permissions but refund ability (given the financial impact) is the most material.  There are many organisations with tens or hundreds of thousands of dollars worth of registration fees for upcoming competitions through the WCA website and this represents a significant risk.

In general, we should limit the access to just those who need it, and further limit ease of doing actions. Obviously WCAT, WEC, WRC shouldn't be able to just refund registration fees. I get that while WST admins and WRT have database access and could do it anyway, my view is that we shouldn't make it accessible through a nice UI. My biggest concern isn't on say an individual WRT member, but on someone gaining access to their credentials and having easy ability to do things (e.g. in the case of a data breach, or someone leaving their laptop open at a competition). 

In the event that non organisers/delegates have legitimate need to make refunds out of someone else's stripe balance and they want a nice UI to do it from we can add them as admins of the WCA's stripe platform account. But as I think those legitimate uses are nil I don't think it is necessary. I don't think this is distrust of people we give DB read/write access too, but it is simply reducing risk.

gregorbg:
Hmm... I can clearly see your intention and I appreciate the sense to detail as always.

But if you're concerned about a WRT member's credentials leaking, and someone abusing their access, then shouldn't this same logic apply to an organizer's credentials leaking, and abusing their (legitimate) access to the refunds section?

Also, I feel strongly that at least WST Admins (a trusted inner circle of WST members) need to maintain superuser access to the whole website, at least to help with debugging and solving issues. We cannot figure out why a refund breaks or yields an error if we cannot even access the refunds page to diagnose.

DanielEgdal:
> But if you're concerned about a WRT member's credentials leaking, and someone abusing their access, then shouldn't this same logic apply to an organizer's credentials leaking, and abusing their (legitimate) access to the refunds section?

I think the level of potential damage between the two is uncomparable. If someone gains access to a WRT/WCAT/WRC member account, they can cause damage to hundreds of competitions which are impossible to 'roll-back'. An organiser account leaking does not expose nearly as many competitions.

> Also, I feel strongly that at least WST Admins (a trusted inner circle of WST members) need to maintain superuser access to the whole website, at least to help with debugging and solving issues. We cannot figure out why a refund breaks or yields an error if we cannot even access the refunds page to diagnose.

I am fine with having WST admins have this access. A decrease from ~50 people to ~5 people having this elevated access is still a good improvement from the current statement.

gregorbg:
> I think the level of potential damage between the two is uncomparable. If someone gains access to a WRT/WCAT/WRC member account, they can cause damage to hundreds of competitions which are impossible to 'roll-back'. An organiser account leaking does not expose nearly as many competitions.

I definitely agree that "the level of potential damage between the two is uncomparable". But my point was that the _risk_ of such a leak happening in the first place is identical. So the argument about which roles to give access is not affected by website permission considerations, because (to be blunt) no matter who we end up giving access they all have the same probability of getting their credentials leaked anyways. So from that standpoint it simply doesn't make a difference.

> I am fine with having WST admins have this access. A decrease from ~50 people to ~5 people having this elevated access is still a good improvement from the current statement.

There is `some_user.admin?` to check for WST admin privileges. This method returns true if and only if someone is a member of the corresponding WST Admin group on prod. No other team memberships (like WRT or WCAT, or even Board) grant `admin?` privileges. So you can use that in your PR.

DanielEgdal:
>So the argument about which roles to give access is not affected by website permission considerations, because (to be blunt) no matter who we end up giving access they all have the same probability of getting their credentials leaked anyways. So from that standpoint it simply doesn't make a difference.

My point was that it is a good idea to limit **how many** people have this access. You are right that the likelihood of a leak per account is probably equally likely. But if we increase the amount of people x10 with this access for every single competition, then we do see a massive increase in the risk (albeit, still a low risk, but why increase it?). In general with this kind of sensitive issue, I would think that people who do not need to use the feature should not be able to use it, as it is an unnecessary risk.

I have updated the PR to use some of the predefined functions such that competition managers and WST admins can still use the feature.

gregorbg (approved):

PR #9206, opened 2024-04-09
Venue coordinate/map fixes

kr-matthews:
- Prevent crash when deleting all characters from a coordinate.
- Auto-center the map on the venue marker whenever it moves.
- Rename some constants which briefly slightly confused me.

For some reason, the map geo search doesn't work when I run locally, so I can't check that searching for a location and selecting it still works as expected. But dragging the marker, or directly updating the coordinates, both work as expected for me.

There's still the problem where you can't really type a coordinate from scratch because decimals and extra 0s disappear as soon as you type them. But pasting in a coordinate, or back-filling decimals, negatives, and zeros can also work. I don't have a clean solution yet so I'll save that for a later PR.

### DIFF
--- a/app/webpacker/components/EditSchedule/EditVenues/VenuePanel.js
+++ b/app/webpacker/components/EditSchedule/EditVenues/VenuePanel.js
@@ -36,9 +36,7 @@ function VenuePanel({
   const confirm = useConfirm();
 
   const handleCoordinateChange = (evt, { name, value }) => {
-    dispatch(
-      editVenue(venue.id, name, Number.isNaN(toMicrodegrees(value)) ? 0 : toMicrodegrees(value)),
-    );
+    dispatch(editVenue(venue.id, name, toMicrodegrees(value)));
   };
 
   const handleVenueChange = (evt, { name, value }) => {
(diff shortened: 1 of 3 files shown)


app/webpacker/components/EditSchedule/EditVenues/VenuePanel.js
```diff
@@ -36,7 +36,9 @@ function VenuePanel({
   const confirm = useConfirm();
 
   const handleCoordinateChange = (evt, { name, value }) => {
-    dispatch(editVenue(venue.id, name, toMicrodegrees(value)));
+    dispatch(
+      editVenue(venue.id, name, Number.isNaN(toMicrodegrees(value)) ? 0 : toMicrodegrees(value)),
```
gregorbg, line 40: Could this logic move into the `toMicrodegrees` function?
I understand the reason behind your approach, and your solution is fine except for the fact that right now it's too densely packed into one line. (And you're calling `toMicrodegrees` twice, that seems redundant)

gregorbg (changes requested):

app/webpacker/components/EditSchedule/EditVenues/VenuePanel.js
gregorbg: If you find a better place for the input sanitization, feel free. I just don't want it to happen right inside the setter fn call

gregorbg (approved):

PR #9207, opened 2024-04-09
Reuse/unify controller and view code for Competition Payment Integration setup

gregorbg:
Trying to get rid of a lot of copy/pasted code between Stripe and PayPal.

**This PR does not contain payment logic!** It only concerns the "Connect/Disconnect your account" setup logic.


dunkOnIT:
Test cases:
- [ ] Connect stripe (expect success)
- [ ] Disconnect stripe (expect success)
- [ ] Connect paypal (expect success)
- [ ] Disconnect paypal (expect success)
- [ ] Connect stripe, attempt to connect paypal (expect failure)
- [ ] Connect paypal, attempt to connect stripe (expect failure)
- [ ] Connect stripe, make payment (expect success)
- [ ] Connect paypal, make payment (expect success)

app/models/competition.rb
```diff
@@ -2509,12 +2509,23 @@ def payments_enabled?
...
+    raw_types.map { |type| CompetitionPaymentIntegration::AVAILABLE_INTEGRATIONS.invert[type] }
+  end
+
+  def payment_integration_connected?(integration_name)
+    CompetitionPaymentIntegration.validate_integration_name!(integration_name)
+
+    self.competition_payment_integrations.send(integration_name).exists?
+  end
+
   def stripe_connected?
```
dunkOnIT, line 2523: Do we still need these if we could just call `competition.payment_integration_connected?('stripe') and have a `.to_sym` sanitation?

app/models/competition.rb
```diff
@@ -2509,12 +2509,23 @@ def payments_enabled?
...
-    competition_payment_integrations.stripe.exists?
+    self.payment_integration_connected?(:stripe)
   end
 
   def paypal_connected?
```
dunkOnIT, line 2527: Same question as above

app/models/competition.rb
```diff
@@ -2527,7 +2538,10 @@ def payment_account_for(integration_name)
 
   def disconnect_payment_integration(integration_name)
     CompetitionPaymentIntegration.validate_integration_name!(integration_name)
-    competition_payment_integrations.destroy_by(connected_account_type: CompetitionPaymentIntegration::AVAILABLE_INTEGRATIONS[integration_name])
+
```
dunkOnIT, line 2541: extra line here is a stylistic choice?

app/views/competitions/_nav.html.erb
```diff
@@ -43,7 +43,7 @@
               { text: t('.menu.orga_view'), path: edit_competition_path(@competition), icon: "lock" },
               { text: t('.menu.event_view'), path: edit_events_path(@competition), icon: "cubes" },
               { text: t('.menu.schedule_view'), path: edit_schedule_path(@competition), icon: "calendar" },
-              { text: t('.menu.payment_view'), path: competitions_payment_setup_path(@competition), icon: "cc stripe" },
+              { text: t('.menu.payment_view'), path: competition_payment_integration_setup_path(@competition), icon: "cc stripe" },
```
dunkOnIT, line 46: Do we want to take this opportunity to change the icon away from the stripe icon? (I have it on my backlog checklist so not a nit we're in danger of losing)

app/models/competition.rb
gregorbg: more like readability choice, but yes.

app/views/competitions/_nav.html.erb
gregorbg: sure! Changed it to a money bill, which is more intuitive and less tightly coupled to Stripe

app/models/competition.rb
gregorbg: Actually, a symbol makes sense in this context because it expresses that whatever you're passing is a static/frozen constant. Strings are too "mutable", if you get what I mean

gregorbg: Can add a sanitizer nonetheless

dunkOnIT:
**Review Comments:** (WIP)

- [ ] Should organizers be able to see "Setup Payments" after the comp is announced? At the moment we show the below pictured text when no payment method is connected, which is a bit haphazard.
 - I'd suggest some variation of the following: 
   - When a CPI is connected & comp is announced, **organizers and delegates** can see that stripe/paypal is connected with the "Disconnected" button greyed out (long-term, it would be nice to use the accounts API to show the name of the connected account)
   - When a CPI is **not** connected & comp is announced, **organizers** don't have the "connect payments" menu option"
image

dunkOnIT:
Is that information re the disconnect process needed when no payment method is connected, though? TL;DR of my proposal is not to show payment options to organizers when payments aren't setup (ie, when that info is irrelevant)

app/models/competition.rb
```diff
@@ -2525,12 +2525,23 @@ def payments_enabled?
...
 
+  def connected_payment_integration_types
+    raw_types = self.competition_payment_integrations.pluck(:connected_account_type)
+    raw_types.map { |type| CompetitionPaymentIntegration::AVAILABLE_INTEGRATIONS.invert[type] }
+  end
+
+  def payment_integration_connected?(integration_name)
+    CompetitionPaymentIntegration.validate_integration_name!(integration_name)
+
+    self.competition_payment_integrations.send(integration_name.to_sym).exists?
```
dunkOnIT, line 2536: If I'm understanding correctly, the `send` here is calling a scope on `competition_payment_integrations` (eg, competition_payment_integrations.paypal)? And then we do `exists?` to check if there's anything returned by that scope

dunkOnIT (approved):
LTGM, question about `.send` is for my own understanding, and visibility of payment setup to organizers once comp is announced is up to you, but probably not worth holding up the PR over?

app/models/competition.rb
gregorbg: > If I'm understanding correctly, the `send` here is calling a scope on `competition_payment_integrations` (eg, competition_payment_integrations.paypal)?

This part is correct

> And then we do `exists?` to check if there's anything returned by that scope

This is only half-correct. There will always be something returned by the scope (as long as the method name referencing the scope exists in the first place, which we make sure through `validate_integration_name!`). Then calling the `exists?` method checks whether a *non-zero amount of entries* exists. In other words, whether there is an integration of the said scope type in the database.

PR #9208, opened 2024-04-09
Unify PayPal and Stripe payment controller logic

gregorbg:
Turns PayPal into a first-class citizen of the WCA payment gateway ecosystem.

Massive changes to the code, but actually not that massive changes to the flow. Mostly just shuffling code around. Will review in a session with Duncan and Finn.

Table column renames are tehcnically optional but I feel like they make sense to harmonise Stripe and PayPal as good as possible.


app/models/connected_paypal_account.rb
```diff
@@ -2,4 +2,50 @@
...
+    PaymentIntent.create!(
+      holder: registration,
+      payment_record: paypal_record,
+      client_secret: paypal_record.paypal_id,
+      initiated_by: paying_user,
+      wca_status: paypal_record.determine_wca_status,
+    )
+  end
+
+  def capture_charges(payment_intent)
```
dunkOnIT, line 24: The naming of this function is confusing because of it uses both "capture" and "charge". It being used across both payment providers is also confusing, because in Paypal we have to initiate the capture, and in Stripe we only have to record the charge - it implies an equivalence of functionality which I don't think exists

app/controllers/registrations_controller.rb
```diff
@@ -666,16 +735,18 @@ def refund_payment
       return redirect_to edit_registration_path(registration)
     end
 
-    # Backwards compatibility: We may at some point try to record a refund for a payment that was
-    #   - created before the introduction of receipts
-    #   - but refunded after the new receipts feature was introduced. Fall back to the old stripe_charge_id if that happens.
-    charge_id = payment.receipt&.stripe_id || payment.stripe_charge_id
+    if payment.receipt.nil?
```
dunkOnIT, line 738: Is this even a case we need to handle? (ie, is it just bloating our controller?)

app/controllers/registrations_controller.rb
```diff
@@ -666,16 +735,18 @@ def refund_payment
...
+      # Backwards "compatibility": Before the introduction of receipts, in an old age when we were
+      #   hard-wired to a much older version of the Stripe SDK, we stored charge IDs directly in a legacy column.
+      #   This was so, so long ago that nobody should try and refund one of these payments anyways.
+      flash[:danger] = "The payment you're trying to refund is too old. Please refund from your Stripe dashboard directly."
+      return redirect_to edit_registration_path(registration)
+    end
 
-    refund_receipt = stripe_integration.issue_refund(charge_id, refund_amount)
+    refund_receipt = payment_account.issue_refund(payment.receipt, refund_amount)
```
dunkOnIT, line 746: `refund_receipt` is a bit confusing - perhaps `refund_record`? (as `issue_refund` seems to be returning a PaymentRecord object)

app/controllers/registrations_controller.rb
```diff
@@ -644,15 +710,18 @@ def load_payment_intent
...
+
+    unless payment_account.present?
+      flash[:danger] = "You cannot emit refund for this competition anymore. Please use the integration's dashboard to do so."
       return redirect_to edit_registration_path(registration)
     end
 
-    payment = RegistrationPayment.find(params[:payment_id])
-
     refund_amount_param = params.require(:payment).require(:refund_amount)
     refund_amount = refund_amount_param.to_i
```
dunkOnIT, line 726: Why is it not just:
```ruby
refund_amount = params.require(:payment).require(:refund_amount).to_i
```

(I know it's not part of the diff, but may as well clean it up while we're here)

app/controllers/registrations_controller.rb
```diff
@@ -666,16 +735,18 @@ def refund_payment
...
+      flash[:danger] = "The payment you're trying to refund is too old. Please refund from your Stripe dashboard directly."
+      return redirect_to edit_registration_path(registration)
+    end
 
-    refund_receipt = stripe_integration.issue_refund(charge_id, refund_amount)
+    refund_receipt = payment_account.issue_refund(payment.receipt, refund_amount)
 
-    # Should be the same as `refund_amount`, but by double-converting from the Stripe object
+    # Should be the same as `refund_amount`, but by double-converting from the Payment Gateway object
     # we can also double-check that they're on the same page as we are (to be _really_ sure!)
```
dunkOnIT, line 749: Should we just do the check and then we can remove these 2 lines of comments?

dunkOnIT:
Mostly nits sp far, will chew on the payment logic with a clear head tomorrow

PR #9211, opened 2024-04-10
Update Shakapacker configuration to give us basic SSR capabilities

gregorbg:
Long story short, our config that hooked in `webpack` through the `Shakapacker` gem was terribly, terribly outdated.
Handling it this way follows the `Shakapacker` README and coincidentally gives us the capability of using server-side rendering capabilities of `react-rails`.

It's not used in practice yet, and I cannot figure out why `refresh-react/babel` blows up. But seeing that we didn't even use this plugin at all before, I'm fine leaving it commented out for somebody else to figure out.

### DIFF
--- a/app/webpacker/packs/server_rendering.js
+++ b/app/webpacker/packs/server_rendering.js
@@ -1,0 +1,5 @@
+// By default, this pack is loaded for server-side rendering.
+// It must expose react_ujs as `ReactRailsUJS` and prepare a require context.
+var componentRequireContext = require.context("components", true);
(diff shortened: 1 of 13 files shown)


gregorbg:
Blocked because of a transitive dependency issue on @pmmmwh/react-refresh-webpack-plugin

gregorbg:
Unblocked by simply abandoning the `react-reload` integration. Currently doesn't work well for us anyways because of the `js` vs `jsx` confusion.

gregorbg:
Needed for another PR

PR #9212, opened 2024-04-11
Renamed candidate_delegate to junior_delegate

danieljames-dj:


db/migrate/20240411113530_rename_candidate_to_trainee.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class RenameCandidateToTrainee < ActiveRecord::Migration[7.1]
+  def change
+    RolesMetadataDelegateRegions.where(status: 'candidate_delegate').each do |metadata|
+      metadata.update!(status: RolesMetadataDelegateRegions.statuses[:junior_delegate])
+    end
```
gregorbg, line 7: You can do a bulk update like so: `RolesMetadataDelegateRegions.where(status: 'old').update_all(status: 'new_status')`. This directly translates to a `UPDATE table SET status = 'new' WHERE status = 'old'` query, instead of issung an `UPDATE WHERE id` style query one by one

gregorbg (changes requested):

db/migrate/20240411113530_rename_candidate_to_trainee.rb
danieljames-dj: Oh okay, done. Thanks.

gregorbg (approved):

PR #9215, opened 2024-04-11
Render Registration V2 Panel in old Frontend

FinnIckler:
There is still a lot that can be done to improve this integration. I basically just copy pasted the old code with Rails Routes until it rendered correctly:

- [x] Replace placeholders with correct routes for payments
- [x] Correctly integrate openapi-fetch (I think that requires typescript though)
- [x] Rework folder structure
- [x] Get rid of `externalServiceFetch`
- [x] Correctly set up i18n React
- [x] Replace the Event picker with the one in the new competitions page
- [x] Get rid of the gigantic amount of eslint errors

Would like some feedback on the folder structure


app/webpacker/components/CompetitionsOverview/CompetitionsFilters.js
```diff
@@ -27,7 +27,8 @@ function CompetitionsFilters({
       <Form.Field>
         <EventSelector
           selectedEvents={filterState.selectedEvents}
-          dispatchFilter={dispatchFilter}
+          onEventSelection={dispatchFilter}
+          eventList={WCA_EVENT_IDS}
```
gregorbg, line 31: Can you introduce this as a default value in `EventSelector`'s props or so?
The most common use-case is to be able to pick from ALL events (competition overview, personal records page, rankings pages, ...) so it feels redundant to always import a list of official events and pass it along.

app/webpacker/components/RegistrationsV2/Context/registration_context.jsx
```diff
@@ -0,0 +1,7 @@
+import { createContext } from 'react';
+
+export const RegistrationContext = createContext({
+  registration: null,
+  refetch: () => {},
```
gregorbg, line 5: Not super happy about a method being part of a context, but this is part of a bigger discussion.

app/webpacker/components/RegistrationsV2/Register/RegistrationRequirements.jsx
```diff
@@ -0,0 +1,102 @@
...
+  const [regRequirementsAcknowledged, setRegRequirementsAcknowledged] = useState(false);
+
+  const [showRegRequirements, setShowRegRequirements] = useState(false);
+
+  const { t } = useTranslation();
+
+  const setFromCheckbox = (data, setState) => {
+    const { checked } = data;
+    setState(checked);
+  };
```
gregorbg, line 30: I think we have hooks for that in the monolith React code, under `webpacker/lib/hooks`

app/webpacker/components/RegistrationsV2/Register/RegistrationRequirements.jsx
```diff
@@ -0,0 +1,102 @@
...
+
+export default function RegistrationRequirements({ nextStep }) {
+  const { competitionInfo } = useContext(CompetitionContext);
+
+  const [generalInfoAcknowledged, setGeneralInfoAcknowledged] = useState(false);
+  const [regRequirementsAcknowledged, setRegRequirementsAcknowledged] = useState(false);
+
+  const [showRegRequirements, setShowRegRequirements] = useState(false);
+
+  const { t } = useTranslation();
```
gregorbg, line 25: Not ready for React-i18n yet.

app/webpacker/components/RegistrationsV2/Register/RegistrationRequirements.jsx
```diff
@@ -0,0 +1,102 @@
...
+      setShowRegRequirements(true);
+    }
+  }, [generalInfoAcknowledged, setShowRegRequirements]);
+
+  return (
+    <Segment basic>
+      <Form onSubmit={nextStep}>
+        <Form.Checkbox
+          checked={generalInfoAcknowledged}
+          onClick={(_, data) => setFromCheckbox(data, setGeneralInfoAcknowledged)}
```
gregorbg, line 57: If you use the custom hook instead of `useState` hook (see comment above) you can directly pass in the setter and the hook logic will figure everything out by itself.

app/webpacker/components/RegistrationsV2/Register/RegistrationRequirements.jsx
```diff
@@ -0,0 +1,102 @@
...
+              duration={500}
+            >
+              <Accordion.Content active={showRegRequirements}>
+                <p
+                  dangerouslySetInnerHTML={{
+                    __html: marked(
+                      competitionInfo.extra_registration_requirements,
+                    ),
+                  }}
+                />
```
gregorbg, line 81: I think we have shared components for Markdown, if not this is probably a really good time to introduce them.

app/webpacker/components/RegistrationsV2/Register/RegistrationRequirements.jsx
```diff
@@ -0,0 +1,102 @@
...
+                <Message positive>
+                  <Form.Checkbox
+                    checked={regRequirementsAcknowledged}
+                    onClick={(_, data) => setFromCheckbox(data, setRegRequirementsAcknowledged)}
```
gregorbg, line 85: `useCheckboxState` custom hook, see above.

app/webpacker/components/RegistrationsV2/Register/StepPanel.jsx
```diff
@@ -0,0 +1,69 @@
...
+  key: 'payment',
+  i18nKey: 'competitions.registration_v2.tabs.payment',
+  component: StripeWrapper,
+};
+
+export default function StepPanel() {
+  const { competitionInfo } = useContext(CompetitionContext);
+  const { isRegistered } = useContext(RegistrationContext);
+
+  const { t } = useTranslation();
```
gregorbg, line 30: No React-i18n yet, see above.

app/webpacker/components/RegistrationsV2/api/auth/get_jwt.js
gregorbg, line 1: Quiz question: Do we even need JWT authentication now that the frontend lives "close" to the monolith again?
Other components do backend fetch requests by "stealing" the CSRF token, see `webpacker/lib/requests/fetchWithAuthenticityToken`. That _should_ be enough to authenticate against the backend in this context, or am I missing something?

(Obviously there is a larger discussion to be had about re-introducing JWT auth when moving to a fully independent NextJS backend, but as part of this current re-assimilation effort I'd rather stick with what we have in the existing React codebase and _then_ introduce new standards step by step)

app/webpacker/components/RegistrationsV2/api/helper/error_codes.js
gregorbg, line 1: The competition form introduced a very similar thing with `webpacker/lib/requests/FetchJsonError`, can you reuse/unify it with this code?

app/webpacker/components/RegistrationsV2/api/registration/get/get_registrations.js
gregorbg, line 1: Again, philosophical discussion about the benefits of openapi-fetch when we're not even using TypeScript (yet).
I strongly believe that the microservice should maintain its own openapi-schema (as should the monolith, eventually) but the frontend consuming data in pure JS can simply just "ignore" the fact that there is typing information available and stick to a more consistent approach with what we have already.

app/webpacker/components/RegistrationsV2/messages/LoadingMessage.jsx
gregorbg, line 1: There's `webpacker/components/Requests/Loading.js` already

app/webpacker/lib/i18next.js
gregorbg, line 1: Excited about react-i18n but not excited about introducing it in this specific PR. See also my other comments.

app/webpacker/components/RegistrationsV2/ui/events/messages.js
gregorbg, line 1: This clashes with how we're doing things in the existing React stack. We don't have an event bus anywhere; we just always have an "error" state and if that is null, we don't render anything. If an error occurs, we set that to some not-null value and React re-renders the DOM with the error message active.

gregorbg (commented):
Happy with the folder structure and happy with the relative ease of switching between the two integrations. Good job for coming up with something so quickly!

Going through the specifics of the PR, I realised that the "new new" frontend clashes with a lot of conventions / libraries we use in the "new" (i.e. monolith React) frontend. Namely,
- react-i18n gem
- OpenAPI fetch
- JWT auth
- Event Bus for error messages/flashes

While the first three are in principle good changes that we should adapt, this single PR would massively blow out of proportion if we introduced them all. I am suggesting a migration path where we copy the components and then adjust them to use what we already have (existing monolith i18n approach, existing monolith fetch auth, etc.) and then later introduce these other improvements step by step.

app/webpacker/components/RegistrationsV2/Register/RegistrationRequirements.jsx
```diff
@@ -0,0 +1,102 @@
+import { UiIcon } from '@thewca/wca-components';
```
gregorbg, line 1: Addendum: I would like to avoid the use of `@thewca/wca-components` if its only purpose is the `UiIcon` thingy.
(In general, I am not too happy with the notion of `UiIcon` overall, because it ties `wca-components` to SemUI, which is not cool... But that's a whole different discussion to be had.)
Actionable item in terms of this review: Use SemUI `Icon` component instead and get rid of the `wca-components` dependency in this PR altogether.

package.json
```diff
@@ -26,7 +26,11 @@
     "@fullcalendar/react": "^6.1.11",
     "@fullcalendar/timegrid": "^6.1.11",
     "@rails/ujs": "^7.1.3",
+    "@stripe/react-stripe-js": "^2.7.0",
+    "@stripe/stripe-js": "^3.3.0",
     "@tanstack/react-query": "^5.28.9",
+    "@thewca/wca-components": "^0.6.0",
```
gregorbg, line 32: I would prefer not to include `wca-components` at this point in time. See my lengthier comment above. Is there any use case in this PR other than `UiIcon` and `CubingIcon`?

FinnIckler:
I think I got everything we talked about?

app/webpacker/components/Markdown.jsx
```diff
@@ -0,0 +1,23 @@
...
+    isLoading,
+  } = useQuery({
+    queryKey: ['markdown', id],
+    queryFn: () => window.wca.renderMarkdownRequest(md),
+    staleTime: Infinity,
+  });
+
+  return isLoading ? <Loading /> : (
+    // eslint-disable-next-line react/no-danger
+    <span dangerouslySetInnerHTML={{ __html: html }} />
```
gregorbg, line 21: We have sanitizers, just to be safe. See `webpacker/components/I18nHTMLTranslate.js`

gregorbg:
Yay! Just one tiny remark about the Markdown component

gregorbg (approved):
Approved as part of a large "move over" effort from the other codebase's frontend.
There is a larger picture here, so some parts of this may not be perfect. They will be once all follow-up PRs are done.

PR #9216, opened 2024-04-12
Update WCA logo usages to include ® symbol

nsilvestri:
The Logo Usage Policy states that the registered trademark symbol (®) be used on the WCA logo in all jurisdictions. All modified versions of the logo, including versions with borders and 3D effects, are not covered by our trademark and therefore must be replaced with the registered version (note that color is not included in our trademark claim).

Removal of the logotype in the file `WCAlogo.svg` is for consistency in distributed file types at this moment. Versions with a logotype will be added back in a separate future PR.

Variation files were deleted from `public/files` because we do not want to be distributing versions of our logo which are not covered by our trademark policy.

I have only updated the English locale for this. This will break links on the logo page when using non-English locales; **this is intentional** to prevent any logo distribution of now-unauthorized logos.

This PR is related to #8336 regarding previous discussion on removing unofficial versions of the WCA logo. While concerns about logo readability are valid, legal requirements regarding our trademark and how we use our own logo force us to use the version that we have registered with USPTO.


app/assets/images/android-touch-icon.png
jfly, line 1: There's more going on here than just a ® symbol getting added. I see black borders getting removed.

I see this was discussed briefly on https://github.com/thewca/worldcubeassociation.org/pull/8336, but speaking as an outsider, I can't tell where that conversation landed. My suggestion:

- Either add more context to your original post explaining why the borders and changing
- Or, keep the black borders for now and make that change in a subsequent PR with more context

jfly (commented):
> The Board has requested that the registered trademark symbol (®) in all jurisdictions for the WCA logo.

nit: I think there's a missing word in this sentence.

app/assets/images/android-touch-icon.png
nsilvestri: Nit fixed, accidentally a word.

The reasoning for the border changes was already included in the main post but I've updated the body to be more explicit about the reason. Let me know what you think about the new PR text.

jfly: Much better, thanks!

dunkOnIT:
Have:
- reviewed the code changes to ensure that all updated logo instances correspond to an actually-existing filename
- check locally that the logo renders as expected
- checked the WCA Logo page & links to ensure they all work as expected.

on that basis, it LGTM, but will let @gregorbg approve as it is a significant change.

Also, I pushed a minor change, which updates the `WCA Logo` page from looking like this: 
image

To this: 
image

@nsilvestri happy to revert if you'd prefer. It's arguably out-of-scope as you didn't introduce it, but I figure let's correct this small mistake while we're making changes to the logo page.

nsilvestri:
Changes are fine with me! My to-do list (read: wishlist) include overhauling this page in the future anyway, so minor adjustments like these don't bother me.

gregorbg (approved):
Approved in terms of changes to the program code. I did not check every single SVG and PNG file submitted in this PR, as WMT has probably done so many times.

PR #9217, opened 2024-04-14
Speed up Developer Dump import on Docker MySQL server

gregorbg:
This is based largely on the ideas from two StackExchange posts:
- https://dba.stackexchange.com/a/293790
- https://dba.stackexchange.com/a/98815

Using these improvements, my local machine import time for a fresh dev dump went down from ~1h20m to ~11m (yes, that is more than 700% speedup). The "largest" parameter in use here is 4 gigabytes of RAM; I feel that in 2024 this is reasonable to demand from a developer machine. Especially since this parameter specifies a *maximum* amount of memory that InnoDB *may* allocate. It does not mean that 4GB will be fully blocked at all times while the DB server is running; rather it specifies that a maximum of up to 4GB may be used if necessary.

**IMPORTANT**: It is notoriously hard to debug _exactly_ what is going on under the hood when running the import from a MySQL dump file. The graphic at the beginning of [this DBA answer](https://dba.stackexchange.com/a/83385) shows that InnoDB memory management is actually arcane, so most of the exact values in the `my.cnf` are directly stolen from the recommendations in the answers linked above. It may well be possible that more fine-tuning can speed up the import process even more, but this involves a lot of guesswork (tuning some knobs a little bit, running a full import, and see if things improved further) while for the time being 11 minutes already seems alright.
If you have any special DB engine expertise and can provide some insight / further tuning, please feel free. This also applies if you think "It does not make sense to set parameter X to 4 GB when parameter Y is only limited to 512MB" or similar. Please feel free to criticise the technicalities if you have special knowledge.


dunkOnIT (approved):
The 4gb default requirement seems entirely reasonable to me - if this were ever an issue for somebody, it seems they could always tweak `my.cnf` anyway?

gregorbg:
> The 4gb default requirement seems entirely reasonable to me - if this were ever an issue for somebody, it seems they could always tweak `my.cnf` anyway?

Basically, yes. However this would create a diff that they would constantly have to pay attention not to commit

dunkOnIT:
> Basically, yes. However this would create a diff that they would constantly have to pay attention not to commit
True enough, but I think this is not a major concern
- unless they're me, devs shouldn't need to load the dev dump too often
- having a dev running docker with a machine that doesn't have the necessary RAM to run this is already pretty unlikely

PR #9218, opened 2024-04-14
Ignore CAD data during dev dump

gregorbg:
See title.

The database sanitizer rules for the CAD tables were dated February 2017, at which time we were still running lots of PHP. Now we even have a very simple (in terms of running/starting the thing) Ruby script which didn't exist back then, so I feel like we can remove these auxiliary tables and save filesize and import time.

**IMPLICATION**: Developers who are working on Ranking Tables or the Persons page will need to run CAD on their machines individually. Where should we document this?

Sindenote: This does **not** impact the "Results Dump". That one uses separate sanitizers where the table data is maintained.

### DIFF
--- a/lib/database_dumper.rb
+++ b/lib/database_dumper.rb
@@ -150,40 +150,8 @@ module DatabaseDumper
         },
       ),
     }.freeze,
-    "ConciseAverageResults" => {
-      where_clause: "",
-      column_sanitizers: actions_to_column_sanitizers(
-        copy: %w(
-          average
-          continentId
-          countryId
-          day
-          eventId
-          id
-          month
-          personId
-          valueAndId
-          year
-        ),
-      ),
-    }.freeze,
-    "ConciseSingleResults" => {
-      where_clause: "",
-      column_sanitizers: actions_to_column_sanitizers(
-        copy: %w(
-          best
-          continentId
-          countryId
-          day
-          eventId
-          id
-          month
-          personId
-          valueAndId
-          year
-        ),
-      ),
-    }.freeze,
+    "ConciseAverageResults" => :skip_all_rows,
+    "ConciseSingleResults" => :skip_all_rows,
     "connected_paypal_accounts" => {
       where_clause: "",
       column_sanitizers: actions_to_column_sanitizers(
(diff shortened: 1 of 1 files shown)


dunkOnIT:
Where to document: Somewhere in the [Detailed Contributing Guide](https://docs.worldcubeassociation.org/contributing/detailed_contributing_guide.html) perhaps?

gregorbg:
> * What does CAD mean in this context? ComputeAuxiliaryData?

Yes.

> * It looks like this will remove results from the dev dump - am I correct on that? I'd be a bit concerned about breaking associations which rely on Results, especially given that we're set to improve the load time so tremendously from your other PR

Not exactly. The Results themselves are not removed. CAD is an auxiliary data computation routine that "derives" things like Rankings tables and PB/PR indices based on the data from the `Results` table. That table is still part of the dev dump, even with this change.

In short, the result of "import + run CAD job" will yield the same thing as a previous (pre this change) "import".

dunkOnIT:
Ah ok that makes sense! 

What would the steps be for running CAD locally? I can checkout this branch and try the new process locally to get a sense of what it's like as a relative noob (to CAD jobs at least)

gregorbg:
In a `bin/rails console` prompt, run `ComputeAuxiliaryData.perform_now`.

There is also a way of running this in the UI under the "Admin" panel, but this is only accessible depending on the privileges of the currently logged in user (only WRT and WST have access, and maybe the Board(??))

dunkOnIT:
Have tried the CAD job - took about 9 minutes on my machine. _In principle_, it's still pretty fast and easy.

So my main question is: Will this actually save us time after the changes you just made to the import job? And if not, do we still want to do it?

gregorbg:
That's exactly the question that is up for debate here. This change saves space (which is good for storage costs on AWS, bandwidth costs when downloading, and has speed benefits when devs far away from the US west coast want to download our data) and it definitely does save a non-zero amount of runtime.
Whether that change is significant enough to warrant merging this PR, or maybe whether it's negligible enough, is up for debate.

gregorbg:
Discarded after considering the actual row counts

PR #9221, opened 2024-04-15
Remove team leader section from navigation

danieljames-dj:
This is ready for review, but is marked as draft till April 20. I'll merge this after April 20 (if this is approved).

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -518,10 +518,6 @@ class User < ApplicationRecord
     self.current_team_members.select { |t| t.team_id == team.id && t.team_leader }.count > 0
   end
 
-  def teams_where_is_leader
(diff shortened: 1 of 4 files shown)


gregorbg:
Why is this in Draft state?
And is there an easy, convenient qay to access the equivalent "replacement" panel for Leaders?

danieljames-dj:
> Why is this in Draft state? And is there an easy, convenient way to access the equivalent "replacement" panel for Leaders?

Draft because I've informed Leaders that this will be removed after April 20. I'll undraft after that date.
Replacement: yes, they can open 'Leader Panel' from navigation bar where they can find the Groups Manager page to make the changes.

gregorbg (approved):

PR #9223, opened 2024-04-15
Show only root regions in RegionsAdmin

danieljames-dj:


app/models/user_group.rb
```diff
@@ -167,6 +167,10 @@ def unique_status?(status)
     end
   end
 
+  def is_root_group?
+    UserGroup.root_groups.include?(self)
```
gregorbg, line 171: I'm a bit unhappy about this implementation, because it loads a lot of groups fully into memory to check inclusion.
Shouldn't `self.parent_group_id.nil?` be enough?

gregorbg (changes requested):

gregorbg (approved):

PR #9226, opened 2024-04-16
Replace team_member? with active_roles.any? for checking membership

danieljames-dj:


gregorbg (changes requested):
Can you extract this `active_roles.any? { ...do stuff... }` pattern into a (private) method and just call that method with the corresponding params?

Just like before, where there was a `team_member?` method that was called with the different corresponding Teams as params.

danieljames-dj:
> team_member?

Done

gregorbg (approved):

PR #9227, opened 2024-04-16
Some cleanups targeting final role migration

danieljames-dj:


app/jobs/sync_mailing_lists_job.rb
```diff
@@ -7,7 +7,7 @@ class SyncMailingListsJob < WcaCronjob
   end
 
   def perform
-    GsuiteMailingLists.sync_group("leaders@worldcubeassociation.org", TeamMember.current.in_official_team.leader.map(&:user).map(&:email))
+    GsuiteMailingLists.sync_group("leaders@worldcubeassociation.org", UserGroup.teams_committees.map(&:lead_user).compact.map(&:email))
```
gregorbg, line 10: Does this account for the **current** lead users only?

app/models/user.rb
```diff
@@ -65,9 +65,9 @@ def self.eligible_voters
   end
 
   def self.leader_senior_voters
-    team_leaders = TeamMember.current.in_official_team.leader.map(&:user)
+    team_leaders = UserGroup.teams_committees.map(&:lead_user)
```
gregorbg, line 68: Same question as above (feel free to resolve both by just answering one): Does this account for **current** Team Leaders only?

spec/features/eligible_voters_spec.rb
```diff
@@ -15,7 +15,7 @@
       user.team_members.find_by_team_id(wrc_team_id).update!(end_date: 1.day.ago)
     end
   }
-  let!(:team_leader) { FactoryBot.create(:user, :wrc_member, team_leader: true) }
+  let!(:team_leader) { FactoryBot.create(:user, :wrt_member, team_leader: true) }
```
gregorbg, line 18: Why? This change seems quite arbitrary...

spec/requests/users_spec.rb
```diff
@@ -200,7 +200,7 @@
       # WST is not moderator by default, admin status is granted manually in
       # Discourse
       expect(answer_sso.moderator).to be false
-      expect(answer_sso.add_groups).to eq "wst,delegate"
+      expect(answer_sso.add_groups).to eq "delegate,wst"
```
gregorbg, line 203: Might be cool to change the SSO interface to make sure the list of string tags (`wst`, `delegate`, etc.) is always sorted alohabetically. That way, we have a defined behavior of tests and don't need to worry about test chokes if some order of the rows in the DB changes.

gregorbg (changes requested):

spec/requests/users_spec.rb
danieljames-dj: Okay done

gregorbg (approved):

PR #9228, opened 2024-04-16
Add WCA_REGISTRATIONS_POLL_URL Env Variable

FinnIckler:
Adds these to the terraform, I also already changed the Task Definitions and updated staging to the latest one. Should happen automatically for prod with the next deployment


gregorbg (approved):

FinnIckler:
Ok, I ran terraform without changing the `buildspec_build.yml` just so we can deploy again. Gonna confer with Jonatan about if we should change it or not

FinnIckler:
I updated the terraform to be correct with the current status and removed a wrong redirect (not sure why I was trying to redirect www to auxiliary services, probably copy paste typo)

gregorbg (approved):

PR #9230, opened 2024-04-16
Add V2 Registration Administration Frontend

FinnIckler:
This adds Registration Administration for all and one registration. I have a couple issues currently though which I would like some input on:

- Currently the registration edit page is routed to by using a unique registration id from the DB. The new registration service uses attendee ids instead. This will need changes in the controller to not load @registration when v2 is used, is that fine or do we want to redirect?
- There is some jQuery that is loaded on the onPage 'hook' like `onPage('registrations#edit_registrations', function() {` which currently breaks my rendering, is there a way to load this conditionally?

### DIFF
--- a/app/views/registrations/edit_v2.html.erb
+++ b/app/views/registrations/edit_v2.html.erb
@@ -1,6 +1,4 @@
 <% provide(:title, t('registrations.edit_registration.title', person: @user.person.name, comp: @competition.name)) %>
-<% add_to_packs("auto_numeric") %>
-
 <%= render layout: 'nav' do %>
   <%= react_component('RegistrationsV2/RegistrationEdit', { competitionInfo: @competition.to_competition_info ,
                                                             user: @user }) %>
--- a/app/views/registrations/edit_v2.html.erb
+++ b/app/views/registrations/edit_v2.html.erb
@@ -1,0 +1,7 @@
+<% provide(:title, t('registrations.edit_registration.title', person: @user.person.name, comp: @competition.name)) %>
+<% add_to_packs("auto_numeric") %>
+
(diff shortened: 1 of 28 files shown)


FinnIckler:
I do need some of the functions for the follow up public registrations lists PR, so I'll just wait with pushing that one until this is reviewed

app/webpacker/components/CompetitionsOverview/CompetitionsFilters.js
```diff
@@ -76,7 +76,11 @@ function CompetitionsFilters({
   );
 }
 
-export function EventSelector({ selectedEvents, onEventSelection, eventList = WCA_EVENT_IDS }) {
+export function EventSelector({
+  selectedEvents,
+  onEventSelection, eventList = WCA_EVENT_IDS,
```
gregorbg, line 81: weird line break

app/views/registrations/edit_v2.html.erb
```diff
@@ -0,0 +1,7 @@
+<% provide(:title, t('registrations.edit_registration.title', person: @user.person.name, comp: @competition.name)) %>
+<% add_to_packs("auto_numeric") %>
```
gregorbg, line 2: I don't think you need this pack here, because you're only rendering a React component and that component uses imports through webpack.

The `add_to_packs` hack is only necessary if you want to access Webpack code in Sprockets

gregorbg (commented):
Small nitpicks

FinnIckler:
Thanks for the help with the cleanup, you can merge this now

gregorbg (approved):

PR #9233, opened 2024-04-17
Some cleanups in usages of Team.wxx

danieljames-dj:


gregorbg:
Hm. I get that we need to move away from the `Team.foo` access, but I have a problem with this approach.
The `UserGroup.teams_committees_foo_group` accessors load `foo`'s metadata under the hood, to then look up the group and return it. You use this abstract group to walk back to the metadata again. That's like going from A to B back to A. I see the risk of generating database queries, as I am not sure whether Rails is smart enough to figure this out.

In a development environment, boot up the server, which has the advantage that (in development mode, which is set by default) all database queries are printed to the logs. Trigger any of the actions that are affected in this PR and see if it actually fires three (select metadata, select group based on metadata, select metadata based on group) queries.

If it does, you need to go back to the drawing board and figure out a more clever way of doing things.

danieljames-dj:
In my local, I changed delegates.html.erb as follows:
```
<% UserGroup.teams_committees_group_wct.metadata.email %>
```

Then when I opened `localhost:3000/delegates`, the following were the logs:
```
Started GET "/delegates" for ::1 at 2024-04-17 12:31:27 +0530
  User Load (1.5ms)  SELECT `users`.* FROM `users` WHERE `users`.`id` = 6387 ORDER BY `users`.`id` ASC LIMIT 1
  ↳ lib/middlewares/warden_user_logger.rb:11:in `call'
Processing by StaticPagesController#delegates as HTML
  Rendering layout layouts/application.html.erb
  Rendering static_pages/delegates.html.erb within layouts/application
  GroupsMetadataTeamsCommittees Load (0.8ms)  SELECT `groups_metadata_teams_committees`.* FROM `groups_metadata_teams_committees` WHERE `groups_metadata_teams_committees`.`friendly_id` = 'wct' LIMIT 1
  ↳ app/models/user_group.rb:135:in `teams_committees_group_wct'
  UserGroup Load (0.6ms)  SELECT `user_groups`.* FROM `user_groups` WHERE `user_groups`.`metadata_id` = 1 AND `user_groups`.`metadata_type` = 'GroupsMetadataTeamsCommittees' LIMIT 1
  ↳ app/models/user_group.rb:135:in `teams_committees_group_wct'
  Rendered static_pages/delegates.html.erb within layouts/application (Duration: 8.3ms | Allocations: 1103)
Shakapacker::Compiler - Slow setup for development

Prepare JS assets with either:
1. Running `bin/shakapacker-dev-server`
2. Set `compile` to false in shakapacker.yml and run `bin/shakapacker -w`
[Shakapacker] Everything's up-to-date. Nothing to do
Shakapacker::Compiler - Slow setup for development
```

Hence, looks like it's using only 2 queries and not 3.

gregorbg:
Cool! In this case it actually seems to be cached.

gregorbg (approved):

PR #9234, opened 2024-04-17
Trying out RailsAdmin

dunkOnIT:
WARNING: This broke my database, proceed with caution


gregorbg:
Interestingly enough, adding this gem consistently produces a Segfault when compiling assets

gregorbg:
> did you ever come right with this? Do we want to keep it around?

I think the PR is in a working state now, but by now it needs SCSS deprecation overrides... Meaning RailsAdmin relies on a version of Bootstrap which relies on now-deprecated SCSS features. Not sure if we want to introduce that into our codebase, but on the other hand we're getting rid of that part of our pipeline soon(:tm:) at which point RailsAdmin can do whatever the heck it wants and we're not responsible for their deprecation warnings anymore.

TL;DR Basically fine with merging it, but only if we (or WRT?) are confident that it will yield some benefit for us.

dunkOnIT:
Is it accessible on local dev? I'm happy to give it a look to start off the "will it be useful" conversation - if memory serves, the main advantages it would yield are allowing WST/WRT to access data
- more quickly (not needing to go via the PMA login) 
- more safely (bit wild that we open up the prod db every time - even if we _could_ use the read-only, I think it introduces systemic risk that could be avoided)

If RailsAdmin works as we would hope it does, then it should mean that the only time we need to go into PMA is to run more complex queries on the database, or perform write operations

But we need to verify that it would actually provide those benefits, or some other equally-desireable ones!

dunkOnIT:
I took it for a brief test drive and my main impression is that it's very slow, I presume because of the size of our database. 

I'm not convinced that time trying to optimize would be well-spent - and conversations like [this one](https://www.reddit.com/r/rails/comments/1adqsnv/rails_admin_vs_administrate/) have strong recommendations in favour of essentially rolling your own admin panel, which our project may well be at the size where it is the more appropriate approach. 

My suggestion is that we keep an eye out for common reasons that we need to access the db, and see if (especially the read-only data) might not be easy to expose in a basic admin panel that we could build in a couple of days, and maintain much more easily.

PR #9235, opened 2024-04-17
Removed Team.all_official and showed WCAT instead of their email in competition form

danieljames-dj:


spec/views/teams/index.html.erb_spec.rb
```diff
@@ -5,7 +5,7 @@
 RSpec.describe "teams/index.html.erb" do
   describe "when signed in as an admin" do
     let!(:user) { FactoryBot.create :admin }
-    let!(:teams) { Team.all_official }
+    let!(:teams) { Team.unscoped.all }
```
gregorbg, line 8: This is not an equivalent replacement of the previous code

gregorbg: Okay, thanks!

gregorbg (approved):

PR #9240, opened 2024-04-17
Lock PI row when updating charges

gregorbg:
Seems like there is a weird race condition between the payment gateway return URL and the webhook implementation. This should hopefully prevent concurrent writes :upside_down_face:


gregorbg:
Merging as a hotfix, I just needed the CI to pass

PR #9241, opened 2024-04-18
Migrate WCT China to roles

danieljames-dj:


spec/factories/users.rb
```diff
@@ -98,7 +98,7 @@
 
     trait :wct_china_member do
       after(:create) do |user, options|
-        FactoryBot.create(:team_member, team_id: Team.wct_china.id, user_id: user.id, team_senior_member: options.team_senior_member, team_leader: options.team_leader)
```
gregorbg, line 101: You're losing two important bits of information (senior member, team leader). For WCT China, this is probably okay because they don't have so strict structures, but what about future role migrations?

db/migrate/20240418014842_migrate_wct_china_to_roles.rb
gregorbg, line 1: It may be worth to compare this with the WST Admin migration code and come up with one (or multiple helper) function(s) like "def migrate_team_to_roles" in `lib` somewhere

gregorbg (approved):

db/migrate/20240418014842_migrate_wct_china_to_roles.rb
danieljames-dj: Makes sense, I've moved this method to lib so can be reused for upcoming migrations.

spec/factories/users.rb
danieljames-dj: For WCT China, the status is supposed to be `member` in testcases used so it's fine. For future migration, I'm planning to use multiple traits like `wrt_member`, `wrt_senior_member`, etc.

PR #9242, opened 2024-04-18
Added test for stripe webhook confirmation

dunkOnIT:
Per discussion in Slack, this PR:
- Adds a generator for the Stripe webook payload
- Adds a test that confirms payment by mimicing the Stripe webhook

### DIFF
--- a/spec/requests/registrations_spec.rb
+++ b/spec/requests/registrations_spec.rb
@@ -592,17 +592,22 @@ RSpec.describe "registrations" do
           post registration_payment_intent_path(registration.id), params: {
             amount: registration.outstanding_entry_fees.cents,
           }
+
           payment_intent = registration.reload.payment_intents.first
+          stripe_account_id = competition.payment_account_for(:stripe).account_id
 
           # mimic the user clicking through the interface
           Stripe::PaymentIntent.confirm(
             payment_intent.payment_record.stripe_id,
             { payment_method: 'pm_card_visa' },
-            stripe_account: competition.payment_account_for(:stripe).account_id,
+            stripe_account: stripe_account_id,
           )
 
           # mimic the response that Stripe sends to our webhook upon payment completion
-          post registration_stripe_webhook_path, params: payment_confirmation_webhook(payment_intent.retrieve_remote.to_hash)
+          post registration_stripe_webhook_path, params: payment_confirmation_webhook_as_json(
+            payment_intent.retrieve_remote.to_hash,
+            stripe_account_id,
+          )
 
           expect(registration.reload.outstanding_entry_fees).to eq 0
           expect(registration.paid_entry_fees).to eq competition.base_entry_fee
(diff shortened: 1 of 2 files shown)


spec/requests/registrations_spec.rb
```diff
@@ -1313,3 +1343,90 @@ def refund_response(capture_id)
     ],
   }.to_json
 end
+
+def payment_confirmation_webhook(competition, registration, intent)
```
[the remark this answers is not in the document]
dunkOnIT: Ah that is way cleaner, thank you!

spec/requests/registrations_spec.rb
```diff
@@ -1313,3 +1343,90 @@ def refund_response(capture_id)
...
+    id: "evt_3P6aXQJzvpX2joEA18jzmlxq",
+    object: "event",
+    account: "acct_1FcgXcJzvpX2joEA",
+    api_version: "2023-10-16",
+    created: DateTime.now.to_i,
+    data: {
+      object: {
+        id: intent.payment_record.stripe_id,
+        object: "payment_intent",
+        amount: competition.base_entry_fee_lowest_denomination,
```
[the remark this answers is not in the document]
dunkOnIT: Changed

spec/requests/registrations_spec.rb
```diff
@@ -1313,3 +1343,21 @@ def refund_response(capture_id)
...
+
+def payment_confirmation_webhook(intent)
+  {
+    id: "evt_3P6aXQJzvpX2joEA18jzmlxq",
+    object: "event",
+    account: "acct_1FcgXcJzvpX2joEA",
+    api_version: "2023-10-16",
+    created: DateTime.now.to_i,
+    data: { object: intent },
+    livemode: true,
```
gregorbg, line 1355: Nope.

spec/requests/registrations_spec.rb
```diff
@@ -1313,3 +1343,21 @@ def refund_response(capture_id)
     ],
   }.to_json
 end
+
+def payment_confirmation_webhook(intent)
```
gregorbg, line 1347: The fact that you're serializing to JSON should be reflected in the method name.

Alternatively, don't serialize to JSON in this method and serialize the return value after storing it in a variable instead.

spec/requests/registrations_spec.rb
```diff
@@ -1313,3 +1343,21 @@ def refund_response(capture_id)
...
+  {
+    id: "evt_3P6aXQJzvpX2joEA18jzmlxq",
+    object: "event",
+    account: "acct_1FcgXcJzvpX2joEA",
```
gregorbg, line 1351: This can (and should!) be read from the actual Stripe account ID that was also used to make the requests creating and confirming the PI

gregorbg (changes requested):

spec/requests/registrations_spec.rb
dunkOnIT: Changed

spec/requests/registrations_spec.rb
dunkOnIT: Changed

spec/requests/registrations_spec.rb
dunkOnIT: Changed

dunkOnIT: (and refactored into a variable for this test as the account ID is called elsewhere)

gregorbg (approved):

PR #9246, opened 2024-04-19
Rename columns in the Paypal Records table

gregorbg:
Partial implementation of https://github.com/thewca/worldcubeassociation.org/pull/9208/files


gregorbg:
Yay refactor!

PR #9252, opened 2024-04-22
Add Registration V2 Competitor and Waiting List

FinnIckler:
Needs the useWithUserDataHook and the SortReducer from the other PR.
Another question regards pysch sheets: Are we getting rid of the little Icons on the sidebar if using V2 or do I redirect it to the new combined competitors/psychsheets page?


app/controllers/registrations_controller.rb
```diff
@@ -72,6 +72,10 @@ def index
     @newcomers = @returners = 0
   end
 
+  def waiting
```
gregorbg, line 75: I think you should name this `waiting_list` or similar.

app/webpacker/components/RegistrationsV2/api/registration/get/get_registrations.js
```diff
@@ -1,12 +1,11 @@
...
+import { getPsychSheetForEventUrl, wcaRegistrationUrl } from '../../../../../lib/requests/routes.js.erb';
 import fetchWithJWTToken from '../../../../../lib/requests/fetchWithJWTToken';
 import { fetchJsonOrError } from '../../../../../lib/requests/fetchWithAuthenticityToken';
 
 const confirmedRegistrationsUrl = (competitionId) => `${wcaRegistrationUrl}/api/v1/registrations/${competitionId}`;
-const getPsychSheetForEventUrl = (competitionId, eventId, sortBy) => `${wcaRegistrationUrl}/api/v1/psych_sheet/${competitionId}/${eventId}?sort_by=${sortBy}`;
 const allRegistrationsUrl = (competitionId) => `${wcaRegistrationUrl}/api/v1/registrations/${competitionId}/admin`;
 const singleRegistrationUrl = (competitionId, userId) => `${wcaRegistrationUrl}/api/v1/register?user_id=${userId}&competition_id=${competitionId}`;
-const waitingCompetitiorsUrl = (competitionId) => `${wcaRegistrationUrl}/api/v1/registrations/${competitionId}/waiting`;
+const waitingCompetitorsUrl = (competitionId) => `${wcaRegistrationUrl}/api/v1/registrations/${competitionId}/waiting`;
```
gregorbg, line 8: This seems to be inconsistent with the notion of listing some RegMicroservice URLs in the `routes.js.erb` file in the previous PR

gregorbg (commented):
Re the psych sheet icons:

I have a slight preference towards removing the icons from the sidebar. If you end up finding an easy "if reg_v2 then [blank] else show icons" approach, then consider to show a "info" icon popup with a tooltip text that tells the users where the "new" psych sheets are instead of just straight-out blank space.

But in the end, go with whatever approach works easier for you.

app/webpacker/components/RegistrationsV2/api/registration/get/get_registrations.js
FinnIckler: Not sure which your mean, there were none in the previous PR only the base url

app/views/competitions/_nav.html.erb
```diff
@@ -129,21 +129,32 @@
...
+          if @competition.registrations.any?
             nav_items << {
               text: t('.menu.competitors'),
               path: competition_registrations_path(@competition),
-              icon: "users",
+              icon: "users", # Don't add event icons if v2 is used
               tiny_children: @competition.events.map do |event|
-                { text: event.id, path: competition_psych_sheet_event_path(@competition, event.id), cubing_icon: event.id, title: event.name }
-              end
+                if @competition.uses_new_registration_service?
```
gregorbg, line 145: Potential whoopsie?
To me, this reads "if comp uses the new system, add tiny icons. Otherwise, do nothing". Shouldn't it be the opposite (i.e. use `unless`)? Or am I missing something here?

app/webpacker/components/RegistrationsV2/api/registration/get/get_registrations.js
gregorbg: Ah, okay!

FinnIckler:
@thewca-bot deploy staging

app/views/competitions/_nav.html.erb
```diff
@@ -129,21 +129,32 @@
...
+                  nil
+                else
+                  { text: event.id, path: competition_psych_sheet_event_path(@competition, event.id), cubing_icon: event.id, title: event.name }
+                end
```
gregorbg, line 149: What about `events.map do |event| ... end unless competition_uses_v2?`?

gregorbg: Would save you from explicitly returning nil, and also gets rid of the `.compact` call

app/webpacker/components/RegistrationsV2/Waiting/WaitingList.jsx
```diff
@@ -0,0 +1,50 @@
...
+export default function WaitingList({ competitionInfo }) {
+  const { isLoading: waitingLoading, data: waiting, isError } = useQuery({
+    queryKey: ['waiting', competitionInfo.id],
+    queryFn: () => getWaitingCompetitors(competitionInfo.id),
+    retry: false,
+  });
+
+  const { isLoading: infoLoading, data: registrationsWithUser } = useWithUserData(waiting ?? []);
+
+  // eslint-disable-next-line no-nested-ternary
```
gregorbg, line 17: Definitely not happy with nesting ternaries.
```
if (waitingLoading || infoLoading)` return Loading;
if (isError) return Errored ( <--- we have a shared component for this! )

return (...the actual component...)
```

app/webpacker/components/RegistrationsV2/Registrations/RegistrationList.jsx
```diff
@@ -0,0 +1,315 @@
...
+      });
+      if (sortDirection === 'descending') {
+        return sorted.toReversed();
+      }
+      return sorted;
+    }
+    return [];
+  }, [dataWithUser, sortColumn, sortDirection, psychSheetEvent]);
+
+  // eslint-disable-next-line no-nested-ternary
```
gregorbg, line 150: See my other comment about ternaries.

gregorbg (changes requested):

app/views/competitions/_nav.html.erb
FinnIckler: sadly you can't do `do .. end unless` in an object definition in Ruby it was the first thing I tried as well

gregorbg: Then maybe
```
events.map do |event|
  foo if comp.uses_v2_reg?
end.compact
```
could work at least?

FinnIckler: Making it a variable outside of it though works

gregorbg (approved):

PR #9257, opened 2024-04-24
Added email to api/v0/users/me API

danieljames-dj:
In contacts form, the name and email is prefilled. But I just realized that the email is not getting prefilled. This is because the user serialization is not including email. This PR is adding email to the user serialization if the requested user is `current_user`.

### DIFF
--- a/app/controllers/api/v0/users_controller.rb
+++ b/app/controllers/api/v0/users_controller.rb
@@ -5,7 +5,7 @@ class Api::V0::UsersController < Api::V0::ApiController
     require_user!
     if stale?(current_user)
       # Also include the users current prs so we can handle qualifications on the Frontend
-      show_user(current_user, show_rankings: true, current_user: true)
+      show_user(current_user, show_rankings: true, private_attributes: ['email'])
     end
   end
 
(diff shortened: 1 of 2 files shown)


app/controllers/api/v0/users_controller.rb
```diff
@@ -62,9 +62,13 @@ def token
 
   private
 
-    def show_user(user, show_rankings: false)
```
gregorbg, line 65: The name `current_user` for this flag is definitely confusing. In the context of this controller, this name is already defined to retrieve the actual `User` instance (or `nil` if you're not logged in). So you definitely have to pick another name.

app/controllers/api/v0/users_controller.rb
```diff
@@ -62,9 +62,13 @@ def token
...
+    def show_user(user, show_rankings: false, current_user: false)
       if user
-        json = { user: user }
+        if current_user
+          json = { user: user.serializable_hash(private_attributes: ['email']) }
+        else
+          json = { user: user }
+        end
```
gregorbg, line 71: This if-clause looks pretty repetitive, but I don't have an immediate better replacement in mind.
What do you think about passing in desired private user attributes as an array parameter directly into the method? That way, we avoid the naming discussion from the other comment.

Unsure whether that would be more readable, but it may be worth a try.

gregorbg (changes requested):
Definitely need to change the name of the `current_user` boolean param.
The other comment is more of a philosophical discussion and not a binding statement.

app/controllers/api/v0/users_controller.rb
danieljames-dj: Oh that's better actually. I've done that now. Thanks.

gregorbg (approved):

PR #9258, opened 2024-04-24
Make contact form more cleaner for user

danieljames-dj:
This does two things:
1. Changes the inquiry types to contact recipient, making the form more easier for users, and hence change the options to more easier ones.
2. Won't show user name & email if user is logged in.

Screenshot 2024-04-24 at 8 31 10 AM


gregorbg:
Who approved the deletion of so many contact types? For example, the `media` contact type is completely banished from existence. Whose decision was this?

gregorbg (approved):
Approved based on your quote of the WCT discussion.

PR #9262, opened 2024-04-24
Use Event and Schedule React Component from Registrations

FinnIckler:
Currently untested, will do later


FinnIckler:
I had to change the location of the dropdown a bit, because otherwise we can't localize it (because we still aren't using react i18n and can't render a dropdown in the middle of an i18n string :upside_down_face: )

FinnIckler:
@kr-matthews Does this still convey the ideas you had around the custom timezones?

gregorbg:
Your current way of rendering the timezone dropdown picker has a flaw.
Once you pick "custom", the timezone picker switches into custom mode and offers you a list of all supported timezones. You cannot go back to "your local TZ" mode or "follow htis venue's TZ" mode.

FinnIckler:
I guess I could add a reset button?

gregorbg:
My suggestion: Don't have competing dropdowns.
Keep the old behaviour but decouple it from the need of embedding it into a "nice phrase" for I18N's sake:

```
You are viewing this schedule in the following timezone setting: [Dropdown with "your local"/venue 1/venue 2/.../venue N/"custom"]
{If the above is custom} The current custom timezone setting is [Dropdown]
```

kr-matthews:
I will try and take a look tomorrow, or the day after.

kr-matthews:
I haven't had time to look through the code, but from briefly running locally, the timezone selector seems decent but slightly awkward to me. Once in the custom mode, if I select a time zone corresponding to me or a venue, then the timezone sentence/dropdown disappears, which feels counter-intuitive. I feel like both dropdowns should always be present/visible, but maybe that's just me.

Also, the wording "your local" and "a custom" doesn't sound right to me in the new format.

app/webpacker/components/Schedule/CalendarView.jsx
```diff
@@ -0,0 +1,98 @@
...
+        slotMinTime={calendarStart}
+        slotMaxTime={calendarEnd}
+        slotDuration="00:15:00"
+        height="auto"
+        locale={calendarLocale}
+        timeZone={timeZone}
+        events={fcActivities}
+      />
+      {fcActivities.length === 0 && (
+        <em>No activities for the selected rooms/events.</em>
```
kr-matthews, line 94: This should be `i18n`ed?

app/webpacker/components/Schedule/VenuesAndRooms.jsx
```diff
@@ -0,0 +1,134 @@
...
+    const ids = newVenues.flatMap((venue) => venue.rooms).map((room) => room.id);
+    dispatchRooms({ type: 'reset', ids });
+
+    setActiveVenueIndex(newVenueIndex);
+  };
+
+  return (
+    <>
+      {venueCount > 1 && (
+        <Menu
```
kr-matthews, line 32: This looks weird now, there's no underline of the selected venue, so it's less clear that this is a menu.

image

Compare to calendar vs table menu:

image

app/webpacker/components/Schedule/VenuesAndRooms.jsx
```diff
@@ -0,0 +1,134 @@
...
+          pointing
+          secondary
+          fluid
+          stackable
+          // TODO: can't scroll left when 6+ venues
```
kr-matthews, line 37: This is unfortunately still a problem; see FMC 2024 (with 74 venues!) for example. I did try to fix this myself but got stuck. Someone more familiar with semantic ui might be able to easily fix it though.

image

app/webpacker/components/Schedule/TimeZone.jsx
```diff
@@ -0,0 +1,73 @@
...
+        selection
+        value={activeTimeZoneLocation}
+        onChange={(_, data) => dispatchTimeZone({
+          type: 'update-location',
+          location: data.value,
+          venues,
+        })}
+        options={locationOptions}
+      />
+      {activeTimeZoneLocation === 'custom' && (
```
kr-matthews, line 53: I think this should always be visible. If I'm viewing venue A, and I select venue B's time zone, then I would like to know what time zone that is. It also avoids the unexpected disappearance of the second dropdown.

Maybe combining them into a single sentence? "You are viewing the calendar for the following location's time zone: < "your current location" | venue name | "custom location" > in < time zone >." Or something to that effect. Note the options  "your current location" and "custom location" replacing "your local" and "a custom".

If we prefer to keep them separate, then I would just change "The current custom timezone setting is" to "The calendar is currently being viewed in the time zone:". (Note "timezone" isn't a word, as far as I can tell; it should be time zone.)

app/webpacker/components/Schedule/VenuesAndRooms.jsx
```diff
@@ -0,0 +1,134 @@
...
+              height: '100%',
+            }}
+          >
+            <Form.Checkbox
+              label={
+                // eslint-disable-next-line jsx-a11y/label-has-associated-control
+                <label style={{ color: getTextColor(color) }}>{name}</label>
+              }
+              checked={activeRoomIds.includes(id)}
+              onChange={() => toggleRoom(id)}
```
kr-matthews, line 94: I think it would be more intuitive if the whole box was clickable, not just the checkbox/label. (Yes, I know I wrote this code originally :smiling_face_with_tear:.)

app/webpacker/components/Schedule/TimeZone.jsx
gregorbg: I still don't understand what the purpose behind these TZ selection dropdowns should actually be. And I don't mean "I don't understand" as a polite way of saying that you're wrong and I'm right. I genuinely don't understand.

Say I am selecting some subset of venues, and it shows the calendar view. The only potentially interesting use cases I can think of are:
- How does this schedule "interact" with other venues: I want to see the current schedule but in the time zone of other venues. So there should be one dropdown for selecting which venue's time zone to follow.
- Watching the live stream of a far away competition: I want to see how the schedule looks in relation to my local time zone. So I need one additional dropdown option for setting to my local (browser's) timezone
- Doing some custom calculation shenanigans, giving me the full freedom over selecting timezones.

In summary, I don't understand why you would ever want to have a dropdown offering you to follow a certain venue's TZ _and_ **at the same time** have a second dropdown allowing you to override the selection made by the venue dropdown.

kr-matthews: The use cases you list seem correct. (Plus you might want to plan to watch a live stream next weekend, but you're currently in a different time zone than you will be next weekend, so your local time zone would not be what you want. But that's probably not a common situation. Or you could see how jet lag might affect someone travelling from Germany to Worlds 2025, even if you live in a completely different time zone.)

Basically, WRT had some feedback on the schedule (there was an email thread many months ago), and we settled on the fact that it would be nice to be able to select any time zone. But just having a single (very long) drop-down of all time zones isn't nice. And just having a single drop-down of the venues (plus local) means you don't have all time zones. So I made 2 drop-downs which are synchronized (changing one automatically updates the other appropriately).

So I guess the short answer is that this is the nicest way I could find to allow the user to select any time zone, while still maintaining convenient access to venue time zones. If we don't need the user to be able to see arbitrary time zones then we could remove the custom option and the second drop-down.

gregorbg:
I have added another suggestion for the time zone picker. Basically, you always have one dropdown that allows you to select any timezone.
For following the currently selected venue, there is a "follow selected venue" button which is turned on by default. You can turn it off, which makes a "use this zone" appear in ech venue's individual menu tab.
You can also press a "set to local" button to use your local timezone.

gregorbg:
UI is still a bit rough around the edges and needs proper layout, but here's the proposal:
image

Totally open about ideas for better button placement
image

config/locales/en.yml
```diff
@@ -1915,13 +1914,31 @@ en:
...
+      venue_information_html:
+        one: "You are viewing the schedule for the venue %{venue_name}, the sole venue for this competition."
+        other: "You are viewing the schedule for the venue %{venue_name}, one of %{count} venues for this competition."
+      venue_information_all: "You are viewing the schedule for all %{venueCount} venues at once."
+      timezone_message: "The time zone for this venue is %{timezone}."
+      timezone_setting: "You are viewing this schedule in the following timezone setting:"
+      multiple_timezones_available: "Note that not all venues are in the same time zone -- please be careful!"
+      timezone_set_local: "Set to local"
+      timezone_set_venue: "Set to venue"
+      timezone_follow_venue: "Follow the active venue tab"
```
kr-matthews, line 1939: This wording makes sense once I know the functionality that it's referring to, but when I first saw it I was slightly confused. I thought it was something (maybe notification based? that's what 'follow' usually means online) to do with the specific _current_ active venue tab. I think something along the lines of "Automatically show the active venue's time zone" might be better. It might just be me though.

kr-matthews (commented):
As mentioned in my other review, I can't run locally so I can't verify. But the screenshots on slack, and the code here, both look reasonable.

Thanks for coming up with a better UI than I originally had, I think it's an improvement.

coder13:
Issues that have been raised since merging:
- Cumulative time limits are not surfaced
- Table appears sub-par on mobile.  
image

Because of the way that the table is rendered such that the event name doesn't get duplicated, "First Round" isn't bolded like Second Round is.

I share similar personal sentiment:
image

The alternating row colors is inconsistent across the event name column and the rest.

config/locales/en.yml
```diff
@@ -1915,13 +1914,32 @@ en:
...
+        show_buttons: "Show time zone buttons"
+        use_time_zone: "Use this time zone"
       #context: when displaying the room on the general schedule, we may also display the venue if there are multiple venues
       room_in_venue: "%{room} in %{venue}"
-      venue_information_html: "You are viewing the schedule for the venue %{venue_name}."
-      timezone_message: "The schedule is displayed in the timezone %{timezone}."
-      multiple_venues_available: "This competition has multiple venues, make sure to check out the schedule for the other venues as well!"
+      venue_information_html:
+        one: "You are viewing the schedule for the venue %{venue_name}, the sole venue for this competition."
+        other: "You are viewing the schedule for the venue %{venue_name}, one of %{count} venues for this competition."
```
jonatanklosko, line 1932: Internationalize handles pluralized objects in a special way, but the `venue_information_html` key was regular and is now pluralized, so existing translations for this key cause issues during synchronization. Could you please rename this key to anything else in the English version, so that sync works properly? :)

gregorbg: Sorry, this wasn't intentional! See https://github.com/thewca/worldcubeassociation.org/pull/9402

jonatanklosko: Thanks!

PR #9264, opened 2024-04-24
Added query-params to contact form along with few other changes

danieljames-dj:
This PR aims at following changes related to contact form:
1. In backend, accepts `contactRecipient` instead of `contactType`.
2. In frontend, recognizes if contactRecipient is added in the queryParams. This will allow to replace WRT's contact emails with something like https://worldcubeassociation.org/contact?contactRecipient=results_team.
3. Renamed occurences of contactType with contactRecipient.
4. Moved UserData outside the SubForms folder, because UserData is not actually a subform, instead it is part of the common parent form.
5. Earlier, the subform default values were passed from the parent form. Changed it to the subform doing it's own formValues and form validation, and hence making it easier to expand more subforms in future. In other words, the work in the parent form is reduced and each subform will do some work on their own.

### DIFF
--- a/app/webpacker/components/ContactsPage/SubForms/Competition/index.jsx
+++ b/app/webpacker/components/ContactsPage/SubForms/Competition/index.jsx
@@ -1,23 +1,64 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { FormTextArea } from 'semantic-ui-react';
+import { useQuery } from '@tanstack/react-query';
 import WcaSearch from '../../../SearchWidget/WcaSearch';
 import SEARCH_MODELS from '../../../SearchWidget/SearchModel';
 import I18n from '../../../../lib/i18n';
+import useQueryParams from '../../../../lib/hooks/useQueryParams';
+import { fetchJsonOrError } from '../../../../lib/requests/fetchWithAuthenticityToken';
+import { apiV0Urls } from '../../../../lib/requests/routes.js.erb';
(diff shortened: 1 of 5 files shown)


app/webpacker/components/ContactsPage/SubForms/Competition/index.jsx
```diff
@@ -1,23 +1,64 @@
...
   );
+  const { data: competitionData, isLoading } = useQuery({
+    queryKey: ['competition'],
+    queryFn: () => fetchJsonOrError(apiV0Urls.competitions.info(queryParams.competitionId)),
+    enabled: !!queryParams?.competitionId,
+  });
+
+  useEffect(() => {
+    setSubformValues(formValues);
+    setFormValid(formValues.competition !== null && formValues.message?.length > 0);
```
gregorbg, line 32: This should not be done by manually computing your booleans. SemUI Form elements have a `required` property which does most* of the work for you

*Exception: Checkboxes are a little bit glitched out. Those are the only exception where we have to compute `isValid` booleans ourselves because of a glitch in the SemUI React integration. Everything else should be solvable by just setting the input fields to `required` and calling it a day.

If you can come up with an argument why it's really, _really_ required to do those validations yourself, you should probably do them on the top level and switch business logic based on which `contactRecipient` is selected.

danieljames-dj:
Closing this PR as it's irrelevant now, and has been fixed in other PR(s).

PR #9265, opened 2024-04-24
Replace some usages of teams with Team.banned

danieljames-dj:
current_teams, teams, current_team_members, etc of user model is not used for official teams now (there might be some, but no longer relevant once migration is finished). They are still used for Team.banned though. Hence, changed the unit tests to check from Team.banned.

Also cleaned up some usages with roles methods.

The purpose of this PR to cleanup little more before the final migration.

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -457,10 +457,11 @@ RSpec.describe User, type: :model do
   it "#teams and #current_teams return unique team names" do
     user = FactoryBot.create(:user)
 
-    FactoryBot.create(:team_member, team_id: Team.banned.id, user_id: user.id, start_date: Date.today - 20)
-    FactoryBot.create(:team_member, team_id: Team.banned.id, user_id: user.id, start_date: Date.today - 10)
+    FactoryBot.create(:team_member, team_id: Team.wst.id, user_id: user.id, start_date: Date.today - 20, end_date: Date.today - 10)
+    FactoryBot.create(:team_member, team_id: Team.banned.id, user_id: user.id, start_date: Date.today - 5, end_date: Date.today + 5)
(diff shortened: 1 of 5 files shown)


spec/models/user_spec.rb
gregorbg, line 1: The changes to this file kinda defeat the purpose of this test. In the title, it says `return _unique_ team members`, and the body of the (original, left side diff) test creates two non-overlapping WRT membership periods. But the expectation is that the list of "which teams is this user a member of?" only returns one distinct "WRT" entry desipte the user technically having two WRT roles attached.

Your suggested update only ever adds the person to one team and is thus not equivalent with the original test.

gregorbg (changes requested):

spec/models/user_spec.rb
gregorbg: Hm, still not _exactly_ covering the original case because that made a distinction between teams (current AND PAST memberships) as well as current_teams, but I get that with only one team it can be hard to replicate this behavior. Is there any other team you can use?

(On a sidenote, what makes the "Banned" team so special? WRT and also WRC are going to be migrated very soon, as well as the Banned team which I hope will also be migrated soon. So at the end, this test will have to be adjusted to the Roles system anyways, why not do that right now?)

danieljames-dj: We have the similar tests in roles system (but not exact tests because there is no use-case for these methods in roles): https://github.com/thewca/worldcubeassociation.org/blob/main/spec/models/user_group_spec.rb.

The reason why I chose banned team here is because that is the last team that will be migrated away. For now, I've added WST here expecting WST will be the second last team that will be migrated. Hope this maintains the same test case as it is but with different teams.

gregorbg (approved):

PR #9271, opened 2024-04-26
Added newcomers_allowed field and started adding to competition form

dunkOnIT:
This is per the recent update to WCRP which allows organizers to specify having a WCA ID as a qualification criterion. 

**Note: I'm still going to add a conditional to the field in the competition form that it only shows is v2 registrations are enabled**

2 things: 
1. Implementing the field in the form - I've stumbled around with the React form, but am unsure how to get `newcomersAllowed` into this piece of code without throwing an error: 
```javascript
eventRestrictions: {
        newcomersAllowed,
        earlyPuzzleSubmission,
        qualificationResults,
        eventLimitation,
      },
```
2. Are we concerned at bloat in the competition model with the number of properties it has? I guess this isn't the place to address that, but I'm wondering if at some point we want to store the preferences separately from the "core" competition data - but that's a whole redesign project on its own.

And then there's a lot of nitting that could be done on "newcomer" vs "first-timer" vs "has a WCA ID", as well as where in the competition form this should be displayed.


db/migrate/20240426045248_add_newcomer_allowed_to_competition.rb
```diff
@@ -0,0 +1,6 @@
+class AddNewcomerAllowedToCompetition < ActiveRecord::Migration[7.1]
+  def change
+    add_column :Competitions, :newcomers_allowed, :boolean, default: true
+    Competition.update_all(newcomers_allowed: true)
```
gregorbg, line 4: Not needed b/c of `default: true` above

db/migrate/20240426045248_add_newcomer_allowed_to_competition.rb
```diff
@@ -0,0 +1,6 @@
+class AddNewcomerAllowedToCompetition < ActiveRecord::Migration[7.1]
+  def change
+    add_column :Competitions, :newcomers_allowed, :boolean, default: true
```
gregorbg, line 3: also add `null: false`

gregorbg (changes requested):

db/migrate/20240426045248_add_newcomer_allowed_to_competition.rb
dunkOnIT: Changed

db/migrate/20240426045248_add_newcomer_allowed_to_competition.rb
dunkOnIT: Changed

app/models/competition.rb
```diff
@@ -2284,6 +2284,9 @@ def to_form_data
         "forceComment" => force_comment_in_registration,
       },
       "eventRestrictions" => {
+        "newcomersAllowed" => {
+          "enabled" => newcomers_allowed?,
+          "reason" => newcomers_allowed,
```
gregorbg, line 2289: You're missing a closing bracket here :wink:

dunkOnIT:
- [ ] Make remark required
- [ ] Only show newcomer field if v2 registrations enabled

app/models/competition.rb
dunkOnIT: Changed

gregorbg (approved):

PR #9273, opened 2024-04-26
Make Delegates Table Sortable (#9081)

### LINKED ISSUE #9081: Implement sort feature in All Delegates page
This is a React-only task. The task is to bring sort feature in [DelegatesTable.jsx](https://github.com/thewca/worldcubeassociation.org/blob/main/app/webpacker/components/Delegates/DelegatesTable.jsx).

Also, the sorting has to be enabled only when `isAllNonLeadDelegates` is `true` and hence the sorting will be available only in [All Delegates page](https://www.worldcubeassociation.org/delegates#all) under 'Other Delegates' which is restricted only to few users.

(Later, it can be expanded to other pages if requirement comes)

Feel free to take it up if anybody is willing to work on this. (Please leave a comment here before starting so that we won't have multiple people work on same task)

Carter-Thomas:
Fix #9081, haven't tested this yet as I don't have a functional environment atm, should be fixed though.


app/webpacker/components/Delegates/DelegatesTable.jsx
```diff
@@ -8,44 +8,113 @@ import UserBadge from '../UserBadge';
...
+        }
+        return {
+          column: action.column,
+          data: _.sortBy(state.data, [action.column]),
+          direction: SORT_TYPES.ASCENDING,
+        };
+      default:
+        throw new Error();
+    }
+  };
```
kr-matthews, line 43: It's best to define reducers outside of the component. (To ensure it's not accidentally using any state from the component, and so it's not re-computed every render.)

app/webpacker/components/Delegates/DelegatesTable.jsx
```diff
@@ -8,44 +8,113 @@ import UserBadge from '../UserBadge';
...
+                  sorted={column === 'last_delegated' ? direction : null}
+                  onClick={() =>
+                    dispatch({ type: 'CHANGE_SORT', column: 'last_delegated' })
+                  }
+                >
                   {I18n.t('delegates_page.table.last_delegated')}
                 </Table.HeaderCell>
                 <Table.HeaderCell>
-                  Date Since Last Delegated
+                  Date Since Last Delegated
```
kr-matthews, line 109: Is this space intentional?

FinnIckler:
We have a `createSortReducer` in `webpacker/RegistratonsV2` (feel free to move that file into a more general lib/reducers folder), that does basically the same thing, but doesn't include the data in the state.

Carter-Thomas:
Okay, I think I should've fixed this, let me know if there's anything else that's wrong because I'm still not very confident that this is 100% functional.

FinnIckler:
I'll check out the code and try it

FinnIckler:
I did some clean up and the sorting works. But there are still things that don't work like `total_delegated` is always null (maybe just on local?) and you need to make sure that the additional header columns are only shown if necessary

app/webpacker/components/Delegates/DelegatesTable.jsx
```diff
@@ -1,60 +1,68 @@
...
-  delegates, isAdminMode, isAllLeadDelegates, isAllNonLeadDelegates,
+  delegates,
+  isAdminMode,
+  isAllLeadDelegates,
+  isAllNonLeadDelegates,
 }) {
-  const tableData = useMemo(() => delegates.filter(
-    (delegate) => delegate.metadata.status !== 'trainee_delegate' || isAdminMode,
-  ), [delegates, isAdminMode]);
+  const [sortedData, dispatch] = useReducer(sortReducer, {
```
kr-matthews, line 22: I'd rename `sortedData` to `sortState`, or even just use `{ sortColumn, sortDirection }` directly to avoid naming it.

Carter-Thomas:
> I did some clean up and the sorting works. But there are still things that don't work like `total_delegated` is always null (maybe just on local?) and you need to make sure that the additional header columns are only shown if necessary

Can you test the total_delegated on the current version to see if it's still null? Also, the additional header columns that should be shown conditionally are first_delegated, last_delegated, and total_delegated correct?

PR #9274, opened 2024-04-26
Stripe integration for Registration V2

FinnIckler:
How do we handle the client secret? Do we just generate one for the user when he loads the page? Or do I still have to make a clientSecret URL to get it. It might depend on the registration though, so I think we need to do that first


FinnIckler:
I added Donations and getting the `clientSecret` from the API. 
But there is still the issue that I need to set an `initialAmount` for the Payment Element. Do I return that with the Registration after registering? Doesn't sound too nice.

FinnIckler:
Will it also just work like I do currently? I generate the payment intent without the donation amount, and then set it in the js by setting it in the Form.

gregorbg:
> I added Donations and getting the `clientSecret` from the API. But there is still the issue that I need to set an `initialAmount` for the Payment Element. Do I return that with the Registration after registering? Doesn't sound too nice.

I might be misunderstanding, but isn't this exactly what you were doing all along? When "entering" the payment step, you know how much the user should pay. That is by definition the initial amount that you can pre-load the payment element with.

In the past, this used to be the same amount you were requesting a Stripe PI + Client Secret for. The only difference is that now, there may be a positive amount to _add_ to that. But as far as my understanding goes, the "initial amount" or "checkout total" was always fully specified.

gregorbg:
> Will it also just work like I do currently?

...test it out locally? Staring only at the code, it feels like you should also provide a currency to the Stripe Elements panel, but otherwise yeah this should work. You ask the backend for a Stripe PI Client Secret upon the click event, and if the amount+currency of that PI doesn't match the "temporary amount" that is loaded into the Stripe Elements thingy, the Stripe UI will complain.

FinnIckler:
> > I added Donations and getting the `clientSecret` from the API. But there is still the issue that I need to set an `initialAmount` for the Payment Element. Do I return that with the Registration after registering? Doesn't sound too nice.
> 
> I might be misunderstanding, but isn't this exactly what you were doing all along? When "entering" the payment step, you know how much the user should pay. That is by definition the initial amount that you can pre-load the payment element with.
> 
> In the past, this used to be the same amount you were requesting a Stripe PI + Client Secret for. The only difference is that now, there may be a positive amount to _add_ to that. But as far as my understanding goes, the "initial amount" or "checkout total" was always fully specified.

Currently yes, but if we have stuff like per event pricing, then we can't just use the Base Currency as the payment depends on the registration. Should we just deal with that later?

gregorbg:
> Currently yes, but if we have stuff like per event pricing, then we can't just use the Base Currency as the payment depends on the registration. Should we just deal with that later?

Yes. That is a distinct issue of per-event pricing and totally not within the scope of this PR (which is just about "reinstating" the existing payment functionality without per-event or other fancy fun)

FinnIckler:
https://github.com/thewca/worldcubeassociation.org/assets/10168980/24ecf6c5-3090-4972-817c-001e408bb7a8

FinnIckler:
Finally tested and ready for review!

FinnIckler:
TODOS:
- Hide Progress bar after finishing
- Make Info Boxes Green
- Change Subtotal to non input field

FinnIckler:
@thewca-bot deploy staging

FinnIckler:
@thewca-bot deploy staging

FinnIckler:
@thewca-bot deploy staging

app/webpacker/components/RegistrationsV2/Register/StripeWrapper.jsx
```diff
@@ -1,80 +1,78 @@
...
+  }, [connectedAccountId, stripePublishableKey]);
+  const handleDonation = useCallback(async (_, { value: newDonationAmount }) => {
+    const {
+      api_amounts: { stripe: stripeAmount },
+      human_amount: humanAmount,
+    } = await convertISOAmount(initialAmount + newDonationAmount, competitionInfo.currency_code);
+    setAmount(stripeAmount);
+    setDonationAmount(newDonationAmount);
+    setDisplayAmount(humanAmount);
+  }, [competitionInfo.currency_code, initialAmount]);
```
gregorbg, line 57: I'm a bit unhappy about this `async`-heavy handling of the API call. The biggest problem is that the UI stays interactive while the new amount is loading.

If there is a popular competition, it may be conceivable that psych sheets or whatever else is clogging our infrastructure, and the return value from the conversion API might arrive late. The user types in their credit card details first, then at the very last moment while hovering over the "Pay Now" button sees the donation field, and spontaneously decides to give five dollars.
They then proceed to click the button right away (because they had already entered their credit card details before) but are never actually donating anything because our API didn't return the converted currency amount in time.

gregorbg (changes requested):
Mostly worried about side-effects of the donation button. Otherwise seems good to go!

app/webpacker/components/RegistrationsV2/Register/RegistrationOverview.jsx
```diff
@@ -0,0 +1,67 @@
...
+            </Form>
+          </Segment>
+          { editsAllowed && (
+          <Button
+            primary
+            attached
+            onClick={nextStep}
+          >
+            {i18n.t('registrations.update')}
+          </Button>
```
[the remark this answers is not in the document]
FinnIckler: Changed the button to type submit and moved nextStep to onSubmit

app/webpacker/components/RegistrationsV2/Register/PaymentStep.jsx
```diff
@@ -36,27 +75,52 @@ export default function PaymentStep({
     // your `return_url`. For some payment methods like iDEAL, your customer will
     // be redirected to an intermediate site first to authorize the payment, then
     // redirected to the `return_url`.
-    if (error.type === 'card_error' || error.type === 'validation_error') {
-      dispatch(setMessage(error.message, 'error'));
-    } else {
-      dispatch(setMessage('An unexpected error occurred.', 'error'));
+    if (error) {
+      dispatch(setMessage('registrations.payment_form.errors.stripe_failed', 'error'));
+      console.error(error);
```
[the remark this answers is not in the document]
FinnIckler: I left the console.error here intentionally to get the exact Error for debugging, while only showing the user a translated general error message. Or I could show the error message in english directly from the stripe API somewhere?

app/webpacker/components/RegistrationsV2/Register/StripeWrapper.jsx
```diff
@@ -1,80 +1,78 @@
...
-        'negative',
-      ));
-    },
-  });
+const convertISOAmount = async (amount, currency) => {
+  const { data } = await fetchJsonOrError(
+    paymentDenominationUrl(amount, currency),
+  );
+  return data;
+};
```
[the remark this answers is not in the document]
FinnIckler: sounds reasonable

app/webpacker/components/RegistrationsV2/Register/StripeWrapper.jsx
```diff
@@ -46,32 +41,29 @@ export default function StripeWrapper({
...
-      human_amount: humanAmount,
-    } = await convertISOAmount(initialAmount + newDonationAmount, competitionInfo.currency_code);
-    setAmount(stripeAmount);
+
+  const handleDonation = useCallback(async (newDonationAmount) => {
     setDonationAmount(newDonationAmount);
-    setDisplayAmount(humanAmount);
-  }, [competitionInfo.currency_code, initialAmount]);
+    refetch();
+  }, [refetch]);
```
gregorbg, line 48: I don't think a manual `refetch` trigger is necessary.
Just set the new donation amount using `setDonationAmount`. Doing this results in a new query key, which makes `useQuery` fetch new data (or retrieve information from the instance cache) automatically.

gregorbg (changes requested):

gregorbg:
On an unrelated sidenote: There's a bunch of merge conflicts

app/webpacker/components/RegistrationsV2/Register/StripeWrapper.jsx
```diff
@@ -44,8 +44,7 @@ export default function StripeWrapper({
 
   const handleDonation = useCallback(async (newDonationAmount) => {
```
gregorbg, line 45: WIth the recent change abandoning `refetch`, you don't need `async` anymore. You also don't need this callback anymore, as you can just pass around `setDonationAmount` directly.

gregorbg (changes requested):

app/webpacker/components/RegistrationsV2/Register/StripeWrapper.jsx
```diff
@@ -55,7 +51,7 @@ export default function StripeWrapper({
           options={{ amount: data?.api_amounts.stripe ?? initialAmount, currency: competitionInfo.currency_code.toLowerCase(), mode: 'payment' }}
         >
           <PaymentStep
-            handleDonation={handleDonation}
+            handleDonation={setDonationAmount}
```
gregorbg, line 54: You renamed it to `setDonationAmount` in the function component props, but you still pass it as `handleDonation`...

PR #9282, opened 2024-04-30
Change contact form data to store

danieljames-dj:
This is to avoid complications with contact form, with store the state management will become relatively easy.


app/models/website_contact.rb
danieljames-dj, line 1: The structure of the store is something like:
```
{
  userData: {...},
  contactRecipient: ...,
  [contactRecipient]: {...},
}
```

Since the `contactRecipient` selected (ie, resultsTeam or communicationsTeam) is also a key in the store, to make the code easier, I've renamed the inquiry types and localization keys to lowerCamelCase.

app/webpacker/components/ContactsPage/SubForms/CommunicationsTeam/index.jsx
danieljames-dj, line 1: For now, the components CommunicationsTeam, ResultsTeam and SoftwareTeam looks similar because they have same field. I'm not reusing because they will be further customized - like SoftwareTeam form will have the requestId field (which will be autopopulated) and for ResultsTeam, some customizations will be done, etc.

config/locales/en.yml
danieljames-dj, line 1: I know that editing keys will make it painful for translators, but this won't change hereafter. I'm trying to make the same case everywhere so that these need not be hard-coded.

app/webpacker/components/ContactsPage/SubForms/Competition/index.jsx
```diff
@@ -3,25 +3,28 @@ import { FormTextArea } from 'semantic-ui-react';
...
-  const handleFormChange = (_, { name, value }) => setFormValues(
-    { ...formValues, [name]: value },
-  );
   return (
     <>
       <WcaSearch
         label={I18n.t('page.contacts.form.competition.competition.label')}
         name="competition"
-        value={formValues.competition}
+        value={competition.competition}
```
kr-matthews, line 19: If we have `x.x` then I feel like one of the `x`s is probably not named well.

danieljames-dj: You are right, once https://github.com/thewca/worldcubeassociation.org/pull/9280 is merged, I can rebase and change it to `competition.competitionId`.

app/webpacker/components/ContactsPage/index.jsx
```diff
@@ -6,21 +6,42 @@ import ContactForm from './ContactForm';
...
+          name: userDetails?.user?.name,
+          email: userDetails?.user?.email,
+        },
+        contactRecipient: queryParams?.contactRecipient,
+        competition: {
+          competitionId: queryParams?.competitionId,
+        },
+        communications_team: {},
+        results_team: {},
+        software_team: {},
```
gregorbg, line 33: Do you need these empty initializers?

app/webpacker/components/ContactsPage/UserData.jsx
```diff
@@ -1,24 +1,28 @@
...
-export default function UserData({ formValues, setFormValues }) {
-  const handleFormChange = (_, { name, value }) => setFormValues(
-    { ...formValues, [name]: value },
-  );
+export default function UserData({ userDetails }) {
+  const { userData } = useStore();
+  const dispatch = useDispatch();
+  const handleFormChange = (_, { name, value }) => dispatch(updateUserData(name, value));
+
+  if (userDetails) return null;
```
gregorbg, line 12: I don't understand this. If there is `userData` present, return early and therefor don't render the form at all?!

app/webpacker/components/ContactsPage/ContactForm.jsx
```diff
@@ -55,18 +42,17 @@ export default function ContactForm({ userDetails }) {
...
         return Competition;
+      case CONTACT_RECIPIENTS_MAP.communicationsTeam:
+        return CommunicationsTeam;
+      case CONTACT_RECIPIENTS_MAP.resultsTeam:
+        return ResultsTeam;
+      case CONTACT_RECIPIENTS_MAP.softwareTeam:
+        return SoftwareTeam;
       default:
-        return Wct;
+        return null;
```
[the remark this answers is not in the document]
danieljames-dj: If no contactRecipient is selected, then it should not show anything, hence returning null. Another reason is that eslint will complaint if default is not there.

app/webpacker/components/ContactsPage/ContactForm.jsx
```diff
@@ -102,17 +79,12 @@ export default function ContactForm({ userDetails }) {
               name="contactRecipient"
               value={contactRecipient}
               checked={selectedContactRecipient === contactRecipient}
-              onChange={setSelectedContactRecipient}
+              onChange={() => dispatch(updateContactRecipient(contactRecipient))}
```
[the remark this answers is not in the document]
danieljames-dj: Since this is already inside a `map` block, the `contactRecipient` in this scope is the actual recipient that needs to be changed to. Hence, didn't use the parameter from onChange.

app/webpacker/components/ContactsPage/UserData.jsx
danieljames-dj: It won't render the `UserData` part of the form. For example, if logged-in user goes to /contact, they won't have to enter the name & email as we already have that data. At the same time, if not-logged-in user goes to /contact, they have to enter the name & email.

app/webpacker/components/ContactsPage/index.jsx
danieljames-dj: Actually yes, because in reducer, I'm doing things like:
```
{
  ...state,
  results_team: {
    ...state.results_team,
    message: 'blah',
  }
}
```

app/webpacker/components/ContactsPage/store/reducer.js
```diff
@@ -0,0 +1,63 @@
...
+    },
+  }),
+
+  [UpdateSubFormSoftwareTeam]: (state, { payload }) => ({
+    ...state,
+    software_team: {
+      ...state.software_team,
+      [payload.name]: payload.value,
+    },
+  }),
```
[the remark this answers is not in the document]
danieljames-dj: Oh that's a nice idea, done.

app/webpacker/components/ContactsPage/UserData.jsx
gregorbg: Ah, I was confused between `userDetails` and `userData`. Can you find a better name for (current) `userDetails` to make clear that it's essentially just a "pre-loaded version" of `userData`?

app/webpacker/components/ContactsPage/index.jsx
gregorbg: With the new, section-ized approach it might not be necessary. And of course in the reducer you can always fall back to a default value.

app/webpacker/components/ContactsPage/UserData.jsx
danieljames-dj: Done

app/webpacker/components/ContactsPage/index.jsx
danieljames-dj: Done

app/webpacker/components/ContactsPage/ContactForm.jsx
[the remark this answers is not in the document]
gregorbg: This may "accidentally" work, but it's not conceptually correct for reactive components.
You should always use the value "closest" to the update function.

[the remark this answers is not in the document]
danieljames-dj: Okay done

gregorbg (approved):

PR #9283, opened 2024-04-30
Change contact form recipients to lowerCamelCase to match with store

danieljames-dj:
This is a sub-PR of https://github.com/thewca/worldcubeassociation.org/pull/9282, expecting that PR will take some time for the review to finish. Also, merging this PR will make that PR simpler.

### DIFF
--- a/app/webpacker/components/ContactsPage/ContactForm.jsx
+++ b/app/webpacker/components/ContactsPage/ContactForm.jsx
@@ -17,12 +17,15 @@ import Competition from './SubForms/Competition';
 
 const CONTACT_RECIPIENTS = [
   'competition',
-  'communications_team',
-  'results_team',
-  'software',
+  'communicationsTeam',
+  'resultsTeam',
+  'softwareTeam',
 ];
 
-const CONTACT_RECIPIENTS_MAP = _.keyBy(CONTACT_RECIPIENTS);
+const CONTACT_RECIPIENTS_MAP = CONTACT_RECIPIENTS.reduce((acc, recipient) => {
+  acc[recipient] = _.snakeCase(recipient);
+  return acc;
+}, {});
 
 const SUBFORM_DEFAULT_VALUE = {
   competitionId: null,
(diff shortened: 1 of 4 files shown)


config/locales/en.yml
gregorbg, line 1: Not a fan of camelCase in the locale files. YML should use snake_case by convention.

gregorbg (commented):
I understand your intention behind this PR, but perhaps it would be best to stick with snake_case in the backend and consider something like lodash case converters for the frontend?

https://lodash.com/docs/4.17.15#snakeCase
https://lodash.com/docs/4.17.15#camelCase

config/locales/en.yml
danieljames-dj: Changed

danieljames-dj:
I didn't knew about lodash helper method, that helped. Made the changes, thank you.

app/webpacker/components/ContactsPage/ContactForm.jsx
```diff
@@ -16,12 +16,15 @@ import Competition from './SubForms/Competition';
...
+  'communicationsTeam',
+  'resultsTeam',
+  'softwareTeam',
 ];
 
-const CONTACT_RECIPIENTS_MAP = _.keyBy(CONTACT_RECIPIENTS);
+const CONTACT_RECIPIENTS_MAP = CONTACT_RECIPIENTS.reduce((acc, recipient) => {
+  acc[recipient] = _.snakeCase(recipient);
+  return acc;
+}, {});
```
gregorbg, line 27:
```suggestion
const CONTACT_RECIPIENTS_MAP = _.keyBy(CONTACT_RECIPIENTS, _.camelCase);
```

gregorbg: **NOTE** This only works if you undo the diff above and keep `CONTACT_RECIPIENTS` in snake_case.
Lodash only has a method for turning array into object when you create keys of the object and retain the values. There is no equivalent shorthand method for iterating over an arrays as keys and determining values.

gregorbg (changes requested):

app/webpacker/components/ContactsPage/ContactForm.jsx
danieljames-dj: Done, thanks

gregorbg (approved):

PR #9285, opened 2024-04-30
Migrate some teams to groups

danieljames-dj:
This PR is to migrate some teams to groups. The first commit is very mechanical - it's the same steps done for WRT migration (WRT migration went well, there are no issues so far). The same process will be for every team.

I'm doing one team in a commit because otherwise it's very difficult to understand which part caused that error. Once one team successfully passes all tests, I'll start with next team.

The way I'm choosing next team is the team with less number of presence of Team.wxx.

This PR is ready to be merged any time because it's safer to go few teams to production instead of all at once. If I find this PR approved, I'll merge and deploy and do the remaining in a new PR, and if it's not in approved state, I'll work on the next team migration.


danieljames-dj:
@gregorbg I'm thinking to stop here and check if there are any impacts in the production. If everything is good, I'll go ahead with remaining teams. Hence, can you please review this PR when you are free?

gregorbg (approved):

PR #9287, opened 2024-05-01
Added clarification on ScheduleActivity parent object in error messages

dunkOnIT:
Currently, the Schedule validation error is unclear because we don't know what the parent `holder` object is (it's a polymorphic association). This PR adds clarity by giving the class name of the holder object, and the name of the holder object itself if available.

image

### DIFF
--- a/app/models/schedule_activity.rb
+++ b/app/models/schedule_activity.rb
@@ -22,10 +22,10 @@ class ScheduleActivity < ApplicationRecord
   def included_in_parent_schedule
     return unless errors.blank?
     unless start_time >= holder.start_time
-      errors.add(:start_time, "should be after parent's start_time")
+      errors.add(:start_time, "should be after parent's start_time. Parent: #{holder.class}: #{holder&.name}")
     end
     unless end_time <= holder.end_time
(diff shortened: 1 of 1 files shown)


gregorbg:
I don't think this specific update is needed as per my above comment. We always know that the parent in this instance will be another schedule activity, so there's nothing new we can gain from this.

However, what would indeed be interesting is a more detailed error view that doesn't just take **one** error but rather shows **all** errors and lists which error was triggered by which activity. This will be possible after #9185 is merged.

dunkOnIT:
Isn't the parent a polymorphic `holder`, so we don't necessarily know that the parent will be an activity? I also think exposing the name at least makes it obvious which _specific_ parent activity is causing the issue - which is what the user needs to know to resolve this.

dunkOnIT:
To clarify: I may have brought too much attention to the parent object's class, and not enough to the parent object's name. The point is, if a parent object's time is incompatible with the change, the user needs to know what the parent object is in order to resolve it.

If #9185 is imminent I'm happy to wait on the more comprehensive solution.

PR #9308, opened 2024-05-01
Update links from Delegate Crash Course to Delegate Handbook

nsilvestri:
The Delegate Crash Course has been deprecated in favor of the Delegate Handbook. This PR updates existing references to the DCC to the Handbook where it is still relevant, or removes references to the DCC entirely where they are no longer needed.

### DIFF
--- a/app/views/results_submission/_results_submission_panel.html.erb
+++ b/app/views/results_submission/_results_submission_panel.html.erb
@@ -21,8 +21,8 @@
       <p>Make sure the schedule on the WCA website actually reflects what happened during the competition.</p>
       <p>
         Please also make sure to include any other additional details required by the
-        <%= link_to "'Submitting competition results' section of the delegate crash course",
-          panel_delegate_crash_course_path %>.
+        <%= link_to "'Competition Results' section of the Delegate Handbook",
+          "https://documents.worldcubeassociation.org/edudoc/delegate-handbook/delegate-handbook.pdf#competition-results" %>.
       </p>
 
       <%= simple_form_for results_submission, url: competition_submit_results_path do |f| %>
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -407,6 +407,6 @@ Rails.application.routes.draw do
 
   # Deprecated Links
   get 'teams-committees' => redirect('teams-committees-councils')
-  get 'panel/delegate-crash-course' => redirect('panel/delegate#delegate-crash-course')
+  get 'panel/delegate-handbook' => redirect('panel/delegate#delegate-handbook')
   get 'panel' => redirect('panel/staff')
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -407,6 +407,6 @@ Rails.application.routes.draw do
 
   # Deprecated Links
   get 'teams-committees' => redirect('teams-committees-councils')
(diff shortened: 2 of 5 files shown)


danieljames-dj:
Hi @nsilvestri, may I know by when are you planning to get this PR merged? Does WQAC has any deadline for the same?

The reason why I'm asking is, in /panel, Delegate Panel is going to be migrated to a separate react component. I've already raised the PR to migrate Delegate Panel: https://github.com/thewca/worldcubeassociation.org/pull/9332. Once that is merged, you should be able to update delegate-crash-course to delegate-handbook more easily probably.

nsilvestri:
@danieljames-dj this PR will be pending the document being added to wca-documents, hopefully 2ish weeks? Don't let this PR be a blocker for making any changes to the panel, I can adapt to whatever changes happen in the meantime.

config/routes.rb
```diff
@@ -409,6 +409,6 @@
 
   # Deprecated Links
   get 'teams-committees' => redirect('teams-committees-councils')
-  get 'panel/delegate-crash-course' => redirect('panel/delegate#delegate-crash-course')
+  get 'panel/delegate-handbook' => redirect('panel/delegate#delegate-handbook')
```
danieljames-dj, line 412: Only the RHS needs to be changed. The LHS is actually meant to be the link that existed earlier and doesn't exist now, but still supporting in case the users have them bookmarked.

app/views/results_submission/_results_submission_panel.html.erb
```diff
@@ -21,8 +21,8 @@
       <p>Make sure the schedule on the WCA website actually reflects what happened during the competition.</p>
       <p>
         Please also make sure to include any other additional details required by the
-        <%= link_to "'Submitting competition results' section of the delegate crash course",
-          panel_delegate_crash_course_path %>.
+        <%= link_to "'Competition Results' section of the Delegate Handbook",
+          "https://documents.worldcubeassociation.org/edudoc/delegate-handbook/delegate-handbook.pdf#competition-results" %>.
```
danieljames-dj, line 25: 1. I tried opening this link, but it's not scrolling to the section in the handbook. Is this something that is supported in at least some browsers?
2. Having the same URL hardcoded in two places looks bit odd, but I cannot think of a way to solve it especially because there is a hash parameter. Just mentioning here so that you or somebody else will be able to find a solution.

danieljames-dj (commented):
The code looks good to me. I've raised a comment at the place where I have a slight concern but not sure if it's solvable. I would like @gregorbg to have a look to give the final approval.

Thanks @nsilvestri for working on this. 🎉

app/views/results_submission/_results_submission_panel.html.erb
nsilvestri: 1. Yes; on Firefox the link takes me straight to the section. It should be supported in other browses as AFAIK it's the same as heading tag hashes in URLs, but if it doesn't work in some browser configuration that's nbd to me since otherwise the link works and the section is explicitly called out here.
2. I'm not sure how you'd prefer to handle a constant like this across Rails code and React code, open to suggestions if it's a priority fix. Changing this around would also be easy in a subsequent commit if required.

danieljames-dj:
I'm approving this for now, if any solution for the URL comes up, we can take it up later.

danieljames-dj (approved):

PR #9329, opened 2024-05-01
Give Sidekiq more RAM in Staging

FinnIckler:
allocates 500mb more RAM for sidekiq in staging, as we have seen OOM errors when compiling due to more and more React components.


gregorbg:
Yay! Do we need to update the AWS deployment task ID because of the terraform change?

FinnIckler:
I already updated staging to use the latest task definition

PR #9330, opened 2024-05-01
Fix Polling for V2

FinnIckler:
Had to make a infra change on the API Gateway as well which was just whitelisting the User-Agent Header for CORS requests. I don't think that requires a Terraform change, but I'll run terraform plan later and see if something shows up.

Bonus: fix the error with Registration Action


dunkOnIT (approved):
LGTM!

PR #9331, opened 2024-05-01
Remove /teams page

danieljames-dj:
The page worldcubeassociation.org/teams is currently not used by anybody. I was thinking that this page will be used to access Team.banned. But just found that banned competitors are actually accessed from https://www.worldcubeassociation.org/disciplinary. Hence I thought removing this page will be good. Also, this will help to avoid leaders accidentally going to this page (using some bookmarked URL) instead of the new page.


app/controllers/teams_controller.rb
```diff
@@ -2,13 +2,8 @@
 
 class TeamsController < ApplicationController
   before_action :authenticate_user!
-  before_action -> { redirect_to_root_unless_user(:can_manage_teams?) }, except: [:edit, :update]
```
gregorbg, line 5: I think this safeguard should stay in place, with an adequate replacement of `can_manage_teams?`

app/webpacker/lib/requests/routes.js.erb
```diff
@@ -147,8 +147,6 @@ export const subordinateDelegateClaimsUrl = `<%= CGI.unescape(Rails.application.
 export const subordinateUpcomingCompetitionsUrl = `<%= CGI.unescape(Rails.application.routes.url_helpers.competitions_for_senior_path) %>`;
 export const wfcCompetitionsExportUrl = `<%= CGI.unescape(Rails.application.routes.url_helpers.wfc_competitions_export_path) %>`;
 
-export const teamUrl = (id) => `<%= CGI.unescape(Rails.application.routes.url_helpers.team_path("${id}"))%>`;
```
gregorbg, line 150: Have you made sure that this isn't used across our webpacker code?

gregorbg (changes requested):

app/webpacker/lib/requests/routes.js.erb
danieljames-dj: Yes, it's not used anywhere in our codebase.

app/controllers/teams_controller.rb
danieljames-dj: This safeguard was for `def index`, but the `def index` won't exist after this PR is merged, that is the reason I removed it. There is already safeguard for other actions in the next line.

gregorbg: Ah okay, in that case I misread something in the controller. Thanks for confirming!

gregorbg (approved):

PR #9332, opened 2024-05-01
Create separate panel for Delegates

danieljames-dj:
Currently, most of the panels are already migrated to panel. But there are two panels which are not yet migrated to React - staff panel and delegate panel. This PR is to migrate the Delegate Panel to react. The staff panel will be migrated in a later PR.

### DIFF
--- a/app/webpacker/components/Panel/Delegate/ImportantLinks.jsx
+++ b/app/webpacker/components/Panel/Delegate/ImportantLinks.jsx
@@ -1,0 +1,102 @@
+import React from 'react';
+import { Header, List } from 'semantic-ui-react';
+import { allDelegatePageUrl, countryBandsUrl } from '../../../lib/requests/routes.js.erb';
+
+const disciplinaryIncidentsGuidelinesLink = 'https://documents.worldcubeassociation.org/edudoc/guidelines-for-disciplinary-incidents/guidelines-for-disciplinary-incidents.pdf';
+const wrSubmissionFormLink = 'https://docs.google.com/forms/d/e/1FAIpQLSeJwf6b7yGGyWFhU0xfKSF3ki_nITyIKXVFRP86unb9EYRtHQ/viewform';
+const gearOrderFormLink = 'https://forms.gle/owX3ppZahYkoq9s48';
+const travelReimbursementFormLink = 'https://docs.google.com/forms/d/12tz2I_EeBORm14kQO6ZB5TOp321YbXkIXvrxUNIHxN0/viewform';
+const equipmentFundingFormLink = 'https://docs.google.com/forms/d/e/1FAIpQLSebkWMyG2kRzR3cDm3jXFVMFCwd5u4XI6Yt35givu0SOidpHg/viewform';
+const visualGuideForReg3j = 'https://drive.google.com/file/d/1m6THsA8fXRN7QFM4ApJbm6eVODKGbMLx/view';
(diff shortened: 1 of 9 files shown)


config/routes.rb
```diff
@@ -189,9 +189,9 @@
   post 'competitions/:id/disconnect_stripe' => 'competitions#disconnect_stripe', as: :competition_disconnect_stripe
 
   get 'panel' => 'panel#index'
-  get 'panel/delegate-crash-course', to: redirect('https://documents.worldcubeassociation.org/edudoc/delegate-crash-course/delegate_crash_course.pdf', status: 302)
```
gregorbg, line 192: A lot of Delegates / ROs may have this link bookmarked or otherwise use this link in their internal documentation. I don't think we should remove it prematurely.

gregorbg: Okay, so is there any risk associated with merging this PR right now?

danieljames-dj: Nope, it's risk free.

gregorbg: So what will happen with the countless ROs who may have bookmarked this link that you're deleting in this PR?

nsilvestri: Just chiming in on behalf of WQAC, I am comfortable making any necessary changes to #9308 that come as a result of this PR. Keeping the redirect unchanged here and having the discussion when #9308 is ready to merge would also keep this commit focused on the React changes.

danieljames-dj: Thanks Nick, @gregorbg I've added back redirect at the end of routes.rb under "Deprecated Links" section.

app/webpacker/components/Panel/Delegate/ImportantLinks.jsx
kr-matthews, line 1: Could remove a bunch of repetition by defining
```
function ListItemLink({ title, link }) {
        <List.Item>
          <a
            href={link}
            target="_blank"
            rel="noreferrer"
          >
            {title}
          </a>
        </List.Item>
}
```

gregorbg (changes requested):
Technically okay but you might want to implement Kevin's suggested changes before merging

app/webpacker/components/Panel/Delegate/ImportantLinks.jsx
danieljames-dj: That's nice, thanks. Did some more cleanups.

app/webpacker/components/Panel/Delegate/ImportantLinks.jsx
```diff
@@ -0,0 +1,63 @@
...
+  );
+}
+
+export default function ImportantLinks() {
+  return (
+    <>
+      <Header as="h2">Important Links</Header>
+      <List>
+        {IMPORTANT_LINKS.map(({ title, link }) => (
+          <ListItemLink key={title.replace(/ /g, '')} title={title} link={link} />
```
gregorbg, line 58: Why not just take the `link` URL itself as a key?

gregorbg (approved):
Tiny suggestion that really doesn't matter in the grand scheme of things. Feel free to implement it, or to ignore it and merge as-is :D

app/webpacker/components/Panel/Delegate/ImportantLinks.jsx
danieljames-dj: That's cleaner, accepted the suggestion. :-)

PR #9333, opened 2024-05-02
Add preferred contact mode for teams/committees

danieljames-dj:
Currently the email IDs of all the teams/committees are exposed. There can be some teams/committees who prefers not to show their email address instead allow to contact only via contact form. Then there can be some other teams/committees who prefers not to have their email ID in public. To accommodate all their needs, added a column in groups_metadata_teams_committees.


app/models/groups_metadata_teams_committees.rb
```diff
@@ -3,6 +3,12 @@
 class GroupsMetadataTeamsCommittees < ApplicationRecord
   has_one :user_group, as: :metadata
 
+  enum :preferred_contact_mode, {
+    no_public_way: "no_public_way",
```
gregorbg, line 7: Naming choice here feels a bit odd. What about `private`?

app/webpacker/components/TeamsCommitteesCouncils/GroupPage.jsx
```diff
@@ -1,13 +1,20 @@
...
 import useLoadedData from '../../lib/hooks/useLoadedData';
 import Loading from '../Requests/Loading';
 import Errored from '../Requests/Errored';
 import RolesTable from './RolesTable';
 
+function contactRecipientUrlForGroup(group) {
+  if (group.metadata.friendly_id === 'wrt') {
+    return contactRecipientUrl('results_team');
+  }
+  return contactRecipientUrl('communications_team');
```
gregorbg, line 15: Lots of hard-coded strings. At the very least, the references to `results_team` and `communications_team` should be exposed through some `ContactForm` constants.

Ideally, this would use the new "preferred contact mode" feature like so: (pseudo-code)
```
if (group.metadata.preferred_contact_mode === 'private') { // perhaps store string as constant somewhere
  return contactRecipientUrl(group.metadata.friendly_id);
}
return contactRecipientUrl(someFallback);
```

The parameter to pass to `contactRecipientUrl` is a bit tricky here, because we generally use friendly IDs for T/Cs but the contact form expects humanized values instead (might be worth changing this?!)
Also not sure what `fallback` should be or whether we even need it. If a T/C set their preferred communication mode to `private`, they should definitely have a corresponding contact form recipient. So there really shouldn't be the case where we even need a fallback.

config/locales/en.yml
```diff
@@ -788,7 +789,7 @@ en:
         wac: "The WAC is responsible for maintaining feedback and communication between the WCA Community and the WCA Staff. They maintain the WCA Forums and summarize surveys and attempt to increase reach across the worldwide community."
     contacts:
       title: "WCA Contact Form"
-      faq_note_html: 'Before making an inquiry, you may wish to take a look at our <a href="/faq">Frequently Asked Questions</a>! If you wish to contact a specific committee or team not listed here, you can find their contact information on <a href="/teams-committees-councils">this page</a>.'
+      faq_note_html: 'Before making an inquiry, you may wish to take a look at our <a href="/faq">Frequently Asked Questions</a>! If you wish to contact a specific committee or team not listed here, you can find their contact information on <a href="/teams-committees-councils">this page</a> (Some committee/team might have also opted out from contacting, you will not be able to contact them).'
```
gregorbg, line 792: This extra snippet of text should appear on the T/C overview page directly. Perhaps you can change the previously existing "you can find their contact information on" to something like "you can contact them on <a>this page</a>", which still holds true even with email opt-out (we don't guarantee that you will be able to contact them **via email**, but if they opt out you will still be able to ultimately contact them via the form)

db/migrate/20240502013953_add_preferred_contact_mode_to_groups_metadata_teams_committees.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class AddPreferredContactModeToGroupsMetadataTeamsCommittees < ActiveRecord::Migration[7.1]
+  def change
+    add_column :groups_metadata_teams_committees, :preferred_contact_mode, :string, null: false
+    change_column_default :groups_metadata_teams_committees, :preferred_contact_mode, GroupsMetadataTeamsCommittees.preferred_contact_modes[:email]
+    UserGroup.teams_committees_group_wrt.metadata.update!(preferred_contact_mode: GroupsMetadataTeamsCommittees.preferred_contact_modes[:contact_form])
+    UserGroup.teams_committees_group_wct.metadata.update!(preferred_contact_mode: GroupsMetadataTeamsCommittees.preferred_contact_modes[:contact_form])
+    UserGroup.teams_committees_group_wcat.metadata.update!(preferred_contact_mode: GroupsMetadataTeamsCommittees.preferred_contact_modes[:no_public_way])
```
gregorbg, line 9: Follow-up change if you rename the constant as per my other comment above

gregorbg (changes requested):

app/webpacker/components/TeamsCommitteesCouncils/GroupPage.jsx
danieljames-dj: Now I'm thinking should I change the contact recipient to friendly ID, because it's safer to do this now than later. What do you think?

app/models/groups_metadata_teams_committees.rb
danieljames-dj: Changed to no_contact as per discussion.

db/migrate/20240502013953_add_preferred_contact_mode_to_groups_metadata_teams_committees.rb
danieljames-dj: Changed.

gregorbg (approved):

PR #9335, opened 2024-05-02
Renamed contact recipients to match friendly IDs of teams

danieljames-dj:
As discussed in another PR, we decided to rename recipients to match with friendly ID. This will help in many places probably.

### DIFF
--- a/app/webpacker/components/ContactsPage/ContactForm.jsx
+++ b/app/webpacker/components/ContactsPage/ContactForm.jsx
@@ -12,16 +12,16 @@ import UserData from './UserData';
 import Loading from '../Requests/Loading';
 import { useDispatch, useStore } from '../../lib/providers/StoreProvider';
 import { updateContactRecipient } from './store/actions';
-import CommunicationsTeam from './SubForms/CommunicationsTeam';
+import Wct from './SubForms/Wct';
+import Wrt from './SubForms/Wrt';
+import Wst from './SubForms/Wst';
 import Competition from './SubForms/Competition';
-import ResultsTeam from './SubForms/ResultsTeam';
-import SoftwareTeam from './SubForms/SoftwareTeam';
 
 const CONTACT_RECIPIENTS = [
   'competition',
-  'communications_team',
-  'results_team',
-  'software_team',
+  'wct',
+  'wrt',
+  'wst',
 ];
 
 const CONTACT_RECIPIENTS_MAP = _.keyBy(CONTACT_RECIPIENTS, _.camelCase);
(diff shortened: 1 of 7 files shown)


gregorbg (commented):
The general idea is very nice, but now we suddenly have friendly IDs hard-coded in a lot of places. What if one of them changes one day, or becomes obsolete? (T/C being merged/replaced/whatever...)

How hard would it be to re-use the constants from the database? I.e. `UserGroup.results_team.friendly_id` instead of hard-coding `wrt`

danieljames-dj:
I've reused the constants from database for backend code. But didn't do for frontend because I prefer not to expose the groups/roles data through `wca-data.js.erb` because I prefer only API as it goes through filtering. Is it fine to leave frontend as it is?

app/webpacker/components/ContactsPage/ContactForm.jsx
```diff
@@ -42,12 +42,12 @@ export default function ContactForm({ loggedInUserData }) {
...
-      case CONTACT_RECIPIENTS_MAP.resultsTeam:
-        return ResultsTeam;
-      case CONTACT_RECIPIENTS_MAP.softwareTeam:
-        return SoftwareTeam;
+      case CONTACT_RECIPIENTS_MAP.wct:
+        return Wct;
+      case CONTACT_RECIPIENTS_MAP.wrt:
+        return Wrt;
+      case CONTACT_RECIPIENTS_MAP.wst:
+        return Wst;
```
gregorbg, line 50: The components themselves can probably still be called `SoftwareTeam` etc. (Or maybe even `SoftwareTeamContact`?)
I mean, it literally doesn't matter what the functional React component is called.

gregorbg:
> I've reused the constants from database for backend code. But didn't do for frontend because I prefer not to expose the groups/roles data through `wca-data.js.erb` because I prefer only API as it goes through filtering. Is it fine to leave frontend as it is?

Thanks! I guess it would create too much overhead to expose this data to JS directly right now. But we still have to think about how to handle this in the future, especially when we have a fully independent frontend.

app/webpacker/components/ContactsPage/ContactForm.jsx
danieljames-dj: Yes I agree. But I named as the friendly ID to avoid too many types of namings. I'm fine to change to `SoftwareTeamContact` or `WstContact` or whatever you prefers to if needed.

gregorbg (approved):

danieljames-dj:
Thanks for approving Gregor :-)

PR #9336, opened 2024-05-03
Add request ID to WST contact form

danieljames-dj:
For the 404 and 500 page, now users doesn't have to copy the request ID, instead just clicking the contact button will add the request ID through URL.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -658,9 +658,9 @@ en:
       error_pages:
         #context: The error pages in the case of 500s or 400s
         500_error: "We're sorry, but something went wrong."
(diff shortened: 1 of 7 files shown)


config/locales/en.yml
gregorbg, line 1: You're inconsistent about "report it" and "report it **to the Software Team**"

gregorbg (changes requested):

config/locales/en.yml
danieljames-dj: Oh yes, thanks for pointing. Fixed.

gregorbg (approved):

PR #9340, opened 2024-05-03
Migrate WDPC to roles

danieljames-dj:
WDPC is the only team that doesn't exist anymore. We haven't created user_group as well for this team. So this migration is also creating a user_group for WDPC.

### DIFF
--- a/db/migrate/20240503144557_migrate_wdpc_to_groups.rb
+++ b/db/migrate/20240503144557_migrate_wdpc_to_groups.rb
@@ -8,7 +8,7 @@ class MigrateWdpcToGroups < ActiveRecord::Migration[7.1]
     # email was removed in https://github.com/thewca/worldcubeassociation.org/pull/6065, but I don't
     # think there was any reason to remove it. Hence adding it back.
     metadata = GroupsMetadataTeamsCommittees.create!(email: 'dataprotection@worldcubeassociation.org', friendly_id: team.friendly_id)
-    UserGroup.create!(name: team.name, group_type: UserGroup.group_types[:teams_committees], is_active: false, is_hidden: false, metadata: metadata)
-    migrate_team_members_to_group(Team.c_find_by_friendly_id!('wdpc'), GroupsMetadataTeamsCommittees.find_by(friendly_id: 'wdpc').user_group)
+    group = UserGroup.create!(name: team.name, group_type: UserGroup.group_types[:teams_committees], is_active: false, is_hidden: false, metadata: metadata)
+    migrate_team_members_to_group(team, group)
   end
 end
--- a/db/migrate/20240503144557_migrate_wdpc_to_groups.rb
+++ b/db/migrate/20240503144557_migrate_wdpc_to_groups.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+class MigrateWdpcToGroups < ActiveRecord::Migration[7.1]
+  include RoleMigrationHelper
+
+  def change
+    team = Team.find_by_friendly_id('wdpc')
+    # email was removed in https://github.com/thewca/worldcubeassociation.org/pull/6065, but I don't
+    # think there was any reason to remove it. Hence adding it back.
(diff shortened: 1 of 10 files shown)


db/schema.rb
gregorbg, line 1: This has unrelated changes

db/migrate/20240503144557_migrate_wdpc_to_groups.rb
```diff
@@ -0,0 +1,14 @@
...
+class MigrateWdpcToGroups < ActiveRecord::Migration[7.1]
+  include RoleMigrationHelper
+
+  def change
+    team = Team.find_by_friendly_id('wdpc')
+    # email was removed in https://github.com/thewca/worldcubeassociation.org/pull/6065, but I don't
+    # think there was any reason to remove it. Hence adding it back.
+    metadata = GroupsMetadataTeamsCommittees.create!(email: 'dataprotection@worldcubeassociation.org', friendly_id: team.friendly_id)
+    UserGroup.create!(name: team.name, group_type: UserGroup.group_types[:teams_committees], is_active: false, is_hidden: false, metadata: metadata)
+    migrate_team_members_to_group(Team.c_find_by_friendly_id!('wdpc'), GroupsMetadataTeamsCommittees.find_by(friendly_id: 'wdpc').user_group)
```
gregorbg, line 12: You have `Team.find_by_friendly` already stored in a variable above. Just reuse that

db/migrate/20240503144557_migrate_wdpc_to_groups.rb
```diff
@@ -0,0 +1,14 @@
...
+
+class MigrateWdpcToGroups < ActiveRecord::Migration[7.1]
+  include RoleMigrationHelper
+
+  def change
+    team = Team.find_by_friendly_id('wdpc')
+    # email was removed in https://github.com/thewca/worldcubeassociation.org/pull/6065, but I don't
+    # think there was any reason to remove it. Hence adding it back.
+    metadata = GroupsMetadataTeamsCommittees.create!(email: 'dataprotection@worldcubeassociation.org', friendly_id: team.friendly_id)
+    UserGroup.create!(name: team.name, group_type: UserGroup.group_types[:teams_committees], is_active: false, is_hidden: false, metadata: metadata)
```
gregorbg, line 11: This `create` call returns the Rails model that you can store into a variable and use directly below

db/seeds/user_groups.seeds.rb
```diff
@@ -233,14 +233,20 @@
...
     group_type: :teams_committees,
     is_active: true,
     is_hidden: true,
     metadata: GroupsMetadataTeamsCommittees.find_by!(friendly_id: 'wct_china'),
   )
+  UserGroup.create!(
+    name: 'WCA Data Protection Committee',
+    group_type: :teams_committees,
+    is_active: false,
+    is_hidden: false,
```
gregorbg, line 247: uh oh

gregorbg (changes requested):

db/migrate/20240503144557_migrate_wdpc_to_groups.rb
danieljames-dj: Oh ya, thanks. Done.

db/migrate/20240503144557_migrate_wdpc_to_groups.rb
danieljames-dj: Oops, done.

db/schema.rb
danieljames-dj: My mistake, removed 😅

db/seeds/user_groups.seeds.rb
danieljames-dj: It was intentional actually. WDPC 'team' is hidden because it is no longer active but there was no way in teams table to set active status. Here, `is_active` solves that. Hence, if we decide to show past roles of a user in a public page, we have to show their WDPC role (if exists) as well and for that the group should not be hidden.

gregorbg (approved):

PR #9341, opened 2024-05-03
Remove usages of Team.wrc

danieljames-dj:
Team.wrc was used in few places, and not sure if changing them will cause any issues. So thought of making those changes before migrating WRC to roles.

### DIFF
--- a/spec/controllers/teams_controller_spec.rb
+++ b/spec/controllers/teams_controller_spec.rb
@@ -6,31 +6,6 @@ RSpec.describe TeamsController do
   let(:team) { FactoryBot.create :team }
 
   describe 'GET #edit' do
-    context 'when signed in as a team leader without rights to manage all teams' do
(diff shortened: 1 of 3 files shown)


spec/controllers/teams_controller_spec.rb
```diff
@@ -6,31 +6,6 @@
   let(:team) { FactoryBot.create :team }
 
   describe 'GET #edit' do
-    context 'when signed in as a team leader without rights to manage all teams' do
```
danieljames-dj, line 9: This test is removed because the page is now only for editing banned competitors and the test for that already exists below.

Alternatively creating a similar unit test for user_roles will come up once the migration is completed. Currently since the user_roles_controller is in an intermediate stage, it might be difficult to have unit tests there now.

gregorbg (approved):

PR #9342, opened 2024-05-03
Migrate WST to roles

danieljames-dj:
Migrating WST to roles is considered as tricky because an intermediate stage failed (https://github.com/thewca/worldcubeassociation.org/pull/9198). Checking if the full migration will fix this issue.

### DIFF
--- a/spec/controllers/teams_controller_spec.rb
+++ b/spec/controllers/teams_controller_spec.rb
@@ -46,13 +46,6 @@ RSpec.describe TeamsController do
         expect(team.team_members.first.current_member?).to be false
       end
 
-      it 'cannot demote oneself' do
-        admin_team = admin.teams.first
-        patch :update, params: { id: admin_team.id, team: { team_members_attributes: { "0" => { user_id: admin.id, start_date: admin.team_members.first.start_date, end_date: Date.today-1 } } } }
-        admin_team.reload
-        expect(admin_team.team_members.first.end_date).to eq nil
-      end
-
       it 'cannot set start_date < end_date' do
         member = FactoryBot.create :user
         patch :update, params: { id: team, team: { team_members_attributes: { "0" => { user_id: member.id, start_date: Date.today, end_date: Date.today-1, team_leader: false } } } }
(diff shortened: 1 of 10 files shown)


spec/controllers/teams_controller_spec.rb
```diff
@@ -6,31 +6,6 @@
   let(:team) { FactoryBot.create :team }
 
   describe 'GET #edit' do
-    context 'when signed in as a team leader without rights to manage all teams' do
```
danieljames-dj, line 9: Same reason as mentioned here: https://github.com/thewca/worldcubeassociation.org/pull/9341/files#r1589365799.

Both the PRs need this test to be removed for all tests to pass.

spec/controllers/teams_controller_spec.rb
```diff
@@ -71,13 +46,6 @@
         expect(team.team_members.first.current_member?).to be false
       end
 
-      it 'cannot demote oneself' do
```
danieljames-dj, line 74: This feature is still available in banned competitors, but not used now. Also cannot create an alternate test case because banned people cannot even go to the edit page.

spec/models/user_spec.rb
```diff
@@ -454,17 +454,6 @@
     end
   end
 
-  it "#teams and #current_teams return unique team names" do
```
danieljames-dj, line 457: Again removing it because now there is no teams to replace. Also thought of having similar test in roles, but there is no use-case of getting current groups in roles, only current roles are fetched and there it's supposed to return all roles of a group. So replacement may also be not possible.

gregorbg:
Yay, seems to have worked. Is there any specific reason why you're only moving WST in this PR and nobody else?
Also, has the WST Admin sub-team already been moved over?

danieljames-dj:
> Yay, seems to have worked. Is there any specific reason why you're only moving WST in this PR and nobody else? Also, has the WST Admin sub-team already been moved over?

WST alone is moved because WST is the only team which caused problem when migrated Team.wxx usages to UserGroups.teams_commitees_wxx. Also, WST is one of the three remaining teams to be merged and I thought of doing all three in independent PRs. WDPC PR is already created (and is approved, thanks), WRC will come soon.

danieljames-dj:
Forgot to mention, yes WST admin was already migrated in another PR earlier

gregorbg (approved):

PR #9344, opened 2024-05-04
Some backend cleanups related to contacts page

danieljames-dj:
Currently all the contacts are going through website_contact.rb and that will affect scalability. This PR is to introduce new class for WCT Contact, and if this PR is approved, I'll do the similar thing for other contacts as well.

### DIFF
--- a/app/models/contact_form.rb
+++ b/app/models/contact_form.rb
@@ -8,6 +8,8 @@ class ContactForm < MailForm::Base
   attribute :to_email, validate: :validate_to_email
   attribute :subject, validate: true
 
+  attr_accessor :logged_in_email
+
   def validate_your_email
     errors.add(:your_email, I18n.t('common.errors.invalid')) unless ValidateEmail.valid?(your_email)
   end
--- a/app/models/contact_wct.rb
+++ b/app/models/contact_wct.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+class ContactWct < ContactForm
+  attribute :message, validate: true
+
+  def to_email
+    "contact@worldcubeassociation.org"
+  end
+
+  def subject
+    Time.now.strftime("[WCA Website] General Comment by #{name} on %d %b %Y at %R")
+  end
+end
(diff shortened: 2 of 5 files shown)


app/models/contact_form.rb
```diff
@@ -8,6 +8,8 @@ class ContactForm < MailForm::Base
   attribute :to_email, validate: :validate_to_email
   attribute :subject, validate: true
 
+  attr_accessor :logged_in_email
```
gregorbg, line 11: Where is this value being written?

app/views/mail_form/contact.erb
```diff
@@ -13,10 +13,6 @@
   %></p>
 <% end %>
 
-<% if @resource.request_id.present? %>
-  <p><b>Request ID:</b> <%= @resource.request_id %></p>
-<% end %>
```
gregorbg, line 18: Where did this go?

app/models/contact_wct.rb
```diff
@@ -0,0 +1,13 @@
...
+  attribute :message, validate: true
+
+  def to_email
+    "contact@worldcubeassociation.org"
+  end
+
+  def subject
+    Time.now.strftime("[WCA Website] General Comment by #{name} on %d %b %Y at %R")
+  end
+end
```
gregorbg, line 13: Where was this done previously? How can the current form work without this file?

app/models/contact_form.rb
danieljames-dj: It is written from contacts_controller. It was done earlier for WebsiteContact, but I missed to add it to ContactWct. I've added it now.

app/models/contact_wct.rb
danieljames-dj: This was done in website_contact.rb. I didn't remove it from that file because I am thinking to remove that file completely and create a model for each recipient separately once this PR is merged.

app/views/mail_form/contact.erb
danieljames-dj: In the model, I made `request_id` to `attribute`, so it will get added to the email automatically (because all the attributes are appended in this file above).

gregorbg (approved):

PR #9345, opened 2024-05-04
Change panel to staff panel

danieljames-dj:
Currently the panel page contains only staff panel, so renaming it to staff panel. It's not yet migrated to react, I'll do that later (if nobody does for next few days).


app/controllers/panel_controller.rb
```diff
@@ -4,8 +4,8 @@ class PanelController < ApplicationController
   include DocumentsHelper
 
   before_action :authenticate_user!
-  before_action -> { redirect_to_root_unless_user(:staff_or_any_delegate?) }
```
gregorbg, line 7: Not happy if global permissions are removed.

gregorbg (changes requested):

app/controllers/panel_controller.rb
danieljames-dj: Okay, added back, but using `can_access_panel?` method now as that is more appropriate.

gregorbg (approved):

PR #9346, opened 2024-05-04
Re-enable the "Show Registration Status" button in the new Competition List frontend

gregorbg:
Suddenly had an idea to use a piece of data hydration logic from the Registrations microservice to keep registration status data in the cache. It worked for the backend, and then I found out about an awesome `keepPreviousData` placeholder function for the frontend through ChatGPT. (Scary: This isn't even in Tanstack's current documentation. You can only find it as a "sidenote" in the "Migrate to v5" guide. Really impressive that ChatGPT still knew about this!)

Basically, this batches all competition IDs into an "extra data" request that loads only the registration status in an extra request when the button is toggled. If the toggle is off, the network request is turned off with Tanstack's `enabled` option.

The code design for the backend (returning relatively arbitrary Ruby symbols as "registration status") is more of a proof-of-concept and totally up for debate.

### DIFF
--- a/app/webpacker/components/CompetitionsOverview/CompetitionsView.js
+++ b/app/webpacker/components/CompetitionsOverview/CompetitionsView.js
@@ -1,7 +1,7 @@
 import React, {
   useEffect, useMemo, useReducer, useState,
 } from 'react';
-import { useInfiniteQuery, useQuery } from '@tanstack/react-query';
+import { keepPreviousData, useInfiniteQuery, useQuery } from '@tanstack/react-query';
 import { Container } from 'semantic-ui-react';
 
 import I18n from '../../lib/i18n';
@@ -78,7 +78,7 @@ function CompetitionsView() {
 
   const {
     data: compRegistrationData,
-    isPending: regDataIsPending,
+    isFetching: regDataIsPending,
   } = useQuery({
     queryFn: () => fetchJsonOrError(apiV0Urls.competitions.registrationData, {
       headers: {
@@ -89,6 +89,10 @@ function CompetitionsView() {
     }),
     queryKey: ['registration-info', ...compIds],
     enabled: shouldShowRegStatus,
+    // This is where the magic happens: Using `keepPreviousData` makes it so that
+    //   all previously loaded indicators are held in-cache while the fetcher for the next
+    //   batch is running in the background. (Adding comment here because it's not in the docs)
+    placeholderData: keepPreviousData,
     select: (data) => data.data,
   });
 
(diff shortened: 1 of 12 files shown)


app/webpacker/components/CompetitionsOverview/CompetitionsView.js
```diff
@@ -68,12 +68,41 @@ function CompetitionsView() {
...
+      method: 'POST',
+      body: JSON.stringify({ ids: compIds }),
+    }),
+    queryKey: ['registration-info', ...compIds],
+    enabled: shouldShowRegStatus,
+    // This is where the magic happens: Using `keepPreviousData` makes it so that
+    //   all previously loaded indicators are held in-cache while the fetcher for the next
+    //   batch is running in the background. (Adding comment here because it's not in the docs)
+    placeholderData: keepPreviousData,
+    select: (data) => data.data,
```
kr-matthews, line 96: Do we have to have `data.data`? Does one of them have a better name?

kr-matthews (approved):
I can't run locally (see my comment on Slack) so I can't verify that it works, but the diff looks reasonable.

app/webpacker/components/CompetitionsOverview/CompetitionsView.js
gregorbg: This pattern is a common problem with using your own fetcher methods in Tanstack query. It's even present in their own documentation and other files already suffer from it. We can think about a solution once we've made the full switch.

notrichardpeng:
I found [this](https://github.com/TanStack/query/discussions/6460
) thread, seems that `keepPreviousData` is deprecated and this would achieve the same effect without using it:

```
queryClient.setDefaultOptions({
  placeholderData: (prev) => prev
})
```

gregorbg:
> I found [this](https://github.com/TanStack/query/discussions/6460) thread, seems that `keepPreviousData` is deprecated and this would achieve the same effect without using it:
> 
> ```
> queryClient.setDefaultOptions({
>   placeholderData: (prev) => prev
> })
> ```

Interesting, but actually not directly relevant. The deprecated feature is a keyword directly in the `options` object that you pass to `useQuery`. You used to be able to just do
```
const foo = useQuery({
  queryKey: [...],
  queryFn: () => fetch(...),
  keepPreviousData: true,
});
```

This kind of keyword argument has been deprecated in V4 and removed in V5. Now we're doing the following:
```
const foo = useQuery({
  queryKey: [...],
  queryFn: () => fetch(...),
  placeholderData: keepPreviousData,
});
```
...where `keepPreviousData` is an imported function directly from the Tanstack package, that provides similar functionality as the (once deprecated, now removed) `keepPreviousData` keyword. I think, under the hood it's really just a `export const keepPreviousData = (prev) => prev;` kind of dummy implementation, but I don't care much about these implementation details.

All I care about is the fact that it is indeed **not** deprecated and it does what we need :P

PR #9349, opened 2024-05-05
Change contact API response type to formData

danieljames-dj:
The contact form will soon have file attachment feature (one example: while contacting WRT, attach proof), so to accommodate that, converting the response type of the API body from json to formData.

### DIFF
--- a/app/controllers/contacts_controller.rb
+++ b/app/controllers/contacts_controller.rb
@@ -63,9 +63,11 @@ class ContactsController < ApplicationController
 
   def contact
     formValues = JSON.parse(params.require(:formValues), symbolize_names: true)
-    contact_recipient = formValues.fetch(:contactRecipient)
-    contact_params = formValues.fetch(contact_recipient.to_sym)
-    requestor_details = current_user || formValues.fetch(userData)
+    contact_recipient = formValues[:contactRecipient]
+    contact_params = formValues[contact_recipient.to_sym]
+    requestor_details = current_user || formValues[:userData]
+
+    render status: :bad_request, json: { error: "Invalid arguments" } if contact_recipient.nil? || contact_params.nil? || requestor_details.nil?
 
     case contact_recipient
     when UserGroup.teams_committees_group_wct.metadata.friendly_id
--- a/app/controllers/contacts_controller.rb
+++ b/app/controllers/contacts_controller.rb
@@ -62,7 +62,7 @@ class ContactsController < ApplicationController
   end
 
   def contact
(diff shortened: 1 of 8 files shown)


app/controllers/contacts_controller.rb
```diff
@@ -38,9 +38,10 @@ class ContactsController < ApplicationController
   end
 
   def website_create
-    contact_recipient = params.require(:contactRecipient)
-    contact_params = params.require(contact_recipient)
-    requestor_details = current_user || params.require(:userData)
+    formValues = JSON.parse(params.require(:formValues)).deep_symbolize_keys
```
gregorbg, line 41:
```suggestion
    formValues = JSON.parse(params.require(:formValues), symbolize_names: true)
```

app/controllers/contacts_controller.rb
```diff
@@ -38,9 +38,10 @@ class ContactsController < ApplicationController
...
+    contact_recipient = formValues.fetch(:contactRecipient)
+    contact_params = formValues.fetch(contact_recipient.to_sym)
+    requestor_details = current_user || formValues.fetch(userData)
```
gregorbg, line 44: After passing the JSON string through `JSON.parse`, it is a normal plain Ruby object that can be accessed as such: `formValues[:contactRecipient]` instead of `.fetch(...foo...)`

gregorbg (changes requested):

app/controllers/contacts_controller.rb
danieljames-dj: Oh I didn't knew that can be done to "deep" symbolize keys for nested hash. Accepted suggestion, thanks.

app/controllers/contacts_controller.rb
danieljames-dj: Thing is, I want to throw error if the key is not available, hence thought of using this (just like how we use params.require(:foo) instead of params[:foo].

gregorbg: Not sure if I'm happy with that.
The point why we use `params.require` is that it gives meaningful information to the client. (Since `params` is directly provided by the Rails framework, calling `.require(:foo)` when `:foo` is not set returns a proper HTTP error response code with a proper error message to the client)
With `my_hash.fetch(:some_key)` this is not the case so I don't see the big advantage of using it

danieljames-dj: Okay, that makes sense. I've changed it. I've added a line to return bad_request in case some required argument is missed.

gregorbg (approved):

PR #9350, opened 2024-05-05
Migrate cronjob to send teams-committees changes from team to user_group

danieljames-dj:
Changed the cronjob from referring user_groups instead of teams because all the roles are now in groups


gregorbg (commented):
With this new code in place, shouldn't you be able to remove the old code?

danieljames-dj:
> With this new code in place, shouldn't you be able to remove the old code?

Oops I removed and committed, but forgot to push. Just pushed it.

gregorbg (approved):

PR #9359, opened 2024-05-07
Rewrite the new competitions overview table as actual <Table>

gregorbg:
Instead of focusing on List Items that imitate the existing overview down to the precise CSS classes, I think a table (as in, an actual HTML table element) makes more sense.
It gives us more potential to expand features, makes it easier to add columns, can possibly support sorting by clicking on the header row, etc...


gregorbg:
image

coder13:
What does this change look like on mobile?

gregorbg:
Like this!
image

kr-matthews:
1. I feel like there should be a separate loading indicator per table.

image

2. If it loads some competitions, then I play around with filters, I get a similar loading view as the original page load (above), except now the "v 0" tags are present. The inconsistency feels weird to me.

image

3. The icon makes me think that clicking it will scroll me down. I think a more standard thing to do is "20+"?

image

4. There seems to be a bug where selecting "All" events includes 333mbo and the other old events, so if I click "All" then start turning off individual events until they're all off, no competitions show up. Not related to this PR, though.

image

kr-matthews:
5. I was playing around with ordering (by announcement, then present, then recent) and got a crash to do with trying to `.map` something undefined. But I can't reproduce...

6. I think the icons would look better if they were left-aligned.

image

gregorbg:
Thanks for the extensive review including screenshots!
I've tried to address most of your concerns, the bugfix for the "`.map` something undefined" crash is just a vague guess though as long as there is no clear reconstruction

app/webpacker/components/CompetitionsOverview/ListView.js
```diff
@@ -18,34 +18,37 @@ function ListView({
   hasMoreCompsToLoad,
 }) {
   const { ref: bottomRef, inView: bottomInView } = useInView();
+
   useEffect(() => {
     if (hasMoreCompsToLoad && bottomInView) {
       fetchMoreCompetitions();
     }
-  }, [bottomInView, hasMoreCompsToLoad, fetchMoreCompetitions]);
+  }, [bottomInView, hasMoreCompsToLoad, fetchMoreCompetitions, competitions]);
```
kr-matthews, line 26: An unused dependency is not necessarily wrong, but generally not a good sign I think.

kr-matthews (commented):
I don't have time to run locally today, can do tomorrow though.

gregorbg:
Screenshots of the current working state:

## Desktop
image

## Tablet
image

## Mobile
image

app/webpacker/components/CompetitionsOverview/ListView.js
gregorbg: I get your thinking, and I'm open to suggestions for improvement. Here's the context of this hook:

Tables in the main body (i.e. the actual list of competitions) are supposed to have infinity scrolling. Usually this works by detecting whether the `ref` that we point to the end of the table is in view. The `useInView` hook from `react-intersection-observer` works as expected in this regard.

The problem is that the `useEffect` hook only triggers when one of its dependencies _changes_. If your screen is so large, that loading the next paginated batch of competitions does *not* push the footer of the table out of view (because they all fit on your current screen), the `bottomInView` variable never _changes_. It _stays_ in view but we still want to trigger the "load more" hook.

kr-matthews: Thanks. Makes sense. I think a brief comment in the code would be appropriate.

The only alternative that I can think of is providing some kind of `onSuccess` (or equivalent) callback to the query, which would check `bottomInView` and fetch again if necessary. I don't know if that's possible, or if it would even be better though.

kr-matthews (commented):
My concerns were addressed and things work well locally. I would like that brief comment about the "extra" dependency added.

However, I'm not seeing any competitions on the map. I tried to run `main` to see if it's just this PR, but then docker got messed up somehow and I haven't fixed it yet, so I'm not sure.

gregorbg:
> However, I'm not seeing any competitions on the map.

That's odd, I tried locally and the map works as expected (both on `main` and on this PR branch). So it might be a fluke?

gregorbg:
Merging because this page is still in beta development state and changes won't be reflected on the main website for the average user (plus I basically got Kevin's okay on the code)

PR #9360, opened 2024-05-07
Reset AUTO INCREMENT counter upon CAD computation

gregorbg:
I just noticed that the values in the dev export are getting frighteningly high, because they just kept increasing and increasing every time CAD is run.

This is more of an immediate fix to get us out of the "id column int overflow" hot waters. In the long run, I'd be interested to know why the CAD tables even have an auto increment primary key in the first place.


gregorbg:
Confirmed with WRT that this is okay

PR #9362, opened 2024-05-08
Fix remaining Registration V2 issues

FinnIckler:
About halfway done going through https://docs.google.com/spreadsheets/d/1VQXDGVdYgmoH5hWpUI6wzla6NhZ4rMGbuDcHOc9ODH8/edit#gid=1116831171

Am I correct using
```
pending_registrations = if @competition.uses_new_registration_service?
                                      Microservices::Registrations.registrations_by_competition(@competition.id, 'pending', cache: false)
                                      else
                                        @competition.registrations.pending.count
                                      end
```
here? I am still not 100% sure when to use microservice_registrations instead of calling directly.

### DIFF
--- a/app/views/competitions/_nav.html.erb
+++ b/app/views/competitions/_nav.html.erb
@@ -51,7 +51,7 @@
           }
           if @competition.use_wca_registration?
             pending_registrations = if @competition.uses_new_registration_service?
-                                      Microservices::Registrations.registrations_by_competition(@competition.id, 'pending', cache: false)
+                                      Microservices::Registrations.registrations_by_competition(@competition.id, 'pending', cache: true).length
                                       else
                                         @competition.registrations.pending.count
                                       end
--- a/app/views/competitions/_nav.html.erb
+++ b/app/views/competitions/_nav.html.erb
@@ -50,11 +50,16 @@
               ] : [])
           }
           if @competition.use_wca_registration?
+            pending_registrations = if @competition.uses_new_registration_service?
+                                      Microservices::Registrations.registrations_by_competition(@competition.id, 'pending', cache: false)
+                                      else
+                                        @competition.registrations.pending.count
+                                      end
             nav_items << {
               text: t('.menu.registration'),
               path: competition_edit_registrations_path(@competition),
               icon: "list ul",
-              notification_count: @competition.registrations.pending.count,
+              notification_count: pending_registrations,
               children: [{
                            hidden: true,
                            is_active: lambda do
--- a/app/views/registrations/register.html.erb
+++ b/app/views/registrations/register.html.erb
@@ -31,13 +31,13 @@
       <% end %>
     <% end %>
   <% else %>
-    <% if @competition.uses_new_registration_service? && ( @competition.registration_opened? || @competition.user_can_pre_register?(current_user)) %>
+    <% user_may_register = current_user.cannot_register_for_competition_reasons(@competition).length == 0 %>
+    <% if @competition.uses_new_registration_service? && user_may_register && ( @competition.registration_opened? || @competition.user_can_pre_register?(current_user)) %>
       <%= react_component('RegistrationsV2/Register', { competitionInfo: @competition.to_competition_info ,
                                                         userInfo: @current_user,
                                                         preferredEvents: @current_user.preferred_events.pluck(:id),
                                                         }) %>
     <% else %>
-      <% user_may_register = current_user.cannot_register_for_competition_reasons(@competition).length == 0 %>
       <% if @registration.show_details?(current_user) %>
         <p><%= t 'registrations.greeting', name: current_user.name %></p>
       <% end %>
--- a/app/views/registrations/register.html.erb
+++ b/app/views/registrations/register.html.erb
@@ -31,7 +31,7 @@
       <% end %>
     <% end %>
   <% else %>
-    <% if @competition.uses_new_registration_service? %>
+    <% if @competition.uses_new_registration_service? && @competition.registration_opened? %>
       <%= react_component('RegistrationsV2/Register', { competitionInfo: @competition.to_competition_info ,
                                                         userInfo: @current_user,
                                                         preferredEvents: @current_user.preferred_events.pluck(:id),
--- a/app/views/registrations/register.html.erb
+++ b/app/views/registrations/register.html.erb
@@ -34,8 +34,8 @@
     <% user_may_register = current_user.cannot_register_for_competition_reasons(@competition).length == 0 %>
     <% if @competition.should_render_register_v2?(current_user) %>
       <%= react_component('RegistrationsV2/Register', { competitionInfo: @competition.to_competition_info ,
-                                                        userInfo: @current_user,
-                                                        preferredEvents: @current_user.preferred_events.pluck(:id),
+                                                        userInfo: current_user,
+                                                        preferredEvents: current_user.preferred_events.pluck(:id),
                                                         }) %>
     <% else %>
       <% if @registration.show_details?(current_user) %>
--- a/app/views/registrations/register.html.erb
+++ b/app/views/registrations/register.html.erb
@@ -31,7 +31,7 @@
       <% end %>
     <% end %>
   <% else %>
-    <% if @competition.uses_new_registration_service? && @competition.registration_opened? %>
+    <% if @competition.uses_new_registration_service? && ( @competition.registration_opened? || @competition.user_can_pre_register?(current_user)) %>
       <%= react_component('RegistrationsV2/Register', { competitionInfo: @competition.to_competition_info ,
                                                         userInfo: @current_user,
                                                         preferredEvents: @current_user.preferred_events.pluck(:id),
--- a/app/views/registrations/register.html.erb
+++ b/app/views/registrations/register.html.erb
@@ -32,7 +32,7 @@
     <% end %>
   <% else %>
     <% user_may_register = current_user.cannot_register_for_competition_reasons(@competition).length == 0 %>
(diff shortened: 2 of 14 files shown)


app/views/competitions/_nav.html.erb
```diff
@@ -50,11 +50,16 @@
               ] : [])
           }
           if @competition.use_wca_registration?
+            pending_registrations = if @competition.uses_new_registration_service?
+                                      Microservices::Registrations.registrations_by_competition(@competition.id, 'pending', cache: false)
```
gregorbg, line 54: This is lacking a `.length` or something. Also, why are you not caching?

app/webpacker/components/RegistrationsV2/Register/index.jsx
gregorbg, line 1: Can we set these as global defaults somewhere?
According to Tanstack documentation, you can pass these options directly to the `new QueryClient()` constructor that you provide in the `QueryClientProvider` context.

app/views/registrations/register.html.erb
```diff
@@ -31,13 +31,13 @@
       <% end %>
     <% end %>
   <% else %>
-    <% if @competition.uses_new_registration_service? %>
+    <% user_may_register = current_user.cannot_register_for_competition_reasons(@competition).length == 0 %>
+    <% if @competition.uses_new_registration_service? && user_may_register && ( @competition.registration_opened? || @competition.user_can_pre_register?(current_user)) %>
```
gregorbg, line 35: This is quite a lot of combined checks. Can we move them either into the controller or (better yet!) into the model?

gregorbg (commented):
About using the Microservice interface vs. using the hacked `microservice_registrations` relation: The answer is quite simply that it depends on your filtering needs.
The Rails relation only exists on competitions `foo_competition.microservice_registrations` and users `current_user.microservice_registrations`. The table is only a shallow "there exists something" table, and it does not contain actual data about the registration. As such, you can **not** do things like `current_user.microservice_registrations.where(event_id: '333')`.
If you need this kind of "complex" query, you have to use the Microservice connector directly.

app/views/competitions/_nav.html.erb
FinnIckler: if we cache, this will not be accurate right?

gregorbg: Oh, the name `cache: true` might be a bit misleading here. It controls whether the data returned from the Microservice should also be _written to_ the caching tables or not.

app/views/registrations/register.html.erb
FinnIckler: something like `@competition.should_show_register_v2_panel?(current_user) ` ?

app/webpacker/components/RegistrationsV2/Register/index.jsx
FinnIckler: I created a WCAQueryClientProvider

app/webpacker/components/CompetitionsOverview/CompetitionsFilters.js
```diff
@@ -81,21 +81,23 @@ export function EventSelector({
   onEventSelection,
   eventList = WCA_EVENT_IDS,
   disabled = false,
+  maxEvents = WCA_EVENT_IDS.length,
```
kr-matthews, line 84: Would infinity work better as a default? This default is kind of assuming that the default `eventList` is being used.

app/webpacker/components/CompetitionsOverview/CompetitionsFilters.js
```diff
@@ -81,21 +81,23 @@ export function EventSelector({
...
 }) {
   return (
     <>
       <label htmlFor="events">
         {`${I18n.t('competitions.competition_form.events')}`}
         <br />
-        <Button primary type="button" size="mini" id="select-all-events" onClick={() => onEventSelection({ type: 'select_all_events' })}>{I18n.t('competitions.index.all_events')}</Button>
+        <Button disabled={disabled || selectedEvents.length >= maxEvents} primary type="button" size="mini" id="select-all-events" onClick={() => onEventSelection({ type: 'select_all_events' })}>{I18n.t('competitions.index.all_events')}</Button>
```
kr-matthews, line 91: I'm a little confused why this file in the `CompetitionsOverview` has a translation from the `competition_form`.

kr-matthews, line 91: Why is the select-all button disabled depending on how many events the user has selected. Should it only be disabled if the `eventList` length is longer than the max events?

(And even then, it might still be quickest to select all events then deselect those you don't want, even if that temporarily goes over the max?)

app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
```diff
@@ -183,7 +188,7 @@ export default function CompetingStep({
 
   const handleEventSelection = ({ type, eventId }) => {
```
kr-matthews, line 189: This method is a bit inconsistent now. It's enforcing the max events only for `select_all_events` but not for `toggle_event`.

app/webpacker/components/CompetitionsOverview/CompetitionsFilters.js
FinnIckler: sure, but WCA_EVENT_IDS.length is practically Infinity

app/webpacker/components/CompetitionsOverview/CompetitionsFilters.js
FinnIckler: I think this component was just moved to the competition form folder

FinnIckler: yeah, I will just disable it from the start and add a text underneath it.

app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
FinnIckler: true, but that is prevented in the component

app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
```diff
@@ -380,7 +386,10 @@ export default function CompetingStep({
               fluid
               icon
               labelPosition="left"
-              disabled={isCreating}
+              disabled={
```
FinnIckler, line 389: Switch to onSubmit and button type submit

app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
```diff
@@ -283,8 +289,8 @@ export default function CompetingStep({
                 setGuests(Number.parseInt(data.value, 10));
               }}
               min="0"
-              label={<Label>Guests</Label>}
-              max={competitionInfo.guests_per_registration_limit ?? 10}
+              label={<Label>{i18n.t('activerecord.attributes.registration.guests')}</Label>}
+              max={competitionInfo.guests_per_registration_limit ?? Infinity}
```
FinnIckler, line 293: remove ?? Infinity

FinnIckler:
@thewca-bot deploy staging

FinnIckler:
@thewca-bot deploy staging

FinnIckler:
@thewca-bot deploy staging

app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
```diff
@@ -104,31 +106,27 @@ export default function CompetingStep({
 
   const hasChanges = hasEventsChanged || hasCommentChanged || hasGuestsChanged;
 
-  const commentIsValid = comment.trim() || !competitionInfo.force_comment_in_registration;
```
kr-matthews, line 107: Is this being checked somewhere else?

FinnIckler: yes, this is now checked by the native form element by setting the comment to required and the button to type submit

app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
```diff
@@ -37,9 +37,8 @@ export default function CompetingStep({
 
   const [comment, setComment] = useState('');
   const [selectedEvents, setSelectedEvents] = useState(
-    preferredEvents
-      .filter((event) => competitionInfo.event_ids.includes(event))
-      .slice(0, maxEvents),
+    !competitionInfo.events_per_registration_limit ? preferredEvents
+      .filter((event) => competitionInfo.event_ids.includes(event)) : [],
```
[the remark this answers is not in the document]
FinnIckler: :+1:

app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
```diff
@@ -237,29 +235,41 @@ export default function CompetingStep({
...
+          event.preventDefault();
+          if (shouldShowUpdateButton) {
+            attemptAction(actionUpdateRegistration, { checkForChanges: true });
+          } else if (shouldShowReRegisterButton) {
+            attemptAction(actionReRegister);
+          } else {
+            attemptAction(actionCreateRegistration);
+          }
+        }}
+        >
```
gregorbg, line 248: Move this into its own callback `const` (and potentially even wrap in `useCallback`? Although that's optional if you feel that it doesn't make sense)

gregorbg (changes requested):
One nitpick about inlining long method in React JSX body. Otherwise good to go :+1:

gregorbg (approved):

Issue #9364, opened 2024-05-08
New events & schedule components don't display cumulative time limits


maxidragon:
Example: [Ostróda Open 2024](https://www.worldcubeassociation.org/competitions/OstrodaOpen2024#competition-events)

PR #9365, opened 2024-05-09
Localize time limits including cumulative events

gregorbg:
Fixes https://github.com/thewca/worldcubeassociation.org/issues/9364


kr-matthews:
The competition tab table view also show time limits, but hasn't been updated.

gregorbg:
> The competition tab table view also show time limits, but hasn't been updated.

Good call, thanks! Fixed.

gregorbg:
Merging to avoid further witch hunt

PR #9368, opened 2024-05-09
Fixes #9361: Add confirmation message when email is sent & clears state on submit

### LINKED ISSUE #9361: Contact form doesn't confirm form submission
We're having a lot of duplicate submissions from people who don't realise their form was sent when they click "submit" - we need to display some sort of "Submitted successfully" feedback

danieljames-dj:
This PR adds confirmation message when a form is submitted. This PR also clears the state when the form is submitted.

### DIFF
--- a/app/webpacker/components/ContactsPage/ContactForm.jsx
+++ b/app/webpacker/components/ContactsPage/ContactForm.jsx
@@ -44,6 +44,11 @@ export default function ContactForm({ loggedInUserData }) {
     setContactSuccess(true);
   };
 
+  const recipientChangeHandler = (__, { value }) => {
+    setContactSuccess(false);
+    dispatch(updateContactRecipient(value));
+  };
+
   const SubForm = useMemo(() => {
     if (!selectedContactRecipient) return null;
     switch (selectedContactRecipient) {
@@ -93,7 +98,7 @@ export default function ContactForm({ loggedInUserData }) {
                 name="contactRecipient"
                 value={contactRecipient}
                 checked={selectedContactRecipient === contactRecipient}
-                onChange={(__, { value }) => dispatch(updateContactRecipient(value))}
+                onChange={recipientChangeHandler}
               />
             </FormField>
           ))}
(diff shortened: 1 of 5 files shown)


app/webpacker/components/ContactsPage/ContactForm.jsx
```diff
@@ -30,13 +30,20 @@ export default function ContactForm({ loggedInUserData }) {
   const { save, saving } = useSaveAction();
   const [captchaValue, setCaptchaValue] = useState();
   const [captchaError, setCaptchaError] = useState(false);
+  const [contactSuccess, setContactSuccess] = useState(false);
```
kr-matthews, line 33: This is never rest to false. Maybe it should be reset when the user clicks the send button? Or when they edit any form values? Or after x seconds? Probably the first is easiest/best.

danieljames-dj: Done, took the first one :-)

dunkOnIT:
Have done a basic smoketest locally, functionality looks good! 

* Alternate wording suggestion: "Email sent - we'll get back to you soon."
  * "successfully" is a bit redundant/sounds awkward to me
* one thought I had that you can ignore (just putting it out there as a thought): intuitively, I'm expecting to see the "sent" banner above the submit button, not at the top of the page. Not even sure if it's technically possible to have it above the "submit" button even if we did want it there. I did miss it initially when I submitted, though.

danieljames-dj:
@dunkOnIT thanks for your suggestions. I've made the wording change. I'm not sure of the second part. I'm fine with doing it, I'm very bad in deciding the UI stuffs. I will wait for a 3rd or 4th opinion as well on this. If this is something that we need to implement, I'm happy to do it, just a small change.

dunkOnIT:
Sounds good! If it ends up that my second comment is the only thing blocking this PR, let's just merge it and see how users respond - having _some_ confirmation and clearing the form will be 90% of the win here, my nits definitely shouldn't slow it down.

gregorbg (approved):

PR #9371, opened 2024-05-10
Add localizers for Cutoff and AdvancementCondition

gregorbg:
Stole the Ruby code from `lib/advancement_conditions` and `lib/cutoff.rb`


dunkOnIT (approved):
LGTM
* Tested locally, and the display issue is resolved
* Done a pretty thorough review, and the code all makes sense (obvious disclaimer: I'm not a React-head)

Only thing that doesn't make sense to me is how the booleans on wcaEvent (eg, `wcaEvent.isTimedEvent` are resolved - but they clearly work, seeing as the code works locally. 

(I tried to follow the logic through to `wcif.js` and then to `wca-data.js.erb`, but line 82 stumps me because there doesn't seem to be a `to_json` function defined for the Event model (looks like`official` is just a scope):
```ruby
  official: _.map(<%= Event.official.to_json.html_safe %>, extendEvents),
```
)

gregorbg:
> LGTM
> 
> * Tested locally, and the display issue is resolved
> * Done a pretty thorough review, and the code all makes sense (obvious disclaimer: I'm not a React-head)
> 
> Only thing that doesn't make sense to me is how the booleans on wcaEvent (eg, `wcaEvent.isTimedEvent` are resolved - but they clearly work, seeing as the code works locally.
> 
> (I tried to follow the logic through to `wcif.js` and then to `wca-data.js.erb`, but line 82 stumps me because there doesn't seem to be a `to_json` function defined for the Event model (looks like`official` is just a scope):
> 
> ```ruby
>   official: _.map(<%= Event.official.to_json.html_safe %>, extendEvents),
> ```
> 
> )

`to_json` exists on every Ruby object. It doesn't need to be defined explicitly.

FinnIckler:
*rails Object. Ruby does not provide this natively, but Rails injects it in everything

dunkOnIT:
my dumb ass seeing `to_json` and thinking `to_wcif` lol

PR #9375, opened 2024-05-11
Add 2024-01-01 Regulations files

Randomno:
Looks like there's a PR https://github.com/thewca/wca-regulations/pull/1191 to make this process automatic.


FinnIckler:
Yes, because we are using S3 now doing it this way would do nothing. I'll pick up the other PR instead

PR #9377, opened 2024-05-13
Some cleanups related to groups

danieljames-dj:
Some of the cleanups included in this PR are:
1. Removal of some static methods like UserGroup.delegate_region_groups, in favor of rails-provided static methods like UserGroup.delegate_regions.
2. In User.leader_senior_voters, use associations to make the fetch more efficient.
3. In RolesMetadataTeamsCommittees, avoid using class name to fetch statuses.
4. Use associations for active_roles in user_group
5. Remove unused list API from routes.js.erb

### DIFF
--- a/app/models/roles_metadata_delegate_regions.rb
+++ b/app/models/roles_metadata_delegate_regions.rb
@@ -11,5 +11,8 @@ class RolesMetadataDelegateRegions < ApplicationRecord
 
   has_one :user_role, as: :metadata
   has_one :group, through: :user_role
+  has_one :user, through: :user_role
   has_one :delegate_region, through: :group, source: :metadata, source_type: "GroupsMetadataDelegateRegions"
+
+  scope :senior_delegate, -> { where(status: statuses[:senior_delegate]) }
 end
--- a/app/models/roles_metadata_delegate_regions.rb
+++ b/app/models/roles_metadata_delegate_regions.rb
@@ -13,6 +13,4 @@ class RolesMetadataDelegateRegions < ApplicationRecord
   has_one :group, through: :user_role
   has_one :user, through: :user_role
   has_one :delegate_region, through: :group, source: :metadata, source_type: "GroupsMetadataDelegateRegions"
-
-  scope :senior_delegate, -> { where(status: statuses[:senior_delegate]) }
 end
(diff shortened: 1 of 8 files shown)


app/models/roles_metadata_delegate_regions.rb
```diff
@@ -11,5 +11,8 @@ class RolesMetadataDelegateRegions < ApplicationRecord
 
   has_one :user_role, as: :metadata
   has_one :group, through: :user_role
+  has_one :user, through: :user_role
   has_one :delegate_region, through: :group, source: :metadata, source_type: "GroupsMetadataDelegateRegions"
+
+  scope :senior_delegate, -> { where(status: statuses[:senior_delegate]) }
```
gregorbg, line 17: I am 97% sure you're getting this scope "for free" because you've declared an `enum` with key `senior_delegate: ...foo...` entry above. In other words, `RolesMetadataDelegateRegions.senior_delegate` should exist and work as intended even without defining this scope.

app/models/user_group.rb
```diff
@@ -60,7 +51,7 @@ def users
   end
 
   def active_users
-    self.active_roles.map { |role| role.user }
+    self.active_roles.includes(:user).map(&:user)
```
gregorbg, line 54: I think you can realise this as a relation above.

```ruby
has_many :users, through: :roles
has_many :active_users, through: :active_roles
```

app/models/user_group.rb
```diff
@@ -198,7 +173,7 @@ def senior_delegate
   end
 
   def lead_role
-    self.active_roles.find { |role| role.is_lead? }
+    active_roles.includes(:group, metadata: [:status]).find { |role| role.is_lead? }
```
gregorbg, line 176: Have you tried locally whether this works? Normally, loading polymorphic associations like `metadata` through an `includes` throws errors...

gregorbg (changes requested):

app/models/user_group.rb
gregorbg: Scrap that, I just tried this and surprisingly it worked :O

app/models/roles_metadata_delegate_regions.rb
danieljames-dj: Oh yes, this worked. Thank you.

app/models/user_group.rb
danieljames-dj: Thanks. `has_many :active_users, through: :active_roles, source: :user, class_name: "User"` worked.

app/models/user_group.rb
danieljames-dj: Yes it worked, and as per our discussion in slack, removed :status.

gregorbg (approved):

PR #9378, opened 2024-05-13
Some cleanups related to groups

danieljames-dj:
The cleanups done in this PR are:
- Move filter & sort of roles to UserRole, because `roles_of_group_type` will be moved to a static method in user_group, and that method can be used from different controllers.
- Created few more associations.
- Removed an unused scope in user.rb.

### DIFF
--- a/app/models/user_group.rb
+++ b/app/models/user_group.rb
@@ -19,8 +19,12 @@ class UserGroup < ApplicationRecord
   has_many :direct_child_groups, class_name: "UserGroup", inverse_of: :parent_group, foreign_key: "parent_group_id"
   has_many :roles, foreign_key: "group_id", class_name: "UserRole"
   has_many :active_roles, -> { active }, foreign_key: "group_id", class_name: "UserRole"
+  has_many :roles_of_direct_child_groups, through: :direct_child_groups, source: :roles
+  has_many :active_roles_of_direct_child_groups, -> { active }, through: :direct_child_groups, source: :roles
   has_many :users, through: :roles
-  has_many :active_users, through: :active_roles, source: :user, class_name: "User"
+  has_many :active_users, through: :active_roles, source: :user
+  has_many :users_of_direct_child_groups, through: :roles_of_direct_child_groups, source: :user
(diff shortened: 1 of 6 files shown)


app/models/user_group.rb
```diff
@@ -19,8 +19,12 @@ class UserGroup < ApplicationRecord
   has_many :direct_child_groups, class_name: "UserGroup", inverse_of: :parent_group, foreign_key: "parent_group_id"
   has_many :roles, foreign_key: "group_id", class_name: "UserRole"
   has_many :active_roles, -> { active }, foreign_key: "group_id", class_name: "UserRole"
+  has_many :roles_of_direct_child_groups, through: :direct_child_groups, source: :roles
+  has_many :active_roles_of_direct_child_groups, -> { active }, through: :direct_child_groups, source: :roles
```
gregorbg, line 23: Maybe something like `direct_child_roles` and `active_direct_child_roles`? Not 100% confident about my suggestions, but your current naming scheme is very long as well...

app/models/user_group.rb
```diff
@@ -19,8 +19,12 @@ class UserGroup < ApplicationRecord
...
   has_many :users, through: :roles
-  has_many :active_users, through: :active_roles, source: :user, class_name: "User"
+  has_many :active_users, through: :active_roles, source: :user
+  has_many :users_of_direct_child_groups, through: :roles_of_direct_child_groups, source: :user
+  has_many :active_users_of_direct_child_groups, through: :active_roles_of_direct_child_groups, source: :user
```
gregorbg, line 27: See above for alternative naming suggestions.

danieljames-dj: Done.

app/models/user_group.rb
danieljames-dj: Okay, accepted the name suggestions.

gregorbg (approved):

gregorbg:
Looks okay but has merge conflicts :sweat_smile:

danieljames-dj:
> Looks okay but has merge conflicts 😅

Oops, resolved. Thanks for approving.

PR #9355, opened 2024-05-05
Cleanups related to roles

danieljames-dj:
Some more cleanups are now done. Main things were removal of some unused functions. Also, added roles & active_roles as associations, so method is no longer needed.

PR #9382, opened 2024-05-14
Prevent n+1 query using delegate_role_metadata

viroulep:
It's slightly noticeable on prod when loading a WCIF with a large number of competitors, but in development it's obvious that GET-ing `/api/v0/competitions/CompID/wcif` is unbearably slow!

The WCIF for Euro2024 is currently taking 30s to show on my computer, and this PR put it back to the "usual" 2s (which looks good to me given the overall size of the competition).

Looking at the log without this fix it's pretty obvious what is going on: `Person`'s [to_wcif](https://github.com/thewca/worldcubeassociation.org/blob/1bf137caa153076fe4a4abf077d506b8e9d88614/app/models/user.rb#L1149) uses the competition's `trainee_delegates` which itself uses `trainee_delegate?` which uses `delegate_role_metadata`, and since it's not an eager loaded association it results in a n+1 query.
Since the `delegate_role_metadata` seems like something we pretty much always want if we want the `delegates`, I have added the `includes` directly on the association; an alternative would be to specify it explicitly at all the place we do need that information.

Interestingly, it was not enough to remove the n+1 query: rails translates `delegate_role_metadata.trainee_delegate.any?` into a `select 1 ...`, which is smart if you don't already have the association loaded or if there are a lot of associated rows, but in this case it also ignores the eager loaded data.
I saw elsewhere that `RolesMetadataDelegateRegions.statuses[:trainee_delegate]` was used to compare the status so I went for that.

I'm convinced there are other n+1 queries that were introduced with the changes to the roles structure (just look at the log when loading a competition's page), but I didn't look into them, and they must be fairly less critical than this one.

I'd definitely consider reactivating bullet or something similar; I know it has been disabled by default because it was annoying, but it's an absolutely awesome tool to prevent situation like this (maybe restricting bullet's output to the terminal log would help?).

### DIFF
--- a/app/models/competition.rb
+++ b/app/models/competition.rb
@@ -16,7 +16,7 @@ class Competition < ApplicationRecord
   has_many :competitors, -> { distinct }, through: :results, source: :person
   has_many :competitor_users, -> { distinct }, through: :competitors, source: :user
   has_many :competition_delegates, dependent: :delete_all
-  has_many :delegates, through: :competition_delegates
+  has_many :delegates, -> { includes(:delegate_role_metadata) }, through: :competition_delegates
   has_many :competition_organizers, dependent: :delete_all
   has_many :organizers, through: :competition_organizers
   has_many :media, class_name: "CompetitionMedium", foreign_key: "competitionId", dependent: :delete_all
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -518,7 +518,15 @@ class User < ApplicationRecord
   end
 
   def trainee_delegate?
-    delegate_role_metadata.trainee_delegate.any?
+    # NOTE: `delegate_role_metadata.trainee_delegate.any?`, does fire a db query
+    # even if the `delegate_role_metadata` is eager loaded (because rails
+    # really wants to translate it to a "select 1 ..."; therefore we use a
+    # different implementation when we explicitly eager load roles.
+    if delegate_role_metadata.loaded?
+      delegate_role_metadata.any?(&:trainee_delegate?)
+    else
+      delegate_role_metadata.trainee_delegate.any?
+    end
   end
 
   private def staff_delegate?


app/models/competition.rb
```diff
@@ -16,7 +16,7 @@ class Competition < ApplicationRecord
   has_many :competitors, -> { distinct }, through: :results, source: :person
   has_many :competitor_users, -> { distinct }, through: :competitors, source: :user
   has_many :competition_delegates, dependent: :delete_all
-  has_many :delegates, through: :competition_delegates
+  has_many :delegates, -> { includes(:delegate_role_metadata) }, through: :competition_delegates
```
gregorbg, line 19: Awesome, I never thought about using the `includes` in this way, but it totally makes sense!

app/models/user.rb
```diff
@@ -518,7 +518,12 @@ def any_kind_of_delegate?
   end
 
   def trainee_delegate?
-    delegate_role_metadata.trainee_delegate.any?
+    # NOTE: it used to be `delegate_role_metadata.trainee_delegate.any?`, but
+    # it does fire a db query even if the `delegate_role_metadata` is eager
+    # loaded.
+    delegate_role_metadata.any? do |r|
+      r.status == RolesMetadataDelegateRegions.statuses[:trainee_delegate]
+    end
```
gregorbg, line 526: Because the Delegate status column is backed by a Rails `enum`, it gives you `r.trainee_delegate?` as a convenience method for free. So I think this can be turned into `delegate_role_metadata.any?(&:trainee_delegate?)` as a shorthand.

gregorbg: With that being said, your approach makes it much more inefficient for cases where the data is _not_ pre-loaded. Your above change uses `includes` by default, so it shouldn't be much of a problem, but I guess I am mostly astonished by Rails that it doesn't seem to "respect" pre-loading for the `exists?` method.

gregorbg (commented):
Thank you for your submission and super-thorough explanation!
We're actually in the middle of optimizing the roles queries, see PRs like https://github.com/thewca/worldcubeassociation.org/pull/9378, https://github.com/thewca/worldcubeassociation.org/pull/9355 and https://github.com/thewca/worldcubeassociation.org/pull/9377. These are being developed / submitted almost on a daily basis by Daniel now, so I guess we're on a good path :)

For this PR specifically, I am surprised that `any?` is not optimized for pre-loading. It feels like we're trading one disease here for another (switching from a "inefficient even when pre-loading" to "even more inefficient when _not_ pre-loading") and I cannot help but wonder whether Rails has any remedies for that?

viroulep:
I actually don't know why they can't adjust the implementation under the hood; `association.enum_value.any?` really means "is there any row with that enum value?", and a few guesses I could make are:
  - maybe using the eager loaded array could trigger some callback of some sort?
  - maybe rails has something for associations but not enums?

I honestly don't know; I'm not sure it's a huge issue in this specific case to unconditionally load all roles and doing `any?` over them (since generally for one user it's gonna be a dozens or so at most I guess?), but I'm sure it's absolutely worth it to make sure eager loading works.
I've taken into account the `loaded?` status of the association, but it's not clear to me if spreading such "implementation choices" is really good for small associations like these (read: if we start putting if/else like this everywhere we may or may not use eager loading, it's gonna be a mess).

If I'm not mistaken in both cases we send just one query ("select 1" vs "select *" + `any?`), but the big difference is that for the latter we need to have the objects in memory; so my wild guess would be that it's important to care about this if the number of associated rows can grow beyond expectations.

gregorbg:
> I honestly don't know; I'm not sure it's a huge issue in this specific case to unconditionally load all roles and doing `any?` over them (since generally for one user it's gonna be a dozens or so at most I guess?), but I'm sure it's absolutely worth it to make sure eager loading works.
> I've taken into account the `loaded?` status of the association, but it's not clear to me if spreading such "implementation choices" is really good for small associations like these (read: if we start putting if/else like this everywhere we may or may not use eager loading, it's gonna be a mess).

Fair point, but for the time being I think we should be pragmatic and merge this to allow loading WCIFs fast :smile:

gregorbg (approved):

PR #9385, opened 2024-05-15
Create db/seeds for board & officers

danieljames-dj:
For now, creating just board & officers in db/seeds. If this way of creating user_roles is fine, I'll do the same for other group types as well in upcoming PRs.

### DIFF
--- a/db/seeds/development/user_roles.seeds.rb
+++ b/db/seeds/development/user_roles.seeds.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+after :user_groups do
+  # Board Roles
+  past_board_roles = 9.times.collect { |index| FactoryBot.create!(:board_role, :inactive) }
+  current_board_roles = 4.times.collect { |index| FactoryBot.create!(:board_role, :active) }
+
+  # Officer Roles
+  # Giving officer roles to all board users except a few
+  FactoryBot.create!(:executive_director_role, :inactive, user: past_board_roles[1].user)
+  FactoryBot.create!(:executive_director_role, :inactive, user: past_board_roles[2].user)
+  FactoryBot.create!(:executive_director_role, user: current_board_roles[1].user)
+  FactoryBot.create!(:chair_role, :inactive, user: past_board_roles[2].user)
+  FactoryBot.create!(:chair_role, :inactive, user: past_board_roles[3].user)
+  FactoryBot.create!(:chair_role, user: current_board_roles[1].user)
+  FactoryBot.create!(:vice_chair_role, :inactive, user: past_board_roles[4].user)
+  FactoryBot.create!(:vice_chair_role, :inactive, user: past_board_roles[5].user)
+  FactoryBot.create!(:vice_chair_role, user: current_board_roles[2].user)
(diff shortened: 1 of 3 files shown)


db/seeds/development/user_roles.seeds.rb
gregorbg, line 1: You don't seem to be accessing index `[0]` for either past or current Board roles. Is this intended and if so, why?

spec/factories/user_roles.rb
```diff
@@ -2,15 +2,15 @@
 
 FactoryBot.define do
   factory :user_role do
-    user { FactoryBot.create(:user) }
+    user { rand(2) == 0 ? FactoryBot.create(:user_with_wca_id) : FactoryBot.create(:user) }
```
gregorbg, line 5: I don't like randomness in our test suite. Why do you flip a coin here?

db/seeds/development/user_roles.seeds.rb
danieljames-dj: This is to consider the case where there can be a Board member who doesn't have an officer role. The "all board users except a few" phrase of line 9 is the explanation of the same.

spec/factories/user_roles.rb
danieljames-dj: Thing is, I want to make sure this works irrespective of whether the user has a WCA ID or not. Do you have any alternative suggestion?

FinnIckler: two tests?

db/seeds/development/user_roles.seeds.rb
gregorbg: That comment does not make sense to me in relation to the array usage at all (which you can also judge by the fact that I started this review comment despite reading the code comment that you quoted).

So when you create a Board role, it also generates a new user attached to that role?

danieljames-dj: Okay, I've made the comment more clear. Yes when we create a board role, it also generates a new user attached to that role.

spec/factories/user_roles.rb
danieljames-dj: I was thinking to test both randomly in each and every test where user_role is created. Just like how we create random name and random date. Adding 2 tests for each tests can be painful, so I've decided not to test the user_with_wca_id case. I've removed that.

gregorbg: There is a distinct difference between randomness in test _data_, and randomness in test _cases_.
For example, when you create a dummy competition for tests, the latitude and longitude will be random. That's okay because you're only using this competitions as "some valid competition that I don't really care about", and once you have created it, the randomized lat/long stays the same. It's generated randomly once, saved into the mocked object, and then it stays like that. (If you run into a test case where you care about the exact lat/long, like testing distances in the Competition Series proximity, you will see that these tests very carefully compute distinct lat/long values instead of relying on randomness).
Your random call actually influences test _behavior_, i.e. the thing that you're testing is wildly different. That's not good and it's also the reason why I flagged it during review.

danieljames-dj: Oh okay, understood. Thanks a lot for the detailed explanation. :-)

gregorbg (approved):

PR #9386, opened 2024-05-15
Add GroupsManagerAdmin to WRT panel and few related cleanups

danieljames-dj:
1. GroupsManagerAdmin is required by WRT as well, hence added it to WRT panel as well.
2. Since more and more shared pages are coming, moved GroupsManagerAdmin to a new pages folder inside panel. Newer components will go inside pages folder, and eventually will move the things inside respective folders will also move to pages folder.
3. Realized that `statuses` cannot be used in normal methods, and that was throwing error in `roles_metadata_teams_committees.rb`. Fixed it.
4. Renamed `status_sort_rank` to `status_rank` to use it in `at_least_senior_member?` method.
5. Found a bug that `highest_delegate_role` is actually returning lowest delegate role, fixed it.


app/models/user.rb
```diff
@@ -1325,7 +1325,7 @@ def subordinate_delegates
   end
 
   private def highest_delegate_role
-    delegate_roles.max_by { |role| role.status_sort_rank }
+    delegate_roles.min_by { |role| role.status_rank }
```
gregorbg, line 1328: This change is confusing. The method is called "highest" but you are computing a minimum.
You should **at least** add a comment, or perhaps even reconsider the whole implementation of `status_rank`.

gregorbg (changes requested):

app/models/user.rb
danieljames-dj: Okay added a comment. Thanks.

gregorbg: Why is `status_rank` implemented like this, anyways? In WCA real-world logic (not code), Senior Delegate is considered a "higher" status than Junior Delegate, for example...

gregorbg (approved):

PR #9388, opened 2024-05-15
Don't run SET commands in staging when importing the dev dump

FinnIckler:
RDS doesn't let you run those commands, as these are managed by RDS Config Profiles


FinnIckler:
@thewca-bot deploy staging

FinnIckler:
@thewca-bot deploy staging

gregorbg (approved):

PR #9390, opened 2024-05-16
Fixes #8894: added length checks for all varchar columns

### LINKED ISSUE #8894: Add validation to length of fields (venue link, name reason string, check for any others)
**Describe the bug**
After filling the form and pressing "Create competition" nothing happens.
No errors from the front end at all
There are some errors when doing the POST, as you will see in the screenshots section. 
[Here ](https://pastebin.com/z46ff9ZG)you can see a dump from the "options" JSON, this could help.

**To Reproduce**
Create a competition, nothing too specific,

**Expected behavior**
The competition should be created

**Screenshots**
image
WST3

Desktop (please complete the following information):

OS: Windows
Browser: Brave
Version: 1.56.20 (Chromium 115.0.5790.171)

dunkOnIT:
Fixes #8894

There are a number of varchar database columns which, if their specified length is exceeded, will throw an error which our frontend doesn't know how to handle.

Instead of addressing these invdividually, I am iterating over all string columns and validating their length. Important notes:
* according to ChatGPT, columns_hash is generated by ActiveRecord on startup - we aren't firing calls at the db to ask about the table schema to perform this validation
* if we are happy with this approch, we could also remove `length` validations from some of the fields which do have it enforced (eg, `name`)

**Note:** Also includes some spec changes, including
- removing previously unremoved `focus` tags (these should have been removed in earlier PR's, my bad)
- refactor of request/competition_spec specs to make it possible to add `create` tests in future (I had one, but couldn't reliably trigger it _only_ when the frontend would bug out, so I removed it - leaving the refactor tho)


dunkOnIT:
@gregorbg after a bit of digging, it looks like our default character limit is 191 because we use utf8mb4 encoding, as defined in `config/database.yml`. ChatGPT's explanation is below, TL;DR is that if a string exceeds 191 the column can't be indexed by MySQL. 

Do we care about that in this case? 

Also - upon thinking about string length limits, I'm not a fan of making it super long (mediumtext) because it can lead to issues with display / WCAT needing to moderate how long these fields get.

Rather, I'd suggest we: 
* change the string fields to be varchar(255). This should be plenty for their use cases (venue address, venue details, etc) - double length of fields which have infrequently caused problems in the past
* keep the validations so that no frontend-breaking error is thrown

> Why utf8mb4 Imposes a 191 Character Limit
> 
>     UTF-8 vs. utf8mb4:
>         utf8mb4 is a true 4-byte UTF-8 encoding that supports all Unicode characters, including emojis and various symbols.
>         utf8 (in MySQL) only supports up to 3 bytes per character, which is a limited subset of UTF-8.
> 
>     Index Length Limitation:
>         In MySQL, the maximum size for an index key is 767 bytes for InnoDB tables.
>         With utf8mb4, each character can take up to 4 bytes.
>         To stay within the 767-byte limit for indexed columns, the maximum length for a VARCHAR field is effectively 191 characters (191 * 4 = 764 bytes).
> 
> Rails Handling of utf8mb4
> 
> Rails adjusts the default length of string fields to 191 characters when utf8mb4 is used to ensure compatibility with MySQL's index limitations. This behavior is typically managed by Rails' ActiveRecord and might be explicitly set in initializers.

gregorbg:
Interesting! We should not meddle with the internal defaults of MySQL too much. If Rails is actually smart enough to choose the length based on encoding (I'm surprised it does, but it's a positive surprise to be sure!) then who are we to question these smart assumptions?

To proceed, I think you should migrate the "critical" columns in `Competition` (ie. the ones that have an unreasonably *short* limit) to just be "type string" and let Rails figure out the best limit in that case. Don't artificially set them to 255 or anything, and don't bother about lowering the currently existing 255 columns to the 191 default either.

In other news, why is Rails not validating the string length during save? ChatGPT could know about this as well.

dunkOnIT:
> In other news, why is Rails not validating the string length during save? ChatGPT could know about this as well.

ChatGPT says this isn't part of the default Rails behaviour (a quick Google of [SO questions](https://stackoverflow.com/questions/10988963/rails-activerecord-validation-maximum-length-of-a-string) confirms this, giving a few reasons of which these seem the most compelling:
- auto-validating string length is overly prescriptive (maybe the dev WANTS a db error, or wants to validate for a different length, or whatever)
- separations of concerns between model validations and db constraints (ie, allow the db to enforce its own constraints)

In most applications, I guess it doesn't matter as much _where_ the error is generated, and whether it is a **validation error** specifically which is generated. Our React frontend is more brittle to this distinction than most Rails code (whether backend or frontend) imo

dunkOnIT:
Shipped a migration which updated venueDetails and venueAddress to default strings. Pictured below are all strings with their varchar limits in the comp table - all of the others look deliberately low, so I'm leaving them as is, let me know if you want any others changed.

Also leaving in the validation, as it is still reasonable for a user to exceed some of these limits in "good faith", and we want to give them a clear error when they do imo.

This raises the question, though: Should I remove the other string length validations, given that they are now redundant? (eg, on `name`)
image

db/schema.rb
```diff
@@ -1117,6 +1117,7 @@
     t.datetime "created_at", precision: nil, null: false
     t.datetime "updated_at", precision: nil, null: false
     t.bigint "parent_record_id"
+    t.string "test_string"
```
gregorbg, line 1120: ehehe

app/models/competition.rb
```diff
@@ -221,6 +221,13 @@ class Competition < ApplicationRecord
   validates :event_restrictions_reason, presence: true, if: :event_restrictions?
   validates_inclusion_of :main_event_id, in: ->(comp) { [nil].concat(comp.persisted_events_id) }
 
+  # Ensure that no string fields exceed the varchar limit in the database
+  columns_hash.each do |column_name, column_info|
+    if column_info.type == :string && column_info.limit
```
gregorbg, line 226: Probably wanna limit this to the "interesting" columns, i.e. check that `column_name` is included in some pre-defined list.

gregorbg (changes requested):

app/models/competition.rb
dunkOnIT: I thought about this as well, but my concern is that future devs may come along and add "interesting" columns, and we have no way of drawing their attention to whether the column should be included in this check (eg, our failing test if you don't add a column to the db sanitizer). 

I agree it's overkill to check _all_ columns, but I don't think there are any explicitly _bad_ things about doing it this way?

db/schema.rb
dunkOnIT: de-oopsied

app/models/competition.rb
gregorbg: If a future dev comes along and adds interesting columns, it should also be their responsibility to think about how those interesting columns are being validated.

Actually, I would use your point to form a counter-argument: When introducing a new column, there shouldn't be a (reasonable but) arbitrary validation enforced upon that column out of nowhere

dunkOnIT: Please see latest push - realized we can just add a test for this lol. Thus, the discussion is transmuted into: Which strings should be validated?

gregorbg (approved):

PR #9392, opened 2024-05-17
Ensure trailing slash for all /regulations pages

dunkOnIT:
This may address the issue [raised here](https://github.com/thewca/wca-regulations-compiler/pull/29), where regulation links don't generate properly

### DIFF
--- a/app/controllers/regulations_controller.rb
+++ b/app/controllers/regulations_controller.rb
@@ -21,7 +21,7 @@ class RegulationsController < ApplicationController
     end
   end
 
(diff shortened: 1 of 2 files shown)


dunkOnIT:
@thewca-bot deploy staging

FinnIckler:
I don't know if this is the correct solution, there is a a method in the `regulations_translations_controller` called `ensure_trailing_slash` that is made to handle these cases already

FinnIckler (approved):
LGTM

PR #9394, opened 2024-05-19
Migrate banned competitors to groups

danieljames-dj:
This PR is migration of banned competitors to groups.

This PR is ready for review for existing changes, but there are couple of more changes which might depend on other upcoming PRs:
- Update banned competitors is not working now
- Currently banned competitors list is shown only in leader panel, it has to be shown in some other relevant panels as well
I'm working on the above couple of issues and will submit them soon.
- Add to role change mailers

### DIFF
--- a/app/controllers/api/v0/user_roles_controller.rb
+++ b/app/controllers/api/v0/user_roles_controller.rb
@@ -143,6 +143,10 @@ class Api::V0::UserRolesController < Api::V0::ApiController
     end
   end
 
+  private def changed_value_to_human_readable(changed_value)
+    changed_value.nil? ? 'None' : changed_value
+  end
+
   # update method is written in a way that at a time, only one parameter can be changed. If multiple
   # values needs to be changed, then they need to be sent as separate APIs from the client.
   def update
@@ -264,8 +268,8 @@ class Api::V0::UserRolesController < Api::V0::ApiController
         if changed_parameter.present?
           changes << UserRole::UserRoleChange.new(
             changed_parameter: changed_parameter,
-            previous_value: values[0] || 'None',
-            new_value: values[1] || 'None',
+            previous_value: changed_value_to_human_readable(values[0]),
+            new_value: changed_value_to_human_readable(values[1]),
           )
         end
       end
--- a/app/controllers/api/v0/user_roles_controller.rb
+++ b/app/controllers/api/v0/user_roles_controller.rb
@@ -134,6 +134,15 @@ class Api::V0::UserRolesController < Api::V0::ApiController
     }
   end
 
+  private def changed_key_to_human_readable(changed_key)
+    case changed_key
+    when 'end_date'
+      'End Date'
+    else
+      nil
+    end
+  end
+
   # update method is written in a way that at a time, only one parameter can be changed. If multiple
   # values needs to be changed, then they need to be sent as separate APIs from the client.
   def update
@@ -251,14 +260,7 @@ class Api::V0::UserRolesController < Api::V0::ApiController
 
       role.save!
       role.previous_changes.each do |changed_key, values|
-        changed_parameter = (
-          case changed_key
-          when 'end_date'
-            'End Date'
-          else
-            nil
-          end
-        )
+        changed_parameter = changed_key_to_human_readable(changed_key)
         if changed_parameter.present?
           changes << UserRole::UserRoleChange.new(
             changed_parameter: changed_parameter,
--- a/app/controllers/api/v0/user_roles_controller.rb
+++ b/app/controllers/api/v0/user_roles_controller.rb
@@ -64,6 +64,7 @@ class Api::V0::UserRolesController < Api::V0::ApiController
   def create
     user_id = params.require(:userId)
     group_id = params[:groupId] || UserGroup.find_by(group_type: params.require(:groupType)).id
+    end_date = params[:endDate]
 
     create_supported_groups = [
       UserGroup.group_types[:delegate_regions],
@@ -73,6 +74,7 @@ class Api::V0::UserRolesController < Api::V0::ApiController
       UserGroup.group_types[:teams_committees],
       UserGroup.group_types[:councils],
       UserGroup.group_types[:board],
+      UserGroup.group_types[:banned_competitors],
     ]
     group = UserGroup.find(group_id)
 
@@ -110,6 +112,8 @@ class Api::V0::UserRolesController < Api::V0::ApiController
         metadata = RolesMetadataTeamsCommittees.create!(status: status)
       elsif group.group_type == UserGroup.group_types[:councils]
         metadata = RolesMetadataCouncils.create!(status: status)
+      elsif group.group_type == UserGroup.group_types[:banned_competitors]
+        metadata = RolesMetadataBannedCompetitors.create!
       else
         metadata = nil
       end
@@ -118,6 +122,7 @@ class Api::V0::UserRolesController < Api::V0::ApiController
         user_id: user_id,
         group_id: group_id,
         start_date: Date.today,
+        end_date: end_date,
         metadata: metadata,
       )
     end
@@ -238,6 +243,29 @@ class Api::V0::UserRolesController < Api::V0::ApiController
         end
       else
         return render status: :unprocessable_entity, json: { error: "Invalid parameter to be changed" }
+      end
+    elsif group_type == UserGroup.group_types[:banned_competitors]
+      if params.key?(:endDate)
+        role.end_date = params.require(:endDate)
+      end
+
+      role.save!
+      role.previous_changes.each do |changed_key, values|
+        changed_parameter = (
+          case changed_key
+          when 'end_date'
+            'End Date'
+          else
+            nil
+          end
+        )
+        if changed_parameter.present?
+          changes << UserRole::UserRoleChange.new(
+            changed_parameter: changed_parameter,
+            previous_value: values[0] || 'None',
+            new_value: values[1] || 'None',
+          )
+        end
       end
     else
       return render status: :unprocessable_entity, json: { error: "Invalid group type" }
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -577,6 +577,17 @@ class User < ApplicationRecord
     delegate_roles.select { |role| role.metadata.status == RolesMetadataDelegateRegions.statuses[:senior_delegate] }
   end
 
+  private def groups_with_read_access
+    return "*" if can_edit_any_groups?
+    groups = groups_with_edit_access
+
+    if can_view_banned_competitors?
+      groups += UserGroup.banned_competitors.ids
+    end
+
+    groups
+  end
+
   private def groups_with_edit_access
     return "*" if can_edit_any_groups?
     groups = []
@@ -622,6 +633,9 @@ class User < ApplicationRecord
       },
       can_create_groups: {
         scope: groups_with_create_access,
+      },
+      can_read_groups: {
+        scope: groups_with_read_access,
       },
       can_edit_groups: {
         scope: groups_with_edit_access,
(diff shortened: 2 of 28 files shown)


app/webpacker/components/Panel/pages/BannedCompetitorsPage/BannedCompetitorForm.jsx
```diff
@@ -0,0 +1,84 @@
...
+  const [formValues, setFormValues] = useState({
+    user: null,
+    endDate: banActionRole?.end_date,
+  });
+  const { save, saving } = useSaveAction();
+
+  const handleFormChange = (_, { name, value }) => setFormValues({ ...formValues, [name]: value });
+
+  const createNewBannedCompetitor = () => {
+    // Do action to create new banned competitor
```
kr-matthews, line 23: I don't think this comment is helpful, it just repeats the function name.
```suggestion
```

app/webpacker/components/Panel/pages/BannedCompetitorsPage/BannedCompetitorForm.jsx
```diff
@@ -0,0 +1,84 @@
...
+      groupType: groupTypes.banned_competitors,
+      endDate: formValues?.endDate,
+    }, () => {
+      sync();
+      closeForm();
+    }, { method: 'POST' });
+  };
+
+  const editBannedCompetitor = () => {
+    // Do action to edit new banned competitor
```
kr-matthews, line 35: Likewise. (Plus the word "new" shouldn't be there?)
```suggestion
```

app/webpacker/components/Panel/pages/BannedCompetitorsPage/BannedCompetitorForm.jsx
```diff
@@ -0,0 +1,84 @@
...
+    }, { method: 'PATCH' });
+  };
+
+  const formSubmitHandler = () => {
+    if (banAction === 'new') {
+      createNewBannedCompetitor();
+    } else {
+      editBannedCompetitor();
+    }
+  };
```
kr-matthews, line 50: Does this work?
```suggestion
  const formSubmitHandler = banAction === 'new' ? createNewBannedCompetitor : editBannedCompetitor;
```

app/webpacker/components/Panel/pages/BannedCompetitorsPage/BannedCompetitorForm.jsx
```diff
@@ -0,0 +1,84 @@
...
+        showYearDropdown
+        dateFormatOverride="YYYY-MM-dd"
+        dropdownMode="select"
+        isoDate={formValues?.endDate}
+        onChange={(date) => handleFormChange(null, {
+          name: 'endDate',
+          value: date,
+        })}
+      />
+      <Form.Button onClick={() => closeForm()}>Cancel</Form.Button>
```
kr-matthews, line 80:
```suggestion
      <Form.Button onClick={closeForm}>Cancel</Form.Button>
```

app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -39,7 +39,7 @@ class Api::V0::UserRolesController < Api::V0::ApiController
   # Filter role based on the permissions of the current user.
   private def can_current_user_access(role)
     group = role.group
-    !group.is_hidden || current_user&.has_permission?(:can_edit_groups, group.id)
+    !group.is_hidden || current_user&.has_permission?(:can_read_groups, group.id)
```
gregorbg, line 42: Without any context, this seems like a pretty crazy change: What does "access" (from the method name `can_user_access`) mean in this context?
And why is it suddenly enough to have read-only permissions when previously you needed write permissions?

app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -330,7 +335,8 @@ def update
       else
         return render status: :unprocessable_entity, json: { error: "Invalid parameter to be changed" }
       end
-    else
+    # elsif group_type == UserGroup.group_types[:banned_competitors]
+    # else
```
gregorbg, line 339: Why is there commented-out (dead) code here?

app/models/user.rb
```diff
@@ -570,6 +570,17 @@ def banned_at_date?(date)
     delegate_roles.select { |role| role.metadata.status == RolesMetadataDelegateRegions.statuses[:senior_delegate] }
   end
 
+  private def groups_with_read_access
+    return "*" if can_edit_any_groups?
+    groups = groups_with_edit_access
+
+    if can_view_banned_competitors?
+      groups += UserGroup.banned_competitors.ids
+    end
```
gregorbg, line 579: Why is this case not covered by `groups_with_edit_access` directly above?

db/schema.rb
```diff
@@ -1141,6 +1148,7 @@
     t.datetime "created_at", precision: nil, null: false
     t.datetime "updated_at", precision: nil, null: false
     t.boolean "team_senior_member", default: false, null: false
+    t.bigint "group_id"
```
gregorbg, line 1151: I'm not sure but this feels unrelated?

db/schema.rb
```diff
@@ -1236,6 +1244,8 @@
     t.boolean "cookies_acknowledged", default: false, null: false
     t.boolean "registration_notifications_enabled", default: false
     t.string "otp_secret"
+    t.integer "senior_delegate_id"
+    t.string "delegate_status", limit: 64
```
gregorbg, line 1248: Definitely unrelated

gregorbg (changes requested):

app/controllers/api/v0/user_roles_controller.rb
danieljames-dj: I've changed this a bit in https://github.com/thewca/worldcubeassociation.org/pull/9378 and now this method is living in user_role.rb with the name `can_user_read?`, this should make it more clear.

This method is only used for read API. The reason why `can_edit_groups` is changed to `can_read_groups` is that, earlier both were having same value - ie, a user can edit a group if they can read the hidden group. But for banned competitors it's different - eg: WEC have access to read the hidden group, but not write. So created new permission.

app/controllers/api/v0/user_roles_controller.rb
danieljames-dj: I'll add them once https://github.com/thewca/worldcubeassociation.org/pull/9395 is merged since this PR is dependent on that 😬

app/models/user.rb
danieljames-dj: Because users like WEC members can view banned competitors, but they cannot edit banned competitors.

db/schema.rb
danieljames-dj: Oops removed

db/schema.rb
danieljames-dj: Oops removed

app/controllers/api/v0/user_roles_controller.rb
gregorbg: Now that's merged, yay!

app/controllers/api/v0/user_roles_controller.rb
gregorbg: Okay, makes sense. Thanks for the explanation!

app/models/user.rb
gregorbg: Okay, this makes sense but still looks a bit strange because one might be under the impression that you're giving Banned a kind of "special" treatment. I know this is due to the fact that they are currently the only group that could potentially have read-only access, but at least in the back of your mind think about something more generic for computing permissions, please.

app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -239,6 +244,29 @@ def update
       else
         return render status: :unprocessable_entity, json: { error: "Invalid parameter to be changed" }
       end
+    elsif group_type == UserGroup.group_types[:banned_competitors]
+      if params.key?(:endDate)
+        role.end_date = params.require(:endDate)
+      end
+
+      role.save!
```
danieljames-dj, line 252: role.update can be used, but there are more parameters to be added in the next PR, so updating each inside if block and then role.save at end is ideal here.

app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -239,6 +244,29 @@ def update
...
+            'End Date'
+          else
+            nil
+          end
+        )
+        if changed_parameter.present?
+          changes << UserRole::UserRoleChange.new(
+            changed_parameter: changed_parameter,
+            previous_value: values[0] || 'None',
+            new_value: values[1] || 'None',
```
danieljames-dj, line 266: @gregorbg Could you please review these two lines in more deep? I'm looking to print 'None' whenever the value is nil. This is working fine, but I'm confused whether it will miss any edge cases. Like, I expected `0 || 'None'` to give 'None', but it gives 0. Luckily I wanted 0. (In python it would have taken 'None') Basically I want to take 'None' only if value is `nil`.

gregorbg: Yup, interpreted languages all have their own rules about what is false-y and what isn't.
Ruby is relatively strict in comparison, but also has a lot of convenience methods for you: `my_value.nil?` should do exactly what you want.

gregorbg: You can de-structure the array (similar to JS `const [foo, bar] = my_awesome_array` but slightly different syntax) using `previous_value, new_value = values`. Then you can have a helper function `changed_value_to_human_readable`:

```ruby
def changed_value_to_human_readable(value)
  value.nil? ? 'None' : value
end
```

gregorbg: Using the above method and destructuring, you can then do 
```suggestion
            previous_value: changed_value_to_human_readable(previous_value),
            new_value: changed_value_to_human_readable(new_value),
```
and I think that's reasonably readable

gregorbg (commented):
Code is okay, I've given a suggestion about the snippet that you asked for.

Why are there no tests changing in this PR at all? Banned Competitors is a quite sensitive area of our code, I'd like this backend change to be backed and supported by our existing tests.

app/controllers/api/v0/user_roles_controller.rb
danieljames-dj: Thanks, I've implemented your suggestion.

danieljames-dj (commented):
Thanks Gregor, regarding unit tests - I was planning to have the unit tests for all methods in user_controller as another PR. But since you explicitly mentioned, I've added it for the read APIs. And while writing that, I realized a mistake during rebase - I didn't change `can_write_group` to `can_read_group` because of which banned competitors didn't become accessible to delegates.

spec/controllers/api/v0/user_roles_controller_spec.rb
```diff
@@ -28,6 +29,28 @@
...
+    context 'when user is logged in as a normal user' do
+      sign_in { FactoryBot.create(:user) }
+
+      it 'fetches list of roles of a user' do
+        get :index_for_user, params: { user_id: user_whose_delegate_status_changes.id }
+
+        expect(response.body).to eq(user_whose_delegate_status_changes.active_roles.to_json)
+      end
+
+      it 'fetches list of banned competitos' do
```
gregorbg, line 49: The point of this test is that it does NOT fetch the list of banned competitors :P

gregorbg (approved):
Small nitpick about the naming of the tests, but the rest of the code is okay so I'll approve already and you can merge once you fixed the nitpick

spec/controllers/api/v0/user_roles_controller_spec.rb
danieljames-dj: Thanks, changed. 👍

PR #9395, opened 2024-05-19
Allow change multiple parameters in single user-role update request

danieljames-dj:
Currently using user_roles_controller#update, we can update only one parameter. This PR is to allow editing more than one parameters. This will be helpful in banned competitors because more parameters will be updated there.

### DIFF
--- a/app/controllers/api/v0/user_roles_controller.rb
+++ b/app/controllers/api/v0/user_roles_controller.rb
@@ -143,11 +143,11 @@ class Api::V0::UserRolesController < Api::V0::ApiController
     if group_type == UserGroup.group_types[:delegate_regions]
       if params.key?(:status)
         status = params.require(:status)
-        changes << {
+        changes << UserRole::UserRoleChange.new(
           changed_parameter: 'Status',
           previous_value: I18n.t("enums.user_roles.status.delegate_regions.#{role.metadata.status}", locale: 'en'),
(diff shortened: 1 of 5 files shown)


app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -333,7 +344,7 @@ def update
     else
       return render status: :unprocessable_entity, json: { error: "Invalid group type" }
     end
-    RoleChangeMailer.notify_role_change(role, current_user, changed_parameter, previous_value, new_value).deliver_later
+    RoleChangeMailer.notify_role_change(role, current_user, changes).deliver_now # Change to deliver_later
```
gregorbg, line 347: There's a comment at the end of the line which feels relevant

gregorbg (commented):
For the list of changes, you might want to consider using something like `Struct.new` which can declare utility classes on-the-fly. See `PsychSheet` in `competition.rb` for reference -- it makes future refactors much easier.

app/controllers/api/v0/user_roles_controller.rb
danieljames-dj: Oops sorry I missed to remove it after testing 😅 Removed now.

danieljames-dj (commented):
Thanks for Struct.new idea. I didn't knew it. Used it now.

gregorbg (approved):

PR #9396, opened 2024-05-19
Remove activities from schedule when events/rounds removed

kr-matthews:
This won't help the existing competitions which have 'invalid' activities. I looked into that a bit but I'm not sure how/if that's causing the schedule view to crash, and I won't have time to look into in the near future. And if that's indeed causing the crash then we should fix the UI because other tools might remove events without removing them from the schedule (or alternatively, add more validation to the WCIF when saving it? I don't know much about that though).

### DIFF
--- a/app/views/competitions/edit_events.html.erb
+++ b/app/views/competitions/edit_events.html.erb
@@ -11,7 +11,8 @@
     canAddAndRemoveEvents: current_user.can_add_and_remove_events?(@competition),
     canUpdateEvents: current_user.can_update_events?(@competition),
     canUpdateQualifications: current_user.can_update_qualifications?(@competition),
-    wcifEvents: @competition.events_wcif
+    wcifEvents: @competition.events_wcif,
+    wcifSchedule: @competition.schedule_wcif,
   }, {
     id: "events-edit-area",
   }) %>
--- a/app/webpacker/components/EditEvents/EventPanel/index.js
+++ b/app/webpacker/components/EditEvents/EventPanel/index.js
@@ -38,7 +38,7 @@ export default function EventPanel({
         content: `Are you sure you want to remove all ${pluralize(
           wcifEvent.rounds.length,
           'round',
-        )} of ${event.name}?`,
+        )} of ${event.name}? This will also remove these rounds from the schedule.`,
       })
         .then(() => {
           dispatch(removeEvent(wcifEvent.id));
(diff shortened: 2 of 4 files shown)


gregorbg:
Thanks for the quick fix! I can see that you are passing the schedule part of the WCIF only to remove events if necessary, and then otherwise you never touch it. This is exactly the kind of consistency validation that should be handled by the backend.

But in this concrete case, it raises another interesting issue: Should we even delete these activities? We'regoing through the standard WCIF PATCH endpoint, which is cool but it means that whatever validations we introduce might affect external WCIF PATCHes as well.
This might lead to unforeseen consequences if people PATCH'ing in partial WCIFs don't pay attention, and REST APIs don't have mechanisms to send a "This will have the following side effects, are you sure?" prompt back to the client. So I don't feel happy introducing a straight-out "delete activities when the event is deleted" consistency validation. On the other hand, doing just that feels like "the right thing to do" from a very strict data consistency standpoint.

Opinions?

gregorbg:
If you need a quick solution, then probably we should have the "Show Schedule" component skip activities that it cannot find the matching event for

kr-matthews:
> This is exactly the kind of consistency validation that should be handled by the backend.

I can't speak much about the back-end side of things. But unless a decision is made and the work is done soon, it seems reasonable to have the front-end do this at least temporarily, as I believe most organizers would be doing deletions via the WCA site, and (as I mention below) I think the activities do need to be removed at some point in the process (not _just_ skipped over in the schedule UI).

> But in this concrete case, it raises another interesting issue: Should we even delete these activities?

Ultimately the schedule should be accurate. If we're not going to delete them immediately then we'd need to somehow delete them as part of the results submission process, right?

> If you need a quick solution, then probably we should have the "Show Schedule" component skip activities that it cannot find the matching event for

As I mentioned, I looked into that briefly but couldn't figure out the issue. I can take another look (I agree that that's a good fix to have regardless of whatever else we do), but might not have time to do so for another week or 2. But if we do hide these activities, then it's hard to know (just from the WCA site UI) that they still exist and need deleting.

gregorbg:
This is now actually done "for free" with the linking of activities to rounds in the backend. The change has been made in #11643 and in particular in the `has_many :schedule_activities, dependent: :destroy` association on `round.rb`. It means that whenever a round is deleted (which happens when you sync the Edit Events UI), it automatically pulls all associated events to the grave.

cf. https://github.com/thewca/worldcubeassociation.org/pull/11643/files#diff-70a53df91de6ce17997a4f6d188b36ae01091c5997e1eac68e99e06cc81dd255

PR #9400, opened 2024-05-21
Re-enable Bullet in a slightly less annoying version

gregorbg:
Inspired by https://github.com/thewca/worldcubeassociation.org/pull/9382 I have reconfigured [Bullet](https://github.com/flyerhzm/bullet) to be active by default again. But this time, I also de-activated the `window.alert` style logging, so it "only" logs to the Rails and JS consoles and you can easily access the details there if you want to optimize things.


FinnIckler (approved):

viroulep:
Thanks a lot @gregorbg!

FTR I did not quite realize that when I made the suggestion, but it creates a certain performance hit when generating large WCIF json (at a point where the json is very slow to render even for small comps).
Definitely useful to have this on when developing though, just something to have in mind when working on "big" competitions and that temporarily disabling bullet will help perf if needed :)

PR #9401, opened 2024-05-21
Consistently name activites in English in the EditSchedule component

gregorbg:
This PR splits what is stored internally as "Activity name" (see WCIF spec) and what is displayed in FullCalendar.

Previously, we immediately used the activity name (as per WCIF) as the FC display title. Now, we _compute_ an FC display title based on the activity, which may or may not have something to do with the name (and can be changed relatively easily in the future).

In doing so, I have also tried to unify and de-duplicate places in the code where we compute names.

### DIFF
--- a/app/webpacker/components/EditSchedule/EditActivities/index.js
+++ b/app/webpacker/components/EditSchedule/EditActivities/index.js
@@ -46,7 +46,7 @@ import { friendlyTimezoneName } from '../../../lib/wca-data.js.erb';
 import {
   activityToFcTitle,
   buildPartialActivityFromCode,
-  defaultDurationFromActivityCode,
+  defaultDurationFromActivityCode, FC_ACTIVITY_ATTACHMENT,
   fcEventToActivityAndDates,
   luxonToWcifIso,
 } from '../../../lib/utils/edit-schedule';
@@ -114,7 +114,7 @@ function EditActivities({
         start: activity.startTime,
         end: activity.endTime,
         extendedProps: {
-          activity,
+          [FC_ACTIVITY_ATTACHMENT]: activity,
           matchCount,
         },
       };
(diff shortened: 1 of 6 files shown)


app/webpacker/components/EditSchedule/EditActivities/index.js
```diff
@@ -101,15 +107,14 @@ function EditActivities({
...
-        title: activity.name + (shouldUpdateMatches && matchCount > 0 ? matchesText : ''),
+        title: fcTitle,
         start: activity.startTime,
         end: activity.endTime,
         extendedProps: {
-          activityId: activity.id,
-          activityCode: activity.activityCode,
-          activityName: activity.name,
-          childActivities: activity.childActivities,
+          activity,
```
kr-matthews, line 117: If `activityPickerRef` is only providing a partial activity, would the key make more sense as `partialActivity`?
```suggestion
          partialActivity: activity,
```

gregorbg: I don't think so, because then our helper coe who is able to turn _any_ fcEvent into an activity won't be able to know which key it has to access (might be `activity` might be `partialActivity`)

kr-matthews: Hm. My concern stems from `activity: partialActivity` on line 142 below, which feels wrong. I was thinking every `activity` key would have to be switched to `partialActivity`, since it seems like sometimes the activity is partial and sometimes is full. (I would say a full activity is a partial activity - but not vice-versa of course.)

gregorbg: Attended to the review comments by introducing a shared key across the whole FC integration :smile:

kr-matthews (approved):

PR #9405, opened 2024-05-22
Add ORDER BY clause to developer dump

gregorbg:
This mitigates problems with a self-referential foreign key in `user_groups`. The error occurs mostly when existing groups are re-ordered, like recently there was a new big parent group for Asia and older groups were suddenly referencing that newer parent group.

The INSERT INTO ... SELECT ... FROM` query then tries to insert the rows in primary key order, where some earlier rows point to a primary key that is not inserted yet.

I have considered turning off foreign key checks as an alternative, but that will make debugging exponentially harder when the only error message at the very end of the whole dump is "some foreign key somewhere failed so we couldn't turn foreign key checks back on"

### DIFF
--- a/lib/database_dumper.rb
+++ b/lib/database_dumper.rb
@@ -647,6 +647,8 @@ module DatabaseDumper
       ),
     }.freeze,
     "user_groups" => {
+      # groups have a self-referencing foreign key to their parent group, so we need to make sure that root groups are inserted first
+      order_by_clause: "ORDER BY parent_group_id ASC",
       column_sanitizers: actions_to_column_sanitizers(
         copy: %w(
           id
--- a/lib/database_dumper.rb
+++ b/lib/database_dumper.rb
@@ -1204,8 +1204,11 @@ module DatabaseDumper
         # Some column names like "rank" are reserved keywords starting mysql 8.0 and require quoting.
         quoted_column_list = column_sanitizers.keys.map { |column_name| ActiveRecord::Base.connection.quote_column_name column_name }.join(", ")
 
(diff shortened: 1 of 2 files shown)


lib/database_dumper.rb
```diff
@@ -1204,8 +1159,11 @@ def self.with_dumped_db(dump_config_name, dump_sanitizers, dump_ts_name = nil)
         # Some column names like "rank" are reserved keywords starting mysql 8.0 and require quoting.
         quoted_column_list = column_sanitizers.keys.map { |column_name| ActiveRecord::Base.connection.quote_column_name column_name }.join(", ")
 
-        populate_table_sql = "INSERT INTO #{dump_db_name}.#{table_name} (#{quoted_column_list}) SELECT #{column_expressions} FROM #{source_table} #{table_sanitizer[:where_clause]}"
-        ActiveRecord::Base.connection.execute(populate_table_sql)
+        where_clause_sql = table_sanitizer.fetch(:where_clause, "")
+        order_by_clause_sql = table_sanitizer.fetch(:order_by_clause, "")
+
+        populate_table_sql = "INSERT INTO #{dump_db_name}.#{table_name} (#{quoted_column_list}) SELECT #{column_expressions} FROM #{source_table} #{where_clause_sql} #{order_by_clause_sql}"
```
gregorbg, line 1165: The basic structure of this line hasn't changed. I simply added the `order_by_clause_sql` at the very end, and together with the two new variable declarations above this was enough to throw off Git and block it from matching this line with the original line on the left side of the diff

danieljames-dj (approved):

PR #9411, opened 2024-05-24
Some more cleanups related to roles

danieljames-dj:
1. Moved `roles_of_group_type` to user_group as a static method, so that it can be used wherever the roles of a group_type is called.
2. Used the above method in api_controller.
3. Added some query optimizations in user.rb

### DIFF
--- a/app/controllers/api/v0/api_controller.rb
+++ b/app/controllers/api/v0/api_controller.rb
@@ -156,11 +156,7 @@ class Api::V0::ApiController < ApplicationController
   end
 
   def delegates
-    delegate_users = UserGroup.roles_of_group_type(
-      UserGroup.group_types[:delegate_regions],
-      includes_params: [:group, { user: [:delegate_roles, :delegate_role_metadata, :active_roles] }],
-    ).flat_map(&:user)
-    paginate json: delegate_users
+    paginate json: UserGroup.delegate_regions.flat_map(&:active_users)
   end
 
   def delegates_search_index
--- a/app/controllers/api/v0/api_controller.rb
+++ b/app/controllers/api/v0/api_controller.rb
@@ -156,7 +156,11 @@ class Api::V0::ApiController < ApplicationController
   end
 
   def delegates
-    paginate json: UserGroup.delegate_regions.flat_map(&:active_users)
+    delegate_users = UserGroup.roles_of_group_type(
+      UserGroup.group_types[:delegate_regions],
+      includes_params: [:group, { user: [:delegate_roles, :delegate_role_metadata, :active_roles] }],
+    ).flat_map(&:user)
+    paginate json: delegate_users
   end
 
   def delegates_search_index
--- a/app/controllers/api/v0/user_roles_controller.rb
+++ b/app/controllers/api/v0/user_roles_controller.rb
@@ -37,20 +37,10 @@ class Api::V0::UserRolesController < Api::V0::ApiController
     render json: roles
   end
 
-  private def roles_of_group_type(group_type)
-    group_ids = UserGroup.where(group_type: group_type).pluck(:id)
-    UserRole.where(group_id: group_ids)
-  end
-
   # Returns a list of roles primarily based on groupType.
   def index_for_group_type
     group_type = params.require(:group_type)
-    roles = roles_of_group_type(group_type)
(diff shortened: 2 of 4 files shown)


app/models/user_group.rb
```diff
@@ -180,6 +180,16 @@ def is_root_group?
     parent_group_id.nil?
   end
 
+  def self.roles_of_group_type(group_type, current_user: nil, params: {}, includes_params: [])
+    roles = UserRole.includes(includes_params + [:group]).where(group: { group_type: group_type })
```
gregorbg, line 184: Nitpick: I think `UserRole.includes(:group).includes(includes_params)` is a bit more straight forward than manually glueing arrays together (especially since you cannot be sure that `includes_params` itself will always be an array. Passing hashes as include_params is totally valid in Rails.)

app/controllers/api/v0/api_controller.rb
```diff
@@ -156,7 +156,11 @@ def omni_search
   end
 
   def delegates
-    paginate json: UserGroup.delegate_regions.flat_map(&:active_users)
+    delegate_users = UserGroup.roles_of_group_type(
+      UserGroup.group_types[:delegate_regions],
+      includes_params: [:group, { user: [:delegate_roles, :delegate_role_metadata, :active_roles] }],
+    ).flat_map(&:user)
```
gregorbg, line 162: It used to be `active_users`, now you're (flat-)mapping to just `users`. Is this intended or a whoopsie?

app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -37,20 +37,10 @@ def index_for_group
...
   # Returns a list of roles primarily based on groupType.
   def index_for_group_type
     group_type = params.require(:group_type)
-    roles = roles_of_group_type(group_type)
-
-    # Filter & Sort roles
-    roles = UserRole.filter_roles(roles, current_user, params)
-    roles = UserRole.sort_roles(roles, params[:sort])
-
+    roles = UserGroup.roles_of_group_type(group_type, current_user, params)
```
gregorbg, line 43: See my remark about using `roles_of_group_type` as a "god method" above.
You are of course free to _also_ provide an _additional_ "does it all" method as a convenience, but the basic `roles_of_group_type` should not do filtering and sorting.

app/controllers/api/v0/api_controller.rb
```diff
@@ -156,7 +156,11 @@ def omni_search
   end
 
   def delegates
-    paginate json: UserGroup.delegate_regions.flat_map(&:active_users)
+    delegate_users = UserGroup.roles_of_group_type(
```
gregorbg, line 159: I am not really a fan of providing one "god method" that does selecting, filtering and sorting all in one.
In this particular example, you are not applying any filter params at all, but the `roles_of_group_type` method will still run through the `filter` method needlessly.
This is especially problematic because your current implementation of the `filter_roles` method loops through all roles objects in-memory, so in this specific case for Delegates, the roles are being loaded and looped through (in the `reject` part of the `UserRole.filter_roles` implementation) for literally no reason.
It would be _much_ more efficient to just return the `UserRole.includes(...).where(...)` part directly, especially for a paginated endpoint like this.

gregorbg (changes requested):

app/controllers/api/v0/user_roles_controller.rb
danieljames-dj: Okay done.

danieljames-dj: Okay done

danieljames-dj (commented):
I've only answered to 2 of your comments. Will come back to the other 2 comments soon.

app/controllers/api/v0/api_controller.rb
gregorbg: Kindly pinging this comment :)

app/controllers/api/v0/api_controller.rb
danieljames-dj: For now, I've reverted the changes in this file. I'll come back to it in another PR as I've some plans for this 😬

gregorbg (approved):

PR #9412, opened 2024-05-24
Allow qualification on the day of the deadline

gregorbg:
As per WRT remark. This PR merely swaps a `<` comparison for a `<=` comparison


dunkOnIT (dismissed):
Edit: Oops, only checked diff not tests
LGTM

dunkOnIT:
@gregorbg we may want to add test cases for `on_or_before` in addition to the `before` tests? Or just update the expectation of the old tests, I haven't looked in detail at which is more appropriate

dunkOnIT (approved):

FinnIckler:
Isn't this V1?

dunkOnIT:
I believe v2 is just calling the qualification class to check qualifications, so it should still apply

timreyn:
Can you also update the text on https://www.worldcubeassociation.org/competitions/WC2025#competition-events ? Under "Qualification" it says "by DATE_MINUS_ONE", and at the bottom it says "before".

Thank you!

PR #9416, opened 2024-05-25
Add 'Admin' view to new competitions table

gregorbg:
*CHANGE* Previously, "Admin" used to be a separate view option next to "List" and "Map". But it's actually also a list so I decided to just turn it into a checkbox.

It is only shown when you are logged in as somebody who actually has admin privileges.
image

Clicking the checkbox starts rendering additional admin details like on the old page
image


app/webpacker/components/CompetitionsOverview/CompetitionsFilters.js
```diff
@@ -52,12 +54,22 @@ function CompetitionsFilters({
         </Form.Field>
       </Form.Group>
 
+      {canViewAdminDetails && shouldShowAdminDetails && (
+        <Form.Group>
+          <Form.Field>
+            <AdminStatusButtonGroup filterState={filterState} dispatchFilter={dispatchFilter} />
+          </Form.Field>
+        </Form.Group>
+      )}
```
kr-matthews, line 63: Can we put this below the checkboxes? It feels weird when clicking a checkbox moves that checkbox down, so you can't directly click again to un-check it.

(Just my initial thoughts from running locally; will review the code later.)

app/webpacker/components/CompetitionsOverview/CompetitionsView.js
```diff
@@ -38,8 +38,14 @@ function CompetitionsView() {
...
-    [debouncedFilterState],
+    () => calculateQueryKey(debouncedFilterState, canViewAdminDetails),
+    [debouncedFilterState, canViewAdminDetails],
+  );
+
+  // Need to make sure that people don't "hijack" admin mode by manipulating the URL
+  const shouldShowAdminDetails = useMemo(
+    () => canViewAdminDetails && filterState.shouldShowAdminDetails,
+    [canViewAdminDetails, filterState.shouldShowAdminDetails],
   );
```
kr-matthews, line 49: There's no benefit to memoizing a simple calculation resulting in a boolean.
```suggestion
  const shouldShowAdminDetails = canViewAdminDetails && filterState.shouldShowAdminDetails;
```

app/webpacker/components/CompetitionsOverview/queryUtils.js
```diff
@@ -4,26 +4,32 @@ import { competitionConstants } from '../../lib/wca-data.js.erb';
...
   if (filterState?.timeOrder === 'past') {
     timeKey = `${filterState.selectedYear}`;
   } else if (filterState?.timeOrder === 'custom') {
     timeKey = `start${filterState.customStartDate}-end${filterState.customEndDate}`;
   }
 
+  let adminStatus = filterState?.adminStatus;
+  if (!canViewAdminDetails) {
+    adminStatus = null;
+  }
```
kr-matthews, line 18:
```suggestion
  const adminStatus = canViewAdminDetails ? filterState?.adminStatus : null
```

kr-matthews (commented):
A few files are getting a bit long, it might be good to break some of them up.

gregorbg:
I want to leave code cleanup for when the feature has been sufficiently tested and is about to be launched

PR #9422, opened 2024-05-27
Remove JWT Token from Local Storage on Logout

FinnIckler:
You can't just set a variable because it has to be available on redirect, so I used session. We do the same for the locale on sign in. 

`session.delete(:should_reset_jwt)` is a nice little trick to check if a value is present, but also immediately delete it.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -22,7 +22,7 @@
 
   <% if session.delete(:should_reset_jwt) %>
     <script>
-      localStorage.setItem("jwt", null);
+      localStorage.removeItem("jwt");
     </script>
   <% end %>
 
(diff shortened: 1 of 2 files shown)


app/views/layouts/application.html.erb
```diff
@@ -20,6 +20,12 @@
     };
   </script>
 
+  <% if session.delete(:should_reset_jwt) %>
+    <script>
+      localStorage.setItem("jwt", null);
```
coder13, line 25: Any reason to use this over `deleteItem`?

FinnIckler: It was because I was looking for `deleteItem` in my IDE, but didn't find it. That's because it's `removeItem`

gregorbg (approved):

PR #9424, opened 2024-05-28
Make Register V2 form clearer

FinnIckler:
Mostly string fixes, but also fixes a merge whoopsie and makes negative `RegistrationMessages` not dissappear automatically.
Still a couple todo from the sheet, but needed the PR to clone

### DIFF
--- a/app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
+++ b/app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
@@ -122,7 +122,7 @@ export default function CompetingStep({
             : 'registrations.errors.exceeds_event_limit.other',
           'negative',
           {
-            count: selectedEvents.length,
+            event_limit: maxEvents,
           },
         ));
       } else {
--- a/app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
+++ b/app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
@@ -310,76 +310,42 @@ export default function CompetingStep({
             />
           </Form.Field>
           {isRegistered ? (
-            <>
-              <Message warning icon>
-                <Popup
-                  trigger={<Icon name="circle info" />}
-                  position="top center"
-                  content={
-                    canUpdateRegistration
-                      ? i18n.t('competitions.registration_v2.register.until', {
-                        date: getMediumDateString(
-                          competitionInfo.event_change_deadline_date
-                          ?? competitionInfo.start_date,
-                        ),
-                      })
-                      : i18n.t('competitions.registration_v2.register.passed')
-                  }
-                />
-                <Message.Content>
-                  <Message.Header>
-                    {i18n.t(
-                      `competitions.registration_v2.register.registration_status.${registration.competing.registration_status}`,
-                    )}
-                  </Message.Header>
-                  {/* eslint-disable-next-line no-nested-ternary */}
-                  {canUpdateRegistration
-                    ? i18n.t('registrations.update')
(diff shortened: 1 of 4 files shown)


app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
```diff
@@ -245,49 +270,45 @@ export default function CompetingStep({
...
-            />
+            {!competitionInfo.events_per_registration_limit
+              && (
+              <p
+                dangerouslySetInnerHTML={{
+                  __html: i18n.t('registrations.preferred_events_prompt_html', {
+                    link: `<a href="${userPreferencesRoute}">here</a>`,
+                  }),
+                }}
+              />
```
gregorbg, line 300: We have an `I18nHtmlTranslate` component to avoid repeating this `dangerouslySetInnerHTML` idiom (with some additional sanitization)

app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
```diff
@@ -305,81 +327,46 @@ export default function CompetingStep({
...
-                </Message.Content>
-              </Message>
-
-              <ButtonGroup widths={2}>
-                {shouldShowUpdateButton && (
-                  <>
-                    <Button
-                      primary
-                      disabled={
-                        isUpdating || !canUpdateRegistration || !hasChanges
```
gregorbg, line 352: Why is this check being removed, and why is it safe to do so?

gregorbg (changes requested):

app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
FinnIckler: cool!

app/webpacker/components/CompetitionsOverview/CompetitionsFilters.js
```diff
@@ -88,7 +88,16 @@ export function EventSelector({
       <label htmlFor="events">
         {`${I18n.t('competitions.competition_form.events')}`}
         <br />
-        <Button disabled={disabled || eventList.length >= maxEvents} primary type="button" size="mini" id="select-all-events" onClick={() => onEventSelection({ type: 'select_all_events' })}>{I18n.t('competitions.index.all_events')}</Button>
+        <Popup
+          disabled={!Number.isFinite(maxEvents)}
+          trigger={
+            <span><Button disabled={disabled || eventList.length >= maxEvents} primary type="button" size="mini" id="select-all-events" onClick={() => onEventSelection({ type: 'select_all_events' })}>{I18n.t('competitions.index.all_events')}</Button></span>
```
gregorbg, line 94: Why are you wrapping the button into a `span`?
And if the button is sometimes disabled, then shouldn't the whole popup also be disabled in the first place? Or will SemUI detect this automatically for you?

app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
gregorbg: Okay :+1:

app/webpacker/components/CompetitionsOverview/CompetitionsFilters.js
FinnIckler: because of a bug that disabled buttons do not probive popups. I am setting the popup to disabled on `disabled={!Number.isFinite(maxEvents)}`

dunkOnIT:
@thewca-bot deploy staging

gregorbg (approved):

PR #9407, opened 2024-05-22
Stripe refunds for Registration V2

FinnIckler:
Almost done. Just want to clean up the Form and don't show refunds when none are available (everything has been refunded)

PR #9425, opened 2024-05-28
Unify payment logic for register v1 and v2

FinnIckler:
Refunds to come


app/controllers/registrations_controller.rb
```diff
@@ -558,9 +558,24 @@ def stripe_webhook
...
+    registration = if uses_v2
+                     competition_id, user_id_s = params[:attendee_id].split("-")
+                     user_id = user_id_s.to_i
+                     MicroserviceRegistration.find_by(competition_id: competition_id, user_id: user_id)
+                   else
+                     Registration.includes(:competition).find(params[:id])
+                   end
     @competition = registration.competition
 
+    if registration.user.id == current_user.id
```
gregorbg, line 573: I suppose that this should be `!==` instead?

app/controllers/registrations_controller.rb
```diff
@@ -558,9 +558,24 @@ def stripe_webhook
   end
 
   def payment_completion
-    registration = Registration.includes(:competition).find(params[:id])
+    uses_v2 = params[:attendee_id].present?
```
gregorbg, line 561: Hm... It seems like you're "overloading" the route params. This existing `payment_completion` route definitely requires a registration ID to be set as part of the route, and it cannot be skipped because it's a hard part of the route structure: `/registrations/:id/payment-completion` (or something like that).
So in the new registration frontend, are you gonna fill a fantasy value for `:id` in the URL? Doesn't feel right.

Instead, may I suggest just changing the route altogether so that it does not include the reg ID at all anymore?
You can use `params[:payment_intent]` which will always be provided as `my/cool/url?payment_intent=123` as GET param by Stripe.
Then, you retrieve the matching Stripe Record from our DB (I think we already do that anyways) and then the `payment_intent` of that Stripe Record.
Payment Intents on the WCA side always have a "holder" which is polymorphic, so you can do a type check on the holder (or simply run string comparisons on the `holder_type` column, which feels even more natural) to find out whether it's a V2 comp being redirected

app/controllers/registrations_controller.rb
```diff
@@ -558,9 +558,24 @@ def stripe_webhook
...
+
+
+    registration = if uses_v2
+                     competition_id, user_id_s = params[:attendee_id].split("-")
+                     user_id = user_id_s.to_i
+                     MicroserviceRegistration.find_by(competition_id: competition_id, user_id: user_id)
```
gregorbg, line 567: I'm sorry to say that the `find_by` entrypoint won't work as expected. It does not communicate with the microservice and will only rely on results that were previously cached.

(Ye, I realize that this is a drawback to my glued-together Microservice+AR integration, and it should probably be improved. But for the time being, we have to accept that we should not use the direct finder methods. See my longer comment above for an alternative suggestion.)

app/controllers/registrations_controller.rb
gregorbg: Compare also the Stripe Webhook endpoint, which does not contain any WCA entity's ID in its params and only ever loads the registration based on our PI's holder.

app/controllers/registrations_controller.rb
```diff
@@ -591,12 +606,20 @@ def payment_completion
...
-        ruby_money.cents,
-        ruby_money.currency.iso_code,
-        charge_transaction,
-        current_user.id,
-      )
+      if uses_v2
+        begin
+          Microservices::Registrations.update_registration_payment(attendee_id, charge.id, ruby_money.cents, ruby_money.currency.iso_code, stripe_intent.status)
+        rescue Faraday::Error
+          return redirect_to competition_register_path(competition_id, "registration_unreachable")
```
gregorbg, line 613: Same old issue with redirect: This will redirect you to /competition/register.registration_unreachable` which probably does not do what you want. Set `flash[:error]` instead.

gregorbg (changes requested):

app/controllers/registrations_controller.rb
FinnIckler: well I'm sorry to say that this works already :upside_down_face:, you can check in the old method. I don't mind just using the stripe id instead though

app/controllers/registrations_controller.rb
FinnIckler: yup

app/controllers/registrations_controller.rb
FinnIckler: copy paste whoopsie

app/controllers/registrations_controller.rb
FinnIckler: It's because the Microservice registration is definitely there because it's created when the payment intent is created

app/controllers/registrations_controller.rb
```diff
@@ -558,12 +558,10 @@ def stripe_webhook
   end
 
   def payment_completion
-    registration = Registration.includes(:competition).find(params[:id])
-    @competition = registration.competition
-
     # Provided by Stripe upon redirect when the "PaymentElement" workflow is completed
     intent_id = params[:payment_intent]
     intent_secret = params[:payment_intent_client_secret]
+    @competition = params[:competition_id]
```
gregorbg, line 564: ...ehm??

FinnIckler: This is necessary due to a hen and egg problem, we can only get the registration out of the payment id, but we can't redirect on `unless stored_record.payment_intent?` because we don't know the competition_id yet. Instead I changed the url to be `competitions/:competition_id/payment-completion` (which I think makes more sense anyway) and then we can get the competition_id directly

FinnIckler: I renamed it to be clearer

config/routes.rb
```diff
@@ -41,9 +41,9 @@
     post 'users/authenticate-sensitive' => 'users#authenticate_user_for_sensitive_edit'
     delete 'users/sign-out-other' => 'sessions#destroy_other', as: :destroy_other_user_sessions
   end
-  post 'registration/:id/refund/:payment_id' => 'registrations#refund_payment', as: :registration_payment_refund
   post 'registration/:id/load-payment-intent' => 'registrations#load_payment_intent', as: :registration_payment_intent
-  get 'registration/:id/payment-completion' => 'registrations#payment_completion', as: :registration_payment_completion
+  post 'competitions/:competition_id/:payment_id/refund' => 'registrations#refund_payment', as: :registration_payment_refund
```
gregorbg, line 45: Might I suggest `competitions/:competition_id/refund/:payment_id`?
That way it's clearer which ID is which.

app/webpacker/lib/requests/routes.js.erb
```diff
@@ -84,11 +84,11 @@ export const adminFixResultsUrl = (personId, competition_id = undefined, event_i
   return `<%= CGI.unescape(Rails.application.routes.url_helpers.admin_fix_results_path) %>?${searchParams.toString()}`
 };
 
-export const paymentFinishUrl = (competitionId, userId) => `<%= Rails.application.routes.url_helpers.finish_url(host: EnvConfig.ROOT_URL) %>?attendee_id=${competitionId}-${userId}`
+export const paymentFinishUrl = (competitionId) => `<%= CGI.unescape(Rails.application.routes.url_helpers.registration_payment_completion("${competitionId}", host: EnvConfig.ROOT_URL)) %>`
 
 export const paymentRefundsUrl = (competitionId, userId) => `<%= CGI.unescape(Rails.application.routes.url_helpers.refunds_path) %>?attendee_id=${competitionId}-${userId}`
 
-export const refundPaymentUrl = (competitionId, userId, paymentId, amount) => `<%= CGI.unescape(Rails.application.routes.url_helpers.refund_path) %>?attendee_id=${competitionId}-${userId}&payment_id=${paymentId}&refund_amount=${amount}`
+export const refundPaymentUrl = (paymentId) => `<%= CGI.unescape(Rails.application.routes.url_helpers.registration_payment_refund("${paymentId}")) %>`
```
gregorbg, line 91: Where did the refund amount go?

FinnIckler: going to be communicated via POST

FinnIckler:
Payments via v2 work! To Test refunds I need to wait until https://github.com/thewca/worldcubeassociation.org/pull/9407 is reviewed/merged (only choke tests remaining)

FinnIckler:
Also `registration.record_refund` uses the RegistrationPayment id

gregorbg (approved):

gregorbg (approved):

PR #9431, opened 2024-05-29
Show error when some action fails in banned competitor page

danieljames-dj:
This PR also edits the Errored component to add the error message.

### DIFF
--- a/app/webpacker/components/Panel/pages/BannedCompetitorsPage/BannedCompetitorForm.jsx
+++ b/app/webpacker/components/Panel/pages/BannedCompetitorsPage/BannedCompetitorForm.jsx
@@ -29,9 +29,7 @@ export default function BanendCompetitorForm({
     }, () => {
       sync();
       closeForm();
-    }, { method: 'POST' }, (error) => {
-      setFormError(error);
-    });
+    }, { method: 'POST' }, setFormError);
   };
 
   const editBannedCompetitor = () => {
@@ -40,9 +38,7 @@ export default function BanendCompetitorForm({
     }, () => {
       sync();
       closeForm();
-    }, { method: 'PATCH' }, (error) => {
-      setFormError(error);
-    });
+    }, { method: 'PATCH' }, setFormError);
   };
 
   const formSubmitHandler = banAction === 'new' ? createNewBannedCompetitor : editBannedCompetitor;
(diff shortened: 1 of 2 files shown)


app/webpacker/components/Panel/pages/BannedCompetitorsPage/BannedCompetitorForm.jsx
```diff
@@ -27,7 +29,9 @@ export default function BanendCompetitorForm({
     }, () => {
       sync();
       closeForm();
-    }, { method: 'POST' });
+    }, { method: 'POST' }, (error) => {
+      setFormError(error);
+    });
```
gregorbg, line 34: Just pass in `setFormError` as a callback reference?

gregorbg:
```suggestion
    }, { method: 'POST' }, setFormError);
```

app/webpacker/components/Panel/pages/BannedCompetitorsPage/BannedCompetitorForm.jsx
```diff
@@ -36,12 +40,15 @@ export default function BanendCompetitorForm({
     }, () => {
       sync();
       closeForm();
-    }, { method: 'PATCH' });
+    }, { method: 'PATCH' }, (error) => {
+      setFormError(error);
+    });
```
gregorbg, line 45: see my other comment

app/webpacker/components/Requests/Errored.js
```diff
@@ -3,21 +3,25 @@ import React from 'react';
...
+            Something went wrong while loading the data
+            {componentName && (
+            <>
+              {' '}
+              for the component &apos;
+              {componentName}
+              &apos;
+            </>
+            )}
+          </>
```
gregorbg, line 24: I don't think the error message alone is always clear enough. Remember that this is a pretty generic component that's being used all over the place, and error messages are not always self-contained.
So perhaps stick with the existing text (pre your change) and _append_ the error text additionally, if present?

gregorbg (changes requested):

app/webpacker/components/Panel/pages/BannedCompetitorsPage/BannedCompetitorForm.jsx
danieljames-dj: Done

app/webpacker/components/Panel/pages/BannedCompetitorsPage/BannedCompetitorForm.jsx
danieljames-dj: Done

gregorbg (approved):

PR #9434, opened 2024-05-30
Banned competitors hotfix - edit actions & add to delegate panel

danieljames-dj:
2 hotfixes:
1. Added banned competitors to delegate panel
2. Edit actions were missing for WDC leader, I missed to take element from array.

### DIFF
--- a/app/webpacker/components/Panel/pages/BannedCompetitorsPage/index.jsx
+++ b/app/webpacker/components/Panel/pages/BannedCompetitorsPage/index.jsx
@@ -39,7 +39,7 @@ export default function BannedCompetitorsPage() {
     return <Errored />;
   }
 
-  const canEditBannedCompetitors = loggedInUserPermissions.canEditGroup(bannedGroup.id);
+  const canEditBannedCompetitors = loggedInUserPermissions.canEditGroup(bannedGroup[0].id);
 
   return (
     <>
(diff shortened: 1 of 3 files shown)


app/webpacker/components/Panel/pages/BannedCompetitorsPage/index.jsx
```diff
@@ -39,7 +39,7 @@ export default function BannedCompetitorsPage() {
     return <Errored />;
   }
 
-  const canEditBannedCompetitors = loggedInUserPermissions.canEditGroup(bannedGroup.id);
+  const canEditBannedCompetitors = loggedInUserPermissions.canEditGroup(bannedGroup[0].id);
```
gregorbg, line 42: Where does the array index zero come from? Why not 1 or 2 or 3?

danieljames-dj: Currently there is only one banned group and it is coming as an array. Since there is only one banned group, we are taking the first banned group.

gregorbg: Meh. Fine for now but not ideal for the future.

gregorbg (approved):

Issue #9398, opened 2024-05-20
Remove all code related to workbook assistant


danieljames-dj:
We don't use workbook assistant anymore, hence all the code related to workbook assistant can be cleaned up from the repo.

elfishhenry:
I have no idea how I'm suppose to do this, but I'd like to contribute. I've already sent a pull request removing said files.

gregorbg:
Fixed in #9707

PR #9435, opened 2024-05-30
Removed workbook assistant

elfishhenry:


danieljames-dj:
Thanks @elfishhenry for working on this.
1. Can you please add 'Fixes #9398' to the start of the PR title so that it will automatically fix the bug?
2. Can you please assign the bug (https://github.com/thewca/worldcubeassociation.org/issues/9398) to yourself so that nobody else will do the same task in parallel?
3. It looks like there is an error for i18n, can you please remove the corresponding keys from en.yml as well?

gregorbg (commented):
There are a bunch of whitespace-only changes in this PR. That is, you're changing the indentation of a lot of YAML keys but not the actual string contents themselves. Why?

image

PR #9460, opened 2024-06-02
Changed the way read access is given to roles

danieljames-dj:
There was a request from WDC that the previous probation members should not be visible to delegates. This created a small mess-up in existing system where a user can either read a group completely or cannot read a group at all. This PR aims to create a new way to check if a user can access a role.


app/models/user_role.rb
```diff
@@ -139,10 +139,32 @@ def discourse_user_group
...
+    if group.banned_competitors?
+      if is_active? # The actively banned competitors can be viewed by following set of people:
+        (
+          user.wdc_team? || # 1. WDC
+          user.ethics_committee? || # 2. WEC
+          user.board_member? || # 3. Board
+          user.staff_delegate? || # 4. Staff Delegates
+          user.weat_team? || # 5. WEAT
+          user.results_team? || # 6. WRT
+          user.admin? # 7. Admin
```
danieljames-dj, line 153: This logic can be moved to user.rb, but I felt having this here itself will be better, mainly because this won't be used anywhere else. Also this makes the function more easy to understand.
Same with the else part.

app/models/user_role.rb
```diff
@@ -139,10 +139,32 @@ def discourse_user_group
...
+        )
+      else
+        (
+          user.wdc_team? || # 1. WDC
+          user.ethics_committee? || # 2. WEC
+          user.board_member? || # 3. Board
+          user.weat_team? || # 4. WEAT
+          user.results_team? || # 5. WRT
+          user.admin? # 6. Admin
```
danieljames-dj, line 162: This and the or list in the if block can be combined by moving to a method. But the thing is the list can change in future - like we can reach a stage were one user can read past and not active, while another user can read active but not past. To handle these situations and to make the function more readable, I went for this approach. Please don't ask to move to user.rb :-)

gregorbg (changes requested):
With the bigger picture in mind, I have to insist on keeping the `:can_read_groups` permission approach.
You simply cannot implement "business details" (like, in the paper-world of WCA, who has access to this and that user role?) in the body of the `can_user_read?` method. Your current PR already blows up the line count from 1 line (or 4, if you count comments) to 26. And that's only for one group.

I suggest that you "split" the `can_read_groups` permission into `can_read_groups_current` and `can_read_groups_history` (names are tentative, feel free to come up with something better). Then the whole check in `user_role` will change from `not hidden || has read permission` to `not hidden || has read_history permission || (is_active? && has read_current permission)`

danieljames-dj:
Okay, I've changed the implementation as you suggested.

app/models/user_role.rb
```diff
@@ -139,10 +139,9 @@ def discourse_user_group
   end
 
   def can_user_read?(user)
-    # A user can view a role if:
-    # 1. the role belongs to a non-hidden group, or
-    # 2. the user has read-access to that group.
-    !group.is_hidden || user&.has_permission?(:can_read_groups, group.id)
+    return true unless group.is_hidden # Roles of non-hidden groups are public
+    return false if user.nil? # Roles of hidden groups are visible only to a set of users based on permisssions.
+    user&.has_permission?(is_active? ? :can_read_groups_current : :can_read_groups_past, group.id)
```
gregorbg, line 144: The `user&.has_permission?` null-safe `&` is unnecessary because you check `if user.nil?` in the line directly above

gregorbg: Also, the line would feel less busy if you extract the ternary check to a separate line and store it in a `role_permission` variable. Then just check `user.has_permission?(role_permission, group.id)`

app/models/user.rb
```diff
@@ -638,8 +657,11 @@ def permissions
       can_create_groups: {
         scope: groups_with_create_access,
       },
-      can_read_groups: {
```
gregorbg, line 641: Is the rest of the codebase updated to understand that `can_read_groups` does not exist anymore?

gregorbg: I'm a bit worried because `has_permission?` in user.rb does not check whether the name of the permission is valid.
`some_random_user.has_permission?(:give_me_bananas)` will simply return `false` at the moment without complaining about the (intentionally) ridiculous and non-existent `:give_me_bananas` permission name.

This means that in the context of this PR, all old calls using `current_user.has_permission?(:can_read_groups)` will continue to "work", i.e. they will not throw an error but they will always return false even when the user has full (i.e. past) read access to the group.

app/models/user.rb
```diff
@@ -577,11 +577,30 @@ def banned_at_date?(date)
...
     return "*" if can_edit_any_groups?
     groups = groups_with_edit_access
 
-    if can_view_banned_competitors?
+    if can_view_current_banned_competitors?
+      groups += UserGroup.banned_competitors.ids
+    end
+
+    groups
+  end
```
gregorbg, line 597: Design decision / code philosophy: Should the list of groups with "current" access also include the list of groups with "past" access?

Background: If you have edit access, the groups are also considered for (both cases of) read access.
But if you have full (i.e. past) read access, the group is not considered for restricted (i.e. current) read access.
In my mind, `edit_access > past_read_access > current_read_access`, so every permission group should also consider "higher" permissions. Is this feasible? Or is there some edge case that I'm not considering properly here?

app/models/user.rb
```diff
@@ -577,11 +577,30 @@ def banned_at_date?(date)
...
   end
 
-  private def groups_with_read_access
+  private def can_view_current_banned_competitors?
+    wdc_team? || ethics_committee? || board_member? || staff_delegate? || weat_team? || results_team? || admin?
+  end
+
+  private def can_view_past_banned_competitors?
+    wdc_team? || ethics_committee? || board_member? || weat_team? || results_team? || admin?
+  end
```
gregorbg, line 586: In line with my other, longer comment about "access hierarchy" below: Can this be simplified to an implementation where `can_view_current` is simplified to `can_view_past || staff_delegate?`?

gregorbg (changes requested):
Thanks for the updated implementation, much better now!
Although it raises some implementation concerns that need to be discussed before merging :upside_down_face:

app/models/user_role.rb
danieljames-dj: Okay applied both changes, thanks.

app/models/user.rb
danieljames-dj: Yes these doesn't exist anymore. I made sure that.

app/models/user.rb

app/models/user.rb
danieljames-dj: Agreed, this should cover all cases probably. Cannot think of any edge case where this fails. If somebody can see a past role, they can definitely see the current role.

gregorbg: If you agree, then `groups_with_read_access_for_current` should probably reference `groups_with_read_access_for_past`. But it currently doesn't...

danieljames-dj: Oh yes. Added now. I initially added only to `can_view_current_banned_competitors?`.

gregorbg (approved):

PR #9461, opened 2024-06-02
POC of single user_roles get API

danieljames-dj:
Currently there are three different roles API, and there are different filter parameters for each API. The querying in the backend is not so good. So, this API will do proper pre-filtering before fetching the actual roles, and hence making the querying more faster.

If this approach is fine, I'll deprecate the other three GET APIs and make this as the single API required for roles fetching.

### DIFF
--- a/app/controllers/api/v0/user_roles_controller.rb
+++ b/app/controllers/api/v0/user_roles_controller.rb
@@ -11,7 +11,7 @@ class Api::V0::UserRolesController < Api::V0::ApiController
     User.where(delegate_to_handle_wca_id_claim: user.id).update_all(delegate_id_to_handle_wca_id_claim: nil, unconfirmed_wca_id: nil)
   end
 
-  def pre_filtered_user_roles_active_record
+  private def pre_filtered_user_roles
     active_record = UserRole
     is_active = params.key?(:isActive) ? ActiveRecord::Type::Boolean.new.cast(params.require(:isActive)) : nil
     group_type = params[:groupType]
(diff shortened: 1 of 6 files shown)


app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -11,6 +11,38 @@ class Api::V0::UserRolesController < Api::V0::ApiController
...
+    end
+    if user_id.present?
+      active_record = active_record.where(user_id: user_id)
+    end
+    active_record
+  end
+
+  # Returns a list of roles based on the parameters.
+  def index
+    roles = pre_filtered_user_roles_active_record.all
```
gregorbg, line 36: The `.all` call at the end feels redundant. What happens if you drop it?

app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -11,6 +11,38 @@ class Api::V0::UserRolesController < Api::V0::ApiController
     User.where(delegate_to_handle_wca_id_claim: user.id).update_all(delegate_id_to_handle_wca_id_claim: nil, unconfirmed_wca_id: nil)
   end
 
+  def pre_filtered_user_roles_active_record
```
gregorbg, line 14: You don't have to include `_active_record` in the name

app/models/user_role.rb
```diff
@@ -11,6 +11,7 @@ class UserRole < ApplicationRecord
   delegate :group_type, to: :group
 
   scope :active, -> { where(end_date: nil).or(where.not(end_date: ..Date.today)) }
```
gregorbg, line 13: Crazy Rails magic: (read `inactive.invert_where` as "where NOT inactive")

```suggestion
  scope :active, -> { where(end_date: nil).or(inactive.invert_where) }
```

gregorbg (changes requested):

gregorbg (approved):

PR #9462, opened 2024-06-02
Create db/seeds for necessary roles

danieljames-dj:
Currently db/seeds is not working, this PR is to fix them.

### DIFF
--- a/db/seeds/development/competitions.seeds.rb
+++ b/db/seeds/development/competitions.seeds.rb
@@ -29,7 +29,7 @@ after "development:users", "development:user_roles" do
 
   countries = Country.all
 
-  delegate = RolesMetadataDelegateRegions.includes(:user_role).where.not(status: RolesMetadataDelegateRegions.statuses[:trainee_delegate]).select { |role_metadata| role_metadata.user_role.is_active? }.sample.user
+  delegate = RolesMetadataDelegateRegions.joins(:user_role).not_trainee_delegate.merge(UserRole.active).sample.user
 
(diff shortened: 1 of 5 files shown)


db/seeds/development/competitions.seeds.rb
```diff
@@ -29,7 +29,7 @@ def random_city(country)
 
   countries = Country.all
 
-  delegate = User.find_by(delegate_status: "delegate")
+  delegate = RolesMetadataDelegateRegions.includes(:user_role).where.not(status: RolesMetadataDelegateRegions.statuses[:trainee_delegate]).select { |role_metadata| role_metadata.user_role.is_active? }.sample.user
```
gregorbg, line 32: Using Rails `enum` gives you `not_*` scopes for free, so using `RolesMetadataDelegateRegions.not_trainee_delegate` should work.

gregorbg: Also, you can use the `active` scope on `UserRole` to execute this query directly on DB level. The magic feature is called `merge`, which can combine arbitrary `scope` definitions:
```suggestion
  delegate = RolesMetadataDelegateRegions.joins(:user_role).not_trainee_delegate.merge(UserRole.active).sample.user
```

gregorbg (changes requested):

db/seeds/development/competitions.seeds.rb
danieljames-dj: Wow thanks a lot, this is nice. Accepted the suggestion.

gregorbg (approved):

PR #9508, opened 2024-06-04
Remove teams & team_members tables

danieljames-dj:
Title says everything.


app/models/user.rb
```diff
@@ -1121,7 +1117,7 @@ def url
     only: ["id", "wca_id", "name", "gender",
            "country_iso2", "created_at", "updated_at"],
     methods: ["url", "country", "delegate_status"],
-    include: ["avatar", "teams"],
+    include: ["avatar"],
```
gregorbg, line 1120: This PR would be a good opportunity to think about optimizing the `deprecated_team_roles` method.

As a matter of fact, currently there is a snippet of code in the serialization logic for users that reads
```ruby
if include_teams
  json[:teams] = deprecated_team_roles
end
```

and this will become silent (or "dead") code after this change. Some parts of our application may break if they rely on team roles being serialized to the user json.

gregorbg: (as a matter of fact, that's why some of the non-choke CI tests fail)

config/locales/en.yml
```diff
@@ -2284,66 +2284,11 @@ en:
...
-        name: "WCA Sports Organization Team"
-      #context: WAT
-      wat:
-        name: "WCA Archive Team"
-      #context: Hidden team for list of banned competitors
-      banned:
-        name: "Banned Competitors"
-      #context: Hidden team for WCT China members
-      wct_china:
-        name: "WCT China"
```
gregorbg, line 2341: where did these go? Are there already adequate replacements?

config/locales/en.yml
```diff
@@ -2284,66 +2284,11 @@ en:
...
       operations_html: "The day to day operations are run by the WCA's %{committees_and_teams}."
       teams_committees: "Teams and Committees"
-      #context: Hidden team for Software administrators
-      wst_admin:
-        name: "Software admins"
```
gregorbg, line 2346: see other comment about replacements in the locale file

gregorbg (changes requested):

app/models/user.rb
danieljames-dj: Oh ya I forgot that this was anyway loading made-up teams and not the actual teams. My bad. Added it back.

Regarding optimizing `deprecated_team_roles`, there is already some optimization in https://github.com/thewca/worldcubeassociation.org/pull/9411, but I don't remember how many queries it's taking now. Once that PR is merged, I'll take up that as a different PR.

config/locales/en.yml
danieljames-dj: These were used to show in the teams page probably. They already have a replacement under `page.teams_committees_councils.groups_name`.

config/locales/en.yml
danieljames-dj: This is no longer used. The internal page is anyway showing this unlocalized. For the public facing teams, we already have the replacement as mentioned in previous comment.

spec/models/reassign_wca_id_spec.rb
```diff
@@ -50,7 +50,6 @@
   end
 
   it "can actually reassign wca id" do
-    team_member = FactoryBot.create(:team_member, user_id: account1.id)
```
gregorbg, line 53: You should probably add a replacement dummy data that tests reassignment of new Roles instead of old TeamMember

gregorbg (changes requested):

spec/models/reassign_wca_id_spec.rb
danieljames-dj: Okay done

gregorbg (approved):

PR #9510, opened 2024-06-04
Remove isGroupHidden parameter from post-filter

danieljames-dj:
This PR aims at removing all usages of `isGroupHidden` from the role list APIs. This is achieved by replacing the API with new list API, so that the `isGroupHidden` parameter will be pre-filtered instead of post-filter.

### DIFF
--- a/app/controllers/api/v0/user_roles_controller.rb
+++ b/app/controllers/api/v0/user_roles_controller.rb
@@ -14,6 +14,7 @@ class Api::V0::UserRolesController < Api::V0::ApiController
   private def pre_filtered_user_roles
     active_record = UserRole
     is_active = params.key?(:isActive) ? ActiveRecord::Type::Boolean.new.cast(params.require(:isActive)) : nil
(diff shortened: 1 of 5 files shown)


app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -22,6 +23,9 @@ class Api::V0::UserRolesController < Api::V0::ApiController
     if !is_active.nil?
       active_record = is_active ? active_record.active : active_record.inactive
     end
+    if !is_group_hidden.nil?
+      active_record = active_record.includes(:group).where(group: { is_hidden: is_group_hidden })
+    end
```
danieljames-dj, line 28: I feel there is a better way to handle this. Also, I'm not sure if this works as expected in all cases (worked in all cases in my testing though).

gregorbg: What kind of better way are you looking for?
You're using a very similar approach for `group_type` below, so I don't see what would be wrong with this approach either

danieljames-dj: As we discussed over slack, I'm convinced that this is a good way. :-)

gregorbg (approved):

PR #9524, opened 2024-06-05
Set unique IDs for hidden AutoNumeric fields in Refunds view

gregorbg:
This is actually a bug that exists in Prod, so it's rather urgent.

The hidden fields backing the "old-style" (non-React) AutoNumeric inputs get assigned repetitive IDs. In the old (i.e. current registration system) "Refunds" panel, we render one form per payment, and each payment is structurally identical so the ID `payment_refund_amount` gets assigned to **every** hidden field.

We did not encounter this issue yet because (a) it's very rare that someone has two payments in the first place and (b) that the _second_ (or third or fourth) payment should only be _partially_ refunded.
But if any organizer would have tried that, the second AutoNumeric field's jQuery would have fired on the _first_ hidden field of the page because that's what it finds by ID

### DIFF
--- a/app/inputs/money_amount_input.rb
+++ b/app/inputs/money_amount_input.rb
@@ -18,8 +18,13 @@ class MoneyAmountInput < SimpleForm::Inputs::Base
     # Get the id of the currency selector's selector
     currency_selector = options.delete(:currency_selector)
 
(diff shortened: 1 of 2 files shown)


gregorbg:
Obsolete now that the new reg system is launched

PR #9525, opened 2024-06-06
Increased maximum size of roles list API from 100 to 1000

danieljames-dj:
In all delegates page, there are more than 100 delegates, so this limit needs to be increased to have all delegates there


gregorbg:
You should probably just begin paginating instead of hard-cutting the first N entries.
We have a gem installed in our Ruby backend that makes pagination very painless: Instead of `render json: foo_bar` you call `paginate json: foo_bar` and it does the same as `render` but it handles the SQL pagination and the page size header information that is sent back to the client for you

danieljames-dj:
Pagination is nice, I have changed to that. But still I'm forcing to have 1000 as the per_page limit for just one API call, because otherwise I've to change the UI a lot. Also, for 'download csv' button of the page, I'll have to iteratively call the API till I reach all the delegates which requires lots of coding efforts. Considering this page to be internal, I guess 1000 limit should be fine. Please let me know if you have any suggestions.

gregorbg:
I think the pagination gem has a default for setting `per_page` through your request: https://github.com/davidcelis/api-pagination

gregorbg:
Also, using the `respond_to` approach, you can have the JSON endpoint return paginated content but at the same time provide a CSV endpoint that returns a full list of non-paginated roles.

danieljames-dj:
Thanks for the knowledge on existing `per_page` param. This helped, I've made the change. I think I'll not go ahead with respond_to because the role structure can be different for different group types.

app/webpacker/lib/requests/routes.js.erb
```diff
@@ -195,7 +195,7 @@ export const apiV0Urls = {
     resetClaimCount: (wcaId) => `<%= CGI.unescape(Rails.application.routes.url_helpers.api_v0_wrt_person_reset_claim_count_path("${wcaId}"))%>`,
   },
   userRoles: {
-    list: ({isActive, isGroupHidden, userId, groupId, status, groupType, isLead} = {}, sort) => `<%= CGI.unescape(Rails.application.routes.url_helpers.api_v0_user_roles_path) %>?${jsonToQueryString({ sort, isActive, isGroupHidden, userId, groupId, status, groupType, isLead })}`,
+    list: ({isActive, isGroupHidden, userId, groupId, status, groupType, isLead} = {}, sort, per_page) => `<%= CGI.unescape(Rails.application.routes.url_helpers.api_v0_user_roles_path) %>?${jsonToQueryString({ sort, isActive, isGroupHidden, userId, groupId, status, groupType, isLead, per_page: per_page || 100 })}`,
```
gregorbg, line 198: You could just provide the value `100` as a JS default argument
```suggestion
    list: ({isActive, isGroupHidden, userId, groupId, status, groupType, isLead} = {}, sort, perPage = 100) => `<%= CGI.unescape(Rails.application.routes.url_helpers.api_v0_user_roles_path) %>?${jsonToQueryString({ sort, isActive, isGroupHidden, userId, groupId, status, groupType, isLead, per_page: perPage })}`,
```

gregorbg (changes requested):
Cool! Have you tested that the `paginate` call from `api-pagination` correctly reads the `per_page` param without any further configuration?

app/webpacker/lib/requests/routes.js.erb
danieljames-dj: Oh yes, accepted suggestion.

danieljames-dj:
> Cool! Have you tested that the `paginate` call from `api-pagination` correctly reads the `per_page` param without any further configuration?

Yes I've tested it. It's correctly reading the `per_page` param.

gregorbg (approved):

PR #9531, opened 2024-06-11
More small Registration V2 Fixes

FinnIckler:
- Fixes Guests being errored incorrectly
- Fixed Event Selector showing error before being interacted with if no preferred event is set or a favourite competition is used
- Fixed a bunch of strings
- deleted some unused imports
- flipped the columns of the waiting list table

Needs a small PR on wca-registration to add `registration?.payment.payment_amount_human_readable`([already done, just need to open it](https://github.com/thewca/wca-registration/pull/605))

### DIFF
--- a/app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
+++ b/app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
@@ -341,7 +341,7 @@ export default function CompetingStep({
 
               {shouldShowReRegisterButton && (
               <Button
-                secondary
+                primary
                 disabled={isUpdating}
                 type="submit"
               >
--- a/app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
+++ b/app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
@@ -186,7 +186,6 @@ export default function CompetingStep({
   };
 
   const actionReRegister = () => {
-    dispatch(setMessage('competitions.registration_v2.update.being_updated', 'basic'));
     updateRegistrationMutation({
       user_id: registration.user_id,
       competition_id: competitionInfo.id,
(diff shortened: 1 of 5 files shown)


dunkOnIT:
@thewca-bot deploy staging

config/locales/en.yml
```diff
@@ -1772,8 +1772,8 @@ en:
         acknowledgement: "I have read and acknowledged all information under the 'Register' tab, including entry fees, refund policies and any extra requirements listed above."
         next_step: "Continue to next Step"
       update:
-        being_updated: "Registration is being updated"
-        being_deleted: "Registration is being deleted"
+        being_updated: "Registration is being updated..."
+        being_deleted: "Registration is being deleted..."
```
gregorbg, line 1776: How are we handling `...` ellipsis in other locale string?
I know there is a single Unicode `…` character as opposed to three `.` characters

FinnIckler: seems to be all `...` and no `…`

app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
```diff
@@ -83,15 +86,27 @@ export default function CompetingStep({
...
         {
           ...data.registration,
           payment: registration.payment,
         },
       );
-      nextStep();
+      // Going from pending -> Cancelled
+      if (data.registration.competing.registration_status === 'cancelled') {
+        dispatch(setMessage('competitions.registration_v2.register.registration_status.cancelled', 'positive'));
+        nextStep({ toStart: true });
```
FinnIckler, line 99: I could return here to next one level deeper?

gregorbg (approved):

app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
gregorbg: Seems you already made up your mind :smile:

PR #9534, opened 2024-06-12
Remove limit from venueAddress and venueDetails from Dumper Schema

FinnIckler:
Not sure if a migration is needed


gregorbg (approved):
Thanks for the clean fix!

No migration needed this time because we don't persist any actual data to those tables.
We only load the schema on-the-fly when creating a dump, so this should be fine

PR #9535, opened 2024-06-12
Use Bulk Update in Registration Admin V2 and some string changes

FinnIckler:
I added `registration_full_v2_html` but it's not shown as HTML in the warning, not sure what else I have to do?

Bulk update relies on https://github.com/thewca/wca-registration/pull/605, will split this into another PR after testing is done


FinnIckler:
Seems like i18n can´t figure out that `uses_new_registration_service? ? 'registrations.registration_full_v2_html' : 'registrations.registration_full'` doesn't mean the keys are unused so I will just use an if

gregorbg (approved):

PR #9539, opened 2024-06-13
Make the refund endpoint gateway-agnostic

gregorbg:
Streamline Stripe and PayPal refunds through the exact same API.

Needs to add a few methods to `Connected*Account` and `*Record` because during runtime, `*` could either be `Stripe` or `Paypal` and both need to fulfill the same "interface" (despite the fact that there are no interfaces in Ruby)

Partial implementation of https://github.com/thewca/worldcubeassociation.org/pull/9208


gregorbg:
Tests will inevitable fail because of a quirk that needs https://github.com/thewca/worldcubeassociation.org/pull/9540 to be merged first: The current ``create_paypal_order` implementation is a proof-of-concept that doesn't properly create a PaymentIntent. But the refunds need the payment intent to access the holder registration.

app/webpacker/components/RegistrationsV2/api/payment/get/refundPayment.js
```diff
@@ -7,7 +7,7 @@ export default async function refundPayment({
   amount,
 }) {
   return fetchWithAuthenticityToken(
-    refundPaymentUrl(competitionId, paymentId),
+    refundPaymentUrl(competitionId, 'stripe', paymentId),
```
dunkOnIT, line 10: Looks like this is still hardcoded to Stripe - will we need to update it to also support Paypal in the future? I assume yes, and the reason we aren't doing so is because it's hard to pass a `paypal_enabled?` flag to the frontend?

app/models/connected_stripe_account.rb
```diff
@@ -70,14 +70,16 @@ def prepare_intent(registration, amount_iso, currency_iso, paying_user)
     )
   end
 
-  def issue_refund(charge_id, amount_iso)
-    charge_record = StripeRecord.charge.find_by!(stripe_id: charge_id)
+  def find_payment_record(record_id)
```
dunkOnIT, line 73: Is the naming here appropriate given that we're only returning `charge` payment record objects? I'd expect a method like this to either not limit its search to a scope, or take a `record_type` parameter which defines which scope will be searched.

dunkOnIT (commented):
Also looking good, just a couple of questions. Will need to test it locally when the other PR is merged ofc

app/models/connected_stripe_account.rb
gregorbg: Renamed.

dunkOnIT:
Have tested full and partial refunds with both payment providers - all working as expected.

I did find one pretty niche issue - but i think this is better handled during a more formal integration testing phase:
- User pays via Paypal
- organizer refunds payment
- organizer disconnects paypal and connects Stripe
- User tries to pay via Stripe

Generates the following error stack:
```bash
 Completed 500 Internal Server Error in 78ms (ActiveRecord: 9.0ms | Allocations: 22062)
rails        | 
rails        | 
rails        | [User Id] Request was made by user id: 158816
rails        |   
rails        | NoMethodError (undefined method `account_id' for an instance of PaypalRecord):
rails        |   
rails        | app/models/payment_intent.rb:15:in `account_id'
rails        | app/models/connected_stripe_account.rb:10:in `block in prepare_intent'
rails        | app/models/connected_stripe_account.rb:9:in `prepare_intent'
rails        | app/controllers/registrations_controller.rb:663:in `load_payment_intent'
rails        | lib/middlewares/warden_user_logger.rb:13:in `call'
rails        | lib/middlewares/fix_accept_header.rb:14:in `call'
rails        | 172.18.0.1 - - [29/Jun/2024:05:26:23 UTC] "POST /registration/925874/load-payment-intent/stripe HTTP/1.1" 500 16777
rails        | http://localhost:3000/competitions/NubOpenMissionViejoSummer2024/register -> /registration/925874/load-payment-intent/stripe
```

dunkOnIT (approved):
No longer trying to recycle PI's for a different CPI - approved!

PR #9546, opened 2024-06-17
Implement V2 Register Page User Feedback

FinnIckler:


app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
```diff
@@ -93,11 +96,6 @@ export default function CompetingStep({
           payment: registration.payment,
         },
       );
-      // Going from pending -> Cancelled
-      if (data.registration.competing.registration_status === 'cancelled') {
-        dispatch(setMessage('competitions.registration_v2.register.registration_status.cancelled', 'positive'));
-        return nextStep({ toStart: true });
-      }
```
gregorbg, line 100: Why is this transition/step being dropped?

app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
```diff
@@ -172,15 +170,21 @@ export default function CompetingStep({
...
+        competition_id: competitionInfo.id,
+        competing: {
+          comment: hasCommentChanged ? comment : undefined,
+          event_ids: hasEventsChanged ? selectedEvents : undefined,
+        },
+        guests,
+      });
+    }).catch(() => {
+      nextStep();
     });
```
gregorbg, line 188: Am I reading it correctly that you're still proceeding to the `nextStep` even if the user does NOT confirm (which is what the `catch` branch represents in this case)?

app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
```diff
@@ -350,6 +337,9 @@ export default function CompetingStep({
                   {i18n.t('registrations.update')}
                 </Button>
                 <ButtonOr />
+                <Button secondary onClick={() => nextStep()}>
+                  {i18n.t('competitions.registration_v2.register.view_registration')}
+                </Button>
```
gregorbg, line 342: The naming scheme here feels unintuitive to me. The button to **view** your registration (according to the i18n text, at least) internally invokes `nextStep` which in my mind is a mutation. So a button that touts to "view" actually "mutates". Am I missing something here?

app/webpacker/components/RegistrationsV2/Register/RegistrationOverview.jsx
```diff
@@ -37,15 +86,16 @@ export default function RegistrationOverview({
...
         <Header>{i18n.t('competitions.nav.menu.registration')}</Header>
         <Form onSubmit={nextStep} size="large">
           <FormField>
             <label>
               {i18n.t('activerecord.attributes.registration.registration_competition_events')}
               :
             </label>
-            {registration.competing.event_ids.map((id) => (<EventIcon key={id} id={id} style={{ cursor: 'unset' }} />))}
+            { /* Make sure to keep WCA Event order */}
+            {events.official.flatMap((e) => (registration.competing.event_ids.includes(e.id) ? <EventIcon key={e.id} id={e.id} style={{ cursor: 'unset' }} /> : []))}
```
gregorbg, line 98: Maybe do a `filter` first instead of this "hidden filter" by flat-mapping to an empty array?

config/locales/en.yml
```diff
@@ -1775,6 +1775,7 @@ en:
       update:
         being_updated: "Registration is being updated..."
         being_deleted: "Registration is being deleted..."
+        update_confirm: "Are you sure you want to update your Registration?"
```
gregorbg, line 1778: lower-case `registration`?

app/webpacker/components/RegistrationsV2/Register/RegistrationOverview.jsx
```diff
@@ -63,15 +113,24 @@ export default function RegistrationOverview({
             </label>
             {registration.guests}
           </FormField>
-          { editsAllowed && (
+          <ButtonGroup widths={2}>
```
gregorbg, line 116: What happens when `deleteAllowed` is false, but the button group width is still set to `2`?

gregorbg (changes requested):

app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
FinnIckler: because deletion is now handled on the registration overview

app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
FinnIckler: this brings the user back to the RegistrationOverview page

app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
FinnIckler: no mutation happening, the next step from the Update page is the View registration page

app/webpacker/components/RegistrationsV2/Register/RegistrationOverview.jsx
FinnIckler: I chose to flat map so we only iterate over the array once, sadly JS does not have filterMap

app/webpacker/components/RegistrationsV2/Register/RegistrationOverview.jsx
FinnIckler: let me see

config/locales/en.yml
FinnIckler: fixed

app/webpacker/components/RegistrationsV2/Register/RegistrationOverview.jsx
FinnIckler: still works

dunkOnIT:
@thewca-bot deploy staging

app/webpacker/components/RegistrationsV2/Register/RegistrationOverview.jsx
gregorbg: But you're not iterating over an excessively long array or over something that is costly to compute.

If every array iteration triggered lots and lots of DB calls, then fine. But here, we have a pre-computed list in-memory that is never longer than ~20 entries. Even doing one `includes` call per event is still max 400 calls which literally every processor of any computer less than 20 years old sneezes out in a time faster than I care to find the micro-femto-atto-prefix for :P

TL;DR: Readability and clarity goes over iteration optimization in simple cases like this

FinnIckler: changed!

gregorbg (approved):

PR #9550, opened 2024-06-18
Including users and groups when loading user_roles

danieljames-dj:
The user data and group data is used for post filtering, hence precomputing them for the roles list API.

### DIFF
--- a/app/controllers/api/v0/user_roles_controller.rb
+++ b/app/controllers/api/v0/user_roles_controller.rb
@@ -12,15 +12,12 @@ class Api::V0::UserRolesController < Api::V0::ApiController
   end
 
   private def pre_filtered_user_roles
-    active_record = UserRole
+    active_record = UserRole.includes(:user, :group) # Including user & group for post filtering.
     is_active = params.key?(:isActive) ? ActiveRecord::Type::Boolean.new.cast(params.require(:isActive)) : nil
     is_group_hidden = params.key?(:isGroupHidden) ? ActiveRecord::Type::Boolean.new.cast(params.require(:isGroupHidden)) : nil
     group_type = params[:groupType]
     group_id = params[:groupId]
     user_id = params[:userId]
-
-    # For post filtering, user & group is needed.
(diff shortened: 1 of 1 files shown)


app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -47,8 +50,8 @@ def index
     roles = UserRole.filter_roles(roles, current_user, params)
     roles = UserRole.sort_roles(roles, params[:sort])
 
-    # Limiting to first 100 elements of roles array to avoid serializing of large array.
-    render json: roles.first(100)
+    # Paginating the list by sending only first 100 elements unless mentioned in the API.
+    paginate json: roles
```
gregorbg, line 54: This change seems like it was included in your previous pagination PR.
Is there maybe a chance you didn't include the latest changes from `main` in your currentPR branch?

app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -19,6 +19,9 @@ class Api::V0::UserRolesController < Api::V0::ApiController
     group_id = params[:groupId]
     user_id = params[:userId]
 
+    # For post filtering, user & group is needed.
+    active_record = active_record.includes([:user, :group])
```
gregorbg, line 23: First, `includes` takes varargs, so `.includes(:user, :group)` will work.
Second, if you know that these two are going to be `includes`'d anyways no matter what, then you can just declare `active_record = UserRole.includes(...)` directly at the top of the method above.

gregorbg (changes requested):

app/controllers/api/v0/user_roles_controller.rb
danieljames-dj: Oops I don't know how this came, removed now.

app/controllers/api/v0/user_roles_controller.rb
danieljames-dj: Oh ya, thanks for that. Implemented.

gregorbg (approved):

PR #9556, opened 2024-06-18
Created a new Admin panel

danieljames-dj:
This admin panel is in React. This admin panel will be having all the panel pages, and hence WRT & WST Admin team can view all of them easily in a single panel.

### DIFF
--- a/app/controllers/panel_controller.rb
+++ b/app/controllers/panel_controller.rb
@@ -17,6 +17,7 @@ class PanelController < ApplicationController
   before_action -> { redirect_to_root_unless_user(:can_access_wdc_panel?) }, only: [:wdc]
   before_action -> { redirect_to_root_unless_user(:can_access_wec_panel?) }, only: [:wec]
   before_action -> { redirect_to_root_unless_user(:can_access_weat_panel?) }, only: [:weat]
+  before_action -> { redirect_to_root_unless_user(:can_access_admin_panel?) }, only: [:admin]
 
   def pending_claims_for_subordinate_delegates
     # Show pending claims for a given user, or the current user, if they can see them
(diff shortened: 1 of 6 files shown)


gregorbg:
No objections to the code changes themselves, but the whole admin UI is getting quite "panel-heavy". You're basically introducing a new dropdown menu item (in the top right corner avatar icon dropdown) for every team panel, and WST (as well as WRT) will see every single one of these menu items because they have admin privileges.
Now you're _also_ introducing an extra master panel which holds all panel pages, and it generates an extra dropdown menu entry.

I certainly understand your good intentions, but is there any way we can reduce the amount of panels and panel-panels?

danieljames-dj:
Hey @gregorbg, so my intention is to have only admin panel and WST panel for WST admin teams, no other panels will be shown. If WST admin want to debug any panel page, they can access the same from admin panel because all other panels are actually subset of admin panel. I hope this clarifies.

That being said, even if this PR is merged, still WST admin team members can see all the panels in the navigation bar. This is because some panel pages are yet to be moved to the generic one which I'll be doing in a later stage. Once that is done, other panels will be removed from admin's navigation bar.

My future folder structure will be like:

app/webpacker/components/Panel/
- pages/ (none of the pages in this folder are connected to any teams, it can be added to any team's panels)
  - PanelPage1/
  - PanelPage2/
  - PanelPage3/
- PanelTemplate.jsx (Basic template for each team's panel)
- PanelPages.jsx (List of each panel pages which are in pages folder)
- TeamPanel1.jsx
- TeamPanel2.jsx
- TeamPanel3.jsx

gregorbg (approved):

PR #9564, opened 2024-06-22
Add error to contact form if submission fails

danieljames-dj:
Currently if there is an error for the contact form submission, we are not informing the user of the same. This PR is to achieve that.

### DIFF
--- a/app/webpacker/components/ContactsPage/ContactForm.jsx
+++ b/app/webpacker/components/ContactsPage/ContactForm.jsx
@@ -16,6 +16,7 @@ import Wct from './SubForms/Wct';
 import Wrt from './SubForms/Wrt';
 import Wst from './SubForms/Wst';
 import Competition from './SubForms/Competition';
+import Errored from '../Requests/Errored';
 
 const CONTACT_RECIPIENTS = [
   'competition',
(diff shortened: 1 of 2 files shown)


app/webpacker/components/ContactsPage/ContactForm.jsx
```diff
@@ -66,6 +68,7 @@ export default function ContactForm({ loggedInUserData }) {
   }, [selectedContactRecipient]);
 
   if (saving) return <Loading />;
+  if (saveError) return <Errored error={saveError} />;
```
gregorbg, line 71: I think the error handler of our `useSaveAction` returns the actual error _object_. This means that the `Errored` property here might end up just displaying: "An error occured! Details: [Object object]".

Have you tested this? You can temporarily modify the backend controller to just execute `raise "Something went horribly wrong!"` directly as the first line, to simulate an error. And then see how the frontend behaves.

danieljames-dj: Yes, it returns error object. If I do `raise "Something went horribly wrong!"`, it will show some weird error:
Screenshot 2024-06-24 at 7 06 57 AM

But instead for `return render status: :bad_request, json: { error: "Invalid arguments" }`, I'll get following:
Screenshot 2024-06-24 at 7 08 11 AM

The last one is the current behavior. Is this fine or do you have an alternative suggestion?

(I also just realized I missed to add `return` for that error in contact_controller. Added it now)

gregorbg (approved):

PR #9570, opened 2024-06-23
Add WRT to role change emails

danieljames-dj:
Add WRT to the role change emails. This is to handle mistaken role changes. For example, sometimes leaders will accidentally promote wrong person and then undo it and do the correct action. In these cases, the role table gets manipulated with wrong value, which needs to be fixed by WRT.

### DIFF
--- a/app/mailers/role_change_mailer.rb
+++ b/app/mailers/role_change_mailer.rb
@@ -1,6 +1,14 @@
 # frozen_string_literal: true
 
 class RoleChangeMailer < ApplicationMailer
+  private def wrt_email_recipient
+    UserRole::UserRoleEmailRecipient.new(
+      name: UserGroup.teams_committees_group_wrt.name,
+      email: UserGroup.teams_committees_group_wrt.metadata.email,
+      message: 'Please take action if this role change is inconsistent or accidental.',
+    )
+  end
+
   private def role_metadata(role)
     metadata = {}
     group = role.group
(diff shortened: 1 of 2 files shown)


gregorbg:
Is this really the way to go?

As an alternative, we could consider making it as hard as possible for Leaders to take the wrong action. If it is really sooo easy to mis-promote somebody, then clearly the promotion UI is not good enough.

Also, why is it so hard for Leaders themselves to notice and fix this? What must be fixed in the database that Leaders cannot fix by themselves?

danieljames-dj:
Leader doing a wrong promotion is just an example. There are many other errors possible - like yesterday Board promoted wrong Sachin Arvind to senior delegate, which cannot be avoided so easily. The action done by board when this was realized is that they just changed to correct Sachin Arvind. Now there is a wrong row in our database because of this action and I think this can be solved only by adding WRT to role change emails for few months and maybe WRT come up with some sanity check to catch them.

gregorbg:
Okay, picking the wrong person (same name but different WCA ID I suppose) is a very interesting edge-case. I am also honestly surprised that this happened to the Board! Perhaps we should reconsider our search bar to make the information more accessible and make it much more obvious who you're actually adding, but that is clearly out of scope of this PR.

If you are certain that WRT will be included in **any and all** role change emails, can't you just hard-code WRT as a receiver in the mailer directly?
The text snippet from the `UserRoleEmailRecipient` can also be hard-coded in a similar fashion, below the "for each recipient print some text" loop. Because the text for WRT will always be the same anyways, there's no need to include it in the loop.

danieljames-dj:
Okay, I've moved the WRT contact to a method so it can be re-used. Regarding directly adding to mailer, I'm not sure how it can be done. We are already in mailer, is there anything additional I can do to make it more clean?

gregorbg:
That's a good approach, and we could merge it as-is.
You may consider taking it even one step further: In the calls to the `mail` method at the bottom of each mailer method, you set the recipients as `@to_list.map(&:email)`. You could leave this `map` as-is, but instead of adding WRT to the to-list, you just hard-code WRT as an additional recipient.
The message you're currently using could then be added directly in the `app/views/role_change_mailer` list under the current `@to_list.each` block.

Advantage: It's clearer that WRT is not just "yet another recipient" but actually a hard-coded part of the process. Feels closer to the actual business logic.
Disadvantage: You cannot do the beautiful kind of code reuse you're doing now.

I'm 50/50 myself, see whatever works best for you.

danieljames-dj:
I am more inclined the current way because I see a slight chance that WRT will get irritated if they get lots of emails. If that happens, maybe we will consider making some conditions. So, is it fine to approve as it is now please? :-)

gregorbg (approved):

PR #9573, opened 2024-06-24
Stop syncing Juniors and Trainees to reports@ by default

gregorbg:
As per Delegate Report working group request:
> Allow trainees and juniors to unsubscribe from report emails.

Software review note: The toggle-able checkbox under the personal profile page was always editable even for Trainees and Juniors, see `user.rb` method `editable_personal_preference_fields`. But no matter whether a Junior or a Trainee checked or un-checked this box, they were always hard-wired to be synced to the reports mailing list anyways.

The code snippet that did this hard-syncing is now removed, so the checkbox is actually being respected.

**IMPORTANT**: The checkbox is turned off by default, so this PR includes a migration to opt them in at first and then they can opt out to their hearts desire.


FinnIckler (approved):
LGTM

PR #9574, opened 2024-06-24
Use new working group template for Delegate Reports

gregorbg:
Template at https://docs.google.com/document/d/1k2MwIpb09UNsS-eMlZ2yRlyVqCZj2ouU5Y2OOi_e5Eo/edit

The part about adding the `summary` column to the DB and adding editing the Markdown files is hopefully straight-forward.

Actually detecting whether to _use_ the new summary field is tricky. For historical reasons, we _create_ an empty, default Delegate report the moment that the competition is **created**. So if a Delegate has already planned a competition three months ago, that competition was created in an un-announced state in the database and the report was created along with it, with the old "Venue" section pre-filled. But the competition may not even have happened yet, and the report is submitted next month when this template is already in-use.

If we use a simple `delegate_report.venue.present?` and `delegate_report.summary.present?`, then the result will be wrong because as outlined above, competitions that were created before this PR will have the `venue` DR field set to some empty default, while the `summary` field will be empty.

Perhaps we even need a migration to pre-fill the `summary` field with the default template value for all unposted reports...


FinnIckler (approved):

SAuroux:
There is a bug in there now unfortunately. I am currently writing the report for last weekend's competition and the "summary" section does not save.

gregorbg:
> There is a bug in there now unfortunately. I am currently writing the report for last weekend's competition and the "summary" section does not save.

Thanks for reporting, and sorry for the inconvenience! I have just now pushed a quickfix to production, saving works!

maxidragon:
IMO there are too much dividers now (in the old version), for example https://www.worldcubeassociation.org/competitions/FrenchChampionship2022/report

PR #9576, opened 2024-06-25
Fixed logic of at_least_senior_member?

danieljames-dj:
The logic of status rank was changed few days ago and this was a section which was missed.


gregorbg:
Looks okay, why are there no tests to catch this?

danieljames-dj:
Oh yes, thanks for catching that. Unit tests is important for this method, especially when WRT is also going to use this in one of the upcoming PR. I've done it now.

gregorbg (approved):

PR #9578, opened 2024-06-25
Optimize queries for new competitions list page

gregorbg:
The generic API endpoint was returning way too much data.

This PR introduces a separate `competitions_index` API endpoint which only returns as much data as we need. It also computes admin data only if the current user has the corresponding privileges.

Lastly, this contains a few frontend fixes to only fire loading requests to the new API when no other page request is currently loading. Before, you could trigger simultaneous duplicated requests by scrolling.

### DIFF
--- a/app/webpacker/components/CompetitionsOverview/ListView.js
+++ b/app/webpacker/components/CompetitionsOverview/ListView.js
@@ -21,12 +21,13 @@ function ListView({
   const { ref: bottomRef, inView: bottomInView } = useInView();
 
   useEffect(() => {
-    if (hasMoreCompsToLoad && bottomInView) {
+    if (hasMoreCompsToLoad && bottomInView && !isLoading) {
       fetchMoreCompetitions();
     }
   }, [
-    bottomInView,
     hasMoreCompsToLoad,
+    bottomInView,
+    isLoading,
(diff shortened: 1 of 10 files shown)


app/webpacker/components/CompetitionsOverview/ListView.js
```diff
@@ -21,12 +21,13 @@ function ListView({
...
       fetchMoreCompetitions();
     }
   }, [
-    bottomInView,
     hasMoreCompsToLoad,
+    bottomInView,
+    isLoading,
     fetchMoreCompetitions,
     // The bottom ref can still _stay_ in view even after loading new comps.
     //   In that case, the useEffect will not be triggered, so we introduce this extra dependency.
```
kr-matthews, line 33: Is `competitions` still needed here? Now `isLoading` will change and should trigger another fetch.

kr-matthews (commented):
Did not run locally, but front-end changes look reasonable.

app/webpacker/components/CompetitionsOverview/ListView.js
gregorbg: Good catch, thanks!
Fixed.

PR #9579, opened 2024-06-26
Panel cleanups

danieljames-dj:
This PR is an attempt to simplify the codes related to panels. There are definitely more scope of simplifications, but starting with this for now.

### DIFF
--- a/app/controllers/panel_controller.rb
+++ b/app/controllers/panel_controller.rb
@@ -4,20 +4,8 @@ class PanelController < ApplicationController
   include DocumentsHelper
 
   before_action :authenticate_user!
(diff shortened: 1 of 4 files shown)


gregorbg (approved):
With this change out of the way, I'd love to see a follow-up PR that removes all "single" panels from the menu navigation dropdown if you have access to the "master" panel anyways.

danieljames-dj:
> With this change out of the way, I'd love to see a follow-up PR that removes all "single" panels from the menu navigation dropdown if you have access to the "master" panel anyways.

It's already implemented in this PR. For example, you will be able to see only Leader Panel, Staff Panel, WST Panel and new Admin Panel.

PR #9581, opened 2024-06-26
Added sub-queries for WRT contact form

danieljames-dj:
For WRT contact form, instead of just sending the message, gave options like:
- Edit profile data
- Report an issue in a result
- Other query

For 2nd and 3rd option, we are still inputing only the message, but for 'edit profile data' section, we will ask the user what to change and ask to input that data. For now it's very primitive, will be improved in future after getting more feedback.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -803,7 +803,7 @@ en:
           email:
             label: "Email"
         contact_recipient:
-          label: "Whom do you want to contact?"
+          label: "Who do you want to contact?"
           competition:
             label: "Contact organizers/delegates of a competition (to ask query regarding that particular competition)"
           wct:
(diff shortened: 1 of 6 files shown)


config/locales/en.yml
```diff
@@ -803,7 +803,7 @@ en:
           email:
             label: "Email"
         contact_recipient:
-          label: "Which of the following best describes your inquiry?"
+          label: "Whom do you want to contact?"
```
gregorbg, line 806: Not an English native, but I think it should be "who" instead of "whom"

danieljames-dj: Okay changed.

gregorbg (approved):

PR #9582, opened 2024-06-26
Reduce permissions of WRT member to seniors + leader

danieljames-dj:
Currently WRT members has a huge set of access. This PR aims to restrict the access to just seniors + leader for following:
1. PMA
2. New admin panel


danieljames-dj:
I'm making this as draft because I couldn't test the Generate DB page due to some AWS issue which I've mentioned in wst-troubleshooting channel.

app/models/user.rb
```diff
@@ -493,6 +493,10 @@ def results_team?
     group_member?(UserGroup.teams_committees_group_wrt)
   end
 
+  def senior_results_team?
+    active_roles.any? { |role| role.group == UserGroup.teams_committees_group_wrt && role.metadata.at_least_senior_member? }
```
gregorbg, line 497: Can you realise this using an ActiveRecord query instead of loading the roles into memory?

gregorbg (changes requested):

app/models/user.rb

danieljames-dj:
I've undrafted it as I undo-ed the changes that moves generate DB token page's URL. So that page is still available in old admin panel, I'll get it removed in another PR.

gregorbg (approved):

PR #9591, opened 2024-06-28
Make panel creation easier (POC)

danieljames-dj:
Currently creating a new panel is bit difficult and not so clear. This PR will make the process easier, for example, if a WCT panel has to be created, we just need to add an entry in PanelHelper.

This is a POC, and some panels are not yet deleted from old format - only admin panel & WRT panel are done. If these changes are fine, I'll delete the remaining in a follow-up (single) PR.


app/helpers/panel_helper.rb
```diff
@@ -1,68 +1,60 @@
...
         name: 'New Admin panel',
-        url: Rails.application.routes.url_helpers.panel_admin_path,
+        pages: panel_pages.values,
       },
-      {
-        id: :staff,
+      staff: {
         name: 'Staff panel',
-        url: Rails.application.routes.url_helpers.panel_staff_path,
+        pages: [],
```
danieljames-dj, line 13: The pages are empty just because they are not moved to this format completely. This will work even with empty pages list because these will rely on panel#staff, and not panel#index.

gregorbg (approved):

PR #9616, opened 2024-07-02
Give edit permission to banned competitors for WEC Leader

danieljames-dj:
This was requested by the WDC Leader over chat.


dunkOnIT:
Can we put this request into an email, just so there is a paper trail? Simple "yes" from Shain in email will suffice imo.

dunkOnIT (approved):
Code looks good, tested locally and works as expected

Issue #8419, opened 2023-10-13
Enforce single-quotes in Rubocop


dunkOnIT:
If anyone else wants to pick this up, it requires the following in `.rubocop.yml`:
```yml
Style/StringLiterals:
  Enabled: true
```

And then running `rubocop -a` to deal with any instances where that rule hasn't been enforced.

I'll try do it later today/on Monday

misaka:
PR created as above, it seems like there isn't enough appetite for this, at the moment at least. Happy to contribute further if this changes.

dunkOnIT:
Yeah unfortunately the risk/reward of trying to go about these changes it too high for the team's capacity at the moment. 

If you're interested in contributing on other issues, let me know and I'd be happy to point out some other good first issues!

dunkOnIT:
We are likely too far down the double quotes road to consider this anytime soon

PR #9657, opened 2024-07-07
Enable rubocop Style/StringLiterals cop

misaka:
As mentioned in #8419, this PR enables the `Style/StringLiterals` cop and corrects all the related strings. That issue is old, however, and I don't know if it wasn't done before because opinion shifted on it, or whether it just wasn't a priority. If there's a reason not to do this maybe the issue should be closed down.

There's also the commented out config to enable double-quotes. Is that a suggestion that that's the preferred style (it is for me, for what little it's worth)? Happy to amend / remove as preferred.


misaka:
I've resolved conflicts on this PR. Since this PR touches so many files it's going to go stale quite quickly, if this isn't something the maintainers want addressed now it'll probably be better to close this PR for now instead of updating it with every merged PR.

gregorbg:
Hmm... I like code consistency, but it's incredibly easy for unintended changes to "slip in" with a PR this huge...

And it's not even about trusting you as a contributor. Rubocop screwing up the automated fix routine would already be enough.
I know we have a test suite but I still feel quite ambivalent about this, especially since there is no real benefit... (If double-quote strings used significantly more memory, then maybe that's a different story... But they don't.)

What do others think?

misaka:
Thanks for the response @gregorbg. I understand your reluctance, and don't take it personally at all.

If you and others think there's enough value in enabling this cop, i could take the approach of cutting off small sections and PRing smaller bits of the codebase at a time. It might be easier to review in smaller chunks and we could wait after deploying each one to try to limit the impact if it goes wrong. I don't know what tests you have beyond the rspecs, but I suppose testing with real-world-like data would be useful too. I know this work is probably not great use of time for the core contributors, such as yourself, but it would give me a way to get a bit more familiar with the codebase, and would be willing to do it so I can contribute more over time.

Of course, if the general consensus is that it's just not worth the time / risk, I'll close the PR and try to find somewhere else I can contribute.

gregorbg:
I'll leave the debate open for others to chime in. As far as my personal opinion is concerned, distinguishing between single-quote and double-quote strings doesn't help a great deal in understanding the codebase. A string is still a string. It's the code around the strings that matters.

With that being said, if there is a strong consensus towards introducing this cop (which I would in principle be fine with) then we need to go down the step-by-step road to make the individual PRs more accessible and easily reviewable.

misaka:
I'm going to close this PR down, it seems like there's not enough appetite for this.

PR #9658, opened 2024-07-08
Show local date/time in utc date picker

kr-matthews:
When time is input is active.

Input without time (no change):
image

Input with time:
image

image

### DIFF
--- a/app/webpacker/components/wca/FormBuilder/input/FormInputs.js
+++ b/app/webpacker/components/wca/FormBuilder/input/FormInputs.js
@@ -241,11 +241,11 @@ export const InputDate = wrapInput((props) => {
           isoMaxDate={props.maxDate}
         />
       </Input>
-      {props.dateTime && (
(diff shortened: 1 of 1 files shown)


app/webpacker/components/wca/FormBuilder/input/FormInputs.js
```diff
@@ -221,25 +221,37 @@ export const InputDate = wrapInput((props) => {
...
+      {props.dateTime && (
+        <p className="help-block">
+          In your current time zone:
+          {' '}
+          {props.value && new Date(props.value).toLocaleString(undefined, {
+            dateStyle: 'full',
+            timeStyle: 'short',
+          })}
+        </p>
+      )}
```
gregorbg, line 253: IMO, the whole `p` block only makes sense to show if there is a `props.value` present.

gregorbg (approved):

PR #9659, opened 2024-07-09
Don't show Payment Panel if registration is succeeded

FinnIckler:
Allows organizers to waive payment by accepting the competitor regardless. 
We should probably still add a view registration button on the payment step and a pay button on the registration step?


gregorbg (approved):

FinnIckler:
I added the registration overview under the payment panel for users who want to update their registration even though they haven't paid yet. This does not show when prompted to pay the first time so it shouldn't cause any confusion.

FinnIckler:
Scratch that. After having a call with Duncan, we have decided that instead. You can just click back on Register to update if you haven't paid. This approach also allows people to pay even after they have been accepted (if they decide not to staff for example)

gregorbg (approved):

PR #9661, opened 2024-07-10
Increased coverage for user_roles_controller

danieljames-dj:
This PR aims at increasing the coverage of user_roles_controller. This increased the coverage from 43.6% to 50% (not 100 yet, will achieve that later)

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -506,7 +506,7 @@ class User < ApplicationRecord
   end
 
   def admin?
-    Rails.env.production? && EnvConfig.WCA_LIVE_SITE? ? software_team_admin? : software_team?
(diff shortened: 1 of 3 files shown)


app/models/user.rb
```diff
@@ -506,7 +506,7 @@ def staff?
   end
 
   def admin?
-    Rails.env.production? && EnvConfig.WCA_LIVE_SITE? ? software_team_admin? : software_team?
+    Rails.env.production? && EnvConfig.WCA_LIVE_SITE? ? software_team_admin? : (software_team? || software_team_admin?)
```
danieljames-dj, line 509: This is done because wst_admin_role is not considered as a subset of wst_role, and technically, one person can be part of WST admin and not WST.

gregorbg (approved):

PR #9662, opened 2024-07-10
Allow to enter DOB in Finish unfinished persons

danieljames-dj:
Currently, if we create a newcomer using "Finish unfinished persons" script, and if there is no DOB (for example, if the person is created from the existing results instead of InboxPersons), creation of person fails due to empty DOB.

This change will allow to enter DOB manually, so that we can manually enter DOB in case DOB is not available.


danieljames-dj:
This is no longer relevant, there was another procedure which can be followed to split profiles.

PR #9667, opened 2024-07-10
Migrate to Puma webserver

gregorbg:
Attempt to move our application server to what the cool kids seem to be using these days.

Jokes aside, Unicorn seems unmaintained (last release 2021) and recently had some critical issue with the login when trying to migrate to Rack 3.

As of opening this PR, I didn't run the thread-safety Rubocops yet. Will rely on GitHub's CI to do that for me.


FinnIckler:
The Docker Compose file needs to be changed to
```
    command: >
      bash -c 'bundle install && corepack install && yarn install &&
      rm -f .db-inited &&
      if ! [[ "`mysqlshow --user=root --host=wca_db wca_development`" =~ "Tables" ]] ;
        then echo "Populating development db, this will take a while" && bin/rake db:reset ;
      fi &&
      touch .db-inited &&
      ./bin/bundle exec puma'
```

FinnIckler:
Actually.. rails just does bin/rails server as puma is the default
https://github.com/rails/rails/blob/ef087627f53fac581e6120332c61e96d5e68628c/railties/lib/rails/generators/rails/app/templates/docker-entrypoint.tt#L4
https://github.com/rails/rails/blob/ef087627f53fac581e6120332c61e96d5e68628c/railties/lib/rails/generators/rails/app/templates/Dockerfile.tt#L4

gregorbg:
Thread-safety Rubocops have been taken care of. For the most part, I moved class instance variables that were used as "simplified caches" to the actual Rails cache.

Especially stuff like countries, competition events (333, 444, skewb, etc.) were cached so that we don't fire a humongous amount of `SELECT * FROM Country` calls. These calls are now fetched and inserted into our Rails cache backed by Redis, instead of being stored in a runtime class variable.

Other occurences have been optimised as I saw fit. Country validators were cached previously, but I don't understand why because they're cheap to compute. So now I instantiate them on-the-fly.

FinnIckler:
Did you check performance impact? I remember I tried to do the change to the cachable concern a couple years ago when we introduced Redis, but stuff like WCIF was 10x slower. Definitely was user a cruder approach then you though.

gregorbg:
Right now I want to get tests to pass, so we can do smoke tests. Then performance tests are next.

Depending on your testing back in the day, you may have simply fallen into the trap that in local development, Rails doesn't cache by default. So even if you code a correct `Rails.cache` replacement, it still fires a gazillion `SELECT * FROM Country` calls in local testing, thereby slowing down WCIF.

gregorbg:
@thewca-bot deploy staging

gregorbg:
@thewca-bot deploy staging

gregorbg:
@thewca-bot deploy staging

FinnIckler (approved):
LGTM and staging looks good

PR #9668, opened 2024-07-10
Go back to Rack 3 and fix underlying array issue

gregorbg:
This reverts a previous revert. We actually _want_ Rack 3, we just had to temporarily go back to Rack 2 as a hotfix.

Seems that different browsers interpret the "hidden `false` field" (https://rubydoc.info/gems/actionview/7.1.3.4/ActionView/Helpers/FormBuilder#check_box-instance_method) from Rails form checkboxes differently, and in Rack 3 that difference finally causes a crash.


FinnIckler:
doesn't fix the issue. Looking at the logs, this issue occurs on the redirect, not on the sign in request so I don't think this would cause it

FinnIckler:
Issue is actually that Unicorn does not support Rack 3

PR #9671, opened 2024-07-15
Few more panel cleanups

danieljames-dj:
This PR aims at following cleanups for panels:
1. Removal of `can_access_at_least_one_panel?` from user.rb as it is not necessary.
2. Introduces panel redirect, which will accept a panel page with URL format "/panel/redirect/panel_page" and gets redirected to one of the panels that belong to the user. This will be helpful for adding the URL in pages like 'active roles'.

### DIFF
--- a/app/controllers/panel_controller.rb
+++ b/app/controllers/panel_controller.rb
@@ -46,4 +46,12 @@ class PanelController < ApplicationController
       replica: 2,
     }
   end
+
+  def redirect
+    @panel_page = params.require(:panel_page)
+    panel_with_panel_page = current_user&.panels_with_access&.find { |panel| User.panel_list[panel][:pages].include?(@panel_page) }
+
+    return head :unauthorized if panel_with_panel_page.nil?
+    redirect_to panel_index_path(panel_id: panel_with_panel_page, anchor: @panel_page)
+  end
 end
--- a/app/controllers/panel_controller.rb
+++ b/app/controllers/panel_controller.rb
@@ -49,7 +49,7 @@ class PanelController < ApplicationController
 
   def redirect
     @panel_page = params.require(:panel_page)
-    panel_with_panel_page = current_user&.panels_with_access&.find { |panel| User.panel_list[panel][:pages].include?(@panel_page) }
+    panel_with_panel_page = current_user.panels_with_access&.find { |panel| User.panel_list[panel][:pages].include?(@panel_page) }
 
     return head :unauthorized if panel_with_panel_page.nil?
     redirect_to panel_index_path(panel_id: panel_with_panel_page, anchor: @panel_page)
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1423,10 +1423,6 @@ class User < ApplicationRecord
     end
   end
 
-  def can_access_at_least_one_panel?
-    panels_with_access.any?
-  end
-
   def subordinate_delegates
     delegate_roles
       .filter { |role| role.is_lead? }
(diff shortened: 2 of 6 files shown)


app/models/user.rb
```diff
@@ -671,7 +667,7 @@ def permissions
         scope: can_access_wfc_senior_matters? ? "*" : [],
       },
       can_access_panels: {
-        scope: panels_with_access,
+        scope: panel_list(self).keys,
```
gregorbg, line 670: I suppose this `panel_list(self)` comes out of `app/helpers/panel_helper`.
In that case, please move the method somewhere else (for example, as a static/class method directly in `user.rb` because in Rails conventions, `app/helper` is only meant for **UI** helper functions.

danieljames-dj:
I'm marking this as draft for now, will come back once https://github.com/thewca/worldcubeassociation.org/pull/9672 is merged.

app/models/user.rb
danieljames-dj: Reverted this change as this is outdated

danieljames-dj:
This is ready for review now.

app/controllers/panel_controller.rb
```diff
@@ -46,4 +46,12 @@ def generate_db_token
       replica: 2,
     }
   end
+
+  def redirect
+    @panel_page = params.require(:panel_page)
+    panel_with_panel_page = current_user&.panels_with_access&.find { |panel| User.panel_list[panel][:pages].include?(@panel_page) }
```
gregorbg, line 52: Why are you using nil-safe access here? You have `before_action :authenticate_user` in this controller already, so I don't see how `current_user` could ever be nil.

app/webpacker/components/RolesTab/ActiveRoles.jsx
gregorbg, line 1: What is the advantage of using redirects here? In particular, why is the thing that you were doing before (directly using `panelUrls.foo.bar` not good enough?

gregorbg: Your description from when you raised the PR explains _what_ you're trying to do (I understand that) but I don't see _why_ it is necessary to do in the way you're doing (i.e. why direct linking doesn't work and this redirect approach is used instead)

app/controllers/panel_controller.rb
danieljames-dj: Oh yes, I've removed it now. Thanks for pointing it out.

app/webpacker/components/RolesTab/ActiveRoles.jsx
danieljames-dj: The main reason was that there can be cases (if not now, maybe in future) where the same page might be available in some other panel - like Officers panel or WIC panel, etc. So, if we are not forcing the user to go to a particular panel and instead just ask them to go to a panel page, the chances of failures might be less.

Or another use-case I can think of now is that admin team or WRT senior member clicks the 'edit' button, it will try to go to a panel where they are not part of - like Board panel or Leader panel, etc. But with this feature, they can go to that page using 'new admin panel'.

gregorbg (approved):
Okay for the time being.

In the medium-term future I'd like to see a solution where each panel page is deterministically linked to one panel, and either you have access or you don't.
This notion of "the same panel might be accessible or unaccessible, depending on which route you're accessing it from" has the potential for a lot of chaos if we keep it in our codebase forever.

danieljames-dj:
Thanks @gregorbg for approving. Sure, I'll think about this solution.

PR #9673, opened 2024-07-15
Load MS registration data upon #find call

gregorbg:
The mailers are serializing using Global ID mechanism, which means that `deliver_later` doesn't serialize the actual, nice, enriched MS registration. Rather, it just stores a reference to the row using the primary key, and loads it again later.

That means in order to send mails, we automatically load/enrich MS data when using `.find` directly.

NOTE: This does not affect the `has_many` bulk loading because that one doesn't go through `find` directly. And we have an `unless` guard clause for models that are already loaded :upside_down_face:


FinnIckler (approved):
LGTM

PR #9674, opened 2024-07-15
Export model data to JS instead of letting the JS compilation query the database

gregorbg:
Still not getting rid of `erb-loader` entirely, but at least we don't have to query the database during compilation now.

Only thing that's left now is the Recaptcha key, which is currently stored in Vault, so the asset pipeline in prod would need to access Vault. I guess we can just hard-code that one since it's a **public** key anyways?


dunkOnIT (dismissed):
By my own hand I undo what I have wrought

app/models/competition.rb
dunkOnIT, line 1: Reviewed the changes to how we check the lengths of strings vs the varchar limit, happy with it

lib/static_data/round_types.json
```diff
@@ -0,0 +1,79 @@
...
+    "id": "f",
+    "rank": 99,
+    "name": "Final",
+    "cellName": "Final",
+    "final": true
+  },
+  {
+    "id": "g",
+    "rank": 70,
+    "name": "Semi Final",
```
dunkOnIT, line 68: In the database this is called `Third round` instead of "Semi Final"

lib/static_data/preferred_formats.json
```diff
@@ -0,0 +1,152 @@
...
+    "format_id": "2",
+    "ranking": 2
+  },
+  {
+    "event_id": "333mbf",
+    "format_id": "3",
+    "ranking": 3
+  },
+  {
+    "event_id": "333ft",
```
dunkOnIT, line 108: feet, magic and mbo don't appear in the prod database - why are they here?

lib/static_data/events.json
```diff
@@ -0,0 +1,326 @@
+[
+  {
+    "id": "333",
```
dunkOnIT, line 3: prod database has a `cellName` column, but the JSoN does not - a deliberate exclusion?

lib/static_data/countries.real.json
```diff
@@ -0,0 +1 @@
+../../config/wca-states.json
```
dunkOnIT, line 1: The key in the json is `continent_id`, but in the db it's the `continentId` - will this cause any issues with key access?

lib/static_data/championship_eligible_iso2.json
```diff
@@ -0,0 +1,8 @@
+{
```
dunkOnIT, line 1: I don't see this anywhere in the db? But it looks simple enough that I trust it

gregorbg: The table name in the DB is much longer. But I am confident that the data is correct.

gregorbg: For reference, the table is called `eligible_country_iso2s_for_championship`

lib/static_data/countries.real.json
gregorbg: No, because this symlink points to a file maintained by WRC which we parse manually

lib/static_data/events.json
gregorbg: Yes, `cellName` is being dropped in a migration that is part of this PR

lib/static_data/round_types.json
gregorbg: After clarification with WRT, we decided that "Semi Final" makes more sense.

gregorbg:
@thewca-bot deploy staging

PR #9676, opened 2024-07-16
Remove teams from default serialization

danieljames-dj:
Having the expensive `teams` in default user serialization doesn't looks good so making it optional and add only in publicly exposed APIs.

### DIFF
--- a/app/controllers/api/v0/users_controller.rb
+++ b/app/controllers/api/v0/users_controller.rb
@@ -17,7 +17,7 @@ class Api::V0::UsersController < Api::V0::ApiController
   def show_users_by_id
     user_ids = params.require(:ids)
     users = User.where(id: user_ids)
-    render status: :ok, json: { users: users }
+    render status: :ok, json: { users: users.map { |user| user.serializable_hash(include: ["avatar", "teams"]) } }
   end
 
   def show_user_by_wca_id
@@ -64,7 +64,7 @@ class Api::V0::UsersController < Api::V0::ApiController
 
     def show_user(user, show_rankings: false, private_attributes: [])
       if user
-        json = { user: user.serializable_hash(private_attributes: private_attributes) }
+        json = { user: user.serializable_hash(private_attributes: private_attributes, include: ["avatar", "teams"]) }
         if params[:upcoming_competitions]
           json[:upcoming_competitions] = user.accepted_competitions.select(&:upcoming?)
         end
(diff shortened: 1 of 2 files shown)


gregorbg (changes requested):
1. Rails has its own implementation of `deep_merge` at https://apidock.com/rails/Hash/deep_merge. Your monkey-patch is not necessary.
2. Even if you used the "official" implementation from point (1), the behavior still changes quite significantly. The current `DEFAULT_SERIALIZE_OPTIONS` is a list of **defaults**, i.e. things that are included if nothing else is specified. If you specify your own stuff, it _overrides_ the defaults: "Hey codebase, I know you suggest to include `avatar` and `teams`, but actually in my use-case I only need to include `person` and `some_other_amazing_join_table`. `avatar` and `teams` are both not needed in my special corner case, thank you!". By changing to `deep_merge`, you make it so that the default turns into a **forced** baseline of things that are included, i.e. no matter what you specify in your custom use case serialization, the `avatar` part will **always** be merged on top of it. This is not what we want because it wastes resources in cases where you don't want/need the avatar to be included. In other words, the behavior as it was before (with shallow `merge` instead of `deep_merge`) is _intended_.
3. Have you checked to make sure that no other React components that use internal serialization are breaking? This is not only about public APIs (which you said in your PR description you will adjust later) but also about internal serialization. If we pass something like `react_component("TeamBadges", { user: @user })` to the React helper, the `@user` variable will implicitly be `to_json`ed and if we lose the `teams`, we lose the functionality of the component. It might be tricky to find all places in the code where this breaks.
4. Even if you decide to include `teams` in all places where it's required (fixing point (3) very diligently) the actual implementation still runs through a method called `deprecated_team_roles`. When are you going to clean that up?

danieljames-dj:
As per our discussion in slack, I'll remove the new implementation of merge.

Also, the reason why I'm having the 'teams' getting included in external APIs is that I've verified that it is not used anywhere else. To confirm this, I searched the word 'teams' in our `app/*` folder by excluding `app/webpacker/rails_translations/*, *.css` - got around 206 results and manually went through all of them and confirmed the teams object is not used anywhere.

Coming to your remark #4, I didn't understand what you meant by "the actual implementation still runs through a method called deprecated_team_roles". The method will be called only during the external facing APIs and not anywhere else. Is there anything I'm missing in understanding that?

gregorbg:
You're not missing anything substantial. The method will be called only during the external facing APIs.
But the fact that a method called `deprecated_*` is being called *at all* is what concerns me.

danieljames-dj:
Closing this as the plan has been changed. Instead of fixing this, will be deprecating teams after a public announcement.

PR #9678, opened 2024-07-16
Add competition_id index for registrations and registration_id for registration payments

FinnIckler:
While I was checking the performance of Puma in New Relic, I noticed that Registration.find, especially in conjunction with `RegistrationPayment` has become our slowest queries (16% of all database time). 

Every time we call `Competition.find(...).registrations` we need to scan the whole table.
Similarly with calls to  `pending.with_payments` for RegistrationPayment
Here is an example trace from new relic on a call to `registration_full?`:
```
 No index was used in this part of the query.
    This table was retrieved with a full table scan, which is often quite bad for performance, unless you only retrieve a few rows.
    You can speed up this query by querying only fields that are within the index. Or you can create an index that includes every field in your query, including the primary key.
    Approximately 427354 rows of this table were scanned.
```
Should be a quick little performance win!

### DIFF
--- a/db/migrate/20240716145605_add_more_indexes_to_registrations.rb
+++ b/db/migrate/20240716145605_add_more_indexes_to_registrations.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddMoreIndexesToRegistrations < ActiveRecord::Migration[7.1]
+  def change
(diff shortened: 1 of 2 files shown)


FinnIckler:
image

gregorbg (approved):

PR #9679, opened 2024-07-16
Give more clear error messages when DOBs cannot be parsed

gregorbg:
Previously, not being able to parse a DOB would result in `nil` values being passed to the `User` constructor, which makes the validation complain with "DOB is required".

With this PR, the import endpoint gives a more informative message and shows _which_ DOBs cannot be parsed.


FinnIckler:
seems you got the Puma PR in there ;)

gregorbg:
> seems you got the Puma PR in there ;)

Thanks, fixed!

gregorbg:
Merging as hotfix

PR #9683, opened 2024-07-17
Use Puma as development & testing server

gregorbg:
I call it "Puma light"

The update to the configuration file comes from the in-development branch of Rails 7.2: https://github.com/rails/rails/blob/v7.2.0.beta3/railties/lib/rails/generators/rails/app/templates/config/puma.rb.tt

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -117,6 +117,8 @@ group :development, :test do
   gem 'faker'
   gem 'capybara-screenshot'
 
+  gem 'puma'
+
   gem 'byebug'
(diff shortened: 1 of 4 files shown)


gregorbg:
In the diff, note that the `gem 'puma'` is only included in development and test. The "Files changed" tab on GitHub cuts off the top & bottom so you can't actually see that it's installed under `group :development, :test do`

FinnIckler (approved):
Sounds like another step in the right direction

PR #9685, opened 2024-07-19
Serve React ReCaptcha secret through backend

gregorbg:
The goal here is to be able to compile frontend assets in our CI, which doesn't have access to our Vault credentials storage.
No change in functionality.


FinnIckler (approved):
LGTM

PR #9686, opened 2024-07-19
Attempt to simplify/shorten GitHub CI commands

gregorbg:
The existing state of the CI was basically a remnant from our Travis migration.

Now that we have a much better idea of our own code organization and now that GitHub Actions has matured, I think we can clean this up a little bit.

### DIFF
--- a/app/models/result_validation_form.rb
+++ b/app/models/result_validation_form.rb
@@ -9,7 +9,6 @@ class ResultValidationForm
 
   COMP_VALIDATION_MODES = [["Pick competition(s) manually", COMP_VALIDATION_MANUAL], ["Execute for ALL competitions", COMP_VALIDATION_ALL]].freeze
 
-  ALL_COMPETITIONS_SCOPE = Competition.over.not_cancelled
   ALL_COMPETITIONS_MAX = 500
 
   include ActiveModel::Model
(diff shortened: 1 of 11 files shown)


FinnIckler:
If you want to work on this without just waiting for the CI to see if it works, I recommend act https://github.com/nektos/act

gregorbg:
Right now all the issues that are ocming up are issues that are failing fast (within ~1 minute) so I'm fine.
But thanks for the recommendation, looks interesting!

gregorbg:
Also, depends on changes made in https://github.com/thewca/worldcubeassociation.org/pull/9674

FinnIckler (approved):
LGTM

app/models/result_validation_form.rb
```diff
@@ -9,7 +9,6 @@ class ResultValidationForm
 
   COMP_VALIDATION_MODES = [["Pick competition(s) manually", COMP_VALIDATION_MANUAL], ["Execute for ALL competitions", COMP_VALIDATION_ALL]].freeze
 
-  ALL_COMPETITIONS_SCOPE = Competition.over.not_cancelled
```
FinnIckler, line 12: How did this even work. Wouldn't that variable only contain Competitions that were over/not_cancelled on boot?

gregorbg: No, actually Rails scopes are lazyily.
The thing that was failing was not the loading of actual data but the basic model validations like "which columns exist" etc. Rails wants to execute those as soon as you speak the name of any model, even if you're not fetching data.

PR #9688, opened 2024-07-21
Serialize just author name in posts

danieljames-dj:
For the posts, only author name is used from the author object, hence avoiding serializing author completely.

### DIFF
--- a/app/webpacker/components/PostsWidget.js
+++ b/app/webpacker/components/PostsWidget.js
@@ -62,7 +62,7 @@ function PostsList({
             <Card.Meta>
               Posted by
               {' '}
-              {post.author_name ? post.author_name : 'Unknown'}
+              {post.author_name ?? 'Unknown'}
               {' '}
               on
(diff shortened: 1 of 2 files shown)


app/webpacker/components/PostsWidget.js
```diff
@@ -62,7 +62,7 @@ function PostsList({
             <Card.Meta>
               Posted by
               {' '}
-              {post.author ? post.author.name : 'Unknown'}
+              {post.author_name ? post.author_name : 'Unknown'}
```
gregorbg, line 65: I think this can be written as `post.author_name ?? 'Unknown'`

gregorbg (changes requested):

app/webpacker/components/PostsWidget.js
danieljames-dj: Done, thanks

gregorbg (approved):

PR #9689, opened 2024-07-21
Move panel_list to user.rb

danieljames-dj:
panel_list will be used in model and controller, and not just UI, so moving it to user model.


app/models/user.rb
```diff
@@ -647,8 +646,97 @@ def banned_at_date?(date)
     groups
   end
 
+  def self.panel_list
+    panel_pages = PanelController.panel_pages
```
gregorbg, line 650: Rails has a whole host of methods related to converting strings between camel case / kebap case / snake case / underscore case etc.

I feel like the only purpose of `PanelController.panel_pages` is to store the association between some string keys and the same strings in kebap-case format again as values. So maybe this PR would be a good opportunity to clean that up?

danieljames-dj: Actually not just that. panel_pages is used in many places. The panel_pages is the source of truth for the list of panel pages. This list is used in PanelPages.jsx to map it to the title and the react component/hyperlink. In this case, panel_pages is used to refer to the panel pages in ease. Hope this clarifies.

danieljames-dj: I've made the changes as you suggested.

app/models/user.rb
```diff
@@ -647,8 +646,128 @@ def banned_at_date?(date)
...
+      :importantLinks,
+      :delegateHandbook,
+      :seniorDelegatesList,
+      :leadersAdmin,
+      :boardEditor,
+      :officersEditor,
+      :regionsAdmin,
+      :downloadVoters,
+      :generateDbToken,
+    ].to_h { |panel_page| [panel_page, panel_page.to_s.underscore.dasherize] }
```
gregorbg, line 677: Instead of this manual `to_h` approach, you can also use `index_with` or `index_by`. Both take a lambda function with _one_ parameter, and the difference is that:
- `index_with` turns an array into the hash using the original value as the key and the result of the lambda function as the new value
- and `index_by` uses the original value as the hash value and the result of the lambda function as the new key

https://apidock.com/rails/Enumerable/index_with
https://apidock.com/rails/Enumerable/index_by

gregorbg (changes requested):

app/models/user.rb
danieljames-dj: Thanks a lot, done.

gregorbg (approved):

PR #9694, opened 2024-07-21
Introduce regional mailing lists for Delegate Reports

gregorbg:
The use-case is that users can filter which region (country or continent) they want to receive Delegate Reports for.

Currently, this is implemented via a dropdown in the "Edit Profile" page where you also set the checkbox whether you want to receive reports or not.

The idea behind the mailing lists is the opposite of Delegate Region mailing lists: For the Delegate Regions, you want the child regions to be a member of their parent, so that if you mail a parent region, all child regions also automatically receive the email.

For this feature, we create mailing lists for the countries first, and then we add the continent (their "parent") to the "child" i.e. country mailing list. This makes it so that someone who has subscribed to US reports does **not** get reports when something to the North-America list is posted. But when someone posts a report to the US list, it is also automatically received by the North America list.

Test cases are still pending.


app/jobs/sync_mailing_lists_job.rb
```diff
@@ -11,8 +11,25 @@ def perform
...
+    GsuiteMailingLists.sync_group("reports@worldcubeassociation.org", report_users | User.default_report_receivers)
+
+    Continent.real.each do |continent|
+      continent_mailing = DelegateReport.continent_mailing_list(continent)
+      report_users = User.where(receive_delegate_reports: true, delegate_reports_region: continent).pluck(:email)
+
+      GsuiteMailingLists.sync_group(continent_mailing, report_users | ["reports@worldcubeassociation.org"])
+
+      continent.countries.real.each do |country|
+        country_mailing = DelegateReport.country_mailing_list(country, continent)
```
FinnIckler, line 27: make clear that this is a single address

spec/jobs/sync_mailing_lists_job_spec.rb
```diff
@@ -104,14 +108,27 @@
     expect(GsuiteMailingLists).to receive(:sync_group).with(
       "reports@worldcubeassociation.org",
       a_collection_containing_exactly("seniors@worldcubeassociation.org", "quality@worldcubeassociation.org", "regulations@worldcubeassociation.org",
```
FinnIckler, line 110: tests for Countries (multiple) and Continents (multiple) competitions

FinnIckler (approved):
Just needs a couple more tests and a few clearer names

PR #9697, opened 2024-07-22
Added validation to prevent making group inactive if the group has active roles

danieljames-dj:
This PR is to add validation to prevent making a group inactive if the group has at least one active role.

### DIFF
--- a/app/webpacker/components/Panel/pages/RegionManager/index.jsx
+++ b/app/webpacker/components/Panel/pages/RegionManager/index.jsx
@@ -80,7 +80,7 @@ export default function RegionManager() {
   }, [data]);
 
   if (loading || fetchLoading || saving) return <Loading />;
-  if (error || saveError) return <Errored />;
+  if (error || saveError) return <Errored error={error || saveError} />;
 
   return (
     <>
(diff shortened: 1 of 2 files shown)


app/webpacker/components/Panel/pages/RegionManager/index.jsx
```diff
@@ -22,7 +22,9 @@ const defaultRegion = {
   friendlyId: '',
 };
 
-function UserGroupVisibility({ userGroup, save, sync }) {
+function UserGroupVisibility({
+  userGroup, save, sync, setSaveError,
```
gregorbg, line 26: I don't see any code change where components pass `setSaveError`...
Where are you passing any value along?

danieljames-dj: It's passed in line 134 and 169, but I guess you already saw that, did you mean something else?

gregorbg: Yeah but where is that `setSaveError` that is being passed in those lines actually defined?
Was it already existing before?

danieljames-dj: Yes, it was already existing earlier. But it was not passing the error to the `Errored` component. I've done it now. Still it's not displaying a proper message, but I think that is fine.

gregorbg (approved):

PR #9698, opened 2024-07-22
Add proof attachment field for contact form

danieljames-dj:
In this PR, allowed the user to attach proof for changing profile data. But I couldn't test if the attachment is coming in email as the MailCatcher is not working. This can probably tested only in production. I can see the email in logs, so pretty sure this will not break anything, the worst case will be that the mail will come but without any attachment.

### DIFF
--- a/app/controllers/contacts_controller.rb
+++ b/app/controllers/contacts_controller.rb
@@ -44,7 +44,7 @@ class ContactsController < ApplicationController
     end
   end
 
-  private def contact_wrt(requestor_details, contact_params)
+  private def contact_wrt(requestor_details, contact_params, attachments)
     profile_data_to_change = contact_params[:profileDataToChange]
     maybe_send_contact_email(
       ContactWrt.new(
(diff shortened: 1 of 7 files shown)


app/controllers/contacts_controller.rb
```diff
@@ -55,6 +55,7 @@ class ContactsController < ApplicationController
         new_profile_data: new_profile_data_key_to_value(contact_params[:newProfileData], profile_data_to_change),
         edit_profile_reason: contact_params[:editProfileReason],
         message: contact_params[:message],
+        document: attachments,
```
gregorbg, line 58: You're passing in a plural "attachments" to a singular "document" field -- that is confusing me. Does `attachments` actually contain a list of multiple values?
If Yes, how can it be safely passed to `document` that seems to only contain one value?
If No, why is it called `attachments` instead of `attachment`?

gregorbg (changes requested):
Code changes seem fine but I'm unhappy (or at least confused) about the naming inconsistency between `document`, `attachment` (singular) and `attachments` (plural)

app/controllers/contacts_controller.rb
danieljames-dj: Sorry for the confusion, this only supports only one attachment for now. I have renamed it to `attachment` now.

gregorbg (approved):

PR #9701, opened 2024-07-22
Compile frontend assets during Docker build

gregorbg:
:tada: 

The pervious Dockerfile and Dockerfile.sidekiq have been merged into one. The build actions target different steps from the build process, one for sidekiq and one for the monolith.

- I had no clue what `zip` and `python-is-python3` in the Sidekiq build script were needed for. I copied them both over.
- The Sidekiq container was pointing to our forked, monkey-patched `mailcatcher` for Ruby 3.3 support. In the meantime, a new version with official 3.3 support was released so just pointing back to the original instead
- Fonts are just for the mononlith to display PDFs properly. Sidekiq doesn't need to display any fancy fonts.
- Debian 12 now hosts MariaDB 10.11.6, so the custom install repo patch is not necessary anymore: https://packages.debian.org/bookworm/mariadb-client


FinnIckler:
I've stopped `EnvConfig` from throwing errors for compilation, BUT it still tries to access vault for `AppSecrets` and I don't think setting `throw_exceptions` to false will stop that

FinnIckler:
One solution would be to to also check for ASSET_COMPILATION in app secrets when doing if `Rails.env.production?`. But maybe there is a better way. I can do some research tomorrow morning

gregorbg:
Huh, interesting. I was under the impression that Vault is safe because the asset compilation code doesn't directly access any secrets, but I guess the Rails environment that needs to be set up in general does access secrets. I didn't fully think that through 😅

Maybe the Vault SDK has a stub_responses dummy? Otherwise you could just reroute the "if production the Vault else dotfile" check to an "if production and not compiling_assets" check like you suggested.

FinnIckler:
You would think that. But the issue is that the `require_relative "vault_config"` in the if block immediately tries to connect to vault

FinnIckler:
argh, it uses the old workflow file for some reason

FinnIckler:
With the latest changes it now compiles both assets and exports i18n files.
It currently fails at
```
RUN useradd rails --create-home --shell /bin/bash && \
    chown -R rails:rails vendor db log tmp public app pids
```
with the error 
```
0.116 chown: cannot access 'vendor': No such file or directory
0.116 chown: cannot access 'db': No such file or directory
0.116 chown: cannot access 'log': No such file or directory
0.119 chown: cannot access 'public': No such file or directory
0.119 chown: cannot access 'app': No such file or directory
0.119 chown: cannot access 'pids': No such file or directory
0.119 chown: cannot access '.yarn': No such file or directory
```

So I think we need to look into the Dockerfile to see what actually happens at `COPY --from=build . .`

FinnIckler:
works now! At least the compile process. Yet to be tested on staging

Dockerfile
```diff
@@ -1,5 +1,8 @@
-FROM ruby:3.3.0
+FROM ruby:3.3.0 AS base
 ARG BUILD_TAG=local
+ARG WCA_REGISTRATIONS_URL
+ARG WCA_REGISTRATIONS_POLL_URL
```
gregorbg, line 4: Why do you specifically need these two?
`ROOT_URL` kinda makes sense, but why the microservice URLs and not other env variables?

app/models/regulation.rb
```diff
@@ -23,7 +23,7 @@ def self.reset_regulations
     self.regulations_load_error = nil
   end
 
-  if Rails.env.production?
+  if Rails.env.production? && !ENV.fetch("ASSETS_COMPILATION", false)
```
gregorbg, line 26: Outsource to `EnvConfig`

app_secrets.rb
```diff
@@ -45,8 +45,8 @@ def vault_file(secret_name, file_path, refresh: true)
   end
 end
 
-AppSecrets = SuperConfig.new do
-  if Rails.env.production?
+AppSecrets = SuperConfig.new(raise_exception: !ENV.fetch("ASSETS_COMPILATION", false)) do
+  if Rails.env.production? && !ENV.fetch("ASSETS_COMPILATION", false)
```
gregorbg, line 49: You already have access to `EnvConfig` within this `do` block

config/environments/production.rb
```diff
@@ -99,7 +99,9 @@
   end
 
   # Setup for ActiveStorage.
-  config.active_storage.service = :amazon
+  unless ENV.fetch("ASSETS_COMPILATION", false)
```
gregorbg, line 102: Outsource to `EnvConfig`

env_config.rb
```diff
@@ -2,7 +2,7 @@
 
 require "superconfig"
 
-EnvConfig = SuperConfig.new do
+EnvConfig = SuperConfig.new(raise_exception: !ENV.fetch("ASSETS_COMPILATION", false)) do
```
gregorbg, line 5: Hm, this is the only tricky part where you cannot let EnvConfig parse it for you because you're literally just creating EnvConfig...

Dockerfile
FinnIckler: because these are needed in the webpacker compilation process

FinnIckler: see `application.js.erb`

env_config.rb
```diff
@@ -78,4 +78,7 @@
 
   # For server status
   optional :BUILD_TAG, :string, "local"
+
+  # For Asset Compilation
+  optional :ASSETS_COMPILATION, :boolean, false
```
gregorbg, line 83:
```suggestion
  optional :ASSETS_COMPILATION, :bool, false
```

FinnIckler:
Moved the env variables needed for compilation into a .env file and it successfully compiled. Gonna push the image to staging now. We still need one variable in the Dockerfile (WCA_LIVE_SITE), but that actually makes sense to set there

FinnIckler:
Successfully tested in staging!

PR #9702, opened 2024-07-23
Cache user-groups

danieljames-dj:
Currently whenever a user-group is loaded from the class methods in user_group.rb, a database read happens. This PR aims to avoid that.

### DIFF
--- a/app/models/concerns/cachable.rb
+++ b/app/models/concerns/cachable.rb
@@ -35,7 +35,7 @@ module Cachable
     end
 
     def as_cached
-      self.class.try(self.cachable_id) || self.class.c_find(self.cachable_id)
+      self.class.try(self.cachable_id.to_s) || self.class.c_find(self.cachable_id)
     end
   end
 
--- a/app/models/concerns/cachable.rb
+++ b/app/models/concerns/cachable.rb
@@ -35,7 +35,7 @@ module Cachable
     end
 
     def as_cached
-      self.class.c_find(self.cachable_id)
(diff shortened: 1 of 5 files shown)


danieljames-dj:
My initial attempt to solve this problem was to use a class-level variable, for example - `UserGroup.board.first` will be replaced by `@board_group ||= UserGroup.board.first`. But I'm getting warning - "ThreadSafety/InstanceVariableInClassMethod: Avoid instance variables in class methods."

Other potential solutions are:
1. Using thread-local variables using `Thread.current`.
2. Using Rails.cache - but not sure what can be a good expiry in this case.

Looking for suggestions on what will be a good approach.

FinnIckler:
You can use https://apidock.com/rails/v5.2.3/Module/thread_mattr_accessor like we do in the cachable module

danieljames-dj:
Hi @FinnIckler, thanks for your feedback. I tried using `thread_mattr_accessor`, but I am getting the database row with `id`, `created_at` and `updated_at` as `nil`. But it was working fine if I replace it with `mattr_accessor`. Do you think it's fine to use `mattr_accessor` instead? These are anyway user_groups which will very rarely change and even if they change, it will be changed only through migrations.

FinnIckler:
look into the cachable module how to use it. Something like

gregorbg (approved):
Feels a bit odd to Approve a PR that I contributed to, but Daniel and I both tested this thing and we're happy with how it works. (And GitHub won't let Daniel approve because he's the one who originally raised the PR and users on GH cannot review their own PRs)

PR #9704, opened 2024-07-23
Reuse import logic from v1

FinnIckler:
Not Tested yet, mostly looking for an opinion on if this is good. Will also have to change the export logic to use this format. I don't think it is worth introducing a new format and writing new checks until we allow 3rd parties to just use our API and remove the csv import


FinnIckler:
https://github.com/thewca/wca-registration/pull/627 for the wca-registration side

FinnIckler:
Also, am I seeing this correctly that the Status field that's in the csv export actually doesn't matter? It looks to me that we just say that if the name is in the csv than it's accepted and if it isn't (but it was before) it's deleted.

gregorbg:
Just curious, I think with V3 this PR is completely redundant?

FinnIckler:
yes

PR #9705, opened 2024-07-23
Save Processing in LocalStorage to Show Processing on Refresh

FinnIckler:
Will test later today. But the gist is that I am now saving processing in local storage, to show to the user the processing screen, even if he refreshes and his registration still hasn't been created


kr-matthews (commented):
One thing to be aware of is that `useStoredState` only checks the local storage value once, on initial render. So if you use the hook in 2 different components which are both rendered at once, and update the value in one component, then the state in that component will update and the local storage value will update, but the state in the other component won't update. (I'm not sure whether that will cause problems in this specific case, but generally using the same key in 2 different components is not great.)

Also, will there be issues if someone is processing, logs out, and then someone else logs in and goes to the same competition (or any other similar scenarios)?

FinnIckler:
That's a good point. I've moved the state into the parent component. I also added `user.id` to the key so there is no issues with shared accounts.

FinnIckler:
@thewca-bot deploy staging

gregorbg:
Obvious question, but why does the client have to remember that the registration is currently being processed anyways?

From my gut feeling, we should design the Registration API in a way that the information "This registration is currently processing" can be inferred from the API response. Is there some conceptual problem that prevents us from sending a `processing` status in the API response?

FinnIckler:
The problem is that you cannot inspect the queue to tell you if a specific message is in it or not. And if we want to write it in the DynamoDB Table on create request, a big point of the queue gets circumvented (that we can guarantee only a certain load to the DB).

FinnIckler:
We could write it into Redis I guess?

dunkOnIT:
Any further thoughts on this in light of Finn's responses Gregor? We'd like to get it merged by the load test next week if possible

gregorbg:
No detail thoughts apart from the general remark that I'm quite amazed nobody ever thought about this over the course of one year worth of designing and planning infrastructure :D

This seems like a pretty central feature of any registration system,... I guess the browser will do. Redis can be lost by our own screw-up, browser is completely the user's control.

PR #9709, opened 2024-07-29
Rename `registration_opened?` to `registration_open?`

dunkOnIT:
This has been requested as a pre-requisite for auto-accepts. Basically, in some regions, organizers/delegates don't currently pre-register for their competitions because they don't want to show up on the Competitor list before the competition actually opens (causing confusion for competitors/generally looking a bit weird.
 
So, I've suggested that we only show the "Competitors" tab when registration for a competition is actually open. 

## Implementation
The flaw in this implementation is that if registration closes, and has another `registration_open` date set, then until registration opens again, the Competitor list will not be visible. 

To overcome this, we could add a boolean to the competition table which tracks whether registration has opened at least once. But that would be yet another setting in the competition table, which isn't ideal. I'll ask WCAT about the relevance of this edge case.


app/views/competitions/_nav.html.erb
```diff
@@ -133,15 +133,16 @@
...
           if @competition.use_wca_registration?
             nav_items << {
               text: t('.menu.register'),
               path: competition_register_path(@competition),
               icon: "sign in alt"
             }
           end
-          if @competition.registrations.any?
+
+          if !@competition.registration_not_yet_opened? && @competition.has_any_registrations?
```
gregorbg, line 145: The first part of the `&&` clause uses a `!` on a method that has "not" in its name. So effectively, it's a double negation "If not not yet opened".

Can you look if there's a "if competition registration open" method? Pretty sure there already ought to be one.

gregorbg (changes requested):

app/views/competitions/_nav.html.erb
dunkOnIT: There is some method to the madness here, because we are only checking if the registration _has opened_, not if it is currently open. Here are the old method definitions: 

```ruby
  def registration_opened?
    use_wca_registration? && !cancelled? && !registration_not_yet_opened? && !registration_past?
  end

  def registration_not_yet_opened?
    registration_open && Time.now < registration_open
  end

  def registration_past?
    registration_close && registration_close < Time.now
  end
```
We can't use the old `registration_opened?` because it invokes `registration_past?`, which returns true if the registration is closed. 

My latest commit changes the logic of `registration_opened?` to ONLY check if the registration has opened in the past, and adds a new `registration_is_open?` method to check if registration is _currently_ open

app/models/competition.rb
```diff
@@ -950,6 +950,10 @@ def can_edit_registration_fees?
   end
 
   def registration_opened?
+    use_wca_registration? && !cancelled? && !registration_not_yet_opened?
+  end
+
+  def registration_is_open?
```
gregorbg, line 956: As per Ruby conventions, boolean-esque methods don't incude any modal verbs like `is`, `has` or `should`. The question mark itself is already the "is" part.

app/models/competition.rb
```diff
@@ -950,6 +950,10 @@ def can_edit_registration_fees?
...
     use_wca_registration? && !cancelled? && !registration_not_yet_opened? && !registration_past?
```
gregorbg, line 957: Also, unrelated: If you introduce `registration_opened?` you can just reuse that here:

```suggestion
    registration_opened? && !registration_past?
```

gregorbg (changes requested):
As much as I agree about the confusing logic of the competition method names, I'm really afraid about this change because it
(a) renames an existing method and
(b) introduces a new method that is named exactly like the old one

where especially point (b) is dangerous. Some parts of the code that you did not catch might still reference the old `registration_opened?` implementation which now "accidentally" gets rerouted to a semantically different implementation.

I'd feel much more comfortable if you split htis up into a separate PR that only does renames (basically, a glorified version of point (a) above, maybe also considering the nomenclature of other `registration_foo?` methods) and then after that passes and can be merged, pick this up again with only an implementation for (b)

dunkOnIT:
Ok this is ready to review again - I will add a method to check if registration _has been_ open in a follow up PR

gregorbg (commented):
For the sake of what's to come, might I suggest `registration_currently_open?`?

dunkOnIT:
I was 50/50 between this and `registration_currently_open` - so that swings it.

gregorbg (approved):

PR #9712, opened 2024-07-30
OAuth Login for Staging

dunkOnIT:


FinnIckler:
took over this PR to implement OAuth login instead

FinnIckler:
@thewca-bot deploy staging

FinnIckler:
@thewca-bot deploy staging

dunkOnIT:
@thewca-bot deploy staging

gregorbg (changes requested):

app/views/devise/sessions/new.html.erb
```diff
@@ -40,6 +40,12 @@
           <%= f.submit t('devise.sessions.new.sign_in'), class: "btn btn-primary", tabindex: "3" %>
           <%= t 'wca.devise.no_account' %> <%= link_to t('devise.shared.links.sign_up'), new_user_registration_path %>!
         <% end %>
+        <% unless EnvConfig.WCA_LIVE_SITE? %>
+          <div>
+            or <br>
+            <button class="ui button primary"><%= link_to 'Sign in through Main WCA Page', staging_login_url, style: "color: white" %></button>
+          </div>
```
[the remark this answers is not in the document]
FinnIckler: I switched it to `btn btn-primary`. We could also switch the other button to use the semantic styling

app/controllers/sessions_controller.rb
```diff
@@ -8,6 +8,54 @@ class SessionsController < Devise::SessionsController
...
+      builder.request :json
+      builder.response :json
+
+      # Raises an error on 4xx and 5xx responses.
+      builder.response :raise_error
+
+      # Logs requests and responses.
+      # By default, it only logs the request method and URL, and the request/response headers.
+      builder.response :logger, ::Logger.new($stdout), bodies: true if Rails.env.development?
+    end
```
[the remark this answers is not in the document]
FinnIckler: I introduces a faraday helper. Should I change it in the other places as well? Or in a different PR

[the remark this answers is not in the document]
gregorbg: Different PR will suffice :)

app/views/devise/sessions/new.html.erb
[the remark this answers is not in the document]
dunkOnIT: I suggest:
- Changing wording to "Sign in using your WCA Account"
- Putting this button above the login box / somehow making it more attention-grabby than the login box (it should be the default
- This starts to deviate a lot, but if we could have the sign-in box be accessed by clicking a dropdown which says "Specify account details (only use if you are a developer with access)" or something, that would really help direct people to the OAuth option by default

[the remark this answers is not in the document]
FinnIckler: made it green and moved it to the top

FinnIckler:
image
This is the watermark I just added. Other suggestions welcome

app/controllers/sessions_controller.rb
```diff
@@ -8,6 +8,45 @@ class SessionsController < Devise::SessionsController
...
+
+    results = connection.get("/api/v0/me").body
+
+    user = User.find(results["me"]["id"])
+    if user
+      sign_in(user)
+      redirect_to root_url, notice: "Successfully logged in as #{user.wca_id}"
+    else
+      redirect_to root_url, alert: "Couldn't find your user"
```
gregorbg, line 46: Be a bit more talkative about what that actually means (On staging, dev dumps are imported periodically so data might be out of sync, or we might just be in the middle of importing something

app/controllers/sessions_controller.rb
```diff
@@ -8,6 +8,45 @@ class SessionsController < Devise::SessionsController
...
+
+    # Get /me to figure out which user we are
+    connection = Faraday.new(
+      url: EnvConfig.STAGING_OAUTH_URL,
+      headers: {
+        'Authorization' => "Bearer #{access_token}",
+        'Content-Type' => 'application/json',
+      },
+    ) do |builder|
+      FaradayHelper.default_connection(builder)
```
gregorbg, line 36: Methinks you can pass this method reference directly: `Faraday.new(...foo..., FaradayHelper.method(:default_connection))`

In case of doubt, consult ChatGPT

app/views/devise/sessions/new.html.erb
```diff
@@ -3,44 +3,10 @@
...
-                <%= f.label :remember_me %>
-              </label>
-            </div>
-            <% end %>
-          <%= f.submit t('devise.sessions.new.sign_in'), class: "btn btn-primary", tabindex: "3" %>
-          <%= t 'wca.devise.no_account' %> <%= link_to t('devise.shared.links.sign_up'), new_user_registration_path %>!
-        <% end %>
-      </div>
-    </div>
+    <% if Rails.env.production? && !EnvConfig.WCA_LIVE_SITE? %>
```
gregorbg, line 6: The name(s) ending in _template are a bit misleading/generic. Is there really no better option?

dunkOnIT: Names updated and rendering tested locally

app/controllers/sessions_controller.rb
FinnIckler: I figured out a way as per https://github.com/lostisland/faraday/issues/946#issuecomment-500607890

app/controllers/sessions_controller.rb
FinnIckler: done

gregorbg (approved):
LGTM, needs a merge conflict resolution.

Did you reach a consensus on the watermark? I am generally in favor of it.

PR #9733, opened 2024-08-01
Redirect users to contact form when allow_registration_edits is false and prefill it

FinnIckler:
Gets rid of another of those pesky banners and allows users to contact the organizers with a prefilled template.

@danieljames-dj I know you worked on the contact form, is this the way I should interact with it to prefill it? I noticed you had the message property already there, but you wasn't using it


danieljames-dj (approved):
Hi @FinnIckler, yes the usages of the contact form looks good to me, I didn't had it in the reducer.js just because there wasn't any use-case so far and now you have.
I have also went through all the changes in this PR and overall LGTM, hence approved.

dunkOnIT:
@thewca-bot deploy staging

dunkOnIT:
@thewca-bot deploy staging

dunkOnIT:
@thewca-bot deploy staging

dunkOnIT:
@thewca-bot deploy staging

dunkOnIT:
The text formatting in the prefill could be a lot cleaner. I'd suggest the following format: 

> I would like to update my registration details as follows:
> Events: 3x3x3, 3bld
> Comment: Here's an example comment
> Guests: 7

And keep the behaviour of not including unchanged fields, that's great - eg if no change to guests is made: 

> I would like to update my registration details as follows:
> Events: 3x3x3, 3bld
> Comment: Here's an example comment

app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
```diff
@@ -171,18 +172,23 @@ export default function CompetingStep({
...
+          user_id: registration.user_id,
+          competition_id: competitionInfo.id,
+          competing: {
+            comment: hasCommentChanged ? comment : undefined,
+            event_ids: hasEventsChanged ? selectedEvents : undefined,
+          },
+          guests,
+        });
+      } else {
+        const updateMessage = `${hasCommentChanged ? `Comment to ${comment} ` : ''}${hasEventsChanged ? `Events to ${selectedEvents.map((eventId) => events.byId[eventId].name).join(', ')} ` : ''}${hasGuestsChanged ? `Guests to ${guests}` : ''}`;
```
dunkOnIT, line 189: The text formatting in the prefill could be a lot cleaner. I'd suggest the following 

> I would like to update my registration details as follows:
> Events: 3x3x3, 3bld
> Comment: Here's an example comment
> Guests: 7


And keep the behaviour of not including unchanged fields, that's great - eg if no change to guests is made:

> I would like to update my registration details as follows:
> Events: 3x3x3, 3bld
> Comment: Here's an example comment

app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
```diff
@@ -171,18 +172,23 @@ export default function CompetingStep({
...
+        window.location = contactCompetitionUrl(competitionInfo.id, encodeURIComponent(i18n.t('competitions.registration_v2.update.update_contact_message', { update_params: updateMessage })));
```
dunkOnIT, line 190: Let's also add a banner message on the contact page which says the following: 

> To update your registration, you need to contact the organizers of the competition. We have prefilled the contact form with your requested changes - please submit the form.

dunkOnIT (changes requested):
Disregard the lines I added these comments on :P

dunkOnIT (approved):
Happy with the changes!

Issue #8440, opened 2023-10-31
Add a warning below the WCA ID box for all authorized WCA staff members 


lnzainn:
I believe that at the moment, only the Board, WRT and WST can see the warning (see image below) for special users 

![image (1)](https://github.com/thewca/worldcubeassociation.org/assets/99952754/b19ccd3c-a7d4-4324-a7d2-83a01d471ab8)

I believe having this warning show up for all the staff members (who have the permission to view the edit page of any profile) will be beneficial.

dunkOnIT:
100% agree, and pretty straightforward change for anyone looking to put in a first issue.

lnzainn:
> Should organizers be included in this? The message itself says that they should, however they don’t have permission to view the edit page of any profile.

It doesn't say that organizers should?

Carter-Thomas:
Oh wait, I misread. I thought that special users (which include organizers) were part of the group that were trying to be added.

PR #9734, opened 2024-08-01
Warning for special type of accounts

lnzainn:
#8440

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -988,7 +988,7 @@ class User < ApplicationRecord
 
   private def editable_competitor_info_fields(user)
     fields = Set.new
-    if user == self || admin? || any_kind_of_delegate? || results_team? || communication_team?
+    if user == self || can_edit_any_user?
       unless cannot_edit_data_reason_html(user)
         fields += %i(name dob gender country_iso2)
       end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -714,6 +714,10 @@ class User < ApplicationRecord
     self == user || can_view_all_users? || organizer_for?(user)
   end
 
+  def can_edit_any_user?
+    admin? || any_kind_of_delegate? || can_admin_results? || communication_team?
+  end
+
   def can_change_users_avatar?(user)
     user.wca_id.present? && self.editable_fields_of_user(user).include?(:remove_avatar)
   end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -997,7 +997,7 @@ class User < ApplicationRecord
     if user.wca_id.blank? && organizer_for?(user)
       fields << :name
     end
-    if admin? || any_kind_of_delegate? || results_team? || communication_team?
+    if can_edit_any_user?
       fields += %i(
         unconfirmed_wca_id
       )
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -715,7 +715,7 @@ class User < ApplicationRecord
   end
 
   def can_edit_any_user?
-    admin? || any_kind_of_delegate? || can_admin_results? || communication_team?
+    admin? || any_kind_of_delegate? || results_team? || communication_team?
   end
 
   def can_change_users_avatar?(user)
--- a/app/views/competitions/_nav.html.erb
+++ b/app/views/competitions/_nav.html.erb
@@ -234,7 +234,7 @@
 
     </div>
     <div id="competition-data">
-      <h3 style="<%= @competition.uses_new_registration_service? ? "display:none" : ""  %>">
+      <h3>
         <% if current_user && @competition.upcoming? %>
           <span class="bookmark-icon" id="not-bookmarked" data-toggle="tooltip" data-placement="bottom" title="<%= I18n.t('competitions.competition_info.bookmark') %>" <% if current_user.is_bookmarked?(@competition) %>style="display: none;"<% end %>>
             <%= ui_icon('bookmark') %>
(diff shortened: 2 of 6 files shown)


lnzainn:
@danieljames-dj @dunkOnIT good to merge this?

danieljames-dj:
Thanks @lnzainn for raising this PR. But I really don't think this change aligns completely with what you intended for. By reading the issue that you tagged, I see that "all staff members who have permission to view [...]", but if your change is merged, this will shown to all users who have access to edit the user's account details which include the user themselves as well irrespective of whether they are staff or not).

(in case you are wondering if the user has access to edit their own account details then can they edit the WCA ID as well, the answer is no because WCA ID is added as editable field [only for this set of users](https://github.com/thewca/worldcubeassociation.org/blob/da0a9bdb2829bd853137a902871f33063923d19e/app/models/user.rb#L996).

My suggestion would be to add the same check in `app/views/users/edit.html.erb` as well.

But I also feel that it will be code duplication because you are writing `admin? || any_kind_of_delegate? || results_team? || communication_team?` in multiple places. I would also suggest to move this to a new method named something like `can_edit_any_users?` and use that method in both `app/views/users/edit.html.erb` and inside `editable_competitor_info_fields` method of user.rb.

lnzainn:
Thanks for your suggestion, Daniel. This was something I discussed with Duncan and I get your point. On second thoughts, I agree, only those who can edit should be able to see that message. I'll propose the new changes

dunkOnIT:
My first response was "Best to ask Daniel about this", I should have just stuck with that :laughing:

lnzainn:
I wasn't sure if Board comes under 'admin?' method so I removed results_team? and replaced it with 'can_admin_results?' which I believe contains both board and WRT.

danieljames-dj:
> I wasn't sure if Board comes under 'admin?' method so I removed results_team? and replaced it with 'can_admin_results?' which I believe contains both board and WRT.

Is there any reason why we want Board as well? Usually they won't handle these cases unless they are a delegate as well. If they have any case like this, it will be better if they do it through a local delegate or the necessary team because the chances are less that a non-Delegate Board member knows the correct procedur (they don't need to know as well because it's not part of Board workflows).

app/models/user.rb
```diff
@@ -993,7 +997,7 @@ def editable_fields_of_user(user)
     if user.wca_id.blank? && organizer_for?(user)
       fields << :name
     end
-    if admin? || any_kind_of_delegate? || results_team? || communication_team?
+    if can_edit_any_user?
```
danieljames-dj, line 1000: Can you please use this method in line 991 as well? Something like `user == self || can_edit_any_user?`

lnzainn:
Thanks Daniel for the catch, I have made the necessary changes.

danieljames-dj (approved):
Thanks @lnzainn, LGTM.

app/models/user.rb
```diff
@@ -714,6 +714,10 @@ def can_edit_user?(user)
     self == user || can_view_all_users? || organizer_for?(user)
   end
 
+  def can_edit_any_user?
+    admin? || any_kind_of_delegate? || results_team? || communication_team?
```
dunkOnIT, line 718: @danieljames-dj should `communication_team` have edit permissions in your view? In my mind, they should only need view permissions.

app/models/user.rb
```diff
@@ -984,7 +988,7 @@ def editable_fields_of_user(user)
 
   private def editable_competitor_info_fields(user)
     fields = Set.new
-    if user == self || admin? || any_kind_of_delegate? || results_team? || communication_team?
+    if user == self || can_edit_any_user?
```
dunkOnIT, line 991: Bit weird to have an unless immediately after an if block - I suggest merging them so that the line is:

```ruby
if user == self || can_edit_any_user? || !cannot_edit_data_reason_html(user)
```

dunkOnIT (changes requested):

app/models/user.rb
lnzainn: @dunkOnIT WCT has been regularly handling wca id re assigning tasks which are trivial and can be performed by WRT too. We in fact had a discussion with what should be done but no conclusive solutions. @danieljames-dj should there be another WRT discussion on this? Perhaps if both teams vote together on what should be done, then we maybe will get a permanent solution

danieljames-dj: @dunkOnIT Yes, currently, they handle requests like "change WCA ID to new account" themselves instead of forwarding them to WRT.

danieljames-dj: @lnzainn We can think of having another discussion for sure. But I guess that won't block this PR to be submitted anyway because this PR doesn't change any permissions related stuffs. :-)

dunkOnIT: Ah ok I see - and is this preferred by WRT, or is it worth re-evaluating if this is the best arrangement? (I assume it's preferred by WRT, but sometimes things can just be the way they are because no one's thought about them in a while - eg, WRC having permission to view registrations :P

lnzainn: Not speaking of behalf of WRT, but when someone from wct made a rookie mistake of forwarding such request to wrt, then response from wrt was 'this can be done by wct themselves, wrt intervention is not required' 

Btw I'm that rookie who made this mistake when I had newly joined the team 😂

app/models/user.rb
dunkOnIT: What do you mean by that?

gregorbg (approved):

app/models/user.rb
gregorbg: Daniel is right, you cannot merge anything here. The GitHub diff view might be misleading: There's actually stuff happening inside the outer `if ...` block but _outside_ of the nested `unless` block.

dunkOnIT: Ah sorry I see now! Probably just me misreading the diff.

dunkOnIT (approved):
LGTM!

PR #9738, opened 2024-08-03
Fix registration table stacking

dunkOnIT:
We had the same issue with the Events and Schedule tabs in mobile


FinnIckler (approved):

FinnIckler:
We should do the same for the competitors table

PR #9740, opened 2024-08-05
remove WFC from role start and role change emails

lnzainn:
Request from Daniel James.


danieljames-dj (approved):
Thanks @lnzainn for making this change.

For others: context is that WFC figured out a way of not manually adding delegates to xero contacts all the time. So, WFC no longer needs to receive these two emails.

Issue #9569, opened 2024-06-22
Incomplete ranks shown while CAD is running


jonatanklosko:
We have the manually-triggered background job for computing ranks information (CAD). For each of the ranks tables, it removes all rows, then computes and inserts fresh rows. While the computation is running, the ranks tables have missing information, which surfaces in the UI and API.

Here are a few screenshots while running CAD on the staging, where it can be seen how CAD progresses:

<details><summary>Screenshots</summary>

* Records showing only average as ConciseSingleResults are being computed
  <img width="1512" alt="image" src="https://github.com/thewca/worldcubeassociation.org/assets/17034772/1d757643-7035-415a-9c50-bf21d91d4018">
* Records showing only single as ConciseAverageResults are being computed
  <img width="1512" alt="image" src="https://github.com/thewca/worldcubeassociation.org/assets/17034772/82f5c9da-b6ed-485a-ab42-1b73fef58f80">
* Incremental entries in the profile as RanksSingle are being computed
  <img width="1512" alt="image" src="https://github.com/thewca/worldcubeassociation.org/assets/17034772/9a4ea9e4-030b-40b5-a909-1affc72ce61a">
  <img width="1512" alt="image" src="https://github.com/thewca/worldcubeassociation.org/assets/17034772/7fddaeab-58d2-425f-ad72-1449dee2d305">

</details> 

The same issue affects API endpoints that rely on ranks (records, WCIF personal bests), and that is a bigger problem, because API users definitely don't expect records to suddenly have no entries.

For more context, I believe this is the underlying issue behind a few strange false-positive record tags we've seen on WCA Live https://github.com/thewca/wca-live/issues/145.

Note that some of the pages/endpoints are cached, so keep that in mind when reproducing.

gregorbg:
Interesting... Taking a complete shot in the dark here but would wrapping the whole CAD process into a transaction help?

jonatanklosko:
We already update each table in a transaction, so it should be atomic:

https://github.com/thewca/worldcubeassociation.org/blob/1e83c26f9b204bfbd236c6a074fff309eb9d8a47/lib/auxiliary_data_computation.rb#L15-L19

Just looking at the code I don't have any reasonable idea, unless the queries skip the transaction somehow. One option to test would be to add rollback after the delete and see if that works as expected.

dunkOnIT:
[...]

PR #9742, opened 2024-08-05
Use temp tables for CAD job

dunkOnIT:
This is an example fix for #9569 - I'm well out of my depth here, and this was based entirely on a conversation with ChatGPT. I assume more SQL-literate people will be able to assess its merit, and whether it's worth taking a similar approach for the second job.

EDIT: I do now feel a lot more confident that I actually understand what CAD is about, and what the changes are doing :smile:

### DIFF
--- a/lib/auxiliary_data_computation.rb
+++ b/lib/auxiliary_data_computation.rb
@@ -12,15 +12,10 @@ module AuxiliaryDataComputation
       %w(best ConciseSingleResults),
       %w(average ConciseAverageResults),
     ].each do |field, table_name|
-      ActiveRecord::Base.transaction do
+      begin
         temp_table_name = "#{table_name}_temp"
-
-        # Drop the temporary table if it exists
-        ActiveRecord::Base.connection.execute("DROP TABLE IF EXISTS #{temp_table_name}")
-
-        # Create a copy of the original table structure
         ActiveRecord::Base.connection.execute("CREATE TABLE #{temp_table_name} LIKE #{table_name}")
-
+        ActiveRecord::Base.connection.execute("ALTER TABLE #{temp_table_name} AUTO_INCREMENT = 1")
         ActiveRecord::Base.connection.execute <<-SQL
           INSERT INTO #{temp_table_name} (id, #{field}, valueAndId, personId, eventId, countryId, continentId, year, month, day)
           SELECT
@@ -52,6 +47,8 @@ module AuxiliaryDataComputation
 
         # Drop the old table
         ActiveRecord::Base.connection.execute("DROP TABLE #{table_name}_old")
+      ensure
+        ActiveRecord::Base.connection.execute("DROP TABLE IF EXISTS #{temp_table_name}")
       end
     end
   end
--- a/lib/auxiliary_data_computation.rb
+++ b/lib/auxiliary_data_computation.rb
@@ -13,10 +13,16 @@ module AuxiliaryDataComputation
       %w(average ConciseAverageResults),
     ].each do |field, table_name|
       ActiveRecord::Base.transaction do
-        ActiveRecord::Base.connection.execute "DELETE FROM #{table_name}"
-        ActiveRecord::Base.connection.execute "ALTER TABLE #{table_name} AUTO_INCREMENT = 1"
+        temp_table_name = "#{table_name}_temp"
+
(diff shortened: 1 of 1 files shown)


lib/auxiliary_data_computation.rb
```diff
@@ -13,10 +13,16 @@ def self.compute_concise_results
       %w(average ConciseAverageResults),
     ].each do |field, table_name|
       ActiveRecord::Base.transaction do
```
jonatanklosko, line 15: There's probably no need for a transaction anymore, because DDL statements like `CREATE TABLE` implicitly commit transactions anyway (this varies across DBMS, but I understand [this is the case](https://dev.mysql.com/doc/refman/8.4/en/implicit-commit.html) for MySQL). It should be fine, since we swap the tables only if the new one was successfully computed.

lib/auxiliary_data_computation.rb
```diff
@@ -13,10 +13,16 @@ def self.compute_concise_results
...
-        ActiveRecord::Base.connection.execute "DELETE FROM #{table_name}"
-        ActiveRecord::Base.connection.execute "ALTER TABLE #{table_name} AUTO_INCREMENT = 1"
+        temp_table_name = "#{table_name}_temp"
+
+        # Drop the temporary table if it exists
+        ActiveRecord::Base.connection.execute("DROP TABLE IF EXISTS #{temp_table_name}")
```
jonatanklosko, line 19: Ideally we should never keep this temporary table around. Instead of checking and deleting here, it would be better to make sure the table is always removed at the end of the computation (even if it fails). IIRC we can wrap it like this:

```ruby
begin
  ...
ensure
  ActiveRecord::Base.connection.execute("DROP TABLE IF EXISTS #{temp_table_name}")
end
```

jonatanklosko:
Looks good to me! I think this is the next best alternative to a transaction (which clearly doesn't seem to work as needed), and semantically it makes sense, as we literally recompute the whole table.

dunkOnIT:
Thanks! Will it be necessary to apply the same approach for the `compute_rank_tables` function? I'm not sure what the data source is that WCA Live relies on, but I sort of assume it's that table.

jonatanklosko:
@dunkOnIT yeah we need to so the same for both concise results and ranks. This is not specific to WCA Live. If you look at the screenshots at #9569, issues with both tables surface in the WCA website UI :)

dunkOnIT:
Ok cool - have applied the same approach for `compute_ranks_tables`. I've run `ComputeAuxiliaryData.perform_now` locally and it seems to work, but no idea how I would go about testing this beyond that.

lib/auxiliary_data_computation.rb
dunkOnIT: Fixed, much nicer solution!

lib/auxiliary_data_computation.rb
dunkOnIT: Replaced the `transaction` wrapper with a `begin`/`ensure` wrapper

jonatanklosko (approved):
Looks good! If it computes as expected locally I think we are good. If you want to be extra cautious, you can double-check if it runs on the staging.

With the current code we always use a filled table, so it should be safe to assume it addresses the issue.

dunkOnIT:
@thewca-bot deploy staging

lib/auxiliary_data_computation.rb
jfly: For the record, running code in an `ensure` block doesn't actually guarantee it will happen (the ruby process could crash, someone could trip over a power cord in the data center, etc).

Does this code assume the temp table doesn't exist at the start? I glanced through it and it seems like it might. If I'm right, then the only truly safe option would be to drop the temp table at the start if it exists.

jonatanklosko: If that happens, the next job would just fail and we would know something unexpected went wrong, which I'd say is fine in this case (given how unlikely those events are). But if we want to be absolutely sure, we could run in both places, yes :D

FinnIckler:
CAD did run successfully. Not sure how I check that the results are correct though
Screenshot 2024-08-06 at 12 26 11

jonatanklosko:
@FinnIckler just looking around results and profile pages should be enough. From a brief look, seems all good to me :)

PR #9745, opened 2024-08-06
Skip all rows for microservice_registrations in dev_dump

FinnIckler:
Makes no sense to keep these anyway for the Dev Dump, as they rely on DynamoDB having the same data


dunkOnIT (approved):
LGTM let's merge when the tests pass

PR #9747, opened 2024-08-06
Enable V2 Registrations for my competitions in prod

FinnIckler:
We still need to figure out what to do in development because we don't persist v2 registrations locally right now which causes errors.


FinnIckler:
One solution would be to switch to the pro version of localstack, or https://hub.docker.com/r/gresau/localstack-persist in wca-registrations

gregorbg (approved):

PR #9749, opened 2024-08-07
Don't show register tab before comp is announced

dunkOnIT:


gregorbg:
What is the context/use-case here?
I understand from the title and from the code what this change does, but I don't understand why this wasn't an issue in the legacy V1 registration system

dunkOnIT:
We didn't investigate in too much depth, but the new reg system needs to get the competition details and via the API, and it seems like the API won't return the details of an unannounced comp [or something in the JSON doesn't play nicely with how we're currently reading the API response]. 

Either way, this seemed like the most time-efficient fix.

FinnIckler:
Yes, the v0/competitions/{comp_id} API does only return visible announced competitions

gregorbg (approved):

PR #9752, opened 2024-08-07
Add validations to subforms of contact form

danieljames-dj:
This is to add some client-side validations in the contact form.

### DIFF
--- a/app/webpacker/components/ContactsPage/ContactForm.jsx
+++ b/app/webpacker/components/ContactsPage/ContactForm.jsx
@@ -37,10 +37,16 @@ export default function ContactForm({
   const [contactSuccess, setContactSuccess] = useState(false);
   const contactFormState = useStore();
   const dispatch = useDispatch();
-  const { formValues: { contactRecipient: selectedContactRecipient, userData } } = contactFormState;
+  const {
+    formValues: { contactRecipient: selectedContactRecipient, userData },
+    validSubForm,
+  } = contactFormState;
 
   const isFormValid = (
-    selectedContactRecipient && userData.name && userData.email && captchaValue
+    userData.name && userData.email
+    && selectedContactRecipient
+    && validSubForm[selectedContactRecipient]
+    && captchaValue
   );
 
   const contactSuccessHandler = () => {
--- a/app/webpacker/components/ContactsPage/SubForms/Competition/index.jsx
+++ b/app/webpacker/components/ContactsPage/SubForms/Competition/index.jsx
@@ -1,10 +1,10 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import { FormTextArea } from 'semantic-ui-react';
 import { IdWcaSearch } from '../../../SearchWidget/WcaSearch';
 import SEARCH_MODELS from '../../../SearchWidget/SearchModel';
 import I18n from '../../../../lib/i18n';
 import { useDispatch, useStore } from '../../../../lib/providers/StoreProvider';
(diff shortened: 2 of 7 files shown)


gregorbg (commented):
Why are you manually performing these basic validity checks?

As far as I can tell, everything you're doing is a basic equivalent of "this value has to be present". That can be achieved by setting the form field to `required` and letting the browser do its job.

In any case, I'm always quite allergic to `useEffect` and it doesn't feel right to introduce so many hooks for such a simple purpose.

danieljames-dj:
The main challenge was for the radio button form field like following:
```
          <FormField key={queryType}>
            <Radio
              label={I18n.t(`page.contacts.form.wrt.query_type.${queryType}.label`)}
              name="queryType"
              value={queryType}
              checked={selectedQueryType === queryType}
              onChange={handleFormChange}
            />
          </FormField>
```

Here, `required` won't work because it was bringing the red star to each of the options instead of the form field as such. Do you have any idea on how to do validations for the radio button option?

gregorbg:
> Here, `required` won't work because it was bringing the red star to each of the options instead of the form field as such. Do you have any idea on how to do validations for the radio button option?

Hm, I couldn't find a way to mark the whole radio group as `required` either. According to https://stackoverflow.com/a/8287947, the general approach for HTML forms (not just SemUI specifically) is to just mark one of the radios as required, and that should already let the browser handle the validation.

Of course now it gets a little bit cumbersome because you're looping, but JS `map` function can provide the index of the current item as second argument to the mapping function, maybe you can use that index and set `required={i == 0}` along with a nice comment?

danieljames-dj:
Thanks @gregorbg, I'll try that.

But now I actually have one more (more important) use-case where `required` is not working - `WcaSearch`. Can you please let me know how we can implement `required` for `WcaSearch` or at least for `Dropdown` object (or maybe even `Select`)?

I did my own research on it and ended up [here](https://github.com/Semantic-Org/Semantic-UI-React/issues/1423#issuecomment-285896180) where it mentioned following:

> Regarding form validation behavior associated with the required attribute, I don't think it is idiomatic React to delegate validation to the form controls themselves.

I read some more below it (but not so deeply) and linked ones, but couldn't find any solutions.

Do you have any idea how to handle `Dropdown` validation for this case?

gregorbg:
Thanks for digging up that issue! It is indeed a bit frustrating that SemUI doesn't support this, because a comment further down in the issue that you linked does indicate that the HTML DOM element supports it basically.

And yes, I get the argument about letting React do the job *when it's about complex validations*. If you had to do some stuff like "Verify that the user competed at least 7 times", then I'd be fine with your basic approach of React-sided validations (although I would still debate about the use of `useEffect` specifically, but that's a different topic).
When it comes to a plain and simple "Is this value even present?" aka. "Has the user even interacted with this input in the first place" then I don't feel that it's wrong to use a widely-supported, well-implemented, accessibility-backed browser feature called `required`.

How badly do you need these validations? There are changes regarding our form state libraries on the horizon which could significantly simplify this, but they are at least a couple of weeks if not a few months away...

danieljames-dj:
> How badly do you need these validations? 

One active issue I can say is that if a user doesn't select a competition and try to submit, it will throw error. It's not a really big deal, I'm fine to wait for couple of weeks or months. But I am interested to know - how will the changes in form state library help in this case?

danieljames-dj:
As there was internal announcement that the UI library is going to change, I think the new UI library will be able to handle these things. Expecting that (and to free up my PR quota), I'm closing this PR.

PR #9754, opened 2024-08-07
Serve Assets from Cloudfront

FinnIckler:
Haven't tested it yet, but this should be all the code changes needed.


lib/db_dump_helper.rb
```diff
@@ -46,7 +46,7 @@ def self.dump_developer_db
   end
 
   def self.public_s3_path(file_name)
-    "https://s3.#{EnvConfig.STORAGE_AWS_REGION}.amazonaws.com/#{BUCKET_NAME}/#{file_name}"
+    "#{EnvConfig.ASSET_HOST}/#{file_name}"
```
gregorbg, line 49: Why are you moving the database dumps to the assets host?

FinnIckler: Because we pay 300 bucks a month serving the database dumps directly from s3

FinnIckler: They are already in the asset bucket

gregorbg:
Regarding the flow of the whole build script: At the time of writing this comment, the GitHub CI builds assets and uploads them, and _then_ continues to build server Docker images and trigger the AWS deployment process. But I think that means that the new, updated assets become available much much earlier than the server "expects" them (ie. before the backend that can work with these new assets comes online)

Is this a problem? And if it is, could we solve it by compiling the assets into the Docker container and letting the container upload them during deployment as part of the entrypoint script?

FinnIckler:
That's a good point. If we do that we might as well skip the s3 bucket and just point the cloudfront distribution to /public on the loadbalancer and just invalidate in the entry point

gregorbg:
Does that however alleviate the concerns of Puma not serving the assets well? If Cloudfront points to /public, that same /public is still served by Puma (which sucks at serving static assets), isn't it??

FinnIckler:
That's correct, but it will serve the traffic only once for each geographic area/deploy so it would be much less. S3 upload would still be a pretty clean solution, but it'll be annoying if we ever want to scale because we will delete the assets and reupload them every time we start a new server

FinnIckler:
We could version our assets and then copy over from the current git hash to latest on boot?

gregorbg:
Hm... I think the `/public` solution won't work simply because of a server configuration paradox.

In order to even get the server to return some response under `/public`, you have to set `config.public_file_server.enabled = true`. But our intention behind this PR is to serve asset files through a CDN, which requires(?)* this option to be set to false and to redirect to an asset host instead.

*I'm not 100% sure whether setting/enabling `config.asset_host` to an actual URL also implies that you have to set `public_file_server.enabled` to false. But I feel like setting it to true still defeats the purpose/concept of this feature.

Also, when scaling, and the load balancer points to `/public`, _which_ machine's `/public` is it actually pointing to when running multiple instances? S3 is probably the cleaner solution, and I feel like we're not the only team in the world who wants to serve their assets through S3+Cloudfront in a (semi-)modern AWS environment. Can you do some more googling or perhaps chat with ChatGPT about this?

FinnIckler:
Actually a lot of Guides you find through googling serves from the Machine:
https://headey.net/rails-assets-active-storage-and-a-cloudfront-cdn
https://blog.saeloun.com/2023/04/27/how-to-add-cloudfront-cdn-to-rails-app/
https://aws.amazon.com/blogs/developer/caching-the-rails-asset-pipeline-with-amazon-cloudfront/

This one uses a gem called asset-sync to do S3 + Cloudfront:
https://medium.com/product-musings/ruby-on-rails-aws-how-to-put-assets-on-aws-s3-and-fetch-using-cloudfront-84de9800ce3d

ChatGPT basically uses the same solution I have here, but also versions its asset folders to fix the assets being overwritten on deploy issue. 
You would set `config.assets.prefix = "/assets/{version}"` to fix this (instead of copying over). There's probably a separate config for webpacker.

I also saw that we still need the manifest file in the docker image so Rails knows about the hashed file names. So I will copy that into the Dockerfile

FinnIckler:
Asset sync looks like a cool solution, but the [Repo](https://github.com/AssetSync/asset_sync) looks fairly abandoned

FinnIckler:
I built an image with this setup + assets to S3, but I am getting an `Error during failsafe response: The asset "wca.ico" is not present in the asset pipeline.` error so there is definitely something wrong still. Will continue tomorrow

FinnIckler:
It also looks like that webpacker picks up the production.js config in test which is why all the tests are failing...

FinnIckler:
I just remembered that I didn't include the manifest file in the build as I spun up a new machine to build... So the first error is definitely related to that

FinnIckler:
Running 
1. `RAILS_ENV=production WCA_LIVE_SITE=false ASSETS_COMPILATION=true SECRET_KEY_BASE=1 SHAKAPACKER_ASSET_HOST="https://assets-staging.worldcubeassociation.org/assets/4cf089ef4" BUILD_TAG="4cf089ef4" ASSET_HOST="https://assets-staging.worldcubeassociation.org/assets/" ./bin/rake assets:precompile`
2. `aws s3 sync ./public "s3://assets-staging.worldcubeassociation.org/assets/4cf089ef4"` 
3. `docker build --target monolith --build-arg="BUILD_TAG=4cf089ef4" -t rails-staging .` 
almost works, some of the files do not load and some have the prefix twice in there. I am not 100% sure yet though how `SHAKAPACKER_ASSET_HOST` works, do I even need to set it during the Compilation process?

I also haven't created the prod distribution yet to test the Exports

FinnIckler:
`RAILS_ENV=production WCA_LIVE_SITE=false ASSETS_COMPILATION=true SECRET_KEY_BASE=1 SHAKAPACKER_ASSET_HOST="https://assets-staging.worldcubeassociation.org/" BUILD_TAG="9df8193c9" ./bin/rake assets:precompile`

This one works

FinnIckler:
I've also set everything up so https://assets.worldcubeassociation.org/export/developer/wca-developer-database-dump.zip works!

FinnIckler:
For merging and deployment of this PR I would say we do the following:
- After merge we use the new Deploy Main to Staging workflow to check if it builds and uploads correctly
- If yes we deploy to main
- Then we trigger a new export and see if that is served correctly (only via cloudfront, not via S3)
  - Now that I am thinking about it we need to either start to label our dev export or invalidate the cache after upload

Dockerfile
gregorbg, line 1: GitHub won't let me comment on lines that didn't change, but I guess `RAILS_SERVE_STATIC_FILES="true"` in this build file can then also go.

gregorbg (commented):
Just out of curiosity, where did all the `WCA_LIVE_SITE` flagging go? Seems you deleted a lot of it without apparent replacement

FinnIckler:
It's in the env block on the asset compilation step of the GitHub action now

Dockerfile
FinnIckler: Yeah, right

gregorbg (approved):

PR #9757, opened 2024-08-08
Unify Faraday config under local module instead of global prepend

gregorbg:
CC @FinnIckler


FinnIckler:
looks good! Let me check out the branch and try

app/controllers/api/internal/v1/api_controller.rb
```diff
@@ -13,7 +13,7 @@ def validate_wca_token
     # Renew our token if it has expired or is close to expiring
     Vault.auth_token.renew_self if vault_token_data[:ttl] < 300
 
-    client = Faraday.new(url: EnvConfig.VAULT_ADDR)
+    client = Faraday.new(url: EnvConfig.VAULT_ADDR, &FaradayConfig)
```
FinnIckler, line 16: If you look into the file we actually parse our own requests here

FinnIckler (changes requested):

app/controllers/api/internal/v1/api_controller.rb
gregorbg: Good catch! Fixed :wink:

FinnIckler (approved):
LGTM

PR #9758, opened 2024-08-09
Untangle some threads (aka. clean up caching mechanisms with unnecessary thread_mattr_accessor)

gregorbg:
Turns out we were a bit more careful than we needed to be when introducing strongly thread-safe attribute caches in #9681

The thread-safe accessors can make it so that all threads hold their own, local copies. But in most cases, we actually don't _want_ the threads to hold their own local copies. For example, we don't want the cached 3x3 event to exist N times for N threads.

Further reading on why the GIL is helpful here: https://thoughtbot.com/blog/untangling-ruby-threads

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -234,3 +234,12 @@ Naming/BlockForwarding:
 
 Style/ArgumentsForwarding:
   RedundantBlockArgumentNames: []
+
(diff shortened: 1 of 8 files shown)


FinnIckler (approved):
LGTM

PR #9759, opened 2024-08-11
Implement July 2024 Motions changes: Teams, Commitees, Councils

gregorbg:
- WDC has been renamed to WIC
    - Mail references have been changed from `disciplinary@` to `integrity@`
- WEC has been retired
    - All currently active team members and leaders will have their term ended automatically once this comes online [Should this be set to another concrete date instead of "whatever date this change gets deployed to our servers"?]
    - Permissions regarding voting members and profile access have been absorbed by WIC
    - Members are NOT automatically migrated to WIC [is this correct?]
- Appeals committee has been introduced, mail alias `appeals@`
    - To avoid confusion, I used the handle `wapc` (so that it doesn't shadow the old Council `wac` handle) [is this fine?]
    - No membership information by default, just the empty committee
- WAC (old Advisory Council) has been retired
    - Memberships automatically end the same way that WEC ends [Same question as for WEC about end date applies]
    - The notion of a "Council" still exists even in the updated Motions, so the code revolving around the concept of councils has not been removed
    - Admin permissions for the forum have been transitioned to WCT automatically

Open questions:
- What about the description texts for WIC and WAPC?
    - Currently WIC is a 1:1 copy of the old WDC text with "Disciplicary" replaced by "Integrity"
    - The WEC description text has been deleted without replacement
    - There is no WAPC description text yet, please indicate what we should write here


danieljames-dj (approved):
Thanks @gregorbg for working on this. All looks good to me. I've left a minor comment regarding end date, but you can take a call on that.

Also, I like the idea of naming appeals committee as WAPC instead of WAC, so that it won't cause confusion with WAC (council) at least in the software world.

Regarding method `wic_team?`, name looks better than `integrity_committee?`. I guess you considered the ease to use even though WIC is not a team but a committee. (I think we already have `wrc_team?` which is wrongly called team instead of committee)

lnzainn:
Hi @gregorbg 
Many thanks for working on this. I suggest that to keep things consistent, the end dates of membership for WEC, Advisory Council be set to whenever the motions got passed (since motions proposal was to dissolve em). Does membership date continue for those who were on WDC and now joined WIC? Or that it will reset. 

For the team description, I'll inform both Shain and Stone.

gregorbg:
> Does membership date continue for those who were on WDC and now joined WIC? Or that it will reset.

This PR renames WDC to WIC without any further changes. So if the PR gets merged in its current state, the "new" WIC will have all members from WDC, not more and not less. In other words, WIC **is** WDC because this implementation literally just renames the team but the data backing the team (including memberships and membership history) remains the exact same.

Ex-WDC members will be listed as Ex-WIC members because again, it **is** the same team in terms of code, simply renamed.

slongbehn:
Hi Gregor,

Please use this as the WIC description:

> The WCA Integrity Committee (WIC) performs independent investigations regarding WCA Community members. The subjects of the investigations are incidents during WCA Competitions, incidents related to WCA Competitions, or incidents within the WCA Community. Incidents are alleged violations of the Spirit, Mission, Motions, Policies, Codes and/or Regulations of the WCA.

Additionally, please ensure that disciplinary@ emails do not get removed and that they become aliases for integrity@ so that people's emails do not get lost. Thank you!

gregorbg:
> Hi Gregor,
> 
> Please use this as the WIC description:
> 
> > The WCA Integrity Committee (WIC) performs independent investigations regarding WCA Community members. The subjects of the investigations are incidents during WCA Competitions, incidents related to WCA Competitions, or incidents within the WCA Community. Incidents are alleged violations of the Spirit, Mission, Motions, Policies, Codes and/or Regulations of the WCA.
> 

Thanks! The Board has given me a slightly different description, which is closer to the old WDC description but with key elements like WCA community members and the CoE also included. I went with that instead, please file a complaint if you're not fine with that and we can easily update the description once more :)

> Additionally, please ensure that disciplinary@ emails do not get removed and that they become aliases for integrity@ so that people's emails do not get lost. Thank you!

Yup, that will however depend on some manual configuration in the Google Workspace administration settings and is completely unrelated to code changes per se.

gregorbg:
Going online in coordination with WEAT.

PR #9760, opened 2024-08-11
Update WDC to WIC in visible Delegate Report contexts

nsilvestri:
This PR updates the WDC email to the WIC email when feedback is requested, and updates related strings on Delegate Reports that mention WDC. The intent of this change is to save WIC needing to be manually CC'd on all emails that request WDC feedback.

This does not try to update all instances of WDC to WIC; only the publicly visible ones. Underlying
variables referencing WDC or database migrations establishing team emails remain WDC, even in Delegate Report contexts.


nsilvestri:
nevermind; will be done in #9759

PR #9761, opened 2024-08-12
Show button to switch to new competition page overview

gregorbg:
Title says it all.

Query params are ported from the old to the new UI, so when you start setting up some search filters in the old UI and then switch to the new UI, the search state should persist.

### DIFF
--- a/app/assets/javascripts/competitions.js
+++ b/app/assets/javascripts/competitions.js
@@ -11,6 +11,9 @@ function resizeMapContainer() {
 }
 
 onPage('competitions#index', function() {
+  const queryParams = new URLSearchParams(window.location.search);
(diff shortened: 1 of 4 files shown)


gregorbg:
image

dunkOnIT:
Do we not want to launch the New UI into staging first and get some user feedback on it? This may buffer against a CF thread

dunkOnIT:
@thewca-bot deploy staging

dunkOnIT:
Have added a WCA_LIVE_SITE? flag, which will allow us to merge this into main and make it much easier to keep this feature on staging when deploying other branches. Happy for you to revert if you disagree. We can also JUST use WCA_LIVE_SITE? to determine whether or not to show the new UI, I don't have a strong preference.

PR #9767, opened 2024-08-13
Fix SASS compilation warnings

gregorbg:
Pretty much just followed the recommendation from the deprecation messages: The bllock of rules in `user_badge.scss` has been moved from below the nested block to above the nested block.

For the Sprockets warnings, there's not much we can do because it's bundled code of a third-party gem. I've locked the Dart SASS compiler at a version that will stay compatible with our monolith for the time being.


dunkOnIT (approved):
A blessing upon your house for this

PR #9772, opened 2024-08-13
Add seperate Deploy Main to Staging workflow

FinnIckler:
Introducing the asset precompilation PR, I accidentally made it so prod assets where compiled when using the deploy-on-release github action (Because WCA_LIVE_SITE was set to true when pushing to both staging and prod ECR). Luckily this only affected a couple of links for the registration service which failed when the jwt token didn't matched. 
I split up the deploy to staging and prod workflows which we wanted to do anyway.

### DIFF
--- a/.github/workflows/deploy-on-release.yml
+++ b/.github/workflows/deploy-on-release.yml
@@ -48,4 +48,3 @@ jobs:
       - name: Deploy Sidekiq
         run: |
           aws ecs update-service --cluster wca-on-rails --service wca-on-rails-prod-auxiliary-services --force-new-deployment
-        # There is no pipeline for staging so we manually force to update the image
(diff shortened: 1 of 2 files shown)


FinnIckler:
@thewca-bot deploy staging

FinnIckler:
Just overwriting the current wrong assets

.github/workflows/deploy-on-release.yml
```diff
@@ -51,6 +49,3 @@ jobs:
         run: |
           aws ecs update-service --cluster wca-on-rails --service wca-on-rails-prod-auxiliary-services --force-new-deployment
         # There is no pipeline for staging so we manually force to update the image
```
gregorbg, line 51: If you're removing the step then you should probably also remove the comment ;)

FinnIckler: oops done

PR #9774, opened 2024-08-13
Manually sync WIC to former Ethics Google Group

gregorbg:
Edge case that never happened before because we never (since becoming a nonprofit) experienced the case of one team absorbing another's duties.

I don't think this needs a custom implementation / generic support from the Groups/Roles models. This way of hard-coded semi-manual sync is fine for the forseeable future.


dunkOnIT (approved):
Not an area of the code I know a lot about, but looks reasonable to me

Issue #7994, opened 2023-06-12
Avatar admin page reject box should have a dropdown


danieljames-dj:
Currently in avatar admin page (https://www.worldcubeassociation.org/admin/avatars), if we click 'reject', it will have a text box to provide reason for rejection. Instead of that, it will be good if we have a dropdown with the list of guidelines (in english language) along with an 'other' option which when clicked the user can manually enter the rejection reason.

dunkOnIT:
@FinnIckler we should take this into account in the avatar rework

SAuroux:
I was also thinking about this more often recently as copy & pasting from the avatar guidelines on top of the page is getting increasingly annoying.

dunkOnIT:
No promises, but the odds are good that this will be worked on in the next month or so.

PR #9781, opened 2024-08-15
add checkboxes for avatar reject reasons

Brandon-Reichman:
When clicking reject on the admin avatar page, it displays a list of check boxes with the guidelines. These can be checked to inform the user which guideline their avatar violates. Along with a textbox for additional comments. Additional staff/delegate guidelines show up only under applicable users.

This alleviates the need to copy and pate the guidelines from the top of the page into the textbox.

Check boxes made more sense than a dropdown initially described in https://github.com/thewca/worldcubeassociation.org/issues/7994 as there are often multiple reasons given for rejection and a dropdown allowing multiple selections seemed less intuitive.

### DIFF
--- a/app/controllers/admin/avatars_controller.rb
+++ b/app/controllers/admin/avatars_controller.rb
@@ -16,8 +16,11 @@ module Admin
             user.approve_pending_avatar!
           when "reject"
             user.remove_pending_avatar = true
+            rejection_guidelines = args[:rejection_guidelines] || []
+            additional_reason = args[:rejection_reason].presence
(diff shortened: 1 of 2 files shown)


Brandon-Reichman:
Here is an example of the check boxes when rejecting a normal user.

Screenshot from 2024-08-15 20-42-53

And for a staff/delegate user.

Screenshot from 2024-08-15 20-50-20

After submitting the rejection with the checked reasons above I saw the email, in mailcatcher, with the proper reason.

Screenshot from 2024-08-15 20-50-49

gregorbg (approved):
Thanks, LGTM!

PR #9784, opened 2024-08-15
Enable HMR (Hot Module Reloading) for Webpacker Dev Server

gregorbg:
As per https://github.com/shakacode/shakapacker/blob/main/docs/react.md

### DIFF
--- a/config/shakapacker.yml
+++ b/config/shakapacker.yml
@@ -34,7 +34,7 @@ development:
     host: localhost
     port: 3035
     # Hot Module Replacement updates modules while the application is running without a full reload
-    hmr: false
+    hmr: true
     # Defaults to the inverse of hmr. Uncomment to manually set this.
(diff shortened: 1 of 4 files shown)


FinnIckler:
development.js won't actually be loaded without setting the NODE_ENV to development as I did in my PR

gregorbg:
As much as I appreciate you discovering that fix, your statement is not entirely correct.
Look at the source code of the `bin/shakapacker-dev-server` entry point:
```ruby
#!/usr/bin/env ruby

ENV["RAILS_ENV"] ||= "development"
ENV["NODE_ENV"]  ||= ENV["RAILS_ENV"]
```

FinnIckler:
Oh, I didn't look into there. That's ok then.

FinnIckler (approved):

PR #9785, opened 2024-08-16
V2 Live Feedback

FinnIckler:
- Reduced Padding significantly in the Registration Administration List by setting the Margin to 0
- Replaced Payment Status by Payment Amount and combined registered_on/paid_on like we do in V1
- Used the Link color for the Psych sheet headers and made the events inside not clickable
- Fixed the competitor list not loading when not logged in by not fetching the JWT Token before sending the request

### DIFF
--- a/app/webpacker/components/RegistrationsV2/RegistrationAdministration/AdministrationTableHeader.jsx
+++ b/app/webpacker/components/RegistrationsV2/RegistrationAdministration/AdministrationTableHeader.jsx
@@ -49,23 +49,25 @@ export default function TableHeader({
         >
           {i18n.t('common.user.representing')}
         </Table.HeaderCell>
-        <Table.HeaderCell
-          sorted={sortColumn === 'registered_on' ? sortDirection : undefined}
-          onClick={() => changeSortColumn('registered_on')}
-        >
-          {i18n.t('registrations.list.registered.without_stripe')}
-        </Table.HeaderCell>
-        {competitionInfo['using_payment_integrations?'] && (
-          <>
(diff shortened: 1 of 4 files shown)


FinnIckler:
Depends on this very small PR
https://github.com/thewca/wca-registration/pull/643

dunkOnIT:
@thewca-bot deploy staging

dunkOnIT:
@thewca-bot deploy staging

gregorbg (approved):

dunkOnIT:
### Payments Enabled

- [ ] Change column title "Paid" to "Amount"
- [ ] Show currency shortcode - $50 USD instead of $50 United States Dollar [organizers should be familiar with the currency code for the currency they've selected]
image

### Payments Not Enabled

Totals are not lined up with their columns:
- [ ] Payment amount total appears when there is no payment column at all
- [ ] All subsequent totals are one column later than they should be

image
image

### Psych sheets

I still don't think its obvious enough that the user needs to click on the event column header to get to the psych sheet, although the `link` styling definitely makes it more intuitive that they're clickable.

Perhaps the short-term solution is just to have an info banner that says "Click on an event icon in the header row to see the psych sheet for that event". It's not an elegant approach, but it is effective and for now effective is all we need.

dunkOnIT (changes requested):
See most recent comment for change requests

FinnIckler:
@thewca-bot deploy staging

dunkOnIT (approved):
All requested changes made :]

PR #9786, opened 2024-08-17
Improve timezone display in the Edit Schedule venue panel dropdown

gregorbg:
Export only a list of standardized ISO time zone IDs from the Ruby backend and let the browser handle all the localozation.
Previously, we were displaying English (human-readable) names straight from the Ruby `ActiveSupport` timezone backend.

## Before
image

## After
image


app/webpacker/lib/utils/timezone.js
```diff
@@ -0,0 +1,50 @@
+import { DateTime } from 'luxon';
+
+// Hijacks the JS `Intl` formatting code, which is nornally used to localize whole dates
```
kr-matthews, line 3:
```suggestion
// Hijacks the JS `Intl` formatting code, which is normally used to localize whole dates
```

kr-matthews (commented):
Is this order correct?
image
image
image
image
image

gregorbg:
Yes, this ordering is correct because you're seeing "local" time zones (ie those which match your competition's location) first, followed by all the other time zones. Both heaps of time zones are ordered west to east by themselves, but I've improved the display even further by including headers to make the separation more clear.

kr-matthews (approved):

PR #9787, opened 2024-08-17
Add a button to automatically detect venue timezones based on Lat/Lng

gregorbg:
There are plenty of libraries that can do time zone lookups out there, but most are either abandoned or carry a ~500MB JSON data file around. The one I chose now is relatively lightweight, but needs native Rust bindings which means we need an additional build tooling dependency in our Docker setups :/

Here's a short video clip of me choosing an obviously wrong time zone (Bangkok is not in Canada), and then using the new button to fix it

[recording_20240817-184943.webm](https://github.com/user-attachments/assets/3b4e9ab2-67b6-49fa-a9fa-9c08e7f9807f)


kr-matthews (commented):
Some kind of visual success/failure indication would be good. If the zone input doesn't change than it's unclear whether the value was already correct or it failed to find one.

Why are we introducing a button, rather than automatically calculating it or at least automatically re-calculating it whenever the coordinates change?

gregorbg:
I've adjusted the behavior a bit: The button now instantly loads and suggests a timezone after the coordinates have been changed. The user still needs to accept the suggestion, though

config/routes.rb
```diff
@@ -361,7 +361,8 @@
       get '/persons/:wca_id/results' => "persons#results", as: :person_results
       get '/persons/:wca_id/competitions' => "persons#competitions", as: :person_competitions
       get '/persons/:wca_id/personal_records' => "persons#personal_records", as: :personal_records
-      get '/geocoding/search' => 'geocoding#get_location_from_query', as: :geocoding_search
+      get '/geocoding/search' => 'geocoding#location_from_query', as: :geocoding_search
```
FinnIckler, line 364: Do we really want to keep these routes in the public API? Maybe a post request with a csrf check would be better

gregorbg: The geocoding routes are restricted via `before_action :raise_if_invalid`.
Otherwise, someone would have found out that it's possible to piggyback our Google Maps geocoder for free years ago

app/webpacker/components/EditSchedule/EditVenues/VenuePanel.js
```diff
@@ -148,6 +180,20 @@ function VenuePanel({
...
               onChange={handleVenueChange}
             />
+            <Button
+              floated="right"
+              compact
+              icon
+              labelPosition="left"
+              primary
+              negative={timeZonesError}
+              disabled={timeZonesLoading}
```
kr-matthews, line 190: Maybe additionally disable it if there's no best match?

kr-matthews (approved):

app/webpacker/components/EditSchedule/EditVenues/VenuePanel.js
```diff
@@ -121,7 +121,8 @@ function VenuePanel({
     isError: timeZonesError,
   } = useQuery({
     queryFn: fetchSuggestedTimeZones,
-    queryKey: ['suggested-tz', venue.id, venue.latitudeMicrodegrees, venue.longitudeMicrodegrees],
+    queryKey: ['suggested-tz', venue.latitudeMicrodegrees, venue.longitudeMicrodegrees],
+    enabled: venue.latitudeMicrodegrees && venue.longitudeMicrodegrees,
```
kr-matthews, line 125: Unlikely scenario in practice, but if one of the coordinates is 0 then this will be disabled, I think.

gregorbg:
I just merged main which used a tiny part of this PR to improve timezone communication between the backend and frontend.

This PR now completely and utterly focuses on adding a button, which is not blocking any other venue-related work. CC @kr-matthews :wink: 

(If there is a merge conflict, I will happily take responsibility to resolve it, especially seeing that this is "my" PR anyways)

gregorbg:
(setting this as a draft until we decided how we want to pull in the `clang` package without breaking people's environments)

PR #9790, opened 2024-08-17
Add current directory build context to docker-compose.yml (Fixes #9789)

### LINKED ISSUE #9789: Docker compose fails to start due to missing build context
**Describe the bug**
Attempting to run `docker-compose up` results in
```
ada@framework:~/RubymineProjects/worldcubeassociation.org$ docker-compose up -d
ERROR: The Compose file is invalid because:
Service wca_on_rails has neither an image nor a build context specified. At least one must be provided.
```
I'm running fedora 40, docker-compose 1.29.2, and podman 5.2.0.

0xAda:


dunkOnIT:
Thanks for the contribution! @FinnIckler would you be able to take a look at this?

FinnIckler:
I think this is a podman only issue. I don't think this breaks anything with docker though

FinnIckler (approved):
I'm approving this (because it works for docker as. well), but if you have any other issues that are podman related but work for the rest of the team, keep in mind that we officially only support docker

PR #9791, opened 2024-08-17
Change visibility for "Pending" incidents

Nevseros:
This should (I hope) change visibility for incidents that are not resolved yet -- basically changing the meaning from Pending/Resolved to Draft/Published. @CarterKoala

### DIFF
--- a/app/controllers/incidents_controller.rb
+++ b/app/controllers/incidents_controller.rb
@@ -37,7 +37,7 @@ class IncidentsController < ApplicationController
   def show
     set_incident
     unless @incident.resolved?
-      redirect_to_root_unless_user(:can_view_delegate_matters?)
+      redirect_to_root_unless_user(:can_manage_incidents?)
     end
   end
 


dunkOnIT:
@thewca-bot deploy staging

FinnIckler:
can you merge main first duncan? otherwise I'm sure this will break staging in some way

dunkOnIT:
@thewca-bot deploy staging

app/views/incidents/show.html.erb
```diff
@@ -24,7 +24,7 @@
   </p>
   <% warning_text = "" %>
   <% unless @incident.resolved_at %>
-    <% warning_text += "This incident is pending, a final ruling and public description may not be available.<br/>" %>
+    <% warning_text += "Note: This log has not been posted yet.<br/>" %>
```
gregorbg, line 27: In my opinion, this text "hides" too much of what's going on in the background. Something like "has not been resolved yet, so it is not publicly visible" seems like a fair compromise

gregorbg (changes requested):

app/views/incidents/show.html.erb
dunkOnIT: Changed!

gregorbg (approved):

PR #9795, opened 2024-08-19
Attempt to optimize group_member? a bit

danieljames-dj:
This is an attempt to optimize group_member? method a bit by passing group ID instead of full group.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -452,8 +452,8 @@ class User < ApplicationRecord
     preferred_locale || I18n.default_locale
   end
 
-  private def group_member?(group)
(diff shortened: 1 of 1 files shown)


gregorbg (commented):
I'm sorry to say that this "optimization" does absolutely nothing. You're still fetching the exact same groups as before, you're even passing the exact same groups to the exact same function as before.

The only difference is that you're accessing the ID now immediately _before_ you pass the group into the function, instead of letting the `group_member?` function access it for you. In terms of optimizing the SQL data fetching, this PR does absolutely nothing.

gregorbg:
Closing as NO-OP

PR #9799, opened 2024-08-19
Use a temporary database to reload Dev Dump without bringing the staging server down

FinnIckler:
This allows us to swap the develop database while the server is running. This is quite a lot of code duplication, but the control flow is quite different with a lot of database connection swapping. We can actually `SET foreign_key_checks=0` in RDS (only the `SET global` one's are not, my mistake from April).

We could also combine the two loading task with a reload parameter in the Rake task?


FinnIckler:
Tested and it works (locally)! @gregorbg feel free to merge the two tasks if you want while I am asleep ;)

gregorbg:
> We could also combine the two loading task with a reload parameter in the Rake task?

Yes, please do that :upside_down_face:

lib/tasks/db.rake
```diff
@@ -115,4 +115,109 @@ namespace :db do
...
+            DatabaseDumper.mysql("COMMIT", temp_db_name)
+          end
+
+          # RENAME Database has been removed, that's why we need to swap tables
+          LogTask.log_task "Swapping tables between databases" do
+            ActiveRecord::Base.connection.execute("CREATE DATABASE #{database_name}_old")
+            # Re-establish the connection to the old database so we can swap
+            ActiveRecord::Base.establish_connection(config)
+            # Get the list of tables from the current database using ActiveRecord
+            current_tables = ActiveRecord::Base.connection.execute("SHOW TABLES").map { |row| row[0] }
```
gregorbg, line 171: There is `ActiveRecord::Base.connection.tables` which does the exact same thing for the currently connected DB ;)

FinnIckler: ahhh ofc

FinnIckler:
Merged the tasks!

dunkOnIT (approved):
Tested locally and it works

FinnIckler:
I've deployed the DUMP_HOST Env variable via terraform so this can be safely merged now

PR #9805, opened 2024-08-21
Add script to connect to current prod/staging container 

FinnIckler:
I used the same script as in wca live, but added an environment parameter and some checks if the prerequisites are installed.

### DIFF
--- a/rel/console.sh
+++ b/rel/console.sh
@@ -76,18 +76,22 @@ if [ -z "$environment" ]; then
   usage
 fi
 
-if [[ "$environment" != "production" && "$environment" != "staging" ]]; then
-  printf "${COLOR_RED}Invalid environment: $environment. Must be 'production' or 'staging'.\n"
-  usage
-fi
+case "$environment" in
+  "production")
+    service_name="wca-on-rails-prod"
+    container_name="rails-production"
+  ;;
+  
+  "staging")
+    service_name="wca-on-rails-prod"
+    container_name="rails-production"
+  ;;
 
-# Set container name based on environment
-service_name="wca-on-rails-staging"
-container_name="rails-staging"
-if [ "$environment" = "production" ]; then
-  service_name="wca-on-rails-prod"
-  container_name="rails-production"
-fi
+  *)
+    printf "${COLOR_RED}Invalid environment: $environment. Must be 'production' or 'staging'.\n"
+    usage
+  ;;
+esac
 
 task_arn="$(
   aws ecs list-tasks \
--- a/rel/console.sh
+++ b/rel/console.sh
@@ -1,0 +1,97 @@
+#!/bin/bash
+set -euo pipefail
+
+# This script connects to one of the running ECS tasks and opens an
+# interactive rails shell attached directly to the running system.
+# Idea stolen from https://github.com/thewca/wca-live/blob/main/rel/console.sh
+# You will need to have AWS CLI + Systems Manager Addon installed and credentials configured
+# in a profile called 'wca'
+
+# Colors for output
+COLOR_DEFAULT='\033[0m'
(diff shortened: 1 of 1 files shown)


FinnIckler:
@jonatanklosko as you did the original script. Any comments/reviews?

rel/console.sh
```diff
@@ -0,0 +1,108 @@
...
+usage() {
+  printf "${COLOR_DEFAULT}Usage: $0 -e <environment>\n"
+  printf " -e <environment>  Specify the environment (production or staging)"
+  exit 1
+}
+
+# Check if AWS CLI is installed
+printf "${COLOR_DEFAULT}AWS CLI | "
+command -v aws >/dev/null 2>&1 && status="$?" || status="$?"
+if [[ "${status}" = 0 ]]; then
```
jonatanklosko, line 26: I believe you can do this:

```suggestion
if command -v aws >/dev/null 2>&1; then
```

rel/console.sh
```diff
@@ -0,0 +1,108 @@
...
+  printf "${COLOR_GREEN}OK ${COLOR_DEFAULT}(${smpVersion})\n"
+else
+  printf "${COLOR_RED}Missing\n"
+  exit 1;
+fi
+
+# Check if you are connected to the AWS WCA Account
+printf "${COLOR_DEFAULT}aws wca credentials | "
+account=$(aws iam list-account-aliases --query "AccountAliases[0]" --output text )
+if [[ "${account}" = "thewca" ]]; then
```
jonatanklosko, line 48:
```suggestion
if [[ "${account}" == "thewca" ]]; then
```

With `[[` you can use `==`, which is probably more intuitive. Alternatively, you can do `[ "${account}" = "thewca" ]`, which is sh-compatible. Either way is fine though!

rel/console.sh
```diff
@@ -0,0 +1,108 @@
...
+  usage
+fi
+
+# Set container name based on environment
+service_name="wca-on-rails-staging"
+container_name="rails-staging"
+if [ "$environment" = "production" ]; then
+  service_name="wca-on-rails-prod"
+  container_name="rails-production"
+fi
```
jonatanklosko, line 91: You can combine these:

```suggestion
case "$environment" in
  "production")
    service_name="wca-on-rails-prod"
    container_name="rails-production"
  ;;
  
  "staging")
    service_name="wca-on-rails-prod"
    container_name="rails-production"
  ;;

  *)
    printf "${COLOR_RED}Invalid environment: $environment. Must be 'production' or 'staging'.\n"
    usage
  ;;
esac
```

jonatanklosko (approved):
Just a few code suggestions, but LGTM!

rel/console.sh
jonatanklosko, line 1: Oh sorry, bad copy paste, good you caught it!

rel/console.sh
jonatanklosko: Oh sorry, bad copy paste, good you caught it!

rel/console.sh
```diff
@@ -0,0 +1,111 @@
...
+  printf "${COLOR_GREEN}OK ${COLOR_DEFAULT}(${cliVersion})\n"
+else
+  printf "${COLOR_RED}Missing\n"
+  exit 1
+fi
+
+# Check whether the Session Manager plugin exists
+printf "${COLOR_DEFAULT}Session Manager Plugin | "
+command -v session-manager-plugin >/dev/null 2>&1 && status="$?" || status="$?"
+if [[ "${status}" = 0 ]]; then
```
jonatanklosko, line 36: Same as above :D

```suggestion
if command -v session-manager-plugin >/dev/null 2>&1; then
```

jonatanklosko (approved):

PR #9807, opened 2024-08-22
Fix file URLs for the CDN and move TW Flag Overwrite to webpacker

FinnIckler:
Another two file urls that weren't prefixed with `asset_url`. I also moved the TW Flag overwrite from Sprockets to webpacker because I couldn't figure out why the `ASSET_HOST` was being ignored. Hoping we can move all the CSS to webpacker soon anyway!


FinnIckler:
tests are actually not failing, coveralls seems to be down

dunkOnIT (approved):

PR #9809, opened 2024-08-23
Add an internal route so that v2 reg can query unannounced competitions

dunkOnIT:
This will allow organizers to pre-register before competitions are announced, which is a request which has come through after we hid the Register button before announcement

### DIFF
--- a/app/controllers/api/internal/v1/competitions_controller.rb
+++ b/app/controllers/api/internal/v1/competitions_controller.rb
@@ -24,9 +24,21 @@ class Api::Internal::V1::CompetitionsController < Api::Internal::V1::ApiControll
     end
   end
 
-  private def competition_from_params
+  def qualifications
+    competition = competition_from_params(associations: [:competition_events])
+
+    qualifications = competition.competition_events
+                                .where.not(qualification: nil)
+                                .index_by(&:event_id)
+                                .transform_values(&:qualification)
+                                .transform_values(&:to_wcif)
+
+    render json: qualifications
+  end
+
+  private def competition_from_params(associations: {})
     id = params[:competition_id]
-    competition = Competition.find_by_id(id)
+    competition = Competition.includes(associations).find_by_id(id)
 
     raise WcaExceptions::NotFound.new("Competition with id #{id} not found") unless competition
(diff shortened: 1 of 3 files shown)


dunkOnIT:
@thewca-bot deploy staging

app/controllers/api/internal/v1/competitions_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+                 on_the_spot_registration on_the_spot_entry_fee_lowest_denomination qualification_results event_restrictions
+                 base_entry_fee_lowest_denomination currency_code allow_registration_edits allow_registration_self_delete_after_acceptance
+                 allow_registration_without_qualification refund_policy_percent use_wca_registration guests_per_registration_limit venue contact
+                 force_comment_in_registration use_wca_registration external_registration_page guests_entry_fee_lowest_denomination guest_entry_status
+                 information events_per_registration_limit],
+        methods: %w[url website short_name city venue_address venue_details latitude_degrees longitude_degrees country_iso2 event_ids registration_currently_open?
+                    main_event_id number_of_bookmarks using_payment_integrations? uses_qualification? uses_cutoff? competition_series_ids],
+        include: %w[delegates organizers tabs],
+      }
+      render json: competition.as_json(options)
```
FinnIckler, line 23: Do we not have to_competition_info for this?

dunkOnIT: Ah cool! Wasn't aware of this as the v0 competition API doesn't make use of it either - but have updated both to use `to_competition_info`

FinnIckler: You should check if it has all the fields that we want because we might have added one since I created it

dunkOnIT: I did a detailed check, and `to_competition_info` had more information than the old method - so we're definitely not losing any fields. I didn't save the comparison so can't remember exactly what the field was, but I didn't see any issue with exposing it on the non-authenticated route as well.

FinnIckler (approved):
LGTM

PR #9811, opened 2024-08-24
Edit Profile Form

danieljames-dj:
Currently the edit profile form is merged along with the main contact form, but this is making the main contact form increasingly complex. This PR is an attempt to make 'edit profile' portion of the form into it's own separate form.

~Temporarily, this particular form is not linked anywhere else in website, this can be accessed only if people know the URL. Once this is merged, I'll raise another PR to remove everything related to 'edit profile' from main contact form and instead if the user selects 'edit profile', then will give the user the button to redirect to 'edit profile form' page.~

EDIT: This is now linked with the main contact form.


app/webpacker/components/ContactEditProfilePage/index.jsx
```diff
@@ -0,0 +1,53 @@
...
+    );
+  }
+
+  return (
+    <Container text>
+      <Header as="h2">{i18n.t('page.contact_edit_profile.title')}</Header>
+      <EditProfileFormWithWcaId
+        wcaId={wcaId}
+        loggedInUserData={loggedInUserData}
+        setContactSuccess={setContactSuccess}
```
kr-matthews, line 48: The fact that the parent here is storing a boolean state is irrelevant to the child; the parent could navigate to a different component, for example. Note that the child never needs to (and shouldn't) call `setContactSuccess(false)`.
```suggestion
        onContactSuccess={() => setContactSuccess(true)}
```

app/webpacker/components/ContactEditProfilePage/EditProfileFormWithWcaId.jsx
```diff
@@ -0,0 +1,126 @@
...
+  wcaId: loggedInUserWcaId,
+  loggedInUserData,
+  setContactSuccess,
+  recaptchaPublicKey,
+}) {
+  const [contactEmail, setContactEmail] = useState();
+  const [customWcaId, setCustomWcaId] = useState();
+  const [editProfileReason, setEditProfileReason] = useState();
+  const [editedProfileDetails, setEditedProfileDetails] = useState();
+  const [profileDetailsChanged, setProfileDetailsChanged] = useState(false);
```
kr-matthews, line 23: If a const can be calculated, rather than stored as state, then that's almost always better (via `useMemo` if necessary).
```suggestion
  const profileDetailsChanged = editedProfileDetails && !_.isEqual(editedProfileDetails, profileDetails);
```

That doesn't quite work though, since `profileDetails` is stuck in the child. So I would hoist that (the query/const) up to this component - state should always be at the highest level that it's needed, and it's "needed" here to calculate `profileDetailsChanged`.

If hoisting that up would cause problems, then (in my experience) that means that the components aren't properly thought out, and need to be adjusted.

app/webpacker/components/ContactEditProfilePage/index.jsx
danieljames-dj: Done

app/webpacker/components/ContactEditProfilePage/EditProfileFormWithWcaId.jsx
danieljames-dj: Actually it was not easy, so I finally decided to merge both `EditProfileFormWithWcaId` and `EditProfileForm` into `EditProfileForm`. Now, I guess this problem is solved.

The reason why I splitted them was I was trying to make the react file small. But it was too difficult to manage states, so I thought merging them. I hope this is good, let me know if you have any changes.

As usual, thanks a lot for your review and your work in understanding the code and giving relevant comments.

danieljames-dj:
@thewca-bot deploy staging

maxidragon:
@thewca-bot deploy staging

maxidragon:
(If anyone is wondering why I tried to trigger the deployment, Daniel allowed me to try to check if everyone has access to trigger it)

app/controllers/contacts_controller.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class ContactsController < ApplicationController
+  PERSON_FIELDS_EDIT_REQUESTABLE = [:name, :country_iso2, :gender, :dob].freeze
```
gregorbg, line 4: This should probably(? feel free to argue) be a part of `person.rb`, because that's the model you're actually trying to edit

app/controllers/contacts_controller.rb
```diff
@@ -37,7 +39,7 @@ class ContactsController < ApplicationController
   end
 
   private def new_profile_data_key_to_value(new_profile_data, profile_data_to_change)
-    if profile_data_to_change == 'country'
+    if ['country', 'country_iso2'].include?(profile_data_to_change)
```
gregorbg, line 42: How can both variants be possible? Shouldn't you make sure that the frontend is consistent enough to focus on either one?

(Or am I correct in guessing that one is from legacy Ruby templates and the other is from React?)

app/controllers/contacts_controller.rb
```diff
@@ -99,6 +101,52 @@ def contact
...
+    attachment = params[:attachment]
+    wca_id = formValues[:wcaId]
+    if current_user.present?
+      profile_to_edit = {
+        name: current_user.person.name,
+        country_iso2: current_user.person.country_iso2,
+        gender: current_user.person.gender,
+        dob: current_user.person.dob,
+      }
+    else
```
gregorbg, line 118: Would it not be prudent to perform some kind of sanity checking whether the currently logged in user has actually claimed the WCA ID they're trying to edit?

Right now, you're just happily loading the current user's details without any check, so I could impersonate anybody by simply entering their WCA ID

app/controllers/contacts_controller.rb
```diff
@@ -99,6 +101,52 @@ def contact
...
+      }
+    end
+    changes_requested = PERSON_FIELDS_EDIT_REQUESTABLE
+                        .reject { |field| profile_to_edit[field].to_s == edited_profile_details[field].to_s }
+                        .map { |field|
+                          {
+                            field: field.to_s.humanize,
+                            from: (new_profile_data_key_to_value(profile_to_edit[field], field.to_s) || "Unknown").to_s,
+                            to: new_profile_data_key_to_value(edited_profile_details[field], field.to_s),
+                          }
```
gregorbg, line 134: I veguely remember that you had a utility class for this kind of "from-to-change" in the Roles mailer. Can you reuse that here as well?

app/controllers/contacts_controller.rb
```diff
@@ -99,6 +101,52 @@ def contact
...
+                        }
+
+    maybe_send_contact_email(
```
gregorbg, line 137: What is the `maybe` part about? In which cases do we not send an email?

app/views/mail_form/contact_edit_profile.erb
gregorbg, line 1: Suggestion for the template: If there is an attachment present in the code, add a paragraph that mentions "There is an attachment"

app/webpacker/components/ContactEditProfilePage/EditProfileForm.jsx
```diff
@@ -0,0 +1,185 @@
...
+  const [captchaValue, setCaptchaValue] = useState();
+  const [captchaError, setCaptchaError] = useState(false);
+  const [saveError, setSaveError] = useState();
+  const { save, saving } = useSaveAction();
+  const wcaId = loggedInUserWcaId || customWcaId;
+  const { data, isLoading, isError } = useQuery({
+    queryKey: ['profileData'],
+    queryFn: () => fetchJsonOrError(apiV0Urls.persons.show(wcaId)),
+  }, CONTACT_EDIT_PROFILE_FORM_QUERY_CLIENT);
+  const profileDetails = data?.data?.person;
```
gregorbg, line 50: More line breaks for readability please :face_with_spiral_eyes:

app/webpacker/components/ContactEditProfilePage/EditProfileForm.jsx
```diff
@@ -0,0 +1,185 @@
...
+    save(
+      contactEditProfileActionUrl,
+      formData,
+      onContactSuccess,
+      { method: 'POST', headers: {}, body: formData },
+      setSaveError,
+    );
+  };
+
+  const handleEditProfileReaconChange = (e, { value }) => {
```
gregorbg, line 71: Typo: `Reacon` -> `Reason`

app/webpacker/components/ContactEditProfilePage/EditProfileForm.jsx
```diff
@@ -0,0 +1,185 @@
...
+  wcaId: loggedInUserWcaId,
+  loggedInUserData,
+  onContactSuccess,
+  recaptchaPublicKey,
+}) {
+  const [contactEmail, setContactEmail] = useState();
+  const [customWcaId, setCustomWcaId] = useState();
+  const [editProfileReason, setEditProfileReason] = useState();
+  const [editedProfileDetails, setEditedProfileDetails] = useState();
+  const [proof, setProof] = useState();
```
gregorbg, line 40: Maybe `proofAttachment`?

app/webpacker/components/ContactEditProfilePage/EditProfileForm.jsx
```diff
@@ -0,0 +1,185 @@
...
+            name="dob"
+            control={UtcDatePicker}
+            showYearDropdown
+            dateFormatOverride="YYYY-MM-dd"
+            dropdownMode="select"
+            isoDate={editedProfileDetails?.dob}
+            onChange={(date) => handleFormChange(null, {
+              name: 'dob',
+              value: date,
+            })}
```
gregorbg, line 143: Can you put this into a separate callback please?
In general, `onChange` and similar React component callbacks should only be written in-line if they can (comfortably) fit on one line

app/webpacker/components/ContactEditProfilePage/EditProfileForm.jsx
```diff
@@ -0,0 +1,185 @@
...
+  if (saving || isLoading) return <Loading />;
+  if (saveError || isError) return <Errored />;
+
+  return (
+    <Form onSubmit={formSubmitHandler}>
+      {!loggedInUserData && (
+        <Form.Input
+          label={i18n.t('page.contact_edit_profile.form.contact_email.label')}
+          value={contactEmail}
+          onChange={(e, { value }) => setContactEmail(value)}
```
gregorbg, line 92: If you declare `contactEmail` with our WCA custom `useInputState` instead of `useState`, you can pass the `setContactEmail` setter here directly and it will "just work" :magic_wand:

app/webpacker/components/ContactEditProfilePage/EditProfileForm.jsx
```diff
@@ -0,0 +1,185 @@
...
+          required
+        />
+      )}
+      {!loggedInUserWcaId && (
+        <Form.Field
+          control={IdWcaSearch}
+          name="wcaId"
+          label={i18n.t('page.contact_edit_profile.form.wca_id.label')}
+          value={customWcaId}
+          onChange={(e, { value }) => setCustomWcaId(value)}
```
gregorbg, line 102: `useInputState` for direct setter access, see above

app/webpacker/components/ContactEditProfilePage/EditProfileForm.jsx
```diff
@@ -0,0 +1,185 @@
...
+          <Message
+            error
+            content={i18n.t('page.contact_edit_profile.form.captcha.validation_error')}
+          />
+        )}
+      </Form.Field>
+      <Form.Button
+        type="submit"
+        disabled={
+          !editedProfileDetails || _.isEqual(editedProfileDetails, profileDetails) || !captchaValue
```
gregorbg, line 178: Please put at least the `_.isEqual` check, or potentially even the whole `||` chain, into its own `useMemo` variable

app/webpacker/components/ContactsPage/SubForms/Wrt/index.jsx
```diff
@@ -21,15 +21,18 @@ export default function Wrt() {
...
+    if (!selectedQueryType || selectedQueryType === QUERY_TYPES_MAP.editProfile) return null;
+    return OtherQuery;
   }, [selectedQueryType]);
 
+  useEffect(() => {
+    if (selectedQueryType === QUERY_TYPES_MAP.editProfile) {
+      dispatch(setFormRedirection(SECTION, contactEditProfileActionUrl));
+    } else {
+      dispatch(setFormRedirection(SECTION, null));
+    }
```
gregorbg, line 33: If you calculate the redirect first, and then send one dispatch with the result, you can save yourself from copy-pasting the same `dispatch` structure multiple times.

app/webpacker/components/ContactsPage/SubForms/Wrt/index.jsx
```diff
@@ -21,15 +21,18 @@ export default function Wrt() {
...
-      case QUERY_TYPES_MAP.editProfile:
-        return EditProfileQuery;
-      default:
-        return OtherQuery;
-    }
+    if (!selectedQueryType || selectedQueryType === QUERY_TYPES_MAP.editProfile) return null;
+    return OtherQuery;
   }, [selectedQueryType]);
 
+  useEffect(() => {
```
gregorbg, line 28: I'm not a big fan of `useEffect`. You're trying to react to changes to `selectedQueryType`, so where is this variable being set/written, i.e. where are those changes originally fired? And can you just do what you're doing right here wherever the changes are triggered/fired in the first place?

gregorbg (changes requested):

app/controllers/contacts_controller.rb
danieljames-dj: Oh yes, created similar thing for this as well.

app/controllers/contacts_controller.rb
danieljames-dj: This method was named so from very long time - I think even before I started working on contact form. The `may` is not relevant much because it fails only when there is a server error or validation error, but since it's used in many places I didn't change it. But I can change all of them all-together in any of the future PRs if needed.

app/views/mail_form/contact_edit_profile.erb
danieljames-dj: Okay done, thanks for the suggestion.

app/webpacker/components/ContactEditProfilePage/EditProfileForm.jsx
danieljames-dj: Changed

app/webpacker/components/ContactEditProfilePage/EditProfileForm.jsx
danieljames-dj: Done

app/webpacker/components/ContactEditProfilePage/EditProfileForm.jsx
danieljames-dj: Oops sorry for the typo, changed.

app/webpacker/components/ContactEditProfilePage/EditProfileForm.jsx
danieljames-dj: Done

app/webpacker/components/ContactEditProfilePage/EditProfileForm.jsx
danieljames-dj: Done

app/webpacker/components/ContactsPage/SubForms/Wrt/index.jsx
danieljames-dj: I think I understood what you meant, and I did it in ContactForm.jsx. Please let me know if this is not what you meant.

app/webpacker/components/ContactsPage/SubForms/Wrt/index.jsx
danieljames-dj: Removed useEffect

app/controllers/contacts_controller.rb
gregorbg: Oh, I didn't realize that. Thought you came up with this method.

But then nevermind, never touch a running system :sweat_smile:

gregorbg (approved):

PR #9815, opened 2024-08-26
Gear order form in WCA website

danieljames-dj:
In this PR, the gear order form is brought within WCA website so that there will be more options of customizations - for example, the notification emails can be well formatted in a way that it will reduce the workload of WFC, the user can know the total order value while filling the form, etc.

### DIFF
--- a/app/webpacker/components/ContactGearOrderPage/GearOrderForm.jsx
+++ b/app/webpacker/components/ContactGearOrderPage/GearOrderForm.jsx
@@ -1,0 +1,201 @@
+/* eslint-disable max-len */
+import _ from 'lodash';
+import React, { useEffect, useState } from 'react';
+import {
+  Form, Header, HeaderSubheader, Input, Message, Table,
+  TableHeader,
+  TableHeaderCell,
+  TableRow,
+} from 'semantic-ui-react';
+import useSaveAction from '../../lib/hooks/useSaveAction';
+import Loading from '../Requests/Loading';
+import Errored from '../Requests/Errored';
+import { contactGearOrderUrl } from '../../lib/requests/routes.js.erb';
+import i18n from '../../lib/i18n';
+
+export default function GearOrderForm({ equipments }) {
+  const [orderDetails, setOrderDetails] = useState(equipments.map(() => ({
+    quantity: 0,
+    price: 0,
+  })));
+  const [totalPrice, setTotalPrice] = useState(0);
+  const [formValues, setFormValues] = useState({});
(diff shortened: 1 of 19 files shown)


danieljames-dj:
This form is not 100% complete, currently I'm facing a blocker that if I mark any of the `Form.Input` as `required`, it will become 'red' even before user starts inputting any details. The same issue is happening with competition form as well if I give required.

I remember seeing this while working on Sign Up page migration, but iirc I couldn't find a solution during that time.

I request somebody to unblock this issue. Alternative is to use custom validation to check if all the fields are there.

dunkOnIT:
> it will become 'red' even before user starts inputting any details

Currently there is no workaround for this issue - that's how the form behaves in [v2 Register](https://staging.worldcubeassociation.org/competitions/CiuLaManchaOpen2024/register) as well. I think for now, proceed as if this were not a blocker.

app/webpacker/components/ContactGearOrderPage/index.jsx
```diff
@@ -0,0 +1,57 @@
+import React from 'react';
+import { Container, Header } from 'semantic-ui-react';
+import GearOrderForm from './GearOrderForm';
+import i18n from '../../lib/i18n';
+
+export default function ContactGearOrder() {
+  const equipments = [
```
kr-matthews, line 7: "Equipment" doesn't have a plural form.
And this array could be moved out of the component.
```suggestion
  const equipment = [
```

app/webpacker/components/ContactGearOrderPage/GearOrderForm.jsx
```diff
@@ -0,0 +1,170 @@
...
+  TableHeaderCell,
+  TableRow,
+} from 'semantic-ui-react';
+
+export default function GearOrderForm({ equipments }) {
+  const [orderDetails, setOrderDetails] = useState(equipments.map(() => ({
+    quantity: 0,
+    price: 0,
+  })));
+  const [totalPrice, setTotalPrice] = useState(0);
```
kr-matthews, line 16: Does this work? Direct calculation is better than a state and an effect.
```suggestion
  const totalPrice = _.sumBy(orderDetails, 'price');
```

kr-matthews (changes requested):

danieljames-dj:
I'm closing this for now. WFC is currently relying on the spreadsheet, will check if bringing this to website will improve the process rapidly. If yes, will restart it.

Issue #9788, opened 2024-08-17
New Competition UI Feedback & Changes


dunkOnIT:
See CF thread here: https://www.facebook.com/groups/cyoubxfriends/posts/9026899190669787/

### Further Research
- Search by announcement date may be broken?

### General Comments
- The new view significantly reduces the number of comps you can see without scrolling, and this is highly undesirable
  - Too much spacing in table rows
  - The filters take up too much space
- On mobile, competition rows take up too much space, but also their text is too small/flag icons are too small [compare old and new mobile views to see that old mobile view definitely is more compact and readable]

#### UI Simplifications
- [ ] Remove "search by delegate" entirely, or only have it available in Admin View
- [ ] Remove "show cancelled competitions", or only have it available in Admin View
- [ ] Display the registration icon by default, and remove the toggle

#### Minor Changes
- [ ] Sort years in descending order [so that most recent years are closest to the top]
- [ ] BUG: If you've already selected a region, clicking the dropdown arrow to see the list again (but NOT picking anything), resets the region to "All regions" but displays a blank in the region box. 

### Feature Requests
- [ ] Show events that the comp is hosting in the competition table [perhaps consider using the event toggle that we have in the Registration Admin list?]
- [ ] Consider improved defaults around what competitions are displayed - eg, allow the user to select a region to display comps from and remember that in localstorage
- [ ] Add a true search feature that doesn't just filter loaded competitions, but hits the backend with an API request to return all comps that might match the search criteria

kr-matthews:
I'll start working on some of these.

kr-matthews:
> * Search by announcement date may be broken?

The legacy search-by-announcement omits competitions which are over, whereas the new one includes _all_ competitions. Besides that, they function the same as far as I can tell when playing around with it.

Is that a problem? Seems fine to me.

kr-matthews:
We can save some space by putting the "All" and "Clear" buttons in the same row as all the other buttons, and also shrinking the event boxes a bit.

![image](https://github.com/user-attachments/assets/89633f79-9222-467a-9a0a-139ecdfc8aec)

But before doing any of that, why do we have  separate event selectors for this vs the rest of the website (e.g. selecting events in registration). I don't want to make changes to this selector, just to have it removed and replaced with the other one. This one is much clearer about which events are selected vs no selected.

kr-matthews:
[...]

PR #9816, opened 2024-08-27
New Competition Search UI | various improvements

kr-matthews:
Addresses a few issues from #9788 - see commit list.
~The region dropdown bug isn't fully fixed - I left details about that on the linked issue.~

~Edit: not sure how to best resolve the cyclical dependency. Should I move this (or all) filter constants to a new file?~


kr-matthews:
@gregorbg Any insight on the drop-down icon issue? From the linked issue:

> The dropdown is 'clearable', so when something is selected, there should be an "x" icon to clear it. The component does this via class="dropdown icon clear". For some reason, that's not working, and the dropdown arrow icon is still showing up. Possibly we're missing semantic-ui-css, or need to update semantic-ui-react? I'm stuck here.

gregorbg:
That was actually an interesting issue to debug. I checked that we are definitely using an up-to-date version of the stylesheets, but we're using **Fomantic CSS** (a maintained fork) whereas the clearable feature lives in **Semantic CSS**.

The clearable feature must have been one of the last things they added around the time it was forked, so the CSS styles are missing in Fomantic stylesheets. I have now manually ported them over and pushed a commit, which should fix the isse.

kr-matthews:
> I have now manually ported them over and pushed a commit, which should fix the isse.

Running locally, I see nothing (as opposed to the "x" I should, or the "v" drop-down arrow I was before).

image

I have been having trouble with docker recently, so maybe it's just my setup.

gregorbg:
This most likely means that your browser is caching some stylesheets. Try hitting Ctrl+Shift+R (or the Mac equivalent, which I think is Cmd+Shift+R)

PR #9822, opened 2024-08-28
Optimize competition pages for competitions that use v2

FinnIckler:
There was a lot of unnecessary load surfaced by the load test caused by these calls

- Stopps making multiple calls to the registration service for count if used multiple times by caching for a few seconds
- Doesn't call the registration service on every page load just because it needs to know if we should show the competitors tab (I switched this to `unless @competition.registration_not_yet_opened?` and I honestly think we could even just remove this check entirely 
- Uses the much faster cached `Microservices::Registrations.competitor_count_by_competition(id)` route for `any_registrations?`, which was only called in `_nav`, so we could also remove it if we don't think it'll be useful somewhere else

### DIFF
--- a/app/models/competition.rb
+++ b/app/models/competition.rb
@@ -979,7 +979,7 @@ class Competition < ApplicationRecord
 
   def any_registrations?
     if uses_new_registration_service?
-      self.microservice_registrations.any?
+      Microservices::Registrations.competitor_count_by_competition(id) > 0
     else
       self.registrations.any?
     end
--- a/app/views/competitions/_nav.html.erb
+++ b/app/views/competitions/_nav.html.erb
@@ -142,7 +142,7 @@
             }
           end
 
-          if @competition.any_registrations?
+          unless @competition.registration_not_yet_opened?
             event_icons = @competition.events.map do |event|
               { text: event.id, path: competition_psych_sheet_event_path(@competition, event.id), cubing_icon: event.id, title: event.name }
             end unless @competition.uses_new_registration_service?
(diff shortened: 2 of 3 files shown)


dunkOnIT:
> and I honestly think we could even just remove this check entirely

Hard no - organizers REALLY don't like users being able to see the competitor list before registration is open - ESPECIALLY if the organizers have pre-registered

FinnIckler:
That point is moot though, because the current check `registrations_any?` also shows the competitor list before the registration is open as long as the organizers have pre registered

dunkOnIT:
> That point is moot though, because the current check registrations_any? also shows the competitor list before the registration is open as long as the organizers have pre registered

Ah I see! Until checking the diff I thought we had both the competitor count and the registration open check, and you'd just removed the competitor count. Long story but I had a PR in the works to add a check for if registration had opened, which then got split up and honestly I thought I'd merged it but clearly I hadn't :P The intention was very much to move to the check you have now implemented, because this concern from organizers was a blocker for moving forward with auto-accepts

FinnIckler:
> > That point is moot though, because the current check registrations_any? also shows the competitor list before the registration is open as long as the organizers have pre registered
> 
> Ah I see! Until checking the diff I thought we had both the competitor count and the registration open check, and you'd just removed the competitor count. Long story but I had a PR in the works to add a check for if registration had opened, which then got split up and honestly I thought I'd merged it but clearly I hadn't :P The intention was very much to move to the check you have now implemented, because this concern from organizers was a blocker for moving forward with auto-accepts

I think that was the register tab you changed

dunkOnIT (approved):

PR #9823, opened 2024-08-28
Split up Server into Renderer/API for Staging

FinnIckler:
Starting point see if this works. After testing I will need to split it up into 2 services, because like this the API server will still die if the Rendering server dies because they are part of the same service.


FinnIckler:
mh.. my deploy idea didn't work because it can't access the secrets. I get it, it's in my PR and I could manipulate it, but not sure how to develop these changes well then

FinnIckler:
Even putting secrets into my own branch doesn't work

FinnIckler:
looks like I need to reopen this PR directly from the repo not from a fork

PR #9831, opened 2024-08-30
Update Yarn to 4.4.1

gregorbg:
Also running `yarn dedupe` for convenience.


gregorbg:
Merging as dependency update

PR #9854, opened 2024-09-01
Bump Shakapacker to 8.0.2

gregorbg:
Sigh...

See #9845 and #9842


gregorbg:
Merging as dependency update

PR #9895, opened 2024-09-02
Bump Ruby to version 3.3.5

gregorbg:
Also gets rid of the Bundler and Yarn dependency volumes, which have been causing headaches every time a Ruby version update (or any Docker rebuild, for that matter) rolls around.

See https://stackoverflow.com/questions/77067553/using-gem-volume-breaks-for-native-dependency-updates-in-docker-compose


gregorbg:
Treating as a glorified dependency update

PR #9896, opened 2024-09-02
Load secret key directly instead of secrets.yml file

gregorbg:
Due to deprecations that we were never properly made aware of (grrr Rails...) we have to switch the way `secret_key_base` is being set.

Abridged snippet from the Rails source code:

```ruby
if Rails.env.local? || ENV["SECRET_KEY_BASE_DUMMY"]
  config.secret_key_base ||= generate_local_secret
else
  validate_secret_key_base(ENV["SECRET_KEY_BASE"] || credentials.secret_key_base)
end
```

This means that for local environments, we have to set `config.secret_base_key` explicitly, and the environment variable won't work. In production however, we have to set the environment variable, because setting `config.secret_key_base` won't work.

Someone on the Rails team is trying _really_ hard to dictate how people have to encrypt their secrets, which is worrying me.


gregorbg:
Confirmed to be working on staging (was logged in before deploy, am still logged in after deploy. Confirmed that the PR code has actually been deployed via `/server-status`)

FinnIckler (approved):
LGTM

PR #9903, opened 2024-09-03
enable puma

FinnIckler:
I've been looking at a lot of puma.rb in the last hour or so and just adding the worker count and the development `worker_timeout` (when you are using a debugger) should be everything we need to change from the defaults


FinnIckler:
@thewca-bot deploy staging

gregorbg (commented):
Hm, why are you adding back all these values when the Rails 7.2 update process just recently deleted them?

I'm not saying you're wrong, but In curious as to why Rails would kick these out when it seems like they're actually necessary?!

FinnIckler:
Well the values are actually still needed, rails just wants us to add them via environment variables. At least that's how I understood it. We can do that, but not sure if that is actually cleaner

gregorbg:
Hm, where do these environment variables come from (actual Puma documentation or weird Rails quirk) and where is this documented?

FinnIckler:
If you look at the commit history of the puma.rb.tt it says: https://github.com/rails/rails/commit/142e6ab2c1e3cb7df4b82e20e2a1cd676b755714 

But then we can't do that dynamically depending on the processors right?

FinnIckler:
Also just read this on the puma repo
Many configuration options and Puma features are not available when using rails server. It is recommended that you use Puma's executable instead:

gregorbg:
Hm. I don't think we need to dig so deep in the beginning. Maybe we can switch to the Puma executable once the basic setup works and we start to consider fine-tuning.

Regarding the Rails commit, it looks a bit weird to me... They had an `auto` option in the Puma config file, then they argued that Puma 5 now reads this env variable directly, but only for integers and not for `auto`. Then why the heck did Rails remove their own support for `auto` if there is no replacement in Puma??!

Am I missing something? If not, can you maybe follow up with the Rails team?

FinnIckler:
https://github.com/puma/puma/discussions/3096 This confuses me even more. 
So they say that there is no difference if you config it all using puma.rb, but rails removed most of the config from it?

FinnIckler:
@thewca-bot deploy staging

dunkOnIT (approved):

PR #9906, opened 2024-09-04
Realigning error codes for V2

FinnIckler:
We merged some codes, moved one around and added two new ones


dunkOnIT (changes requested):
It's missing codes -4003, -4004, -4005, otherwise LGTM

FinnIckler:
it's not, if you look at the full files and not just the diff, they are there

dunkOnIT (approved):

FinnIckler:
@jonatanklosko Is this the correct way to move a translation from one key to the other, or will this break internationalize?

FinnIckler:
Cool, it wasn't so this should be fine

PR #9907, opened 2024-09-04
Refetch data on registration v2 update 

FinnIckler:
Because moving people to and from the waiting list causes the waiting list position to change for potentially everyone on the waiting list we need to refetch the data.


dunkOnIT:
@thewca-bot deploy staging

FinnIckler:
I'm gonna change this to refetching the data on every update. This is because Organizers expect all the data to be fresh after they have updated (i.e the payment status changed, new registration be listed etc.), currently they need to manually refresh the page to get the latest data. This isn't super important when the competition has been open for a while, but is important at the start of a competition opening. 
We can reconsider using websockets/graphql updates for this once we get into the weeds with wca-live

dunkOnIT (approved):
LGTM as far as I can follow React code

PR #9911, opened 2024-09-05
Add Rejected Status to V2

FinnIckler:
Adds the option to reject a competitor from the frontend and explains the difference between cancelled and rejected registrations. 

Do we want to do a separate email for rejected? Or do we depend on organizers to notify them with the reason.


dunkOnIT (approved):
Can't comment on React code but functionality is tested

PR #9912, opened 2024-09-05
Fix Sorting by WCA id not working in v2 if newcomers registered

FinnIckler:
Sorts so newcomers are on the top


kirjavascript (approved):

dunkOnIT (approved):
Tested locally and it works as expected

PR #9913, opened 2024-09-06
Fixes V2 Registration Message to move to the middle when scrolling

FinnIckler:
Moves the Sticky out of the RegistrationMessage (so it can be used without it like in the Register Step) and moves the ref to the top. I think the issue was that the div that contained the sticky changed size or something...


dunkOnIT (approved):
Functionality works as expected - can't comment on the code

PR #9914, opened 2024-09-07
Show total delegated count only if it exists

danieljames-dj:
Currently if the number of delegated competitions is not available, it's showing empty in the email. This PR is to show the number of delegated competitions only if the value is actually there.


gregorbg (commented):
As an alternative suggestion, `my_hash.compact` will throw away all entries where the value is `nil`. So you could just pass `metadata.compact` to the mailer and that takes care of the problem without the explicit `metadata` check

danieljames-dj:
> As an alternative suggestion, `my_hash.compact` will throw away all entries where the value is `nil`. So you could just pass `metadata.compact` to the mailer and that takes care of the problem without the explicit `metadata` check

Okay done

gregorbg (approved):

PR #9915, opened 2024-09-07
Form bug fix and adding WCA ID

danieljames-dj:
This PR does two things:
1. Fixes a bug: Changed `profile_data_to_change.titleize` to `profile_data_to_change&.titleize`. This is because if any query type other than profile_data_change is selected, it will crash. This is now fixed.
2. Added WCA ID to contact form email for WRT. (This has small style issue - in the email it shows 'Wca' instead of 'WCA ID', I didn't work to fix that, instead will fix it later, because having WCA ID in contact form email will be really helpful for WRT.

### DIFF
--- a/app/controllers/contacts_controller.rb
+++ b/app/controllers/contacts_controller.rb
@@ -50,8 +50,9 @@ class ContactsController < ApplicationController
       ContactWrt.new(
         name: requestor_details[:name],
         your_email: requestor_details[:email],
+        wca_id: User.find_by(email: requestor_details[:email])&.wca_id || 'None',
         query_type: contact_params[:queryType].titleize,
-        profile_data_to_change: profile_data_to_change.titleize,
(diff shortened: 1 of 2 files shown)


app/controllers/contacts_controller.rb
```diff
@@ -50,8 +50,9 @@ class ContactsController < ApplicationController
       ContactWrt.new(
         name: requestor_details[:name],
         your_email: requestor_details[:email],
+        wca_id: User.find_by(email: requestor_details[:email])&.wca_id || 'None',
```
gregorbg, line 53: Why don't you let the user enter the WCA ID in the frontend for themselves?

danieljames-dj: If user is already logged-in, we will already have that data. We can get WCA ID from user if the user is not logged in. I'll consider that in another PR later, hope that is fine.

gregorbg (approved):

PR #9916, opened 2024-09-08
Fix editing custom activities

kr-matthews:
When we switched to always giving English names to to activities in the schedule editor, custom activities were not taken into account.

There should probably be some tests for editing custom activities?

### DIFF
--- a/app/webpacker/components/EditSchedule/EditActivities/EditActivityModal.js
+++ b/app/webpacker/components/EditSchedule/EditActivities/EditActivityModal.js
@@ -7,7 +7,18 @@ import {
 } from 'semantic-ui-react';
 import { DateTime } from 'luxon';
 import useInputState from '../../../lib/hooks/useInputState';
-import { commonActivityCodes } from '../../../lib/utils/wcif';
+
+const commonActivityCodes = {
+  'other-registration': 'On-site registration',
+  'other-checkin': 'Check-in',
+  'other-tutorial': 'Tutorial for new competitors',
+  'other-multi': 'Cube submission for 3x3x3 Multi-Blind',
+  'other-breakfast': 'Breakfast',
+  'other-lunch': 'Lunch',
+  'other-dinner': 'Dinner',
+  'other-awards': 'Awards',
+  'other-misc': 'Other',
+};
 
 const otherActivityCodeOptions = Object.entries(commonActivityCodes)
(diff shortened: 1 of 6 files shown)


gregorbg (commented):
Hm, there's a subtle problem to this approach. The function `humanizeActivityCode` itself is not hard-coded to English. As you can see, it uses `I18n.t` which is as translation-friendly as it gets.

The place where English is being forced right now is in `edit-schedule.js` function `buildPartialActivityFromCode`. There, we temporarily enable English, then call the `t` functions, then enable back any language that was chosen before our English hack.

Now your hard-coded `commonAcitvityCodes` approach hard-codes English directly at the level of `humanizeActivityCode`, which I feel is not in line with the existing implementation. You should return these hard-coded levels directlly in the `edit-schedule` code, because only that page is restricted to English. The general notion of "humanizing" an activity code should not be restricted to English.

If you want to do it _really_ neatly, you should include the common activity codes in `en.yml` and reference their (new) keys through `I18n.t`

kr-matthews:
> If you want to do it _really_ neatly, you should include the common activity codes in `en.yml` and reference their (new) keys through `I18n.t`

Tried to do this - not sure whether I did it right, I get `[missing "en.activity.multi" translation]` when running locally.

gregorbg:
The first step is to export them into JS by editing `config/i18n.yml`. We only export a limited subset of keys from the `YML` files to the frontend, in order to limit the size of the JSON that we have to serve over the network. You have to add a pattern matching the keys you want to export.

As to marking the keys as used, that happens automatically if you spell out the keys. For example, `I18n.t('activity.lunch')` will automatically mark `en.activity.lunch` as used. If you use dynamic keys like here in your PR, then you have two options:

1. Tell the code explicitly which keys are being used through a code comment: `// i18n-tasks-use t('my.awesome.key')`
2. (Preferrable here) If there are a bunch of keys and you don't want to check in 20 code comments, you can just add the whole group via a wildcard pattern in `i18n-tasks.yml`

gregorbg (approved):

gregorbg:
Code looking good now! The modal will still have the hard-coded localization coming from `commonActivityCodes`. Can you please turn that hash/object into an array (which only contains the keys) and refer to "proper" i18n within in `EditActivityModal` as well?

PR #9935, opened 2024-09-09
Move PMA and Sidekiq out of Main Staging Service

FinnIckler:
Allows us to scale the staging monolith separately and to finally reach infrastructure parity between staging and prod (but just running more container on prod). 
I moved some RAM around to make this happen; mainly giving Sidekiq Prod 1GB less RAM (totally fine, we are still wayyy over it's RAM usage) and making the staging monolith use an eighth of what prod is using (our desired state being 8 containers on our current instance), 
Leaving this PR in draft until https://github.com/thewca/worldcubeassociation.org/pull/9827 because both change the RAM values and I will make sure these PRs are compatible


infra/wca_on_rails/production/auxiliary_services.tf
```diff
@@ -13,15 +13,15 @@ resource "aws_ecs_task_definition" "auxiliary" {
   execution_role_arn = aws_iam_role.task_execution_role.arn
   task_role_arn      = aws_iam_role.task_role.arn
 
-  cpu = "2048"
-  memory = "7861"
+  cpu = "1536"
```
FinnIckler, line 16: this CPU value was set wrongly, we were always using 1536 CPU

infra/wca_on_rails/production/rails.tf
```diff
@@ -206,14 +206,14 @@ resource "aws_ecs_task_definition" "this" {
   task_role_arn      = aws_iam_role.task_role.arn
 
   cpu = "8192"
-  memory = "30000"
+  memory = "31634"
```
FinnIckler, line 209: This RAM value was also set wrongly (we were leaving 1.6GB of RAM on the table, not that that was ever an issue...)

FinnIckler:
image
Already successfully running all services

gregorbg (approved):

PR #9942, opened 2024-09-11
Fixing "missing zh-TW.cutoff.time.one" error

hanwu85:
"one" in the cutoff is missing in zh-TW.
I think this can fix the error.

### DIFF
--- a/config/locales/zh-TW.yml
+++ b/config/locales/zh-TW.yml
@@ -239,14 +239,17 @@ zh-TW:
     #original_hash: cf458e7
     time:
       zero: '%{count}次還原 < %{time}'
+      one: '%{count}次還原 < %{time}'
       other: '%{count}次還原 < %{time}'
     #original_hash: b49d930
     moves:
       zero: '%{count}次還原 < %{moves} 步'
+      one: '%{count}次還原 < %{moves} 步'
       other: '%{count}次還原 < %{moves} 步'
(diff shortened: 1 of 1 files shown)


gregorbg (approved):
I wonder why Internationalize (seemingly) didn't pick this up.
Do you have any clue, @jonatanklosko?

jonatanklosko:
If you look at other pluralization in the file, there's only "zero"/"other" in all cases.

According to rails-i18n (which internationalize uses as the reference) the forms should just be "other" ([ref](https://github.com/svenfuchs/rails-i18n/blob/master/rails/pluralization/zh-TW.rb)). Previously it was "one"/"other" and they [changed](https://github.com/svenfuchs/rails-i18n/pull/768) 6 years ago. Internationalize always asks for "zero" in addition.

In other words, the current state is expected.

gregorbg:
Hm. If `one` is not being used at all, then why are you trying to add it? @hanwu85

jonatanklosko:
Looks like the JS i18n pluralization rules are not aligned with the ones we use in Rails. From [the docs](https://github.com/fnando/i18n?tab=readme-ov-file#pluralization) it looks like by default the rules match English, so we may need to register rules for other languages manually, perhaps using `useMakePlural` as suggested in the docs.

gregorbg:
Closing in favor of #9945

PR #9943, opened 2024-09-11
Run 8 small containers instead of 1 big one

FinnIckler:
Same specs, I removed the block that sets the web concurrency because we are running a single worker everywhere now, which is the default when you don't set `WEB_CONCURRENCY` as per rails default puma.rb
```
# You can control the number of workers using ENV["WEB_CONCURRENCY"]. You
# should only set this value when you want to run 2 or more workers. The
# default is already 1.
```
Next step could be to move the prod containers to the t3 medium capacity provider, which would allow us to scale prod (wow). But let's monitor this for a while.

### DIFF
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -35,22 +35,6 @@ port ENV.fetch("PORT", 3000)
 # Allow puma to be restarted by `bin/rails restart` command.
 plugin :tmp_restart
 
-if ENV["RAILS_ENV"] == "production"
-  require "concurrent-ruby"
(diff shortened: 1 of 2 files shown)


FinnIckler:
I'm in a bit of a pickle in deploying this one, as I need to run the terraform apply only **after** this is merged, so we don't run 8 workers on each container the next time we deploy. We could deploy this in a call together I guess.

dunkOnIT (approved):

PR #9944, opened 2024-09-11
Add new Relic Metrics for Registrations

FinnIckler:
Adds a job that publishes hourly metrics about how many registrations have been processed in the last hour and how many potential registrations we can expect in the next hour. I am adding these now so we'll have plenty of data to look at when we work on scaling. I'll add these to a dashboard once they have run once


gregorbg (commented):
Cool idea, I think that the number of competitions opening registration within the next 60 minutes would also be useful to have

FinnIckler:
Done!

gregorbg (commented):
It's 2024, Rails' AR queries can do `BETWEEN` natively:
`Competition.where(registration_open: Time.now..60minutes.from_now)` :wink:

app/jobs/publish_registration_metrics.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+class PublishRegistrationMetrics < WcaCronjob
+  before_enqueue do
+    # This information is not useful for us in staging
+    throw :abort unless EnvConfig.WCA_LIVE_SITE?
+  end
+
+  def record_last_60_minutes_registrations
+    count = Registration.where('created_at >= ?', 60.minutes.ago).count
```
gregorbg, line 10: Fun fact: You can even map this to a AR native query! `Registration.where(created_at: 60.minutes.ago..)` works because Ruby supports open-ended ranges as a language feature: `a..` implicitly evaluates to `a..INFINITY` and Rails is smart enough to detect the infinity and map it to `>=` in SQL

gregorbg (changes requested):

app/jobs/publish_registration_metrics.rb
gregorbg: Sidenote: We do not use this feature consistently throughout the codebase because it is _relatively_ new (since Rails 6 or so) and we have queries which definitely predate that.

FinnIckler: I think rubocop-rails would be able to figure that out

gregorbg (approved):

app/jobs/publish_registration_metrics.rb
gregorbg: Yes, but last time we tried to introduce it it suggested like a gazillion changes which would have easily blown the 1000 line diff mark

PR #9945, opened 2024-09-12
Use Unicode pluralizers in JS frontend i18n

gregorbg:
Courtesy of https://github.com/eemeli/make-plural/tree/main/packages/plurals, which is not the exact same implementation as our backend project but at least relies on the same Unicode standard.

Also cleaning up the date-fns locale code a tiny bit while we're at it (essentially just moved the function calls to another block without modifying the calls themselves)

Should solve the problem(s) described in https://github.com/thewca/worldcubeassociation.org/pull/9942


jonatanklosko (approved):
:shipit:

PR #9948, opened 2024-09-14
Allow V2 Competitions to add On the Spot Registrations

FinnIckler:
Needs https://github.com/thewca/wca-registration/pull/688

### DIFF
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -210,16 +210,25 @@ class RegistrationsController < ApplicationController
     end
     ActiveRecord::Base.transaction do
       user, locked_account_created = user_for_registration!(params[:registration_data])
-      registration = @competition.registrations.find_or_initialize_by(user_id: user.id)
-      raise I18n.t("registrations.add.errors.already_registered") unless registration.new_record?
-      registration_comment = params.dig(:registration_data, :comments)
-      registration.assign_attributes(comments: registration_comment) if registration_comment.present?
-      registration.assign_attributes(accepted_at: Time.now, accepted_by: current_user.id)
-      params[:registration_data][:event_ids]&.each do |event_id|
-        competition_event = @competition.competition_events.find { |ce| ce.event_id == event_id }
-        registration.registration_competition_events.build(competition_event_id: competition_event.id)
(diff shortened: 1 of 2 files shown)


lib/microservices/registrations.rb
```diff
@@ -48,6 +52,21 @@ def self.registration_connection
...
 
+    # rubocop:disable Metrics/ParameterLists
+    def self.add_registration(competition_id, user_id, event_ids, comment, competing_status, current_user)
+      response = self.registration_connection.post(self.add_registration_path(competition_id)) do |req|
+        req.body = { competition_id: competition_id,
+                     user_id: user_id,
+                     event_ids: event_ids,
+                     comment: comment,
+                     competing_status: competing_status,
+                     current_user: current_user }.to_json
```
gregorbg, line 63: Is the `to_json` actually necessary here? I feel like we already configured the JSON modules to auto-encode both request and response

app/controllers/registrations_controller.rb
```diff
@@ -210,16 +210,25 @@ def do_add
...
-        competition_event = @competition.competition_events.find { |ce| ce.event_id == event_id }
-        registration.registration_competition_events.build(competition_event_id: competition_event.id)
+      if @competition.uses_new_registration_service?
+        Microservices::Registrations.add_registration(@competition.id,
+                                                      user.id,
+                                                      params[:registration_data][:event_ids],
+                                                      params[:registration_data][:comments],
+                                                      "accepted",
+                                                      @current_user.id)
+      else
```
gregorbg, line 220: I don't have the "Hide whitespace" feature on my phone, so GitHub makes it seem like all these lines changed. I suspect that the only actual change is indentation because of the if-else, can you please confirm?

FinnIckler: yes. I just enabled white whitespace and all those changes dissappeared

FinnIckler: image

lib/microservices/registrations.rb
FinnIckler: I just tested and it looks like it doesn't

gregorbg (approved):

PR #9965, opened 2024-09-18
Properly localize round names in frontend

gregorbg:
With the introduction of Puma, it was possible for API requests to be served by a thread that has a different locale set than the user who is currently browsing.

We already have logic to localize round names in the frontend on-the-fly, so let's use it!

### DIFF
--- a/app/controllers/api/v0/competitions_controller.rb
+++ b/app/controllers/api/v0/competitions_controller.rb
@@ -76,15 +76,11 @@ class Api::V0::CompetitionsController < Api::V0::ApiController
       {
         id: round&.id,
         roundTypeId: round_type.id,
-        # Also include the (localized) name here, we don't have i18n in js yet.
-        name: round&.name || "#{event.name} #{round_type.name}",
         results: results.sort_by { |r| [r.pos, r.personName] },
       }
     end
     render json: {
       id: event.id,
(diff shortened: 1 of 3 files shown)


FinnIckler:
We then should remove the name in v0/competitions_controller, it even mentions the name is only there because "we don' t have i18n n js yet." 

```
 def event_results
    competition = competition_from_params(associations: [:rounds])
    event = Event.c_find!(params[:event_id])
    results_by_round = competition.results
                                  .where(eventId: event.id)
                                  .group_by(&:round_type)
                                  .sort_by { |round_type, _| -round_type.rank }
    rounds = results_by_round.map do |round_type, results|
      # I think all competitions now have round data, but let's be cautious
      # and assume they may not.
      # round data.
      round = competition.find_round_for(event.id, round_type.id)
      {
        id: round&.id,
        roundTypeId: round_type.id,
        # Also include the (localized) name here, we don't have i18n in js yet.
        name: round&.name || "#{event.name} #{round_type.name}",
        results: results.sort_by { |r| [r.pos, r.personName] },
      }
    end
    render json: {
      id: event.id,
      # Also include the (localized) name here, we don't have i18n in js yet.
      name: event.name,
      rounds: rounds,
    }
  end
```

FinnIckler (approved):

PR #9967, opened 2024-09-19
Map `rejected` competing_status to `deleted` WCIF status

dunkOnIT:
This was causing errors in production as `rejected` had no mapping. 

We'll need to think about whether and how to update the WCIF statuses to line up with the registration statuses.


gregorbg (approved):
Thanks for detecting this so quickly!

PR #9970, opened 2024-09-19
Squeeze together new competitiontable on mobile

gregorbg:
Custom CSS yay!

localhost_3000_competitions_legacy=off

### DIFF
--- a/app/assets/stylesheets/competitions.scss
+++ b/app/assets/stylesheets/competitions.scss
@@ -347,6 +347,19 @@ $venue-map-wrapper-height: 400px;
   }
 }
 
+tr.competition-info {
+  &.mobile-compact {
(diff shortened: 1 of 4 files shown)


FinnIckler:
That label still overlaps, or is this screenshot not the newest state? Also weird triangle underneath it

gregorbg:
What do you mean by "the label still overlaps"?
I don't see any overlap in the screenshot, but maybe I'm blind... 🙈

The "weird triangle" is a graphical feature by SemUI. It's supposed to be a ribbon that "wraps around" the border of the table from behind.

FinnIckler:
ohhhh that's what it's supposed to be. I thought the label was overlapping the row and then the triangle was some error

FinnIckler (approved):

PR #9974, opened 2024-09-20
Use m6i.large as the base instance instead of m6i.2xlarge

FinnIckler:
The m6i.2xlarge just has 3 network interfaces so we can't run 8 containers on it like we planned. Instead we now run the equivalent 4 m6i.large instances and run 2 containers on each.

### DIFF
--- a/infra/wca_on_rails/production/rails.tf
+++ b/infra/wca_on_rails/production/rails.tf
@@ -273,13 +273,8 @@ resource "aws_ecs_service" "this" {
   enable_execute_command = true
 
   ordered_placement_strategy {
-    type  = "spread"
-    field = "attribute:ecs.availability-zone"
-  }
-
-  ordered_placement_strategy {
-    type  = "spread"
-    field = "instanceId"
+    type  = "binpack"
+    field = "memory"
   }
 
   load_balancer {
--- a/infra/wca_on_rails/production/rails.tf
+++ b/infra/wca_on_rails/production/rails.tf
@@ -206,14 +206,14 @@ resource "aws_ecs_task_definition" "this" {
   task_role_arn      = aws_iam_role.task_role.arn
 
   cpu = "1024"
-  memory = "3954"
+  memory = "3910"
 
   container_definitions = jsonencode([
     {
       name              = "rails-production"
       image             = "${var.shared.ecr_repository.repository_url}:latest"
       cpu    = 1024
-      memory = 3954
+      memory = 3910
       portMappings = [
         {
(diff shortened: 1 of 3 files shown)


infra/wca_on_rails/production/rails.tf
```diff
@@ -206,14 +206,14 @@ resource "aws_ecs_task_definition" "this" {
   task_role_arn      = aws_iam_role.task_role.arn
 
   cpu = "1024"
-  memory = "3954"
+  memory = "3910"
```
gregorbg, line 209: Where does this change come from?

infra/wca_on_rails/production/rails.tf
```diff
@@ -206,14 +206,14 @@ resource "aws_ecs_task_definition" "this" {
...
 
   container_definitions = jsonencode([
     {
       name              = "rails-production"
       image             = "${var.shared.ecr_repository.repository_url}:latest"
       cpu    = 1024
-      memory = 3954
+      memory = 3910
```
gregorbg, line 216: (see above)

infra/wca_on_rails/staging/auxiliary_services.tf
```diff
@@ -104,6 +104,11 @@ resource "aws_ecs_service" "auxiliary" {
 
   enable_execute_command = true
 
+  deployment_circuit_breaker {
+    enable   = true
+    rollback = false
+  }
```
gregorbg, line 110: What does this do? I cannot infer it from the name alone

infra/wca_on_rails/production/rails.tf
FinnIckler: the m6i.large instance reserves slightly more RAM for the ECS Agent then the m6i.2xlarge, so I had to tweak this number to fit in 2 container per instance.

infra/wca_on_rails/staging/auxiliary_services.tf
FinnIckler: This is good to have in every ECS Service.
It stops non Blue Green deployments after they failed 3 times. I changed this in the console for the staging auxiliary service after I saw that it failed deployment 600 times due to the staging redis being half full. It's not technically part of this PR, but the terraform files need to always match the state of what's on AWS or they will revert it.

infra/wca_on_rails/production/rails.tf
FinnIckler: this need to match the number above

gregorbg (approved):

PR #9977, opened 2024-09-20
Automatically refresh when registration opens

FinnIckler:
https://github.com/user-attachments/assets/f59f58d5-a716-4bec-bbd5-49e8d5e8bf78

Uses the old message if the competition opens in more than one hour, then it switches to the countdown.


app/webpacker/components/RegistrationsV2/Register/RegistrationClosedMessage.jsx
```diff
@@ -0,0 +1,48 @@
...
+import React, { useEffect, useState } from 'react';
+import { Message } from 'semantic-ui-react';
+import I18nHTMLTranslate from '../../I18nHTMLTranslate';
+
+export default function RegistrationClosedMessage({
+  registrationStart,
+  onTimerEnd,
+}) {
+  const [timeLeft, setTimeLeft] = useState({
+    days: 0, hours: 0, minutes: 0, seconds: 0,
```
kr-matthews, line 11: This should have the correct initial value. Especially since there's a `timeLeft.hours < 1` check below.

app/webpacker/components/RegistrationsV2/Register/RegistrationClosedMessage.jsx
```diff
@@ -0,0 +1,48 @@
...
+      const diff = start.diff(now, ['days', 'hours', 'minutes', 'seconds']).toObject();
+      setTimeLeft({
+        days: Math.floor(diff.days),
+        hours: Math.floor(diff.hours),
+        minutes: Math.floor(diff.minutes),
+        seconds: Math.floor(diff.seconds),
+      });
+    };
+
+    const intervalId = setInterval(updateTimer, 1000); // Update every second
```
kr-matthews, line 35: I'm not sure how important this is for performance or anything, but if there's more than 1 hour left then you could set the next update to happen when there's an hour left. Although then the `days`, `hours`, `minutes`, and `seconds` values will be incorrect, so maybe this is a bad suggestion.

app/webpacker/components/RegistrationsV2/Register/RegistrationClosedMessage.jsx
```diff
@@ -0,0 +1,48 @@
...
+    updateTimer(); // Initial call to set the state right away
+
+    return () => clearInterval(intervalId); // Cleanup interval on unmount
+  }, [registrationStart, onTimerEnd, start]);
+
+  return (
+    <Message color="blue">
+      { timeLeft.hours < 1
```
kr-matthews, line 43:
```suggestion
      { timeLeft.days < 1 && timeLeft.hours < 1
```
otherwise 1 day 0 hours 1 minute 1 second would satisfy this.

app/webpacker/components/RegistrationsV2/Register/index.jsx
```diff
@@ -53,22 +58,39 @@ function Register({
...
+    setTimerEnded(true);
+  }, [setTimerEnded]);
+
+  if (isFetching) {
+    return <Loading />;
+  }
+
+  if (userCanPreRegister(userInfo, competitionInfo) || competitionInfo['registration_currently_open?'] || timerEnded) {
+    return (
+      isFetching ? <Loading />
```
kr-matthews, line 71: The is-fetching case is handled above now, right? Doesn't need to be here.

kr-matthews (changes requested):

app/webpacker/components/RegistrationsV2/Register/index.jsx
FinnIckler: copy paste whoopsie

app/webpacker/components/RegistrationsV2/Register/RegistrationClosedMessage.jsx
FinnIckler: mh.. I don't really expect anyone to sit on the page if it's more than an hour away

app/webpacker/components/RegistrationsV2/Register/RegistrationClosedMessage.jsx
FinnIckler: I call updateTimer immediately in the useEffect. isn't that enough? I haven't seen any rendering issues on local

app/webpacker/components/RegistrationsV2/Register/RegistrationClosedMessage.jsx
FinnIckler: I guess `const [timeLeft, setTimeLeft] = useState(start.diff(DateTime.local(), ['days', 'hours', 'minutes', 'seconds']).toObject());` this would be enough though

app/webpacker/components/RegistrationsV2/Register/RegistrationClosedMessage.jsx
FinnIckler: I guess `const [timeLeft, setTimeLeft] = useState(start.diff(DateTime.local(), ['days', 'hours', 'minutes', 'seconds']).toObject());` this would be enough though

kr-matthews: It's working now, but if someone comes along and modifies the effect, there's a chance of them introducing a bug. For example, any new effects which depend on these values will be triggered on the initial render when they're all 0.
If something can be done without an effect (such as setting this initial value) then it's almost always better to do without.

app/webpacker/components/RegistrationsV2/Register/RegistrationClosedMessage.jsx
```diff
@@ -0,0 +1,48 @@
...
+
+  // Update the timer every second
+  useEffect(() => {
+    const updateTimer = () => {
+      const now = DateTime.local();
+
+      if (now >= start) {
+        onTimerEnd();
+        return;
+      }
```
kr-matthews, line 24: Not a big deal, but it's best to have each effect do 1 effect, so I would separate this onto its own effect (dependent on all the time values - if they're all 0 then call `onTimerEnd`). (Which is an example of why they should be initialized properly.)

FinnIckler: I added the second effect

app/webpacker/components/RegistrationsV2/Register/RegistrationClosedMessage.jsx
FinnIckler: makes sense, I changed it! Will retest on monday

FinnIckler:
still works!

dunkOnIT:
@kr-matthews just checking if your requested changes have been addressed?

kr-matthews:
> @kr-matthews just checking if your requested changes have been addressed?

yes

dunkOnIT:
Awesome! If you mark you changes as resolved/submit a review for the React side of things that would be great

kr-matthews (approved):
Didn't test locally, but the react code looks reasonable

app/models/competition.rb
```diff
@@ -732,7 +732,7 @@ def uses_new_registration_service?
   end
 
   def should_render_register_v2?(user)
-    uses_new_registration_service? && user.cannot_register_for_competition_reasons(self).empty? && (registration_currently_open? || user_can_pre_register?(user))
```
gregorbg, line 735: Why is this change necessary, what does it do in the context of the PR?

app/views/registrations/register.html.erb
```diff
@@ -10,7 +10,7 @@
   <% opens_in_days = distance_of_time_in_words_to_now(@competition.registration_open) %>
   <% closes_in_days = distance_of_time_in_words_to_now(@competition.registration_close) %>
 
-  <% if @competition.registration_not_yet_opened? %>
+  <% if !@competition.uses_new_registration_service? && @competition.registration_not_yet_opened? %>
```
gregorbg, line 13: I suspect this is womehow linked to the above change in `should_render_register_v2?`

app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
```diff
@@ -31,7 +32,7 @@ const maxCommentLength = 240;
 const potentialWarnings = (competitionInfo) => {
   const warnings = [];
   // Organizer Pre Registration
-  if (!competitionInfo['registration_currently_open?']) {
+  if (DateTime.fromISO(competitionInfo.registration_open) > DateTime.now()) {
```
gregorbg, line 35: Too much parsing logic inside an `if` clause. I feel like we already have a method for that, let me check...

gregorbg: Yes, we have `hasPassed` and `hasNotPassed` in `lib/utils/dates.js`

app/webpacker/components/RegistrationsV2/Register/RegistrationClosedMessage.jsx
```diff
@@ -0,0 +1,48 @@
...
+import React, { useEffect, useState } from 'react';
+import { Message } from 'semantic-ui-react';
+import I18nHTMLTranslate from '../../I18nHTMLTranslate';
+
+export default function RegistrationClosedMessage({
+  registrationStart,
+  onTimerEnd,
+}) {
+  const start = DateTime.fromISO(registrationStart);
+  const [timeLeft, setTimeLeft] = useState(start.diff(DateTime.local(), ['days', 'hours', 'minutes', 'seconds']).toObject());
```
gregorbg, line 11: This should definitely be a `util` function

app/webpacker/components/RegistrationsV2/Register/RegistrationClosedMessage.jsx
```diff
@@ -0,0 +1,48 @@
...
+
+  useEffect(() => {
+    if (timeLeft.days === 0
+      && timeLeft.hours === 0
+      && timeLeft.minutes === 0
+      && timeLeft.seconds === 0) {
+      onTimerEnd();
+    }
+  }, [onTimerEnd, timeLeft.days, timeLeft.hours, timeLeft.minutes, timeLeft.seconds]);
```
gregorbg, line 20: Can you just use `timeLeft` as a whole as dependency?

app/webpacker/components/RegistrationsV2/Register/RegistrationClosedMessage.jsx
```diff
@@ -0,0 +1,48 @@
...
+
+  useEffect(() => {
+    const updateTimer = () => {
```
gregorbg, line 23: Why are you defining this as a local function within `useEffect`?

app/webpacker/components/RegistrationsV2/Register/RegistrationClosedMessage.jsx
```diff
@@ -0,0 +1,48 @@
...
+      const now = DateTime.local();
+
+      const diff = start.diff(now, ['days', 'hours', 'minutes', 'seconds']).toObject();
```
gregorbg, line 26: Use a shared function, see above

gregorbg: And on that note, you could apply `Math.floor` inside that util function already, so that you can use the return value of the function as param to `setTimeLeft` directly below

app/webpacker/components/RegistrationsV2/Register/RegistrationClosedMessage.jsx
```diff
@@ -0,0 +1,48 @@
...
+    updateTimer(); // Initial call to set the state right away
+
+    return () => clearInterval(intervalId); // Cleanup interval on unmount
+  }, [registrationStart, onTimerEnd, start]);
+
+  return (
+    <Message color="blue">
+      { timeLeft.days < 1 && timeLeft.hours < 1
+        ? <I18nHTMLTranslate i18nKey="competitions.registration_v2.register.will_open_countdown" options={{ minutes: timeLeft.minutes, seconds: timeLeft.seconds }} />
+        : <I18nHTMLTranslate i18nKey="registrations.will_open_html" options={{ days: start.toRelative(), time: DateTime.fromISO(registrationStart).toLocaleString(DateTime.DATETIME_FULL) }} />}
```
gregorbg, line 45: Isn't `const start` above exactly the same as `DateTime.fromISO(registrationStart)` so that you could rewrite this to `start.toLocaleString(...)`?

app/webpacker/components/RegistrationsV2/Register/index.jsx
```diff
@@ -8,6 +8,9 @@ import StoreProvider, { useDispatch } from '../../../lib/providers/StoreProvider
 import messageReducer from '../reducers/messageReducer';
 import WCAQueryClientProvider from '../../../lib/providers/WCAQueryClientProvider';
 import ConfirmProvider from '../../../lib/providers/ConfirmProvider';
+import RegistrationClosedMessage from './RegistrationClosedMessage';
+
+const userCanPreRegister = (user, competitionInfo) => competitionInfo.organizers.some((o) => o.id === user.id);
```
gregorbg, line 13: Why only organizers and not Delegates?

app/webpacker/components/RegistrationsV2/Register/index.jsx
```diff
@@ -53,22 +58,36 @@ function Register({
...
 
+  const onTimerEnd = useCallback(() => {
+    setTimerEnded(true);
+  }, [setTimerEnded]);
+
+  if (isFetching) {
+    return <Loading />;
+  }
+
+  if (userCanPreRegister(userInfo, competitionInfo) || competitionInfo['registration_currently_open?'] || timerEnded) {
```
gregorbg, line 69: Lots of calculations inside one `if` statement, not happy about that

gregorbg (changes requested):
I feel like you're doing three changes in this PR.
1. Reduce the amount of data returned by the internal `User` API
2. Implement a timer for registration opening
3. Enable pre-registration for competition managers

(1) Is very trivial but important, and it's unfortunate that this change is being blocked by discussions about (2) and (3). So it should be its own PR.

(3) needs its own design discussion, especially because I think it depends on another PR in the microservice. If that PR gets delayed, it would also block features (1) and (2), so it should be its own PR.

(2) is the core essence of this PR, and I have added lots of comments. Please split out the other things into separate PRs

app/models/competition.rb
FinnIckler: Because before we showed the Registration Opens in x in Ruby only and now we want to show it in react

app/views/registrations/register.html.erb
FinnIckler: yes

app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
FinnIckler: done

app/webpacker/components/RegistrationsV2/Register/RegistrationClosedMessage.jsx
FinnIckler: I had it auto generated by Rubymine, but it looks like eslint doesn't complain if I just use the whole object (also I googled and it looks like it's only an issue if the reference changes, with I think isn't the case here?)

app/webpacker/components/RegistrationsV2/Register/RegistrationClosedMessage.jsx
FinnIckler: done

app/webpacker/components/RegistrationsV2/Register/RegistrationClosedMessage.jsx
FinnIckler: true! must have been before I refactored it

app/webpacker/components/RegistrationsV2/Register/index.jsx
FinnIckler: should be better when I move userCanPreRegister to Ruby?

app/webpacker/components/RegistrationsV2/Register/RegistrationClosedMessage.jsx
```diff
@@ -0,0 +1,38 @@
...
+import { Message } from 'semantic-ui-react';
+import I18nHTMLTranslate from '../../I18nHTMLTranslate';
+import { fullTimeDiff } from '../../../lib/utils/dates';
+
+export default function RegistrationClosedMessage({
+  registrationStart,
+  onTimerEnd,
+}) {
+  const start = DateTime.fromISO(registrationStart);
+  const [timeLeft, setTimeLeft] = useState(start.diff(DateTime.local(), ['days', 'hours', 'minutes', 'seconds']).toObject());
```
gregorbg, line 12: Why is this not using the new `fullTimeDiff` util function, now that we have it?

gregorbg (changes requested):

app/webpacker/components/RegistrationsV2/Register/RegistrationClosedMessage.jsx
FinnIckler: good catch

gregorbg (approved):

FinnIckler:
I'll do one more test and then I'll merge it

FinnIckler:
confirmed still working

PR #9978, opened 2024-09-20
Use v2 Registration System for newly-created competitions

dunkOnIT:
Just a migration that changes the default for `uses_v2_registration` to `true`. Tested locally and it works. We can merge and deploy on Monday, I'll draft an email as well that we can send to staff and RO's.

### DIFF
--- a/app/models/competition.rb
+++ b/app/models/competition.rb
@@ -2664,6 +2664,18 @@ class Competition < ApplicationRecord
     competition_payment_integrations.destroy_all
   end
 
+  def can_change_registration_system?
+    return false unless registration_not_yet_opened?
+
+    if uses_new_registration_service?
+      return true
+    elsif self.registrations.any?
+      return false
+    else
+      return true
+    end
+  end
+
   # Our React date picker unfortunately behaves weirdly in terms of backend data
   def self.date_json_schema(string_format)
     {
--- a/app/models/competition.rb
+++ b/app/models/competition.rb
@@ -969,10 +969,6 @@ class Competition < ApplicationRecord
     registration_open && Time.now < registration_open
   end
 
-  def after_registration_open?
-    registration_open && Time.now > registration_open
-  end
-
   def registration_past?
     registration_close && registration_close < Time.now
   end
@@ -2669,11 +2665,7 @@ class Competition < ApplicationRecord
   end
 
   def can_change_registration_system?
-    return false if after_registration_open?
-
-    return true if uses_new_registration_service?
-
-    self.registrations.none?
+    registration_not_yet_opened? && (uses_new_registration_service? || self.registrations.empty?)
   end
 
   # Our React date picker unfortunately behaves weirdly in terms of backend data
--- a/app/models/competition.rb
+++ b/app/models/competition.rb
@@ -2601,7 +2601,7 @@ class Competition < ApplicationRecord
       showAtAll: form_data.dig('admin', 'isVisible'),
       being_cloned_from_id: form_data.dig('cloning', 'fromId'),
       clone_tabs: form_data.dig('cloning', 'cloneTabs'),
-      uses_v2_registrations: form_data.dig('admin', 'usesV2Registrations')
+      uses_v2_registrations: form_data.dig('admin', 'usesV2Registrations'),
     }
   end
 
--- a/app/models/competition.rb
+++ b/app/models/competition.rb
@@ -969,6 +969,10 @@ class Competition < ApplicationRecord
     registration_open && Time.now < registration_open
   end
 
+  def after_registration_open?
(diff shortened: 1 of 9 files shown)


FinnIckler:
probably just need to set it to false as the default in the competition factory though

gregorbg (changes requested):

config/locales/en.yml
```diff
@@ -1364,6 +1364,7 @@ en:
       in_progress_at_wca_live_html: "This competition is ongoing. You can check the live results <a href='%{link_here}'>here</a>!"
       reg_opens_too_early: "Registration is set to start in less than the required 48 hours."
       registration_already_closed: "The deadline to register has already passed. Please set it to be in the future."
+      uses_v2_reg: "This competition is using the new Registration System. Please contact WCAT if you believe this competition should use the old system instead."
```
gregorbg, line 1367: Suuuuuper nitpick, but could you please not use `reg` in this key? (Properly name it `registration` instead)

gregorbg (commented):
Sorry you have to suffer through another 18 minutes of CI for this nitpick, but I feel like I need to play devil's advocate in the interest of long term consistency

config/locales/en.yml
dunkOnIT: Fixed

app/models/competition.rb
```diff
@@ -563,6 +563,7 @@ def registration_full_message
 
   def reg_warnings
     warnings = {}
+    warnings[:uses_v2] = I18n.t('competitions.messages.uses_v2_registrations') if uses_new_registration_service?
```
gregorbg, line 566: More nitpick: This should also share the same name with the field.
I think it's not technically a strict requirement, but certainly nice for other developers. Otherwise, V2 could be anything, perhaps the V2 database or V2 schedule editor.

config/locales/en.yml
```diff
@@ -1693,6 +1695,7 @@ en:
         admin:
           is_confirmed: ""
           is_visible: ""
+          uses_v_2_registrations: ""
```
gregorbg, line 1698: Maybe use this footnote text to explain that switching systems while registration is open is painful and strongly discouraged?

gregorbg (changes requested):

config/locales/en.yml
dunkOnIT: The implementation of bringing in additional properties (like competitor count, or checking for registration open date) is not trivial for me, so not going to attempt it without further discussion. Have added the footnote in the meantime

FinnIckler:
What about changing the Dev dump sanitizers to always return true for `uses_v2_registration` as well, means we don't have to activate it anymore locally

dunkOnIT:
I think it's worth keeping those on the old reg system until we bring v2 back into the monolith, given that it makes it more complex for someone to run the container locally on default settings.

gregorbg:
> I think it's worth keeping those on the old reg system until we bring v2 back into the monolith, given that it makes it more complex for someone to run the container locally on default settings.

Very strongly agree.

app/models/competition.rb
```diff
@@ -2661,6 +2664,18 @@ def disconnect_all_payment_integrations
     competition_payment_integrations.destroy_all
   end
 
+  def can_change_registration_system?
+    return false unless registration_not_yet_opened?
```
gregorbg, line 2668: The implicit NOT in `unless` and the (explicit) NOT in the method name cancel each other out. Is there a method like `registration_already_open?`?

app/models/competition.rb
```diff
@@ -2661,6 +2664,18 @@ def disconnect_all_payment_integrations
...
+
+    if uses_new_registration_service?
+      return true
+    elsif self.registrations.any?
+      return false
+    else
+      return true
```
gregorbg, line 2675: As an exercise, try merging this into one boolean expression with `&&` and/or `||` operators 😉

app/models/competition.rb
dunkOnIT: Created `after_registration_open`. To satisfy this. Naming is a bit tricky, as we're trying to communicate that the registration_open date is in the past, not just whether registration is _currently_ open

app/models/competition.rb
dunkOnIT: Took it down to 3 separate booleans - I like the readability/clarity of not combining statements which are logically separate but can change if you prefer to compound them

dunkOnIT:
@thewca-bot deploy staging

dunkOnIT:
@thewca-bot deploy staging

dunkOnIT:
@thewca-bot deploy staging

gregorbg (approved):
:ship:

PR #9994, opened 2024-09-23
Add missing V2 Localization

FinnIckler:
This should be all public pages

### DIFF
--- a/app/webpacker/components/RegistrationsV2/Register/StepPanel.jsx
+++ b/app/webpacker/components/RegistrationsV2/Register/StepPanel.jsx
@@ -10,28 +10,25 @@ import { hasPassed } from '../../../lib/utils/dates';
 const requirementsStepConfig = {
   key: 'requirements',
   description: 'Accept competition terms',
-  i18nKey: 'competitions.registration_v2.requirements.title',
+  i18nKey: 'competitions.registration_v2.register.panel.requirements',
   component: RegistrationRequirements,
 };
 
 const competingStepConfig = {
   key: 'competing',
-  i18nKey: 'competitions.nav.menu.register',
-  description: 'Choose your events',
+  i18nKey: 'competitions.registration_v2.register.panel.register',
   component: CompetingStep,
 };
 
 const paymentStepConfig = {
   key: 'payment',
-  description: 'Enter billing information',
-  i18nKey: 'registrations.payment_form.labels.payment_information',
+  i18nKey: 'competitions.registration_v2.register.panel.payment',
   component: StripeWrapper,
 };
 
 const registrationOverviewConfig = {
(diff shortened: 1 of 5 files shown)


app/webpacker/components/RegistrationsV2/Register/StepPanel.jsx
```diff
@@ -10,28 +10,25 @@ import { hasPassed } from '../../../lib/utils/dates';
 const requirementsStepConfig = {
   key: 'requirements',
   description: 'Accept competition terms',
-  i18nKey: 'competitions.registration_v2.requirements.title',
+  i18nKey: 'competitions.registration_v2.register.panel.requirements',
```
dunkOnIT, line 13: This should perhaps be .requirements.title?

config/locales/en.yml
```diff
@@ -1845,13 +1852,25 @@ en:
...
         no_self_update: "Contact the Organization Team to update your registration."
       register:
+        panel:
+          approval:
+            title: "Approval"
+            description: "By organization team"
+          payment:
+            title: "Payment information"
+            description: "Enter billing information"
+          register:
```
dunkOnIT, line 1862: Suggest changing this to `competing` instead of `register`

app/webpacker/components/RegistrationsV2/Register/StepPanel.jsx
```diff
@@ -10,28 +10,25 @@ import { hasPassed } from '../../../lib/utils/dates';
 const requirementsStepConfig = {
   key: 'requirements',
   description: 'Accept competition terms',
```
dunkOnIT, line 12: I think this description should be deleted?

dunkOnIT:
We still need to localize:
- `expandableColumns` in RegistrationAdmin
- "Enable Waiting List Edit Mode" in line 396 of RegistrationAdministrationList.jsx

FinnIckler:
This is because we originally didn't have the waiting list position in the i18n key and it hasn't been updated for Brazilian Portuguese. On normal Portuguese it has it `waiting_list: 'Está em Lista de Espera na posição %{waiting_list_position}.'`
That is kinda annoying....

FinnIckler:
Done! Let's email the translators after this is merged

FinnIckler:
No, it's here https://github.com/thewca/worldcubeassociation.org/blob/9885460adb28a81254b5219cb4af62b5500fb097/app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationList.jsx#L91

FinnIckler:
And for the brazillian translation, we should just tell the translators to check if some of the translations need to be updated.

dunkOnIT (approved):

PR #9998, opened 2024-09-26
Add registration V2 models back into the monolith

FinnIckler:
Work in Progress, but I wanted to already create it so @gregorbg can implement his ideas about a better relational model for lanes. 
I added a Job to create a v2 registration as well, which is about what it will look like later for event registration


FinnIckler:
We found a better way to do this 🎉

PR #9999, opened 2024-09-27
Pull gem and Yarn installation into separate `environment` Docker service

gregorbg:
Problem: Rails, Webpacker and Sidekiq all need our Ruby runtime. Currently, all three download the same gems independently, and without caching them into a volume.

Solution: Let one central `environment` service download everything (and also configure it, courtesy of the new `bin/setup` script in Rails 7) and also cache it via mounted volumes.

Note: Previously, we had problems with native bindings in a shared volume. Whenever Ruby was updated, the native gems that were built for the previous version caused errors/crashes. This is now fixed through a (somewhat complex) `entrypoint` script, that loads the gems into per-version directories.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,44 +1,61 @@
+x-environment: &env
+  RAILS_ENV: development
+
 x-app: &app
   build:
     dockerfile: Dockerfile.dev
     context: .
   working_dir: /app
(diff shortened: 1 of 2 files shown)


FinnIckler (approved):
LGTM

PR #381, opened 2016-02-07
Make Competitions.id immutable

timhabermaas:
Letting the end user change primary keys was a rather unfortunate design
decision which led to some weird corner cases. Especially keeping
foreign keys in sync was painful. This PR fixes that and introduces
another column: `friendly_id` which is meant to be shown and edited by the
user. `friendly_id` is generated using the same rules as `id`.

Note: `id` is still generated before validation in order to not break
existing assumptions/code.

:bangbang: **Not meant to be merged before #407 and #319**. :bangbang:

PR #10000, opened 2024-09-27
Migrate Competition model primary key to stable AUTO_INCREMENT integer

gregorbg:
Should be the last of our PHP legacy problems.

This will obviously require massive code changes throughout the entire Rails app (`find` has to be replaced by `find_by`) but we can also get rid of problems like `with_old_id` and all the hacks that surround it.

WIP but it seemed appropriate for PR number 10.000 :smile:

### DIFF
--- a/db/migrate/20240816041854_add_auto_increment_to_competitions.rb
+++ b/db/migrate/20240816041854_add_auto_increment_to_competitions.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+class AddAutoIncrementToCompetitions < ActiveRecord::Migration[7.1]
+  def up
+    remove_foreign_key :microservice_registrations, :Competitions
+
+    execute "ALTER TABLE Competitions DROP PRIMARY KEY"
+
+    rename_column :Competitions, :id, :competition_id
+    add_index :Competitions, :competition_id, unique: true
+
+    add_column :Competitions, :id, :bigint, first: true
+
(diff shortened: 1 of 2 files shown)


jfly:
Wow! This is huge.

gregorbg:
Recording this here for further research / history / bookkeeping later on: https://github.com/thewca/worldcubeassociation.org/pull/381

app/controllers/api/v0/competitions_controller.rb
```diff
@@ -242,7 +242,7 @@ def update_wcif
 
   private def competition_from_params(associations: {})
     id = params[:competition_id] || params[:id]
-    competition = Competition.includes(associations).find_by(id: id)
+    competition = Competition.includes(associations).find(id)
```
gregorbg: There is a very notable difference between `find_by(id: id)` and `find(id)`: In case of doubt, ie when passing an invalid or non-existing ID, the `find_by` will return `nil` and resume code execution, whereas `find` throws and halts execution (it is the only method in Rails AR that does throw despite not being named `find!`, for historical reasons).

In very "neuralgic" spots like this, you should probably honor the original author's intention of "gently" failing and returning `nil`, thus using `find_by(competition_id: id)`

FinnIckler: of course!

FinnIckler:
I'm just going to delete the final failing test, because it is currently not possible to do an id change and an event change at the same time on the front end

FinnIckler:
or actually I'll change it to be a "changes the competition_id of competition_events" test like the others

PR #10016, opened 2024-09-30
Add new V2 models to the monolith

FinnIckler:
Adds the registration history, rejected/waitilsted status and waiting list to the V1 models, adds a lane library and a job to add registration from a queue (sidekiq locally).

I think for the migration we can remove accepted_at/accepted_by and the other status columns because that is now stored in the registration history. Instead I would propose a status column.


FinnIckler:
Not sure why the tests failed, I added waiting_list to the unclonable attributes

gregorbg:
You were a bit too eager to please here, it doesn't _need_ to be added to (un)cloneable attributes. These are matched against `Competition.column_names`, i.e. things that the competition table directly holds the reference to. But in the case of waiting lists, the waiting list table is the one that holds the reference so the Competition model does not need to bother.

It _does_ need to define a cloning behavior, but not a cloning attribute.

FinnIckler:
ahh. that failure text was misleading then. Test are passing now

app/jobs/add_registration_job.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class AddRegistrationJob < ApplicationJob
+  def perform(lane, competition_id, user_id, lane_params)
+    ActiveRecord::Base.transaction do
+      if lane == "competing"
+        registration = Registration.build(competition_id: competition_id, user_id: user_id)
+        Registrations::CompetingLane.create!(registration, lane_params[:competing], user_id, competition_id)
```
gregorbg, line 8: My suggestion: Use the `lane` parameter to perform some kind of (very rudimentary) lookup to find the module/class that handles this lane, and then give the lane a method along the lines of `process`.

The implementation can then do whatever the heck it wants with the raw payload. In this case, the body of that method over in the `lib` code should be the one to create the registration entity.

Because strictly speaking, the `Registration` model in the legacy backend is (mostly, except for the `is_competing` flag which was hacked in waaaaay later as a low-effort quick and easy approach for Continental and World staff planning) actually a `competing_registration` model. So it would make sense imo if it lives under the auspices of the competing lane

gregorbg (commented):
Leaving this as a non-binding comment because the design is still very much up for debate at this point in time

app/jobs/add_registration_job.rb
```diff
@@ -0,0 +1,16 @@
...
+  def perform(lane_name, competition_id, user_id, lane_params)
+    lane_model_name = lane_name.upcase_first
+
+    lane = Registrations::Lanes::class_eval(lane_model_name)
+    if lane.nil?
+      raise "No Lane exists for #{lane_model_name}"
+    end
+
+    ActiveRecord::Base.transaction do
+      lane.process!(lane_params[:competing], user_id, competition_id)
```
gregorbg, line 13: Nitpick: You don't know (in general) what kind of lane this will be, so you cannot assume that the key `competing` is relevant here.
You should just pass the whole params to the lane and let it do its own thing.

gregorbg (changes requested):

app/jobs/add_registration_job.rb
FinnIckler: yeah, I even changed this in my API branch.

gregorbg (approved):

lib/registrations/lanes/competing.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module Registrations
+  module Lanes
+    module Competing
+      def self.process!(lane_params, user_id, competition_id)
+        registration = Registration.build(competition_id: competition_id,
+                                          user_id: user_id,
+                                          comments: lane_params[:competing][:comment] || '',
+                                          guests: lane_params[:guests] || 0)
```
FinnIckler, line 10: this is on purpose btw. currently guests is not part of the `[:competing]` payload right now (but probably will be in the future when API Design is finalized). I think our idea was that guests could be generalized for staff only registrations and such.

PR #10019, opened 2024-10-01
Send WRC Report Notifications to Webhook

dunkOnIT:


app/controllers/delegate_reports_controller.rb
```diff
@@ -51,6 +51,13 @@ def update
         if @competition.end_date >= DelegateReport::REPORTS_ENABLED_DATE
           CompetitionsMailer.notify_of_delegate_report_submission(@competition).deliver_later
           CompetitionsMailer.wrc_delegate_report_followup(@competition).deliver_later
+
+          Faraday.post(EnvConfig.WRC_WEBHOOK_URL) do |req|
```
FinnIckler, line 55: External API Request like that should be part of a job, to not slow down the server if the request is slow.

dunkOnIT: Ah great point thank you!

app/models/competition.rb
```diff
@@ -1411,7 +1411,7 @@ def wca_live_link
   end
 
   def results_submitted?
-    !results_submitted_at.nil?
+    true
```
gregorbg, line 1414: Whoopsie from testing

app/controllers/api/v0/competitions_controller.rb
```diff
@@ -4,6 +4,14 @@ class Api::V0::CompetitionsController < Api::V0::ApiController
   # Enable CSRF protection if we use cookies based user instead of OAuth one.
   protect_from_forgery if: -> { current_user.present? }, with: :exception
 
+  def wrc
+    competition = competition_from_params
+    delegate_report = competition.delegate_report
+
+
+    render json: wrc_information.to_json
```
gregorbg, line 12: What is `wrc_information` and where does it come from?

app/controllers/api/v0/competitions_controller.rb
```diff
@@ -4,6 +4,14 @@ class Api::V0::CompetitionsController < Api::V0::ApiController
   # Enable CSRF protection if we use cookies based user instead of OAuth one.
   protect_from_forgery if: -> { current_user.present? }, with: :exception
 
+  def wrc
```
gregorbg, line 7: What does this route do?

gregorbg (changes requested):

app/controllers/api/v0/competitions_controller.rb
dunkOnIT: This was just from playing around with payload generation

app/controllers/api/v0/competitions_controller.rb
dunkOnIT: See other comment

env_config.rb
gregorbg, line 1: Don't add the username and password to `env_config`, add them to `secrets` instead using the `vault :FOO_BAR` convention.

(And don't forget to actually upload to Vault before/upon deployment of course)

gregorbg (changes requested):

env_config.rb
dunkOnIT: user/pass moved to `secrets`, and uploaded to Vault

dunkOnIT:
Test failure appears to be completely unrelated? Let me know if there are any suspicions for what's going wrong, otherwise I can try and spend a bit of time on it

env_config.rb
gregorbg, line 1: Right now you're only adding the value for non-production environments. You definitely need to also add it in prod, otherwise the value won't "exist" there in terms of the code not being able to access it even if it is properly defined in Terraform.

gregorbg (changes requested):

env_config.rb
dunkOnIT: Fixed, and defined the variable in what I think is the right terraform file as well

env_config.rb
```diff
@@ -19,6 +19,7 @@
     mandatory :DATABASE_AWS_REGION, :string
     mandatory :DATABASE_WRT_USER, :string
     optional :PAYPAL_BASE_URL, :string ## TODO: Change to mandatory when launching paypal
+    optional :WRC_WEBHOOK_URL, :string, ''
```
gregorbg, line 22: Not sure if I'm happy with `optional` in the prod context, we usually want the startup to crash (and thus fail the B/G deployment, causing no apparent harm to the Average Joe website user) when we forgot to configure something

infra/wca_on_rails/production/variables.tf
```diff
@@ -108,3 +108,9 @@ variable "shared" {
   })
   description = "The shared resources between Environments"
 }
+
+variable "WRC_WEBHOOK_URL" {
+  description = "The URL to send delegate report webhook notifications for WRC to"
+  type = string
+  default = "https://joba.me/wca/reports_webhook"
```
gregorbg, line 115: Summoning @FinnIckler regarding Terraform specifics but I feel like we have different files for the different envs and we don't generally use the default value here...?

env_config.rb
```diff
@@ -19,6 +19,7 @@
     mandatory :DATABASE_AWS_REGION, :string
     mandatory :DATABASE_WRT_USER, :string
     optional :PAYPAL_BASE_URL, :string ## TODO: Change to mandatory when launching paypal
+    mandatory :WRC_WEBHOOK_URL, :string, ''
```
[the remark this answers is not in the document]
dunkOnIT: Damn ofc, classic Duncan behaviour

app/models/delegate_report.rb
```diff
@@ -86,4 +86,19 @@ def posted=(new_posted)
     self.posted_at = (new_posted ? Time.now : nil)
     self.posted_by_user_id = current_user&.id
   end
+
+  def feedback_requests
```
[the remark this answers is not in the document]
dunkOnIT: Good point - added

gregorbg (approved):
Fine by me, but please wait for Finn to approve the Terraform vars

infra/wca_on_rails/production/variables.tf
FinnIckler: yeah, I will create it in the different files

FinnIckler:
I added the environment variables and ran terraform

PR #10022, opened 2024-10-02
Add Registration V2 API Routes to the monolith

FinnIckler:
I still need to get payments to work. Creating as Draft to allow Duncan to add the registration checker and route tests.

I will also revert all the javascript changes in the end (it's just so I can test that the frontend works with the migrated routes).


FinnIckler:
Now that I refactored the update method to please rubocop, you can see that the tests just pass, as all factories still think they are running on v1

app/controllers/api/v1/registrations/registrations_controller.rb
```diff
@@ -35,23 +33,27 @@ def show
 
   def create
     # Currently we only have one lane
+    # QUESTION: Should we be calling registration_params[:competing]?
     if params[:competing]
+      # QUESTION: Should this not be params[:competing].permit?
```
FinnIckler, line 38: No, because guests are currently not in the competing object

app/controllers/api/v1/registrations/registrations_controller.rb
```diff
@@ -35,23 +33,27 @@ def show
 
   def create
     # Currently we only have one lane
+    # QUESTION: Should we be calling registration_params[:competing]?
```
FinnIckler, line 36: we could, but the I think we can just remove `registration_params` as this is going to be dynamic depending on the lanes used

lib/wca_exceptions.rb
```diff
@@ -32,4 +32,23 @@ def initialize(error_str)
...
+      @error = error
+      @data = data
+    end
+  end
+
+  class BulkUpdateError < StandardError
+    attr_reader :status, :errors
+
+    def initialize(status, errors)
+      @status = status
```
FinnIckler, line 50: why did you not super here?

dunkOnIT: I can't remember the exact chain of reasoning, but because `errors` is an array not a string it produces an error if I call super (as you can't call super without arguments). This is the same workaround I used in the microservice, but we can tweak the design a bit more if needed

.rubocop.yml
```diff
@@ -235,6 +235,10 @@ Naming/BlockForwarding:
 Style/ArgumentsForwarding:
   RedundantBlockArgumentNames: []
 
+Lint/MissingSuper:
+  Exclude:
+    - 'lib/wca_exceptions.rb'
+
```
gregorbg, line 241: Veeeeeery unhappy with this. Strong bias towards finding a better solution that doesn't need a Rubocop rule, unless it's unreasonably painful and costs you days upon days.

FinnIckler: agree! I'm pretty sure this is easily done

dunkOnIT: Ok I've fixed it - I admit I was being lazy

config/locales/en.yml
```diff
@@ -1898,6 +1898,7 @@ en:
         event_limit: "This competition limits the amount of events you can register for to %{max_events}."
       errors:
         events: "Loading Events failed, please try again"
+        bulk_update: "%{count} errors detected in request. See response payload for more information."
```
FinnIckler, line 1901: This is only shown for API requests right? Otherwise this would be a pretty useless error message in the frontend.

gregorbg (approved):

PR #10024, opened 2024-10-04
Make database setup optional when running rspec

dunkOnIT:
Just a draft PR for now. I'm doing a bit of profiling on our test startup, because even with SKIP_PRETEST_SETUP, we still take ~10 seconds which is much longer than I'd expect. Will use this PR to note my findings for future reference


dunkOnIT:
Whether or not SKIP_PRETEST_SETUP is set to true doesn't seem to be making much of a difference. I still suspect there's something going on with database cleaner.

dunkOnIT:
It's not the test file loading (bearing in mind that I'm only loading certain files): 
Loading test files...
/app/spec/requests/registrations/registration_payments_spec.rb loaded in 0.000468596 seconds
/app/spec/requests/registrations/registrations_spec.rb loaded in 0.041066249 seconds
All test files loaded in 0.041617773 seconds

gregorbg (approved):

PR #10026, opened 2024-10-04
Hotfix: Use a DST-less timezone in frontend tests

gregorbg:
California just turned their clocks due to Daylight Savings (at the time of writing this description message). This causes our test suite to fail "out of nowhere", because the string literaly "Pacific Daylight Time" cannot be found in the frontend.

Also sneaking in a tiny improvement to how available/eligible timezones are being passed from the backend to the frontend, stolen from https://github.com/thewca/worldcubeassociation.org/pull/9787

### DIFF
--- a/app/models/competition.rb
+++ b/app/models/competition.rb
@@ -1025,13 +1025,11 @@ class Competition < ApplicationRecord
   end
 
   def country_zones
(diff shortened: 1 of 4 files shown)


dunkOnIT (approved):

dunkOnIT:
Did a basic test of schedule functionality and all seems to still work as expected!

PR #10028, opened 2024-10-05
Split out editing venues from editing schedule

kr-matthews:
image

image

The styling is a bit off with the "Venues + Add a venue" and the text at the top of each tab being a bit close to the content below (formerly there was an accordion).

Now each folder has an `index` and a (single) component subfolder with another `index`, so it might be worth bringing those subfolder contents up and having a single `index` (maybe as a follow-up PR?).

There's some minor duplication now, since everything was duplicated then trimmed away, mostly in the 2 top-level `index` files.

I removed a scss file because it didn't appear to be doing anything, but I'm not 100% confident in this.


app/webpacker/lib/utils/wcif.js
```diff
@@ -22,11 +23,11 @@ export function useSaveWcifAction() {
       options = {},
       onError = alertWcifError,
     ) => {
-      const url = `/api/v0/competitions/${competitionId}/wcif`;
+      const url = patchWcifUrl(competitionId, patchOpts);
 
       save(url, wcifData, onSuccess, options, onError);
     },
-    [save],
+    [save, patchOpts],
```
kr-matthews, line 30: You're passing `patchOpts` as an object (`useSaveWcifAction({ skipSchedule: true })`) which means this will be a fresh object (reference) each render, so this `saveWcif` will re-compute every time, negating the use of `useCallback`. If any code is relying on `saveWcif` never re-computing (i.e. using `saveWcif` in a dependency array somewhere), then this will break that code.

This can be avoided by defining `const wcifSaveOptions = { skipSchedule: true }` outside the component. But it doesn't prevent anyone in the future from passing in an object 'incorrectly.'

(I'm not sure if the default value `{}` is a fresh object each time?)

Making the options a parameter to `save`, instead of the hook itself, might work.

kr-matthews (commented):
I tried and **nothing** on the venues page nor on the schedule page saves - when I refresh, the original (old) data reappears.
I can successfully edit events.

app/webpacker/lib/utils/wcif.js
gregorbg: Fixed! Should now be less recompute-heavy. And this incidentally also fixes the error with the backend not saving, because the PATCH request was accidentally passing `undefined` in the URL under certain circumstances.

kr-matthews: Everything looks good to me, and works well as far as I can tell.

I wonder if specifying what _should_ be saved, rather than what should _not_ be saved, might be better though? In the past I have been editing groups in delegate dashboard, and also editing the schedule on the site, and overwritten myself, for example.

kr-matthews: No wait, adding rooms might still cause conflicts when editing in different browser tabs. I will investigate more tomorrow. But your changes certainly haven't made anything worse.

kr-matthews: Things that can go wrong with 2 tabs open for venues/schedule respectively:
- Delete a venue/room & save, then edit the schedule in that room & save -> 'invalid json' error
- Copy an existing room (in venue editor) & save, then edit anything in the schedule & save -> new copied room is gone
- Maybe other scenarios

Simple things like editing the fields of venues/rooms while editing the schedule _seem_ to work fine, as far as I can tell.

kr-matthews:
What's the plan with this? As expected, there are conflicts due to #9787.

gregorbg:
I think I managed to resolve all merge conflicts. Still want to do more manual testing next week, and perhaps get a tad bit of feedback from Delegates/Organizers before merging this.

Also, I know the current WCIF parameter API is not ideal, so I wanna wait a few more days to see if I can come up with something.

kr-matthews:
Should I close this and forget about it, or resolve the conflicts?

PR #10049, opened 2024-10-07
Execute check records script from Read Replica

gregorbg:
Attempt to accelerate Check Records by running the expensive read operation off the Read Replica.

Does not work (quite as easily) for Create Newcomers, because that requires immediate write operations in batches which are reflected in subsequent read operations. (In other words, the data written immediately needs to be read).

Check Records however reads all the data in one batch, lets WRT assign overrides, and "just" stores them. No pagination or anything.

### DIFF
--- a/app/models/check_regional_records_form.rb
+++ b/app/models/check_regional_records_form.rb
@@ -10,6 +10,8 @@ class CheckRegionalRecordsForm
   end
 
   def run_check
-    CheckRegionalRecords.check_records(self.check_event_id, self.competition_id)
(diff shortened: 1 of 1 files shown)


FinnIckler:
What about adding the optimized query that Eleanor made from the WRT thread?

gregorbg:
That requires significantly more effort to implement properly. This PR is more of a low-hanging fruit

FinnIckler (approved):
fair, LGTM!

PR #10051, opened 2024-10-08
Trying to get capybara working locally

dunkOnIT:


dunkOnIT:
Try and use the approach in #10003 

Here's Github's install script for chromedriver: https://github.com/actions/runner-images/blob/main/images/ubuntu/scripts/build/install-google-chrome.sh#LL14

dunkOnIT:
We're replacing apparition with playwright, making this unnecessary

PR #10052, opened 2024-10-09
Switched competition_id foreign key in microservice_registrations comp cascade

dunkOnIT:
As discussed in meeting, this fixes an issue where competition id's could not be updated by admins if a microservice registration existed for them.

Tested locally and works as expected.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.2].define(version: 2024_09_30_132809) do
+ActiveRecord::Schema[7.2].define(version: 2024_10_09_111904) do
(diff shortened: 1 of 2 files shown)


db/schema.rb
```diff
@@ -1318,7 +1318,7 @@
 
   add_foreign_key "active_storage_attachments", "active_storage_blobs", column: "blob_id"
   add_foreign_key "active_storage_variant_records", "active_storage_blobs", column: "blob_id"
-  add_foreign_key "microservice_registrations", "Competitions", column: "competition_id"
+  add_foreign_key "microservice_registrations", "Competitions", column: "competition_id", on_update: :cascade
```
gregorbg, line 1321: Now you were just lazy and didn't actually run it, so the delete cascade doesn't show in `schema.rb` :P

gregorbg (changes requested):

db/schema.rb
dunkOnIT: :innocent:

gregorbg (approved):

PR #10058, opened 2024-10-11
Create Shoryuken Worker configuration

FinnIckler:
This would be the configuration for shoryuken, no service for the worker yet, but I can add that on tuesday. 
I added the gem in the production section only.


FinnIckler:
https://github.com/ruby-shoryuken/shoryuken/pull/777
There is a PR out there to get the Rails 7.2 Support for Shoryuken going (sadly hasn't been looked at yet). At least it looks to be super easy

gregorbg (approved):

PR #10076, opened 2024-10-15
Filter v2 registrations correctly 

### LINKED ISSUE #10045: `Past Competitions` list: Don't show comps where status is pending, cancelled or rejected

FinnIckler:
We had the TODO there for a while now. fixes #10045


gregorbg:
We kinda get that "for free" when the microservice backporting is done, don't we?

FinnIckler:
yes we do

PR #10080, opened 2024-10-16
New comp overview: Filter for events and cancelled status in the backend

gregorbg:
Previously, we still did events and cancelled filters in the frontend, and everything else in the backend.
With this change, everything should be backend-based for the overview page.


gregorbg:
Pure backend fix, merging to facilitate launch of newer PRs

Issue #10073, opened 2024-10-15
Give regional delegates access to the "admin view" of delegates' competitions tasks


mirandapedro:
I can't link it or describe it precisely, but senior delegates have a page (which I believe might just be the standard competition page) in which they can see the status of delegates' tasks in past competitions. If the results and report are posted, when were they posted, and if delegates were late on the tasks.

 Giving regional delegates access to this view/page would be useful to allow them to oversee delegates in their region and make better decisions on promotions from junior to full delegates.

Instead of recommending a promotion only for teams to shoot it down (e.g. due to tardiness), regional delegates could make a more informed recommendation to the senior delegate/team/weat.

Issue #10072, opened 2024-10-15
Give regional delegates access to unconfirmed competitions from their region


mirandapedro:
Currently, regional delegates do not have access to unconfirmed/draft competitions which delegates in their region are finalising before sending them over to the WCAT. This prevents them from supervising competitions in this stage and advising delegates on unconfirmed competitions if needed.

Giving regional delegates access to unconfirmed competitions would solve this by allowing regional delegates to oversee all competitions in their area from the get-go.

I believe Senior Delegates can already access these competitions.

PR #10081, opened 2024-10-16
Give more competition permissions to Regional Delegates

gregorbg:
With this PR, Regional Delegates will have permission to...
1. Access the 'Admin' part of the Competitions Overview page (which Senior Delegates already have access to)
2. Manage competitions of their subordinate Delegates (most notably, the "Edit Competition" form) the same way that Senior Delegates already can

In short, this PR gives Regional Delegates some access rights which Senior Delegates already have.

Fixes https://github.com/thewca/worldcubeassociation.org/issues/10073
Fixes https://github.com/thewca/worldcubeassociation.org/issues/10072

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -836,6 +836,7 @@ class User < ApplicationRecord
       competition.organizers.include?(self) ||
       competition.delegates.include?(self) ||
       competition.delegates.flat_map(&:senior_delegates).compact.include?(self) ||
+      competition.delegates.flat_map(&:regional_delegates).compact.include?(self) ||
       wic_team?
     )
   end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -492,6 +492,10 @@ class User < ApplicationRecord
     senior_delegate_roles.any?
   end
 
+  def regional_delegate?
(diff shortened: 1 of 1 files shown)


EdHollingdale:
Why do Regional Delegates need more permissions? The website should be reducing the number of people with various permissions rather than increasing it IMO.

mirandapedro:
Thank you so much for working on this, Gregor!

Ed, please see the issues Gregor mentioned for more details. In short, as a regional delegate, I feel like these permissions would be useful for:
- allowing regional delegates to help delegates in their region whenever necessary (e.g. reviewing details of a competition before it's confirmed)
- giving oversight tools to the regional delegate (e.g. making sure deadlines are met, easily accessing information for promotions)

EdHollingdale:
Thanks Pedro. They seem like sensible reasons - now that you point it out I think Regional Delegates being able to help via a review of a competition before it is announced makes sense. As does having more visibility of the upcoming competitions for their subordinate Delegates.

danieljames-dj (approved):

KerrieJarman:
The Board is supportive of this change

PR #10082, opened 2024-10-16
Fix competition form not rendering if the id hasn't been set yet

FinnIckler:


lib/microservices/registrations.rb
```diff
@@ -85,6 +85,7 @@ def self.update_registration_payment(attendee_id, payment_id, iso_amount, curren
     # rubocop:enable Metrics/ParameterLists
 
     def self.competitor_count_by_competition(competition_id)
+      return 0 if competition_id.nil? || competition_id.empty?
```
gregorbg, line 88: alternative: `return 0 unless competition_id.present?`

FinnIckler: true, I could have also used `blank?`

FinnIckler:
tests seems to fail because of some mailing list hotfix?

FinnIckler:
merging as a hotfix as the failed tests are also on main and we want to allow people to create competitions again

Issue #10085, opened 2024-10-16
Country name displayed in random language (v2 registrations list)


maxidragon:
It doesn't happen on every load, my friend told me about it and I encountered it after a few refreshes. 

Note: It has happened only after changing locale to PL (I assume it may happen for all locales except English??)

![image](https://github.com/user-attachments/assets/2df3b14c-542a-4e1a-8c51-f94497774c50)

FinnIckler:
I assume a simple one line change [here](https://github.com/thewca/worldcubeassociation.org/blob/30de3c6d82d100ebe026be6dfbf591645db97c19/app/webpacker/components/RegistrationsV2/Registrations/RegistrationList.jsx#L258) to not use the name, but to use the I18n mapping from iso2 will suffice. I can do that tomorrow

maxidragon:
This issue can be closed as #10087 has been merged.

PR #10087, opened 2024-10-17
Fix countries being displayed in random languages for V2 Lists

FinnIckler:
Same issue as we had with the activity names. Hope this is the last spot we use i18n strings generated from the backend in API calls


FinnIckler:
fixes #10085

PR #10089, opened 2024-10-17
Move Computing Record Markers to a job

FinnIckler:
This seems to work, but I wouldn't mind some more thorough testing my WRT. Also open to any API changes on the Job. 
Also this currently does not have re-run functionality, as I wasn't sure if we need it


FinnIckler:
@thewca-bot deploy staging

FinnIckler:
Superseeded

PR #10091, opened 2024-10-18
Do not show the main event warning if competition has only one event

maxidragon:
See title. It's quite clearly that if the event is the only event, it was treated in a special way and there is no need to show this warning.


danieljames-dj:
Hi @maxidragon can you please let me know if this was discussed in any WRT emails or slack? I've been away for few days, so might have missed if there was one.

This is really good change actually and I'm totally fine with this change if this has been already informed to the team.

maxidragon:
Yeah, this is the request from @c-goodyear

danieljames-dj (approved):

PR #10092, opened 2024-10-19
Distinguish between V1 and V2 competitions for Competition ID changes

gregorbg:
Follow-up to https://github.com/thewca/worldcubeassociation.org/pull/10059.

Makes it so that V1 competitions can change IDs freely, just as they used to.
The "don't change ID when V2" limitation now actually checks for V2 reg system. If yes, then registrations are allowed only after announcement of the competition.

### DIFF
--- a/app/models/competition.rb
+++ b/app/models/competition.rb
@@ -614,7 +614,14 @@ class Competition < ApplicationRecord
   end
 
   def user_can_pre_register?(user)
-    (delegates.include?(user) || trainee_delegates.include?(user) || organizers.include?(user)) && self.confirmed_or_visible?
+    # The user has to be either a registered Delegate or organizer of the competition
+    is_competition_manager = delegates.include?(user) || trainee_delegates.include?(user) || organizers.include?(user)
+    # We allow pre-registration at any time for the old registration system (because we have control over the foreign keys there)
+    #   Otherwise, if it's using the new system, we only allow registrations when it's announced
+    #   because the probability of an ID change is pretty low in that case and when it does happen, WST can handle it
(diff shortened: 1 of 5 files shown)


gregorbg:
> Am I correct that we're removing `usesV2Registrations` from the React hydration because it is instead being retrieved in the competition JSON? ie, in the `useFormData` method?

Yes, almost correct. The crucial detail you're missing (and that I have been missing as well) is that the question "does this competition use V2 reg?" is answered in the JSON-ified competition already. In `to_form_data`, there is a field `usesV2Registrations` which contains exactly this bit of information.
So in the React hydration, we implicitly pass the information "does it use V2?" already simply by passing the form-ified competition. It was redundant all along to pass this information separately and nobody on the team has noticed that until now :P

gregorbg:
In other words, the reason why we even can retrieve it from `useFormData` is because we're passing it in the form-ified `to_form_data` competition during the React component hydration.

dunkOnIT:
Cool that makes sense! I tested this locally, and was still able to change the ID of a v2 competition after announcement even if it had a registration. Will deploy to staging and see if I can replicate there.

dunkOnIT:
@thewca-bot deploy staging

gregorbg:
> Cool that makes sense! I tested this locally, and was still able to change the ID of a v2 competition after announcement even if it had a registration. Will deploy to staging and see if I can replicate there.

I assume you mean "change the ID of a v1 competition"?!

dunkOnIT:
No, I meant v2 competition - ie, there was a bug. However, deploying to staging I wasn't able to change a v2 competition after announcement - so I'm pretty sure there was an issue with my local setup/caching

PR #10095, opened 2024-10-21
Fix for users self-deleting their avatar

gregorbg:
Was accidentally fired against the wrong endpoint, resulting in a silent error.
Now using the correct endpoint and also a bit of cleanup convenience.


gregorbg:
Merging as hotfix

PR #10096, opened 2024-10-21
Index avatars by their own ID instead of user WCA ID

gregorbg:
see title. Now that our storage mechanism doesn't depend on WCA ID anymore, we already have users uploading profile pictures which don't have an ID attached. This causes errors in the admin submission form because `null` WCA IDs get converted to the empty string, making the form part `avatars[2012BILL01][action] = :approve` look like `avatars[][action] = :reject`.


gregorbg:
Merging as hotfix

PR #10117, opened 2024-10-21
Panel page to anonymize account-only users

danieljames-dj:
Currently only users with WCA ID can be anonymized and for account-only anonymization we need to use PMA.

Also, the AnonymizeUser component is made in such a way that this can be integrated in tickets page as well. That is the reason why AnonymizeUser and AnonymizationScriptPage are not made into single component.

There is also a future plan of integrating AnonymizeProfile (which is currently in rails) to AnonymizationScriptPage and then there will be only one anonymization script.


app/webpacker/components/Panel/pages/AnonymizationScriptPage/index.jsx
```diff
@@ -0,0 +1,32 @@
...
+  return userId
+    ? (
+      <AnonymizeUser
+        userId={userId}
+        onSuccess={() => setSuccess(true)}
+      />
+    )
+    : (
+      <IdWcaSearch
+        label="Enter the user to ban"
```
maxidragon, line 25:
```suggestion
        label="Enter the user to anonymize"
```

app/controllers/panel_controller.rb
```diff
@@ -46,4 +53,22 @@ def generate_db_token
       replica: 2,
     }
   end
+
+  def anonymize_user
+    user_id = params.require(:userId)
+    user = User.find(user_id)
+    user.update!(
```
gregorbg, line 60: If this `update!` fails, Rails will throw a hard error. This hard error will render an HTML error page, which JSON endpoints cannot handle very well.

A better pattern is to use `user_changed = user.update(...)`: When you use `update` method without the `!`, it will return a boolean indicating whether or not the validations that run while saving are successful.
Then you can react like `render json: { success: user_changed }`, and if `user_changed` is false you could even include `user.errors` to communicate the validation errors as a "clean" JSON to the frontend

app/webpacker/components/Panel/pages/AnonymizationScriptPage/AnonymizeUser.jsx
```diff
@@ -0,0 +1,58 @@
...
+import { apiV0Urls, actionUrls, adminAnonymizePersonUrl } from '../../../../lib/requests/routes.js.erb';
+import Loading from '../../../Requests/Loading';
+import Errored from '../../../Requests/Errored';
+import useSaveAction from '../../../../lib/hooks/useSaveAction';
+
+const ANONYMOUS_ACCOUNT_EMAIL_ID_SUFFIX = '@worldcubeassociation.org';
+const ANONYMOUS_ACCOUNT_NAME = 'Anonymous';
+const ANONYMOUS_ACCOUNT_DOB = '1954-12-04';
+const ANONYMOUS_ACCOUNT_GENDER = 'o';
+const ANONYMOUS_ACCOUNT_COUNTRY_ISO2 = 'US';
```
gregorbg, line 13: Can you export them from Rails to the frontend using our ugly `wca-data.js.erb` bridge?

app/webpacker/components/Panel/pages/AnonymizationScriptPage/AnonymizeUser.jsx
```diff
@@ -0,0 +1,58 @@
...
+            ? 'The user has both account and profile.' : 'The user has only account.'}
+        </List.Item>
+      </List>
+      {accountAnonymized && <Message positive>The user is already anonymous.</Message>}
+      {!accountAnonymized && userData.wca_id && (
+        <Button href={adminAnonymizePersonUrl}>Go to Anonymize Person script</Button>
+      )}
+      {!accountAnonymized && !userData.wca_id && (
+        <Button onClick={anonymizeAccountAction}>Anonymize account</Button>
+      )}
```
gregorbg, line 55: This can be `accountAnonymized ? (<SomeComponent />) : (<><SomeOtherButton label="Foo" /><SomeOtherButton label="Bar" /></>)`

app/controllers/api/v0/users_controller.rb
```diff
@@ -17,10 +17,16 @@ def show_user_by_id
   def show_users_by_id
     user_ids = params.require(:ids)
     users = User.where(id: user_ids)
+    private_attributes = []
+    if current_user && current_user.can_admin_results?
+      # These attributes are needed to check anonymization status during anonymization.
+      private_attributes = %w[dob email]
+    end
```
gregorbg, line 24: I'm not too wild about the way you're "hacking" confidential data into this endpoint. Where is it being called?

gregorbg (changes requested):

app/controllers/api/v0/users_controller.rb
danieljames-dj: I have changed the way this is done, now no more needed.

app/controllers/panel_controller.rb
```diff
@@ -8,8 +8,15 @@ class PanelController < ApplicationController
...
   before_action -> { redirect_to_root_unless_user(:has_permission?, 'can_access_panels', :staff) }, only: [:staff]
   before_action -> { redirect_to_root_unless_user(:has_permission?, 'can_access_panels', :admin) }, only: [:generate_db_token]
+  before_action -> { redirect_to_root_unless_user(:has_permission?, 'can_access_panels', :wrt) }, only: [:anonymize_user]
   before_action -> { redirect_to_root_unless_user(:can_access_senior_delegate_panel?) }, only: [:pending_claims_for_subordinate_delegates]
 
+  ANONYMOUS_ACCOUNT_EMAIL_ID_SUFFIX = '@worldcubeassociation.org'
+  ANONYMOUS_ACCOUNT_NAME = 'Anonymous'
+  ANONYMOUS_ACCOUNT_DOB = '1954-12-04'
+  ANONYMOUS_ACCOUNT_GENDER = 'o'
+  ANONYMOUS_ACCOUNT_COUNTRY_ISO2 = 'US'
```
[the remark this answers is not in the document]
danieljames-dj: Okay, moved it to `user` model.

app/controllers/panel_controller.rb
danieljames-dj: Okay, thanks. Implemented.

app/webpacker/components/Panel/pages/AnonymizationScriptPage/AnonymizeUser.jsx
danieljames-dj: No longer needed in frontend.

app/webpacker/components/Panel/pages/AnonymizationScriptPage/AnonymizeUser.jsx
danieljames-dj: I've changed the UI completely. So guess this is no longer needed.

app/controllers/users_controller.rb
```diff
@@ -1,9 +1,10 @@
 # frozen_string_literal: true
 
 class UsersController < ApplicationController
-  before_action :authenticate_user!, except: [:select_nearby_delegate, :acknowledge_cookies]
+  before_action :authenticate_user!, except: [:select_nearby_delegate, :acknowledge_cookies, :anonymize]
```
gregorbg, line 4: Why are you skipping authentication for the `anonymize` step?
It most definitely should only be available to authenticated users

app/webpacker/components/Panel/pages/AnonymizationScriptPage/index.jsx
gregorbg, line 1: I don't like the notion of "swapping out" the searchbox with the actual content.

What if you have several people with similar names and then you mis-click in the dropdown? You would have to reload the whole page to fix your mistake.

gregorbg: Always displaying the search box, and then only showing the workbench if a user has been selected, seems like a better alternative

gregorbg: Or, even better but only possible if it makes sense for your component: Always show your workbench, but only in a blank/empty state if no user is selected. But I'll be happy with the previous version as well, if this idea is too complicated

app/webpacker/components/Tickets/TicketWorkbenches/AnonymizationTicketWorkbench/AccountAnonymization.jsx
```diff
@@ -0,0 +1,40 @@
...
+      null,
+      () => setCompleted(true),
+      { method: 'POST' },
+    );
+  };
+
+  return (
+    <>
+      <Header as="h4">Account anonymization</Header>
+      {completed && <p>Account anonymization completed.</p>}
```
gregorbg, line 27: This could use a `Message` component with `success` prop.

app/controllers/users_controller.rb
```diff
@@ -373,4 +374,24 @@ def acknowledge_cookies
     end
     true
   end
+
+  def anonymize
```
gregorbg, line 378: Regarding your WRT workflows, how does this relate to anonymizing a `Person`?
I know that your existing anon logic is quite complex in terms of "does this person have posts on the forum, which comps is this person registered for, etc". These are all checks which _technically_ belong to `User`, not to Person. But you're running them all in the anonymizePerson step, so I want to make sure that this "simple" user anonymization here doesn't break any existing assumptions or workflows.

gregorbg (changes requested):

app/webpacker/components/Panel/pages/AnonymizationScriptPage/index.jsx
danieljames-dj: Okay, I like the idea. Implemented.

app/webpacker/components/Tickets/TicketWorkbenches/AnonymizationTicketWorkbench/AccountAnonymization.jsx
danieljames-dj: Done

gregorbg (approved):

PR #10120, opened 2024-10-21
Speed up records precompute query

gregorbg:
I feel a little bit sorry for Eleanor after she spent so much time on much more sophisticated optimizations...

But it turns out that this plain and simple restructuring speeds up the query by factor ~3 on my local system. So I'll take that as a win for the time being, and continue to work on a proper table caching mechanism while this PR buys us a few days time.

# Before
```sql
SELECT r.eventId, r.countryId, MIN(r.best) AS `value`
FROM (
  SELECT `Results`.`eventId`, `Results`.`competitionId`, `Results`.`countryId`, `Results`.`best`
  FROM `Results`
  WHERE `Results`.`best` > 0
) AS r
INNER JOIN (
  SELECT `Competitions`.`id`
  FROM `Competitions`
  WHERE `Competitions`.`end_date` < '2013-06-01' # This seemingly hard-coded value is read directly from the `Competitions` table one step before
) AS c
  ON c.id = r.competitionId
INNER JOIN (
  SELECT `competition_events`.`event_id`
  FROM `competition_events`
  WHERE `competition_events`.`competition_id` = 'ErfurtOpen2013'
) AS ce
  ON ce.event_id = r.eventId
GROUP BY r.eventId, r.countryId
```

Runtime on my local machine: `2301.3ms`

# After
```sql
SELECT eventId, Results.countryId, MIN(best) AS `value`
FROM `Results`
INNER JOIN `Competitions` `competition`
  ON `competition`.`id` = `Results`.`competitionId`
WHERE `Results`.`best` > 0
  # The following two values are both fetched directly from the `Competitions` table immediately before this query is being executed
  AND `Results`.`eventId` IN ('333', '222', '444', '555', '666', '777', '333bf', '333fm', '333oh', 'clock', 'minx', 'pyram', 'sq1')
  AND `competition`.`end_date` < '2013-06-01'
GROUP BY `Results`.`eventId`, `Results`.`countryId`
```

Runtime on my local machine: `756.6ms`

### DIFF
--- a/db/migrate/20241021145709_add_results_country_event_index.rb
+++ b/db/migrate/20241021145709_add_results_country_event_index.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddResultsCountryEventIndex < ActiveRecord::Migration[7.2]
+  def change
+    add_index :Results, [:eventId, :countryId]
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.2].define(version: 2024_10_09_111904) do
+ActiveRecord::Schema[7.2].define(version: 2024_10_21_145709) do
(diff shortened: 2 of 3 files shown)


gregorbg:
Superseded by 10122

PR #10121, opened 2024-10-22
Basic implementation to block banned user Discourse SSO

dunkOnIT:
This is a very manual and blunt method - reasoning as follows:
1. WIC has waited a long time for this feature, and there are competitor safety concerns that should be addressed
2. I think a more elegeant solution will probably take a couple of weeks of deliberation, given that it is neither high priority nor trivial to change. 

### Suggested Changes to Handling of Banned Competitors
Expanding on (2) above, currently we have an enum which defines the scope of a user's ban. I think an enum is the wrong approach because it scales badly if we need to add more scopes (there are n! permutations, I think). Very reasonably, we could have the following scopes:
- forums
- attending
- competing
- staffing

Which leads to like ~10 permutations or something (given that attending would cover competing and staffing).

Instead, I think we should have a boolean column for each ban scope, and let WIC set these as they desire. (With perhaps some logic to automatically tick mark `competing` and `staffing` as true if `attending` is selected).

Assuming we agree on this design, it will require: 
1. Confirmation with WIC (should be quick)
2. Migration for the banned_competitor_metadata model
3. Population of the ban scopes for banned competitors
4. Change to the banned competitors list display to show scopes not enums
5. Change to the ban competitor interface to show scopes not enums

None of this is impossible, but I don't think we'll get it all done this week, and having a hotfix in place in the meantime will remove some of the pressure. I am still happy to take responsibility for making sure that our temporary solution stays temporary, though.

### DIFF
--- a/.env.test
+++ b/.env.test
@@ -35,3 +35,4 @@ STAGING_OAUTH_SECRET=example-secret
 AVATARS_PUBLIC_STORAGE=local
 AVATARS_PRIVATE_STORAGE=local_private
 DUMP_HOST=https://assets.worldcubeassociation.org
+BANNED_FORUM_USER_IDS=777777,777778
--- a/.env.test
+++ b/.env.test
@@ -35,4 +35,3 @@ STAGING_OAUTH_SECRET=example-secret
 AVATARS_PUBLIC_STORAGE=local
 AVATARS_PRIVATE_STORAGE=local_private
 DUMP_HOST=https://assets.worldcubeassociation.org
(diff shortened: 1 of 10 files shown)


FinnIckler:
Do you want to add a starburst alert to signal to users that they are banned?

dunkOnIT:
Good idea - starburst added

dunkOnIT:
@thewca-bot deploy staging

gregorbg (approved):

PR #10122, opened 2024-10-22
Move Regional Records computation to Auxiliary table

gregorbg:
Supersedes https://github.com/thewca/worldcubeassociation.org/pull/10089, although I'm very grateful for the initial inspiration :smile: 
Supersedes https://github.com/thewca/worldcubeassociation.org/pull/10120

The basic problem is that the query to pre-load existing records was breaking our necks. The simple solution of this PR is to combine `Results` data and `Competitions` data into one common, `JOIN`ed table which can be indexed.
That way, we can essentially reuse the same queries as before (with minor modifications to the table name etc.) but the indexes will make them reasonably fast.

# Before
```sql
SELECT r.eventId, r.countryId, MIN(r.best) AS `value`
FROM (
  SELECT `Results`.`eventId`, `Results`.`competitionId`, `Results`.`countryId`, `Results`.`best`
  FROM `Results`
  WHERE `Results`.`best` > 0
) AS r
INNER JOIN (
  SELECT `Competitions`.`id`
  FROM `Competitions`
  WHERE `Competitions`.`end_date` < '2024-08-10' # This seemingly hard-coded value is read directly from the `Competitions` table one step before
) AS c
  ON c.id = r.competitionId
INNER JOIN (
  SELECT `competition_events`.`event_id`
  FROM `competition_events`
  WHERE `competition_events`.`competition_id` = 'StevenageAugust2024'
) AS ce
  ON ce.event_id = r.eventId
GROUP BY r.eventId, r.countryId
```

Benchmark on my local machine: 11390.9ms

# After

:warning: Please do not be confused by the table being referenced as `Results`. This is a necessary hack because otherwise the Rails model engine gets confused! It is only an alias that is freely choosable, and in terms of understanding the contents of this query, you can mentally replace `Results` with `pancake`.

```sql
SELECT eventId, countryId, MIN(best) AS `value`
FROM regional_records_lookup AS Results # This alias is here only to make sure that Ruby doesn't get confused
WHERE `Results`.`best` > 0
  AND `Results`.`competitionEndDate` < '2024-08-10'
  AND `Results`.`eventId` IN ('333', '222', '444', '555', '666', '777', '333bf', '333fm', '333oh', 'clock', 'minx', 'pyram', 'sq1')
GROUP BY `Results`.`eventId`, `Results`.`countryId`
```

The relevant index is on `(eventId, countryId, competitionEndDate, best)` (as well as `(eventId, countryId, competitionEndDate, average)` respectively for averages)

Benchmark on my local machine: 1804.1ms

### DIFF
--- a/app/views/admin/check_regional_records.html.erb
+++ b/app/views/admin/check_regional_records.html.erb
@@ -31,7 +31,10 @@
 
     <%= alert :warning do %>
       This script relies on auxiliary tables which are computed as part of <%= link_to "running CAD", admin_compute_auxiliary_data_path, target: :_blank %>.
-      <b>It will not deliver accurate results for specific competitions if CAD hasn't been run after the results posting.</b>
+      <br/>
+      <b>The detected records will not automatically be up-to-date with the <code>Results</code> table by default.</b>
+      <br/>
+      Instead, they are up to date with CAD (last successful run: <%= @cad_timestamp %>)
     <% end %>
 
     <hr/>
--- a/app/views/admin/check_regional_records.html.erb
+++ b/app/views/admin/check_regional_records.html.erb
@@ -42,6 +42,7 @@
     <%= simple_form_for @check_records_request, url: admin_override_regional_records_path, method: :get do |f| %>
       <%= f.input :competition_id, as: :competition_id, only_one: true, label: "Competition ID", hint: "Leave blank to check for all competitions" %>
       <%= f.input :event_id, as: :events_picker, allowed_events: Event.all, only_one: true, include_all: true, label: "Event", hint: "First symbol means 'all events'" %>
+      <%= f.input :refresh_index, as: :boolean, label: "Refresh index for selected competition(s)", hint: "This will attempt to compute the lookup on-the-fly only if a specific competition has been selected above. It can lead to timeouts, please rely on CAD when that happens." %>
 
       <%= f.button :submit, value: "Run check", class: "btn-primary" %>
     <% end %>
--- a/app/views/admin/check_regional_records.html.erb
+++ b/app/views/admin/check_regional_records.html.erb
@@ -29,6 +29,11 @@
       huge).
     </p>
 
+    <%= alert :warning do %>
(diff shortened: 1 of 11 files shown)


app/views/admin/check_regional_records.html.erb
```diff
@@ -29,6 +29,11 @@
       huge).
     </p>
 
+    <%= alert :warning do %>
+      This script relies on auxiliary tables which are computed as part of <%= link_to "running CAD", admin_compute_auxiliary_data_path, target: :_blank %>.
+      <b>It will not deliver accurate results for specific competitions if CAD hasn't been run after the results posting.</b>
```
danieljames-dj, line 34: Maybe replace "after the results posting" with "after the 'import results' step", so that it won't get confused with last step?

lib/db_helper.rb
```diff
@@ -10,4 +10,25 @@ def self.execute_sql(sql)
...
+
+    ActiveRecord::Base.connection.execute("CREATE TABLE #{temp_table_name} LIKE #{table_name}")
+    ActiveRecord::Base.connection.execute("ALTER TABLE #{temp_table_name} AUTO_INCREMENT = 1")
+
+    ActiveRecord::Base.transaction do
+      yield temp_table_name
+    end
+
+    # Atomically swap the tables
+    ActiveRecord::Base.connection.execute("RENAME TABLE #{table_name} TO #{old_table_name}, #{temp_table_name} TO #{table_name}")
```
danieljames-dj, line 26: Just curious - why don't we instead delete the table_name and rename the temp_table to table_name (in a single transaction)?

danieljames-dj (approved):

lib/db_helper.rb
gregorbg: Because in most database systems, statements that fundamentally alter the structure of the DB (like `CREATE TABLE`, `DROP TABLE`, etc.) implicitly cause a transaction commit. In other words, dropping a table inside a transaction and later rolling back the transaction to "undo dropping" is impossible.

For MySQL specifically, see https://dev.mysql.com/doc/refman/8.4/en/implicit-commit.html

PR #10123, opened 2024-10-22
Add V2 Polling Lambda

FinnIckler:
Split from #10058. Depends on that being merged first because of the SQS queue, but I have already deployed all the infrastructure to test, so this works. 

Review needed for the ruby changes, i.e the hook and the lambda function


app/jobs/add_registration_job.rb
```diff
@@ -2,7 +2,9 @@
 
 class AddRegistrationJob < ApplicationJob
   before_enqueue do |job|
-    Rails.cache.write("#{job.arguments[1]}-#{job.arguments[2]}-processing", true)
+    competition_id = job.arguments[1]
+    user_id = job.arguments[2]
```
gregorbg, line 6: Ruby 101: `competition_id, user_id = job.arguments`

FinnIckler: Ruby 102, arrays begin at 0 :P and I don't think `_,competition_id, user_id = job.arguments` looks that great?

FinnIckler: Alternatively I could change the order of the job params

gregorbg: Ruby 103: `competition_id, user_id = job.arguments.drop(1)`

gregorbg: And to be honest your underscore approach isn't too bad either.

FinnIckler: ok, if you fine with that, i changed it to that

gregorbg (approved):

PR #10124, opened 2024-10-23
Made status to english in role change mailer's metadata

danieljames-dj:


dunkOnIT:
Why is this needed?

danieljames-dj:
> Why is this needed?

Because in emails sometimes we are getting the status in different language - like when somebody is making the change from non-english locale.

dunkOnIT:
Ah I see! So we're essentially telling i18n "give us this key in English, not whatever language setting the user triggering it is using"?

Any suggest on how I can reproduce this locally to test?

danieljames-dj:
Correct. In local you can sign in as a WST senior (basically yourself), change to non English language, go to 'New Admin Panel' and go to "/panel/admin#groups-manager-admin" and promote someone from member to non-member.

(I've commented immediately after your comment, but somehow I added my comment after editing your comment and today when I checked I see my comment under your comment and I realized the mistake, really sorry for that 😅)

gregorbg (approved):

PR #10127, opened 2024-10-23
V2/V3 Compability Changes

FinnIckler:
Allows us to run V2 and V3 at the same time with the frontend choosing the correct route to hit depending on the registration_version field (from the other PR)


FinnIckler:
There is definitely much more clean up possible, but I think I should do that in a separate PR

env_config.rb
```diff
@@ -61,6 +60,11 @@
     mandatory :WCA_REGISTRATIONS_BACKEND_URL, :string
   end
 
+  if is_compiling_assets
```
gregorbg, line 63: Why only when compiling assets? Shouldn't these be available during runtime as well?

app/controllers/payment_controller.rb
```diff
@@ -5,15 +5,21 @@ def available_refunds
...
 
-      ms_registration = MicroserviceRegistration.includes(:competition, :payment_intents)
-                                                .find_by(competition_id: competition_id, user_id: user_id)
-      return render status: :bad_request, json: { error: "Registration not found" } unless ms_registration.present?
+      registration = if competition.registration_version_v3?
+                       Registration.includes(:payment_intents)
+                                   .find_by(competition_id: competition_id, user_id: user_id)
+                     else
+                       MicroserviceRegistration.includes(:payment_intents)
+                                               .find_by(competition_id: competition_id, user_id: user_id)
```
gregorbg, line 15: Can leverage https://github.com/thewca/worldcubeassociation.org/pull/10133 once that's merged

app/controllers/registrations_controller.rb
```diff
@@ -545,6 +545,7 @@ def stripe_webhook
           #   and Stripe tries again after an exponential backoff. So we (erroneously!) record the creation timestamp
           #   in our DB _after_ the backed-off event has been processed. This can lead to a wrong registration order :(
           stored_payment.update!(created_at: audit_event.created_at_remote)
+          stored_intent.holder.add_history_entry({ payment_status: stored_intent.wca_status, iso_amount: ruby_money.cents }, "stripe_webhook", audit_event.id, 'Payment')
```
gregorbg, line 548: Hmm... `Registration` already has a `record_payment` method, which currently is nothing more than a glorified "insert new registration_payment entity".

Could you look into possibly augmenting that method, so that it automatically also creates an history entry?
Or even more, could you consider adding `on_create` hooks to `RegistrationPayment` that automatically record some history entry? (The challenge there is probably going to be to pass along the webhook event that triggered the history write, so don't try too hard if the solution isn't obvious :sweat_smile: )

app/controllers/registrations_controller.rb
```diff
@@ -746,6 +748,7 @@ def refund_payment
         payment_record.registration_payment.id,
         current_user.id,
       )
+      registration.add_history_entry({ payment_status: "refund", iso_amount: amount_left }, "user", current_user.id, 'Refund')
```
gregorbg, line 751: Same as before, see whether it would make sense to merge this into the `record_refund` method directly above

app/controllers/registrations_controller.rb
```diff
@@ -621,6 +622,7 @@ def payment_completion
           charge_transaction,
           current_user.id,
         )
+        registration.add_history_entry({ payment_status: stored_intent.wca_status, iso_amount: ruby_money.cents }, "user", current_user.id, 'Payment')
```
gregorbg, line 625: (same as above about potetially merging this with `record_payment`)

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationList.jsx
```diff
@@ -219,8 +219,8 @@ export default function RegistrationAdministrationList({ competitionInfo }) {
               - DateTime.fromISO(b.competing.registered_on).toMillis();
           case 'paid_on_with_registered_on_fallback':
           {
-            const hasAPaid = a.payment.payment_status === 'succeeded';
-            const hasBPaid = b.payment.payment_status === 'succeeded';
+            const hasAPaid = a.payment?.has_paid === 'succeeded';
+            const hasBPaid = b.payment?.has_paid === 'succeeded';
```
gregorbg, line 223: These two comparisons are never gonna be true :stuck_out_tongue_closed_eyes:

env_config.rb
FinnIckler: no, they are only defined in .env.assets.{environment} which is only loaded during asset compilation.

app/controllers/payment_controller.rb
```diff
@@ -5,15 +5,21 @@ def available_refunds
...
+                     end
```
gregorbg, line 16:
```suggestion
registration_scope = competition.uses_microservice_registrations? ? MicroserviceRegistration : Registration
registration = registration_scope.includes(:payment_intents).find_by(competition: competition, user_id: user_id)
```

env_config.rb
gregorbg: Huh, then I'm afraid you have to explain to me why the polling URL is only a known env variable during asset compilation and not during runtime. How is the frontend supposed to call the Polling Lambda during runtime if the URL is not defined?!

FinnIckler: Because the frontend is already compiled during runtime!

app/controllers/registrations_controller.rb
FinnIckler: makes sense, sounds easy to do

app/controllers/payment_controller.rb
FinnIckler: nice!

gregorbg (approved):

FinnIckler:
let me test first if it everything is working in the frontend with v2 before you approve

gregorbg:
> let me test first if it everything is working in the frontend with v2 before you approve

Sorry, I was a bit eager because I saw you fixed all of my suggestions from the previous review. So I approved as in "Now I'm happy with the code changes compared to my previous review" 😇

FinnIckler:
V2 works.. Now testing V3

FinnIckler:
V3 works by my smoketests. We also got a new test choke?

FinnIckler:
@thewca-bot deploy staging

FinnIckler:
mh.. deployment fails

FinnIckler:
oh I know why

FinnIckler:
@thewca-bot deploy staging

dunkOnIT:
Ah, I see why the "Users aren't allowed to change events when cancelling" check was commented out - it exceeds the Rubocop rule for how many parameters can be passed to a function. 

I think there are legitimate questions about how necessary this check is, and I'm not seeing any immediate ways to refactor this which aren't very unwieldy (the lack of instance variables definitely hurts us here).

dunkOnIT:
@thewca-bot deploy staging

dunkOnIT:
@thewca-bot deploy staging

env_config.rb
```diff
@@ -30,9 +30,13 @@
     mandatory :VAULT_AWS_REGION, :string
     mandatory :TASK_ROLE, :string
     mandatory :WCA_REGISTRATIONS_URL, :string
-    mandatory :WCA_REGISTRATIONS_POLL_URL, :string
```
dunkOnIT, line 33: just confirming this should be removed / that it won't have any effects on v2 in production?

dunkOnIT: Handled in line 36 and at the end of the file

FinnIckler:
@thewca-bot deploy staging

gregorbg (approved):

FinnIckler:
@thewca-bot deploy staging

FinnIckler:
@thewca-bot deploy staging

FinnIckler:
@thewca-bot deploy staging

FinnIckler:
@thewca-bot deploy staging

PR #10131, opened 2024-10-25
Clean up EditPerson page

danieljames-dj:
This PR aims at cleaning up EditPerson page a bit. Earlier everything was in single file and now it's split into two files.

### DIFF
--- a/app/webpacker/components/Panel/pages/EditPersonPage/EditPersonForm.jsx
+++ b/app/webpacker/components/Panel/pages/EditPersonPage/EditPersonForm.jsx
@@ -1,16 +1,13 @@
 import React, { useState, useEffect } from 'react';
 import {
-  Button, Form, Icon, Item, Message,
+  Button, Form, Icon, Message,
 } from 'semantic-ui-react';
 import _ from 'lodash';
 import { adminCheckRecordsUrl, apiV0Urls } from '../../../../lib/requests/routes.js.erb';
 import useSaveAction from '../../../../lib/hooks/useSaveAction';
-import WcaSearch from '../../../SearchWidget/WcaSearch';
-import SEARCH_MODELS from '../../../SearchWidget/SearchModel';
 import Loading from '../../../Requests/Loading';
 import I18n from '../../../../lib/i18n';
 import { genders, countries } from '../../../../lib/wca-data.js.erb';
-import useQueryParams from '../../../../lib/hooks/useQueryParams';
 import useLoadedData from '../../../../lib/hooks/useLoadedData';
 import Errored from '../../../Requests/Errored';
 import UtcDatePicker from '../../../wca/UtcDatePicker';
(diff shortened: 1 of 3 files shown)


app/webpacker/components/Panel/pages/EditPersonPage/index.jsx
kr-matthews, line 1: I don't see `setLoading(false)` ever get called - how does this component get out of the loading state?
I guess it was already like this before the changes, so presumably it works.

app/webpacker/components/Panel/pages/EditPersonPage/EditPersonForm.jsx
```diff
@@ -100,20 +98,32 @@ function EditPersonForm({ wcaId, clearWcaId, setResponse }) {
...
+          success={response.success}
+          error={!response.success}
+          content={response.message}
+        >
+          <Message.Content>
+            {response.success && (
+            <>
+              Success!
+              <br />
+            </>
```
kr-matthews, line 112: Should be indented, same with the the fragment below.
```suggestion
              <>
                Success!
                <br />
              </>
```

danieljames-dj: Thanks Kevin, applied the change.

app/webpacker/components/Panel/pages/EditPersonPage/index.jsx
danieljames-dj: I'm using the custom hook [useQueryParams](https://github.com/thewca/worldcubeassociation.org/blob/main/app/webpacker/lib/hooks/useQueryParams.js) which I created few months ago. So, when the queryParams are changed, it will refresh the page, hence the loading will stop.

In more detail:
Initially the form is loaded without WCA ID, the user will search for the person and when they click submit, it will start loading, search for WCA ID of that person and once it receives, it will load again with the WCA ID in query parameter, then everything will be happening inside EditPersonForm.

kr-matthews: Ok, that makes this component (and any other which uses the hook) pretty hard to understand, since it's non-standard behaviour and is not visible. If possible, it would be best to refactor the hook at some point.
But that's not in the scope of this PR.

kr-matthews (approved):

app/webpacker/components/Panel/pages/EditPersonPage/index.jsx
danieljames-dj: Thanks, Kevin for approving the PR. I'll definitely keep this in my mind and this should get removed as the tickets project progress.

The backstory of this non-standard behavior was that, this was in rails and we migrated to react, but after migration WRT found one missing important feature which was to pre-select a WCA ID if WCA ID is in URL (because WRT receives the direct edit link in email). To implement this feature, this non-standard way was the only way that came to my mind at that point.

I don't have clear idea how this non-standard way will be removed completely in tickets project, but I can explain my path in tickets project. Currently this single big component is split into EditPersonForm and EditPersonPage. The tickets UI will be re-using EditPersonForm and then the edit person requests that comes through tickets will be handled in tickets UI. Then I'll try to figure out some way we can avoid using EditPersonPage completely - so that all requests will come through tickets (also have additional benefit - details like "who made that change", "when was the change made", etc will be tracked there). Once we figure out a way for all request to come through tickets, EditPersonPage can be removed, and hence the non-standard behavior will also be removed.

PR #10132, opened 2024-10-25
Introduce registration version column instead of uses_v2_registrations

FinnIckler:
To allow us to switch over to 'v3' in stages

### DIFF
--- a/app/models/competition.rb
+++ b/app/models/competition.rb
@@ -96,7 +96,9 @@ class Competition < ApplicationRecord
     restricted: 2,
   }, prefix: true
 
-  enum :registration_version, [:v1, :v2, :v3], prefix: true, default: :v2
+  NEW_REG_SYSTEM_DEFAULT = :v2
+
+  enum :registration_version, [:v1, :v2, :v3], prefix: true, default: NEW_REG_SYSTEM_DEFAULT
 
   CLONEABLE_ATTRIBUTES = %w(
     cityName
@@ -634,7 +636,6 @@ class Competition < ApplicationRecord
   def build_clone
     Competition.new(attributes.slice(*CLONEABLE_ATTRIBUTES)).tap do |clone|
       clone.being_cloned_from_id = id
-      clone.registration_version = :v2
 
       Competition.reflections.each_key do |association_name|
         case association_name
@@ -741,11 +742,11 @@ class Competition < ApplicationRecord
   end
 
   def enable_v2_registrations!
-    update_column :registration_version, :v2
+    update_column :registration_version, NEW_REG_SYSTEM_DEFAULT
   end
 
   def uses_new_registration_service?
-    self.registration_version_v2?
+    self.registration_version_v2? || self.registration_version_v3?
   end
 
   def should_render_register_v2?(user)
@@ -2409,7 +2410,7 @@ class Competition < ApplicationRecord
       "admin" => {
         "isConfirmed" => confirmed?,
         "isVisible" => showAtAll?,
-        "registrationVersion" => registration_version.to_s,
+        "usesNewRegistrationSystem" => uses_new_registration_service?,
       },
       "cloning" => {
         "fromId" => being_cloned_from_id,
@@ -2616,7 +2617,7 @@ class Competition < ApplicationRecord
       showAtAll: form_data.dig('admin', 'isVisible'),
       being_cloned_from_id: form_data.dig('cloning', 'fromId'),
       clone_tabs: form_data.dig('cloning', 'cloneTabs'),
-      registration_version: form_data.dig('admin', 'registrationVersion'),
+      registration_version: form_data.dig('admin', 'usesNewRegistrationSystem') ? NEW_REG_SYSTEM_DEFAULT : :v1,
     }
   end
 
@@ -2848,7 +2849,7 @@ class Competition < ApplicationRecord
           "properties" => {
             "isConfirmed" => { "type" => "boolean" },
             "isVisible" => { "type" => "boolean" },
-            "usesV2Registrations" => { "type" => "boolean" },
+            "usesNewRegistrationSystem" => { "type" => "boolean" },
           },
         },
         "cloning" => {
--- a/app/models/competition.rb
+++ b/app/models/competition.rb
@@ -96,7 +96,7 @@ class Competition < ApplicationRecord
     restricted: 2,
   }, prefix: true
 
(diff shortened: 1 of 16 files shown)


FinnIckler:
I'll do the form changes tomorrow afternoon

app/models/competition.rb
```diff
@@ -96,6 +96,8 @@ class Competition < ApplicationRecord
     restricted: 2,
   }, prefix: true
 
+  enum :registration_version, { v1: 'v1', v2: 'v2', v3: 'v3' }
```
gregorbg, line 99: In "simple" cases like this, you can use `enum :my_column, [:yes, :no, :maybe]` instead. The database level will be an integer corresponding to array indices, but if you access `some_model.my_column` Rails will convert and return it to you as `:yes`.

gregorbg: Independent remark: You can add the option `prefix: true` to this. Any enum in Rails generates magic methods that are based on the enum's entries: An entity with `enum :my_field, [:yes, :no]` will automatically get methods like `my_entity.yes?` and `my_entity.no?`. By adding the `prefix` option, the name of the enum is prefixed like `my_entity.my_field_yes?`.
In this case, I think `my_comp.registration_system_v2?` is much more concise (and much less misleading!) than `my_comp.v2?`

app/models/competition.rb
```diff
@@ -739,11 +741,11 @@ def trainee_delegate_ids
   end
 
   def enable_v2_registrations!
-    update_column :uses_v2_registrations, true
+    update_column :registration_version, 'v2'
   end
 
   def uses_new_registration_service?
-    self.uses_v2_registrations
+    self.registration_version == 'v2'
```
gregorbg, line 748: See my other comment above: If you use `prefix` on the enum, you can change this to `self.registration_version_v2?`

app/models/competition.rb
gregorbg: Yet another independent remark: You can pass `default: :v2` as an option to the enum call. I think it's self-explanatory what that does.

gregorbg (changes requested):
Am I crazy, or did we have some other places in the code where we explicitly pass `uses_v2_registration: true`?!
Please do a global string search for the name of the old boolean column to confirm. I reckon some tests and factories might also need to be adjusted.

app/models/competition.rb
FinnIckler: done, but should we not set it explicitly in the form instead of setting a default?

app/models/competition.rb
```diff
@@ -739,11 +741,11 @@ def trainee_delegate_ids
   end
 
   def enable_v2_registrations!
-    update_column :uses_v2_registrations, true
+    update_column :registration_version, 'v2'
```
gregorbg, line 744: Now with the integer-baser enum column this will probably(?) break. Unless Rails is smart enough to convert between string and symbol.

app/models/competition.rb
gregorbg: No, the competition has to be initialized with "some value" before it's being serialized and passed to the form.
On database level, zero (which means `v1`) should be the default, just in case. That's our hard backup. On Rails level, you can set V2 as default.
We are already effectively treating the current boolean field's `true` as a default, except that Rails didn't offer a standardized way of distinguishing between database-level and model-level defaults. Now with enums, it does. And I think we should leverage that.

FinnIckler:
I added the changes to the Form and CTRL-F for uses_v2_registrations only finds the migrations

FinnIckler:
I implemented it my way to give us an easier way to switch between systems, but I get that we can always just use the rails console if needed (as it might be confusing for users)

FinnIckler:
Again, I can't approve myself, but I'm fine with this implementation

gregorbg:
> I implemented it my way to give us an easier way to switch between systems, but I get that we can always just use the rails console if needed (as it might be confusing for users)

Making it easy to switch systems is my intention as well. But I think our interpretation of "easy" is not the same here :sweat_smile: WCAT doesn't need to know that there is a further distinction between V2 and V3. They only need one checkbox where "off = old" and "on = new". We can simply change the value of the `NEW_REG_SYSTEM_DEFAULT` constant in the backend, thereby controlling what this checkbox internally does.

gregorbg:
Furthermore, I have tried to drill down the distinction between V2 and V3 in the backend even further. This is important for places where we need to decide whether a route hits the microservice or not.

Other places, for example methods that decide whether to render React or the old frontend, don' need to care whether it's V2 or V3, they only care that it's "not V1". At the end of the day, I think this facilitates migration.

app/models/competition.rb
```diff
@@ -738,16 +741,16 @@ def trainee_delegate_ids
...
-  def enable_v2_registrations!
-    update_column :uses_v2_registrations, true
+  def uses_microservice_registrations?
+    self.registration_version_v2?
   end
 
-  def uses_new_registration_service?
-    self.uses_v2_registrations
+  def uses_new_registration_system?
+    self.uses_microservice_registrations? || self.registration_version_v3?
```
dunkOnIT, line 749: `registration_version_v3` is provided automatically by the enums?

gregorbg: Yes!

app/models/competition.rb
```diff
@@ -96,6 +96,10 @@ class Competition < ApplicationRecord
     restricted: 2,
   }, prefix: true
 
+  NEW_REG_SYSTEM_DEFAULT = :v2
```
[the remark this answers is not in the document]
gregorbg: Well because this is only for distinguishing whether the "new registration system" (as it appears in the competition form) maps to v2 or v3. The old legacy system will always stay as v1

FinnIckler (commented):
Method changes look good to me. 
Technically it's always possible to switch from v1 to v3 and back again, but I don't think we need to overcomplicate the competition form like that

gregorbg:
> Method changes look good to me.

Thanks!

> Technically it's always possible to switch from v1 to v3 and back again, but I don't think we need to overcomplicate the competition form like that

Both V2 and V3 are detected as "new system" in terms of the WCAT frontend, so it would be a simple matter of removing the checkbox to go from V3 to V1. The only thing that's "unintuitive" is that ticking the checkbox again would enable V2 as-is. But TBH when we are manually marking competitions as V3 we will probably hand-pick a few where we are sure that WCAT won't un-tick then re-tick them without telling us. And once we're rolling out V3 as default this will become a non-issue.

gregorbg (approved):
Approving on behalf of Finn and myself.

PR #10133, opened 2024-10-25
Only Load MS Registration if needed in init_stripe

FinnIckler:
Attempt to reduce more unnecessary traffic to the microservice


gregorbg (commented):
Shouldn't `MicroserviceRegistration.find_by || competition.microservice_registrations.includes.find_by` do just fine?

gregorbg:
I took the liberty of refactoring your idea into a general-purpose "lazy load" method that can be used in the mailer as well -- whoever is faster loads once, and then it's already done.

FinnIckler (commented):
cool! I can't review because this is technically my PR

gregorbg (approved):

PR #10135, opened 2024-10-28
Block unwanted file types from avatar upload

gregorbg:
One of the gazillion reasons to hate Apple! When switching to the new avatar backend, we introduced support for "web images" according to Rails -- which in turn basically just relies on MIME types and standards.

Apple's "High Efficiency Image Codec" somehow managed to get approved as "web image" in that standard, but those money-greedy bastards of course didn't release their codec properly. So while it's _technically_ adhering to the standard, in practice most browsers don't support it because they don't know how to decode the image stream client-side.

Now we have to fall back to hard-blocking HEIC files, sigh...

### DIFF
--- a/app/models/user_avatar.rb
+++ b/app/models/user_avatar.rb
@@ -32,8 +32,8 @@ class UserAvatar < ApplicationRecord
 
   MAX_UPLOAD_SIZE = 2.megabytes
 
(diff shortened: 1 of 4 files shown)


FinnIckler (approved):
I assume we keep the defaults from https://github.com/rails/rails/blob/e6429269fd5a8a7d728557f4e7d7f82c0ad1478c/activestorage/lib/active_storage/engine.rb#L49 and don't overwrite them. If so LGTM!

PR #10152, opened 2024-10-28
Revert to v1 registrations if not using WCA registration

dunkOnIT:
This will become unnecessary once we move to v3, but for now the default to v2 makes it impossible to import registrations for the comps not using WCA registrations

Have tested locally and can confirm:
- If "Use WCA registration" is ticked, it uses v2 registrations
- If "Use WCA registration" is NOT ticked, it uses v1 registratiosn


app/models/competition.rb
```diff
@@ -2548,6 +2548,8 @@ def set_form_data(form_data, current_user)
       end
 
       assign_attributes(Competition.form_data_to_attributes(form_data))
+
+      self.uses_v2_registrations = false unless form_data.dig('website', 'usesWcaRegistration')
```
gregorbg, line 2552: At this point, the attributes from the form data have already been assigned (in the line directly above), so you might as well do `unless self.use_wca_registration`.

gregorbg (commented):
Hmm... You definitely earn a "short and sweet" prize for coming up with this idea (and extra kudos for finding this place in the code to begin with, the WCAT form loading is a horrible mess!)

However I'm not convinced that it's the most clean place to implement this. In particular, it causes unintuitive side-effects for people in the UI. If you want to dabble with some React code, look at (for example) `FormSections/CompetitorLimit` to see how we enable/disable certain sections based on other parts of the form. It would definitely be better for the users to straight-out disable this in the UI already, to minimize confusion (otherwise, it would make it seem to someone who is explicitly turning on the checkbox in the UI that "the form isn't saving properly" because during the proper save, the checkbox value is being overridden without clear communication to the user)

if you don't have the time or motivation to look into this, let me know and I'll hijack the PR.

dunkOnIT:
> otherwise, it would make it seem to someone who is explicitly turning on the checkbox in the UI that "the form isn't saving properly"

This concern isn't quite clicking with me? As I understand it, the only people who we expose the "enable v2" checkbox to are WCAT - and we can just "manually" communicate with them about this change. 

In the case of a normal user creating a competition - they would just uncheck the "Use WCA registration" checkbox and without them being aware of it, the default uses_v2_registration value is overridden. I don't see this as a concern/see how it would lead to confusion?

app/models/competition.rb
dunkOnIT: Right you are!

gregorbg:
This needs to be updated now that #10132 is merged.

> This concern isn't quite clicking with me? As I understand it, the only people who we expose the "enable v2" checkbox to are WCAT - and we can just "manually" communicate with them about this change.
>
> In the case of a normal user creating a competition - they would just uncheck the "Use WCA registration" checkbox and without them being aware of it, the default uses_v2_registration value is overridden. I don't see this as a concern/see how it would lead to confusion?

I will accept this explanation only if you add a comment stating that this line is a V2 hack that can be removed after full V3 rollout :smile:

gregorbg (approved):

Issue #9463, opened 2024-06-02
Tickets project


danieljames-dj:
Tickets project is a project that is aimed to automate/simplify works done by WCA staffs. The following is the database architecture of the MVP:

![tickets](https://github.com/thewca/worldcubeassociation.org/assets/9512698/a21ce4ec-8ab1-4d7e-85b2-04b3d3ccf4c1)

This Github issue is to discuss regarding this design and the plan of implementation.

gregorbg:
Some initial thoughts:
- Stakeholders should have an `is_active` flag. It might be that someone initially takes charge of a ticket and then later a second person comes in and reviews the work. Or maybe some T/C initially received the ticket, but then noticed that another T/C is more suitable to handle it, so the stakeholder "responsibility" gets transfered.
- comments should have both a stakeholder ID and an actual user ID. Since only a user can make a comment, it is important to record that user and then display it as "comment by user ID 1234 on behalf of stakeholder gorup WCAT" or something
- status should probably have more entries. Things that come to mind immediately are `rejected` and `no_action`. But this can be easily implemented later down the road and is not an inherent design flaw.
    - On that note, I'm not sure how much we can even standardise the `status` field at all? If we want to use a ticketing system throughout the website, then various teams will have various workflows with wildly different statuses. Some tickets for some workflows might need a `pending_review` status, others might not.
    - In line with this thought, what purpose does the general `status` column serve at all? If different workflows for different tickets can have wildly different status requirements, why even try to unify them?
        -  Maybe the `status` flag should be moved to the metadata objects like `ticket_wst_contacts` or `tickets_edit_person_name`?
        - Maybe we should consider introducing different terminology for the "raw, technical status" of a ticket (is it open? is it closed?) and the "internal business logic status" (i.e. steps that depend on the actual type of the ticket, where WRT anonymizations have different "status" steps from WCAT competition announcements). Calling both of them "status" will be hugely confusing
- We **definitely** need a separate table keeping track of changes to tickets. Something along the lines of `ticket_history`. The MVP would be to record changes in ticket `status` field -- whenever the `status` changes, there has to be an entry to match this status change with information on who made this status change and why
    - This will end up being a very similar structure to `ticket_comments`. Essentially, a change log entry is just a "glorified comment". So maybe think about a clever way of reusing these tables. Maybe just have the change log entry point to a comment instead of storing its own comments?
[...]

PR #10156, opened 2024-10-29
Tickets for edit persons

danieljames-dj:
This is the first PR related to tickets and the implementation plan was explained [here](https://github.com/thewca/worldcubeassociation.org/issues/9463#issuecomment-2438433185).

Screenshots
(The UI is internal, so the `TicketHeader` and `TicketLogs` components are pretty basic and is not stressing on minor things like 'capitalizing open to Open', 'having human readable timestamp in logs', etc)

Edit profile form submission email:
Screenshot 2024-11-22 at 8 41 19 AM

Ticket page
Screenshot 2024-11-22 at 8 41 51 AM


danieljames-dj:
This is WIP, but still open for review for the files except TicketPlayground.

The main works pending here are:
1. Creation of ticket when somebody submits change request through form. (this is blocked on https://github.com/thewca/worldcubeassociation.org/pull/9811)
2. Changing of link in the contact email which we get (blocked on same above PR)
3. Implementation of TicketPlayground

I'll be working on 3 soon, for 1 & 2 I'll wait for the PR to get merged.

(Ignore the conflict, I'll resolve this later. I didn't resolve now because this won't get merged soon)

app/models/ticket_stakeholder.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  belongs_to :ticket
+
+  def stakeholder
+    if stakeholder_type == TicketStakeholder.stakeholder_types[:user_group]
+      UserGroup.find(stakeholder_id)
+    elsif stakeholder_type == TicketStakeholder.stakeholder_types[:user]
+      User.find(stakeholder_id)
+    end
+  end
```
gregorbg, line 22: Rails has `polymorphic` associations for exactly this purpose. You don't need to write this yourself, just use `belongs_to :stakeholder, polymorphic: true` above.

app/models/ticket.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+class Ticket < ApplicationRecord
+  enum :ticket_type, {
+    edit_person: "edit_person",
+  }
```
gregorbg, line 6: Does this column provide any actional value, on top of `metadata_type`?
Or in other words, can you think of use-cases where your suggested `ticket_type` is significantly different from `metadata_type` (defined by the polymorphic `metadata` association below)?

app/controllers/tickets_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+    respond_to do |format|
+      format.html do
+        @ticket_id = params.require(:id)
+        render :index
+      end
+      format.json do
+        ticket = Ticket.find(params.require(:id))
+        requester_stakeholders = ticket.user_stakeholders(current_user)
+
+        return head :unauthorized if requester_stakeholders.empty?
```
gregorbg, line 14: Why? If there are no (active) stakeholders, why does that imply that the user doesn't have authorization to view the ticket?

app/controllers/tickets_controller.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+class TicketsController < ApplicationController
+  def show
+    respond_to do |format|
+      format.html do
+        @ticket_id = params.require(:id)
+        render :index
```
gregorbg, line 8: In line with my other comment below, this `index` view should probably be renamed `show` to match the controller route.

app/controllers/tickets_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+    return head :unauthorized unless ticket.action_allowed?(:update_status, current_user)
+
+    ActiveRecord::Base.transaction do
+      ticket.metadata.update!(status: ticket_status)
+      TicketLog.create!(
+        ticket_id: ticket.id,
+        log: "Ticket status changed to #{ticket_status} by #{current_user.name}.",
+      )
+    end
+    status_changed = ticket.metadata.update(status: ticket_status)
```
gregorbg, line 36: You're executing `ticket.metadata.update` twice: Once with an `!` within the transaction, and then once immediately after the transaction.

But it's the same ticket and the same `ticket_status` variable being passed, so I wonder if this is a whoopsie?

gregorbg: Also, I want to point out that the return value of `update` does not indicate whether the status _changed_. The returned boolean only indicates whether the write operation to the DB was successful or not. It may be that nothing changed, in which case a kind of "no-op" SQL update will be performed, and the method will still return `true`.

The most notable use-case for this return value is validations: If you pass arguments to `update` that would make your model invalid (as defined by your own validations on the AR model), then the return value will be `false`.

app/controllers/tickets_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  def update_status
+    ticket = Ticket.find(params.require(:ticket_id))
+    ticket_status = params.require(:ticket_status)
+
+    return head :unauthorized unless ticket.action_allowed?(:update_status, current_user)
+
+    ActiveRecord::Base.transaction do
+      ticket.metadata.update!(status: ticket_status)
```
gregorbg, line 30: Here's the same(?) update as below, see my other comment for details

app/models/ticket.rb
```diff
@@ -0,0 +1,45 @@
...
+  has_many :ticket_stakeholders
+  belongs_to :metadata, polymorphic: true
+
+  def user_stakeholders(user)
+    return [] if user.nil?
+    ticket_stakeholders.select do |stakeholder|
+      if stakeholder.stakeholder_type == :user
+        stakeholder.stakeholder_id == user.id
+      elsif stakeholder.stakeholder_type == TicketStakeholder.stakeholder_types[:user_group]
+        user.active_roles.any? { |role| role.group_id == stakeholder.stakeholder_id }
```
gregorbg, line 18: I think `user.active_roles.where(group_id: stakeholder.stakeholder_id).any?` might be more efficient.

app/models/ticket.rb
```diff
@@ -0,0 +1,45 @@
...
+      (
+        stakeholder.stakeholder_type == TicketStakeholder.stakeholder_types[:user_group] &&
+        metadata.action_user_groups(action).include?(stakeholder.stakeholder_id)
+      )
+    end
+  end
+
+  def url
+    Rails.application.routes.url_helpers.ticket_url(id)
+  end
```
gregorbg, line 34: Hm, I'm not a big fan of "forcing" routes into AR models.

There's a reason that you have to call it in such a complicated way through `Rails.application.routes.url_helpers`, and that reason is: Helpers aren't supposed to be available in AR models.

Most places in our code that need this URL (like views, email templates, etc.) have these helpers available and mounted without needing the `Rails.application.routes.url_helpers` hack, so maybe just use `ticket_url` directly there, instead of "sideloading" it here.

app/models/ticket_stakeholder.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+class TicketStakeholder < ApplicationRecord
+  enum :stakeholder_type, {
+    user: "user",
+    user_group: "user_group",
+  }
```
gregorbg, line 7: The only small downside to my polymorphic suggestion below: Rails does not support enums on polymorphic `*_type` columns. That's not your fault, it's a definition by Rails that somebody just said "This isn't supported". It's a shame, I would love to have it, but in this case I think the benefits of polymorphic automations and other convenience features outweigh the loss of this one enum.

app/models/tickets_edit_person.rb
```diff
@@ -0,0 +1,81 @@
+# frozen_string_literal: true
+
+class TicketsEditPerson < ApplicationRecord
+  self.table_name = "tickets_edit_person"
```
gregorbg, line 4: Why do you have to override the table name here?

app/models/tickets_edit_person.rb
```diff
@@ -0,0 +1,81 @@
...
+        when :dob
+          ticket_metadata.update!(
+            previous_dob: change[:from],
+            new_dob: change[:to],
+          )
+        end
+      end
+
+      ticket = Ticket.create!(
+        name: "Edit Profile request by #{wca_id}",
```
gregorbg, line 53: Why are we storing humanized English names in the DB?
This should be inferred on-the-fly through I18n strings in the frontend at the moment the name needs to be displayed.

app/models/tickets_edit_person.rb
```diff
@@ -0,0 +1,81 @@
...
+        ticket_type: Ticket.ticket_types[:edit_person],
+        metadata: ticket_metadata,
+      )
+
+      TicketStakeholder.create!(
+        ticket_id: ticket.id,
+        stakeholder_id: UserGroup.teams_committees_group_wrt.id,
+        stakeholder_type: TicketStakeholder.stakeholder_types[:user_group],
```
gregorbg, line 61: If you "polymorphize" this association, as discussed above, you can just plug in `stakeholder: UserGroup.teams_committees_group_wrt` and call it a day.

app/models/tickets_edit_person.rb
```diff
@@ -0,0 +1,81 @@
...
+        connection: TicketStakeholder.connections[:assigned],
+        is_active: true,
+      )
+      TicketStakeholder.create!(
+        ticket_id: ticket.id,
+        stakeholder_id: requester.id,
+        stakeholder_type: TicketStakeholder.stakeholder_types[:user],
```
gregorbg, line 68: Again, "polymorphization" would allow you to pass `stakeholder: requester`.

app/models/tickets_edit_person.rb
```diff
@@ -0,0 +1,81 @@
...
+        connection: TicketStakeholder.connections[:cc],
+        is_active: true,
+      )
+
+      TicketLog.create!(
+        ticket_id: ticket.id,
+        log: "Ticket created.",
```
gregorbg, line 75: Again, not super excited about localized strings in the DB. Since this is "only" logs, I'd be leaning to accept it, but can you maybe spend some 10 or so minutes to think about an `enum` set of log changes?

In my mind, typical log actions include "created", "moved", "reassigned" and "closed", all of which can easily be enumized.

db/migrate/20241029105239_create_tickets_edit_person.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class CreateTicketsEditPerson < ActiveRecord::Migration[7.2]
```
gregorbg, line 3: I'm not too wild about the repeated nature of `previous_*` and `new_*` fields.
Every time you want to extend/augment the capability of this ticket type in the future, you'd have to write a migration with two new columns.

Instead, what would you think about an auxiliary table along the lines of "EditPersonChanges", with a foreign key pointing to this table, and three columns `field_name`, `old_value`, `new_value`?

`field_name` can be enumized, to allow some degree of control over which fields are "allowed" to be edited, but at the same time also make future extensions very easy (just add an entry to the enum, which is much less invasive than a two column migration)

And if you want, you might even store the `old_value` and `new_value` as JSON to allow various field types. Just an idea :)

db/schema.rb
gregorbg, line 1: Sidenote regarding this file and the corresponding migration files:
I made a lot of comments above in the AR model files about table structure, modelling choices, etc. but I did not take the time to make matching comments in the raw SQL schema here. I want to reach consensus on the other discussions first, without constantly needing to update two threads at once. (And then I trust that you can always just make the according schema and migration changes yourself :innocent: )

gregorbg (changes requested):
Cool, I like the overall structure! Lots of detail questions and some model suggestions, but all in all on a good path!

app/controllers/tickets_controller.rb
danieljames-dj: Oh sorry this was a whoopsie. I've removed one. Ya I knew the returned boolean just represents whether it's success or not, I shouldn't have named it `status_changed`. Anyway I have hard-coded `render json: { success: true }` now.

app/controllers/tickets_controller.rb
danieljames-dj: Removed the other one, thank you.

app/models/ticket.rb
danieljames-dj: Oh yes, thanks.

db/schema.rb
danieljames-dj: Thanks, I've updated this file too now.

app/controllers/tickets_controller.rb
gregorbg: Okay, I understand the context now. Based on your explanation, shouldn't it be something like `return head :unauthorized unless requester_stakeholders.include?(current_user)`?

app/models/ticket.rb
gregorbg: Hm, I see that enums are useful for switch-case, but this doesn't change the fact that the contained information is basically redundant.

In the places where you're currently doing the switch-case, can you instead maybe try to do something like `if ticket.metadata.is_a? EditPersonTicket` or similar?

app/models/tickets_edit_person.rb
gregorbg: Hm, I think you can be more loose on the names. It doesn't necessarily have to contain `Ticket` as prefix.
Something like `EditPersonRequest` will be fine. Or if you really want the "ticket" link, you could do `EditPersonTicket`, which Rails correctly maps to `edit_person_tickets`.

But speaking from our experience with the Groups tables so far, I don't think you have to forcefully include `Ticket` everywhere.

gregorbg: Actually, a request to change some of your personal details _is not_ a ticket. The request is _handled via_ your ticket system, but that doesn't make the request itself become a ticket. (Just as an argument in favor of dropping `Ticket`)

app/models/tickets_edit_person.rb
gregorbg: Hm... I can definintely make an exception about localized strings in the DB when it comes to `ticket_comments` later. There, users can write in whatever English they want and we store it.

In this case, for `ticket_logs`, I want to explore other options first. What kind of log entries do you expect in the future? I think having `ticket_logs` as a "clean" base table which only tracks standardized changes is desirable, because it helps getting an overview.

gregorbg: If you really want to include lots of custom statuses, then I would recommend splitting it into two columns:
One enumized column that tracks basic, common, unified lifecycle changes (stuff like "created", "reassigned", "closed", etc. that all tickets everywhere share) and you simply add _one_ enum entry there that says `custom_change`.

Then, you introduce a second column called "details" or something in the `ticket_logs` table which will be null by default and which won't be needed for plain "opened" or "closed" logs, but where users can optionally enter more details if they want.

gregorbg: But then again, we have to make sure that they don't "abuse" the custom log state as a separate, second "comment" field. I.e. it would need to be totally clear what goes into `ticket_logs` and what goes into `ticket_comments`. Is this even clearly defined at this point in time?

gregorbg: (sorry for the spam) Actually, we could improve the flow even more: Instead of a column that contains "additional details" in the logs, you could add a nullable column that _optionally_ holds a foreign key that points to a `ticket_comment` if some user wants to have a "custom explanation" for a status change. That way, we can nicely link `ticket_logs` and `ticket_comments`.

And I can still get my standardized (enumized) lifecycle changes in `ticket_logs` :innocent:

app/webpacker/components/Tickets/TicketPlaygrounds/EditPersonTicketPlayground.jsx
```diff
@@ -0,0 +1,39 @@
...
+
+  return (
+    <EditPersonForm
+      wcaId={ticket.metadata.wca_id}
+      onSuccess={closeTicket}
+    />
+  );
+}
+
+export default function EditPersonTicketPlayground({ ticketDetails, sync }) {
```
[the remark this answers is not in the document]
gregorbg: Not sure if that's ideal. In software, "Playground" often means "try out stuff without real consequences".
What about "workbench", "backoffice", or "panel"?

db/migrate/20241124050607_create_tickets_edit_person_fields.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class CreateTicketsEditPersonFields < ActiveRecord::Migration[7.2]
+  def change
+    create_table :tickets_edit_person_fields do |t|
+      t.bigint :tickets_edit_person_id, null: false
+      t.string :field_name, null: false
+      t.string :old_value, null: false
+      t.string :new_value, null: false
```
gregorbg, line 9: Shouldn't this be JSON?

db/migrate/20241124050607_create_tickets_edit_person_fields.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class CreateTicketsEditPersonFields < ActiveRecord::Migration[7.2]
+  def change
+    create_table :tickets_edit_person_fields do |t|
+      t.bigint :tickets_edit_person_id, null: false
```
gregorbg, line 6: I think this can be `t.references :tickets_edit_person`

db/migrate/20241029105222_create_ticket_logs.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class CreateTicketLogs < ActiveRecord::Migration[7.2]
+  def change
+    create_table :ticket_logs do |t|
+      t.bigint :ticket_id, null: false
```
gregorbg, line 6: `t.references :ticket` should work.

db/migrate/20241029105150_create_ticket_stakeholders.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class CreateTicketStakeholders < ActiveRecord::Migration[7.2]
+  def change
+    create_table :ticket_stakeholders do |t|
+      t.bigint :ticket_id, null: false
```
gregorbg, line 6: `t.references :ticket`, see other comments.

db/migrate/20241029105150_create_ticket_stakeholders.rb
```diff
@@ -0,0 +1,14 @@
...
+      t.integer :stakeholder_id, limit: 8, null: false
+      t.string :stakeholder_type, null: false
```
gregorbg, line 8: These two can become `t.references :stakeholder, polymorphic: true`

db/migrate/20241029104311_create_tickets.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class CreateTickets < ActiveRecord::Migration[7.2]
+  def change
+    create_table :tickets do |t|
+      t.string :ticket_type, null: false
+      t.bigint :metadata_id, null: false
+      t.string :metadata_type, null: false
```
gregorbg, line 8: `t.references :metadata, polymorphic: true`

app/models/tickets_edit_person.rb
maxidragon: I agree with Gregor, we should avoid localized strings in DB whenever possible

db/migrate/20241029105222_create_ticket_logs.rb
danieljames-dj: Done

db/migrate/20241124050607_create_tickets_edit_person_fields.rb
danieljames-dj: Done

db/migrate/20241124050607_create_tickets_edit_person_fields.rb
danieljames-dj: Actually string is also working fine because the request from frontend is anyway coming in string format. Currently there is no use-case to convert it to date or any other type. But it may come in future for this same ticket type. Even then, having this as string will be sufficient.

app/models/ticket.rb
gregorbg: Yeah, that's basically an enum without enforcing it in Rails.
What about moving it into a constant like
```TICKET_TYPES = {
  edit_person: "EditPersonTicket",
  anonymization: "AnonymizationTicket",
}.freeze
```

app/controllers/tickets_controller.rb
gregorbg: Okay, makes more sense now. What about including this "WRT example" from your most recent comment as a code comment as well?

app/models/tickets_edit_person.rb
gregorbg: I don't have a clear/strong preference about how to proceed here. Having to override `table_name` always feels a bit weird, but on the other hand your arguments in favor of the override are also very reasonable.

Fine to keep the `self.table_name` override as is, or go with another name as you see fit.

app/models/tickets_edit_person.rb
gregorbg: It seems like you want to log changes to tickets quite thoroughly. That's good and I support your basic intention, and I can tell you that simply logging (primary key id, ticket_id, change_string) won't be enough then. Believe me, it would become a nightmare to properly track and parse and organize these changes later on.

Please take a step back to the drawing board and think about the structure and information of `ticket_logs` on a DB design level again. What information do you want to track, and what is the most structured way to track this information?

From your existing description, I think
- `action_type`
- `acting_user` (foreign key ID)
- `acting_stakeholder` (foreign key ID)
    - There should probably be both -- one reference to an actual user that can be tracked, and one reference to a stakeholder on behalf of which the user was acting
    - This is the same line of thought as `ticket_comments`. Both have two fields (user_id, stakeholder_id). Makes four references in total and is a bit convoluted to track. Can you make a suggestion on how to "unify" these two independent fields? (i.e., store them only in one place, and reference it in the other place)

are all necessary. About the details: You should probably track details as well, for example by using a `changed_field` and `changed_value` logic. These can be custom, but still "somewhat" easy to parse.

db/migrate/20241124050607_create_tickets_edit_person_fields.rb
gregorbg: Then at least change it to `t.text`, because the length of a `string` field in the DB is quite limited.

app/controllers/tickets_controller.rb
danieljames-dj: Okay, I've added comments 2-3 lines above - before initializing `requester_stakeholders`. Hope it's fine.

app/models/ticket.rb
danieljames-dj: That's better. Changed, thank you.

app/models/tickets_edit_person.rb
danieljames-dj: > Re how to avoid two references being referenced in two tables
How about having one more table named `ticket_users` with columns `user_id` and `stakeholder_id`? And then we can reference that table to both `comments` and `logs`? Basically we can have a method like following in `ticket_user.rb`:

```
def self.get(user_id, stakeholder_id)
  ticket_user = TicketUser.find_by(user_id: user_id, stakeholder_id: stakeholder_id)
  ticket_user ? ticket_user : TicketUser.new(user_id: user_id, stakeholder_id: stakeholder_id)
end
```

Will this be good to go?

I personally find this bit complicated though, but can't think of a better case.

Also, just re-thinking this again - do we really need `stakeholder_id` in the comments and logs? Is it fine if we just have `user_id`? In the UI, while mentioning the user name, we can add the user's relevant roles as well (for both logs and comments) so that whoever wants to know the stakeholder role connected with the user can get to know from there itself.

What do you think?

gregorbg: Hm, this boils down to "organization-level" decisions pretty quickly. The question that we have to consider completely irrespective of the code or programming is "Do we want to keep track of the concrete stakeholder on behalf of which the user acted?"

I can think of scenarios where a WRT member files a request to have their own personal details edited (you should know that can happen :wink:), in which case this person would become a stakeholder as a simple competitor, but also be a stakeholder as a WRT member potentially (as far as the code is concerned) eligible to handle the ticket. Or a WCAT member who also happens to be a Delegate might be responsible for a competition and then submit it to WCAT, at which point they would become responsible for their own ticket.

How do we want to deal with this kind of situation on an organizational level?
1. We could just flat-out prevent it. Like build some exceptions into the stakeholder query that groups which the user is a member of should not be considered if the user is also the one who filed the ticket.
2. (my preferred option) always mark actions on tickets with the stakeholder attached. Like "this user acted on behalf of this stakeholder".

Regarding (2), it is totally up for debate and I am really open to change my mind and be convinced whether we want to keep track of this information or not. But _if_ we decide that we want it, then there's no way that we can prevent the double `(user_id, stakeholder_id)` bookkeeping.

danieljames-dj: Yes I agree there will be cases where a user will be part of multiple stakeholder_types. But my point was that intuitively we can know on whose behalf the action was done through the `action_type`. But when I think again, it may not be practical when tickets grow so much. Hence, having both `user_id` and `stakeholder_id` in `ticket_logs` makes sense to me.

I'll go with option (2) as you suggested.

I understood your comment regarding `comments` and theoretically I think it's a good idea. I'm not 100% convinced practically because I somehow feel making `ticket_logs` as a more important thing than `ticket_comments` may not be so good, but I might be wrong. I'll think of it again once I start working on comments in a later PR.

I'll go ahead with implementation of `ticket_logs` table with columns `action_type`, `acting_user_id`, `acting_stakeholder_id`. Hope this is fine, thank you.

danieljames-dj:
@thewca-bot deploy staging

danieljames-dj:
@thewca-bot deploy staging

PR #10157, opened 2024-10-30
Disable frontend competitor limit check if no competitor limit is enforced

dunkOnIT:
This mirrors the change pushed to the registration microservice - tested locally and works as expected in cases where a competitor limit is and isn't set.


kr-matthews:
I didn't look closely, but wouldn't it make more sense if `spotsRemaining` were re-defined in such a way that it's `Infinity` if there is no competitor limit?

dunkOnIT:
Great point, have just made that change!

gregorbg (approved):

PR #10158, opened 2024-10-30
Unify AWS_REGION variables

FinnIckler:
Setting to draft until [#10058](https://github.com/thewca/worldcubeassociation.org/pull/10058) and #10123 are merged, just because that makes it easier for me to run terraform.

We can safely remove all the references as per https://docs.aws.amazon.com/sdk-for-ruby/v3/developer-guide/setup-config.html#aws-ruby-sdk-setting-region -> Setting the Region using environment variables


gregorbg:
What's the status on this?

FinnIckler:
Let's merge this in a call just because we have to run terraform and merge the code at the same time before we can deploy again.

gregorbg (approved):

FinnIckler:
mh.. it doesn't seem to work, I will have to closely monitor a deployment tomorrow

FinnIckler:
@thewca-bot deploy staging

FinnIckler:
@thewca-bot deploy staging

FinnIckler:
I successfully deployed it and ran terraform in a non destructive way (meaning I only added the AWS_REGION variable for now and not removed the others). Staging does not do as many AWS Interactions as prod, but I did successfully connect to S3 for regulations, so I would call it a win.
Let's merge and deploy this on monday so we can intervene if I missed something.

Issue #8884, opened 2024-02-07
Automatically make competitions visible when announcing competitions


slongbehn:
![image](https://github.com/thewca/worldcubeassociation.org/assets/136729888/f201e2a2-ed55-4295-aee9-6c2f9800259b)

**Is your feature request related to a problem? Please describe.**
When WCAT clicks "announce" on a competition, we then have to make it visible and save the competition. Occasionally, a member either forgets to do this or doesn't save correctly. While most organizers notice this, one did not for several weeks. As far as I am aware, there has never been a time where a competition was announced and was not immediately made visible to the public.

**Describe the solution you'd like**
It would be awesome for WCAT quality of life to automatically make a competition visible when announcing a competition.

**Describe alternatives you've considered**
The alternative is to continue to manually do it. It's fine, would just be nice to not have to :)

Thanks for your time!

dunkOnIT:
Very reasonable, should be a pretty straightforward fix - happy to accept PR's against this, and provide guidance if necessary

RamanTheDeveloper:
Hi, can I get assigned to this task?

lnzainn:
@slongbehn should the 'Make this competition visible to the public' still be there? I assume it may be needed for emergency purposes.

slongbehn:
@lnzainn Yes, the checkbox ideally still there just in case. Thanks!

PR #10166, opened 2024-11-01
Making comps visible after announcing

pranav-027:
Making competitions visible after announcing. Saves a step for WCAT
solves issue : https://github.com/thewca/worldcubeassociation.org/issues/8884

Demo: https://youtu.be/97qNRyvvTiA

### DIFF
--- a/app/controllers/competitions_controller.rb
+++ b/app/controllers/competitions_controller.rb
@@ -769,7 +769,7 @@ class CompetitionsController < ApplicationController
       return render json: { error: "Already announced" }
     end
 
-    competition.update!(announced_at: Time.now, announced_by: current_user.id)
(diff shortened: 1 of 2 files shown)


gregorbg (changes requested):
Thank you for your contribution! So far, you have successfully taught the frontend to act like the competition has the `visible` boolean set to `true`.

But you're actually missing the coresponding change in the backend. Please adjust the Rails controller so that it actually writes the `showAtAll` property to the database when the `/competitions/:competition_id/announce` route is being called in `def announce`.

(For context: If you repeat exactly the steps from your PoC video, and then reload the page using "Refresh" in your browser, you will find that the value in the checkbox is suddenly gone.)

pranav-027:
@gregorbg Done!

gregorbg (approved):

Issue #10074, opened 2024-10-15
Add 


GenTheThief:
When looking at a competition page, it's always frustrating to have to go to the competitor page to see how many people have registered. Particularly for major championships when there are typically 500-1000 people registered, it can sometimes take minutes for the page to load so I can scroll to the bottom of the page to see how close the competition is to filling up. For smaller comps sometimes, it can be an unnecessary step to see how full it is.

It always seems so tantalizing that the competition sidebar has the competitor limit listed as well as the number of bookmarks. A solution to my problem would be to list how many people are currently registered right there on the main page, something like this:

<img width="351" alt="Screenshot number of competitors registered" src="https://github.com/user-attachments/assets/d5ddfe73-455c-4bcc-8a37-0e6a20f433ef">

It also seems unnecessary to still include the number of bookmarks a competition has after registration has opened. Before registration opens, it can be useful to see if the demand is more than the limit, but after it opens, the more pertinent number is how much space is left.

An alternative way to display the information would be like this, with the number of competitors listed as a fraction of the competitor limit.
<img width="328" alt="Screenshot number of competitors registered 2" src="https://github.com/user-attachments/assets/79dfa8ec-f77c-483a-bbd2-3139b856c2f4">

One problem I can think of is that the main page would need to constantly update to reflect the changing number of competitors. I don't know how much extra stress this would cause, but I don't imagine it would be a terrible load on the system.
The main page shouldn't need to regularly update on its own either -- each time a delegate or organizer accepts a competitor, as the competitor list updates, the main page can receive a +1 update as well.

FinnIckler:
The newly designed competition page will include this, so not sure if we will invest time to add this to the old page. 

If a volunteer wants to pick this up: Don't worry about the database impact, we already load the number of people registered on the page to display the warning if the competition is full.

PR #10167, opened 2024-11-01
Registration count added on competition info page

pranav-027:
Solves issue: https://github.com/thewca/worldcubeassociation.org/issues/10074

image

image

### DIFF
--- a/app/views/competitions/_competition_info.html.erb
+++ b/app/views/competitions/_competition_info.html.erb
@@ -134,6 +134,13 @@
       <% end %>
 
       <dl class="dl-horizontal">
+        <dt><%= t '.registered_competitors' %></dt>
+        <dd>
+          <%= competition.registrations.accepted.count %>
+        </dd>
+      </dl>
+
+      <dl class="dl-horizontal">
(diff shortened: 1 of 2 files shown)


gregorbg (changes requested):
Thank you for your contribution! Unfortunately, we're in the middle of migrating between registration systems, so this feature will not work as expected for competitions that are using our temporary "external registration service".

I'm sorry to say that it's way too complicated to explain what "external registration service" is under a simple GitHub PR, but the good side is that this PR will continue to work once we've fully completed the migration in ~1 week from now.

If you want to continue submitting PRs, I recommend that you first comment on the corresponding issue to get some feedback or "Okay, go ahead" from us. That way. we can minimize the risk of you accidentally working on some feature that is obsolete :upside_down_face:

pranav-027:
Sure @gregorbg, will take care of that in the future! I will leave this PR open then

pranav-027:
Hello @gregorbg,

I didn’t give that much thought. The same information is already available in the Competitors tab; this will only save a click and a scroll. I am just showing that information on the General page. As for the part you mentioned about being misleading, it still applies if someone checks registration count in the Competitors tab.

PR #10168, opened 2024-11-01
Only run SubmitResultsNagJob in Prod

FinnIckler:
Just had to wait 10 minutes until sidekiq send 50000 nags so I think it's fair to say we don't need to run this unless we are on prod

### DIFF
--- a/app/jobs/submit_report_nag_job.rb
+++ b/app/jobs/submit_report_nag_job.rb
@@ -1,6 +1,11 @@
 # frozen_string_literal: true
 
 class SubmitReportNagJob < WcaCronjob
+  before_enqueue do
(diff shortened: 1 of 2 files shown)


gregorbg:
by "send" in this case you mean to Mailcatcher, I suppose?

FinnIckler:
Yes. Rubymine was basically unusable during this period (that's also kind of an optimization issue for them), but it's still annoying because this happens just because you had sidekiq running locally at a certain time

gregorbg (approved):

gregorbg:
On second thought, you may want to include a similar check on SubmitReportNagJob as well

PR #10169, opened 2024-11-01
V2/Edit: Allow Organizers to Edit Registrations after deadline has passed

FinnIckler:
I don't know why I implemented it in that way, because organizers should be able to overwrite it (and it works in the backend).


FinnIckler:
Should I add a warning to the update prompt?

gregorbg:
> Should I add a warning to the update prompt?

Yes please!

FinnIckler:
I added a message, not 100% sure about the wording, maybe this is something another team should decide?

gregorbg:
Hm, let's not over-bureaucratize things. Launch it with this wording and if anyone has a problem with it, we can still change it in a follow-up PR.

gregorbg (approved):

gregorbg (approved):

PR #10170, opened 2024-11-01
Update About Page WCA Member Count

StanFromIreland:
Updated the WCA member count from 200,000 members to 245,000 as currently there are 245,631 WCA members and the statistic has not been updated for around a year from what i can tell.

image


StanFromIreland:
Hello @gregorbg, unfortunately I am not familiar with ruby however I will research the topic and see if I am able to do anything. [For now I have updated the statistic in all the translations. ](https://github.com/thewca/worldcubeassociation.org/pull/10200)

PR #10173, opened 2024-11-02
Set DOB to 1954-12-04 during anonymization

maxidragon:
Setting DOB to `nil` for accounts produces an error when trying to fetch WCIF.

For more information, refer to November sanity check email thread.


dunkOnIT (approved):
Change approved by WRT in "Sanity Check - NOVEMBER 2024", and a database update has already been run

PR #10192, opened 2024-11-04
Set BUNDLE_PATH in Docker Compose dev file

gregorbg:
See title (and diff).


FinnIckler (approved):
LGTM, I've been doing this on my local for a while

PR #10198, opened 2024-11-04
Add competing_status column to registration

FinnIckler:
This just adds the column and does nothing with it as this would generate a lot of conflicts with #10127 . So either we merge this first and then change it in  #10127 or we merge  #10127 and then I do the changes here

### DIFF
--- a/app/models/registration.rb
+++ b/app/models/registration.rb
@@ -1,10 +1,7 @@
 # frozen_string_literal: true
 
 class Registration < ApplicationRecord
-  scope :pending, -> {
-    where(accepted_at: nil, deleted_at: nil, is_competing: true)
-      .or(where(registration_version: :v3, competing_status: 'pending'))
-  }
+  scope :pending, -> { where(accepted_at: nil, deleted_at: nil, is_competing: true) }
   scope :accepted, -> { where.not(accepted_at: nil).where(deleted_at: nil) }
   scope :deleted, -> { where.not(deleted_at: nil) }
   scope :cancelled, -> { where(competing_status: 'cancelled') }
--- a/app/models/registration.rb
+++ b/app/models/registration.rb
@@ -230,9 +230,9 @@ class Registration < ApplicationRecord
   end
 
   def add_history_entry(changes, actor_type, actor_id, action, timestamp = Time.now.utc)
-    new_entry = registration_history_entries.create(actor_type: actor_type, actor_id: actor_id, action: action)
+    new_entry = registration_history_entries.create(actor_type: actor_type, actor_id: actor_id, action: action, created_at: timestamp)
     changes.each_key do |key|
-      new_entry.registration_history_change.create(value: changes[key], key: key, created_at: timestamp)
+      new_entry.registration_history_change.create(value: changes[key], key: key)
     end
   end
 
--- a/app/models/registration.rb
+++ b/app/models/registration.rb
@@ -1,7 +1,10 @@
 # frozen_string_literal: true
 
 class Registration < ApplicationRecord
-  scope :pending, -> { where(accepted_at: nil).where(deleted_at: nil).where(is_competing: true).or(competing_status: 'pending') }
+  scope :pending, -> {
+    where(accepted_at: nil, deleted_at: nil, is_competing: true)
+      .or(where(competing_status: 'pending'))
+  }
   scope :accepted, -> { where.not(accepted_at: nil).where(deleted_at: nil) }
   scope :deleted, -> { where.not(deleted_at: nil) }
   scope :cancelled, -> { where(competing_status: 'cancelled') }
--- a/app/models/registration.rb
+++ b/app/models/registration.rb
@@ -3,7 +3,7 @@
 class Registration < ApplicationRecord
   scope :pending, -> {
     where(accepted_at: nil, deleted_at: nil, is_competing: true)
-      .or(where(competing_status: 'pending'))
+      .or(where(registration_version: :v3, competing_status: 'pending'))
   }
   scope :accepted, -> { where.not(accepted_at: nil).where(deleted_at: nil) }
   scope :deleted, -> { where.not(deleted_at: nil) }
--- a/app/models/registration.rb
+++ b/app/models/registration.rb
@@ -1,11 +1,12 @@
 # frozen_string_literal: true
 
 class Registration < ApplicationRecord
-  scope :pending, -> { where(accepted_at: nil).where(deleted_at: nil).where(rejected_at: nil).where(waitlisted_at: nil).where(is_competing: true) }
+  scope :pending, -> { where(accepted_at: nil).where(deleted_at: nil).where(is_competing: true).or(competing_status: 'pending') }
   scope :accepted, -> { where.not(accepted_at: nil).where(deleted_at: nil) }
   scope :deleted, -> { where.not(deleted_at: nil) }
-  scope :rejected, -> { where.not(rejected_at: nil) }
-  scope :waitlisted, -> { where.not(waitlisted_at: nil) }
+  scope :cancelled, -> { where(competing_status: 'cancelled') }
+  scope :rejected, -> { where(competing_status: 'rejected') }
+  scope :waitlisted, -> { where(competing_status: 'waiting_list') }
   scope :non_competing, -> { where(is_competing: false) }
   scope :not_deleted, -> { where(deleted_at: nil) }
   scope :with_payments, -> { joins(:registration_payments).distinct }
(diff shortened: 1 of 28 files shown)


db/migrate/20241104161404_add_competing_status_enum.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class AddCompetingStatusEnum < ActiveRecord::Migration[7.2]
+  def up
```
gregorbg, line 4: You can use `def change` here. `add_column` is reversible

db/migrate/20241104161404_add_competing_status_enum.rb
```diff
@@ -0,0 +1,12 @@
...
+    add_column :registrations, :competing_status, :string, default: Registrations::Helper::STATUS_PENDING, null: false
+
+    # Update values based on the old boolean column
+    Registration.all.each do |r|
```
gregorbg, line 8: For bulk updates like this, that cannot be done in pure SQL, you can use `Registration.find_each`. It's more memory-efficient by batching the model load

lib/tasks/registration_version.rake
```diff
@@ -0,0 +1,84 @@
...
+
+          # Point any payments to the new holder
+          if competition.using_payment_integrations?
+            payment_intents = registration.payment_intents
+            if payment_intents.present?
+              payment_intents.update_all(holder_id: new_registration.id)
+            end
+          end
+
+          new_registration.save!
```
gregorbg, line 72: I don't think you need this `save!` here. The `create` call above already persists to the database, and in the following `if` statement (that re-attaches payments) you're not changing or touching any attributes on the registration itself (only on the field(s) of `PaymentIntent` that _point to_ the registration, but the registration model itself couldn't care less).

Thus, I think the `save!` as is now will always become a NOOP. Shout if I'm missing something.

lib/tasks/registration_version.rake
```diff
@@ -0,0 +1,84 @@
...
+        competition.microservice_registrations.each do |registration|
+          new_registration = Registration.create(competition_id: competition_id,
+                                                 user_id: registration.user_id,
+                                                 comments: registration.comments,
+                                                 guests: registration.guests,
+                                                 competing_status: registration.competing_status,
+                                                 administrative_notes: registration.administrative_notes,
+                                                 registration_competition_events: registration.event_ids.map do |event_id|
+                                                   RegistrationCompetitionEvent.create(competition_event: competition.competition_events.find { |ce| ce.event_id == event_id })
+                                                 end)
```
gregorbg, line 62: Could you move this block into its own separate line please?

lib/tasks/registration_version.rake
```diff
@@ -0,0 +1,84 @@
...
+  end
+
+  task :migrate_v2_v3, [:competition_id] => [:environment] do |_, args|
+    competition_id = args[:competition_id]
+
+    if competition_id.blank?
+      abort "Competition id is required"
+    end
+
+    competition = Competition.find(competition_id)
```
gregorbg, line 41:
```suggestion
    competition = Competition.includes(:competition_events).find(competition_id)
```

lib/tasks/registration_version.rake
```diff
@@ -0,0 +1,84 @@
...
+      abort "Competition #{competition_id} not found"
+    end
+
+    unless competition.registration_version_v2?
+      abort "Competition #{competition_id} is not on version 2"
+    end
+
+    LogTask.log_task("Migrating Registrations for Competition #{competition_id}") do
+      ActiveRecord::Base.transaction do
+        competition.microservice_registrations.each do |registration|
```
gregorbg, line 53:
```suggestion
        competition.microservice_registrations.includes(:payment_intents).each do |registration|
```

lib/tasks/registration_version.rake
```diff
@@ -0,0 +1,84 @@
...
+                                                   RegistrationCompetitionEvent.create(competition_event: competition.competition_events.find { |ce| ce.event_id == event_id })
+                                                 end)
+
+          # Point any payments to the new holder
+          if competition.using_payment_integrations?
+            payment_intents = registration.payment_intents
+            if payment_intents.present?
+              payment_intents.update_all(holder_id: new_registration.id)
+            end
+          end
```
gregorbg, line 70: Technically, you could do all of this in one short and concise `registration.payment_intents&.update_all(...) if competition.using_payment_integrations?`. Although I'm debating with myself whether that's too dense or not.

In any case, the current state is a tad bit too verbose. Maybe see whether you can reach a compromise between these two variants.

lib/tasks/registration_version.rake
```diff
@@ -0,0 +1,84 @@
...
+
+          new_registration.save!
+
+          ms_history = Microservices::Registrations.history_by_id(registration.attendee_id)
+          ms_history['entries'].each do |entry|
+            new_registration.add_history_entry(entry['changed_attributes'], entry['actor_type'], entry['actor_id'], entry['action'], entry['timestamp'])
+          end
+        end
+
+        competition.update_column :registration_version, :v3
```
gregorbg, line 80: Not strictly necessary but thought I'd still drop it here so you can learn about Rails enum magic:
`competition.registration_version_v3!` fires an SQL update directly, and is magically defined by the Rails enum for you.

lib/tasks/registration_version.rake
```diff
@@ -0,0 +1,84 @@
...
+
+    unless competition.registration_version_v1?
+      abort "Competition #{competition_id} is not on version 1"
+    end
+
+    LogTask.log_task("Migrating Registrations for Competition #{competition_id}") do
+      ActiveRecord::Base.transaction do
+        competition.registrations.each do |registration|
+          registration.update_column :competing_status, registration.compute_competing_status
+          registration.add_history_entry(registration.attributes.to_h, "rake task", "WST", "V2 Migration")
```
gregorbg, line 26: I am 97% sure that `atrributes` is already a hash, so you shouldn't :tm: need `to_h` here.

gregorbg: Also, what additional benefit does this history entry have?

gregorbg, line 26: Organizers who look at the history have no clue what a "rake task" is. Maybe go for something slightly more descriptive like "Automated migration script"

lib/tasks/registration_version.rake
```diff
@@ -0,0 +1,84 @@
...
+        end
+
+        competition.update_column :registration_version, :v3
```
gregorbg, line 29: (see remark in V2->V3 path below)

app/models/registration.rb
```diff
@@ -1,11 +1,12 @@
...
 class Registration < ApplicationRecord
-  scope :pending, -> { where(accepted_at: nil).where(deleted_at: nil).where(rejected_at: nil).where(waitlisted_at: nil).where(is_competing: true) }
+  scope :pending, -> { where(accepted_at: nil, deleted_at: nil, is_competing: true) }
   scope :accepted, -> { where.not(accepted_at: nil).where(deleted_at: nil) }
   scope :deleted, -> { where.not(deleted_at: nil) }
-  scope :rejected, -> { where.not(rejected_at: nil) }
-  scope :waitlisted, -> { where.not(waitlisted_at: nil) }
+  scope :cancelled, -> { where(competing_status: 'cancelled') }
+  scope :rejected, -> { where(competing_status: 'rejected') }
+  scope :waitlisted, -> { where(competing_status: 'waiting_list') }
```
gregorbg, line 9: Defining an `enum` in Rails gives you scopes (that are named like the keys of the hash) for free!

gregorbg (changes requested):

app/models/registration.rb
FinnIckler: yes, I use that in some of the methods, was just using following the convention of the other two scopes. Should we migrate those away in v3 cleanup and just use the built in enum scopes everywhere?

lib/tasks/registration_version.rake
FinnIckler: just giving a starting point for the registration history. I could also just use the accepted_at/accepted_by/deleted_at/deleted_by/created_at to create it?

lib/tasks/registration_version.rake
FinnIckler: otherwise the add_registration_history function doesn't work

lib/tasks/registration_version.rake
gregorbg: Yeah, since we have that data why not use it?

lib/tasks/registration_version.rake
gregorbg: But... why?!? Calling `Registration.create` _definitely_ writes to the DB

FinnIckler:
@thewca-bot deploy staging

gregorbg:
Tests pass! :tada:

dunkOnIT:
@thewca-bot deploy staging

dunkOnIT:
@thewca-bot deploy staging

dunkOnIT:
@thewca-bot deploy staging

FinnIckler:
@thewca-bot deploy staging

dunkOnIT:
@thewca-bot deploy staging

gregorbg (approved):

PR #10200, opened 2024-11-04
Update About Page WCA Member/Region Count (Translations)

StanFromIreland:
As I did with my [previous pull request](https://github.com/thewca/worldcubeassociation.org/pull/10170) where I updated the WCA member count I have now gone and done the same for all of the translations that had the not updated version, for some I also changed the region count from 140 to the current 158.


gregorbg:
Wow, you didn't have to do that! It is always enough to just touch the `en.yml` file.

If that file changed in our git working tree between the last server deployment and the current server deployment, our translators mailing list will receive a notification asking them to provide updated translation strings.

StanFromIreland:
@gregorbg I found it simple enough to just save the translators time so they can translate things that I can't (stuff that isn't numbers). I hope that's ok?

gregorbg:
> @gregorbg I found it simple enough to just save the translators time so they can translate things that I can't (stuff that isn't numbers). I hope that's ok?

In principle, Yes, I mean, we certainly appreciate your work. The problem here is that I assume you are not fluent in all of these languages -- and indeed languages can get _very_ weird when it comes to handling numbers. It's as easy as decimal units changing in Japanese, and can reach points of different plural cases for different quantities in Arabic and in some South Slavic languages. (We've had issues with this before, see https://github.com/svenfuchs/rails-i18n/issues/986 for example)

So all I'm trying to say is that "Oh, it's only numbers, every language does maths!" is not as evident as it may seem at first glance :upside_down_face:

StanFromIreland:
Ah, In that case I can only confirm it for the Polish translation as I am fluent in Polish. French case is also fine I believe though my French is quite elementary.

gregorbg:
Please revert the changes to the other locales, otherwise I will close the PR and just let the translators do their job

PR #10201, opened 2024-11-05
Registration Model v3 Update Tests

dunkOnIT:
Implements basic update tests. Still need to add waiting list tests

Note that I also disable random test order for local - happy to create a separate PR if we'd prefer that


spec/spec_helper.rb
```diff
@@ -86,7 +86,7 @@
   # order dependency and want to debug it, you can fix the order by providing
   # the seed, which is printed after each run.
   #     --seed 1234
-  config.order = :random
+  config.order = :random unless Rails.env.local?
```
gregorbg, line 89: Whoopsie ;)

app/models/waiting_list.rb
gregorbg, line 1: Suggestion: Since `holder` is already polymorphic (meaning, it's not necessarily a `Registration` by design), you could just rename all arguments and references below to `holder_id`.

dunkOnIT: `holder` is a competition, though, not a registration

gregorbg: Huh.

gregorbg: Maybe `entry_id`, then?

dunkOnIT: Good idea - changed!

gregorbg (approved):

PR #10202, opened 2024-11-06
Configure sidekiq-cron in its own file

gregorbg:
Was not explicitly mentioned in the `CHANGELOG.md` of sidekiq-cron when we migrated to RCv2: https://github.com/thewca/worldcubeassociation.org/pull/10163


FinnIckler (approved):
That's why cron jobs where running for me...

PR #10208, opened 2024-11-07
Replace manual stringified query parts by Rails ORM

gregorbg:
Hotfix because WFC currently cannot download Dues exports.

The stack trace is very cryptic: `ActionView::Template::Error (undefined method `instantiate' for nil)`

During my investigation, I found out by trial and error that the problem goes away when removing certain parts of the query that fetches competitions from the DB in the controller. Upon closer inspection, it was exactly those parts that contained manual, stringified statements.

Rails 7.2.2 must have done something :tm: to these styles of (admittedly old) queries without properly communicating it.

The "new" version is functionally exactly identical to the existing one, except that it lets the ORM fully handle gluing the raw SQL together. Instead of `posted_at  >= ? AND posted_at <= ?` it now uses `posted_at BETWEEN ? AND ?` under the hood, which is semantically identical (`BETWEEN` is inclusive in MySQL)

### DIFF
--- a/app/controllers/wfc_controller.rb
+++ b/app/controllers/wfc_controller.rb
@@ -19,10 +19,10 @@ class WfcController < ApplicationController
     response.headers["Content-Disposition"] = "attachment; filename=\"wfc-competitions-export-#{from}-#{to}.tsv\""
     @competitions = Competition
                     .select(select_attributes)
-                    .includes(:delegates, :championships, :organizers, :events)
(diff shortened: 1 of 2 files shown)


FinnIckler:
any chance we can add a test to this?

gregorbg:
Test for what? Test that the file downloads correctly?

FinnIckler (approved):
I'll approve it due to it being critical, but if it really is critical we should have a test for it that at least checks for a 200 and a correct content type

gregorbg:
Your wish be my command :wink: Test case added.

PR #10218, opened 2024-11-11
Fix bridge between Stripe webhook and microservice

gregorbg:
The microservice call in the Stripe webhook was completely garbled, using properties and methods on our internal model objects that definitely don't exist :O Makes me wonder what we all drank when we reviewed that PR

### DIFF
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -530,10 +530,11 @@ class RegistrationsController < ApplicationController
       stored_intent = stored_record.payment_intent
 
       stored_intent.update_status_and_charges(stripe_intent, audit_event, audit_event.created_at_remote) do |charge_transaction|
-        if stored_intent.holder.is_a? Registration
-          ruby_money = charge_transaction.money_amount
+        ruby_money = charge_transaction.money_amount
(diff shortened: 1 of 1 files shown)


gregorbg:
We haven't noticed this until now because 99.9% of payments were "fast enough" to be picked up by the redirect route, which works beautifully. With this PR, the webhook PR imitates the same list of arguments in the hopes that it will work equally well.

dunkOnIT (approved):
LGTM - as far as I can tell, the issues were the following?
- calling `stored_intent.id` instead of `charge_transaction.id` - which would have given the id of our payment_intent object rather than the id of the stripe charge
- calling `stored_intent.status` instead of `stripe_intent.status` - which would just have written the status we already had, instead of the new status received in the webhook event

gregorbg:
The problem is that previously, we were using `stripe_intent.holder.attendee_id` when we should have used `stored_intent.holder.attendee_id`.

`stripe_intent` is an API object directly returned by the Stripe SDK. It is not related to our database at all, so of course it doesn't have any `holder` property nor does it know what an `attendee_id` is. It was just fundamentally wrong in the code leading to a crash that never could have worked in the first place.

PR #10231, opened 2024-11-11
Mount apt-relevant folders in Docker Compose

gregorbg:
The packages themselves all live in `/usr`, but the information about which packages are installed live in separate directories.
In other words, if you only mount `/usr` as shared volume, the package contents will be there but the images don't know that they are there...


gregorbg:
This is blisfully unnecessary! Turns out that since our `environment` step is the one that fetches _and builds_ the dependencies, all downstream containers can use the pre-built bindings.

PR #10232, opened 2024-11-11
Use wca_status instead of Stripe's native status in payment_completion message

gregorbg:
Small part of https://github.com/thewca/worldcubeassociation.org/pull/9208, should be pretty self-explanatory.
We already have and use `wca_status` since a long while, so let's use it in places where it makes sense.

### DIFF
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -415,15 +415,14 @@ class RegistrationsController < ApplicationController
       #   this behavior differs and we overwrite created_at manually, see #stripe_webhook above.
     end
 
-    # Payment Intent lifecycle as per https://stripe.com/docs/payments/intents#intent-statuses
-    case stored_intent.payment_record.stripe_status
+    case stored_intent.wca_status
     when 'succeeded'
       flash[:success] = t("registrations.payment_form.payment_successful")
-    when 'requires_action'
+    when 'pending'
       # Customer did not complete the payment
       # For example, 3DSecure could still be pending.
       flash[:warning] = t("registrations.payment_form.errors.payment_pending")
-    when 'requires_payment_method'
+    when 'created'
       # Payment failed. If a payment fails, it is "reset" by Stripe,
       # so from our end it looks like it never even started (i.e. the customer didn't choose a payment method yet)
       flash[:error] = t("registrations.payment_form.errors.payment_reset")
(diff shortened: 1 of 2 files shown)


app/controllers/registrations_controller.rb
```diff
@@ -628,21 +628,23 @@ def payment_completion
...
       flash[:success] = t("registrations.payment_form.payment_successful")
-    when 'requires_action'
-      # Customer did not complete the payment
-      # For example, 3DSecure could still be pending.
+    when 'pending'
       flash[:warning] = t("registrations.payment_form.errors.payment_pending")
-    when 'requires_payment_method'
-      # Payment failed. If a payment fails, it is "reset" by Stripe,
-      # so from our end it looks like it never even started (i.e. the customer didn't choose a payment method yet)
+    when 'created'
```
dunkOnIT, line 638: I don't fully follow why a `created` status should mean that we display "The payment was rejected by Stripe" - I understand this is just porting what the old system was doing, just want to confirm that the same logic should still apply? 

ie, are there reasons besides the payment being rejected that we might end up in the `created` wca_status?

gregorbg: We map to wca `created` only from Stripe status `requires_payment_method`. And that Stripe status is documented as follows:

> If the payment attempt fails (for example due to a decline), the PaymentIntent’s status returns to requires_payment_method so that the payment can be retried.

See the bottom half of https://docs.stripe.com/payments/paymentintents/lifecycle

In essence, `created` means "has been initiated on the payment provider's side **but nothing else happened with it yet**". In particular, even the user trying to interact with the PI by entering an expired credit card, or waiting for funds to be transferred, would count as "some interaction" with the PI, and would thus move the lifecycle forward. `created` really means _the very initial state_ and that can only be reached if a PI has been reset.

dunkOnIT: Makes sense, thanks! Still need to do some basic testing before approving, should get to it this evening

dunkOnIT:
Question that comes up in testing: Currently, we're displaying these as flash messages - which means, they're only displayed 
- briefly
- upon progressing to the next step

This means that if the user's payment is in a non-succeeded state, they are generally not able to see our message explaining that state (as often a Stripe payment error will not even trigger a flash message to be rendered).

Instead, do we perhaps want to display them as messages, which are always available on the Payment Step screen?

dunkOnIT (approved):

PR #10234, opened 2024-11-11
Unify payment completion entry point

gregorbg:
Partial implementation of #9208 

Uses a "default parameter" in the URL (per https://guides.rubyonrails.org/routing.html#bound-parameters) so that existing Stripe redirects aren't broken, but new requests can pass in that parameter later when we merge PayPal upstream.


app/controllers/registrations_controller.rb
```diff
@@ -576,52 +576,59 @@ def stripe_webhook
...
     end
 
-    stored_intent.update_status_and_charges(stripe_intent, current_user) do |charge_transaction|
+    registration = stored_intent.holder
+
+    stored_intent.update_status_and_charges(remote_intent, current_user) do |charge_transaction|
       ruby_money = charge_transaction.money_amount
 
-      if uses_v2
+      if registration.is_a? MicroserviceRegistration
```
dunkOnIT, line 624: do these references to MicroserviceRegistration still apply, or should just be assuming v3 is the default? (Or, using the new checker methods at least)

app/controllers/registrations_controller.rb
```diff
@@ -576,52 +576,59 @@ def stripe_webhook
...
         begin
           Microservices::Registrations.update_registration_payment(
```
dunkOnIT, line 626: Also I think this method is now outdated given that we're using v3 registrations?

dunkOnIT:
Generally the functionality looks good - want to do a bit more playing around with triggering specific errors (missing payment intent, CPI, etc) but I think we need a path forward with the changes to the MicroserviceRegistration code first. I can come back around and try an update there if I have spare time, but I figure my time is better spent elsewhere at this point

dunkOnIT (approved):
Tested locally and positive case is working fine! 

I tried to repro some failure cases accounted for in the code by removing database entries (payment intents/stripe records/CPI's), but it seems like doing these and then trying to pay is either already accounted for[1] or causes the code to fail before getting to the 

[1] if I use Stripe fail details to get a WCA PI in the db, delete that WCA PI and then pay with good details, we handle this by just creating a new WCA PI. If I instead delete the PI payment record, the code fails before hitting payment_completion

If I could suspend execution as the payment_completion request is received by the Rails server I could test further, but haven't had success with attempting to do that in the past (although I've never tried _very_ hard - can put more effort into that if desired)

PR #10235, opened 2024-11-11
Add safeguards for production avatars that aren't exported to local environments

gregorbg:
A potential long-term solution would be to export `active_record_attachments` in the dev export. But to be honest, that doesn't quite feel right. They're uploaded to production for a reason, and we had nasty incidents in the past (uploading derogatory and/or anti-semitic profile pictures via staging) when sharing assets between environments.

Of course, nothing goes against _reading_ prod avatars on local, but that requires a more sophisticated solution and think-through. This one here is an intermediate fix so that people with newer avatars can log in on local and staging again in the first place.


dunkOnIT (approved):
Don't have the full context but seems reasonable

PR #10236, opened 2024-11-12
Make comment counter more clear by moving it to the label

### LINKED ISSUE #10215: Character limit on "additional comments to the organizer" is too big
image
It's not immediately clear that "0 / 240" refers to the characters in the text box above it. There's a few reasons IMO:

* The counter just looks like regular text -- it's the same font size and not italics.
* The counter is too far away from the textbox, being in its own `<p></p>` creates a big gap.
* It doesn't say that it's a character count, so it sort of just looks like 2 random numbers

Some example pages about design patterns for character limits:
https://developer.blackbaud.com/skyux/components/character-count?docs-active-tab=design
https://heymichellemac.com/design-pattern-character-limits
https://design-system.service.gov.uk/components/character-count/

FinnIckler:
fixes #10215


FinnIckler:
comment mobile
comment non movile

PR #10237, opened 2024-11-12
Default competition_factory to registration_version = :v3

dunkOnIT:


FinnIckler:
looks like it's all front end tests that are failing?

dunkOnIT:
Yep, all the frontend tests + a mailer render, which surprises me a little. My suggestion here is to rewrite the tests for v3 once we get capybara working locally, which it sounds like Gregor is close to doing?

gregorbg:
The mailer failure is actually a bug (from the initial introduction of V3). Hotfix per https://github.com/thewca/worldcubeassociation.org/commit/b44fcb5a177b1003eedcd567f4c03e97461d836e already pushed, if you rebase this PR then you should be good

FinnIckler:
That was not a bug, V3 uses the same new edit registration url as V2!
We can only change that when we have all competitions off v1

spec/features/register_for_competition_spec.rb
```diff
@@ -2,43 +2,68 @@
...
-      expect(page).to have_text("Registration form submitted - make sure to follow all instructions from the Registration Requirements below to ensure your registration is complete.")
+      expect(page).to have_text("Your registration is processing...")
+      perform_enqueued_jobs
+      expect(page).to have_text("Your registration is pending approval by the organizers.")
       registration = competition.registrations.find_by_user_id(user.id)
-      expect(registration).not_to eq nil
+      expect(registration).not_to be_nil
     end
 
     scenario "User registers for a competition with invalid guest info" do
```
dunkOnIT, line 40: EDIT: Especially given the test on Line 89, I suggest removing the guests part of this entirely and just making it a general "User submits invalid registration" smoketest

We have no way of knowing whether the registration is actually failing because of the invalid guest data - if events aren't selected, it may well just fail both times because the event selection is invalid. I think it's worth either asserting the presence of specific error text for guests, or just calling this "User can't register without selecting events

gregorbg: In this test, we don't care why exactly the registration is nil. We just care _that_ it is nil, which we can be sure because we never selected any events.

I have renamed and commented more thoroughly, does that address your concerns?

FinnIckler (approved):

PR #10238, opened 2024-11-12
Set v3 registrations as default for newly-created comps

dunkOnIT:
Is it this easy? Have tested locally and it seems to work.


gregorbg:
We initially designed the `enum` in the _hopes_ that it would be this easy, now I'm excited to see that it seems our hope has come true!

gregorbg (approved):

PR #10240, opened 2024-11-12
Remove all references to external v2 registration service

FinnIckler:
To be merged after all v2 comps are on v3

### DIFF
--- a/app/webpacker/components/RegistrationsV2/api/mocks/polling_mock.js
+++ b/app/webpacker/components/RegistrationsV2/api/mocks/polling_mock.js
@@ -2,9 +2,10 @@ import { getSingleRegistration } from '../registration/get/get_registrations';
 
 export default async function pollingMock(
   userId,
+  competition,
 ) {
   // Now that we are doing more things on Registration create we have to poll ourselves
-  const registration = await getSingleRegistration(userId);
+  const registration = await getSingleRegistration(userId, competition);
 
   return {
     processing: !registration,
--- a/app/webpacker/components/RegistrationsV2/Register/RegistrationOverview.jsx
+++ b/app/webpacker/components/RegistrationsV2/Register/RegistrationOverview.jsx
@@ -44,7 +44,7 @@ export default function RegistrationOverview({
   const queryClient = useQueryClient();
 
   const { mutate: deleteRegistrationMutation, isPending: isDeleting } = useMutation({
-    mutationFn: () => updateRegistration(competitionInfo, {
+    mutationFn: () => updateRegistration({
       user_id: registration.user_id,
       competition_id: competitionInfo.id,
       competing: {


app/models/registration.rb
```diff
@@ -293,6 +293,7 @@ def registration_history
 
   def to_v2_json(admin: false, history: false, pii: false)
     base_json = {
+      user: user,
```
gregorbg, line 296: Seems like you're sneaking in a feature addition?

app/models/competition.rb
```diff
@@ -1933,10 +1914,10 @@ def persons_wcif(authorized: false)
       } },
       :wcif_extensions,
     ]
-    # V2 registrations store the event IDs in the microservice data, not in the monolith
-    includes_associations << :events unless self.uses_microservice_registrations?
 
-    registrations_relation = self.uses_microservice_registrations? ? self.microservice_registrations : self.registrations
+    includes_associations << :events
+
+    registrations_relation = self.registrations
```
gregorbg, line 1920: You don't need to explicitly declare this into a variable anymore, you can just use `self.registrations` directly at the call-site because (I think) it's only being used once.

app/models/competition.rb
```diff
@@ -1933,10 +1914,10 @@ def persons_wcif(authorized: false)
       } },
       :wcif_extensions,
     ]
-    # V2 registrations store the event IDs in the microservice data, not in the monolith
-    includes_associations << :events unless self.uses_microservice_registrations?
 
-    registrations_relation = self.uses_microservice_registrations? ? self.microservice_registrations : self.registrations
+    includes_associations << :events
```
gregorbg, line 1918: The array `includes_associations` is literally being defined in the line directly above.
Now that the `if` / ternary check is not necessary anymore, you can just as well directly include `:events` in the list again

app/models/competition.rb
```diff
@@ -2088,13 +2069,13 @@ def set_wcif_events!(wcif_events, current_user)
...
   def update_persons_wcif!(wcif_persons, current_user)
-    registrations_relation = self.uses_microservice_registrations? ? self.microservice_registrations : self.registrations
+    registrations_relation = self.registrations
     registration_includes = [
       { assignments: [:schedule_activity] },
       :user,
       :wcif_extensions,
     ]
-    registration_includes << :registration_competition_events unless self.uses_microservice_registrations?
+    registration_includes << :registration_competition_events
```
gregorbg, line 2078: See my two remarks about inlining `registration_relation` and `registration_includes`.

app/models/registration.rb
gregorbg: Ah, now I see why you're doing this. It saves you one `getUserInfo` call in the frontend.

gregorbg: Although I do want to point out that serializing a user is quite "expensive", I think we include stuff like team membership by default (for legacy reasons). This will definitely slow down the loading even more.

app/webpacker/components/RegistrationsV2/api/mocks/polling_mock.js
```diff
@@ -2,19 +2,10 @@ import { getSingleRegistration } from '../registration/get/get_registrations';
...
-    return {
-      status: {
-        competing: registration?.competing.registration_status ?? 'processing',
-      },
-      queue_count: Math.round(Math.random() * 10),
-    };
-  }
   return {
     processing: !registration,
     queue_count: Math.round(Math.random() * 10),
```
gregorbg, line 11: Lol.

db/migrate/20241112161111_remove_microservice_v2_tables.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class RemoveMicroserviceV2Tables < ActiveRecord::Migration[7.2]
+  def change
+    drop_table :microservice_registrations
```
gregorbg, line 5: This will probably cause hard crashes when the migration has been run but AWS hasn't switched the traffic over yet. Maybe save this for a follow-up PR?

app/models/registration.rb
FinnIckler: true, that's why we removed a bunch of stuff in the users info route, so I will make sure it also includes only the data we needed

db/migrate/20241112161111_remove_microservice_v2_tables.rb
FinnIckler: makes sense. removed!

PR #10241, opened 2024-11-12
Additional Waiting List Tests

dunkOnIT:
## Questions:
- [ ] What behaviour do we want if a payload is received that contains a waiting list position, but the registration isn't on the waiting list yet? 
  - [ ] Reject it in the checker?
  - [ ] Add it to the waiting list at the end?
  - [ ] Add it to the waiting list in the given position?


FinnIckler:
I think those questions are part of a larger API modelling discussion

gregorbg:
In my opinion there's only really two options here:
1. Reject the competition because of the waiting list part of the payload explicitly
2. Drop/ignore the waiting list part of the payload implicitly

gregorbg:
If a competition is not on the waiting list -- and, crucially, if the update payload does **not** specify that it wants to be put on the waiting list -- then there should be no additional information about waiting list positions.

As Finn said, this then only boils down to a general decision about "complain loudly" (and thereby implicitly don't process the waiting list part) or "ignore quietly" (and thereby implicitly don't process the waiting list part).

Adding the user in any capacity or any position to the waiting list when the same update payload does not explicitly specify "please put me to status `waiting_list`" gets a big no-no from me.

dunkOnIT:
I'm partial to loudly shouting "hey, this registration isn't on the waiting list", which would be 2 tests: 
1. Throw an error in the registration_checker if we get a waiting list position but no competing_status: waiting_list
1. Throw an error in the `waiting_list.rb` model if we call `add` on a Registration without a `waiting_list` competing_status

gregorbg:
> I'm partial to loudly shouting "hey, this registration isn't on the waiting list", which would be 2 tests:

Okay!

> This leads to a second question: What if we get a payload which contains a `waiting_list` competing_status AND a waiting_list_position? Do we
> 
> * add it to the waiting list at the given position? Or,
> * reject it because adding and moving to different positions should be done in separate steps?

Please go with `add it to the waiting list at the given position`. The rationale here should be that the payload itself _should_ support this in the backend. If we have strong opinions about what users of our website are(n't) allowed to do, then we should block these actions in the frontend.

dunkOnIT:
> If we have strong opinions about what users of our website are(n't) allowed to do, then we should block these actions in the frontend.

My only concern with this is that we also at some point want this API to be accessible by third parties - so if there's an action we don't want to support, it also needs to be enforced at the API level (at least in my view)

gregorbg:
Just curious, what is the status on this? (Especially seeing that we discussed about waiting list issues for over one hour yesterday)

dunkOnIT:
Busy finalizing this. A few design decisions I'm making: 
- **Passing waiting_list the `registration` object instead of an `entry_id`:** 
  -  I've added a check in the `waiting_list.add` method that the registration's status must be waiting_list - however, when we call `waiting_list.add` the registration object hasn't yet been saved
  - Another option here would be to finish updating the registration before adding it to the waiting list - problem with this is that we're determining the changes for the history entry during the db transaction - and we'd like to include the waiting list position in the history (which we can only do if the reg has been added to the waiting list)
  - This also saves us a database call by using the in-memory object instead of looking up the reg again in `waiting_list.add`
  - I've changed this for the `remove` and `move` functions as well, just for the sake of consistency

TODO: 
- [ ] Add a validation that a reg must have a `waiting_list_position` in order to be saved with the `waiting_list` status
- [ ] Add waiting list position to the `changes` hash passed to `add_history_entry`

dunkOnIT:
Ready for review

app/models/waiting_list.rb
gregorbg, line 1: I have to veto this in its current state. Passing the whole registration specifically and assuming knowledge about its properties defeats the whole idea of having WaitingList be generic.
You could either:
- Perform the `registration.competing_status == Registrations::Helper::STATUS_WAITING_LIST` check outside of this model, immediately before the `add` method is being called
- Or add a `can_be_waitlisted?` method (or similar) to each and every model that should support waitlisting (kinda like an `interface` definition except Ruby doesn't support interfaces)

gregorbg: To be clear, I'm not at all against passing in the whole entity, I just don't want this class to make any assumptions about what kind of entity it is.

gregorbg: But then again, as far as I can tell, if you skip that one `raise` check that I'm complaining about then suddenly all the code ever needs is the ID of the entity so passing the whole entity becomes redundant again.

gregorbg: I'll leave the details and choice up to you, the only thing I am adamant about is not having registration-specific business logic inside this generic (polymorphic-capable) `WaitingList` class.

gregorbg (changes requested):

app/models/waiting_list.rb
dunkOnIT: I'm preferring the `can_be_waitlisted?` solution, because there's no restriction on where in the code `waiting_list.add()` could be called from - even though at the moment we only use it in `lanes/competing.rb` - and so I prefer being able to enforce restrictions about what can be added to the waiting list from one central place.

spec/requests/api_registrations_spec.rb
```diff
@@ -299,35 +299,34 @@
     end
 
     it 'accepts competitors from the waiting list' do
-      waiting_list = FactoryBot.create(:waiting_list, holder: competition)
-      waiting_list.add(registration1.user_id)
-      waiting_list.add(registration2.user_id)
-      waiting_list.add(registration3.user_id)
+      waitlisted1 = FactoryBot.create(:registration, competition: competition)
+      waitlisted2 = FactoryBot.create(:registration, competition: competition)
+      waitlisted3 = FactoryBot.create(:registration, competition: competition)
```
gregorbg, line 304: These don't have the `:waiting_list` trait, but I feel like they should? (Based on the fact that in the previous state of things, you had added them to the waiting list manually)

app/models/waiting_list.rb
gregorbg, line 1: Why not just rename from `entry_id` to `entry`?

gregorbg: Cool!

app/models/waiting_list.rb
```diff
@@ -3,30 +3,33 @@
...
 
-  def remove(entry_id)
-    update_column :entries, entries - [entry_id]
+  def remove(item)
+    update_column :entries, entries - [item.id]
   end
 
-  def add(entry_id)
+  def add(item)
+    item.can_be_waitlisted! # can_be_waitlisted! raises an error if it fails
```
gregorbg, line 11: Ruby trick: You can do `item.try(:can_be_waitlisted!)` which (obviously) _tries_ to call it if the method is defined, but doesn't explode if you forget to implement it.

app/models/registration.rb
```diff
@@ -420,6 +420,11 @@ def series_registration_info
                                    .join(" + ")
   end
 
+  def can_be_waitlisted!
```
gregorbg, line 423: Not too wild about the `can_be_` and `!` combination. I know that the usage of `!` is a bit overloaded between "throws an error" and "mutates in-place", but perhaps something like `check_waitlist!`-ish can do?

gregorbg (changes requested):

app/models/waiting_list.rb
dunkOnIT: Good point!

spec/requests/api_registrations_spec.rb
dunkOnIT: Great catch - I've added assertions to check their status before the test runs as well

app/models/waiting_list.rb
dunkOnIT: Implemented

app/models/registration.rb
dunkOnIT: Gone with `ensure_waitlist_eligibility!` which feels like it avoids sounding like a boolean check while keeping the method purpose clear

spec/models/registration_spec.rb
```diff
@@ -509,10 +508,22 @@
...
+        expect(reg5.waiting_list_position).to eq(5)
+
+        expect(waiting_list.entries.count).to eq(5)
+      end
+
       it 'removes from waiting list' do
         reg4.update_lanes!({ user_id: reg4.user.id, competing: { status: 'pending' } }.with_indifferent_access, reg4.user)
 
-        expect(reg4.waiting_list_position).to eq(nil)
+        expect(reg4.reload.waiting_list_position).to eq(nil)
```
gregorbg, line 526: Curious: Why is a `reload` necessary here when it wasn't necessary before?

spec/models/registration_spec.rb
```diff
@@ -70,7 +70,6 @@
     registration.user = user
     expect(registration).to be_valid
   end
-
```
gregorbg, line 73: Why did this poor blank line have to die? :sob:

gregorbg (changes requested):

spec/models/registration_spec.rb
dunkOnIT: I eat blank lines for breakfast (fixed)

spec/models/registration_spec.rb
dunkOnIT: Apparently is isn't? Fixed

dunkOnIT: Speculation: LIkely added the reload in an attempt to fix the test which was failing for an unrelated reason, and forgot to remove it once the test was passing again

gregorbg (approved):

PR #10243, opened 2024-11-13
Improve registration migration task process visibility 

dunkOnIT:


dunkOnIT:
@thewca-bot deploy staging

PR #10246, opened 2024-11-13
Set competing_status in import and add

FinnIckler:


app/controllers/registrations_controller.rb
```diff
@@ -183,6 +183,7 @@ def do_import
           end
         end
         registration.save!
+        registration.add_history_entry({ event_ids: registration.event_ids }, "user", current_user.id, "CSV Import")
```
gregorbg, line 186: Nitpick: Are you sure Rails polymorphic associations work with lowercase `u` in `user`?
The AR model class is called `User`, I genuinely don't know whether Rails is smart enough to "skip over" the lowercase here.

app/controllers/registrations_controller.rb
```diff
@@ -222,12 +223,13 @@ def do_add
...
         registration_comment = params.dig(:registration_data, :comments)
         registration.assign_attributes(comments: registration_comment) if registration_comment.present?
-        registration.assign_attributes(accepted_at: Time.now, accepted_by: current_user.id)
+        registration.assign_attributes(accepted_at: Time.now, accepted_by: current_user.id, competing_status: Registrations::Helper::STATUS_ACCEPTED)
         params[:registration_data][:event_ids]&.each do |event_id|
           competition_event = @competition.competition_events.find { |ce| ce.event_id == event_id }
           registration.registration_competition_events.build(competition_event_id: competition_event.id)
         end
         registration.save!
+        registration.add_history_entry({ event_ids: registration.event_ids }, "user", current_user.id, "OTS Form")
```
gregorbg, line 232: Nitpick about lowercase, see above

app/controllers/registrations_controller.rb
FinnIckler: this is not a polymorphic association at all, but just a string

gregorbg: But then how can the frontend correctly display the user's name instead of just the literal string `user` plus the ID `7121`?

gregorbg: Oh, that's surprising. So basically the frontend completely ignores the `actor_type` column and just assumes that whatever ID is written there is also a user ID?

gregorbg: You will get all of this "for free" if you set the field to a polymorphic association in Rails :wink:

gregorbg: But I agree that's out of scope for this registration

gregorbg (approved):

PR #10247, opened 2024-11-13
Skip payment and history migration if non-competing reg

dunkOnIT:


lib/tasks/registration_version.rake
```diff
@@ -88,9 +88,11 @@ namespace :registration_version do
           puts "Registration built: #{new_registration.inspect}"
           new_registration.save!
 
+          next unless registration.is_competing? # We don't need to migrate history or payments for non-competing registrations
+
           # Point any payments to the new holder
           if competition.using_payment_integrations?
-            payment_intents = registration.payment_intents
+            payment_intents = registration&.payment_intents
```
gregorbg, line 95: How can this `&` be necessary?

gregorbg (changes requested):

lib/tasks/registration_version.rake
FinnIckler, line 95: this is not needed right?

dunkOnIT: Ah it's from an earlier version, removed

gregorbg (approved):

PR #10248, opened 2024-11-13
Remove all references to v1

FinnIckler:
To be merged after all the competitions have been migrated


FinnIckler:
The two failing tests are
- my competition string, I am changing this in #10364 anyway so I will wait until that is done and
- a bunch of unused I18n keys... not sure what the best way to check how much of these are actually used in the frontend

config/routes.rb
gregorbg, line 1: The old V1 URLs can probably be deleted from this file

gregorbg: Also, why are we continuing the V2/V3 schema and just "disguising" it as V1?
The proper V1 schema is like /registrations/:primary_key_reg_id

config/locales/en.yml
```diff
@@ -1462,7 +1420,6 @@ en:
       claim_wca_id_html: "If this is not your first competition, associate your WCA ID to your WCA account %{here}."
       register_link_html: "Register for this competition %{here}."
       register_below_html: "Register for this competition below."
-      part_of_a_series: "This competition is part of a Competition Series."
```
gregorbg, line 1465: How is this reflected in the new UI?

config/locales/en.yml
```diff
@@ -1302,22 +1265,17 @@ en:
...
         show_donation: "I would like to make an additional optional donation that will go only to the organization team"
         payment_information: "Payment information"
-        payment_button: "Pay your fees with Stripe"
         ajax_error: "WCA server error"
         payment_service_error: "Payment service error"
         subtotal: "Subtotal"
       alerts:
-        not_a_number: "Please enter a valid amount."
         amount_too_low: "The total amount must be at least the registration fees to be paid."
-        amount_rather_high: "You are about to pay at least double the required amount for your registration to proceed. Please confirm the payment by pressing OK."
```
gregorbg, line 1320: I feel like we should include this alert. At least "steal" the string here and wrap it in a "V3 TODO"

gregorbg: But then again, temporarily removing it will prompt the translators to update their translation files, and quick translators like pt and th will remove it, and later they will get an email asking them to add it back again

config/locales/en.yml
```diff
@@ -1302,22 +1264,17 @@ en:
       button_text: "Pay now!"
       hints:
         donation: "Optional donation that will go only to the organization team"
-        payment_button: "Your payment information never reaches the WCA website. It is securely sent to Stripe, which handles the payment process and gets back to us."
```
gregorbg, line 1305: We should probably keep this(?)
Should be easy to add as a subtitle below the Stripe panel

config/locales/en.yml
```diff
@@ -1238,37 +1219,24 @@ en:
...
-    registered_with_account_html: "This person registered with an account. You can edit their personal information %{here}."
     sign_in: "Sign in"
     profile: "profile"
     panel_title: "Registration details"
     register: "Register!"
-    refund: "Refund Payment"
-    refund_confirmation: "Are you sure you want to refund this payment?"
     update: "Update Registration"
-    delete: "Delete"
-    delete_confirm: "Are you sure you want to delete your registration?"
```
gregorbg, line 1270: Do we have a confirmation process for self-deletions in the new system? If yes, where does that get its string from?

config/locales/en.yml
```diff
@@ -1238,37 +1219,24 @@ en:
...
-      n_events: "Number of events"
     new_registration:
       title: "Register for %{comp}"
-    registered_with_account_html: "This person registered with an account. You can edit their personal information %{here}."
     sign_in: "Sign in"
     profile: "profile"
     panel_title: "Registration details"
     register: "Register!"
-    refund: "Refund Payment"
-    refund_confirmation: "Are you sure you want to refund this payment?"
```
gregorbg, line 1267: Do we have a confirmation flow for refunds in the current system?

config/locales/en.yml
```diff
@@ -1238,37 +1219,24 @@ en:
...
       payment_requested_on: "Payment requested on: %{date}"
       payment_completed_on: "Payment completed on: %{date}"
-      #context: Counts how many parallel registrations in the same Series there are. Will display a number behind this label.
-      series_registrations: "Series registrations"
-      series_registrations_hint_html: "Hint: Series registrations are being displayed as <i>%{description}</i>"
       total: "Total"
-      n_events: "Number of events"
     new_registration:
       title: "Register for %{comp}"
-    registered_with_account_html: "This person registered with an account. You can edit their personal information %{here}."
```
gregorbg, line 1261: How is this represented in the new system?

config/locales/en.yml
```diff
@@ -1238,37 +1219,24 @@ en:
...
-      non_competing: "Non-competing registrations"
       not_paid: "Not paid"
       registered:
         without_stripe: "Registered on"
         with_stripe: "Paid on"
       #context: date is the registration creation date, eg: "2016-12-01 16:18:56 UTC"
       payment_requested_on: "Payment requested on: %{date}"
       payment_completed_on: "Payment completed on: %{date}"
-      #context: Counts how many parallel registrations in the same Series there are. Will display a number behind this label.
-      series_registrations: "Series registrations"
```
gregorbg, line 1255: Do we show parallel Series registrations in the new system?

config/locales/en.yml
```diff
@@ -1238,37 +1219,24 @@ en:
       export_csv: "Export to CSV"
       email: "Email"
       approve: "Approve"
-      reject: "Reject"
-      delete: "Delete"
       waiting_list: "Waiting List"
       approved_registrations: "Approved registrations"
-      deleted_registrations: "Deleted registrations"
-      non_competing: "Non-competing registrations"
```
gregorbg, line 1246: Do we show these in the new system?

config/locales/en.yml
```diff
@@ -1156,20 +1150,7 @@ en:
...
       not_using_wca: "This competition is not using WCA registration"
-      accepted_and_mailed:
-        one: "Registration accepted! Email notification sent."
-        other: "Registrations accepted! Email notifications sent."
-      rejected_and_mailed:
-        one: "Registration moved to waiting list"
-        other: "Registrations moved to waiting list"
-      deleted_and_mailed:
-        one: "Registration deleted"
-        other: "Registrations deleted"
```
gregorbg, line 1167: Do we have confirmation messages in the new system?

gregorbg (commented):
I noticed quite a lot of I18n strings that get deleted now but that are actually useful/relevant.

I'm rather strict on those because deleting them now and adding them back later when we have time, causes double work for fast translators :( But of course I'm open for discussion

config/routes.rb
FinnIckler: yes. I thought I would do that by reviving the other PR

config/locales/en.yml
FinnIckler: Yes, but not including the and email notification sent bit

config/locales/en.yml
FinnIckler: We currently do not

config/locales/en.yml
FinnIckler: no, I'm not sure what this did, but we don't allow registering for both competitions in a series

config/locales/en.yml
FinnIckler: Like this 
image

FinnIckler: oh actually we still use this string

FinnIckler: image

config/locales/en.yml
FinnIckler: you are right we do use this string

config/locales/en.yml
FinnIckler: accidental deletion

config/locales/en.yml
FinnIckler: yes we can add that to the PaymentStep

config/locales/en.yml
FinnIckler: yes we can add that to the PaymentStep

dunkOnIT:
Potential issues:
- Missing translation for `en.registrations.flash.registered`
- After paying for a registration I get redirected and my url has "default.host" instead of "localhost:3000"

config/locales/en.yml
FinnIckler: done

config/locales/en.yml
FinnIckler: done

app/webpacker/components/RegistrationsV2/RegistrationEdit/Refunds.jsx
```diff
@@ -75,6 +77,21 @@ function RefundRow({
...
+    refundMutation({
+      competitionId,
+      userId,
+      paymentId: refund.payment_id,
+      paymentProvider: refund.payment_provider,
+      amount: amountToRefund,
+    });
+  }).catch(() => {
+
+  });
```
gregorbg, line 94: Why is there an empty catch here?

FinnIckler: I thought useConfirm always needed it, but I guess not

app/webpacker/components/RegistrationsV2/Register/StripeWrapper.jsx
```diff
@@ -45,6 +46,15 @@ export default function StripeWrapper({
   return (
     <>
       <Header>Payment</Header>
+      <Message positive>
+        {I18n.t('registrations.payment_form.hints.payment_button')}
+      </Message>
+      { data?.api_amounts.stripe > 2 * initialAmount
```
gregorbg, line 52: You shouldn't rely on Stripe-specific values here. I think since we have the `initialAmount` from the backend already (which is guaranteed to be ISO because it's served through our backend money gem) you can just use `donationAmount > initialAmount` as condition. What do you think?

gregorbg (changes requested):

app/webpacker/components/RegistrationsV2/Register/StripeWrapper.jsx
FinnIckler: sure

config/locales/en.yml
FinnIckler: added todo

config/locales/en.yml
FinnIckler: added todo

gregorbg (approved):

PR #10255, opened 2024-11-15
Migrate assignments and WCIF extensions in V2->V3 routine

gregorbg:


gregorbg:
Closing because of weird runtime quirk. Will reopen soon.

PR #10275, opened 2024-11-18
Sidekiq-cron: Move from GitHub source tree to V2 releases

gregorbg:
Follow-up to https://github.com/thewca/worldcubeassociation.org/pull/10163


gregorbg:
The reason why we pointed to the GitHub source tree is that we explicitly _wanted_ V2 pre-releases.
Now that V2 has been officially released, the risk is maybe not zero but at least negligible.

Plus staging deployment+test takes too long :stuck_out_tongue:

PR #10278, opened 2024-11-18
Add WCIF sanity check in V2->V3 migration

gregorbg:


lib/tasks/registration_version.rake
```diff
@@ -160,6 +178,23 @@ namespace :registration_version do
...
+          puts wcif_v2.to_json
+          puts wcif_v3.to_json
+
+          raise "The WCIF output did not match. Logging debug details: First WCIF is v2, second WCIF is v3"
+        end
+
+        puts "WCIF sanity check has completed succesfully. Continue the migration? [y/n]"
+        user_confirm = gets.chomp.downcase
+
+        unless user_confirm
```
FinnIckler, line 195: don't we need to check if this is 'y' here?

PR #10282, opened 2024-11-19
Some fixes for edit profile form

danieljames-dj:
This PR aims at fixing some bugs of edit profile form noticed by WRT:

1. Edit reason was not there in the email - this was a big miss by me.
2. The email always mentioned there is a proof even if there wasn't. It was because of sending `undefined` from client.
3. Made name a required field.
4. Made country to english instead of user's locale

### DIFF
--- a/app/controllers/contacts_controller.rb
+++ b/app/controllers/contacts_controller.rb
@@ -38,7 +38,7 @@ class ContactsController < ApplicationController
 
   private def new_profile_data_key_to_value(new_profile_data, profile_data_to_change)
     if profile_data_to_change == 'country_iso2'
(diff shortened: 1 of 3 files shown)


maxidragon (commented):
I think DOB should be required as well - currently it's possible to send a form with an empty DOB

danieljames-dj:
> I think DOB should be required as well - currently it's possible to send a form with an empty DOB

Actually adding `required` won't work. I should do some other approach for that. I'll take care of that in a separate PR.

maxidragon (approved):
Tested the changes locally, LGTM.

PR #10284, opened 2024-11-19
Update WRT email nag wording

c-goodyear:
To avoid the Delegate basically saying "Results are already submitted, thanks." and getting to the _why_ they were not submitted issue.


gregorbg:
I'm going to assume because you're the WRT Leader that this is WRT-approved :wink:

PR #10294, opened 2024-11-21
Custom schedule calendar event rendering

### LINKED ISSUE #10002: Line break on calendar view
**Describe the bug**
In the "Calendar" view of the schedule, for rounds that last 20 minutes, the text is entirely on one line. This is good.

For rounds that last 25 minutes, there is a line break between the times and the name of the round. This is not quite big enough, so the line of text with the name of the round gets cut off.

**To Reproduce**
1. Add a round that is 25 minutes long.

**Expected behavior**
25-minute rounds should have text only on one line; 30-minute rounds can have a line break.

**Screenshots**
image

**Desktop (please complete the following information):**
 - OS: Linux
 - Browse: Chrome
 - Version: 129

**Additional context**
Add any other context (for instance the language you're using the site in) about the problem here.

kr-matthews:
Customize the styling depending on the length of the event, rather than relying on full calendar to do everything. Made 10min font size smaller, and adjusted line heights and how many lines to use for other lengths. One size effect is that the time text is slightly bigger now, which seems good/fine to me.

If/when this is merged, it will be easy to implement a tool-tip on the events, so that people can hover/tap on events to see the full time/title/etc. if the box happens to be too small. (We didn't do this originally because we didn't want to do custom event rendering just to get a tool-tip; but now it's worth it.) The changes can also be applied to the edit schedule view (it would be nice to be able to read 10 minute events when creating a schedule, especially since a tutorial is often that long). And on a related note, it would be good for mobile view to use a separate calendar per day, so that the days can stack on top of each other instead of squishing way too much.

Before:
image

image

After:
image

image

Examples of all event lengths:
image

image

### DIFF
--- a/app/webpacker/components/Schedule/CalendarView.jsx
+++ b/app/webpacker/components/Schedule/CalendarView.jsx
@@ -1,7 +1,7 @@
-import luxonPlugin from '@fullcalendar/luxon3';
+import luxonPlugin, { toLuxonDateTime } from '@fullcalendar/luxon3';
 import FullCalendar from '@fullcalendar/react';
 import timeGridPlugin from '@fullcalendar/timegrid';
-import { DateTime } from 'luxon';
+import { DateTime, Interval } from 'luxon';
 import React from 'react';
 import {
   earliestTimeOfDayWithBuffer,
@@ -92,10 +92,54 @@ export default function CalendarView({
         locale={calendarLocale}
         timeZone={timeZone}
         events={fcActivities}
+        // custom rendering of event content
+        eventContent={(args) => (
+          <CalendarEventView {...args} />
+        )}
       />
       {fcActivities.length === 0 && (
         <em>{I18n.t('competitions.schedule.no_activities')}</em>
       )}
     </>
+  );
+}
+
+// TODO: check events crossing midnight
+function CalendarEventView({ event, timeText, view }) {
+  const startLuxon = toLuxonDateTime(event.start, view.calendar);
+  const endLuxon = toLuxonDateTime(event.end, view.calendar);
+  const interval = Interval.fromDateTimes(startLuxon, endLuxon);
+  const lengthInMin = interval.length('minutes');
+
+  return (
+    <div className='fc-event-main-frame' style={{ overflow: 'hidden' }}>
+      {lengthInMin < 15 && (
+        <div style={{ whiteSpace: 'nowrap', fontSize: '70%', lineHeight: '1.2em' }}>
+          {timeText} - {event.title}
+        </div>
+      )}
+      {15 <= lengthInMin && lengthInMin < 20 && (
+        <div style={{ whiteSpace: 'nowrap', fontSize: '90%' }}>
+          {timeText} - {event.title}
+        </div>
+      )}
+      {20 <= lengthInMin && lengthInMin < 25 && (
+        <div style={{ whiteSpace: 'nowrap' }}>
+          {timeText} - {event.title}
+        </div>
+      )}
+      {25 <= lengthInMin && lengthInMin < 30 && (
+        <div style={{ lineHeight: '1.4em' }}>
+          <div style={{ whiteSpace: 'nowrap' }}>{timeText}</div>
+          <div>{event.title}</div>
+        </div>
+      )}
+      {30 <= lengthInMin && (
+        <div>
+          <div style={{ whiteSpace: 'nowrap' }}>{timeText}</div>
+          <div>{event.title}</div>
+        </div>
+      )}
+    </div>
   );
 }
--- a/app/webpacker/components/Schedule/CalendarView.jsx
+++ b/app/webpacker/components/Schedule/CalendarView.jsx
@@ -115,7 +115,7 @@ function CalendarEventView({ event, timeText, view }) {
   return (
     <div className='fc-event-main-frame' style={style}>
       <InnerEventContent
-        lengthInMin={lengthInMin}
+        onlyOneLine={lengthInMin < 25}
         title={event.title}
         timeText={timeText}
       />
@@ -123,8 +123,8 @@ function CalendarEventView({ event, timeText, view }) {
   );
 }
 
-function InnerEventContent({ lengthInMin, timeText, title }) {
-  if (lengthInMin < 25) {
+function InnerEventContent({ onlyOneLine, timeText, title }) {
+  if (onlyOneLine) {
     return (
       <>{timeText} - {title}</>
     );
--- a/app/webpacker/components/Schedule/CalendarView.jsx
+++ b/app/webpacker/components/Schedule/CalendarView.jsx
@@ -113,14 +113,12 @@ function CalendarEventView({ event, timeText, view }) {
   const style = getEventStyle(lengthInMin);
 
   return (
-    <div className='fc-event-main-frame' style={{ overflow: 'hidden' }}>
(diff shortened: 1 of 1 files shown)


gregorbg (commented):
Thanks for taking a look! Sorry if I'm a bit obnoxiuos, but what is the difference here? I'm having a very hard time trying to spot the difference in your Before/After...

In terms of code, where are the CSS styles in `getEventStyle` coming from? Did you steal them from the FullCalendar source code or did you come up with them yourself?

kr-matthews:
> Thanks for taking a look! Sorry if I'm a bit obnoxiuos, but what is the difference here? I'm having a very hard time trying to spot the difference in your Before/After...

The before/after pictures primarily show that not much has changed and I didn't break anything. The changes mainly make a difference for short events, and for that I only provided an 'after' picture (directly under "Examples of all event lengths:"). I should have added a 'before' picture, my bad.

I don't have a picture handy, but before (i.e. on the current site) 10-minute events are unreadable because the (single) line of text is cut off, and 25-minute events break onto 2 lines and the second text line is cut off. (See the linked issue for a picture of that.)

> In terms of code, where are the CSS styles in `getEventStyle` coming from? Did you steal them from the FullCalendar source code or did you come up with them yourself?

I came up with them myself by some experimenting. For example, in some cases changing the line height seemed preferable to shrinking the text size. Styling is not my strong point so I'm very open to changes/improvements.

gregorbg:
> I came up with them myself by some experimenting. For example, in some cases changing the line height seemed preferable to shrinking the text size. Styling is not my strong point so I'm very open to changes/improvements.

Wow, definitely kudos to you for finding that out! The point is, I'd like to avoid custom event renderers at all, if possible.

> I don't have a picture handy, but before (i.e. on the current site) 10-minute events are unreadable because the (single) line of text is cut off, and 25-minute events break onto 2 lines and the second text line is cut off. (See the linked issue for a picture of that.)

This problem description sounds like something that FullCalendar should fix on their side. Did you look through their Issues list to see whether similar behavior/glitches were reported to them? What did they say, i.e. did they give a clear and unequivocal "won't fix"?

it is worth noting that FullCalendar 7 is very short of being released (they were up to the third beta iteration before Christmas) and the rendering and layouting engine there is going to change completely, according to their announcements. I don't know yet what that means in our concrete case, but anyhow this issue should have a solution somewhere on the spectrum of "simply wait until FC7 is released and hope that the issue fixes itself" and "report an issue to FC and let them fix it upstream".

I'm willing to entertain these CSS fixes only as a _very_ last resort.

kr-matthews:
I did look at similar FullCalendar issues a year ago when we first wrote the schedule in React, and I don't remember seeing anything exactly about this, but there were similar issues that apparently weren't going to be fixed any time soon. Hopefully they will be in the new release now. I didn't check again before doing this work, I'll take another look. If the release is soon then waiting for it sounds good.

kr-matthews:
`fullcalendar` version `7.0.0-rc.0` came out 2 months ago. There doesn't seem to be any documentation for it yet. It seems like a step backwards, as far as rendering short events go.

I see many relevant github issues/comments, most of which are old (2021 and earlier) but a few from late 2024.

Before/After:
image image

gregorbg:
> `fullcalendar` version `7.0.0-rc.0` came out 2 months ago. There doesn't seem to be any documentation for it yet. It seems like a step backwards, as far as rendering short events go.

Exactly this kind of shortcoming is why the FullCalendar people have decided to halt the release cycle of 7.0 and redo their rendering engine (more or less) from scratch: https://github.com/fullcalendar/fullcalendar/issues/7907

I'm still keeping my hopes up that this kind of problem will "fix itself" with those upgrades. If we come back to this issue in two months with the "new" v7 and realize it's still broken, we might just consider adjusting the default slot height (decreasing it to make one existing slot represent less minutes, thereby stretching events longer).

Feel free to set this to draft in the meantime, and sorry for the constant delay :sweat_smile:

gregorbg:
Incredible but true: FullCalendar v7 was released a couple of hours ago at https://github.com/fullcalendar/fullcalendar/releases/tag/v7.0.0.

Does this final release version of v7 fix your concerns regarding the rendering engine? The update seems quite massive (cf. the waiting time until full release), so maybe it's worth porting to v7 with feature-parity first, and then tackle this PR again separately.

kr-matthews:
I tried following the migration guide to update to v7, but couldn't get it to work (stuck trying to install `temporal-polyfill`).

gregorbg:
FWIW, I have blocked the v7 installation on our Webpacker legacy code (per https://github.com/thewca/worldcubeassociation.org/commit/07f17b98654c153da11754e9c13f2c56b1e3417c), but I have also updated our NextJS page to FullCalendar v7 already: https://github.com/thewca/worldcubeassociation.org/commit/7ed8594c7cc2867c902c804d8f92524e202e8abf.

You can see the Worlds 2025 schedule in action, rendered by FullCalendar v7, at https://next.worldcubeassociation.org/competitions/WC2025/schedule. If you want, you can absolutely play around with it in your local dev environment! Also note that FullCalendar v7 revamped the notion of "Themes" entirely, per https://fullcalendar.io/docs/custom-themes (and a Theme Explorer available at https://themes.fullcalendar.io/).

I'd be interested to know:
- How much better/worse did the final release of v7 actually turn out?
- How eady/hard is it to play around with the theme customization? My strong preference would be to use the [render hook index](https://fullcalendar.io/docs/render-hook-index) over custom CSS files, but I am open to the latter if you can make a strong case on why it is necessary (or at least _significantly_ easier)
- Bonus: Can we tie Chakra's design system into the theming engine? Maybe an LLM can answer this better...

kr-matthews:
> You can see the Worlds 2025 schedule in action, rendered by FullCalendar v7, at https://next.worldcubeassociation.org/competitions/WC2025/schedule. If you want, you can absolutely play around with it in your local dev environment! Also note that FullCalendar v7 revamped the notion of "Themes" entirely, per https://fullcalendar.io/docs/custom-themes (and a Theme Explorer available at https://themes.fullcalendar.io/).

I included Worlds 2025 screenshots mostly to show that there weren't significant changes caused by my PR. I was mostly trying to fix the rendering of short (10-25 minute) events, as shown in my custom schedule screenshots.

I'm not sure how to run/access the new site locally (edit: Duncan has explained to me how), but from looking at other competitions (e.g. BCChampionship2026), it does appear that events of all normal durations (10 min, 15 min, 20 min, 25 min, etc.) are rendering nicely. I think v7 resolves the original issue #10002.

So I guess we can close this PR now. Or maybe, since the old schedule won't be getting further updates and will "soon" be deleted, there's no harm in having the custom styling until it dies?

> I'd be interested to know:
> 
> * How much better/worse did the final release of v7 actually turn out?
> * How eady/hard is it to play around with the theme customization? My strong preference would be to use the [render hook index](https://fullcalendar.io/docs/render-hook-index) over custom CSS files, but I am open to the latter if you can make a strong case on why it is necessary (or at least _significantly_ easier)
> * Bonus: Can we tie Chakra's design system into the theming engine? Maybe an LLM can answer this better...

Once I know how to run locally, and have some time, I could look into these. It's been a while since I've worked with theming though, and I'm not at all familiar with the design/theming of the new site.

In any case, adding a tool-tip (or something similar) with information about the round (advancement, cut-off, etc.) _should_ be possible now, and would be a good next step I think.

PR #10296, opened 2024-11-21
Send emails to organizer and attendee upon registration

dunkOnIT:


FinnIckler:
oh we just never implemented it

FinnIckler (approved):

dunkOnIT:
Yep pretty much! The other emails also need some testing, but figured I should get this out ASAP

PR #10297, opened 2024-11-22
Get Processing Status from Redis into the frontend

FinnIckler:
I wouldn't mind also getting the registration into the page and conditionally checking for if processing only if no registration exists, but I would need to touch the v1
`@registration = @competition.registrations.find_or_initialize_by(user_id: current_user.id, competition_id: @competition.id)`
line which sounds dangerous for the little time we have to get this out


FinnIckler:
@thewca-bot deploy staging

app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
```diff
@@ -79,6 +80,10 @@ export default function CompetingStep({
 
   const [processing, setProcessing] = useState(false);
```
gregorbg, line 81: What goes against just using the `irProcessing` that is passed in from the backend as initial value for this state?

FinnIckler: I tried to fix that with the useEffect, but it still occurs

gregorbg: But the loop is being caused by the `useEffect`. My point was to remove the `useEffect` and _instead_ use the initialization value here.

FinnIckler: it is not. I had it exactly like you proposed previously and it caused the same loop

FinnIckler: https://github.com/thewca/worldcubeassociation.org/pull/10297/commits/fc0048e35ff3b8793a4c2af269d871cd93c011b4
see first commit

gregorbg:
I don't see an "easy" solution here. The problem is that the "backend" version of `isProcessing` is taken as a ground truth, and once the polling actually finds out that the processing is done, this statically rendered flag from the backend is suddenly "out of date", meaning you'd have to ignore it completely once `onProcessingComplete` fired.

The technical problem is that within the body of `onProcessingComplete`, you're reloading the live-queried version of the registration, which lives much higher in the DOM tree and thus causes a full re-render. The re-render then tries to take this "outdated" backend information back into account.

I can see two solutions, with neither of them being overly amazing:
1. Introducing another "tripwire" style state that gets set to `true` once the `onProcessingComplete` hook actually fired, and when that tripwire state is `true`, the `isProcessing` flag from the backend gets fully ignored. Basically a short-circuiting mechanism. Nasty hack but works.
2. Moving the current, frontend-only `processing` state (the one created by the `useState` hook) up several levels to the same place where the `registration` is queried. That would mean you have to pass the `setProcessing` setter down several levels to be able to use it in the `onProcessingComplete` hook, which is also nasty. But at least it would avoid the re-render logic and you could then indeed just initialize `useState` with the backend value as indicated above.

The ideal solution (which requires significant effort though) would of course be to move everything into a shared `useContext` state that can be pulled in on-demand, completely making these render dependencies obsolete.

FinnIckler:
It works locally, but polling works a bit different there. So we would need to try it out on staging too

FinnIckler:
@thewca-bot deploy staging

app/controllers/registrations_controller.rb
```diff
@@ -269,6 +269,9 @@ def do_add
 
   def register
     @competition = competition_from_params
+    if current_user
+      @is_processing = Rails.cache.read(CacheAccess.registration_processing_cache_key(@competition.id, current_user.id)) || false
+    end
```
gregorbg, line 274:
```suggestion
    @is_processing = current_user.present? && Rails.cache.read(CacheAccess.registration_processing_cache_key(@competition.id, current_user.id))
```

gregorbg (changes requested):

app/controllers/registrations_controller.rb
FinnIckler: does nil evaluate to false when passed to react?

gregorbg: How would my suggested version ever evaluate to nil?

FinnIckler: oh I guess when Rails.cache.read returns nil then false is returned?

gregorbg: Yup `true && nil` is false because `nil` acts as false-y in this context

gregorbg (approved):

app/controllers/registrations_controller.rb
FinnIckler: well.. actually `true && nil` is `true`

gregorbg (approved):

PR #10299, opened 2024-11-22
Always show queue count and change processing message

FinnIckler:


dunkOnIT (approved):

config/locales/en.yml
```diff
@@ -1911,7 +1911,7 @@ en:
           waiting_list: "You are on the Waiting List at position %{waiting_list_position}."
         disclaimer: "Submission of Registration does not mean approval to compete"
         processing: "Your registration is processing..."
-        processing_longer: "Processing is taking longer than usual, don't go away!"
+        processing_longer: "Your registration has been received, and will be processed soon - this can take up to 20 seconds if our servers are busy."
```
gregorbg, line 1914: Do we really want to commit ourselves to a 20 second SLA?
I feel like "this can take a moment if our servers are busy" is more appropriate

dunkOnIT: Agreed

FinnIckler: I think the queue Count number going down will already give the user a good indication on how long it will take

dunkOnIT: Change made - should we resolve?

dunkOnIT: We could also add the "You can safely close this window - but refreshing won't make things go any faster!" line?

FinnIckler: but that's not true right now. if you close the window it will 'look' like you have to register again

gregorbg (approved):

PR #10300, opened 2024-11-22
Optimize admin registration panel query

gregorbg:
Preloads everything necessary to load the admin dashboard without individual queries

### DIFF
--- a/app/models/registration.rb
+++ b/app/models/registration.rb
@@ -294,7 +294,7 @@ class Registration < ApplicationRecord
   end
 
   def registration_history
-    registration_history_entries.includes(:registration_history_changes).map do |r|
+    registration_history_entries.map do |r|
       changed_attributes = r.registration_history_changes.each_with_object({}) do |change, attrs|
         attrs[change.key] = if change.key == 'event_ids'
                               JSON.parse(change.value) # Assuming 'event_ids' is stored as JSON array in `to`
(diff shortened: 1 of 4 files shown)


app/models/registration.rb
```diff
@@ -313,49 +313,44 @@ def registration_history
   end
 
   def to_v2_json(admin: false, history: false, pii: false)
+    private_attributes = pii ? %w[dob email] : nil
```
FinnIckler, line 316: this won't work because the pii data is assumed to be at the base of the json and not in the user field

FinnIckler: To fix it you would need to include the frontend changes from this PR
https://github.com/thewca/worldcubeassociation.org/pull/10286/files

app/models/registration_history_entry.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
 class RegistrationHistoryEntry < ActiveRecord::Base
-  has_many :registration_history_change, dependent: :destroy
+  has_many :registration_history_changes, dependent: :destroy
```
FinnIckler, line 4: how does this not break anything?

app/models/registration.rb
gregorbg: Thanks, done! :heavy_check_mark:

app/models/registration_history_entry.rb
gregorbg: Rails is flexible. It tries both singular and plural for `has_many` associations, but plural is "more correct" in terms of conventions.

FinnIckler: so add_history_entry will still work even though it does 
`registration_history_change.create`
?

dunkOnIT:
Tested locally and it seems to work, although I don't have any v3 comps on the scale of Worlds currently in my db

app/models/registration_history_entry.rb
gregorbg: No, you have to commit to either singluar or plural.
I CTRL+F'ed through the entire codebase to find all usages of the old singluar `registration_history_change` and replaced them. You will see in the diff for this very PR that `add_history_entry` was also changed accordingly :smile:

FinnIckler: oh, I missed that

dunkOnIT:
How's it working?

FinnIckler (approved):
Speed looking great!

PR #10301, opened 2024-11-24
Few cleanups in contact edit profile form

danieljames-dj:
This PR aims at few cleanups and a minor change which will be needed for first PR of tickets.

The cleanups include removal of few fields from `ContactWrt` class, which was missed in one of the previous PR.

The minor change is that the "changes" was pushed to an array after converting to human readable format. But now that logic is moved to the email body because tickets need the raw format bit early.

### DIFF
--- a/app/controllers/contacts_controller.rb
+++ b/app/controllers/contacts_controller.rb
@@ -36,25 +36,13 @@ class ContactsController < ApplicationController
     )
   end
 
-  private def new_profile_data_key_to_value(new_profile_data, profile_data_to_change)
-    if profile_data_to_change == 'country_iso2'
-      Country.find_by(iso2: new_profile_data).name_in(:en)
-    else
-      new_profile_data
-    end
-  end
-
   private def contact_wrt(requestor_details, contact_params, attachment)
-    profile_data_to_change = contact_params[:profileDataToChange]
     maybe_send_contact_email(
       ContactWrt.new(
         name: requestor_details[:name],
         your_email: requestor_details[:email],
         wca_id: User.find_by(email: requestor_details[:email])&.wca_id || 'None',
         query_type: contact_params[:queryType].titleize,
-        profile_data_to_change: profile_data_to_change&.titleize,
-        new_profile_data: new_profile_data_key_to_value(contact_params[:newProfileData], profile_data_to_change),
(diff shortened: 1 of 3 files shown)


gregorbg (changes requested):
I'm not a big fan (at all) of loading model classes and "converting" changes into human-readable form on-the-fly directly in the email template.

You should _at least_ outsource this to some helper functions.

danieljames-dj:
> I'm not a big fan (at all) of loading model classes and "converting" changes into human-readable form on-the-fly directly in the email template.
> 
> You should _at least_ outsource this to some helper functions.

Okay changed, thanks.

app/controllers/contacts_controller.rb
```diff
@@ -99,6 +87,27 @@ def contact
     end
   end
 
+  private def changes_requested_humanized(changes_requested)
+    changes_requested.map do |change|
+      case change[:field]
+      when :country_iso2
+        change_from = Country.find_by(iso2: change[:from]).name_in(:en)
+        change_to = Country.find_by(iso2: change[:to]).name_in(:en)
```
gregorbg, line 95: Now you're still doing the same thing twice. What about a function that takes "field name, and some value (no matter whether it's the old or the new value" to do the complicated logic? You can then _call_ this helper twice, once for the old and once for the new.

In other words, the "heavy" logic of converting a country iso code to its English name should only be coded once. You are free to call this logic twice, but coding it twice is not very clean 😅

app/controllers/contacts_controller.rb
```diff
@@ -99,6 +87,27 @@ def contact
...
+      when :gender
+        change_from = User::GENDER_LABEL_METHOD.call(change[:from].to_sym)
+        change_to = User::GENDER_LABEL_METHOD.call(change[:to].to_sym)
```
gregorbg, line 98: (essentially the same as the remark for country data above.

gregorbg (changes requested):

app/controllers/contacts_controller.rb
danieljames-dj: Done

app/controllers/contacts_controller.rb
danieljames-dj: Done

gregorbg (approved):

PR #10317, opened 2024-11-25
Only display guest field in Register form if guests enabled

dunkOnIT:
This fixes the fact that we don't hide the "Guests" field if "Don't ask about guests" is selected in the competition form

### DIFF
--- a/app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
+++ b/app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
@@ -337,20 +337,23 @@ export default function CompetingStep({
               error={competitionInfo.force_comment_in_registration && comment.trim().length === 0 && i18n.t('registrations.errors.cannot_register_without_comment')}
             />
           </Form.Field>
-          <Form.Field>
-            <label>{i18n.t('activerecord.attributes.registration.guests')}</label>
-            <Form.Input
-              id="guest-dropdown"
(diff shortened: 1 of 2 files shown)


app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
```diff
@@ -337,20 +337,23 @@ export default function CompetingStep({
...
-              value={guests}
-              onChange={(event, data) => {
-                setGuests(Number.parseInt(data.value, 10));
-              }}
-              min="0"
-              max={competitionInfo.guests_per_registration_limit ?? 99}
-              error={Number.isInteger(competitionInfo.guests_per_registration_limit) && guests > competitionInfo.guests_per_registration_limit && i18n.t('competitions.competition_info.guest_limit', { count: competitionInfo.guests_per_registration_limit })}
-            />
-          </Form.Field>
+          {console.log(competitionInfo.guest_entry_status)}
```
gregorbg, line 340: Whoopsie :wink:

gregorbg (changes requested):

app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
dunkOnIT: Oopsie removed :D

gregorbg (approved):

PR #10319, opened 2024-11-25
Add optional argument to run bash in console.sh

FinnIckler:
For when we want to run rake tasks on the server


dunkOnIT (approved):
Tested locally, works perfectly!

PR #10321, opened 2024-11-25
Change `console.sh` command to `bash` not `rails c`

dunkOnIT:
Tested locally and it works - this lets us run rake tasks etc using the same commands that we would locally


FinnIckler:
I opened this https://github.com/thewca/worldcubeassociation.org/pull/10319 a couple of hours ago

PR #10323, opened 2024-11-25
Use a Context for fetching Registrations

FinnIckler:
From my basic testing everything seems to work as expected. Would like someone more confident in Context design weigh in if stuff like 
```
const isRegistered = Boolean(registration) && registration.competing.registration_status !== 'cancelled';
const isAccepted = isRegistered && registration.competing.registration_status === 'accepted';
const isRejected = isRegistered && registration.competing.registration_status === 'rejected';
```
makes sense to expose in the Context or not


app/webpacker/components/RegistrationsV2/lib/RegistrationProvider.js
```diff
@@ -0,0 +1,60 @@
...
+  const value = {
+    isRegistered,
+    isAccepted,
+    isRejected,
+    hasPaid,
+    registration,
+    refetchRegistration,
+    isFetching,
+    isProcessing,
+    setIsProcessing,
```
kr-matthews, line 44: I'm not super familiar with the details here, but it seems weird to expose `setIsProcessing` to any component that wants it. It feels like this context/hook should be handling that internally and just expose `isProcessing`.

Possibly the query and effect in `Processing.jsx` could/should be moved here to accomplish that?

kr-matthews (commented):
> Would like someone more confident in Context design weigh in if stuff like
> 
> ```
> const isRegistered = Boolean(registration) && registration.competing.registration_status !== 'cancelled';
> const isAccepted = isRegistered && registration.competing.registration_status === 'accepted';
> const isRejected = isRegistered && registration.competing.registration_status === 'rejected';
> ```
> 
> makes sense to expose in the Context or not

Makes sense to me :+1:

app/webpacker/components/RegistrationsV2/lib/RegistrationProvider.js
FinnIckler: mh.. that is true, my goal was to allow for it to be set by the backend as well (if someone refreshes), but that can be accomplished without exposing the isProcessing method

app/webpacker/components/RegistrationsV2/lib/RegistrationProvider.js
```diff
@@ -0,0 +1,97 @@
...
+import { setMessage } from '../Register/RegistrationMessage';
+import pollRegistrations from '../api/registration/get/poll_registrations';
+
+const REFETCH_INTERVAL = 3000;
+
+const RegistrationContext = createContext();
+
+export default function RegistrationProvider({ competitionInfo, userInfo, children }) {
+  const dispatch = useDispatch();
+  const [isProcessing, setIsProcessing] = useState(false);
```
kr-matthews, line 16: It would be helpful to distinguish between polling and processing (though the corresponding booleans are often equal). I think many uses of `isProcessing` (including this state) should be `isPolling`, and `startProcessing` should be `startPolling`. This hook/provider should still expose `isProcessing`, but via `pollingData && !pollingData.processing` rather than the (renamed) `isPolling` state.

I didn't really understand what the code was trying to do until I made this distinction in my head.

app/webpacker/components/RegistrationsV2/Register/Processing.jsx
```diff
@@ -1,31 +1,18 @@
...
-  });
   useEffect(() => {
-    const processingComplete = competitionInfo.registration_version === 'v2' ? data?.status?.competing === 'pending'
-      : data && !data.processing;
-    if (processingComplete) {
+    if (!isProcessing) {
       onProcessingComplete();
     }
-  }, [competitionInfo.registration_version, data, onProcessingComplete]);
+  }, [isProcessing]);
```
kr-matthews, line 13: I think this could be hoisted up to `CompetingStep`, since that component now has direct access to `isProcessing`. `Processing` can just be static UI with no side effects.

kr-matthews (dismissed):

kr-matthews:
> This works, but now `isProcessing: pollingStatus !== 'success' || pollingData.processing` feels a bit weird to me because `isProcessing` will be true before we start polling (you have to do this otherwise isProcessing will be false during the fetching of the data). Maybe processing should be a string with 'pending', 'processing' and 'processed' instead?

Hm I thought `isPolling` would be only inside the provider, and not shared via the hook (only `isProcessing` and `startPolling` provided). The point of polling is to figure out whether you're done processing right? Components should only care whether the registration is processing, not how we're figuring that out (i.e. with polling). Although I guess that's already broken if they have access to `startPolling`...

I feel like maybe the provider should be responsible for doing the registration actions too, e.g. `refetchRegistration` that `CompetingStep` is currently calling. Then `CompetingStep` doesn't have to observe `isProcessing` and `isPolling`, and can just observe `registration` or some other property and see directly from that when to call `nextStep()` (I think? I'm not familiar with the backend/processing part of registrations, so maybe I don't know what I'm talking about).

FinnIckler:
Seems like this still works after the merge. Maybe we should use that test matrix again to test, after the registration error message refactor?

app/webpacker/components/RegistrationsV2/lib/RegistrationProvider.js
```diff
@@ -0,0 +1,103 @@
...
+      {children}
+    </RegistrationContext.Provider>
+  );
+}
+
+export const useRegistration = () => {
+  const context = useContext(RegistrationContext);
+  if (!context) {
+    throw new Error('useRegistration must be used within a RegistrationProvider');
+  }
```
gregorbg, line 101: Is this really necessary? I don't feel like we're doing this for any other contexts, or are we?

app/webpacker/components/RegistrationsV2/lib/RegistrationProvider.js
```diff
@@ -0,0 +1,103 @@
...
+    hasPaid,
+    isAccepted,
+    isFetching,
+    isRegistered,
+    isRejected,
+    isPolling,
+    refetchRegistration,
+    registration,
+    pollingData,
+    startPolling]);
```
gregorbg, line 88:
```suggestion
    startPolling,
  ]);
```

app/webpacker/components/RegistrationsV2/lib/RegistrationProvider.js
```diff
@@ -0,0 +1,103 @@
...
+      dispatch(
+        showMessage(
+          `competitions.registration_v2.errors.${error?.response?.data?.json?.error || 'unknown'}`,
+          'negative',
+        ),
+      );
+    },
+  });
+
+  const isRegistered = Boolean(registration) && registration.competing.registration_status !== 'cancelled';
```
gregorbg, line 60: Why is the explicit cast to a boolean necessary here?

app/webpacker/components/RegistrationsV2/lib/RegistrationProvider.js
```diff
@@ -0,0 +1,103 @@
...
+  const stopPolling = useCallback(() => {
+    setIsPolling(false);
+  }, [setIsPolling]);
+
+  const { data: pollingData, status: pollingStatus } = useQuery({
+    queryKey: ['registration-status-polling', userInfo.id, competitionInfo.id],
+    queryFn: async () => pollRegistrations(userInfo.id, competitionInfo),
+    refetchInterval: REFETCH_INTERVAL,
+    onSuccess: () => {
+      setPollCounter(pollCounter + 1);
```
gregorbg, line 32:
```suggestion
      setPollCounter((prevCounter) => prevCounter + 1);
```

gregorbg: Makes it a little less dependent on `pollCounter`

app/webpacker/components/RegistrationsV2/RegistrationEdit/RegistrationEditor.jsx
```diff
@@ -67,7 +63,7 @@ export default function RegistrationEditor({ competitor, competitionInfo }) {
     },
     onSuccess: (data) => {
       queryClient.setQueryData(
-        ['registration-admin', competitionInfo.id, competitor.id],
+        ['registration', competitionInfo.id, competitor.id],
```
gregorbg, line 66: Will this cause problems if you're "meddling" with the query state inside the context?

app/webpacker/components/RegistrationsV2/Register/Processing.jsx
```diff
@@ -1,30 +1,13 @@
...
-  });
-  useEffect(() => {
-    const processingComplete = data && !data.processing;
-    if (processingComplete) {
-      onProcessingComplete();
-    }
-  }, [data, onProcessingComplete]);
   return (
-    <Modal open={data?.status?.competing !== 'pending' || !data?.processing} dimmer="blurring">
+    <Modal open dimmer="blurring">
```
gregorbg, line 10: Why is it suddenly viable to keep the modal always open?

app/webpacker/components/RegistrationsV2/lib/RegistrationProvider.js
FinnIckler: doesn't look like it is

app/webpacker/components/RegistrationsV2/Register/Processing.jsx
FinnIckler: because of 
```
{isPolling && (
        <Processing
          competitionInfo={competitionInfo}
          user={user}
        />
      )}
```

app/webpacker/components/RegistrationsV2/RegistrationEdit/RegistrationEditor.jsx
kr-matthews: It would probably be better if the context/hook was doing the mutation internally, and providing `updateRegistrationMutation` and `isUpdating` to this component.

app/webpacker/components/RegistrationsV2/Register/Processing.jsx
```diff
@@ -1,30 +1,13 @@
-import { useQuery } from '@tanstack/react-query';
-import React, { useEffect, useState } from 'react';
+import React from 'react';
 import { Message, Modal } from 'semantic-ui-react';
 import I18n from '../../../lib/i18n';
-import pollRegistrations from '../api/registration/get/poll_registrations';
+import { useRegistration } from '../lib/RegistrationProvider';
 
-const REFETCH_INTERVAL = 3000;
+export default function Processing() {
```
kr-matthews, line 6: Might make more sense to name this component/file `ProcessingModal`.

app/webpacker/components/RegistrationsV2/Register/RegistrationOverview.jsx
```diff
@@ -13,21 +13,21 @@ import { useDispatch } from '../../../lib/providers/StoreProvider';
...
   const hasRegistrationEditDeadlinePassed = hasPassed(
     competitionInfo.event_change_deadline_date ?? competitionInfo.start_date,
   );
 
-  const isRejected = registration.competing.registration_status === 'rejected';
-
   const deleteAllowed = (competitionInfo.competitor_can_cancel === 'always')
-    || (competitionInfo.competitor_can_cancel === 'not_accepted' && registration.competing.registration_status !== 'accepted')
+    || (competitionInfo.competitor_can_cancel === 'not_accepted' && !isAccepted)
     || (competitionInfo.competitor_can_cancel === 'unpaid' && !registration.payment?.has_paid);
```
kr-matthews, line 31: `useRegistration` provides `hasPaid`

app/webpacker/components/RegistrationsV2/Register/index.jsx
```diff
@@ -122,7 +107,6 @@ function Register({
             preferredEvents={preferredEvents}
             competitionInfo={competitionInfo}
             registration={registration}
```
kr-matthews, line 109: StepPanel could access registration itself via the hook.

kr-matthews (commented):
I could look into encapsulating the update registration mutation and the polling if you want, maybe as a follow-up PR since this one has been up for a while and seems to work right now.

app/webpacker/components/RegistrationsV2/Register/index.jsx
FinnIckler: it actually does, just didn't remove that

app/webpacker/components/RegistrationsV2/RegistrationEdit/RegistrationEditor.jsx
FinnIckler: I tried this out, but I need to call `nextStep` in the `onSuccess`, and I can't do that if I move it to the Context.

app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
```diff
@@ -94,6 +93,13 @@ export default function CompetingStep({
     }
   }, [isRegistered, registration, setSelectedEventIds]);
 
+  useEffect(() => {
+    if (isPolling && !isProcessing) {
+      refetchRegistration();
+      nextStep();
+    }
+  }, [isPolling, isProcessing, nextStep, refetchRegistration]);
```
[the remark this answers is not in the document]
FinnIckler: same, I can't call `nextStep` from the provider

PR #10328, opened 2024-11-26
Add support for newcomer month

dunkOnIT:
Questions:
- Do we want some kind of validation on when to allow newcomer spots to be reserved? I see these options:
  1. [what I'm currently doing] just setting a constant to true/false when newcomer month comps are allowed at all
  2. Setting NEWCOMER_MONTH_START/END_DATE and checking if the comp's {start date/end date?} falls inside that range
  3. Having a boolean on the competition table to mark a comp as a newcomer comp - but I sort of prefer this being implicitly defined by whether or not there are newcomer spots reserved


app/webpacker/components/CompetitionForm/FormSections/CompetitorLimit.js
```diff
@@ -17,6 +17,7 @@ export default function CompetitorLimit() {
       <ConditionalSection showIf={hasLimit}>
         <InputNumber id="count" min={0} />
         <InputTextArea id="reason" />
+        <InputNumber id="newcomerReservedSpots" min={0} />
```
dunkOnIT, line 20: Calculate and display the max number of newcomer reserved spots allowed per the 50% WCRP exception limit

app/models/competition.rb
```diff
@@ -188,6 +189,9 @@ class Competition < ApplicationRecord
   MAX_MARKDOWN_LENGTH = 255
   MAX_COMPETITOR_LIMIT = 5000
   MAX_GUEST_LIMIT = 100
+  NEWCOMER_MONTH_ENABLED = true
```
gregorbg, line 192: Do we really need to hard-code this? We could just leave the form field as `enabled` all the time and have WCAT manually verify.

At least until we come up with a smarter solution of "marking" newcomer month start and end dates in the DB

app/models/competition.rb
```diff
@@ -242,6 +246,30 @@ class Competition < ApplicationRecord
...
+
+    max_newcomer_spots = (competitor_limit * MAX_NEWCOMER_SPOTS_RESERVED_FRACTION).to_i
+    if newcomer_reserved_spots > max_newcomer_spots
+      errors.add(:newcomer_reserved_spots, 'cant reserve more than 50% of spots for newcomers')
+    end
+  end
+
+  # TODO: Make sure this query is optimized
+  def newcomers_competing
+    registrations.competing_status_accepted.includes(:user).select { |registration| registration.user.newcomer? }
```
gregorbg, line 265: Now that we're fully committed to V3, this can be `registrations.accepted...`

gregorbg: Regarding optimization: That depends on what you're trying to achieve.
- If you really want to get the whole list of users, to access their properties (name, dob, email, whatever) or display them in a frontend _fully_, then this is as optimized as it can get
- If the only purpose of this method is to obtain a _count_ of newcomers, then you effectively discard all of the "hard work" of optimizing and preloading. A more effective way would need a `newcomer` **scope** on `User` to then do something along the lines of `registrations.accepted.joins(:user).merge(User.newcomer).count`

spec/factories/competitions.rb
```diff
@@ -99,8 +99,17 @@
     refund_policy_percent { 0 }
     guests_entry_fee_lowest_denomination { 0 }
 
+    # TODO: This should change to :v3
```
gregorbg, line 102: I think it has already if you merge in `main` now :wink:

gregorbg (changes requested):

dunkOnIT:
I've had to disable a CyclomaticComplexity warning on `validate_update_status` - I think `validate_update_status` is due for a refactor, and #10519 will track doing that - but it is beyond the scope of this PR.

app/models/competition.rb
dunkOnIT: Updated!

dunkOnIT:
@gregorbg tests passing, ready for review

db/migrate/20241126084446_add_newcomer_reserved_spots_to_competitions.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddNewcomerReservedSpotsToCompetitions < ActiveRecord::Migration[7.2]
+  def change
+    add_column :Competitions, :newcomer_reserved_spots, :integer, default: 0, null: false
```
gregorbg, line 5: You could use `null` here as a way of saying "I don't want to specify such a limit".
Although admittedly, `0` serves the same effect, so I have only a slight bias towards an actual `null` column

lib/registrations/registration_checker.rb
```diff
@@ -184,6 +185,19 @@ def validate_update_status!(new_status, competition, current_user, target_user,
         raise WcaExceptions::RegistrationError.new(:forbidden, Registrations::ErrorCodes::ALREADY_REGISTERED_IN_SERIES) if
           new_status == Registrations::Helper::STATUS_ACCEPTED && existing_registration_in_series?(competition, target_user)
 
+        if new_status == Registrations::Helper::STATUS_ACCEPTED && competition.competitor_limit_enabled?
+          raise WcaExceptions::RegistrationError.new(:forbidden, Registrations::ErrorCodes::COMPETITOR_LIMIT_REACHED) if
+            competition.registrations.competing_status_accepted.count >= competition.competitor_limit
+
+          if !target_user.newcomer? && competition.newcomer_reserved_spots > 0
```
gregorbg, line 192: In line with my other comment above, this check would make slightly more sense if you use `null` instead of a zero value.

app/webpacker/components/CompetitionForm/FormSections/CompetitorLimit.js
gregorbg: Is this comment a mental TODO?

app/models/user.rb
```diff
@@ -387,6 +388,10 @@ def country
     Country.find_by_iso2(country_iso2)
   end
 
+  def newcomer?
+    person.nil? || wca_id.start_with?(Time.current.year.to_s)
+  end
```
gregorbg, line 393: This is a little bit ambiguous definition of the term "Newcomer".
In the intuitive sense, a newcomer only denotes a person who has never competed before. The second part after the `||` is specific to WCA Newcomer Month. Can we rename the method or maybe split it in two?

app/models/user.rb
```diff
@@ -57,6 +57,7 @@ class User < ApplicationRecord
   has_many :user_avatars, dependent: :destroy, inverse_of: :user
 
   scope :confirmed_email, -> { where.not(confirmed_at: nil) }
+  scope :newcomers, -> { where(wca_id: nil).or(where('wca_id LIKE ?', "#{Time.current.year}%")) }
```
gregorbg, line 60: Same as my other remark below about the definition "Newcomer".

app/models/competition.rb
```diff
@@ -242,6 +246,24 @@ class Competition < ApplicationRecord
   validates :external_website, :external_registration_page, length: { maximum: MAX_URL_LENGTH }
   validates :contact, length: { maximum: MAX_MARKDOWN_LENGTH }
 
+  validate :validate_newcomer_reserved_spots
+  private def validate_newcomer_reserved_spots
+    return unless competitor_limit.present?
+
+    max_newcomer_spots = (competitor_limit * MAX_NEWCOMER_SPOTS_RESERVED_FRACTION).to_i
```
gregorbg, line 253: Do the rules of Newcomer Month (or WCRP or whatnot) somehow indicate whether this should be rounded up or down?
In any case, I think it would be good to make that explicit by using either `floor` or `ceil` (or perhaps even `round` FWIW)

app/models/competition.rb
gregorbg: You marked this conversation as resolved but the constant is still there. It is completely unused as far as I can tell.

gregorbg (changes requested):

lib/registrations/registration_checker.rb
dunkOnIT: See comment above

app/webpacker/components/CompetitionForm/FormSections/CompetitorLimit.js
dunkOnIT: I honestly have no idea - it seems like a nice idea, but newcomer month is soon enough that I'd say let's just get this out the door and worry about QOL if another newcomer month ever comes around

app/models/competition.rb
dunkOnIT: It doesn't - I've set it to `floor`, on the assumptions that (a) it doesn't matter, and (b) if it did, we'd rather have 1 more flexible spot than 1 specifically reserved for newcomers - given that the newcomer reservation portion is already very generous

app/models/competition.rb
```diff
@@ -242,6 +246,24 @@ class Competition < ApplicationRecord
   validates :external_website, :external_registration_page, length: { maximum: MAX_URL_LENGTH }
   validates :contact, length: { maximum: MAX_MARKDOWN_LENGTH }
 
+  validate :validate_newcomer_reserved_spots
+  private def validate_newcomer_reserved_spots
+    return unless competitor_limit.present?
```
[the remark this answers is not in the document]
dunkOnIT: Done

app/models/user.rb
gregorbg: Rather than a comment, you should rename this method. If the name is `newcomer?` then there is **definitely** a strong intuitional bias towards "has never competed before".

The first thing that comes to mind is `newcomer_month_eligible?`

app/webpacker/components/CompetitionForm/FormSections/CompetitorLimit.js
gregorbg: Okidoki

db/migrate/20241126084446_add_newcomer_reserved_spots_to_competitions.rb
gregorbg: The other PR will probably be merged first :wink:

spec/models/user_spec.rb
```diff
@@ -791,4 +791,21 @@
       expect(user.teams_committees_at_least_senior_roles).not_to include(wrt_role)
     end
   end
+
+  describe 'newcomer?' do
```
gregorbg, line 795: Relevant if we rename the method.

spec/factories/registrations.rb
```diff
@@ -38,10 +38,14 @@
       competing_status { Registrations::Helper::STATUS_WAITING_LIST }
     end
 
-    trait :newcomer do
+    trait :first_timer do
```
gregorbg, line 41: I like this as a term, but I would use it the other way around!
As in, `newcomer` is someone who really has absolutely zero competition experience, while `first_timer` is someone who competed for the first time this year.

But I'm happy to defer to consensus and a broader vote

gregorbg (changes requested):

app/models/competition.rb
```diff
@@ -189,6 +190,9 @@ class Competition < ApplicationRecord
   MAX_MARKDOWN_LENGTH = 255
   MAX_COMPETITOR_LIMIT = 5000
   MAX_GUEST_LIMIT = 100
+  NEWCOMER_MONTH_ENABLED = true
+  MAX_NEWCOMER_SPOTS_RESERVED_FRACTION = 0.5
```
gregorbg, line 197: Basically, we should also change this to `MAX_NEWCOMER_MONTH_SPOTS_RESERVED_FRACTION`, but that becomes unfortunately long. Any ideas for a shorter name?

app/models/user.rb
```diff
@@ -61,6 +61,8 @@ class User < ApplicationRecord
   has_many :user_avatars, dependent: :destroy, inverse_of: :user
 
   scope :confirmed_email, -> { where.not(confirmed_at: nil) }
+  scope :newcomers, -> { where(wca_id: nil) }
+  scope :newcomer_month_eligible, -> { where(wca_id: nil).or(where('wca_id LIKE ?', "#{Time.current.year}%")) }
```
gregorbg, line 65: You can actually define this as `-> { newcomers.or(...) }`

app/models/competition.rb
```diff
@@ -246,6 +250,23 @@ class Competition < ApplicationRecord
...
+  end
+
+  def newcomers_competing_count
+    registrations.accepted.joins(:user).merge(User.newcomers).count
+  end
+
+  def newcomer_month_reserved_spots_remaining
+    newcomer_month_eligible_competitors = registrations.accepted.joins(:user).merge(User.newcomer_month_eligible).count
+    newcomer_reserved_spots - newcomer_month_eligible_competitors
+  end
```
gregorbg, line 268: Can you move these into functions under `registration.rb` similar to the already existing `def accepted_count` and `def accepted_and_paid_pending_count`?

gregorbg: In the latter case of `newcomer_month_reserved_spots_remaining`, it will suffice to only move the count itself into `registration.rb` and leave the subtraction here

app/models/user.rb
```diff
@@ -397,6 +399,10 @@ def country
     Country.find_by_iso2(country_iso2)
   end
 
+  def newcomer?
```
gregorbg, line 402: Just curious, are you using this method anywhere except maybe in tests?

db/migrate/20241126084446_add_newcomer_reserved_spots_to_competitions.rb
gregorbg: Now that other PR is merged, so you can go full steam ahead with `null` support here

lib/registrations/registration_checker.rb
gregorbg: Now that other PR is merged, so you can rethink `null` support

app/models/competition.rb
```diff
@@ -173,6 +173,7 @@ class Competition < ApplicationRecord
     event_change_deadline_date
     competition_series_id
     registration_version
+    newcomer_reserved_spots
```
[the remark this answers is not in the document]
dunkOnIT: Done

app/models/competition.rb
dunkOnIT: Changed to `NEWCOMER_MONTH_RESERVATIONS_FRACTION` which isn't _much_ shorter but perhaps acceptably not-massive. My next suggestions would be just `NEWCOMER_MONTH_RESERVATIONS` (ie, drop the "fraction")

app/models/user.rb
dunkOnIT: Ooh, that's nice - changed!

app/models/competition.rb
dunkOnIT: Moved!

app/models/user.rb
dunkOnIT: Using it in tests, but also in `registration_checker:196`:
```ruby
          if !target_user.newcomer? && competition.newcomer_month_reserved_spots > 0
```

gregorbg:
Best wishes from RuboCop :wink:

app/models/user.rb
gregorbg: Then that reference should probably be `if !target_user.newcomer_month_eligible?`, or not??!

lib/registrations/registration_checker.rb
```diff
@@ -188,6 +188,18 @@ def validate_update_status!(new_status, competition, current_user, target_user,
         raise WcaExceptions::RegistrationError.new(:forbidden, Registrations::ErrorCodes::ALREADY_REGISTERED_IN_SERIES) if
           new_status == Registrations::Helper::STATUS_ACCEPTED && existing_registration_in_series?(competition, target_user)
 
+        if new_status == Registrations::Helper::STATUS_ACCEPTED && competition.competitor_limit_enabled?
+          raise WcaExceptions::RegistrationError.new(:forbidden, Registrations::ErrorCodes::COMPETITOR_LIMIT_REACHED) if
+            competition.registrations.competing_status_accepted.count >= competition.competitor_limit
+
+          if !target_user.newcomer? && competition.newcomer_month_reserved_spots > 0
```
gregorbg, line 195: See my other remark above

spec/controllers/competitions_controller_spec.rb
```diff
@@ -885,10 +885,10 @@
         expect(competition.results_posted_by).to eq wrt_member.id
       end
 
-      it "sends notifications of id claim possibility to newcomers" do
+      it "sends notifications of id claim possibility to first-timers" do
         competition = FactoryBot.create(:competition, :registration_open)
-        FactoryBot.create_list(:registration, 2, :accepted, :newcomer, competition: competition)
-        FactoryBot.create_list(:registration, 3, :pending, :newcomer, competition: competition)
+        FactoryBot.create_list(:registration, 2, :accepted, :first_timer, competition: competition)
+        FactoryBot.create_list(:registration, 3, :pending, :first_timer, competition: competition)
```
gregorbg, line 891: This is now a bit inconsistent with the `newcomer` / `newcomer_month_eligible` thingy above

gregorbg (changes requested):

spec/controllers/competitions_controller_spec.rb
dunkOnIT: Yep this is the result of an over-eager grep/replace, I think

lib/registrations/registration_checker.rb
dunkOnIT: Definitely agree - why aren't my tests failing then?

lib/registrations/registration_checker.rb
```diff
@@ -188,6 +188,18 @@ def validate_update_status!(new_status, competition, current_user, target_user,
         raise WcaExceptions::RegistrationError.new(:forbidden, Registrations::ErrorCodes::ALREADY_REGISTERED_IN_SERIES) if
           new_status == Registrations::Helper::STATUS_ACCEPTED && existing_registration_in_series?(competition, target_user)
 
+        if new_status == Registrations::Helper::STATUS_ACCEPTED && competition.competitor_limit_enabled?
+          raise WcaExceptions::RegistrationError.new(:forbidden, Registrations::ErrorCodes::COMPETITOR_LIMIT_REACHED) if
+            competition.registrations.competing_status_accepted.count >= competition.competitor_limit
```
gregorbg, line 193: Are you trying to tell me we never checked competitor limits before?!

gregorbg (changes requested):

lib/registrations/registration_checker.rb
```diff
@@ -182,12 +182,21 @@ def contains_organizer_fields?(request, organizer_fields)
...
+
+          if competition.enforce_newcomer_month_reservations? && !target_user.newcomer_month_eligible?
+            available_spots = competition.competitor_limit - competition.registrations.competing_status_accepted.count
+
+            raise WcaExceptions::RegistrationError.new(:forbidden, Registrations::ErrorCodes::NO_UNRESERVED_SPOTS_REMAINING) if
+              (available_spots <= competition.newcomer_month_reserved_spots_remaining) && competition.registration_currently_open?
```
gregorbg, line 196: Why are you bothering to check `competition.registration_currently_open?` here, but not during the "general" competitor limit check above?

From my gut feeling, it should either be both or none.

lib/registrations/registration_checker.rb
```diff
@@ -182,12 +182,21 @@ def contains_organizer_fields?(request, organizer_fields)
...
-          competition.registrations.accepted.count >= competition.competitor_limit
         raise WcaExceptions::RegistrationError.new(:forbidden, Registrations::ErrorCodes::ALREADY_REGISTERED_IN_SERIES) if
           new_status == Registrations::Helper::STATUS_ACCEPTED && existing_registration_in_series?(competition, target_user)
 
+        if new_status == Registrations::Helper::STATUS_ACCEPTED && competition.competitor_limit_enabled?
+          raise WcaExceptions::RegistrationError.new(:forbidden, Registrations::ErrorCodes::COMPETITOR_LIMIT_REACHED) if
+            competition.registrations.competing_status_accepted.count >= competition.competitor_limit
+
+          if competition.enforce_newcomer_month_reservations? && !target_user.newcomer_month_eligible?
+            available_spots = competition.competitor_limit - competition.registrations.competing_status_accepted.count
```
gregorbg, line 193: I think you can reuse the `competition.registrations.competing_status_accepted.count` part. If you wanna be very sneaky, you can even reuse the whole `available_spots` variable, and for the "general" check above you throw if `available_spots <= 0`.

gregorbg: (but the latter is really just personal taste, feel free to keep it the way it is as you see fit.)

app/models/registration.rb
```diff
@@ -320,6 +320,10 @@ def self.accepted_and_paid_pending_count
     accepted_count + pending.with_payments.count
   end
 
+  def self.newcomer_month_eligible_competitors_count
+    accepted.joins(:user).merge(User.newcomer_month_eligible).count
```
gregorbg, line 324:
```suggestion
    joins(:user).merge(User.newcomer_month_eligible).accepted_count
```

lib/registrations/registration_checker.rb
dunkOnIT: If registration is closed and the only people left for an organizer to accept are non-newcomers, we don't want the organizer's attempt to accept them being blocked by the newcomer limit. 

Although now that I've typed it out I'm not fully convinced - because it may well be the case that once registration is closed the organizer still wants to respect the newcomer month reserved spots

PR #10330, opened 2024-11-27
Fix recompute_timestamps not setting the other timestamp to nil

FinnIckler:
This isn't entirely needed, because you usually don't go from deleted -> accepted, but if an organizer accidentally deleted someone and puts them back to accepted the registration will still be marked as deleted

### DIFF
--- a/app/models/registration.rb
+++ b/app/models/registration.rb
@@ -64,8 +64,10 @@ class Registration < ApplicationRecord
       self.deleted_at = nil
     when Registrations::Helper::STATUS_ACCEPTED
       self.accepted_at = DateTime.now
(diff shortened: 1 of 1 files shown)


gregorbg (commented):
I feel like I had this in my initial PR and then either you or I myself removed it for "some reason" :tm:

How does V1 handle this kind of situation?

FinnIckler:
V1 does it the same
```
if !registration.accepted?
          if registration.update(accepted_at: Time.now, accepted_by: current_user.id, deleted_at: nil)
```

gregorbg (approved):

Issue #10333, opened 2024-11-28
In the new registration system DOB and Email are exported as 'undefined' in registration csv


pranav-027:
**Describe the bug**
After exporting the registration CSV in the new reg system DOB and emails are exported as undefined.

FinnIckler:
ah yeah we moved those from registration.email to registration.user.email

PR #10334, opened 2024-11-28
Fixed dob and email showing as undefined in exported csv

pranav-027:
registration.email -> registration.user.email
registration.dob -> registration.user.dob

fixes: https://github.com/thewca/worldcubeassociation.org/issues/10333


dunkOnIT (approved):
Thanks for the fix!

PR #10335, opened 2024-11-28
unify edit endpoints

FinnIckler:
It feels kinda silly getting the registration and then only using the id, but we do need the competition as well. 
So we could instead fetch the competition separately? Or we just use the registration from the backend...


gregorbg:
Why is this breaking V1?

FinnIckler:
you are right. you can just edit V1 Registrations In the V3 Edit Form no problem

gregorbg:
Why are you entirely removing the V1 edit form anyways? It has its own legacy V1 routes since forever, which can continue to exist.

FinnIckler:
because I thought we don't want any `if registration_version_v3?` dead code anymore

gregorbg:
Yes, but that's for the other PR. This one should only concern cleaning up routes

FinnIckler:
So I still duplicate the routes but just make the v3 route use an id and then merge it together in a later PR?

gregorbg:
My point was to remove the v2-style routes that pass competition ID+user ID and use V1-style routes for the V3 components as well

FinnIckler:
I misunderstood then, I thought you wanted to just have one route. Let me reopen this PR with only those changes

gregorbg:
Well, it will end up being just one route - that is, V3 will switch to use the V1 style routes and V1 will keep its V1 routes.

So there's only V1-style routes which identify the registration by primary key ID in the end

Issue #10336, opened 2024-11-29
CSV export: Region names with commas overflow to next column


dunkOnIT:
Report received:
"some regions (for example, Hong Kong, China or Macau, China) are parsed by most applications (e.g. google sheets) as being 2 columns wide, effectively making records for people from those regions not matching the correct fields. Additionally, on the exports, the birthdate and email fields are undefined for every entry.

To replicate both issues, you can look at the registrations for [this competition](https://www.worldcubeassociation.org/competitions/BerkeleyDecTricubealon2024/edit/registrations)"

I suspect this is because the comma in region names which have commas is not being escaped - @pranav-027 is that an issue you'd like to take a look at?

pranav-027:
Hello @dunkOnIT. Sure I would take a look. You can assign it to me

pranav-027:
Fixed: https://github.com/thewca/worldcubeassociation.org/pull/10337

PR #10337, opened 2024-11-29
Fixed countryId overflow in registration csv for countries with comma in name

pranav-027:
Fixes: https://github.com/thewca/worldcubeassociation.org/issues/10336

Screenshot: 
image

I feel it's a reasonable fix for now. But in the future the function could be re-written to eliminate all possible errors


dunkOnIT (approved):
Thanks for the quick fix!

Issue #10320, opened 2024-11-25
Multiple Iterations of Senior / regional delegate emails when organizer addition/removal notification is sent.


pranav-027:
**Describe the bug**
Someone reported this on Discord: when an organizer is added or removed, or when the competitions are confirmed, the email contains duplicate iterations of the emails of senior/regional delegates.

**Screenshots**
Screenshot from Discord: 
![1](https://github.com/user-attachments/assets/b1030e12-117b-46fb-8100-74caecfa39ed)

Screenshots from my local testing:
![image](https://github.com/user-attachments/assets/c1a9ca06-9365-49c4-b75a-ba04ca29568e)

![image](https://github.com/user-attachments/assets/febcefdf-4274-4735-9210-49770586bbc7)

gregorbg:
This error occurs because of a quirk in our Delegate role metadata join.

Some users are marked as Delegates multiple times, most notably if they have Senior/Regional responsibilities. In that case, there's one entry for them as a "normal" Delegate and once more in their extended role (Senior Delegate or Regional Delegate). This "ruins" the JOIN under the hood, making some rows appear multiple times.

I will point you to a "proper" fix directly on the relevant PR.

PR #10339, opened 2024-11-30
Fixed multiple senior email iterations in organizer addition/removal notification

pranav-027:
Fixes: https://github.com/thewca/worldcubeassociation.org/issues/10320

image

I wasn't able to find the root cause, but this seems like a good fix for the places where emails are visually represented. The emails are still duplicated in the recipient list, but thankfully, the email client handles that duplication.

### DIFF
--- a/app/models/competition.rb
+++ b/app/models/competition.rb
@@ -14,7 +14,7 @@ class Competition < ApplicationRecord
   has_many :competitors, -> { distinct }, through: :results, source: :person
   has_many :competitor_users, -> { distinct }, through: :competitors, source: :user
   has_many :competition_delegates, dependent: :delete_all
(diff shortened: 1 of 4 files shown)


gregorbg (changes requested):
Formally requesting the `distinct` directly in the `competition.rb` model as indicated in my previous comment.

pranav-027:
Fixed! 

image

image

gregorbg (commented):
Super nitpick, but I think `includes(...).distinct` is a bit more efficient (it will allow Rails to consider the preloads before collapsing down the rows)

pranav-027:
fixed and tested!

gregorbg (approved):

PR #10351, opened 2024-12-02
Add WRT to sso_moderator

maxidragon:
This change was approved by @c-goodyear (WRT Leader) and @lnzainn  (WCT Leader)

Is there any way to test it locally?

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -203,7 +203,7 @@ class UsersController < ApplicationController
   end
 
   private def sso_moderator?(user)
-    user.communication_team?
+    user.communication_team? || user.results_team?
   end
 
   def sso_discourse


maxidragon:
> Oof, if you _really_ want to test this, you'd have to spin up an entire Discourse forum instance locally.
> None of us was ever brave enough to even consider trying that, but just looking at the code I think this doesn't need thorough testing.

Okay, thanks!

> I'm curious though, what is the context of this change? Why does WRT need this permission?

The context is that WRT needs moderator privileges to perform anonymizations. It’s easier to just add us to the code because, for some unknown reason, Discourse revokes manually granted privileges after logging out. As a result, WRT members can not even log in on another browser without losing their permissions.

gregorbg:
> Discourse revokes manually granted privileges after logging out

This is technically not correct. Discourse overrides the privileges upon _logging in again_. Every single Discourse login makes Discourse "phone home" to our WCA SSO, and that WCA SSO (currently, before this PR is merged) tells Discourse that WRT members are not moderators, so the SSO payload of the renewed login _after you previously logged out_ comes back as `moderator: false`

c-goodyear:
@gregorbg I can confirm I was aware, and if any member tried to acquire permissions like this without a discussion, they would quite quickly no longer be a WRT member 😁

lnzainn:
Can confirm that I was aware. Thanks Gregor for approving, and Maks for the PR

PR #10353, opened 2024-12-02
Sort by name if country or event count is the same in Competitor List

FinnIckler:


gregorbg (approved):
Okay we'll just leave (1) as a future optimization then

Issue #10283, opened 2024-11-19
Edit Profile form - don't allow 'multiple countries'


danieljames-dj:
The solution is to replace the country selector of [EditProfileForm](https://github.com/thewca/worldcubeassociation.org/blob/main/app/webpacker/components/ContactEditProfilePage/EditProfileForm.jsx) with existing [CountrySelector](https://github.com/thewca/worldcubeassociation.org/blob/main/app/webpacker/components/CountrySelector/CountrySelector.js).

If anybody has a better solution, please comment out.

If anybody is interested to work on it, please go ahead after dropping a comment here.

Note: Please test the functionality using MailCatcher once this change is implemented.

MichaScant:
Hey, can I work on this issue?

danieljames-dj:
@MichaScant Thanks for your interest, sure you can work on this issue. Please let us know if you face any issues. I'm assigning this issue to you.

MichaScant:
Hello, are there instructions I can follow when testing my code with MailCatcher?

maxidragon:
> Hello, are there instructions I can follow when testing my code with MailCatcher?

If you are running a website with Docker, you can view all emails sent from the app using MailCatcher. Simply visit http://localhost:1080 in your browser.

PR #10354, opened 2024-12-02
updated country selector

MichaScant:
Fixed the issue, where we shouldn't allow multiple countries to be selected. To fix this I replaced the country selector of [EditProfileForm](https://github.com/thewca/worldcubeassociation.org/blob/main/app/webpacker/components/ContactEditProfilePage/EditProfileForm.jsx) with the existing [CountrySelector](https://github.com/thewca/worldcubeassociation.org/blob/main/app/webpacker/components/CountrySelector/CountrySelector.js).

Issue: #10283


app/webpacker/components/ContactEditProfilePage/EditProfileForm.jsx
```diff
@@ -82,7 +77,10 @@ export default function EditProfileForm({
   };
 
   const handleFormChange = (e, { name: formName, value }) => {
-    setEditedProfileDetails((prev) => ({ ...prev, [formName]: value }));
+    setEditedProfileDetails((prev) => {
+      const newState = { ...prev, [formName]: value };
+      return newState;
+    });
```
gregorbg, line 83: I don't think putting it on two lines is necessary

app/webpacker/components/ContactEditProfilePage/EditProfileForm.jsx
```diff
@@ -102,13 +100,10 @@ export default function EditProfileForm({
...
       />
-      <Form.Select
-        options={countryOptions}
+      <CountrySelector
         label={i18n.t('activerecord.attributes.user.country_iso2')}
-        name="country_iso2"
-        search
-        value={editedProfileDetails?.country_iso2}
-        onChange={handleFormChange}
+        selected={editedProfileDetails?.country_iso2}
```
gregorbg, line 105: `selected` is not a supported property by `CountrySelector`. You should pass it as `countryIso2`

app/webpacker/components/ContactEditProfilePage/EditProfileForm.jsx
```diff
@@ -102,13 +100,10 @@ export default function EditProfileForm({
         onChange={handleFormChange}
         required
       />
-      <Form.Select
-        options={countryOptions}
+      <CountrySelector
         label={i18n.t('activerecord.attributes.user.country_iso2')}
```
gregorbg, line 104: `label` is not a supported property by `CountrySelector`.

app/webpacker/components/ContactEditProfilePage/EditProfileForm.jsx
```diff
@@ -102,13 +100,10 @@ export default function EditProfileForm({
...
-        name="country_iso2"
-        search
-        value={editedProfileDetails?.country_iso2}
-        onChange={handleFormChange}
+        selected={editedProfileDetails?.country_iso2}
+        onChange={(event, data) => handleFormChange(null, { name: 'country_iso2', value: data.value })}
```
gregorbg, line 106: If you keep passing in `name` (not sure why you deleted that in the first place), then just keeping `onChange={handleFormChange}` should work

gregorbg (changes requested):

gregorbg (approved):
Thanks a lot! Unfortunately, you cannot just rename the `I18n` import due to historical reasons. So I went ahead and reverted that.

Otherwise LGTM, thank you for your contribution!

PR #10364, opened 2024-12-03
New tooltips for new competing statuses in my competitions

### LINKED ISSUE #10361: "My competitions" status shows "you are currently on the waiting list" when waiting for registration approval
**Describe the bug**
On the My Competitions page, if you are waiting for a registration to be approved, when hovering over the status icon it says "You are currently on the waiting list." even if the competition is not full and you are not on a waiting list.

**To Reproduce**
1. Register for a competition that requires approval.
2. Go to My Competitions.
3. Hover over the hourglass symbol next to the competition.
4. The message is as described.

**Expected behavior**
It should say something like "Your registration is pending approval."

**Screenshots**
Screenshot 2024-12-03 135824

**Desktop**
 - OS: Windows 10
 - Browser: Firefox
 - Version: 133.0

FinnIckler:
Any input on the wording @dunkOnIT ?
fixes #10361

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1393,7 +1393,7 @@ en:
       reg_opens_too_early: "Registration is set to start in less than the required 48 hours."
       registration_already_closed: "The deadline to register has already passed. Please set it to be in the future."
       tooltip_registered: "You are registered."
-      tooltip_pending: "Your registration has not been looked at by the organizers or is incomplete. Please check the competition page for details"
+      tooltip_pending: "Your registration is awaiting manual approval. Please ensure you have followed any payment instructions in the Competition Page."
       tooltip_waiting_list: "You are currently on the waiting list."
       tooltip_deleted: "Unfortunately your registration has been deleted. Please check your email for details, or contact the organiser(s)."
       confirmed_visible: "This competition is confirmed and visible."
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1393,7 +1393,7 @@ en:
       reg_opens_too_early: "Registration is set to start in less than the required 48 hours."
       registration_already_closed: "The deadline to register has already passed. Please set it to be in the future."
       tooltip_registered: "You are registered."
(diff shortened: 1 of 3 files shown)


spec/views/competitions/my_competitions.html.erb_spec.rb
```diff
@@ -23,7 +23,7 @@
 
   it "shows you are on the waiting list" do
     render
-    expect(rendered).to match 'You are currently on the waiting list'
+    expect(rendered).to match 'Your registration has not been looked at by the organizers or is incomplete. Please check the competition page for details This competition is not confirmed and not visible.'
```
gregorbg, line 26: There seems to be a copy/paste whoopsie going on here? Superfluous `This competition is not confirmed and not visible.` dangling at the end...

gregorbg (changes requested):

dunkOnIT:
Have updated the pending tooltip warning

config/locales/en.yml
```diff
@@ -1393,6 +1393,7 @@ en:
       reg_opens_too_early: "Registration is set to start in less than the required 48 hours."
       registration_already_closed: "The deadline to register has already passed. Please set it to be in the future."
       tooltip_registered: "You are registered."
+      tooltip_pending: "Your registration is awaiting manual approval. Please ensure you have followed any payment instructions in the Competition Page."
```
gregorbg, line 1396: Not only **payment** instructions -- it could really be _any_ instructions (for example, "Don't forget to enter a venue name in the comment field or else we cannot approve you")

gregorbg (changes requested):

config/locales/en.yml
dunkOnIT: I'm just a bit worried about cluttering this field - and I'm assuming that if a user has paid  but not fulfilled some other requirement, organizers will reach out to the competitor for the missing information

dunkOnIT: Perhaps: 
> Please ensure you have followed any instructions (eg, payment) listed in the Competition Page

gregorbg (approved):

PR #10367, opened 2024-12-03
Avoid showing EditPersonTicketWorkbench if the ticket is closed

danieljames-dj:
Currently the edit form is visible in ticket even for closed tickets. It will be good if we avoid showing the edit form (which is there inside the workbench) if the ticket is already closed.

### DIFF
--- a/app/webpacker/lib/wca-data.js.erb
+++ b/app/webpacker/lib/wca-data.js.erb
@@ -223,4 +223,5 @@ export const avatarImageTypes = <%= Rails.application.config.active_storage.web_
 
 // ----- TICKETS -----
 export const ticketTypes = <%= Ticket::TICKET_TYPES.to_json %>;
+export const ticketStatuses = <%= Ticket::TICKET_TYPES.transform_values { |value| value.constantize.statuses }.to_json %>;
 export const ticketLogActionTypes = <%= TicketLog.action_types.to_json %>;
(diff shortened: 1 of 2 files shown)


app/webpacker/lib/wca-data.js.erb
```diff
@@ -223,4 +223,5 @@ export const avatarImageTypes = <%= Rails.application.config.active_storage.web_
 
 // ----- TICKETS -----
 export const ticketTypes = <%= Ticket::TICKET_TYPES.to_json %>;
+export const ticketStatuses = <%= Ticket::TICKET_TYPES.transform_values { |value| value.constantize.statuses }.to_json %>;
```
gregorbg, line 226:
```suggestion
export const ticketStatuses = <%= Ticket::TICKET_TYPES.transform_values { |value| value.safe_constantize&.statuses }.to_json %>;
```

gregorbg: Just in case one status isn't quite working as expected :wink:

gregorbg (changes requested):

app/webpacker/lib/wca-data.js.erb
danieljames-dj: Thanks, done.

gregorbg (approved):

PR #10368, opened 2024-12-03
 Make psychsheets sort by average by default

pranav-027:
Making new psych sheet behavior consistent with old one

### DIFF
--- a/app/webpacker/components/RegistrationsV2/Registrations/RegistrationList.jsx
+++ b/app/webpacker/components/RegistrationsV2/Registrations/RegistrationList.jsx
@@ -98,13 +98,7 @@ export default function RegistrationList({ competitionInfo }) {
   const [psychSheetEvent, setPsychSheetEvent] = useState();
   const [psychSheetSortBy, setPsychSheetSortBy] = useState('average');
 
-  useEffect(() => {
-    if (psychSheetEvent === '333mbf') {
-      setPsychSheetSortBy('single');
-    } else {
-      setPsychSheetSortBy('average');
-    }
-  }, [psychSheetEvent]);
+
 
   const { isLoading: isLoadingPsychSheet, data: psychSheet } = useQuery({
     queryKey: [
--- a/app/webpacker/components/RegistrationsV2/Registrations/RegistrationList.jsx
+++ b/app/webpacker/components/RegistrationsV2/Registrations/RegistrationList.jsx
@@ -96,15 +96,7 @@ export default function RegistrationList({ competitionInfo }) {
   const changeSortColumn = (name) => dispatch({ type: 'CHANGE_SORT', sortColumn: name });
 
   const [psychSheetEvent, setPsychSheetEvent] = useState();
-  const [psychSheetSortBy, setPsychSheetSortBy] = useState('average');
-
-  useEffect(() => {
-    if (psychSheetEvent === '333mbf') {
-      setPsychSheetSortBy('single');
-    } else {
-      setPsychSheetSortBy('average');
-    }
-  }, [psychSheetEvent]);
+  const psychSheetSortBy = useMemo(() => (psychSheetEvent === '333mbf' ? 'single' : 'average'), [psychSheetEvent]);
 
   const { isLoading: isLoadingPsychSheet, data: psychSheet } = useQuery({
     queryKey: [
@@ -124,7 +116,6 @@ export default function RegistrationList({ competitionInfo }) {
 
   const registrationsWithPsychsheet = useMemo(() => {
     if (psychSheet !== undefined) {
-      setPsychSheetSortBy(psychSheet.sort_by);
       return psychSheet.sorted_rankings.map((p) => {
         const registrationEntry = registrations.find((r) => p.user_id === r.user_id);
         return { ...p, ...registrationEntry };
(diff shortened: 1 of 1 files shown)


app/webpacker/components/RegistrationsV2/Registrations/RegistrationList.jsx
```diff
@@ -95,7 +96,15 @@ export default function RegistrationList({ competitionInfo }) {
...
-  const [psychSheetSortBy, setPsychSheetSortBy] = useState('single');
+  const [psychSheetSortBy, setPsychSheetSortBy] = useState('average');
+
+  useEffect(() => {
+    if (psychSheetEvent === '333mbf') {
+      setPsychSheetSortBy('single');
+    } else {
+      setPsychSheetSortBy('average');
+    }
+  }, [psychSheetEvent]);
```
gregorbg, line 107: This is not a good usage of `useEffect`. In essence, you're only recomputing the initial value for `useMemo` above. So, you should replace this whole effect hook with a single line along the lines of `const default sort = isMultiBld ? 'single' : 'average'` and then pass that new value as initializer: `useMemo(psychSheetEvent)`

gregorbg (changes requested):

app/webpacker/components/RegistrationsV2/Registrations/RegistrationList.jsx
pranav-027: Done!

app/webpacker/components/RegistrationsV2/Registrations/RegistrationList.jsx
```diff
@@ -95,7 +96,7 @@ export default function RegistrationList({ competitionInfo }) {
   const changeSortColumn = (name) => dispatch({ type: 'CHANGE_SORT', sortColumn: name });
 
   const [psychSheetEvent, setPsychSheetEvent] = useState();
-  const [psychSheetSortBy, setPsychSheetSortBy] = useState('single');
+  const psychSheetSortBy = useMemo(() => (psychSheetEvent === '333mbf' ? 'single' : 'average'), [psychSheetEvent]);
```
maxidragon, line 99: IIRC in the old system all blindfolded events were sorted by single, not just `333mbf`

app/webpacker/components/RegistrationsV2/Registrations/RegistrationList.jsx
```diff
@@ -115,7 +117,6 @@ export default function RegistrationList({ competitionInfo }) {
 
   const registrationsWithPsychsheet = useMemo(() => {
     if (psychSheet !== undefined) {
-      setPsychSheetSortBy(psychSheet.sort_by);
```
gregorbg, line 118: Why is this line being deleted?

app/webpacker/components/RegistrationsV2/Registrations/RegistrationList.jsx
```diff
@@ -94,8 +94,10 @@ export default function RegistrationList({ competitionInfo }) {
   const { sortColumn, sortDirection } = state;
   const changeSortColumn = (name) => dispatch({ type: 'CHANGE_SORT', sortColumn: name });
 
+  const blindFoldedEvents = ['333mbf', '333bf', '444bf', '555bf'];
```
gregorbg, line 97: You don't need to hardcode this. We have events data available as `import { events } from 'lib/wca-data.js.erb'` (note you might need to change the relative path to the file)

app/webpacker/components/RegistrationsV2/Registrations/RegistrationList.jsx
```diff
@@ -94,8 +94,10 @@ export default function RegistrationList({ competitionInfo }) {
...
+
   const [psychSheetEvent, setPsychSheetEvent] = useState();
-  const [psychSheetSortBy, setPsychSheetSortBy] = useState('single');
+  const psychSheetSortBy = useMemo(() => (blindFoldedEvents.includes(psychSheetEvent) ? 'single' : 'average'), [psychSheetEvent]);
```
gregorbg, line 100: Using my suggestion from above, you can compute this as `events.byId[psychSheetEvent].recommendedFormat().sortBy`

gregorbg (changes requested):

app/webpacker/components/RegistrationsV2/Registrations/RegistrationList.jsx
pranav-027: Otherwise, react throws an error that `setPsychSheetSortBy` is not defined

gregorbg: Ah, because you're changing to a `useMemo` above. I see. But that means when loading a psych sheet from the backend, you cannot change the sort order anymore. I feel like this will break lots of existing use-cases

(for example, this will break proper sorting on events like 333 which have genuine use-cases of using either single _or_ average for sorting)

gregorbg:
Kindly ping @pranav-027 

I think the approach from my most recent review will make the code more efficient and future-proof (if we ever decide to add events)

pranav-027:
Hello @gregorbg, I didn't get time to look into this because of other commitments. I will implement all the suggested changes soon.

gregorbg:
What was that "close and then reopen immediately" about?

pranav-027:
I accidentally deleted branch from my IDE and then restored it

pranav-027:
Created new PR cuz my IDE is crazy: https://github.com/thewca/worldcubeassociation.org/pull/10626

PR #10369, opened 2024-12-04
Select stakeholder if a user is part of multiple stakeholders

danieljames-dj:
There are different cases where the user needs to represent a single stakeholder - for example, once we have the feature to change status, the person who change should represent one stakeholder so that we can add that in the logs.

In this PR, if the user is part of multiple stakeholders in a ticket, we will ask the user to select one of them.

### DIFF
--- a/app/webpacker/components/Tickets/index.jsx
+++ b/app/webpacker/components/Tickets/index.jsx
@@ -37,6 +37,20 @@ function SkateholderSelector({ stakeholderList, setUserSelectedStakeholder }) {
   );
 }
 
+function TicketContent({ ticketDetails, currentStakeholder, sync }) {
+  return (
+    <>
+      <TicketHeader ticketDetails={ticketDetails} />
+      <TicketWorkbench
+        ticketDetails={ticketDetails}
+        sync={sync}
+        currentStakeholder={currentStakeholder}
+      />
+      <TicketLogs logs={ticketDetails.ticket.ticket_logs} />
+    </>
+  );
+}
+
 export default function Tickets({ id }) {
   const {
     data: ticketDetails, sync, loading, error,
@@ -56,16 +70,21 @@ export default function Tickets({ id }) {
       <Container fluid>
         {userSelectedStakeholder && (
           <Message>
-            {`You are currently viewing the ticket as stakeholder "${userSelectedStakeholder.stakeholder.name}". If you wish to change, please refresh the page.`}
+            {`You are currently viewing the ticket as stakeholder "${userSelectedStakeholder.stakeholder.name}".`}
+            <Button
+              onClick={() => setUserSelectedStakeholder(false)}
+            >
+              Click here to change
+            </Button>
           </Message>
         )}
-        <TicketHeader ticketDetails={ticketDetails} />
-        <TicketWorkbench
-          ticketDetails={ticketDetails}
-          sync={sync}
-          currentStakeholder={currentStakeholder}
-        />
-        <TicketLogs logs={ticketDetails.ticket.ticket_logs} />
+        {!shouldUserSelectStakeholder && (
+          <TicketContent
+            ticketDetails={ticketDetails}
+            currentStakeholder={currentStakeholder}
+            sync={sync}
+          />
+        )}
       </Container>
       <Modal open={shouldUserSelectStakeholder}>
         <Modal.Header>Select stakeholder</Modal.Header>
--- a/app/webpacker/components/Tickets/index.jsx
+++ b/app/webpacker/components/Tickets/index.jsx
@@ -1,5 +1,7 @@
-import React from 'react';
-import { Container } from 'semantic-ui-react';
+import React, { useState } from 'react';
+import {
+  Button, Container, Dropdown, Message, Modal,
+} from 'semantic-ui-react';
 import useLoadedData from '../../lib/hooks/useLoadedData';
 import { actionUrls } from '../../lib/requests/routes.js.erb';
 import Loading from '../Requests/Loading';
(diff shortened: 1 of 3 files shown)


app/webpacker/components/Tickets/index.jsx
```diff
@@ -8,19 +10,72 @@ import TicketHeader from './TicketHeader';
...
-    <Container fluid>
-      <TicketHeader ticketDetails={ticketDetails} />
-      <TicketWorkbench ticketDetails={ticketDetails} sync={sync} />
-      <TicketLogs logs={ticketDetails.ticket.ticket_logs} />
-    </Container>
+    <>
+      <Container fluid>
+        {userSelectedStakeholder && (
+          <Message>
+            {`You are currently viewing the ticket as stakeholder "${userSelectedStakeholder.stakeholder.name}". If you wish to change, please refresh the page.`}
```
gregorbg, line 59: Refreshing the page _really_ shouldn't be necessary in React. That's what we have "state" and "useMemo" for. If the dropdown value changes, the rest of the DOM can immediately react to it.

If necessary, move the stakeholder selector higher in the component hierarchy, so that the value is determined first, and then the children can react as needed.

gregorbg (changes requested):

app/webpacker/components/Tickets/index.jsx
danieljames-dj: Okay I knew react can handle it, but tbh I was just lazy to implement that part 😬 I've done it now. Thank you.

app/webpacker/components/Tickets/index.jsx
```diff
@@ -8,19 +10,91 @@ import TicketHeader from './TicketHeader';
 import TicketWorkbench from './TicketWorkbench';
 import TicketLogs from './TicketLogs';
 
+function SkateholderSelector({ stakeholderList, setUserSelectedStakeholder }) {
+  const [selectedOption, setSelectedOption] = useState(stakeholderList[0]);
```
gregorbg, line 14: We have a custom `useInputState` hook for form inputs

app/webpacker/components/Tickets/index.jsx
```diff
@@ -8,19 +10,91 @@ import TicketHeader from './TicketHeader';
...
+        to visit the ticket page.
+      </p>
+      <Dropdown
+        options={stakeholderList.map((requesterStakeholder) => ({
+          key: requesterStakeholder.id,
+          text: requesterStakeholder.stakeholder.name,
+          value: requesterStakeholder,
+        }))}
+        value={selectedOption}
+        onChange={(__, { value }) => setSelectedOption(value)}
```
gregorbg, line 28: You can directly pass `onChange={setSelectedOption}` if you use the hook that I suggested above in my other comment

app/webpacker/components/Tickets/index.jsx
```diff
@@ -8,19 +10,91 @@ import TicketHeader from './TicketHeader';
...
 export default function Tickets({ id }) {
   const {
     data: ticketDetails, sync, loading, error,
   } = useLoadedData(actionUrls.tickets.show(id));
 
+  const [userSelectedStakeholder, setUserSelectedStakeholder] = useState();
+  const currentStakeholder = userSelectedStakeholder || ticketDetails?.requester_stakeholders[0];
+  const shouldUserSelectStakeholder = (
+    !userSelectedStakeholder
+    && ticketDetails?.requester_stakeholders?.length > 1);
```
gregorbg, line 63: This is more like `userHasToSelectStakeholder`, right?

app/webpacker/components/Tickets/index.jsx
```diff
@@ -8,19 +10,91 @@ import TicketHeader from './TicketHeader';
...
+          <Message>
+            {`You are currently viewing the ticket as stakeholder "${userSelectedStakeholder.stakeholder.name}".`}
+            <Button
+              onClick={() => setUserSelectedStakeholder(false)}
+            >
+              Click here to change
+            </Button>
+          </Message>
+        )}
+        {!shouldUserSelectStakeholder && (
```
gregorbg, line 81: It was very hard for me to figure out during review why you only show the ticket body if the user doesn't have to select a stakeholder.
Actually it's more like "When the user has selected a stakeholder or when we can derive the stakeholder automatically". Is there any way you can come up with a better boolean combination + clearer naming?

gregorbg: For example, you could make it so that `requester_stakeholder[0]` is only used as default if there is only one stakeholder in the first place. Then this check would be as easy as `currentStakeholder && (<TicketContent>)` which is far more readable

app/webpacker/components/Tickets/index.jsx
```diff
@@ -8,19 +10,91 @@ import TicketHeader from './TicketHeader';
...
+          <TicketContent
+            ticketDetails={ticketDetails}
+            currentStakeholder={currentStakeholder}
+            sync={sync}
+          />
+        )}
+      </Container>
+      <Modal open={shouldUserSelectStakeholder}>
```
gregorbg, line 89: It seems that you "force" this modal open whenever the user has to make a choice. This is good, but have you made sure that the default SemUI behavior prevents the user from closing the modal by clicking outside the modal window frame?

I know it's definintely possible to set this "forced modal" mode in SemUI, I just don't know whether it's the default or not.

gregorbg (changes requested):

app/webpacker/components/Tickets/index.jsx
danieljames-dj: I did a bit of research. I initially thought that by default it was forced modal as I couldn't close the modal by clicking outside or pressing escape button.

But after research I realized that it was acting as forced modal because I didn't had `onClose` in the modal. When I printed logs in `onClose`, it got printed when I click outside or press escape. So, I made `closeOnEscape` and `closeOnDimmerClick` to `false`. I didn't write `onClose` because I don't have to do any action there.

app/webpacker/components/Tickets/index.jsx
danieljames-dj: Okay, accepted your suggestion. Hope this is more readable now. Please let me know if any more changes are required.

app/webpacker/components/Tickets/index.jsx
danieljames-dj: After some cleanups this variable was removed.

gregorbg (approved):

PR #10372, opened 2024-12-05
Allowing Incedent log visible to staff/public or editors

MichaScant:
Added a drop-down menu to each incident log, allowing you to set its permissions to be either public (all can see), draft (WCA members and those with permission to edit the log can see), or staff (only staff can see).

### DIFF
--- a/app/controllers/incidents_controller.rb
+++ b/app/controllers/incidents_controller.rb
@@ -38,8 +38,16 @@ class IncidentsController < ApplicationController
 
   def show
     set_incident
-    unless @incident.visible_to?(current_user)
+
+    # Check visibility and redirect based on user role
+    if @incident.visibility == 'draft' && !current_user&.can_manage_incidents?
       redirect_to_root_unless_user(:can_manage_incidents?)
+      return
+    end
+
+    if @incident.visibility == 'staff' && !current_user&.staff?
+      redirect_to_root_unless_user(:staff?)
+      nil
     end
   end
 
@@ -88,12 +96,6 @@ class IncidentsController < ApplicationController
 
   def update
     set_incident
-    if @incident.update(incident_params)
-      flash[:success] = "Incident was successfully updated."
-      redirect_to @incident
-    else
-      render :edit
-    end
   end
 
   def destroy
--- a/app/controllers/incidents_controller.rb
+++ b/app/controllers/incidents_controller.rb
@@ -13,11 +13,13 @@ class IncidentsController < ApplicationController
 
   def index
     base_model = Incident.includes(:competitions, :incident_tags)
-    if current_user&.can_view_delegate_matters?
-      @incidents = base_model.all
-    else
-      @incidents = base_model.resolved
-    end
+    @incidents = if current_user&.can_manage_incidents? # WRC members see all
+                   base_model.all
+                 elsif current_user&.can_view_delegate_matters? # Staff see staff + public
+                   base_model.staff_visible
+                 else # Public users only see public
+                   base_model.publicly_visible
+                 end
 
     respond_to do |format|
       format.html do
@@ -36,8 +38,8 @@ class IncidentsController < ApplicationController
 
   def show
     set_incident
-    unless @incident.resolved?
-      redirect_to_root_unless_user(:can_view_delegate_matters?)
+    unless @incident.visible_to?(current_user)
+      redirect_to_root_unless_user(:can_manage_incidents?)
     end
   end
 
@@ -111,14 +113,17 @@ class IncidentsController < ApplicationController
     end
 
     def incident_params
-      params.require(:incident).permit(
+      permitted_params = [
         :title,
         :private_description,
         :private_wrc_decision,
         :public_summary,
         :tags,
         :digest_worthy,
-        incident_competitions_attributes: [:id, :competition_id, :comments, :_destroy],
-      )
+        :visibility,
+        { incident_competitions_attributes: [:id, :competition_id, :comments, :_destroy] },
+      ]
+
+      params.require(:incident).permit(permitted_params)
     end
 end
--- a/app/models/incident.rb
+++ b/app/models/incident.rb
@@ -7,7 +7,7 @@ class Incident < ApplicationRecord
 
   accepts_nested_attributes_for :incident_competitions, allow_destroy: true
 
-  attribute :visibility, :integer, default: 0
+  attribute :visibility, :integer, default: 1
 
   enum visibility: {
     draft: 0,          # Only visible to WRC members
@@ -17,11 +17,10 @@ class Incident < ApplicationRecord
 
   scope :publicly_visible, -> { where(visibility: :visible) }
   scope :staff_visible, -> { where(visibility: [:staff, :visible]) }
-  scope :wrc_visible, -> { all } # WRC can see all incidents
+  scope :wrc_visible, -> { all }
 
   validate :digest_sent_at_consistent
   validates_presence_of :title
-  validate :visibility_transition_valid
 
   include Taggable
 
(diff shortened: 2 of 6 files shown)


app/controllers/incidents_controller.rb
```diff
@@ -13,11 +13,13 @@ class IncidentsController < ApplicationController
...
     base_model = Incident.includes(:competitions, :incident_tags)
-    if current_user&.can_manage_incidents?
-      @incidents = base_model.all
-    else
-      @incidents = base_model.resolved
-    end
+    @incidents = if current_user&.can_manage_incidents? # WRC members see all
+                   base_model.all
+                 elsif current_user&.can_view_delegate_matters? # Staff see staff + public
+                   base_model.staff_visible
```
gregorbg, line 19: The name of the scope (aka `staff_visible`) implies that "if you're Staff in any capacity, you should be able to see this incident". The check you're performing however is whether the current user can access _Delegate Matters_, which is distinct from being _staff_.

For example, Software Team members are staff but they cannot see Delegate matters. Same goes for WSOT or WMCT for example.

So either you change the check to something that relates to the user being _staff_, or you should rename the visibility scope for the incidents to something like `delegate_visible` (name tentative, feel free to come up with something better)

app/controllers/incidents_controller.rb
```diff
@@ -36,6 +38,18 @@ def index
...
+    end
+
+    if @incident.visibility == 'staff' && !current_user&.staff?
+      redirect_to_root_unless_user(:staff?)
+      nil
+    end
+
     unless @incident.resolved?
       redirect_to_root_unless_user(:can_manage_incidents?)
     end
```
gregorbg, line 55: This existing check should probably be "collapsed" into the two new if-checks -- that is, the visibility flag should be the indicator for access now.

(I'm assuming that once an incident is resolved, its visibility is set to public accordingly)

app/controllers/incidents_controller.rb
```diff
@@ -88,12 +102,6 @@ def mark_as
 
   def update
     set_incident
-    if @incident.update(incident_params)
-      flash[:success] = "Incident was successfully updated."
-      redirect_to @incident
-    else
-      render :edit
-    end
```
gregorbg, line 96: Why are you removing this logic? This would make it so that incidents don't get persisted to the database :O

app/controllers/incidents_controller.rb
```diff
@@ -113,14 +121,17 @@ def set_incident
...
         :public_summary,
         :tags,
         :digest_worthy,
-        incident_competitions_attributes: [:id, :competition_id, :comments, :_destroy],
-      )
+        :visibility,
+        { incident_competitions_attributes: [:id, :competition_id, :comments, :_destroy] },
+      ]
+
+      params.require(:incident).permit(permitted_params)
```
gregorbg, line 135: Why are you refactoring this into a separate new line, instead of "just" adding `:visibility` to the existing parameter list?

app/models/incident.rb
```diff
@@ -7,7 +7,17 @@ class Incident < ApplicationRecord
 
   accepts_nested_attributes_for :incident_competitions, allow_destroy: true
 
-  scope :resolved, -> { where.not(resolved_at: nil) }
+  attribute :visibility, :integer, default: :visible
```
gregorbg, line 10: I don't think this is necessary, why did you add it here?

app/models/incident.rb
```diff
@@ -34,8 +44,8 @@ def digest_sent_at_consistent
     if digest_sent_at && !digest_worthy
       errors.add(:digest_sent_at, "can't be set if digest_worthy is false.")
     end
-    if digest_sent_at && !resolved_at
-      errors.add(:digest_sent_at, "can't be set if incident is not resolved.")
+    if digest_sent_at && !visible?
```
gregorbg, line 47: This change resonates with my other remark about "merging" the ifs in the controller permissions check

app/models/incident.rb
```diff
@@ -7,7 +7,17 @@ class Incident < ApplicationRecord
...
   accepts_nested_attributes_for :incident_competitions, allow_destroy: true
 
-  scope :resolved, -> { where.not(resolved_at: nil) }
+  attribute :visibility, :integer, default: :visible
+
+  enum visibility: {
+    draft: 0,          # Only visible to WRC members
+    staff: 1,          # Visible to WCA staff members only
+    visible: 2,        # Visible to everyone
+  }
```
gregorbg, line 16: If the `enum`ized column is an integer, then you can define the enum values as an array
```suggestion
  enum visibility: [
    :draft,          # Only visible to WRC members
    :staff,          # Visible to WCA staff members only
    :visible,        # Visible to everyone
  ]
```

app/models/incident.rb
```diff
@@ -7,7 +7,17 @@ class Incident < ApplicationRecord
 
   accepts_nested_attributes_for :incident_competitions, allow_destroy: true
 
-  scope :resolved, -> { where.not(resolved_at: nil) }
+  attribute :visibility, :integer, default: :visible
+
+  enum visibility: {
+    draft: 0,          # Only visible to WRC members
+    staff: 1,          # Visible to WCA staff members only
+    visible: 2,        # Visible to everyone
```
gregorbg, line 15: "visibility level visible" isn't a very amazing name. What about something like `public` instead?

app/models/incident.rb
```diff
@@ -7,7 +7,17 @@ class Incident < ApplicationRecord
...
+  }
+
+  scope :publicly_visible, -> { where(visibility: :visible) }
+  scope :staff_visible, -> { where(visibility: [:staff, :visible]) }
+  scope :wrc_visible, -> { all }
```
gregorbg, line 20: Rails `enum` gives you scopes by default that you can reuse here. Read more at https://api.rubyonrails.org/v7.1/classes/ActiveRecord/Enum.html

app/models/incident.rb
```diff
@@ -95,4 +105,19 @@ def serializable_hash(options = nil)
...
+    when "staff"
+      user&.staff? || user&.can_manage_incidents? # Staff and WRC can see staff incidents
+    when "draft"
+      user&.can_manage_incidents? # Only WRC can see drafts
+    end
+  end
+
+  scope :publicly_visible, -> { where(visibility: :visible) } # Public sees only visible
+  scope :staff_visible, -> { where(visibility: [:staff, :visible]) } # Staff sees staff + visible
+  scope :wrc_visible, -> { all } # WRC sees everything
```
gregorbg, line 122: This is (accidental?) copy-pasta from above

app/models/incident.rb
```diff
@@ -95,4 +105,19 @@ def serializable_hash(options = nil)
...
+  def visible_to?(user)
+    case visibility
+    when "visible"
+      true # Everyone can see public incidents
+    when "staff"
+      user&.staff? || user&.can_manage_incidents? # Staff and WRC can see staff incidents
+    when "draft"
+      user&.can_manage_incidents? # Only WRC can see drafts
+    end
+  end
```
gregorbg, line 118: Interestingly, the checks you're doing here are different from the controller (see my other, longer comment about `can_view_delegate_matters`) but I feel like they _should_ be the same.

Perhaps a common check method can be reused both here and in the controller?

app/views/incidents/_form.html.erb
```diff
@@ -4,6 +4,19 @@
...
     <%= f.input :title %>
+    <% if current_user.can_manage_incidents? %>
+      <div class="form-group" style="margin-top: 10px;">
+        <label class="col-sm-2 control-label">Visibility</label>
+        <div class="col-sm-10">
+          <%= f.select :visibility, 
+              Incident.visibilities.map { |k, _| [k.titleize, k] },
+              { include_blank: false },
+              { class: 'ui dropdown',
+                style: 'min-width: 80px; max-height: 40px;' } %>
```
gregorbg, line 15: I feel like `simple_form` should be able to detect enums by itself? Feel free to consult the documentation: https://github.com/heartcombo/simple_form or ChatGPT (about simple_form specifically)

app/views/incidents/show.html.erb
```diff
@@ -12,6 +12,20 @@
...
+  <% if current_user&.can_manage_incidents? %>
+    <div style="margin-top: 10px; display: flex; align-items: center;">
+      <strong style="margin-right: 10px;">Visibility:</strong>
+      <%= form_for @incident, html: { style: 'display: inline-flex; align-items: center;' } do |f| %>
+        <%= f.select :visibility, 
+            Incident.visibilities.map { |k, _| [k.titleize, k] },
+            { selected: @incident.visibility },
+            { class: 'ui dropdown',
+              style: 'min-width: 80px; max-height: 40px;',
+              onchange: 'this.form.submit()' } %>
```
gregorbg, line 24: Why is this `onchange` listener necessary? At first glance, it doesn't feel correct to say "This form should be submitted every time a new dropdown value is selected"

gregorbg (changes requested):
Thank you for your contribution! This is quite a complex issue leaning on many different topics (Enums, permissions, frontend layout) so I left quite a few comments. Feel free to ask back if you're not sure how to proceed!

PR #10375, opened 2024-12-06
Edit person - DOB validation warning in ticket workbench

danieljames-dj:


danieljames-dj:
Currently only DOB validations are done. This is almost a POC, and would like to do the same thing for name as well. The reason why I made the change in PersonValidators is because I don't want to copy paste the existing logic, instead re-use the logic somehow. Suggestions are welcome.

lib/results_validators/persons_validator.rb
```diff
@@ -28,6 +28,15 @@ class PersonsValidator < GenericValidator
...
     SINGLE_NAME_WARNING = "'%{name}' has only one name. Please confirm that this is indeed the competitor's full name according to an official document."
+    UNKNOWN_DOB_WARNING = "Some unknown DOB warning is thrown, please check with WST."
+
+    DOB_VALIDATION_ISSUES = [
+      :dob_jan_one,
+      :dob_very_young,
+      :dob_not_so_young,
+    ].freeze
+
+    DOB_VALIDATION_ISSUES_MAP = DOB_VALIDATION_ISSUES.to_h { |issue| [issue, issue] }
```
gregorbg, line 39: Why do you force the values into a hash with themselves?

lib/results_validators/persons_validator.rb
```diff
@@ -74,27 +104,33 @@ def run_validation(validator_data)
...
+                                                 name: p.name)
+            when DOB_VALIDATION_ISSUES_MAP[:dob_not_so_young]
+              @warnings << ValidationWarning.new(:persons, competition.id,
+                                                 NOT_SO_YOUNG_PERSON_WARNING,
+                                                 name: p.name)
+            else
+              @warnings << ValidationWarning.new(:persons, competition.id,
+                                                 UNKNOWN_DOB_WARNING,
+                                                 name: p.name)
+            end
```
gregorbg, line 131: Hm, I'm not a big fan of this "outsourcing" of certain validations.

In general, if some validations are so important that they are being treated differently from the rest (in this case, `dob` validations are being treated very differently / "more special" from the other `Persons` validators) they should ideally have their own validator.
But I understand that this is a deep-reaching change especially regarding the way that WRT works, so I won't force you to go down that path.

But what I do want to avoid is your current mapping. My suggestion: Find a backwards-compatible way to allow passing a more abstract "type" flag into `ValidationWarning` and `ValidationError`. (where "backwards-compatible way" in this case means "way that avoids having to touch _all_ instances of ValidationWarning/ValidationError across _all_ validators in one PR, and just add it optionally to those warnings/errors you care about").
You can then (a) check for the presence (not-null) of the new `type` property, and (b) use that to communicate the error contents to the frontend.

Basically, don't change the core logic of the validators at all, "only" add the additional parameter to `ValidationError` / `ValidationWarning` that allows you to identify the type of warning/error.

gregorbg (changes requested):

lib/results_validators/persons_validator.rb
danieljames-dj: To access them with ease - like `DOB_VALIDATION_ISSUES_MAP[:dob_jan_one]`. But anyway I've removed these lines, so most probably won't be relevant anymore.

lib/results_validators/persons_validator.rb
danieljames-dj: Okay, I've done as you suggested. But adding a new field to `ValidationIssue` is technically not possible as far as I understood, because the initialization is as follows:

```
def initialize(kind, competition_id, message, **message_args)
```

I definitely need one more field `type` to understand what the error is in the client side. For now I'm living without the `type` field, but it is limiting the features a lot for me. I'll discuss with you over slack and check what can be done for `type` and implement in another PR.

app/webpacker/components/Tickets/TicketWorkbenches/EditPersonTicketWorkbench.jsx
```diff
@@ -1,8 +1,24 @@
...
+  const { ticket } = ticketDetails;
+  const {
+    data: validators, loading, error,
+  } = useLoadedData(actionUrls.tickets.editPersonValidators(ticket.id));
+
+  if (loading) return <Loading />;
+  if (error) return <Errored />;
+
+  return validators.dob.map((validator) => (
+    <Message warning>{validator.message}</Message>));
```
gregorbg, line 20: Nit: There should be a line-break before the last `)`

lib/results_validators/persons_validator.rb
```diff
@@ -74,26 +95,18 @@ def run_validation(validator_data)
...
-            @warnings << ValidationWarning.new(:persons, competition.id,
-                                               DOB_0101_WARNING,
-                                               name: p.name)
-          end
           if p.gender.blank?
             @warnings << ValidationWarning.new(:persons, competition.id,
                                                EMPTY_GENDER_WARNING,
                                                name: p.name)
           end
-          # Competitor less than 3 years old are extremely rare, so we'd better check these birthdate are correct.
```
gregorbg, line 87: The information contained in this comment is lost.

gregorbg: (I think you should copy it to the new code snippet)

gregorbg (approved):
Works for the time being.

I think before you start hooking other validations into your ticketing system, we should take a look at why `competition_id` is a required parameter for the validation issues anyways, and whether there are more clever ways of handling this information. (For example, by looping over competitions, run validations for each competition, and then storing the results in a `{competition_id: [validation_results] }` style hash)

app/webpacker/components/Tickets/TicketWorkbenches/EditPersonTicketWorkbench.jsx
danieljames-dj: Oops, thanks for pointing. Fixed.

lib/results_validators/persons_validator.rb
danieljames-dj: added to the new code, thank you.

PR #10379, opened 2024-12-07
Prevent cancelled user from registering if registration is closed

dunkOnIT:
Tim has been struggling with this issue when they close World's reg, and I realized that the problem is that users with cancelled registrations shouldn't be able to register after reg is closed - but our system views them as having a registration, so we still render the Register panel and accept the "update".


app/views/registrations/register.html.erb
```diff
@@ -32,7 +32,7 @@
     <% end %>
   <% else %>
     <% user_may_register = current_user.cannot_register_for_competition_reasons(@competition).length == 0 %>
-    <% if @competition.should_render_register_v2?(current_user) %>
+    <% if @competition.should_render_register_v2?(current_user) && !@registration.competing_status_cancelled? %>
```
FinnIckler, line 35: this doesn't render the registration panel at all now when a registration is cancelled, which is not correct

dunkOnIT: Ah of course - didn't realise the check was happening in the React component. I think I've moved it to the correct place now?

FinnIckler:
Nope. This will still not show the registration component when a registration is cancelled, but the registration period is still open. The correct solution is to migrate the Registration Closed message to react and then check if the registration is closed and the user is not registered or cancelled and then render that. I can do that on Monday if that is fine

FinnIckler:
If you want to do it this weekend we can just merge only the backend and just show the cancelled people the error until monday

dunkOnIT:
Ah yep of course, very silly. No urgency on this as I think they're all sorted for Worlds now, but removed the frontend change so you can either add it to this PR/raise it in a new one

db/schema.rb
gregorbg, line 1: Whoopsie from your other PR?

gregorbg (changes requested):

db/schema.rb
dunkOnIT: Ah wait I know exactly how it happened, I was keeping schema.rb uncommitted in my local repo and ran `git add .` after running rubocop

dunkOnIT: Fixed

gregorbg (approved):

PR #10381, opened 2024-12-07
Add basic auto-accept for paid registrations

dunkOnIT:
Lots to do here, will probably just handle basic pending -> accepted case in this PR.

## Implementation Thoughts

- I'm adding an `auto_accept` method on the `registration` model. I think this should ultimately be where our auto_accept logic lives, and that is what will get called even if we have multiple entry points for auto_accepting. For example:
  - creation of a RegistrationPayment has an after_create hook that calls `registration.auto_accept`
  - if we want to do a BulkAutoAccept job, it can just iterate through the pending registrations in order of payment and call `auto_accept` on each of them

## This PR 
- [x] Add `auto-accept` and `auto-accept-cutoff` settings to competition table
- [x] Make them available in the competition form
- [x] Add tests for auto-accept functionality
- [x] Add tests for auto-accept cutoff
- [ ] Hide it behind a feature flag? Or just give organizers a way to turn it off

## Next PR 
- [ ] Give organizers a way to toggle auto-accept on/off

## Later PR's 
- [ ] Auto-accept users onto waiting list when Accepted is full
  - [ ] Should they be accepted onto the waiting list if the auto-accept-cutff has been reached?
  - [ ] If so, then we should also implement a warning when an organizer tries to approve a user from the pending list while there are users in the waiting list
- [ ] Auto-accept users from waiting list if a spot becomes available in the accepted list (ie, if an attendee's status changes away from `accepted`

## Questions:
- Should we default to 0 or NULL if no auto_accept_threshold is set
- Think carefully about naming of db fields


dunkOnIT:
Tests:
- [ ] logic of when `registration.auto_accept` should/shouldn't allow the user to be auto-accepted
  - [ ] before registration_open
  - [ ] after registration_close
  - [ ] registration is not paid for (and all the edge cases here - donation, refunds, repayment, etc)
  - [ ] competitor limit is reached / already registered for series / newcomer limit
  - [ ] don't auto-accept from non-pending state
  - [ ] respect the enabled/disabled toggle
  - [ ] respect the disable auto-accept threshold (whether or not auto-accept is acutally disabled)
- [ ] check that auto_accept runs after successful payment
- [ ] check that auto_accept doesn't run after failed payment
- [ ] when auto-accept threshold is reached, auto_accept is toggled off
  - [ ] auto_accept is ONLY toggled off when the threshold is reached - not if registrations are over the threshold?

dunkOnIT:
Closed in favour of 10415

PR #10400, opened 2024-12-09
Remove Microservice registrations and registration Version

FinnIckler:


db/schema.rb
```diff
@@ -1023,13 +1011,11 @@
...
     t.integer "guests", default: 0, null: false
-    t.datetime "accepted_at", precision: nil
-    t.integer "accepted_by"
-    t.datetime "deleted_at", precision: nil
-    t.integer "deleted_by"
     t.text "roles"
     t.boolean "is_competing", default: true
     t.text "administrative_notes"
+    t.datetime "rejected_at"
+    t.datetime "waitlisted_at"
```
gregorbg, line 1018: Where did these two come from?

FinnIckler: good question haha

gregorbg:
Converting to draft until we figure out a "clean" migration strategy that doesn't loose the information encapsulated in those columns

PR #10402, opened 2024-12-09
Migrate Users List to React

FinnIckler:
Basically the same as the persons page with different columns

image

### DIFF
--- a/app/assets/javascripts/users.js
+++ b/app/assets/javascripts/users.js
@@ -1,44 +1,44 @@
-onPage('users#edit, users#update', () => {
+onPage('users#edit, users#update', function() {
   // Hide/show avatar picker based on if the user is trying to to remove
   // the current avatar.
-  $('input#user_remove_avatar').on('change', (e) => {
-    const toDelete = e.currentTarget.checked;
+  $('input#user_remove_avatar').on("change", function(e) {
+    var toDelete = e.currentTarget.checked;
     $('.form-group.user_avatar').toggle(!toDelete);
-  }).trigger('change');
+  }).trigger("change");
 
-  const $approve_wca_id = $('#approve-wca-id');
-  const $unconfirmed_wca_id = $('#user_unconfirmed_wca_id');
-  const $unconfirmed_wca_id_profile_link = $('a#unconfirmed-wca-id-profile');
-  $approve_wca_id.on('click', (e) => {
-    $('#user_wca_id').val($unconfirmed_wca_id.val());
+  var $approve_wca_id = $('#approve-wca-id');
+  var $unconfirmed_wca_id = $("#user_unconfirmed_wca_id");
+  var $unconfirmed_wca_id_profile_link = $("a#unconfirmed-wca-id-profile");
+  $approve_wca_id.on("click", function(e) {
+    $("#user_wca_id").val($unconfirmed_wca_id.val());
     $unconfirmed_wca_id.val('');
     $unconfirmed_wca_id.trigger('input');
   });
-  $unconfirmed_wca_id.on('input', (e) => {
-    const unconfirmed_wca_id = $unconfirmed_wca_id.val();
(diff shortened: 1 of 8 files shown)


app/webpacker/components/Users/List/index.jsx
```diff
@@ -0,0 +1,93 @@
...
+    </WCAQueryClientProvider>
+  );
+}
+
+function PersonList() {
+  const [query, setQuery] = useState('');
+  const [page, setPage] = useState(1);
+  const [region, setRegion] = useState('all');
+
+  const debouncedSearch = useDebounce(query, 1000);
```
gregorbg, line 28: We have a `WCA_DEFAULT_DEBOUNCE` or similar constant defined somewhere :tm: 
If you look for all usages of `useDebounce` (there really aren't that many) you'll quickly find it

app/webpacker/components/Users/api/getUsers.js
```diff
@@ -0,0 +1,16 @@
...
+import { usersUrl } from '../../../lib/requests/routes.js.erb';
+// eslint-disable-next-line import/prefer-default-export
+export async function getPersons(page, region, query) {
+  const { data } = await fetchJsonOrError(
+    `${usersUrl}?search=${query}&sort=name&order=asc&offset=${(page - 1) * 10}&limit=10&region=${region ?? 'all'}`,
+    {
+      headers: {
+        Accept: 'application/json, text/javascript, */*; q=0.01',
+        'Content-Type': 'application/json',
+        'X-Requested-With': 'XMLHttpRequest',
```
gregorbg, line 11: :(

FinnIckler: well I tried it without it and that doesn't work

FinnIckler: even though this does respond.to json

app/webpacker/components/Users/List/index.jsx
FinnIckler: We actually do not. Everyone seems to be defining their own debounce time
image

FinnIckler: we have window.wca.TEXT_INPUT_DEBOUNCE_MS for the old js code which is 250ms which I think is to low for this usecase

app/assets/javascripts/users.js
```diff
@@ -1,104 +1,45 @@
...
 
   // Show avatar removal confirmation form
-  $(document).ready(function(){
-    $('.remove-avatar').click(function(){
+  $(document).ready(() => {
+    $('.remove-avatar').click(() => {
       $('.remove-avatar-confirm').show();
     });
   });
 });
```
gregorbg, line 45: A bit nitpicky, but can you please revert this part of the file changes?
The only reason I'm asking is because I don't want the git blame to become "polluted" on this ancient part of the code base. It's important to be able to trace back changes and find out who wrote this and why, precisely _because_ it's ancient and arcane.

gregorbg (changes requested):

app/webpacker/components/Users/List/index.jsx
gregorbg: Oh wow, that's funny :joy: Then maybe go with either 300 or 600 based on what we already have?

app/assets/javascripts/users.js
FinnIckler: I was hoping I will kill the rest of this file this week

app/webpacker/components/Users/List/index.jsx
FinnIckler: 600 feels decent

app/assets/javascripts/users.js
FinnIckler: I will add all the files in assets/javascript to the eslint ignore so this doesn't happen again

gregorbg (approved):

PR #10408, opened 2024-12-09
Migrate Post Form to React

FinnIckler:
Only a little jQuery deleted sadly. 

image

image

TODO: That date picker how long something should be stickied for


FinnIckler:
Added the Datepicker!

app/webpacker/components/Posts/PostForm.jsx
```diff
@@ -0,0 +1,141 @@
...
+export default function PostForm({
+  header, allTags, post,
+}) {
+  const [formPost, setFormPost] = useState(post);
+  const [formTitle, setFormTitle] = useInputState(formPost?.title ?? '');
+  const [formBody, setFormBody] = useInputState(formPost?.body ?? '');
+  const [formTags, setFormTags] = useState(formPost?.tags_array ?? []);
+  const [formIsStickied, setFormIsStickied] = useCheckboxState(formPost?.sticky ?? false);
+  const [formShowOnHomePage, setFormShowOnHomePage] = useCheckboxState(formPost?.show_on_homepage ?? true);
+  const [unstickAt, setUnstickAt] = useState(formPost?.unstick_at ?? null);
```
kr-matthews, line 24: Feels weird to have states for both the (form)post, and most of the properties of the post. Besides initializing the other states (which happens initially, so doesn't need `formPost` to be a state), `formPost` is only used to check whether we're creating or editing, get the id (if we're editing), and show the url at the end.

It seems like we could get rid of `formPost` and instead just have 2 more states for the post id and url; then update those 2 states in the `onSuccess` functions below.

app/webpacker/components/Posts/PostForm.jsx
```diff
@@ -0,0 +1,141 @@
...
+          <a href={formPost.url}>here</a>
+        </Message>
+        )}
+        { postUpdated && (
+          <Message positive>
+            Post successfully updated. View it
+            {' '}
+            <a href={formPost.url}>here</a>
+          </Message>
+        )}
```
kr-matthews, line 136: If you create a post then edit it, won't both messages be visible?

Side note: Formatting (indentation and spacing) seems to be off here, and I've noticed it in other PRs too.

kr-matthews (changes requested):

app/webpacker/components/Posts/PostForm.jsx
FinnIckler: yeah makes sense

app/webpacker/components/Posts/PostForm.jsx
```diff
@@ -0,0 +1,143 @@
...
+            </Message.List>
+          </Message>
+        )}
+        { postCreated && !postUpdated && (
+        <Message positive>
+          Post successfully created. View it
+          {' '}
+          <a href={postURL}>here</a>
+        </Message>
+        )}
```
kr-matthews, line 131: As an example, I'm pretty sure this is standard linting, not sure why your linter wouldn't be doing it.
```suggestion
        {postCreated && !postUpdated && (
          <Message positive>
            Post successfully created. View it
            {' '}
            <a href={postURL}>here</a>
          </Message>
        )}
```

app/webpacker/components/Posts/PostForm.jsx
```diff
@@ -0,0 +1,143 @@
...
+        { postUpdated && (
+          <Message positive>
+            Post successfully updated. View it
+            {' '}
+            <a href={postURL}>here</a>
```
kr-matthews, line 136:
```suggestion
            <a href={postURL}>here</a>.
```
and same above on line 129.

app/webpacker/components/Posts/PostForm.jsx
FinnIckler: mh.. my linter doesn't revert it, but maybe we just don't have that in the eslintrc? we are stuck on a quite old version of eslint

gregorbg (approved):

gregorbg:
@kr-matthews Is it fine to dismiss your review on this one or do you want to actually re-review?

kr-matthews (approved):

PR #10410, opened 2024-12-10
Do not attempt to process duplicate registrations

dunkOnIT:
Changes as discussed in meeting. One other thing I did was change the query to check if a registration exists from a `find_by().present?` to a `where().exists?` - ChatGPT suggests this is better as it doesn't load the data into memory, but you two are probably better able to adjudicate that than me.


app/jobs/add_registration_job.rb
```diff
@@ -17,9 +17,10 @@ def self.prepare_task(user_id, competition_id)
   end
 
   def perform(lane_name, competition_id, user_id, lane_params)
-    # Hotfix for mass stuck duplicate registrations in the queue
-    registration = Registration.find_by(competition_id: competition_id, user_id: user_id)
-    return if registration.present?
+    if Registration.where(competition_id: competition_id, user_id: user_id).exists?
```
FinnIckler, line 20: Isn't this just Registration.exists?(competition_id: competition_id, user_id: user_id)
At least it looks like it from this https://apidock.com/rails/ActiveRecord/Base/exists%3F/class

dunkOnIT: Ah good call! Changed

gregorbg (approved):

PR #10411, opened 2024-12-10
Move Registration Error Messages to React

FinnIckler:
I choose to do multiple ifs because it was getting complicated.
Also: is it time to only render the registration requirements in the Panel, or do we want to continue having them on top?


app/views/competitions/_nav.html.erb
```diff
@@ -267,11 +267,12 @@
         <% end %>
       <% end %>
 
-      <% @competition.warnings_for(current_user).each do |field, message| %>
-        <%= alert :warning, message, note: true %>
+      <% unless defined?(@skip_warnings )%>
```
gregorbg, line 270: Why are you wrapping this in a `defined?` call? The variable `@skip_warnings` itself is a boolean, so `unless @skip_warnings` should do.

FinnIckler: I think that caused the tests to fail? `unless @skip_warnings` was my original try, let me try again

FinnIckler: Interestingly you can't do that because `@undefined_variable == false` -> false

dunkOnIT:
Re the discussion in #10605 - what's the status on this PR / can I pick it up to help get it over the line?

gregorbg:
> Re the discussion in #10605 - what's the status on this PR / can I pick it up to help get it over the line?

Talked to Finn, this can basically go ahead. In terms of review, we need to make super duper extra sure that all warnings are represented properly.

The main issue is that the warnings were previously generated in the backend in a very convoluted way -- i.e. there was one central "please give me all warnings" method in a **model** that glued together i18n strings. Grr.....
Unfortunately people probably have come to expect/rely on these, so we need to make triple-sure we're adequately covering all edge cases and stuff.

dunkOnIT (approved):
Tested with the following cases - had to fix a couple of minor issues with cannotRegisterReasons display

image

dunkOnIT:
> Does the last item in your testing matrix indicate that you waited for the countdown to pass and then it correctly "flipped" to show the registration panel?

Correct! Those test cases weren't actually explicitly recorded - but I did wait for the countdowns both for opening and closing, and the panel "flipped" as expected in both cases

PR #10413, opened 2024-12-11
Move Caching Registrations in V2 from HTML to API Level

FinnIckler:


app/controllers/api/v1/registrations/registrations_controller.rb
```diff
@@ -89,8 +89,16 @@ def validate_bulk_update_request
...
   def list
     competition_id = list_params
-    registrations = Registration.accepted.where(competition_id: competition_id)
-    render json: registrations.includes(:user).map { |r| r.to_v2_json }
+    competition = Competition.find(competition_id)
+    registrations = competition.registrations.accepted
+    payload = Rails.cache.fetch([
+                                  "registrations_v2_list",
+                                  competition.event_ids,
+                                  registrations.joins(:user).order(:id).pluck(:id, :updated_at, :"users.updated_at"),
```
gregorbg, line 97: Pro tip for cross-table access in Rails:

```suggestion
                                  registrations.joins(:user).order(:id).pluck(:id, :updated_at, user: [:updated_at]),
```

gregorbg: That way, you're using the same symbol in the `pluck` and in the `join`, which also coincides with the name of the `belongs_to` relation, making future refactors potentially much easier

gregorbg (approved):

PR #10415, opened 2024-12-11
V3/auto accept

dunkOnIT:
This replaces #10381 - I needed to make thewca the base repo for the first PR so that the diff on the 2nd PR wouldn't be absurdly long like the last time I branched off an in-review PR.

-------

Lots to do here, will probably just handle basic pending -> accepted case in this PR.

## Implementation Thoughts

- I'm adding an `auto_accept` method on the `registration` model. I think this should ultimately be where our auto_accept logic lives, and that is what will get called even if we have multiple entry points for auto_accepting. For example:
  - creation of a RegistrationPayment has an after_create hook that calls `registration.auto_accept`
  - if we want to do a BulkAutoAccept job, it can just iterate through the pending registrations in order of payment and call `auto_accept` on each of them

## This PR 
- [x] Add `auto-accept` and `auto-accept-cutoff` settings to competition table
- [x] Make them available in the competition form
- [x] Add tests for auto-accept functionality
- [x] Add tests for auto-accept cutoff
- [ ] Hide it behind a feature flag? Or just give organizers a way to turn it off

## Next PR 
- [ ] Give organizers a way to toggle auto-accept on/off

## Later PR's 
- [x] Auto-accept users onto waiting list when Accepted is full
  - [ ] Should they be accepted onto the waiting list if the auto-accept-cutff has been reached?
  - [ ] If so, then we should also implement a warning when an organizer tries to approve a user from the pending list while there are users in the waiting list
- [x] Auto-accept users from waiting list if a spot becomes available in the accepted list (ie, if an attendee's status changes away from `accepted`

## Questions:
- Should we default to 0 or NULL if no auto_accept_threshold is set
- Think carefully about naming of db fields


FinnIckler:
You could have still branched off your PR but squashed your previous commits using rebase (which you can then drop after merge)

PR #10416, opened 2024-12-11
Auto Accept PR 2: Frontend & Waiting List Integration

dunkOnIT:
## Frontend Changes 
- [x] Give organizers a way to toggle auto-accept off
- [x] Add disable auto accept button for organizers when auto accept is enabled
- [x] Add enable/disable auto accept settings in the competition form
- [x] Make the "Disable Auto Accept" button actually do something when it is clicked, to indicate whether or not auto accept was successfully disabled

## Backend Changes
- [x] Auto-accept users onto waiting list when Accepted is full
  - [x] ~Should they be accepted onto the waiting list if the auto-accept-cutff has been reached?~ No - consulted people on discord
- [x] Auto-accept users from waiting list if a spot becomes available in the accepted list (ie, if an attendee's status changes away from `accepted`
- [x] Add bulk auto accept class method
- [x] Add bulk auto accept route
- [ ] Test if registration history and state change emails get created/sent as expected

## Future PR 
- [ ] Move waiting list to above the accepted list if it has at least one registration || competitor_limit has been reached | auto_accept is enabled and auto_accept_disable_threshold has been reached
- [ ] Warn an organizer if they're trying to accept a user from the Pending list while there are registrations on the Waiting List
- [ ] Give organizers a way to toggle auto-accept on


dunkOnIT:
Waiting list: 
- Accept a user from the waiting list if a spot becomes available on the Accepted list
- [ ] Model Tests
  - [ ] allow auto-accepting the first position on the waiting list
- [ ] Integration Tests
  - [ ] Trigger an auto_accept job from the competing lane if a user's status changes away from `accepted`

dunkOnIT:
## Bulk auto accept class method tests:
- [ ] Accepts competitors from the waiting list
  - [ ] in order of waiting list
- [ ] Accepts pending competitors
  - [ ] in order of payment
- [ ] Accepts up to but not exceeding the competition limit
  - [ ]  those not accepted follow the expected order (later on waiting list = not accepted, later payment = not accepted)

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationList.jsx
```diff
@@ -128,6 +130,8 @@ export default function RegistrationAdministrationList({ competitionInfo }) {
 
   const actionsRef = useRef();
 
+  const [autoAcceptEnabled, setAutoAcceptEnabled] = useState(competitionInfo.auto_accept_registrations)
```
kr-matthews, line 133: Can we use `competitionInfo` as the source of truth?

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationList.jsx
```diff
@@ -160,6 +164,22 @@ export default function RegistrationAdministrationList({ competitionInfo }) {
     },
   });
 
+
+  const { mutate: disableAutoAcceptMutation, isPending: isUpdating } = useMutation({
+    mutationFn: disableAutoAccept,
+    onError: (data) => {
+      const { error } = data.json;
```
kr-matthews, line 171: `error` is never used.

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationList.jsx
dunkOnIT: Currently it is being hydrated on render (not queried via API), so there's no way to refetch it currently. Assuming we could refetch it, how would you suggest we then update whether or not the "Disable auto accept" button is rendered? (my understanding is that React doesn't detect changes in the state of variables)

FinnIckler: if we would fetch it, we would rely on the react Query to handle the state being up to date.

kr-matthews: Yeah it would be pretty simple if we switch to fetching it with react query, I think.

Having a situation where there are 2 sources of truth (`competitionInfo.auto_accept_registrations` and `autoAcceptEnabled`), and which could disagree, is not good - leaves room for future bugs, etc.

dunkOnIT:
Ok I'm now combining a useQuery and prop strategy by setting `competitionInfo` as a state. Just useQuery was causing a lot of issues with competitionInfo not being loaded initially, and it seemed better to just hydrate it rather than add a bunch of conditionals on everything trying to access competitionInfo. 

Current issues: 
- [ ] "Disable Auto Accept" button not disappearing when competitionInfo reloads - the change in state is not being picked up by autoAcceptEnabled
- [ ] Getting inconsistent results with disableAutoAcceptMutation - it often triggers the onError block even though the API call is succeeding - so perhaps it isn't `await`ing the result of disableAutoAccept? I'm not quite sure what's happening there.

In any case, we're close - should finish it off by tomorrow morning. Open to input on whether there's a better way to handle `competitionInfo` with the dual hydration/fetch strategy

FinnIckler:
the issue with it not being hydrated at first is because you are not waiting for competitionInfo to be loaded before rendering. You could just fetch it in a parent component, or use competitionInfo?. access everywhere in the component where you are fetching it

package-lock.json
gregorbg, line 1: Seems like someone executed `npm install` when it should have been `yarn install`? :smiling_imp:

gregorbg (changes requested):

package-lock.json
dunkOnIT: Ok I _think_ I've fixed it? Just removed `package-lock.json`, but also had to fix a merge conflict at `yarn.lock:125` so might be worth taking a look at that as well

dunkOnIT:
@kr-matthews would you be able to help resolve the merge conflict in `RegistrationAdministrationList`? Particularly in line ~580 - it seems like how we render the lists may have changed?

kr-matthews:
> @kr-matthews would you be able to help resolve the merge conflict in `RegistrationAdministrationList`? Particularly in line ~580 - it seems like how we render the lists may have changed?

Will take a look tomorrow. The headings/lists have just been moved from the return to an array, the contents should all be the same.

kr-matthews:
> @kr-matthews would you be able to help resolve the merge conflict in `RegistrationAdministrationList`? Particularly in line ~580 - it seems like how we render the lists may have changed?

The import conflict is easy to fix: accordion, button, and icon should all be new imports.
For the other one, I think git might only be confused because you added a few new lines within the return, but the surrounding lines were moved in my work. If you remove those new lines (before trying to merge with main), the conflict will probably go away?
image

dunkOnIT:
Thanks, conflicts resolved!

app/controllers/api/v0/competitions_controller.rb
```diff
@@ -184,6 +184,14 @@ def update_wcif
     }
   end
 
+  def disable_auto_accept
+    competition = competition_from_params
+    require_can_manage!(competition)
+
+    competition.update!(auto_accept_registrations: false)
+    render json: { status: 'success' }
+  end
```
gregorbg, line 193: Why is this part of the public `v0` API? Should API users be able to disable auto-accept from the outside world? (FTR: I don't think they should, but perhaps I'm overlooking something)

app/models/competition.rb
```diff
@@ -382,6 +384,25 @@ def advancement_condition_must_be_present_for_all_non_final_rounds
     end
   end
 
+  validate :auto_accept_validations
+  private def auto_accept_validations
+    errors.add(:auto_accept_registrations, I18n.t('competitions.errors.must_use_wca_registration')) if
+      auto_accept_registrations && !use_wca_registration
```
gregorbg, line 390: I think you can use `auto_accept_registrations?` here, to make it more explcitly clear that you're checking for (non-zero) presence

app/models/competition.rb
```diff
@@ -382,6 +384,25 @@ def advancement_condition_must_be_present_for_all_non_final_rounds
...
+
+    errors.add(:auto_accept_registrations, I18n.t('competitions.errors.auto_accept_limit')) if
+      auto_accept_disable_threshold > 0 &&
+      competitor_limit.present? &&
+      auto_accept_disable_threshold >= competitor_limit
+
+    errors.add(:auto_accept_registrations, I18n.t('competitions.errors.auto_accept_not_negative')) if auto_accept_disable_threshold < 0
+
+    if auto_accept_registrations_changed?
```
gregorbg, line 399: Why do you need to check this additionally?

gregorbg: (i.e. can't you just run the validations inside this `if` globally?)

app/models/competition.rb
```diff
@@ -481,6 +502,7 @@ def confirmed_or_visible?
     self.confirmed? || self.showAtAll
   end
 
+  # TODO: Consider refactoring this once auto-accept is implemented
```
gregorbg, line 505: Refactor to what purpose?

app/models/competition.rb
```diff
@@ -2923,4 +2953,8 @@ def self.form_data_json_schema
       },
     }
   end
+
+  def auto_accept_threshold_reached?
+    auto_accept_disable_threshold > 0 && auto_accept_disable_threshold <= registrations.competing_status_accepted.count
+  end
```
gregorbg, line 2959: This method feels a bit "lonely" down here, can you find a better place for it? (Somewhere above among methods with a similar purpose perhaps)

app/models/registration.rb
```diff
@@ -336,6 +338,14 @@ def self.newcomer_month_eligible_competitors_count
     end
   end
 
+  validate :does_not_exceed_competitor_limit
+  private def does_not_exceed_competitor_limit
+    return unless competition&.competitor_limit.present?
+    return unless competing_status == Registrations::Helper::STATUS_ACCEPTED
+    errors.add(:competitor_limit, I18n.t('registrations.errors.competitor_limit_reached')) if
+      competition.registrations.competing_status_accepted.count >= competition.competitor_limit
+  end
```
gregorbg, line 347: I have an eery feeling that I asked this in other PRs, but... Shouldn't we already be checking this right now (ie before merging your PR, in the status quo production code?)

app/models/registration.rb
```diff
@@ -445,4 +455,44 @@ def ensure_waitlist_eligibility!
   def serializable_hash(options = nil)
     super(DEFAULT_SERIALIZE_OPTIONS.merge(options || {}))
   end
+
+  def auto_accept
+    return log_error('Auto-accept is not enabled for this competition.') unless competition.auto_accept_registrations
```
gregorbg, line 460:
```suggestion
    return log_error('Auto-accept is not enabled for this competition.') unless competition.auto_accept_registrations?
```

gregorbg: On a databse level, booleans are actually just `TINYINT(1)`, i.e. boolean-esque values encoded as either `0` or `1`. Using `if competition.auto_accept_registrations` relies on the fact that `0` "accidentally" behaves like `false` and `1` "accidentally" behaves like `true`.
With the question mark, Rails will properly convert it for you.

app/models/registration.rb
```diff
@@ -445,4 +455,44 @@ def ensure_waitlist_eligibility!
...
+    return log_error('Can only auto-accept pending registrations or first position on waiting list') unless
+      competing_status_pending? || (competing_status_waiting_list? && waiting_list_position == 1)
+    return log_error("Competition has reached auto_accept_disable_threshold of #{competition.auto_accept_disable_threshold} registrations") if
+      competition.auto_accept_threshold_reached?
+    return log_error('Competitor still has outstanding registration fees') if outstanding_entry_fees > 0
+    return log_error('Cant auto-accept while registration is not open') if !competition.registration_currently_open?
```
gregorbg, line 466:
```suggestion
    return log_error('Cant auto-accept while registration is not open') unless competition.registration_currently_open?
```

app/models/registration.rb
```diff
@@ -445,4 +455,44 @@ def ensure_waitlist_eligibility!
...
+        'Auto-accept',
+      )
+    else
+      update_lanes!(
+        { user_id: user_id, competing: { status: Registrations::Helper::STATUS_ACCEPTED } }.with_indifferent_access,
+        'Auto-accept',
+      )
+    end
+  rescue ActiveRecord::RecordInvalid => e
+    log_error("Auto accept error for registration #{id}: #{e}")
```
gregorbg, line 480: Nope. We don't catch "raw" errors like that. Instead, we fix what's causing the error :upside_down_face: 
(as an alternative, soft `update` instead of `update!` returns a boolean that indicates whether the save operation was successful or not)

app/models/registration.rb
```diff
@@ -445,4 +455,44 @@ def ensure_waitlist_eligibility!
...
+  end
+
+  private def log_error(error)
+    Rails.logger.error(error)
+    false
+  end
```
gregorbg, line 486: Really not a fan of this logging approach, especially the fact that you're (meaninglessly??) returning `false` at the end?!

gregorbg: What is the underlying intention here? Some sort of record of what auto-accept "thinks"?

My point is that with all other validations, and even with errors raised by the `registration_checker` itself, we never "loop" them through `Rails.logger` manually. So I'm inclined to say that there must be some other way.

app/models/registration.rb
```diff
@@ -445,4 +455,44 @@ def ensure_waitlist_eligibility!
...
+
+  private def waiting_list_auto_accept_check
```
gregorbg, line 488: Why is the registration that is currently being accepted triggering auto-accept on the first registration of the waiting list?

app/models/waiting_list.rb
```diff
@@ -6,7 +6,7 @@ class WaitingList < ActiveRecord::Base
   def remove(entry)
     update_column :entries, entries - [entry.id]
   end
-
+
```
gregorbg, line 9: Meaningless blank space, surprised that RuboCop doesn't beat you over the head with it

app/views/registrations/edit_registrations.html.erb
```diff
@@ -1,5 +1,5 @@
 <% provide(:title, I18n.t('registrations.list.title', comp: @competition.name)) %>
 
 <%= render layout: 'nav' do %>
-  <%= react_component('RegistrationsV2/RegistrationAdministration', { competitionInfo: @competition.to_competition_info }) %>
+  <%= react_component('RegistrationsV2/RegistrationAdministration', { competitionId: @competition.id }) %>
```
gregorbg, line 4: Why is this change in approach necessary? And will it perhaps be "fixed" by the Registration-Context PR?

app/webpacker/components/CompetitionForm/FormSections/Admin.js
```diff
@@ -1,16 +1,23 @@
...
   if (!isPersisted || !isAdminView) return null;
 
   return (
     <SubSection section="admin">
       <InputBoolean id="isConfirmed" />
       <InputBoolean id="isVisible" />
+      <InputBooleanSelect id="autoAcceptEnabled" required />
+      <ConditionalSection showIf={admin.autoAcceptEnabled}>
+        <InputNumber id="autoAcceptDisableThreshold" />
+      </ConditionalSection>
```
gregorbg, line 20: Why do you need to control these through the WCAT Competition Form?
It feels like it would be much easier if the organizers can control these relevant form fields through the registration admin panel directly.

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationContainer.jsx
```diff
@@ -0,0 +1,44 @@
...
+  const dispatchStore = useDispatch();
+
+  const fetchCompetitionInfo = async () => getCompetitionInfo(competitionId);
+
+  const {
+    isLoading: isCompetitionInfoLoading,
+    data: competitionInfo,
+    refetch: refetchCompetitionInfo,
+  } = useQuery({
+    queryKey: ['competitionInfo', 'test'],
```
gregorbg, line 19: `'test'` still contained in your reviewable code.
This should contain `competitionId` instead.

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationContainer.jsx
```diff
@@ -0,0 +1,44 @@
...
+    queryFn: fetchCompetitionInfo,
+    refetchOnWindowFocus: false,
+    refetchOnReconnect: false,
+    staleTime: Infinity,
+    refetchOnMount: 'always',
+    retry: false,
```
gregorbg, line 25: That's a whole bunch of options, why exactly are they necessary?
I feel like they should be provided by the default `WcaQueryClientProvider`

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationList.jsx
```diff
@@ -522,7 +540,19 @@ export default function RegistrationAdministrationList({ competitionInfo }) {
   );
 
   return (
-    <Segment loading={isMutating}>
+    <Segment loading={isMutating} style={{ overflowX: 'scroll' }}>
+      { competitionInfo.auto_accept_registrations && (
+        <Button
+          disabled={isUpdating}
+          color="red"
```
gregorbg, line 547:
```suggestion
          negative
```

app/webpacker/components/RegistrationsV2/RegistrationAdministration/index.jsx
```diff
@@ -17,7 +17,7 @@ export default function RegistrationEdit({ competitionInfo }) {
             <Sticky context={ref} offset={60}>
               <RegistrationMessage />
             </Sticky>
-            <RegistrationAdministrationList competitionInfo={competitionInfo} />
+            <RegistrationAdministrationContainer competitionInfo={competitionId} />
```
gregorbg, line 20:
```suggestion
            <RegistrationAdministrationContainer competitionId={competitionId} />
```

db/schema.rb
```diff
@@ -82,8 +82,10 @@
     t.boolean "forbid_newcomers", default: false, null: false
     t.string "forbid_newcomers_reason"
     t.integer "registration_version", default: 0, null: false
-    t.integer "competitor_can_cancel", default: 0, null: false
     t.integer "newcomer_month_reserved_spots"
+    t.boolean "auto_accept_registrations", default: false, null: false
+    t.integer "auto_accept_disable_threshold", default: 0, null: false
+    t.integer "competitor_can_cancel", default: 0, null: false
```
gregorbg, line 88: Why is this column jumping to a different position? Feels like merge-whoopsie

yarn.lock
gregorbg, line 1: Definitely unrelated changes

gregorbg (changes requested):

app/webpacker/components/RegistrationsV2/RegistrationAdministration/index.jsx
dunkOnIT: Change made during merge with main

db/schema.rb
dunkOnIT: Does the order here matter? I just removed the merge markers and committed the resulting order

yarn.lock
dunkOnIT: I'm just committing the result of running `docker compose up` - what _should_ be happening here? `yarn.lock` not showing up at all? 

It's possible this was introduced when Finn was talking me through (trying to) get `eslint` working locally

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationList.jsx
dunkOnIT: Change made

yarn.lock
gregorbg: Yes, it shouldn't show up here at all. `yan.lock` is a dependency lockfile: It lists which yarn dependencies are currently being targeted ("locked") by your project, to make builds reproducible. It also lists transitive dependencies, dependencies of transitive dependencies, etc. Basically a full dependency "tree" of whatever ecosystem the file relates to.
(We also have a lockfile called `Gemfile.lock` for Ruby, but `yarn.lock` right here is specifically for JS)

It should only ever change when adding or updating dependencies, so I guess while trying to get `eslint` working you issued some commands along the lines of `yarn update` or so.

Definitely shouldn't be part of this diff though, because you're not adding new packages or requiring a certain update

app/controllers/api/v0/competitions_controller.rb
dunkOnIT: Organizers/delegates should be able to disable auto accept - this is the endpoint that receives the call when "Disable auto accept" is clicked on the Reg Admin page. 

We could make it part of the v1 API instead - but I figured that given it's in the same namespace as the WCIF PATCH and non-public WCIF list it was in the appropriate place?

app/models/competition.rb
dunkOnIT: Changed

app/models/competition.rb
dunkOnIT: The intention here is to ensure that auto_accept can only be enabled when there are no paid-pending/waitlisted registrations while registration is not full. 

I don't think we want that to run all the time - _after_ auto_accept is enabled there may well be legitimate reasons for an organizer to want to move a paid registration from accepted to pending, and we don't want to break the competition when that happens.

app/models/competition.rb
dunkOnIT: I genuinely have no idea - comment removed, and lesson learned about including the details of my thoughts when I make comments like this

app/models/competition.rb
dunkOnIT: Moved it up to join the `registration_full?` and `registration_full_and_accepted?` methods, where it seems like the same theme sort of exists (asking questions about the state of the comp's registrations versus the competition's limits)

app/models/registration.rb
dunkOnIT: Updated

app/models/registration.rb
dunkOnIT: We are already checking this in the registration_checker - what you asked about was a refactoring of where that check happened, which made it appear like it was newly added.

In this case, auto-accepts bypass the registration checker, and so we need to have a registration-level validation to ensure that the competitor limit is respected.

app/models/registration.rb
dunkOnIT: I think we had a conversation about this at the time - although "at the time" was December so it's hazy to me. It doesn't feel great and I'm open to better solutions - or perhaps the notion that we don't actually need to store why a reg wasn't auto-accepted

The intention of this is to record the reasons why auto_accept would not trigger. This information would otherwise not be visible, as even if we raised an error here, it would not be visible to the user[1]

We want to record these reasons so that we can more easily debug any cases where auto_accept doesn't accept someone when it otherwise might appear that it should have[2]. However, there's no need or desire for this to be user-facing - as accepting isn't a user-facing process

[1] because this is a process which is at best indirectly triggered by the user - in the case that they pay for a registration - and in other cases, triggered by the actions of a different user (someone cancelling a registration prompting an auto-accept from the waiting list).

[2] organizer comes to us and says "this person wasn't auto-accepted but I think they should have been", it might be very hard/impossible for us to repro that case and figure out for what reason the system chose not to auto-accept

app/models/registration.rb
dunkOnIT: The problem is we don't necessarily know what might cause the error, and if not for catching and logging the error, we would have no visibility of that error at all. 

The call chain is: 
- update_lanes!
- Registrations::Lanes::Competing.update! [note that `update!` is NOT an ActiveRecord method - it is a custom-defined function in competing.rb - maybe it needs to be renamed to `update_lane`?]
- registration.save!

We could instead call `registration.save`, and then perhaps propagate the `registration` object up the call chain so that we can call `registration.errors.full_messages` on it to populate the log_error call - but that might be downstream of the log_error discussion

app/views/registrations/edit_registrations.html.erb
dunkOnIT: (to discuss in meeting_

app/controllers/api/v0/competitions_controller.rb
dunkOnIT: TODO

app/models/competition.rb
dunkOnIT: - [ ] only enforce this if it changed to `true`

app/models/registration.rb
dunkOnIT: - [ ] add a comment that this is a redundant check with the registration_checker, and we want to move the check out of the registration checker when possible because of Fin's PR beiongmerged

app/models/registration.rb
dunkOnIT: - [ ] Change this to `competing_status_previously_was`
- [ ] Try some `attempt_to_fill_opened_spot` to clarify what this method is trying to do

app/views/registrations/edit_registrations.html.erb
dunkOnIT: - [ ] Implement a route that returns the auto_accept_enabled state

PR #10418, opened 2024-12-11
Migrate Competition Overview Page to React

FinnIckler:
Currently wip porting the registration requirements over

There is an issue that it only does one markdown request at once, but a bunch of fields are in markdown... so not sure what to do about that.

Progress:
image


FinnIckler:
Migrated all information over. Now onto the styling
Screenshot 2024-12-12 at 15 47 54
Screenshot 2024-12-12 at 15 48 03

FinnIckler:
Moving it to the grid with the right sizes was quite easy actually. But I don't really know what the correct equivalent of a Descriptive List is in Semantic

FinnIckler:
I can rebuild it with another grid, maybe that looks good enough

FinnIckler:
Screenshot 2024-12-12 at 21 42 04
I was able to replicate it with a grid, but I did have to set the padding to 0 manually

FinnIckler:
Screenshot 2024-12-13 at 00 11 24
Screenshot 2024-12-13 at 00 11 08
Almost everything moved over to grid

app/webpacker/components/CompetitionTabs/Requirements/index.jsx
gregorbg, line 1: Lots of dangling `br`, won't comment on all of them.
Not sure what you're trying to solve, but a simple `<List>` (which does _not_ have enumeration bullet points by default in SemUI!) might already do the trick in terms of line breaks.

app/webpacker/components/CompetitionTabs/index.jsx
```diff
@@ -0,0 +1,95 @@
...
+import CompetitionTab from './CompetitionTab';
+import EventsTable from './EventsTable';
+import Schedule from './Schedule';
+import WCAQueryClientProvider from '../../lib/providers/WCAQueryClientProvider';
+import './style.css';
+import TimeLimitCutoffInfo from './TimeLimitCutoffInfo';
+
+const updatePath = (tabSlug) => {
+  window.history.replaceState({}, '', `${window.location.pathname}#${tabSlug}`);
+};
```
gregorbg, line 13: Do we have shared `lib` functions for this? I feel like one or two other parts of our frontend code use similar features already

app/webpacker/components/CompetitionTabs/GeneralInfoTab.jsx
gregorbg, line 1: Not happy about the repetitive `Grid`ding in here. Have you (a) tried out SemUI `List` using `List.Header`, `List.Description` and perhaps even `List.Icon` and (b) if a didn't work, can you please move the Grid cells into an "internal" component that you can reuse at least within this page?

gregorbg (changes requested):

app/webpacker/components/CompetitionTabs/Requirements/RegistrationFeeRequirements.jsx
```diff
@@ -0,0 +1,27 @@
...
+              ),
+          })
+          : I18n.t('competitions.competition_info.no_entry_fee')}
+      </p>
+      {competition.competition_events.map((event) => (event['has_fee?'] ? (
+        <dl key={event.id}>
+          <dt>{event.event.name}</dt>
+          <dd>{isoMoneyToHumanReadable(event.fee)}</dd>
+        </dl>
+      ) : null))}
```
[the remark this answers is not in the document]
FinnIckler: this is just not migrated yet!

app/webpacker/components/CompetitionTabs/Requirements/GuestRequirements.jsx
```diff
@@ -0,0 +1,31 @@
...
+    return (
+      <>
+        {I18n.t('competitions.competition_info.guests_pay', {
+          guests_base_fee:
+            isoMoneyToHumanReadable(
+              competition.guests_entry_fee_lowest_denomination,
+              competition.currency_code,
+            ),
+        })}
+        <br />
```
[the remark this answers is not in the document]
FinnIckler: the <br /> are from the original file

app/webpacker/components/CompetitionTabs/GeneralInfoTab.jsx
FinnIckler: It does not work with a List, but I did do a bunch of refactoring to reuse the grid components

FinnIckler:
image
Mobile view

app/webpacker/components/CompetitionTabs/Requirements/index.jsx
FinnIckler: managed to remove all those brs using a list

app/webpacker/components/CompetitionTabs/index.jsx
FinnIckler: There is a function in filterUtils.js but it uses searchParams and not Hash so it works completely different

app/webpacker/components/CompetitionTabs/Requirements/RegistrationFeeRequirements.jsx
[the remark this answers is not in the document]
FinnIckler: we don't have any way to currently set event based registration fees right? So can this not be safely removed?

app/webpacker/components/BookmarkIcon/index.jsx
```diff
@@ -0,0 +1,49 @@
...
+  );
+}
+
+function BookmarkIcon({ competitionInitiallyBookmarked, competitionId }) {
+  const [isBookmarked, setIsBookmarked] = useState(competitionInitiallyBookmarked);
+
+  const { mutate: bookmarkCompetitionMutation, isPending: isBookmarking } = useMutation({
+    mutationFn: bookmarkCompetition,
+    onSuccess: () => {
+      setIsBookmarked(!isBookmarked);
```
kr-matthews, line 25: It's safer to explicitly set it, just in case the state somehow gets out of sync. It might be possible for the user to click again while the initial mutation is processing or something, which might flip it twice? (Might be worth disabling the icon while it's loading?)
```suggestion
      setIsBookmarked(true);
```
Likewise below, but with false.

Also, could be simplified to one line:
```
onSuccess: () => setIsBookmarked(true),
```

app/webpacker/components/CompetitionTabs/Requirements/GuestRequirements.jsx
kr-matthews, line 1: This seems like a mix of an if and a ternary, where the second case of the ternary is also conditional. It might be clearer as 3 if statements.

kr-matthews (commented):
(Only viewed some of the files. Will do the rest later, if the PR is still open.)

gregorbg:
Opinions?
image

kr-matthews (changes requested):
~I haven't fully reviewed everything yet.~

I'd be in favour of having the main, exported function at the top of files.

app/webpacker/components/CompetitionTabs/WinnerTable.jsx
```diff
@@ -0,0 +1,81 @@
...
+          {results.map((r) => {
+            const attempts = [r.value1, r.value2, r.value3, r.value4, r.value5];
+            const bestResult = _.max(attempts);
+            const worstResult = _.min(attempts);
+            const bestResultIndex = attempts.findIndex((a) => a === bestResult);
+            const worstResultIndex = attempts.findIndex((a) => a === worstResult);
+            return (
+              <Table.Row>
+                <Table.Cell>
+                  <a href={competitionAllResultsUrl(competition.id, r.event.id)}>
```
kr-matthews, line 52: I'm getting an error when I try to view results for a particular event.

FinnIckler:
> ~I haven't fully reviewed everything yet.~
> 
> I'd be in favour of having the main, exported function at the top of files.

Will that not cause a 'referenced before defined' eslint error?

FinnIckler:
Moved the numbers on the right to a Statisticgroup
image

Also added a 'Registrations' statistic that is shown instead of the competitor count when the competition isn't finished yet

app/webpacker/components/CompetitionTabs/GeneralInfo/GeneralInfoTab.jsx
```diff
@@ -0,0 +1,136 @@
...
+import I18n from '../../../lib/i18n';
+import Markdown from '../../Markdown';
+import RegistrationRequirements from '../Requirements/RegistrationRequirements';
+import { getFullDateTimeString } from '../../../lib/utils/dates';
+import WinnerTable from '../WinnerTable';
+import InformationList from './InformationList';
+import DateAddressContact from './DateAddressContact';
+import InformationEvents from './InformationEvents';
+
+function RegistrationTime({ competition }) {
```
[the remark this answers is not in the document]
FinnIckler: I removed the seconds. About the GMT-8 vs. PST, I don't know why that happens for you, but https://github.com/moment/luxon/blob/master/docs/formatting.md shows that DATETIME_FULL should use the offset name and not just the offset

Issue #10419, opened 2024-12-12
Add "Setup" and "Teardown" as default options in the Schedule editor 


dunkOnIT:
Request came from this thread: "Fwd: [WCA Website] Software Comment by Braden Richards on 05 Dec 2024 at 16:28"
> This isn't a bug report or API question, but I wanted to ask how easy it would be to add new sections to the Edit Schedule UI for Setup/Teardown time periods. While it's not very time consuming to select 'Other' and then manually type out Setup or Teardown, but I feel like since it's such a common occurrence at almost every competition, reducing the amount of clicks to add setups or teardowns to comps. If this isn't easy to do, I get that, I just figured I would suggest this since it's something I've wanted for quite a while.

PR #10437, opened 2024-12-16
added setup and teardown options 

darshanvss:
Added Setup and teardown options under custom activities #10419

image


gregorbg:
I'm not quite sure whether "teardown" is the correct word to use here, as it sounds to me like somebody will move in with sledgehammers and wrecking balls and demolish the whole venue. I am not a native speaker though so happy to defer to other people's expertise.

Also, if we want to go by the book (which I believe we should) a companion PR to https://github.com/thewca/wcif/blob/master/specification.md#other-activities needs to be opened.

darshanvss:
@gregorbg we could maybe change it to 'Take down' or 'Dismantle' ,  or probably 'Pack up' ,'Clear up'.

I'll update the PR once we decide what it should be. I'll raise a companion PR in wcif as well

gregorbg:
Let's just go with teardown. Enough native English speakers convinced me that it's not as daunting as it sounds.

gregorbg (approved):

dunkOnIT:
Did anyone test this in frontend? There are no translation keys for the new menu options #10539 
image

gregorbg:
Feel free to add some ;)

PR #10451, opened 2024-12-18
Edit Person Requests page in WRT panel

danieljames-dj:
This PR creates a new page in WRT panel which lists all the tickets of type 'edit person'. The `TicketsList` component is made in a way that this can be reused for other components as well in future.

### DIFF
--- a/app/controllers/tickets_controller.rb
+++ b/app/controllers/tickets_controller.rb
@@ -20,7 +20,7 @@ class TicketsController < ApplicationController
     status = params[:status]
     tickets = tickets.select do |ticket|
       if status
-        ticket.metadata.status == status
+        ticket.metadata&.status == status
       else
         true
       end
@@ -35,6 +35,10 @@ class TicketsController < ApplicationController
     sort_param = params[:sort] || ''
     tickets = sort(tickets, sort_param, SORT_WEIGHT_LAMBDAS)
 
+    # paginate won't help in improving efficiency here because we are fetching all the tickets
+    # and then filtering and sorting them. We can't use the database to do this because the
+    # filtering and sorting is based on the metadata of the ticket.
+    # TODO: Check the feasibility of using the database to filter and sort the tickets.
     paginate json: tickets
   end
 
--- a/app/controllers/tickets_controller.rb
+++ b/app/controllers/tickets_controller.rb
@@ -1,6 +1,43 @@
 # frozen_string_literal: true
 
 class TicketsController < ApplicationController
+  include Rails::Pagination
+
+  SORT_WEIGHT_LAMBDAS = {
+    createdAt:
+      lambda { |ticket| ticket.created_at },
+  }.freeze
+
+  def index
+    tickets = Ticket
+
+    # Filter based on params
+    type = params[:type]
+    if type
+      tickets = tickets.where(metadata_type: type)
+    end
+
+    status = params[:status]
+    tickets = tickets.select do |ticket|
+      if status
+        ticket.metadata.status == status
+      else
+        true
+      end
+    end
+
+    # Filter based on current_user's permission
+    tickets = tickets.select do |ticket|
+      ticket.can_user_access?(current_user)
+    end
+
+    # Sort
+    sort_param = params[:sort] || ''
+    tickets = sort(tickets, sort_param, SORT_WEIGHT_LAMBDAS)
+
+    paginate json: tickets
+  end
+
   def show
     respond_to do |format|
       format.html do
--- a/app/models/ticket.rb
+++ b/app/models/ticket.rb
@@ -35,6 +35,10 @@ class Ticket < ApplicationRecord
     end
   end
 
+  def can_user_access?(user)
+    user_stakeholders(user).any?
+  end
+
   DEFAULT_SERIALIZE_OPTIONS = {
     include: %w[ticket_logs metadata],
   }.freeze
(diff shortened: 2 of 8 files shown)


app/models/ticket.rb
```diff
@@ -22,6 +22,10 @@ def action_allowed?(action, user)
     end
   end
 
+  def can_user_access?(user)
+    user_stakeholders(user).any?
```
gregorbg, line 26: I vaguely remember that you did a similar check in other, already-existing controller methods.
Can you go back and check whether you can re-use this new method in other places, now that you've created it?

gregorbg: Also, I want to note that this query can become very expensive very quickly. In a lot of developers' minds (who aren't familiar with the in-depth implementation for your tickets) the intuition is that "a simple boolean method that gives a Yes-No answer surely will just to some simple `==` comparison(s)". But the truth is that even in this early stage, the `user_stakeholders` already does a lot of heavy lifting, loading roles and stuff under the hood.

So _at least_ a warning comment should be warranted, and potentially we even need to think about a better optimization.

app/webpacker/components/Panel/pages/EditPersonRequestsPage/index.jsx
```diff
@@ -0,0 +1,23 @@
...
+        status={ticketStatuses.edit_person.open}
+        sort="createdAt:desc"
+      />
+      <Header>Closed Tickets</Header>
+      <TicketsList
+        type={ticketTypes.edit_person}
+        status={ticketStatuses.edit_person.closed}
+        sort="createdAt:desc"
+      />
+    </>
```
gregorbg, line 21: Can't you just loop over `ticketStatuses.edit_person` and create this copy-pasted "Header+TicketsList" thing on-the-fly?

app/controllers/tickets_controller.rb
```diff
@@ -1,6 +1,43 @@
...
+    # Filter based on params
+    type = params[:type]
+    if type
+      tickets = tickets.where(metadata_type: type)
+    end
+
+    status = params[:status]
+    tickets = tickets.select do |ticket|
+      if status
+        ticket.metadata.status == status
```
gregorbg, line 23: If `metadata` is generic, how do you even know it will always have a `status` variable?

And if you know that every metadata definitely always has a `status` variable, then why is it part of `metadata` and not part of the `Ticket` entity itself?

app/controllers/tickets_controller.rb
```diff
@@ -1,6 +1,43 @@
...
+    end
+
+    # Filter based on current_user's permission
+    tickets = tickets.select do |ticket|
+      ticket.can_user_access?(current_user)
+    end
+
+    # Sort
+    sort_param = params[:sort] || ''
+    tickets = sort(tickets, sort_param, SORT_WEIGHT_LAMBDAS)
```
gregorbg, line 36: Where is this `sort` method coming from?

app/controllers/tickets_controller.rb
```diff
@@ -1,6 +1,43 @@
...
+
+    paginate json: tickets
```
gregorbg, line 38: Paginate won't really help you if you're already loading everything from the DB into memory anyways. In this case, pagination turns into "nice cosmetic" for slicing out the part of the pre-loaded array that the frontend wants, but before it can slice, all tickets need to be loaded anyways.

This is a huge waste of resources. `paginate` can really unfold its magic when you're building a database query and then passing on that query builder directly. Because then the pagination will translate this to a `LIMIT ... OFFSET ...` query making it beautifully efficient.

gregorbg (changes requested):

app/models/ticket.rb
danieljames-dj: The PR is merged, so removed this method. (looks like I forgot to reply the previous message, it was still in draft. Sending all of them together now)

app/controllers/tickets_controller.rb
danieljames-dj: Oh yes, it's not necessary that `status` is there always theoretically. I've added `&`.

app/controllers/tickets_controller.rb
danieljames-dj: It's the same `sort_helper` we used for roles.

app/controllers/tickets_controller.rb
danieljames-dj: I understand that paginate won't help in efficiency here. The reason why I added paginate is for the same purpose why we have paginate in user_roles: in case we decide to list the tickets in multiple pages instead of single page, this will be helpful.

I understand that if I need to show 100 tickets, in backend it will fetch all the 10000 tickets and then just cut the first 100. But I'm just focussing on frontend for now. I've added a comment for the same. Hope it's fine, let me know if I can do something to make it better.

(I have a feeling that some kind of relation stuffs like I did in `can_user_access?` might help, but I didn't spend time for that yet. Also, I have a feeling that the filtering and sorting is going to be much more complicated in future like user_roles, so in future this might become infeasible)

gregorbg: Okay. I am not 100% happy with that reasoning, especially the part about "I have a feeling that the filtering and sorting is going to be much more complicated in future like user_roles, so in future this might become infeasible". If our database models are so convoluted that SQL-level filtering is impossible, we will have to redesign the database models.

But anyways, since you're integrating this feature step by step and the scope is quite limited for now, I'll approve.

gregorbg (approved):

danieljames-dj:
Thanks Gregor

PR #10452, opened 2024-12-18
Migrate Avatar Approval To React

FinnIckler:
image
image
image

I decided against a carousel because Semantic doesn't have one. I tried out `pure-react-carousel` but it's a lot of work to make it look good with semantic. I think having the buttons on the bottom is fine


FinnIckler:
Todo is figuring out what the best way to update the component after submitting is so only the deferred avatars are shown

app/webpacker/components/Panel/pages/ApprovePictures/AvatarCard.jsx
```diff
@@ -0,0 +1,170 @@
...
+            {user.name}
+          </a>
+        </CardHeader>
+        { user['staff_or_any_delegate?'] && (
+          <CardMeta>
+            <strong>Staff Member or Trainee Delegate</strong>
+            {' '}
+            - see guidelines above
+          </CardMeta>
+        )}
```
kr-matthews, line 137: Could we only show the guidelines when the user is a staff, instead of always showing them and conditionally showing this message?

app/webpacker/components/Panel/pages/ApprovePictures/AvatarCard.jsx
```diff
@@ -0,0 +1,170 @@
...
+      </ButtonGroup>
+    </Segment>
+  );
+}
+
+export default function AvatarCard({
+  user, onReject, onApprove, onDefer,
+}) {
+  const [activeAvatarIndex, setActiveAvatarIndex] = useState(0);
+  const [avatarState, setAvatarState] = useState('');
```
kr-matthews, line 57: This is essentially duplicating state, because e.g. whenever this is set to 'defer', `onDefer` is also called. So the parent knows `avatarState` and should be able to provide it.

It's not quite that simple because of the additional `chooseRejectionReason`. But that could be converted into its own (boolean) state I think, while `avatarState` is hoisted up.

app/webpacker/components/Panel/pages/ApprovePictures/index.jsx
```diff
@@ -0,0 +1,96 @@
...
+
+function ApprovePictures({ pendingUsers }) {
+  const [approvedAvatars, setApprovedAvatarIds] = useState([]);
+  const [rejectedAvatars, setRejectedAvatarIds] = useState([]);
+  const [deferredAvatars, setDeferredAvatarIds] = useState([]);
+
+  const onApprove = useCallback((avatar) => {
+    setApprovedAvatarIds((prev) => (prev.includes(avatar.id) ? prev.filter((a) => a.id !== avatar.id)
+      : [...prev, avatar]));
+  }, [setApprovedAvatarIds]);
```
kr-matthews, line 45: As far as I can see, `onApprove` can only be called if the avatar is not already approved, so the ternary condition would always be false when this callback is called?
It doesn't really make sense for something called `onApprove` to remove the argument from the approved list, to me.

app/webpacker/components/Panel/pages/ApprovePictures/index.jsx
```diff
@@ -0,0 +1,96 @@
...
+    <WCAQueryClientProvider>
+      <ApprovePictures pendingUsers={pendingUsers} />
+    </WCAQueryClientProvider>
+  );
+}
+
+function ApprovePictures({ pendingUsers }) {
+  const [approvedAvatars, setApprovedAvatarIds] = useState([]);
+  const [rejectedAvatars, setRejectedAvatarIds] = useState([]);
+  const [deferredAvatars, setDeferredAvatarIds] = useState([]);
```
kr-matthews, line 40: I assume each avatar should be in at most one of these lists? A reducer could enforce that, but it might be over-kill in this case.

Also: inconsistency in the names - `...Avatars` vs `AvatarIds`.

Also also: the third state is unused?

maxidragon:
I think it's a good idea to move this to WRT Panel now. @danieljames-dj, what do you think?

danieljames-dj:
> I think it's a good idea to move this to WRT Panel now. @danieljames-dj, what do you think?

Ya I agree. Looks like Finn is in right direction as he created the component in `Panel/pages` folder. It's not necessary that moving to WRT panel needs to be done in this PR itself, it can be done in a later PR as well because it will involve removing it from 'old admin panel' as well and having those codes too here might distract the reviewer a bit.

app/webpacker/components/Panel/pages/ApprovePictures/index.jsx
FinnIckler: I will remove all these state because we don't actually need batch processing

app/webpacker/components/Panel/pages/ApprovePictures/AvatarCard.jsx
FinnIckler: not sure what you mean

FinnIckler:
Changed the PR, so instead of batch processing the avatar request and then pressing submit, you just immediately approve/reject and avatar (after you choose a reason for reject) and the Avatar is removed from the list.

app/webpacker/components/Panel/pages/ApprovePictures/AvatarCard.jsx
kr-matthews: `- see guidelines above` is referring to the items below `<Header as="h4">Additional guidelines for Staff Members</Header>` in the index file. I'm suggesting only show that header and those items if the user is a staff/delegate, instead of always showing the guidelines and sometimes showing this warning.

FinnIckler: Well the problem with that is that there might be 10 request, but only one staff. The index page would need to determine if any of the ten people are staff to display the additional guidelines, which I think will be quite jarring of an UI update once that one staff avatar request is worked off

gregorbg (approved):

PR #10453, opened 2024-12-19
Sidekiq staging improvements

FinnIckler:
Aligns the Staging/Production Sidekiq config by giving it the same amount of RAM and its own Redis


FinnIckler:
I already ran terraform on this so it should just come in effective with the next deployment

gregorbg (approved):

FinnIckler:
Should have said you also need to deploy to staging again

PR #10456, opened 2024-12-21
Edit profile form for delegates

danieljames-dj:
This PR aims to open up the edit profile form for delegates, so that delegates can request for profile updates through form instead of sending email to WRT.


app/webpacker/components/ContactsPage/ContactForm.jsx
```diff
@@ -47,7 +50,10 @@ export default function ContactForm({
   const contactFormState = useStore();
   const dispatch = useDispatch();
   const { formValues: { contactRecipient: selectedContactRecipient, userData } } = contactFormState;
-  const formRedirection = useMemo(() => getFormRedirection(contactFormState.formValues));
+  const formRedirection = useMemo(
+    () => getFormRedirection(contactFormState.formValues),
+    [contactFormState.formValues],
+  );
```
kr-matthews, line 56: This calculation isn't expensive and the const isn't used in any dependency arrays, so I don't see any reason to memo this.

app/webpacker/components/ContactEditProfilePage/index.jsx
```diff
@@ -8,31 +8,88 @@ import Loading from '../Requests/Loading';
...
+          wcaId={wcaId}
+          onContactSuccess={() => setContactSuccess(true)}
+          recaptchaPublicKey={recaptchaPublicKey}
+        />
+      )}
+    </Container>
+  );
+}
+
+export default function ContactEditProfilePage({ loggedInUserId, recaptchaPublicKey }) {
```
kr-matthews, line 85: It's nice to have the main/exported component at the top, and the helper components below.

app/webpacker/components/ContactsPage/ContactForm.jsx
danieljames-dj: Okay removed.

app/webpacker/components/ContactEditProfilePage/index.jsx
danieljames-dj: Okay changed, I actually thought the reverse - like the exported component should be at bottom - just like how we have main method at bottom of programs.

kr-matthews: I think that's often the case because in some languages anything that a main method references has to have already been defined, so you have to put it at the bottom. Which isn't the case with React.

The point of this file is to provide (export) a component, so the most useful thing to see first is that component. You can then scroll down further to see more details if you want.

But it's not a big deal.

app/webpacker/components/ContactEditProfilePage/index.jsx
```diff
@@ -45,6 +36,39 @@ export default function ContactEditProfilePage({ loggedInUserId, recaptchaPublic
...
+    />
+  );
+}
+
+function ContactEditProfileSelf({ setContactSuccess, recaptchaPublicKey }) {
+  const { data: loggedInUserData, isLoading, isError } = useQuery({
+    queryKey: ['userData'],
+    queryFn: () => fetchJsonOrError(apiV0Urls.users.me.userDetails),
+  }, CONTACT_EDIT_PROFILE_QUERY_CLIENT);
+  const wcaId = loggedInUserData?.data?.user?.wca_id;
```
gregorbg, line 60: Am I correct that you're running this whole query just to get the current user's WCA ID?

app/webpacker/components/ContactEditProfilePage/index.jsx
```diff
@@ -45,6 +36,39 @@ export default function ContactEditProfilePage({ loggedInUserId, recaptchaPublic
...
+        recaptchaPublicKey={recaptchaPublicKey}
+      />
+    );
+  }
+  return (
+    <ContactEditProfileSelf
+      setContactSuccess={setContactSuccess}
+      recaptchaPublicKey={recaptchaPublicKey}
+    />
+  );
```
gregorbg, line 52: I'm not a big fan of this duplication approach. It should definitely be possible to have just one `ContactEditProfile` component, and give it a WCA ID search input which either happens to be pre-filled (and disabled, in case of self-edit) or enabled and open for the user to search whathever ID they want to edit.

app/webpacker/components/ContactEditProfilePage/index.jsx
```diff
@@ -56,3 +80,40 @@ export default function ContactEditProfilePage({ loggedInUserId, recaptchaPublic
...
+      <Message error>
+        <I18nHTMLTranslate i18nKey="page.contact_edit_profile.no_permission_error" />
+      </Message>
+    );
+  }
+
+  return (
+    <Container text>
+      <Header as="h2">{I18n.t('page.contact_edit_profile.title')}</Header>
+      <WcaSearch
```
gregorbg, line 102: In relation to my other comment above, we also have a `IdWcaSearch` component (in the same file that defined `WcaSearch` which is purely ID-based and may help you here, because it seems that `person` is a dummy value whose only purpose is to store a WCA ID which you later pass down to the edit component.

gregorbg (changes requested):
Please find a way to handle the frontend rendering without copy-pasta of the "self-edit or other-edit" component.

Backend is fine.

app/webpacker/components/ContactEditProfilePage/index.jsx
danieljames-dj: Okay, I merged both of them. I split them because I feel they were bit complicated when together.

app/webpacker/components/ContactEditProfilePage/index.jsx
danieljames-dj: Yes, in future maybe we can have a hook like `useLoggedInUserPermission` to get the WCA ID to make it simpler.

app/webpacker/components/ContactEditProfilePage/index.jsx
danieljames-dj: Okay, done.

app/webpacker/components/ContactEditProfilePage/index.jsx
```diff
@@ -8,21 +8,37 @@ import Loading from '../Requests/Loading';
...
+  const editOthersProfileMode = Boolean(queryParams.editOthersProfile);
   const { data: loggedInUserData, isLoading, isError } = useQuery({
     queryKey: ['userData'],
     queryFn: () => fetchJsonOrError(apiV0Urls.users.me.userDetails),
-    enabled: !!loggedInUserId,
+    enabled: (
+      // If not logged in, fetching WCA ID of logged in user is not possible.
+      !!loggedInUserId
+      // If the user needs to edit other's profile, then fetching own WCA ID is not needed.
+       || !editOthersProfileMode
```
gregorbg, line 29: It took me a while to see thrrough this. I think a better explanation of the `!` should be: If the user is *not* editing somebody else's profile, then we need to fetch their own WCA ID.

gregorbg: (The problem is that your comment explains when the fetching should be DISabled, but the `useQuery` function needs to be passed information when it is ENabled. So I was confused why your comment says "if the user needs to edit other's profile", while the code directly below says "NOT editing other's profile")

app/webpacker/components/ContactsPage/ContactForm.jsx
```diff
@@ -47,7 +50,7 @@ export default function ContactForm({
   const contactFormState = useStore();
   const dispatch = useDispatch();
   const { formValues: { contactRecipient: selectedContactRecipient, userData } } = contactFormState;
-  const formRedirection = useMemo(() => getFormRedirection(contactFormState.formValues));
+  const formRedirection = getFormRedirection(contactFormState.formValues);
```
gregorbg, line 53: Why did you get rid of the `useMemo`? You should of course pass a dependency array correctly (probably my bad for letting that slip through in the initial review of this page) but otherwise I feel like `formValues` can change relatively often so memoing the result feels like a good idea.

gregorbg (changes requested):
Cool! Just a few nits left now.

app/webpacker/components/ContactsPage/ContactForm.jsx
kr-matthews: I suggested it in my review:
> This calculation isn't expensive and the const isn't used in any dependency arrays, so I don't see any reason to memo this.

gregorbg: Ah, that's interesting! My understanding was always that `memo`ing stuff is worth it if the _inputs_ to the calculation change (semi-)frequently, not whether the _output_ of the memo is then later used as a dependency down the line. Would you care to elaborate? I'm eager to learn more! :smile:

kr-matthews: The vast majority of consts we use in components are very quick/easy to compute, so it doesn't really matter if they get recomputed frequently. In simple cases, `memo`ing them might even (technically) slow things down, since React needs to look at the dependency array and check if anything has changed and also stores the old value even when it's not needed (but the slowing down would be negligible).

As far as performance goes, the more often the inputs change, the less useful memoizing is, because you use the cached value less frequently and need to recompute anyway.

Some of the first results from a google search about when to use `useMemo`:

https://maxrozen.com/understanding-when-use-usememo
> don't use useMemo until you notice parts of your app are frustratingly slow.

https://blog.logrocket.com/when-not-to-use-usememo-react-hook/
> Most methods on JavaScript data types are optimized, e.g. Array.map, Object.getOwnPropertyNames(), etc. If you’re performing an operation that’s not expensive (think Big O notation), then you don’t need to memoize the return value. The cost of using useMemo [may outweigh](https://kentcdodds.com/blog/usememo-and-usecallback) the cost of reevaluating the function.

kr-matthews: And hopefully it's clear that `memo`ing is (potentially) helpful when a calculation is expensive; and that if you're using an array/object/function in another dependency array then it needs `memo`ing otherwise that other dependency array would trigger an update every single render.

gregorbg (approved):

PR #10457, opened 2024-12-21
Unify update_status_and_charges for generic provider support

gregorbg:
Partial implementation of #9208 

The idea is to pull out all platform-specific behavior from `update_stripe_status_and_charges`, and implement that in the CPI instead. The payment intent's `update_status_and_charges` then gets the CPI as an additional param, which it can use to call the platform-specific functionality as sort of like an "interface"

Edit: For review, I recommend the "Hide Whitespace" feature in the diff settings


dunkOnIT (approved):
LGTM and tested locally with succeeded and failed payments, plus refunds - all works as expected

PR #10458, opened 2024-12-21
Make `payment_integration` a required parameter in the payment completion route

gregorbg:
Follow-up to #10234

The optional URL parameter that was in that PR (to maintain backwards compatibility) now becomes a required parameter.
All usages throughout the code have been adjusted to pass `'stripe'` (or the functionally equivalent `:stripe`) when calling the URL helpers

### DIFF
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -378,11 +378,8 @@ class RegistrationsController < ApplicationController
     competition_id = params[:competition_id]
     competition = Competition.find(competition_id)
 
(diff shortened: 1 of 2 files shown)


dunkOnIT (approved):
Tested locally and works as expected :) Assume we want to give it a few more days before deploying, but figured it was an easy enough review while I waited for CAD on staging

gregorbg:
Let's wait for another day or two now that we have #10477

dunkOnIT:
Should we merge+deploy now?

gregorbg:
Yay let's go!

PR #10459, opened 2024-12-22
Clean up avatar inclusion in new competition overview backend

gregorbg:
This is a follow-up of the avatars migration which went by unnoticed because the page isn't actively used in prod and only in "preview" state.

Avatars used to be directly attached to the `user` class, now they are a separate DB entity which we have to preload.


gregorbg:
Pure backend optimization

PR #10476, opened 2024-12-25
Store country band details in database

danieljames-dj:
This PR aims at removing some hard-coding related to dues details. This will help to make the changes quickly in case we need to make some change in dues in future.

### DIFF
--- a/app/controllers/country_bands_controller.rb
+++ b/app/controllers/country_bands_controller.rb
@@ -10,7 +10,7 @@ class CountryBandsController < ApplicationController
 
   def edit
     @number = id_from_params
-    unless CountryBandDetail.distinct.pluck(:number).include?(@number)
+    unless CountryBandDetail.exists?(number: @number)
       flash[:danger] = "Unknown band number"
       return redirect_to country_bands_path
     end
--- a/app/controllers/country_bands_controller.rb
+++ b/app/controllers/country_bands_controller.rb
@@ -10,7 +10,7 @@ class CountryBandsController < ApplicationController
 
   def edit
     @number = id_from_params
-    unless CountryBand::BANDS.keys.include?(@number)
+    unless CountryBandDetail.distinct.pluck(:number).include?(@number)
       flash[:danger] = "Unknown band number"
       return redirect_to country_bands_path
     end
--- a/app/models/country_band.rb
+++ b/app/models/country_band.rb
@@ -2,13 +2,15 @@
 
 class CountryBand < ApplicationRecord
   belongs_to :country, foreign_key: :iso2, primary_key: :iso2
+  has_many :country_band_details, foreign_key: :number, primary_key: :number
   validates_inclusion_of :iso2, in: Country::WCA_COUNTRY_ISO_CODES
+  validates_inclusion_of :number, in: (0..5).to_a
 
   def country
     Country.find_by_iso2(self.iso2)
   end
 
   def active_country_band_detail
-    CountryBandDetail.active.find_by(number: self.number)
+    country_band_details.active.first
   end
 end
--- a/app/models/country_band.rb
+++ b/app/models/country_band.rb
@@ -4,7 +4,11 @@ class CountryBand < ApplicationRecord
   belongs_to :country, foreign_key: :iso2, primary_key: :iso2
   has_many :country_band_details, foreign_key: :number, primary_key: :number
   validates_inclusion_of :iso2, in: Country::WCA_COUNTRY_ISO_CODES
-  validates_inclusion_of :number, in: (0..5).to_a
+  validates :number, numericality: {
+    only_integer: true,
+    greater_than_or_equal_to: 0,
+    less_than_or_equal_to: 5,
+  }
 
   def country
     Country.find_by_iso2(self.iso2)
--- a/app/models/country_band.rb
+++ b/app/models/country_band.rb
@@ -1,46 +1,9 @@
 # frozen_string_literal: true
 
 class CountryBand < ApplicationRecord
-  BANDS = {
-    0 => {
-      value: 0.00,
-    },
-    1 => {
-      value: 0.19,
-    },
-    2 => {
-      value: 0.32,
-    },
-    3 => {
-      value: 0.45,
-    },
-    4 => {
-      value: 2.28,
-    },
-    5 => {
-      value: 3.00,
-    },
-  }.freeze
-
-  # According to WCA's current dues policy, the due amount per competitor is equivalent
-  # to this percent of registration fee. Only used if this due amount per competitor is
-  # larger than the due amount per competitor calculated from the competition's country band.
-  PERCENT_REGISTRATION_FEE_USED_FOR_DUE_AMOUNT = 0.15
-
-  def self.percent_registration_fee_used_for_due_amount(country_band)
-    return 0 if country_band.nil?
-    if country_band >= 3
-      0.15
-    elsif country_band >= 1
-      0.05
-    else
-      0.00
-    end
-  end
-
   belongs_to :country, foreign_key: :iso2, primary_key: :iso2
+  belongs_to :country_band_detail, foreign_key: :number, primary_key: :number
   validates_inclusion_of :iso2, in: Country::WCA_COUNTRY_ISO_CODES
-  validates_inclusion_of :number, in: BANDS.keys.freeze
 
   def country
     Country.find_by_iso2(self.iso2)
--- a/app/models/country_band.rb
+++ b/app/models/country_band.rb
@@ -2,10 +2,13 @@
 
 class CountryBand < ApplicationRecord
   belongs_to :country, foreign_key: :iso2, primary_key: :iso2
-  belongs_to :country_band_detail, foreign_key: :number, primary_key: :number
   validates_inclusion_of :iso2, in: Country::WCA_COUNTRY_ISO_CODES
 
   def country
     Country.find_by_iso2(self.iso2)
+  end
+
+  def active_country_band_detail
+    CountryBandDetail.active.find_by(number: self.number)
   end
 end
--- a/app/models/country_band_detail.rb
+++ b/app/models/country_band_detail.rb
@@ -1,0 +1,6 @@
+# frozen_string_literal: true
+
+class CountryBandDetail < ApplicationRecord
+  scope :active, -> { where(end_date: nil).or(inactive.invert_where) }
+  scope :inactive, -> { where(end_date: ..Date.today) }
+end
--- a/app/models/country_band_detail.rb
+++ b/app/models/country_band_detail.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
 class CountryBandDetail < ApplicationRecord
-  scope :active, -> { where(end_date: nil).or(inactive.invert_where) }
+  scope :active, -> { where(end_date: nil).or(where.not(end_date: ..Date.today)) }
   scope :inactive, -> { where(end_date: ..Date.today) }
 end
--- a/app/views/country_bands/_show_band.html.erb
+++ b/app/views/country_bands/_show_band.html.erb
@@ -1,12 +1,12 @@
 <div class="row list-group-item country-band">
   <div class="col-xs-12 text-center title">
-    <%= t("country_bands.band_title", number: number) %>
-    (<%= t("country_bands.due_value", value: data[:value]) %>)
+    <%= t("country_bands.band_title", number: country_band_detail.number) %>
+    (<%= t("country_bands.due_value", value: country_band_detail.due_amount_per_competitor_in_cents.to_f / 100) %>)
     <% if current_user&.can_admin_finances? %>
-      <%= link_to(ui_icon("pencil alt"), edit_country_band_path(id: number)) %>
+      <%= link_to(ui_icon("pencil alt"), edit_country_band_path(id: country_band_detail.number)) %>
     <% end %>
   </div>
-  <% (@country_bands_by_number[number] || []).map(&:country).sort_by(&:name).each do |c| %>
+  <% (@country_bands_by_number[country_band_detail.number] || []).map(&:country).sort_by(&:name).each do |c| %>
     <div class="col-xs-6 col-sm-4 col-md-3 col">
       <div class="country continent-<%= continent_id_as_class(c.continent.id) %>">
         <div class="flag"><%= flag_icon c.iso2 %></div>
--- a/app/views/country_bands/_show_band.html.erb
+++ b/app/views/country_bands/_show_band.html.erb
@@ -1,7 +1,7 @@
 <div class="row list-group-item country-band">
   <div class="col-xs-12 text-center title">
     <%= t("country_bands.band_title", number: country_band_detail.number) %>
-    (<%= t("country_bands.due_value", value: country_band_detail.due_amount_per_competitor_us_cents.to_f / 100) %>)
+    (<%= t("country_bands.due_value", value: country_band_detail.due_amount_per_competitor_us_cents / 100.0) %>)
     <% if current_user&.can_admin_finances? %>
       <%= link_to(ui_icon("pencil alt"), edit_country_band_path(id: country_band_detail.number)) %>
     <% end %>
--- a/app/views/country_bands/_show_band.html.erb
+++ b/app/views/country_bands/_show_band.html.erb
@@ -1,7 +1,7 @@
 <div class="row list-group-item country-band">
   <div class="col-xs-12 text-center title">
     <%= t("country_bands.band_title", number: country_band_detail.number) %>
-    (<%= t("country_bands.due_value", value: country_band_detail.due_amount_per_competitor_in_cents.to_f / 100) %>)
+    (<%= t("country_bands.due_value", value: country_band_detail.due_amount_per_competitor_us_cents.to_f / 100) %>)
     <% if current_user&.can_admin_finances? %>
       <%= link_to(ui_icon("pencil alt"), edit_country_band_path(id: country_band_detail.number)) %>
     <% end %>
--- a/db/migrate/20241225031242_populate_country_band_details.rb
+++ b/db/migrate/20241225031242_populate_country_band_details.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class PopulateCountryBandDetails < ActiveRecord::Migration[7.2]
-  def change
+  def up
     CountryBandDetail.create!(
       number: 0,
       start_date: '2018-01-01',
@@ -38,5 +38,9 @@ class PopulateCountryBandDetails < ActiveRecord::Migration[7.2]
       due_amount_per_competitor_in_cents: 300,
       due_percent_registration_fee: 15,
     )
+  end
+
+  def down
+    CountryBandDetail.delete_all
   end
 end
--- a/db/migrate/20241225031242_populate_country_band_details.rb
+++ b/db/migrate/20241225031242_populate_country_band_details.rb
@@ -5,37 +5,37 @@ class PopulateCountryBandDetails < ActiveRecord::Migration[7.2]
     CountryBandDetail.create!(
       number: 0,
       start_date: '2018-01-01',
-      due_amount_per_competitor_in_cents: 0,
+      due_amount_per_competitor_us_cents: 0,
       due_percent_registration_fee: 0,
     )
     CountryBandDetail.create!(
       number: 1,
       start_date: '2018-01-01',
-      due_amount_per_competitor_in_cents: 19,
+      due_amount_per_competitor_us_cents: 19,
       due_percent_registration_fee: 5,
     )
     CountryBandDetail.create!(
       number: 2,
       start_date: '2018-01-01',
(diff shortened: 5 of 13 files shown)


app/controllers/country_bands_controller.rb
```diff
@@ -10,7 +10,7 @@ def index
 
   def edit
     @number = id_from_params
-    unless CountryBand::BANDS.keys.include?(@number)
+    unless CountryBandDetail.distinct.pluck(:number).include?(@number)
```
gregorbg, line 13: In case of CountryBand, where there's only like 5 or so entries the whole time, loading all values from the DB into memory won't be a big problem. But in general, loading the whole column into memory just for a simple `include?` check is a _big_ problem, so let me show you the cool SQL way
```suggestion
    unless CountryBandDetail.exists?(number: @number)
```

gregorbg: This (roughly, not sure how Rails does internal optimizations exactly) translates to a `SELECT 1 FROM country_bands WHERE number = 3` query

db/migrate/20241225031242_populate_country_band_details.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+class PopulateCountryBandDetails < ActiveRecord::Migration[7.2]
+  def change
```
gregorbg, line 4: This should be `def up`. Then you can also create a `def down` which does `CountryBandDetail.delete_all` or similar.

lib/dues_calculator.rb
```diff
@@ -24,10 +24,12 @@ def self.dues_per_competitor_in_usd(country_iso2, base_entry_fee_lowest_denomina
...
     input_money_us_dollars = Money.new(base_entry_fee_lowest_denomination, currency_code).exchange_to("USD")
 
-    registration_fee_dues_us_dollars = input_money_us_dollars * CountryBand.percent_registration_fee_used_for_due_amount(country_band)
-    country_band_dues_us_dollars = country_band.present? && country_band > 0 ? CountryBand::BANDS[country_band][:value] : 0
-    # times 100 because Money require lowest currency subunit, which is cents for USD
-    country_band_dues_us_dollars_money = Money.new(country_band_dues_us_dollars * 100, "USD")
+    country_band_detail = CountryBandDetail.find_by(number: country_band)
+    return nil unless country_band_detail
+
+    registration_fee_dues_us_dollars = input_money_us_dollars * country_band_detail.due_percent_registration_fee.to_f/100
```
gregorbg, line 30: Spaces around the `/` division sign please

lib/dues_calculator.rb
```diff
@@ -24,10 +24,12 @@ def self.dues_per_competitor_in_usd(country_iso2, base_entry_fee_lowest_denomina
     DuesCalculator.update_exchange_rates_if_needed
     input_money_us_dollars = Money.new(base_entry_fee_lowest_denomination, currency_code).exchange_to("USD")
 
-    registration_fee_dues_us_dollars = input_money_us_dollars * CountryBand.percent_registration_fee_used_for_due_amount(country_band)
-    country_band_dues_us_dollars = country_band.present? && country_band > 0 ? CountryBand::BANDS[country_band][:value] : 0
-    # times 100 because Money require lowest currency subunit, which is cents for USD
-    country_band_dues_us_dollars_money = Money.new(country_band_dues_us_dollars * 100, "USD")
+    country_band_detail = CountryBandDetail.find_by(number: country_band)
+    return nil unless country_band_detail
```
gregorbg, line 28: Is our code prepared to handle a `nil` return value? The previous implementation was returning `0` If the `country_band.present?` check failed.

gregorbg: Also, you can check `country_band_detail.present?`. It's a bit more readable than this "hacky" null check.

gregorbg (changes requested):

lib/dues_calculator.rb
```diff
@@ -20,16 +20,19 @@ def self.dues_for_n_competitors(country_iso2, base_entry_fee_lowest_denomination
 
   def self.dues_per_competitor_in_usd(country_iso2, base_entry_fee_lowest_denomination, currency_code)
     country_band = CountryBand.find_by(iso2: country_iso2)&.number
+    country_band_detail = CountryBandDetail.find_by(number: country_band)
```
gregorbg, line 23: You're completely ignoring your new `belongs_to` association here.

gregorbg (changes requested):

app/models/country_band.rb
```diff
@@ -1,46 +1,9 @@
...
-      0.15
-    elsif country_band >= 1
-      0.05
-    else
-      0.00
-    end
-  end
-
   belongs_to :country, foreign_key: :iso2, primary_key: :iso2
+  belongs_to :country_band_detail, foreign_key: :number, primary_key: :number
```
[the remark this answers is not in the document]
danieljames-dj: Or in other words, I'll not have tight coupling between CountryBand and CountryBandDetail, instead CountryBand table will represent the band number of a country (maybe in future we can introduce start_date and end_date as well if we need the history of bands of a country). From the CountryBandDetails table, we can get the active details of a country band.

lib/dues_calculator.rb
danieljames-dj: Oh sorry, changed it.

lib/dues_calculator.rb
```diff
@@ -20,16 +20,19 @@ def self.dues_for_n_competitors(country_iso2, base_entry_fee_lowest_denomination
...
 
     DuesCalculator.update_exchange_rates_if_needed
-    input_money_us_dollars = Money.new(base_entry_fee_lowest_denomination, currency_code).exchange_to("USD")
 
-    registration_fee_dues_us_dollars = input_money_us_dollars * CountryBand.percent_registration_fee_used_for_due_amount(country_band)
-    country_band_dues_us_dollars = country_band.present? && country_band > 0 ? CountryBand::BANDS[country_band][:value] : 0
-    # times 100 because Money require lowest currency subunit, which is cents for USD
-    country_band_dues_us_dollars_money = Money.new(country_band_dues_us_dollars * 100, "USD")
+    # Calculation of 'registration fee dues'
+    registration_fee_dues = Money.new(registration_fees * (country_band_detail&.due_percent_registration_fee.to_f || 0) / 100, "USD")
```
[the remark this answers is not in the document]
danieljames-dj: Cleaned it a bit

lib/dues_calculator.rb
```diff
@@ -20,16 +20,19 @@ def self.dues_for_n_competitors(country_iso2, base_entry_fee_lowest_denomination
...
 
-    [registration_fee_dues_us_dollars, country_band_dues_us_dollars_money].max
+    # Calculation of 'country band dues'
+    country_band_dues = Money.new(country_band_detail&.due_amount_per_competitor_in_cents || 0, "USD")
```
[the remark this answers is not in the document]
danieljames-dj: Cleaned it a bit

app/views/country_bands/_show_band.html.erb
```diff
@@ -1,12 +1,12 @@
 <div class="row list-group-item country-band">
   <div class="col-xs-12 text-center title">
-    <%= t("country_bands.band_title", number: number) %>
-    (<%= t("country_bands.due_value", value: data[:value]) %>)
+    <%= t("country_bands.band_title", number: country_band_detail.number) %>
+    (<%= t("country_bands.due_value", value: country_band_detail.due_amount_per_competitor_in_cents.to_f / 100) %>)
```
gregorbg, line 4: I'm 60% sure you don't need the `to_f`

app/models/country_band_detail.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+class CountryBandDetail < ApplicationRecord
```
gregorbg, line 3: You can specify a `belongs_to :country_band` association by providing `primary_key: :number` and `foreign_key: :number` I think.

gregorbg: Don't confuse these with SQL `PRIMARY KEY` and `FOREIGN KEY`. The options I'm talking about here are (easily misunderstandable) Rails AR vocabulary, which means "the key [i.e. column] where I look up information in the foreign [i.e. pointing to] model, or in myself [i.e. primary]".

app/models/country_band.rb
```diff
@@ -1,48 +1,14 @@
...
   validates_inclusion_of :iso2, in: Country::WCA_COUNTRY_ISO_CODES
-  validates_inclusion_of :number, in: BANDS.keys.freeze
 
   def country
     Country.find_by_iso2(self.iso2)
   end
+
+  def active_country_band_detail
+    CountryBandDetail.active.find_by(number: self.number)
+  end
```
gregorbg, line 13: As per my other comment in `CountryBandDetail`, I think you can solve this via a `has_many` relation.

db/schema.rb
```diff
@@ -665,6 +665,16 @@
     t.datetime "updated_at", null: false
   end
 
+  create_table "country_band_details", charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
+    t.integer "number", null: false
+    t.date "start_date", null: false
+    t.date "end_date"
+    t.integer "due_amount_per_competitor_in_cents", null: false
```
gregorbg, line 672: Maybe name this column `due_amount_per_competitor_us_cents` to make clear that the currency is pre-defined?

lib/dues_calculator.rb
```diff
@@ -19,17 +19,22 @@ def self.dues_for_n_competitors(country_iso2, base_entry_fee_lowest_denomination
...
-    country_band_dues_us_dollars = country_band.present? && country_band > 0 ? CountryBand::BANDS[country_band][:value] : 0
-    # times 100 because Money require lowest currency subunit, which is cents for USD
-    country_band_dues_us_dollars_money = Money.new(country_band_dues_us_dollars * 100, "USD")
+    country_band = CountryBand.find_by(iso2: country_iso2)
+    country_band_detail = country_band&.active_country_band_detail
+    registration_fees = Money.new(base_entry_fee_lowest_denomination, currency_code).exchange_to("USD")
+
+    # Calculation of 'registration fee dues'
+    due_percent_registration_fee = country_band_detail&.due_percent_registration_fee.to_f || 0
+    registration_fee_dues = Money.new(registration_fees * due_percent_registration_fee / 100, "USD")
```
gregorbg, line 30: I think you can do calculations with the money object directly:
```
my_fee = Money.new(2000, "USD")
some_percentage = my_fee * 0.8
```

lib/dues_calculator.rb
```diff
@@ -19,17 +19,22 @@ def self.dues_for_n_competitors(country_iso2, base_entry_fee_lowest_denomination
...
-    registration_fee_dues_us_dollars = input_money_us_dollars * CountryBand.percent_registration_fee_used_for_due_amount(country_band)
-    country_band_dues_us_dollars = country_band.present? && country_band > 0 ? CountryBand::BANDS[country_band][:value] : 0
-    # times 100 because Money require lowest currency subunit, which is cents for USD
-    country_band_dues_us_dollars_money = Money.new(country_band_dues_us_dollars * 100, "USD")
+    country_band = CountryBand.find_by(iso2: country_iso2)
+    country_band_detail = country_band&.active_country_band_detail
+    registration_fees = Money.new(base_entry_fee_lowest_denomination, currency_code).exchange_to("USD")
+
+    # Calculation of 'registration fee dues'
+    due_percent_registration_fee = country_band_detail&.due_percent_registration_fee.to_f || 0
```
gregorbg, line 29: Again, I think you don't need the `to_f`. Ruby is very flexible with numbers.

app/models/country_band.rb
```diff
@@ -1,48 +1,14 @@
...
-    elsif country_band >= 1
-      0.05
-    else
-      0.00
-    end
-  end
-
   belongs_to :country, foreign_key: :iso2, primary_key: :iso2
   validates_inclusion_of :iso2, in: Country::WCA_COUNTRY_ISO_CODES
-  validates_inclusion_of :number, in: BANDS.keys.freeze
```
gregorbg, line 43: Can you hard-code the range from 0 to 5 instead of just completely deleting the line?
I know hard-coding is not very pretty, but effectively it was already hard-coded before (through an intermediate step which was getting the keys from the hard-coded hash)

spec/models/country_band_spec.rb
```diff
@@ -15,6 +15,6 @@
 
   it "invalidates band with invalid band id" do
     cb = CountryBand.new(number: 6, iso2: "HELLO")
-    expect(cb).to be_invalid_with_errors(number: ["is not included in the list"])
+    expect(cb).to be_invalid_with_errors(country: ["must exist"])
```
gregorbg, line 18: Regarding my other comment above, feel free to re-add the previous "number not included in the list" validation.

gregorbg: Also, why did the `country` validation that you're testing now not exist/fail before?

gregorbg (changes requested):

app/models/country_band.rb
danieljames-dj: Okay done.

app/models/country_band_detail.rb
danieljames-dj: Okay done

app/models/country_band.rb
danieljames-dj: Done

app/models/country_band_detail.rb
danieljames-dj: Oh wait, removed it for now. But looks like the association is not used anywhere, is it still needed to have this association here?

spec/models/country_band_spec.rb
danieljames-dj: Added back that error. I just realized the country validation test was already there in the unit test just before this.

danieljames-dj (commented):
@gregorbg Thanks for your review, I've partially attended your comments. But since some changes are clashing with changes in https://github.com/thewca/worldcubeassociation.org/pull/10767, I'll wait for that PR to get merged before attending the remaining.

app/models/country_band.rb
```diff
@@ -1,48 +1,16 @@
...
-    else
-      0.00
-    end
-  end
-
   belongs_to :country, foreign_key: :iso2, primary_key: :iso2
+  has_many :country_band_details, foreign_key: :number, primary_key: :number
   validates_inclusion_of :iso2, in: Country::WCA_COUNTRY_ISO_CODES
-  validates_inclusion_of :number, in: BANDS.keys.freeze
+  validates_inclusion_of :number, in: (0..5).to_a
```
gregorbg, line 7: Rails has special `numericality` validators for cases like this: https://guides.rubyonrails.org/active_record_validations.html#numericality

db/schema.rb
danieljames-dj: Okay changed.

danieljames-dj (commented):
One more issue remaining, I've pinged you in slack.

UPDATE: The one issue is resolved, it was related to `invert_where` and the details are explained in https://github.com/thewca/worldcubeassociation.org/pull/10818.

app/views/country_bands/_show_band.html.erb
gregorbg: Then divide by `100.0`

lib/dues_calculator.rb
gregorbg: But `x / 100.0` will return `0.05` :wink:

app/views/country_bands/_show_band.html.erb
danieljames-dj: Okay done

spec/models/country_band_spec.rb
```diff
@@ -14,7 +14,7 @@
   end
 
   it "invalidates band with invalid band id" do
-    cb = CountryBand.new(number: 6, iso2: "HELLO")
-    expect(cb).to be_invalid_with_errors(number: ["is not included in the list"])
```
gregorbg, line 18: Can you please keep this country-related smoke check in a separate test?

lib/dues_calculator.rb
```diff
@@ -9,16 +9,20 @@ def self.dues_per_competitor(country_iso2, base_entry_fee_lowest_denomination, c
   end
 
   def self.dues_per_competitor_in_usd(country_iso2, base_entry_fee_lowest_denomination, currency_code)
-    country_band = CountryBand.find_by(iso2: country_iso2)&.number
+    country_band = CountryBand.find_by(iso2: country_iso2)
+    country_band_detail = country_band&.active_country_band_detail
+    registration_fees = Money.new(base_entry_fee_lowest_denomination, currency_code).exchange_to("USD")
```
gregorbg, line 14: The name should make clear that we're dealing with the reg fees in US dollars.

lib/dues_calculator.rb
```diff
@@ -9,16 +9,20 @@ def self.dues_per_competitor(country_iso2, base_entry_fee_lowest_denomination, c
   end
 
   def self.dues_per_competitor_in_usd(country_iso2, base_entry_fee_lowest_denomination, currency_code)
```
gregorbg, line 11: You're doing an awful lot of `&.` and `|| 0` safe navigation throughout the whole method. Most of if seems to be related to the fact that `country_band_detail` can be null. When and under which circumstances can this happen? And if it happens, can you just `return` early and then after the return be sure that the CB details exist?

gregorbg (changes requested):

spec/models/country_band_spec.rb
danieljames-dj: That is already there in the previous unit test, is that fine?

lib/dues_calculator.rb
danieljames-dj: Okay renamed.

lib/dues_calculator.rb
danieljames-dj: > When and under which circumstances can this happen

I think it happens only if there is no active country_band_detail for a country.

I've returned early as you suggested.

lib/dues_calculator.rb
```diff
@@ -9,16 +9,22 @@ def self.dues_per_competitor(country_iso2, base_entry_fee_lowest_denomination, c
...
-    country_band = CountryBand.find_by(iso2: country_iso2)&.number
+    country_band = CountryBand.find_by(iso2: country_iso2)
+    country_band_detail = country_band&.active_country_band_detail
+    registration_fees_in_usd = Money.new(base_entry_fee_lowest_denomination, currency_code).exchange_to("USD")
 
-    input_money_us_dollars = Money.new(base_entry_fee_lowest_denomination, currency_code).exchange_to("USD")
+    return nil if country_band_detail.nil?
```
gregorbg, line 16: If you move this statement one line higher, you can save yourself an extra call to `exchange_to` :wink:

lib/dues_calculator.rb
```diff
@@ -9,16 +9,22 @@ def self.dues_per_competitor(country_iso2, base_entry_fee_lowest_denomination, c
...
 
-    registration_fee_dues_us_dollars = input_money_us_dollars * CountryBand.percent_registration_fee_used_for_due_amount(country_band)
-    country_band_dues_us_dollars = country_band.present? && country_band > 0 ? CountryBand::BANDS[country_band][:value] : 0
-    # times 100 because Money require lowest currency subunit, which is cents for USD
-    country_band_dues_us_dollars_money = Money.new(country_band_dues_us_dollars * 100, "USD")
+    # Calculation of 'registration fee dues'
+    due_percent_registration_fee = country_band_detail.due_percent_registration_fee
```
gregorbg, line 19: Is declaring this variable super-duper necessary? It's only used once below...

lib/dues_calculator.rb
```diff
@@ -9,16 +9,22 @@ def self.dues_per_competitor(country_iso2, base_entry_fee_lowest_denomination, c
...
+    registration_fee_dues = registration_fees_in_usd * due_percent_registration_fee / 100.0
 
-    [registration_fee_dues_us_dollars, country_band_dues_us_dollars_money].max
+    # Calculation of 'country band dues'
+    due_amount_per_competitor_us_cents = country_band_detail.due_amount_per_competitor_us_cents
```
gregorbg, line 23: Same as above. this variable seems kinda redundant now

gregorbg (changes requested):

lib/dues_calculator.rb
danieljames-dj: Okay, I thought of keeping inits at the beginning. I've moved the above line below now (same as you suggested). Hope it is fine now.

lib/dues_calculator.rb
danieljames-dj: Okay removed.

lib/dues_calculator.rb
danieljames-dj: Done

spec/models/country_band_spec.rb
gregorbg: Okidoki!

gregorbg (approved):

PR #10477, opened 2024-12-26
More cleanly transition to mandatory :payment_integration parameter in URL

gregorbg:
Basically a copy-cat of #10458, except that the parameter in the URL is not marked as required yet.

This will make it so that if anybody starts a Stripe session with the old (paremeter-less, because optional) URL _during deployment_, the route would still resolve after coming back to the newly deployed server.

Highly unlikely circumstance anyways (somebody would need to perform their 3DS authorization _exactly_ during the time window that AWS swaps out our ECS instances and switches over the traffic) but with payments you can never be safe enough.

Since the original #10458 is already reviewed and approved, I will merge this as soon as tests pass.

### DIFF
--- a/app/webpacker/components/RegistrationsV2/Register/PaymentStep.jsx
+++ b/app/webpacker/components/RegistrationsV2/Register/PaymentStep.jsx
@@ -65,7 +65,7 @@ export default function PaymentStep({
       elements,
       clientSecret,
       confirmParams: {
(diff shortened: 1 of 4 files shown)


gregorbg:
Merging based on the review of #10458

This PR literally cherry-picks the exact same commits (and squashes them, so the hashes won't match. But the diff does!)

PR #10412, opened 2024-12-11
Precompute "popular" Records/Results pages in Sidekiq as part of CAD

gregorbg:
Basically to remedy the issue that several threads get fried when people try to access Records pages directly after CAD has been run.

We pre-compute and insert the most popular pages (i.e. the default filter selection that you would see by just opening "Records", but not sophisticated queries like "Female Pyraminx Average History for Guatemala") and insert those into the cache using the same key which the page is expected to fetch just a moment later.

As an additional optimization, we could outsource this new "CAD part" into a separate job so that CAD runtimes aren't skewed. But then the cache timestamp is already invalid and we might not be fast enough with pre-computing until the statistics fetischists out there start bombarding us

PR #10491, opened 2024-12-30
Replace CAD *Ranks table calculation with SQL window function

gregorbg:
Kinda-followup to #10412

While testing the queries over in that other PR, I noticed how the `RanksSingle` and `RanksAverage` computations were kinda slow, and after staring at the code for a while I realized that it's really just computing numeric rankings per (world, continent, country) each.

Window functions in MySQL have been around for long enough, and MySQL 8 is pretty much standard everywhere now, so I believe it's fair to use this. See https://dev.mysql.com/doc/refman/8.4/en/window-function-descriptions.html#function_rank and https://dev.mysql.com/doc/refman/8.4/en/window-functions-usage.html for details.

We still need to discuss with WRT whether these queries are actually equivalent, so I'm setting this to Draft for the time being.


Baguettely:
This should work to correctly capture the edge cases for nationality changes. I haven't been able to run a full regression test because a) it times out on phpmyadmin and b) I can't compare the run time for SQL vs Ruby

```
# Get all personal records, grouping for each person by event, event+continent, and event+continent+country
WITH personal_bests AS (
   SELECT personId, 
      eventId,
      continentId,
      countryId, 
      MIN(best) AS value
   FROM ConciseSingleResults
   GROUP BY personId,
      eventId,
      continentId,
      countryId
   WITH ROLLUP
   HAVING eventId IS NOT NULL
),
# Calculate all rankings via RANK(). worldRank ranks over the best results for each person by event. continentRank ranks over the best result for each person by event+continent. countryRank ranks over the best result for each person by event+continent+country. If someone changes country and/or continent, their results from the previous region still count towards the rankings for others in that region, however do not count towards their own regional ranking. Regional rankings are marked as NULL if they do not match the person's present region of representation.

rankings_calculations AS (
   SELECT personId,
      eventId,
      value,
      CASE WHEN pb.continentId IS NULL
         THEN RANK() OVER(PARTITION BY eventId
            ORDER BY CASE WHEN pb.continentId IS NULL THEN 0 ELSE 1 END, value)
         END AS worldRank,
      CASE WHEN pb.countryId IS NULL AND co.continentId = pb.continentId
         THEN RANK() OVER(PARTITION BY eventId, pb.continentId
            ORDER BY CASE WHEN pb.countryId IS NULL AND pb.continentId IS NOT NULL THEN 0 ELSE 1 END, value) 
         END AS continentRank,
      CASE WHEN pb.countryId IS NOT NULL AND p.countryId = pb.countryId
          THEN RANK() OVER(PARTITION BY eventId, pb.countryId
             ORDER BY CASE WHEN pb.countryId IS NOT NULL THEN 0 ELSE 1 END, value)
          END AS countryRank
   FROM personal_bests pb
   JOIN Persons p
      ON p.wca_id = pb.personId
         AND p.subId = 1
   JOIN Countries co
     ON co.id = p.countryId
)

# Group again by person to combine the world/continent/country rankings for each person. If someone has changed region of representation, their ranking is per the fastest result achieved under their present region. If they have not achieved any results in an event since changing their region, their rank is stored as 0.
SELECT personId,
   eventId,
   MIN(value) AS best,
   MIN(worldRank) AS worldRank,
   IFNULL(MIN(continentRank), 0) AS continentRank,
   IFNULL(MIN(countryRank), 0) AS countryRank
FROM ranking_calculations
GROUP BY personId,
   eventId
ORDER BY eventId,
   worldRank
```

gregorbg:
Thanks a lot for your query @Baguettely! I hijacked it and tried to rework it into three different subqueries. I then had Gemini run it over to make it even more SQL-scheduler-friendly, came up with some indexes (akin to the speedup indexes on `regional_records_lookup`).

I think it does what it's supposed to, and it gets the tests passing locally. For what it's worth, here's a fully glued-together version for the `concise_single_results` and `single_ranks` table:

```sql
WITH personal_bests AS (
    SELECT person_id, event_id, continent_id, country_id, MIN(best) AS value
    FROM concise_single_results
    GROUP BY person_id, event_id, continent_id, country_id
    WITH ROLLUP
    HAVING event_id IS NOT NULL
       AND person_id IS NOT NULL
),
current_persons AS (
    SELECT wca_id, country_id, countries.continent_id
    FROM persons
        INNER JOIN countries ON persons.country_id = countries.id
    WHERE persons.sub_id = 1
),
world_ranks AS (
    SELECT person_id, event_id, value,
        RANK() OVER (PARTITION BY event_id ORDER BY value) AS world_rank
    FROM personal_bests
    WHERE continent_id IS NULL
        AND country_id IS NULL
),
continent_ranks AS (
    SELECT person_id, event_id, continent_id, value,
        RANK() OVER (PARTITION BY event_id, continent_id ORDER BY value) AS continent_rank
    FROM personal_bests
    WHERE country_id IS NULL
       AND continent_id IS NOT NULL
),
country_ranks AS (
    SELECT person_id, event_id, country_id, value,
        RANK() OVER (PARTITION BY pb.event_id, pb.country_id ORDER BY pb.value) AS country_rank
    FROM personal_bests pb
    WHERE pb.country_id IS NOT NULL
)
SELECT cp.wca_id AS person_id,
       wr.event_id,
       wr.value AS best,
       wr.world_rank AS world_rank,
       COALESCE(cr.continent_rank, 0) AS continent_rank,
       COALESCE(nr.country_rank, 0) AS country_rank
FROM current_persons cp
    INNER JOIN world_ranks wr ON cp.wca_id = wr.person_id
    LEFT JOIN continent_ranks cr
        ON cp.wca_id = cr.person_id AND wr.event_id = cr.event_id AND cp.continent_id = cr.continent_id
    LEFT JOIN country_ranks nr
        ON cp.wca_id = nr.person_id AND wr.event_id = nr.event_id AND cp.country_id = nr.country_id
ORDER BY wr.event_id, world_rank;
```

At least locally, this gets the tests passing. Let me know what you think!

PR #10493, opened 2024-12-31
Only show competing registrations in Competitor list

dunkOnIT:
TL;DR - I'm operating on the assumption that devs should explicitly _request_ the presence of non-competing registrations when working with registration data, which seems much less likely to lead to issues than non-competing registrations being returned when they don't realize they've implicitly requested them

Currently, we show non-competing registrations in the Competitor list - eg, Amanda Campbell in [Warm Up Seattle 2025](https://www.worldcubeassociation.org/competitions/WarmUpSeattle2025/registrations).

This is because our `accepted` scope doesn't check whether a user `is_competing` or not. This is technically correct, as non-competing lanes will also have registration_statuses, but in practice I'm concerned that most developers (myself included) wouldn't realize that calling the `accepted` scope is also returning non-competing registrations.

Based on this, I think the **best** solution would be to:
- Renamed `accepted` scope to `accepted_registrations`
- Add an `accepted_competitors` scope which returns `competing_status: 'accepted', is_competing: true`

However, that would mean
- renaming ~40 instances where the `accepted` scope is used, which carries some risk and will take time
- that we should also consider changing the other scopes to match this new naming convention
- we are preparing for a future (more formal support for non-competing registrations) that isn't here yet - so perhaps we're over-engineering

Thus, the actually-submitted PR returns all accepted & competing registrations in the `accepted` scope - which is what I think developers expect when they use that scope. (Because if they want non-competing registrations, they'll access them directly using the `non-competing` scope)

### DIFF
--- a/app/controllers/api/v1/registrations/registrations_controller.rb
+++ b/app/controllers/api/v1/registrations/registrations_controller.rb
@@ -90,7 +90,7 @@ class Api::V1::Registrations::RegistrationsController < Api::V1::ApiController
   def list
     competition_id = list_params
     competition = Competition.find(competition_id)
-    registrations = competition.registrations.accepted
+    registrations = competition.registrations.accepted.competing
     payload = Rails.cache.fetch([
                                   "registrations_v2_list",
                                   competition.id,
--- a/app/models/registration.rb
+++ b/app/models/registration.rb
@@ -3,7 +3,7 @@
 class Registration < ApplicationRecord
   # TODO: Reg-V3 Cleanup: Remove all these and use the competing_status_{status} scopes
   scope :pending, -> { where(competing_status: 'pending') }
-  scope :accepted, -> { where(competing_status: 'accepted') }
+  scope :accepted, -> { where(competing_status: 'accepted', is_competing: true) }
   scope :cancelled, -> { where(competing_status: 'cancelled') }
   scope :rejected, -> { where(competing_status: 'rejected') }
   scope :waitlisted, -> { where(competing_status: 'waiting_list') }
--- a/app/models/registration.rb
+++ b/app/models/registration.rb
@@ -3,11 +3,12 @@
 class Registration < ApplicationRecord
   # TODO: Reg-V3 Cleanup: Remove all these and use the competing_status_{status} scopes
   scope :pending, -> { where(competing_status: 'pending') }
-  scope :accepted, -> { where(competing_status: 'accepted', is_competing: true) }
+  scope :accepted, -> { where(competing_status: 'accepted') }
   scope :cancelled, -> { where(competing_status: 'cancelled') }
   scope :rejected, -> { where(competing_status: 'rejected') }
   scope :waitlisted, -> { where(competing_status: 'waiting_list') }
   scope :non_competing, -> { where(is_competing: false) }
+  scope :competing, -> { where(is_competing: true) }
   scope :not_cancelled, -> { where.not(competing_status: 'cancelled') }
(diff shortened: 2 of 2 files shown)


gregorbg:
I understand the motivation for this fix, but I have to veto the current implementation.

By your logic, you'd have to adjust the `pending`, `cancelled`, etc. scopes as well, and this is not something I'm willing to do.
Filtering for a certain status should only filter for status. If the results are unexpected at some point in the code (in our case, the API that feeds data to the frontend) then the query should be manipulated _at the call-site_ where the "unexpected" results are being generated.

For a short-term fix, you can introduce another `only_competing` (name tentative) scope as `scope :only_competing, -> { where(is_competing: true) }`. Then, in the API that lists registrations, you can change the call to `Registration.accepted.only_competing.where(competition_id: 'FooOpen2000')`, which will correctly apply _both_ conditions.

In the long term, we should accept that this is just the way things are now. The flag `is_competing` is a remnant of older times, when we didn't have "proper" support for lanes. Now that the concept of lanes is in our heads, we probably need to "upgrade" this boolean to an enum instead. `is_competing = false` implies `reg_type = competing` and `is_competing = true` implies `reg_type = staff`.

We then "expose" the `reg_type` to the frontend and the React frontend can decide (based on a case study and design prototypes that we have yet to do!) where and how to display these in the frontend.

gregorbg (changes requested):
See my longer comment in the issue tree on why I don't think this is an acceptable solution, even for a short-term hotfix

coder13:
Ideas:
We should introduce some terminology:
- **competitor**- someone with an accepted registration with events
- **registrant**- someone with a registration (not delegates/organizers who didn't register, people with non-null registrantIds
- **persons**
  - For the purpose of the public wcif, everyone with accepted registrations and registrantIds
  - For the purpose of the private wcif, everyone and anyone associated with a comp regardless of if they have a registration

Bonus: **non-competing-staff**- people with an accepted registration, 0 events, and marked as staff (or in today's case, `is_competing` is false)

gregorbg:
Yup, that terminology more or less coincides with our vision of a "reg_status" that I outlined in my previous post.

Briefly put, we need to say goodbye to the notion that a `Registration` is always forcibly someone who _competes_ at the event. (Thereby implicitly affecting the terminology).

In the long run, a "registration" should just be a tiny bit of information that says "this person attends this event in some capacity". The only things that these registrations can have in common (due to their abundance of potential capacities that you could attend the event as) are
1. Who is attending
2. What competition they're attending
3. What capacity you're attending the event as (a string-ish enum)
4. What "lifecycle state" their registration is in (accepted, deleted, etc.). Note that all registrations have this property, no matter which reg_status they have in point (3) above

Everything else needs to be stored and managed as "metadata" to a registration, which of course also affects WCIF. But we shouldn't model our internal database in a way that makes it easiest to squeeze into existing WCIF assumptions. Rather, we should introduce breaking changes in WCIF (new schema version) if our internal models require us to do so. The current ability to sync what effectively is a competing registration and just "hack" it as non-competing by marking it with a boolean that was a hotfix proposed by Tim so he can have an easier time using his third party tools for some US Nats is definitely not a sustainable solution

dunkOnIT:
I like the chained-scope solution. I do still worry that this will lead devs to unwittingly include non-competing registrations in their accepted scopes, but this is probably something we'll just need to watch out for until we've been able to modernize the non-competing registration support.

gregorbg (approved):

Issue #1356, opened 2017-02-22
Unclear behavior when competition's currency changes when a payment has been made


viroulep:
Just thought about this: I don't think we handle well the case where the organizer or delegate change the currency when a payment has been made!
The refund behavior is well defined since we refund a Stripe charge, however we don't take into account multiple currencies in some part of the website, eg in `paid_entry_fees`:
```ruby
  def paid_entry_fees
    Money.new(
      registration_payments.sum(:amount_lowest_denomination),
      # FIXME: see https://github.com/thewca/worldcubeassociation.org/issues/1257
      (competition.currency_code.blank? ? Money.default_currency : competition.currency_code),
    )
  end
```
Note how we just make a sum, assuming everything is in the same currency.

Suggestion: make the currency immutable as long as anyone has a non zero balance for the competition.

larspetrus:
Yeah, I agree with immutability!

PR #10497, opened 2024-12-31
Can only update a competition's currency code if there are no unrefunded payments

dunkOnIT:
Addresses #1356 - in draft because I've asked WCAT if this is something that's still needed.

If this is needed, then I think this implementation _may_ be a bit thin. If we get the mega-edge-case of a user having been over-refunded, and another user still having a not-yet-refunded payment, the total amount paid for a competition will be $0 even though there are still two unresolved payments.

As such, it might be better to `find` a registration with a non-zero sum of its `registration_payments`

Also - it seems like we don't track the overall status of a payment (eg, whether it has been refunded) - we only infer this by summing registration_payments related to a given registration.

### DIFF
--- a/app/models/competition.rb
+++ b/app/models/competition.rb
@@ -353,6 +353,12 @@ class Competition < ApplicationRecord
     end
   end
 
+  validate :payments_nil_to_change_currency
+  private def payments_nil_to_change_currency
+    return unless currency_code_changed?
+    errors.add(:currency_code, I18n.t('competitions.errors.currency_cant_change')) if total_payment_amount != 0
+  end
+
   def has_any_round_per_event?
     competition_events.map(&:rounds).none?(&:empty?)
   end
@@ -2686,6 +2692,10 @@ class Competition < ApplicationRecord
 
   def disconnect_all_payment_integrations
     competition_payment_integrations.destroy_all
+  end
+
+  def total_payment_amount
+    registrations.joins(:registration_payments).sum('registration_payments.amount_lowest_denomination')
   end
 
   # Our React date picker unfortunately behaves weirdly in terms of backend data
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1933,6 +1933,7 @@ en:
         -6002: "You need to finish your registration before you can pay"
     #context: and when an error occured
     errors:
+      currency_cant_change: "You cannot change the currency until all outstanding payments have been refunded."
       invalid_name_message: "must end with a year and must contain only alphanumeric characters, dashes(-), ampersands(&), periods(.), colons(:), apostrophes('), and spaces( )"
       cannot_manage: "Cannot manage competition."
       cannot_delete_public: "Cannot delete a competition that is publicly visible."


gregorbg:
Hm, the original issue also mentions another interesting edge case: Simple questions like "How much income does this competition have?" also rely on a (competition-)global currency. So even if there are no unrefunded payments, you would still sum up existing payments and then treat them as the new currency.
TL;DR I think we should just straight out forbid changing the currency as long as payments exist.

gregorbg:
Another alternative would be to only allow updating the currency if all payments are _fully_ refunded.

dunkOnIT:
> Another alternative would be to only allow updating the currency if all payments are fully refunded. If I have ten payments worth 1500 Something-Dollars each, and I refund every last one, then the sum evaluates to zero, which pretty much has the same meaning in any currency.

I think this is what I'm already doing? If you check the diff this is my logic to check for "unrefunded" payments: 
``
registrations.joins(:registration_payments).sum('registration_payments.amount_lowest_denomination')
```

**Comment by dunkOnIT** [comment]
Blocked as it needs signoff from WCAT

**Comment by dunkOnIT** [comment]
Unblocked - WCAT is happy with it, and the team agreed internally that it is necessary because we have areas of the code that assume a consistent currency code when doing things like totalling the registration fees received for a given competition

File: spec/controllers/competitions_controller_spec.rb [hunk 1]
```diff
@@ -401,17 +401,26 @@
         expect(competition.currency_code).to eq "EUR"
       end
 
-      it "can update the registration fees when there is any payment" do
+      it "can update the registration fee amount when there is any payment", :only do
```
gregorbg: Stray `only` tag

spec/controllers/competitions_controller_spec.rb
```diff
@@ -401,17 +401,26 @@
...
+        update_params = build_competition_update(competition, entryFees: { baseEntryFee: previous_fees + 10 })
         patch :update, params: update_params, as: :json
         expect(response).to be_successful
         competition.reload
         expect(competition.base_entry_fee_lowest_denomination).to eq previous_fees + 10
-        expect(competition.currency_code).to eq "EUR"
+      end
+
+      # We rely on a consistent currency code for many functions - such as summing total registration fees
+      it "cannot update the registration fee currency when there is any payment", :only do
```
gregorbg: Stray `only` tag

app/models/competition.rb
```diff
@@ -2658,6 +2665,10 @@ def disconnect_all_payment_integrations
     competition_payment_integrations.destroy_all
   end
 
+  def total_payment_amount
+    registrations.joins(:registration_payments).sum('registration_payments.amount_lowest_denomination')
```
gregorbg: For better readability, you can add a `has_many :registration_payments, through: :registrations` association above, and then simply do `registration_payments.sum(:amount_lowest_denomination)` here.

app/models/competition.rb
```diff
@@ -379,6 +379,13 @@ def advancement_condition_must_be_present_for_all_non_final_rounds
     errors.add(:registration_close, I18n.t('competitions.errors.registration_already_closed')) if !editing_user.can_admin_competitions? && registration_range_specified? && registration_past?
   end
 
+  validate :payments_nil_to_change_currency
+  private def payments_nil_to_change_currency
+    return unless currency_code_changed?
```
gregorbg: This can directly become an `if: :currency_code_changed?` option in the `validate` declaration

gregorbg (changes requested):

spec/controllers/competitions_controller_spec.rb
dunkOnIT: removed

spec/controllers/competitions_controller_spec.rb
dunkOnIT: removed

app/models/competition.rb
dunkOnIT: Changed

app/models/competition.rb
dunkOnIT: Changed

PR #10509, opened 2025-01-02
Add an option to reset failed cronjobs from the UI

gregorbg:
Mostly for WRT when CAD occasionally crashes.

image


gregorbg:
Merging on my authority

PR #10510, opened 2025-01-02
Remove i18n-country-translations as it is not used anywhere

danieljames-dj:
I'm not 100% sure about this change, I couldn't find any usage like given in [README of the gem](https://github.com/thewca/i18n-country-translations). I searched for `I18n.t(:` and couldn't find any relevant details, so I believe this is no longer needed.

The reason why I want this to be removed is, when I install `currencylayer` gem, I got the following error:

```
Could not find compatible versions



Because every version of i18n-country-translations depends on i18n >= 0.9.3, < 2

  and money >= 6.5.0, < 6.8.2 depends on i18n >= 0.6.4, <= 0.7.0,

  every version of i18n-country-translations is incompatible with money >= 6.5.0, < 6.8.2.

And because every version of currencylayer depends on money ~> 6.5.0,

  every version of i18n-country-translations is incompatible with currencylayer >= 0.

So, because Gemfile depends on currencylayer >= 0

  and Gemfile depends on i18n-country-translations >= 0,

  version solving has failed.
```

I believe the problem is that `i18n-country-translations` requires old version of `money` while `currencylayer` requires newer version of `money`. Removing `i18n-country-translations` (if not used) will allow to upgrade money to newer version and hence allow to install `currencylayer`.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -83,7 +83,6 @@ gem 'blocks'
 gem 'rack-cors', require: 'rack/cors'
 gem 'api-pagination'
 gem 'daemons'
-gem 'i18n-country-translations', github: 'thewca/i18n-country-translations'
 gem 'http_accept_language'
 gem 'twitter_cldr'
 # version explicitly specified because Shakapacker wants to keep Gemfile and package.json in sync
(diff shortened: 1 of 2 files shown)


danieljames-dj:
Looks like it's actually used. Will try some other way.

PR #10518, opened 2025-01-03
Show Registration status in comp overview page by default

gregorbg:
Promised follow-up to #10461

Only feature we're losing now is the "hydrated caching" in Redis. Registration values are now always pulled from the database freshly, and TBH I'm not sure whether that's a good thing.


gregorbg:
I've added an attempt at providing cached registration status even in the default index, which should hopefully alleviate the burden on our DB

kr-matthews:
Is this checkbox meant to be there twice?

Besides that, front-end code looks good.

image

gregorbg:
Thanks, that was obviously a merge whoopsie which I've fixed now.

PR #10520, opened 2025-01-03
Change RubyMoney bank to currencylayer

danieljames-dj:
The bank is changed so that all our currencies will be supported.


lib/dues_calculator.rb
```diff
@@ -1,12 +1,6 @@
 # frozen_string_literal: true
 
 module DuesCalculator
-  def self.update_exchange_rates_if_needed
-    if !Money.default_bank.rates_updated_at || Money.default_bank.rates_updated_at < 1.day.ago
-      Money.default_bank.update_rates
-    end
-  end
```
gregorbg, line 8: Why does this need to go? Especially for `CurrencyLayer` which is rate-limited (we pay to get a maximum amount of 10,000 requests per month) having a caching mechanism like that would be lovely.

config/initializers/ruby_money.rb
```diff
@@ -1,8 +1,16 @@
...
 Money.locale_backend = :i18n
 
-eu_bank = EuCentralBank.new
-Money.default_bank = eu_bank
+mclb = Money::Bank::CurrencylayerBank.new
+mclb.access_key = ENV.fetch("CURRENCY_LAYER_API_KEY", "")
+mclb.currencylayer = true
+mclb.ttl_in_seconds = 86_400
+mclb.cache = 'rails_money_cache.txt' # To be replaced with proper cache path.
+mclb.update_rates
```
gregorbg, line 12: This would mean rates are only updated once per server boot. We sometimes go for days without redeploying, which in some cases may mean exchange rates are up to ~10% out of sync.

(See also my other comment about more dynamic updating below)

config/initializers/ruby_money.rb
```diff
@@ -1,8 +1,16 @@
...
+
 Money.locale_backend = :i18n
 
-eu_bank = EuCentralBank.new
-Money.default_bank = eu_bank
+mclb = Money::Bank::CurrencylayerBank.new
+mclb.access_key = ENV.fetch("CURRENCY_LAYER_API_KEY", "")
+mclb.currencylayer = true
+mclb.ttl_in_seconds = 86_400
+mclb.cache = 'rails_money_cache.txt' # To be replaced with proper cache path.
```
gregorbg, line 11: Does the gem only support file caching? Would love to use Redis for this instead

config/initializers/ruby_money.rb
```diff
@@ -1,8 +1,16 @@
...
+require 'money/bank/currencylayer_bank'
+
 Money.locale_backend = :i18n
 
-eu_bank = EuCentralBank.new
-Money.default_bank = eu_bank
+mclb = Money::Bank::CurrencylayerBank.new
+mclb.access_key = ENV.fetch("CURRENCY_LAYER_API_KEY", "")
+mclb.currencylayer = true
+mclb.ttl_in_seconds = 86_400
```
gregorbg, line 10: Where did this value come from?

config/initializers/ruby_money.rb
```diff
@@ -1,8 +1,16 @@
...
 
+require 'money/bank/currencylayer_bank'
+
 Money.locale_backend = :i18n
 
-eu_bank = EuCentralBank.new
-Money.default_bank = eu_bank
+mclb = Money::Bank::CurrencylayerBank.new
+mclb.access_key = ENV.fetch("CURRENCY_LAYER_API_KEY", "")
+mclb.currencylayer = true
```
gregorbg, line 9: What does this flag do and why is it needed? From my perspective as an outside reviewer who hasn't done any research on CurrencyLayer, this line of code reads like "Oh dear CurrencyLayer gem, which gives me a class called `CurrencylayerBank`, please _really_ **really** access CurrencyLayer by the way!".

config/initializers/ruby_money.rb
```diff
@@ -1,8 +1,16 @@
 # frozen_string_literal: true
 
+require 'money/bank/currencylayer_bank'
+
 Money.locale_backend = :i18n
 
-eu_bank = EuCentralBank.new
-Money.default_bank = eu_bank
+mclb = Money::Bank::CurrencylayerBank.new
+mclb.access_key = ENV.fetch("CURRENCY_LAYER_API_KEY", "")
```
gregorbg, line 8: This should be `AppSecrets.CURRENCYLAYER_API_KEY` instead. We have a top-level file called `app_secrets.rb` where you can see how we configure secrets access.

In dev and test, the values for these secrets are loaded from environment variables, which means placing it in `.env.development` and `.env.test` is correct. For production (and staging), you tell me the real access key through a (semi-)secure channel, and I insert it into a Vault runtime that feeds our production servers

.env.test
```diff
@@ -35,3 +35,4 @@ STAGING_OAUTH_SECRET=example-secret
 AVATARS_PUBLIC_STORAGE=local
 AVATARS_PRIVATE_STORAGE=local_private
 DUMP_HOST=https://assets.worldcubeassociation.org
+CURRENCY_LAYER_API_KEY=9d254ec37e1591eec964dac947fafa09
```
gregorbg, line 38: Are these keys (here in `test` and same in `dev`) random dummy values or are they actual access keys?

gregorbg (changes requested):

config/initializers/ruby_money.rb
danieljames-dj: Okay I didn't knew we were having redis. I couldn't get to see anywhere officially how to use it.

I tried `mclb.cache = :redis_cache_store, { url: EnvConfig.CACHE_REDIS_URL }` and I am not sure how to test if it worked.

lib/dues_calculator.rb
danieljames-dj: Including the details in the other comment.

config/initializers/ruby_money.rb
danieljames-dj: I couldn't find a proper documentation on these part, but my best guess is that if `ttl_in_seconds` is set, whenever we ask the library to do an exchange, if the previous exchange was fetched before x seconds, it will invalidate the cache and fetch the newer rates. I feel so because if this is not how it works, it doesn't make sense to have `ttl_in_seconds`. What do you think?

danieljames-dj: I just verified from the library's [source code](https://github.com/phlegx/money-currencylayer-bank/blob/1d4763654511a2355a0bd22db8e0fec7b2fa2a94/lib/money/bank/currencylayer_bank.rb#L160) that my guess is correct.

.env.test
gregorbg: We actually use access keys from testing or dummy accounts for Stripe and PayPal. Look into `.env.test`, you will see that a lot of `PAYPAL_*` variables exist. These are from "real" accounts, which means we have actual username/password to log in to PayPal dashboard. But if you do actually log in, you will see that these accounts are marked as test accounts by PayPal, and you cannot send actual, real money with them. Only sandbox money. So that's why the accounts are not real, they are only "real".

danieljames-dj: Okay, I think currencylayer doesn't have a test account as it doesn't do any sensitive things. So will it make sense to create two currencylayer accounts for WCA - one for test and one for production?

gregorbg: Hm, now that I think about it, I have another idea...
What would you think about using the `if Rails.env.test?` check to use the `EuCentralBank` in local environments and switch on the CurrencyLayer in production only?

gregorbg: You can introduce a flag in `EnvConfig` (file `env_config.rb`) where developers can switch on CurrencyLayer manually if they want, for local testing and stuff

danieljames-dj: That is a great idea, I've now implemented it. :-)

config/initializers/ruby_money.rb
```diff
@@ -1,8 +1,28 @@
...
-Money.default_bank = eu_bank
+if Rails.env.test? || Rails.env.development?
+  eu_bank = EuCentralBank.new
+  Money.default_bank = eu_bank
+else
+  mclb = Money::Bank::CurrencylayerBank.new
+  mclb.access_key = AppSecrets.CURRENCY_LAYER_API_KEY
+  mclb.currencylayer = true
+  mclb.ttl_in_seconds = 86_400
+  mclb.cache = proc.new do |payload|
```
gregorbg, line 15: Shouldn't this be `Proc.new` (note the capitalization)?

config/initializers/ruby_money.rb
```diff
@@ -1,8 +1,28 @@
 # frozen_string_literal: true
 
+require 'money/bank/currencylayer_bank'
+
 Money.locale_backend = :i18n
 
-eu_bank = EuCentralBank.new
-Money.default_bank = eu_bank
+if Rails.env.test? || Rails.env.development?
+  eu_bank = EuCentralBank.new
```
gregorbg, line 8: You should also `update_rates` the same way you're doing for the new CurrencyLayer bank below

config/initializers/ruby_money.rb
```diff
@@ -1,8 +1,28 @@
 # frozen_string_literal: true
 
+require 'money/bank/currencylayer_bank'
+
 Money.locale_backend = :i18n
 
-eu_bank = EuCentralBank.new
-Money.default_bank = eu_bank
+if Rails.env.test? || Rails.env.development?
```
gregorbg, line 7: There is `Rails.env.local?` which is exactly an equivalent of "test or dev"

gregorbg (changes requested):

config/initializers/ruby_money.rb
danieljames-dj: Oh yes, done.

config/initializers/ruby_money.rb
danieljames-dj: Oh that's nice, updated.

config/initializers/ruby_money.rb
danieljames-dj: Actually rubocop asked to replace `Proc.new` with `proc`. After re-reading again, I realized it's `proc` and not `proc.new`. I cannot test it right now as my access key has expired. Once the budget is approved, I'll test again using a new test key created with WCA's account.

danieljames-dj (commented):
Thanks @gregorbg for your review. I hope this is now fine codewise, let me know if not. I'll ping you once the budget is approved for the same so that I can test it in staging.

gregorbg (commented):
Code fine, pending valid API keys

app_secrets.rb
```diff
@@ -106,6 +107,7 @@ def vault_file(secret_name, file_path, refresh: true)
     mandatory :OIDC_SECRET_KEY, :string
     mandatory :STAGING_OAUTH_CLIENT, :string
     mandatory :STAGING_OAUTH_SECRET, :string
+    mandatory :CURRENCY_LAYER_API_KEY, :string
```
gregorbg, line 110: Just curious, why is this `mandatory` when we fall back to EuCentralBank outside of prod anyways?

danieljames-dj: You are right, it's not needed. I removed itnow.

gregorbg: ...that means you probably also don't need the `.env` sample values anymore, or do you?

danieljames-dj: Okay I thought env should be added there, my bad - it's my first encounter with env stuff. I've removed them now.

gregorbg: You need `.env.development` entries, if you actually read `AppSecrets.WHATEVER` in development. But in this case, `CURRENCY_LAYER_API_KEY` is only ever read in Prod, and never accessed in Dev or Test. So you don't need to fill values in for `.env.test` nor for `.env.development`.

gregorbg (approved):

app_secrets.rb
danieljames-dj: Got it, thanks for the info.

gregorbg:
For the record, the code is fine so I gave approval but we still need to install the credentials into Vault manually.
Hence not merging yet.

gregorbg:
Keys are registered, ready for :ship:ing

PR #10526, opened 2025-01-04
Bump `@cubing/icons` to `v3.0.4`

lgarron:
This should be a compatible replacement. I've done a basic check that the icons show up properly, and a full switch on the WCA website will help shake out potential issues.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "@babel/preset-env": "^7.26.9",
     "@babel/preset-react": "^7.26.3",
     "@babel/runtime": "^7.27.0",
-    "@cubing/icons": "^2.0.2",
+    "@cubing/icons": "^3.0.4",
     "@fullcalendar/core": "^6.1.17",
     "@fullcalendar/interaction": "^6.1.17",
     "@fullcalendar/luxon3": "^6.1.17",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1876,10 +1876,10 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@cubing/icons@npm:^2.0.2":
-  version: 2.0.2
-  resolution: "@cubing/icons@npm:2.0.2"
-  checksum: 10c0/a4c35e91e1707229beab507d4e2c97511e3518376885d2de53a6603134bb8bc5213155843589abc0df3ca734f866c5baad5e840f2df1d480ea72454c53049ac7
+"@cubing/icons@npm:^3.0.4":
+  version: 3.0.4
+  resolution: "@cubing/icons@npm:3.0.4"
+  checksum: 10c0/4985d1f1c06c24eea0f5bcb44bab8d398c7a508b9dabdf191fc77e9cfd13a05d480ca2e2ae111b0cba688a52f78f40679922104460238d7a151a964635ac8db8
   languageName: node
   linkType: hard
 
@@ -10240,7 +10240,7 @@ __metadata:
     "@babel/preset-env": "npm:^7.26.9"
     "@babel/preset-react": "npm:^7.26.3"
     "@babel/runtime": "npm:^7.27.0"
-    "@cubing/icons": "npm:^2.0.2"
+    "@cubing/icons": "npm:^3.0.4"
     "@fullcalendar/core": "npm:^6.1.17"
     "@fullcalendar/interaction": "npm:^6.1.17"
     "@fullcalendar/luxon3": "npm:^6.1.17"


app/webpacker/packs/global_styles.js
```diff
@@ -1,5 +1,5 @@
 import 'flag-icons/css/flag-icons.css';
-import '@cubing/icons';
+import '@cubing/icons/css';
```
lgarron, line 2: This is the forward-compatible CSS import path:

- https://icons.cubing.net/#npm-usage
- https://github.com/cubing/icons/blob/main/package.json#L21-L23

lgarron: urgh, this worked just fine in dev `webpack` doesn't seem to handle this properly. Lemme see if I can add a shim using a folder.

lgarron: I've tried adding a workaround in `@cubing/icons@v2.1.0-pre.3` but this is still broken.

Unfortunately, I no longer use or debug webpack and don't want to spend any time debugging webpack, so I'm not sure what to do. https://webpack.js.org/guides/package-exports/ suggest this should be working.

lgarron:
I also haven't figure out what's going on here: https://github.com/thewca/worldcubeassociation.org/blob/8290946f174944c85a5bafb79775c3ef193e0a03/app/assets/stylesheets/pdf.css.scss#L11

The string `cubing-icons` doesn't appear anywhere else in the repo, so I have no idea where that's pointing. 😳

app/webpacker/packs/global_styles.js
lgarron: Trying a rollback of this (which I would be very unhappy about, but #yolo).

```suggestion
import '@cubing/icons';
```

lgarron:
SASS is now erroring in CI without printing an error (it's just printing a warning). No idea what's going on with that. 🤷

gregorbg:
> SASS is now erroring in CI without printing an error (it's just printing warnings). No idea what's going on with that. 🤷

I'm getting something that looks very much like an error:
image

This is related to the stylesheets for the "print competition information as PDF" feature, which unfortunately predates every active member on WST, myself included.

@jonatanklosko might know more about this, or he might not. Otherwise, @viroulep is probably the person to talk to.

jonatanklosko:
No idea, but the package was already `@cubing/icons` before and the import worked, so perhaps it would work as it was?

lgarron:
> I'm getting something that looks very much like an error:

Ah, somehow I missed that? Anyhow, looks like `cubing-icons` and `@cubing/icons` both error in that location now. I don't see how the former should ever have worked, so I'm still [not sure what to do](https://en.wiktionary.org/wiki/Chesterton%27s_fence).

viroulep:
You can find the file by looking at the repository in its state when I introduced the pdf export: https://github.com/thewca/worldcubeassociation.org/blob/c15867cfe662906f65fe7d90c74a636c04cbc470/WcaOnRails/vendor/assets/stylesheets/cubing-icons.css

I have no idea what happened to this file though.

lgarron:
> You can find the file by looking at the repository in its state when I introduced the pdf export: https://github.com/thewca/worldcubeassociation.org/blob/c15867cfe662906f65fe7d90c74a636c04cbc470/WcaOnRails/vendor/assets/stylesheets/cubing-icons.css
> 
> I have no idea what happened to this file though.

Yeah, thins would make a lot more sense if this file still existed. Now that's no longer in the tree, I don't know how it's still load-bearing. 😳

gregorbg:
There's also this: https://github.com/thewca/worldcubeassociation.org/blob/8290946/app/assets/stylesheets/pdf.css.scss#L9-L11

lgarron:
@jfly poked me about this, so let me describe my understanding of the situation.

As far as I can tell, we have two problems stemming from the same line:

> I also can't figure out what's going on here:
> 
> https://github.com/thewca/worldcubeassociation.org/blob/8290946f174944c85a5bafb79775c3ef193e0a03/app/assets/stylesheets/pdf.css.scss#L11
> 
> The string `cubing-icons` doesn't appear anywhere else in the repo, so I have no idea where that's pointing. 😳

1. This line depends breaking package abstraction (e.g. violating the semantics of [conditional exports](https://nodejs.org/api/packages.html#conditional-exports)). This makes me grumbly, but it's easy to implement a hack that works.
2. The code assumes that it can load a single CSS file. `@cubing/icons` no longer publishes CSS with inline fonts, as this is an anti-pattern.

For the second point, I think the options are:

- Implement module resolution for the PDF code. Probably prohibitive for now.
- Write some sort of build step to inline the font in the CSS so it looks the same to the PDF generation code. This is trivial with something like https://esbuild.github.io/content-types/#data-url , but we don't use `esbulid` and I don't know where in the build pipeline to put it.
- Implement a compromise that explicitly loads the `.woff2` font for the PDF at a path that the CSS can consume.

lgarron:
> * Write some sort of build step to inline the font in the CSS so it looks the same to the PDF generation code. This is trivial with something like https://esbuild.github.io/content-types/#data-url , but we don't use `esbulid` and I don't know where in the build pipeline to put it.

Specifically:

```shell
mkdir temp && cd temp
npm install esbuild @cubing/icons
npx esbuild --bundle "@cubing/icons/css" --loader:.woff2=base64 > cubing-icons-with-font-inlined.css
```

Or in JS/TS:

```js
import { build } from "esbuild";

await build({
  bundle: true,
  entryPoints: ["@cubing/icons/css"],
  loader: {".woff2": "base64"}
})
```

lgarron:
I strongly disagree with this linter error:

```
Errors on modified lines:
/home/runner/work/worldcubeassociation.org/worldcubeassociation.org/app/assets/stylesheets/pdf.css.scss: line 11, col 1, error - Unexpected empty comment (scss/comment-no-empty)
```

But I'll squash the comment parts together to make it happy for now.

lgarron:
> Okay, I tried to update this in a commit. Let's see if it passes CI.

🥳

I was able to get local dev working by switching to an older commit. Currently trying to figure out how to get a PDF to display.

lgarron:
> Currently trying to figure out how to get a PDF to display.

Okay, a *lot* of things went wrong, but I was finally able to construct a competition with a schedule.

lgarron:
So, with some difficulty I've been able to test how this affects PDF.

It seems that the icons don't render *even if the font is inlined*. This is tricky to debug, as I can't actually load the relevant HTML for the PDF easily.

@viroulep, is there a way for me to serve the HTML that is used to render the PDF, so I can access it in a browsers?

lgarron:
Okay, with a lot of pain I'm able to make slow progress debugging. My current assessment is that this is not working because:

1. We're rendering PDFs `wkhtmltopdf`, which is a dead project: https://github.com/wkhtmltopdf/wkhtmltopdf
2. The rendering engine in `wkhtmltopdf` is either out of date or is not configured to render WOFF2.
3. The setup does not work without fonts inlined into the CSS.

To give an idea of how out of date that is: WebKit introduced WOFF2 support back in 2016, somewhere **8 or 9 years ago**.
(Some sources imply it might not have been available on all platforms until 2018, but that's still a long time ago.)

However, the PDF *does* seem to work with WOFF and inlined fonts.

I'm thinking of introducing a build in `@cubing/icons/css/legacy-inline-woff` that provides this, with the hope that the WCA website will not need to rely on it indefinitely. (After all, relying on `wkhtmltopdf` indefinitely will at some point become a fairly serious security concern. And we can also hopefully clean up a bunch of the bundling if development can move from `webpack` to `esbuild`.)

@gregorbg, how does that sound to you?

gregorbg:
Well as far as I can tell this "competition info PDF" is the only place where we ever rely on any server-side PDF generation library. And so it should be fairly easy :tm: to move away from `wkhtmltopdf`, which I would be happy to get done sooner rather than later.

What do cool kids use these days for PDF generation? Is there any library or backend that would make your life easier in terms of using the fonts?

lgarron:
> What do cool kids use these days for PDF generation? Is there any library or backend that would make your life easier in terms of using the fonts?

My go-to for server-side page rendering is [Playwright](https://playwright.dev/). This is a bit heavy, but it's much more likely to be evergreen and it can be isolated in a container. It's also comparatively easy to write robust scripts for.

gregorbg:
Okay, integrating Playwright is definitely possible but not trivial. Will need to sit down and play with the integration over a free weekend.

Do you want to go ahead and merge this PR in the meantime? Or do you prefer to wait until I can get around to integrating Playwright?

lgarron:
> Okay, integrating Playwright is definitely possible but not trivial. Will need to sit down and play with the integration over a free weekend.

Sounds good, feel free to crib from the `cubing.js` screenshot example:

https://github.com/cubing/cubing.js/blob/aa33f2514800634121f5c9ef4d3b50a978ecf7d1/script/bin/screenshot.ts

> Do you want to go ahead and merge this PR in the meantime? Or do you prefer to wait until I can get around to integrating Playwright?

This PR will currently render broken icons. I'd have to add the legacy hack in `@cubing/icons` first. Lemme know if you want that and I'll go ahead.

gregorbg:
Hey! I just took a stab at replacing wkhtmltopdf with something less dated, and it turns out it's not quite as easy as I thought.

Our current approach for rendering PDFs uses `wicked_pdf`, which is specifically wired up (and hot-glued) to work only with `wkhtmltopdf`, and they specifically stated that they don't have plans to change that: https://github.com/mileszs/wicked_pdf/issues/1081

The issue with migrating to another engine is that doing so would fundamentally change the way we handle PDF generation. `wicked_pdf` currently intercepts requests that specify `format: :pdf`, then calls `wkhtmltopdf` on the HTML that would normally be served by Rails, and returns the resulting PDF as output. Note that no extra request is at play here.

If we wanted to continue doing server-side rendering with Puppeteer or Playwright or whatever, we would need to provide an endpoint on our server that exposes the "raw" HTML that makes up the PDF, let Playwright capture that (**which implies an extra request**), pipe that into an output file, and then let Rails read that output file and return it as output.

Overall feels quite complicated and is definitely a long-term project to replace. Especially seeing that we're in the middle of migrating our entire frontend stack, I'd rather not divert more resources to this issue and think about a proper solution in NextJS later :tm:

TL;DR: Would you mind providing a legacy `woff` file please? :innocent:

lgarron:
@gregorbg: Thanks for giving it a go.

It sounds like the state of the ecosystem has some really gnarly security implications. I don't quite understand the subtleties of why `wicked_pdf` is able to avoid an extra request in a way that rendering using Playwright cannot, but I trust that if there's a simple resolution you would have found it.

I'd rather add a workaround for now than for the WCA to be stuck unable to update `@cubing/icons`, so I've discussed with @jfly and will add a `@cubing/icons/css/legacy-inline-woff` build and export.

gregorbg:
Okay now I actually took the time to play around with Playwright in more detail. I have managed to configure my local setup to let Playwright generate PDFs using `page.pdf` (reference: https://playwright.dev/docs/api/class-page#page-pdf)

What I've found is that even with the latest Playwright (v1.51.1 at the time of writing this comment), **WOFF2 fonts are not rendered/embedded correctly into the PDF**. I have confirmed this suspicion by leaving my Playwright setup as-is, rolling back to `@cubing/icons` version `1.1.3`, et voilà, icons are showing in the PDF as expected.

It would seem that Chromium has a problem with embedding WOFF2 fonts into PDF files. I haven't been able to find clear, well-documented sources on this, but occasional mentions like https://github.com/diegomura/react-pdf/issues/334 and https://peterdev.pl/how-to-set-a-font-in-pdf/ do seem to confirm this observation.

In essence, even implementing a modern Playwright stack won't give us `@cubing/icons` v3 support (with PDF rendering). After a little bit of "anecdotal googling", I am under the impression that custom icons these days expose their own SVG for others to consume, especially in React: https://github.com/react-icons/react-icons and https://www.chakra-ui.com/docs/components/icon#custom-svg both demonstrate how this is done in a headless way.

Seeing that we're going to migrate our whole stack to React anyways, exporting the SVGs directly may be worth a shot. What do you think?

For reference, I am attaching two PDFs. Both have been generated by the exact same Playwright setup, only difference is that one uses `@cubing/icons@1.1.3` and the other one uses it `@3.0.4`.

[Playwright_Cubing1-1-3.pdf](https://github.com/user-attachments/files/19547714/Playwright_Cubing1-1-3.pdf)
[Playwright_Cubing3-0-4.pdf](https://github.com/user-attachments/files/19547715/Playwright_Cubing3-0-4.pdf)

gregorbg:
> why `wicked_pdf` is able to avoid an extra request in a way that rendering using Playwright cannot

`wicked_pdf` is based on a locally installed binary executable of `wkhtmltopdf`, which is able to ingest HTML source code as a string (or by piping it in from a file) and turning that into a PDF. No network request required.

Playwright on the other hand runs its own browser which needs to "visit" the webpage it wants to turn into a PDF. So the user makes a request to `MyCompetition2020.pdf`, and that prompts Playwright to "browse" an internal route which it can then turn into a PDF.

I have been able to optimize this to a degree where Rails renders the template into a string and I pass that to the Playwright browser directly. But this still leaves us with some "marginal" network requests for loading the stylesheets and image resources. You _could_ inline those, but you'd have to make some strict assumptions about which container/cluster you're running on, where to load these files from, etc.

`wicked_pdf` was able to skip this because it assumed that all necessary assets are present on-disk. Playwright would be a nice chance to get rid of these assumptions, at the cost of an extra network request.

gregorbg:
Okay nevermind, it's Rails :sweat_smile: Loading the icons from your CDN via `<link rel="stylesheet" href="https://cdn.cubing.net/v0/css/@cubing/icons/css">` does work in the PDF export... Sorry that testing this out hadn't occured to me earlier!

On a sidenote, I'd still be very interested in getting SVG exports for React, but I appreciate that this is a completely different topic and should go under a separate issue.

lgarron:
> What I've found is that even with the latest Playwright (v1.51.1 at the time of writing this comment), **WOFF2 fonts are not rendered/embedded correctly into the PDF**. I have confirmed this suspicion by leaving my Playwright setup as-is, rolling back to `@cubing/icons` version `1.1.3`, et voilà, icons are showing in the PDF as expected.
> 
> It would seem that Chromium has a problem with embedding WOFF2 fonts into PDF files. I haven't been able to find clear, well-documented sources on this, but occasional mentions like [diegomura/react-pdf#334](https://github.com/diegomura/react-pdf/issues/334) and https://peterdev.pl/how-to-set-a-font-in-pdf/ do seem to confirm this observation.

Well, that's disconcerting. Thanks for the research on that. 😳

I suppose WOFF is only marginally larger than WOFF2. The same can't be said of `.eot` or `.ttf`, so that really only leaves us with `.woff` as a "good" choice for PDF support.

If PDF generation requires it for the time being, I think it's reasonable for the package to include `.woff` as a fallback to `.woff2` — with a proper bundler this should still serve `.woff2` efficiently to clients that support it. Do you have an easy way to confirm whether Playwright will automatically load the `.woff` as a fallback when rendering PDFs?

For the short term, it sounds like we still need the legacy inline WOFF, so I'll keep working on that.

> After a little bit of "anecdotal googling", I am under the impression that custom icons these days expose their own SVG for others to consume, especially in React: https://github.com/react-icons/react-icons and https://www.chakra-ui.com/docs/components/icon#custom-svg both demonstrate how this is done in a headless way.
> 
> Seeing that we're going to migrate our whole stack to React anyways, exporting the SVGs directly may be worth a shot. What do you think?

If there is an ecosystem-portable way to do this, I'm open to it!

(I would be strongly against anything React-specific, but perhaps there are advances in server-compatible conventions I don't know about.)

lgarron:
> Okay nevermind, it's Rails 😅 Loading the icons from your CDN via `<link rel="stylesheet" href="https://cdn.cubing.net/v0/css/@cubing/icons/css">` does work in the PDF export... Sorry that testing this out hadn't occured to me earlier!

Ah, so WOFF2 should be fine for Playwright?

gregorbg:
Yup! And I even found a way to get rid of the hack that augmented the Rails assets path... Here's a PDF built with Playwright and `@cubing/icons` v3 :tada: 

[XianCherryBlossom2025_en_2025-03-11T12:03:22Z_false.pdf](https://github.com/user-attachments/files/19548399/XianCherryBlossom2025_en_2025-03-11T12.03.22Z_false.pdf)

gregorbg:
Opening a PR for this is going to take a hot moment though, because we now have to consider how to best integrate Playwright into our existing infrastructure.

lgarron:
> Yup! And I even found a way to get rid of the hack that augmented the Rails assets path... Here's a PDF built with Playwright and `@cubing/icons` v3 🎉

Quite some timing. I just got done adding the implementing `legacy-inline-woff` build. 😆

> Opening a PR for this is going to take a hot moment though, because we now have to consider how to best integrate Playwright into our existing infrastructure.

Would you prefer that to the `legacy-inline-woff` build?

I don't know if I can commit much time to helping, but I'd certainly be glad to support it if I can.

gregorbg:
Playwright is now running productively. This also means that the PDF styles and assets are delivered the "normal" way through Webpacker now, and this whole PR should boil down to a simple version bump in `package.json`

gregorbg (approved):

lgarron:
@gregorbg I didn't have a chance to check the merged code in dev yet, is it working properly?

gregorbg:
Yes, I tested it locally before approving, and after deployment I was also able to confirm that it works on the prod system as expected.

For reference, I just dropped the cached PDF on `OldenburgOsterOpen2025` and regenerated it, you can see the result at https://www.worldcubeassociation.org/competitions/OldenburgOsterOpen2025.pdf

PR #10527, opened 2025-01-04
Improve/fix registration status icon text

kr-matthews:
This competition is in just over a month. Previously it also said "in 1 month".

image

This competition is tomorrow. Previously it would have said "in 2 hours".

image

Unfortunately Luxon doesn't seem to be using the locale, so that will need to be fixed if possible (in another PR). Also, other languages have the extra "in" too, it seems (and missing periods).

image


kr-matthews:
Actually these strings are also used in the 'all competitions' page (both the old and new UI), so this PR just shifts the extra 'in' problem. However, that page is correctly translating the 'days' message. So we probably want a consistent approach between the two - using luxon in both (with correct locale), or explicitly calculating the number of days and directly translating that (note this never would show "1 month", it's always days). For the former approach, I'm not sure how to make those changes in ruby, so someone else would have to pick this up.

app/webpacker/components/MyCompetitions/UpcomingCompetitionTable.jsx
```diff
@@ -23,11 +23,18 @@ const competingStatusIcon = (competingStatus) => {
...
+  const toRelativeOptions = {
+    // don't be more precise than "days" (i.e. no hours/minutes/seconds)
+    unit: ["years", "quarters", "months", "weeks", "days"],
+    // round up, e.g. in 8 hours -> pads to 1 day 8 hours -> rounds to "in 1 day"
+    padding: 24 * 60 * 60 * 1000,
+  };
+
   if (competition.registration_status === 'not_yet_opened') {
-    return I18n.t('competitions.index.tooltips.registration.opens_in', { duration: DateTime.fromISO(competition.registration_open).toRelative() });
+    return I18n.t('competitions.index.tooltips.registration.opens_in', { duration: DateTime.fromISO(competition.registration_open).toRelative(toRelativeOptions) });
```
gregorbg, line 34: I think in terms of Luxon respecting the locale, it's (supposed to be) as simple as passing the locale ISO code to `fromISO` like so: `DateTime.fromISO(dateIsoString, { locale: 'fr' })`. Cf. https://moment.github.io/luxon/#/intl?id=setting-the-locale

We expose the currently seleted locale as `I18n.locale` on the same I18n object that also does `t` and `tArray`.

app/webpacker/components/MyCompetitions/UpcomingCompetitionTable.jsx
```diff
@@ -23,11 +23,18 @@ const competingStatusIcon = (competingStatus) => {
...
   }
   if (competition.registration_status === 'past') {
-    return I18n.t('competitions.index.tooltips.registration.closed', { days: DateTime.fromISO(competition.start_date).toRelative() });
+    return I18n.t('competitions.index.tooltips.registration.closed', { days: DateTime.fromISO(competition.start_date).toRelative(toRelativeOptions) });
```
gregorbg, line 37: See above for passing locale information to Luxon

app/webpacker/components/MyCompetitions/UpcomingCompetitionTable.jsx
gregorbg: Interestingly, even the `toRelativeOptions` object supports a `locale` key.

gregorbg (commented):
Added some comments about locale handling

gregorbg:
About the handling of `in`: This is a "devil or the deep blue sea" scenario. Luxon does have `luxonDt.diffNow().toHuman()` method, which (with a little bit of fiddling in terms of passing in the correct `config` options) would boild down to `3 days, 8 hours` without the "in".

However, for most languages there's a notable difference because of grammar. In German for example, if you just enumerate the time units in a "raw" fashion, it's `6 Monate, 3 Tage` (six months and three days). In case of a relative time difference, as in "the event happens **in 6 months and 3 days**" you'd have to change the declension to `in 6 Monate**_n_** und 3 Tage**_n_**`.

So I personally feel like it's best to let the Luxon library handle the "in" and "ago" and nasty declensions, and remove them from our translatable I18n keys like you did here.

app/webpacker/components/MyCompetitions/UpcomingCompetitionTable.jsx
kr-matthews: Fixed by a separate PR.

kr-matthews:
Another PR fixed the wrong locale issue.
I've updated this PR to take that into account. And instead of 'fixing' the existing strings, I've created new ones. Because the existing strings are also used elsewhere, and there the "in" is required. Alternatively, we could update those instances to handle "in" differently and not introduce any new strings.

gregorbg:
> Alternatively, we could update those instances to handle "in" differently and not introduce any new strings.

I would prefer that, if it's not too much to ask. When you get started and you realize you're wasting more than say ~45 minutes and you have to dig into the source code of some old. deprecated Ruby library, then stop what you're doing, post another comment here and I'll approve the additional transations :wink:

kr-matthews:
> > Alternatively, we could update those instances to handle "in" differently and not introduce any new strings.
> 
> I would prefer that, if it's not too much to ask. When you get started and you realize you're wasting more than say ~45 minutes and you have to dig into the source code of some old. deprecated Ruby library, then stop what you're doing, post another comment here and I'll approve the additional transations 😉

The only ruby use is in the old competitions overview code. We recently switched over to the new react version. Can we just remove that old code now? I'm under the impression we won't be switching back to it again.

kr-matthews:
I reverted to the existing strings and updated them. The old competitions overview ruby code will be wrong, but it doesn't matter since it's being removed and can't be seen in the meantime anyway.

I also adjusted it so that for registration open it is precise (since it's a date time), but for competition start the precision is at most days (since it's a date, not a date time).

wst-c
wst-b
wst-1
wst-a
wst-A
wst-B
wst-C
wst-D
wst-3
wst-2

gregorbg (approved):

kr-matthews:
Can we merge this? Or are we waiting for the other PR which removes the old code, to be safe?

gregorbg:
I went ahead and merged the other PR, so this is now good to :ship:

PR #10537, opened 2025-01-06
Just fixing capitalization

JC01010:


dunkOnIT (changes requested):
Good instinct to change it, I'd go further and say this is a bad place for contraction. 

Suggested change: "I DO NOT want to specify a competitor limit"

gregorbg (approved):
Approving based on Duncan's previous comment

PR #10538, opened 2025-01-06
Show notification bubble in new approveAvatars panel

gregorbg:
Porting over the small "bubble" that showed how many were pending.

The backend code looks a bit complicated, but the idea is this: There should be a method generating "notification count" for a given panel page. But if I do this like `{page1: User.count, page2: Competition.count, page5: AwesomeStuff.joins(...).where(...).count}`, it means a lot of useless things will be computed when the user only has access to `page1`.
So instead of returning the counts, I will return the _instruction_ how to compute the count, then filter that for the panels which the user has access to, then compute.

I'm all ears for better ideas @danieljames-dj

### DIFF
--- a/app/controllers/panel_controller.rb
+++ b/app/controllers/panel_controller.rb
@@ -22,6 +22,10 @@ class PanelController < ApplicationController
     panel_details = User.panel_list[@panel_id.to_sym]
     @pages = panel_details[:pages]
     @title = panel_details[:name]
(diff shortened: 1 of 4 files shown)


danieljames-dj (approved):
This looks great. I cannot think of a better idea.

`User.panel_notifications.slice(*@pages).transform_values(&:call)` - this was amazing. My solution would have been to do some switch case and probably would have made it more complicated. I liked your solution better.

Issue #10522, opened 2025-01-03
Boxes in the Manage events tab i too small


AxelFl:
**Describe the bug**
Boxes in the mange events tab are too small to show the entire drop down menu and requires scrolling where I believe none should be necessary. This is more QOL, but should be possible to fix.

**To Reproduce**
1. Go to 'Manage events' tab in a competition
2. Click on a "Cutoff" or "To advance" condition
3. Click on "Cutoff Format" or "Type"

**Expected behavior**
I think it would be more helpful if the entire drop down could be seen.

**Screenshots**
<img width="668" alt="Screenshot 2025-01-03 at 18 04 34" src="https://github.com/user-attachments/assets/09b8090c-273c-477d-a9aa-3f347ebe56fa" />
<img width="635" alt="Screenshot 2025-01-03 at 18 04 29" src="https://github.com/user-attachments/assets/765f57b4-61f6-4e86-9afc-0080d5a50acf" />

**Desktop (please complete the following information):**
 - OS: macOS 15.0
 - Browser: Firefox
 - Version 133.0.3 (aarch64)

dunkOnIT:
Adding the title from #10338 here as they're the same issue but you've documented it better: "Events menu drop-down boxes not scrollable on mobile, and the panel could be larger overall on all platforms (so that scrolling isn't necessary)"

PR #8765, opened 2024-01-24
Add Scrolling Functionality to Cumulative Selection

Carter-Thomas:
(#8634) Had to delete my branch that initially had this change due to unrelated merge, should be correct now.

PR #10544, opened 2025-01-06
Let dropdowns expand beyond Modal body in EditEvents

gregorbg:
Really quick fix for #10522, although I'm not sure whether this breaks anything else. I did a few basic sanity checks, but I'm sure experienced Delegates have other use-cases for this panel...

Git Blame shows that we introduced the `scrolling` property only in #8765, which was targeted towards a specific issue with scrolling in a long list of cumulative rounds. I have changed the code so that each nested modal can decide by itself whether it wants to be scrolling or not, although the new default value is `false` (as opposed to before. when _every_ modal was `scrolling` by default, which caused the issue with dropdowns)


kr-matthews:
This is also an issue with editing advancement conditions, so the same changes should be applied there. But even if I add `contentScrolling` there, it still looks like this:

image

image

gregorbg:
The point is that *enabling* `contentScrolling` is what *causes* the dropdown glitch. In a sense, enabling the scrollbar is precisely what "swallows" the dropdown.

The solution is to make sure that it is *disabled* for popups where you don't want this problem to happen. That's why the default value is set to `false`, and you should only enable it at all on dropdowns that have a lot of content (like the radio buttons in the cumulative cutoff window).

kr-matthews (approved):
Oops, yeah I misunderstood. Looks good.

PR #10370, opened 2024-12-04
Migrate the My Competition Table to React

FinnIckler:
Much cleaner implementation of my PR a year ago (can't believe it's been that long)

image

PR #10545, opened 2025-01-06
Improving mobile view of new "My competitions" table

davidkaralli:
* Added unstackable and singleLine attributes

### DIFF
--- a/app/webpacker/components/MyCompetitions/UpcomingCompetitionTable.jsx
+++ b/app/webpacker/components/MyCompetitions/UpcomingCompetitionTable.jsx
@@ -66,80 +66,82 @@ export default function UpcomingCompetitionTable({
   }
 
   return (
-    <Table basic>
-      <Table.Header>
-        <Table.Row>
-          { shouldShowRegistrationStatus && <Table.HeaderCell collapsing /> }
-          <Table.HeaderCell>
-            {I18n.t('competitions.competition_info.name')}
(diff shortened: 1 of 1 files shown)


gregorbg:
Thank you for your contribution! We had initially decided against this (see https://github.com/thewca/worldcubeassociation.org/pull/10370#issuecomment-2548508730) and there were no significant complaints or uproar. So I'd like to understand your intention a bit more, what's your motivation for this change?

davidkaralli:
Hi Gregor, thanks for reviewing! Currently, the "My competitions" table is confusing and difficult to navigate on mobile. I think the reason there haven't been complaints until now is that the new table wasn't deployed until the past day? (I'm not sure when updates to the WCA website are deployed, but no one I'm aware of noticed this until a few hours ago.)

As for complaints from users, here's a comment from Calvin Nielson on Cyoubx's Friends echoing a similar sentiment. I also had some fellow delegates on a voice call with similar views about the new table.

Screenshot 2025-01-06 012840

Jason Chang also noticed that the registration status icon (first column) has been clipped into what looks like an out-of-place hyphen (see the attached image), which isn't reflected in the image in the original pull request. Although IMO, fixing this detail alone wouldn't fix the primary problem here, which is that this isn't how users expect a table to look.

IMG_3800

FinnIckler (approved):
I will take on the rest of the tasks

PR #10547, opened 2025-01-06
Don't show delegate Report Icons for Organizers

FinnIckler:
Introduces a new `can_view_delegate_report` permission to accomplish this.

Bonus fix: the registration status text was not being shown because it was still doing the pre-refactor `registrationStatus?.competing_status`


app/models/user.rb
```diff
@@ -751,6 +751,9 @@ def permissions
       can_view_delegate_admin_page: {
         scope: can_view_delegate_matters? ? "*" : [],
       },
+      can_view_delegate_report: {
+        scope: can_view_delegate_matters? ? "*" : delegated_competitions.pluck(:id),
```
gregorbg, line 755: This won't quite achieve what you want, because literally any Delegate will have `can_view_delegate_matters?` return the value `true`. So in the version as it currently stands, any Delegate will have permission to access any and all Delegate reports

gregorbg: Also, I think the `has_many :delegated_competitions` automagically gives you `delegated_competition_ids` (but only 90% sure, please confirm)

gregorbg (changes requested):

app/models/user.rb
gregorbg: Yes okay, they have _permission_ to view all old Delegate Reports. But how useful will it be for a Delegate to see a report of an event they only attended as a competitor?

I might be overthinking things here, how did the old Bootstrap version of the UI determine whether to display this icon?

FinnIckler: it uses that exact same method

gregorbg: Actually, the snippet that you posted reveals an important distinction: The old UI determined whether to show the report or not using the `can_view_delegate_report?` method that you quoted. This method takes into account whether or not the report is posted at the time of calling the method!

Your approach in this PR just uses a flat-out "can this user view Delegate matters", which returns a flat-out `true` for any Delegate. So in the old version, a logged-in Delegate would _not_ see icons for unposted reports for other comps. In your suggested fix, they do.

gregorbg: I think the suggestion here might be to introduce more granular permissions, like the (old) method names `can_view_delegate_report?` and `can_edit_delegate_report?` imply,

app/views/competitions/_my_competitions_table.html.erb
gregorbg, line 1: This table is still used in `for_senior.rb`, you cannot delete it yet.

gregorbg (changes requested):

app/views/competitions/_my_competitions_table.html.erb
FinnIckler: can't I just make that render the my competitions table?

PR #10548, opened 2025-01-06
Migrate mysql to 8.4 in Docker compose

FinnIckler:
Currently Upgrading staging but it seems to be working (staging replica already updated). Locally we need to add `--mysql-native-password=ON` as it's now deprecated and will be removed in version 9.

We can migrate to the new `caching_sha2_password` but that means every developer will need to recreate their database.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -108,11 +108,13 @@ services:
         condition: service_completed_successfully
 
   wca_db:
+    # You will have to run docker exec -it database mysql to connect to the database and then
+    # ALTER USER 'root'@'%' IDENTIFIED WITH caching_sha2_password BY '';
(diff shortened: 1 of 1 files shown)


FinnIckler:
Another thing is that https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md currently uses mysql 8.0.40 so we could consider switching to another gh action that allows us to control mysql versions like https://github.com/marketplace/actions/actions-setup-mysql

gregorbg:
I am very cautious about using core software setup steps that are not maintained by the official vendors.

For Ruby, there's a GH Action made by the Ruby folks. Same for NPM/Yarn. By any chance, is there an official MySQL or MariaDB action?

gregorbg (approved):

PR #10550, opened 2025-01-07
Fix psych sheet footer

kr-matthews:
[Screencast from 2025-01-06 04:47:37 PM.webm](https://github.com/user-attachments/assets/c24f37f7-6c93-4e5d-b160-863969bbbddc)

The last commit is moving code, so might be good to look at the diff without that.

### DIFF
--- a/app/webpacker/components/RegistrationsV2/Registrations/RegistrationList.jsx
+++ b/app/webpacker/components/RegistrationsV2/Registrations/RegistrationList.jsx
@@ -24,61 +24,6 @@ import { EventSelector } from '../../CompetitionsOverview/CompetitionsFilters';
 
 const sortReducer = createSortReducer(['name', 'country', 'total']);
 
-function FooterContent({
-  registrations, psychSheetEvent, competitionInfo,
-}) {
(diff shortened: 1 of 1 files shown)


coder13:
Does it make sense to show "0 first-timers + 18 returners = 18 people" on the psych sheet pages or just the competitors list?

For the psych sheet, we could show "X / Y competitors" if we want to show something.

kr-matthews:
> Does it make sense to show "0 first-timers + 18 returners = 18 people" on the psych sheet pages or just the competitors list?
> 
> For the psych sheet, we could show "X / Y competitors" if we want to show something.

I'm not sure what would be best. There's another issue about this page looking bad on mobile, and that's in part because this footer text makes the first column too wide. So I would save this for a discussion & separate PR.

gregorbg (approved):

gregorbg:
> I'm not sure what would be best. There's another issue about this page looking bad on mobile, and that's in part because this footer text makes the first column too wide. So I would save this for a discussion & separate PR.

Fine by me, feel free to merge at your own discretion.

PR #10551, opened 2025-01-07
Move event selector to components/wca

kr-matthews:
Since it's used everywhere, not just in `CompetitionsOverview`.


gregorbg (approved):
Thanks, long overdue!

PR #10569, opened 2025-01-07
Migrate Rankings Table to React

FinnIckler:
We completely forgot about this one!
image
This already works fine if we just want to get rid of the bootstrap table. But I think we should also just invest the time and fix up the filters to get the speed boost from not rerendering the page


FinnIckler:
image
I added the functionality of filters and loading the data from the frontend. 
Styling for the filters are still TODO though

FinnIckler:
I think removing the segment makes sense
image

FinnIckler:
Screenshot 2025-01-10 at 14 58 47
Styled the filters based on the competitions page

FinnIckler:
Screenshot 2025-01-10 at 16 33 35
Here is the by region view

FinnIckler:
I think that means I am done?

app/webpacker/components/wca/EventSelector.jsx
kr-matthews, line 1: Looks like only indentation changes, and the indentation was correct before.

app/webpacker/lib/requests/routes.js.erb
```diff
@@ -1,3 +1,4 @@
+import {countries} from "../wca-data.js.erb";
```
kr-matthews, line 1: Doesn't look like this is used.

app/webpacker/components/Results/Rankings/index.jsx
```diff
@@ -0,0 +1,74 @@
...
+}
+
+export function Rankings({
+  initialEvent, initialRegion, initialRankingType, initialGender, initialShow,
+}) {
+  const [event, setEvent] = useState(initialEvent);
+  const [region, setRegion] = useState(initialRegion ?? 'all');
+  const [rankingType, setRankingType] = useState(initialRankingType);
+  const [gender, setGender] = useState(initialGender);
+  const [show, setShow] = useState(initialShow ?? 'Persons');
```
kr-matthews, line 35: This feels a bit off. Generally the only time you might want to pass an initial state and then create a new independent state is when there's some kind of form where the user can change things, I think.

Ideally, either the back-end is providing the (up-to-date) state, removing the need for these `useState`s, or the front-end is handling everything, removing the need for the `initial...` proprs in `Rankings`.

app/webpacker/components/Results/Rankings/index.jsx
```diff
@@ -0,0 +1,74 @@
...
+
+  const filterState = useMemo(() => ({
```
kr-matthews, line 37: A reducer might be good here? Then only one setter/callback needs to be passed to the `ResultsFilter` component.
Setters aren't really 'state' so it feels off to include them in `filterState`.

app/webpacker/components/Results/Rankings/RankingsTable.jsx
```diff
@@ -0,0 +1,125 @@
...
+  );
+}
+
+function ResultRow({
+  result, competition, rank, isAverage, show, country,
+}) {
+  const attempts = [result.value1, result.value2, result.value3, result.value4, result.value5];
+  const bestResult = _.max(attempts);
+  const worstResult = _.min(attempts);
+  const bestResultIndex = attempts.findIndex((a) => a === bestResult);
```
kr-matthews, line 26:
```suggestion
  const bestResultIndex = attempts.indexOf(bestResult);
```
and the line below

app/webpacker/components/Results/Rankings/RankingsTable.jsx
```diff
@@ -0,0 +1,125 @@
...
+      {show === 'by region' ? <CountryCell country={country} />
+        : <Table.Cell textAlign="center">{rank}</Table.Cell> }
+      <Table.Cell>
+        <a href={personUrl(result.personId)}>{result.personName}</a>
+      </Table.Cell>
+      <Table.Cell>
+        {formatAttemptResult(result.value, result.eventId)}
+      </Table.Cell>
+      {show !== 'by region'
+      && <CountryCell country={country} />}
```
kr-matthews, line 39:
```suggestion
      {show !== 'by region' && <CountryCell country={country} />}
```
or
```suggestion
      {show !== 'by region' && (
          <CountryCell country={country} />
      )}
```
would be standard formatting. I think we should improve our linter. The consecutive `))` `)}` before the table row ends is also hard to read.

app/webpacker/components/Results/Rankings/RankingsTable.jsx
```diff
@@ -0,0 +1,125 @@
...
+        country = { name: I18n.t('results.table_elements.world') };
+      } else if (index >= firstContinentIndex && index < firstCountryIndex) {
+        country = continents.real.find((c) => c.id === country.continentId);
+      }
+
+      previousValue = value;
+      previousRank = rank;
+
+      return (
+        <ResultRow
```
kr-matthews, line 87: JSX isn't usually produced inside a `useMemo`. I'm not entirely sure why.
I would memoize an array of row data objects, rather than an array of `<ResultsRow>`s, and map over that in the `RankingsTable` return (calling `<ResultRow>`s there).

app/webpacker/components/Results/Rankings/RankingsTable.jsx
```diff
@@ -0,0 +1,125 @@
...
+    let rowsToMap = rows;
+    let firstContinentIndex = 0;
+    let firstCountryIndex = 0;
+    if (show === 'by region') {
+      [rowsToMap, firstContinentIndex, firstCountryIndex] = rows;
+    }
+
+    let previousValue = 0;
+    let previousRank = 0;
+    return rowsToMap.map((result, index) => {
```
kr-matthews, line 70: This isn't very React-like (i.e. using `let`). I would use `Array.reduce`. Instead of maintaining `previousRank` etc., check whether the previous entry in the accumulator has the same value; if so, use the rank from that entry, otherwise use the size of the accumulator. It's possible that that might end up being more awkward or harder to understand (I'm not entirely sure how the countries/continents are being handled), but usually it's the better option.

app/webpacker/components/Results/resultsFilter.jsx
```diff
@@ -0,0 +1,86 @@
...
+  } = filterState;
+  const regionIso2 = useMemo(() => {
+    if (region === 'world') {
+      return 'all';
+    }
+    const iso2 = countries.real.find((country) => country.id === region)?.iso2;
+    if (iso2) {
+      return iso2;
+    }
+    return region;
```
kr-matthews, line 31:
```suggestion
    return countries.real.find((country) => country.id === region)?.iso2 || region;
```

app/webpacker/components/Results/resultsFilter.jsx
```diff
@@ -0,0 +1,86 @@
...
+  }, [region]);
+  return (
+    <Segment raised>
+      <Form>
+        <Form.Field>
+          <EventSelector
```
kr-matthews, line 37: Maybe we should add a new prop `title` with default value `"Events"`, and use `title="Event"` here, since you can only select one.

If so, it would make sense to use the title `"Event"` in the psych sheet too.

Alternatively, we could automatically make the title `"Event"` if the 'All' button is hidden?

app/webpacker/components/Results/resultsFilter.jsx
```diff
@@ -0,0 +1,86 @@
...
+            selectedEvents={[event]}
+            onEventSelection={({ eventId }) => setEvent(eventId)}
+            hideAllButton
+            hideClearButton
+          />
+          <RegionSelector
```
kr-matthews, line 43: This component should be moved to `components/wca` (maybe in separate PR; the `dispatchFilter` prop should probably be changed to `onChange`, since you don't need a reducer/dispatcher to use this reusable component).

app/webpacker/components/Results/resultsFilter.jsx
```diff
@@ -0,0 +1,86 @@
...
+                {I18n.t('results.selector_elements.type_selector.single')}
+              </Button>
+              <Button active={rankingType === 'average'} onClick={() => setRankingType('average')}>{I18n.t('results.selector_elements.type_selector.average')}</Button>
+            </ButtonGroup>
+          </Form.Field>
+          {/* <Form.Field width={1}> */}
+          {/*   <ButtonGroup> */}
+          {/*    <Button>All years</Button> */}
+          {/*   </ButtonGroup> */}
+          {/* </Form.Field> */}
```
kr-matthews, line 65: Remove?

kr-matthews (changes requested):
It takes about a minute to load the data when I change any filters. Is that just a running-locally problem?

app/webpacker/components/Results/Rankings/index.jsx
FinnIckler: Well the backend is providing the data read from the URL. We could parse the URL from the frontend I guess

app/webpacker/components/Results/resultsFilter.jsx
FinnIckler: I kept this in as a reminder to include it when we bring back the year filter

FinnIckler:
> It takes about a minute to load the data when I change any filters. Is that just a running-locally problem?

Yes and no. It is quite slow to compute rankings always, which is why we precompute the pages in prod after every cad, but current local state makes it about 5 * worse because of the Debugging output and it also currently serializes the competition object very expensively

dunkOnIT:
@thewca-bot deploy staging

FinnIckler:
@thewca-bot deploy staging

FinnIckler:
> > It takes about a minute to load the data when I change any filters. Is that just a running-locally problem?
> 
> Yes and no. It is quite slow to compute rankings always, which is why we precompute the pages in prod after every cad, but current local state makes it about 5 * worse because of the Debugging output and it also currently serializes the competition object very expensively

Just improved the speed by 10x

FinnIckler:
fixed the mobile issues

FinnIckler:
I think that the region -> persons issue might be related to the useMemo jsx thing. At least I don't understand how it's happening because the data is correct, so it must be a jsx rendering issue

dunkOnIT:
@thewca-bot deploy staging

FinnIckler:
@thewca-bot deploy staging

app/webpacker/components/Results/Rankings/index.jsx
```diff
@@ -8,44 +10,85 @@ import Loading from '../../Requests/Loading';
...
-    setRegion,
-    rankingType,
-    setRankingType,
-    gender,
-    setGender,
-    show,
-    setShow,
-  }), [event, gender, rankingType, region, show]);
+  // Use the reducer
+  const [filterState, dispatch] = useReducer(filterReducer, initialState);
```
kr-matthews, line 76: You can remove the `initialState` const and just do
```suggestion
  const [filterState, dispatch] = useReducer(filterReducer, window.location.href, parseInitialStateFromUrl);
```
The value will only be calculated on initial render, so it's `memo`ed for free.

app/webpacker/components/Results/Rankings/index.jsx
```diff
@@ -8,44 +10,85 @@ import Loading from '../../Requests/Loading';
...
+
+  const filterActions = useMemo(
```
kr-matthews, line 78: I was thinking you could instead pass `dispatch` to `ResultsFilter`, and directly use e.g. `dispatch({ type: ActionTypes.SET_EVENT, payload: event })` in the `onEventSelection` prop of `EventSelector`.
But maybe the current way is better because then `ResultsFilter` doesn't need to know about the reducer?

app/webpacker/components/Results/Rankings/index.jsx
kr-matthews: I think it's worth it for the consistency, yeah.

app/webpacker/components/Results/Rankings/RankingsTable.jsx
```diff
@@ -18,15 +18,16 @@ function CountryCell({ country }) {
 }
 
 function ResultRow({
-  result, competition, rank, isAverage, show, country,
+  result, competition, rank, isAverage, show, country, key,
```
kr-matthews, line 21: I'm pretty sure this isn't how to use keys. The key should just be provided to `ResultRow`, but not explicitly used in `ResultRow`.

kr-matthews: Yeah, you _cannot_ access a prop named `key`; it will always be undefined. It's a reserved name, basically.

kr-matthews: > I fixed that issue by setting <Table.Body key={show}> to force a rerender of the table when the show mode is changed. @kr-matthews if you want to investigate this more feel free, but I spend an hour and that was the only thing that worked

The issue is that you have duplicate keys, specifically the `result.id` `6350611` is coming up multiple times, both when 'show' is 'results' and 'by region'. For the latter, I believe this is because the WR is also the NAR and the "USR", so that key comes up 3 times. I'm not sure why it's coming up twice for 'results'.

app/webpacker/components/Results/Rankings/RankingsTable.jsx
```diff
@@ -57,69 +57,80 @@ function ResultRow({
...
   rows, competitionsById, isAverage, show,
 }) {
-  const r = useMemo(() => {
+  const results = useMemo(() => {
     let rowsToMap = rows;
     let firstContinentIndex = 0;
     let firstCountryIndex = 0;
     if (show === 'by region') {
       [rowsToMap, firstContinentIndex, firstCountryIndex] = rows;
     }
```
kr-matthews, line 66: I would do something like this to remove these `let`s too, but it's somewhat personal preference.
```suggestion
    const isByRegion = show === 'by region';
    const rowsToMap = isByRegion ? rows[0] : rows;
    const firstContinentIndex = isByRegion ? rows[1] : 0;
    const firstCountryIndex = isByRegion ? rows[2] : 0;
```
If the value can be calculated directly, then it's usually clearer, and removes the possibility of a future PR updating the `let` in an obscure/unexpected place. And if you can't directly calculate the value, then it's usually because you're using a loop instead of a reducer/map, as was the case below.

kr-matthews: Or if you prefer,
```suggestion
    const isByRegion = show === 'by region';
    const [rowsToMap, firstContinentIndex, firstCountryIndex] = isByRegion ? rows : [rows, 0, 0];
```

kr-matthews: For clarity, it might help to split the region case out into a separate case/helper function. Since the rank stuff is only used for persons/results, and the country/continent stuff is only used for by region. Not sure if it's worth it at this point though.

app/webpacker/components/Results/Rankings/RankingsTable.jsx
```diff
@@ -18,15 +18,16 @@ function CountryCell({ country }) {
...
+    .filter(Boolean);
   const bestResult = _.max(attempts);
   const worstResult = _.min(attempts);
-  const bestResultIndex = attempts.findIndex((a) => a === bestResult);
-  const worstResultIndex = attempts.findIndex((a) => a === worstResult);
+  const bestResultIndex = attempts.indexOf(bestResult);
+  const worstResultIndex = attempts.indexOf(worstResult);
   return (
-    <Table.Row>
+    <Table.Row key={key}>
```
kr-matthews, line 30:
```suggestion
    <Table.Row>
```

app/webpacker/components/Results/Rankings/RankingsTable.jsx
```diff
@@ -57,69 +57,80 @@ function ResultRow({
...
+            <>
+              <Table.HeaderCell>{I18n.t('results.table_elements.solves')}</Table.HeaderCell>
+              <Table.HeaderCell />
+              <Table.HeaderCell />
+              <Table.HeaderCell />
+              <Table.HeaderCell />
+            </>
+          )}
+        </Table.Header>
+        <Table.Body key={show}>
```
kr-matthews, line 119:
```suggestion
        <Table.Body>
```

kr-matthews (changes requested):

dunkOnIT:
Some more issues: 
- [ ] (fig 1) Have to scroll horizontally to see the results on mobile - we should be able to see at least the name and the time achieved. One option is just making the table wider (see Fig 2 for reference to old table)
- [ ] Multi BLD average button shows the round info from when the person achieved the MBLD single oop

Fig 1: 
image

Fig 2: 
image

app/webpacker/components/Results/Rankings/RankingsTable.jsx
FinnIckler: well this is what actually fixes the stuck table row issue

app/webpacker/components/Results/Rankings/index.jsx
FinnIckler: yes, especially because `ResultsFilter` will be reused in records

app/webpacker/components/Results/Rankings/RankingsTable.jsx
kr-matthews: No it really is a duplicate key (result id and show), the console was printing warnings.

FinnIckler: oh I get it now. The issue is that the duplicate rows are not across categories, but in the by-region category which react then doesn't clean up I guess.

FinnIckler: fixed by adding country.name

kr-matthews: That doesn't fix it when 'show' is 'results'. I'm still seeing the duplicate key warning:
`Warning: Encountered two children with the same key, `6350611-100 results-China`.` This is at `results/rankings/333/single?show=100+results` i.e. load the rankings page and click 'Results'.

FinnIckler: I have now used tanstack's react table to handle the key generation

app/webpacker/components/Results/Rankings/RankingsTable.jsx
```diff
@@ -0,0 +1,187 @@
...
+      return acc;
+    }, []);
+  }, [competitionsById, rows, show]);
+
+  const columns = useMemo(() => {
+    const commonColumns = [
+      {
+        accessorKey: 'rank',
+        header: show !== 'by region' ? '#' : I18n.t('results.table_elements.region'),
+        cell: (info) => info.getValue(),
```
kr-matthews, line 103: I think this is the default so you can omit `cell` here and everywhere else.

app/webpacker/components/Results/Rankings/RankingsTable.jsx
```diff
@@ -0,0 +1,187 @@
...
+      accessorKey: 'competition.name',
+      header: I18n.t('results.table_elements.competition'),
+      cell: (info) => info.getValue(),
+    });
+
+    if (isAverage) {
+      commonColumns.push(
+        ...Array(5).fill({
+          accessorKey: 'solves', // Adjust as needed
+          header: I18n.t('results.table_elements.solves'),
```
kr-matthews, line 135: Only the first of these 5 should have a header.

kr-matthews (commented):
I'm still confused why there was a duplicate key before - doesn't that mean that a result was showing up in the table twice? I don't understand how switching to react table fixed it either.

FinnIckler:
> I'm still confused why there was a duplicate key before - doesn't that mean that a result was showing up in the table twice? I don't understand how switching to react table fixed it either.

Well I assumed that there was a edge case where the current key wasn't unique enough and react table takes care of that

FinnIckler:
Also I wasn't just switching to React Table to fix that. I was talking to Gregor and the goal is to use it everywhere

app/controllers/results_controller.rb
```diff
@@ -156,7 +156,27 @@ def rankings
...
+    respond_to do |format|
+      format.html {}
+      format.json do
+        cached_data = Rails.cache.fetch ["results-page-api", *@cache_params, @ranking_timestamp] do
+          rows = DbHelper.execute_cached_query(@cache_params, @ranking_timestamp, @query)
+          comp_ids = rows.map { |r| r["competitionId"] }.uniq
+          if @is_by_region
+            rows = compute_rankings_by_region(rows, @continent, @country)
+          end
+          competitions_by_id = Competition.where(id: comp_ids).index_by(&:id).transform_values { |comp| comp.as_json(methods: %w[country], include: [], only: %w[cellName id]) }
```
gregorbg, line 173: Do you really need the whole country? Usually the `iso2` is enough, pulling the `country.name` from the backend to the frontend results in wrong results because of the stateful locale on API servers.

gregorbg: You can then use the `iso2` in the frontend to access even the name via `countries.byIso2[some_iso2].name` or such

app/controllers/results_controller.rb
```diff
@@ -356,4 +376,43 @@ def records
       params[:show] = nil
     end
   end
+
+  private def compute_rankings_by_region(rows, continent, country)
```
gregorbg, line 380: Out of sheer laziness: Is this a 1:1 copy from the helper or did you modify anything?

app/webpacker/components/CompetitionsOverview/CompetitionsFilters.js
```diff
@@ -159,7 +159,7 @@ function TimeOrderButtonGroup({ filterState, dispatchFilter }) {
   return (
     <>
       <label htmlFor="state">{I18n.t('competitions.index.state')}</label>
-      <Button.Group id="state" size="small" compact primary>
+      <Button.Group id="state" size="large" primary>
```
gregorbg, line 162: Why? This change can potentially affect other pages as well...

app/webpacker/components/Results/Rankings/RankingsTable.jsx
```diff
@@ -0,0 +1,186 @@
...
+import CountryFlag from '../../wca/CountryFlag';
+import { continents, countries } from '../../../lib/wca-data.js.erb';
+import { personUrl } from '../../../lib/requests/routes.js.erb';
+
+function CountryCell({ country }) {
+  return (
+    <Table.Cell textAlign="left">
+      {country.iso2 && <CountryFlag iso2={country.iso2} />}
+      {' '}
+      {country.name}
```
gregorbg, line 16: :warning: unreliable, see my other remark above.

app/webpacker/components/Results/Rankings/RankingsTable.jsx
```diff
@@ -0,0 +1,186 @@
...
+
+      const rank = value === previousValue ? previousRank : index + 1;
+      const tiedPrevious = rank === previousRank;
+
+      let country = countries.real.find((c) => c.id === result.countryId);
+      if (index < firstContinentIndex) {
+        country = { name: I18n.t('results.table_elements.world') };
+      } else if (index >= firstContinentIndex && index < firstCountryIndex) {
+        country = continents.real.find((c) => c.id === country.continentId);
+      }
```
gregorbg, line 83: Any chance you can get rid of this last `let`?
Introduce a (file-private) helper fuction if it helps.

app/webpacker/components/Results/Rankings/RankingsTable.jsx
```diff
@@ -0,0 +1,186 @@
...
+      acc.push({
+        result,
+        competition,
+        country,
+        rank,
+        tiedPrevious,
+        key: `${result.id}-${show}-${country.name}`,
+      });
+
+      return acc;
```
gregorbg, line 94: Would this work with `[...acc, { newObjectStuff: 123 }]`?
Not sure whether that introduces performance issues, but I'm usually not a big fan of stateful calls like `push` or `pop`.

app/webpacker/components/Results/Rankings/RankingsTable.jsx
```diff
@@ -0,0 +1,186 @@
...
+      commonColumns.push({
+        accessorKey: 'solves',
+        header: I18n.t('results.table_elements.solves'),
+      });
+      commonColumns.push(
+        ...Array(4).fill({
+          accessorKey: '',
+          header: ' ',
+        }),
+      );
```
gregorbg, line 136: Tiny comment what this is about please

app/webpacker/components/Results/Rankings/RankingsTable.jsx
```diff
@@ -0,0 +1,186 @@
...
+              ))}
+            </Table.Row>
+          ))}
+        </Table.Header>
+        <Table.Body>
+          {table.getRowModel().rows.map((row) => {
+            const {
+              country, result, competition, rank, tiedPrevious,
+            } = row.original;
+            return (
```
gregorbg, line 169: Line breaks are free! :money_mouth_face:

app/webpacker/components/Results/resultsFilter.jsx
```diff
@@ -0,0 +1,95 @@
...
+              } else {
+                setRegion(countries.byIso2[r]?.id ?? r);
+              }
+            }}
+          />
+        </Form.Field>
+        <Form.Group widths="equal">
+          <Form.Field>
+            <label>{I18n.t('results.selector_elements.type_selector.type')}</label>
+            <ButtonGroup primary compact widths={2}>
```
gregorbg, line 60: The second button is only there for non-mbf events though?

app/webpacker/lib/requests/routes.js.erb
```diff
@@ -306,6 +306,25 @@ export const bulkUpdateRegistrationUrl = `<%= CGI.unescape(Rails.application.rou
...
+    queryParams.append('region', region);
+  }
+
+  if (gender !== 'All') {
+    queryParams.append('gender', gender);
+  }
+
+  if (show !== '100 persons') {
+    queryParams.append('show', show);
+  }
```
gregorbg, line 323: If you use this for querying the JSON, you can just always append the parameters. The controller will internally append them anyways, so if you already append them here it saves you a few if-queries for the same end result.

gregorbg (commented):
Please also merge `main` so that you can get the linter to rumble about your JSX files :wink:

app/webpacker/components/CompetitionsOverview/CompetitionsFilters.js
FinnIckler: you did this change!

app/controllers/results_controller.rb
FinnIckler: 1:1 copy

app/webpacker/lib/requests/routes.js.erb
FinnIckler: I was thinking it would be slightly less annoying for people sharing links

app/webpacker/components/CompetitionsOverview/CompetitionsFilters.js
gregorbg: There's a chance >50% that was an accident.

app/controllers/results_controller.rb
FinnIckler: well the issue is that you can only get the countryId and not the iso2 without calling the country method. I introduced a countries.byId to fix this, which I need to use in multiple places anyway

app/webpacker/components/Results/Rankings/RankingsTable.jsx
FinnIckler: nope, this is actually already a country that comes from the frontend

app/webpacker/components/Results/Rankings/RankingsTable.jsx
FinnIckler: it would, don't think it's a performance issue considering it's like at most 200 rows

app/webpacker/components/Results/resultsFilter.jsx
FinnIckler: I think it looks nicer though with widths={2} even for mbf
image
vs
image

app/webpacker/components/Results/Rankings/RankingsTable.jsx
FinnIckler: done

kr-matthews:
> > I'm still confused why there was a duplicate key before - doesn't that mean that a result was showing up in the table twice? I don't understand how switching to react table fixed it either.
> 
> Well I assumed that there was a edge case where the current key wasn't unique enough and react table takes care of that

My point is that a duplicate key would mean a result was showing up twice, since the key was using the id, which seems wrong - how could such an edge case exist? The switch to react table doesn't change that fact.

FinnIckler:
> > > I'm still confused why there was a duplicate key before - doesn't that mean that a result was showing up in the table twice? I don't understand how switching to react table fixed it either.
> > 
> > 
> > Well I assumed that there was a edge case where the current key wasn't unique enough and react table takes care of that
> 
> My point is that a duplicate key would mean a result was showing up twice, since the key was using the id, which seems wrong - how could such an edge case exist? The switch to react table doesn't change that fact.

Did you actually see any duplicate results? I assume this is a problem with the query when one average contains two singles that are shown in the results view. I can dig deeper tomorrow

kr-matthews:
> Did you actually see any duplicate results? I assume this is a problem with the query when one average contains two singles that are shown in the results view. I can dig deeper tomorrow

No, but I didn't look very closely/carefully. This was for displaying 3x3x3 single. If the id of a result is coming from the average, then that could explain it. But in that case, it indicates that the key needs to be updated. If moving to react table has resolved the issue, despite a duplicate key existing, then surely it's not using the key and we can remove the key (I see it's still in the code)? (And if it is using the key, then how is it not getting duplicates?)

Basically, it might (currently) work, but there's definitely a (potentially minor) underlying issue somewhere, because the key isn't unique.

dunkOnIT:
@thewca-bot deploy staging

PR #10570, opened 2025-01-08
Create can_user_access? for ticket

danieljames-dj:
This PR aims to create a method to check if a user has access. This also optimizes the time complexity of checking if they have access.

### DIFF
--- a/app/controllers/tickets_controller.rb
+++ b/app/controllers/tickets_controller.rb
@@ -11,7 +11,7 @@ class TicketsController < ApplicationController
         ticket = Ticket.find(params.require(:id))
 
         # Currently only stakeholders can access the ticket.
-        return head :unauthorized if ticket.can_user_access?(current_user)
+        return head :unauthorized unless ticket.can_user_access?(current_user)
 
         render json: {
           ticket: ticket,
--- a/app/controllers/tickets_controller.rb
+++ b/app/controllers/tickets_controller.rb
@@ -9,22 +9,13 @@ class TicketsController < ApplicationController
       end
       format.json do
         ticket = Ticket.find(params.require(:id))
-        # requester_stakeholders will have the list of stakeholders where the requester is part of.
-        # For example, if a normal user X requests for a change by creating a ticket, the
-        # stakeholders list will be [X, WRT] (WRT is added as stakeholder because WRT is
-        # responsible for taking action on the ticket). If a WRT member fetches the ticket data,
-        # the value of requester_stakeholders will be [WRT] and if the normal user fetches the
-        # ticket data, the value of requester_stakeholders will be [X]. If the ticket is created by
-        # a WRT member, then the value requester_stakeholders will be [X, WRT] because the user can
-        # be any of the two stakeholders.
-        requester_stakeholders = ticket.user_stakeholders(current_user)
 
         # Currently only stakeholders can access the ticket.
-        return head :unauthorized if requester_stakeholders.empty?
+        return head :unauthorized if ticket.can_user_access?(current_user)
 
         render json: {
           ticket: ticket,
-          requester_stakeholders: requester_stakeholders,
+          requester_stakeholders: ticket.user_stakeholders(current_user),
         }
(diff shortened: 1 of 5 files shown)


app/models/ticket_stakeholder.rb
```diff
@@ -9,6 +9,14 @@ class TicketStakeholder < ApplicationRecord
   belongs_to :ticket
   belongs_to :stakeholder, polymorphic: true
 
+  scope :belongs_to_user, lambda { |user_id|
+    where(stakeholder_id: user_id)
```
gregorbg, line 13: This should require an additional `where(stakeholder_type: 'User')` similar to the `belongs_to_groups` scope below, doesn't it?

app/models/ticket_stakeholder.rb
```diff
@@ -9,6 +9,14 @@ class TicketStakeholder < ApplicationRecord
...
+  }
+
+  scope :belongs_to_groups, lambda { |group_ids|
+    where(stakeholder_type: "UserGroup").where(stakeholder_id: group_ids)
```
gregorbg, line 17: Nifty Rails trick: `where` supports an unlimited number of arguments. `where(foo: 123).where(bar: nil)` can be written as `where(foo: 123, bar: nil)`

app/models/ticket.rb
```diff
@@ -9,13 +9,28 @@ class Ticket < ApplicationRecord
...
+  # [X, WRT] (WRT is added as stakeholder because WRT is responsible for taking action on the
+  # ticket). If a WRT member fetches the ticket data, the value of user_stakeholders will be [WRT]
+  # and if the normal user fetches the ticket data, the value of user_stakeholders will be [X]. If
+  # the ticket is created by a WRT member, then the value user_stakeholders will be [X, WRT] because
+  # the user can be any of the two stakeholders.
   def user_stakeholders(user)
     return [] if user.nil?
     ticket_stakeholders.select do |ticket_stakeholder|
       user.active_roles.where(group: ticket_stakeholder.stakeholder).any? || user == ticket_stakeholder.stakeholder
     end
```
gregorbg, line 23: I think this existing method body may now be able to benefit from your new scopes. I might be reading things wrong here, but isn't this just `ticket_stakeholders.belongs_to_user(user) + ticket_stakeholders.belongs_to_group(user.active_groups)``?

app/models/ticket_stakeholder.rb
danieljames-dj: Thanks for that, done.

app/models/ticket_stakeholder.rb
danieljames-dj: Oh I had it initially TBH, I don't know how it disappeared 😅
Added it now.

app/models/ticket.rb
danieljames-dj: Oh yes, I didn't think of that. Thanks for pointing it, done.

app/controllers/tickets_controller.rb
```diff
@@ -9,22 +9,13 @@ def show
...
-        # responsible for taking action on the ticket). If a WRT member fetches the ticket data,
-        # the value of requester_stakeholders will be [WRT] and if the normal user fetches the
-        # ticket data, the value of requester_stakeholders will be [X]. If the ticket is created by
-        # a WRT member, then the value requester_stakeholders will be [X, WRT] because the user can
-        # be any of the two stakeholders.
-        requester_stakeholders = ticket.user_stakeholders(current_user)
 
         # Currently only stakeholders can access the ticket.
-        return head :unauthorized if requester_stakeholders.empty?
+        return head :unauthorized if ticket.can_user_access?(current_user)
```
gregorbg, line 14: Ehrm... I was about to merge this until I realized: Shouldn't this be an `unless`?!

danieljames-dj: Oops sorry yes. This was a big miss, sorry. I've fixed it now.

PR #10572, opened 2025-01-08
Added name argument to validators for ticket validations

danieljames-dj:
One drawback of this is still js cannot replace "%{name}" with name. So, this doesn't fix the UI issue. This should be solved when moved to i18n. This PR aims at sending `name` parameter for `dob_validations` method.

### DIFF
--- a/app/controllers/tickets_controller.rb
+++ b/app/controllers/tickets_controller.rb
@@ -49,7 +49,8 @@ class TicketsController < ApplicationController
     ticket.metadata.tickets_edit_person_fields.each do |edit_person_field|
       case edit_person_field[:field_name]
       when TicketsEditPersonField.field_names[:dob]
-        dob_validation_issues = ResultsValidators::PersonsValidator.dob_validations(Date.parse(edit_person_field[:new_value]), { name: ticket.metadata.wca_id })
+        dob_to_validate = Date.parse(edit_person_field[:new_value])
+        dob_validation_issues = ResultsValidators::PersonsValidator.dob_validations(dob_to_validate, nil, name: ticket.metadata.wca_id)
       end
     end
 
--- a/app/controllers/tickets_controller.rb
+++ b/app/controllers/tickets_controller.rb
@@ -49,7 +49,7 @@ class TicketsController < ApplicationController
     ticket.metadata.tickets_edit_person_fields.each do |edit_person_field|
       case edit_person_field[:field_name]
       when TicketsEditPersonField.field_names[:dob]
-        dob_validation_issues = ResultsValidators::PersonsValidator.dob_validations(Date.parse(edit_person_field[:new_value]))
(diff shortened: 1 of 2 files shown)


lib/results_validators/persons_validator.rb
```diff
@@ -41,22 +41,22 @@ def include_persons?
       true
     end
 
-    def self.dob_validations(dob, competition_id = nil, name = nil)
+    def self.dob_validations(dob, message_args = {}, competition_id = nil)
```
gregorbg, line 44: I think you can make it more Ruby-esque by something like...
```suggestion
    def self.dob_validations(dob, competition_id = nil, **message_args)
```

gregorbg: If you take this suggestion, then below you can "spread" the parameters into the `ValidationIssue` class as well: `ValidationWarning.new(:persons, competition_id, DOB_0101_WARNING, **message_args)`

lib/results_validators/persons_validator.rb
```diff
@@ -101,7 +101,7 @@ def run_validation(validator_data)
                                                name: p.name)
           end
 
-          PersonsValidator.dob_validations(p.dob, competition.id, p.name).each do |validation|
+          PersonsValidator.dob_validations(p.dob, { name: p.name }, competition.id).each do |validation|
```
gregorbg, line 104: For my suggestion above, this could be changed to `PersonsValidator.dob_validations(p.dob, competition.id, name: p.name)`

app/controllers/tickets_controller.rb
```diff
@@ -49,7 +49,7 @@ def edit_person_validators
     ticket.metadata.tickets_edit_person_fields.each do |edit_person_field|
       case edit_person_field[:field_name]
       when TicketsEditPersonField.field_names[:dob]
-        dob_validation_issues = ResultsValidators::PersonsValidator.dob_validations(Date.parse(edit_person_field[:new_value]))
+        dob_validation_issues = ResultsValidators::PersonsValidator.dob_validations(Date.parse(edit_person_field[:new_value]), { name: ticket.metadata.wca_id })
```
gregorbg, line 52: If you like my other suggestion about the method signature, then this can become...
```suggestion
        dob_validation_issues = ResultsValidators::PersonsValidator.dob_validations(Date.parse(edit_person_field[:new_value]), name: ticket.metadata.wca_id)
```

gregorbg: Sidenote: You should probably extract the `Date.parse` into its own line, helps with debugging :)

lib/results_validators/persons_validator.rb
danieljames-dj: Okay done.

lib/results_validators/persons_validator.rb
danieljames-dj: Done

app/controllers/tickets_controller.rb

gregorbg (approved):

PR #10573, opened 2025-01-08
Check competition dates for upcoming comps while banning

danieljames-dj:
This was an edge case noticed by Shain and the WIC also has an urgent case which is blocked because of this

### DIFF
--- a/spec/controllers/api/v0/user_roles_controller_spec.rb
+++ b/spec/controllers/api/v0/user_roles_controller_spec.rb
@@ -125,6 +125,15 @@ RSpec.describe Api::V0::UserRolesController do
         expect(response_json["error"]).to eq "The user has upcoming competitions: #{upcoming_comps_for_user.join(', ')}. Before banning the user, make sure their registrations are deleted."
       end
 
+      it "can ban a user if the user's upcoming competitions are after end date" do
+        post :create, params: {
+          userId: user_to_be_banned_with_future_comps.id,
+          groupType: UserGroup.group_types[:banned_competitors],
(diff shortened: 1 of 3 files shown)


gregorbg:
Could you please explain this edge case?

danieljames-dj:
> Could you please explain this edge case?

If we are trying to ban a competitor for 1 year, and if the competitor has a registration after 1 year, still they should be allowed to be banned because their upcoming competition is not within the ban window.

spec/controllers/api/v0/user_roles_controller_spec.rb
```diff
@@ -125,6 +125,15 @@
         expect(response_json["error"]).to eq "The user has upcoming competitions: #{upcoming_comps_for_user.join(', ')}. Before banning the user, make sure their registrations are deleted."
       end
 
+      it 'can ban a user if the user\'s upcoming competitions are after end date' do
```
gregorbg, line 128: You can make this double quotes, then you don't have to mask out the apostrophe in `don't` 
```suggestion
      it "can ban a user if the user's upcoming competitions are after end date" do
```

gregorbg (commented):
You need to also fix the other test case which is failing ("cannot ban a user if the user have any upcoming competitions"), otherwise fine to merge :+1:

spec/controllers/api/v0/user_roles_controller_spec.rb
danieljames-dj: Done

danieljames-dj:
> You need to also fix the other test case which is failing ("cannot ban a user if the user have any upcoming competitions"), otherwise fine to merge 👍

Fixed it, changed the code a bit to fix the test.

gregorbg (approved):

PR #10575, opened 2025-01-09
Fix Person trying to serialize non-existing methods in User

gregorbg:
For some reason, we have the following code snippet in the `serializable_hash` method of `Person`:

```
# If there's a user for this Person, merge in all their data,
# the Person's data takes priority, though.
(user || User.new).serializable_hash(user_override_options).merge(json)
```

The code comment explains what the intention is, but the implementation is problematic because it directly passes down the serialization options from Person down to User. But `Person` has plenty of methods and properties (which the developer might want to serialize) that don't exist on `User` at all.

The only reason why this didn't break so far is because nobody ever happened to request (i.e. try and serialize) a property of `Person` that isn't available from `User`.

The monkey-patched method is essentially a short-hand for:
```
{
  only: USER_PERSON_COMMON_OPTIONS[:only] & options[:only],
  methods: USER_PERSON_COMMON_OPTIONS[:methods] & options[:methods],
  includes: USER_PERSON_COMMON_OPTIONS[:includes] & options[:includes],
}
```
with a little bit of null-safety built in. The intention is to merge these two arrays together, but only keep the union (i.e. common values shared between two arrays) of the value params. That way, we can effectively restrict which params should be passed down to `User`.

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -314,7 +314,7 @@ class Person < ApplicationRecord
 
     # If there's a user for this Person, merge in all their data,
     # the Person's data takes priority, though.
-    user_override_options = USER_COMMON_SERIALIZE_OPTIONS.merge_union(options)
+    user_override_options = USER_COMMON_SERIALIZE_OPTIONS.merge_union(options&.stringify_keys)
     (user || User.new).serializable_hash(user_override_options).merge(json)
   end
 end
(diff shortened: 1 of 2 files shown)


danieljames-dj (approved):
I understood almost everything, the monkey_patches code is bit confusing for me, but I think I understood it.

PR #10577, opened 2025-01-09
Switch flag to disable Comp Overview React on demand

gregorbg:
Effectively launching the new competition overview page

This is the least-intrusive-but-also-easy-to-understand diff that I could come up with (everything with even less changed lines would have been an awkwardly convoluted `if !@disabled_unless` blabla)


dunkOnIT (approved):
Tested locally and working as expected!

PR #10584, opened 2025-01-09
Remove @@character_set_server latin1 check

FinnIckler:
We don't need this anymore as we don't run any PHP.

### DIFF
--- a/app/controllers/server_status_controller.rb
+++ b/app/controllers/server_status_controller.rb
@@ -92,6 +92,7 @@ class MysqlSettingsCheck < StatusCheck
   EXPECTED_MYSQL_SETTINGS = {
     "@@innodb_ft_min_token_size" => 2,
     "@@ft_min_word_len" => 2,
+    "@@character_set_server" => "utf8mb4",
   }.freeze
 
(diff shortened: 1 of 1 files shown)


FinnIckler:
I don't think that makes sense, these checks are for settings that are changed from the default, but `utf8mb4` is the default

gregorbg:
In that case, having a check would make sure that we're not deviating from the default.
Plus, if MySQL ever decided to change it to some other silly value, we would just casually accept that "because it's the default", while in reality this might screw us over.

It's especially important for WRT who take a great deal of care to make sure that local names in non-English alphabets are accurate, so I really don't want to mess with encoding. Please add a check for it.

FinnIckler:
I added the check, but this setting is only used for CREATE TABLE statements that have no charset specified. By setting
```
  encoding: utf8mb4
  collation: utf8mb4_unicode_ci
```
in db.yml this is automatically done. That's why this still worked fine even though we had set it to latin1 before. So even if mysql9 or whatever changes the value, we will still be fine

gregorbg (approved):

PR #10592, opened 2025-01-10
Psych sheet columns and headers

### LINKED ISSUE #10552: Psych Sheet Changes
## High

- [ ] Order psych sheet by mean/average for events where mean/average is used for ranking
- [ ] Show position on psych sheet in a column on the left of the table (ie, to the left of the user's name), rather than in the middle of the table where it currently is

## Medium

- [ ] Show position on psych sheet in a column on the left of the table (ie, to the left of the user's name), rather than in the middle of the table where it currently is
- [ ] Alternate the colours of rows in the psych sheet table

kr-matthews:
Partially resolves 10552.

Move rank to first column, mark columns as disabled (cannot be clicked (to sort)) when applicable, and re-introduce ability to click an event column to open the psych sheet for that event.

image

[Screencast from 2025-01-09 05:18:46 PM.webm](https://github.com/user-attachments/assets/eb31292b-8814-47c2-893a-d4333c214846)

Future PR: Given a particular event, it might be good to have a single call to get the psych sheet data, and then have the front end handle sorting by single vs average?

Also, ties are not currently shown as the same rank:

image

### DIFF
--- a/app/webpacker/components/RegistrationsV2/Registrations/RegistrationList.jsx
+++ b/app/webpacker/components/RegistrationsV2/Registrations/RegistrationList.jsx
@@ -31,13 +31,12 @@ export default function RegistrationList({ competitionInfo }) {
     retry: false,
   });
 
-  const [state, dispatch] = useReducer(sortReducer, {
+  const [{ sortColumn, sortDirection }, sortDispatch] = useReducer(sortReducer, {
     sortColumn: 'name',
     sortDirection: undefined,
   });
 
-  const { sortColumn, sortDirection } = state;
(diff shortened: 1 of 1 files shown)


app/webpacker/components/RegistrationsV2/Registrations/RegistrationList.jsx
```diff
@@ -301,14 +309,18 @@ function FooterContent({
 
   return (
     <Table.Row>
+      {!isAllCompetitors && (
```
gregorbg, line 312: Nit: This would boil down to `isPsychSheet` again, no?

gregorbg (approved):

PR #10598, opened 2025-01-10
Only disable qualification button if qualifications are enforced

### LINKED ISSUE #10593: Register UI doesn't allow users to select events they don't meet qualification for, even if qualifications aren't enforced

dunkOnIT:
[Austrlian Nationals](https://www.worldcubeassociation.org/competitions/AustralianNationals2025/register) allows registration without meeting the qualification requirements, and this is allowed by our backend, but the event buttons are disabled in the frontend in all cases, meaning users can't register even if qualifications aren't enforced

Fixes #10593

### DIFF
--- a/app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
+++ b/app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
@@ -297,6 +297,8 @@ export default function CompetingStep({
             warning
             list={formWarnings}
           />
+          {console.log("comp info")}
+          {console.log(competitionInfo)}
           <Form.Field required error={hasInteracted && selectedEvents.length === 0}>
             <EventSelector
               onEventSelection={handleEventSelection}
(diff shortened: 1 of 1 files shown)


app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
```diff
@@ -297,18 +297,24 @@ export default function CompetingStep({
             warning
             list={formWarnings}
           />
+          {console.log("comp info")}
```
FinnIckler, line 300: console.logs

dunkOnIT: every time

FinnIckler:
I am still a bit confused about `allow_registration_without_qualification` it just means they'll handle qualifications manually at the end?

dunkOnIT:
Yep exactly

FinnIckler (approved):

PR #10599, opened 2025-01-10
Date Range locale fixes

FinnIckler:
I translated the dateRange function from time will tell into luxon, but I assume we don't actually need to be this flexible and I can remove most of the options?


FinnIckler:
Ok, you can actually just use luxon interval to make this really simple

gregorbg (approved):

PR #10605, opened 2025-01-12
Reject user registration edits if comp has started

dunkOnIT:
We missed this case because we always checked edits against the registration edit deadline. Another option here could be to default the edit deadline to the comp's start date

### DIFF
--- a/app/models/competition.rb
+++ b/app/models/competition.rb
@@ -1170,6 +1170,7 @@ class Competition < ApplicationRecord
   # can registration edits be done right now
   # must be allowed in general, and if the deadline field exists, is it a date and in the future
   def registration_edits_allowed?
+    return false if started?
     self.allow_registration_edits &&
       (!has_event_change_deadline_date? || !event_change_deadline_date.present? || event_change_deadline_date > DateTime.now)
   end


gregorbg:
> because we always checked edits against the registration edit deadline

You mean in the legacy system?

app/models/competition.rb
```diff
@@ -1170,6 +1170,7 @@ def has_event_change_deadline_date?
   # can registration edits be done right now
   # must be allowed in general, and if the deadline field exists, is it a date and in the future
   def registration_edits_allowed?
+    return false if started?
     self.allow_registration_edits &&
       (!has_event_change_deadline_date? || !event_change_deadline_date.present? || event_change_deadline_date > DateTime.now)
```
gregorbg, line 1175: Shouldn't this be a simple matter of...
```suggestion
    !started? && self.allow_registration_edits &&
      (!has_event_change_deadline_date? || !event_change_deadline_date.present? || event_change_deadline_date > DateTime.now)
```

gregorbg (changes requested):

app/models/competition.rb
dunkOnIT: Yep I like that better!

gregorbg (commented):
On second thought, not quite sure whether this is the right place for this check. The method name `allowed` is a bit ambiguous, in my mind it refers to the "legal" permission to make edits per WCRP and competition settings. It should distinguish between comps who allow edits in general, vs. competitions who opted out of edits _in the first place_.

The start date could also be checked elsewehere, i.e. at the call-site where the changes are being made. And then just return early instead of continuing to process.

Anyhow, feel free to argue/debate. But no matter what conclusion we reach, this PR should definitely be backed by tests. The fact that tests passed before this change and tests continue to pass after this change indicates that our existing tests are not very helpful or conclusive.

dunkOnIT:
Agreed, this should absolutely be supported by tests - rare "Duncan doesn't submit tests" moment. 

> The start date could also be checked elsewehere, i.e. at the call-site where the changes are being made. And then just return early instead of continuing to process.

I think the Competition model is the appropriate place for this check - the method is being called from registration_checker:81, which already has a lot of logic going on - so I feel strongly that the we should be delegating the responsibility of handling the specifics of whether a comp may be edited to the comp itself.

> The method name allowed is a bit ambiguous, in my mind it refers to the "legal" permission to make edits per WCRP and competition settings. It should distinguish between comps who allow edits in general, vs. competitions who opted out of edits in the first place.

The intention of the method - in my mind at least - is determine whether registrations may be edited for _any_ reason. If we want to know specifically whether edits are allowed per competition settings, we can just call `self.allow_registration_edits`. The only other reasons registrations may not be editable at the competition-level is because of the edit being made outside of an allowed date range - which is what the rest of the method is checking [essentially, the comp having already started is just another limit on _when_ registration edits can be made].

I'm not aware of any WCRP limitations on registrations being editable at the competition-level [beyond dates] - but if you have specific examples in mind, let me know.

app/models/competition.rb
```diff
@@ -1564,7 +1564,7 @@ def ineligible_events(user)
   end
 
   def started?
-    start_date.present? && start_date < Date.today
+    start_date.present? && start_date <= Date.today
```
gregorbg, line 1567: Oh, this may have veeery unforessen consequences or side-effects. TBH I'm surprised that tests even pass. Why is this change necessary in the first place?

app/models/competition.rb
```diff
@@ -1169,8 +1169,8 @@ def has_event_change_deadline_date?
 
   # can registration edits be done right now
   # must be allowed in general, and if the deadline field exists, is it a date and in the future
-  def registration_edits_allowed?
-    self.allow_registration_edits &&
+  def registration_edits_permitted?
```
gregorbg, line 1172: Maybe bake `currently` into the name since we're checking against `DateTime.now`?

spec/lib/registrations/registration_checker_spec.rb
```diff
@@ -797,6 +797,24 @@
         end
       end
 
+      it 'user cant change events after comp has started', :tag do
```
gregorbg, line 800: `:tag` :smiling_imp:

gregorbg (changes requested):

app/models/competition.rb
dunkOnIT: > Why is this change necessary in the first place?

Currently, the check only considers a comp started if the current date is _after_ the start date - but, if the current date _is_ the start date, I think it is fair to consider a competition started. As it currently stands, a user could still edit their registration on the first day of a competition

> Oh, this may have veeery unforessen consequences or side-effects

Agreed - for a first step I'll take a look at where `.started?` is called and see if I notice any places this might cause issue

dunkOnIT: Update: I only find 3 instances of `started?` being invoked - it might be worth confirming this using a grep on your side as well 
- `competition_spec.rb:202`
- `competition.rb:1173` (in the `registration_edits_currently_allowed?` method)
- In its method definition

Following `registration_edits_currently_permitted` leads to `can_edit_registrations?` in `user.rb`, which leads to a bunch of test cases and one use in `register.html.erb` - which I'm pretty sure is deprecated now that we're on the new registration system?

gregorbg: Well then I suppose we should work on merging https://github.com/thewca/worldcubeassociation.org/pull/10411 first, because that will get rid of most `register.html.erb` code.

dunkOnIT: Confirming that `can_edit_registrations?` is now only used in tests after merging in main - so the splash zone for changing `started?` is only the `registration_checker`, which is well-tested

PR #10618, opened 2025-01-13
Allow user to access their own DOB

maxidragon:
@danieljames-dj discovered that users cannot access their own DOB because it is restricted to WRT only. As a result, we (WRT) have received some empty forms, which were not actually empty. Users had entered the correct DOB, but after backend comparison, it was determined that there was no change, leading to an empty email notification.

In this PR, I've allowed users to access their own `private_attributes`, so their DOB will be prefilled in the edit profile form.


danieljames-dj (commented):
Thanks @maxidragon. I agree with the changes. But since this is having changes in sensitive information, I prefer to get approval from @gregorbg.

app/controllers/api/v0/persons_controller.rb
```diff
@@ -17,7 +17,7 @@ def index
   def show
     person = Person.current.includes(:user, :ranksSingle, :ranksAverage).find_by_wca_id!(params[:wca_id])
     private_attributes = []
-    if current_user && current_user.can_admin_results?
+    if current_user && (current_user.can_admin_results? || current_user&.person == person)
```
gregorbg, line 20: `current_user.person` is by definition identified by the User's `wca_id`. You also have an ID in `params[:wca_id]`. So doing `current_user.wca_id == params[:wca_id]` is a lot cheaper and efficient to execute

gregorbg (changes requested):

gregorbg:
Also, question that I genuinely don't know the answer to:
Should users see their own `incorrect_wca_id_claim_count`?

danieljames-dj:
> Also, question that I genuinely don't know the answer to: Should users see their own `incorrect_wca_id_claim_count`? Because with this PR, they would. But I'm honestly not sure if that's a good thing.

Great catch. There is nothing wrong in providing, but it's currently not necessary. Maybe can we add just DOB for self users? WDYT @maxidragon?

maxidragon:
Yeah, it's not necessary, I'll modify the code to add just DOB for normal users

app/controllers/api/v0/persons_controller.rb
```diff
@@ -17,8 +17,12 @@ def index
...
     person = Person.current.includes(:user, :ranksSingle, :ranksAverage).find_by_wca_id!(params[:wca_id])
     private_attributes = []
-    if current_user && current_user.can_admin_results?
-      private_attributes = %w[incorrect_wca_id_claim_count dob]
+    if current_user
+      if current_user.wca_id == params[:wca_id]
+        private_attributes = %w[dob]
+      elsif current_user.can_admin_results? || current_user.any_kind_of_delegate?
+        private_attributes = %w[incorrect_wca_id_claim_count dob]
+      end
```
danieljames-dj, line 25: LGTM, but actually even delegate doesn't actually need `incorrect_wca_id_claim_count`.

Also how about storing `params[:wca_id]` to a variable as it is used more than once?

Also I was thinking to remove the nested if with following:

```
if current_user&.wca_id == wca_id || current_user&.any_kind_of_delegate?
  private_attributes = %w[dob]
elsif current_user&.can_admin_results?
  private_attributes = %w[incorrect_wca_id_claim_count dob]
end
```

But I am not sure which is better - @gregorbg might be able to help.

app/controllers/api/v0/persons_controller.rb
```diff
@@ -17,8 +17,13 @@ def index
   def show
     person = Person.current.includes(:user, :ranksSingle, :ranksAverage).find_by_wca_id!(params[:wca_id])
     private_attributes = []
-    if current_user && current_user.can_admin_results?
-      private_attributes = %w[incorrect_wca_id_claim_count dob]
+    if current_user
+      wca_id = params[:wca_id]
```
gregorbg, line 21: You're extracting the param into its own variable here, but you completely miss the other usage in line 18. Please declare the variable there and use it everywhere :)

gregorbg (changes requested):

app/controllers/api/v0/persons_controller.rb
maxidragon: Missed that one, thanks!

app/controllers/api/v0/persons_controller.rb
```diff
@@ -15,10 +15,15 @@ def index
...
+    person = Person.current.includes(:user, :ranksSingle, :ranksAverage).find_by_wca_id!(wca_id)
     private_attributes = []
-    if current_user && current_user.can_admin_results?
-      private_attributes = %w[incorrect_wca_id_claim_count dob]
+    if current_user
+      if current_user&.wca_id == wca_id || current_user&.any_kind_of_delegate?
+        private_attributes = %w[dob]
+      elsif current_user&.can_admin_results?
+        private_attributes = %w[incorrect_wca_id_claim_count dob]
+      end
```
gregorbg, line 26: Also, more nit (sorry!): You're wrapping this whole block inside an `if current_user` check. So you already know that `current_user` won't be nil, which makes the `&.` completely unnecessary.

Either:
- Get rid of the `&.` and use simple `.`
- or Drop the outer `if  current_user` completely

(both will effectively do the same thing and both will work, because `if nil` is false in Ruby)

gregorbg (changes requested):

gregorbg:
(Had to merge `main` again to get rid of a very weird CI quirk. Nothing wrong with your code!)

PR #10625, opened 2025-01-15
Psych sheet current user highlight/info/scroll

kr-matthews:
~The built-in `active` prop for a table row is barely noticeable (in the docs it's noticeable, so I'm not sure if we have some conflicting custom css or something?), so I did a custom background colour.~

**Updated videos in a comment below.**

User competing in some but not all events:

[Screencast from 2025-01-14 04:09:16 PM.webm](https://github.com/user-attachments/assets/2f908113-e6bd-405b-be43-756e6a082d09)

User competing in events without an official result already:

[Screencast from 2025-01-14 04:10:25 PM.webm](https://github.com/user-attachments/assets/be2a3e37-7a0f-45ec-a573-6f1c3dc56809)

Not logged in:

[Screencast from 2025-01-14 04:14:33 PM.webm](https://github.com/user-attachments/assets/fb333a52-b66b-4372-87f6-f5b76be95c17)


gregorbg:
I've just pushed a CSS rule that will override Bootstrap, so if you use SemUI's `active` it should work as expected now.

kr-matthews:
> I've just pushed a CSS rule that will override Bootstrap, so if you use SemUI's `active` it should work as expected now.

Thanks. It works. Any chance there's a similar issue making the table stripes less noticeable?

coder13:
My feelings are mixed on taking up the vertical space just for this functionality.

I don't have any different suggestions on this functionality.

I also feel we should not present the ranking information "Your rank is 19." as a sentence and rather it should be a more simplified view of "Rank: 19 / Average: xx.yy / Single: xx.yy" or similar.

kr-matthews:
> My feelings are mixed on taking up the vertical space just for this functionality.
> 
> I don't have any different suggestions on this functionality.
> 
> I also feel we should not present the ranking information "Your rank is 19." as a sentence and rather it should be a more simplified view of "Rank: 19 / Average: xx.yy / Single: xx.yy" or similar.

One of the other tasks is to move the "10 first-timers + 75 returners = 85 people" out of the first column's footer, because you can't see it without scrolling to the bottom and it makes the first column unnecessarily wide. So I was thinking that that information could also move up to this box, in which case the vertical space is 'less wasteful' since it's showing more information.

I don't mind truncating the sentences like that, or keeping them, but it would be good to get feedback from anyone else on this.

An alternative would be to just have a floating button and no text/information above the table. Or moving the current user's row to the top of the table, but this would be awkward since the table is sortable and half the point is to see how you compare to those around you, so I'm not in favour of this.

gregorbg:
> I don't mind truncating the sentences like that, or keeping them, but it would be good to get feedback from anyone else on this.

Speaking from my personal perspective (ie this is not a binding Team Leader instruction): I always like to have "statistical" information like the one you're mentioning be as condensed as possible. It reduces the workload on translators for information that easily translates across cultures without the need for fully phrased sentences.

And also yes, it saves space which will delight all those who want to squeeze as much information on their phone screen as humanly possible. And judging by some other recent React page launches, there are plenty of these people out there.

kr-matthews:
Updated:

[Screencast from 2025-01-20 06:50:19 PM.webm](https://github.com/user-attachments/assets/565cb783-ba7c-4ebf-82df-6cf654df3eb2)

app/views/registrations/index.html.erb
```diff
@@ -1,5 +1,5 @@
 <% provide(:title, I18n.t('registrations.list.title', comp: @competition.name)) %>
 
 <%= render layout: "nav" do %>
-  <%= react_component('RegistrationsV2/Registrations', { competitionInfo: @competition.to_competition_info }) %>
+  <%= react_component('RegistrationsV2/Registrations', { competitionInfo: @competition.to_competition_info, userInfo: @current_user }) %>
```
gregorbg, line 4: Fully serializing a user is usually expensive, because the resulting JSON contains a _lot_ of information by default. Depending on what information you need, you can do:

```suggestion
  <%= react_component('RegistrationsV2/Registrations', { competitionInfo: @competition.to_competition_info, userInfo: @current_user.as_json(
    only: [...names of database columns...],
    methods: [...names of methods...],
    include: [...ORM relations to include, like current teams for example...]
  ) }) %>
```

app/views/registrations/index.html.erb
```diff
@@ -1,5 +1,7 @@
 <% provide(:title, I18n.t('registrations.list.title', comp: @competition.name)) %>
 
 <%= render layout: "nav" do %>
-  <%= react_component('RegistrationsV2/Registrations', { competitionInfo: @competition.to_competition_info }) %>
+  <%= react_component('RegistrationsV2/Registrations', { competitionInfo: @competition.to_competition_info, userInfo: @current_user.as_json(
+    only: %w[id],
+  )}) %>
```
gregorbg, line 6: Now if you really only need the user's ID, you can just pass it as `userId` prop, no?

gregorbg (approved):

PR #10626, opened 2025-01-15
PsychSheet default sort fix

pranav-027:


gregorbg (commented):
Just curious, why do you need the new `isBlindfolded` information?

pranav-027:
To check if its a blindfolded event as those need to be sorted by single

gregorbg:
To clarify, in my previous comment (on the other PR that was eaten by your IDE) I suggested using `events.byId[psychSheetEvent].recommendedFormat().sortBy`, which is based solely on existing data.

In which cases does my "receommended format approach" and your "blindfolded approach" deliver different results, and why do these cases matter?

pranav-027:
Hello @gregorbg, I tried your approach
```jsx
events.byId[psychSheetEvent].recommendedFormat().sortBy
```
Getting error `Cannot read properties of undefined (reading 'recommendedFormat')`

gregorbg:
In that case, you should debug what's going on. I'm pretty confident the general approach should work.

The error indicates that it cannot read `events.byId[psychSheetEvent]` properly. Since the `events.byId` hash is used in a gazillion places, the error will probably be that `psychSheetEvent` is not initialized properly

pranav-027:
Hello @gregorbg, It's working as expected now. Thanks for suggestion

gregorbg (approved):

PR #10628, opened 2025-01-15
Give organizer access to "bulk auto accept" 

dunkOnIT:
Splitting out the bulk auto accept logic from my other PR


dunkOnIT:
Blocked until #10416 is merged

dunkOnIT:
Merged as #11233 instead

PR #10086, opened 2024-10-17
Add inline images to Delegate Reports

gregorbg:
Adds the ability to directly force image uploads in a Delegate Report, instead of "butchering" it through Markdown.
Advantage 1: We can verify that a Delegate indeed uploaded images before submitting the report.
Advantage 2: We can query these images later, aggregate them, move them elsewhere without much hassle, etc.

PR #10641, opened 2025-01-15
Include ImageMagick in GitHub CI runner

gregorbg:
Because GH threw it out without clearly announcing it. Yay!

(Note that this makes our tests fail. I merged #10086 *after tests passed* because it was still on the old 22.04 runner. But that CI run happened a few weeks ago already. Now after the merge the `main` CI ran on the new 24.04 runner and suddenly failed on the exact same code because ImageMagick was missing)


gregorbg:
Also see https://github.com/actions/runner-images/issues/10772

FinnIckler:
how is this handled in the dockerfile?

PR #10644, opened 2025-01-15
Port edit profile to react

FinnIckler:
Have only done the scaffolding so far, the pro is that the RolesTab  and the Edit Avatar Tab is already migrated


FinnIckler:
image
The disabled state of Semantic is pretty bad and also makes the labels disabled...

FinnIckler:
Should be feature complete now. There is definitely some room for refactoring like the creating a shared component for the authenticity token hidden field that is required for rails form requests

FinnIckler:
Will also need to have some custom styling for the main page

FinnIckler:
General Page:
image
Confirmation Prompt:
image
Sensitive Change:
image
2FA Page:
image

PR #10645, opened 2025-01-15
Remove old admin panel

danieljames-dj:
The old panel is removed in this PR. The pages of the old admin panel are yet to be migrated to React, they will be done in upcoming PRs. This PR aims at disconnecting those pages from the panel and get an independent page to all of them.


gregorbg (commented):
Just curious, what is your point/intention with this PR? In terms of code, we're not gaining anything, and you might just as well "properly" migrate the pages to React one by one

danieljames-dj:
> Just curious, what is your point/intention with this PR? In terms of code, we're not gaining anything, and you might just as well "properly" migrate the pages to React one by one

This PR is just to decouple everything from old admin panel and connect to new admin/WRT panel. I'm already working on the react migration in parallel - I started with 'run validators' page.

gregorbg (approved):

PR #10646, opened 2025-01-16
Fix date format for DatePicker component parser

gregorbg:
It seems that our DatePicker library wants the "year" part to be specified as lowercase `yyyy` rather than standard JS uppercase `YYYY`.

I could not find a precise documentation explaining why (or what the difference is). I just noticed locally that uppercase throws errors in the console while lowercase works without any flaws. Also, the examples at https://reactdatepicker.com/ all use lowercase `yyyy`, I couldn't find any uppercase `YYYY` at all in there.


Goosly (approved):

gregorbg:
Merging based on Manu's review

Issue #10539, opened 2025-01-06
Setup and Teardown activities missing en.yml keys


dunkOnIT:
![image](https://github.com/user-attachments/assets/a11036a0-3f8e-4434-bbca-5348c22378e1)

PR #10649, opened 2025-01-18
Fixing issues around new activity codes Teardown and Setup

pranav-027:
fixes: https://github.com/thewca/worldcubeassociation.org/issues/10539 and https://github.com/thewca/worldcubeassociation.org/issues/10650


gregorbg (approved):
Thanks, the backend must have been an oversight from the initial feature merge

PR #10652, opened 2025-01-19
Migate Run Validators page to react

danieljames-dj:
This is part of migrating all panel pages to react


app/webpacker/components/Panel/pages/RunValidatorsPage/RunValidatorsForm.jsx
```diff
@@ -0,0 +1,158 @@
...
+          fluid
+          multiple
+          selection
+          value={selectedValidators}
+          onChange={setSelectedValidators}
+        />
+        <Form.Checkbox
+          label={`Apply fix when possible (List of validators with automated fix: ${
+            VALIDATORS_WITH_FIX.map(validatorNameReadable).join(', ')
+          })`}
```
gregorbg, line 144: Bit awkward to have such a (potentially) long list as part of the checkbox label. Can you add them as an extra text paragraph below?

gregorbg: SemUI `Header` as `h4` or `h5` with only the `subheader` and no "real" header should do: https://react.semantic-ui.com/elements/header/#content-subheader

app/webpacker/components/Panel/pages/RunValidatorsPage/ValidationListView.jsx
```diff
@@ -0,0 +1,56 @@
...
+function ValidationText({ validationData, group, showCompetitionNameOnOutput }) {
+  return (
+    <>
+      {showCompetitionNameOnOutput && (
+        <>
+          [
+          <a href={competitionUrl(validationData.competition_id)}>
+            {validationData.competition_id}
+          </a>
+          {'] '}
```
gregorbg, line 50: Why do you need the `[]`? Are they absolutely essential for user communication?

app/webpacker/components/Panel/pages/RunValidatorsPage/ValidationOutput.jsx
```diff
@@ -0,0 +1,90 @@
...
+import Errored from '../../../Requests/Errored';
+import ValidationListView from './ValidationListView';
+
+export default function ValidationOutput({
+  validationOutput, isFetching, isError, showCompetitionNameOnOutput,
+}) {
+  if (isFetching) return <Loading />;
+  if (isError) return <Errored />;
+
+  if (!isFetching && !isError && !validationOutput) {
```
gregorbg, line 13: At this point, you already know that `!isFetching && !isError` will always be true. Otherwise, the code would have returned already. So just checking `!validationOutput` should be enough.

gregorbg (changes requested):

app/webpacker/components/Panel/pages/RunValidatorsPage/CompetitionRangeSelector.jsx
```diff
@@ -0,0 +1,68 @@
...
+import Errored from '../../../Requests/Errored';
+
+const RUN_VALIDATORS_QUERY_CLIENT = new QueryClient();
+const MAX_COMPETITIONS_PER_QUERY = 50;
+
+export default function CompetitionRangeSelector({ range, setRange }) {
+  const [startDate, setStartDate] = useState(range?.startDate);
+  const [endDate, setEndDate] = useState(range?.endDate);
+
+  const enableCompetitionListFetch = Boolean(startDate && endDate);
```
kr-matthews, line 16: I'm strongly in favour of naming things based on what they are, not how they are used.
```suggestion
  const bothDatesAreSelected = Boolean(startDate && endDate);
```

app/webpacker/components/Panel/pages/RunValidatorsPage/CompetitionRangeSelector.jsx
```diff
@@ -0,0 +1,68 @@
...
+import getCompetitionList from './api/getCompetitionList';
+import Loading from '../../../Requests/Loading';
+import Errored from '../../../Requests/Errored';
+
+const RUN_VALIDATORS_QUERY_CLIENT = new QueryClient();
+const MAX_COMPETITIONS_PER_QUERY = 50;
+
+export default function CompetitionRangeSelector({ range, setRange }) {
+  const [startDate, setStartDate] = useState(range?.startDate);
+  const [endDate, setEndDate] = useState(range?.endDate);
```
kr-matthews, line 14: Why do we have these states here in the component? If we're receiving `range` and `setRange` then the parent should be updating `range` when `setRange` is called, and `const startDate = range?.startDate` will work.

app/webpacker/components/Panel/pages/RunValidatorsPage/CompetitionRangeSelector.jsx
```diff
@@ -0,0 +1,68 @@
...
+      <Form.Field
+        label="End Date"
+        control={UtcDatePicker}
+        showYearDropdown
+        dateFormatOverride="yyyy-MM-dd"
+        dropdownMode="select"
+        isoDate={endDate}
+        onChange={setEndDate}
+      />
+      {enableCompetitionListFetch && (
```
kr-matthews, line 53: Is this required? I think removing it and have `competitionList && ...` for the third case would be better, assuming it works.

app/webpacker/components/Panel/pages/RunValidatorsPage/ValidationListView.jsx
```diff
@@ -0,0 +1,56 @@
...
+  }
+};
+
+export default function ValidationListView({ validations, showCompetitionNameOnOutput, type }) {
+  const listByGroup = useMemo(() => _.groupBy(validations, 'kind'), [validations]);
+
+  return (
+    <>
+      {Object.entries(listByGroup).map(([group, list]) => (
+        <>
```
kr-matthews, line 22: I'm pretty sure the console will be warning you that you need a key here.

app/webpacker/components/Panel/pages/RunValidatorsPage/ValidationListView.jsx
```diff
@@ -0,0 +1,56 @@
...
+          <Header as="h5">{`${headingPrefixForType(type)} ${group}`}</Header>
+          <List bulleted>
+            {list.map((validationData) => (
+              <ListItem>
```
kr-matthews, line 26: And here.

app/webpacker/components/Panel/pages/RunValidatorsPage/RunValidatorsForm.jsx
danieljames-dj: Done

app/webpacker/components/Panel/pages/RunValidatorsPage/ValidationOutput.jsx
danieljames-dj: Oh yes, I missed it for some reason. Changed it.

danieljames-dj (commented):
Thanks for the detailed review, @gregorbg. One of the main change that I did post last commit is that I removed `result_validation_form`. I'm not sure if you are against it, but just to strengthen my argument, I like to highlight that the reason I felt to remove `result_validation_form` is because this was a helper model for the rails UI, and the rails UI no longer exists, and everything done inside `result_validation_form` can now be done in 2-3 lines.

app/webpacker/components/Panel/pages/RunValidatorsPage/CompetitionRangeSelector.jsx
danieljames-dj: Thanks for the suggestion, changed it.

app/webpacker/components/Panel/pages/RunValidatorsPage/ValidationListView.jsx
danieljames-dj: The console was filled with warnings and errors nowadays, so I'm not looking at it nowadays. Thanks for pointing it, I've made the change.

app/webpacker/components/Panel/pages/RunValidatorsPage/ValidationListView.jsx
danieljames-dj: Done

app/webpacker/components/Panel/pages/RunValidatorsPage/CompetitionRangeSelector.jsx
danieljames-dj: Done

app/webpacker/components/Panel/pages/RunValidatorsPage/CompetitionRangeSelector.jsx
danieljames-dj: Done

app/controllers/api/v0/competitions_controller.rb
```diff
@@ -37,6 +37,14 @@ def competition_index
              include: serial_includes
   end
 
+  def competition_count
+    start_date = params.require(:startDate)
+    end_date = params.require(:endDate)
+
+    count = Competition.where("start_date >= ? AND end_date <= ?", start_date, end_date).count
+    render json: { count: count }
+  end
```
gregorbg, line 46: This is a highly specific endpoint for internal uses, I'd rather not have that as part of the public API.

app/controllers/panel_controller.rb
```diff
@@ -50,6 +50,48 @@ def generate_db_token
...
+
+  private def competition_ids_in_range(range)
+    start_date = range[:startDate]
+    end_date = range[:endDate]
+    Competition.over
+               .not_cancelled
+               .where(start_date: start_date..)
+               .where(start_date: ..end_date)
+               .order(:start_date)
+               .ids
```
gregorbg, line 80: Can you combine this with the logic for `competition_count` above?
Like, have some `competitions_between(start, end)` method which returns `Competition.over.not_cancelled.where(...).order(...)` and then one endpoint returns `competitions_between(foo, bar).count` and the other one returns `competitions_between(foo, bar).ids`

gregorbg: This seems like the perfect use case for an AR `scope` by the way :)

app/controllers/panel_controller.rb
```diff
@@ -50,6 +50,48 @@ def generate_db_token
...
+  end
+
+  def validators_for_competition_list
+    competition_ids = params.require(:competitionIds).split(',')
+    validators_for_competition_ids(competition_ids)
+  end
+
+  def validators_for_competitions_in_range
+    range = JSON.parse(params.require(:competitionRange)).transform_keys(&:to_sym)
```
gregorbg, line 89: ...Why are you not simply passing in `startDate` and `endDate` as params directly?

app/webpacker/components/Panel/pages/RunValidatorsPage/CompetitionRangeSelector.jsx
```diff
@@ -0,0 +1,67 @@
...
+
+export default function CompetitionRangeSelector({ range, setRange }) {
+  const bothDatesAreSelected = Boolean(range?.startDate && range?.endDate);
+
+  const {
+    data: competitionCount, isLoading, isError,
+  } = useQuery({
+    queryKey: ['competitionCountInRange', range?.startDate, range?.endDate],
+    queryFn: () => getCompetitionCount(range?.startDate, range?.endDate),
+    enabled: bothDatesAreSelected,
```
gregorbg, line 19: If the query is only enabled when both dates are selected, then you should be able to drop the `?` at least from the `queryFn`

app/webpacker/components/Panel/pages/RunValidatorsPage/CompetitionRangeSelector.jsx
```diff
@@ -0,0 +1,67 @@
...
+import { Form } from 'semantic-ui-react';
+import { QueryClient, useQuery } from '@tanstack/react-query';
+import UtcDatePicker from '../../../wca/UtcDatePicker';
+import getCompetitionCount from './api/getCompetitionCount';
+import Loading from '../../../Requests/Loading';
+import Errored from '../../../Requests/Errored';
+
+const RUN_VALIDATORS_QUERY_CLIENT = new QueryClient();
+
+export default function CompetitionRangeSelector({ range, setRange }) {
```
gregorbg, line 11: Why are you passing in `range` instead of just passing the dates separately?

app/webpacker/components/Panel/pages/RunValidatorsPage/CompetitionRangeSelector.jsx
```diff
@@ -0,0 +1,67 @@
...
+        control={UtcDatePicker}
+        showYearDropdown
+        dateFormatOverride="yyyy-MM-dd"
+        dropdownMode="select"
+        isoDate={range?.startDate}
+        onChange={(date) => {
+          setRange({
+            ...range,
+            startDate: date,
+          });
```
gregorbg, line 35: See my point about passing the values separately

app/webpacker/components/Panel/pages/RunValidatorsPage/CompetitionRangeSelector.jsx
```diff
@@ -0,0 +1,67 @@
...
+        showYearDropdown
+        dateFormatOverride="yyyy-MM-dd"
+        dropdownMode="select"
+        isoDate={range?.endDate}
+        onChange={(date) => {
+          setRange({
+            ...range,
+            endDate: date,
+          });
+        }}
```
gregorbg, line 50: (ibid)

app/webpacker/components/Panel/pages/RunValidatorsPage/CompetitionRangeSelector.jsx
```diff
@@ -0,0 +1,67 @@
...
+      )}
+    </>
+  );
+}
+
+function CompetitionCountViewer({ isLoading, isError, competitionCount }) {
+  if (isLoading) return <Loading />;
+  if (isError) return <Errored />;
+  return `The checks will run for ${competitionCount} competitions`;
+}
```
gregorbg, line 67: This seems a bit redundant to put into an extra component. Either move the whole query down here (bonus: You can already be sure that both dates will definitely be selected!) or move this small text snippet up into the parent.

app/webpacker/components/Panel/pages/RunValidatorsPage/RunValidatorsForm.jsx
```diff
@@ -0,0 +1,158 @@
...
+import { ALL_VALIDATORS, VALIDATORS_WITH_FIX } from '../../../../lib/wca-data.js.erb';
+import { IdWcaSearch } from '../../../SearchWidget/WcaSearch';
+import SEARCH_MODELS from '../../../SearchWidget/SearchModel';
+import CompetitionRangeSelector from './CompetitionRangeSelector';
+import useCheckboxState from '../../../../lib/hooks/useCheckboxState';
+import runValidatorsForCompetitionList from './api/runValidatorsForCompetitionList';
+import runValidatorsForCompetitionsInRange from './api/runValidatorsForCompetitionsInRange';
+import ValidationOutput from './ValidationOutput';
+
+const validatorNameReadable = (validatorName) => validatorName.split('::')[1];
```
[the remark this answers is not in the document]
gregorbg: Hm, in your code you're still splitting at `::` (which not everybody will know what that is or where it's coming from) and you're skipping the first part (by accessing `[1]` directly, and ignoring `[0]`). So I insist on a short explanatory code comment :grin:

app/webpacker/components/Panel/pages/RunValidatorsPage/RunValidatorsForm.jsx
```diff
@@ -0,0 +1,165 @@
...
+
+const VALIDATOR_OPTIONS = ALL_VALIDATORS.map((validator) => ({
+  key: validator,
+  text: validatorNameReadable(validator),
+  value: validator,
+}));
+
+const COMPETITION_SELECTION_OPTIONS_TEXT = {
+  manual: 'Pick competition(s) manually',
+  range: 'Competition between dates',
```
gregorbg, line 27: Nit: Competition**s**

app/webpacker/components/Panel/pages/RunValidatorsPage/RunValidatorsForm.jsx
```diff
@@ -0,0 +1,165 @@
...
+    <WCAQueryClientProvider>
+      <RunValidatorsForm />
+    </WCAQueryClientProvider>
+  );
+}
+
+function RunValidatorsForm({ competitionIds }) {
+  const [
+    selectedCompetitionSelectionOption,
+    setSelectedCompetitionSelectionOption,
```
gregorbg, line 43: If the names get so long that you have to break them into two lines, you should probably think about shorter names.

app/webpacker/components/Panel/pages/RunValidatorsPage/RunValidatorsForm.jsx
```diff
@@ -0,0 +1,165 @@
...
+  // enableCompetitionEditor says whether competition list editor should be enabled or not. If the
+  // list of competitions is passed as parameter, then the editor need not be shown.
+  const enableCompetitionEditor = !competitionIds;
+
+  // Competition name needs to be shown on output only when the script is not ran just for a single
+  // competition.
+  const showCompetitionNameOnOutput = !(
+    selectedCompetitionSelectionOption === 'manual'
+    && selectedCompetitionIds.length === 1
+  );
```
gregorbg, line 88:
```suggestion
  const showCompetitionNameOnOutput =
    selectedCompetitionSelectionOption !== 'manual'
    || selectedCompetitionIds.length > 1;
```

app/webpacker/components/Panel/pages/RunValidatorsPage/RunValidatorsForm.jsx
```diff
@@ -0,0 +1,165 @@
...
+                    value={key}
+                    checked={selectedCompetitionSelectionOption === key}
+                    onChange={setSelectedCompetitionSelectionOption}
+                  />
+                </FormField>
+              ))}
+            </FormGroup>
+            {(
+              selectedCompetitionSelectionOption === 'manual'
+            ) && (
```
gregorbg, line 111: Why are you wrapping `()` around a single simple `===` check?

app/webpacker/components/Panel/pages/RunValidatorsPage/RunValidatorsForm.jsx
```diff
@@ -0,0 +1,165 @@
...
+                value={selectedCompetitionIds}
+                onChange={setSelectedCompetitionIds}
+                model={SEARCH_MODELS.competition}
+                multiple
+                required
+              />
+            )}
+            {(
+              selectedCompetitionSelectionOption === 'range'
+            ) && (
```
gregorbg, line 125: Again, useless wrapping the boolean check in `()`

app/webpacker/components/Panel/pages/RunValidatorsPage/ValidationListView.jsx
```diff
@@ -0,0 +1,56 @@
...
+import { Header, List, ListItem } from 'semantic-ui-react';
+import I18n from '../../../../lib/i18n';
+import { competitionUrl } from '../../../../lib/requests/routes.js.erb';
+
+const headingPrefixForType = (type) => {
+  switch (type) {
+    case 'info': return 'Information for';
+    case 'warning': return 'Warnings detected in';
+    case 'error': return 'Errors detected in';
+    default: return 'Unknown detected in';
```
gregorbg, line 12: Change this text to make it more obvious something wrong happened. Something along the lines `Please contact WST because an unknown response (${type}) was given in`

app/webpacker/components/Panel/pages/RunValidatorsPage/ValidationListView.jsx
gregorbg: Huh, interesting that the `[` above is acceptable but `]` is not. Are there HTML codes (similar to `&apost;`) that you could use?

app/webpacker/components/Panel/pages/RunValidatorsPage/ValidationOutput.jsx
```diff
@@ -0,0 +1,90 @@
...
+  validationOutput, isPending, isError, showCompetitionNameOnOutput,
+}) {
+  if (isPending) return <Loading />;
+  if (isError) return <Errored />;
+
+  if (!validationOutput) {
+    return (
+      <Message>Please run the validators to see the output.</Message>
+    );
+  }
```
gregorbg, line 17: What if the validators have been run but still didn't give any output?

app/webpacker/components/Panel/pages/RunValidatorsPage/ValidationListView.jsx
```diff
@@ -0,0 +1,56 @@
...
+  }
+};
+
+export default function ValidationListView({ validations, showCompetitionNameOnOutput, type }) {
+  const listByGroup = useMemo(() => _.groupBy(validations, 'kind'), [validations]);
+
+  return (
+    <>
+      {Object.entries(listByGroup).map(([group, list]) => (
+        <Fragment key={group}>
```
gregorbg, line 22: Instead of `Fragment`, you could consider wrapping this in a `Message`. SemUI Messages have `info`, `warning` and `error` states with nice color codings :)

app/webpacker/components/Panel/pages/RunValidatorsPage/ValidationOutput.jsx
```diff
@@ -0,0 +1,90 @@
...
+    <>
+      <p>Please pay attention to the warnings below:</p>
+      <ValidationListView
+        validations={validationOutput.warnings}
+        showCompetitionNameOnOutput={showCompetitionNameOnOutput}
+        type="warning"
+      />
+    </>
+  );
+}
```
gregorbg, line 90: Quite a lot of repetition, any chance to unify this?
You could consider changing the English texts a little, like `No messages of type ${type} detected in the results`, to make reusing the components a bit easier

app/webpacker/components/Panel/pages/RunValidatorsPage/api/getCompetitionCount.js
```diff
@@ -0,0 +1,7 @@
+import { fetchJsonOrError } from '../../../../../lib/requests/fetchWithAuthenticityToken';
+import { viewUrls } from '../../../../../lib/requests/routes.js.erb';
+
+export default async function getCompetitionCount(startDate, endDate) {
+  const { data } = await fetchJsonOrError(viewUrls.competitions.countInRange(startDate, endDate));
+  return data?.count;
```
gregorbg, line 6: Hm, now you're accessing the data already again. If only the count is so important, then why not just return it directly from the backend?
(Yes, numbers by themselves are valid JSON. Not everything has to be `{}` or `[]` on top level)

gregorbg (changes requested):

app/webpacker/components/Panel/pages/RunValidatorsPage/ValidationOutput.jsx
danieljames-dj: Okay done

app/webpacker/components/Panel/pages/RunValidatorsPage/CompetitionRangeSelector.jsx
gregorbg: If you want to maintain it as two separate states anyways, then that feels like a strong argument for passing them in as individual states as well.

> Hence I thought of passing the range as it is.
I don't understand the "hence", because it implies a logical connection. But if your original intention is to maintain one state each, then the logical thing would also be to pass one state each. (And that's exactly what I suggested in my original comment as well). Am I missing something?

app/webpacker/components/Panel/pages/RunValidatorsPage/ValidationListView.jsx
gregorbg: Whatever works for you! I won't block this PR only because of this design choice.
But I do want to point out that you can have a list inside the body of a `Message`: https://react.semantic-ui.com/collections/message/#types-list

app/webpacker/components/Panel/pages/RunValidatorsPage/CompetitionRangeSelector.jsx
danieljames-dj: As per our chat in slack, I've created `setStartDate` and `setEndDate` methods and used that instead of inline logic.

app/webpacker/components/Panel/pages/RunValidatorsPage/CompetitionRangeSelector.jsx
danieljames-dj: Changed.

app/webpacker/components/Panel/pages/RunValidatorsPage/CompetitionRangeSelector.jsx
danieljames-dj: Changed.

app/webpacker/components/Panel/pages/RunValidatorsPage/ValidationListView.jsx
danieljames-dj: Oh you meant to have one `Message` for one section right? That actually makes sense. I thought each point should have been a message, which I felt a bit as an overkill. I've made the change. Thank you.

app/webpacker/components/Panel/pages/RunValidatorsPage/CompetitionRangeSelector.jsx
kr-matthews: I haven't seen that slack discussion, but I don't understand the point of keeping the dates in a single range object/state. The 2 dates are passed individually to the back-end, and in the front-end the 2 dates are always updated separately (as far as I can see, correct me if I'm wrong). It's extra effort/code. It's quite minor though.

danieljames-dj: Sorry for not explaining it in detail here. In `RunValidatorsForm`, I'm having two ways of selecting competitions: through competition ID and through range. For each of these ways, I'm maintaining **one** state - `selectedCompetitionIds` and `selectedCompetitionRange`.

If I need to split `startDate` and `endDate`, I have to split them in `RunValidatorsForm` and I personally like not to split `selectedCompetitionRange` because in that class I want to consider that as a single state.

It's not like I won't split in `RunValidatorsForm`, but I prefer not to split. If you or Gregor still ask me to split, I'll definitely go ahead and split it.

app/controllers/api/v0/competitions_controller.rb
gregorbg: But you forgot to remove the code here :wink:

gregorbg (changes requested):

app/controllers/panel_controller.rb
```diff
@@ -50,6 +50,45 @@ def generate_db_token
...
+  private def validators_for_competition_ids(competition_ids)
+    validators = params.require(:selectedValidators).split(',').map(&:constantize)
+    apply_fix_when_possible = params.require(:applyFixWhenPossible)
+
+    results_validator = ResultsValidators::CompetitionsResultsValidator.new(
+      validators,
+      check_real_results: true,
+      apply_fixes: apply_fix_when_possible,
+    )
+    results_validator.validate(competition_ids)
```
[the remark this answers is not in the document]
gregorbg: > I didn't use the `result_validation_form.rb` because it was having too many things. I agree those were good, but I feel that if I forcefully use that model, it may not be so clean. So just took the above two lines from that.
> 
> I just now saw couple of validations from the model which are also relevant, I've now implemented those validations in the client side.
> 
> I understand the risk of `constantize`, I now see how the `result_validation_form` is doing it - I have now changed to a similar logic, so no more use of `constantize`.
> 
> Also, now I think technically `result_validation_form.rb` is not needed anymore as it was used to support the rails ERB code, and that code is no longer there. So I've removed `result_validation_form.rb` after making necessary changes.

Up at line 54 there is still one use of `constantize` :upside_down_face:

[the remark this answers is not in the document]
danieljames-dj: Oops I don't remember why I wrote "so no more use of `constantize`". I've changed it now.

gregorbg (approved):
Code fine for me, please merge and deploy this only when you have time to "monitor" the changes and react to potential crashes

PR #10654, opened 2025-01-19
Migrate anonymization script to react

danieljames-dj:
Currently there are two pages of anonymization - one for anonymizing profile (in rails) and another for anonymizing account (in react). This PR merges both to a single react script.


app/webpacker/components/Tickets/TicketWorkbenches/AnonymizationTicketWorkbenchForWrt/PerformManualChecks.jsx
```diff
@@ -0,0 +1,32 @@
...
+      <ListItem>
+        Request data removal from OAuth Access Grants.
+      </ListItem>
+      <ListItem>
+        Inspect external websites of competitions for data usage. If so, instruct the website to
+        remove the person&apos;s data.
+      </ListItem>
+      <ListItem>
+        For recently competed competitions (Past 1 month), after anonymizing the person&apos;s data,
+        synchrozie the results on WCA Live.
```
maxidragon, line 28:
```suggestion
        synchronize the results on WCA Live.
```

Maybe it's finally time to fix this typo?

app/webpacker/components/Tickets/TicketWorkbenches/AnonymizationTicketWorkbenchForWrt/PerformManualChecks.jsx
```diff
@@ -0,0 +1,32 @@
...
+      </ListItem>
+      <ListItem>
+        Request data removal from OAuth Access Grants.
+      </ListItem>
+      <ListItem>
+        Inspect external websites of competitions for data usage. If so, instruct the website to
+        remove the person&apos;s data.
+      </ListItem>
+      <ListItem>
+        For recently competed competitions (Past 1 month), after anonymizing the person&apos;s data,
```
maxidragon, line 27: Maybe it's worth mentioning that currently comps older than 3 months are removed automatically from Live?

app/controllers/tickets_controller.rb
```diff
@@ -97,4 +97,167 @@ def edit_person_validators
...
+              )
+            end
+          end
+        end
+        # Anonymize person's data in Persons for subid 1
+        person.update(
+          wca_id: new_wca_id,
+          name: User::ANONYMOUS_NAME,
+          gender: User::ANONYMOUS_GENDER,
+          dob: nil
```
maxidragon, line 231: I think it's better to use `User::ANONYMOUS_DOB` here rather than `nil`, as this will be raised in sanity checks (see Sanity Check - DECEMBER 2024).

app/controllers/tickets_controller.rb
```diff
@@ -97,4 +97,167 @@ def edit_person_validators
...
+            if p.countryId == current_country_id
+              p.delete
+            else
+              current_sub_id += 1
+              current_country_id = p.countryId
+              p.update(
+                wca_id: new_wca_id,
+                name: User::ANONYMOUS_NAME,
+                gender: User::ANONYMOUS_GENDER,
+                dob: nil,
```
maxidragon, line 220: See another comment about DOB, same here

app/webpacker/components/Panel/pages/AnonymizationScriptPage/index.jsx
```diff
@@ -1,25 +1,53 @@
...
-        model={SEARCH_MODELS.user}
+      <FormGroup grouped>
+        <div>Select where to search for</div>
+        {[SEARCH_MODELS.user, SEARCH_MODELS.person].map((searchModel, index) => (
+          <FormField key={searchModel}>
+            <Radio
+              label={MODEL_NAME[index]}
+              value={searchModel}
+              checked={model === searchModel}
+              onChange={setModel}
```
kr-matthews, line 29: You could both call `setModel` and reset the search input here. This would allow you to remove the effect above. Generally, if you can do something without an effect, then you should.

app/webpacker/components/Tickets/TicketWorkbenches/AnonymizationTicketWorkbenchForWrt/index.jsx
```diff
@@ -0,0 +1,48 @@
...
+  } = useQuery({
+    queryKey: ['anonymizeDetails', userId, wcaId],
+    queryFn: () => getDetailsBeforeAnonymization(userId, wcaId),
+    enabled: Boolean(userId || wcaId),
+  }, ANONYMIZATION_QUERY_CLIENT);
+
+  if (isLoading) return <Loading />;
+  if (isError) return <Errored />;
+
+  if (!userId && !wcaId) {
```
kr-matthews, line 26: The important thing is that `data` exists, right? (Which is equivalent, given the `enabled` value above, but that could change in the future.)
```suggestion
  if (!data) {
```

app/controllers/tickets_controller.rb
danieljames-dj: Thanks for pointing that, @maxidragon. I've fixed it as you suggested.

app/controllers/tickets_controller.rb

app/webpacker/components/Panel/pages/AnonymizationScriptPage/index.jsx
danieljames-dj: Okay done

app/webpacker/components/Tickets/TicketWorkbenches/AnonymizationTicketWorkbenchForWrt/PerformManualChecks.jsx
danieljames-dj: I'm not sure about that actually. Is that how it is happening? In that case, we need to do for past 3 months right?

app/webpacker/components/Tickets/TicketWorkbenches/AnonymizationTicketWorkbenchForWrt/PerformManualChecks.jsx
danieljames-dj: Good catch 😅 fixed.

app/webpacker/components/Tickets/TicketWorkbenches/AnonymizationTicketWorkbenchForWrt/index.jsx
danieljames-dj: Yes you are right, changed.

app/webpacker/components/Tickets/TicketWorkbenches/AnonymizationTicketWorkbenchForWrt/PerformManualChecks.jsx
maxidragon: Yup, it was implemented recently.

danieljames-dj: So that means "Past 1 month" needs to be replaced with "Past 3 months" and in brackets we can mention WCA Live will auto-clear after 3 months, am I right?

maxidragon: Yes, you're right

danieljames-dj: Done

app/webpacker/components/Panel/pages/AnonymizationScriptPage/index.jsx
```diff
@@ -1,25 +1,56 @@
...
+import { IdWcaSearch } from '../../../SearchWidget/WcaSearch';
 import SEARCH_MODELS from '../../../SearchWidget/SearchModel';
 import useInputState from '../../../../lib/hooks/useInputState';
-import AnonymizationTicketWorkbench from '../../../Tickets/TicketWorkbenches/AnonymizationTicketWorkbench';
+import AnonymizationTicketWorkbenchForWrt from '../../../Tickets/TicketWorkbenches/AnonymizationTicketWorkbenchForWrt';
+
+const MODEL_NAME = {
+  [SEARCH_MODELS.user]: 'User',
+  [SEARCH_MODELS.person]: 'Person',
+};
```
gregorbg, line 11: What about using the keys of this object as a separate `AVAILABLE_MODELS` const (name tentative)?
Would be useful in a few places further down

app/webpacker/components/Panel/pages/AnonymizationScriptPage/index.jsx
```diff
@@ -1,25 +1,56 @@
...
 
 export default function AnonymizationScriptPage() {
-  const [user, setUser] = useInputState();
+  const [model, setModel] = useInputState(SEARCH_MODELS.user);
```
gregorbg, line 14: If you have a constant list, you can just use `AVAILABLE_MODELS[0]` here

app/webpacker/components/Panel/pages/AnonymizationScriptPage/index.jsx
```diff
@@ -1,25 +1,56 @@
...
+  };
 
   return (
     <>
-      <WcaSearch
-        label="Enter the user to anonymize"
-        model={SEARCH_MODELS.user}
+      <FormGroup grouped>
+        <div>Select where to search for</div>
+        {[SEARCH_MODELS.user, SEARCH_MODELS.person].map((searchModel, index) => (
```
gregorbg, line 26: Here is another place where a list of models available for anon would help.

app/webpacker/components/Panel/pages/AnonymizationScriptPage/index.jsx
```diff
@@ -1,25 +1,56 @@
...
+      {model === SEARCH_MODELS.user && (
+        <AnonymizationTicketWorkbenchForWrt
+          userId={searchInput}
+        />
+      )}
+      {model === SEARCH_MODELS.person && (
+        <AnonymizationTicketWorkbenchForWrt
+          wcaId={searchInput}
+        />
+      )}
```
gregorbg, line 53:
```suggestion
      <AnonymizationTicketWorkbenchForWrt
          userId={model === SEARCH_MODELS.user && searchInput}
          wcaId={model === SEARCH_MODELS.person && searchInput}
        />
```

gregorbg: (You will need to try out whether that works as expected, though)

app/webpacker/components/Tickets/TicketWorkbenches/AnonymizationTicketWorkbenchForWrt/AnonymizeAction.jsx
```diff
@@ -0,0 +1,43 @@
...
+import { Button, Confirm, Message } from 'semantic-ui-react';
+import useSaveAction from '../../../../lib/hooks/useSaveAction';
+import Loading from '../../../Requests/Loading';
+import { actionUrls } from '../../../../lib/requests/routes.js.erb';
+
+export default function AnonymizeAction({ userId, wcaId }) {
+  const [confirmOpen, setConfirmOpen] = useState(false);
+  const [completed, setCompleted] = useState(false);
+
+  const { save, saving } = useSaveAction();
```
gregorbg, line 11: I think you can use Tanstack mutation here

app/webpacker/components/Tickets/TicketWorkbenches/AnonymizationTicketWorkbenchForWrt/PerformManualChecks.jsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import { List, ListItem } from 'semantic-ui-react';
+
+export default function PerformManualChecks() {
+  // TODO: Add necessary links in this page.
```
gregorbg, line 5: Is this TODO supposed to be here? Or did you just forget to take care of it before opening the PR?

app/webpacker/components/Tickets/TicketWorkbenches/AnonymizationTicketWorkbenchForWrt/VerifyAnonymizeDetails.jsx
```diff
@@ -0,0 +1,28 @@
...
+const anonymizationType = (data) => {
+  if (data.user_details && !data.person_details) {
+    return 'Account Only';
+  } if (data.person_details && !data.user_details) {
+    return 'Profile Only';
+  } if (data.user_details && data.person_details) {
+    return 'Account & Profile';
+  }
+  return 'Unknown';
+};
```
gregorbg, line 13: Just a minor comment, because it's not my power to decide: I am confused about the terminology here. In the entry point to the script (further up, where you have a search input) you call it "User" and "Person". Here, you suddenly talk about "Profile" and "Account". Maybe this PR is a good chance to clean up some terminology?

app/webpacker/components/Tickets/TicketWorkbenches/AnonymizationTicketWorkbenchForWrt/index.jsx
```diff
@@ -0,0 +1,48 @@
...
+
+const ANONYMIZATION_QUERY_CLIENT = new QueryClient();
+
+export default function AnonymizationTicketWorkbenchForWrt({ userId, wcaId }) {
+  const {
+    data, isLoading, isError,
+  } = useQuery({
+    queryKey: ['anonymizeDetails', userId, wcaId],
+    queryFn: () => getDetailsBeforeAnonymization(userId, wcaId),
+    enabled: Boolean(userId || wcaId),
```
gregorbg, line 20: I think `userId || wcaId` is already enough in this case, because of the `||` operator

app/webpacker/components/Tickets/TicketWorkbenches/AnonymizationTicketWorkbenchForWrt/index.jsx
```diff
@@ -0,0 +1,48 @@
...
+      <Header as="h5">Step 2: Review system generated checks</Header>
+      <ReviewSystemGeneratedChecks data={data} />
+
+      <Header as="h5">Step 3: Perform manual checks</Header>
+      <PerformManualChecks data={data} />
+
+      <Header as="h5">Step 4: Anonymize Action</Header>
+      <AnonymizeAction data={data} userId={userId} wcaId={wcaId} />
+
+    </Container>
```
gregorbg, line 46: Not sure if this is something that you would be interested in, but SemUI has this lovely `Step` component which would improve UX here a lot: https://react.semantic-ui.com/elements/step/

gregorbg: (I do appreciate that this is a WRT-only panel though, which means only ~0.005% of website users end up here. So I won't block this PR just because of this. But may still be useful though!)

app/controllers/tickets_controller.rb
gregorbg, line 1: Can you please give me a short summary of where these methods come from, and how much (if anything) was copy/pasted from the previous anon logic?

I can see that you deleted the existing model file that held all the backend steps logic, but I don't want to compare exactly each line from the old file and this file, especially since it seems that you have changed the flow a little bit

gregorbg (changes requested):

app/webpacker/components/Tickets/TicketWorkbenches/AnonymizationTicketWorkbenchForWrt/index.jsx
danieljames-dj: Okay done

app/webpacker/components/Tickets/TicketWorkbenches/AnonymizationTicketWorkbenchForWrt/AnonymizeAction.jsx
danieljames-dj: Done

danieljames-dj (commented):
Thanks for the review, @gregorbg. After making those changes, I realized that this can be done in bit more cleaner way by doing the major checks related to anonymizations in the server side. I've made that change and this has made some significant changes in the code compared to last review, sorry for that. But I hope you will find this approach more cleaner than earlier.

app/webpacker/components/Tickets/TicketWorkbenches/AnonymizationTicketWorkbenchForWrt/index.jsx
gregorbg: Huh, interesting

config/routes.rb
```diff
@@ -217,6 +216,10 @@
     get 'reassign-connected-wca-id' => 'admin#reassign_wca_id', as: :admin_reassign_wca_id
   end
   get 'panel-page/:id' => 'panel#panel_page', as: :panel_page
+  scope 'tickets' do
+    get 'details_before_anonymization' => 'tickets#details_before_anonymization', as: :tickets_details_before_anonymization
+    post 'anonymize' => 'tickets#anonymize', as: :tickets_anonymize
+  end
```
gregorbg, line 222: You can move this under `resources :tickets` below. No need to open the same scope twice.

app/controllers/tickets_controller.rb
```diff
@@ -106,4 +106,189 @@ def edit_person_validators
...
+      return true
+    end
+
+    if user.nil? && person.nil?
+      render status: :unprocessable_entity, json: {
+        error: "User ID and WCA ID is not provided.",
+      }
+      return true
+    end
+    false
```
gregorbg, line 141: Why are you returning `false` by default?

app/controllers/tickets_controller.rb
```diff
@@ -106,4 +106,189 @@ def edit_person_validators
...
+  end
+
+  def details_before_anonymization
+    user, person = get_user_and_person
```
gregorbg, line 145: Ruby methods don't start with `get_` or `set_` by convention. Plain Ruby would just be `def user_and_person`.
Since we're working in the context of Rails, where you're loading entities from a database, names like `find_user_and_person` or `user_and_person_from_params` are acceptable compromises.

app/controllers/tickets_controller.rb
```diff
@@ -106,4 +106,189 @@ def edit_person_validators
...
+    anonymization_checks = {}
+    message_args = {}
+    action_items = []
+    non_action_items = []
+
+    # 1. Check whether the user is currently banned or not.
+    anonymization_checks[:user_currently_banned] = user&.banned?
+
+    # 2. Check whether the user was banned in the past.
+    anonymization_checks[:user_banned_in_past] = user&.banned_in_past?
```
gregorbg, line 157: You can do this directly while creating the object above

```suggestion
    anonymization_checks = {
      user_currently_banned: user&.banned?,
      user_banned_in_past: user&.banned_in_past?,
    }
```

app/controllers/tickets_controller.rb
```diff
@@ -106,4 +106,189 @@ def edit_person_validators
...
+    anonymization_checks[:user_has_active_oauth_access_grants] = access_grants&.any?
+    message_args[:access_grants] = access_grants
+
+    # 8. Check whether there are any competitions with external websites.
+    competitions_with_external_website = person&.competitions&.select { |c| c.external_website.present? }
+    anonymization_checks[:competitions_with_external_website] = competitions_with_external_website&.any?
+    message_args[:competitions_with_external_website] = competitions_with_external_website
+
+    # 9. Check whether there are any recent competitions data to be removed from WCA Live.
+    anonymization_checks[:recent_competitions_data_to_be_removed_wca_live] = recent_competitions_3_months&.any?
```
gregorbg, line 188: Same here: You can compute this in-line (I just cannot be bothered to make code suggestions for every single item)

gregorbg: With that being said, there's definitely too much happening inside the controller here. Controllers are only "glue" between models and frontend views, so you can do some loading and a liiiitle bit of conditional fetching. But "strong" logic like this definitely belongs inside models.

I envision something along the lines of:
```
# In user.rb

def anonymization_checks
  {
    user_currently_banned: self.banned?,
    user_banned_in_past: self.banned_in_past?,
  }
end

# In person.rb
# ...similar definition as above

# In controller.rb
user_checks = user&.anonymization_checks || {}
person_checks = person&.anonymization_checks || {}

anon_checks = user_checks.merge(person_checks)
```

app/controllers/tickets_controller.rb
```diff
@@ -106,4 +106,189 @@ def edit_person_validators
...
+    render json: {
+      user: user,
+      person: person,
+      action_items: action_items,
+      non_action_items: non_action_items,
+      message_args: message_args,
+    }
+  end
+
+  def anonymize
```
gregorbg, line 201: Again, definitely too much business logic happening inside the controller. Please move this into either a module in `lib` or into methods in `person`/`user`. Whatever seems appropriate to you

app/controllers/tickets_controller.rb
```diff
@@ -106,4 +106,189 @@ def edit_person_validators
...
+
+    # 8. Check whether there are any competitions with external websites.
+    competitions_with_external_website = person&.competitions&.select { |c| c.external_website.present? }
+    anonymization_checks[:competitions_with_external_website] = competitions_with_external_website&.any?
+    message_args[:competitions_with_external_website] = competitions_with_external_website
+
+    # 9. Check whether there are any recent competitions data to be removed from WCA Live.
+    anonymization_checks[:recent_competitions_data_to_be_removed_wca_live] = recent_competitions_3_months&.any?
+
+    anonymization_checks.each { |key, value| (value ? action_items : non_action_items) << key }
```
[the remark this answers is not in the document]
gregorbg: If I'm reading this correctly, then I think you want https://docs.ruby-lang.org/en/3.4/Enumerable.html#method-i-partition

app/controllers/tickets_controller.rb
gregorbg: As a rule of thumb, code along the lines of
```ruby
model_id = params.require(:id)
my_model = MyModel.find(model_id)

if some_condition
  my_model.update(foo: ..., bar: ...)
else
  my_model.delete
end

return status: :ok
```
is generally acceptable in a controller. Our codebase isn't the perfect rolemodel everywhere (especially keep your fingers from `competitions_controller` in that regard!) but your body here is definitely too much.

app/webpacker/components/Panel/pages/AnonymizationScriptPage/index.jsx
```diff
@@ -1,24 +1,55 @@
...
+  },
+  {
+    id: SEARCH_MODELS.person,
+    name: 'Person',
+  },
+];
 
 export default function AnonymizationScriptPage() {
-  const [user, setUser] = useInputState();
+  const [activeModelIndex, setActiveModelIndex] = useInputState(0);
```
gregorbg, line 20: You can use this for a `const activeModel = useMemo(() => AVAILABLE_MODELS[activeModelIndex], [activeModelIndex]);`

app/webpacker/components/Tickets/TicketWorkbenches/AnonymizationTicketWorkbenchForWrt/ActionItems.jsx
```diff
@@ -0,0 +1,112 @@
...
+
+export default function ActionItems({ actionItemList, messageArgs }) {
+  return (
+    <List bulleted>
+      {actionItemList.map((actionItem) => (
+        <List.Item>
+          <ActionItemContent
+            actionItem={actionItem}
+            messageArgs={messageArgs}
+          />
```
gregorbg, line 13: I think you would be much better off letting I18n handle this. It would definitely save you from the `messageArgs` with the tons of `?` safe operations, and it would also allow you to unify `ActionItems` and `NonActionItems` like so:
```yml
en:
  anonymization:
    user_currently_banned:
      true: This user is currently banned and cannot be anonymized.
      false: This user is currently not banned.
```

app/webpacker/components/Tickets/TicketWorkbenches/AnonymizationTicketWorkbenchForWrt/index.jsx
```diff
@@ -0,0 +1,109 @@
...
+import NonActionItems from './NonActionItems';
+
+const ANONYMIZATION_QUERY_CLIENT = new QueryClient();
+
+const STEPS = [
+  '1. Verify details to be anonymized',
+  '2. Non-action items to review',
+  '3. Action items to perform',
+  '4. Anonymize Action',
+];
```
gregorbg, line 22: Are the numeral prefixes necessary? SemUI has `ordered` for the `Step` component: https://react.semantic-ui.com/elements/step/#groups-ordered, which should do the trick

spec/controllers/tickets_controller_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+      expect(person.reload.dob).to eq User::ANONYMOUS_DOB.to_date
+    end
+
+    it "can anonymize account data" do
+      user = FactoryBot.create(:user_with_wca_id)
+      FactoryBot.create(:result, person: user.person)
+
+      post :anonymize, params: { userId: user.id }
+
+      expect(!!response).to eq true
```
gregorbg, line 81: I know you're only copying the old test body, but for what it's worth you might also try and improve it:
`expect(response).to be_successful` is a good starting point

spec/controllers/tickets_controller_spec.rb
gregorbg, line 1: In general, `reload`ing is only necessary once:
```ruby
person = Person.new(not_anonymize)

post :anonymize_path, person_id: person.id

person.reload # Stores the new information in-memory for all `person` attributes
expect(person.name).to eq User::ANOONYMOUS_NAME
expect(person.wca_id).to include('ANON')
```

gregorbg: (Yes, I know you just copied old tests but if you're touching the code anyways then you could as well improve it)

gregorbg (changes requested):

app/controllers/tickets_controller.rb
```diff
@@ -229,66 +180,22 @@ def anonymize
...
-        )
+        user_to_anonymize.anonymize
       end
     end
 
+    puts('DJDJDJ start')
+    puts(person)
+    puts(person&.reload)
+    puts(person&.reload&.wca_id)
+    puts('DJDJDJ end')
```
gregorbg, line 194: :upside_down_face:

gregorbg (changes requested):

app/webpacker/components/Tickets/TicketWorkbenches/AnonymizationTicketWorkbenchForWrt/AnonymizeAction.jsx
```diff
@@ -0,0 +1,56 @@
+import { useMutation } from '@tanstack/react-query';
+import React, { useState } from 'react';
+import { Button, Confirm, Message } from 'semantic-ui-react';
+import Errored from '../../../Requests/Errored';
+import Loading from '../../../Requests/Loading';
+import anonymize from './api/anonymize';
+
+export default function AnonymizeAction({ userId, wcaId }) {
+  const [confirmOpen, setConfirmOpen] = useState(false);
+  const [completed, setCompleted] = useState(false);
```
kr-matthews, line 10: I think this can be replaced with `useMutation`'s `isSuccess`.

app/webpacker/components/Tickets/TicketWorkbenches/AnonymizationTicketWorkbenchForWrt/AnonymizeAction.jsx
```diff
@@ -0,0 +1,56 @@
...
+  const [newWcaId, setNewWcaId] = useState();
```
kr-matthews, line 11: And this can be replaced with the mutation's `data.new_wca_id`.

app/webpacker/components/Panel/pages/AnonymizationScriptPage/index.jsx
kr-matthews: I don't think `useMemo` is necessary. The computation is not expensive, and the model object is defined outside the component so will be the same object reference each render (unless the index changes).

danieljames-dj: Oops ya, you are right. Removed it.

app/webpacker/components/Tickets/TicketWorkbenches/AnonymizationTicketWorkbenchForWrt/AnonymizeAction.jsx
danieljames-dj: Oh that's nice, changed. Thanks.

app/webpacker/components/Tickets/TicketWorkbenches/AnonymizationTicketWorkbenchForWrt/AnonymizeAction.jsx
danieljames-dj: Oh that's nice too, changed. Thanks.

config/routes.rb
gregorbg: OK!

gregorbg (approved):

PR #10672, opened 2025-01-20
Add jsx to overcommit

FinnIckler:
this actually passes for me in overcommit locally, but as I understand it overcommit only runs on the files that are commited?

### DIFF
--- a/app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationList.jsx
+++ b/app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationList.jsx
@@ -1,4 +1,4 @@
-import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query';
+import { useMutation, useQuery } from '@tanstack/react-query';
 import React, {
   useMemo, useReducer, useRef,
 } from 'react';
@@ -120,8 +120,6 @@ export default function RegistrationAdministrationList({ competitionInfo }) {
     initialExpandedColumns,
   );
 
-  const queryClient = useQueryClient();
-
   const [editable, setEditable] = useCheckboxState(false);
 
   const dispatchStore = useDispatch();
@@ -199,7 +197,8 @@ export default function RegistrationAdministrationList({ competitionInfo }) {
             return a.user.wca_id.localeCompare(b.user.wca_id);
           }
           case 'country':
-            return countries.byIso2[a.user.country.iso2].name.localeCompare(countries.byIso2[b.user.country.iso2].name);
+            return countries.byIso2[a.user.country.iso2].name
+              .localeCompare(countries.byIso2[b.user.country.iso2].name);
           case 'events':
             return a.competing.event_ids.length - b.competing.event_ids.length;
           case 'guests':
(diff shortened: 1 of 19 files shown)


gregorbg:
The commit hook only runs for files that are part of that respective commit you're in the middle of committing.
As you can see, our CI runs on the whole codebase.

FinnIckler:
I fixed most of the linting errors. But we do still have a bunch of label related a11y errors. Not sure how to fix those because just adding a htmlFor doesn't work because of it being a custom component. So this requires actual code changes

FinnIckler:
Should I just do those changes in another PR?

gregorbg:
> Should I just do those changes in another PR?

For `RegistrationEditor`, the errors are actually trivial to fix since `Forn.TextArea` and other `Form.Foo` controls take a `label` property that you can use and which will internally wire everything correctly.

For `RegistrationOverview`, tbh I feel like these form fields shouldn't even be form fields because they are never interactive. Just view-only strings, which are probably best in a `<List>` anyways.

gregorbg:
And then as per your most recent commit there's still a handful other errors, which should be straight-forward to fix.

FinnIckler:
yeah, the other errors are in the same file as the label errors which why I couldn't fix them

FinnIckler:
Did the last two. I added a space between the event icons because the formatting was different in the list items

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationList.jsx
```diff
@@ -286,19 +285,21 @@ export default function RegistrationAdministrationList({ competitionInfo }) {
...
         competing: {
-          waiting_list_position: waitingSorted[result.destination.index].competing.waiting_list_position,
+          waiting_list_position: waitingSorted[result.destination.index]
+            .competing.waiting_list_position,
         },
       }],
     }, {
       onSuccess: () => {
-        // We need to get the information for all Competitors if you change the waiting list position
+        // We need to get the info for all Competitors if you change the waiting list position
```
gregorbg, line 302: This has to be the most sneaky way to drop below the 100 char radar that I've seen in the past few months

app/webpacker/components/Schedule/TimeZone.jsx
```diff
@@ -46,7 +46,7 @@ export default function TimeZoneSelector({
         search
         selection
         value={activeTimeZone}
-        onChange={(_, data) => setActiveTimeZone(data.value)}
+        onChange={(__, data) => setActiveTimeZone(data.value)}
```
gregorbg, line 49: Calling this `_e` should also work, and is at least not quite as "silent" as your double-underscore approach.

app/webpacker/components/Schedule/TimeZone.jsx
```diff
@@ -71,7 +71,7 @@ export default function TimeZoneSelector({
         <Checkbox
           label={I18n.t('competitions.schedule.timezone_follow_venue')}
           checked={followVenueSelection}
-          onChange={(_, data) => setFollowVenueSelection(data.checked)}
+          onChange={(__, data) => setFollowVenueSelection(data.checked)}
```
gregorbg, line 74: `_e`, see above.

gregorbg (changes requested):

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationList.jsx
FinnIckler: it was 101 chars before :sob:

gregorbg (approved):
Fine for this PR, but the CI also has some weaker warnings in the Overcommit output. Please address those in a follow-up PR as soon as is convenient.

app/webpacker/components/TeamsCommitteesCouncils/index.jsx
```diff
@@ -46,12 +46,12 @@ export default function TeamsCommitteesCouncils({ canViewPastRoles }) {
   const hashIsValid = Boolean(activeGroup);
 
   useEffect(
-    function sanitizeHash() {
```
kr-matthews, line 49: The linter doesn't allow named functions in effects? Naming the effect is useful.

PR #10673, opened 2025-01-21
Fixed 'Registrations' button on my comp page to redirect to correct page

pranav-027:
'Registrations' button on my competitions page was redirecting to the registrations page instead of edit/registrations page.


gregorbg (approved):
Would like some feedback from another person because I personally don't use this page so much so not sure if this is truly the intended behavior

maxidragon:
> Would like some feedback from another person because I personally don't use this page so much so not sure if this is truly the intended behavior

This link is visible only for organizers/delegates/admins so I think edit registrations page makes much more sense

pranav-027:
Hello @gregorbg, here is the code from the old ruby page: https://github.com/thewca/worldcubeassociation.org/blob/9ace921b464156342cb839d21754aa37601c6b42/app/views/competitions/_my_competitions_table.html.erb#L69

Which uses `competition_edit_registrations_path` for the registration button. I hope this clears your doubt.

PR #10674, opened 2025-01-21
Clean up old comp overview code

gregorbg:
The new overview page seems to be going well (enough), so it's time to rip out the old code!
Also, it will make https://github.com/thewca/worldcubeassociation.org/pull/10527 much easier to implement.

The only bit of non-delete changes that I did are old backwards compatibilities for PHP links. We would probably have received complaints by now if this was a wide-spread issue, but I thought why not, if it's cheap...


gregorbg:
Note: As per the diff, a handful of tests are removed without replacement. This is because the competitions frontend is now served by `Competition.search`, which is already tested in `api_competitions_controller_spec` under the `describe 'GET #index'` scenario.

The only difference with the `competition_index` endpoint for the comp overview is that the latter serializes less fields to save on runtime and bandwidth. But the internal `Competition.search` mechanism is exactly the same, so any tests would be mere duplicates.

kr-matthews (approved):
I can't comment on the back-end changes, but the other changes look fine and it runs locally.

PR #10675, opened 2025-01-21
Fix weird but necessary edge case in serializable_opts monkey-patch

gregorbg:
Follow-up to https://github.com/thewca/worldcubeassociation.org/pull/10575 to cover an edge-case where nested options are passed down as hash instead of array.

As noted in that other PR, it is a horrible, _horrible_ idea to squeeze-merge two different ORM models together, and it's an even worse idea to pass around serialization options from one to the other. But our APIs have grown to expect this behavior, so we cannot remove this serialization nightmare without breaking API contracts.

I don't expect anybody to review or even understand this patch, I really just need to see whether the CI passes.


gregorbg:
Nevermind, found a simpler solution. Still don't think I need a review as long as CI says OK

PR #10678, opened 2025-01-22
Clean up CountrySelector

gregorbg:
Use SemUI's out-of-the-box flag support, and move from custom `countryOptions` to a pre-defined list

### DIFF
--- a/app/webpacker/components/CountrySelector/CountrySelector.js
+++ b/app/webpacker/components/CountrySelector/CountrySelector.js
@@ -2,6 +2,7 @@ import React from 'react';
 import { Form } from 'semantic-ui-react';
 
 import { countries } from '../../lib/wca-data.js.erb';
+import I18n from '../../lib/i18n';
 
 const countryOptions = countries.real.map((country) => ({
   key: country.iso2,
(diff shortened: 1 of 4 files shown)


danieljames-dj (approved):
LGTM, and thanks for using it in many places. I see that you used it in EditPersonForm but not in EditProfileForm, I guess there is some limitations for EditProfileForm or was it just a miss?

(Sorry for having both EditPersonForm and EditProfileForm which names very similar. EditPersonForm is internal and is to actually make the edit, and EditProfileForm is external and is to request edits)

gregorbg:
> LGTM, and thanks for using it in many places. I see that you used it in EditPersonForm but not in EditProfileForm, I guess there is some limitations for EditProfileForm or was it just a miss?

I think both are using `CountrySelector`? The difference is probably that one is _already_ using it even before this PR, and the other one is being actively changed in this PR

danieljames-dj:
> > LGTM, and thanks for using it in many places. I see that you used it in EditPersonForm but not in EditProfileForm, I guess there is some limitations for EditProfileForm or was it just a miss?
> 
> I think both are using `CountrySelector`? The difference is probably that one is _already_ using it even before this PR, and the other one is being actively changed in this PR

Oops my bad. I was checking in my local, and it was couple of weeks old branch. The other CountrySelector was implemented a week ago.

PR #10680, opened 2025-01-22
Consistently send continent name in English to WRC API

gregorbg:
See title. Diff should also be self-explanatory.


dunkOnIT (approved):
We are so back

Issue #10596, opened 2025-01-10
/competitions page: Post-transition fixes


dunkOnIT:
## High 

- [x] Many reports of the competition date being returned in the wrong locale [#10599]
- [x] Search bar reloads after the first character typed, instead of waiting for "Enter"/search trigger
- [x] [fig 2] Time remaining to comp start not being passed in i18n key [#10527]

## Medium

- [ ] Sorting by "By announcement" used to NOT show older and completed competitions. It's no longer the case and it's annoying since few competitions get announced several months back and there's no option to filter out only ongoing and upcoming comps right now
- [x] (fig 1) In Admin View, "Results submitted" date always shows pending, even when results have been posted [#10682]

Fig 1:
![Image](https://github.com/user-attachments/assets/37d3e397-7a94-4885-be9c-099dae287730)

Fig 2:

![Image](https://github.com/user-attachments/assets/f16d5745-a5d9-4b26-b0ee-43e2601994c2)

kr-matthews:
> Search bar reloads after the first character typed, instead of waiting for "Enter"/search trigger

This is intentional. It's debounced, so if you type a word then it won't search until you stop typing (unless you type slowly). If this is causing issues (I don't think it is) then we could increase the debounce time.

dunkOnIT:
I definitely think it is causing issues - have you tried it on your end? Typing at a reasonable speed I can't get more than 1-3 characters in

kr-matthews:
https://github.com/user-attachments/assets/35ff4795-b88a-4fa2-859f-0c2ad39a51b8

Is it possible you're using the old competitions page? This was definitely a major issue there.

dunkOnIT:
I'm on Firefox, maybe it's more of an issue there? The UX isn't great, with it constantly refreshing/trying to search and defocusing the search bar. I'm not opposed to debouncing, but a longer debounce delay would make sense imo

https://github.com/user-attachments/assets/fe8db06a-cc34-403c-b323-9e33910f4d9d

gregorbg:
The fact that Firefox is even reloading the whole page in the first place is utterly wrong.

FinnIckler:
I was wondering that too, because I have the same reloading issue

kr-matthews:
@dunkOnIT does it happen when you run #10674? The old code would reload the page, I wonder if that's somehow still coming through on firefox?

dunkOnIT:
Yep it's working fine now that #10674 is deployed!

PR #10682, opened 2025-01-22
Fix dates in competitions admin view

maxidragon:
Fixes second medium sub-task from #10596

> In Admin View, "Results submitted" date always shows pending, even when results have been posted


gregorbg (approved):
I took the liberty of slimming down the serialization a bit, otherwise LGTM :cat:

PR #10683, opened 2025-01-22
Expose announced_at in non-admin view

gregorbg:
Otherwise, the return value is `null` and the I18N library interprets this as "variable missing in translation string".


kr-matthews (changes requested):
The date needs formatting
image

gregorbg:
Done! My computer won't let me take screenshots because as soon as I move the mouse to select the rectangle to crop, the popup also disappears :sweat_smile:

gregorbg:
Merging immediately as the review concern has been clearly addressed and this unblocks other PRs

PR #10685, opened 2025-01-24
WCA Live Integration Pre Work #2: Add liveResult and liveAttempt models

FinnIckler:
Also confirmed working using my test page. 
wca live's result table additionally has 
```
      t.integer :ranking
      t.integer :best, null: false
      t.integer :average, null: false
      t.string :single_record_tag, limit: 255
      t.string :average_record_tag, limit: 255
      t.boolean :advancing, default: false, null: false
      t.boolean :advancing_questionable, default: false, null: false
```
But I am not sure which one of these we should also use and which one we should just compute on the fly.


app/models/live_attempt.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+class LiveAttempt < ApplicationRecord
+  include Comparable
+
+  belongs_to :live_result
+  # If the Attempt has been replaced, it no longer points to a live_result, but instead is being pointed to
```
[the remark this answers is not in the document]
FinnIckler: We do not delete a `live_attempt` even if a user corrects it, in that case the corrected attempt links to the attempt that replaced it

FinnIckler:
I've had the factories and some basic tests. There will be a ton more when validating cutoffs and time limits will be introduced in a future pr

app/jobs/add_live_result_job.rb
```diff
@@ -4,8 +4,22 @@ class AddLiveResultJob < ApplicationJob
   self.queue_adapter = :shoryuken if WcaLive.sqs_queued?
   queue_as EnvConfig.LIVE_QUEUE if WcaLive.sqs_queued?
 
+  # params: { results, round_id, user_id, entered_by }
   def perform(params)
```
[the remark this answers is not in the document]
gregorbg: (that's a feature of `ActiveJob` specifically, not a general Ruby thing)

app/jobs/add_live_result_job.rb
```diff
@@ -4,8 +4,22 @@ class AddLiveResultJob < ApplicationJob
...
-    ActionCable.server.broadcast(WcaLive.broadcast_key(params[:round_id]),
-                                 { attempts: params[:results], user_id: params[:user_id] })
+    results = params[:results]
+    attempts = results.map.with_index(1) { |r, i| LiveAttempt.build(result: r, attempt_number: i) }
+    round = Round.find(params[:round_id])
+    event = round.event
+    format = round.format
+
+    r = Result.build({ value1: results[0], value2: results[1] || 0, value3: results[2] || 0, value4: results[3] || 0, value5: results[4] || 0, event_id: event.id, round_type_id: round.round_type_id, format_id: format.id })
```
gregorbg, line 15: Why are you using `build` instead of `new` here?

app/models/live_result.rb
```diff
@@ -0,0 +1,52 @@
...
+    include: %w[event live_attempts round],
+  }.freeze
+
+  def serializable_hash(options = nil)
+    super(DEFAULT_SERIALIZE_OPTIONS.merge(options || {}))
+  end
+
+  def result_id
+    id
+  end
```
gregorbg, line 28: alias please

app/models/live_result.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+class LiveResult < ApplicationRecord
+  has_many :live_attempts, -> { where(replaced_by: nil).order(:attempt_number) }
+
+  after_save :notify_users
+
+  belongs_to :registration
+
+  belongs_to :entered_by, class_name: 'User', foreign_key: 'entered_by_id'
```
gregorbg, line 10: I think Rails is smart enough to infer the `foreign_key`, as long as it's "name of relation + _id"

app/models/live_result.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  belongs_to :round
+
+  has_one :event, through: :round
+
+  DEFAULT_SERIALIZE_OPTIONS = {
+    only: %w[ranking registration_id round live_attempts round best average single_record_tag average_record_tag advancing advancing_questionable entered_at entered_by_id],
+    methods: %w[event_id attempts result_id],
+    include: %w[event live_attempts round],
```
gregorbg, line 19: This is a heck of a lot of data you're serializing for each and every single scorecard. I won't stand in your way for your PoC experiments, but please think about how to slim this down. Especially for `event`, just sending along `event_id` and accessing the static list of WCA events in JS should be enough.

For the round, I'd also argue that a round ID should do, and you keep a lookup somewhere else in your React component.

db/migrate/20250124154917_create_live_result_tables.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+class CreateLiveResultTables < ActiveRecord::Migration[7.2]
+  def change
+    create_table :live_results do |t|
+      t.references :registration, null: false, index: true
+      t.references :round, null: false, index: true
+      t.references :entered_by, type: :integer, null: false, foreign_key: { to_table: :users }
+      t.datetime :entered_at, null: false
```
gregorbg, line 9: Is it sufficient to keep this information per scorecard?
I'm thinking about FMC and MBLD where you enter results with long gaps in between, and these gaps also make it plausible that individual scores are entered by different people.

gregorbg: Also related about the timestamp: I assume this is the time when the scorecard was first inserted, but what if we just change one score later? Where is that being recorded?

db/migrate/20250124154917_create_live_result_tables.rb
```diff
@@ -0,0 +1,30 @@
...
+      t.integer :best, null: false
+      t.integer :average, null: false
+      t.string :single_record_tag, limit: 255
+      t.string :average_record_tag, limit: 255
+      t.boolean :advancing, default: false, null: false
+      t.boolean :advancing_questionable, default: false, null: false
+      t.timestamps
+    end
+
+    add_index :live_results, [:registration_id, :round_id], unique: true
```
gregorbg, line 20: I think this can just be `t.index` inside of `create_table`, but feel free to prove me wrong

db/schema.rb
```diff
@@ -1084,6 +1115,7 @@
     t.text "round_results", size: :medium
     t.integer "total_number_of_rounds", null: false
     t.string "old_type", limit: 1
+    t.boolean "is_open", default: false, null: false
```
gregorbg, line 1118: I don't see a corresponding migration for this. Potential whoopsie from your other PR?

db/migrate/20250124154917_create_live_result_tables.rb
FinnIckler: I was thinking I would just overwrite entered_by and entered_at with the latest person. I think that's how jonatan does it

db/schema.rb
FinnIckler: yeah, that should probably go into the open round pr instead

app/jobs/add_live_result_job.rb
FinnIckler: when using new it gives me an error saying `Found extra argument. Required 0 ` I think because there is no initializer?

gregorbg (approved):

PR #10688, opened 2025-01-26
Fix psych sheet crash when selecting event while loading

kr-matthews:
I think it would be a lot cleaner to have separate components for the registrations list and the psych sheet, now that they're significantly different (originally, it was just contracting a bunch of columns in the middle into one column, basically). I'll do that in a follow-up PR, unless anybody raises a concern.

Edit: to reproduce crash, go to worlds competitor list and click an event in the event selector before the initial data finishes loading.


gregorbg (approved):
No concerns re your proposal to have separate components!

PR #10719, opened 2025-01-27
Bump Ruby to version 3.4.1

gregorbg:
See title.


FinnIckler:
Ruby 3.4 is not supported for lambda yet https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html
so you can't change the lambda gemfile

FinnIckler (approved):

Issue #10503, opened 2025-01-01
Mega-issue: Registration Email Changes


dunkOnIT:
This is a combination of a number of different issues: #7388, #6822, #5328, #3121

- [ ] Overall: Once the below issues are handled, there should be a general review of the emails we send - ensuring that all status changes are covered.

## High
- [x] Change Pending Email away from the "Waiting List" email (currently the user is sent the "Waiting List" email when they get moved to competing_status: pending - but only when moved _back_ to pending, not when they initially sign up)
- [x] When the user re-registers after cancelling, just send the same "You've registered!" email that we did the first time - currently we send them the bad "Waiting List" email

## Medium: 
- [ ] Consider adding a delay (~2-5min?) when sending status change updates to competitors, in case the organizer has to move them around a bit/makes a mistake the first time they move them (this should make it so that only the user's _final_ status is emailed, instead of receiving one email for every status change)
- [ ] Waiting List email: Currently we don't send any email when the user is placed on the waiting list. The email should: Tell the user they're on the waiting list, what this means, and their current numbered position on the waiting list [#6822]
- [ ] Email the user telling them their registration is cancelled when they cancel their registration - currently we don't send any notification of this to the user
- [ ] Allow organizers to specify a cancellation reason when cancelling/rejecting a registration, which gets sent to the user in the status change email [#5328]
- [ ] Notify organizers of edits to a user's registration
- [ ] Include the comment of a registration in the notification email when a user cancels their registration [#7388]
- [ ] Create a differentiated "registration rejected" email

## Low: 
- [ ] Give organizers greater control over email notifications: Allow them to either choose a batched email, turn emails on/off, or only sign up for certain notifications (first implement auto-accept, as this removes a lot of the need for organizers to receive emails saying that a user has signed up)  [#3121, #7388]

dunkOnIT:
#10720 addresses high priority issues

dunkOnIT:
Reopening as there are still other unaddressed items here

PR #10720, opened 2025-01-27
Bring registration emails in line with v2 statuses

dunkOnIT:
Addresses #10503

This addresses a few different issues: 
- The email we were sending people when they were moved from Cancelled -> Pending was an old email that still viewed the Pending list as the Waiting List - I've removed that
- When moved from Cancelled -> Pending, a user should get a fresh "you just registered" email, not a "you're now pending" email - fixed that
  -  Same should apply for Rejected, so I added that case
- We weren't sending anyone an email when they got moved to the _actual_ Waiting List - so I've added an email for that

## TODO: 
- [x] Get tests passing
- [x] Add mailer tests for waitlisted email

### DIFF
--- a/spec/mailers/registrations_mailer_spec.rb
+++ b/spec/mailers/registrations_mailer_spec.rb
@@ -26,7 +26,7 @@ RSpec.describe RegistrationsMailer, type: :mailer do
       # scope this call, but rather compare the result to the expected locale.
       expect(mail_new.subject).to eq(I18n.t('registrations.mailer.new.mail_subject', comp_name: registration.competition.name, locale: :fr))
       expect(mail_accepted.subject).to eq(I18n.t('registrations.mailer.accepted.mail_subject', comp_name: registration.competition.name, locale: :fr))
-      expect(mail_waitlisted.subject).to eq(I18n.t('registrations.mailer.waitlisted.mail_subject', comp_name: registration.competition.name, locale: :fr))
+      expect(mail_waitlisted.subject).to eq(I18n.t('registrations.mailer.waiting_list.mail_subject', comp_name: registration.competition.name, locale: :fr))
       expect(mail_deleted.subject).to eq(I18n.t('registrations.mailer.deleted.mail_subject', comp_name: registration.competition.name, locale: :fr))
     end
 


spec/mailers/registrations_mailer_spec.rb
```diff
@@ -15,7 +15,6 @@
     let(:registration) { FactoryBot.create(:registration, user: french_user, competition: competition_with_organizers) }
     let(:mail_new) { RegistrationsMailer.notify_registrant_of_new_registration(registration) }
     let(:mail_accepted) { RegistrationsMailer.notify_registrant_of_accepted_registration(registration) }
-    let(:mail_pending) { RegistrationsMailer.notify_registrant_of_pending_registration(registration) }
```
gregorbg, line 18: Why is this disappearing without replacement? There is `notify_registrant_of_waitlisted_registration` now, which you could well put here.

spec/requests/api_registrations_spec.rb
```diff
@@ -164,6 +164,49 @@
       registration = Registration.find_by(user_id: user.id, competition_id: favourites_reg.competition.id)
       expect(registration.event_ids.sort).to eq(new_event_ids.sort)
     end
+
+    it 'user gets registration email if they cancel and re-register' do
+      perform_enqueued_jobs do
```
gregorbg, line 169: You only need to wrap `perform_enqueued_jobs` around those parts of your test that actually enqueue the job. So in this case, it would probably be the `patch` below.

gregorbg: Another option is to just write your test as usual, and call `perform_enqueued_jobs` directly as a method to "flush" the job queue, without passing a `do` block

spec/requests/api_registrations_spec.rb
```diff
@@ -164,6 +164,49 @@
...
+
+        expect(response.status).to eq(200)
+
+        email = ActionMailer::Base.deliveries.last
+        expect(email.subject).to eq(I18n.t('registrations.mailer.new.mail_subject', comp_name: registration.competition.name))
+      end
+    end
+
+    it 'user gets registration email if they were rejected and get moved to pending' do
+      perform_enqueued_jobs do
```
gregorbg, line 190: Same as above.

gregorbg (changes requested):

spec/requests/api_registrations_spec.rb
dunkOnIT: Done

spec/requests/api_registrations_spec.rb
dunkOnIT: Done

spec/mailers/registrations_mailer_spec.rb
dunkOnIT: The header test seems to rely on a french translation key which we don't have - should I add it in in order for the test to pass? I can find some french people to consult with

gregorbg: Nah, just a comment that we should add this as soon as French is translated should suffice

gregorbg: On the other hand, French does get updated rather infrequently.
If you do indeed have contacts to French speakers (ideally, cubers) it would be great if you could add the relevant snippets.

gregorbg (changes requested):
Just the merge conflict and learning French, then we should be good to go :D

spec/mailers/registrations_mailer_spec.rb
dunkOnIT: Asked for a translation from some French delegates :+1:

dunkOnIT: And merge is resolved

gregorbg (approved):

PR #10727, opened 2025-01-28
Refactor psych sheet

kr-matthews:
To separate the psych sheet code from the competitor list code, and to isolate the UI from the (conditional) logic.

It's nearly possible to fully separate the psych sheet data from the registrations data... except that the former is missing the country iso2. If it had that, then the queries/resulting data could be fully separated too, which would be nice.

Besides the 2 items below, the functionality should be identical.

Bonus: Fix a bug where if you initially click on an event column header, the psych sheet will load with question marks instead of data.

Also: Hide the 'clear' button from the event selector when there's nothing to clear.


gregorbg:
Should I push some code to this PR to expose the countryIso2 in the psych sheet data?

gregorbg:
> Should I push some code to this PR to expose the countryIso2 in the psych sheet data?

Since it was literally a one-liner, I just went ahead and pushed that. The psych sheet data returned by the existing endpoint will now also contain `country_iso2` in addition to `country_id`.

kr-matthews:
> > Should I push some code to this PR to expose the countryIso2 in the psych sheet data?
> 
> Since it was literally a one-liner, I just went ahead and pushed that. The psych sheet data returned by the existing endpoint will now also contain `country_iso2` in addition to `country_id`.

Thanks, but it didn't work. I fixed though (hopefully properly).

I still have some more follow-up refactoring to do in this PR, later tonight.

kr-matthews:
I don't know what to do about the lint errors. We can have non-camel-case properties directly from the back-end, but not directly create/modify them in the front-end? I just want to apply a mapper to the psych sheet query data so that it's the same format as the registrations data, allowing easy re-use of helper functions.

gregorbg:
Huh, I see that I messed up the header but `country_iso2` should really work. Let me see what's going on there...
(Also while I'm at that I will take a look at the linting issue)

gregorbg:
I've pushed a poor man's solution for the linting problem. ESLint is happy with _reading_ from non-camel-case properties, he just doesn't like you _storing_ them in non-camel-case variables.

(Also, just for reference, yes we will use "proper" serialization with snakeCase properties in our frontend data bindings once we migrate to React. Everything happening in the `webpacker` folder is a pretty "intermittent" solution anyways)

gregorbg (approved):

PR #10728, opened 2025-01-28
Add show_reg_status to URL params in competitions overview

maxidragon:


gregorbg:
Thanks, but this is kinda obsolete since we're planning on removing the checkbox and just showing the reg status all the time in https://github.com/thewca/worldcubeassociation.org/pull/10518

maxidragon:
Oops, I was searching if there are any others PR's about that and somehow missed that one. I'll close this PR

PR #10731, opened 2025-01-28
Remove 2 unused routes

danieljames-dj:
These two routes were introduced in https://github.com/thewca/worldcubeassociation.org/pull/7726, but looks like they are not used anywhere (99%, leaving the remaining 1% to the reviewer 😬), so removing it.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -291,8 +291,6 @@ Rails.application.routes.draw do
   get '/admin/do_generate_public_export' => 'admin#do_generate_public_export'
   get '/admin/override_regional_records' => 'admin#override_regional_records'
   post '/admin/override_regional_records' => 'admin#do_override_regional_records'
-  get '/admin/finish_persons' => 'admin#finish_persons'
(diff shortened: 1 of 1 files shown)


gregorbg:
Why do you believe that these routes are unused?
`Create Newcomers` is still very much relevant for WRT overall, isn't it?

danieljames-dj:
> Why do you believe that these routes are unused? `Create Newcomers` is still very much relevant for WRT overall, isn't it?

I was doing the react migration for create newcomers entry page, and I was going through the whole flow, and couldn't find the usages of the mentioned URLs anywhere. I couldn't find them in the https://github.com/thewca/worldcubeassociation.org/pull/7726 as well. So I thought it's unused.

gregorbg:
I see that you just opened that other migration PR.

May I suggest you close this and just remove the routes over in that other PR? That way, if CI and manual checks on the other PR work, we can be reasonably confident that the routes are indeed unused.

danieljames-dj:
Okay, done. I raised this separately because this was not part of 'migration'.

PR #10732, opened 2025-01-28
Migrate Create newcomers entry page to react

danieljames-dj:
This PR aims at converting the create-newcomers page to react


app/webpacker/components/Panel/pages/CreateNewComersPage/index.jsx
gregorbg, line 1: You can definitely use SemUI components here.

app/webpacker/components/Panel/PanelPages.jsx
```diff
@@ -169,7 +169,7 @@ export default {
   },
   [PANEL_PAGES.createNewComers]: {
     name: 'Create New Comers',
-    link: createNewComersUrl,
+    component: CreateNewComersPage,
```
gregorbg, line 172: Not 100% sure but I think the correct capitalization is just "Newcomers".

app/controllers/admin_controller.rb
```diff
@@ -418,22 +418,17 @@ def do_anonymize_person
     render 'anonymize_person'
   end
 
-  def finish_unfinished_persons
-    @finish_persons = FinishPersonsForm.new(
-      competition_ids: params[:competition_ids] || nil,
-    )
+  private def competition_list_from_string(competition_ids_string)
+    (competition_ids_string || '').split(',').uniq.compact
```
gregorbg, line 422: You can make sure that there's always a valid string being passed to this method via stuff like `params.fetch(:foo, 'my_default')`

app/webpacker/components/Panel/pages/CreateNewComersPage/CompetitionsInput.jsx
```diff
@@ -0,0 +1,39 @@
...
+        multiple
+        value={competitionIds}
+        onChange={setCompetitionIds}
+        label="Competition(s)"
+      />
+      <Header as="h4">
+        <HeaderSubheader>
+          Leave blank to check for all competitions
+        </HeaderSubheader>
+      </Header>
```
gregorbg, line 29: Creating a header only for showing a subheader (with no "main" header) smells like bad code to me. What is your goal here, and why is a simple text paragraph not good enough?

gregorbg (changes requested):

app/webpacker/components/Panel/PanelPages.jsx
danieljames-dj: Oh yes, changed. (there are some places where it's still wrong, I'll change that in a separate PR after this is merged, because those are not related to migration)

app/webpacker/components/Panel/pages/CreateNewComersPage/CompetitionsInput.jsx
danieljames-dj: I wanted to show it as 'hint' like thing, so I followed your suggestion in another PR ([LINK](https://github.com/thewca/worldcubeassociation.org/pull/10652#discussion_r1922373248)). Did I implement this wrongly?

app/webpacker/components/Panel/pages/CreateNewComersPage/index.jsx
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
+import { Header } from 'semantic-ui-react';
+import CompetitionsInput from './CompetitionsInput';
+
+export default function CreateNewComersPage() {
+  return (
+    <>
+      <Header>Create New Comers</Header>
```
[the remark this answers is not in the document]
danieljames-dj: Done

app/webpacker/components/Panel/pages/CreateNewComersPage/index.jsx
danieljames-dj: Done, I assume <p> doesn't have any alternative and hence can be used.

app/webpacker/components/Panel/PanelPages.jsx
```diff
@@ -169,7 +169,7 @@ export default {
   },
   [PANEL_PAGES.createNewComers]: {
     name: 'Create New Comers',
```
maxidragon, line 171:
```suggestion
     name: 'Create newcomers',
```
This should be changed as well ;-)

danieljames-dj: Okay now since you said it, I've changed it :-)

As mentioned in https://github.com/thewca/worldcubeassociation.org/pull/10732#discussion_r1932464070, I wanted to change them separately all together in another PR after this is merged. For example, the line above the line you highlighted - for that to be changed, I need to change at many places which will make this PR more confusing and those changes will not be aligned to the PR title.

app/views/admin/_import_inbox_step.erb
```diff
@@ -49,7 +49,7 @@
         All newcomers seem to have been assigned WCA IDs.
         Please <%= link_to "delete the inbox rows", competition_admin_delete_inbox_data_path(@competition, model: :inbox_person), class: 'inbox-trigger btn btn-primary', data: { jquery_method: :delete } %>.
       <% else %>
-        Please <%= link_to "assign WCA IDs to newcomers", admin_complete_persons_path(finish_persons_form: { competition_ids: @competition.id }), target: '_blank', class: 'btn btn-primary' %>
+        Please <%= link_to "assign WCA IDs to newcomers", admin_complete_persons_path(competition_ids: [@competition.id]), target: '_blank', class: 'btn btn-primary' %>
```
gregorbg, line 52: Why do you need to wrap this in an array? Passing it as a string should be just fine

app/views/admin/complete_persons.html.erb
```diff
@@ -1,18 +1,18 @@
 <% provide(:title, 'Finish unfinished persons') %>
 
 <div class="container">
-  <% competition_ids = @finish_persons.competition_ids %>
+  <% competition_ids = @competition_ids %>
```
gregorbg, line 4: This line is redundant now, if you already have a `@competition_ids` variable then use it everywhere.

gregorbg: No need to re-assign it to "itself" with another name

app/views/admin/complete_persons.html.erb
```diff
@@ -1,18 +1,18 @@
...
 
   <h2>
     <%= yield(:title) %>
-    <%= link_to "Go back", admin_finish_unfinished_persons_path(competition_ids: competition_ids), class: "btn btn-default" %>
+    <%= link_to "Go back", panel_page_path(id: User.panel_pages[:createNewComers], competition_ids: competition_ids), class: "btn btn-default" %>
```
gregorbg, line 8: Does this work as expected when `competition_ids` is an already-split array?

gregorbg (changes requested):

app/views/admin/_import_inbox_step.erb
gregorbg: Well, right here you're passing it to a `*_path` Rails helper, which encodes it into a URL anyways. So when reading it again in the controller, won't it end up as a string in each case?

Sorry my previous comment wasn't specific, but what I'm surprised about is the fact that you explicitly wrap it in an array _just before encoding it into a string URL_.

What is the difference in terms of the resulting URL anyways?

app/views/admin/complete_persons.html.erb
```diff
@@ -1,25 +1,23 @@
 <% provide(:title, 'Finish unfinished persons') %>
 
 <div class="container">
-  <% competition_ids = @finish_persons.competition_ids %>
-
   <h2>
     <%= yield(:title) %>
-    <%= link_to "Go back", admin_finish_unfinished_persons_path(competition_ids: competition_ids), class: "btn btn-default" %>
+    <%= link_to "Go back", panel_page_path(id: User.panel_pages[:createNewComers], competition_ids: @competition_ids.join(',')), class: "btn btn-default" %>
```
gregorbg, line 6: In regards to my other comment above: Kindly pointing out that here, you're fine with comma-glueing, but above in the other URL helper you insist on array-wrapping :wink:

app/views/admin/complete_persons.html.erb
```diff
@@ -1,25 +1,23 @@
...
-      <% @finish_persons.competitions.each do |competition_id| %>
+      <% @competition_ids.each do |competition_id| %>
         <li><code><%= competition_id %></code></li>
       <% end %>
     </ul>
   <% end %>
 
   <%= simple_form_for :person_completions, url: admin_complete_persons_path do |f| %>
-    <%= f.hidden_field :competition_ids, value: competition_ids %>
+    <%= f.hidden_field :competition_ids, value: @competition_ids %>
```
gregorbg, line 20: Again, this creates an `input` field in the DOM tree. How is this being handled if `@competition_ids` is an array?

gregorbg: (In other words: Should you manually glue them together with `join(',')` or is that not necessary?

app/webpacker/components/Panel/pages/CreateNewcomersPage/CompetitionsInput.jsx
```diff
@@ -0,0 +1,39 @@
...
+      />
+      <Header as="h4">
+        <HeaderSubheader>
+          Leave blank to check for all competitions
+        </HeaderSubheader>
+      </Header>
+      <Button
+        primary
+        size="big"
+        href={viewUrls.admin.completePersons(competitionIds.length > 0 ? competitionIds : null)}
```
gregorbg, line 33: Why is your `jsonToQueryString` in the JS route definition not clever enough to catch the case when `competitionIds` is an empty array?

gregorbg: (and can you adjust it easily to make it clever?)

gregorbg (changes requested):

app/views/admin/complete_persons.html.erb
danieljames-dj: Thanks for the catch. I fixed it now.

I tested with one competition and no competition. When I input no competition, it will run this script for all competitions, and I wrongly assumed that multiple competitions case will also be covered there.

app/webpacker/components/Panel/pages/CreateNewcomersPage/CompetitionsInput.jsx
danieljames-dj: I just realized that `jsonToQueryString` was able to handle empty array, so I removed the check now. :-)

app/webpacker/components/Panel/pages/CreateNewcomersPage/CompetitionsInput.jsx
```diff
@@ -0,0 +1,39 @@
...
+} from 'semantic-ui-react';
+import { IdWcaSearch } from '../../../SearchWidget/WcaSearch';
+import SEARCH_MODELS from '../../../SearchWidget/SearchModel';
+import { viewUrls } from '../../../../lib/requests/routes.js.erb';
+import useQueryParams from '../../../../lib/hooks/useQueryParams';
+import useInputState from '../../../../lib/hooks/useInputState';
+
+export default function CompetitionsInput() {
+  const [queryParams] = useQueryParams();
+  const competitionIdsFromQuery = queryParams?.competition_ids?.split(',').filter(Boolean);
```
gregorbg, line 13:
```suggestion
  const competitionIdsFromQuery = queryParams?.competition_ids?.split(',')?.filter(Boolean);
```

gregorbg (changes requested):
Nitpick about a null-safe in JS, otherwise looking good!

app/webpacker/components/Panel/pages/CreateNewcomersPage/CompetitionsInput.jsx
danieljames-dj: Done

gregorbg (approved):

PR #10734, opened 2025-01-29
Add name validations for edit person tickets

danieljames-dj:
Currently the tickets page shows DOB validations, this PR aims at showing name validations as well.

### DIFF
--- a/lib/results_validators/persons_validator.rb
+++ b/lib/results_validators/persons_validator.rb
@@ -37,6 +37,14 @@ module ResultsValidators
       true
     end
 
+    def self.roman_readable_part(name)
+      if name.include? " ("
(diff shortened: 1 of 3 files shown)


lib/results_validators/persons_validator.rb
gregorbg, line 1: Can I trust you that you only moved code around, not invented new code? (except for creating `roman_readable_part`, which is technically a new method but actually it's just reusing code so it's fine...)

Don't feel like comparing every single line

danieljames-dj: Yes, you can trust. I didn't change any logic.

gregorbg (approved):

Issue #10409, opened 2024-12-09
Grammatically incorrect error message


OWelton-Rosie:
**Describe the bug**
After playing around with the WCA Website and trying to access the delegate Panel by entering the URL worldcubeassociation.org/panel, I was greeted by an error message that said, "_You are not allowed to has permission_." 

**To Reproduce**
1. Sign out of your WCA account
2. Go to [worldcubeassociation.org/panel](worldcubeassociation.org/panel)
3. See error

**Screenshots**
![Screenshot 2024-12-10 11 27 45 AM](https://github.com/user-attachments/assets/e579d973-fba7-4928-a680-2bad0a99e50e)

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser: Chrome
 - Version: Unsure, but it's the latest version of Chrome. 

**Additional context**
- I'm using the WCA Website in English. 
- I realise that this is a very minor change, but I'm unsure of how to resolve it.

dunkOnIT:
Anyone wanting to work on this: It's not just changing an i18n string, it appears to originate in `application_controller#redirect_to_root_unless_user` - whatever action it is getting passed isn't combining nicely with the partial flash message defined in that method

pranav-027:
Hello @dunkOnIT I looked into it the code. It converts the action into a string, in this case, has_permission? -> "has permission" and combines it with "You are not allowed to". 

What would be the best solution here? I can think of either 

- Refactoring the method name (doesn't sound like a nice idea given it has wider usage)
- Having a custom error message for this has_permission? action

dunkOnIT:
Hi Pranav - I like the idea of having a custom error message if `has_permission` is passed. Let's go ahead with that!

PR #10735, opened 2025-01-29
Fixed grammatically incorrect message while accessing pages without permission

pranav-027:
Fixes: https://github.com/thewca/worldcubeassociation.org/issues/10409

image

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -96,7 +96,7 @@ class ApplicationController < ActionController::Base
       if redirecting
         flash[:danger] = case action
                          when :has_permission?
-                           t("errors.messages.no_permission.no_permission")
+                           t("errors.messages.no_permission")
                          else
                            "You are not allowed to #{action.to_s.sub(/^can_/, '').chomp('?').humanize.downcase}"
                          end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -96,7 +96,7 @@ class ApplicationController < ActionController::Base
       if redirecting
         flash[:danger] = case action
                          when :has_permission?
-                           t("application.no_permission")
+                           t("errors.messages.no_permission.no_permission")
                          else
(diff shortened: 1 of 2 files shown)


config/locales/en.yml
```diff
@@ -2879,3 +2879,5 @@ en:
       missing_scrambles_for_multi_error: "[%{round_id}] While you may have multiple groups in 3x3x3 Multi-Blind, at least one of the groups must contain scrambles for all attempts."
       multiple_fmc_groups_warning: "[%{round_id}] There are multiple groups of FMC used. If one group of FMC was used, please use the Scrambles Matcher to uncheck the unused scrambles. Otherwise, please include a comment to WRT explaining why multiple groups of FMC were used."
       wrong_number_of_scramble_sets_error: "[%{round_id}] This round has a different number of scramble sets than specified on the Manage Events tab. Please adjust the number of scramble sets in the Manage Events tab to the number of sets that were used."
+  application:
```
gregorbg, line 2882: You already have the `common` and `homepage` keys. There is also `errors.messages`. Feel free to use either of those, no need to introduce a new "application" key.

gregorbg (changes requested):

config/locales/en.yml
pranav-027: Done

app/controllers/application_controller.rb
```diff
@@ -94,7 +94,12 @@ def doorkeeper_unauthorized_render_options(error: nil)
     def redirect_to_root_unless_user(action, *)
       redirecting = !current_user&.send(action, *)
       if redirecting
-        flash[:danger] = "You are not allowed to #{action.to_s.sub(/^can_/, '').chomp('?').humanize.downcase}"
+        flash[:danger] = case action
```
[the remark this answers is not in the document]
pranav-027: Oops. I have changed it to if now

gregorbg (approved):

Issue #10562, opened 2025-01-07
Automatically add an administrative note when a formerly-banned competitor registers


dunkOnIT:

slongbehn:
Please make sure this is NOT visible by organizers!

dunkOnIT:
This is currently not possible - we'd need to add a new, delegates-only field to the Registration (which is possible). Before doing that, I just want to check if this is something WDC definitely wants? Or if there's another way you'd like us to notify delegates (eg email)

slongbehn:
1. This is a highly requested feature from delegates, and this would massively improve the support they’re receiving from the WIC. 
2. It does not need to be realized in this way. Personally, I think it would be easier if listed delegates were emailed a notice. How feasible is that?
3. For future use, we want to add warnings to a different list that uses this same feature. Like “This competitor has received an official warning from The WIC.” It’s very difficult for delegates to be informed of the competitor travels a lot.

pranav-027:
Can I work on this? Or is it still blocked?

PR #10738, opened 2025-01-29
Email delegates when a formerly-banned competitor registers

pranav-027:
Fixes: https://github.com/thewca/worldcubeassociation.org/issues/10562

Email look like this:
image

The red area is competitor name and WCA ID

### DIFF
--- a/app/mailers/registrations_mailer.rb
+++ b/app/mailers/registrations_mailer.rb
@@ -84,10 +84,12 @@ class RegistrationsMailer < ApplicationMailer
   def notify_delegates_of_formerly_banned_user_registration(registration)
     @registration = registration
     to = registration.competition.competition_delegates.map(&:user).map(&:email)
-    mail(
-      to: to,
-      reply_to: UserGroup.teams_committees_group_wic.metadata.email,
-      subject: "A formerly-banned competitor just registered for #{registration.competition.name}",
-      ) unless to.empty?
+    unless to.empty?
+      mail(
+        to: to,
+        reply_to: UserGroup.teams_committees_group_wic.metadata.email,
+        subject: "A formerly-banned competitor just registered for #{registration.competition.name}"
+      )
+    end
   end
 end
--- a/app/mailers/registrations_mailer.rb
+++ b/app/mailers/registrations_mailer.rb
@@ -88,7 +88,7 @@ class RegistrationsMailer < ApplicationMailer
       mail(
         to: to,
         reply_to: UserGroup.teams_committees_group_wic.metadata.email,
-        subject: "A formerly-banned competitor just registered for #{registration.competition.name}"
+        subject: "A formerly-banned competitor just registered for #{registration.competition.name}",
       )
     end
   end
--- a/app/mailers/registrations_mailer.rb
+++ b/app/mailers/registrations_mailer.rb
@@ -88,12 +88,11 @@ class RegistrationsMailer < ApplicationMailer
     if to.empty?
       nil
     else
-    mail(
-      to: to,
-      reply_to: "integrity@worldcubeassociation.org",
-      subject: "A formerly-banned competitor just registered for #{registration.competition.name}",
-    )
+      mail(
+        to: to,
+        reply_to: "integrity@worldcubeassociation.org",
+        subject: "A formerly-banned competitor just registered for #{registration.competition.name}",
+      )
     end
   end
-
 end
--- a/app/mailers/registrations_mailer.rb
+++ b/app/mailers/registrations_mailer.rb
@@ -80,4 +80,20 @@ class RegistrationsMailer < ApplicationMailer
       subject: "Unlock your new account on the WCA website",
     )
   end
+
+  def notify_delegates_of_formerly_banned_user_registration(registration)
+    @registration = registration
+    delegate_ids = registration.competition.competition_delegates.map(&:delegate_id)
+    to = User.where(id: delegate_ids).map(&:email)
+    if to.empty?
+      nil
+    else
+    mail(
+      to: to,
+      reply_to: "integrity@worldcubeassociation.org",
+      subject: "A formerly-banned competitor just registered for #{registration.competition.name}",
+    )
+    end
+  end
+
 end
--- a/app/mailers/registrations_mailer.rb
+++ b/app/mailers/registrations_mailer.rb
@@ -83,16 +83,11 @@ class RegistrationsMailer < ApplicationMailer
 
   def notify_delegates_of_formerly_banned_user_registration(registration)
     @registration = registration
-    delegate_ids = registration.competition.competition_delegates.map(&:delegate_id)
-    to = User.where(id: delegate_ids).map(&:email)
-    if to.empty?
-      nil
-    else
-      mail(
-        to: to,
-        reply_to: "integrity@worldcubeassociation.org",
-        subject: "A formerly-banned competitor just registered for #{registration.competition.name}",
-      )
-    end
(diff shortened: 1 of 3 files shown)


app/views/registrations_mailer/notify_delegates_of_formerly_banned_user_registration.erb
```diff
@@ -0,0 +1,12 @@
...
+  <%= @registration.name %>
+  <% if @registration.user.wca_id %>
+    (<%= wca_id_link @registration.user.wca_id -%>)
+  <% end %>
+  just registered for <%= @registration.competition.name %>.
+</p>
+
+<p>
+You can check the ban scope and reason
+  <%= link_to "here", panel_index_url(panel_id: :delegate, anchor: User.panel_pages[:bannedCompetitors]) %>
```
danieljames-dj, line 11: Instead of `panel_index_url`, you can use `panel_page_path(id: User.panel_pages[:bannedCompetitors]`.

app/mailers/registrations_mailer.rb
```diff
@@ -80,4 +80,19 @@ def notify_registrant_of_locked_account_creation(user, competition)
...
     )
   end
+
+  def notify_delegates_of_formerly_banned_user_registration(registration)
+    @registration = registration
+    delegate_ids = registration.competition.competition_delegates.map(&:delegate_id)
+    to = User.where(id: delegate_ids).map(&:email)
+    if to.empty?
+      nil
+    else
```
danieljames-dj, line 90: Maybe `unless to.empty?`

app/mailers/registrations_mailer.rb
```diff
@@ -80,4 +80,19 @@ def notify_registrant_of_locked_account_creation(user, competition)
       subject: "Unlock your new account on the WCA website",
     )
   end
+
+  def notify_delegates_of_formerly_banned_user_registration(registration)
+    @registration = registration
+    delegate_ids = registration.competition.competition_delegates.map(&:delegate_id)
```
danieljames-dj, line 86: You can directly get the delegate users here itself by doing `competition_delegates.map(&:user)`. And actually even email by adding `.map(&:email)`.

app/mailers/registrations_mailer.rb
```diff
@@ -80,4 +80,19 @@ def notify_registrant_of_locked_account_creation(user, competition)
...
+    to = User.where(id: delegate_ids).map(&:email)
+    if to.empty?
+      nil
+    else
+      mail(
+        to: to,
+        reply_to: "integrity@worldcubeassociation.org",
```
danieljames-dj, line 93: You can use `UserGroup.teams_committees_group_wic.metadata.email` to get the email.

danieljames-dj (commented):
Didn't check completely, nor did check what are the test failures. I'm leaving comments for which I've reviewed.

app/views/registrations_mailer/notify_delegates_of_formerly_banned_user_registration.erb
pranav-027: ActionView::Template::Error: undefined method `panel_page_path'

app/mailers/registrations_mailer.rb
pranav-027: done

app/mailers/registrations_mailer.rb
pranav-027: done

app/mailers/registrations_mailer.rb
pranav-027: Done, I saw some other mailers using direct emails I'll fix them also in different PR

app/views/registrations_mailer/notify_delegates_of_formerly_banned_user_registration.erb
danieljames-dj: Oh sorry, even `panel_page_url(id: User.panel_pages[:bannedCompetitors]` not working?

pranav-027: This worked thanks!

pranav-027:
> Didn't check completely, nor did check what are the test failures. I'm leaving comments for which I've reviewed.

@danieljames-dj this might explain earlier test failures: https://github.com/thewca/worldcubeassociation.org/pull/10736

danieljames-dj (approved):

lib/registrations/lanes/competing.rb
```diff
@@ -19,6 +19,10 @@ def self.process!(lane_params, user_id, competition_id)
         registration.save!
         RegistrationsMailer.notify_organizers_of_new_registration(registration).deliver_later
         RegistrationsMailer.notify_registrant_of_new_registration(registration).deliver_later
+        user = User.find_by_id(user_id)
```
gregorbg, line 22: No need to load the user manually from the DB, since you already have the registration ready at this point. So accessing `registration.user` should be perfectly fine.

lib/registrations/lanes/competing.rb
```diff
@@ -19,6 +19,10 @@ def self.process!(lane_params, user_id, competition_id)
...
+        if user&.banned_in_past?
```
gregorbg, line 23: In line with my other comment above: Why are you using null-safe access here? We don't allow registrations without a user account, so (unless I' overlooking something super-obvious) `user` should always be defined and not-null at this point.

gregorbg (changes requested):

lib/registrations/lanes/competing.rb
pranav-027: Thanks for reminding of this this has been fixed

lib/registrations/lanes/competing.rb
pranav-027: I had thought what if find_by_id fails to retrieve User. But as its not used anymore this has been removed

app/views/registrations_mailer/notify_delegates_of_formerly_banned_user_registration.erb
```diff
@@ -0,0 +1,12 @@
...
+  <%= @registration.name %>
+  <% if @registration.user.wca_id %>
+    (<%= wca_id_link @registration.user.wca_id -%>)
+  <% end %>
+  just registered for <%= @registration.competition.name %>.
+</p>
+
+<p>
+You can check the ban scope and reason
+  <%= link_to "here", panel_page_url(id: User.panel_pages[:bannedCompetitors]) %>
```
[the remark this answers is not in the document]
pranav-027: Yes! confirmed on staging using a Trainee delegates account.

app/views/registrations_mailer/notify_delegates_of_formerly_banned_user_registration.erb
```diff
@@ -0,0 +1,12 @@
+<p>
+  <%= @registration.name %>
+  <% if @registration.user.wca_id %>
+    (<%= wca_id_link @registration.user.wca_id -%>)
```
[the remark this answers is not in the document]
pranav-027: Actully I had no idea it was copy pasted from:
https://github.com/thewca/worldcubeassociation.org/blob/4eaeb2d13d28b0b1a4ef182d41737b39357ed8f7/app/views/registrations_mailer/notify_organizers_of_new_registration.html.erb#L4

But I searched about it and found that it ensures that no extra whitespace is added after the line.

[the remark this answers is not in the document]
gregorbg: Aha! Thanks for looking that up :)

gregorbg (approved):

PR #10750, opened 2025-02-03
Default datepicker to iso-8601 date format

dunkOnIT:
It seems that react-datepicker internally defaults to en-US locale if no dateFormatOverride is given. Given that we have an international userbase and date formats can be ambiguous, I believe iso8601 is a more appropriate default.

I've taken the simplest approach - manually replacing the fallback to datepicker's default with a fallback to "yyyy-MM-dd". There are other options: 
- always pass the user's locale to datepicker
- change datepicker's internal default


gregorbg:
I'd argue passing the locale makes more sense... Oh look! We're already doing that :D
https://github.com/thewca/worldcubeassociation.org/blob/main/app/webpacker/lib/i18n.js#L4

dunkOnIT:
Lol, missed that. In that case, the question becomes: Why are non-US users being shown US formats? My guess would be that our only locale option is `English`, which means `US English` - if so, would we be open to adding a `British English` locale?

PR #10765, opened 2025-02-03
Prevent user from registering if registration has passed

dunkOnIT:
Prevents payment after registration close in 3 ways: 
1. Disables and marks inactive the payment window - so that the user can't access the payment UI 
2. Renders a "can't pay" message instead of the payment panel - in case the user gets to the payment panel, they can't use it
3. Rejects payment_ticket endpoint requests `if registration_past?` - so that if the user does somehow submit a request, it won't work

### DIFF
--- a/app/models/competition.rb
+++ b/app/models/competition.rb
@@ -1831,10 +1831,12 @@ class Competition < ApplicationRecord
                              reverse, field = part.match(/^(-)?(\w+)$/).captures
                              [field.to_sym, reverse ? :desc : :asc]
                            end
+                           # rubocop:disable Style/HashSlice
                            #   RuboCop suggests using `slice` here, which is a noble intention but breaks the order
                            #   of sort arguments. However, this order is crucial (sorting by "name then start_date"
                            #   is different from sorting by "start_date then name") so we insist on doing it our way.
                            .select { |field, _| orderable_fields.include?(field) }
+                           # rubocop:enable Style/HashSlice
                            .to_h
     else
       order = { start_date: :desc }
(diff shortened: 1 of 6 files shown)


app/models/competition.rb
```diff
@@ -1831,12 +1831,10 @@ def self.search(query, params: {}, managed_by_user: nil)
                              reverse, field = part.match(/^(-)?(\w+)$/).captures
                              [field.to_sym, reverse ? :desc : :asc]
                            end
-                           # rubocop:disable Style/HashSlice
```
dunkOnIT, line 1834: This change was not intentional - I believe it was made by rubocop?

dunkOnIT: (there is a matching change a few lines lower down)

app/webpacker/components/RegistrationsV2/Register/PaymentStep.jsx
```diff
@@ -86,6 +88,14 @@ export default function PaymentStep({
     setIsLoading(false);
   };
 
+  if (!competitionInfo['registration_currently_open?']) {
+    return (
+      <Message color="red">
+        <I18nHTMLTranslate i18nKey="registrations.payment_form.errors.registration_closed" />
```
gregorbg, line 94: You don't need the component if you don't inject any HTML fragments. Your new i18n key is just a plain and simple string without any replacements, so `I18n.t('foo.bar')` will do

gregorbg (changes requested):

app/webpacker/components/RegistrationsV2/Register/PaymentStep.jsx
dunkOnIT: Fixed!

gregorbg (approved):

PR #10767, opened 2025-02-04
Dues estimator to reduce currencylayer usage

danieljames-dj:
For getting currency details, we moved to currencylayer API which is paid. Currently, the dues is shown on competition edit page, but it may not be noticed every time by the user. So it has been replaced with a button now as shown below:

Screenshot 2025-02-04 at 06 33 28

When the user clicks the button, they will be able to see the dues estimate. This view not only shows the dues estimate, but also allows user to change the competitor count temporarily and see what is the dues for that competitor count:

Screenshot 2025-02-04 at 06 35 50

The future plan of this component will be to explain the dues calculation method used in this component, so that delegate can easily understand it from here, and hence reduce queries to WFC on 'why is the dues xyz for n competitors'.

### DIFF
--- a/app/webpacker/components/CompetitionForm/FormSections/DuesEstimate.jsx
+++ b/app/webpacker/components/CompetitionForm/FormSections/DuesEstimate.jsx
@@ -1,0 +1,51 @@
+import React from 'react';
+import { Input, Modal } from 'semantic-ui-react';
+import { QueryClient, useQuery } from '@tanstack/react-query';
+import useInputState from '../../../lib/hooks/useInputState';
+import { calculateDuesUrl } from '../../../lib/requests/routes.js.erb';
+import { currenciesData } from '../../../lib/wca-data.js.erb';
+import Loading from '../../Requests/Loading';
+import Errored from '../../Requests/Errored';
+import { fetchJsonOrError } from '../../../lib/requests/fetchWithAuthenticityToken';
+
+const CALCULATE_DUES_QUERY_CLIENT = new QueryClient();
+
+export default function DuesEstimate({
+  close, countryId, currencyCode, baseEntryFee, competitorLimit,
+}) {
+  const [competitorCount, setCompetitorCount] = useInputState(competitorLimit || 0);
+  const currencyInfo = currenciesData.byIso[currencyCode] || currenciesData.byIso.USD;
+
+  const {
+    data, isLoading, isError,
+  } = useQuery({
+    queryKey: [countryId, currencyCode, baseEntryFee],
+    queryFn: () => fetchJsonOrError(calculateDuesUrl(countryId, currencyCode, baseEntryFee)),
+  }, CALCULATE_DUES_QUERY_CLIENT);
+
+  const { dues_value: duesValue } = data?.data || {};
+
+  if (isLoading) return <Loading />;
+  if (isError) return <Errored error="Dues fetching failed..." />;
+
+  return (
+    <Modal
+      open
+      onClose={close}
+      closeIcon
+    >
+      <Modal.Header>Dues Estimate</Modal.Header>
+      <Modal.Content>
+        <Input
+          label="Competitor count"
+          type="number"
+          value={competitorCount}
+          onChange={setCompetitorCount}
+        />
+        <p>
+          {`Dues for ${competitorCount} competitors (approximately): ${currencyInfo?.symbol || ''}${(duesValue * competitorCount).toFixed(2)}`}
+        </p>
+      </Modal.Content>
+    </Modal>
+  );
+}
--- a/app/webpacker/components/CompetitionForm/FormSections/DuesEstimate.jsx
+++ b/app/webpacker/components/CompetitionForm/FormSections/DuesEstimate.jsx
@@ -19,7 +19,7 @@ export default function DuesEstimate({
   const {
     data, isLoading, isError,
   } = useQuery({
-    queryKey: [countryId, currencyCode, baseEntryFee],
+    queryKey: ['dues-estimate', countryId, currencyCode, baseEntryFee],
     queryFn: () => fetchJsonOrError(calculateDuesUrl(countryId, currencyCode, baseEntryFee)),
   }, CALCULATE_DUES_QUERY_CLIENT);
 
--- a/app/webpacker/components/CompetitionForm/FormSections/DuesEstimate.jsx
+++ b/app/webpacker/components/CompetitionForm/FormSections/DuesEstimate.jsx
@@ -24,6 +24,7 @@ export default function DuesEstimate({
   }, CALCULATE_DUES_QUERY_CLIENT);
 
   const { dues_value: duesValue } = data?.data || {};
+  const totalDues = (duesValue * competitorCount || 0) / currencyInfo.subunitToUnit;
 
   if (isLoading) return <Loading />;
   if (isError) return <Errored error="Dues fetching failed..." />;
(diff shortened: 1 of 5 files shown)


app/webpacker/components/CompetitionForm/FormSections/DuesEstimate.jsx
```diff
@@ -0,0 +1,48 @@
...
+import Errored from '../../Requests/Errored';
+
+export default function DuesEstimate({
+  close, countryId, currencyCode, baseEntryFee, competitorLimit,
+}) {
+  const [competitorCount, setCompetitorCount] = useInputState(competitorLimit || 0);
+  const currencyInfo = useMemo(
+    () => (currenciesData.byIso[currencyCode] || currenciesData.byIso.USD),
+    [currencyCode],
+  );
```
kr-matthews, line 17: I don't think there's any need to memoize this.

app/webpacker/components/CompetitionForm/FormSections/DuesEstimate.jsx
```diff
@@ -0,0 +1,48 @@
...
+
+  const {
+    data, loading, error,
+  } = useLoadedData(calculateDuesUrl(countryId, currencyCode, baseEntryFee));
```
kr-matthews, line 21: Can we use `useQuery` here? `useLoadedData` has some issues and I think the plan is to replace it at some point anyway.

app/webpacker/components/CompetitionForm/FormSections/DuesEstimate.jsx
danieljames-dj: Okay removed.

app/webpacker/components/CompetitionForm/FormSections/DuesEstimate.jsx
danieljames-dj: Okay, replaced with `useQuery`. I actually moved the API call from `RegistrationFees` component.

app/webpacker/components/CompetitionForm/FormSections/DuesEstimate.jsx
```diff
@@ -0,0 +1,51 @@
...
+import { Input, Modal } from 'semantic-ui-react';
+import { QueryClient, useQuery } from '@tanstack/react-query';
+import useInputState from '../../../lib/hooks/useInputState';
+import { calculateDuesUrl } from '../../../lib/requests/routes.js.erb';
+import { currenciesData } from '../../../lib/wca-data.js.erb';
+import Loading from '../../Requests/Loading';
+import Errored from '../../Requests/Errored';
+import { fetchJsonOrError } from '../../../lib/requests/fetchWithAuthenticityToken';
+
+const CALCULATE_DUES_QUERY_CLIENT = new QueryClient();
```
kr-matthews, line 11: Usually we wrap the component in a query client. I don't really know what the difference is though.

app/webpacker/components/CompetitionForm/FormSections/DuesEstimate.jsx
```diff
@@ -0,0 +1,51 @@
...
+export default function DuesEstimate({
+  close, countryId, currencyCode, baseEntryFee, competitorLimit,
+}) {
+  const [competitorCount, setCompetitorCount] = useInputState(competitorLimit || 0);
+  const currencyInfo = currenciesData.byIso[currencyCode] || currenciesData.byIso.USD;
+
+  const {
+    data, isLoading, isError,
+  } = useQuery({
+    queryKey: ['calculateDues'],
```
kr-matthews, line 22: Does this work? The keys need to include everything that the query depends on. Otherwise, if the query is re-run with a different say base entry fee, then it will retrieve the previous cached data (based on the previous base entry fee), since the keys match. I think.

danieljames-dj: Thanks for that. I've changed it, I'm pretty new to useQuery so didn't knew the theory much.

Apparently this was working with constant queryKey as well apparently. I changed the registration fees and opened the component again, and the output is different, and not from cached. I'm not sure why, maybe in local there is no cache.

kr-matthews: I think that since the query client is re-created every time the component is first rendered (since it's defined in the dues estimate file rather than wrapping the entire form), so it's a fresh instance with no cache each time? And you can't change any of the keys without first closing the component and then re-opening it fresh.

Generally, I think the best approach is a single query client at the root of the app, but we can't do that since rails is rendering react components individually, so we don't have a single root in the react world (if I understand correctly).

danieljames-dj: That makes sense. I guess these will be changed after the nextJS integration.

app/webpacker/components/CompetitionForm/FormSections/DuesEstimate.jsx
gregorbg: We wrap it everywhere we can. To my knowledge, there is one instance of a `const` client being passed around, which is `WcaSearch` (and `IdWcaSearch` in the same file). That's because these are form components where we cannot be sure in which contexts they are used, and under what circumstances a client may or may not be available.

app/webpacker/components/CompetitionForm/FormSections/DuesEstimate.jsx
```diff
@@ -0,0 +1,57 @@
...
+export default function DuesEstimate({
+  close, countryId, currencyCode, baseEntryFee, competitorLimit,
+}) {
+  const [competitorCount, setCompetitorCount] = useInputState(competitorLimit || 0);
+  const currencyInfo = currenciesData.byIso[currencyCode] || currenciesData.byIso.USD;
+
+  const {
+    data, isLoading, isError,
+  } = useQuery({
+    queryKey: [countryId, currencyCode, baseEntryFee],
```
gregorbg, line 22: Super nitpick, but you should definitely still include something like `'dues-estimate'` or a similar string literal in the key.

If you only pass in the parameters relevant for `queryFn`, then you won't be able to distinguish _what_ is being fetched. An integer `42` might be a user ID, or a registration ID, or a schedule event ID, or whatever.

In your case, it's not crucial because you're using an isolated `CALCULATE_DUES_QUERY_CLIENT` which doesn't share its memory at all. But if you use an application-wide shared client, then you need the queryKey to be `['user', userId]` or `['registration', registrationId]` instead of just `[userId]`.

lib/dues_calculator.rb
```diff
@@ -1,15 +1,9 @@
...
-  def self.dues_for_n_competitors(country_iso2, base_entry_fee_lowest_denomination, currency_code, n)
+  def self.dues_per_competitor(country_iso2, base_entry_fee_lowest_denomination, currency_code)
     dues_per_competitor_in_usd_money = dues_per_competitor_in_usd(country_iso2, base_entry_fee_lowest_denomination, currency_code)
-    if dues_per_competitor_in_usd_money.present?
-      (dues_per_competitor_in_usd_money * n).exchange_to(currency_code)
-    else
-      nil
-    end
-  rescue CurrencyUnavailable
-    nil
```
gregorbg, line 12: I think this exception will still happen in local development when using the `EuCentralBank` exchange rates, so you should not delete the `rescue` part.

gregorbg (changes requested):

app/webpacker/components/CompetitionForm/FormSections/DuesEstimate.jsx
danieljames-dj: Okay done, added.

lib/dues_calculator.rb
danieljames-dj: Okay added back. This method is used only for the competition edit page, and I was thinking to show error there in case the currency is not available. Showing 0 dues is also fine, so I've added it back.

gregorbg (approved):

PR #10772, opened 2025-02-04
Auto-close registrations after enough paid registrations

dunkOnIT:


app/models/competition.rb
```diff
@@ -2883,4 +2899,9 @@ def self.form_data_json_schema
       },
     }
   end
+
+  def auto_close
```
gregorbg, line 2903: This only attempts to auto-close, so the name should be `attempt_auto_close`. With that in mind, you should return a boolean to indicate whether it auto-closed or not. That could be as simple as `threshold_reached && update!(...)` because the latter `update` call also returns a boolean in Rails if the update was successful.
Lastly, it does modify the competition in-place so the method name should end with a `!` per Ruby conventions.

app/models/competition.rb
```diff
@@ -925,6 +936,8 @@ def receive_registration_emails=(r)
     update_column :external_website, nil
   end
 
+  after_save :auto_close, if: -> { saved_change_to_auto_close_threshold? }
```
gregorbg, line 939: Super-duper unhappy about `after_save`s in our huge competition god model. Is this absolutely necessary for the feature to work?

As I understand it, checking in the `after_create` of the competition payments should be enough.

gregorbg: Also, if we decide to keep it: `if: :saved_change_to_auto_close_threshold?` should suffice.

db/migrate/20250204104514_add_auto_close_to_competitions_table.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddAutoCloseToCompetitionsTable < ActiveRecord::Migration[7.2]
+  def change
+    add_column :Competitions, :auto_close_threshold, :integer, default: 0, null: false
```
gregorbg, line 5: Why is `0` a good choice for a default value here, instead of just letting it be `null`? The latter seems like the perfect choice to express "I don't want to use any auto-close".

gregorbg (changes requested):

app/models/competition.rb
dunkOnIT: The intention here is to prevent the organizer changing the threshold to/below the number of paid registrations - but I realize that this should just be a model validation instead.

dunkOnIT: Validation added

app/models/competition.rb
dunkOnIT: Changes made

db/migrate/20250204104514_add_auto_close_to_competitions_table.rb
dunkOnIT: We don't currently have a way of setting values to `NULL` in `InputNumber` in the competition form, so once you enter a numeric value there is no way to disable that field. I've started using 0 as a default to get around this, but I strongly agree `NULL` would be more appropriate - I'm happy to take a detour to make this change in the FormInputs element now that I'm a little more React literate

gregorbg: Yay! But you gotta pay attention that some parts of the form _want_ to avoid `NULL` and instead explicitly set stuff to 0. So you cannot just straight-out interpret 0 as NULL in each case.

My suggestion here is to pass in an additional prop, happy to hop on a call and whip this up together because the wrapped form components are a bit tricky

dunkOnIT:
Blocked pending a change to allow setting competition form fields to `NULL`

app/webpacker/components/wca/FormBuilder/input/FormInputs.js
```diff
@@ -196,7 +196,7 @@ export const InputTextArea = wrapInput((props) => (
 
 export const InputNumber = wrapInput((props) => {
   const onChangeNumber = useCallback((e, { value: newValue }) => {
-    const convertedNumber = Number(newValue);
+    const convertedNumber = (props.nullable && newValue === '') ? '' : Number(newValue);
```
gregorbg, line 199: This is essentially correct, but very dense. Can you split the logic into two or even three lines perhaps?

app/webpacker/components/wca/FormBuilder/input/FormInputs.js
```diff
@@ -206,14 +206,14 @@ export const InputNumber = wrapInput((props) => {
       name={props.htmlName}
       type="number"
       label={props.attachedLabel}
-      value={props.value}
+      value={props.value ?? ''}
```
gregorbg, line 209: (Didn't run the code locally, so may be painfully obvious) Why is this `??` fallback necessary?

gregorbg (changes requested):

app/webpacker/components/wca/FormBuilder/input/FormInputs.js
dunkOnIT: Oop it isn't - just a holdover from when I wasn't doing things properly on 199 - removed

app/webpacker/components/wca/FormBuilder/input/FormInputs.js
dunkOnIT: Updated!

spec/models/competition_spec.rb
```diff
@@ -1584,4 +1584,96 @@ def change_and_check_activities(new_start_date, new_end_date)
...
+      comp.update_column(:auto_close_threshold, 5)
+      FactoryBot.create(:registration, :paid, competition: comp)
+
+      expect(comp.registration_past?).to eq(true)
+    end
+
+    it 'only auto-closes if the registrations are paid registrations' do
+      FactoryBot.create_list(:registration, 5, competition: auto_close_comp)
+      expect(auto_close_comp.registration_past?).to eq(false)
+    end
```
gregorbg, line 1632: I think these tests should differentiate between the hook triggering auto-close, and somebody explicitly trying to call the `auto_close!` method. Perhaps duplicate the tests, and in one batch just test whether the hook does its thing, but also have a second batch of similar/identical tests where you explicitly call the auto_close method and assume that it does/doesn't work (depending on the individual test case)

gregorbg (changes requested):

spec/models/competition_spec.rb
dunkOnIT: Great point - I've separated the tests into: 
- tests against the function itself in `competition_spec`
- a test asserting that the hook does actually call `attempt_auto_close!` in `registration_spec`

app/models/competition.rb
```diff
@@ -2893,4 +2909,11 @@ def self.form_data_json_schema
       },
     }
   end
+
+  def attempt_auto_close!
+    return false if auto_close_threshold.nil?
+    threshold_reached = registrations.with_payments.count >= auto_close_threshold && auto_close_threshold > 0
+    # update!(closing_full_registration: true, registration_close: Time.now) if threshold_reached
+    threshold_reached && update!(closing_full_registration: true, registration_close: Time.now)
```
gregorbg, line 2917: The `!` version of `update!` has a (potentially) dangerous implication: If, for some reasons, the validations for the `update` payload (the updated attributes) fail, it will throw a hard error.

And since this method is being called in the `after_create` hook of `registration_payment`, this would result in the creating transaction for that payment to be halted and rolled back. So if push came to shove, people could be unable to pay because of an oversight in our tests / edge cases that we haven't considered.

Fun fact: You have tests which try for both return values (`true` and `false`) but the `false` only ever happens when the first condition `threshold_reached` falsifies the entire `&&`. In reality, the signature is like "true or throw" more than an actually legitimate "true or false".

gregorbg: TL;DR:
```suggestion
    threshold_reached && update(closing_full_registration: true, registration_close: Time.now)
```

gregorbg (changes requested):

app/models/competition.rb
dunkOnIT: Updated

app/webpacker/components/CompetitionForm/FormSections/CompetitorLimit.js
```diff
@@ -18,6 +18,9 @@ export default function CompetitorLimit() {
         <InputNumber id="count" min={0} />
         <InputTextArea id="reason" />
       </ConditionalSection>
+      <ConditionalSection showIf={hasLimit}>
+        <InputNumber id="autoCloseThreshold" min={1} nullable />
+      </ConditionalSection>
```
gregorbg, line 23: I feel like these two can be merged into the same `ConditionalSection`

gregorbg: (both have `hasLimit` exactly set as their `showIf`)

gregorbg (changes requested):

app/webpacker/components/CompetitionForm/FormSections/CompetitorLimit.js
dunkOnIT: Updated

config/locales/en.yml
```diff
@@ -1720,6 +1721,7 @@ en:
           enabled: "A Competitor Limit is required. Only simultaneous FMC competitions with multiple locations can have no Competitor Limit."
           count: "The number of competitors allowed in this competition. For now this number is informational only, and does not yet prevent more people from registering. We are working on adding explicit support for this to the registration flow, but it requires some other work first."
           reason: "What is the reason for the limit on competitors? Please fill this out in English!"
+          auto_close_threshold: "Registration will automatically close when this number of paid registrations is received. Leave at 0 to disable, ignore if you aren't using integrated payments."
```
gregorbg, line 1724: Super nitpicky but with our `null` support now, is this still accurate?

app/models/competition.rb
```diff
@@ -321,6 +322,18 @@ def has_administrative_notes?
     end
   end
 
+  validate :auto_close_threshold_validations, if: -> { auto_close_threshold.present? }
```
gregorbg, line 325: `if: :auto_close_threshold?` should do:
image

gregorbg (changes requested):

app/models/competition.rb
```diff
@@ -321,6 +322,18 @@ def has_administrative_notes?
     end
   end
 
+  validate :auto_close_threshold_validations, if: -> { auto_close_threshold? }
```
gregorbg, line 325: No, I literally meant...
```suggestion
  validate :auto_close_threshold_validations, if: :auto_close_threshold?
```

gregorbg: For context, Ruby/Rails can use symbols to invoke methods, but these obviously only work for simple use-cases (i.e. if the check that you want to use is actually its own method). Opening lambdas with `-> { }` is really only worth it if you're introducing complex logic

gregorbg (changes requested):

app/models/competition.rb
dunkOnIT: Ah of course! That's much simpler

gregorbg (approved):

app/models/registration_payment.rb
```diff
@@ -29,7 +29,11 @@ def payment_status
...
   end
 
-  private def attempt_auto_close
-    registration.competition.attempt_auto_close!
+  private def auto_close_hook
+    if amount_lowest_denomination > 0
+      return registration.consider_auto_close
+    else
+      return false
+    end
```
gregorbg, line 37: FWIW you don't need the `if` here because I already added an `unless:` check in the `after_create` routine

dunkOnIT: oh nice!

app/models/registration.rb
```diff
@@ -398,11 +398,8 @@ def permit_user_cancellation?
   end
 
   def consider_auto_close
-    if outstanding_entry_fees == 0
-      return competition.attempt_auto_close!
-    else
-      return false
-    end
+    return competition.attempt_auto_close! if outstanding_entry_fees == 0
+    false
```
gregorbg, line 402:
```suggestion
    outstanding_entry_fees.zero? && competition.attempt_auto_close!
```

gregorbg: Whenever you have `if`s flying around, and you manually return booleans at the end, then there's a ~75% chance that your expression can be condensed into the `if`s.
Manually returning "naked" booleans is aaaalmost never necessary.

gregorbg (changes requested):

app/models/registration.rb
dunkOnIT: Thanks!

gregorbg (approved):

PR #10776, opened 2025-02-05
WCA Live Result Admin #1: Submit and Updating Results

FinnIckler:
Simple Submitting/Updating Results and double checking them. Does not include calculating advancing/rankings/records. Those will be follow up PRs (which I have already done, but I want to add tests).

Have not checked if I deleted everything related to those features, will do after dinner.


app/controllers/live_controller.rb
```diff
@@ -0,0 +1,103 @@
...
+    registration_id = params.require(:registration_id)
+
+    if LiveResult.exists?(round_id: round_id, registration_id: registration_id)
+      return render json: { status: "result already exist" }, status: :unprocessable_entity
+    end
+
+    AddLiveResultJob.perform_now({ results: results,
+                                   round_id: round_id,
+                                   registration_id: registration_id,
+                                   entered_by: current_user })
```
gregorbg, line 24: Regarding my other comment at https://github.com/thewca/worldcubeassociation.org/pull/10685#discussion_r1944347347: You would also need to update the call here and get rid of the hash `{` braces.

app/controllers/live_controller.rb
```diff
@@ -0,0 +1,103 @@
...
+  def update_result
+    results = params.require(:attempts)
+    round = Round.find(params.require(:round_id))
+    registration_id = params.require(:registration_id)
+
+    result = LiveResult.includes(:live_attempts).find_by(round: round, registration_id: registration_id)
+
+    unless result.present?
+      return render json: { status: "result does not exist" }, status: :unprocessable_entity
+    end
```
gregorbg, line 38: I think this can be `return render json: ... unless result.present?`. Surprised that RuboCop isn't complaining.

app/controllers/live_controller.rb
```diff
@@ -0,0 +1,103 @@
...
+        new_result = LiveAttempt.build(result: r, attempt_number: i)
+        different_result&.update(replaced_by_id: new_result.id)
+        new_result
+      end
+    end
+
+    # TODO: What is the best way to do this?
+    r = Result.build({ value1: results[0], value2: results[1], value3: results[2], value4: results[3] || 0, value5: results[4] || 0, event_id: round.event.id, round_type_id: round.round_type_id, format_id: round.format_id })
+
+    result.update(average: r.compute_correct_average, best: r.compute_correct_best, live_attempts: new_attempts, entered_at: Time.now.utc, entered_by: current_user)
```
gregorbg, line 57: Should we send a timestamp from the frontend when the query was fired?
If we're running this in a queue later, the "now" here in the backend could be significantly different from the "now" happening for the people in front of their screens.

app/controllers/live_controller.rb
```diff
@@ -0,0 +1,103 @@
...
+      same_result = previous_attempts.find_by(result: r, attempt_number: i)
+      if same_result.present?
+        same_result
+      else
+        different_result = previous_attempts.find_by(attempt_number: i)
+        new_result = LiveAttempt.build(result: r, attempt_number: i)
+        different_result&.update(replaced_by_id: new_result.id)
+        new_result
+      end
+    end
```
gregorbg, line 52: Note to self: I have a hunch there might be an even more efficient way to handle this.

gregorbg: Will play around later when I have more headspace.

app/controllers/live_controller.rb
```diff
@@ -0,0 +1,103 @@
...
+
+    unless round.round_can_be_opened?
+      flash[:danger] = "You can't open this round yet"
+      return redirect_to live_schedule_admin_path(competition_id: competition_id)
+    end
+
+    round.update(is_open: true)
+    flash[:success] = "Successfully opened round"
+    redirect_to live_schedule_admin_path(competition_id: competition_id)
+  end
```
gregorbg, line 88: Isn't this part of a separate PR?

app/models/round.rb
```diff
@@ -121,6 +121,42 @@ def name
...
+  def registrations
+    if number == 1
+      Registration.joins(:registration_competition_events)
+                  .where(
+                    competition_id: competition_event.competition_id,
+                    competing_status: 'accepted',
+                    registration_competition_events: { competition_event_id: competition_event_id }
+                  ).includes([:user])
+    else
+      previous_round = Round.joins(:competition_event).find_by(competition_event: { competition_id: competition_event.competition_id, event_id: event.id }, number: number - 1)
```
gregorbg, line 133:
```suggestion
      previous_round = Round.joins(:competition_event).find_by(competition_event: { competition_id: competition_event.competition_id, event_id: competition_event.event_id }, number: number - 1)
```

app/models/round.rb
```diff
@@ -121,6 +121,42 @@ def name
...
+      previous_round.live_results.where(advancing: true).includes(:registration).map(&:registration)
```
gregorbg, line 134: Ooof, that's a lot of database load. Can we somehow pluck registration IDs with a single `SELECT` and then load registrations based on that?

app/models/round.rb
```diff
@@ -121,6 +121,42 @@ def name
...
+    end
+  end
+
+  def registrations_with_wcif_id
```
gregorbg, line 138: Why do you need both `registrations` and also `registrations_with_wcif_id`?

gregorbg: Also, we should probably take this horrendous query as an incentive to think about a better way for coming up with registration IDs.

app/webpacker/components/Live/Admin/DoubleCheck/index.jsx
```diff
@@ -0,0 +1,151 @@
...
+  return Array(size).fill(0);
+}
+
+function Competitors({
+  competitionId, competitors, round, eventId,
+}) {
+  const event = events.byId[eventId];
+  const solveCount = event.recommendedFormat().expectedSolveCount;
+  const [currentIndex, setCurrentIndex] = useState(0);
+  const [registrationId, setRegistrationId] = useState(competitors[0].id);
```
gregorbg, line 42: What if `competitors` is empty?

app/webpacker/components/Live/Admin/DoubleCheck/index.jsx
```diff
@@ -0,0 +1,151 @@
...
+      return competitors.find((r) => r.id === results[currentIndex].registration_id);
+    }
+    return {};
+  }, [competitors, currentIndex, results]);
+
+  useEffect(() => {
+    if (results) {
+      setAttempts(results[currentIndex].attempts.map((a) => a.result));
+    }
+  }, [currentIndex, results]);
```
gregorbg, line 85: This looks like you can achieve the same thing without `useEffect`.

gregorbg: Isn't this essentially a `useMemo` depending on `results`?

app/webpacker/components/Live/Admin/DoubleCheck/index.jsx
```diff
@@ -0,0 +1,151 @@
...
+        { currentIndex !== 0 && <Button onClick={() => setCurrentIndex((oldIndex) => oldIndex - 1)}>{'<'}</Button>}
+      </Grid.Column>
+      <Grid.Column width={7}>
+        <Segment loading={isPendingUpdate}>
+          <Form>
+            <Form.Select
+              label="Competitor"
+              placeholder="Competitor"
+              value={currentCompetitor.id}
+              search={(inputs, value) => inputs.filter((d) => d.text.toLowerCase().includes(value.toLowerCase()) || parseInt(value, 10) === d.registrationId)}
```
gregorbg, line 103: I'm surprised that SemUI doesn't handle lowercasing natively. I am 97% sure I've added myself as a dummy organizer to comps before by using `gregor billing`. Does the `deburr` option change anything?

app/webpacker/components/Live/Admin/DoubleCheck/index.jsx
```diff
@@ -0,0 +1,151 @@
...
+              onChange={handleRegistrationIdChange}
+              options={competitors.map((p) => ({
+                key: p.id,
+                value: p.id,
+                registrationId: p.registration_id,
+                text: `${p.user.name} (${p.registration_id})`,
+              }))}
```
gregorbg, line 110: These should probably be `memo`ed somewhere.

app/webpacker/components/Live/Admin/DoubleCheck/index.jsx
```diff
@@ -0,0 +1,151 @@
...
+            />
+            {Array.from(zeroedArrayOfSize(solveCount).keys()).map((index) => (
```
gregorbg, line 112: You're building an array based on keys of a zeroed array? Since when do arrays have keys anyways? And what are you actually trying to do?

gregorbg: This looks like one of those typical features that lodash has a helper fpr

app/webpacker/components/Live/Admin/DoubleCheck/index.jsx
```diff
@@ -0,0 +1,151 @@
...
+        <Card fluid raised>
+          <Card.Header>
+            {currentIndex + 1}
+            {' '}
+            of
+            {' '}
+            {results.length}
+          </Card.Header>
+          <Card.Content>
+            <Header>{round.name}</Header>
```
gregorbg, line 139: You just completed a `Card.Header` above and the first thing you're doing in the body is opening a `Header`. Why not just move `round.name` into the card header?

app/webpacker/components/Live/Admin/Results/index.jsx
```diff
@@ -0,0 +1,200 @@
...
+        eventId={eventId}
+        competitors={competitors}
+      />
+    </WCAQueryClientProvider>
+  );
+}
+
+function zeroedArrayOfSize(size) {
+  return Array(size).fill(0);
+}
```
gregorbg, line 33: Repeated from above, this should probably go into a util function.

app/webpacker/components/Live/Admin/Results/index.jsx
```diff
@@ -0,0 +1,200 @@
...
+            </Header>
+
+            {error && <Message error content={error} />}
+            {success && <Message success content={success} />}
+
+            <Form.Select
+              label="Competitor"
+              placeholder="Competitor"
+              value={registrationId}
+              search={(inputs, value) => inputs.filter((d) => d.text.toLowerCase().includes(value.toLowerCase()) || parseInt(value, 10) === d.registrationId)}
```
gregorbg, line 157: Same remark about lowercasing as above.

app/webpacker/components/Live/Admin/Results/index.jsx
```diff
@@ -0,0 +1,200 @@
...
+              onChange={handleRegistrationIdChange}
+              options={competitors.toSorted((a, b) => a.id - b.id).map((p) => ({
+                key: p.id,
+                value: p.id,
+                registrationId: p.registration_id,
+                text: `${p.user.name} (${p.registration_id})`,
+              }))}
```
gregorbg, line 164: Same remark about pre-computing as above.

app/webpacker/components/Live/Admin/Results/index.jsx
```diff
@@ -0,0 +1,200 @@
...
+            />
+            {Array.from(zeroedArrayOfSize(solveCount).keys()).map((index) => (
```
gregorbg, line 166: Same "huh" as above

app/webpacker/components/Live/Admin/Results/index.jsx
```diff
@@ -0,0 +1,200 @@
...
+        roundId, registrationId, competitionId, attempts,
+      });
+    } else {
+      mutateSubmit({
+        roundId, registrationId, competitionId, attempts,
+      });
+    }
+  };
+
+  return (
```
gregorbg, line 141: Just from the fact that I wrote a lot of "same as above" comments below: How much code is this actually sharing with `DoubleCheck`?

app/webpacker/components/Live/components/ResultsTable.jsx
```diff
@@ -0,0 +1,172 @@
...
+
+  return competitorResults[0][sortBy];
+};
+
+export const rankingCellStyle = (result) => {
+  if (!result) {
+    return {};
+  }
+  if (result.advancing) {
+    return { backgroundColor: 'rgb(0, 230, 118)' };
```
gregorbg, line 23: The green tone should probably be stored somewhere, especially since you're reusing it with transparency again below

app/webpacker/components/Live/components/ResultsTable.jsx
```diff
@@ -0,0 +1,172 @@
...
+    case 'PR': {
+      styles.backgroundColor = 'rgb(66, 66, 66)';
+      break;
+    }
+    default: {
+      return {};
+    }
+  }
+  return styles;
+};
```
gregorbg, line 70: Not a fan of big custom styles at all. SemUI has a `Label` component, can that work?
For specific colors, I'll agree to some CSS styling but `display: block` and others should definitely be left to SemUI.

gregorbg (changes requested):

app/controllers/live_controller.rb
FinnIckler: updating is currently not done in a queue, maybe it should? But we aren't doing that for registrations for example (but there updating doesn't create anything new)

app/controllers/live_controller.rb
FinnIckler: oh yeah, that was accidental

app/models/round.rb
FinnIckler: yeah, I was thinking something like that too

app/models/round.rb
FinnIckler: registrations -> loading data for front office, registrations_with_wcif_id -> loading data for back office

app/webpacker/components/Live/Admin/DoubleCheck/index.jsx
FinnIckler: this creates an array with the values [0,1,2,3,4,5]

app/webpacker/components/Live/Admin/Results/index.jsx
FinnIckler: this definitely can be combined

app/webpacker/components/Live/components/ResultsTable.jsx
FinnIckler: yes, this was a fairly lazy way of recreating the same thing as wca live

app/models/round.rb
FinnIckler: you can even do `Round.joins(:competition_event).find_by(competition_event: competition_event)`

app/webpacker/components/Live/Admin/DoubleCheck/index.jsx
FinnIckler: I added a fallback in the parent component

app/webpacker/components/Live/Admin/DoubleCheck/index.jsx
FinnIckler: I moved the whole component out

app/webpacker/components/Live/Admin/DoubleCheck/index.jsx
gregorbg: But `currentIndex` is also just a React state. If you put that into the memo dependencies, it will automatically update. I therefore insist this `useEffect` is unnecessary.

FinnIckler: I don't get it: 
```
  useEffect(() => {
    if (results) {
      setAttempts(results[currentIndex].attempts.map((a) => a.result));
    }
  }, [currentIndex, results]);
```
and 
```
 useMemo(() => {
    if (results) {
      setAttempts(results[currentIndex].attempts.map((a) => a.result));
    }
  }, [currentIndex, results]);
```
are functionally identical. and I don't think a useMemo is supposed to trigger a side effect like that

FinnIckler: and you can't just do 
```
const attempts = useMemo(() => {
    if (results) {
      results[currentIndex].attempts.map((a) => a.result);
    }
  }, [currentIndex, results]);
```
Because attempts needs to be a state so it can be manipulated from the form

app/webpacker/components/Live/Admin/DoubleCheck/index.jsx
FinnIckler: done, but it does need a break tag currently. This can be fixed when the "x of y" string will be in i18n

app/webpacker/components/Live/Admin/DoubleCheck/index.jsx
```diff
@@ -0,0 +1,151 @@
...
+    });
+  };
+
+  const handleAttemptChange = (index, value) => {
+    const newAttempts = [...attempts];
+    newAttempts[index] = value;
+    setAttempts(newAttempts);
+  };
+
+  const handleRegistrationIdChange = useCallback((_, { value }) => {
```
kr-matthews, line 72: Does this need to be a callback? I think a regular const anonymous function would be fine.

app/webpacker/components/Live/Admin/DoubleCheck/index.jsx
```diff
@@ -0,0 +1,151 @@
...
+import getRoundResults from '../../api/getRoundResults';
+import Loading from '../../../Requests/Loading';
+
+export default function Wrapper({
+  competitionId, competitors, round,
+}) {
+  return (
+    <WCAQueryClientProvider>
+      <Competitors
+        eventId={round.event.id}
```
kr-matthews, line 31: This prop could be removed since it's derived from another prop (`round`).

app/webpacker/components/Live/Admin/DoubleCheck/index.jsx
kr-matthews, line 46: Can the registration id be derived from the current index and the results?

app/webpacker/components/Live/Admin/DoubleCheck/index.jsx
kr-matthews: Can this be something like `attempts.mapIndexed((attempt, index) =>`?

app/webpacker/components/Live/Admin/DoubleCheck/index.jsx
kr-matthews: Is the query ever re-fetched? i.e. Does `results` ever change? If not, then you indeed do need a side effect to set `attempts` when the data first loads, but you could explicitly update `attempts` when the "<" or ">" button is clicked instead of it being a side effect of the index changing. That would minimize the use of effects (even though it's the same number of `useEffects`, that effect would only be happening once).

app/webpacker/components/Live/Admin/DoubleCheck/index.jsx
FinnIckler: it could, but we want to be able to change the `registrationId` independently for cases like "correct result, but wrong id input"

app/webpacker/components/Live/Admin/DoubleCheck/index.jsx
FinnIckler: I use `_.times` for this now. You can't use `attempts.mapIndexed((attempt, index)` because the attempts might be incomplete

config/routes.rb
```diff
@@ -150,7 +150,16 @@
   get 'competitions/edit/registration-collisions-json' => 'competitions#registration_collisions_json', as: :registration_collisions_json
   get 'competitions/edit/series-eligible-competitions-json' => 'competitions#series_eligible_competitions_json', as: :series_eligible_competitions_json
 
-  get 'competitions/:competition_id/live/admin' => 'live#schedule_admin', as: :live_schedule_admin
+  if WcaLive.enabled?
+    get 'competitions/:competition_id/live/rounds/:round_id/admin' => 'live#admin', as: :live_admin_round_results
+    get 'competitions/:competition_id/live/rounds/:round_id/admin/check' => 'live#double_check', as: :live_admin_check_round_results
+    get 'competitions/:competition_id/live/admin' => 'live#schedule_admin', as: :live_schedule_admin
+    get 'competitions/:competition_id/live/admin' => 'live#schedule_admin', as: :live_schedule_admin
```
gregorbg, line 157: Whoopsie (double route)

gregorbg (approved):

PR #10777, opened 2025-02-05
WCA Live Result Admin #2: Schedule Admin View

FinnIckler:
Has the Admin schedule where you can open rounds. Will add some tests to this once the db pr is merged


app/models/round.rb
```diff
@@ -133,6 +133,25 @@ def advancement_condition_to_s(short: false)
...
+  def score_taking_done?
+    is_open && competitors_live_results_entered == total_registrations
+  end
+
+  def round_can_be_opened?
+    return false if is_open
+    return true if number == 1
+    previous_round = Round.joins(:competition_event).find_by(competition_event: { competition_id: competition_event.competition_id, event_id: event.id }, number: number - 1)
+    previous_round.score_taking_done?
+  end
```
gregorbg, line 149: Isn't this just `number == 1 || (!is_open && ...)`?

gregorbg (changes requested):
Not much to discuss here, but while we're at it we should probably find a better name for this whole "open" concept.

It worked while WCA Live was separate, because ever special term had to be interpreted with regards to live results entry. Now that we're moving the functionality back into the monolith, "open" suddenly is a bit too generic. Open for what? For registration? For complaints? For editing?

Suggestions:
- `scoretaking_open`
- `scoretaking_active`
- `live_results_open`
- `data_entry_ongoing`
- permutations of the prefixes and suffixes above

app/models/round.rb
FinnIckler: no. that would always return true for rounds with `number == 1`. But these are not openable if they are already opened. it actually used to be this
```
def round_can_be_opened?
    return false if is_open
    return !is_open if number == 1
    previous_round = Round.joins(:competition_event).find_by(competition_event: { competition_id: competition_event.competition_id, event_id: event.id }, number: number - 1)
    previous_round.score_taking_done?
  end
```
but that `!is_open `is redundant after the first check

FinnIckler:
I do still want to keep this PR for the Admin schedule page. But the logic can be changed to just check if the previous round has been done scoretaking which is like an implicit is_open

gregorbg (approved):

FinnIckler:
So the `has_many :registrations, through: :competition_event` is about 50x slower as my previous approach, but I am sure you can use some includes to fix it. What is annoying is that previously developers wouldn't need to know that they had to do that when using the method

FinnIckler:
The culprit seems to be that it's running this query for every round
```
Registration Count (1513.6ms)  SELECT COUNT(*) FROM `registrations` INNER JOIN `registration_competition_events` ON `registrations`.`id` = `registration_competition_events`.`registration_id` INNER JOIN `competition_events` ON `registration_competition_events`.`competition_event_id` = `competition_events`.`id` WHERE `competition_events`.`id` = 129185 AND `registrations`.`competing_status` = 'accepted'
```
instead of the much (1000x) faster
```
`Registration Count (0.9ms)  SELECT COUNT(*) FROM `registrations` INNER JOIN `registration_competition_events` ON `registration_competition_events`.`registration_id` = `registrations`.`id` WHERE `registrations`.`competition_id` = 'AachenOpen2025' AND `registrations`.`competing_status` = 'accepted' AND `registration_competition_events`.`competition_event_id` = 129185`
```

gregorbg:
That sounds like a missing index

gregorbg:
And indeed, it was!

```sql
Registration Count (0.8ms)  SELECT COUNT(*) FROM `registrations` INNER JOIN `registration_competition_events` ON `registrations`.`id` = `registration_competition_events`.`registration_id` INNER JOIN `competition_events` ON `registration_competition_events`.`competition_event_id` = `competition_events`.`id` WHERE `competition_events`.`id` = 129185 AND `registrations`.`competing_status` = 'accepted'
```

PR #10778, opened 2025-02-06
Use tanstack table for Competitors

kr-matthews:
See slack discussion about my concerns.

### DIFF
--- a/app/webpacker/components/RegistrationsV2/Registrations/Competitors.jsx
+++ b/app/webpacker/components/RegistrationsV2/Registrations/Competitors.jsx
@@ -1,26 +1,95 @@
 import { useQuery } from '@tanstack/react-query';
-import React, {
-  useMemo,
-  useReducer,
-} from 'react';
+import React from 'react';
 import {
   Flag, Segment, Table,
 } from 'semantic-ui-react';
 import _ from 'lodash';
 import {
+  flexRender, getCoreRowModel, getSortedRowModel, useReactTable
+} from '@tanstack/react-table';
+import {
   getConfirmedRegistrations,
 } from '../api/registration/get/get_registrations';
-import createSortReducer from '../reducers/sortReducer';
 import EventIcon from '../../wca/EventIcon';
 import { personUrl } from '../../../lib/requests/routes.js.erb';
 import I18n from '../../../lib/i18n';
 import { countries } from '../../../lib/wca-data.js.erb';
-import { getPeopleCounts, getTotals, getUserPositionInfo } from './utils';
+import { expandSortTerm, getPeopleCounts, getTotals, getUserPositionInfo } from './utils';
 import PreTableInfo from './PreTableInfo';
 import Errored from '../../Requests/Errored';
 import Loading from '../../Requests/Loading';
 
-const sortReducer = createSortReducer(['name', 'country', 'total']);
+const getColumns = (eventIds) => [
+  {
+    accessorKey: 'user.name',
+    header: I18n.t('activerecord.attributes.registration.name'),
+    cell: (info) => info.getValue() ? (
+      <a href={personUrl(info.row.original.user.wca_id)}>
+        {info.getValue()}
+      </a>
+    ) : (
+      info.getValue()
+    ),
+    footer: (info) => {
+      const registrationCount = info.table.getCoreRowModel().rows.length;
+      return (
+        `${
+          registrationCount
+        } ${
+          I18n.t(
+            'registrations.registration_info_people.person',
+            { count: info.table.getCoreRowModel().rows.length },
+          )
+        }`
+      )
+    },
+  },
+  {
+    id: 'country',
+    header: I18n.t('activerecord.attributes.user.country_iso2'),
+    accessorFn: (row) => countries.byIso2[row.user.country.iso2].name,
+    cell: (info) => (
+      <>
+        <Flag
+          className={info.row.original.user.country.iso2.toLowerCase()}
+        />
+        {info.getValue()}
+      </>
+    ),
+    footer: (info) => {
+      const { countryCount } = getTotals(
+        info.table.getCoreRowModel().rows.map((row) => row.original),
+        eventIds,
+      );
+      return `${I18n.t('registrations.list.country_plural', { count: countryCount })}`
+    }
+  },
+  ...eventIds.map((id) => ({
+    id,
+    header: <EventIcon id={id} size="1em" className="selected" />,
+    enableSorting: false,
+    accessorFn: (row) => row.competing.event_ids.includes(id),
+    cell: (info) => (
+      info.getValue() && <EventIcon id={id} size="1em" hoverable={false} />
+    ),
+    footer: (info) => (
+      info.table.getCoreRowModel().rows.filter((row) => row.getValue(id)).length
+    ),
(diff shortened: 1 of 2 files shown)


app/webpacker/components/RegistrationsV2/Registrations/Competitors.jsx
```diff
@@ -1,26 +1,95 @@
...
+  {
+    accessorKey: 'user.name',
+    header: I18n.t('activerecord.attributes.registration.name'),
+    cell: (info) => info.getValue() ? (
+      <a href={personUrl(info.row.original.user.wca_id)}>
+        {info.getValue()}
+      </a>
+    ) : (
+      info.getValue()
+    ),
```
gregorbg, line 32: I don't get this with regards to the ternary condition. If `getValue` is truthy, you return an `<a>` link, but if it's falsey you just return... that falsey value explicitly? Why not just `info.getValue() && (<a>)`?

app/webpacker/components/RegistrationsV2/Registrations/Competitors.jsx
```diff
@@ -1,26 +1,95 @@
...
+    footer: (info) => {
+      const registrationCount = info.table.getCoreRowModel().rows.length;
+      return (
+        `${
+          registrationCount
+        } ${
+          I18n.t(
+            'registrations.registration_info_people.person',
+            { count: info.table.getCoreRowModel().rows.length },
```
gregorbg, line 41: You can just pass `registrationCount` here.

app/webpacker/components/RegistrationsV2/Registrations/Competitors.jsx
```diff
@@ -1,26 +1,95 @@
...
+          )
+        }`
+      )
+    },
+  },
+  {
+    id: 'country',
+    header: I18n.t('activerecord.attributes.user.country_iso2'),
+    accessorFn: (row) => countries.byIso2[row.user.country.iso2].name,
```
gregorbg, line 50: Clever, I think I should copy that in my Records PR :D

app/webpacker/components/RegistrationsV2/Registrations/Competitors.jsx
```diff
@@ -1,26 +1,95 @@
...
+          className={info.row.original.user.country.iso2.toLowerCase()}
+        />
+        {info.getValue()}
+      </>
+    ),
+    footer: (info) => {
+      const { countryCount } = getTotals(
+        info.table.getCoreRowModel().rows.map((row) => row.original),
+        eventIds,
+      );
```
gregorbg, line 63: Hm, this is indeed quite unintuitive.

app/webpacker/components/RegistrationsV2/Registrations/Competitors.jsx
```diff
@@ -1,26 +1,95 @@
...
+      return `${I18n.t('registrations.list.country_plural', { count: countryCount })}`
+    }
+  },
+  ...eventIds.map((id) => ({
+    id,
```
gregorbg, line 68: Maybe something like `event-${id}`?

app/webpacker/components/RegistrationsV2/Registrations/Competitors.jsx
```diff
@@ -101,164 +150,68 @@ export default function Competitors({
...
-                <Table.Cell
-                  textAlign="center"
-                  key={`registration-table-row-${registration.user.id}-${id}`}
+            </Table.Row>
+          )}
+        </Table.Body>
+        <Table.Footer>
+          {table.getFooterGroups().map(footerGroup => (
+            <Table.Row key={footerGroup.id}>
+              {footerGroup.headers.map(header => (
```
gregorbg, line 203: Do they really call this `footerGroup.headers`?!

app/webpacker/components/RegistrationsV2/Registrations/Competitors.jsx
```diff
@@ -101,164 +150,68 @@ export default function Competitors({
...
+                  textAlign={header.column.columnDef.meta?.style?.textAlign}
+                >
+                  {flexRender(header.column.columnDef.header, header.getContext())}
+                </Table.HeaderCell>
+              ))}
+            </Table.Row>
+          ))}
+        </Table.Header>
+        <Table.Body>
+          {table.getSortedRowModel().rows.length > 0 ? (
```
gregorbg, line 174: Can you return early if no columns are found?

gregorbg (commented):
Okay, I see your point much more clearly now. The configuration is indeed cumbersome, and especially in terms of footer counts the setup is _much_ more complicated than I anticipated.

I'll leave it up to you whether you want to use it and merge this PR or if you prefer to discard it.

app/webpacker/components/RegistrationsV2/Registrations/Competitors.jsx
kr-matthews: Pretty sure I just put the wrong condition, it should be checking that the wca id exists.

app/webpacker/components/RegistrationsV2/Registrations/Competitors.jsx
kr-matthews: Two things here:
1. `getTotals` is a helper function which is shared with the psych sheet which is using the original registrations data. So to keep it as a shared helper function I had to pass in the original rows. That helper could be modified though, to take in the tanstack table data in some other format.
2. I made the cell value just the country name, to get automatic sorting capabilities from tanstack table. I could instead make the cell value be the country object. That would make getting the flag classname simpler, and possibly make the footer calculation easier. But, then a custom sort function would need to be passed in.

Even with those changes, the footer would still not be the most intuitive thing. And this just highlights that there are so many ways to approach tables that it will be hard for us to be consistent across the codebase - especially with non-WST contributors.

The existing (non-tanstack) code is quite easy to understand, and you can explicitly see the data being sorted/etc. If you want to update the footer, it's easy you just go to the footer component and that component has the data immediately available.

As I mentioned elsewhere, I really like that the column definitions centralize the header/body/footer of a column in one spot. But that's about it. The automatic sorting/filtering _can_ be nice - but not if the cell values are 'complicated, such as names with links or countries with flags, etc.

So I'm undecided whether I want to finish off this PR, or cancel it. :shrug:

app/webpacker/components/RegistrationsV2/Registrations/Competitors.jsx
kr-matthews: Do you mean rows? Returning early would omit the header/footer. That could work, I guess. I prefer having the header/footer, and I think many of our other tables do that, including as the psych sheet.

app/webpacker/components/RegistrationsV2/Registrations/Competitors.jsx
kr-matthews: Unfortunately. https://tanstack.com/table/v8/docs/framework/react/examples/basic

gregorbg:
I would still be very interested in a conversation with TanStack about some of our core requirements that arose from this PR, and whether they plan to add any of them as features in the newer releases. But in the meantime, I don't think this needs to be merged just so that we can use their library at all cost.

PR #10781, opened 2025-02-07
Order items moved to waiting list by payment timestamp

dunkOnIT:
Initially, users were added to the waiting list in the order they were selected (ticked) by the organizer. Now, when items users are added in Waiting List, they are sorted according to the following rules: 
- paid registrations before unpaid registrations
- paid registrations sorted by payment timestamp
- unpaid registrations sorted by when they were selected (this behaviour still makes sense in cases where payment is handled off-site, and registration timestamp has no bearing on who should be approved in what order) 

Finn and I also discussed 'previewing' the new state of the waiting list - but that's hard to do, and there isn't user support for it, so I think this change is sufficient for now.

User feedback informing these changes: 
image

image

### DIFF
--- a/app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
+++ b/app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
@@ -67,32 +67,6 @@ export default function RegistrationActions({
     .map((userId) => userEmailMap[userId])
     .join(',');
 
-  const moveToWaitingList = (attendees) => {
-    const attendeesWithRegistrations = attendees
-      .map((user_id) => {
-        const registration = registrations.find((reg) => reg.user_id === user_id);
-        return {
-          user_id,
-          registration,
-          paymentUpdatedAt: registration?.payment?.updated_at || null,
-        };
-      });
-
-    // Sort paid registrations by their payment timestamp
-    const paid = attendeesWithRegistrations.filter(({ paymentUpdatedAt }) => paymentUpdatedAt);
-    // Leave unpaid registrations in the order of selection
-    const unpaid = attendeesWithRegistrations.filter(({ paymentUpdatedAt }) => !paymentUpdatedAt);
-
-    paid.sort((a, b) => {
-      if (!a.paymentUpdatedAt) return 1; // Push entries without payment to the end
-      if (!b.paymentUpdatedAt) return -1;
-      return new Date(a.paymentUpdatedAt) - new Date(b.paymentUpdatedAt);
-    });
-
-    const combined = paid.concat(unpaid)
-    changeStatus(combined.map(({user_id}) =>  user_id ), "waiting_list");
-  }
-
   const changeStatus = (attendees, status) => {
     updateRegistrationMutation(
       {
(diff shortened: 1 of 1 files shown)


app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
```diff
@@ -85,7 +85,30 @@ export default function RegistrationActions({
...
     );
-  };
+  }, [competitionInfo.id]);
+
+  const moveToWaitingList = useCallback((attendees) => {
+    const registrationsByUserId = _.groupBy(registrations, 'user_id');
+
+    const attendeesWithRegistrations = attendees
+      .map((userId) => {
+        const registration = registrationsByUserId[userId]?.[0];
```
FinnIckler, line 95: you shouldn't need the ?. acess here

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
```diff
@@ -85,7 +85,30 @@ export default function RegistrationActions({
         },
       },
     );
-  };
+  }, [competitionInfo.id]);
+
+  const moveToWaitingList = useCallback((attendees) => {
+    const registrationsByUserId = _.groupBy(registrations, 'user_id');
```
FinnIckler, line 91: actually, now that you have this. You should be able to skip computing `attendeesWithRegistrations` at all and just use `registrationsByUserId` like 
    const [paid, unpaid] = _.partition(attendees, ({ user_id }) => registrationsByUserId[user_id].registration?.payment?.updated_at);

dunkOnIT: Done!

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
```diff
@@ -67,7 +67,7 @@ export default function RegistrationActions({
     .map((userId) => userEmailMap[userId])
     .join(',');
 
-  const changeStatus = (attendees, status) => {
+  const changeStatus = useCallback((attendees, status) => {
```
kr-matthews, line 70: I don't think this or the new function needs to be memoized.

kr-matthews (commented):
Why is this being done in the front-end?

dunkOnIT:
> Why is this being done in the front-end?

Finn and I talked through this, and concluded that we don't want the backend to alter the order that change requests are given to it, given that this will ultimately be an public-facing API.

There is definitely an argument to be made for still achieving this in the backend through a parameter or a separate route though. I can ask Gregor his thoughts on Monday.

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
dunkOnIT: Changed

gregorbg:
In terms of API design, I strongly agree with the idea that waiting list items should be inserted exactly in the order that you pass them to the endpoint. Hence having the frontend do the sorting makes sense from that perspective.

However, I also hear the criticism that the frontend is not the right place for data sorting. I really like the idea of (optionally!) specifying sort parameters (or even a simple `auto_sort=1`) as a simple `?` GET param when submitting the request.

How plausible / feasible is that? @dunkOnIT

dunkOnIT:
Very feasible, working on the changes now

dunkOnIT:
As discussed in the call today, I've created #10809 to note that we want to ultimately have this handled in the backend, but the effort/reward of implementing it doesn't make sense now - so this PR is ready for review/approval

gregorbg (approved):

PR #10783, opened 2025-02-07
Fix dates in competitions admin view

maxidragon:
This fixes the bug with results submitted on the competition day and showed as submitted 1 day after.

### DIFF
--- a/app/webpacker/lib/utils/competition-table.js
+++ b/app/webpacker/lib/utils/competition-table.js
@@ -94,7 +94,7 @@ export function numberOfDaysAfter(competition, refDate) {
 
   const numberOfDays = parsedStartDate.diff(parsedRefDate, 'days').days;
 
-  return Math.ceil(Math.abs(numberOfDays));
+  return Math.floor(Math.abs(numberOfDays));
 }
 
 export function timeDifferenceAfter(competition, refDate) {


app/webpacker/lib/utils/competition-table.js
```diff
@@ -94,7 +94,7 @@ export function numberOfDaysAfter(competition, refDate) {
 
   const numberOfDays = parsedStartDate.diff(parsedRefDate, 'days').days;
 
-  return Math.ceil(Math.abs(numberOfDays));
+  return Math.floor(Math.abs(numberOfDays));
```
gregorbg, line 97: Could you add a comment explaing why `floor` makes sense here?
Otherwise in the future, somebody might find an unrelated bug and just change this back to `ceil` accidentally.

gregorbg (changes requested):

gregorbg (changes requested):

gregorbg (approved):

maxidragon:
@gregorbg Sometimes it still displays the wrong number of days. In the legacy view, it showed 7 and 0 days for all these comps.

image

Do you have any ideas on how to fix it? The only thing I can think of is that we might be calculating these dates in local timezones instead of UTC.

gregorbg:
Hm. Look at the raw data first! What is the exact timestamp that is being transmitted by the backend? That information will certainly give you a lead about the proper fix.

PR #10786, opened 2025-02-08
Delegate report changes on my competitions page

pranav-027:
- Delegate report section in past competitions
- Warning logo only for past comps

### DIFF
--- a/app/webpacker/components/MyCompetitions/PastCompetitionTable.jsx
+++ b/app/webpacker/components/MyCompetitions/PastCompetitionTable.jsx
@@ -65,7 +65,7 @@ export default function PastCompetitionsTable({
               competitionId={competition.id}
               permissions={permissions}
               isReportPosted={competition['report_posted?']}
-              isPastCompetition={true}
+              isPastCompetition
             />
           </Table.Row>
         ))}
--- a/app/webpacker/components/MyCompetitions/PastCompetitionTable.jsx
+++ b/app/webpacker/components/MyCompetitions/PastCompetitionTable.jsx
@@ -13,6 +13,7 @@ export default function PastCompetitionsTable({
   permissions,
   fallbackMessage = null,
 }) {
+  const canViewDelegateReport = permissions.can_view_delegate_report.scope === '*' || competitions.some((c) => permissions.can_view_delegate_report.scope.includes(c.id));
   if (competitions.length === 0 && fallbackMessage) {
     return (
       <Message info>
@@ -37,6 +38,13 @@ export default function PastCompetitionsTable({
           <Table.HeaderCell />
           <Table.HeaderCell />
           <Table.HeaderCell />
+          {canViewDelegateReport && (
+            <>
+              <Table.HeaderCell />
+              <Table.HeaderCell />
+              <Table.HeaderCell />
+            </>
+          )}
         </Table.Row>
       </Table.Header>
 
@@ -61,7 +69,7 @@ export default function PastCompetitionsTable({
                 />
               )}
             </Table.Cell>
-            <ReportTableCell competitionId={competition.id} permissions={permissions} isReportPosted={competition['report_posted?']} />
+            <ReportTableCell competitionId={competition.id} permissions={permissions} isReportPosted={competition['report_posted?']} canViewDelegateReport={canViewDelegateReport} />
           </Table.Row>
         ))}
       </Table.Body>
--- a/app/webpacker/components/MyCompetitions/PastCompetitionTable.jsx
+++ b/app/webpacker/components/MyCompetitions/PastCompetitionTable.jsx
@@ -38,13 +38,6 @@ export default function PastCompetitionsTable({
           <Table.HeaderCell />
           <Table.HeaderCell />
           <Table.HeaderCell />
-          {canViewDelegateReport && (
-            <>
-              <Table.HeaderCell />
-              <Table.HeaderCell />
-              <Table.HeaderCell />
-            </>
-          )}
         </Table.Row>
       </Table.Header>
 
--- a/app/webpacker/components/MyCompetitions/PastCompetitionTable.jsx
+++ b/app/webpacker/components/MyCompetitions/PastCompetitionTable.jsx
@@ -13,7 +13,6 @@ export default function PastCompetitionsTable({
   permissions,
   fallbackMessage = null,
 }) {
-  const canViewDelegateReport = permissions.can_view_delegate_report.scope === '*' || competitions.some((c) => permissions.can_view_delegate_report.scope.includes(c.id));
   if (competitions.length === 0 && fallbackMessage) {
     return (
       <Message info>
@@ -62,7 +61,12 @@ export default function PastCompetitionsTable({
                 />
               )}
             </Table.Cell>
-            <ReportTableCell competitionId={competition.id} permissions={permissions} isReportPosted={competition['report_posted?']} canViewDelegateReport={canViewDelegateReport} />
+            <ReportTableCell
+              competitionId={competition.id}
+              permissions={permissions}
+              isReportPosted={competition['report_posted?']}
+              isPastCompetition={true}
+            />
           </Table.Row>
         ))}
       </Table.Body>
--- a/app/webpacker/components/MyCompetitions/TableCells.jsx
+++ b/app/webpacker/components/MyCompetitions/TableCells.jsx
@@ -50,8 +50,7 @@ export function ReportTableCell({
               </a>
             )}
           />
-          { isPastCompetition && !isReportPosted &&
-            (permissions.can_edit_delegate_report.scope === '*' || permissions.can_edit_delegate_report.scope.includes(competitionId))
+          { isPastCompetition && !isReportPosted && (permissions.can_edit_delegate_report.scope === '*' || permissions.can_edit_delegate_report.scope.includes(competitionId))
             && (
             <Popup
               content={I18n.t('competitions.my_competitions_table.edit_report')}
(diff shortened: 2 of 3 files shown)


app/webpacker/components/MyCompetitions/PastCompetitionTable.jsx
```diff
@@ -37,6 +38,13 @@ export default function PastCompetitionsTable({
           <Table.HeaderCell />
           <Table.HeaderCell />
           <Table.HeaderCell />
+          {canViewDelegateReport && (
+            <>
+              <Table.HeaderCell />
+              <Table.HeaderCell />
+              <Table.HeaderCell />
+            </>
+          )}
```
gregorbg, line 47: Why are you adding three header cells here, and where are the corresponding three body cells?

app/webpacker/components/MyCompetitions/TableCells.jsx
```diff
@@ -53,7 +53,7 @@ export function ReportTableCell({
           />
           )}
 
-          {(permissions.can_edit_delegate_report.scope === '*' || permissions.can_edit_delegate_report.scope.includes(competitionId))
+          {!isReportPosted && (permissions.can_edit_delegate_report.scope === '*' || permissions.can_edit_delegate_report.scope.includes(competitionId))
```
gregorbg, line 56: In what way is this boolean check related to the issue you're trying to solve with this PR?

gregorbg (changes requested):

app/webpacker/components/MyCompetitions/PastCompetitionTable.jsx
pranav-027: I had copy-pasted code form upcoming competitions at the start. I have removed it now

app/webpacker/components/MyCompetitions/TableCells.jsx
pranav-027: The edit report button is still shown even if the reports are posted and clicking on it will lead to the following error:
image

gregorbg: Okay, makes sense, thanks!

app/webpacker/components/MyCompetitions/PastCompetitionTable.jsx
```diff
@@ -61,7 +62,7 @@ export default function PastCompetitionsTable({
                 />
               )}
             </Table.Cell>
-            <ReportTableCell competitionId={competition.id} permissions={permissions} isReportPosted={competition['report_posted?']} />
+            <ReportTableCell competitionId={competition.id} permissions={permissions} isReportPosted={competition['report_posted?']} canViewDelegateReport={canViewDelegateReport} />
```
gregorbg, line 65: Hmm... Looking a little deeper at the code of `ReportTableCell`, it seems that component always gets passed the current `permissions` as props, but never really uses them unless the `canView` boolean allows it.

Ironically, that same boolean is also based directly on some calculation in the `permissions` (in both cases where this component is even used!)
Would you mind refactoring `ReportTableCell` to compute this `canViewReport` value internally?

pranav-027: Hello @gregorbg, I looked into it and tried implementing it, but `canViewDelegateReport` is computed based on both `permissions` and `competitions`. Do you recommend passing `competitions` as Props in `ReportTableCell`?

gregorbg: `ReportTableCell` already gets passed a specific competition ID. You can just check for that one competition in the `scope` of the permissions instead of iterating over all competitions.

That way, you don't have to worry about passing down `competitions` as prop.

app/webpacker/components/MyCompetitions/TableCells.jsx
```diff
@@ -51,9 +50,7 @@ export function ReportTableCell({
               </a>
             )}
           />
-          )}
-
-          {(permissions.can_edit_delegate_report.scope === '*' || permissions.can_edit_delegate_report.scope.includes(competitionId))
+          { !isReportPosted && (permissions.can_edit_delegate_report.scope === '*' || permissions.can_edit_delegate_report.scope.includes(competitionId))
```
gregorbg, line 53: You can reuse `canViewDelegateReport` here. And since this whole big return statement is wrapped into one gigantic `if`, you can just drop this part of the `&&` condition altogether because it will always be true at this point.

app/webpacker/components/MyCompetitions/UpcomingCompetitionTable.jsx
```diff
@@ -82,14 +80,8 @@ export default function UpcomingCompetitionTable({
...
-            {canViewDelegateReport && (
-              <>
-                <Table.HeaderCell />
-                <Table.HeaderCell />
-                <Table.HeaderCell />
-              </>
-            )}
-
+            <Table.HeaderCell />
+            <Table.HeaderCell />
```
gregorbg, line 84: You're removing the `canViewDelegateReport` branch which rendered three more headers, but you only add two more new headers. Where did the third one go?

gregorbg: Also, I think it does make sense to keep the general `canViewDelegateReport` check here. In the original definition from above, it uses `competitions.some` to find out whether there is _any_ Delegate Report that you can view / access.
- If there is, you have to add the header cells as placeholders.
- If there isn't, then you don't need to add the header cells because they will never be populated at all (not even for one single competition)

app/webpacker/components/MyCompetitions/UpcomingCompetitionTable.jsx
```diff
@@ -135,7 +127,6 @@ export default function UpcomingCompetitionTable({
                   <ReportTableCell
                     competitionId={competition.id}
                     permissions={permissions}
-                    canViewDelegateReport={canViewDelegateReport}
                   />
```
gregorbg, line 130: In line with my other comment above, this should probably also be wrapped in a `canViewDelegateReport && ()` check, to make sure that the number of header cells and the number of body cells matches.

app/webpacker/components/MyCompetitions/TableCells.jsx
```diff
@@ -35,14 +35,13 @@ export function DateTableCell({ competition }) {
 }
 
 export function ReportTableCell({
-  permissions, competitionId, isReportPosted, canViewDelegateReport,
+  permissions, competitionId, isReportPosted, isPastCompetition,
 }) {
+  const canViewDelegateReport = permissions.can_view_delegate_report.scope === '*' || permissions.can_view_delegate_report.scope.includes(competitionId);
   if (canViewDelegateReport) {
```
gregorbg, line 41: What if `canViewDelegateReport` is false? I think we should just return empty cells in that case, to make sure we still match the expected number of cells with respect to the table headers.

gregorbg (changes requested):
Making sure that the number of columns in the header stays in sync with the numbers of cells from the body is complicated :sweat_smile:

app/webpacker/components/MyCompetitions/TableCells.jsx
pranav-027: It was for permission of editing report i have created a new cosnt `canEditDelegateReport` for it

app/webpacker/components/MyCompetitions/UpcomingCompetitionTable.jsx
pranav-027: My interpretation was 3 cells for - View ,edit and warning logo and we don't want show warning logo for upcoming competitions

app/webpacker/components/MyCompetitions/UpcomingCompetitionTable.jsx
pranav-027: done

app/webpacker/components/MyCompetitions/TableCells.jsx
pranav-027: done

app/webpacker/components/MyCompetitions/TableCells.jsx
gregorbg: cool!

app/webpacker/components/MyCompetitions/UpcomingCompetitionTable.jsx
gregorbg: Now there's still one header cell permanently disappearing, at least in my Git diff view

pranav-027: added back

gregorbg (changes requested):
Okay, aaaalmost there! Now the header definition in the "Upcoming" table _always_ contains three cells, as soon as  `canViewDelegateReport` is true.
But in the cell body, two of those three cells are only rendered if some very complex `can_organize_competitions` checks pass.

(Sorry I cannot directly comment in the code because the body cells are not part of the Git diff...)

pranav-027:
Okay got it! Do you think we should change the header definition to compile the first two cells separately? Or we should change the logic in body cells?

gregorbg:
Hm... No strong preference. If I had to choose, we should probably make the header cells conditional.
As in, the body already does some permission check -- we use the same one except for one single competition ID (in the body) we check for `allCompetitionIds.some(...)`

pranav-027:
Its done and while implementing I found out one more issue in this. The Edit and Registrations buttons should be visible to organizers also (according to the old ERB file). So I have changed the checks from `can_organize_competitions` to `can_administer_competitions`

gregorbg (approved):
Cool, thanks for bearing with all of my nitpicks!

kr-matthews:
I think this PR has caused the missing cells in the body?

image

Issue #10554, opened 2025-01-07
Edit Registration Page Improvements


dunkOnIT:
## High:

- [ ] Add WCA ID in brackets next to name (and hyperlink to profile)
- [ ] Make user's name a hyperlink to the Edit Profile page? (or figure out a similar way to let the organizer get to a user's "Edit Profile" page)

## Medium: 

- [ ] Display dates of history events as ISO-8601 (YYYY-MM-DD)

## Low:
- [ ] When cancelling a user's registration from the edit menu, all their registration data (event ids, comments, guests) is cleared from the form even though it is (presumably) still stored in the database. The data should remain, as it makes it inconvenient for an organizer who cancelled by accident to reinstate someone's registration

PR #10788, opened 2025-02-08
dates of registration history events as ISO-8601

pranav-027:
subtask of: https://github.com/thewca/worldcubeassociation.org/issues/10554
image

### DIFF
--- a/app/webpacker/lib/utils/dates.js
+++ b/app/webpacker/lib/utils/dates.js
@@ -101,9 +101,12 @@ export const getFullDateTimeString = (dateTime, timeZone = 'local') => DateTime.
   .setZone(timeZone)
   .toLocaleString(DateTime.DATETIME_FULL_WITH_SECONDS);
 
-export const getIsoDateWithTimeString = (dateTime, timeZone = 'local') => DateTime.fromISO(dateTime)
+export const getIsoDateString = (dateTime, timeZone = 'local') => DateTime.fromISO(dateTime)
   .setZone(timeZone)
-  .toISODate() + ' ' + DateTime.fromISO(dateTime)
+  .toISODate();
+
+
+export const getTimeWithSecondsString = (dateTime, timeZone = 'local') => DateTime.fromISO(dateTime)
   .setZone(timeZone)
   .toFormat(DateTime.TIME_WITH_SECONDS);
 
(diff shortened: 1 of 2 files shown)


FinnIckler:
What about a full time stamp here?

pranav-027:
> What about a full time stamp here?

image

You can see the time when you hover over date

gregorbg (changes requested):
Hm, not a big fan of hiding the time stamp by default, as this is very crucial information (as in, you maybe don't need Registration History at all in your daily operations as organizer, but *if* you need it then you definitely also need the exact time).

Please explore the different variations of `toLocaleString` and the corresponding `DateTime` constants.

gregorbg:
I also can see that the concrete choice of words in the issue text is a bit misleading, as in "Format the dates in ISO [but leave the time as-is!]"

pranav-027:
@gregorbg is this fine:
image

app/webpacker/lib/utils/dates.js
```diff
@@ -101,6 +101,10 @@ export const getFullDateTimeString = (dateTime, timeZone = 'local') => DateTime.
   .setZone(timeZone)
   .toLocaleString(DateTime.DATETIME_FULL_WITH_SECONDS);
 
+export const getIsoDateString = (dateTime, timeZone = 'local') => DateTime.fromISO(dateTime)
+  .setZone(timeZone)
+  .toFormat('yyyy-MM-dd HH:mm:ss');
```
gregorbg, line 106: Yup, looking good now, but why are you manually glueing together this format string?
Have you looked at https://moment.github.io/luxon/#/formatting?id=presets whether there is any `toLocaleString` that represents your date appropriately? And if not, why is no format from that list good enough?

pranav-027: Yes I had checked all presets and DATETIME_SHORT_WITH_SECONDS (10/14/1983, 1:30:23 PM) or DATETIME_MED_WITH_SECONDS (Oct 14, 1983, 1:30:23 PM) looks good but none of them are in ISO format

gregorbg: Okay, then what about a combination of `toISODate()` and `DateTime.TIME_WITH_SECONDS`? Glueing these two together would still be preferable over a completely hand-painted format string.

gregorbg (approved):

PR #10807, opened 2025-02-10
Allow organizers to prevent any paid user from cancelling

dunkOnIT:
I'm adding an enum so that we can give organizers more control over which users can cancel their registrations. 

TODO:  
- [x] Add enum
- [x] Replace all existing functionality of old field
  - [x] Change tests
  - [x] Replace all instances in code
- [x] Add validations for this new field (if any - haven't thought of any yet besides perhaps `use_wca_registration`)
- [x] Add tests against `restrict_any_paid` field - this is where we add the new feature of preventing cancellations by users who have paid
- [x] Update competition form
- [x] Frontend change to disable Cancel button in the payment case
- [x] Test out in frontend

This will deprecate `allow_self_delete_after_acceptance`, which we can remove in a follow-up migration after a few weeks of this running without issue in the wild.


app/models/competition.rb
```diff
@@ -95,6 +95,8 @@ class Competition < ApplicationRecord
     restricted: 2,
   }, prefix: true
 
+  enum :cancellation_restrictions, [:no_restrictions, :restrict_accepted, :restrict_paid], prefix: :registration_cancel
```
gregorbg, line 98: I don't think you need to include the word `restrict` in the values for an enum which is already called `restrictions`.

db/migrate/20250210071551_expand_cancellation_restrictions_for_competitions.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class ExpandCancellationRestrictionsForCompetitions < ActiveRecord::Migration[7.2]
+  def up
+    add_column :Competitions, :cancellation_restrictions, :integer, default: 0, null: false
+    Competition.where(allow_registration_self_delete_after_acceptance: false)
+               .update_all(cancellation_restrictions: 1)
```
gregorbg, line 7: You can use the enum value literal here.

gregorbg (changes requested):
I'm not too wild about the naming scheme in general. Why are you coming at this from the "restrictions" perspective? In my opinion, the natural question to ask is "Are competitors allowed to cancel their own registrations?". And if that's the question, then the answer should be:
- Yes, anytime.
- Yes, but only if they aren't accepted yet.
- Yes, but only if they haven't paid yet.

In your changes to the YAML file, you use descriptions which indicate when people are NOT allowed to cancel, but perhaps it's more accessible if you specify when you DO allow people to cancel.

app/models/competition.rb
dunkOnIT: What I found in practice is that calling `competition.{enum_value}` makes it unclear what `no_restrictions` means in that context - I'll see how it comes out by taking the "when can they cancel" approach

db/migrate/20250210071551_expand_cancellation_restrictions_for_competitions.rb
dunkOnIT: Updated

dunkOnIT:
Re naming: I think your suggestion (as I understood it) simplifies the naming in the model quite a bit, and I like it. 

In the context of the competition form, I do think that the organizer wants to make an assertion about when the user CAN'T cancel - otherwise, they wouldn't be touching the field at all. From that perspective, I like wording that leads with the restriction being applied - ie, paid/accepted users can't cancel. A wording/information schema that puts the relevant info (paid/accepted) at the end of the sentence is harder to parse than one that puts it at the start. 

Happy to put more effort into this thinking along your lines - just my 2c after playing with the wording in the comp form for a bit

app/models/competition.rb
```diff
@@ -95,6 +95,8 @@ class Competition < ApplicationRecord
     restricted: 2,
   }, prefix: true
 
+  enum :user_can_cancel, [:always, :not_accepted, :unpaid], prefix: :can_cancel
```
gregorbg, line 98: If you just pass `prefix: true`, then the name of the enum will become the prefix. Arguably cleaner and there's not much of a difference between `user_can_cancel_always` and `can_cancel_always`.

gregorbg: Also, maybe `competitor` instead of `user`?

app/models/user.rb
```diff
@@ -916,12 +916,6 @@ def can_edit_registration?(registration)
     can_manage_competition?(registration.competition) || (registration.user_id == self.id && editable_by_user)
   end
 
-  def can_delete_registration?(registration)
-    # A registration can only be deleted by a user after it has been accepted if the organizers allow
-    can_delete_by_user = (!registration.accepted? || registration.competition.registration_delete_after_acceptance_allowed?)
-    can_manage_competition?(registration.competition) || (registration.user_id == self.id && can_delete_by_user)
-  end
-
```
gregorbg, line 924: Why not keep this method and use it in the registration checker, instead of having all incredibly dense logic in oh-so-many lines of code over there?

lib/registrations/registration_checker.rb
```diff
@@ -206,7 +206,8 @@ def validate_update_status!(new_status, competition, current_user, target_user,
 
         # Raise an error if competition prevents users from cancelling a registration once it is accepted
         raise WcaExceptions::RegistrationError.new(:unauthorized, Registrations::ErrorCodes::ORGANIZER_MUST_CANCEL_REGISTRATION) if
-          !competition.allow_registration_self_delete_after_acceptance && registration.accepted?
+        (competition.can_cancel_not_accepted? && registration.accepted?) ||
+        (competition.can_cancel_unpaid? && registration.paid_entry_fees > 0)
```
gregorbg, line 210: (this is the snippet I'm talking about in my other comment)

spec/models/user_spec.rb
```diff
@@ -665,32 +665,6 @@
     end
   end
 
-  describe "can self-delete registration" do
```
gregorbg, line 668: (You may be able to salvage these tests if you decide to accept my other suggestion above)

gregorbg (changes requested):

app/models/competition.rb
dunkOnIT: I like `competitor` instead of `user`, and agree that specifying a different prefix is unnecessary with the new name

app/models/user.rb
dunkOnIT: Good suggestion! Implemented, but with the twist of putting it on the `registration` object which seems to make more sense in the new framework :D 

In general, there is a lot of scope to do more of this, but it's in #10864 as low priority given that it's more tech debt/refactoring work than critical to functionality - `registration_checker` (at least at the moment) is fairly maintainable

spec/models/user_spec.rb
gregorbg: Kindly ping re salvaging tests. These would need to be ported to `registration_spec` accordingly, but with the trouble of `enums` returning raw ints vs returning the "nice-ified" literal value, and the open question of whether that literal is a string or a sym, having tests won't hurt.

app/models/registration.rb
```diff
@@ -386,6 +386,19 @@ def events_to_associated_events(events)
     end
   end
 
+  def permit_user_cancellation?
+    case competition.competitor_can_cancel
```
gregorbg, line 390: Could you either do `.to_sym` and check symbols (like `where :always`) or do `to_s` please?
Just want to make sure we don't run into unexpected behavir because in Rails `"foo" == :foo` evaluates to `false`.

gregorbg (changes requested):

spec/models/user_spec.rb
dunkOnIT: The tests are covered in registration_checker_spec - the intention is to move them into registration_spec eventually as part of a larger refactor of the checker, but for now all the test cases I'm concerned about are covered

gregorbg (commented):
I was a few seconds short of merging this, but then it occured to me: This changes user interaction with the competition form significantly!

Previously, the default for the old `allow_registration_self_delete_after_acceptance` used to be `false`. That means, if someone created a new competition and then ignored the relevant field in the form ("I know it's false by default, which matches what I always want for my competitions, so I just scroll over it") competitors would not be allowed to cancel.

That same behavior would now lead to the competitors being allowed to cancel anytime under any circumstances (default value for `competitor_can_cancel` is `0` which translates to `:always`). So the person from my first example who casually creates a competition might run into a complete misunderstanding.

Now of course we want to live in an ideal world where every Delegate and every organizers reads the full form carefully. But we are both adult enough to know how many people consciously read a 5-screen-long webform (that they use on a weekly basis) in great detail every single time. But this is something even WCAT (whose job it is to read the forms carefully) cannot catch easily because they don't know what the Delegate's original intention was. From WCAT's perspective, the default choice in the dropdown (when you ignore the field and don't touch it) could be the legitimate, willful intention of the Delegate who submitted the form.

Options:
1. Send an email. Hope people will read it.
2. Change the defaults around
    1. Please don't pick a non-zero default value for the raw DB column. Change the order of enum entries if anything
3. Rename the whole enum again (not preferrable because we had already settled on a good name)

app/models/registration.rb
```diff
@@ -386,6 +386,19 @@ def events_to_associated_events(events)
...
+  def permit_user_cancellation?
+    case competition.competitor_can_cancel.to_sym
+    when :always
+      true
+    when :not_accepted
+      !accepted?
+    when :unpaid
+      paid_entry_fees == 0
+    else
+      false
```
gregorbg, line 398: FWIW this state is never reachable. Rails `enum` introduces validations under the hood that make sure you can never set any (in this case, integer) value that is not covered by the enum.

Do we _want_ cases where the cancellation policy is just "straight-out nope"? In that case, we need a separate `never` enum constant.

gregorbg (changes requested):
Another comment, indicating that we (probably?) need a fourth `never` state in the enum. And then we could fix my previous concerns rather easily by flipping the order of entries in the enum around, i.e. listing the new `never` first.

That would mean that peeps who rely on the default value get the exact same behavior as it used to be with the old boolean implementation

gregorbg (approved):

PR #10813, opened 2025-02-11
Fix null wca_id in search users

maxidragon:
I think adding a link to the WCA ID instead of the name makes more sense because some users don't have a WCA ID, and currently, clicking on their name points to https://worldcubeassociation.org/persons/null


gregorbg (approved):

gregorbg:
I just merged main in the hopes that this will fix the intermittent test failure

PR #10815, opened 2025-02-11
WCA Live: Recompute Advancing on update

FinnIckler:
Currently can't get tests to run due to (I think?) some issues how we associate registrations with events


app/models/live_result.rb
```diff
@@ -42,7 +45,56 @@ def complete?
...
     def notify_users
       ActionCable.server.broadcast("results_#{round_id}", serializable_hash)
     end
+
+    def recompute_advancing
+      round_results = LiveResult.where(round: round)
+      round_results.update_all(advancing: false)
+
+      missing_attempts = round.total_registrations - round_results.count
+      puts(round.registrations.inspect)
```
gregorbg, line 61: :smiling_imp:

FinnIckler: me trying to debug the issue before I added another expect

app/models/live_result.rb
```diff
@@ -36,13 +38,65 @@ def potential_score
     complete? ? self[rank_by.to_sym] : best_possible_score
   end
 
+  def best_possible_score
+    1
+  end
```
gregorbg, line 43: Why can't this be a constant, along the lines of `BEST_POSSIBLE_SCORE`?

FinnIckler: it can!

gregorbg (approved):

PR #10819, opened 2025-02-12
WCA Live: Podiums and Competitors Views

FinnIckler:
Three Easy displaying pages. Podiums relies on `ResultsTable` from the Admin PR, but I am putting this out there already for a potential review while I am on vacation

### DIFF
--- a/app/webpacker/components/Live/ByPerson/index.jsx
+++ b/app/webpacker/components/Live/ByPerson/index.jsx
@@ -47,8 +47,7 @@ function PersonResults({
               </Table.Row>
             </Table.Header>
             <Table.Body>
-              {eventResults.map((r) => {
-                const { round, attempts } = r;
+              {eventResults.map(({ round, attempts }) => {
                 return (
                   <Table.Row>
                     <Table.Cell width={1}>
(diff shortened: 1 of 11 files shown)


app/webpacker/components/Live/ByPerson/index.jsx
```diff
@@ -0,0 +1,78 @@
...
+                    {num + 1}
+                  </Table.HeaderCell>
+                ))}
+                <Table.HeaderCell>Average</Table.HeaderCell>
+                <Table.HeaderCell>Best</Table.HeaderCell>
+              </Table.Row>
+            </Table.Header>
+            <Table.Body>
+              {eventResults.map((r) => {
+                const { round, attempts } = r;
```
kr-matthews, line 51:
```suggestion
              {eventResults.map(({ round, attempts }) => {
```

kr-matthews: Ah oops I didn't notice `r` was used elsewhere, sorry.

gregorbg:
(no need to be as strict as usual on the review here, we're just prototyping at this stage. More thorough UI polishes are due before any of this potentially goes live.)

gregorbg (approved):

PR #10828, opened 2025-02-13
tried out a validation for the image

dunkOnIT:
We had an issue in a comp where a PDF file was accidentally uploaded to a delegate report in the Setup Images section, breaking the page. ChatGPT guided me to try and add a validation for the image type, but doing so yields the following error. My suspicion is that we should handle this in the frontend, which ChatGPT didn't have any inspiringly good ideas for. 

My suspicion is this is a "React will fix" issue, but posting in case the code is a useful starting point, and because we don't want to need to manually delete _too_ many images via the console
 
image

### DIFF
--- a/app/models/delegate_report.rb
+++ b/app/models/delegate_report.rb
@@ -61,6 +61,15 @@ class DelegateReport < ApplicationRecord
     end
   end
 
+  validate :setup_image_format, if: -> { setup_images.attached? }
+  private def setup_image_format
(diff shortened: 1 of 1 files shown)


gregorbg:
Thanks for the initial effort! I'm commandeering this one because ChatGPT doesn't know that we use the [activestorage-validator gem](https://github.com/aki77/activestorage-validator).

gregorbg (approved):

PR #10831, opened 2025-02-14
Implement ticket comments

danieljames-dj:
This is the last thing in tickets design as mentioned here: https://github.com/thewca/worldcubeassociation.org/issues/9463.

(There are more ticket metadata yet to come though)


db/migrate/20250214015507_create_ticket_comments.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class CreateTicketComments < ActiveRecord::Migration[7.2]
+  def change
+    create_table :ticket_comments do |t|
+      t.references :ticket, null: false
+      t.string :comment
```
gregorbg, line 7: `string` translates to `VARCHAR(255)` I believe, This is most likely too small, you should consider going with `t.text` instead

config/routes.rb
```diff
@@ -206,6 +206,7 @@
   resources :tickets, only: [:index, :show] do
     post 'update_status' => 'tickets#update_status', as: :update_status
     get 'edit_person_validators' => 'tickets#edit_person_validators', as: :edit_person_validators
+    resources :comments, only: [:index, :create], module: :tickets
```
gregorbg, line 209: What does `module: :tickets` do? I cannot find `module` being used anywhere else in our routes file (unless CTRL+F betrayed me) so I'm wondering whether it's necessary.

app/webpacker/components/Tickets/index.jsx
```diff
@@ -51,12 +53,25 @@ function TicketContent({ ticketDetails, currentStakeholder, sync }) {
...
     </>
   );
 }
 
-export default function Tickets({ id }) {
+export default function Wrapper(args) {
+  return (
+    <WCAQueryClientProvider>
+      {/* eslint-disable-next-line react/jsx-props-no-spreading */}
+      <Tickets {...args} />
```
gregorbg, line 69: Why are you not spreading out the args directly in the definition of `Wrapper`, like any other React component would?

app/webpacker/components/Tickets/index.jsx
```diff
@@ -51,12 +53,25 @@ function TicketContent({ ticketDetails, currentStakeholder, sync }) {
         sync={sync}
         currentStakeholder={currentStakeholder}
       />
+      <TicketComments
+        ticketId={ticketDetails.ticket.id}
+        currentStakeholder={currentStakeholder}
+      />
       <TicketLogs logs={ticketDetails.ticket.ticket_logs} />
```
gregorbg, line 60: In `TicketLogs`, you already hard-pass the logs. In `TicketComments`, you only pass the ID and fetch the comments dynamically. Both approaches are fine by themselves, but why do you mix/combine them in this unusual way?

app/webpacker/components/Tickets/api/getComments.js
```diff
@@ -0,0 +1,9 @@
+import { fetchJsonOrError } from '../../../lib/requests/fetchWithAuthenticityToken';
+import { viewUrls } from '../../../lib/requests/routes.js.erb';
+
+export default async function getComments({ ticketId }) {
+  const { data } = await fetchJsonOrError(
+    viewUrls.tickets.listComments(ticketId),
+  );
+  return data?.comments || [];
```
gregorbg, line 8: Not a big fan of already accessing the data here. Do we do this in other places throughout the code?

app/webpacker/components/Tickets/TicketComments.jsx
```diff
@@ -0,0 +1,67 @@
...
+            <CommentContent>
+              <CommentAuthor as="a">{comment.acting_user.name}</CommentAuthor>
+              <CommentMetadata>
+                <div>{DateTime.fromISO(comment.created_at).toLocal().toRelative()}</div>
+              </CommentMetadata>
+              <CommentText>{comment.comment}</CommentText>
+            </CommentContent>
+          </Comment>
+        ))}
+      </CommentGroup>
```
gregorbg, line 56: Not a big deal, but: You can access all of the `Comment` "sub-components" as `Comment.Avatar` etc. (note the dot behind `Comment`). This will save you a lot of imports.

app/webpacker/components/Tickets/TicketComments.jsx
```diff
@@ -0,0 +1,67 @@
...
+import getComments from './api/getComments';
+import Loading from '../Requests/Loading';
+import Errored from '../Requests/Errored';
+import '../../stylesheets/semantic/components/comment.min.css';
+
+export default function TicketComments({ ticketId, currentStakeholder }) {
+  const [createComment, setCreateComment] = useState(false);
+
+  const {
+    data: comments, isLoading, isError, refetch, isRefetching,
```
gregorbg, line 25: Why do you have to export `isRefetching` with the **re** part specifically? Shouldn't `isFetching` as a replacement for `isLoading` and `isRefetching` be good?

gregorbg: See https://tanstack.com/query/latest/docs/framework/react/reference/useQuery for further information on the meaning of the different `is...` flags

app/webpacker/components/Tickets/TicketCommentCreate.jsx
```diff
@@ -0,0 +1,54 @@
...
+import Loading from '../Requests/Loading';
+import Errored from '../Requests/Errored';
+
+export default function TicketCommentCreate({
+  open, onClose, ticketId, currentStakeholder, refetchComments,
+}) {
+  const [comment, setComment] = useInputState();
+
+  const {
+    mutate: doCreateComment,
```
gregorbg, line 16: Instead of the `do...` prefix, maybe call it `createCommentMutation`? That also avoids the naming collision with the API function.

app/models/ticket_comment.rb
```diff
@@ -0,0 +1,16 @@
...
+  belongs_to :ticket
+  belongs_to :acting_user, class_name: 'User', foreign_key: :acting_user_id
+
+  DEFAULT_SERIALIZE_OPTIONS = {
+    include: %w[acting_user],
+  }.freeze
+
+  def serializable_hash(options = nil)
+    json = super(DEFAULT_SERIALIZE_OPTIONS.merge(options || {}))
+    json[:class] = self.class.to_s.downcase
```
gregorbg, line 13: If I'm not mistaken, this will take the class name `TicketComment`, and convert it to downcase like `ticketComment`. Why not `dasherize` or `underscore` or any other casing?

And on that note, why do you need to export the class name anyways?

app/models/ticket_comment.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class TicketComment < ApplicationRecord
+  belongs_to :ticket
+  belongs_to :acting_user, class_name: 'User', foreign_key: :acting_user_id
```
gregorbg, line 5: Bit surprised that the `foreign_key` is necessary. Usually Rails assumes `#{name_of_link}_id` as default. Have you tried without it?

db/migrate/20250214015507_create_ticket_comments.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class CreateTicketComments < ActiveRecord::Migration[7.2]
+  def change
+    create_table :ticket_comments do |t|
+      t.references :ticket, null: false
```
gregorbg, line 6: Why no foreign key here?

gregorbg: (You do have foreign keys below for hte other two `references`, that's why I'm asking)

gregorbg (changes requested):

db/migrate/20250214015507_create_ticket_comments.rb
danieljames-dj: I actually forgot it TBH, I've added it now.

app/webpacker/components/Tickets/index.jsx
gregorbg: Rule of thumb: If `eslint` complains about it (which in this case you can see by the fact that you had to introduce `eslint-disable-next-line` comment) then there's a good reason why it complains.
In 9 out of 10 cases, you should follow whatever ESLint tells you instead of suppressing the warning.

app/models/ticket_comment.rb
gregorbg: > I simply copied the thing from there without reading it completely.

😐 Copying from other places is totally fine, but only if you understand what you're copying!

app/webpacker/components/Tickets/TicketCommentCreate.jsx
```diff
@@ -0,0 +1,54 @@
...
+
+  const {
+    mutate: createCommentMutation,
+    isPending,
+    isError,
+  } = useMutation({
+    mutationFn: createComment,
+    onSuccess: () => {
+      setComment('');
+      refetchComments();
```
gregorbg, line 23: You shouldn't need to do this here. Your thinking probably comes from the old `useLoadedData` and `sync`, which was fine because there was no other way.

Now we're starting to use `tanstack-query` more and more, which has much better features and is more powerful. So you can do something you didn't do before: You can access the `QueryClient` (the thing that stores and basically "caches" backend API responses using query keys) and tell it to replace data directly.

gregorbg: I will make a suggestion around the whole mutation to show you how this basically looks like. It is only a skeleton, and I will leave some gaps in the details so that you can fill them in yourself as an exercise to get familiar with the "new way" of doing things.

app/webpacker/components/Tickets/TicketCommentCreate.jsx
```diff
@@ -0,0 +1,54 @@
...
+      onClose();
+    },
+  });
```
gregorbg, line 26:
```suggestion
const queryClient = useQueryClient();

const {
  mutate: createCommentMutation,
  isPending,
  isError,
} = useMutation({
  mutationFn: createComment,
  onSuccess: (returnValueOfMutationFn, parametersPassedToMutationFn) => {
    setComment('');
    queryClient.setQueryData(['my-query-key', myQueryId, someOtherId], (previousData) => [...previousData, parametersPassedToMutationFn]);
    onClose();
  },
});
```

gregorbg: Essentially, you're writing into the client-side cache storage, which lets you completely drop the `refetchComments` part and save one "fetch from backend -> set loading state -> display" cycle. In particular, it saves network traffic and database load. Hooray!

gregorbg: (Note that in the future, you may run across a very special edge case where it is necessary to manually refetch. It's not 100% forbidden to do that. But in at least 8/10 cases, like simple "add a comment" or "manipulate an existing comment", you can easily modify the data in-place, without having to refetch the exact same comment which you just added anyways)

gregorbg: Reference: https://tanstack.com/query/latest/docs/framework/react/guides/mutations#mutation-side-effects

app/webpacker/components/Tickets/TicketComments.jsx
```diff
@@ -0,0 +1,53 @@
...
+import getComments from './api/getComments';
+import Loading from '../Requests/Loading';
+import Errored from '../Requests/Errored';
+import '../../stylesheets/semantic/components/comment.min.css';
+
+export default function TicketComments({ ticketId, currentStakeholder }) {
+  const [createComment, setCreateComment] = useState(false);
+
+  const {
+    data: { comments } = { comments: [] }, isFetching, isError, refetch,
```
gregorbg, line 15: If the `comments` are really all that you care about, then you should pass a `select` option: https://tanstack.com/query/latest/docs/framework/react/guides/render-optimizations#select

app/webpacker/components/Tickets/TicketComments.jsx
```diff
@@ -0,0 +1,53 @@
...
+  } = useQuery({
+    queryKey: ['ticket-comments', ticketId],
+    queryFn: () => getComments({ ticketId }),
```
gregorbg, line 18:
```suggestion
    queryKey: ['ticket-comments', ticketId],
    queryFn: () => getComments({ ticketId }),
    select: (data) => data.comments,
```

gregorbg: see above. Note that I wrote this code blindly on GH, without actually testing whether it covers your use-case. Please treat this suggestion with caution :)

gregorbg (changes requested):

app/controllers/ticket_comments_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+  before_action :authenticate_user!
+
+  def index
+    ticket = Ticket.find(params.require(:ticket_id))
+
+    # Currently only stakeholders can access the ticket comments.
+    return head :unauthorized unless ticket.can_user_access?(current_user)
+
+    comments = ticket.ticket_comments.order(created_at: :desc)
+    paginate json: comments
```
gregorbg, line 15: This pagination is kinda pointless if you don't query for individual pages in the frontend.

app/webpacker/components/Tickets/TicketComments.jsx
```diff
@@ -0,0 +1,53 @@
...
+import '../../stylesheets/semantic/components/comment.min.css';
+
+export default function TicketComments({ ticketId, currentStakeholder }) {
+  const [createComment, setCreateComment] = useState(false);
+
+  const {
+    data: comments, isFetching, isError, refetch,
+  } = useQuery({
+    queryKey: ['ticket-comments', ticketId],
+    queryFn: () => getComments({ ticketId }),
```
gregorbg, line 18: This "plain" fetching is dangerous, now that you've added pagination: Whenever there are more than 20 (or whatever is the page limit) comments, you would just happily display these 20 without realizing that there's more.

Either get rid of the pagination (for now) or implement it properly, please :)

gregorbg (changes requested):

app/controllers/ticket_comments_controller.rb
danieljames-dj: Okay removed pagination, will implement this in frontend later and add it.

app/webpacker/components/Tickets/TicketComments.jsx
danieljames-dj: Okay, removed pagination for now.

app/controllers/ticket_comments_controller.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+class TicketCommentsController < ApplicationController
+  include Rails::Pagination
```
gregorbg, line 4: Now this inclusion is useless :upside_down_face:

gregorbg (changes requested):

app/controllers/ticket_comments_controller.rb
danieljames-dj: Whoopsieee, removed.

gregorbg (approved):

Issue #10542, opened 2025-01-06
Registration Admin Page Improvements


dunkOnIT:
## High

- [ ] Add popup telling the user that registrations will be added to the waiting list in the order they were selected
- [ ] Improve loading times and navigation (we just need to do _something_ - pagination + search/filter?) to make it more usable, and then this will move down to Medium) [#10524]

## Medium
- [ ] Add an index/number showing the position in the list - for Waiting List, this should explicitly show the waiting list position
- [ ] Add explanation that waiting list is ordered by the order users are added, not by payment timestamp
- [ ] Status change buttons get in the way when you select a registration in mobile. One solution here is to show icons across the top of the screen instead of the button with all the text (to reproduce: Mobile view in a competition list -> Scroll down so that the page is full of registrations -> select the 2nd registration from the top -> it will be blocked by the reg action buttons)

jacklee1792:
Noticed the same issue with the Finland flag, the `fi` class name is also provided by the flag-icons CSS and collides with the `fi` name from Fomantic-UI. You can observe that the behaviour is correct when using `finland flag` instead of `fi flag`

dunkOnIT:
Sounds good - shoot me a ping if you hit any backend changes you need

PR #10836, opened 2025-02-14
Registration admin: hide-able tables

kr-matthews:
And move waitlist above accepted.

And fix the footer (had wrong number of cells sometimes).

[Screencast from 2025-02-14 12:30:56 PM.webm](https://github.com/user-attachments/assets/691edda1-f563-49d7-a429-b9a8d6442fee)

### DIFF
--- a/app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationTableFooter.jsx
+++ b/app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationTableFooter.jsx
@@ -13,9 +13,9 @@ const moneyCountHumanReadable = (registrations, competitionInfo) => {
 };
 
 export default function RegistrationAdministrationTableFooter({
-  columnsExpanded, registrations, competitionInfo, eventsToggled,
+  columnsExpanded, registrations, competitionInfo,
 }) {
-  const { dob, events, comments } = columnsExpanded;
+  const { events: eventsAreExpanded, comments: commentsAreVisible } = columnsExpanded;
 
   const newcomerCount = registrations.filter(
     (reg) => !reg.user.wca_id,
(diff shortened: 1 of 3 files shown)


gregorbg:
> And move waitlist above accepted.

Where was this decided? I am happy to defer to general consensus, but at least to me personally it feels **hugely** unintuitive to have the waitlist above accepted.

(In my mind, I would perceive the first table row as "first place, first to be accepted once a spot clears up" on the waiting list, and the last table row as "last place, virtually no chance to make it to accepted" on the waiting list. If you move Waitlist above Accepted, it would make the poor sap from last place visually look very very close to the accepted list. Is this just me, or did I perhaps miss a memo?)

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationTableFooter.jsx
```diff
@@ -43,18 +44,24 @@ export default function RegistrationAdministrationTableFooter({
           registrations.length - newcomerCount
         } Returners = ${registrations.length} People`}
       </Table.Cell>
-      <Table.Cell>{`${countryCount}  Countries`}</Table.Cell>
-      <Table.Cell />
-      { competitionInfo['using_payment_integrations?'] && <Table.Cell>{moneyCountHumanReadable(registrations, competitionInfo)}</Table.Cell>}
-      { eventsToggled ? competitionInfo.event_ids.map((evt) => (
+      <Table.Cell>{`${countryCount} Countries`}</Table.Cell>
+      <Table.Cell key="registered on" />
```
gregorbg, line 48: Out of interest: Why are you passing in a key outside of a loop?

gregorbg (approved):
Code looks fine, but I'd still like to find about why the order was switched before merging.

kr-matthews:
> > And move waitlist above accepted.
> 
> Where was this decided? I am happy to defer to general consensus, but at least to me personally it feels **hugely** unintuitive to have the waitlist above accepted.
> 
> (In my mind, I would perceive the first table row as "first place, first to be accepted once a spot clears up" on the waiting list, and the last table row as "last place, virtually no chance to make it to accepted" on the waiting list. If you move Waitlist above Accepted, it would make the poor sap from last place visually look very very close to the accepted list. Is this just me, or did I perhaps miss a memo?)

It's listed in #10542. I'll also be adding waitlist position explicitly to the waitlist table in a future PR (probably the next PR), so hopefully that addresses your concern.

The order makes a lot of sense to me (and others, as far as I'm aware) because registrations flow from pending -> waitlist -> accepted -> cancelled/rejected, potentially skipping some steps along the way.

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationTableFooter.jsx
kr-matthews: Basically as an equivalent to
```
// this footer cell corresponds to the 'registered on' column
<Table.Cell />
```
since comments have a tendency to be ignored/missed and get out-of-date. Obviously this could still happen, but it seems less likely.

gregorbg:
Okay, I'm happy to take your explanation and the Issue as an argument in favor of merging!

PR #10837, opened 2025-02-15
Unify flag components (& fix Finland)

kr-matthews:
In particular, fixes Finland flag, and show flags with tooltip for multi-country instances.

Before:
before before2
After: 
after after2

Screenshot from 2025-02-14 19-44-18

Formerly, half our flags were using our custom `RegionFlag` (formerly named `CountryFlag`), and half were using `Flag` from Semantic UI. The former were working well, the latter were smaller, messed up Finland, and didn't show anything for multi-country regions.

I'm not sure when `comp.country_iso2` would be `null`, but I included it since there were existing `?`s.

### DIFF
--- a/app/webpacker/components/wca/RegionFlag.js
+++ b/app/webpacker/components/wca/RegionFlag.js
@@ -3,27 +3,28 @@ import { Popup } from 'semantic-ui-react';
 import classnames from 'classnames';
 import { countries } from '../../lib/wca-data.js.erb';
 
-export default function CountryFlag({ iso2, withoutTooltip = false }) {
+/** Works with countries, continents, and world. */
+export default function RegionFlag({ iso2, withoutTooltip = false }) {
   if (withoutTooltip) {
-    return <CountryFlagWithoutTooltip iso2={iso2} />;
+    return <RegionFlagWithoutTooltip iso2={iso2} />;
   }
 
-  return <CountryFlagWithTooltip iso2={iso2} />;
+  return <RegionFlagWithTooltip iso2={iso2} />;
 }
 
-function CountryFlagWithTooltip({ iso2 }) {
+function RegionFlagWithTooltip({ iso2 }) {
   return (
     <Popup
       id="resultCountryFlagTooltip"
       position="top center"
       content={countries.byIso2[iso2].name}
-      // trigger <CountryFlagWithoutTooltip iso2={iso2} /> doesn't work (??)
+      // trigger <RegionFlagWithoutTooltip iso2={iso2} /> doesn't work (??)
       trigger={<span className={classnames('fi', `fi-${iso2.toLowerCase()}`)} />}
     />
   );
 }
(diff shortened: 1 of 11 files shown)


app/webpacker/components/wca/RegionFlag.js
```diff
@@ -0,0 +1,30 @@
...
+  return <RegionFlagWithTooltip iso2={iso2} />;
+}
+
+function RegionFlagWithTooltip({ iso2 }) {
+  return (
+    <Popup
+      id="resultCountryFlagTooltip"
+      position="top center"
+      content={countries.byIso2[iso2].name}
+      trigger={<span><RegionFlagWithoutTooltip iso2={iso2} /></span>}
```
gregorbg, line 21: Why the extra `span`?

gregorbg (commented):
Cool! I almost wanna approve, but not without having a discussion about `flag-icons`. What is wrong with using the `Flag` component from SemUI?
If it's only about Finland, then I would be happy to contribute a CSS override hack for that ISO2 overload, and call it a day.

app/webpacker/components/wca/RegionFlag.js
kr-matthews: For some reason, the `trigger` needs to explicitly start with a `span`, or it doesn't work. There's a comment about it in the `EventSelector` component. I cannot explain why.

kr-matthews:
> Cool! I almost wanna approve, but not without having a discussion about `flag-icons`. What is wrong with using the `Flag` component from SemUI? If it's only about Finland, then I would be happy to contribute a CSS override hack for that ISO2 overload, and call it a day.

We were already using both version, about equally. The benefits of the one I picked are Finland, the ability to handle continents/the world (for multi-location competitions), trying to avoid needing to meddle with css, and they just look bigger and nicer to me.

I have no problem switching back to Semantic UI. And if we do that, then it'll be easier because it's all centralized in one component now. But we'd need to fix Finland and add continent/world support. If you think it's best, feel free to contribute that to this branch/PR, I'm not sure how to proceed.

gregorbg:
> I'm not sure how to proceed.

Compromise: Use `fi` for now, but add a note that we need a more "React-esque" flag integration when moving to NextJS+Chakra.

Also, please add another comment about the weird span thing. That also feels like a glitch that Chakra can resolve "for free"

app/webpacker/components/wca/RegionFlag.js
kr-matthews: I added a comment, but there are a lot of instances in the codebase where we wrap a trigger in a span without any explanation.

gregorbg: These are not our immediate concern. Probably a SemUI glitch that we can get rid of "for free" when moving frameworks.

gregorbg (approved):

PR #10838, opened 2025-02-15
Add requested changes list to edit profile ticket workbench

maxidragon:


app/models/tickets_edit_person.rb
```diff
@@ -60,4 +60,14 @@ def self.create_ticket(wca_id, changes_requested, requester)
...
     end
   end
+
+  DEFAULT_SERIALIZE_OPTIONS = {
+    include: %w[tickets_edit_person_fields],
+  }.freeze
+
+  def serializable_hash(options = nil)
+    json = super(DEFAULT_SERIALIZE_OPTIONS.merge(options || {}))
+    json[:class] = self.class.to_s.downcase
```
danieljames-dj, line 70: This line is actually not needed. Sorry that I suggested you this change over chat, but I came to know late that this line is not needed.

app/webpacker/components/Tickets/TicketWorkbenches/EditPersonTicketWorkbench.jsx
```diff
@@ -26,6 +26,28 @@ function EditPersonValidations({ ticketDetails }) {
   ));
 }
 
+function EditPersonRequestedChangesList({ ticketDetails }) {
+  const { ticket } = ticketDetails;
+  const { tickets_edit_person_fields: requestedChanges } = ticket.metadata;
```
danieljames-dj, line 31: If `ticketDetails` is used only for `requestedChanges`, then maybe you can pass that alone to this class.

app/webpacker/components/Tickets/TicketWorkbenches/EditPersonTicketWorkbench.jsx
```diff
@@ -49,6 +79,9 @@ function EditPersonTicketWorkbenchForWrt({ ticketDetails, actingStakeholderId, s
       <EditPersonValidations
         ticketDetails={ticketDetails}
       />
+      <EditPersonRequestedChangesList
+        requestedChanges={ticketDetails.ticket.metadata?.tickets_edit_person_fields}
```
danieljames-dj, line 83: I assume `ticket` is already available, so `ticket?.metadata?...` should also work I guess.

danieljames-dj (approved):

app/webpacker/components/Tickets/TicketWorkbenches/EditPersonTicketWorkbench.jsx
```diff
@@ -49,6 +79,9 @@ function EditPersonTicketWorkbenchForWrt({ ticketDetails, actingStakeholderId, s
       <EditPersonValidations
         ticketDetails={ticketDetails}
       />
+      <EditPersonRequestedChangesList
+        requestedChanges={ticket.metadata?.tickets_edit_person_fields}
```
gregorbg, line 83: The `?` in here means you expect that the result can be `null` sometimes. But your component above (`EditPersonRequestedChangesList`) does not handle the case where its param (`requestedChanges`) indeed becomes null. It would just crash trying to map over the (potentially non-existent) list.

gregorbg (changes requested):

gregorbg (approved):

PR #10839, opened 2025-02-16
Combined edit location and change region to one

danieljames-dj:
Merged 'edit location' and 'change region' to single action as this was creating two different roles and WRT has to manually delete extra role from database.

### DIFF
--- a/app/controllers/api/v0/user_roles_controller.rb
+++ b/app/controllers/api/v0/user_roles_controller.rb
@@ -188,6 +188,19 @@ class Api::V0::UserRolesController < Api::V0::ApiController
     return head :unauthorized unless current_user.has_permission?(:can_edit_groups, role.group.id)
 
     if group_type == UserGroup.group_types[:delegate_regions]
+      new_role = UserRole.new(
+        user_id: role.user.id,
+        group_id: role.group.id,
+        start_date: Date.today,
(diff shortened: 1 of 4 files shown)


app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -188,82 +188,57 @@ def update
...
-          )
-        end
-      else
-        return render status: :unprocessable_entity, json: { error: "Invalid parameter to be changed" }
+      return render status: :unprocessable_entity, json: { error: "No valid parameters to be changed" } if changes.empty?
+      return head :unauthorized unless current_user.has_permission?(:can_edit_groups, new_role.group_id)
+
+      ActiveRecord::Base.transaction do
+        role.update!(end_date: Date.today)
+        new_role_metadata.save!
```
gregorbg, line 239: This `save!` is redundant. Rails will automatically save all nested models during the `new_role.save!` call below.

gregorbg (changes requested):

app/controllers/api/v0/user_roles_controller.rb
danieljames-dj: Done, thank you

gregorbg (approved):

PR #10840, opened 2025-02-16
Refactor Event Selector to use callbacks

kr-matthews:
Potential follow-up PR: The `activeIdReducer` from Schedule/index could be generalized into a hook and re-used in RegistrationEditor and CompetingStep.


gregorbg (approved):
Thanks a lot! I also like the idea of using a hook for the ID dispatcher, as this is indeed a rather common pattern that will help avoid code duplication. But please don't give it too high of a priority (in other words, feel free to work on it whenever I fall behind on reviews again :sweat_smile:) Registration Admin fixes are more important!

PR #10842, opened 2025-02-17
Bump Semantic React to beta v3

gregorbg:
That... Seems to have just worked?! :sweat_smile: 

Migration guide at https://react.semantic-ui.com/migration-guide/ essentially only mentions the usage of `Ref`. Did that, Webpacker compiles. It also mentions `Visibility` but we don't use that component anyways...


gregorbg:
For reviewers: Due to the fact that you don't need the `Ref` component as an "outer bracket" around ref-ed components anymore, it means that the line count seems high because a lot of nested component indentation changed. Best view the diff without whitespace.

kr-matthews:
[Screencast from 2025-02-16 07:27:17 PM.webm](https://github.com/user-attachments/assets/f44f7e50-80bb-485b-9c6d-02a59b345119)

image

This does not happen on staging.

kr-matthews:
> This does not happen on staging.

Hm but it does happen when I run `main` locally. :thinking:

gregorbg:
> Hm but it does happen when I run `main` locally. 🤔

Does it happen on `main` as well? It kinda feels unrelated to the SemUI `Ref` changes. Since the "drop zone" itself works, that means the `Ref` is correctly wired up and everything after that should be an existing error that would also occur on v2 on `main`.

gregorbg:
Found an issue with `Sticky` that is not trivial to fix. Reverting to Draft for the time being.

kr-matthews:
> > Hm but it does happen when I run `main` locally. 🤔
> 
> Does it happen on `main` with V2 as well?

I'm not sure what you mean. I'm seeing this on `main` (v2) and on your branch (v3 beta).
But yes, it sounds like it's not actually related to these changes.

PR #10866, opened 2025-02-17
Fix error from large partial refunds

dunkOnIT:
## Overview 

Currently, issuing a partial refund which exceeds the remainder value after the refund[1] causes an error that blocks rendering of the Refund page until refresh.

This is because the refund amount is not being updated with the refetch of the new refund data - so the new `max` value is less than the original refund value, causing an error.

```javascript
<AutonumericField
  currency={refund.currency_code.toUpperCase()}
  value={amountToRefund}
  onChange={setAmountToRefund}
  max={refund.ruby_amount_refundable}
/>
```

[1] ie, Refund amount $12 on $22 total leaves $10 remaining afterwards - $12 > $10, causing the error

## Trial and Error

First, I tried a useEffect to update the `amountToRefund` with a useEffect as follows: 
```javascript
useEffect(() => {
    setAmountToRefund(refund.ruby_amount_refundable);
  }, [refund.ruby_amount_refundable, setAmountToRefund]);
```

That hasn't worked. I'm now experimenting with "watching" other values in the useEffect to see if that might do anything

### DIFF
--- a/app/webpacker/components/RegistrationsV2/RegistrationEdit/Refunds.jsx
+++ b/app/webpacker/components/RegistrationsV2/RegistrationEdit/Refunds.jsx
@@ -1,5 +1,5 @@
 import { useMutation, useQuery } from '@tanstack/react-query';
-import React from 'react';
+import React, { useEffect } from 'react';
 import {
   Button, Header, Message, Table,
 } from 'semantic-ui-react';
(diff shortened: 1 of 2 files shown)


gregorbg:
Please no `useEffect` 😇
Without looking at the code in-depth, my instinctive solution is that you should have a `useState` which represents the refundable amount and the endpoint to handle the refund returns a "amount that is still available to refund after the currently requested refund was processed" from the backend. This return value can be accessed in the mutation's `onSuccess` hook and written to the state.

dunkOnIT:
Can't approve as it's my own PR, but tested and happy with the changes!

kr-matthews:
What about setting the value to 0 when the mutation is triggered?
And then maybe restoring it if the mutation fails.

gregorbg (approved):

PR #10867, opened 2025-02-17
"Wca" -> "WCA ID" in WRT contact form

pranav-027:


app/models/contact_wrt.rb
```diff
@@ -13,4 +13,9 @@ def to_email
   def subject
     Time.now.strftime("[WCA Website] Results Team Comment by #{name} on %d %b %Y at %R")
   end
+
+  def self.human_attribute_name(attr, options = {})
```
gregorbg, line 17: Definitely don't override Rails-internal methods unless you really, _really_ know what you're doing!
The documentation for `human_attribute_name` will reveal to you that the implementation of this method actually accesses localized strings in `en.yml`.

The correct solution here is to leave all Ruby source code in peace and just add a new translation key with the desired value.

gregorbg (changes requested):

gregorbg (approved):

PR #10870, opened 2025-02-18
Fix selected email address list

kr-matthews:
An oversight when the rejected state was added. I didn't see any other analogous omissions while briefly checking around.


gregorbg (approved):
Thanks!

PR #10871, opened 2025-02-18
Refactor messages

kr-matthews:
~**I've marked this as a draft, and need input from people who are more familiar with the back-end/error messages.**~

The main work here is refactoring the message reducer/UI to be more general: when a new message is shown it's added to the existing messages instead of replacing them, positive messages are cleared after 4 seconds (regardless of new messages appearing), and messages are dismissable.

~Using this, in the registration admin lists I added the user's name to the error message. I'm not very confident in this part as I changed an error message - I don't know if that error message might show up somewhere else and would no longer make sense, and I don't know if/which other messages might show up here and will also need adjusting. But hopefully this is the right direction to go.~

Edit: Note that all of the error messages are written as if being shown directly to a competitor registering - e.g. use "you" - but many also appear when an organizer/delegate is handling registrations, in which case this doesn't make sense. Hopefully this can also be cleared up in this PR.

If modifying the error messages to include the name is not appropriate, then I could instead modify the message reducer to consume a string instead of an id, and add the name in manually (i.e. `'${I18n.t(errorMessage)} (${userName})'` or something.

Edit: Done. ~Edit: If this is going to take time, I could remove showing the user name/string edits, and just make this PR be the message logic refactoring.~

[Screencast from 2025-02-17 09:48:02 PM.webm](https://github.com/user-attachments/assets/c31474ea-cad1-4839-84f2-bcadab483ce8)

### DIFF
--- a/app/webpacker/components/RegistrationsV2/Register/RegistrationMessage.jsx
+++ b/app/webpacker/components/RegistrationsV2/Register/RegistrationMessage.jsx
@@ -3,6 +3,10 @@ import React, { useEffect } from 'react';
 import { useDispatch, useStore } from '../../../lib/providers/StoreProvider';
 import I18n from '../../../lib/i18n';
 
+/**
+ * You may pass an array of keys - if so, also pass an
+ * array of params of the same length.
+ */
 export const setMessage = (key, type, params) => ({
   payload: {
     key,
@@ -33,12 +37,13 @@ export default function RegistrationMessage() {
   if (!message?.key) return null;
 
   if (Array.isArray(message.key)) {
-    return message.key.map((key) => (
+    return message.key.map((key, index) => (
       <Message
+        key={key}
         positive={message.type === 'positive'}
         negative={message.type === 'negative'}
       >
-        {I18n.t(key, message.params)}
+        {I18n.t(key, (message.params ?? [])[index] ?? {})}
       </Message>
     ));
   }
--- a/app/webpacker/components/RegistrationsV2/Register/RegistrationMessage.jsx
+++ b/app/webpacker/components/RegistrationsV2/Register/RegistrationMessage.jsx
@@ -41,7 +41,7 @@ export default function RegistrationMessage() {
       style={{ margin: messages.length === 1 ? 0 : undefined }}
       positive={type === 'positive'}
       negative={type === 'negative'}
-      onDismiss={() => dispatch(clearMessage(id))}
+      onDismiss={type === 'negative' && (() => dispatch(clearMessage(id)))}
     >
       {I18n.t(key, params)}
     </Message>
--- a/app/webpacker/components/RegistrationsV2/Register/RegistrationMessage.jsx
+++ b/app/webpacker/components/RegistrationsV2/Register/RegistrationMessage.jsx
@@ -12,7 +12,7 @@ export const showMessages = (messages) => ({ newMessages: messages });
 
 export const clearAllMessages = () => ({ messages: [] });
 
-// const clearMessage = (id) => ({ toClear: [id] });
+const clearMessage = (id) => ({ toClear: [id] });
 
 const clearMessages = (ids) => ({ toClear: ids });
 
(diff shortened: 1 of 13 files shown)


app/webpacker/components/RegistrationsV2/reducers/messageReducer.jsx
```diff
@@ -1,6 +1,37 @@
-const messageReducer = (state, { payload }) => ({
-  ...state,
-  message: { key: payload.key, type: payload.type, params: payload.params },
-});
+const messageReducer = (state, { messages, newMessages, toClear }) => {
```
gregorbg, line 1: Where did the `payload` go? I was always under the assumption that this is a React convention (or even hard requirement?!) for the `useReducer` hook

config/locales/en.yml
gregorbg, line 1: I'd rather not touch registration messages in this PR. More explanation in my general comment.

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationList.jsx
```diff
@@ -87,6 +87,12 @@ const partitionRegistrations = (registrations) => registrations.reduce(
   },
 );
 
+const getUserName = (registrations, userId) => {
+  const registration = registrations.find(({ user_id: id }) => id === userId);
+  if (!registration) return null;
+  return registration.user.name;
```
gregorbg, line 93: What about `return registration?.user?.name`? That should also return `null` in case of doubt.

app/webpacker/components/RegistrationsV2/Register/RegistrationMessage.jsx
```diff
@@ -3,53 +3,47 @@ import React, { useEffect } from 'react';
...
+  }) => (
     <Message
-      style={{ margin: 0 }}
-      positive={message.type === 'positive'}
-      negative={message.type === 'negative'}
+      key={id}
+      style={{ margin: messages.length === 1 ? 0 : undefined }}
+      positive={type === 'positive'}
+      negative={type === 'negative'}
+      onDismiss={() => dispatch(clearMessage(id))}
```
gregorbg, line 44: Does passing an `onDismiss` automatically give the message a small "cross" icon to clear it?
And would it make sense to make positive messages non-dismissable since they clear after 4 seconds anyways?

gregorbg (commented):
Yay!

About changing the error messages: I think this is best for a follow-up PR because of translations. Every string we touch needs to be re-translated, and I'd rather be careful about that because the registration panel is such a "critical" part of our system

app/webpacker/components/RegistrationsV2/Register/RegistrationMessage.jsx
kr-matthews: Yes.
I thought about it originally. I've done it now.

gregorbg (approved):

app/webpacker/components/RegistrationsV2/reducers/messageReducer.jsx
gregorbg: Huh. Interesting! I am very open to these ideas, although I do want to point out that the `dispatch` originates from React itself: `useReducer` [returns a store and a dispatch function](https://react.dev/reference/react/useReducer) by default.

kr-matthews: Yes, we can't avoid a reducer, but my point is to avoid exposing it. Right now, we have a generic `StoreProvider.js` which we use every time we distribute state via a context, and I would rather have a dedicated provider for each use-case. (Not a big deal though, especially since we already have the latter established. More of a side note than a request to change.)

Since it's generic, the best we can do is provide `dispatch` to the components that use them. If we had a dedicated provider, then we could internally use that dispatch function to create the specific actions available and provide those. This is exactly what my recent `useOrderedSet` hook does (the only difference is that we're not distributing that to components via a context, it's just being used in specific components when necessary).

In this case, instead of `export const showMessages = (messages) => ({ newMessages: messages });` (top-level export) and the generic `export const useDispatch = () => useContext(StoreContext)[1];` (return value of hook), we would define those internally and provide them in the hook return just like `useOrderedSet`: `const clear = useCallback(() => dispatch({ type: 'clear' }), []);` (defined inside the hook) and `return { asArray: array, size, has, clear, update, add, remove, toggle };` (return value of hook).

PR #10873, opened 2025-02-18
Add `registered_at` timestamp

dunkOnIT:
## Notes: 

Everything here seems to work, the main question I have is when to populate `registered_at` - I see two options: 

1. Time at which the reg request was submitted?
  - _Just because someone submits a request doesn't mean they've registered_
2. Current time at the point of creating the registration?
  - _This is what I intuitively want to use, but then: Why are we not using created_at with an increased precision value?_

Until we have an answer to this, I'm using a very direct approach - getting the db to populate `registered_at` when the row is created. I don't love this approach, because it isn't exposed in the code/editable later - but something like an `after_create` hook feels very redundant (given that we have `created_at`) with my current understanding, so I just went with something simple while we discuss the implementation.

## TODO

- [x] Add & populate migration
- [x] Populate `registered_at` when reg is created
- [x] Display `registered_at` in the frontend


db/migrate/20250218070141_add_registered_at_timestamp.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+class AddRegisteredAtTimestamp < ActiveRecord::Migration[7.2]
+  def up
+    add_column :registrations, :registered_at, :datetime, precision: 6, default: -> { 'NOW(6)' }
```
gregorbg, line 5: What happens when you just drop the `precision` key altogether?

gregorbg, line 5: Also, not happy about a low-level default value :/ Is there really no other way?

dunkOnIT: My understanding is that's equivalent to `precision: nil` - which means it defaults to MySQL's default, which is precision: 0 

[MySQL documentation](https://dev.mysql.com/doc/refman/8.4/en/date-and-time-type-syntax.html) supports this

image

gregorbg: Okay, convinced

gregorbg: Still curious though about the default value. I'd rather mark this as non-null and backfill accordingly.
(I think if you create the whole column as non-null, it will get the Unix epoch or some other silly default, but then you backfill with actual contents directly below, so we're good)

dunkOnIT: Sorry just didn't get around to changing this! That was just a placeholder to populate the rows with a timestamp while we decided how to populate it. Will be replaced by an explicitly-defined registered_at timestamp per our Slack discussin

dunkOnIT: That should be better now :)

gregorbg (approved):

PR #10874, opened 2025-02-18
allow mailto in comp tabs

pranav-027:
This was reported on discord today: https://discord.com/channels/1121050787074805830/1121171516105699408/1339851598699954178


gregorbg:
Please avoid Discord links. It is not an official WCA channel, and not all WST members (or even wider WCA Staff) are registered there.

PR #10880, opened 2025-02-18
Admin Edit Registration Improvements

dunkOnIT:
Some minor improvements, including:
- Header:
  - use the editPersonUrl link for the edit link
  - only show the ({wca_id}) if the user has a wca_id
- change the update confirmation text to be about updating the user's registration
- only show 'Registration is updating...' if the user submits the changes

Asked Discord users re the order of header elements - poll after an hour: 
image

(There was a previous poll for Edit/competitor name/(WCA ID) vs (WCA ID)/competitor name/edit, but the former won that handily


gregorbg:
I feel like at least one of the two things you were trying here has been fixed through other PRs in the meantime, how relevant is this PR still?

dunkOnIT:
Yes, header has already been changed in a separate PR - this PR still contains 
- an important fix to the editPersonUrl (current it just points to personUrl)
- minor text improvements

Which are unique and worth merging

gregorbg:
> * an important fix to the editPersonUrl (current it just points to personUrl)

I do not see that change in the current diff.

dunkOnIT:
Oops you're right, that was in a different PR. In that case, just minor text fixes.

gregorbg (approved):

PR #10885, opened 2025-02-19
Add option to copy new value to form in edit profile workbench

maxidragon:


tmp/.keep
gregorbg, line 1: Please don't remove files out of nowhere

bin/setup
```diff
@@ -28,7 +28,7 @@ FileUtils.chdir APP_ROOT do
   # end
 
   puts "\n== Preparing database =="
-  system! 'bin/rails db:prepare'
+  #system! 'bin/rails db:prepare'
```
gregorbg, line 31: Unrelated change

app/webpacker/components/Panel/pages/EditPersonPage/EditPersonForm.jsx
```diff
@@ -85,6 +90,12 @@ export default function EditPersonForm({ wcaId, onSuccess, showDestroyButton = f
     }, { method: 'PUT' }, (error) => setResponse({ success: false, message: `${error}` }));
   };
 
+  useEffect(() => {
+    if (defaultValues) {
+      setEditedUserDetails((prev) => ({ ...prev, ...defaultValues }));
+    }
+  }, [defaultValues]);
```
gregorbg, line 97: This is not what useEffect was designed for. Is there really no other way?

gregorbg (commented):
Here are some initial thoughts. Also, please try to open PRs with at least a very basic description of what's happening (as long as the diff is more than ~3 lines).
Screenshots or videos would also be appreciated wherever it makes sense (for example, changing the text of a button does not need a screenshot, but your current PR probably does.)

tmp/.keep
maxidragon: My bad, restored

bin/setup
maxidragon: My bad, fixed

maxidragon:
> Here are some initial thoughts. Also, please try to open PRs with at least a very basic description of what's happening (as long as the diff is more than ~3 lines).
> Screenshots or videos would also be appreciated wherever it makes sense (for example, changing the text of a button does not need a screenshot, but your current PR probably does.)

Edited the description, I've uploaded the video

app/webpacker/components/Panel/pages/EditPersonPage/EditPersonForm.jsx
danieljames-dj: After thinking again, I got one more solution - for this I would like to rename `defaultValues` to `forcedValues`. The `forcedValues` will initially be `{}`.

When a field is populated, it will become like `{name: "John Doe"}`.

Now in `<Form.Input.../>`, we can have the value as `{forcedValues?.name || <whatever is already there>}` and `disabled` will be `{forcedValues?.name || <whatever is already there>}`. This way, if there is a forced value, that will be having highest priority and user won't be able to edit it as well.

While calling the API, instead of sending `person` as `editedUserDetails`, we might have to send `{...editedUserDetails, ...forcedValues}`.

This might be complicated, I'm not sure. @gregorbg WDYT?

maxidragon: > One way I can think of where we can avoid `useEffect` is - pass the `editedUserDetails` and `setEditedUserDetails` from the parent component. But since this form is used in two different places, I'm bit reluctant to use that way and will be more inclined towards using `useEffect`.

I've also considered this but I think this would make the component much more complicated and harder to use in other places, so I agree with you.

I have no strong preference regarding your second comment, both solutions would probably work but the second one looks more complicated than mine. Let's wait for Gregor's opinion, if you guys think that that would be better, I'm more than happy to implement it in that way.

gregorbg:
Okay so here's the thing: If you desperately need this and you say WRT is suffocating under the workload unless we get this PR merged in the next 48 hours, then I'll review it in its current state.

Otherwise: You need to make clear that `EditPersonForm` is actually doing two things right now:
1. Fetching person data, converting that backend information into a form state
2. Rendering an actual form (that is, the inputs that the user can interact with) based on the state derived in (1).

The clean solution is to split these two aspects into two separate components. And then the "inner" component derived from point (2) gets passed the current form state and the `updateFormValues` callback as props. This will allow you to control the form state from the outside, and will let the buttons here directly manipulate the form state as well. No need for `useEffect` hooks anymore.

Does that idea make sense or would you like some more in-depth explanation?

maxidragon:
This is definitely not urgent and can be implemented as you suggested. I'll look into it later and let you know if I need a detailed explanation.

maxidragon:
This PR is no longer needed after changes made by Daniel in #12363

PR #10887, opened 2025-02-20
Registration admin color-/icon-coded sections/actions with dropdown

kr-matthews:
# Edit: See new updates below.

My intention was to change the yellow color (since contrast isn't good enough with white text) and improve the button UI on mobile, but then this happened. I'll improve the buttons next.

- Reorder action buttons to the same order as the tables.
- Adjust the colors (and some icons) of these buttons.
- Add colors to the tables & section headers too, to make it easier to find sections at a glance, and to easily see which table you're in mid-table.
- Disable the missing button instead of hiding it, when only 1 registration is selected.

These are the available colors. I didn't use brown because the light version (which appears in the left column of the table) looks the same as the light version of orange.

image

If the section headers are too much, then just coloring the text is another option. Or keeping them black, but then you can't see the color at all when collapsed.

alt-header

It's also possible to color the entire table, but I think that's way too much:

alt-table

Preview of next PR - grouping action buttons into a dropdown:

next


gregorbg:
I love it! As far as my personal taste goes, I think coloring the headers is a bit too much. But I definitely want more opinions on this change, as this is not a matter that I have "jurisdiction" over. (But I do agree that fully coloring the tables is _way_ too much :joy: )

kr-matthews:
If the problem with yellow is that you can't read text on it, then just don't put text on it.

Updated to make color less prominent, use icons/colors/names consistently, avoid button/scroll jumping on mobile, show selected count when moving.

Still room for improvement - it would be nice if the actions fit into a single line on mobile, maybe by just have 3 icons for export, email, and move, and then the actual options show up as dropdown options just like move current does. And the success message overlaps the buttons awkwardly.

image
image
image
image

[Screencast from 2025-02-23 09:50:40 AM.webm](https://github.com/user-attachments/assets/48a51a2d-ec45-4baf-b02c-391eabf969ca)

For reference, currently it's very easy to accidentally move 1 competitor because the buttons suddenly appear just as you try to tap a second competitor.

[Screencast from 2025-02-23 10:00:17 AM.webm](https://github.com/user-attachments/assets/07a4700d-eca2-4c1f-aaa3-b8393aad9fe9)

kr-matthews:
Oh and this changes a bunch of translation keys/strings. They don't seem to be very well organized overall :frowning_face:.

kr-matthews:
Old strings:
image
image


New strings:
image
image

Open to other suggestions.

PR #10893, opened 2025-02-24
Extract region selector

kr-matthews:
There should be no visible changes to the site.

I can't see the "x" to clear the input, though clicking where it should be works. I feel like we had the same (or at least a very similar) issue a while ago and Gregor fixed it?

Turns out this is actually using Semantic UI for flags, so it has the other style that we got rid of (with the broken Finland flag). I'll see later if that's an easy fix.

Edit: We also have `CountrySelector`, which is obviously very similar. I will combine that with this component too in a follow-up PR, probably by adding a `countriesOnly` or `hideContinents` prop.


gregorbg:
The region selector issue was already handled here: https://github.com/thewca/worldcubeassociation.org/pull/9816#issuecomment-2348038587

From that comment, it feels like the solution we found there should also work here. Let me spin up your PR and find out why it doesn't...

gregorbg:
Apparently, in that other PR, I had only touched the font styles for the RTL-mode of the Dropdown (must have been an oversight/sloppiness on my part, although I have no clue why it worked back then...?)

I have now ported over the font style for the "normal" (ie Left-to-Right) Dropdown as well, which should fix the icon issue.

gregorbg (approved):

Issue #10543, opened 2025-01-06
Register Page Bugs/Improvements


dunkOnIT:
## Medium
- [ ] "Show the user info about their payment in the Payment and View Registration screens:
  - CASE - REG FEE ONLY: You have paid ${amount} of the ${reg_fee} registration fee
  - CASE - FEE + DONATION: You have paid a total of ${total_amount}. This covers the registration fee of ${reg_fee}, and a donation of ${donation_amount}."
- [ ] Quite a few people are getting the "Lots of registrations in queue" message when there clearly aren't a lot of registrations in the queue

## Low
- [ ] After registration has closed, shows "Registration will open in -25 minutes -18 seconds."
- [ ] When a cancelled user is re-registering, prefill the fields (events, guests, comments) with the values from their cancelled registration
- [ ] "Payment Information" doesn't render for a user who has paid, cancelled and then clicks on ""Payment information"" before submitting a new registration. Console error:
```bash
TypeError: C[w] is undefined
    StepPanel.jsx:111"
```
- [ ] If a payment is pending, inform the user of this on the Payment step
  - Not quite sure what I meant when I first wrote this - maybe if the payment's status is pending?

PR #10905, opened 2025-02-24
Enforce a reasonable guest count in registration backend

dunkOnIT:
This aims to prevent the issues we've been having of guest counts larger than a 32-bit integer breaking our registration queue. 

Some implementation notes: 
- We could limit this to whatever our max 32-bit integer size is - but it seems better to prevent people from trolling to that degree in the first place
- `DEFAULT_GUEST_LIMIT` could be stored on the `Registration` model instead of in the `registration_checker`
- Guest limit is largely arbitrary and doesn't necessarily align with frontend validations


gregorbg:
Please gather some (semi-representative) feedback from organizers what a reasonable upper limit would be.

You don't have to go to great lengths mapping out which regions and whom to reach out, mail ROs with a 14 day deadline, etc... Just a plain survey with 10 or so answers should be enough to give us a feeling of what's needed.

lib/registrations/registration_checker.rb
```diff
@@ -3,6 +3,8 @@
 module Registrations
   class RegistrationChecker
     COMMENT_CHARACTER_LIMIT = 240
+    DEFAULT_GUEST_LIMIT = 1000
```
gregorbg, line 6: You can expose this value to the frontend via `app/webpacker/lib/wca-data.js.erb`. That way the frontend form can already perform a `max` check on the number input

dunkOnIT:
I will note that this is a bit downstream of the broader questions we've had about guests, and the fact that it's strange that we're displaying the guest input _at all_ if no guest limit is being enforced. 

But I think it still makes sense to make these changes to hotfix the issue, as the deeper redesign of guests has tripped us up in the past

gregorbg:
> the fact that it's strange that we're displaying the guest input _at all_ if no guest limit is being enforced.

Again, this is a misunderstanding of `guests_enabled`. The naming is confusing, I admit that. But we have to distinguish between:
- The organizer wants to know how many guests people are bringing (without enforcing any limit, so that for example they can decide how many chairs to set up the morning of the competition)
- The organizer wants to know how many guests people are bringing **so that they can enforce a guest limit**
- The organizer genuinely doesn't care at all about how many guests people are bringing. They don't even want to bother asking for the number input.

To a certain degree, organizers are to blame (if they enable "Ask about guests", then that is what they get). We have to take the blame for not modeling the current boolean choices intuitively enough, and perhaps we should come up with better names. This would also help us decide whether we need to introduce an extra column to cover all the cases that I outlined above.

app/webpacker/lib/wca-data.js.erb
```diff
@@ -193,6 +193,12 @@ export const groupedChampionshipTypes = {
   national: _.map(_.sortBy(countries.real, 'name'), 'iso2'),
 }
 
+// ----- REGISTRATIONS  -----
+
+export const commentCharacterLimit = '<%= Registrations::RegistrationChecker::COMMENT_CHARACTER_LIMIT  %>'
```
gregorbg, line 198: Unrelated change, methinks

gregorbg (changes requested):

app/models/registration.rb
```diff
@@ -43,7 +43,6 @@ class Registration < ApplicationRecord
   validates_numericality_of :guests, less_than_or_equal_to: :guest_limit, if: :check_guest_limit?
 
   after_save :mark_registration_processing_as_done
-
```
FinnIckler, line 46: add back white space

lib/registrations/registration_checker.rb
```diff
@@ -129,6 +131,9 @@ def validate_qualifications!(request, competition, target_user)
       def validate_guests!(guests, competition)
         raise WcaExceptions::RegistrationError.new(:unprocessable_entity, ErrorCodes::INVALID_REQUEST_DATA) if guests < 0
         raise WcaExceptions::RegistrationError.new(:unprocessable_entity, ErrorCodes::GUEST_LIMIT_EXCEEDED) if competition.guest_limit_exceeded?(guests)
+        byebug
```
FinnIckler, line 134: accidental debug statement

app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
```diff
@@ -299,6 +299,9 @@ export default function CompetingStep({
   ]);
 
   const guestsRestricted = competitionInfo.guest_entry_status === 'restricted';
+  const guestLimit = competitionInfo.guests_per_registration_limit != null && guestsRestricted
```
gregorbg, line 302: You're in JS so this should probably be `!== null`

gregorbg (changes requested):

app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
dunkOnIT: Fixed!

lib/registrations/registration_checker.rb
dunkOnIT: Removed

app/models/registration.rb
dunkOnIT: Done

app/webpacker/components/RegistrationsV2/api/helper/error_codes.js
```diff
@@ -23,5 +23,6 @@ export const ORGANIZER_MUST_CANCEL_REGISTRATION = -4009;
 export const INVALID_WAITING_LIST_POSITION = -4010;
 export const QUALIFICATION_NOT_MET = -4012;
 export const REGISTRATION_ALREADY_EXISTS = -4014;
+export const REASONABLE_GUEST_COUNT = -4015;
```
kr-matthews, line 26:
```suggestion
export const UNREASONABLE_GUEST_COUNT = -4015;
```
?

app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
```diff
@@ -397,8 +400,8 @@ export default function CompetingStep({
                   setGuests(Number.parseInt(data.value, 10));
                 }}
                 min="0"
-                max={guestsRestricted && (competitionInfo.guests_per_registration_limit ?? 99)}
-                error={guestsRestricted && Number.isInteger(competitionInfo.guests_per_registration_limit) && guests > competitionInfo.guests_per_registration_limit && I18n.t('competitions.competition_info.guest_limit', { count: competitionInfo.guests_per_registration_limit })}
+                max={guestLimit}
+                error={guestsRestricted && guests > guestLimit && I18n.t('competitions.competition_info.guest_limit', { count: guestLimit })}
```
kr-matthews, line 404: Don't we want to show a (different) error message when guests are not restricted? Might as well handle it in the front end since we have the limit (and are using it as `max`).

config/locales/en.yml
```diff
@@ -1926,6 +1926,7 @@ en:
         -4011: "You can only accept the first person on the waiting list"
         -4012: "You are not qualified to compete in the selected events"
         -4014: "You are already registered for this competition."
+        -4015: "Please enter a reasonable amount of guests, or contact the organizers if you need to bring an extraordinary amount of guests."
```
kr-matthews, line 1929: This message should make it clear that the competitor should register with 9 guests now, and also email the organizers. They might miss a spot if they don't register now, and even if not, the organizers can't change the site to allow them to register with >9 guests.

app/webpacker/components/RegistrationsV2/api/helper/error_codes.js
dunkOnIT: Good suggestion, thanks!

app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
dunkOnIT: Gregor and I decided to increase the limit to 99, for the purposes of getting this (potentially production-breaking) problem fixed - there is definitely more we can do here to improve QoL for organizers and competitors - I've updated #10543 with a low-priority item that covers thinking more deeply about handling this well

config/locales/en.yml
dunkOnIT: See other comment

gregorbg (approved):

PR #10918, opened 2025-02-25
Move Guest Registration Check to Validations

### LINKED ISSUE #10864: Refactor `registration_checker` to make more use of Registration model & validations

FinnIckler:
Just doing guests for now as an example how we should convert the others


FinnIckler:
The last failing tests are actually old registration tests that fail, which makes sense because it expects different errors. There is also some tests where the competition is nil.

lib/registrations/registration_checker.rb
```diff
@@ -129,10 +129,18 @@ def validate_qualifications!(request, competition, target_user)
       end
 
       def validate_guests!(guests, competition)
-        raise WcaExceptions::RegistrationError.new(:unprocessable_entity, ErrorCodes::INVALID_REQUEST_DATA) if guests < 0
-        raise WcaExceptions::RegistrationError.new(:unprocessable_entity, ErrorCodes::GUEST_LIMIT_EXCEEDED) if competition.guest_limit_exceeded?(guests)
-        raise WcaExceptions::RegistrationError.new(:unprocessable_entity, ErrorCodes::UNREASONABLE_GUEST_COUNT) if
-          guests > DEFAULT_GUEST_LIMIT && !competition.guest_entry_status_restricted?
+        r = Registration.new(guests: guests, competition: competition)
```
gregorbg, line 132: The target idea here is that we do **not** have to create an "on-the-fly" registration in every `validate_*` method. Rather, we should create the registration exactly as the user/request wishes but without persisting it to the DB, and then "pass it around" before calling the final `save`.

In other words, in an ideal world this `r` Registration already exists and can be passed into the method. Right now, it's okay to create on the fly because this PR is more of a proof-of-concept.

app/models/registration.rb
```diff
@@ -48,8 +48,9 @@ class Registration < ApplicationRecord
   end
 
   validates :guests, numericality: { greater_than_or_equal_to: 0 }
-
-  validates :guests, numericality: { less_than_or_equal_to: :guest_limit, if: :check_guest_limit? }
+  validates :guests, numericality: { less_than_or_equal_to: :guest_limit, if: :check_guest_limit?, frontend_code: Registrations::ErrorCodes::GUEST_LIMIT_EXCEEDED }
+  validates :guests, numericality: { equal_to: 0, unless: :guests_allowed?, frontend_code: Registrations::ErrorCodes::GUEST_LIMIT_EXCEEDED }
+  validates :guests, numericality: { less_than_or_equal_to: Registrations::RegistrationChecker::DEFAULT_GUEST_LIMIT, if: :guests_unrestricted?, frontend_code: Registrations::ErrorCodes::UNREASONABLE_GUEST_COUNT }
```
gregorbg, line 53: We should probably work more with comments explaining the "business logic" idea behind each validation. But in principle, this is how you would pass information around.

lib/registrations/registration_checker.rb
```diff
@@ -129,10 +129,18 @@ def validate_qualifications!(request, competition, target_user)
...
+
+        unless r.valid?
+          guest_error_details = r.errors.details[:guests].first
+
+          if guest_error_details.present?
+            frontend_code = guest_error_details[:frontend_code] || Registrations::ErrorCodes::INVALID_REQUEST_DATA
+
+            # Assumption: If a model validation fails, it should always be HTTP status 422 (unprocessable entity)
```
gregorbg, line 140: Please challenge this assumption at your convenience

gregorbg (commented):
I took a shot at redoing a lot of the manual/custom validation logic. Not sure yet if tests will pass, but putting my thoughts here anyways while CI is running.

As you can see, a lot of checks can be reduced to "primitive" Ruby checks. Even more complex logic like individual guest limits per competition can be relegated to these ActiveRecord checkers, and I think we should do that because that's as standardized as it's going to get. The more we rely on standardized validation logic, the better.

Note for example that I was able to break up the whole `guest_limit_exceeded?` method into custom, atomic validations. This refactoring means that the old return value, which was `guests_not_allowed_but_coming || guests_exceeding_limit` is now broken up into two validations. Before, when the method returned `true`, it would only let you know that _either_ you had guests coming that weren't allowed at all, _or_ your guests were exceeding the limit. But you could not have known which one of these two conditions it was that triggered the `||`.
Now, there are two independent validations for these two cases, and you will know exactly which one failed.

gregorbg:
Interesting: The only test that is failing now is an old one that predates the V2 / React registration system.
It was setting its guests to one million, and expecting that to be valid when guest restrictions are not enforced.

We changed this logic a few weeks ago, to cap registrations at 99 by default. But this logic was only ever implemented in the custom, finger-painted `registration_checker`, and so the old tests that were built for ActiveRecord validations still act based on the assumption that one million is a valid guest count.

My 0.02USD is that we should adjust the expectation of the test.

FinnIckler:
That's what I meant with:

> The last failing tests are actually old registration tests that fail, which makes sense because it expects different errors. There is also some tests where the competition is nil.

gregorbg:
Something along the lines of this should do, but I am open to debate about the base assumption of the test
image

gregorbg (approved):

PR #10919, opened 2025-02-26
Change `nullable` field to actually pass `null` to backend

dunkOnIT:


gregorbg:
I remember you said this causes glitches with the actual value display in the frontend...?

gregorbg (approved):

PR #10921, opened 2025-02-27
Reg admin toggles: 2 per row on mobile

kr-matthews:
I can't figure out how to make it dynamic - 3 would fit in landscape (but not portrait). Before/after:

b1 a1

b2 a2

b3 a3

Alternatively, switch to checkboxes and do 3 per row - I don't like it:
alt

You can also make them `inline`, but then there's 0 padding/margin and they're way too close - but they do take up much less space. (Didn't take a screenshot.)

### DIFF
--- a/app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationList.jsx
+++ b/app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationList.jsx
@@ -501,17 +501,16 @@ export default function RegistrationAdministrationList({ competitionInfo }) {
   return (
     <Segment loading={isMutating} style={{ overflowX: 'scroll' }}>
       <Form>
-        <Form.Group widths="equal">
+        <Form.Group unstackable widths="2">
           {Object.entries(expandableColumns).map(([id, name]) => (
-            <Form.Field key={id}>
-              <Checkbox
-                name={id}
-                label={name}
-                toggle
-                checked={expandedColumns[id]}
-                onChange={() => dispatchColumns({ column: id })}
-              />
-            </Form.Field>
+            <Form.Checkbox
+              key={id}
+              name={id}
+              label={name}
+              toggle
+              checked={expandedColumns[id]}
+              onChange={() => dispatchColumns({ column: id })}
+            />
           ))}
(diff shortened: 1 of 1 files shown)


app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationList.jsx
```diff
@@ -501,17 +501,16 @@ export default function RegistrationAdministrationList({ competitionInfo }) {
   return (
     <Segment loading={isMutating} style={{ overflowX: 'scroll' }}>
       <Form>
-        <Form.Group widths="equal">
+        <Form.Group unstackable widths="2">
```
gregorbg, line 504: SemUI is not very good at responsive forms. Good news: Chakra will handle this one million times better. Bad news: There is no better solution within our current constraints.

One alternative that I can think of is `Form.Group inline widths="equal">`. This will still stack the toggles on mobile, but because of the `inline` tag they're using much less space.

kr-matthews: I think I tried that, but they were literally touching each other, which wasn't enough space to accurate tap one with a thumb. I can try again tomorrow though.

gregorbg: That happens when you do `inline` *and* `unstackable`, because then they are not stacked, and the line breaks are super-duper narrow when using `inline`. If you allow them to stack however (by dropping the `unstackable`) then SemUI adds whitespace.

kr-matthews: Oh, I tried that too. (The `widths` doesn't seem to make any difference.) I think it's a similar amount of space as my current solution? I don't like it as much since there's a big empty space, but if you think it's better then I'll change it.
inline2
inline1

gregorbg: Okay I'll take your solution :D

gregorbg (approved):

PR #10922, opened 2025-02-27
Omit row color if using payment but competitor hasn't paid

kr-matthews:
Not sure how to test this since locally no competitions seem to be using payment. But it _should_ look something like this, when the top few people have paid and the rest haven't:
image
image

Could also color delegates and organizers, since they don't need to pay? Not sure how easy that would be. Probably best to keep it simple.

Open to other suggestions for distinguishing them. The main problem is that on a phone you can't see the checkbox column at the same time as the payment column:
image


kr-matthews:
I forgot to only do this for the pending table. Will fix tomorrow.
I guess it could be applied to all 5 tables, but then you lose the advantage of knowing which table you're in at a glance.

FinnIckler:
You can setup payments locally in the site bar as usual and then just register with one of stripes test credit cards (42424242424242)

kr-matthews:
I can link stripe, but I can't make a payment because my local host doesn't work with https.
Still, I'm reasonably confident that this is correct. Could someone confirm?
PS: you missed a 42 :smile:

gregorbg:
Oh, that's curious. HTTP(S) has nothing to do with local testing overall. Only very specific methods like Apple Pay cannot be tested locally because of missing HTTPS. The panel itself, and other methods like the credit card thing should render just fine even on local.

kr-matthews:
> Oh, that's curious. HTTP(S) has nothing to do with local testing overall. Only very specific methods like Apple Pay cannot be tested locally because of missing HTTPS. The panel itself, and other methods like the credit card thing should render just fine even on local.

Things render fine. It's when I click 'pay now'.

Trying again on chrome, getting a different result from last time. The button is replaced with a loading bar which stays there indefinitely; console:
image

Trying on Firefox:
image

I don't think it's worth the time to pursue this.

gregorbg:
Ah, that's the exact same kind of error as the one we tried to debug before:
The frontend wants JSON but for some reason the backend hard-crashes and renders out an HTML error page, which the frontend then cannot ingest because it still wants JSON.

gregorbg (approved):

gregorbg:
(But I agree it's not worth debugging this for the sake of merging the PR)

PR #10926, opened 2025-02-27
Fix the anonymization script issue for access grants

danieljames-dj:
This PR is to fix the issue where the anonymization script UI breaks as it was trying to access `application` of `access_grants` ([code link](https://github.com/thewca/worldcubeassociation.org/blob/a0ae42f8c3d51baf3da60af46b3a2e5d2fae8efb/app/webpacker/components/Tickets/TicketWorkbenches/AnonymizationTicketWorkbenchForWrt/index.jsx#L185C29-L185C40)).

I am not sure if this fix will work, I couldn't test it in local as I don't have any idea how to populate access grants.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1503,7 +1503,7 @@ class User < ApplicationRecord
 
   def anonymization_checks_with_message_args
     access_grants = oauth_access_grants
-                    .select { |access_grant| !access_grant.revoked_at.nil? }
+                    .where.not(revoked_at: nil)
                     .map do |access_grant|
                       access_grant.as_json(
                         include: {
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1502,7 +1502,22 @@ class User < ApplicationRecord
   end
 
   def anonymization_checks_with_message_args
-    access_grants = oauth_access_grants&.select { |access_grant| !access_grant.revoked_at.nil? }
+    access_grants = oauth_access_grants
+                    .select { |access_grant| !access_grant.revoked_at.nil? }
+                    .map do |access_grant|
+                      access_grant.as_json(
+                        include: {
+                          application: {
+                            only: [:name, :redirect_uri],
+                            include: {
+                              owner: {
+                                only: [:name, :email],
+                              },
+                            },
+                          },
+                        },
+                      )
+                    end
 
     [
       {
@@ -1513,6 +1528,7 @@ class User < ApplicationRecord
       },
       {
         access_grants: access_grants,
+        oauth_applications: oauth_applications,
       },
     ]
   end


danieljames-dj:
@thewca-bot deploy staging

app/models/user.rb
```diff
@@ -1502,7 +1502,22 @@ def location
   end
 
   def anonymization_checks_with_message_args
-    access_grants = oauth_access_grants&.select { |access_grant| !access_grant.revoked_at.nil? }
+    access_grants = oauth_access_grants
+                    .select { |access_grant| !access_grant.revoked_at.nil? }
```
gregorbg, line 1506: Just curious: Why do you limit this to access grants where `revoked_at` is NOT nil?
This means you will only see grants which have `revoked_at` present, i.e. ones which have already been revoked.

gregorbg: On that note, the select can be a tad bit easier to parse if you use `access_grant.revoked_at.present?` because you can skip the `!` in that case.

gregorbg (commented):
Code is fine, just one genuine question that I would like to understand :upside_down_face:

app/models/user.rb
gregorbg: Huh. To be honest I don't remember that I originally gave this comment to Jacob. Was really surprised to see my own reviews there 😅😅

My guess (and I really mean **guess** here!) is that you only need to care about grants which have already been revoked, because the user requesting Anon might not be aware of them anymore.
All currently active grants (ie those where revoked_at is nil) can be taken care of by the user themselves.

With that being said, the approach with filtering for not-nil obviously worked fine for many years. So it might be worth checking on WRT-internal channels if anybody knows anything about this procedure.
If necessary, you can also try reaching out to Jacob in retirement and see if he still remembers.

danieljames-dj: > ... the select can be a tad bit easier to parse ...

I just figured out `oauth_access_grants.where.not(revoked_at: nil)` might also be a good option. I'm not sure if it'll work fine in staging. I'll test in staging once and confirm.


Re decision on whether we need all access grants or just the ones with `revoked_at` value as `nil` or just the ones with `revoked_at` value as not `nil` - I will check with the WRT and if change is required, I'll make it in a later PR. Is it fine to approve this PR as it is now so that it will allow WRT to anonymize in case some request arise? Currently the UI is breaking when `access_grants` is available for a user.

danieljames-dj:
@thewca-bot deploy staging

danieljames-dj:
@thewca-bot deploy staging

danieljames-dj:
@thewca-bot deploy staging

app/models/user.rb
danieljames-dj: > I'm not sure if it'll work fine in staging. I'll test in staging once and confirm.

I've now confirmed that this code is working fine in staging.

gregorbg (approved):

PR #10929, opened 2025-02-28
Added method competition_list_from_string back (hotfix)

danieljames-dj:
I added this method in https://github.com/thewca/worldcubeassociation.org/pull/10732 but accidentally removed in https://github.com/thewca/worldcubeassociation.org/pull/10654 during rebase.

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -382,6 +382,10 @@ class AdminController < ApplicationController
     Competition.includes(associations).find_by_id!(params[:competition_id])
   end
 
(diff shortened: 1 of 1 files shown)


dunkOnIT:
This is also removing a lot of code from admin_controller, not just adding the new method - is that intentional?

danieljames-dj:
> This is also removing a lot of code from admin_controller, not just adding the new method - is that intentional?

Oops not really. I made the change in some other branch, and this caused while switching between the branches. Thanks for pointing it out.

dunkOnIT (approved):

PR #10931, opened 2025-02-28
Reg admin column sorting

kr-matthews:
- Allow sorting by amount paid, and by event (it's not useless, I would have liked this functionality when creating groups for a favourites competition).
- Disable other columns.

Currently we sort registrations, then partition them into tables, then pass the data to tables. One awkward consequence is that all tables share the same sort state. In a follow-up PR, I'll change this to partition -> pass to tables -> sort individually.

### DIFF
--- a/app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationList.jsx
+++ b/app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationList.jsx
@@ -32,6 +32,7 @@ const sortReducer = createSortReducer([
   'country',
   'paid_on_with_registered_on_fallback',
   'registered_on',
+  'amount',
   'events',
   'guests',
   'paid_on',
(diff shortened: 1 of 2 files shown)


app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationList.jsx
```diff
@@ -213,10 +222,29 @@ export default function RegistrationAdministrationList({ competitionInfo }) {
...
+              const aHasEvent = a.competing.event_ids.includes(sortColumn);
+              const bHasEvent = b.competing.event_ids.includes(sortColumn);
+
+              if (aHasEvent && !bHasEvent) {
+                return -1;
+              }
+              if (!aHasEvent && bHasEvent) {
+                return 1;
+              }
+              return 0;
```
gregorbg, line 243: A little bit "hacky" but quite a lot shorter: `Number(bHasEvent) - Number(aHasEvent)`

gregorbg (approved):

PR #10933, opened 2025-02-28
Migrate CAD page and data exports page to react

danieljames-dj:
This PR aims at migrating the Controller for cronjobs, and hence the two pages that use it.

Screenshots:

Screenshot 2025-03-03 at 06 05 53

Screenshot 2025-03-03 at 06 06 37

Screenshot 2025-03-03 at 06 06 54

Screenshot 2025-03-03 at 07 20 56


danieljames-dj:
@thewca-bot deploy staging

danieljames-dj:
@thewca-bot deploy staging

danieljames-dj:
@thewca-bot deploy staging

app/controllers/panel_controller.rb
```diff
@@ -59,4 +59,27 @@ def panel_page
     query_params = request.query_parameters.except(:id)
     redirect_to panel_index_path(panel_id: panel_with_panel_page, anchor: panel_page_id, **query_params)
   end
+
+  private def cronjob_from_params
+    cronjob_class_name = params.require(:cronjob_class_name)
+    Object.const_get(cronjob_class_name)
```
gregorbg, line 65: Again, I am not too happy about `const_get` because of injection. You can take most of the functions here (at least `run` and `reset`, along with a `cronjob_from_name` helper) and move them to a separate `lib` module similar to `ResultsValidators::Utils`

app/jobs/wca_cronjob.rb
```diff
@@ -116,5 +116,21 @@ def reset_error_state!
         last_error_message: nil,
       )
     end
+
+    def serialize
```
gregorbg, line 120: Why do you need this custom serialization logic? `my_cronjob_class.cronjob_statistics` gives you a plain normal AR model, which you can use for plain normal `as_json` and `serializable_hash` shenanigans

gregorbg: (ironically you're including `cronjob_statistics` in your serialization response, whereas most of the other methods like `in_progress`, `scheduled`, `finished` and many others are just `delegate`d to that same `cronjob_statistics` entity.)

app/webpacker/components/Panel/views/CronJobStatus/CronjobActions.jsx
```diff
@@ -0,0 +1,75 @@
...
+  const queryClient = useQueryClient();
+  const {
+    mutate: runCronjobMutation,
+    isLoading: isRunLoading,
+    isError: isRunError,
+  } = useMutation({
+    mutationFn: runCronjob,
+    onSuccess: (updatedCronjobDetails) => queryClient.setQueryData(
+      ['cronjob-details', cronjobClassName],
+      () => updatedCronjobDetails,
```
gregorbg, line 28: In such a simple case, where you can ignore the existing data and you just want to overwrite it completely, you can pass `updatedCronjobDetails` by itself, without wrapping it into an "empty" function.

app/webpacker/components/Panel/views/CronJobStatus/CronjobActions.jsx
```diff
@@ -0,0 +1,75 @@
...
+      <Button
+        disabled={isDoItDisabled}
+        onClick={() => runCronjobMutation({ cronjobClassName })}
+      >
+        Do it!
+      </Button>
+      <Button
+        disabled={showDebugInfo}
+        onClick={() => setShowDebugInfo(true)}
+      >
```
gregorbg, line 65: You can use the `toggle` feature of SemUI here.

app/webpacker/components/Panel/views/CronJobStatus/CronjobActions.jsx
```diff
@@ -0,0 +1,75 @@
...
+      </Button>
+      <Button
+        disabled={isDoItDisabled}
+        onClick={() => runCronjobMutation({ cronjobClassName })}
+      >
+        Do it!
+      </Button>
+      <Button
+        disabled={showDebugInfo}
+        onClick={() => setShowDebugInfo(true)}
```
gregorbg, line 64: Also, as far as I can tell this button only allows you to switch things on, but never to switch things off.
You could consider `(debugWas) => !debugWas` (name tentative, feel free to come up with something better) to turn the button into a true toggle.

app/controllers/panel_controller.rb
gregorbg: During the course of the review, I doscovered that there is already `JobUtils`, so please use that!

app/webpacker/components/Panel/views/CronJobStatus/index.jsx
```diff
@@ -0,0 +1,109 @@
...
+import { DateTime } from 'luxon';
+import _ from 'lodash';
+import I18n from '../../../../lib/i18n';
+import getCronjobDetails from './api/getCronjobDetails';
+import Loading from '../../../Requests/Loading';
+import Errored from '../../../Requests/Errored';
+import CronjobActions from './CronjobActions';
+
+const CRONJOB_MESSAGE_TYPES = ['warning', 'info', 'error', 'success'];
+const CRONJOB_MESSAGE_TYPES_MAP = _.keyBy(CRONJOB_MESSAGE_TYPES, (item) => item);
```
gregorbg, line 15: I think this is taking type safety a bit too far. We already know we're definitely going to use TypeScript in NextJS, at which point this kind of "type-safe map accessing" will become completely redundant.

gregorbg: If you feel very strongly about it, then I won't block the whole PR because of this.
But it just feels like you're creating too much work for such a simple purpose.

app/webpacker/components/Panel/views/CronJobStatus/index.jsx
```diff
@@ -0,0 +1,109 @@
...
+      <Header>{`Controls for ${cronjobClassName}`}</Header>
+      <Button primary onClick={refetch}>Refresh</Button>
+      <Message info>
+        Steps involved in this cronjob:
+        <List bulleted>
+          {I18n.tArray('cronjobs.compute_auxiliary_data.steps').map((step) => (
+            <List.Item key={step}>{step}</List.Item>
+          ))}
+        </List>
+      </Message>
```
gregorbg, line 94: Nitpick: SemUI has `SegmentGroup` which can replicate the old look and feel: https://react.semantic-ui.com/elements/segment/#groups-segments

I don't feel strongly about this (ie your PR won't be blocked because of this) but it does look a bit confusing to have two `Message`s (which could both be `info` depending on the exact case) directly above each other.

config/locales/en.yml
```diff
@@ -2883,3 +2883,17 @@ en:
...
+        '2': "Computes the lookup for assigning record markers."
+        '3': "Deletes cached results (so they can be cached again)."
+    dump_developer_database:
+      steps:
+        '1': "Copies a redacted version of our entire database into an SQL file."
+        '2': "Zips the file and makes the ZIP file available for public download."
+    dump_public_results_database:
+      steps:
+        '1': "Generates the public Results Export in SQL and TSV formats."
+        '2': "Zips the file and makes the ZIP file available for public download."
```
gregorbg, line 2899: Not sure if this interesting for the general public. Usually only WRT sees these pages, and they can all speak English well enough to understand this.

config/routes.rb
```diff
@@ -302,10 +303,7 @@
   get '/admin/fix_results_selector' => 'admin#fix_results_selector', as: :admin_fix_results_ajax
   get '/admin/person_data' => 'admin#person_data'
   get '/admin/do_compute_auxiliary_data' => 'admin#do_compute_auxiliary_data'
```
gregorbg, line 305: Why is this route not being deleted along with the others?

gregorbg (changes requested):

config/routes.rb
danieljames-dj: Because this route is still used for import results page.

gregorbg: Gotcha!

app/webpacker/components/Panel/views/CronJobStatus/CronjobActions.jsx
```diff
@@ -0,0 +1,78 @@
...
+      <Button
+        disabled={isDoItDisabled}
+        onClick={() => runCronjobMutation({ cronjobName })}
+      >
+        Do it!
+      </Button>
+      <Button
+        toggle
+        active={showDebugInfo}
+        onClick={() => setShowDebugInfo(!showDebugInfo)}
```
gregorbg, line 67:
```suggestion
        onClick={() => setShowDebugInfo((debugInfoWas) => !debugInfoWas)}
```

app/webpacker/components/Panel/views/CronJobStatus/index.jsx
gregorbg: Yes, I think that will be fine.

gregorbg: For future reference, please note that `CRONJOB_MESSAGE_TYPES_MAP` is only used within this file. If...
- it holds shared constants that are `export`ed
- or it reads data from the backend (ie currently our `wca-data.js-erb` hack)

then it makes sense to "secure" the data access like that. So your approach is not "forbidden" in general.

But in this particular case, you're using four hand-written literal string constants and then wrapping them in an array with key access. This is not much better than just writing the string literals anyways.

config/locales/en.yml
gregorbg: Yup!

gregorbg (changes requested):

app/webpacker/components/Panel/views/CronJobStatus/CronjobActions.jsx
danieljames-dj: Okay changed.

app/webpacker/components/Panel/views/CronJobStatus/index.jsx
danieljames-dj: Understood, changed it.

config/locales/en.yml
danieljames-dj: Okay done

gregorbg (approved):
Code is fine by me, if you're happy with your recent Staging testing yesterday then feel free to merge

danieljames-dj:
Thanks @gregorbg. I'm fairly confident with the changes, but would like to test in staging once. I couldn't test in staging yesterday because login was not working in staging (I should have informed the team, but I forgot). Is the login issue a known issue? If it takes time, then I'll merge it and test it in production (this is anyway internal and affects only WRT. I can quickly take action if something goes wrong).

gregorbg:
Huh, no. Staging login _should_ work...
CC @FinnIckler

FinnIckler:
It's because the staging database is very empty. Maybe a botched reload?

danieljames-dj:
I'll go ahead and merge for now as I'm fairly confident about the code. I'll make sure to deploy and test immediately.

PR #10936, opened 2025-03-02
Absorb CountrySelector into RegionSelector

kr-matthews:
Little bit awkward as some change handlers are more naturally set up to handle an event than be directly passed the value. I thought about giving the possibility of accepting either type of handler, but we want to intercept 'clear' events and replace the empty string with our chosen default value, so that would work cleanly anyway as far as I can see.

Note that the label's ternary operator default value is actually returning "Region" in both cases, since we currently 'translate' country as "Region". But in theory it's choosing between country and region depending on the dropdown options.

`DuesRedirect` was using the wrong prop name to pass the region value, so it's been broken this whole time.

Next steps: ~Investigate replacing the `flag` use in RegionSelector with our own RegionFlag.~ _Done in this PR._ And move GenderSelector to /wca. (How come /wca isn't called /common?)


app/webpacker/components/wca/RegionSelector.jsx
```diff
@@ -1,54 +1,85 @@
...
+      name={name}
+      search
+      selection
+      clearable={
+        (nullable && region !== null) || (defaultValue && region !== defaultValue)
+      }
+      value={region}
+      options={onlyCountries ? countryOptions : regionsOptions}
+      // clearing calls onChange with the empty string; catch and replace it
+      onChange={(_, data) => onRegionChange(data.value || defaultValue)}
```
gregorbg, line 80: I think for the sake of reusability, you should not ignore the event and instead propagate the "long form" of the setter (the one reacting to `(evt, { value })`'s `value`. Outside code that used to have a "plain" setter can then "wrap" this SemUI-esque function signature when needed.

See also how `useInputState` achieves a similar result -- it effectively returns a setter function that can be called in two different ways.

gregorbg (commented):
I'm still not 100% settled on the reusability approach, but I have a strong instinct towards not ignoring the signature that SemUI provides by default.

In essence, my argument is that SemUI is the "longer" or "more detailed" signature, and it's easier to mask out / hide parts of that longer signature when needed (instead of currently relying on the plain setter everywhere and "inventing" `null` events when the original code was expecting the more complex SemUI signature). I have added a note about how we do this in other places of the codebase, maybe this can help arrive at a more concise solution.

But this isn't a hard directive, I want to arrive at a consensus that everybody is happy with and that will be sustainable to use in the long run

app/webpacker/components/wca/RegionSelector.jsx
kr-matthews: That makes sense :+1:

app/webpacker/components/Users/List/index.jsx
```diff
@@ -40,7 +40,7 @@ function PersonList() {
       <Header>
         Users
       </Header>
-      <RegionSelector region={region} onRegionChange={setRegion} />
+      <RegionSelector region={region} onRegionChange={(e, { value }) => setRegion(value)} />
```
gregorbg, line 43: This can now benefit from `useInputState`

app/webpacker/components/Results/ResultsFilter.jsx
```diff
@@ -60,7 +60,7 @@ export default function ResultsFilter({
         <Form.Field>
           <RegionSelector
             region={regionIso2}
-            onRegionChange={(r) => setRegion(getRegionIdWithFallback(r))}
+            onRegionChange={(e, { value }) => setRegion(getRegionIdWithFallback(value))}
```
gregorbg, line 63: This can... maybe...? benefit from `useInputState`, although I'm not sure what `getRegionIdWithFallback` does. Is it compatible / can it be replaced with the default value of your refactored country selector?

app/webpacker/components/Persons/NewPersonForm/NewPersonForm.jsx
```diff
@@ -85,10 +86,12 @@ function NewPersonForm({
             gender={gender}
             onChange={setGender}
           />
-          <CountrySelector
-            countryIso2={countryIso2}
+          <RegionSelector
+            label={I18n.t('activerecord.attributes.user.country_iso2')}
+            onlyCountries
+            region={countryIso2}
+            onRegionChange={(e, { value }) => setCountryIso2(value)}
```
gregorbg, line 93: This could now benefit from `useInputState`

app/webpacker/components/Persons/NewPersonForm/NewPersonForm.jsx
```diff
@@ -24,7 +25,7 @@ function NewPersonForm({
   const [semiId, setSemiId] = useInputState('');
   const [wcaId, setWcaId] = useInputState('');
   const [name, setName] = useInputState('');
-  const [countryIso2, setCountryIso2] = useInputState('');
+  const [countryIso2, setCountryIso2] = useState('');
```
gregorbg, line 28: See other comment below, changing this to a "plain" state isn't necessary anymore, you can keep it as `useInputState` as it was.

app/webpacker/components/Persons/List/index.jsx
```diff
@@ -41,7 +41,7 @@ function PersonList() {
       <Header>
         {I18n.t('layouts.navigation.persons')}
       </Header>
-      <RegionSelector region={region} onRegionChange={setRegion} />
+      <RegionSelector region={region} onRegionChange={(e, { value }) => setRegion(value)} />
```
gregorbg, line 44: This could now benefit from `useInputState`

app/webpacker/components/Panel/pages/EditPersonPage/EditPersonForm.jsx
```diff
@@ -51,6 +51,16 @@ export default function EditPersonForm({ wcaId, onSuccess, showDestroyButton = f
...
 
+  const handleDobChange = (date) => handleFormChange(null, {
+    name: 'dob',
+    value: date,
+  });
+
+  const handleCountryChange = (e, { value }) => handleFormChange(e, {
+    name: 'representing',
+    value,
+  });
```
gregorbg, line 62: I think this extra callback is redundant now. If you propagate the `name` to the country selector directly as prop, it will be part of the `data` payload (that is now "nicely" propagated since your last commit)

app/webpacker/components/Panel/pages/DuesRedirect/index.jsx
```diff
@@ -17,12 +17,17 @@ export default function DuesRedirect() {
...
+  const [open, setOpen] = useState(false);
+  const [toDeleteId, setToDeleteId] = useState();
+  const [formData, setFormData] = useState({ redirectType: 'Country' });
 
   const handleFormChange = (_, { name, value }) => setFormData({ ...formData, [name]: value });
 
+  const handleRegionChange = (e, { value }) => handleFormChange(e, {
+    name: 'redirectFromCountryIso2',
+    value,
+  });
```
gregorbg, line 29: Again, probably redundant now if you propagate the `name` to the country selector

app/webpacker/components/EditSchedule/EditVenues/VenuePanel.js
```diff
@@ -145,6 +145,11 @@ function VenuePanel({
     }
   };
 
+  const handleRegionChange = (e, { value }) => handleVenueChange(e, {
+    name: 'countryIso2',
+    value,
+  });
```
gregorbg, line 151: Again, this should be redundant if you propagate `name` to the country selector

app/webpacker/components/ContactEditProfilePage/EditProfileForm.jsx
```diff
@@ -80,6 +80,11 @@ export default function EditProfileForm({
     value: date,
   });
 
+  const handleCountryChange = (e, { value }) => handleFormChange(e, {
+    name: 'country_iso2',
+    value,
+  });
```
gregorbg, line 86: Redundant if you propagate `name` to the country selector

gregorbg (changes requested):
Cool, thanks for that change set! Now there are a bunch of opportunities for cleanup all throughout.

Note that I have (completely arbitrarily) reviewed the files from bottom to top, not sure if GitHub shows my comments in the order of the files or in the order of the timestamp when I created them. If it's the former, then some remarks like "Again" or "See below" may not work (or have to be mentally replaced with "See above"). Sorry 😐

app/webpacker/components/Results/ResultsFilter.jsx
kr-matthews: The results page is storing the region as the country name, instead of iso2. `getRegionIdWithFallback` translates from the iso2 to the name. (And also translates continents/world.)

gregorbg (approved):

PR #10938, opened 2025-03-02
Reg Admin warning when skipping waitlisted registrations

kr-matthews:
Initially I showed the name of the first person skipped over, but I think this is better.

It's purely a front-end warning, just click Ok to ignore it.

image

Note: I changed the message slightly after recording the video below - see screenshot above. And the final time, I clicked Yes instead of cancel.

[Screencast from 2025-03-02 02:40:53 PM.webm](https://github.com/user-attachments/assets/f87bc4d1-8346-4e3c-8b9d-cab8d098356f)

I think an analogous warning for moving registrations to the waitlist would make sense? So, if there are unselected completed pending registrations with an earlier payment (or registered, if not using payments) timestamp, then show a warning.

### DIFF
--- a/app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
+++ b/app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
@@ -135,7 +135,7 @@ export default function RegistrationActions({
     if (skippedWaitlistRegistration) {
       const { name } = skippedWaitlistRegistration.user;
       confirm({
-        content: `You've skipped over ${name} on the waitlist. Are you sure you want to approve the other registration(s) instead?`,
+        content: I18n.t('competitions.registration_v2.list.waitlist.skipped_warning', { name }),
       }).then(
         () => changeStatus(idsToAccept, 'accepted'),
       ).catch(() => null);
--- a/app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
+++ b/app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
@@ -8,10 +8,12 @@ import { countries } from '../../../lib/wca-data.js.erb';
 import {
   APPROVED_COLOR, APPROVED_ICON,
   CANCELLED_COLOR, CANCELLED_ICON,
+  getSkippedWaitlistRegistration,
   PENDING_COLOR, PENDING_ICON,
   REJECTED_COLOR, REJECTED_ICON,
   WAITLIST_COLOR, WAITLIST_ICON,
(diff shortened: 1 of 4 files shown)


app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
```diff
@@ -124,7 +127,21 @@ export default function RegistrationActions({
...
+
+    if (skippedWaitlistCount > 0) {
+      confirm({
+        content: I18n.t(
+          'competitions.registration_v2.list.waitlist.skipped_warning',
+          { count: skippedWaitlistCount },
+        ),
+      }).then(
+        () => changeStatus(idsToAccept, 'accepted'),
+      ).catch(() => null);
```
kr-matthews, line 143: I get an error without the catch, here and in every other use of `useConfirm` in the code, but it doesn't happen on the actual website.

config/locales/en.yml
```diff
@@ -1824,6 +1824,7 @@ en:
           information: "Pending should only be used for incomplete (eg, unpaid) registrations - please use the Waiting List if your competition is too full to accept new competitors."
         waitlist:
           title: "Waiting List Registrations"
+          skipped_warning: "You've skipped %{count} registration(s) on the Waiting List. Are you sure you want to approve the other registration(s) instead?"
```
gregorbg, line 1827: Why did you decide against the pluralization feature (separate sub-keys for `one`, `many`, `other`, etc) in this case?

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
gregorbg: Yeah maybe I should look into updating the actual `ConfirmProvider` to fix the underlying issue. Must have something to do with `StrictMode` being enabled/disabled.

config/locales/en.yml
kr-matthews: The second "registration(s)" is independent of the first. It's also not entirely clear how it should work. If you take positions 2, 3, and 8 on the waitlist, does "other registration(s)" refer to just 8, or all 3? You're accepting 8 instead of 1, vs accepting 2, 3, and 8 instead of 1, 2, and 3.
I also saw some other "(s)"s in the file so assumed it was ok.

I can pluralize the first one if you want. And the second if I know how it should work.

kr-matthews: I guess another solution would be to re-phrase the second sentence.

gregorbg: Mh, no hard feelings here. Your argument is strong enough to convince me that we don't need to go the sophisticated route :)

gregorbg (approved):

PR #10940, opened 2025-03-03
Change front-facing occurences of 'Staff' to 'volunteer'

gregorbg:
As per our most recent Motions update.

Note that this does _not_ include code references. These can be handled during a follow-up PR.


dunkOnIT (approved):
LGTM, and double-checked by grepping the repo for `staff` and ensuring there were no non-code "staff" references (with the exception of translations - see next comment)

dunkOnIT:
The Swedish and Danish translations still have 1-2 keys that use "staff" instead of volunteer. Is this something we want to manually override, or leave up to translators?

gregorbg:
Jag pratar inte svenska :sweden: 
Leave that up to the translators, this is consistent with how we've dealt with similar changes in the past

PR #10952, opened 2025-03-03
Reg Admin independent table sorting

kr-matthews:
Before this, clicking on a column header in the waitlist would sort the other 4 tables - but not the waitlist table - by that column.


kr-matthews:
> (lazily assuming that the big blocks of code in `utils/registrationAdmin.js` are just copy/pasted from previously reviewed code)

Well I did replace one `if (registrations) {` with a `?.` and a `?? []`.

PR #10954, opened 2025-03-03
Implement client-based resource access for Stripe API

gregorbg:
As per https://github.com/stripe/stripe-ruby/wiki/Migration-guide-for-v13

I also opted to introduce a test helper for the numerous occasions of `Stripe::PaymentIntent.confirm` in our code.

### DIFF
--- a/app/models/connected_stripe_account.rb
+++ b/app/models/connected_stripe_account.rb
@@ -59,10 +59,7 @@ class ConnectedStripeAccount < ApplicationRecord
 
     # Create the PaymentIntent, overriding the stripe_account for the request
     # by the connected stripe account for the competition.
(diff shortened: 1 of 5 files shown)


gregorbg:
For reviewers: Probably the most notable change is that the `stripe_account` property (which we use to perform actions on behalf of the connected account) is now "baked into" the `Stripe::StripeClient` instance that we're passing around using `stripe_client`. So it is not necessary anymore to manually specify `stripe_account: fooBar` in every single API call.

dunkOnIT (approved):
Code changes look good, and I've done some basic testing on local for paying/refunding/paying again to make sure charges, payments and refunds are all still behaving as expected

PR #10961, opened 2025-03-04
Use 'clamp' option instead of manually clamping AN in useEffect

gregorbg:
This should remedy the "lag" reported by some users when entering currencies very fast.


dunkOnIT:
Not sure if it's the same issue you're referring to, but it also fixes the problem of not being able to type more than 1 digit before the AN field formats/defocuses

dunkOnIT (approved):

gregorbg:
Yes that's exactly the issue I was referring to

PR #10965, opened 2025-03-05
Delegate Next Authorization to WCA

FinnIckler:
This now works


FinnIckler:
Removed all my API Client shenanigans and moved profile scope to optional

gregorbg (approved):

PR #10966, opened 2025-03-06
Reg Admin: more skip warnings

kr-matthews:
I have not tested a competition with payments and some paid registrations - with later registrations having paid and earlier registrations not - since I was having trouble paying locally. But it's using the same sort as the table so I can't see how it would go wrong (famous last words).

[Screencast from 2025-03-05 10:32:35 PM.webm](https://github.com/user-attachments/assets/d6665648-8141-49a4-bded-881e1520d949)

After re-ordering table:

[Screencast from 2025-03-05 10:36:16 PM.webm](https://github.com/user-attachments/assets/2717a538-d961-4ced-9cac-ec9d817815eb)


kr-matthews:
I would like confirmation that we want these skip warnings, before I go through the effort of resolving the conflicts.

gregorbg:
I personally don't have strong feelings about these warnings, but then again I haven't actively organized in a while. Others should chime in, especially CC @dunkOnIT

dunkOnIT:
Warnings are **definitely desireable:** (we had this discussion on Discord a few months ago)
- when a non-waiting-list-leader from Waiting List -> Approved (whether or not payments are enabled)

Warnings are **probably desireable:** (I have posted a poll on Discord to confirm this suspicion)
- when moving a user with a later payment timestamp than another user from Pending -> Approved

Warnings are **very UNdesireable:**
- when moving users from Pending -> Accepted and payments aren't enabled (because the registration timestamp has very little bearing on when payment via 3rd party service might have happened)

As far as I can see, the current PR will yield the `approve_skipped_warning` even for competitions where payments aren't enabled - so I think this behaviour should be changed/exempted. Otherwise, I'm happy in principle and we can wait on Discord feedback to confirm. 

I also have some suggested wording changes I'll add shortly.

config/locales/en.yml
```diff
@@ -1824,6 +1824,8 @@ en:
         pending:
           title: "Pending Registrations"
           information: "Pending should only be used for incomplete (eg, unpaid) registrations - please use the Waiting List if your competition is too full to accept new competitors."
+          waitlist_skipped_warning: "You've skipped %{count} earlier Pending registration(s). Are you sure you want to move the other registration(s) to the Waiting List instead?"
```
dunkOnIT, line 1827:
```suggestion
You've skipped %{count} earlier Pending registration(s). Are you sure you want to move this/these registration(s) to the Waiting List instead?
```

config/locales/en.yml
```diff
@@ -1824,6 +1824,8 @@ en:
...
+          approve_skipped_warning: "You've skipped %{count} earlier Pending registration(s). Are you sure you want to accept the other registration(s) instead?"
```
dunkOnIT, line 1828:
```suggestion
approve_skipped_warning: "You've skipped %{count} earlier Pending registration(s). Are you sure you want to accept this/these registration(s) instead?"
```

kr-matthews:
> Otherwise, I'm happy in principle and we can wait on Discord feedback to confirm.

Any updates?

dunkOnIT:
> Any updates?
Yep, overwhelming support for warning when moving Pending -> Approved
image

There was the point raised that for comps that waive payments in some cases (eg, RO membership) this could get annoying - it would be nice to store a user preference for this warning, but at this stage I'm a little hesitant to go down that route given that we don't have a good user preferences framework at the moment.

dunkOnIT:
Thanks for requesting the review! I missed your earlier comment. I'm happy with the warnings and think they'll definitely be appreciated by organizers.

I see two opportunities for (small) improvements in follow-up PRs: 
1. **Handling on unpaid-pending registrations when payment is enabled:** These registrations give a "there are {n} earlier registrations" warning when trying to approve them - I think instead the warning should be that there are paid registrations which haven't yet been approved. We don't care about the order in which unpaid+pending registrations are accepted, but we do care if paid+pending registrations are out of order
2. **The ability to disable/mute these warnings: ** I'm not even 100% convinced this is a good idea, but in some cases - particularly countries that give free entry to RO members (like France) these warnings will get annoying. We might want a "Don't show this warning again" checkbox that saves that preference in localStorage. Ideally we'd be able to set a preference at the account-level, but we aren't there yet in terms of our backend - so I think localStorage is our best bet in terms of user preferences for the near/medium-term future

dunkOnIT (approved):
Fine by me - @gregorbg shout if you have any concerns re the code

PR #10975, opened 2025-03-06
Added auto_accept to competition model

dunkOnIT:


spec/factories/competitions.rb
```diff
@@ -280,7 +289,7 @@
 
     trait :registration_not_opened do
       registration_open { 1.weeks.from_now.change(usec: 0) }
-      registration_close { 4.weeks.from_now.change(usec: 0) }
+      registration_close { 3.weeks.from_now.change(usec: 0) }
```
gregorbg, line 292: Why?

spec/factories/competitions.rb
```diff
@@ -101,6 +101,15 @@
...
     registration_version { :v3 }
 
+    trait :auto_accept do
+      use_wca_registration { true }
+      auto_accept_registrations { true }
+    end
+
+    trait :allow_self_delete do
+      competitor_can_cancel { :always }
+    end
```
gregorbg, line 111: Where is this seemingly unrelated trait coming from?

gregorbg (changes requested):

spec/factories/competitions.rb
dunkOnIT: Removed - it is only relevant when API tests get added

config/locales/en.yml
```diff
@@ -1944,6 +1944,11 @@ en:
         -6002: "You need to finish your registration before you can pay"
     #context: and when an error occurred
     errors:
+      must_use_wca_registration: "Auto-accept can only be used if you are using the WCA website for registrations"
+      auto_accept_limit: "Limit for auto-accepted registrations must be less than the competitor limit"
+      auto_accept_not_negative: "Limit for auto-accepted registrations cannot be less than 0."
+      auto_accept_accept_paid_pending: "Cant enable auto-accept if there are paid-pending registrations - either accept them or move them to the waiting list"
+      auto_accept_accept_waitlisted: "Cant enable auto-accept - please accept as many users from the Waiting List as possible."
```
gregorbg, line 1951: Nitpick: `Can't`

gregorbg (changes requested):

config/locales/en.yml
gregorbg: (in both keys)

PR #10977, opened 2025-03-07
Auto accept backend implementation

dunkOnIT:


app/models/registration.rb
```diff
@@ -240,12 +240,12 @@ def registration_history
                             end
       end
       {
-        changed_attributes: changed_attributes,
+        changed_attributes: changed_attributes.with_indifferent_access,
```
dunkOnIT, line 243: Added to make accessing the attributes in tests more consistent

app/models/registration_payment.rb
```diff
@@ -9,6 +9,7 @@ class RegistrationPayment < ApplicationRecord
   belongs_to :refunded_registration_payment, class_name: 'RegistrationPayment', optional: true
   has_many :refunding_registration_payments, class_name: 'RegistrationPayment', inverse_of: :refunded_registration_payment, foreign_key: :refunded_registration_payment_id, dependent: :destroy
 
+  after_create :auto_accept_hook
```
dunkOnIT, line 12: Currently I'm performing a check for prod in the `attempt_auto_accept` method - but I could also do it here instead/as well

spec/factories/competitions.rb
```diff
@@ -280,7 +291,7 @@
 
     trait :registration_not_opened do
       registration_open { 1.weeks.from_now.change(usec: 0) }
-      registration_close { 4.weeks.from_now.change(usec: 0) }
+      registration_close { 3.weeks.from_now.change(usec: 0) }
```
dunkOnIT, line 294: This change may be reverted when 10975 is finalized

dunkOnIT:
Tests will fail until #11122 is merged into main

dunkOnIT:
Ok updated this as far as it can go with the current state of the validation migration. Because apply_payload doesn't update the status, there are some tests failing. 

However, this is enough for us to review the implementation in principle. I did also wonder whether it might be better to call `update_registration_allowed` directly until all registration checker checks are migrated across to the model, but not sure if we will be in a situation where we can merge this PR _without_ all validations having been migrated

app/models/registration.rb
```diff
@@ -351,6 +351,17 @@ def self.newcomer_month_eligible_competitors_count
...
+  # functionality exists to use validations to raise errors on submitted registrations
+  validate :does_not_exceed_competitor_limit
+  private def does_not_exceed_competitor_limit
+    return if competition&.competitor_limit.blank?
+    return unless competing_status == Registrations::Helper::STATUS_ACCEPTED
+
+    errors.add(:competitor_limit, I18n.t('registrations.errors.competitor_limit_reached')) if
+      competition.registrations.accepted_and_competing_count >= competition.competitor_limit
+  end
+
```
gregorbg, line 364: There is already `def cannot_exceed_competitor_limit` which does exactly the same, except:
- It only triggers when _changing_ the status to accepted (ie trying to currently accept a registration)
- It stores the error under `competing_status`, not under `competitor_limit`.

I'm quite adamant about that second point, because `competitor_limit` isn't even a field on a registration. About the first point, is that causing any troubles for you? You definitely shouldn't have to duplicate a validation like that. If anything, we should try to re-use the existing one so that it makes sense for you to use it as well

app/models/registration.rb
```diff
@@ -517,4 +528,53 @@ def competition_events_changed?
...
+    update_payload = build_auto_accept_payload
+    if Registrations::RegistrationChecker.apply_payload(self, update_payload).valid?
+      update_lanes!(
+        update_payload,
+        'Auto-accept',
+      )
+      true
+    else
+      log_auto_accept_failure(accept_registration.errors)
```
gregorbg, line 549: What is `accept_registration` and where does it come from?

app/models/competition.rb
```diff
@@ -507,7 +507,9 @@ def registration_full_and_accepted?
   end
 
   def auto_accept_threshold_reached?
-    auto_accept_disable_threshold > 0 && auto_accept_disable_threshold <= registrations.competing_status_accepted.count
+    auto_accept_disable_threshold.present? &&
```
gregorbg, line 510:
```suggestion
    auto_accept_disable_threshold? &&
```

app/models/competition.rb
```diff
@@ -507,7 +507,9 @@ def registration_full_and_accepted?
...
+      auto_accept_disable_threshold > 0 &&
```
gregorbg, line 511:
```suggestion
      auto_accept_disable_threshold.positive? &&
```

app/models/competition.rb
```diff
@@ -507,7 +507,9 @@ def registration_full_and_accepted?
...
+      auto_accept_disable_threshold <= registrations.competing_status_accepted.count
```
gregorbg, line 512:
```suggestion
      auto_accept_disable_threshold <= registrations.accepted_and_competing_count
```

app/models/competition.rb
FinnIckler: We gotta turn on that rubocop

app/models/registration.rb
dunkOnIT: In my defense this was added before `cannot_exceed_competitor_limit` existed

app/models/registration.rb
dunkOnIT: It was left orphaned during a refactor - fixed now, but this `else` block isn't under test until the registration model validates status

app/models/competition.rb
dunkOnIT: Updated this instance, and another 2 instances elsewhere in the file

app/models/competition.rb
dunkOnIT: Changed

dunkOnIT:
Currently blocked until we fix the issues that will the `:disabled` tests to be un-disabled - predominently, normalizing waiting list handling such that 
- we reproduce the current waiting list -> accepted issues
- waiting list reg can be approved

app/models/registration.rb
gregorbg: Not a big fan of this. What does `index_by` above give you by default, and can you use something like `symbolize_keys` to make it consistent?

app/models/registration.rb
```diff
@@ -521,4 +521,56 @@ def competition_events_changed?
   def serializable_hash(options = nil)
     super(DEFAULT_SERIALIZE_OPTIONS.merge(options || {}))
   end
+
+  def attempt_auto_accept
+    return false if Rails.env.production? && EnvConfig.WCA_LIVE_SITE?
```
gregorbg, line 526: This should be moved into a `Registration.AUTO_ACCEPT_ENABLED` setting, similar to how we already have `PaypalInterface.paypal_disabled?`

app/models/registration.rb
```diff
@@ -521,4 +521,56 @@ def competition_events_changed?
...
+
+    failure_reason = auto_accept_failure_reason
+    if failure_reason.present?
+      log_auto_accept_failure(failure_reason)
+      return false
+    end
+
+    update_payload = build_auto_accept_payload
+    auto_accepted_registration = Registrations::RegistrationChecker.apply_payload(self, update_payload)
```
gregorbg, line 535: Can you specify the cloning behavior explicitly here? Would make it easier if we ever decide to change the default value

app/models/registration.rb
```diff
@@ -521,4 +521,56 @@ def competition_events_changed?
...
+    end
+  end
+
+  private def auto_accept_failure_reason
+    return 'Competitor still has outstanding registration fees' if outstanding_entry_fees > 0
+    return 'Auto-accept is not enabled for this competition.' unless competition.auto_accept_registrations?
+    return 'Can only auto-accept pending registrations or first position on waiting list' unless competing_status_pending? || (competing_status_waiting_list? && waiting_list_position == 1)
+    return "Competition has reached auto_accept_disable_threshold of #{competition.auto_accept_disable_threshold} registrations" if competition.auto_accept_threshold_reached?
+
+    'Cant auto-accept while registration is not open' unless competition.registration_currently_open?
```
gregorbg, line 555: I18N :upside_down_face:

app/models/registration.rb
```diff
@@ -521,4 +521,56 @@ def competition_events_changed?
...
+      true
+    else
+      log_auto_accept_failure(auto_accepted_registration.errors.messages.values.flatten)
+      false
+    end
+  end
+
+  private def auto_accept_failure_reason
+    return 'Competitor still has outstanding registration fees' if outstanding_entry_fees > 0
+    return 'Auto-accept is not enabled for this competition.' unless competition.auto_accept_registrations?
```
gregorbg, line 551: How would the code even reach this point? `auto_accept_failure_reason` is only called when auto accept is attempted. And auto accept is only attempted if it's turned on, or not?

app/models/registration.rb
```diff
@@ -521,4 +521,56 @@ def competition_events_changed?
...
+    return 'Can only auto-accept pending registrations or first position on waiting list' unless competing_status_pending? || (competing_status_waiting_list? && waiting_list_position == 1)
```
gregorbg, line 552: (sidenote: You could move the waiting list part into `def waiting_list_leader?` or smth)

app/models/registration_payment.rb
gregorbg: Doing it here as well sounds like a good idea, you can never be sure enough.

app/models/registration.rb
```diff
@@ -521,4 +521,56 @@ def competition_events_changed?
...
+      return false
+    end
+
+    update_payload = build_auto_accept_payload
+    auto_accepted_registration = Registrations::RegistrationChecker.apply_payload(self, update_payload)
+
+    if auto_accepted_registration.valid?
+      update_lanes!(
+        update_payload,
+        'Auto-accept',
```
gregorbg, line 540: This should go into a constant. Something like `AUTO_ACCEPT_HISTORY_ID`. A few lines below you're logging errors as `Auto Accept` without dash, and it would be a nightmare to detect the difference between the two to determine whether something was actioned by auto-accept.

gregorbg, line 540: This should go into a constant. Something like `AUTO_ACCEPT_HISTORY_ID`. A few lines below you're logging errors as `Auto Accept` without dash, and it would be a nightmare to detect the difference between the two to determine whether something was actioned by auto-accept.

app/models/registration.rb
```diff
@@ -521,4 +521,56 @@ def competition_events_changed?
...
+    return "Competition has reached auto_accept_disable_threshold of #{competition.auto_accept_disable_threshold} registrations" if competition.auto_accept_threshold_reached?
+
+    'Cant auto-accept while registration is not open' unless competition.registration_currently_open?
+  end
+
+  private def log_auto_accept_failure(reason)
+    add_history_entry(
+      { auto_accept_failure_reasons: reason },
+      'System',
+      'Auto accept',
```
gregorbg, line 562: (See above for constant)

app/models/registration.rb
```diff
@@ -521,4 +521,56 @@ def competition_events_changed?
...
+      'System reject',
```
gregorbg, line 563: Maybe I18N?

config/locales/en.yml
```diff
@@ -396,6 +396,12 @@ en:
         extra_file: "Any extra file?"
         start_date: "Acknowledged since"
         end_date: "Acknowledged until"
+    errors:
+      models:
+        registration:
+          attributes:
+            competing_status:
+              exceeding_competitor_limit: "Competitor limit reached."
```
gregorbg, line 404: Why are you introducing such a deeply nested group?

app/models/registration.rb
```diff
@@ -67,7 +67,7 @@ class Registration < ApplicationRecord
   end
 
   def update_lanes!(params, acting_user)
-    Registrations::Lanes::Competing.update!(params, self, acting_user.id)
```
gregorbg, line 70: Why are you removing the `.id` here? The `update!` method does need the ID, not the whole entity

gregorbg (changes requested):

app/models/registration.rb
dunkOnIT: Ok so "why did this work" has a few interrelated answers:
- acting_user_id was only used to write the user_id to the registration history
- tests didn't assert the contents of the registration history (I've since added a test to fix this) 
- `update_lanes!` is only used in 2 places in the code since the refactor of lanes/competing.rb: 
  - in tests
  - in this auto accept code

We should update `update_lanes!` invocations to use `update_raw` instead - my suggestion is that this happens in a follow-up PR, given that this PR is already pretty busy? I can do it here too, it'll just increase the diff (particularly in the tests)

app/models/registration.rb
dunkOnIT: Added `auto_accept_enabled?` method to contain this

app/models/registration.rb
dunkOnIT: I don't follow this request - what is being cloned, and what default value might we change? 

The only thing I can think of is the contents of the update payload, but that's just abstracted in a very easy-to-reach method if we ever want to change it

app/models/registration.rb
dunkOnIT: Done

app/models/registration.rb
dunkOnIT: Question here: Will this mean that we'll end up saving translated (ie, non-`en`) strings into the registration history? If so, I think we need to be careful about _what_ locale gets stored - a French user signing up for a US comp shouldn't cause the auto-accept message to be in French. 

Something I have been thinking about is saving the translation keys themselves into the registration history, and displaying them localized in the user's browser - but in this case, we should be saving the key, not the i18n'd string

app/models/registration.rb
dunkOnIT: Done

app/models/registration.rb
```diff
@@ -521,4 +521,56 @@ def competition_events_changed?
...
+      )
+      true
+    else
+      log_auto_accept_failure(auto_accepted_registration.errors.messages.values.flatten)
+      false
+    end
+  end
+
+  private def auto_accept_failure_reason
+    return 'Competitor still has outstanding registration fees' if outstanding_entry_fees.positive?
```
dunkOnIT, line 554: You asked about i18n-ing these strings, although I can't find the comment at the moment for some reason. My other comment re translations in registration history applies here too.

app/models/registration.rb
dunkOnIT: As discussed in Huddle, added a check in RegistrationPayment, but leaving this here for redundancy

config/locales/en.yml
dunkOnIT: Resolved - key changed to registrations.errors, and explicitly passing `message` in the AR error definition

app/models/registration.rb
gregorbg: I'm happy to get rid of `update_lanes!`, and you're welcome to do that in a follow-up PR.

However, please keep in mind that `update_raw` loads a separate copy of the registration from the database and operates on that volatile copy. (This is why we previously had that infamous `reload` that nobody knew how to get rid of). In contrast, `update!` has the advantage of operating directly on an existing `Registration` model (which you can see here with the `self` being passed directly into the method).

So in that follow-up PR, just be careful about which one you're using

app/models/registration.rb
gregorbg: I mean that `apply_payload` has a `clone` parameter that you should explicitly specify, instead of relying on the implicit `false` default from `def self.apply_payload(registration, raw_payload, clone: true)`

app/models/registration.rb
gregorbg: I found the other comment, replied directly below there. But keeping this conversation open as a marker so that you know whatever conclusion we reach there has to be applied here as well

app/models/registration.rb
gregorbg: Good point! I really don't want to store localized strings in the DB at all, except maybe for human-written `comments` from a Registration or something similar.

Maybe the way to go about this is to pass in values to the reg history that we interpret as "error codes"? That way you'd have a standardized list that you can just "pipe through" in the API and then later localize in the frontend

app/models/registration_payment.rb
```diff
@@ -9,6 +9,7 @@ class RegistrationPayment < ApplicationRecord
   belongs_to :refunded_registration_payment, class_name: 'RegistrationPayment', optional: true
   has_many :refunding_registration_payments, class_name: 'RegistrationPayment', inverse_of: :refunded_registration_payment, foreign_key: :refunded_registration_payment_id, dependent: :destroy
 
+  after_create :auto_accept_hook, if: -> { registration.auto_accept_in_current_env? && registration.competition.auto_accept_registrations? }
```
gregorbg, line 12:
```suggestion
  delegate auto_accept_in_current_env?, :auto_accept_registrations?, to: :registration
  
  # in registration.rb, you'd also need the following delegate:
  # delegate auto_accept_registrations?, to: :competition

  after_create :auto_accept_hook, if:  [:auto_accept_in_current_env?, :auto_accept_registrations?]
```

lib/registrations/lanes/competing.rb
```diff
@@ -42,13 +42,17 @@ def self.update!(update_params, registration, current_user_id)
           changes[:event_ids] = registration.changed_event_ids if registration.changed_event_ids.present?
 
           registration.save!
-          registration.add_history_entry(changes, 'user', current_user_id, Registrations::Helper.action_type(update_params, current_user_id))
+          if acting_entity_id == 'auto-accept'
```
gregorbg, line 45:
```suggestion
          if acting_entity_id == Registration::AUTO_ACCEPT_ENTITY_ID
```

lib/registrations/lanes/competing.rb
```diff
@@ -42,13 +42,17 @@ def self.update!(update_params, registration, current_user_id)
...
+            registration.add_history_entry(changes, 'System', acting_entity_id, Registrations::Helper.action_type(update_params, acting_entity_id))
```
gregorbg, line 46: Where does this `'System'` come from and is it standardized somewhere?

gregorbg (changes requested):

app/models/registration.rb
dunkOnIT: Makes sense! I have set `clone: false` explicitly

app/models/registration.rb
dunkOnIT: Refactored to push error codes into the reg history - frontend work can happen in a subsequeny PR, that I'd argue isn't even necessary for launch given that their primary use is for us to debug auto accept issues

app/models/registration_payment.rb
dunkOnIT: Committed a version with the comment on line 14/15 (but with the suggested delegation in registration.rb)

lib/registrations/lanes/competing.rb
dunkOnIT: Changed

lib/registrations/lanes/competing.rb
dunkOnIT: Is it now!

app/models/registration.rb
```diff
@@ -245,7 +252,7 @@ def registration_history
         actor_id: r.actor_id,
         timestamp: r.created_at,
         action: r.action,
-      }
+      }.with_indifferent_access
```
gregorbg, line 255: Not happy about this in our main code. Why do you need it?

app/models/registration.rb
```diff
@@ -521,4 +529,62 @@ def competition_events_changed?
...
+  end
+
+  private def build_auto_accept_payload
+    status = if competition.registration_full_and_accepted? && competing_status_pending?
+               Registrations::Helper::STATUS_WAITING_LIST
+             else
+               Registrations::Helper::STATUS_ACCEPTED
+             end
+
+    { user_id: user_id, competing: { status: status } }.with_indifferent_access
```
gregorbg, line 588: Again, not happy about `with_indifferent_access`.

gregorbg (commented):
Looking all good now in terms of structure, still a bit wary about `with_indifferent_access` though.

gregorbg:
Oh and you also have a good amount of merge conflicts that need to be resolved :upside_down_face:

app/models/registration.rb
dunkOnIT: The hash we were getting from this without intervention is the following: 
```ruby
{changed_attributes: {"auto_accept_failure_reasons" => "-7001"}, actor_type: "System", actor_id: "auto-accept", timestamp: Tue, 06 May 2025 03:19:41.320953000 UTC +00:00, action: "System reject"}
```

This was happening because `changed_attributes` was a hash populated with strings from the database - I'm now calling `symbolize_keys` on it when it's created, which fixes the issue

gregorbg: Muuuch better :relieved:

app/models/registration.rb
dunkOnIT: Fixed! Had to use `deep_stringify_keys` as the hash is 2 levels deep

gregorbg: Why do you need to have stringified keys here? Usually symbols are the more modern/ergonomic choice

gregorbg: Also, if you really, _really_, _absolutely_ need string keys (you shouldn't), then you can directly create the hash with string keys: `{ "user_id" => user_id, "competing" => ... }`

dunkOnIT: This is building a payload to be sent to `apply_payload` - which is built to accept `params`, which is string keys by default. So makes sense to me to keep the keys as strings

We could also handle this in apply_payload instead, but I think our naming/context makes it clear enough to expect to be dealing with string keys in that function?

gregorbg: Ah, so that's kinda our fault. The `apply_payload` method mostly gets passed stuff from the controller's `params`, which are usually handed down with `indifferent_access` already baked in. So if anything, it's my fault for not refactoring it properly when I rebuilt `apply_payload` :P

gregorbg: In that case, please at least go with instantiating string keys directly, instead of instantiating symbol keys and then direclty converting them to string keys on the spot

dunkOnIT: It is now being instantiated with string keys :+1:

gregorbg (approved):

config/locales/en.yml
```diff
@@ -1959,6 +1960,11 @@ en:
         -6001: "WCA Payment is not enabled for this competition"
         -6002: "You need to finish your registration before you can pay"
         -6003: "No outstanding payment found - this is probably because you have already paid for your registration."
+        -7001: "Competitor still has outstanding registration fees"
+        -7002: "Auto-accept is not enabled for this competition."
+        -7003: "Can only auto-accept pending registrations or first position on waiting list"
+        -7004: "Competition has reached auto_accept_disable_threshold"
```
AlexandreOndet, line 1966: Is it normal that this doesn't have brackets ? If `auto_accept_disable_threshold` should be replaced by its value, it probably should, no?

dunkOnIT: I'm no longer passing in the value of `auto_accept_disable_threshold`, given that we're now rendering an i18n string instead of a string directly. It could definitely be done, it's just _slightly_ more complicated and I really wanted to get this PR merged

dunkOnIT: I'm no longer passing in the value of `auto_accept_disable_threshold`, given that we're now rendering an i18n string instead of a string directly. It could definitely be done, it's just _slightly_ more complicated and I really wanted to get this PR merged

PR #10979, opened 2025-03-07
Add ticket comment email notification

danieljames-dj:
This PR implements comments email notification for tickets. A sample email screenshot is attached below:

Screenshot 2025-03-07 at 09 06 49

### DIFF
--- a/app/controllers/ticket_comments_controller.rb
+++ b/app/controllers/ticket_comments_controller.rb
@@ -28,8 +28,6 @@ class TicketCommentsController < ApplicationController
       acting_stakeholder_id: acting_stakeholder_id,
     )
 
-    ticket.ticket_stakeholders.each { |stakeholder| stakeholder.notify_new_comment(comment) }
-
     render json: comment
   end
 end
--- a/app/controllers/ticket_comments_controller.rb
+++ b/app/controllers/ticket_comments_controller.rb
@@ -28,6 +28,8 @@ class TicketCommentsController < ApplicationController
       acting_stakeholder_id: acting_stakeholder_id,
     )
 
+    ticket.ticket_stakeholders.each { |stakeholder| stakeholder.notify_new_comment(comment) }
+
     render json: comment
   end
 end
--- a/app/mailers/tickets_mailer.rb
+++ b/app/mailers/tickets_mailer.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+class TicketsMailer < ApplicationMailer
+  def notify_create_ticket_comment(ticket_comment, stakeholder)
+    @recipient_name = stakeholder.name
+    @ticket_id = ticket_comment.ticket.id
+    @ticket_comment = ticket_comment
+
+    mail(
+      to: [stakeholder.email],
+      subject: "[Ticket #{ticket_comment.ticket.id}] New comment added",
+    )
+  end
+end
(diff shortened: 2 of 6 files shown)


gregorbg:
What kind of problem are you trying to solve with this? I'm a bit worried that this could flood peoples' inboxes once we start relying more on Tickets...

danieljames-dj:
> What kind of problem are you trying to solve with this? I'm a bit worried that this could flood peoples' inboxes once we start relying more on Tickets...

Basically wanted to notify people that there is a new comment so that an action can be taken. I understand it will flood the inboxes, but I'm assuming that every comment on one thread will stay in one thread.

gregorbg:
In which use-case does an email notification make sense?
Or, to put it differently, which comments are so important that you need to have an email notification about it?

danieljames-dj:
I'll explain with the example of "Edit person" tickets. If somebody requests to change their country, then if there is some mistake in the query and if WRT wants to get it clarified, WRT will post a comment in the ticket. At that time, the user should be notified so that they can reply to it.

Similarly, for future "Competition announcement" tickets, it's very common that WCAT might have some clarification questions for the Delegates. At that time, the Delegates need to be notified that there is a new comment. Similarly, when the delegate replies back, the WCAT should be notified too.

I'm also having some ideas to send notifications to frequent ticket user groups (like WRT, WCAT, etc) via Slack (and continue the email notification for remaining stakeholders). If you agree with this idea, I'll be happy to consider WRT as a 'frequent ticket user group' now itself (in this same PR or in the very next PR) and have their notifications to Slack.

gregorbg:
Here's a suggestion to reduce noise: When creating a new comment, you don't need to loop through the stakeholders and notify each of them individually. (But that's what you're doign right now...)

Instead, the person who just submitted the comment probably doesn't need to be emailed about it. It's enough if they are listed in the `Reply-To` header. If the person is acting on behalf of a group stakeholder, put the respective group as `Reply-To` instead.

And also, definitely try to keep every relevant recipient in the same email. Two reasons:
1. Our email quota in AWS is limited. Sending the same email X times instead of sending it once with X recipients listed makes a huge difference for us
2. When one of the recipients replies (and is smart enough to use Reply-All, which will at least be the case with WCA-internal tickets) then all other recipients will stay in the loop

gregorbg:
P.S.: Slack sounds nice but that's for a follow-up PR :upside_down_face:

danieljames-dj:
Thanks @gregorbg. I didn't knew about the AWS email quota. Also I prefer not to have the "communication between stakeholders" in two places - instead have it only in the tickets page. I want to use the email as just "notification" for now. So is it fine if I don't use "Reply-to" and add every stakeholders to bcc?

Also I agree that the person who added the comment doesn't need the notification if the person is sending on behalf of a stakeholder of type "User". I'll try to have that implemented.

One more technical query: A few minutes ago it came to my mind that sending this notification using `after_create` will be way better, and it need not come in the controller. Do you agree with changing to `after_create` in the model file?

gregorbg:
With `after_create`, you will lose the information about who is currently making the change

danieljames-dj:
> With `after_create`, you will lose the information about who is currently making the change

We have it within the `comment.acting_user`.

gregorbg:
Okay, in that case go for it!

danieljames-dj:
Done, thank you.

gregorbg:
Random thought: If this mail goes out to users to notify them about WRT decisions (about nationality change, etc.) then should you put the strings in `en.yml` instead so they can be translated?

danieljames-dj:
But the problem is that WRT won't be able to understand non-English languages. If it was sent to stakeholders separately, then technically we can provide different locales to different users.

Also, usually communications with WRT is usually in English.

gregorbg (approved):

PR #10980, opened 2025-03-07
Reg Admin: scroll-to section and sections summary

### LINKED ISSUE #15107: Review Kevin Reg Admin PR

kr-matthews:
Some ideas. The main point is to allow scrolling to a section. Then I thought having information about the section would be good (and might alleviate the need to scroll), so I added that as a separate thing. I could combine them by making the summary table rows clickable and removing the "scroll to" menu option.

~I also combined the two email options into a dropdown to save space (on mobile), but I suspect there may be more complaints about needing 2 clicks (though I think we only got 1 complaint about moving registrations?). I'm thinking about reverting it.~

~I'm not very happy about the count in "Move ${count} to" since the export and email options should also have it for consistency. Having it in them all 3 doesn't make sense, so having it once, outside the buttons somewhere would make sense. But I can't come up with a good place.~

Thoughts? ~Any ideas to fix the Popup table on mobile (see last video)?~

It's a bit annoying that we aren't consistently using keys across the front-end code, e.g. waiting vs waitlist, accepted vs approved. I can 'fix' it, but I think that would require changing a few translation keys, which isn't ideal. _Edit: I kind of fixed it, via a helper function to translate status key to translation key._

:warning: **Images and screenshots here are out-of-date, see the latest comment at the bottom.** :warning:

image
image

Competition without competitor limit:
image
image

Competition with competitor limit:
image
image

A small offset added to the scroll-to functionality, otherwise the header is hidden under the buttons...
[Screencast from 2025-03-06 10:30:57 PM.webm](https://github.com/user-attachments/assets/bcdec4df-80a2-489f-922a-07fa143c1b30)
...although that happens on mobile anyway :\ and the popup summary table doesn't quite work.
[Screencast from 2025-03-06 10:35:03 PM.webm](https://github.com/user-attachments/assets/7acde978-6e79-4cb0-a175-3b517cb290fb)


app/webpacker/components/Persons/NewPersonForm/NewPersonForm.js
```diff
@@ -140,7 +137,7 @@ function NewPersonForm({
             name,
             dob,
             gender,
-            country_id: countryIdForIso2(countryIso2),
+            countryId: countryIdForIso2(countryIso2),
```
gregorbg: Just curious, why did you change this from `snake_case` to `camelCase`? The backend usually expects snake case and I don't see any relevant changes to the backend in this PR...

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
```diff
@@ -342,7 +429,50 @@ export default function RegistrationActions({
...
+  partitionedSelectedIds,
+  partitionedRegistrations,
+  partitionedMaximums,
+  selectedCount,
+  registrationCount,
+  withSelectedCounts,
+  withMaximums,
+}) {
+  return (
+    <Table basic="very">
```
gregorbg: All of the below table components can use dot notation on the original `Table` import, for example `Table.Header` or `Table.HeaderCell` (and others). This saves you a lot of import statements at the top.

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationList.jsx
```diff
@@ -525,12 +541,14 @@ export default function RegistrationAdministrationList({ competitionInfo }) {
...
 function SectionToggle({
-  icon, title, inParens, color,
+  icon, title, inParens, color, sectionRef,
 }) {
   return (
-    <Header as="span" size="large">
-      <Icon name={icon} color={color} />
-      {`${title} (${inParens})`}
-    </Header>
+    <span ref={sectionRef} style={{ scrollMarginTop: '7em' }}>
```
gregorbg: Does the `Header` component itself not support `ref` natively? If not, I think SemUI has a `Ref` component that you can use to pass refs down within their own framework

gregorbg: (we do actually use that component in a couple of places already, so just search for `<Ref` throughout our codebase and you will see)

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationList.jsx
```diff
@@ -68,6 +68,15 @@ export default function RegistrationAdministrationList({ competitionInfo }) {
...
   const actionsRef = useRef();
 
+  const pendingRef = useRef();
+  const waitlistRef = useRef();
+  const approvedRef = useRef();
+  const cancelledRef = useRef();
+  const rejectedRef = useRef();
+  const tableRefs = useMemo(() => ({
+    pendingRef, waitlistRef, approvedRef, cancelledRef, rejectedRef,
+  }), []);
```
gregorbg, line 81: I like the concise notation, but I am a bit afraid it would make the code "vulnerable" to refactors. What about using the actual `competition_status` enum values (like `"pending"`, `"approved"`, etc) as keys?
That would make it a bit more verbose here but more standardized and easier to maintain in the future

app/webpacker/components/Persons/NewPersonForm/NewPersonForm.js
kr-matthews: Uh this whole file looks like it might just be a mishandled rebase conflict. I didn't check carefully before pushing, since it's just a draft and I'm more looking for feedback on the UI rather than on the code, at this point.

kr-matthews:
Not 100% done (tooltips are sometimes not visible, the table doesn't work well on mobile, email tooltip missing text, etc.) but would like feedback on the direction of the UI in particular. Thinking of adding total money to the summary table too, but don't want it to get too crowded.

[Screencast from 2025-09-07 08:29:32 PM.webm](https://github.com/user-attachments/assets/1480eeac-0f86-45bc-a47a-ee596fbd7b1e)


[Screencast from 2025-09-07 08:23:36 PM.webm](https://github.com/user-attachments/assets/692df1de-2a2d-41bd-9fcd-cfeebc6dc97e)

kr-matthews:
⚠️ Edit: these videos are out-of-date. ⚠️

Only issue is that the notifications are hidden behind the buttons. I'll fix that next time.

(Oops I forgot to check the box to record the cursor position... Hopefully it's obvious.)

[10980.webm](https://github.com/user-attachments/assets/e1b47223-9110-4e33-9a53-2c32f0d9903a)

[noncompeting.webm](https://github.com/user-attachments/assets/f1cd12b2-a2b0-44e3-b5d2-7c7f10d25cab)

The table didn't work on mobile because it was automatically becoming a stackable table. So I used a grid instead. The grid is by default wide (about twice as wide as the mobile screen width), so I added an explicit width. If there's a better way, let me know.

The popups need to be on top, otherwise they interfere with the dropdowns. To have them on top, I had to move the buttons down.

I could also add another "$" button which would show money totals per registration status (when payments are enabled).

kr-matthews:
Message moved up:

image

gregorbg (commented):
I'm fine with the code but I need some degree of user-testing on this. My personal experience with the day-to-day usage of the Reg Admin page is limited.

@dunkOnIT Can you play around with this to give a third-party opinion on whether these changes all make sense from a UX perspective?

dunkOnIT:
As mentioned in DM's to Kevin, I really like the changes - especially the infobox/summary box showing all the registrations! 

Notes/suggestions from me: 
- [ ] Do you think we could combine the infobox pictured above and the scroll-to functionality? The idea being that if the cursor is over the modal that comes up, it stays up - and the icon can be clickd on to scroll to a particular section
- [ ] Add explainer text for "Approved": "Approved registrations are confirmed to be able to participate in the competition in their registered events." 
- [ ] Infobox table sizing in Desktop view looks a little squished - I think either less vertical padding in the cells, or more width in the columns, would alleviate this 

@kr-matthews feedback finalized - apologies, forgot to remove my "still in progress" disclaimer at the top of this comment

kr-matthews:
I got rid of the info icon & popup table, and just put the counts as a hint/description on the move-to options. This is simpler than a table. It doesn't include the competitor limit any more, but that was a bit awkward anyway as it only applied to at most one section.

image

https://github.com/user-attachments/assets/b9d359ba-62cb-4fd1-b303-792b28f2d2a7

https://github.com/user-attachments/assets/23df93bd-276a-434d-b00b-1e3d20d79de4

The only downside is that I had to manually specify a (min) width for the menu items. As mentioned [in the documentation](https://semantic-ui.com/modules/dropdown.html#:~:text=Using%20a%20description,Filter%20Tags), this is necessary (when the item is in a menu). I added enough width to handle 4-digit registration totals (e.g., "1000" or "1000 of 1000"). It's not ideal though.

image

I also added the approved-registrations text that Duncan requested, and renamed a few translation keys which I had pluralized (apparently not renaming them would cause issues for translators), and fixed "e.g.," usages.

dunkOnIT:
Amazing, thanks Kevin! Happy with it from a functionality perspective :+1:

kr-matthews:
I don't understand why the tests keep failing

FinnIckler:
Tests are failing because you changed the spelling in the en.yml 
`(eg., Stripe) to (e.g., Stripe)`. Let me push a fix

gregorbg (approved):

config/locales/en.yml
```diff
@@ -1203,7 +1203,10 @@ en:
       country_plural:
         one: "1 region"
         other: "%{count} regions"
-      export_csv: "Export to CSV"
+      export_to_csv:
+        zero: "Export all registrations to CSV"
```
gregorbg: The CI failure is due to using `zero` here. English grammar does not have a "zero" case in the stricter sense (per Unicode CLDR regulations) so you must not use it here.

If you want/need this "all" option, then you have to store it as a separate `export_all_to_csv` key and perform the `if length === 0` check in the application code

gregorbg: (This is something we introduced rather recently, because we're migrating to a new translation pipeline that is much stricter with CLDR rules than our old tools used to be)

PR #10982, opened 2025-03-07
Add a WCA API Client to make authenticated Requests against the WCA API from next

FinnIckler:
Also includes some example Documentation and an example usage in the `usePermissions` hook.

To be discussed on monday


next-frontend/src/lib/wca/users/usePermissions.ts
```diff
@@ -0,0 +1,46 @@
...
+    return {
+      permissions: NO_PERMISSIONS,
+      isLoading: false,
+    };
+  }
+
+  return {
+    isLoading,
+    permissions: request?.data || NO_PERMISSIONS
+  }
```
kr-matthews, line 45: We wouldn't want to access permissions while they're loading right? I would just return `{ isLoading: true }` if it's loading, and `{ isLoading: false, permissions: request.data }` otherwise.

To avoid any awkwardness with having to check `isLoading` every time before accessing permissions, a wrapper can wait for loading to finish, and only render children after.

I did this in a work project and it worked well.

kr-matthews (commented):
Can we set up linting sooner rather than later?

next-frontend/src/lib/wca/users/usePermissions.ts
FinnIckler: how would such a wrapper look like? would that use the react 19 use hook? I saw something in the next doc, but I think it would require a paradigm shift for how we do requests.

kr-matthews:
```
PermissionWrapper({ children }) {
  const permissions = usePermissions();
  if (permissions.isLoading) return <LoadingUI />
  return (
    <PermissionProvider value={permissions.value}>
      {children}
    </PermissionProvider>
  )
}
```
and then access them in the children via the context.

Apply at the router root (preferable), or a particular router branch (next best), or on each individual component that needs permission, depending on how things are set up.

I'm not familiar with `use` (but it's apparently _not_ a hook).

app/controllers/api/v0/api_controller.rb
```diff
@@ -232,15 +232,19 @@ def export_public
...
   # From: https://github.com/doorkeeper-gem/doorkeeper#authenticated-resource-owner
   private def current_api_user
     @current_api_user ||= User.find_by_id(doorkeeper_token&.resource_owner_id) if doorkeeper_token&.accessible?
   end
 
   private def require_user!
-    raise WcaExceptions::MustLogIn.new if current_api_user.nil? && current_user.nil?
-    current_api_user || current_user
+    raise WcaExceptions::MustLogIn.new if current_user.nil?
+    current_user
```
gregorbg, line 247: In Ruby, you can implement this pattern as `current_user || raise "Nope."`

next-frontend/package.json
```diff
@@ -9,6 +9,8 @@
     "lint": "next lint"
   },
   "dependencies": {
+    "@tanstack/react-query": "^5.67.2",
+    "openapi-fetch": "^0.13.4",
```
gregorbg, line 13: Alphabetical order not kept. This is going to break with dependency updates eventually, creating an unnecessary diff in the future

next-frontend/src/app/layout.tsx
```diff
@@ -28,11 +30,15 @@ export default function RootLayout({
...
-        </AuthProvider>
+        <WCAQueryClientProvider>
+          <AuthProvider>
+            <UiProvider>
+              <PermissionProvider>
+                {children}
+              </PermissionProvider>
+            </UiProvider>
+          </AuthProvider>
+        </WCAQueryClientProvider>
```
gregorbg, line 41: Provider pyramid, yay!

next-frontend/src/auth.ts
```diff
@@ -12,4 +12,17 @@ export const { handlers, signIn, signOut, auth } = NextAuth({
       clientSecret: process.env.OIDC_CLIENT_SECRET,
     },
   ],
+  callbacks: {
+    jwt({ token, trigger, session, account }) {
+      if(account?.access_token){
+        return { ...token, accessToken: account?.access_token }
```
gregorbg, line 18: Inside this `if` you don't need the `?.` safe operator anymore, do you?

next-frontend/src/lib/wca/wcaAPI.ts
```diff
@@ -0,0 +1,5 @@
+import createClient from "openapi-fetch";
+import type { paths } from "./wcaSchema"; // generated by openapi-typescript
+
+export const unauthenticatedClient = createClient<paths>({ baseUrl: "http://localhost:3000/api/v0/" });
```
gregorbg, line 4: Why does the authorized one get `application/json` headers, but the unauthenticated version doesn't?

next-frontend/src/providers/PermissionProvider.tsx
```diff
@@ -0,0 +1,73 @@
...
+      canReadGroupPast: group => Boolean(rawPermissions && rawPermissions.can_read_groups_past.scope.includes(group)),
+      canRequestToEditProfile: profile => Boolean(rawPermissions && allOrSpecificScope(profile, rawPermissions.can_request_to_edit_others_profile.scope)),
+    }
+  }, [request])
+
+  if(isLoading) {
+    return (
+      <p>Loading...</p>
+    )
+  }
```
gregorbg, line 66: Still not too wild about rendering a DOM node while loading. Shouldn't the `Boolean(rawPermissions && actualPermissionCheck(...))` trick be enough?

gregorbg (commented):
Some comments while you're still working diligently :wink:

next-frontend/package.json
FinnIckler: yeah I noticed myself due to next crashing (I think I messed up the yarn lock in merge)

next-frontend/src/providers/PermissionProvider.tsx
FinnIckler: it might create jumpy UIs were it renders some elements differently a second time if we don't

next-frontend/src/lib/wca/wcaAPI.ts
FinnIckler: good point

app/controllers/api/v0/api_controller.rb
FinnIckler: that raises a syntax error

FinnIckler: works with brackets though

next-frontend/src/providers/PermissionProvider.tsx
gregorbg: How so?

FinnIckler:
As I feared some of the API related tests are failing now. Gonna look into that now. The Competition test page is there now though

FinnIckler:
3) is an easy fix, but 1) and 2) are a bit puzzling to me. Seems to be not noticing the manage_competition scope for some reason

next-frontend/src/providers/PermissionProvider.tsx
kr-matthews: Besides potential UI jumping around, passing 'incorrect' permissions to components seems bad. There could also be (potentially, in the future) side-effects depending on these permissions, which would be triggered with the wrong values.

next-frontend/src/providers/CompetitionProvider.tsx
```diff
@@ -0,0 +1,30 @@
...
+  })
+
+  if(isLoading) {
+    return (
+      <p>Loading...</p>
+    )
+  }
+
+  return (
+    <CompetitionContext.Provider value={request!.data!}>
```
kr-matthews, line 26: Probably should have some error handling instead of `!`?

FinnIckler:
I've added an example of SSRing the Competition Test page

gregorbg (approved):

PR #10983, opened 2025-03-08
Disable event update button on register step after event update deadline

pranav-027:
Users can send event update requests even after the deadline, and even after the competition, by using the register subtab
image

This competition took place on 22-23rd of feb
image


app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
```diff
@@ -303,6 +303,10 @@ export default function CompetingStep({
     ? competitionInfo.guests_per_registration_limit
     : defaultGuestLimit;
 
+  const hasRegistrationEditDeadlinePassed = hasPassed(
+    competitionInfo.event_change_deadline_date ?? competitionInfo.start_date,
```
kr-matthews, line 307: I think `hasPassed` is designed to handle date times, but `start_date` is just a date with no time, right? Is this going to have different results for people in different time zones?

pranav-027: Umm, not quite sure about it. I just wanted to mimic the same behavior as approval tab so the logic is taken from that page only -
https://github.com/thewca/worldcubeassociation.org/blob/260b9efab4d6db2ae481e6d0000d806259ec8292/app/webpacker/components/RegistrationsV2/Register/RegistrationOverview.jsx#L23-L25

PR #10984, opened 2025-03-08
Reg Admin: search

kr-matthews:
Took a long time to figure out that we weren't importing the Search styling...

Search by name, wca id, email, comment, or admin note.
Display these things, plus amount paid.
Clicking a results navigates to the edit registration page (and then pressing back in the browser takes you back to reg admin lists).

Also added some dividers.

_Edit: I forgot to do translations._

Note that the rendering of the individual registration results is automatic, and requires **string** values for title, description, price, and image. We can customize it, which would allow us to render more nicely (e.g. put comments and admin note on new lines, add the country flag, etc.), but then we'd have to do all the rest of the styling too. I did this for the category, which is how I was able to show icons there instead of just text. It actually works if you pass in JSX instead of strings, but there are warnings in the console, and some internal checks use `===` which won't work (though I'm not sure how that would matter to us for this use-case?).

image
image
image
image
image

image

Removing `fluid` makes it nicer(?) on a computer, but worse on phone (adds a small horizontal scroll amount):
image
image


app/webpacker/lib/utils/registrationAdmin.js
```diff
@@ -13,6 +13,50 @@ export const APPROVED_ICON = 'check';
...
+    case 'cancelled':
+      return CANCELLED_ICON;
+
+    case 'rejected':
+      return REJECTED_ICON;
+
+    default:
+      return undefined;
+  }
+};
```
gregorbg, line 58: Just curious: Could these helper methods be useful in other parts of the frontend code as well?

gregorbg (commented):
Re rendering of the components: Did you take a look at how `SearchWidget` does it? And could any of the predefined `*Item` components there be of any help to you?

app/webpacker/lib/utils/registrationAdmin.js
kr-matthews: Yes, definitely.
I was hesitant to do too many changes here since I have several other pending PRs. But at some point I'll re-use these.

kr-matthews:
> Re rendering of the components: Did you take a look at how `SearchWidget` does it? And could any of the predefined `*Item` components there be of any help to you?

I have now. They seem to just be basic combinations of divs and spans, without any styling. So if we use one of those then we would lose the styling for bigger/bold header, aligning everything, etc. (Or build a new one, with custom styling, but we want to avoid that right.)
The only thing I would do differently if building it myself is to put the comment/admin note on their own lines, and add the country flag somewhere. Which doesn't seem worth it.

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationSearch.jsx
```diff
@@ -0,0 +1,104 @@
...
+              name: user.name,
+              wcaId: user.wca_id,
+              email: user.email,
+              comment: competing.comment,
+              note: competing.admin_comment,
+            },
+          })),
+        },
+      ]),
+    ),
```
gregorbg, line 55: Lodash has `_.mapValues`, which may be an idiomatic alternative to what you're trying to do here

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationSearch.jsx
```diff
@@ -0,0 +1,104 @@
...
+
+      return Object.fromEntries(
+        Object.entries(options).map(([status, { name, results }]) => [
+          status,
+          {
+            name,
+            results: _.filter(results, isMatch),
+          },
+        ]).filter(([, { results }]) => results.length > 0),
+      );
```
gregorbg, line 74: Same remark about `_.mapValues`, as well as `_.pickBy` (which is not-so-aptly named but effectively just `filterValues`)

gregorbg (approved):
Just two tiny ideas about using Lodash for some more idiomatic code, but otherwise LGTM :D

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationSearch.jsx
kr-matthews: I don't think that will work here because I need the key in the (new) value: `name: status`. This is used to get the icon when rendering the category. I don't see any other lodash functions which would work either.

The status could be found in any of the registrations, but there might be 0 registrations in the category (in which case we filter out the category, so I guess that wouldn't matter) and also we have different keys for the front-end and back-end (e.g. `waiting_list` vs `waiting` - see the `partitionRegistrations` function).

I'll probably do a follow-up PR here to clean up the naming and use the new icon/color helper functions (as you suggested in an earlier comment). But I can't directly align the keys them because we can't use snake case.

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationSearch.jsx
kr-matthews: Neat, done. However, my local db is broken so I'm not able to verify my changes work correctly.

kr-matthews:
Merged with main and updated some minor bugs. It's working as expected:

[Screencast from 2025-04-24 06:57:27 PM.webm](https://github.com/user-attachments/assets/8fa98089-4cc5-4f41-befb-e6ad60e6e75d)

gregorbg:
Coolio, deploying now :tada:

PR #10987, opened 2025-03-09
Reg Admin: Also sort unpaid registrations on moving to waitlist & add info sentence

kr-matthews:
I also just noticed that when adding to the waitlist, we're not sorting the unpaid registrations (only the paid registrations). We definitely should be - or change the UI to make it clear that order of selection matters.

Screenshot from 2025-03-08 18-46-31


gregorbg (approved):

gregorbg:
I'm fine with this PR as-is, do you want to refine it more based on your PR description comment?

gregorbg (approved):

Issue #10928, opened 2025-02-28
Hide schedule link input from delegate report


kr-matthews:
It hasn't been editable for years, no need to show it. All current and future competitions will have a pre-defined link to the schedule.

![Image](https://github.com/user-attachments/assets/2a6722ae-87b8-4c6b-95f6-e69041b3a4d5)

PR #11001, opened 2025-03-10
removed edit schedule URL form delegate reports

pranav-027:
Fixed https://github.com/thewca/worldcubeassociation.org/issues/10928

image


app/views/delegate_reports/edit.html.erb
```diff
@@ -12,8 +12,6 @@
     <% if @delegate_report.posted? %>
       <%= f.input :discussion_url %>
     <% end %>
-    <% @competition.delegate_report.schedule_url = link_to_competition_schedule_tab(@competition) if @competition.has_schedule? %>
-    <%= f.input :schedule_url, readonly: @competition.has_schedule? %>
```
gregorbg, line 16: You cannot remove this altogether! We still need to store the information in the backend, because there are older competitions (predating the WCA website schedule feature) where it definitely makes sense to include and manipulate this field.

You can however turn the field into a hidden field if there is a (WCA website internal) schedule attached

gregorbg (changes requested):

app/views/delegate_reports/edit.html.erb
pranav-027: Done

gregorbg (commented):
Getting there! Although in its current form, the schedule URL wouldn't be passed to the backend at all for modern competitions. I think it's safer to use `f.hidden` in that case.

pranav-027:
@gregorbg Done

pranav-027:
ping @gregorbg

gregorbg:
Fine by me but has merge conflicts now, because we renamed the `has_schedule` method

pranav-027:
Merge ready!

PR #11002, opened 2025-03-10
Update browserslist and caniuse-lite definitions

gregorbg:
Attempt to fix the occasional error that users have been reporting about "not seeing anything on the competitions overview page". From our investigation so far, this is due to old browsers not having `toSorted()` on arrays.

We can make reasonable effort by polyfilling, although I'm not going to polyfill everything down to Chrome 90. We have to draw a reasonable line somewhere instead of blowing up our bundle size meaninglessly. Since JSON does not support comments, I will explain my suggestion here:

```
// Support every browser with > 0.2% market share, according to `caniuse`. 
"> 0.2%",
// Support all 5 last versions of (major) browsers
"last 5 versions",
// Support Firefox long-term releases
"Firefox ESR",
// Do NOT support dead browsers, even if they have a big market share (Sorry IE11)
"not dead",
// Do NOT bother supporting Opera Mini, which has a horrible ECMA implementation
"not op_mini all"
```

See https://browsersl.ist/#q=%3E+0.2%25%2C+last+5+versions%2C+Firefox+ESR%2C+not+dead%2C+not+op_mini+all

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,7 +101,11 @@
     "whatwg-fetch": "^3.6.20"
   },
   "browserslist": [
-    "defaults"
+    "> 0.2%",
+    "last 5 versions",
+    "Firefox ESR",
+    "not dead",
+    "not op_mini all"
   ],
   "devDependencies": {
     "@pmmmwh/react-refresh-webpack-plugin": "^0.5.15",
--- a/yarn.lock
+++ b/yarn.lock
@@ -3912,14 +3912,7 @@ __metadata:
   languageName: node
   linkType: hard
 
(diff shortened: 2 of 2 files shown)


gregorbg:
I do not understand why this lockfile error keeps happening. A local `yarn install` does not yield these results. Even `yarn install --immutable` does not throw any error for me.

Deleting the GitHub caches also did not help.

gregorbg:
Now the problem just magically resolved itself. Maybe it was related to caches after all...?

FinnIckler:
the diff does not slightly different for some reason

FinnIckler (approved):

PR #11004, opened 2025-03-10
Add barebones Chakra UI setup to NextJS

gregorbg:
Nothing exciting yet, just the core package setup. Pretty much everything is copy/pasta from https://www.chakra-ui.com/docs/get-started/frameworks/next-app

### DIFF
--- a/next-frontend/package.json
+++ b/next-frontend/package.json
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "dev": "next dev --turbopack",
+    "dev": "next dev",
     "build": "next build",
     "start": "next start",
     "lint": "next lint"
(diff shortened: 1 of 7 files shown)


next-frontend/package.json
```diff
@@ -3,18 +3,23 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "dev": "next dev --turbopack",
```
FinnIckler, line 6: I knew we would have to disable the experimental turbopack support sooner or later...

next-frontend/src/app/layout.tsx
```diff
@@ -25,9 +26,13 @@ export default function RootLayout({
   children: React.ReactNode;
 }>) {
   return (
-    <html lang="en">
+    <html suppressHydrationWarning>
```
FinnIckler, line 29: why do we need this?

gregorbg: Not a clever answer but: Because Chakra UI says so.
https://www.chakra-ui.com/docs/get-started/frameworks/next-app#setup-provider

PR #11008, opened 2025-03-11
Move timeout to ECS instance

FinnIckler:
This will not make our overall deployment slower as this is a maximum value, if the server stops before that, it will not affect anything.


gregorbg (approved):

FinnIckler:
I have run terraform. I assume it will still take a couple of deployments until all of our instances are replaced

PR #11009, opened 2025-03-11
Prevent organizer from setting event limit to 0

dunkOnIT:
Until this PR, an organizer who set and un-set event restrictions would still have the value of 0 in the database for events_per_registration_limit - and this would trigger an error when they tried to register. 

This PR:
- adds some tests which were missing against validations which already existed
- prevents non-nil events_per_registration_limit if event_restrictions is false
- InputNumber for events_per_registration limit in competition form
  - sets the minimum value to 1 
  - makes it nullable


gregorbg:
Input fixes and validations plus tests are definitely cool, but please also consider the place where the frontend check blew up at runtime! i.e., the number input and reg checker validations during registration submission should only execute the `>` comparison if this limit is enabled in the first place

dunkOnIT:
registration_checker updates to only enforce `events_per_registration_limit` if `event_restrictions`

lib/registrations/registration_checker.rb
```diff
@@ -113,7 +113,8 @@ def validate_create_events!(request, competition)
           event_ids.present? && competition.events_held?(event_ids)
 
         event_limit = competition.events_per_registration_limit
-        raise WcaExceptions::RegistrationError.new(:forbidden, Registrations::ErrorCodes::INVALID_EVENT_SELECTION) if event_limit.present? && event_ids.count > event_limit
+        raise WcaExceptions::RegistrationError.new(:forbidden, Registrations::ErrorCodes::INVALID_EVENT_SELECTION) if
+          competition.event_restrictions && event_limit.present? && event_ids.count > event_limit
```
gregorbg, line 117:
```suggestion
          competition.event_restrictions? && event_limit.present? && event_ids.count > event_limit
```

gregorbg (changes requested):

gregorbg (approved):

PR #11011, opened 2025-03-12
Add person DOB to anonymization dashboard

danieljames-dj:
This was a bug in the previous anonymization dashboard migration because of which the DOB of persons is not shown. I've fixed it now.

There was a similar case in another UI earlier, so I have attempted to reuse the logic.

### DIFF
--- a/app/controllers/tickets_controller.rb
+++ b/app/controllers/tickets_controller.rb
@@ -142,6 +142,8 @@ class TicketsController < ApplicationController
     user, person = user_and_person_from_params
     return if check_errors(user, person)
 
+    person_private_attributes = person&.private_attributes_for_user(current_user)
+
     user_anonymization_checks, user_message_args = user&.anonymization_checks_with_message_args
     person_anonymization_checks, person_message_args = person&.anonymization_checks_with_message_args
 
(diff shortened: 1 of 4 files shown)


app/controllers/tickets_controller.rb
```diff
@@ -154,7 +156,7 @@ def details_before_anonymization
 
     render json: {
       user: user,
-      person: person,
+      person: person&.serializable_hash(private_attributes: person_private_attributes),
```
gregorbg, line 159:
```suggestion
      person: person&.as_json(private_attributes: person_private_attributes),
```

gregorbg: Not a big difference, but more Rails-esque

app/models/person.rb
```diff
@@ -365,6 +365,16 @@ def anonymize
...
   end
 
+  def private_attributes_for_user(user)
+    if user&.wca_id == wca_id || user&.any_kind_of_delegate?
+      %w[dob]
+    elsif user&.can_admin_results?
+      %w[incorrect_wca_id_claim_count dob]
+    else
+      []
+    end
```
gregorbg, line 375: I think you can avoid a whole bunch of `&.` safe-navigation by just checking immediately if the user is nil: `return [] if user.nil?`

gregorbg (changes requested):

app/controllers/tickets_controller.rb
danieljames-dj: Okay changed.

app/models/person.rb
danieljames-dj: Okay changed.

gregorbg (approved):

PR #11015, opened 2025-03-12
Nextjs lint

FinnIckler:
Add linting to nextjs. Also added a separate next js github action that currently only checks lint


FinnIckler:
Note: stylistic is the new way of doing eslint rules like eslint-typescript, see https://eslint.style/guide/getting-started

PR #11016, opened 2025-03-12
Install rubocop plugins but don't turn any rules on yet

FinnIckler:
We will be turning these on in batches over the next days


FinnIckler:
Phew.. That was all

gregorbg (approved):
I'm fine with the code, although I didn't bother to check every single rule exclusion. Trusting you and the CI on this one!

PR #11019, opened 2025-03-12
Activate another three performance cops

FinnIckler:
I tried to keep the changeset to a similar size than the other one (30 lines) so this one has only three cops


FinnIckler:
Love that we already have test failures, at least proves that we are turning them on correctly in batches like this...
I assume this is because .sum is trying to coerce it into Integer instead of just calling plus so we can't use it for the solve times

PR #11028, opened 2025-03-13
Five small rails cops

FinnIckler:
Most of the changeset is moving all the status codes from numbers to symbols

@gregorbg What is your opinion on https://www.rubydoc.info/gems/rubocop-rails/2.5.2/RuboCop/Cop/Rails/EnumHash
I do kinda agree that it makes it less error prone

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -295,9 +295,6 @@ Rails/I18nLocaleAssignment:
 Rails/Blank:
   Enabled: false
 
-Rails/Output:
-  Enabled: false
-
 Rails/SquishedSQLHeredocs:
   Enabled: false
 
@@ -382,9 +379,6 @@ Rails/WhereEquals:
 Rails/CompactBlank:
   Enabled: false
 
-Rails/EnumHash:
-  Enabled: false
-
 Rails/ReadWriteAttribute:
   Enabled: false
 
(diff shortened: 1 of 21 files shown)


dunkOnIT (approved):
I'm trusting the cop's judgement a little bit on `RedundantSafeNavigation`, but otherwise looks reasonable!

FinnIckler:
I assume `!comments&.strip.present?` is fine because `nil.present?` is defined

gregorbg:
> I assume `!comments&.strip.present?` is fine because `nil.present?` is defined

Yes, exactly! You can verify this by literally calling `nil.present?` from a `bin/rails c` (and observe that it doesn't crash/raise)

> @gregorbg What is your opinion on https://www.rubydoc.info/gems/rubocop-rails/2.5.2/RuboCop/Cop/Rails/EnumHash
> I do kinda agree that it makes it less error prone

Hmm... I get the intention behind the cop, I'm just not sure how much sense it makes to redefine already defined enums. Can you keep it disabled until we've made up our minds? (Keeping two-three cops total on `disabled` won't do any harm I think)

gregorbg:
Thanks!
By the way the title of this PR could be something like a children's song, or perhaps a cringy comedy movie :joy:

FinnIckler:
I'm really running out of titles too

PR #11029, opened 2025-03-13
Activate Rspec format cops

FinnIckler:
This changeset looks daunting, but it only affects specs by:
- changing all `eq` to `be`
- changing all `to_not` to `not_to`
- adds some whitespace

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -465,9 +465,6 @@ RSpec/DescribeClass:
 RSpec/ContextWording:
   Enabled: false
 
-RSpec/BeEq:
-  Enabled: false
-
 RSpec/EmptyLineAfterHook:
   Enabled: false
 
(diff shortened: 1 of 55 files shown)


FinnIckler:
Ah sorry, it's not actually all instances. It just checks if it can use eq instead. see

```
RSpec/BeEq
Check for expectations where ‘be(…)` can replace `eq(…)`.

The ‘be` matcher compares by identity while the `eq` matcher compares using `==`. Booleans and nil can be compared by identity and therefore the `be` matcher is preferable as it is a more strict test.
```

gregorbg:
Yup! Duncan, your instinct is spot-on for cases like `my_array = [1,2,3]; expect(my_array).to be([1,2,3]);`. That would probably fail, but as Finn's quote clearly says the literal boolean `true` and `false` as well as `nil` are exempt because under the hood (in the Ruby interpreter) these all reference the **same** exact (internal) object

gregorbg (approved):

PR #11032, opened 2025-03-13
Activate SQL Squish cop

FinnIckler:
squish just removes all whitespace at the start and the end of the string https://apidock.com/rails/v7.1.3.4/String/squish
`Rails/Output` was already activated, it was just eaten in a merge


gregorbg (approved):

gregorbg:
Lucky SQL ain't Python... Or otherwise these whitespace changes would probably make our database explode

PR #11033, opened 2025-03-13
Rails pick, pluck and delegate cops

FinnIckler:
We might want to move all the delegate calls together to the top of the file? not sure

### DIFF
--- a/app/models/live_result.rb
+++ b/app/models/live_result.rb
@@ -33,9 +33,7 @@ class LiveResult < ApplicationRecord
     super(DEFAULT_SERIALIZE_OPTIONS.merge(options || {}))
   end
 
-  def event_id
-    event.id
-  end
+  delegate :id, to: :event, prefix: true
 
   def to_solve_time(field)
     SolveTime.new(event_id, field, send(field))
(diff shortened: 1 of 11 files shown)


app/models/live_result.rb
```diff
@@ -33,9 +33,7 @@ def serializable_hash(options = nil)
     super(DEFAULT_SERIALIZE_OPTIONS.merge(options || {}))
   end
 
-  def event_id
-    event.id
-  end
+  delegate :id, to: :event, prefix: true
```
gregorbg, line 36: Wow, this cop is really smart to detect that even with `prefix: true` :O

gregorbg:
> We might want to move all the delegate calls together to the top of the file? not sure

No hard feelings here, either. TBH I'm surprised that there isn't a cop for that :joy:
Or maybe there is, and you just haven't activated it yet?

FinnIckler:
> > We might want to move all the delegate calls together to the top of the file? not sure
> 
> No hard feelings here, either. TBH I'm surprised that there isn't a cop for that 😂 Or maybe there is, and you just haven't activated it yet?

There are two order related ones, so maybe

gregorbg (approved):

PR #11034, opened 2025-03-13
Activate Order cops

FinnIckler:
Just moves around some methods


gregorbg (commented):
One disadvantage of this approach is that it separates the hook like `validate :my_custom_method` from the definition `def my_custom_method`. Does the Styleguide say anything about how to best handle that?

FinnIckler:
It doesn't, but the Rails documentation all have it in the private method block apart from the definition

gregorbg (approved):

Issue #10916, opened 2025-02-25
Add note to delegate report template reminding delegates to exclude sensitive incidents and identifying information on WIC feedback incidents


weatherman223:
**Is your feature request related to a problem? Please describe.**

Delegates and the WIC have recently noticed an uptick in delegates including extremely sensitive incidents and identifying information in WCA competition reports when asking for WIC feedback. Since these reports are disseminated to hundreds of WCA volunteers and also have the potential to be leaked to the community, it seems necessary to include a note and reminder to prevent oversharing these incidents. 

**Describe the solution you'd like**

A solution that could potentially be supported would be to add a note to the delegate report template that reminds delegates to do the following:

1. Anonymize competitors information when reporting disiplinary incidents, and send a follow up email to WIC on the report thread with reports removed from CC with identifying details. 

2. Omit sensitive incidents from reports and email these separately to the WIC. 

**Describe alternatives you've considered**

A staff wide announcement could be used, but can be easily forgotten by WCA volunteers. A consistent note would serve better.

weatherman223:
A suggested note that could be added:

“For incidents requiring WIC feedback, please omit any identifying information from the incident being reports and follow up with the WIC with the competitor’s information after the report has been posted. Make sure you exclude “reports” from the cc field.

If the incident is sensitive (including but not limited to crimes, harassment, assault, etc), please exclude this from your report and email the WIC separately.”

slongbehn:
> “For incidents requiring WIC feedback, please omit any identifying information from the incident being reported and follow up with the WIC with the competitor’s information after the report has been posted. Make sure you exclude “reports” from the cc field.
> 
> If the incident is sensitive (including but not limited to crimes, harassment, assault, etc), please exclude this from your report and email the WIC separately.”

Not ready to ask delegates to fully omit information until WIC has better training/guidance, as well as better handling of a "watchlist" of sorts. Here's a suggested change for now:

> If the WIC incident is sensitive (including but not limited to crimes, harassment, assault, etc.), please exclude this from your report and **email the WIC separately**. For other WIC incidents, default to a brief, anonymized version of the incident and follow-up with the WIC privately. The WIC will still provide feedback on anonymized versions for general awareness.

Ideally I'd like to see this as a pop-up if the WIC feedback box is checked, but otherwise please add text above that.

![Image](https://github.com/user-attachments/assets/563ca2f7-1bb0-4488-8d0c-f5b44257b079)

[...]

PR #11039, opened 2025-03-14
WIC warning for sensitive incident

pranav-027:
fixes: https://github.com/thewca/worldcubeassociation.org/issues/10916


config/locales/en.yml
```diff
@@ -501,7 +501,7 @@ en:
         wrc_feedback_requested: ""
         wrc_incidents: "List of numbers referencing any incidents that you would like the WRC's feedback on."
         wic_feedback_requested: ""
-        wic_incidents: "List of numbers referencing any incidents that you would like the WIC's feedback on."
+        wic_incidents: "List of numbers referencing any incidents that you would like the WIC's feedback on. \nIf the WIC incident is sensitive (including but not limited to crimes, harassment, assault, etc.), please exclude this from your report and email the WIC separately. For other WIC incidents, default to a brief, anonymized version of the incident and follow-up with the WIC privately with further details. The WIC will still provide feedback on anonymized versions for general awareness. For all WIC incidents involving a WCA Volunteer, please exclude all details from the report, and contact the WIC privately."
```
gregorbg, line 504: Please don't use `\n` in localized strings. If anything, you should probably use `<br/>`, but I'm not sure whether a line break is crucial here.

gregorbg: Also, if you do end up using HTML tags inside the string, you need to rename it to `wic_incidents_html` so that Rails renders it correctly (if a translation key does not end with `_html`, then it will print the tag as if it were simple plain text)

gregorbg: Also, the new parahrapgh that you're adding here is exactly the same as in the JS alert box above. If you move this message into a separate translation key, you could inject it into the HTML/JS above.

app/views/delegate_reports/edit.html.erb
```diff
@@ -64,3 +64,16 @@
...
 <% end %>
+
+
+<script>
+  document.addEventListener("DOMContentLoaded", function () {
+    const checkbox = document.getElementById("wic_feedback_checkbox");
+
+    checkbox.addEventListener("change", function () {
+      if (this.checked) {
+        alert("If the WIC incident is sensitive (including but not limited to crimes, harassment, assault, etc.), please exclude this from your report and email the WIC separately. For other WIC incidents, default to a brief, anonymized version of the incident and follow-up with the WIC privately with further details. The WIC will still provide feedback on anonymized versions for general awareness. For all WIC incidents involving a WCA Volunteer, please exclude all details from the report, and contact the WIC privately.");
```
gregorbg, line 75: This should use a localized version: `alert("<%= t('foo.bar.wic_incidents_warning') %>");`

config/locales/en.yml
gregorbg: Also, I'm not sure whether it's necessary to have the message as part of the alert and as part of the bottom description. If you really want both, you can introduce the alert text as a separate translation and then reference it here: `List of numbers referencing any incidents that you would like the WIC's feedback on. <strong>%{wic_incident_alert}</strong>`

If you do this, then the automated "guess the correct key" mechanism for `simple_form` won't work anymore, because you have to add the `wic_incident_alert` yourself. You can easily override any form hint like so: `<%= f.input :some_field, hint: I18n.t('my.custom.key') %>`

app/views/delegate_reports/edit.html.erb
gregorbg: Also, if you use `confirm` instead of `alert`, you will have a return value (coming from the `confirm` function itself) that you can use to override the `checked` property of the checkbox. That is, if somebody clicks "Cancel" on the alert window that pops up, it never gets checked :upside_down_face:

pranav-027: Yeah, nice idea. Implemented!

config/locales/en.yml
```diff
@@ -501,7 +501,8 @@ en:
         wrc_feedback_requested: ""
         wrc_incidents: "List of numbers referencing any incidents that you would like the WRC's feedback on."
         wic_feedback_requested: ""
-        wic_incidents: "List of numbers referencing any incidents that you would like the WIC's feedback on."
+        wic_incidents: "List of numbers referencing any incidents that you would like the WIC's feedback on. <strong>%{wic_incident_alert}</strong>"
```
gregorbg, line 504: I am still pretty sure you need to call this `wic_incidents_html` (both here and in the code where you call `I18n.t('...')`). Otherwise the HTML renderer will just print the literal `<strong>` string as readable text to your screen, without treating it as parseable HTML

gregorbg (changes requested):

pranav-027:
@gregorbg, here is a small demo

https://github.com/user-attachments/assets/0b10bfe1-9963-411f-8889-78373bae3002

config/locales/en.yml
gregorbg: Hm :thinking: I wonder why it works though, normally the `_html` convention is "baked into" the Rails I18N engine...

gregorbg: Ah, it's because you're manually doing `.html_safe` above. Calling the key `foo_bar_html` would do that automagically for you :wink:

gregorbg (changes requested):

gregorbg:
Huh, it shouldn't be like that... Let me try this out for myself for a moment

gregorbg:
> Am I missing something here?

Yes, but it's very subtle. Also took me a while to find. TL:DR There is a difference between `I18n.t('my.key')` and `t('my.key')`. Only the latter respects the `_html` suffix.

Background: `I18n.t` is the abstract translation engine that is available everywhere, even in controllers, models, lib code, wherever you want basically. But inside `.html.erb` views, you also specifically have the `t()` helper (**without** the `I18n` prefix!) that handles additional, view-related things like HTML markup for you.

So adding `_html` to the key as suffix and then also changing from `I18n.t` to just using `t` does the trick.

pranav-027:
@gregorbg  done!!

gregorbg (approved):

PR #11043, opened 2025-03-14
Activate Migration Cops

FinnIckler:
This touches every migration file, but this was how gitlab did this 8 years ago https://gitlab.com/gitlab-org/gitlab-foss/-/merge_requests/4559

And I do agree that we don't need to run any formatting cops on old migrations

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -287,24 +287,6 @@ Rails/PluralizationGrammar:
 Rails/I18nLocaleAssignment:
   Enabled: false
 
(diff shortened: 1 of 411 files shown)


gregorbg:
I agree on the basic principles. Good job spotting that GitLab PR!

However, is the extra empty line between the two "magic comments" actually necessary?

FinnIckler:
let me check

gregorbg (approved):

PR #11044, opened 2025-03-14
Even more rails auto cops

FinnIckler:
Only a couple more remaining


gregorbg:
What is the difference between `flash` and `flash.now`?

FinnIckler:
flash should only be used when redirecting, flash.now should be used when rendering. 

from the docs
` The flash is typically used for setting messages (e.g. notices and alerts) in a controller action before redirecting to an action that displays the message to the user.`

`[flash.now](https://api.rubyonrails.org/v8.0.2/classes/ActionDispatch/Flash/FlashHash.html#method-i-now) is used to make the flash values available in the same request`

FinnIckler:
Interesting that it didn't figure it out even though the method ends in
    `redirect_to competition_register_path(competition_id)`

app/controllers/users_controller.rb
```diff
@@ -105,7 +105,7 @@ def authenticate_user_for_sensitive_edit
       session[:last_authenticated_at] = Time.now
     end
     on_failure = -> do
-      flash[:danger] = I18n.t("users.edit.sensitive.failure")
+      flash.now[:danger] = I18n.t("users.edit.sensitive.failure")
```
gregorbg, line 108: This is also a redirect method, I think

gregorbg (commented):
Looking at how many exceptions we need, you should probably keep `Rails/ActionControllerFlashBeforeRender` disabled and put it together with the "hashes vs arrays for enums" rule as something we should deal with later on

gregorbg (approved):

PR #11046, opened 2025-03-14
Activate FactoryBot Syntax Cops

FinnIckler:
Adds Parentheses to every FactoryBot method and removes the need to do `FactoryBot::` for every call


FinnIckler:
Looks like all these `sign_in { create(:admin) }` need to be letted before

FinnIckler:
mh.. just letting it didn't work. I am not sure if that is just because we aren't following best practices with that sign_in function or what

dunkOnIT:
Replaced by #11303 (opened a new PR because I wanted to enable the cops from scratch for testing/troubleshooting purposes)

PR #11048, opened 2025-03-14
Small manual cops

FinnIckler:
Starting on the cops that need some manual attention. 
I removed 
  `before_action -> { redirect_to_root_unless_user(:has_permission?, 'can_access_panels', :staff) }, only: [:staff]`
and
`before_action :current_user_can_admin_finances!, only: [:panel_wfc]`

because I think those controller actions don't exist anymore and they were triggering the only have before_actions for controller actions that are defined in the controller cop


spec/support/locale_reset.rb
gregorbg, line 1: Good candidate for file-wide exclusion in `rubocop.yml`

app/controllers/regulations_controller.rb
gregorbg, line 1: Good candidate for file-wide exclusion in `rubocop.yml`

gregorbg (changes requested):
If the whole purpose of a file is a helper to render inline, you can just list that file as exclusion:
```yml
Rails/ObscureRule:
    Exclude:
        - 'app/helpers/my_inline_renderer.rb'
```

gregorbg (commented):
The two `between` instances have already been fixed on `main` a couple of hours ago, did you merge?

FinnIckler:
I thought I did. let me do again

FinnIckler:
now those changes are away

PR #11050, opened 2025-03-14
Turn on automated cops from rubocop_todo

FinnIckler:
I have run out of automated cops from my main list, but these were still easy pickings

### DIFF
--- a/app/models/competition.rb
+++ b/app/models/competition.rb
@@ -1653,7 +1653,7 @@ class Competition < ApplicationRecord
   )
 
   # rubocop:disable Lint/StructNewOverride
-  # this does overwrite sort_by, but as it's never used in an array, it should be fine
+  # this does overwrite sort_by because the frontend relies on the field, but as it's never used in an array, it should be fine
   PsychSheet = Struct.new(
     :sorted_rankings,
     :sort_by,
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -24,6 +24,11 @@ AllCops:
     - 'db/results_dump_schema.rb'
     - '.git-hooks/**/*'
 
+Style/StringConcatenation:
+  # We need to use + here because it's a custom operator TODO: At least until we only render from NExt
+  Exclude:
+  - 'app/helpers/application_helper.rb'
+
 Bundler/OrderedGems:
   Enabled: false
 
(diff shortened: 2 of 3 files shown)


.git-hooks/pre_commit/illegal_strings.rb
```diff
@@ -10,7 +10,7 @@ def run
         "\t" => "tab",
         "\r" => "carriage return",
         "\uFEFF" => "byte order marker (BOM)",
-        ("WCA " + "id") => "We prefer 'WCA ID', see https://github.com/thewca/worldcubeassociation.org/issues/268",
+        "WCA id" => "We prefer 'WCA ID', see https://github.com/thewca/worldcubeassociation.org/issues/268",
```
gregorbg, line 13: I think this will fail because it lints itself

app/controllers/api/v0/competitions_controller.rb
```diff
@@ -205,8 +205,8 @@ def update_wcif
 
   private def require_scope!(scope)
     require_user!
-    if current_api_user # If we deal with an OAuth user then check the scopes.
-      raise WcaExceptions::BadApiParameter.new("Missing required scope '#{scope}'") unless doorkeeper_token.scopes.include?(scope)
+    if current_api_user && doorkeeper_token.scopes.exclude?(scope) # If we deal with an OAuth user then check the scopes.
```
gregorbg, line 208: Not sure if Doorkeeper token has an `exclude` method because it's not an array

app/models/competition.rb
```diff
@@ -1658,12 +1652,15 @@ def organizers_or_delegates
     keyword_init: true,
   )
 
+  # rubocop:disable Lint/StructNewOverride
+  # this does overwrite sort_by, but as it's never used in an array, it should be fine
```
gregorbg, line 1656: And also the frontend relies on the field being called `sort_by`

.git-hooks/pre_commit/illegal_strings.rb
FinnIckler: yeah, I laughed out loud when I saw that

app/controllers/api/v0/competitions_controller.rb
FinnIckler: it does `include Enumerable` though?

gregorbg (approved):

gregorbg (approved):

Issue #8299, opened 2023-09-10
Automated actions when a user is banned


dunkOnIT:
Upon being banned, users should be de-registered from any upcoming competitions taking place before the end of their ban period, and delegates of the competition + the competitor should be emailed informing them of this change - this would save WDC some work.

Email thread reference: "Scope of Bans Applied by WDC"

dunkOnIT:
This can be achieved with the Registration Service API once it is launched.

slongbehn:
@dunkOnIT 

I would like the changes not to be emailed to the competitor, as the WIC needs to be responsible for communicating this to ensure it's carefully communicated.

Emailing the listed delegates would still be helpful, with WIC in CC (absolutely NO organizers that aren't delegates). It would be good to add something indicating that it's urgent in the email subject, and something like the following in the body:

> [Competitor] has been banned until [date]. This message indicates the competitor is registered for [competition]. In order to initiate the ban, the Integrity Committee asks you to urgently delete their registration, regardless of status or the deadline to delete registrations. 
>
> If payment has been made, a full refund is recommended, however it is ultimately up to the organizational team to provide a full refund or follow the current refund policy on the website. 
>
> The WIC will be in touch with relevant parties soon with details of this sanction. Please reply to this email if you have any questions in the meantime.

pranav-027:
@slongbehn 

![Image](https://github.com/user-attachments/assets/fbb82f3d-cde0-44fb-9c2e-a3048af25cbc)

Email will look like this. Lmk if subject and email is fine

CarterKoala:
Maybe nitpicky, but I'm not a fan of the use of "till" here (as it's generally used in more informal contexts), I would prefer "until" instead.

pranav-027:
> Maybe nitpicky, but I'm not a fan of the use of "till" here (as it's generally used in more informal contexts), I would prefer "until" instead.

changed

slongbehn:
Thanks for pinging! Just want to check - this only goes to the delegates right? The organizers cannot receive this email.

Also, it might be good to tell them whether the registration needs to be "Cancelled" or "Rejected". I'm not sure if it matters in order to enact a ban, but I expect that to be a question. Could you double check if it matters? They should probably be rejected so they can't re-register.

pranav-027:
> Thanks for pinging! Just want to check - this only goes to the delegates right? The organizers cannot receive this email.

[...]

PR #11051, opened 2025-03-14
Automatically delete regs of upcoming comps after banning user

pranav-027:
fixes: https://github.com/thewca/worldcubeassociation.org/issues/8299

image

### DIFF
--- a/app/controllers/api/v0/user_roles_controller.rb
+++ b/app/controllers/api/v0/user_roles_controller.rb
@@ -91,14 +91,15 @@ class Api::V0::UserRolesController < Api::V0::ApiController
       user = User.find(user_id)
       ban_reason = params[:banReason]
       scope = params[:scope]
-      upcoming_comps_for_user = user.competitions_registered_for.not_over.merge(Registration.not_cancelled)
+      upcoming_registrations_for_user = user.registrations.not_cancelled
       if end_date.present?
-        upcoming_comps_for_user = upcoming_comps_for_user.between_dates(Date.today, end_date)
+        upcoming_registrations_for_user = upcoming_registrations_for_user.joins(:competition).merge(Competition.between_dates(Date.today, end_date))
       end
-      unless upcoming_comps_for_user.empty?
-        return render status: :unprocessable_entity, json: {
-          error: "The user has upcoming competitions: #{upcoming_comps_for_user.pluck(:id).join(', ')}. Before banning the user, make sure their registrations are deleted.",
-        }
+      unless upcoming_registrations_for_user.empty?
+        upcoming_registrations_for_user.each do |registration|
+          registration.update!(competing_status: Registrations::Helper::STATUS_CANCELLED)
+          RegistrationsMailer.notify_delegates_of_registration_deletion_of_banned_competitor(registration, end_date).deliver_later
+        end
       end
     end
 
--- a/app/mailers/registrations_mailer.rb
+++ b/app/mailers/registrations_mailer.rb
@@ -92,4 +92,18 @@ class RegistrationsMailer < ApplicationMailer
       )
     end
   end
+
+  def notify_delegates_of_registration_deletion_of_banned_competitor(registration, end_date)
+    @registration = registration
+    @end_date = end_date
+    to = registration.competition.competition_delegates.map { |x| x.user.email }
+    unless to.empty?
+      mail(
+        to: to,
+        cc: UserGroup.teams_committees_group_wic.metadata.email,
+        reply_to: UserGroup.teams_committees_group_wic.metadata.email,
+        subject: "Competitor Ban Notice – Registration Deleted for #{registration.competition.name}",
+      )
+    end
+  end
 end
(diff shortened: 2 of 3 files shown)


app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -91,14 +91,15 @@ def create
...
-        upcoming_comps_for_user = upcoming_comps_for_user.between_dates(Date.today, end_date)
+        upcoming_registrations_for_user = upcoming_registrations_for_user.joins(:competition).merge(Competition.between_dates(Date.today, end_date))
       end
-      unless upcoming_comps_for_user.empty?
-        return render status: :unprocessable_entity, json: {
-          error: "The user has upcoming competitions: #{upcoming_comps_for_user.pluck(:id).join(', ')}. Before banning the user, make sure their registrations are deleted.",
-        }
+      unless upcoming_registrations_for_user.empty?
+        upcoming_registrations_for_user.each do |registration|
+          registration.update!(competing_status: Registrations::Helper::STATUS_CANCELLED)
```
gregorbg, line 100: We actually have `STATUS_REJECTED` for cases like this: `cancelled` would allow the competitor themselves to re-register again, whereas `rejected` does not.

gregorbg (approved):

gregorbg (changes requested):
Whoops wrong button I meant to Request Changes on my previous review. Sorry!

app/controllers/api/v0/user_roles_controller.rb
pranav-027: Done!

dunkOnIT:
We should also wait for WIC signoff before merging this

pranav-027:
Hello @gregorbg , @dunkOnIT this is approved by @slongbehn

spec/controllers/api/v0/user_roles_controller_spec.rb
gregorbg, line 1: You should probably add more tests.
1. To check that the previous registration has actually been rejected
2. To check that the emails are getting sent out

app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -76,12 +76,13 @@ def create
...
-      upcoming_comps_for_user = upcoming_comps_for_user.between_dates(Date.today, end_date) if end_date.present?
-      unless upcoming_comps_for_user.empty?
-        return render status: :unprocessable_entity, json: {
-          error: "The user has upcoming competitions: #{upcoming_comps_for_user.pluck(:id).join(', ')}. Before banning the user, make sure their registrations are deleted.",
-        }
+      upcoming_registrations_for_user = user.registrations.not_cancelled
+      upcoming_registrations_for_user = upcoming_registrations_for_user.joins(:competition).merge(Competition.between_dates(Date.today, end_date)) if end_date.present?
+      unless upcoming_registrations_for_user.empty?
+        upcoming_registrations_for_user.each do |registration|
+          registration.update!(competing_status: Registrations::Helper::STATUS_REJECTED)
```
gregorbg, line 83: Using `update!` here can be potentially dangerous because it will throw an error if the updated registration is not considered valid. From the top of my head, I can easily think about edge cases like "The self-cancellation policy does not allow removing paid registrations, so because this registration is paid already, it will be considered invalid when trying to set the status to `rejected`".
And then the error would break the whole validation flow and abort the controller method early with an unhandled exception.
- This can have dangerous side-effects when trying to run the ban again after the error has been fixed (do we then have two ban roles, because the first one was already created before the error was raised?)
- There is no "nice" error reporting for the user, the page will just straight-out crash.

The "cheapest" solution would probably be to just wrap this `each` loop into a transaction. But then the concern about the rest of the method being duplicated would still exist.

Another possibility would be to save and explicitly use `skip_validations`, but I'm not sure whether I like that. We may have to do that though because WIC has "absolute power" to ban competitors no matter if that action of banning them would normally be considered a "valid registration change" or not. But I want to consider this only as a last resort.

gregorbg (changes requested):
Thanks for the changes, I'm very cautious about hard-updating the registrations. Details are in my code comment in `user_roles_controller`.

app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -115,7 +122,7 @@ def create
         start_date: Date.today,
         end_date: end_date,
         metadata: metadata,
-      )
+        )
```
danieljames-dj: Is this accidental change?

app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -79,9 +79,16 @@ def create
...
+        ActiveRecord::Base.transaction do
+          upcoming_comps_for_user.each do |registration|
+            begin
+              registration.update!(competing_status: Registrations::Helper::STATUS_REJECTED)
+              RegistrationsMailer.notify_delegates_of_registration_deletion_of_banned_competitor(registration, end_date).deliver_later
+            rescue ActiveRecord::RecordInvalid => e
+              raise ActiveRecord::Rollback, "Registration update failed: #{e.message}"
+            end
+          end
+        end
```
danieljames-dj: How about moving these to a separate private method?

PR #11052, opened 2025-03-15
Restore continent-filtering by removing reverting rubocop change

dunkOnIT:


gregorbg:
Bit busy today outside of work, but I took the liberty of quickly "commandeering" this PR.
The issue here was that the SQL builder as suggested by RuboCop tried to search the `continent` field directly in the `Competitions` table, which does indeed not exist. But Rails does allow "nesting" since quite a while, which fixes the issue more adequately.

Rule of thumb: Disabling RuboCop is only a (very) last resort. If there's a suggestion, chances are we can follow it.

gregorbg (approved):
Duncan, depending on how available you are please feel free to add a test that queries compettions through this API and filters by continent.

If you're also busy, then by all means feel free to merge this as-is and consider it a hotfix :)

dunkOnIT:
Merging and deploying now to hotfix, but will add a test against the API in a couple of hours!

Issue #8312, opened 2023-09-15
View upcoming registrations of a user in anonymize panel


danieljames-dj:
In anonymize admin panel (https://www.worldcubeassociation.org/admin/anonymize_person), it will be good if we have a link that will list the upcoming registrations of the user. This will help WRT to confirm that the person that is going to be anonymized is not registered in any upcoming competitions.

The logic behing [my competitions](https://www.worldcubeassociation.org/competitions/mine) can be reused somewhere I guess.

pranav-027:
Hello @danieljames-dj,

![Image](https://github.com/user-attachments/assets/fd1cde2d-bbbb-4c40-b4b1-33a227ac8e91)

I have implemented this on step 3. Let me know if text is correct.

maxidragon:
> Hello [@danieljames-dj](https://github.com/danieljames-dj),
> 
> ![Image](https://github.com/user-attachments/assets/fd1cde2d-bbbb-4c40-b4b1-33a227ac8e91)
> 
> I have implemented this on step 3. Let me know if text is correct.

nit: "Take" shouldn't be capitalized IMO

PR #11054, opened 2025-03-15
List upcoming registrations of a person in anonymize panel

pranav-027:
Fixes: https://github.com/thewca/worldcubeassociation.org/issues/8312

This has been added on step 3
image


app/webpacker/components/Tickets/TicketWorkbenches/AnonymizationTicketWorkbenchForWrt/index.jsx
```diff
@@ -227,6 +227,19 @@ function ActionItemContent({ actionItem, messageArgs }) {
           </List>
         </>
       );
+    case 'person_has_upcoming_registered_competitions':
+      return (
+        <>
+          This person has upcoming registered competitions. Please Take care of it.
```
maxidragon, line 233:
```suggestion
          This person has upcoming registered competitions. Please take care of it.
```

pranav-027: Done

app/models/person.rb
```diff
@@ -301,6 +301,7 @@ def best_averages_by(target_date)
   def anonymization_checks_with_message_args
     recent_competitions_3_months = competitions&.select { |c| c.start_date > (Date.today - 3.month) }
     competitions_with_external_website = competitions&.select { |c| c.external_website.present? }
+    upcoming_registered_competitions = user.competitions_registered_for.not_over.merge(Registration.not_cancelled)
```
danieljames-dj, line 304: Registrations belong to user, not person. This way we might miss registrations of newcomer users who requested anonymization. Can you please move this to similar method in `user.rb`?

app/models/person.rb
```diff
@@ -309,12 +310,14 @@ def anonymization_checks_with_message_args
         person_competed_in_last_3_months: recent_competitions_3_months&.any?,
         competitions_with_external_website: competitions_with_external_website&.any?,
         recent_competitions_data_to_be_removed_wca_live: recent_competitions_3_months&.any?,
+        person_has_upcoming_registered_competitions: upcoming_registered_competitions.any?,
```
danieljames-dj, line 313: As per my other comment, can we rename this to `user_has_...`?

danieljames-dj (commented):
Thanks @pranav-027 for working on this. I've left a comment, other than that it looks good.

app/models/person.rb
pranav-027: Done

app/models/person.rb
pranav-027: Done

danieljames-dj (approved):
LGTM

app/models/user.rb
```diff
@@ -1500,6 +1500,7 @@ def location
   end
 
   def anonymization_checks_with_message_args
+    upcoming_registered_competitions = competitions_registered_for.not_over.merge(Registration.not_cancelled)
```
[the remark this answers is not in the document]
pranav-027: `not_cancelled` is defined as `scope :not_cancelled, -> { where.not(competing_status: 'cancelled') }` 

This will only exclude the `cancelled` status, meaning `rejected` will still be included.

Am I missing something here?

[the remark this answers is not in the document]
pranav-027: Ping @gregorbg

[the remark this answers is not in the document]
gregorbg: Well it is my understanding that `rejected` registrations count as "deleted" conceptually, so you don't need to take care of them.

My point with the original comment was that as of now, you're only excluding `cancelled` status specifically. Meaning `rejected` would still be included and displayed as "Hey this is something you need to take care of". And in reality, `rejected` is just as good as `cancelled`, i.e. you don't need to take care of it.

But maybe I'm missing something and you actually _do_ need to take care of `rejected` registrations...?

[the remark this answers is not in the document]
gregorbg: Well since the term `cancelled` refers to a very specific registration status, changing the scope `not_cancelled` to expand beyond the `cancelled` itself definitely doesn't make sense at all :sweat_smile: 

So your current approach should be fine.

app/models/user.rb
```diff
@@ -1448,10 +1449,12 @@ def anonymization_checks_with_message_args
         user_banned_in_past: banned_in_past?,
         user_may_have_forum_account: true,
         user_has_active_oauth_access_grants: access_grants.any?,
+        user_has_upcoming_registered_competitions: upcoming_registered_competitions.any?,
       },
       {
         access_grants: access_grants,
         oauth_applications: oauth_applications,
+        upcoming_registered_competitions: upcoming_registered_competitions,
```
gregorbg, line 1457: Come to think of it... I believe this is using the default serializer for the competition, which includes a **LOT** of unnecessary extra data (venue information, who are the Delegates, etc...) in your use case.

Please change it so that only the actually necessary information (name, ID) is serialized

gregorbg (changes requested):

app/models/user.rb
pranav-027: Done!

gregorbg (approved):

PR #11079, opened 2025-03-17
Use more Chakra components on dummy pages

gregorbg:
Just a little bit of "extended cleanup"


FinnIckler (approved):
The js tests already passed...

PR #11081, opened 2025-03-17
Reexamine disabled cops part 1

FinnIckler:
Should we prepend the unused variables with _ or just deleted them when they are not used?


gregorbg (commented):
I like the `_` prefix, because it works no matter how many or which variable is being (un)unsed. The "drop it" approach only works if the unused variable doesn't have other following (used) variable declarations, and it could seriously break if you drop an unused variable in front of a used one.

I am however very sceptical about the `alias`. It seems like RuboCop corrects it to be `alias original_method new_method`, which really doesn't make sense to me because `new_method` doesn't exist yet?! I mean, the reason why it was the symbol `:new_method` before is because it uses the symbol as a _name_ to declare some new alias, but now it makes a call as though `new_method` already existed?! Is this some Ruby interpreter exception/hack?!

gregorbg:
Just checked and you can indeed run this without symbols, as in `alias new_method_name existing_method_name`. I think that's wildly confusing and you should configure the cop to prefer `alias_method`: https://docs.rubocop.org/rubocop/cops_style.html#enforcedstyle_-prefer_alias_method-stylealias

gregorbg (approved):

PR #11083, opened 2025-03-17
Activate Style/GuardClause

FinnIckler:
this is also something we usually point out in a review. But it is a pretty big changeset.

Rubocop documentation says:
```
Autocorrect works in most cases except with if-else statements that contain logical operators such as `foo || raise('exception')`
```
which I don't think we do?

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -82,9 +82,6 @@ Style/TrailingCommaInHashLiteral:
   # non-empty array and hash literals.
   EnforcedStyleForMultiline: comma
 
-Style/GuardClause:
-  Enabled: false
-
 Metrics/AbcSize:
   Enabled: false
 
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -220,9 +220,6 @@ Style/YodaCondition:
 Style/FormatStringToken:
   Enabled: false
 
-Layout/EmptyLineAfterGuardClause:
-  Enabled: false
-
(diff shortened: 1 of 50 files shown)


FinnIckler:
ok my manual fixes seems to have not worked, as it broke the tests. maybe @gregorbg you have an idea how to fix this or I will just wrap it in a rubocop disable

gregorbg:
Abra kadabra :magic_wand:

FinnIckler:
even with -a it produces the same changeset

gregorbg (changes requested):
I added some sample comment about a "popular" pattern generated by the cop's autocorrect.

Instead of a simple `do_something if condition` there's a lot of
```
return unless condition
do_something
```

This happens because before this cop was turned on, the code looked like
```
if condition
  do_something
end
```

And the cop auto-corrects that into the two-liner. Is there any option to turn on guard clauses as postfix?

FinnIckler:
I think we will get that benefit if we turn on this first https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/IfUnlessModifier

gregorbg:
In that case be my guest ;)

FinnIckler:
Activating the cops in this order cut down the changes by a lot. I also activated Layout/EmptyLineAfterGuardClause for better readability

gregorbg (approved):

PR #11084, opened 2025-03-17
Move comment validations out of registration checker

FinnIckler:
Using the same methods as before. Let's see if this is all correct

### DIFF
--- a/app/webpacker/components/RegistrationsV2/Register/RegistrationOverview.jsx
+++ b/app/webpacker/components/RegistrationsV2/Register/RegistrationOverview.jsx
@@ -107,7 +107,7 @@ export default function RegistrationOverview({
                 {I18n.t('activerecord.attributes.registration.comments')}
                 :
               </List.Header>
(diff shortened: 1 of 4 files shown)


app/webpacker/components/RegistrationsV2/Register/RegistrationOverview.jsx
```diff
@@ -107,7 +107,7 @@ export default function RegistrationOverview({
                 {I18n.t('activerecord.attributes.registration.comments')}
                 :
               </List.Header>
-              {registration.competing.comment.length > 0 ? registration.competing.comment : I18n.t('competitions.schedule.rooms_panel.none')}
+              {registration.competing.comment?.length > 0 ? registration.competing.comment : I18n.t('competitions.schedule.rooms_panel.none')}
```
gregorbg, line 110: This frontend change is necessary because we're adding `strip_attributes` to the backend in this PR, and `strip_attributes` turns "blank-only" values into `nil` in Ruby (which gets serialized as `null` to the frontend)

gregorbg (approved):

PR #11086, opened 2025-03-18
Disable image upload in tickets comment

danieljames-dj:
Currently in tickets, while adding a new comment we can upload image. But this is not a necessary feature, and allows user to upload documents.

Recently there was an incident where WRT asked the requestor via email to submit their ID proof, and what WRT meant is to submit it as a reply to the email. But they thought of doing it via tickets, and added a comment there. (The comment looks like the user has uploaded an image, but we were not able to see the image in the comment anyway. Mostly it's because we are just printing the comment body, I didn't spend time to investigate this, maybe I'll try it later).

One approach to tackle this was to add an option in existing `MarkdownEditor`, but I feel most of the features in `MarkdownEditor` is not meant for general use-case. So I thought instead of using `MarkdownEditor`, I'll directly use `SimpleMdeReact` ([official documentation link](https://www.npmjs.com/package/react-simplemde-editor#controlled-usage)).

### DIFF
--- a/app/webpacker/components/wca/FormBuilder/input/MarkdownEditor.js
+++ b/app/webpacker/components/wca/FormBuilder/input/MarkdownEditor.js
@@ -3,6 +3,8 @@ import SimpleMDE from 'react-simplemde-editor';
 import { fetchWithAuthenticityToken } from '../../../../lib/requests/fetchWithAuthenticityToken';
 import 'easymde/dist/easymde.min.css';
 
+const DEFAULT_EDITOR_OPTIONS = { enableImageUpload: true };
+
 function insertText(editor, markup, promptText) {
   const cm = editor.codemirror;
 
@@ -28,7 +30,7 @@ function insertText(editor, markup, promptText) {
   cm.focus();
 }
 
-function getOptions(disabled) {
+function getOptions(editorOptions) {
   const table = {
     name: 'table-custom',
     action: '\n\n| Column 1 | Column 2 | Column 3 |\n| -------- | -------- | -------- |\n| Text     | Text      | Text     |\n\n',
@@ -69,16 +71,22 @@ function getOptions(disabled) {
 
   const textFormattings = ['bold', 'italic', 'heading'];
   const textStructures = ['quote', 'unordered-list', 'ordered-list', table];
-  const uploadsAndInserts = ['link', 'upload-image', map, youtube];
+  const inserts = ['link', map, youtube];
+  const uploadImageTool = editorOptions?.enableImageUpload ? ['upload-image'] : [];
   const previews = ['preview', 'side-by-side', 'fullscreen'];
   const helps = ['guide'];
   const toolbar = [
     ...textFormattings,
     '|', ...textStructures,
-    '|', ...uploadsAndInserts,
+    '|', ...inserts, ...uploadImageTool,
     '|', ...previews,
     '|', ...helps,
   ];
+
+  const imageOptions = editorOptions.enableImageUpload ? {
+    status: ['upload-image'],
+    uploadImage: true,
+  } : {};
 
   function previewRender(plainText, preview) {
     const previewTarget = preview;
@@ -100,8 +108,8 @@ function getOptions(disabled) {
     spellChecker: false,
     promptURLs: true,
     previewRender,
-    status: ['upload-image'],
-    uploadImage: true,
+    status: [],
+    uploadImage: false,
     async imageUploadFunction(file, onSuccess, onError) {
       const formData = new FormData();
       formData.append('image', file);
@@ -113,7 +121,7 @@ function getOptions(disabled) {
         onError(e);
       }
     },
-    disabled,
+    ...imageOptions,
   };
 }
 
@@ -121,8 +129,9 @@ export default function MarkdownEditor({
   id,
   value,
   onChange,
+  editorOptions = DEFAULT_EDITOR_OPTIONS,
 }) {
-  const options = useMemo(() => getOptions(false), []);
+  const options = useMemo(() => getOptions(editorOptions), [editorOptions]);
   const mdChange = useCallback((text) => onChange(null, { value: text }), [onChange]);
 
   return (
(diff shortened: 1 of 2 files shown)


gregorbg (changes requested):
It should be an easy exercise for you to add an `imageUpload` boolean prop to the existing `MarkdownEditor` and configure simpleMde accordingly. If you scroll through the existing `MarkdownEditor.js` file, I'm sure you will spot the place where picture upload is enabled/disabled pretty quickly

danieljames-dj:
Thanks @gregorbg. As mentioned in the last paragraph of the description, I've already considered that - but I was not sure which would be an ideal way. I've anyway implemented it now.

One info: `getOptions` method currently expects one parameter `disabled`. I need to add one more parameter for upload. But I feel this will lead to add more and more parameters in future. So I thought adding all the options as an object named `editorOptions` to `MarkdownEditor`. Also, there is no prop named `disabled` within `SimpleMDE`, and I tested by setting `disabled` as `true`. The editor was still working. Since the `disabled` is not used anywhere, I've just removed it for now.

One question: Currently the default editor option is `{ enableImageUpload: true }`. But do you think it should be false by default, and manually set true at every usages?

gregorbg:
I think in order to avoid breaking existing pages you should leave the default as `true`

app/webpacker/components/wca/FormBuilder/input/MarkdownEditor.js
```diff
@@ -100,8 +108,8 @@ function getOptions(disabled) {
     spellChecker: false,
     promptURLs: true,
     previewRender,
-    status: ['upload-image'],
-    uploadImage: true,
+    status: [],
+    uploadImage: false,
```
gregorbg, line 112: You can directly pass `editorOptions.enableImageUpload` here

app/webpacker/components/wca/FormBuilder/input/MarkdownEditor.js
```diff
@@ -100,8 +108,8 @@ function getOptions(disabled) {
     spellChecker: false,
     promptURLs: true,
     previewRender,
-    status: ['upload-image'],
-    uploadImage: true,
+    status: [],
```
gregorbg, line 111: This list can be built using `[enableImageUpload && 'upload-image']` and `filter(Boolean)`

app/webpacker/components/wca/FormBuilder/input/MarkdownEditor.js
```diff
@@ -69,17 +71,23 @@ function getOptions(disabled) {
...
-    '|', ...uploadsAndInserts,
+    '|', ...inserts, ...uploadImageTool,
     '|', ...previews,
     '|', ...helps,
   ];
 
+  const imageOptions = editorOptions.enableImageUpload ? {
+    status: ['upload-image'],
+    uploadImage: true,
+  } : {};
```
gregorbg, line 89: With my other two suggestions, you won't need this hash at all

app/webpacker/components/wca/FormBuilder/input/MarkdownEditor.js
```diff
@@ -69,17 +71,23 @@ function getOptions(disabled) {
...
-  const uploadsAndInserts = ['link', 'upload-image', map, youtube];
+  const inserts = ['link', map, youtube];
+  const uploadImageTool = editorOptions?.enableImageUpload ? ['upload-image'] : [];
   const previews = ['preview', 'side-by-side', 'fullscreen'];
   const helps = ['guide'];
   const toolbar = [
     ...textFormattings,
     '|', ...textStructures,
-    '|', ...uploadsAndInserts,
+    '|', ...inserts, ...uploadImageTool,
```
gregorbg, line 81: Why don't you spread `uploadImageTool` into `inserts` directly?

app/webpacker/components/wca/FormBuilder/input/MarkdownEditor.js
```diff
@@ -113,16 +121,17 @@ function getOptions(disabled) {
...
-    disabled,
+    ...imageOptions,
   };
 }
 
 export default function MarkdownEditor({
   id,
   value,
   onChange,
+  editorOptions = DEFAULT_EDITOR_OPTIONS,
```
gregorbg, line 132: Not sure whether such a complex configuration object is needed right now. We don't use this editor a lot to be honest, so you might be trying to future-proof a use-case that will actually never happen (for reference, I think we never ever touched it before this PR of yours)

Anyhow, this thing alone won't block your PR, so you're free to keep it if you feel strongly about this idea. My other remarks are more important

gregorbg: Actually, on second thought here's an argument against your configuration hash idea (sorry!):
If you pass a complex configuration hash, the user is responsible for just "knowing" how the options are called and doing research on which options to pass.
If you have a property `imageUploadEnabled`, it's very clear from the name alone what that does, and only one developer (right now, you :smile: ) is responsible for mapping this boolean property to the correct internal configuration settings. Once that mapping is done (and the library doesn't introduce breaking changes), everybody else can use it "for free"

danieljames-dj: Okay done.

app/webpacker/components/wca/FormBuilder/input/MarkdownEditor.js
danieljames-dj: Done

app/webpacker/components/wca/FormBuilder/input/MarkdownEditor.js
danieljames-dj: Done

app/webpacker/components/wca/FormBuilder/input/MarkdownEditor.js
danieljames-dj: Okay, removed.

app/webpacker/components/wca/FormBuilder/input/MarkdownEditor.js
danieljames-dj: Ya I can do that too, done.

app/webpacker/components/wca/FormBuilder/input/MarkdownEditor.js
```diff
@@ -100,8 +103,8 @@ function getOptions(disabled) {
     spellChecker: false,
     promptURLs: true,
     previewRender,
-    status: ['upload-image'],
-    uploadImage: true,
+    status: [imageUploadEnabled && UPLOAD_IMAGE_KEY].filter(Boolean),
```
danieljames-dj, line 106: I see that this and line 74 are doing the same thing. But I couldn't find a good variable name for both in common, mainly because both are different use-cases - one is toolbar, and other is editor status.

gregorbg (approved):

PR #11087, opened 2025-03-18
Move events checks to model validations

gregorbg:
Another step towards shortening `RegistrationChecker`.

Ironically, this one adds quite a bit of complexity to the building/parsing mechanism within `RegistrationChecker`, but I am confident that:
1. This is only a migration pain and this complexity will naturally disapear when the migration is complete
2. There are huge benefits, for example now we can check which events somebody is (not) qualified for "for free"


lib/registrations/registration_checker.rb
```diff
@@ -2,20 +2,41 @@
...
+
+      if competing_payload&.key?('event_ids')
+        desired_events = competing_payload['event_ids']
+
+        competition_events_lookup = registration.competition.competition_events.where(event_id: desired_events).index_by(&:event_id)
+        competition_events = desired_events.map { competition_events_lookup[it] }
+
+        rce_build_params = competition_events.map { { competition_event: it } }
+        registration.registration_competition_events = registration.registration_competition_events.build(rce_build_params)
+      end
```
gregorbg, line 25: This is just a fancy way of saying "try and load the `competition_events` based on WCA Event ID, or fall back to `nil` if the competition doesn't list it"

PR #11088, opened 2025-03-18
Move user_can_create_registration to a before action

FinnIckler:
Will probably have to move the tests to request tests

### DIFF
--- a/app/controllers/api/v1/registrations/registrations_controller.rb
+++ b/app/controllers/api/v1/registrations/registrations_controller.rb
@@ -5,6 +5,7 @@ class Api::V1::Registrations::RegistrationsController < Api::V1::ApiController
   # The order of the validations is important to not leak any non public info via the API
   # That's why we should always validate a request first, before taking any other before action
   # before_actions are triggered in the order they are defined
+  before_action :user_can_create_registration, only: [:create]
   before_action :validate_create_request, only: [:create]
   before_action :validate_show_registration, only: [:show]
   before_action :validate_list_admin, only: [:list_admin]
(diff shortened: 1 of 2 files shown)


FinnIckler:
As expected, a lot of these tests now need to move to request tests, because the checks considering the current_user have all moved to the before_action

app/controllers/api/v1/registrations/registrations_controller.rb
```diff
@@ -54,8 +55,24 @@ def create
...
+    @target_user = User.find(params.require('user_id'))
+    @competition = Competition.find(params.require('competition_id'))
+    raise WcaExceptions::RegistrationError.new(:forbidden, Registrations::ErrorCodes::REGISTRATION_ALREADY_EXISTS) if
+      Registration.exists?(competition: @competition, user: @target_user)
+
+    # Only the user themselves can create a registration for the user
+    raise WcaExceptions::RegistrationError.new(:unauthorized, Registrations::ErrorCodes::USER_INSUFFICIENT_PERMISSIONS) unless @current_user.id == @target_user.id
+
+    # Only organizers can register when registration is closed, and they can only register for themselves - not for other users
+    raise WcaExceptions::RegistrationError.new(:forbidden, Registrations::ErrorCodes::REGISTRATION_CLOSED) unless @competition.registration_currently_open? || organizer_modifying_own_registration?(@competition, @current_user, @target_user)
```
gregorbg, line 68: Not sure how nitpicky we want to get here, but at this point target_user == current_user has already been established (otherwise the previous exception would've been thrown). So checking `can_manage_competition` should be enough.

gregorbg: I am aware that this would make validation checks order-dependent, so this is not a strong preference or requirement to change it. More of an FYI in case you have any thoughts about it.

gregorbg (approved):
Didn't compare every line of the moved tests exactly, I trust that you know how to c/p ;)

app/controllers/api/v1/registrations/registrations_controller.rb
FinnIckler: these are already order dependent so we do not leak ban information. E.g checking if you are the target user before checking you are banned. so this is fine? I'll check if the tests still pass

PR #11090, opened 2025-03-18
Enable all Rake Rubocops

gregorbg:
The diff about `rehash_links` is a pure copy-paste.


lib/tasks/rehash_competition_tab_links.rake
```diff
@@ -1,5 +1,50 @@
 # frozen_string_literal: true
 
+def rehash_links(model, column, legacy_verifier)
+  puts "Re-hashing field #{column} in model #{model}"
```
FinnIckler, line 4: how are these puts allowed with Rails/Output enabled

FinnIckler (approved):

PR #11091, opened 2025-03-18
Update ko translation.

goomseo:
- Update ko.yml
- Create time_will_tell.ko.yml
- Create carrierwave.ko.yml

Translator: MinSeo Gu (2014GUMI01)


.gitignore
```diff
@@ -60,3 +60,6 @@ yarn-debug.log*
 !.yarn/releases
 !.yarn/sdks
 !.yarn/versions
+
+# .DS_Store files
+.DS_Store
```
gregorbg, line 65: This is something that you typically want to include in your [global gitignore](https://gist.github.com/subfuzion/db7f57fff2fb6998a16c). I don't think any Git project on planet earth wants to version `DS_Store`, so using your global git configuration is a good way to suppress it altogether.

This also means that I'd rather not have this be included here. If we start letting developers add all of their OS-specific and IDE-specific config flags and files, our gitignore would quickly become a mess.

gregorbg (changes requested):
Thank you for your effort! Just one technical remark regarding the gitignore, otherwise we're all good!

gregorbg (approved):

.gitignore
goomseo: Just reverted .gitignore file and made my global gitignore. Your comment taught me a lot—I'm not that familiar with Git yet. Thank you!

By the way, can I ask if the [translators](https://www.worldcubeassociation.org/translators) page is updated automatically?

PR #11095, opened 2025-03-18
Reg Admin: non-competing section

kr-matthews:
Back-end change based on slack thread. Feel free to modify this PR if there's a better way.

Competition without any non-competing registrations:
image

And with:
image

Cannot move people to it:
image

It _looks_ like you can move them, but they aren't passed on to the back-end when you click (similar to how the "move x to" count is wrong if you have anyone selected from the destination section). I might be able to improve that in one of my other draft PRs, but it seems unimportant for now.
image

### DIFF
--- a/app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationList.jsx
+++ b/app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationList.jsx
@@ -193,7 +193,6 @@ export default function RegistrationAdministrationList({ competitionInfo }) {
               onSelect={selectedIds.add}
               onUnselect={selectedIds.remove}
               onToggle={selectedIds.toggle}
-              competition_id={competitionInfo.id}
               competitionInfo={competitionInfo}
               color={PENDING_COLOR}
               distinguishPaidUnpaid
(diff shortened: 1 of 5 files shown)


dunkOnIT:
Would it be possible to not show the "Move to..." option for non-competing registrations? 

Besides that, tested locally and LGTM!

kr-matthews:
> Would it be possible to not show the "Move to..." option for non-competing registrations?

We could disable the move option when only non-competing registrations are selected. But the user could select a non-competing registration and another registration, in which case the button would be enabled. So it seemed simpler for now to not bother, especially since the non-competing user is ignored in the move action. I'm also working on redoing these action buttons in #10980 (currently a draft, waiting for some feedback, and will need to redo it after other things are merged which cause conflicts), so I figure further refinement could be done there.

Note: removing the checkbox would not be a good solution because it's also used to selectively download/email registrations.

dunkOnIT:
Fair points! Have asked Gregor to take a look at this for final sign-off

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationList.jsx
```diff
@@ -342,15 +347,48 @@ export default function RegistrationAdministrationList({ competitionInfo }) {
...
+              {I18n.t('competitions.registration_v2.list.non_competing.information')}
+            </Header.Subheader>
+            <RegistrationAdministrationTable
+              columnsExpanded={expandedColumns}
+              registrations={nonCompeting}
+              selected={partitionedSelectedIds.nonCompeting}
+              onSelect={selectedIds.add}
+              onUnselect={selectedIds.remove}
+              onToggle={selectedIds.toggle}
+              competition_id={competitionInfo.id}
```
gregorbg, line 375: Underscore in a React component prop?! Is that typo or actually intended?

FinnIckler:
Good point, not sure what a good solution is then

FinnIckler (approved):

PR #11097, opened 2025-03-19
Reg Admin: Fix skip/over limit check order

kr-matthews:
As mentioned on slack.


FinnIckler (approved):
Backend fixes are also merged now

PR #11098, opened 2025-03-19
Prevent bulk accepting over the competitor limit

dunkOnIT:
Per discussion in Slack: https://worldcubeassociation.slack.com/archives/C024FPQPL87/p1742340885173609

### DIFF
--- a/lib/registrations/registration_checker.rb
+++ b/lib/registrations/registration_checker.rb
@@ -49,16 +49,20 @@ module Registrations
     end
 
     def self.bulk_update_allowed!(bulk_update_request, current_user)
+      update_requests = bulk_update_request['requests']
       raise WcaExceptions::BulkUpdateError.new(:bad_request, [Registrations::ErrorCodes::INVALID_REQUEST_DATA]) if
-        bulk_update_request['requests'].blank?
+        update_requests.blank?
 
       competition = Competition.find(bulk_update_request['competition_id'])
+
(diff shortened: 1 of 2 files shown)


lib/registrations/registration_checker.rb
```diff
@@ -68,6 +72,15 @@ def self.bulk_update_allowed!(bulk_update_request, current_user)
     end
 
     class << self
+      def will_exceed_competitor_limit?(update_requests, competition)
+        registrations_to_be_accepted = update_requests.count { |r| r.dig('competing', 'status') == Registrations::Helper::STATUS_ACCEPTED }
+        total_accepted_registrations_after_update = competition.registrations.competing_status_accepted.count + registrations_to_be_accepted
```
gregorbg, line 77:
```suggestion
        total_accepted_registrations_after_update = competition.registrations.accepted_count + registrations_to_be_accepted
```

gregorbg (changes requested):
(also kindly pointing out that we will get this "for free" when doing model validations. But I understand we need a more immediate fix now)

dunkOnIT:
Yep agreed! In fact, it wouldn't have worked if the second auto accept PR was merged - this just causes it to fail more gracefully than it would if the model validation in that PR (not the ones you/Finn are working on) had prevented this

gregorbg (approved):

PR #11102, opened 2025-03-19
Activate remaining Naming cops

FinnIckler:
Currently allowing one letter names c, n and id. Mostly because I couldn't figure out what c means for all those distance methods.

All methods I deleted had nothing referencing it so it wasn't worth renaming them


FinnIckler:
just saw that the `is_active`/`is_hidden` `user_group` values are actually database columns

gregorbg:
What's wrong with that naming anyways? On a database level, the `is_` prefix is pretty standard.
Same goes for `has_permission`, I personally feel that `fulfills_permission` is actually a huge step backwards. I think this cop complains because in Rails you usually do `completed?` instead of `has_completed`. But since "to have permission" is a phrasal expression in English, I don't see a good way around it.
Native speakers please feel free to change my mind.

FinnIckler:
there is a allow list we can put is_active and has_permissions into if we want

app/controllers/admin_controller.rb
```diff
@@ -218,13 +218,6 @@ def fix_results_selector
     render partial: "fix_results_selector"
   end
 
-  def edit_person
```
gregorbg, line 221: Why does this method have to go?

app/controllers/panel_controller.rb
```diff
@@ -62,7 +62,7 @@ def generate_db_token
     )
     results_validator.validate(competition_ids)
     render json: {
-      has_results: results_validator.has_results?,
+      has_results: results_validator.results?,
```
gregorbg, line 65: Oof, close call. Only `results?` is like... Yeah, Results. What about them?

Or in other words, nothing about a `?` suffix inherently screams `has`. It might be a shorthand for `is_` or `can_` or `did_` as well FWIW. Any chance for a better rename without introducing `has_results` as a cop exception?

gregorbg: (although for the sake of reviewing I will accept a cop exception as a last resort)

app/controllers/regional_organizations_controller.rb
```diff
@@ -53,7 +53,7 @@ def update
 
   def destroy
     @regional_organization = regional_organization_from_params
-    if @regional_organization.is_pending?
+    if @regional_organization.pending?
```
gregorbg, line 56: Not very firm in ROs. But this might potentially be a DB column...?

app/models/competition.rb
```diff
@@ -407,7 +403,7 @@ def advancement_condition_must_be_present_for_all_non_final_rounds
       registrations.waitlisted.count > 0 && !registration_full_and_accepted?
   end
 
-  def has_any_round_per_event?
+  def any_round_empty?
     competition_events.map(&:rounds).none?(&:empty?)
```
gregorbg, line 407: Isn't this more like `no_rounds_empty?`?!

app/models/competition.rb
```diff
@@ -1058,19 +1054,11 @@ def country_zones
...
-    self.events.include?(event)
-  end
-
-  def has_base_entry_fee?
+  def base_entry_fee_set?
     base_entry_fee.nonzero?
   end
 
-  def has_fees?
+  def paid?
```
gregorbg, line 1061: Not a fan of this rename. `paid?` in the past tense implies that an action had happened which caused... the competition to now be considered as paid...??!

app/models/competition.rb
```diff
@@ -1082,11 +1070,11 @@ def external_registration_page_required?
     confirmed? && !use_wca_registration && created_at.present? && created_at > Date.new(2018, 12, 31)
   end
 
-  def has_rounds?
+  def any_rounds?
     rounds.any?
   end
 
-  def has_schedule?
+  def schedule_set?
```
gregorbg, line 1077: Technically, `any_venues?` fits closest.

app/models/light_result.rb
```diff
@@ -27,24 +27,24 @@ class LightResult
...
-    @personName = r["personName"]
-    @pos = r["pos"]
-    @personId = r["personId"]
-    @regionalSingleRecord = r["regionalSingleRecord"]
-    @regionalAverageRecord = r["regionalAverageRecord"]
-    @country = Country.c_find(r["countryId"])
-    @format = Format.c_find(r["formatId"])
-    @round_type = RoundType.c_find(r["roundTypeId"])
-    @event = Event.c_find(r["eventId"])
+  def initialize(params)
```
gregorbg, line 30: Something like `raw_result` or `result_data` or `result_payload`?

app/models/results_submission.rb
```diff
@@ -12,7 +12,7 @@ class ResultsSubmission
   validates :schedule_url, presence: true, url: true
 
   validate do
-    if results_validator.has_errors?
+    if results_validator.errors?
```
gregorbg, line 15: Same as `results?` above. Perhaps `any_errors?`

app/models/round.rb
```diff
@@ -185,7 +185,7 @@ def score_taking_done?
     competitors_live_results_entered == total_accepted_registrations
   end
 
-  def has_undef_tl?
+  def undef_tl?
```
gregorbg, line 188: I'm personally fine with this, but in the interest of avoiding abbreviations what about `time_limit_undefined?`

lib/results_validators/utils.rb
```diff
@@ -12,7 +12,7 @@ module Utils
       ScramblesValidator,
     ].freeze
 
-    VALIDATORS_WITH_FIX = ALL_VALIDATORS.select(&:has_automated_fix?).freeze
+    VALIDATORS_WITH_FIX = ALL_VALIDATORS.select(&:automated_fix?).freeze
```
gregorbg, line 15: What about `automated_fix_possible?`? Don't feel very strongly about this one though, because it's quite long. Feel free to dismiss.

gregorbg (changes requested):

app/controllers/admin_controller.rb
FinnIckler: I didn't find any uses in the code for it

app/controllers/panel_controller.rb
FinnIckler: any_results?

app/models/competition.rb
FinnIckler: well I was more thinking about free service vs. paid service.

app/models/results_submission.rb
FinnIckler: erorred? I think is good here

lib/results_validators/utils.rb
FinnIckler: automatically_fixable?

app/controllers/regional_organizations_controller.rb
FinnIckler: no, it's defined
```
def pending?
    start_date.nil?
  end
```

app/controllers/admin_controller.rb
gregorbg: Then it could certainly go in a different PR, but it feels very out of scope for this current PR

app/models/competition.rb
```diff
@@ -1058,19 +1054,11 @@ def country_zones
...
-  def has_events_with_ids?(event_ids)
-    (event_ids - events.ids).empty?
-  end
-
-  def has_event?(event)
-    self.events.include?(event)
-  end
-
-  def has_base_entry_fee?
+  def base_entry_fee_set?
```
gregorbg, line 1057: The method name suggests "the value is set", but it actually checks "the value is set and non-zero". What about `base_entry_fee_nonzero?`?

(which would probably end up becoming `delegate :nonzero?, to: :base_entry_fee, prefix: true` anyways)

app/models/competition.rb
```diff
@@ -1155,15 +1143,15 @@ def pending_results_or_report(num_days)
   end
 
   # does the competition have this field (regardless of whether it's a date or blank)
-  def has_event_change_deadline_date?
+  def event_change_deadline_date_set?
```
gregorbg, line 1146: What about `..._present?`?

app/models/competition.rb
```diff
@@ -1251,26 +1239,22 @@ def colliding_registration_start_competitions
...
       )
   end
 
-  def has_date?
+  def date_set?
     !start_date.nil? || !end_date.nil?
   end
 
-  def has_registration_start_date?
+  def registration_start_date_set?
```
gregorbg, line 1254: `..._present?`, see above.

app/models/competition.rb
```diff
@@ -1284,7 +1268,7 @@ def date_range
     ApplicationController.helpers.wca_date_range(self.start_date, self.end_date)
   end
 
-  def has_date_errors?
+  def contains_date_errors?
```
gregorbg, line 1271: Sneaky, `contains` is just a convoluted way of saying `has`, and that in turn is considered "banned". So you're just dancing around the cliff here.

app/models/competition_event.rb
```diff
@@ -32,7 +32,7 @@ def currency_code
     competition&.currency_code
   end
 
-  def has_fee?
+  def paid?
```
gregorbg, line 35: See my other remark above about the name `paid` being ambiguous.

app/models/user.rb
```diff
@@ -1137,7 +1137,7 @@ def notify_of_id_claim_possibility(competition)
     CompetitionsMailer.notify_users_of_id_claim_possibility(self, competition).deliver_later if !wca_id && !unconfirmed_wca_id
   end
 
-  def is_bookmarked?(competition)
+  def bookmarked?(competition)
```
gregorbg, line 1140: What about `competition_bookmarked?`?

lib/results_validators/generic_validator.rb
```diff
@@ -9,15 +9,15 @@ def initialize(apply_fixes: false)
...
 
-    def has_warnings?
+    def warnings?
       @warnings.any?
     end
 
-    def has_infos?
+    def infos?
       @infos.any?
     end
```
gregorbg, line 22: What about `any_..?`?

app/models/competition.rb
```diff
@@ -298,10 +298,6 @@ def number_of_events
     persisted_events_id.length
   end
 
-  def has_administrative_notes?
-    registrations.any? { |registration| registration.administrative_notes.present? }
-  end
-
```
gregorbg, line 304: This method is okay to remove because it would otherwise conflict with the new cops

app/models/competition.rb
```diff
@@ -1058,19 +1054,11 @@ def country_zones
     self.longitude_microdegrees = @longitude_degrees * 1e6 unless @longitude_degrees.nil?
   end
 
-  def has_events_with_ids?(event_ids)
-    (event_ids - events.ids).empty?
-  end
-
-  def has_event?(event)
-    self.events.include?(event)
-  end
```
gregorbg, line 1067: Again, these are fine to remove because otherwise they would conflict with the new cops

gregorbg (changes requested):
Still lots of individual comments, sorry! Naming stuff is hard.

Also, in general I am not a fan of the `xy_set?` nomenclature. That feels way too vague in Rails jargon. What about `xy_present?` instead (in cases where that actually makes sense)? It has an already established meaning and it feels less "generic" than `set`

app/models/competition.rb
FinnIckler: doesn't look like that works
```
Competition.find("WC2025").base_entry_fee_nonzero?
  Competition Load (0.5ms)  SELECT `Competitions`.* FROM `Competitions` WHERE `Competitions`.`id` = 'WC2025' LIMIT 1
=> #<Money fractional:12000 currency:USD>
```
But I am fine with the name

app/models/competition.rb
```diff
@@ -1058,19 +1054,9 @@ def country_zones
...
+  delegate :nonzero?, to: :base_entry_fee, prefix: true
 
-  def has_base_entry_fee?
-    base_entry_fee.nonzero?
-  end
-
-  def has_fees?
+  def payment_required?
```
gregorbg, line 1059: That is a very, very broad method name. It could also imply something like "Is there a CPI connected, is that connection valid, how is the registration lane configured, does the organizer get waived fees, etc...". The previous name of `has_fees?` was actually quite good because it closely describes the content of the method. What about...
- `entry_fee_nonzero?`
- `entry_fee_present?`

gregorbg: ChatGPT also suggested the following...
- `fees_required?`
- `paid_entry?`
- `entry_requires_fee?`

app/models/competition.rb
```diff
@@ -1155,15 +1141,15 @@ def pending_results_or_report(num_days)
   end
 
   # does the competition have this field (regardless of whether it's a date or blank)
-  def has_event_change_deadline_date?
+  def event_change_deadline_date_present?
```
gregorbg, line 1144: Maybe `..._required?`? I'm a bit dumb-struck myself because the previous version with `has_` was also not checking the field in question (`event_change_deadline_date`) at all...

app/models/competition.rb
```diff
@@ -1251,24 +1237,20 @@ def colliding_registration_start_competitions
...
       Math.sqrt(
-        (((c.longitude_radians - longitude_radians) * Math.cos((c.latitude_radians + latitude_radians)/2)) ** 2) +
-        ((c.latitude_radians - latitude_radians) ** 2),
+        (((competition.longitude_radians - longitude_radians) * Math.cos((competition.latitude_radians + latitude_radians)/2)) ** 2) +
+        ((competition.latitude_radians - latitude_radians) ** 2),
       )
   end
 
-  def has_date?
+  def date_set?
```
gregorbg, line 1248: Isn't this the exact same as the old `has_defined_dates?` (at the current state of in-progress review: `dates_set?`) above?!

app/models/competition_event.rb
```diff
@@ -32,7 +32,7 @@ def currency_code
     competition&.currency_code
   end
 
-  def has_fee?
+  def payment_required?
```
gregorbg, line 35: Same reasoning as above: `payment_required?` is much too broad.
Whatever you come up with on the bigger `competition` object can probably be used here as well.

gregorbg (changes requested):

app/models/competition.rb
FinnIckler: yeah, looks like this is a duplicate

app/controllers/application_controller.rb
```diff
@@ -9,6 +9,9 @@ class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
 
   prepend_before_action :set_locale, unless: :api_request?
+  # The API should only ever respond in English
+  prepend_before_action :set_default_locale, if: :api_request?
+
```
gregorbg, line 14: Merge whoopsie?

app/controllers/application_controller.rb
```diff
@@ -114,7 +117,7 @@ def storable_location?
     end
 
     def api_request?
-      request.fullpath.include?('/api/')
+      request.fullpath.include?('/api/') || request.fullpath.include?('/oauth/')
```
gregorbg, line 120: More merge whoopsie?

PR #11103, opened 2025-03-19
Add Washington, D.C. as valid location for comp

mckeenicholas:


gregorbg:
The answer is like half-obvious, but did we really never hold any competition in DC so far?

mckeenicholas:
Appears so, the following results query returns nothing:
`SELECT * FROM Competitions WHERE cityName LIKE '%Washington%' AND longitude > -80000000;`

lib/country_city_validators/us_city_validator.rb
```diff
@@ -62,9 +62,13 @@ module CountryCityValidators
     U.S.\ Virgin\ Islands
   ).to_set
 
+  US_DISTRICTS = %w(
+    District\ Of\ Columbia
```
timreyn, line 66: "D.C." feels a lot more natural to me than spelling out "District of Columbia", consider switching to that?

gregorbg:
I am fine with the code changes per se, but I don't feel qualified to make a call on "D.C." vs "District of Columbia". Maybe as a kind of middle ground you could allow both? (ie list both in the array, even though they technically refer to the same thing)

zekemackay06:
Chiming in as I recently made a lot of fixes to old US comp venue information. I would like to make sure that this formatting stays as correct as possible going forward.

1. No, there has somehow never been a DC comp. The closest thing is a location in a multi-location comp.

2. D.C. is more widely used, but 'District of Columbia' is the technical equivalent to the other state names, which are used in cityName for everywhere else in the US. 

I think it should be 'District of Columbia' to match with the other state names. "Washington, District of Columbia" may not be the most common way to refer to the city, but it isn't unclear.

stonepayne1:
Any movement on this? As a resident, my vote would also be 'District of Columbia' as it would remain in line with other names.

gregorbg:
I am fine with any consensus, but if we really want to go with `District of Columbia` then the current status quo of this PR (as of writing this very comment) has to be changed.

@mckeenicholas Are you available to do that soon(ish) or should I jump in?

mckeenicholas:
I can push a change in the next few minutes.

lib/country_city_validators/us_city_validator.rb
```diff
@@ -62,9 +62,13 @@ module CountryCityValidators
     U.S.\ Virgin\ Islands
   ).to_set
 
+  US_DISTRICTS = %w(
+    District\ of\ Columbia.
```
maxidragon, line 66:
```suggestion
    District\ of\ Columbia
```

gregorbg (approved):

PR #11105, opened 2025-03-20
Add and use `entry_fee_with_donation`

dunkOnIT:
Per discussion with Gregor, this PR does the following: 
- Introduces basic tests against `payment_ticket` and `payment_denomination`
- Adds an `entry_fee_with_donation` method which returns a Ruby money object representing the sum of donation + competition's entry fee
- Refactors `payment_ticket` and `payment_denomination` to use `entry_fee_with_donation`
- Alters the frontend calls of `payment_denomination` to align with the new parameters it wants


gregorbg (commented):
Too late for an in-depth review, but overall looking fabulous!

Just one nit about the naming: `lowest_denomination` is a historical artifact, perhaps it comes from some old DB columns. Who determines what lowest is? And more important, who determines that a currency cannot go any lower?

--> `iso_denomination` and `iso_amount` are much clearer IMO. And the Ruby-Money gem has strong guarantees about ISO adherence, so it's "only" a rename, not actual calculations that you would have to do. WDYT?

dunkOnIT:
I agree with your points about `iso`, with a preference for _prefixing_ with `iso`, as that (to me at least) more intuitively communicates that the value in question is being given in iso format. 

Have updated the PR to that effect, which also included changed the parameter name in payment_ticket - note that I haven't tested this in the frontend yet

app/controllers/api/v1/registrations/registrations_controller.rb
```diff
@@ -135,11 +135,12 @@ def validate_payment_ticket_request
   end
 
   def payment_ticket
-    donation = params[:donation_iso].to_i || 0
-    amount_iso = @competition.base_entry_fee_lowest_denomination
-    currency_iso = @competition.currency_code
+    donation = params[:iso_donation].to_i || 0
```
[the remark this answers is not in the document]
dunkOnIT: Changed to use `iso_donation_amount`

also changed `ruby_money_amount` to just `ruby_money` given that it is precisely _not_ just an amount

app/controllers/registrations_controller.rb
```diff
@@ -260,15 +260,14 @@ def register
   end
 
   def payment_denomination
-    ruby_denomination = params.require(:amount)
-    currency_iso = params.require(:currency_iso)
-
-    ruby_money = Money.new(ruby_denomination, currency_iso)
+    registration = registration_from_params
+    donation = params[:iso_donation].to_i || 0
```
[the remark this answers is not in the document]
dunkOnIT: Changed to `iso_donation_amount`

app/models/registration.rb
```diff
@@ -244,6 +251,7 @@ def to_v2_json(admin: false, history: false, pii: false)
     private_attributes = pii ? %w[dob email] : nil
 
     base_json = {
+      id: self.id,
```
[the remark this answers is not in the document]
dunkOnIT: Fair point - I've refactored to use the competition id and user id to find the registration

config/routes.rb
```diff
@@ -43,7 +43,7 @@
   post 'competitions/:competition_id/refund/:payment_integration/:payment_id' => 'registrations#refund_payment', as: :registration_payment_refund
   get 'competitions/:competition_id/payment-completion/:payment_integration' => 'registrations#payment_completion', as: :registration_payment_completion
   post 'registration/stripe-webhook' => 'registrations#stripe_webhook', as: :registration_stripe_webhook
-  get 'registration/payment-denomination' => 'registrations#payment_denomination', as: :registration_payment_denomination
+  get 'registration/:id/payment-denomination' => 'registrations#payment_denomination', as: :registration_payment_denomination
```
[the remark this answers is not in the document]
dunkOnIT: See above comment

app/controllers/api/v1/registrations/registrations_controller.rb
```diff
@@ -135,11 +135,12 @@ def validate_payment_ticket_request
   end
 
   def payment_ticket
-    donation = params[:donation_iso].to_i || 0
-    amount_iso = @competition.base_entry_fee_lowest_denomination
-    currency_iso = @competition.currency_code
+    iso_donation_amount = params[:iso_donation].to_i || 0
```
gregorbg, line 138: What about renaming the `params` key as well?

gregorbg: In the `payment_denomination` endpoint below, you are actually using `iso_donation_amount` in `params` as well :upside_down_face:

app/webpacker/components/RegistrationsV2/Register/StripeWrapper.jsx
```diff
@@ -10,9 +10,9 @@ import { fetchJsonOrError } from '../../../lib/requests/fetchWithAuthenticityTok
 import { paymentDenominationUrl } from '../../../lib/requests/routes.js.erb';
 import { useRegistration } from '../lib/RegistrationProvider';
 
-const convertISOAmount = async (amount, currency) => {
+const convertISOAmount = async (competitionId, userId, isoDonation) => {
   const { data } = await fetchJsonOrError(
-    paymentDenominationUrl(amount, currency),
+    paymentDenominationUrl(competitionId, userId, isoDonation),
```
gregorbg, line 15: Again, the reference to `amount` is missing

spec/models/registration_spec.rb
```diff
@@ -661,4 +661,14 @@
     expect(reg).not_to be_valid
     expect(reg.errors[:registered_at]).to include("can't be blank")
   end
+
+  describe '#entry_fee_with_donation' do
+    it 'returns a hash of entry fee and currency code' do
```
gregorbg, line 666: It doesn't return a `hash` though?!

spec/requests/api_registrations_spec.rb
```diff
@@ -452,17 +452,153 @@
     end
   end
 
+  def stub_successful_stripe_payment_intent(amount, currency)
```
gregorbg, line 455: Why do you need this here and how does this affect existing Stripe tests?

gregorbg (changes requested):

spec/requests/api_registrations_spec.rb
dunkOnIT: When testing the creation of payment tickets, we call the Stripe API which makes the tests significantly slower. I've included a test on line 526 which does query against the StripeAPI, to make sure that we are still constructing the payload appropriately, but for everything else (ie where behaviour of the Stripe API is not under test) I use `stub_successful_stripe_payment_intent` to mock the Stripe API instead of making live calls against it

spec/models/registration_spec.rb
dunkOnIT: Before I thought of using `Money` - fixed!

app/webpacker/components/RegistrationsV2/Register/StripeWrapper.jsx
dunkOnIT: Fixed

app/controllers/api/v1/registrations/registrations_controller.rb
dunkOnIT: Fixed, and fixed in routes.js.erb

dunkOnIT:
I also renamed the state variables `donationAmount` and `setDonationAmount` to `isoDonationAmount` and `setIsoDonationAmount` respectively - not sure if this is going to far into unrelated changes or not, but I've tested the frontend and it works

gregorbg (changes requested):

spec/requests/api_registrations_spec.rb
dunkOnIT: Removed the Stripe stubbing

app/webpacker/components/RegistrationsV2/Register/StripeWrapper.jsx
```diff
@@ -26,15 +26,15 @@ export default function StripeWrapper({
...
 
   const { registration } = useRegistration();
 
   const {
     data, isFetching,
   } = useQuery({
-    queryFn: () => convertISOAmount(initialAmount + donationAmount, competitionInfo.currency_code),
-    queryKey: ['displayAmount', initialAmount + donationAmount, competitionInfo.currency_code],
+    queryFn: () => convertISOAmount(competitionInfo.id, registration.user_id, isoDonationAmount),
+    queryKey: ['displayAmount', isoDonationAmount, competitionInfo.currency_code],
```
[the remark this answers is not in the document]
dunkOnIT: Query key updated to remove currency_code and add user_id and competition_id

dunkOnIT:
Re adding the safe accessor in this line: 
```ruby
iso_donation_amount = params[:iso_donation_amount]&.to_i || 0
```
Rubocop has the following to say: 
```bash
Lint/RedundantSafeNavigation: Redundant safe navigation with default literal detected.
```

gregorbg:
> Rubocop has the following to say:
> 
> ```shell
> Lint/RedundantSafeNavigation: Redundant safe navigation with default literal detected.
> ```

Huh. Must be one of those "`nil` actually has a `to_i` method defined" Ruby quirks. Interesting, thanks for pointing that out!

gregorbg (approved):

PR #11109, opened 2025-03-22
Add error column to competition dues export & moved row generation to model instead of view file

danieljames-dj:
WFC is facing issue while downloading export if the export includes Venezuela competitions. The reason is that Venezuela competitions are using VEF currency which is deprecated currency since 2018. (WFC is enquiring with Delegate why they used the deprecated currency).

Along with this change, I've added couple of error columns in the export, because WFC is going to trust the dues calculation by our system, and if the errors are silently converted to zeroes, it might lead to mistakes in WFC's side. The error columns will help WFC to understand that something is wrong.

With this change, `DuesCalculator.dues_per_competitor_in_usd` is no longer going to be error-free always, instead it will throw error if something goes wrong. The error will be catched before generating the columns for a competition.

### DIFF
--- a/app/controllers/wfc_controller.rb
+++ b/app/controllers/wfc_controller.rb
@@ -11,8 +11,7 @@ class WfcController < ApplicationController
     select_attributes = [
       :id, :name, :start_date, :end_date,
       :country_id, :announced_at, :results_posted_at,
-      :currency_code, :base_entry_fee_lowest_denomination,
-      "count(distinct persons.id) as num_competitors"
+      :currency_code, :base_entry_fee_lowest_denomination
     ]
     from = params.require(:from_date)
     to = params.require(:to_date)
--- a/app/controllers/wfc_controller.rb
+++ b/app/controllers/wfc_controller.rb
@@ -18,6 +18,7 @@ class WfcController < ApplicationController
     response.headers["Content-Disposition"] = "attachment; filename=\"wfc-competitions-export-#{from}-#{to}.tsv\""
     @competitions = Competition
                     .select(select_attributes)
+                    .select('COUNT(DISTINCT persons.id) AS num_competitors')
                     .includes(:delegates, :championships, :organizers, :events, organizers: [:wfc_dues_redirect])
                     .left_joins(:competitors)
                     .group(:id)
--- a/app/controllers/wfc_controller.rb
+++ b/app/controllers/wfc_controller.rb
@@ -8,10 +8,10 @@ class WfcController < ApplicationController
   end
 
   def competition_export
-    select_attributes = [
-      :id, :name, :start_date, :end_date,
-      :country_id, :announced_at, :results_posted_at,
-      :currency_code, :base_entry_fee_lowest_denomination
+    select_attributes = %i[
+      id name start_date end_date
+      country_id announced_at results_posted_at
+      currency_code base_entry_fee_lowest_denomination
     ]
     from = params.require(:from_date)
     to = params.require(:to_date)
--- a/app/controllers/wfc_controller.rb
+++ b/app/controllers/wfc_controller.rb
@@ -8,17 +8,17 @@ class WfcController < ApplicationController
   end
 
   def competition_export
-    select_attributes = %i[
-      id name start_date end_date
-      country_id announced_at results_posted_at
-      currency_code base_entry_fee_lowest_denomination
+    select_attributes = [
+      :id, :name, :start_date, :end_date,
+      :country_id, :announced_at, :results_posted_at,
+      :currency_code, :base_entry_fee_lowest_denomination,
+      "count(distinct persons.id) as num_competitors"
     ]
     from = params.require(:from_date)
     to = params.require(:to_date)
     response.headers["Content-Disposition"] = "attachment; filename=\"wfc-competitions-export-#{from}-#{to}.tsv\""
     @competitions = Competition
                     .select(select_attributes)
-                    .select('COUNT(DISTINCT persons.id) AS num_competitors')
                     .includes(:delegates, :championships, :organizers, :events, organizers: [:wfc_dues_redirect])
                     .left_joins(:competitors)
                     .group(:id)
--- a/app/models/competition.rb
+++ b/app/models/competition.rb
@@ -2219,16 +2219,8 @@ class Competition < ApplicationRecord
   end
 
   def export_for_dues_generation
-    begin
-      dues_per_competitor_in_usd = DuesCalculator.dues_per_competitor_in_usd(self.country_iso2, self.base_entry_fee_lowest_denomination.to_i, self.currency_code)
-      error = 'None'
-    rescue Money::Currency::UnknownCurrency, CurrencyUnavailable, Money::Bank::UnknownRate
-      dues_per_competitor_in_usd = 0
-      error = 'Unknown/unavailable currency/rate'
-    rescue ArgumentError
-      dues_per_competitor_in_usd = 0
-      error = 'Invalid arguments'
-    end
+    error = DuesCalculator.error_in_dues_calculation(self.country_iso2, self.currency_code)
+    dues_per_competitor_in_usd = error.nil? ? DuesCalculator.dues_per_competitor_in_usd(self.country_iso2, self.base_entry_fee_lowest_denomination.to_i, self.currency_code) : 0
 
     [
       id, name, country.iso2, continent.id,
--- a/app/models/competition.rb
+++ b/app/models/competition.rb
@@ -2223,16 +2223,13 @@ class Competition < ApplicationRecord
 
     begin
       dues_per_competitor_in_usd = DuesCalculator.dues_per_competitor_in_usd(self.country_iso2, self.base_entry_fee_lowest_denomination.to_i, self.currency_code)
-      error_in_dues_calculation = false
-      error_invalid_arguments = false
+      error = 'None'
     rescue Money::Currency::UnknownCurrency, CurrencyUnavailable, Money::Bank::UnknownRate
       dues_per_competitor_in_usd = 0
-      error_in_dues_calculation = true
-      error_invalid_arguments = false
+      error = 'Unknown/unavailable currency/rate'
     rescue ArgumentError
       dues_per_competitor_in_usd = 0
-      error_in_dues_calculation = false
-      error_invalid_arguments = true
+      error = 'Invalid arguments'
     end
 
     [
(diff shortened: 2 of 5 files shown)


gregorbg (changes requested):
I understand the intention behind your PR, but I'm not a fan of catching errors in general. See my inline comments for details and alternative suggestions.

danieljames-dj (commented):
Thanks for the review, @gregorbg. I understand your indention of returning `nil` instead of throwing error. But I feel that in my use-case, throwing might be slightly better. I can explain the reason - since WFC will soon start trusting solely on the dues calculation by the software, it will be great if WFC is informed "is something wrong?". It will be more great if WFC is also informed "what went wrong?", so that WFC can take action accordingly.

Another option is to return multiple values - first value as dues, and second value as error. But I feel that is more worse than throwing.

Can you please let me know if my reasoning is reasonable?

Sidenote: WFC is blocked with few dues because of this PR, but now I feel this might require more discussion. I'll not rush this PR, instead will raise another 1-line change to unblock WFC.

app/models/competition.rb
```diff
@@ -2245,9 +2245,32 @@ def find_round_for(event_id, round_type_id, format_id = nil)
...
-    dues = DuesCalculator.dues_per_competitor_in_usd(self.country_iso2, self.base_entry_fee_lowest_denomination.to_i, self.currency_code)
-    dues.presence || 0
+  def export_for_dues_generation
+    num_competitors = competitors.count
+
+    begin
+      dues_per_competitor_in_usd = DuesCalculator.dues_per_competitor_in_usd(self.country_iso2, self.base_entry_fee_lowest_denomination.to_i, self.currency_code)
+      error_in_dues_calculation = false
+      error_invalid_arguments = false
+    rescue Money::Currency::UnknownCurrency, CurrencyUnavailable, Money::Bank::UnknownRate
```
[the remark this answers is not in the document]
gregorbg: Pinging this comment

lib/dues_calculator.rb
```diff
@@ -4,15 +4,15 @@ module DuesCalculator
   def self.dues_per_competitor(country_iso2, base_entry_fee_lowest_denomination, currency_code)
     dues_per_competitor_in_usd_money = dues_per_competitor_in_usd(country_iso2, base_entry_fee_lowest_denomination, currency_code)
     dues_per_competitor_in_usd_money&.exchange_to(currency_code)
-  rescue CurrencyUnavailable
+  rescue Money::Currency::UnknownCurrency, CurrencyUnavailable, Money::Bank::UnknownRate, ArgumentError
```
gregorbg: For all of the money-related errors, see my other comment above

app/models/competition.rb
```diff
@@ -2208,9 +2208,29 @@ def find_round_for(event_id, round_type_id, format_id = nil)
     end
   end
 
-  def dues_per_competitor_in_usd
-    dues = DuesCalculator.dues_per_competitor_in_usd(self.country_iso2, self.base_entry_fee_lowest_denomination.to_i, self.currency_code)
-    dues.presence || 0
+  def export_for_dues_generation
+    num_competitors = competitors.count
```
gregorbg: I think this will generate one extra query per row. Why did you get rid of the `COUNT(DISTINCT persons.id)` in the `select_attributes` of the controller? Did I tell you that during review? :sweat_smile:

gregorbg (changes requested):

app/models/competition.rb
danieljames-dj: No, you didn't tell. Actually I want to do the generation of each row in the rails side instead of HTML side. I removed it as I thought I won't be able to do the same thing in rails. But after some more research, I was able to figure out a way. Now this should not generate extra query.

app/models/competition.rb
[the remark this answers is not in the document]
danieljames-dj: Yes, we do have with `Money` library:

- `Money::Currency.table.key?(:aed)`
- `Money.default_bank.get_rate('EUR', 'USD').nil?`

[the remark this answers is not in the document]
danieljames-dj: Just to confirm my understanding: you'd like `dues_calculator` to return `nil` on error, and for me to implement `if-else` checks based on that `nil` return. Is that correct?

I'm curious about the benefits of this approach compared to **throwing errors directly** from `dues_calculator`, especially since the `Money` library handles them so conveniently. Is there a specific **software engineering best practice** that this pattern adheres to?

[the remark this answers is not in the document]
gregorbg: I can't quote any literature. Just my own thinking: If you already know that "currency doesn't exist" is an error case that can come up, then your code can handle this edge case explicitly. That's simply my definition of "robustness".

Catching errors is like saying "Oh, I write my main code completely ignoring the possibility that it can happen, and then at the end I add this catch-block just in case".

But since you're already expecting the error to be raised, you might as well write your own robust code to stop the error from being raised in the first place. Like the software version of real-life "disaster prevention"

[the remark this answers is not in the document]
danieljames-dj: Okay, I agree on it. I tried implementing, but I feel I am not doing it the right way. Can you please let me know is this what you meant?

Also, in my previous comment I said that there is a method to check if rate exists (`Money.default_bank.get_rate('EUR', 'USD').nil?`), but this is actually not working as it is throwing `CurrencyUnavailable` error if currency is not there. Our [custom library](https://github.com/phlegx/money-currencylayer-bank/blob/master/lib/money/bank/currencylayer_bank.rb) also doesn't have any method to handle this case.

Will it make sense to add try-catch within `DuesCalculator.error_in_dues_calculation` method?

app/controllers/wfc_controller.rb
```diff
@@ -8,17 +8,17 @@ def panel
...
+      id name start_date end_date
+      country_id announced_at results_posted_at
+      currency_code base_entry_fee_lowest_denomination
     ]
     from = params.require(:from_date)
     to = params.require(:to_date)
     response.headers["Content-Disposition"] = "attachment; filename=\"wfc-competitions-export-#{from}-#{to}.tsv\""
     @competitions = Competition
                     .select(select_attributes)
+                    .select('COUNT(DISTINCT persons.id) AS num_competitors')
```
gregorbg: This new solution seems to me like it's exactly the same as the old solution. (Except that you wrote `COUNT` and `DISTINCT` in capital letters now). Is this change to this file strictly necessary?

danieljames-dj: Oh yes, it's not needed. I've changed it now.

app/models/competition.rb
```diff
@@ -2208,9 +2208,19 @@ def find_round_for(event_id, round_type_id, format_id = nil)
...
+    error = DuesCalculator.error_in_dues_calculation(self.country_iso2, self.currency_code)
+    dues_per_competitor_in_usd = error.nil? ? DuesCalculator.dues_per_competitor_in_usd(self.country_iso2, self.base_entry_fee_lowest_denomination.to_i, self.currency_code) : 0
+
+    [
+      id, name, country.iso2, continent.id,
+      start_date, end_date, announced_at, results_posted_at,
+      Rails.application.routes.url_helpers.competition_url(id), num_competitors, delegates.reject(&:trainee_delegate?).map(&:name).sort.join(","),
+      currency_code, base_entry_fee_lowest_denomination, Money::Currency.new(currency_code).subunit_to_unit,
+      championships.map(&:championship_type).sort.join(","), exempt_from_wca_dues?, organizers.map(&:name).sort.join(","),
+      dues_per_competitor_in_usd * num_competitors, dues_payer_name, dues_payer_email, dues_payer_is_combined_invoice?, country.try(:band).try(:number) || 0,
```
gregorbg: The relation `has_one :band` is definitely defined on `Country` model, so `try` is not necessary here. I think what you mean is `country.band&.number || 0`?

lib/dues_calculator.rb
```diff
@@ -4,16 +4,14 @@ module DuesCalculator
   def self.dues_per_competitor(country_iso2, base_entry_fee_lowest_denomination, currency_code)
     dues_per_competitor_in_usd_money = dues_per_competitor_in_usd(country_iso2, base_entry_fee_lowest_denomination, currency_code)
     dues_per_competitor_in_usd_money&.exchange_to(currency_code)
-  rescue CurrencyUnavailable
-    nil
   end
 
   def self.dues_per_competitor_in_usd(country_iso2, base_entry_fee_lowest_denomination, currency_code)
+    return nil if DuesCalculator.error_in_dues_calculation(country_iso2, currency_code)
```
gregorbg:
```suggestion
    return nil if DuesCalculator.error_in_dues_calculation(country_iso2, currency_code).present?
```

gregorbg (changes requested):

app/models/competition.rb
danieljames-dj: I actually didn't try to improve this part of the existing code, I just copied it from `app/views/wfc/competition_export.csv.erb`. Now when I checked, yes `band` is definitely defined on `Country`, but the `country_band` definitely has `number` too, so just `country.band.number` will work? If yes, do we actually need the error in `error_in_dues_calculation` where it checks if country band exists?

lib/dues_calculator.rb
```diff
@@ -23,7 +21,18 @@ def self.dues_per_competitor_in_usd(country_iso2, base_entry_fee_lowest_denomina
...
+
+  def self.error_in_dues_calculation(country_iso2, currency_code)
+    country_band = CountryBand.find_by(iso2: country_iso2)
+    if country_band.nil?
+      "Country band not found."
+    elsif country_band.active_country_band_detail.nil?
+      "Country band detail not found for #{country_band.iso2}."
+    elsif !Money::Currency.table.key?(currency_code.downcase.to_sym)
+      "Currency #{currency_code} is not supported."
+    elsif Money.default_bank.get_rate(currency_code, 'USD').nil?
```
danieljames-dj, line 34: @gregorbg This will throw `CurrencyUnavailable` error, and I couldn't find a way to catch that error. So is it fine to have `rescue` within this method and return the error message?

app/models/competition.rb
gregorbg: Well, the `has_one` can still return `nil` depending on whether we actually have a corresponding row in the database or not.

lib/dues_calculator.rb
gregorbg: Hm, I just checked the source code of `EuCentralBank` gem, and their code design is a bit weird (return `true` or throw, instead of return `true` or `false`)

Here's the snippet:
```ruby
def check_currency_available(currency)
  currency_string = currency.to_s
  return true if currency_string == "EUR"
  return true if CURRENCIES.include?(currency_string)
  raise CurrencyUnavailable, "No rates available for #{currency_string}"
end
```

You could try to run similar checks here, but as I recall we're even using a different bank in production so maybe the code would become too bloated... I will let you handle it with `rescue` if you add a code comment that blames this `rescue` on the poor code design of EuCentralBank :P

danieljames-dj: Okay done

app/models/competition.rb
danieljames-dj: Oops ya, my bad. I was probably absent-minded in my previous comment. I've now changed it.

PR #11121, opened 2025-03-24
Always compute registrationCurrentlyOpen on the frontend

FinnIckler:
makes sure this is up to date not just on page load


gregorbg (commented):
Please also remove the backend `registration_currently_open?` field from the serialization, so that nobody gets any... ideas... in the future

gregorbg (approved):

PR #11122, opened 2025-03-24
Only include `is_competing: true` registrations in competitor limit check

dunkOnIT:
This fixes an issue Cailyn is having where the she cannot accept the 300th competitor, because the non-competing registration in `WarmUpSeattle2025` is still counted in the competitor_limit check

Note: This should also be added in: 
- [x] Bulk update validation - although this seems to have disappeared from the registration_checker file??? 
- [ ] Competitor limit validation in #10977

### DIFF
--- a/lib/registrations/registration_checker.rb
+++ b/lib/registrations/registration_checker.rb
@@ -116,7 +116,7 @@ module Registrations
 
         if new_status == Registrations::Helper::STATUS_ACCEPTED && competition.competitor_limit_enabled?
           raise WcaExceptions::RegistrationError.new(:forbidden, Registrations::ErrorCodes::COMPETITOR_LIMIT_REACHED) if
-            competition.registrations.competing_status_accepted.count >= competition.competitor_limit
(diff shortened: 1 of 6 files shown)


lib/registrations/registration_checker.rb
```diff
@@ -116,7 +116,7 @@ def validate_update_status!(new_status, competition, current_user, target_user,
 
         if new_status == Registrations::Helper::STATUS_ACCEPTED && competition.competitor_limit_enabled?
           raise WcaExceptions::RegistrationError.new(:forbidden, Registrations::ErrorCodes::COMPETITOR_LIMIT_REACHED) if
-            competition.registrations.competing_status_accepted.count >= competition.competitor_limit
+            competition.registrations.competing_status_accepted.competing.count >= competition.competitor_limit
```
FinnIckler, line 119: why not use the method you just created above?

dunkOnIT: Good point, changed!

gregorbg (approved):

PR #11123, opened 2025-03-24
Add example Persons Results page for Next

FinnIckler:
I took the response straight from the persons.rb controller. I do not think we should be doing this as this causes a 50000 line pretty printed JSON for my profile (which takes 600ms even when cached and like 7 seconds on first load), but it does enable to work on the new persons page.

We can definitely reduce the load time on the Rails end by serializing better, but it will still be quite slow for next to digest this data. 

The better solution is probably to only server side render the basic person info and have a paginated (or infinite scroll?) result loading. Because 99.9% of that 50k line JSON is results. 

 Even if we do want to SSR results, we should only do it for the first event and pull in the rest.


FinnIckler:
looks like we do already have API endpoints for all of this, so I will remodel this PR accordingly

FinnIckler:
much better and much more performant now with the already existing APIs and fetching results from the client. 
I only like that `Api::V0::PersonsController#show` is still doing 17 queries, but I don't know how to write the includes to fix that

config/routes.rb
```diff
@@ -343,16 +343,6 @@
...
-    namespace :internal do
-      namespace :v1 do
-        get '/users/:id/permissions' => 'permissions#index'
-        get '/competitions/:competition_id' => 'competitions#show'
-        get '/competitions/:competition_id/qualifications' => 'competitions#qualifications'
-        post '/users/competitor-info' => 'users#competitor_info'
-        post '/mailers/registration' => 'mailers#registration'
-        post '/payment/init_stripe' => 'payment#init_stripe'
-      end
-    end
```
gregorbg, line 355: Where did these routes go?

config/initializers/cors.rb
```diff
@@ -0,0 +1,38 @@
...
+        '/api/v1/*',
+        headers: %w[Origin X-Requested-With Content-Type Accept Authorization],
+        methods: %i[get post delete put patch options head],
+        expose: %w[Total Per-Page Link],
+        max_age: 0,
+        credentials: false,
+      )
+    end
+  end
+end
```
gregorbg, line 38: What about flipping the structure, something like...
```ruby
Rails.application.config.middleware.insert_before 0, Rack::Cors do
  allow do
    origins '*'
    
    if Rails.env.development?
      ...
    else
      ...
    end
  end
end
```

next-frontend/src/lib/wca/persons/getResultsByPerson.tsx
```diff
@@ -0,0 +1,8 @@
+import { serverClient } from "@/lib/wca/wcaAPI";
+import { cache } from "react";
+
+export const getResultsByPerson = cache(async (wca_id: string) => {
```
gregorbg, line 4: Mh, not happy about snake-casing in JS function params

next-frontend/src/lib/wca/wcaAPI.ts
gregorbg, line 1: Just curious, how did these changes come about?

config/routes.rb
FinnIckler: well, I started out this PR by adding the results route as an internal route. Then I noticed that we had all these dead routes under the internal namespace that we forgot to delete in v1 cleanup

gregorbg (approved):

PR #11126, opened 2025-03-25
Clean up competition form

gregorbg:
This started as a "let's move from `useSaveAction` to mutations!" and ended up with a bigger refactor than I care to admit.

Nothing about the form has changed from an end-user perspective. It should :tm: behave exactly the same. Major things that have changed code-wise:
- The form fully relies on Tanstack `react-query` now.
- The `admin` section has been removed from the main competition serialization routine. It only contained WCAT information like "Is this comp confirmed?" or "Is this comp marked as publicly visible?", which now lives under its own separate `confirmation_data` endpoint with its own JSON mini-format.
- A lot of code has been moved around between files and from one component to another. Roughly 60% of the diff is just purely moved code or at least moved code with a variable rename in there.


dunkOnIT:
Frontend testing done - all looks good except for one issue: 

- When trying to `Confirm` the comp, we get an error from the backend that `refund_policy_percent` must be a numeric value
- Even though the frontend field has `defaultValue={0}` specified, the field appears to have a `NULL` value until it the user puts a different number in
- Not sure if this was an issue before these changes or not
- **Suggested fix:** Migrate `refund_policy_percent` to have a `default` value of `0`
  - If we wanted to propagate this change back in time, we could replace `NULL` with `0` in the `refund_policy_percent` column - but this doesn't seem necessary given that this issue only affects newly-created competitions

dunkOnIT (approved):

FinnIckler:
I'm doing the last react on rails migration now and I noticed that this PR moved the index.js to MainForm.js.

So 
```
<% add_to_packs("wca_maps") %>

<%= react_component("CompetitionForm", {
  competition: @competition.to_form_data,
  storedEvents: @competition.events,
  isAdminView: @competition_admin_view,
  isPersisted: @competition.persisted?,
  isSeriesPersisted: @competition.competition_series&.persisted? || false,
  isCloning: @competition.being_cloned_from.present?,
}) %>
```
in `_competition_form.html.erb` is dead code. Am I correct? @gregorbg

PR #11140, opened 2025-03-25
Normalize Bundler 2.6 lockfile

gregorbg:
Every time you run `bundle install` locally, it currently gives the following error message (with the exact platform replaced by your local platform):
```
The following platform specific gems are getting installed, yet the lockfile includes only their generic ruby version:
 * nokogiri-1.18.5-x86_64-linux-gnu
Please run `bundle lock --normalize-platforms` and commit the resulting lockfile.
Alternatively, you may run `bundle lock --add-platform <list-of-platforms-that-you-want-to-support>`
```

So I did that and ran `bundle lock --normalize-platforms`. Now we officially support everything Bundler supports. Is that alright or do we want to hand-pick? We probably only need to support GNU Linux because that's what our container builds do and that's what Docker Compose for local dev does.

And since Bundler 2.6 introduced lockfile hash sum support, I also went ahead and introduced that per https://bundler.io/blog/2024/12/19/bundler-v2-6.html.

And updated from Bundler 2.6.3 to Bundler 2.6.6.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1008,4 +1008,4 @@ DEPENDENCIES
   wkhtmltopdf-binary-ng
 
 BUNDLED WITH
-   2.6.3
+   2.6.6
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -314,6 +314,16 @@ GEM
     faraday-retry (2.2.1)
       faraday (~> 2.0)
     ffi (1.17.1)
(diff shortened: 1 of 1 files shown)


FinnIckler:
I think we should go for everything that bundler supports. For example I had the issue that my bundler file changes when I run bundle install on my ARM Mac (if I want to install rubocop). Probably a similar issue if someone runs Windows on ARM or something.

PR #11147, opened 2025-03-25
Separate status update permissions checks and validations

dunkOnIT:
@gregorbg per your request this just splits the permissions and validation checks up into two separate functions. I have (redundantly) left them both in the `validate_update_status` function, just to make it obvious what the names of the new functions are.


dunkOnIT:
> Yay! That looks... deceptively easy 😅
Yeah it turns out I'd already essentially structured it the way we want - only going into the user permissions level if the payload was itself valid, so it was pretty easy. The merge conflicts were harder :P

gregorbg (approved):

PR #11149, opened 2025-03-25
Rate limit explorations

FinnIckler:
We had a report months ago that we should rate limit the sign in/up route which I am right limiting here to 10 a minute.

I think rate limiting the API to 1 request a second over a 60 second period is probably also reasonable?


dunkOnIT:
I think both limits are reasonable

gregorbg (approved):

PR #11154, opened 2025-03-26
Move registration_checker competing status validations to model

gregorbg:
Part of our fast-paced effort to get rid of (the essence of) `registration_checker`.
Adds validations for competitor limit, newcomer spots, series registrations. But only triggers those if you're setting the status to `accepted`.

The series registration check was historically appended under `competition_id`, but I think we should change it to `competing_status`. You are allowed to _register_ for multiple competitions in a Series after all. You're just not allowed to be _accepted_ for more than one at a time. So the competing status is what breaks the camels neck, while the competition that you're trying to be accepted for is legit in itself.

### DIFF
--- a/app/models/registration.rb
+++ b/app/models/registration.rb
@@ -471,12 +471,16 @@ class Registration < ApplicationRecord
     competing_status_changed? && (competing_status_cancelled? || competing_status_rejected?)
   end
 
-  validate :not_changing_events_when_cancelling, if: [:trying_to_cancel?, :competition_events_changed?], unless: :new_record?
+  validate :not_changing_events_when_cancelling, if: [:trying_to_cancel?, :tracked_event_ids?, :competition_events_changed?]
   private def not_changing_events_when_cancelling
-    errors.add(:competition_events, :cannot_change_when_cancelling, message: I18n.t('registrations.errors.cannot_change_events_when_cancelling'), frontend_code: Registrations::ErrorCodes::INVALID_REQUEST_DATA)
+    errors.add(:competition_events, :cannot_change_events_when_cancelling, message: I18n.t('registrations.errors.cannot_change_events_when_cancelling'), frontend_code: Registrations::ErrorCodes::INVALID_REQUEST_DATA)
   end
 
   attr_writer :tracked_event_ids
+
+  def tracked_event_ids?
+    @tracked_event_ids.present?
+  end
 
   def tracked_event_ids
     @tracked_event_ids ||= self.event_ids
--- a/app/models/registration.rb
+++ b/app/models/registration.rb
@@ -14,6 +14,7 @@ class Registration < ApplicationRecord
   scope :not_cancelled, -> { where.not(competing_status: 'cancelled') }
   scope :with_payments, -> { joins(:registration_payments).distinct }
   scope :wcif_ordered, -> { order(:id) }
+  scope :might_attend, -> { where(competing_status: ['accepted', 'waiting_list']) }
 
   belongs_to :competition
   belongs_to :user, optional: true # A user may be deleted later. We only enforce validation directly on creation further down below.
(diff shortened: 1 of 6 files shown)


app/models/registration.rb
```diff
@@ -409,35 +410,63 @@ def consider_auto_close
...
 
-    if registration_status.nil?
-      return sibling_registrations
-             .joins(:competition)
-             .order(:start_date)
-    end
+    # There are a limited number of "reserved" spots for newcomer_month_eligible competitions
+    # We know that there are _some_ available_spots in the comp available, because we passed the competitor_limit check above
+    # However, we still don't know how many of the reserved spots have been taken up by newcomers, versus how many "general" spots are left
+    # For a non-newcomer to be accepted, there need to be more spots available than spots still reserved for newcomers
```
gregorbg, line 432: This comment has been ported over entirely from `registration_checker`.

app/models/registration.rb
```diff
@@ -409,42 +403,100 @@ def consider_auto_close
...
   def ensure_waitlist_eligibility!
     raise ArgumentError.new("Registration must have a competing_status of 'waiting_list' to be added to the waiting list") unless
       competing_status == Registrations::Helper::STATUS_WAITING_LIST
   end
 
+  def trying_to_cancel?
+    competing_status_changed? && (competing_status_cancelled? || competing_status_rejected?)
+  end
+
+  validate :not_changing_events_when_cancelling, if: [:trying_to_cancel?, :tracked_event_ids?, :competition_events_changed?]
```
dunkOnIT, line 473: Do we need `:tracked_event_ids?` in here? It seems to be checking for presence, but 
- the way tracked_event_ids is defined makes me think it will always be present? There's no way for a registration to be created without event_ids? 
- Shouldn't `competition_events_changed?` return `false` if there are no `tracked_event_ids`?

lib/registrations/registration_checker.rb
```diff
@@ -12,13 +12,16 @@ def self.apply_payload(registration, raw_payload)
         competing_payload = raw_payload['competing']
         comment = competing_payload&.dig('comment')
         organizer_comment = competing_payload&.dig('organizer_comment')
+        competing_status = competing_payload&.dig('status')
 
         new_registration.comments = comment if competing_payload&.key?('comment')
         new_registration.administrative_notes = organizer_comment if competing_payload&.key?('organizer_comment')
+        new_registration.competing_status = competing_status if competing_payload&.key?('status')
```
dunkOnIT, line 19: Could we not just do `competing_status.present?`

app/models/registration.rb
gregorbg: Yes, we do need it, because `competition_events_changed?` otherwise triggers too early. It has a side-effect due to the inherently caching nature of Rails, and the only way to circumvent that would be by reloading the association fresh every time the comparison is run. In the big picture, that would be sinfully expensive in terms of DB calls.

Briefly put, think of this presence check as "Do we have anything to compare against in the first place?". And only if the answer to that question is Yes, we proceed with running the actual comparison.

(The last elements of the `if` array don't get checked as soon as one of them is falsey)

lib/registrations/registration_checker.rb
gregorbg: In this specific case, we could. But in general we can't, because for some fields (like organizer comments for example, where the organizer accidentally entered something wrong and wants to undo the comment) we want to allow deliberately setting the value to `nil`. But `nil.present?` returns false, so the explicitly, intently passed `nil` would never be written.
In the context of competing_status specifically, `nil` doesn't make sense but for the sake of consistency we use `key?` like everywhere else.

app/models/registration.rb
dunkOnIT: How would we end up in the situation where we don't have anything to compare against - ie, where `tracked_event_ids` is `nil`?

gregorbg: Virtually anywhere in our code. The only case when we do have something to compare is when we intentionally write the value from the outside (`attr_writer` above). And currently, that only happens in `registration_checker`s request loading routine. All other APIs that do somethingTM with registrations don't write this and thus shouldn't be compared.

For reference, even a simple `FactoryBot.create(:registration)` in our test suite triggers validations, and it will needlessly run this comparison if the check you're commenting on isn't there.

dunkOnIT (approved):

app/models/registration.rb
dunkOnIT: This took me a second to understand, so posting my explanation here in case anyone looks at this thread in the future: 
- We have the function tracked_event_ids (line 484)
- Separately, we have the instance variable @tracked_event_ids - this is what the attr_writer accesses
- tracked_event_ids method will return @tracked_event_ids if it is present, otherwise it will fall back to the registrations event_ids association
- tracked_event_ids? is checking the present of the @tracked_event_ids instance variable specifically - and in cases where that variable is nil, no further comparison/action is needed

PR #11159, opened 2025-03-26
Move waiting list position validations to Rails model

gregorbg:
Introduces a new `Waitlistable` trait that manages the validations. Should play nicely with our generic approach to waiting lists.

The hook should be able to do the waiting list management in an easy `after_save` hook as well, but I disabled that one for now so that we don't introduce too many changes in one PR.

But in theory, this could even replace the `try(:ensure_waitlist_eligibility)` approach from #10241 by using some form of `def waitlistable_id` and passing that directly to the waiting list interface.


FinnIckler (approved):
LGTM

PR #11160, opened 2025-03-26
Rename Scramble Table Columns to follow Rails standards

FinnIckler:


FinnIckler:
I tried changing the type to `:primary_key`, but rails complained that there was already a primary key (even though I was changing the column that was the primary key...).

gregorbg (approved):

Issue #11156, opened 2025-03-26
Update CSV export


dunkOnIT:
- [ ] Add user_id field
- [ ] Add "competing_status" field (ie, a field that gives the full name of the status, not just the 1-letter abbreviation)

pranav-027:
@dunkOnIT Sure, assigning it to me. Can you confirm what user_id is supposed to be? registration id? id from the user table? or just serial number

I also found out this one bug- where the cell header is shifted, probably because of my recent development. fixing it also!
![Image](https://github.com/user-attachments/assets/841ff9a3-8bf9-4505-97f1-3150da8db3d9)

dunkOnIT:
Great, thanks! `user_id` should be id from the user table - I believe it's in the `Registration` object 

One last thing - could you also differentiate between the `registration_timestamp` and `payment_timestamp`? ie, each one should have a column

Issue #8401, opened 2023-10-02
Registration export should have a 'd' status for deleted registrations


OJCubing:
On the registrations page for a competition, if I click select all on waiting list, approved registrations and deleted registrations, then Export to CSV, the Status column has only two values: p (for pending; everyone on the waiting list) and a (for approved; **both** everyone on the approved and deleted registrations list). I'm not sure if this is a bug or just a lacking feature, but it seems clear to me that the deleted registrations should have a status value of d.

dunkOnIT:
This will be addressed in as part of #8288, marked as in progress

PR #11165, opened 2025-03-27
CSV export update

pranav-027:
fixes: https://github.com/thewca/worldcubeassociation.org/issues/11156
image

### DIFF
--- a/app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
+++ b/app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
@@ -16,7 +16,7 @@ import {
 import { useConfirm } from '../../../lib/providers/ConfirmProvider';
 
 function V3csvExport(selected, registrations, competition) {
-  let csvContent= `Status,Name,Country,WCA ID,Birth Date,Gender,${competition.event_ids.join(',')},Email,Guests,IP,Registration Date Time (UTC),Payment Date Time(UTC),User Id, Registration Status\n`;
+  let csvContent= `Status,Name,Country,WCA ID,Birth Date,Gender,${competition.event_ids.join(',')},Email,Guests,IP,Registration Date Time (UTC),Payment Date Time(UTC),User Id,Registration Status\n`;
   registrations
     .filter((r) => selected.length === 0 || selected.includes(r.user_id))
     .forEach((registration) => {
--- a/app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
+++ b/app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
@@ -16,7 +16,7 @@ import {
 import { useConfirm } from '../../../lib/providers/ConfirmProvider';
 
 function V3csvExport(selected, registrations, competition) {
-  let csvContent= `Status,Name,Country,WCA ID,Birth Date,Gender,${competition.event_ids.join(',')},Email,Guests,IP,Registration Date Time (UTC),Payment Date Time(UTC),User Id,Registration Status\n`;
+  let csvContent = `Status,Name,Country,WCA ID,Birth Date,Gender,${competition.event_ids.join(',')},Email,Guests,IP,Registration Date Time (UTC),Payment Date Time(UTC),User Id,Registration Status\n`;
   registrations
     .filter((r) => selected.length === 0 || selected.includes(r.user_id))
     .forEach((registration) => {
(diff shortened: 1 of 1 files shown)


dunkOnIT (changes requested):
Thanks a lot for this! Please could you: 
- [ ] Have competing_status be a _new_ column, in addition to the existing `status` column we have (we don't want to break anyone's existing template/script)
- [ ] Add a `registration_timestamp` column - so that organizers have this information in addition to payment timestamps

pranav-027:
@dunkOnIT, Done as suggested. 

About the status code field can we add code for cancelled and rejected?
```jsx
const regStatusMap = {
  accepted: 'a',
  pending: 'p',
  cancelled: 'd',
  waiting: 'p',
  rejected: 'd'
};

regStatusMap[registration.competing.registration_status]
```

something like this?

dunkOnIT:
Yeah the status map is a great idea, let's add that!

pranav-027:
@dunkOnIT Done!

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
```diff
@@ -15,14 +15,19 @@ import {
 } from '../../../lib/utils/registrationAdmin';
 import { useConfirm } from '../../../lib/providers/ConfirmProvider';
 
+const registrationStatusMap = {
+  accepted: 'a',
+  pending: 'p',
+  cancelled: 'd',
+  waiting_list: 'p',
+  rejected: 'd',
+};
```
gregorbg, line 24: This is nice from code layout perspective, but where does `d` come from? The old code does `registration.competing.registration_status === 'accepted' ? 'a' : 'p'`, so the only possible values were `a` and `p`.

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
```diff
@@ -15,14 +15,19 @@ import {
...
 function V3csvExport(selected, registrations, competition) {
-  let csvContent = 'data:text/csv;charset=utf-8,';
-  csvContent
-    += `Status,Name,Country,WCA ID,Birth Date,Gender,${competition.event_ids.join(',')},Email,Guests,IP,Registration Date Time (UTC)\n`;
+  let csvContent = `User Id,Status,Name,Country,WCA ID,Birth Date,Gender,${competition.event_ids.join(',')},Email,Registration Status,Guests,IP,Registration Date Time (UTC),Payment Date Time (UTC)\n`;
```
gregorbg, line 26: Order is important here!

gregorbg (changes requested):
I am very picky because of backwards compatibility. Two remarks:
1. I'm afraid that we may break people's setups if we introduce a new status `d` in an old column. As far as I can see, only `a` and `p` were used before, which is effectively a wannabe-boolean (either it's one or it's the other, there is no third option). We may veeeeeery subtly break some hobby developers' code when they do `if csv_status == "a" else // doing stuff assuming it must be "p"`
2. Adding columns in the middle of the existing columns can break implementations that rely on the order of columns. Yes, it's not clean to rely on the order of columns in the first place, but not every reg system developer is an experienced software veteran with years of experience. Again, I wouldn't be surprised if some people try accessing `row[0]` expecting to get the `Status`, but now they will suddenly get a user ID instead.
Solution: If you want to add new single-letter codes, add them as a new column. Definitely leave the old column, and introduce the new one. Give people time to adjust to the new one and then phase out the old one after a while.
Also, add new columns at the end.

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
pranav-027: Done!

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
pranav-027: I got this idea of  `d`  from https://github.com/thewca/worldcubeassociation.org/issues/8401. But I have removed it now if it's not acceptable.

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
```diff
@@ -16,9 +16,7 @@ import {
 import { useConfirm } from '../../../lib/providers/ConfirmProvider';
 
 function V3csvExport(selected, registrations, competition) {
-  let csvContent = 'data:text/csv;charset=utf-8,';
```
gregorbg, line 19: Sidenote, why did this MIME type and UTF-8 preamble have to go?
I feel like the answer is "because we are generating it outside of this method now". But then, how is it not completely broken because the header would appear twice (once generated outside and once inside the method) right now??

gregorbg: so it is completely broken right now...

gregorbg (approved):

PR #11167, opened 2025-03-27
Edit Registration page minor changes

dunkOnIT:
2 changes: 
- [ ] Move strings in Edit Registration to i18n keys
- [ ] Change header to have a link to the profile, and an edit link

### DIFF
--- a/app/webpacker/components/RegistrationsV2/RegistrationEdit/RegistrationEditor.jsx
+++ b/app/webpacker/components/RegistrationsV2/RegistrationEdit/RegistrationEditor.jsx
@@ -207,9 +207,11 @@ export default function RegistrationEditor({ competitor, competitionInfo }) {
         )}
         <Header>
           {`${competitor.name} `}
-          (<a href={personUrl(competitor.wca_id)} target="_blank" rel="noreferrer" className="hide-new-window-icon">{competitor.wca_id}</a>)
+          (
+            <a href={personUrl(competitor.wca_id)} target="_blank" rel="noreferrer" className="hide-new-window-icon">{competitor.wca_id}</a>
+          )
           {' ' /* Necessary to space the icon away from the wca_id */ }
-          <a href={personUrl(competitor.wca_id)} target="_blank" rel="noreferrer" className="hide-new-window-icon"><Icon name="edit"/></a>
+          <a href={personUrl(competitor.wca_id)} target="_blank" rel="noreferrer" className="hide-new-window-icon"><Icon name="edit"/ ></a>
         </Header>
         <Form.Field required error={selectedEventIds.size === 0}>
           <EventSelector
--- a/app/webpacker/components/RegistrationsV2/RegistrationEdit/RegistrationEditor.jsx
+++ b/app/webpacker/components/RegistrationsV2/RegistrationEdit/RegistrationEditor.jsx
@@ -8,7 +8,9 @@ import React, {
 import {
   Button,
   Form,
-  Header, List,
+  Header,
+  Icon,
+  List,
   Message,
   Segment,
 } from 'semantic-ui-react';
@@ -18,7 +20,7 @@ import { showMessage } from '../Register/RegistrationMessage';
 import Loading from '../../Requests/Loading';
 import EventSelector from '../../wca/EventSelector';
 import Refunds from './Refunds';
-import { editPersonUrl } from '../../../lib/requests/routes.js.erb';
+import { personUrl, editPersonUrl } from '../../../lib/requests/routes.js.erb';
 import { useConfirm } from '../../../lib/providers/ConfirmProvider';
 import I18n from '../../../lib/i18n';
 import RegistrationHistory from './RegistrationHistory';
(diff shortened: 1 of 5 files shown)


app/webpacker/components/RegistrationsV2/RegistrationEdit/RegistrationEditor.jsx
```diff
@@ -199,13 +201,40 @@ export default function RegistrationEditor({ competitor, competitionInfo }) {
           </Message>
         )}
         {registrationEditDeadlinePassed && (
+          // i18n-tasks-use t('registrations.errors.edit_deadline_passed')
```
FinnIckler, line 204: you only need i18n-tasks-use for dynamic uses. This should be fine

dunkOnIT: Changed, thanks!

config/locales/en.yml
```diff
@@ -2237,6 +2244,10 @@ en:
     connect_wca_id: "Connect your WCA ID to your account!"
   payments:
     #context: localized names of payment providers
```
kr-matthews, line 2246: This is meant to be attached to `payment_providers`.

app/webpacker/components/RegistrationsV2/RegistrationEdit/RegistrationEditor.jsx
```diff
@@ -200,12 +202,38 @@ export default function RegistrationEditor({ competitor, competitionInfo }) {
...
-            <strong>Changes should only be made in extraordinary circumstances.</strong>
+            {I18n.t('registrations.errors.edit_deadline_passed')}
           </Message>
         )}
-        <Header>{competitor.name}</Header>
+        <Header>
+          {`${competitor.name} `}
+          (
+          {/* The onClick functions prevent Semantic UI from rendering large icons indicating new
+          tabs */}
```
gregorbg, line 212: This sounds very wrong. I suspect that it has something to do with how you're using `onClick`.

app/webpacker/components/RegistrationsV2/RegistrationEdit/RegistrationEditor.jsx
```diff
@@ -200,12 +202,38 @@ export default function RegistrationEditor({ competitor, competitionInfo }) {
...
+          <a
+            href={personUrl(competitor.wca_id)}
+            onClick={(e) => {
+              e.preventDefault();
+              window.open(personUrl(competitor.wca_id), '_blank', 'noopener,noreferrer');
+            }}
```
gregorbg, line 218: Why are you doing `onClick` here in the first place?
For reference, I think our entire React codebase right now doesn't have a single instance of an `a` tag overriding `onClick`. And unless you have a very, _very_, *very* good reason, it should also stay that way :smile:

gregorbg: If your only intention is to open the link in a new tab, then `<a href={personUrl(...)} target="_blank" rel="noreferrer">` will do

app/webpacker/components/RegistrationsV2/RegistrationEdit/RegistrationEditor.jsx
```diff
@@ -200,12 +202,38 @@ export default function RegistrationEditor({ competitor, competitionInfo }) {
...
+          <a
+            href={editPersonUrl(competitor.id)}
+            aria-label="Edit"
+            role="menuitem"
+            className="edit-link"
+            onClick={(e) => {
+              e.preventDefault();
+              window.open(editPersonUrl(competitor.id), '_blank', 'noopener,noreferrer');
+            }}
+          >
```
gregorbg, line 233: See above about overriding `onClick`

config/locales/en.yml
gregorbg: I second that. You have to move this comment down to where it originally belonged

gregorbg (changes requested):

app/webpacker/components/RegistrationsV2/RegistrationEdit/RegistrationEditor.jsx
dunkOnIT: With normal `<a href=... target="_blank" rel="noreferrer">` tags, semUI automatically inserts these icons indicating that the link will be opened in a new tab: 
image

That used this code: 
```jsx
          <a href={personUrl(competitor.wca_id)} target="_blank" rel="noreferrer">{competitor.wca_id}</a>
          <a href={personUrl(competitor.wca_id)} target="_blank" rel="noreferrer"><Icon name="edit" /></a>
```

However, overriding it the way I did[1] means that semUI doesn't insert those icons, making the header much more readable:  
image


[1] following ChatGPT's suggestion ofc - I don't know enough JS to do that myself, so it is very possible there are better methods of avoiding these icons

Overriding it the way I have here

config/locales/en.yml
dunkOnIT: Fixed

app/webpacker/components/RegistrationsV2/RegistrationEdit/RegistrationEditor.jsx
gregorbg: Oh, that's isn't SemUI... It's us! :sweat_smile: https://github.com/thewca/worldcubeassociation.org/blob/main/app/assets/stylesheets/links.scss

You can get rid of these icons by either not opening the thing in a new tab (duh) or by adding `className="hide-new-window-icon"`

dunkOnIT: Best news ever! Updated!

app/webpacker/components/RegistrationsV2/RegistrationEdit/RegistrationEditor.jsx
dunkOnIT: Updated

gregorbg (approved):

PR #11173, opened 2025-03-28
Reuse registration payload writing logic in create/update steps

gregorbg:
For the previous `registration_checker` migrations, we created logic that allows "applying" a payload to a registration.
Incidentally, that's just the same thing that the queue job handler / updater is doing, except that it was doing it separately.

This PR aims for some basic code reuse and cleanup of now-duplicated features. It does not yet touch on transaction/rollback-based validation.

(Didn't run CI locally, let's see how many tests will fail first try!)

### DIFF
--- a/app/models/concerns/waitlistable.rb
+++ b/app/models/concerns/waitlistable.rb
@@ -36,13 +36,14 @@ module Waitlistable
     validates :waiting_list_present?, presence: { if: :waiting_list_position?, frontend_code: Registrations::ErrorCodes::INVALID_REQUEST_DATA }
 
     after_save :commit_waitlist_position, if: :waiting_list_persisted?
+    after_commit :clear_tracked_waitlist_position!, on: :update
 
     def commit_waitlist_position
-      self.waiting_list_position = @tracked_waiting_list_position
+      self.waiting_list_position = self.tracked_waitlist_position
     end
 
     private def clear_tracked_waitlist_position!
-      @tracked_waiting_list_position = nil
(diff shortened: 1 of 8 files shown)


app/models/concerns/waitlistable.rb
```diff
@@ -63,7 +54,34 @@ def waitlistable?
...
+      should_add = self.waitlistable? && target_position.nil?
+      should_move = self.waitlistable? && target_position.present?
+      should_remove = !self.waitlistable? && target_position.present?
+
+      self.waiting_list.add(self) if should_add
+      self.waiting_list.move_to_position(self, target_position) if should_move
+      self.waiting_list.remove(self) if should_remove
+
+      # Now that we definitely updated the "real" WL, we can forget about our tracking safely.
+      self.clear_tracked_waitlist_position!
```
dunkOnIT, line 84: Explore whether we can make this conditional on a certainty of whether _something_ happened to persist a change to the database in lines 79-81

gregorbg: Done! https://github.com/thewca/worldcubeassociation.org/pull/11173/commits/6586e164f2e5d0f0b9c2b50d6932486d85c03581

gregorbg: And improved! https://github.com/thewca/worldcubeassociation.org/pull/11173/commits/2158abdd6acbaed013e4b061fa5162def7ada4cb

dunkOnIT (approved):
Tested over the weekend but forgot to approve - I think it's good to go!

PR #11179, opened 2025-03-29
Include toSorted and toReversed polyfills manually

gregorbg:
The explanation for this PR is given directly in the code comment. We can remove these polyfills (revert this PR, unless more polyfills have been added in the meantime) when a new Shakapacker version is released.


gregorbg:
Pushing as a hotfix (and because I know that there will be a Shakapacker update soon that will allow us to get rid of this patch)

PR #11184, opened 2025-03-30
Reset change tracking after cloning in RegistrationChecker

gregorbg:
Re email thread `BUG: Can't update registration events for comp that is full`: The subject of the email is a surprisingly accurate description of the underlying bug.

The registration checker creates in-memory clones of registrations, so that we don't have to worry about accidental write operations. This cloning however behaves in a way that even an *already accepted* registration being cloned thinks that its status *changed* to being accepted.
In case of already full competitions, this leads to the original error that any change whatsoever to the registration is being rejected, because the "competitor limit reached" check kinda "pollutes" the validation result.

More details are in the code comment for those who are interested.

### DIFF
--- a/lib/registrations/registration_checker.rb
+++ b/lib/registrations/registration_checker.rb
@@ -5,6 +5,15 @@ module Registrations
     def self.apply_payload(registration, raw_payload)
       # Duplicate everything to make sure we don't trigger unwanted DB write operations
       registration.deep_dup.tap do |new_registration|
+        # Weird quirk in Rails: `deep_dup` initializes a "blank" Registration under the hood
+        #   and _then_ fills its attributes with the model it's being cloned from. In practice,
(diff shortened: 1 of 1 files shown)


gregorbg:
Sidenote: This quirk in the cloning behavior also affects any other registration attribute, for example comments. A comment is `nil` by default, so deep-cloning a registration that already has a comment `Hello` attached will report "The comment changed from `nil` to `Hello`" on the clone.
For attributes such as comments, this does not have a significant impact right now, because we currently don't trigger any validations based on "has the comment changed?". If we did, the bug there would haunt us all the same.

dunkOnIT (approved):
Tested locally and it works

FinnIckler:
Don't we want to add a new test a la "can still update registration if competition is full"?

PR #11209, opened 2025-03-31
Mount Swagger UI for our OpenAPI spec

gregorbg:
Just had a feeling this will come in handy later down the line. Using https://www.npmjs.com/package/swagger-ui-react and mounting our OpenAPI document publicly.

image


next-frontend/package.json
```diff
@@ -43,5 +45,8 @@
     "prettier": "^3.5.3",
     "typescript": "^5"
   },
+  "scarfSettings": {
```
FinnIckler, line 48: what are scraf settings and why do we need to disable them here?

gregorbg: It disables automated telemetry gathering. Unfortunately, `package.json` doesn't support comments so I couldn't explain that directly in the file

FinnIckler (approved):

PR #11212, opened 2025-03-31
Admin Search (search with email) in anonymization page

danieljames-dj:
This is an attempt to use a new search endpoint through which we can search for emails and there is no cache. This new search endpoint will be used if `emailSearchEnabled` is enabled.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -4,6 +4,7 @@ class UsersController < ApplicationController
   before_action :authenticate_user!, except: %i[select_nearby_delegate acknowledge_cookies]
   before_action :check_recent_authentication!, only: %i[enable_2fa disable_2fa regenerate_2fa_backup_codes]
   before_action :set_recent_authentication!, only: %i[edit update enable_2fa disable_2fa]
+  before_action -> { redirect_to_root_unless_user(:can_admin_results?) }, only: %i[admin_search]
 
   RECENT_AUTHENTICATION_DURATION = 10.minutes.freeze
 
@@ -364,5 +365,15 @@ class UsersController < ApplicationController
       return false
     end
     true
+  end
+
+  def admin_search
+    query = params[:q]&.slice(0...SearchResultsController::SEARCH_QUERY_LIMIT)
+
+    return render status: :bad_request, json: { error: "No query specified" } unless query
+
+    render status: :ok, json: {
+      result: User.search(query, params: params).limit(SearchResultsController::SEARCH_RESULT_LIMIT),
+    }
   end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1182,12 +1182,16 @@ class User < ApplicationRecord
   end
 
   def self.search(query, params: {})
-    users = Person.includes(:user).current
-    # We can't search by email on the 'Person' table
-    search_by_email = false
-    unless ActiveRecord::Type::Boolean.new.cast(params[:persons_table])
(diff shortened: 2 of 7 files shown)


app/controllers/users_controller.rb
```diff
@@ -365,4 +366,20 @@ def acknowledge_cookies
     end
     true
   end
+
+  def admin_search
```
FinnIckler, line 370: this should have a authorization check right?

danieljames-dj: Oh yes, right. I've added now. Thanks.

app/models/user.rb
```diff
@@ -1168,12 +1168,16 @@ def self.trainee_delegate_ids
...
 
   def self.search(query, params: {})
-    users = Person.includes(:user).current
-    # We can't search by email on the 'Person' table
-    search_by_email = false
-    unless ActiveRecord::Type::Boolean.new.cast(params[:persons_table])
+    search_by_email = ActiveRecord::Type::Boolean.new.cast(params[:email])
+    admin_search = ActiveRecord::Type::Boolean.new.cast(params[:adminSearch])
+
+    return User.where(email: query) if admin_search && search_by_email
```
gregorbg, line 1174: Why is it necessary to return early if you're (admin) searching via email? What prevents admins from searching by WCA ID and email at the same time, for example?

app/models/user.rb
```diff
@@ -1168,12 +1168,16 @@ def self.trainee_delegate_ids
...
+
+    if ActiveRecord::Type::Boolean.new.cast(params[:persons_table])
```
gregorbg, line 1176: This thing within the `if` can probably be converted to its own `searching_persons_table` boolean variable. That same variable can then be used above already as `search_by_email = searching_persons_table && ActiveRecord::Type::Boolean.new.cast(...)`

gregorbg (changes requested):

app/models/user.rb
danieljames-dj: Irrespective of whether search by email or not, I'm introducing here `admin_search` which is supposed to be optimized. In this case, I'm searching by email, and I just want to return the actual value. Later, if the admin is searching by WCA ID, I can implement something which returns only one result and hence reduce unwanted querying in the whole DB.

app/controllers/users_controller.rb
```diff
@@ -4,8 +4,10 @@ class UsersController < ApplicationController
   before_action :authenticate_user!, except: [:select_nearby_delegate, :acknowledge_cookies]
   before_action :check_recent_authentication!, only: [:enable_2fa, :disable_2fa, :regenerate_2fa_backup_codes]
   before_action :set_recent_authentication!, only: [:edit, :update, :enable_2fa, :disable_2fa]
+  before_action -> { redirect_to_root_unless_user(:can_admin_results?) }, only: [:admin_search]
 
   RECENT_AUTHENTICATION_DURATION = 10.minutes.freeze
+  DEFAULT_SEARCH_RESULT_LIMIT = 20
```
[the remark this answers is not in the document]
danieljames-dj: I copied it from `api_controller`. I guess you wanted to suggest `SearchResultsController::SEARCH_RESULT_LIMIT` and not `SearchResultsController::SEARCH_QUERY_LIMIT`. I've changed it to `SearchResultsController::SEARCH_RESULT_LIMIT`.

app/models/user.rb
danieljames-dj: Changed to own boolean as you suggested, also noted the sidenote.

gregorbg (approved):

PR #11218, opened 2025-03-31
 Rename concise_single_records and concise_average_records table columns to follow Rails Conventions

FinnIckler:


FinnIckler:
This PR also does not affect the Results Export!

gregorbg (approved):

PR #11226, opened 2025-04-02
Update Yarn to v4.8.1

gregorbg:
See title.


gregorbg:
Treating as a (glorified) dependency update

PR #11227, opened 2025-04-02
New Brand Guide Implementation into Next Frontend

ando527:
Welcome to the first commit using the new WCA Brand styleguide and guidelines!

Things included here:

### Chakra Theme File
This includes all the colours and styling needed to implement majority of the core components. Links, Buttons, Cards, Tables, Badges, Accordion, Tabs etc...

### Icon Library
The new brand package comes with a full suite of icons for use around the website, and these have been implemented using the CreateIcon utility within Chakra
image
I have also included the event-icons package using the CreateIcon

### Fonts
This Commit doesn't include any Font Files, as the licensing agreement does not allow them to be uploaded anywhere public (like this github).

### RandomBackground.tsx
Included in the layout.tsx file, this is a background element derived from the brand guide.
image

### Header and Footer
Basic layouts of the Header and Footer have been implemented, with the Header Links, Dropdowns and Icons working with the new Payload CMS setup


next-frontend/src/app/(wca)/navbar.tsx
```diff
@@ -9,12 +9,197 @@ import { RefreshRouteOnSave } from "@/components/RefreshRouteOnSave";
...
+
+interface IconDisplayProps {
+  name: IconName;
+}
+
+const IconDisplay: React.FC<IconDisplayProps> = ({ name }) => {
+  const IconComponent = iconMap[name];
+
+  if (!IconComponent) {
+    return <div>Icon not found</div>; // Optional fallback
```
gregorbg, line 186: Make that fallback actually optional, using an extra boolean `fallback: true` prop

next-frontend/src/app/(wca)/navbar.tsx
```diff
@@ -9,12 +9,197 @@ import { RefreshRouteOnSave } from "@/components/RefreshRouteOnSave";
...
+import VenueIcon from "@/components/icons/VenueIcon";
+import WcaDelegatesIcon from "@/components/icons/WcaDelegatesIcon";
+import WcaDocsIcon from "@/components/icons/WcaDocsIcon";
+import WcaLiveIcon from "@/components/icons/WcaLiveIcon";
+import WcaOfficersAndBoardIcon from "@/components/icons/WcaOfficersAndBoardIcon";
+import WeiboIcon from "@/components/icons/WeiboIcon";
+import XIcon from "@/components/icons/XIcon";
+import YouTubeIcon from "@/components/icons/YoutubeIcon";
+
+const iconMap: Record<string, React.ComponentType> = {
```
gregorbg, line 94: Can you create a union type for the keys? Something like `type IconName = "Bookmark" | "Clone" | ...`.

Might seem a bit verbose at first glance, but will save pulled hair down the line

next-frontend/src/app/(wca)/navbar.tsx
```diff
@@ -9,12 +9,197 @@ import { RefreshRouteOnSave } from "@/components/RefreshRouteOnSave";
...
   const payload = await getPayload({ config });
   const navbar = await payload.findGlobal({ slug: "nav" });
 
   return (
-    <HStack borderBottom="md" padding="3" justifyContent="space-between">
+    <HStack
+      borderBottom="md"
+      padding="3"
+      justifyContent="space-between"
+      bg="bg"
```
gregorbg, line 201: Curious that this is not the default...

next-frontend/src/app/(wca)/page.tsx
```diff
@@ -19,8 +112,133 @@ export default function Home() {
...
+      </HStack>
+      <Card.Root>
+        <Card.Body>
+          <Box mb="4">
+            <Text fontSize="xl" fontWeight="bold">
+              WCA Icon Gallery
+            </Text>
+          </Box>
+          <SimpleGrid columns={{ base: 2, sm: 3, md: 4, lg: 6 }} spacing={6}>
+            {[
```
gregorbg, line 149: Apologies if I'm missing something obvious, but what goes against just `Object.values(iconMap)` here?

next-frontend/src/components/Footer.tsx
```diff
@@ -0,0 +1,93 @@
...
+    <Center borderTopWidth={3} padding={3} mt={5}>
+      <VStack>
+        {/* Social Media Icons */}
+        <HStack>
+          <ChakraLink asChild variant="plainLink">
+            <Link href="https://instagram.com" target="_blank">
+              <IconButton variant="ghost">
+                <InstagramIcon />
+              </IconButton>
+            </Link>
```
gregorbg, line 32: Chakra usually renders the link inside the button, see https://www.chakra-ui.com/docs/components/button#as-link. Does that mess with the icon display of `IconButton`?

gregorbg: (If you end up switching, note that this comment also applies to all other footer buttons but I won't copy-paste the same comment five times)

next-frontend/src/components/icons/AboutTheRegulationsIcon.tsx
```diff
@@ -0,0 +1,22 @@
...
+const AboutTheRegulationsIcon = createIcon({
+  displayName: "AboutTheRegulationsIcon",
+  viewBox: "0 0 42 45",
+  path: (
+    <>
+      <path
+        d="M0,45v-4.74h28v4.74H0ZM13.18,33.51L0,20.13l4.9-5.09,13.3,13.38-5.02,5.09ZM28,18.47L14.82,4.97,19.83,0l13.18,13.38-5.02,5.09ZM38.73,42.63L8.28,11.72l3.27-3.32,30.45,30.91-3.27,3.32Z"
+        fill="currentColor"
+      />
+    </>
```
gregorbg, line 14: This outer `<></>` wrapping seems redundant, and I'm surprised that the Linter didn't punish you (or rather, ChatGPT...) for it.

I will accept it because of the sheer amount of changes that would otherwise be necessary, and because of the fact that these files don't need to be maintained (regularly), but maybe worth looking into whether the linter can detect and fix stuff like this

gregorbg: (not repeating the same comment for the 120 other icons)

next-frontend/src/components/icons/events/_222Icon.tsx
gregorbg, line 1: Did you steal these from the `@cubing/icons` repository? We just formally started a discussion about exposing SVGs there: https://github.com/cubing/icons/issues/151

next-frontend/src/globals/Nav.ts
```diff
@@ -1,5 +1,89 @@
 import { Block, GlobalConfig } from "payload";
 
+const iconOptions = [
```
gregorbg, line 3: Ah, here's _almost_ the type defintion I asked for above.
Unfortunately, you cannot include types as dropdown choice, but if you build an `iconMap` anyways you can "hack" your way around that restriction by using `Object.keys(iconMap)` here.

gregorbg (changes requested):

next-frontend/src/components/icons/AboutTheRegulationsIcon.tsx
ando527: Components have to export one Element - so icons like this one: next-frontend/src/components/icons/MenuIcon.tsx require this wrapper since they have multiple svg draw elements. ChatGPT has just added this in for all icons for consistency. 

I don't think that should be a problem, but could be fixed if required

next-frontend/src/components/icons/events/_222Icon.tsx
ando527: Yes, thats where these files originated, but for now, happy to only be using these icons as placeholders until that discussion gets resolved. 

Also note, these icons may be a little outdated - I downloaded these last year to use in my capacity as a delegate and didn't grab a fresh download before implementing these into the project

next-frontend/src/app/(wca)/navbar.tsx
ando527: Default was transparent, I wanted to align the RandomBackground to the top of the page, rather than sitting underneath the Navbar. Adding this background in was to make sure the random background didn't show through the navbar

ando527: image

next-frontend/src/app/(wca)/page.tsx
ando527: Nothing, I created this page before I did the `iconMap`. This stuff will move to a styleguide page eventually (Similar to my testing playground) - When it does I can swap to using `iconMap` :)

next-frontend/src/components/Footer.tsx
ando527: Yep, My bad, have fixed this now!
[Commit here](https://github.com/thewca/worldcubeassociation.org/pull/11227/commits/933c416c70b78aa7879071c02eb4fdfab974a4d5)
The old code also had nested focusable items, which was a bit weird for accessibility reasons. Good pickup

next-frontend/src/app/(wca)/navbar.tsx
ando527: Done! [here](https://github.com/thewca/worldcubeassociation.org/pull/11227/commits/b80ac448484f0cc874d07330288b37bbc47ea30e)

next-frontend/src/app/(wca)/navbar.tsx
ando527: Sure thing!
[here](https://github.com/thewca/worldcubeassociation.org/pull/11227/commits/b80ac448484f0cc874d07330288b37bbc47ea30e)

next-frontend/src/components/icons/AboutTheRegulationsIcon.tsx
gregorbg: Okay, consistency is a good argument. I know that our old linter (in the intermediate React code from the current monolith) would have complained about the cases where it's indeed redundant, but if we ever decide to introduce similar checks here they should be auto-fixable.

next-frontend/src/components/icons/events/_222Icon.tsx
gregorbg: Yes definitely placeholder right now. Would it be alright to copy/paste a fresh set of icons from the current state of the repo as a compromise?

ando527: Sure, Ill get onto that now

next-frontend/src/app/(wca)/page.tsx
gregorbg: What keeps you from using `iconMap` keys right now? I mean, the map already exists, you may have to export it but otherwise should be a no-brainer...?

next-frontend/src/components/icons/events/_222Icon.tsx
ando527: [done](https://github.com/thewca/worldcubeassociation.org/pull/11227/commits/699f6e4c31b4572c1c98d8f8956d2a7081a55fca)

next-frontend/src/app/(wca)/page.tsx
ando527: Implemented this using iconMap :)
[here](https://github.com/thewca/worldcubeassociation.org/pull/11227/commits/699f6e4c31b4572c1c98d8f8956d2a7081a55fca)

gregorbg (approved):

PR #11230, opened 2025-04-02
Rename Results and Inbox Results Table Columns to follow Rails Conventions

FinnIckler:
this was a lot. I'm sure this has still a good amount of test failures. 

This is the last Table from the orginal https://github.com/thewca/worldcubeassociation.org/pull/8077 PR though. So if that didn't miss anything we should be done :tada:


gregorbg (approved):

lib/check_regional_records.rb
```diff
@@ -138,12 +138,12 @@ def self.check_records(event_id = nil, competition_id = nil)
         model_comp = Competition.find(competition_id)
         event_filter = event_id || model_comp.event_ids
 
-        previous_min_results = Result.select("eventId, countryId, MIN(#{value_column}) AS `value`")
-                                     .from("#{LOOKUP_TABLE_NAME} AS Results")
+        previous_min_results = Result.select("r.event_id, r,country_id, MIN(#{value_column}) AS `value`")
```
pranav-027, line 141: @gregorbg @FinnIckler please push a hotfix  `r.country_id` 
Check regional record is not working

PR #11232, opened 2025-04-03
React migration of fix-results panel page

danieljames-dj:
New UI screenshot:

Screenshot 2025-04-03 at 09 17 55


app/webpacker/components/Panel/pages/FixResultsPage/index.jsx
```diff
@@ -0,0 +1,156 @@
...
+        ...prevFormValues,
+        resultId: undefined,
+      };
+    }
+    return prevFormValues;
+  }), [formValues]);
+
+  const handleFormChange = (_, { name, value }) => setFormValues(
+    (prev) => ({ ...prev, [name]: value }),
+  );
```
kr-matthews, line 79: This function should calculate the new form values and pass them to `setFormValues`, instead of passing in an intermediary value and then updating that via a side effect. That will be easier to understand, have fewer re-renders, and be less likely to accidentally get into an infinite loop.

kr-matthews (changes requested):

app/webpacker/components/Panel/pages/FixResultsPage/index.jsx
danieljames-dj: Okay, does that mean that I should have separate handler for each inputs? (I'm fine with it, but would like to double confirm if I understood it correctly)

danieljames-dj: Oh this is actually nice, useEffect is not needed anymore. Changed it.

gregorbg (commented):
Curious about the API design. Frontend looking fine otherwise, I'm very happy about the absence of `useEffect` :upside_down_face:

app/webpacker/components/Panel/pages/FixResultsPage/api/getCompetitions.js
```diff
@@ -0,0 +1,17 @@
...
+async function getCompetitions({ wcaId }) {
+  const { data } = await fetchJsonOrError(
+    viewUrls.persons.results(wcaId),
+  );
+  const resultsList = data || {};
+  const competitionsList = _.uniqBy(resultsList, 'competition_id')
+    .map((item) => ({
+      competitionId: item.competition_id,
+      competitionName: item.competition_name,
+    })).reverse();
```
gregorbg: Just curious: What is the purpose of `reverse` here? If it's just a matter of sorting the dropdown options "more conveniently", then you should handle that outside of the data fetching layer.

app/webpacker/components/Panel/pages/FixResultsPage/api/getEvents.js
```diff
@@ -0,0 +1,12 @@
+import { fetchJsonOrError } from '../../../../../lib/requests/fetchWithAuthenticityToken';
+import { viewUrls } from '../../../../../lib/requests/routes.js.erb';
+
+async function getEvents({ wcaId, competitionId }) {
+  const { data } = await fetchJsonOrError(
+    viewUrls.persons.results(wcaId, competitionId),
+  );
```
gregorbg: You can actually use the `getResults` API function (that you're adding in this PR as well) for this

app/webpacker/components/Panel/pages/FixResultsPage/api/getCompetitions.js
```diff
@@ -0,0 +1,17 @@
+import { fetchJsonOrError } from '../../../../../lib/requests/fetchWithAuthenticityToken';
+import { viewUrls } from '../../../../../lib/requests/routes.js.erb';
+
+async function getCompetitions({ wcaId }) {
+  const { data } = await fetchJsonOrError(
+    viewUrls.persons.results(wcaId),
+  );
```
gregorbg: Same as other comment: Reuse `getResults`

app/webpacker/components/Panel/pages/FixResultsPage/index.jsx
```diff
@@ -0,0 +1,161 @@
...
+
+  const {
+    data: resultsList,
+    isFetching: resultsFetching,
+    isError: resultsError,
+  } = useQuery({
+    queryKey: ['fix-results-rounds', wcaId, competitionId, eventId],
+    queryFn: () => getResults({ wcaId, competitionId, eventId }),
+    enabled: !!eventId,
+  });
```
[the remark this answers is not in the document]
gregorbg: Cool, now it's much more concise and streamlined. I still wonder though: Does introducing `useQuery` three times imply that each query is firing its own network request against the same endpoint?

That would be three network requests. Maybe `react-query` is smart enough to cache these calls internally (so it doesn't fire three requests against the exact same route) but otherwise one solution could be to fetch the data once and then do all the Lodash `_.uniqBy` etc. shenanigans here in the React component.

app/webpacker/components/Panel/pages/FixResultsPage/api/getResults.js
gregorbg, line 1: I am not sure whether you need to split out all three calls to three different files, seeing that they're practically doing the same thing (except two of the three calls are doing some post-processing on the data).

It would probably be more economical to have all three methods in one file. This will also make reusing the core `getResults` method easier, and maybe reveal some more opprtunities for refactoring the queries to use less `useQuery`.

gregorbg (changes requested):

app/webpacker/components/Panel/pages/FixResultsPage/api/getResults.js
danieljames-dj: Okay understood. I was bit confused what was the right approach as all the three were using same API. Anyway I have moved everything to single file. I'm replying regarding the `useQuery` in the other comment.

app/webpacker/components/Panel/pages/FixResultsPage/index.jsx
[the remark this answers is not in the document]
danieljames-dj: I'm not sure if I understood you correctly. The three network requests are not happening in parallel. Only one of them is happening at once - for example, when WCA ID is selected in the form, first API is triggered, when a competition is selected, the second API is triggered, and when an event is selected, the third API is triggered.

Also, if I merge all of them to single `useQuery`, I don't think it will work. Because let's assume I selected an event. Now if I change my mind, and if I want to change the selected competition, the competition dropdown should have the list of competitions which was generated before selecting the event. If I use single `useQuery`, it will change the competition list to only those competitions where that competitor has competed that particular event.

In case I misunderstood your query, please let me know.

app/webpacker/components/Panel/pages/FixResultsPage/api/getCompetitions.js
danieljames-dj: Merged it to a new file.

app/webpacker/components/Panel/pages/FixResultsPage/api/getEvents.js
danieljames-dj: Merged it to a new file.

app/webpacker/components/Panel/pages/FixResultsPage/api/getCompetitions.js
danieljames-dj: Yes, it's just a matter of sorting the dropdown options. I've moved it to the rendering part.

app/webpacker/components/Panel/pages/FixResultsPage/index.jsx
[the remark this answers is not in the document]
gregorbg: No it's fine, I hadn't considered the edge case that you outlined about switching events midway

PR #11233, opened 2025-04-03
Bulk auto accept registrations

dunkOnIT:
Depends on #10977

Once we have backend support for auto accept, we want to allow organizers to "bulk auto accept" registrations - this will be a precursor to "live" auto accepting registrations.
- **Bulk auto accept:** Organizer clicks a button, and all eligible competitors are moved to either Accepted or Waiting List
- **Live auto accept:** After a competitor pays, the system automatically determines what their registration status should be, and changes it without the need for organizer intervention

This PR does the following: 
- adds tests defining auto accept behaviour
- adds functions fulfilling those test requirements
- exposes an API route for bulk auto accept (and refactors slightly the permissions validations for registration admin routes)
- adds a button in the Registrations page to trigger Bulk Auto Accept


dunkOnIT:
Getting some very strange behaviour in the `registration_payment#auto_accept_hook` conditions. Consider this setup: 

```ruby
  delegate :auto_accept_in_current_env?, :auto_accept_registrations, to: :registration
  after_create :auto_accept_hook, if: :should_auto_accept?

  
  private def should_auto_accept?
    auto_accept_in_current_env? && auto_accept_registrations && Registration::LIVE_AUTO_ACCEPT_ENABLED
  end

  private def auto_accept_hook
    puts "auto accepting for reg with status: #{registration.competing_status}"
    puts should_auto_accept?
    registration.attempt_auto_accept
  end
```

Output from those puts statements is: 
```bash
auto accepting for reg with status: waiting_list
false
```

So even though it sees that `should_auto_accept?` is false in the method scope, in the scope of the after_create declaration it clearly isn't. 

For the purpose of getting this reviewed, and because disabling live auto accept is temporary, I am just adding a guard clause to the `auto_accept_hook` for `LIVE_AUTO_ACCEPT_ENABLED`

app/models/registration.rb
```diff
@@ -247,12 +250,12 @@ def registration_history
       changed_attributes = r.registration_history_changes.index_by(&:key).transform_values(&:parsed_value).symbolize_keys
 
       {
-        changed_attributes: changed_attributes,
+        changed_attributes: changed_attributes.with_indifferent_access,
```
gregorbg, line 253: Veto :P

app/models/registration.rb
```diff
@@ -247,12 +250,12 @@ def registration_history
...
         actor_type: r.actor_type,
         actor_id: r.actor_id,
         timestamp: r.created_at,
         action: r.action,
-      }
+      }.with_indifferent_access
```
gregorbg, line 258: Veto :P

app/models/registration.rb
```diff
@@ -531,6 +534,34 @@ def serializable_hash(options = nil)
...
   end
 
+  def self.bulk_auto_accept(competition)
+    if competition.waiting_list.present?
+      competition.waiting_list.entries.each do |r_id|
+        Registration.find(r_id).attempt_auto_accept
+        # It isnt clear to me why this is needed - but without it, the auto-accepted items getting removed from the waiting list reappear
+        # on the waiting list in the test body, even when calling waiting_list.reload there
+        # I have also tried assigning competition.waiting_list.entries to a variable and iterating over that, but it didnt help
+        competition.waiting_list.reload
```
gregorbg, line 544: Can you point me to a specific test case that should work but doesn't without this line? That would make it easier for me to debug

app/models/registration.rb
```diff
@@ -531,6 +534,34 @@ def serializable_hash(options = nil)
     super(DEFAULT_SERIALIZE_OPTIONS.merge(options || {}))
   end
 
+  def self.bulk_auto_accept(competition)
+    if competition.waiting_list.present?
+      competition.waiting_list.entries.each do |r_id|
```
gregorbg, line 539: I think you can save that `if` by doing something like `competition.waiting_list&.entries&.each`, although I'm not sure whether that's easier to read or not

app/models/registration.rb
```diff
@@ -531,6 +534,34 @@ def serializable_hash(options = nil)
...
+                                   .competing_status_pending
+                                   .with_payments
+                                   .sort_by { |registration| registration.last_positive_payment.updated_at }
+
+    sorted_pending_registrations.each(&:attempt_auto_accept)
+  end
+
+  def last_positive_payment
+    registration_payments
+      .where('amount_lowest_denomination > 0')
```
gregorbg, line 559:
```suggestion
      .where.not(amount_lowest_denomination: ..0)
```

app/models/registration.rb
```diff
@@ -531,6 +534,34 @@ def serializable_hash(options = nil)
...
+      .order(updated_at: :desc)
+      .limit(1)
```
gregorbg, line 561: `limit(1)` is unnecessary when doing `first`.

dunkOnIT:
Reproducing test quirk: 
- Make the changes applied in [this commit](https://github.com/thewca/worldcubeassociation.org/pull/11233/commits/624262ca8a37b903759d1eee1834fa912c13573f) 
- Run `bin/spring rspec spec/models/registration_spec.rb --fail-fast`; expected test failure is #bulk_auto_accept::when competitor limit::accepts waitlisted registrations (registration_spec.rb:877) 
- Interesting note, running the test case in isolation (`bin/spring rspec spec/models/registration_spec.rb --fail-fast --tag only`) does not result in a failure

app/models/registration.rb
dunkOnIT: Veto acknowledged! Looks like a merge fail

app/models/registration.rb
dunkOnIT: Veto acknowledged! Looks like a merge fail

app/models/registration.rb
dunkOnIT: `spec/requests/api_registrations_spec.rb:1379` (PATCH #bulk_auto_accept::triggers bulk auto accept via API route)

app/models/registration.rb
dunkOnIT: Changed

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationList.jsx
```diff
@@ -89,6 +90,20 @@ export default function RegistrationAdministrationList({ competitionInfo }) {
...
+    onError: () => {
+      dispatchStore(showMessage(
+        'competitions.registration_v2.auto_accept.cant_bulk_auto_accept',
+        'negative',
+      ));
+    },
+    onSuccess: async () => {
+      dispatchStore(showMessage('competitions.registration_v2.auto_accept.bulk_auto_accepted', 'positive'));
+      await refetch();
+    },
```
gregorbg, line 104: Why `async`/`await` here? I know that `refetch` itself is an async method, but we don't get any benefit from waiting for the result (afaik)

dunkOnIT: This might be my JS inexperience - in my mind if refetch() is async then we have to use async/await? 

My desired outcome is the following: 
- refetch the registrations after 
- have the various registration list components be `loading` until we're finished fetching the registrations - so `await`ing the results does naively seem like something I want to do?

dunkOnIT:
Also, I thought about adding a validation which asserts that the competition must be using integrated payments to enable auto accept - but I realized that it's silly because most people create comps (and would enable a setting like auto accept) before connecting payments. 

We could have it be something that users have to first enable payments for, and then come back to the competition form - but this seems like annoying UX, compared to perhaps just allowing the user to uselessly enable it (with a hint that it will only work with integrated payments) (and perhaps a message like "You have auto-accept enabled, but no payment integration selected")

gregorbg:
> Also, I thought about adding a validation which asserts that the competition must be using integrated payments to enable auto accept - but I realized that it's silly because most people create comps (and would enable a setting like auto accept) before connecting payments.
> 
> We could have it be something that users have to first enable payments for, and then come back to the competition form - but this seems like annoying UX, compared to perhaps just allowing the user to uselessly enable it (with a hint that it will only work with integrated payments) (and perhaps a message like "You have auto-accept enabled, but no payment integration selected")

We deal with this kind of situation by using the "Is the competition confirmed and submitted to WCAT" flag (in Ruby: `my_comp.confirmed?` and `my_comp.confirmed_or_visible?`) to enforce certain validations only upon confirmation. Basically it's a matter of using the `validate :auto_accept_foo, if: :confirmed_or_visible?` flag to "weaken" your validation to only fire upon confirmation.

That way, Delegates can make chill edits to their confirmation but still be reminded about "correctness" once they're actually ready to submit. The UX infrastructure to display these "confirmation errors" in the competition form is already in place.

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationList.jsx
dunkOnIT: Aha ok that makes sense, thanks for the explanation! 

I have tried both, and I'm keeping async/await because without them the UI re-enables _slightly_ before the `refetch` has reflected in the registration tables - ie, they enable and then the registrations move to their new list. With async/await, when the tables re-enable (ie exit the loading state) the registrations are already in their correct/new list.

gregorbg: That... Shouldn't happen :sweat_smile: The "outermost" call to your mutation isn't blocking, so forcing the inner `onSuccess` hook to block and wait should have zero influence on the outside world code flow. Let me try and investigate.

gregorbg: Aha, very interesting! Turns out this is tied to the internal React rendering cycle: https://tkdodo.eu/blog/mastering-mutations-in-react-query#common-gotchas and the section on `onSuccess` at https://tanstack.com/query/latest/docs/framework/react/reference/useMutation are an interesting read.

gregorbg: Probably the best compromise here is to _return_ the `Promise` generated by calling `refetch`. That way, you don't have to care about `await`ing the result yourself, and thus you don't have to mark your method as `async`. The mutation's inner workings will take care of "cleanly" resolving the value for you: https://tkdodo.eu/blog/mastering-mutations-in-react-query#awaited-promises

dunkOnIT: Ah yep adding a `return` makes it work as expected!

app/controllers/competitions_controller.rb
```diff
@@ -610,7 +610,7 @@ def user_preferences
     render json: competition.form_user_preferences(current_user)
   end
 
-  before_action -> { require_user_permission(:can_manage_competition?, competition_from_params) }, only: [:announcement_data]
+  before_action -> { require_user_permission(:can_manage_competition?, competition_from_params) }, only: [:confirmation_data]
```
dunkOnIT, line 613: Unrelated change - I noticed this when I was playing around with headers. Happy to remove and put in a separate PR if necessary, but it seems like a clear copy/paste error to me - looking at the full context of the file should make this obvious (`announcement_data` already has an identical before_action elsewhere, but `confirmation_data` doens't)

app/models/competition.rb
```diff
@@ -2847,7 +2858,8 @@ def self.form_data_json_schema
           "type" => "object",
           "properties" => {
             "isConfirmed" => { "type" => "boolean" },
-            "isVisible" => { "type" => "boolean" },
+            "autoAcceptEnabled" => { "type" => %w[boolean null] },
+            "autoAcceptDisableThreshold" => { "type" => %w[integer null] }, "isVisible" => { "type" => "boolean" }
```
gregorbg, line 2862: I think you're missing a line break here. On that note: Why did you add the autoAccept fields "sandwiched" between `isConfirmed` and `isVisible`, instead of just appending them behind `isVisible`?

gregorbg: Oh, this is actually a duplicate now that you're exposing this under `competitorLimit` as well

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationList.jsx
```diff
@@ -395,7 +410,21 @@ export default function RegistrationAdministrationList({ competitionInfo }) {
...
   return (
-    <Segment loading={isMutating}>
+    <Segment loading={isMutating || isAutoAccepting}>
+      {competitionInfo.auto_accept_registrations && (
+        <Button
+          disabled={isAutoAccepting}
+          color="green"
+          onClick={() => {
+            bulkAutoAcceptMutation(competitionInfo.id);
+          }}
```
gregorbg, line 420:
```suggestion
          onClick={() => bulkAutoAcceptMutation(competitionInfo.id)}
```

gregorbg (changes requested):

app/models/competition.rb
dunkOnIT: Resolved

app/models/competition.rb
```diff
@@ -2847,7 +2858,6 @@ def self.form_data_json_schema
           "type" => "object",
           "properties" => {
             "isConfirmed" => { "type" => "boolean" },
-            "isVisible" => { "type" => "boolean" },
```
gregorbg, line 2850: This line deletion seems unrelated, although I'm aware that we're probably not really using it anywhere

gregorbg (changes requested):

config/routes.rb
```diff
@@ -353,6 +353,7 @@
           end
 
           collection do
+            patch '/bulk_auto_accept', to: 'registrations#bulk_auto_accept', as: :bulk_auto_accept
```
gregorbg, line 356: Nit: You don't need the leading slash. Rails will use slashes as "glue" for nested routes by default (note that no other route in the `collection` specifies leading slashes)

gregorbg: Also, I don' think you need to specify the `as:` alias. (Because the route param and the controller action are both already called exactly `bulk_auto_accept` anyways). If in doubt, use `bin/rails routes` to let Rails show you a full list including all helper names

gregorbg (changes requested):

PR #11236, opened 2025-04-03
Add a customisable dropdown menu to the registration page

simonkellly:
This adds an option in the competition form to include a dropdown menu with a title and options, which may/may not be mandatory to fill out for competitors.

The key use case here is for things such as allocating t-shirt sizes or other per competitor options. 

At the moment this is usually done by asking for a mandatory comment. This request is frequently ignored as people rush to register and don't know they have to write something (typing random letters/hi into the box to bypass the erros).

This would hopefully allow for more flexibility in these requests while making the UX better for registrants, and life a lot simpler for those managing these competitions (especially once automatic registration becomes a thing).

From what I can see almost all of the changes that need changes in order to allow this have been made, but there is a chance I have missed some, especially as I haven't worked with the new reg system very much.

**Screenshots**
image
image
image
image


OJCubing:
As well as T-Shirts, this would be so useful for multiple venue FMC comps! Would save so much admin overhead.

danieljames-dj (changes requested):
I'm bit concerned about the database design. The main restriction here is that it allows only a single dropdown per competition. What if a competition wants multiple dropdown are required?

Similarly, I am also having concern regarding glueing this with registration. My suggestion on the DB would be to have following tables:

1. `competition_registration_dropdowns`: Main table with columns `competition_id` and`dropdown_name`
2. `competition_registration_dropdown_options`: To store the options
3. `registration_dropdown_responses`: To store the registration dropdown responses

This is just my suggestion, it'll be great if we can have a discussion on the design first within WST.

Thanks a lot for your efforts, this is a great idea overall.

PR #11237, opened 2025-04-04
Remove old rankings and records tables

FinnIckler:
I think we didn't remove them because of the tests? But we migrated these months ago


FinnIckler:
actually it's just i18n keys

danieljames-dj (approved):

danieljames-dj:
I manually verified in local if any of the removed files are used, and indeed none are used. So code LGTM. As Finn pointed out, there are some i18n keys to be removed as well.

PR #11274, opened 2025-04-07
Rename Attempt to RoundResultsAttempt

FinnIckler:
Now that we want to move the `valueX` Results columns to it's own Attempts Table we need to rename this Class.

The long term solution is to replace this with the LiveAttempt Table anyway

### DIFF
--- a/lib/round_results.rb
+++ b/lib/round_results.rb
@@ -54,7 +54,7 @@ class RoundResult
     self.new(
       person_id: json_obj['personId'],
       ranking: json_obj['ranking'],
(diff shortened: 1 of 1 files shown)


danieljames-dj (approved):

gregorbg:
Are you sure beyond any doubt that this is not referenced in any polymorphic `*_type` column in the DB?

FinnIckler:
The whole thing gets chucked into `serialize :round_results, coder: RoundResults` so no type columns here

PR #11292, opened 2025-04-08
Get rid of light result and merge result_methods into resultable

FinnIckler:


app/models/concerns/resultable.rb
```diff
@@ -173,4 +172,75 @@ def to_s(field)
...
+
+  def counting_solve_times
+    unless @counting
+      @counting = []
+      solve_times.each_with_index do |solve_time, i|
+        @counting << solve_time if trimmed_indices.exclude?(i) && i < format.expected_solve_count
+      end
+    end
+    @counting
+  end
```
gregorbg, line 237: This looks like a good use-case of `filter_map`. Something along the lines of `@counting ||= solve_times.each_with_index.filter_map { |solve_time, i| solve_time if ... }`

app/models/concerns/resultable.rb
```diff
@@ -173,4 +172,75 @@ def to_s(field)
   def hlp
     ActionController::Base.helpers
   end
+
+  def wca_id
+    @personId || self.personId
```
gregorbg, line 177: Now that `light_result` is gone, I think you don't need `@personId` anymore. You need to confirm this with a little bit of CTRL+F, but if I'm right then this should boil down to be a simple `alias_attribute`.

gregorbg (changes requested):

spec/models/light_result_spec.rb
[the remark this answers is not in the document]
FinnIckler: I thought they were just duplicated tests, but maybe they are not. Should be easily moved to result spec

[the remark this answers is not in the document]
FinnIckler: I moved the tests. some of them were actually invalid now (Light Results had no validation, so there was a 333 round that was best of 3), but I fixed those too

app/models/concerns/resultable.rb
```diff
@@ -173,4 +172,69 @@ def to_s(field)
...
+    else
+      sorted_solves = sorted_solves_with_index
+      trimmed_solves_with_index = sorted_solves[0...format.trim_fastest_n]
+      trimmed_solves_with_index += sorted_solves[(sorted_solves.length - format.trim_slowest_n)...sorted_solves.length]
+      trimmed_solves_with_index.map { |_, i| i }
+    end
+  end
+
+  def counting_solve_times
+    @counting ||= solve_times.each_with_index.with_object([]) do |(solve_time, i), arr|
```
gregorbg, line 228: Why `each_with_object` instead of `filter_map`?

gregorbg (changes requested):

app/models/concerns/resultable.rb
gregorbg: cf. https://github.com/thewca/worldcubeassociation.org/pull/11295... I would prefer the previous loop over `each_with_object`, but then again I am still convinced that `filter_map` should do

gregorbg (approved):

PR #11293, opened 2025-04-08
Add new Attempts table

FinnIckler:
Adds the new Attempts Table for Results. We can do `has_many :attempts` without breaking anything because we overwrite it 
```
def attempts
    [value1, value2, value3, value4, value5]
  end
```
here anyway.

The last two rake tasks are more for testing with smaller results sets, especially during development.


app/models/result.rb
```diff
@@ -19,6 +19,19 @@ class Result < ApplicationRecord
...
+
+  after_commit :create_or_update_attempts
+
+  private def create_or_update_attempts
+    (1..5).each do |n|
+      value = public_send(:"value#{n}")
+      break if value.zero?
+
+      Attempt.upsert({ value: value, attempt_number: n, result_id: id })
+    end
```
gregorbg, line 32: Loop first, then do one `upsert_all`

gregorbg: https://apidock.com/rails/v7.1.3.4/ActiveRecord/Persistence/ClassMethods/upsert_all

app/models/attempt.rb
```diff
@@ -0,0 +1,18 @@
...
+  belongs_to :result
+
+  validates :value, presence: true
+  validates :value, numericality: { only_integer: true }
+  validates :attempt_number, numericality: { only_integer: true }
+  validates :result_id, uniqueness: { scope: [:attempt_number] }
+
+  def <=>(other)
+    value <=> other.value
+  end
```
gregorbg, line 17: Just curious, does `delegate :<=>, to: :value` work?

FinnIckler: sadly not. 
`comparison of Attempt with Attempt failed (ArgumentError)
`

gregorbg:
Totally open-ended discussion: What about naming this `ResultAttempt` instead?

I mean, we all know what "Attempt" probably means but at the end of the day that's a pretty generic English noun. I don't have strong arguments for either version, just wanted to put this here and see what others think.

gregorbg (approved):

gregorbg (approved):

PR #9737, opened 2024-08-02
Split Payment Information from Extra Registration Requirements

FinnIckler:
This allows for showing the user how to pay externally as part of the Registration Step Panel. Setting to draft because I am still waiting on WCAT on some implementation Details

TODO:
- Make sure the Registration Create Form works as expected 
- Finalize Design of External Payment Step
- Save the external payment 'approval' step in Localstorage or maybe even in the registration service?

PR #11299, opened 2025-04-09
Add a Manual External Payment Integration

FinnIckler:
Replacement for #9737 

This is still WIP because I didn't add any DB changes yet. I would like to have some feedback on this before.

I could not use the already existing `connect_payment_integration` method because that requires a redirect, which doesn't work in react world. I also think it doesn't really make sense, because there are no accounts to save.

I added a new page to input the payment information text and needed payment reference instead.

### DIFF
--- a/app/controllers/competitions_controller.rb
+++ b/app/controllers/competitions_controller.rb
@@ -259,6 +259,21 @@ class CompetitionsController < ApplicationController
     render :edit
   end
 
+  def payment_integration_manual_setup
+    @competition = competition_from_params
+  end
+
+  def add_manual_payment_setup
+    @competition = competition_from_params
+    payment_info = params[:payment_info]
+    payment_reference = params[:payment_reference]
+
+    ManualPaymentIntegration.create(payment_info: payment_info, payment_reference: payment_reference)
+
+    render json: { status: "ok" }
+  end
+
+
   def payment_integration_setup
     @competition = competition_from_params
 
--- a/app/controllers/competitions_controller.rb
+++ b/app/controllers/competitions_controller.rb
@@ -273,7 +273,6 @@ class CompetitionsController < ApplicationController
     render json: { status: "ok" }
   end
 
-
   def payment_integration_setup
     @competition = competition_from_params
 
--- a/app/models/manual_payment_integration.rb
+++ b/app/models/manual_payment_integration.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+class ManualPaymentIntegration < ApplicationRecord
+  def find_payment(record_id)
+    ManualPaymentRecord.find(record_id)
+  end
+
+  def self.generate_onboarding_link(competition_id)
+    Rails.application.routes.url_helpers.competition_manual_payment_setup_url(competition_id)
+  end
+end
--- a/app/models/manual_payment_record.rb
+++ b/app/models/manual_payment_record.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+class ManualPaymentRecord < ApplicationRecord
+  has_one :registration
+
+  validates :registration_id, presence: true
+  validates :payment_reference, presence: true
+end
--- a/app/webpacker/components/RegistrationsV2/Register/ManualPaymentStep.jsx
+++ b/app/webpacker/components/RegistrationsV2/Register/ManualPaymentStep.jsx
@@ -1,0 +1,55 @@
+import React, { useCallback } from 'react';
+import {
+  Form,
+  Message,
+  Segment,
+} from 'semantic-ui-react';
+import { useMutation } from '@tanstack/react-query';
+import I18n from '../../../lib/i18n';
+import { hasPassed } from '../../../lib/utils/dates';
+import useInputState from '../../../lib/hooks/useInputState';
+import { fetchJsonOrError } from '../../../lib/requests/fetchWithAuthenticityToken';
+import { addPaymentReferenceUrl } from '../../../lib/requests/routes.js.erb';
+
+export default function ManualPaymentStep({
+  userInfo,
+  competitionInfo,
+  paymentInfo,
+  nextStep,
+}) {
+  const [paymentReference, setPaymentReference] = useInputState('');
+
+  const { mutate } = useMutation({
+    mutationFn: ({ reference }) => fetchJsonOrError(
+      addPaymentReferenceUrl(competitionInfo.id, userInfo.id),
+      {
+        method: 'POST',
+        body: JSON.stringify({ paymentReference: reference }),
+      },
+    ),
+    onSuccess: () => nextStep(),
+  });
+
+  const handleSubmit = useCallback(async (e) => {
+    e.preventDefault();
+    mutate({ reference: paymentReference });
+  }, [mutate, paymentReference]);
+
+  if (hasPassed(competitionInfo.registration_close)) {
+    return (
+      <Message color="red">{I18n.t('registrations.payment_form.errors.registration_closed')}</Message>
+    );
+  }
+
+  return (
+    <Segment>
+      {paymentInfo.payment_info}
+      <Form id="manual-payment-form" onSubmit={handleSubmit}>
+        <Form.Field>
+          <label htmlFor="paymentReference">{paymentInfo.payment_reference}</label>
+          <Form.Input id="paymentReference" value={paymentReference} onChange={setPaymentReference} />
+        </Form.Field>
+      </Form>
+    </Segment>
+  );
+}
--- a/app/controllers/api/v1/registrations/registrations_controller.rb
+++ b/app/controllers/api/v1/registrations/registrations_controller.rb
@@ -9,8 +9,8 @@ class Api::V1::Registrations::RegistrationsController < Api::V1::ApiController
   before_action :validate_create_request, only: [:create]
   before_action :validate_show_registration, only: [:show]
   before_action :validate_list_admin, only: [:list_admin]
-  before_action :ensure_registration_exists, only: [:update]
-  before_action :user_can_modify_registration, only: [:update]
+  before_action :ensure_registration_exists, only: [:update, :add_payment_reference]
+  before_action :user_can_modify_registration, only: [:update, :add_payment_reference]
   before_action :validate_update_request, only: [:update]
   before_action :user_can_bulk_modify_registrations, only: [:bulk_update]
   before_action :validate_bulk_update_request, only: [:bulk_update]
@@ -84,6 +84,12 @@ class Api::V1::Registrations::RegistrationsController < Api::V1::ApiController
       return render json: { status: 'ok', registration: updated_registration.to_v2_json(admin: true, history: true) }, status: :ok
     end
     render json: { status: 'bad request', message: 'You need to supply at least one lane' }, status: :bad_request
+  end
+
+  def add_payment_reference
+    ManualPaymentRecord.create(registration: @registration, payment_reference: params.require[:payment_reference])
+
+    render json: { status: 'ok' }, status: :ok
   end
 
   def ensure_registration_exists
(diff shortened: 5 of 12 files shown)


FinnIckler:
image
image

There are still a ton of small changes to be made, like adding to the CSV export and changing the single registration edit page

FinnIckler:
Payment step during the registration:
image

Payment step after payment:
image

gregorbg:
(for the record: I had a lengthy discussion with Finn during our regular call, there are gonna be major reworks on this PR)

app/models/manual_payment_integration.rb
```diff
@@ -0,0 +1,62 @@
...
+      payment_reference: payment_reference,
+    }
+  end
+
+  def self.connect_integration(params)
+    ManualPaymentIntegration.create(payment_information: params[:payment_information], payment_reference: params[:payment_reference])
+  end
+
+  def dashboard
+  end
```
gregorbg, line 61: What does this empty method do, and why is it needed?

app/models/manual_payment_integration.rb
```diff
@@ -0,0 +1,62 @@
...
+  def self.generate_onboarding_link(competition_id)
+    Rails.application.routes.url_helpers.competition_manual_payment_setup_url(competition_id)
+  end
+
+  def account_details
+    {
+      payment_info: payment_information,
+      payment_reference: payment_reference,
+    }
+  end
```
gregorbg, line 54: That should be replaceable by simple serialization logic

app/models/competition.rb
```diff
@@ -1812,14 +1816,26 @@ def to_competition_info
...
+  def payment_integration_type
+    return nil unless using_payment_integrations?
+
+    CompetitionPaymentIntegration::AVAILABLE_INTEGRATIONS.key(competition_payment_integrations.first.connected_account_type)
+  end
+
+  def manual_payment_details
+    return nil unless using_manual_payment?
+
+    competition_payment_integrations.first.connected_account.account_details
```
gregorbg, line 1836:
```suggestion
    competition_payment_integrations.first.connected_account.as_json
```

gregorbg: (in line with my other comment about serialization)

app/models/registration.rb
```diff
@@ -255,6 +255,10 @@ def registration_history
     end
   end
 
+  def payment_reference
+    registration_payments.first&.receipt&.payment_reference
```
gregorbg, line 259: Why is only the first payment relevant here?

app/webpacker/components/RegistrationsV2/Register/ManualPaymentStep.jsx
```diff
@@ -0,0 +1,75 @@
...
+export default function ManualPaymentStep({
+  competitionInfo,
+}) {
+  const { hasPaid, registration } = useRegistration();
+
+  const [paymentReference, setPaymentReference] = useInputState(hasPaid ? registration.payment.payment_reference : '');
+  const [paymentConfirmation, setPaymentConfirmation] = useCheckboxState(hasPaid);
+
+  const { data, isLoading } = useQuery({
+    queryKey: ['manual-payment'],
```
gregorbg, line 28: Are you sure this static query key is sufficient?

app/models/manual_payment_integration.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  def self.connect_integration(params)
+    ManualPaymentIntegration.create(payment_information: params[:payment_information], payment_reference: params[:payment_reference])
```
gregorbg, line 57:
```suggestion
    ManualPaymentIntegration.create(**params)
```

gregorbg: (You may need `params.permit!`, Rails is often very strict with accessing `params` directly)

gregorbg (changes requested):
No strong feelings about the way you simulate the remote update, I think that's already in a good spot. But a few other remarks here and there.

gregorbg (changes requested):

pranav-027:
This looks like a great feature, Thank you WST for building this. I have one question: Can a user attach images to the payment information box? If not, is it possible to have that feature? It will be helpful for the UPI payment type, which has high usage in India

gregorbg:
> Can a user attach images to the payment information box? If not, is it possible to have that feature?

Technically, yes. Practically, meh. It is always complicated and cumbersome to host our own image uploads, with filetypes and screen sizes and preview thumbnails and whatnot. The effort it takes to implement may outweigh the benefits.

> It will be helpful for the UPI payment type, which has high usage in India

Can you please elaborate on this? What good will an image upload do, that a simple "copy/paste of your completed payment receipt number" cannot do?

pranav-027:
> Can you please elaborate on this? What good will an image upload do, that a simple "copy/paste of your completed payment receipt number" cannot do?

In India, we have UPI payments via QR code, which involves a customer using a UPI-enabled app to scan a QR code displayed by a merchant, enter the payment amount, and authorize the transaction using their UPI PIN. It would be great if organizers could upload an image of their payment QR code in the information box from their side. Image upload doesn’t need to be enabled for users who will be making the payments.

gregorbg:
Thanks for the clarification! In that case I may have misunderstood. An image upload for the individual competitors who pay is indeed too complicated to realistically implement.

But an image upload for organizers is relatively easy and may in fact already work now! Markdown editors usually support image upload, if not it's easy for us to switch them on.

FinnIckler:
yes, this just supports image uploads as part of the markdown

dunkOnIT:
@thewca-bot deploy staging

pranav-027:
I think this warning: 
image
needs to be made generic after this feature.

Also, is there any discussion with WCAT to mandate payment integration if a competition has registration fees?

FinnIckler:
The original PR had that, but it required a lot of manual work by WCAT. The goal here is that we can make organizer onboard gradually and then make it compulsory at a later date

gregorbg:
If we really wanted to make this mandatory, then we also need to consider the "pay in person at the day of the competition at the venue" option. I guess they can easily fit their description, including which payment methods will be accepted at the venue, into the text box of this PR, but they won't need to collect any payment reference beforehand.

So maybe a "I don't want to collect any payment reference" optional checkbox would be prudent?

dunkOnIT:
@thewca-bot deploy staging

dunkOnIT:
@thewca-bot deploy staging

dunkOnIT:
@thewca-bot deploy staging

dunkOnIT:
@thewca-bot depoly staging

FinnIckler:
you had a typo there :P

dunkOnIT:
@thewca-bot deploy staging

FinnIckler:
I just noticed that we need to implement showing the payment reference in the admin table again

app/controllers/competitions_controller.rb
```diff
@@ -272,21 +276,34 @@ def payment_integration_setup
...
 
-    if payment_integration == 'paypal' && PaypalInterface.paypal_disabled?
+    if payment_integration == :paypal && PaypalInterface.paypal_disabled?
       flash[:error] = 'PayPal is not yet available in production environments'
       return redirect_to competitions_payment_setup_path(competition)
     end
 
-    connector = CompetitionPaymentIntegration::AVAILABLE_INTEGRATIONS[payment_integration.to_sym].safe_constantize
-    account_reference = connector&.connect_account(params)
+    if CompetitionPaymentIntegration::AVAILABLE_INTEGRATIONS[payment_integration].nil?
```
dunkOnIT, line 288: We need to check for the validity of the integration here, before we try and assign it to a variable. Old-line 284 actually fails if the integration is unrecognized because `safe_constantize` can't be called on `nil`

app/controllers/competitions_controller.rb
```diff
@@ -272,21 +276,34 @@ def payment_integration_setup
...
+      render plain: "Payment Integration #{payment_integration} not Found", status: :not_found and return
+    end
 
-    raise ActionController::RoutingError.new("Payment Integration #{payment_integration} not Found") if account_reference.blank?
+    connector = CompetitionPaymentIntegration::AVAILABLE_INTEGRATIONS[payment_integration].safe_constantize
+    integration_reference = connector&.connect_integration(params)
 
-    competition.competition_payment_integrations.new(connected_account: account_reference)
+    raise ActionController::RoutingError.new("No integration reference submitted") if integration_reference.blank?
```
dunkOnIT, line 295: Wording on the old message was incorrect - it wasn't checking for the payment integration type, it was checking for the presence of the account reference

dunkOnIT:
Two related notes on the implementation here: 
- **Should we be storing `amount_iso` and `currency_code` on the payment record?** I'm not a big fan of this given that we're only inferring the information from the competition's settings, and so are very likely to end up in the situation where we're making confident assertions about payments amounts that are incorrect (already, I can give the example of CubingZA, who charge different amounts through their external payment service based on the number of days the competitor is attending the competition
  - At the very least, I'd say we shouldn't show this info to the user if there's some technological reason why it has to be stored (eg, we expect all payment_records to have those fields). But if we aren't showing it to the user, I'd again make the case that we should try not to store it at all
- **Do we need to/should we be creating a PaymentIntent for manual payments?** It seems like all we need to be doing is checking/asserting the presence of a PaymentRecord, so the PaymentIntent seems redundant. At the same time, I recognize that we don't want to deviate from the current workflow too much, so this might well be a very shortsighted thought.

app/models/manual_payment_record.rb
```diff
@@ -0,0 +1,35 @@
...
+  WCA_TO_MANUAL_PAYMENT_STATUS_MAP = {
+    created: %w[created],
+    succeeded: %w[succeeded],
+  }.freeze
+
+  has_one :registration_payment, as: :receipt
+  has_one :payment_intent, as: :payment_record
+
+  def determine_wca_status
+    WCA_TO_MANUAL_PAYMENT_STATUS_MAP[status.to_sym]
```
dunkOnIT, line 13: This used to have the same contents as `status` - I've changed it to pull the wca_status from the defined mapping

app/models/payment_intent.rb
```diff
@@ -100,12 +101,16 @@ def update_intent_details(api_record)
...
+      when 'StripeRecord'
         errors.add(:wca_status, "#{wca_status} is not compatible with StripeRecord status: #{payment_record.stripe_status}") unless
           StripeRecord::WCA_TO_STRIPE_STATUS_MAP[wca_status.to_sym].include?(payment_record.stripe_status)
-      elsif payment_record_type == 'PaypalRecord'
+      when 'PaypalRecord'
         errors.add(:wca_status, "#{wca_status} is not compatible with PaypalRecord status: #{payment_record.paypal_status}") unless
           PaypalRecord::WCA_TO_PAYPAL_STATUS_MAP[wca_status.to_sym].include?(payment_record.paypal_status)
+      when 'ManualPaymentRecord'
+        errors.add(:wca_status, "#{wca_status} is not compatible with ManualPaymentRecord status: #{payment_record.status}") unless
+          ManualPaymentRecord::WCA_TO_MANUAL_PAYMENT_STATUS_MAP[wca_status.to_sym].include?(payment_record.status)
```
dunkOnIT, line 113: Previously this was calling `payment_record.paypal_status` - updated to just call `status` of the payment_record itself

app/models/registration.rb
```diff
@@ -249,6 +249,11 @@ def registration_history
     end
   end
 
+  def payment_reference
+    # TODO: We currently have no
```
dunkOnIT, line 253: Incomplete comment?

spec/requests/api_registrations_spec.rb
```diff
@@ -1275,54 +1275,69 @@
...
   describe 'GET #payment_ticket' do
-    let(:competition) { create(:competition, :registration_open, :with_organizer, :stripe_connected) }
-    let(:reg) { create(:registration, :pending, competition: competition) }
     let(:headers) { { 'Authorization' => fetch_jwt_token(reg.user_id) } }
 
-    it 'successfully builds a payment_intent via Stripe API' do
-      get api_v1_registrations_payment_ticket_path(competition_id: competition.id), headers: headers
-      expect(response).to be_successful
-    end
+    context 'stripe connect payment integration' do
```
dunkOnIT, line 1280: I moved everything that was here into this context block, and added a new context block for manual payment integrations - this makes the diff appear larger/more significant than it is

PR #11302, opened 2025-04-10
OTS Page Access Restrictions

### LINKED ISSUE #11238: Prevent OTS form from adding any new registrations if competition does not allow OTS or the reg period is over
Organizers/Delegates can add new registrations through the OTS form even when the registration period is over or that the competition doesn't allow OTS. If the OTS is not allowed, then the OTS form shouldn't be accessible in the first place.

dunkOnIT:
Partially resolves #11238

Currently, organizers/delegates can access the OTS page (`/competition/registrations/add`) and add registrations
- at any time - even after comp has ended and reg has closed
- whether or not OTS is enabled

These permissions were too loose. We have [confirmed with WCAT](https://worldcubeassociation.slack.com/archives/C06N7NM6MJ6/p1744394470092349?thread_ts=1744272016.181829&cid=C06N7NM6MJ6) in Slack that the rules for using the OTS form should be: 
> reg open | ots enabled -> ots form should work

This PR enforces that in the frontend and backend.

In the medium-term, we plan to move the OTS form to use the v1 API - and these validations will need to be moved there.

**Note:** This effectively changes the do_add route to be a dedicated add_on_the_spot route - which it already is, as all other registrations go via the v1 API, and this route should soon be deprecated entirely in favour of the v1 API.


app/controllers/registrations_controller.rb
```diff
@@ -147,11 +147,21 @@ def add
...
     @competition = competition_from_params
+
     if @competition.registration_full?
       flash[:danger] = I18n.t("registrations.mailer.deleted.causes.registrations_full")
       redirect_to competition_path(@competition)
       return
+    elsif !@competition.registration_currently_open? && !@competition.on_the_spot_registration?
+      flash[:danger] = I18n.t("registrations.add.ots_not_enabled")
+      redirect_to competition_path(@competition)
+      return
```
gregorbg, line 158:
```suggestion
      return redirect_to competition_path(@competition)
```

gregorbg: (yes, I admit the existing code above is not exactly a role model)

app/views/registrations/add.html.erb
```diff
@@ -3,6 +3,10 @@
 <%= render layout: "nav" do %>
   <% if @competition.registration_full? %>
     <%= alert :warning, t('registrations.mailer.deleted.causes.registrations_full'), note: true %>
+  <% elsif !@competition.registration_currently_open? && !@competition.on_the_spot_registration? %>
+    <%= alert :warning, t('registrations.add.ots_not_enabled') %>
+  <% elsif @competition.probably_over? %>
+    <%= alert :warning, t('registrations.add.competition_over') %>
```
gregorbg, line 9: About the two `alert`s: The other existing alert above has `note: true`, what does that do and is it useful here as well?

spec/requests/registrations_spec.rb
```diff
@@ -441,22 +441,24 @@
     end
 
     context "when signed in as competition manager" do
+      let(:do_add_competition) { FactoryBot.create(:competition, :registration_open, :with_delegate, :visible) }
```
gregorbg, line 444: I find that name confusing, is it something I'm not getting as a non-native?

gregorbg: `do_add` has nothing to do with competitions or OTS in my mind

gregorbg (changes requested):

spec/requests/registrations_spec.rb
dunkOnIT: `do_add` is the name of the OTS route in the controller - but I agree it's confusing. Renamed to `ots_competition` - basically just need a variable name that differentiates it from the default competition variable

app/views/registrations/add.html.erb
gregorbg: Thans for finding out what it does. What about the other part of my question:

> and is it useful here as well?

I'm genuinely curious to hear what you think, because I would prefer to stay consistent with the alerts that we already have

dunkOnIT: I don't think it is particularly useful - and if we want consistency, I'd suggest removing it from the existing warning here. "Note" is used to indicate some kind of additional information that might be of interest[1]. The fact that we are not rendering the OTS panel and displaying this message instead makes this alert the **main message** - so prefixing it with "Note" doesn't make sense in my view.

[1] "Note: You can still register, but you might end up on the waiting list"

config/locales/en.yml
```diff
@@ -1327,6 +1327,8 @@ en:
       title: "Add registration for %{comp}"
       info: "This will create an approved registration. Use this form to add walk-ins during the competition."
       add: "Add registration"
+      ots_not_enabled: "Registration is closed and on-the-spot registration is not enabled - a Delegate must enable it from the Organizer View"
```
gregorbg, line 1330: The message should not be restrictive about where or how you make the edits. For what it's worth, you _could_ email WCAT, get lucky that they read your mail within 10 minutes, and get it fixed that way. And that would (theoretically) work.

```suggestion
      ots_not_enabled: "Registration is closed and on-the-spot registration is not enabled - you must edit the competition to enable it first"
```

app/controllers/registrations_controller.rb
```diff
@@ -147,11 +147,20 @@ def add
...
       flash[:danger] = I18n.t("registrations.mailer.deleted.causes.registrations_full")
       redirect_to competition_path(@competition)
       return
+    elsif !@competition.registration_currently_open? && !@competition.on_the_spot_registration?
+      flash[:danger] = I18n.t("registrations.add.ots_not_enabled")
+      return redirect_to competition_path(@competition)
+    elsif @competition.probably_over?
+      flash[:danger] = I18n.t("registrations.add.competition_over")
+      redirect_to competition_path(@competition)
+      return
```
gregorbg, line 161:
```suggestion
      return redirect_to competition_path(@competition)
```

app/controllers/registrations_controller.rb
```diff
@@ -147,11 +147,20 @@ def add
 
   def do_add
     @competition = competition_from_params
+
     if @competition.registration_full?
       flash[:danger] = I18n.t("registrations.mailer.deleted.causes.registrations_full")
       redirect_to competition_path(@competition)
       return
```
gregorbg, line 154: (sidenote, as it's not strictly part of the PR but yay consistency!)
```suggestion
      return redirect_to competition_path(@competition)
```

app/views/registrations/add.html.erb
gregorbg: Agreed, feel free to remove it from the other messages for consistency.

gregorbg (changes requested):

app/views/registrations/add.html.erb
dunkOnIT: Removed

gregorbg (approved):

PR #11303, opened 2025-04-11
Rubocop-rspec enable

dunkOnIT:


spec/controllers/admin_controller_spec.rb
```diff
@@ -4,10 +4,11 @@
 
 RSpec.describe AdminController, type: :controller do
   describe 'merge_people' do
-    sign_in { FactoryBot.create :admin }
+    let(:admin) { create(:admin) }
+    before { sign_in admin }
```
gregorbg, line 8: At the risk of overlooking something from a conversation on the other (original) PR: Why is `sign_in { create(:admin) }` as a simple replacement not working?

gregorbg (commented):
Very curious about the `sign_in` changes, otherwise this seems to be a cut-and-dry Rubocop autofix PR

spec/controllers/admin_controller_spec.rb
dunkOnIT: Trying to invoke `create` directly without a `before` block yields the standard "this method is only accessible within the scope of a let or before block" error that RSpec often throws. 

Asking ChatGPT, it seems to be because of the scoping of invoking FactoryBot directly vs using a method that we have access to because FactoryBot was `required` in another file.

I won't pretend to fully understand the nuances of rspec's contexts, but this explanation makes enough sense to me to be fine with it - to be honest, I'm more surprised that sign_in worked _without_ a `before` block

gregorbg: Okay, sorry in case my previous comment was misunderstandable. Why is `before { sign_in create(:admin) }` not working?

(Especially seeing that `sign_in { FactoryBot.create :admin }` and variations thereof are definitely working in the hundreds in our current CI. The fact that just removing `FactoryBot.` prefix here seemingly breaks things is crazy to me)

dunkOnIT: Damn you're absolutely right, it does work - I just didn't think of that. Will make the changes, possibly in a new PR depending what is easier to modify.

dunkOnIT:
Closed in favour of #11338

PR #11306, opened 2025-04-12
Registration Edit: profile & edit user links

kr-matthews:
For competitors with IDs, we had "Name (ID) Icon" with both ID and Icon linking to the profile.
For newcomers, we had "Name () Icon" with nothing in the parens and Icon linking to `persons/null`.

Now, the ID links to the profile, or just says 'first-timer' if there's no id. And Icon links to the edit user page in both cases.
I also removed the message about registering with an account because that applies to everyone (not helpful), and the link is already built into the icon.

### Old

Same link twice:
image

Empty () and broken link:
image

### New

image

image


dunkOnIT (approved):
Damn I thought I'd fixed the blank brackets issue in the initial PR.

Nice changes, thanks a lot!

dunkOnIT:
Note: Tested locally and all works as expected, but not deploying as I'm away for the weekend in a couple of hours so won't be able to support any deployment issues

PR #11336, opened 2025-04-15
Changed homepage link to point directly to international store

dunkOnIT:
Per Hariprasad's request in "Removal of Indian store link", this PR restores the original functionality of the "Link to shop" button on the homepage - linking directly to the international shop. Also, `/merch` now redirects to the international shop as well, in case anyone accesses it directly.

The only reason /merch was introduced was to let users choose between the International and Indian stores. 

`merch.html.erb` not removed from static pages so that it is easy to re-enable when the Indian store comes back online.

### DIFF
--- a/app/views/posts/homepage.html.erb
+++ b/app/views/posts/homepage.html.erb
@@ -16,7 +16,7 @@
           <%= t('homepage.competitions.text_html',
                 wca_live_link: raw(link_to("WCA Live", "https://live.worldcubeassociation.org"))) %>
         <% end %>
-        <%= render "box", title: t('homepage.merchandise.title'), path: 'https://shop.worldcubeassociation.org', link_title: t('homepage.merchandise.link_title') do %>
+        <%= render "box", title: t('homepage.merchandise.title'), path: '/merch', link_title: t('homepage.merchandise.link_title') do %>
           <%= t('homepage.merchandise.text_html') %>
         <% end %>
         <%= render "box", title: t('homepage.records.title'), path: records_path do %>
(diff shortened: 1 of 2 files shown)


gregorbg (changes requested):
Kinda works, but now you're "orphaning" the Merch page. Is this the way we want to go? I'd suggest "cleanly" removing the Indian section from the merch view, especially as they have already hinted that it will come back soon.

dunkOnIT:
Two reasons to not make changes to the merch page: 
1. With the Indian page gone, there is no need for a dedicated merch page - it's just an extra, unnecessary click and thus better not to have it
2. "Orphaning" the merch page makes it very very easily to re-enable the Indian store - not that it would be particularly difficult otherwise, but another 2-line change is preferable to playing with HTML in my view. 

Basically, I don't see any need to point the user to the Merch page, besides the fact that we have it - which in my mind is not a good enough reason to send the user to an unnecessary page

app/views/posts/homepage.html.erb
```diff
@@ -16,7 +16,7 @@
           <%= t('homepage.competitions.text_html',
                 wca_live_link: raw(link_to("WCA Live", "https://live.worldcubeassociation.org"))) %>
         <% end %>
-        <%= render "box", title: t('homepage.merchandise.title'), path: '/merch', link_title: t('homepage.merchandise.link_title') do %>
+        <%= render "box", title: t('homepage.merchandise.title'), path: 'https://shop.worldcubeassociation.org', link_title: t('homepage.merchandise.link_title') do %>
```
gregorbg, line 19: With the redirect you got set up under `routes.rb`, this change now technically becomes obsolete. Leaving it at the previous `path: '/merch'` will trigger the redirect.

dunkOnIT: Fair point - changed

gregorbg (approved):

PR #11337, opened 2025-04-15
Refactor step panel to pull available steps from the backend

FinnIckler:
Currently I am just sending a hardcoded string from the backend. A more nested json structure will make sense.
The last step related business logic I want to get rid of is the initial index calculation

```
if (isPolling) {
  return steps.findIndex((step) => step === competingStepConfig);
}

if (registrationFinished || isAccepted || isRejected) {
  return steps.findIndex((step) => step === registrationOverviewConfig);
}

return steps.findIndex(
  (step) => step === (isRegistered ? paymentStepConfig : requirementsStepConfig),
);
```

It would be easy to just use the same logic and return a `currentStepIndex` from the `registration_config` route. Thoughts?


gregorbg:
> It would be easy to just use the same logic and return a `currentStepIndex` from the `registration_config` route. Thoughts?

What exactly does that index represent? If the user is opening the page for the first time, the index should of course be `0`. If the user already registered before, and a registration is present, the frontend could generically and "cleanly" (ie without hard-coded business logic about the individual steps) compute it along the lines of "the first regular step that reports 'not complete', or if all steps report complete, then the first post-step, or else the last step" or something along those lines.

The backend doesn't feel like the right place to copy the snippet you quoted into, because then you're mixing config with state. (In extreme cases, you'd have to retrieve the `Registration` entity in the config endpoint, just to compute the index. But in my view, the config endpoint should never retrieve the actual registration payload at all).

FinnIckler:
That is a good idea! The only issue is waived payment again. I think we can do if registrationFinished -> RegistrationOverview otherwise show first non completed step

app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
```diff
@@ -262,16 +265,16 @@ export default function CompetingStep({
...
-    if (competitionInfo['uses_qualification?']) {
+    if (currentStepParameters.qualifications) {
       selectedEventIds.update(
-        competitionInfo.event_ids.filter((e) => isQualifiedForEvent(
+        currentStepParameters.event_ids.filter((e) => isQualifiedForEvent(
           e,
-          qualifications.wcif,
-          qualifications.personalRecords,
+          currentStepParameters.qualifications,
+          personalRecords,
```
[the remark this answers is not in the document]
FinnIckler: true, haven't gotten there yet

app/webpacker/components/RegistrationsV2/Register/RegistrationOverview.jsx
```diff
@@ -85,7 +89,7 @@ export default function RegistrationOverview({
       )}
       <Segment loading={isDeleting}>
         <Header>{I18n.t('competitions.nav.menu.registration')}</Header>
-        <Form onSubmit={nextStep} size="large">
+        <Form onSubmit={() => jumpToStepByKey('competing')} size="large">
```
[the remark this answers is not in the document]
FinnIckler: we got make it jump to the first step that is not disabled?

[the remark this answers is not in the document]
gregorbg: Yeah, something along those lines. My preference is to not have a hard-coded step ID at this point, I'll leave it to you to decide what other alternative works best for you.

app/webpacker/components/RegistrationsV2/Register/RegistrationOverview.jsx
```diff
@@ -68,7 +72,7 @@ export default function RegistrationOverview({
       .then(() => (deleteAllowed
         ? deleteRegistrationMutation()
         : window.location = contactCompetitionUrl(competitionInfo.id, encodeURIComponent(I18n.t('competitions.registration_v2.update.delete_contact_message')))))
-      .catch(() => nextStep({ refresh: true }));
+      .catch(() => refreshStep());
```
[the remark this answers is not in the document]
gregorbg: then why have a `catch` at all, if all it consists of is just a noop?

FinnIckler:
I don't think this is better than the one with the indexes in the methods, but feel free to come up with a better way to calculate it that keeps the logic that we currently have here https://github.com/thewca/worldcubeassociation.org/blob/cc79584d65b25c78883f58512b04031279c9076c/app/webpacker/components/RegistrationsV2/Register/StepPanel.jsx#L35

dunkOnIT:
Comments after clicking around a bit: 
- [ ] [Vid 1] The behaviour when trying to click back on the Payment Step is very strange. I suggest we either: 
  - Disable the payment tab after payment
  - Make a "Payment View" that you can click on the Payment Step

However, I'd suggest a different approach from making going back in the steps more friendly: Don't allow the user to go back in the Steps unless one of the steps has been invalidated (ie, payment cancelled, or registration deleted). 

**Strong opinion**: User should be able to see summary details of all steps on the Overview panel - ie, clicking back to Payment Step should not be necessary for seeing payment info
**Weaker opinion:** When a user wants to update information from a Step, they shouldn't go back in the process - that step is already submitted and going back in the flow makes sense. The submission of an update is an addendum to the already-completed flow, so rendering the interface for updating details of a step in the Overview Step makes sense to me - and also helps us avoid a lot of awkward/confusing cases when we're allowing a user to go back to the Register step to edit, but then they skip over the Payment step. 

This does start to fall apart when, for example, updating events yields extra charges that make an updated payment necessary. So I'm sort of talking myself out of this weaker opinion - but leaving the comment here in case it sparks useful discussion.


Vid 1: 

https://github.com/user-attachments/assets/5057f67f-6753-4b8c-822f-b0e3c3727579

gregorbg:
Resetting this to draft until we figure out a plan of action based on Duncan's extensive review

app/controllers/api/v1/registrations_controller.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
-class Api::V1::RegistrationsController < Api::V1::ApiController
-  skip_before_action :validate_jwt_token, only: [:index]
+class Api::V1::Registrations::RegistrationsController < Api::V1::ApiController
```
gregorbg: The `Registrations::` prefix here doesn't seem right. The folder structure is only `api/v1/registrations_controller`, and the module declaration should match that.

app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
```diff
@@ -260,16 +264,16 @@ export default function CompetingStep({
   };
 
   const onAllEventsClick = () => {
-    if (competitionInfo['uses_qualification?']) {
+    if (currentStepParameters.qualifications_wcif) {
```
gregorbg: I think we should keep relying on the WCAT-configured boolean here, instead of implicitly assuming "Ah if a qualifications WCIF exists, then we should probably enforce it"

app/views/registrations/register.html.erb
```diff
@@ -26,14 +26,9 @@
...
-                                                    connectedAccountId: @competition.payment_account_for(:stripe)&.account_id,
-                                                    qualifications: {
-                                                      wcif: @competition.qualification_wcif,
-                                                      personalRecords: {
-                                                        single: current_user.person&.ranks_single&.map(&:to_wcif) || [],
-                                                        average: current_user.person&.ranks_average&.map(&:to_wcif) || [],
-                                                      }
+                                                    personalRecords: {
+                                                      single: current_user&.ranks_single&.map(&:to_wcif) || [],
+                                                      average: current_user&.ranks_average&.map(&:to_wcif) || [],
```
gregorbg: I think `preferredEvents` and `personalRecords` should be part of the competing step config

app/webpacker/components/RegistrationsV2/Register/StripePaymentStep.jsx
```diff
@@ -96,11 +97,21 @@ function PaymentStep({
...
   const [isLoading, setIsLoading] = useState(false);
   const [isDonationChecked, setDonationChecked] = useCheckboxState(false);
 
+  useEffect(() => {
+    // TODO When we add per Event Payment this logic needs to also check
+    //  if an additional payment is needed
+    if (hasPaid) {
+      nextStep();
+    }
+  }, [nextStep, hasPaid]);
```
gregorbg: I feel like we got rid of this at some point, did the merge let it creep back in or is my memory wrong?

app/webpacker/components/RegistrationsV2/Register/StripePaymentStep.jsx
```diff
@@ -162,17 +173,17 @@ function PaymentStep({
...
+              id="donationInputField"
+              onChange={(_, { value }) => setIsoDonationAmount(value)}
+              currency={competitionInfo.currency_code}
+              value={isoDonationAmount}
+              label={(
+                <Label>
+                  {I18n.t('registrations.payment_form.labels.donation')}
+                </Label>
+              )}
+            />
```
gregorbg: Odd indentation changes

app/webpacker/components/RegistrationsV2/Register/StripeWrapper.jsx
gregorbg, line 1: The whole file is eerily similar to `StripePaymentStep`. Methods like `convertISOAmount` are duplicated, and the parameters of the default export function are 100% identical. I feel like something went wrong here during merge.

app/webpacker/components/RegistrationsV2/Register/StripePaymentStep.jsx
```diff
@@ -96,11 +97,21 @@ function PaymentStep({
   const stripe = useStripe();
   const elements = useElements();
   const dispatch = useDispatch();
-  const { registration } = useRegistration();
+  const { nextStep } = useStepNavigation();
```
gregorbg: This is currently conflicting with the `nextStep` that is still being passed in as prop above. This observation confirms my suspicion that something went very wrong with `StripeWrapper` during merge.

app/webpacker/components/RegistrationsV2/Register/RegistrationOverview.jsx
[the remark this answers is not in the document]
gregorbg: Kindly ping, since the step navigation context now exposes `jumpToFirstIncompleteStep`

app/views/registrations/register.html.erb
FinnIckler: What about `userCanPreRegister` and `cannotRegisterReasons` ?

gregorbg: For now, that is fine as-is, because those two are supposed to act globally on the whole registration lane.
As soon as we introduce support for multiple lanes, we need to rethink them as well. But not necessary right now.

app/webpacker/components/RegistrationsV2/Register/RegistrationOverview.jsx
```diff
@@ -99,7 +99,7 @@ export default function RegistrationOverview({
       )}
       <Segment loading={isDeleting}>
         <Header>{I18n.t('competitions.nav.menu.registration')}</Header>
-        <Form onSubmit={() => jumpToStepByKey('competing')} size="large">
+        <Form onSubmit={() => jumpToFirstIncompleteStep} size="large">
```
gregorbg: Either add `jumpToFirstIncompleteStep()` or remove the `() =>` at the front 😛

app/controllers/api/v1/registrations_controller.rb
```diff
@@ -76,6 +76,11 @@ def show_by_user
     render json: @registration.to_v2_json(admin: true)
   end
 
+  def registration_config
+    competition = Competition.find(params[:competition_id])
+    render json: competition.available_registration_lanes(@current_user), status: :ok
```
gregorbg: `status: :ok` is the default and thereby redundant

app/models/competition.rb
```diff
@@ -1767,6 +1767,45 @@ def self.search(query, params: {}, managed_by_user: nil)
...
+                      methods: [:qualification_wcif, :event_ids],
+                      include: [])
+    user_params = {
+      preferredEvents: current_user.preferred_events.pluck(:id),
+      personalRecords: {
+        single: current_user.ranks_single&.map(&:to_wcif) || [],
+        average: current_user.ranks_average&.map(&:to_wcif) || [],
+      },
+    }
+    competition_params.merge!(user_params)
```
gregorbg: `merge` (without the exclamation point) should be sufficient here. `merge!` modifies `competition_params` in-place, but since you're caring only about the overall return value, the result of `merge` should be enough.

FinnIckler:
Fixed two points from Duncan's list, but we need to work on saving t&c/requirements approval so we can implement
`isCompleted: (stepPayload)` for the `requirementsStepConfig`

dunkOnIT (approved):
QA completed - happy to merge from my side!

PR #11339, opened 2025-04-16
Mailcatcher fix: Manually require dependencies

dunkOnIT:
Mailcatcher was yielding errors indicating that its dependencies weren't being included. I wasn't able to figure out from a cursory glance why this wasn't the case - perhaps some issue with Ruby 3.4.2[1]? 

I took the "obvious" approach of just manually installing the dependencies in the bash commands until it worked - and now it's working locally again. Open to more refined approaches if anyone can see what the underlying issue might be! 

[1] Finn pointed out in Slack that my Rails seemed to think it was still v3.4.0 - this is the output in the Docker terminal, which makes me think my Docker is on the right Ruby version: 
```bash
duncan@duncan-elitebook:~$ docker exec -it rails bash
root@1e12c9feb5cc:/app# ruby -v
ruby 3.4.2 (2025-02-15 revision d2930f8e7a) +PRISM [x86_64-linux]
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -111,7 +111,9 @@ services:
     ports:
       - "1080:1080"
     command: >
-      bash -c 'gem install mailcatcher &&
+      bash -c 'gem install eventmachine -v "~> 1.0" &&
+      gem install thin -v "~> 1.8" &&
+      gem install mailcatcher &&
       mailcatcher --foreground --ip=0.0.0.0'
 
   wca_sidekiq:


gregorbg (changes requested):
I am happy to see that you managed to figure out a solution for your local installation, but this is not a sustainable fix for our global dev environment.

`gem install mailcatcher` definitely resolves the dependencies (in this case, `thin` and `eventmachine`) for you. Manually specifying them for everyone on the team is not necessary.

I'd be willing to take ~30 minutes to try and debug this on a call with you later this week, but this PR definitely does not need to be merged for everyone. It is a bug/choke/... in your specific local Docker that makes it break, I am absolutely confident beyond any doubt that `gem install mailcatcher` is supposed to work without having to also list individual dependencies upfront.

FinnIckler:
Just saying that I have been having the same issue for the last 3-4 months.

Also: https://staging.worldcubeassociation.org:444/ is running and is just doing
```
if [ "$WCA_LIVE_SITE" = "0" ]; then
  gem install mailcatcher
  mailcatcher --http-ip=0.0.0.0 --no-quit
fi
```

dunkOnIT:
Yeah, Finn's comment when I asked on the Slack were why I thought this might be a greater-than-Duncan issue

PR #11347, opened 2025-04-18
Fix for create newcomers and import results page

pranav-027:
Fix for create newcomers and import results page


gregorbg:
Thanks for your submission, all pushed through via hotfix.

PR #11350, opened 2025-04-20
Corrected delegate edits schema to use competitorCanCancel

dunkOnIT:
We recently introduced the ability for delegates to edit fields in the Competition Form after a comp has been announced. The editable fields are defined in `models/competition.rb#self.delegate_edits_json_schema`. 

This schema still made use of a deprectaed field for determining whether users can self-cancel their registration - deprecated field is `allow_self_delete_after_acceptance`, new field to use is `competitor_can_cancel` - meaning that delegates were getting errors when trying to change their self-cancel settings.

This PR updates the schema to support `competitor_can_cancel`, which (tested locally) resolves the errors delegates are receiving.

Follow-up: Mark `allow_self_delete_after_acceptance` as deprecated, or delete it outright. I'll assess how widely used it is in the code and draft a PR.


gregorbg (commented):
Forgive me if I'm overlooking something completely obvious, but the original "main" schema also still uses `allowSelfDeleteAfterAcceptance`...?!

gregorbg:
Okay, apparently GitHub's diff view is drunk

gregorbg:
Could it be that your local `main` branch is spectacularly out of date?
image

gregorbg:
Oh no, it's actually our code... The value is wrong in the original schema, but we don't use `"additionalProperties" => false` over there...

dunkOnIT:
Oh interesting. How does it work at all then?

gregorbg:
Simply put, it works because it is never checked. Crucially, the big master schema does **not** specify `additionalProperties: false`. That means, whenever a change with `competitorCanCancel` comes through, the schema checker sees "Oh there's an extra property that I don't have any schema specifications for. It's an `additionalProperty` which is currently not forbidden, so it's fine".

Likewise, it was still holding an expected schema for the old `allowCompetitorSelfDelete`, but since merging that other PR a few weeks/months ago, that property never appeared in the actual payload again. That's not a problem though because we did not mark it as being required.

gregorbg (approved):

PR #11351, opened 2025-04-20
Remove `allow_registration_self_delete_after_acceptance` from `competition` table

dunkOnIT:
This field was deprecated by the introduction of `competitor_can_cancel` (#10807). I've grepped for calls to this property in the codebase and found none besides
- a validation against the field's value
- usage in migrations

so looks safe to remove outright.

There are also still keys with this field's name in i18n files - not sure how these get resolved/removed. The relevant keys seem to be removed from en.yml, but that doesn't appear to automatically propagate to other i18n files?

### DIFF
--- a/app/models/competition.rb
+++ b/app/models/competition.rb
@@ -126,7 +126,6 @@ class Competition < ApplicationRecord
     on_the_spot_registration
     on_the_spot_entry_fee_lowest_denomination
     allow_registration_edits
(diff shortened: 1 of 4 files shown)


gregorbg (commented):
I will gladly take a rare opportunity to remove something from the `competitions` table rather than adding something :D

gregorbg (approved):

PR #11355, opened 2025-04-20
Fix a variable name typo in development seeds

bohdancho:
This makes `docker exec -it rails bash -c "bin/rake db:reset"` work again.


gregorbg:
Thanks! Instead of adding the underscore where you did, you should rather remove it at the place it came from I think 🤔

bohdancho:
Agreed! Here's a fix.

dunkOnIT (approved):

gregorbg (approved):
Thanks, must have been an oversight from our recent `snake_case` migration wave.

Issue #10915, opened 2025-02-25
Profile Page


dunkOnIT:

dunkOnIT:
Feedback/suggestion received via Discord:
> I'd suggest making it so that if the user is an organiser, the tab is titled "Competitions Organised", and if the user is a delegate, it's title "Competitions Organised/Delegated", with icons (or visual cues of some kind) to differentiate the competitions delegated from the competitions organised.

OWelton-Rosie:
Hi Duncan,

Thanks for adding my DM to this issue thread.

I'd suggest having a line break and a heading (or an <hr> and a heading) to show the difference between WCA and RO roles to prevent confusion. 

I would also suggest making it so that if the user clicks/taps on an RO role badge, it takes them to the RO's website.

pranav-027:
I have implemented the following things in https://github.com/thewca/worldcubeassociation.org/pull/11357

- Number of competitions as listed organizer
- Overview of competitions as listed organizer

According to my analysis, the profile badge depends on user roles. Are we planning to introduce a new organizer role?


> I'd suggest making it so that if the user is an organiser, the tab is titled "Competitions Organised", and if the user is a delegate, it's title "Competitions Organised/Delegated", with icons (or visual cues of some kind) to differentiate the competitions delegated from the competitions organised.

I haven't thought much about this my first thought was to just create a separate section for delegated comps. But if anyone has any design ideas, I'd love to implement them!

PR #11357, opened 2025-04-20
organized competitions on persons page

pranav-027:
Started work on: https://github.com/thewca/worldcubeassociation.org/issues/10915

https://github.com/user-attachments/assets/5410391c-1fa5-42d3-b30e-16e2a35f2dba


maxidragon:
Is there any point in working on this now? The page will be migrated to React/Next.js and redesigned, so I think it makes much more sense to do it after the migration as it's definetely not an urgent feature

gregorbg:
Works in principle, although I'm not too happy with the simple shape of the table in the bottom details section. Especially on large screens, it shows only the name and a loooooot of blank space.

I'd much rather think about this as part of the redesign process, cf. https://github.com/thewca/worldcubeassociation.org/pull/11344

PR #11380, opened 2025-04-21
Implement forum age restriction

dunkOnIT:
In the email thread "Adding Age Requirement to WCA Forum", WCT has requested that we implement an age requirement of 13 years old for signing up to the forum - this will align us with COPPA. From a data protection perspective, WRT confirmed that we can use age data to enforce this type of restriction.

This PR blocks SSO if the user's age is below 13, following a similar pattern to how we prevent banned users from registering. Importantly, this PR also adds a `misc` category for i18n keys - I couldn't see a better category to put a one-off forum key in.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -74,6 +74,8 @@ class User < ApplicationRecord
   ANONYMOUS_GENDER = 'o'
   ANONYMOUS_COUNTRY_ISO2 = 'US'
 
+  FORUM_AGE_REQUIREMENT = 13
+
   def self.eligible_voters
     [
       UserGroup.delegate_regions,
@@ -490,12 +492,22 @@ class User < ApplicationRecord
     user.senior_delegates.include?(self)
   end
 
-  def banned?
-    group_member?(UserGroup.banned_competitors.first)
+  def age_in_years
+    age = Date.today.year - dob.year
+    age -= 1 if Date.today < dob.advance(years: age)
+    age
+  end
+
+  def below_forum_age_requirement?
+    age_in_years < FORUM_AGE_REQUIREMENT
   end
 
   def forum_banned?
     current_ban&.metadata&.scope == 'competing_and_attending_and_forums'
+  end
+
+  def banned?
+    group_member?(UserGroup.banned_competitors.first)
   end
 
   def banned_in_past?
--- a/db/migrate/20250421053202_add_invoice_items_table.rb
+++ b/db/migrate/20250421053202_add_invoice_items_table.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+class AddInvoiceItemsTable < ActiveRecord::Migration[7.2]
+  def change
+    create_table :invoice_items do |t|
+      t.belongs_to :registration
+      t.integer :amount_lowest_denomination
+      t.string :currency_code
+      t.integer :status
+      t.string :display_name
+      t.timestamps
+    end
+  end
+end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -264,6 +264,9 @@ class UsersController < ApplicationController
     if current_user.forum_banned?
       flash[:alert] = I18n.t('registrations.errors.banned_html').html_safe
       return redirect_to new_user_session_path
+    elsif current_user.below_forum_age_requirement?
+      flash[:alert] = I18n.t('registrations.errors.forum_age_requirement').html_safe # TODO: Consider alternative i18n key path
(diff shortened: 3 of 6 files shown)


app/models/user.rb
```diff
@@ -490,14 +492,24 @@ def senior_delegate_for?(user)
     user.senior_delegates.include?(self)
   end
 
-  def banned?
-    group_member?(UserGroup.banned_competitors.first)
+  def age_in_years
+    age = Date.today.year - dob.year
```
gregorbg, line 496: `dob` is not necessarily set. Users who "just" sign up without ever having competed before are not required to enter their DOB.

This value is only "backfilled" as part of the "Claim WCA ID" process. Note that users can claim a WCA ID directly while signing up/creating their account, but that doesn't change the fact that this field is only filled if a WCA ID has been claimed.

app/models/user.rb
```diff
@@ -490,14 +492,24 @@ def senior_delegate_for?(user)
...
+    age -= 1 if Date.today < dob.advance(years: age)
```
gregorbg, line 497: What?!

app/models/user.rb
```diff
@@ -490,14 +492,24 @@ def senior_delegate_for?(user)
...
   end
 
   def forum_banned?
     current_ban&.metadata&.scope == 'competing_and_attending_and_forums'
   end
 
+  def banned?
+    group_member?(UserGroup.banned_competitors.first)
+  end
+
```
gregorbg, line 512: This feels unrelated

db/migrate/20250421053202_add_invoice_items_table.rb
gregorbg, line 1: This is definitely unrelated

gregorbg (changes requested):
I don't really care about the I18N key (ie it's fine by me).

The much less trivial change about this feature is that we now require users to have an associated WCA ID in order to use the forums. Otherwise we don't have any (verified) DOB information. Details are in my inline code comments.

app/models/user.rb
dunkOnIT: Hmm, interesting. Obviously I need to add `nil` support, but more deeply we need to get a clear set of requirements from WCT on how to handle this - requested via email

app/models/user.rb
dunkOnIT: Restructured to make it easier to grok - originally, this was advancing the dob to the current year to compare the current date to the user's date of birth. New version makes that more explicit

db/migrate/20250421053202_add_invoice_items_table.rb
dunkOnIT: Removed

gregorbg:
Setting this to draft until the issues around WCA IDs have been resolved.

lnzainn:
@gregorbg @dunkOnIT we have discussed and the consensus is to allow users to access it as long as they have provided date of birth that is > 13 and that having WCA ID is not mandatory.

gregorbg:
Thank you, this unfortunately makes the implementation a bit more tricky because we currently don't store DOBs for users who have never competed before. It's not rocket science to add this feature, and the Board is also strongly in favor of prioritizing this PR/feature, but it will still take a non-trivial amount of time to figure this one out.

dunkOnIT:
Marked as blocked until we come to consensus on:
- priority of this feature from the Board's perspective
- implications for WCA's data model more broadly if date of birth information can be stored pre-verification from delegates

dunkOnIT:
Unblocked per Callum's email in "Adding Age Requirement to WCA Forum". 

My latest commit redirects the user to the Edit Profile if their `dob` is not filled in, and prompts them to do populate it with a flash message - this should be sufficient to bring the PR in line with requirements

gregorbg:
Implementation looks good code-wise. Do we really want to redirect people to the edit form if they don't have a DOB set? That's like inviting them to cheat and fill in a fantasy value.

dunkOnIT:
At the point of redirecting them, we aren't telling them that there's an age restriction - only that they need a dob, and I think redirecting them to the edit panel is nice QOL in that context.

Only once they have a dob, do we then tell them that they can't access the forum - and that message redirects them to a new session, not to the edit page.

gregorbg (approved):
Please coordinate with WCT whether and when this should be merged

lnzainn:
@dunkOnIT I tried testing this and it looks like a WCA account without a WCA ID is still able to access the forum even if the specified date is not below 13

dunkOnIT:
@lnzainn Sorry, I forgot to deploy! Please try it again and let me know if you have any further issues

PR #11338, opened 2025-04-16
rubocop-factorybot: syntax and consistent parenthesis

dunkOnIT:
Replaces #11303, putting the `create` call in the `before` block instead of in a separate `let`.

PR #11386, opened 2025-04-22
Enable FactoryBot cops

dunkOnIT:
Opened a new PR (and closed #11338) because it's easier to do the replaces from scratch than resolve merge conflicts


gregorbg (approved):
I'm approving this based on its "predecessor" PR where the line change count was exactly identical

PR #11387, opened 2025-04-22
Enable Capybara RuboCops

gregorbg:
See title. All auto-correctable fixes except a few where I had to swap two lines

### DIFF
--- a/spec/features/competition_management_spec.rb
+++ b/spec/features/competition_management_spec.rb
@@ -111,7 +111,7 @@ RSpec.feature "Competition management", :js do
       expect(page).to have_current_path(edit_competition_path("NewId2016"), wait: SLUGGISH_WAIT_TIME)
 
       expect(page).to have_text("This competition is not visible to the public.")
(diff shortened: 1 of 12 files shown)


spec/features/competition_management_spec.rb
```diff
@@ -193,7 +193,8 @@ def wca_registration_checkbox
       competition = FactoryBot.create(:competition, :with_delegate, id: "OldId2016", guests_entry_fee_lowest_denomination: 666)
       visit edit_competition_path(competition)
 
-      expect(page).not_to have_text("Display message for free guest entry")
+      expect(page).to have_text(competition.name)
```
dunkOnIT, line 196: Where did this come from? Looks like a new line

dunkOnIT: Makes sense that the cop requires a positive assertion to succeed before we're "allowed" to make an assertion about the lack of presence of a particular element

dunkOnIT (approved):

PR #11390, opened 2025-04-23
Enable Rubocop WordArray and SymbolArray plus related Style cop

gregorbg:
The whole diff comes from the auto-correct feature, and should hopefully be self-explanatory


dunkOnIT (approved):
I've skimmed the diff, but not reviewed each line in depth. The line count makes sense, because we're losing 9 lines in `rubocop.yml`.

Before merging, you'll also want to check these commits I made (first one includes some inconsistent whitespace which is resolve in the 2 subsequent commits):
- [05f97a9](https://github.com/thewca/worldcubeassociation.org/pull/11390/commits/05f97a96db1b51fd747e526b829b06d1610fa732)
- [05f97a9](https://github.com/thewca/worldcubeassociation.org/pull/11390/commits/05f97a96db1b51fd747e526b829b06d1610fa732)
- https://github.com/thewca/worldcubeassociation.org/pull/11390/commits/383c82b490a394b23a0e25610855c3f9ee69af14

db/migrate/20250428123246_add_competing_status_index.rb
dunkOnIT, line 1: Did this creep in because I had it in my local branch? If so, is it a bad thing, or would it be applied when it merges with main in any case?

gregorbg: What do you mean? After merging, of course the cop will apply to new changes as well (if you rerun Rubocop, that is)

dunkOnIT: I didn't merge anything, so I was worried that it was a file I had locally that wasn't "hidden" by git when changing branches (as sometimes seems to happen with migrations)

But I assume this is moot now that main is merged in anyway

PR #11394, opened 2025-04-23
Activate SymbolProc Cop

gregorbg:
Everything in the diff comes from the auto-correct routine. Added a few exceptions, notably:
- The Capybara DSL
- The configuration initializers

:warning: This was using the **unsafe** auto-correct, read more about the safety here: https://docs.rubocop.org/rubocop/1.75/cops_style.html#safety-stylesymbolproc (I don't think this affects us, but I might be overlooking something)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -170,7 +170,15 @@ Style/WordArray:
   Enabled: false
 
 Style/SymbolProc:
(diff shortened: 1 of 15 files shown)


dunkOnIT (approved):
#

PR #11398, opened 2025-04-23
Reject payment ticket when `outstanding_entry_fees` == 0

dunkOnIT:
Currently, if a user has fully paid for their registration, there is nothing stopping them from submitting another payment for the same registration[1]. At present, we have no use case for multiple payments against a single registration, and it is causing issues for organizers, so this "hotfix" bakes in the explicit assumption that a user should not be allowed to submit payments when `outstanding_entry_fees` is 0.

In the future, this can be handled more subtly when we use invoice_items and can determine the individual payment state of a specific invoice item. 

Additionally, I have noticed that the PaymentStep is not rendering errors from `getPaymentTicket` correctly - I will be doing a follow-up PR to fix this, but that will involve frontend changes that I don't want to slow down this hotfix.

[1] Apart from the frontend _usually_ not showing them the payment form - but this can be trivially circumvented by opening two windows both showing the payment panel, paying through the first one, and then paying through the second one. Users should not be able to do this, even by accident.


app/controllers/api/v1/registrations/registrations_controller.rb
```diff
@@ -221,11 +221,11 @@ def validate_payment_ticket_request
     competition_id = params[:competition_id]
     @competition = Competition.find(competition_id)
     return render_error(:forbidden, Registrations::ErrorCodes::PAYMENT_NOT_ENABLED) unless @competition.using_payment_integrations?
-
```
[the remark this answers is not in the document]
dunkOnIT: It's a reorganization for readability - grouping the competition-related errors and registration-related errors together in a more obvious way: 
```ruby
@competition = Competition.find(competition_id)
    return render_error(:forbidden, Registrations::ErrorCodes::PAYMENT_NOT_ENABLED) unless @competition.using_payment_integrations?
    return render_error(:forbidden, Registrations::ErrorCodes::REGISTRATION_CLOSED) if @competition.registration_past?

    @registration = Registration.find_by(user: @current_user, competition: @competition)
    return render_error(:forbidden, Registrations::ErrorCodes::PAYMENT_NOT_READY) if @registration.nil?
    render_error(:forbidden, Registrations::ErrorCodes::ALREADY_PAID) if @registration.outstanding_entry_fees == 0
```

spec/requests/api_registrations_spec.rb
```diff
@@ -1312,16 +1312,27 @@
       expect(payment_record.currency_code).to eq("usd")
     end
 
-    it 'refuses ticket create request if registration is closed' do
-      closed_comp = create(:competition, :registration_closed, :with_organizer, :stripe_connected)
-      closed_reg = create(:registration, :pending, competition: closed_comp)
+    describe 'refuse ticket create request' do
+      it 'refuses ticket if registration already paid', :only do
```
gregorbg, line 1316:
```suggestion
      it 'refuses ticket if registration already paid' do
```

gregorbg, line 1316:
```suggestion
      it 'if registration already paid', :only do
```

spec/requests/api_registrations_spec.rb
```diff
@@ -1312,16 +1312,27 @@
...
-      get api_v1_registrations_payment_ticket_path(competition_id: closed_comp.id), headers: headers
+        body = response.parsed_body
+        expect(response).to have_http_status(:forbidden)
+        expect(body).to eq({ error: Registrations::ErrorCodes::NO_OUTSTANDING_PAYMENT }.with_indifferent_access)
+      end
 
-      body = response.parsed_body
-      expect(response).to have_http_status(:forbidden)
-      expect(body).to eq({ error: Registrations::ErrorCodes::REGISTRATION_CLOSED }.with_indifferent_access)
+      it 'refuses ticket create request if registration is closed' do
```
gregorbg, line 1325:
```suggestion
      it 'if registration is closed' do
```

gregorbg (changes requested):

app/webpacker/components/RegistrationsV2/api/helper/error_codes.js
```diff
@@ -27,3 +27,4 @@ export const REGISTRATION_ALREADY_EXISTS = -4014;
 export const REASONABLE_GUEST_COUNT = -4015;
 export const PAYMENT_NOT_ENABLED = -6001;
 export const PAYMENT_NOT_READY = -6002;
+export const DUPLICATE_PAYMENT = -6003;
```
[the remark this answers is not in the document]
dunkOnIT: Reworded per this suggestion - both the error code name and the i18n string

spec/models/registration_spec.rb
gregorbg, line 1: These changes don't really feel related...?

dunkOnIT: I can't see how they are related, but I can say with confidence that these tests were failing because the auto close hooks were being sent to a different instance of the same model object (ie, same `id`, different `object_id`). Wasn't able to figure out why the behaviour changed, so I changed the tests to accommodate this

gregorbg: Then I'm afraid we will have to dig deeper. There are currently zero instances of `expect_any_instance_of` in our entire codebase, I'm not going to allow this pattern to start now. It's the kind of thing that can get out of control real quick and later break our necks when we think "but we had tests for this, didn't we??" when the tests were actually operating on two completely separate database entities.

gregorbg: My initial suspicion would be that this has something to do with the changes to the `registration_payments` factory.

gregorbg (changes requested):

spec/models/registration_spec.rb
dunkOnIT: Yep makes sense! Converting to draft while I work on review feedback for invoice items and auto-accept

spec/models/registration_spec.rb
```diff
@@ -597,7 +597,9 @@
     it 'positive registration_payment calls registration.consider_auto_close' do
       competition = create(:competition)
       reg = create(:registration, competition: competition)
-      expect(reg).to receive(:consider_auto_close)
+      expect_any_instance_of(Registration).to receive(:consider_auto_close) do |instance|
```
gregorbg, line 600: This still has `expect_any_instance_of` which I am still vetoing :P
Am I missing something here since you said "Tests have been fixed"?

gregorbg (changes requested):

spec/models/registration_spec.rb
```diff
@@ -631,7 +631,7 @@
 
     it 'calls competition.attempt_auto_close! if reg is fully paid' do
       competition = create(:competition)
-      expect(competition).to receive(:attempt_auto_close!).exactly(1).times
+      expect(competition).to receive(:attempt_auto_close!)
```
gregorbg, line 634: Just curious, why is the "exactly 1 time" assumption not working anymore? And can we reasonably compute what the new count should be instead?

dunkOnIT: We can, and I've included it for both method calls! Not sure why one didn't have it before

gregorbg:
Yay! Now you just need to get rid of your `:tag` thingy ;P

gregorbg (approved):

Issue #11399, opened 2025-04-23
Add edit icon on the website posts


lnzainn:
![Image](https://github.com/user-attachments/assets/d7dcb602-5248-43b8-af09-c0107657cfa8)

It would be an amazing QOL if there was an edit icon (pencil logo) on the right hand side of the posts. This feature used to exist but then was removed maybe during the react transition.

FinnIckler:
it does exists in https://www.worldcubeassociation.org/posts/ though?

lnzainn:
It used to be there when you opened the individual posts too.

PR #11409, opened 2025-04-25
add edit text as icon fallback

dunkOnIT:
Hotfix for #11399 - I gave myself 30 minutes to try and look into/resolve this, and found the following: 
- We are still rendering posts with Ruby, not React - despite other parts of the Post stack using React
- It appears that the icon `icon pencil alternate` is not being found; I'm not sure why, although I suspect we may have removed FontAwesome, which is where the icon seems to be from? 
  - The reason we still have pencil icons in `/posts` (ie, the list view) is because that uses the `PostsWidget` React component. 

**My fix:** Adding "Edit" in the icon's span ensures that something is rendered so that WCT can continue to edit posts from within the posts page, as requested by Zain. 

**Better Fixes:**
- Figure out why the icon isn't being rendered and resolve that, or
- Ideally, rewrite this page in React
  - Not particularly difficult, but it includes permissions check that make such a rewrite just-untrivial-enough that I don't want to delve into it now

### DIFF
--- a/app/views/posts/_post.html.erb
+++ b/app/views/posts/_post.html.erb
@@ -9,7 +9,7 @@
       <% if current_user&.can_create_posts? %>
         <span class="pull-right">
           <%= link_to edit_post_path(post.slug) do %>
-            <span class="icon pencil alternate"></span>
+            <span class="icon pencil alternate">Edit</span>
           <% end %>
           <% if post.persisted? %>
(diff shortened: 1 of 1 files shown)


gregorbg:
The same bug affects the "Delete" icon, so here's a more permanent fix: Replace the `span` tag by an `i` tag.

dunkOnIT:
Fixed by changing <i> tags:
image

gregorbg (approved):

Issue #11388, opened 2025-04-22
Users are unable to update their registration while it is pending or waitlisted.


pranav-027:
Users are not able to edit their registrations while it is pending and they are being redirected to contact form even if competition settings are -
![image](https://github.com/user-attachments/assets/813ea84d-4e58-48a2-9c15-e69f4f1829ec)

dunkOnIT:
Can you give me a user and competition ID?

pranav-027:
https://www.worldcubeassociation.org/competitions/PuneSummerOpen2025

I tried for myself and also by registering as a new account

pranav-027:
https://www.worldcubeassociation.org/competitions/PuneSummerOpen2025

I tried for myself and also by registering as a new account

PR #11410, opened 2025-04-25
allow pending or waitlist registrations to be edited by competitiors

pranav-027:
fixes: https://github.com/thewca/worldcubeassociation.org/issues/11388

When the option is:
image

### DIFF
--- a/app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
+++ b/app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
@@ -199,11 +199,8 @@ export default function CompetingStep({
     guests,
   ]);
 
-  const canEditRegistration = (competitionInfo, registration) => {
-    return (
-      competitionInfo.allow_registration_edits || ['pending', 'waiting_list'].includes(registration.competing.registration_status)
-    );
-  };
+  const canEditRegistration = (competitionInfo, registration) =>
+    competitionInfo.allow_registration_edits || ['pending', 'waiting_list'].includes(registration.competing.registration_status);
 
 
   const actionUpdateRegistration = useCallback(() => {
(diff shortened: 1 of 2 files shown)


app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
```diff
@@ -199,11 +199,15 @@ export default function CompetingStep({
     guests,
   ]);
 
+  const canEditRegistration = (compInfo, regInfo) => (
+    compInfo.allow_registration_edits || ['pending', 'waiting_list'].includes(regInfo.competing.registration_status)
```
gregorbg, line 203: I think this comparison of the registration status should at the very least be extracted into its own variable. Or even better yet, be part of the `RegistrationContext`!

gregorbg (changes requested):
Left a comment in the code about the implementation.

About the general effect of this: Are we absolutely 100% sure (in terms of WCRP and real-life registration flows) that editing your registration is definitely fine while on the waiting list or while pending? I'm talking only about the real-life "WCA logic" here, not about the code.

app/webpacker/components/RegistrationsV2/Register/CompetingStep.jsx
pranav-027: Done!

pranav-027:
> About the general effect of this: Are we absolutely 100% sure (in terms of WCRP and real-life registration flows) that editing your registration is definitely fine while on the waiting list or while pending? I'm talking only about the real-life "WCA logic" here, not about the code.

Well, the option says "Competitors may NOT edit their registration **once it has been accepted**"; this implies that edits have to be restricted after they have been accepted!  I checked WCRP and didn't find anything relevant. I have mailed WCAT asking a clarification about it. Let's see what they think about this

pranav-027:
@gregorbg WCAT confirmed this as an intended functionality!

image

gregorbg (changes requested):
Thanks, now that you're doing the same `bla... || isPending || isWaitingList` in two different places, you might as well extract that to its own (memoed) variable.

pranav-027:
Done!

gregorbg (approved):

PR #11433, opened 2025-04-28
Fix unintended null pointer error in Stripe Webhook

gregorbg:
With the recent Stripe API upgrade (see https://github.com/thewca/worldcubeassociation.org/pull/11298) the Webhook now sends some new events that we don't want/need to handle but that are hard to filter.

Our existing filter mechanism only returned 404 if the `StripeRecord` was missing on an event that was handled by us. However, for events that are *not* handled by us, it's actually much, much more likely (and also not harmful at all) that we don't have the `StripeRecord` on file. Ironically, in that case we were trying to load the connected account on a Stripe record that we did not find, leading to errors.

TL;DR: The point in time when we abort with a `404` changes to a tiny bit earlier. The overall webhook structure is completely unaffected.

### DIFF
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -313,18 +313,18 @@ class RegistrationsController < ApplicationController
     stripe_intent = event.data.object # contains a polymorphic type that depends on the event
     stored_record = StripeRecord.find_by(stripe_id: stripe_intent.id)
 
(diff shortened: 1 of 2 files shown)


gregorbg:
Viewing the diff without whitespace changes makes it very obvious what this PR does / tries to do.

dunkOnIT (approved):

PR #11435, opened 2025-04-28
Don't serialize registration_status for past competitions in MyComps page

gregorbg:
We used a default set of options for all competitions appearing on the My Competititons page. But for past competitions, we don't actually need the registration status, so we can save a lot of `COUNT` SQL calls by just skipping this field. This should be beneficial mostly for people with a lot of past competitions.


dunkOnIT (changes requested):
Yields this error when testing locally - tried docker compose down/up, which didn't help, and confirmed that it's specific to this PR by checking that the error does not appear when I'm in `main`: 
image

gregorbg:
Very obscure error message, but should be fixed now!

PR #11436, opened 2025-04-29
Be less strict on Edit Events Deadline in competition form soft edits

### LINKED ISSUE #11416: Allow extension of event change deadline once it has passed
Currently, delegates cannot add OTS registration after the event change deadline has passed since the deadline cannot be extended. This deadline is not particularly important to WCAT and should be extend-able at any time.

Currently, you can only extend the event change deadline before it has passed. We would like to change this so that you can extend/remove this deadline even after it has passed.

gregorbg:
Fixes #11416

Essentially does two things:
1. Lift the restriction that `eventChangeDeadlineDate` can only be touched if it hasn't passed yet. All other date-related restrictions (for example, cannot set a deadline to an earlier date than the original deadline, even if the old and new deadline dates haven't passed yet) stay in place
2. Adds a validation that OTS cannot be enabled if Edit Events has already passed. This avoids scenarios where Delegates soft-edit to enable OTS, then go to the OTS page, then try to add someone, get a hard error from the Reg model validation, and have to go back to make another soft edit to `eventChangeDeadlineDate`.

### DIFF
--- a/app/models/competition.rb
+++ b/app/models/competition.rb
@@ -1199,8 +1199,8 @@ class Competition < ApplicationRecord
     self.on_the_spot_registration_changed? && self.on_the_spot_registration?
   end
 
-  validate :on_the_spot_registration_must_be_valid
-  private def on_the_spot_registration_must_be_valid
+  validate :enforce_edit_deadline_ots_consistency
+  private def enforce_edit_deadline_ots_consistency
     errors.add(:on_the_spot_registration, I18n.t('competitions.errors.on_the_spot_with_past_event_change_deadline')) if enabling_on_the_spot_registration? && event_change_deadline_date&.past?
   end
 
--- a/app/models/competition.rb
+++ b/app/models/competition.rb
@@ -1195,6 +1195,15 @@ class Competition < ApplicationRecord
     errors.add(:event_change_deadline_date, I18n.t('competitions.errors.event_change_deadline_after_end_date')) if event_change_deadline_date > end_date.to_datetime.end_of_day
   end
 
+  def enabling_on_the_spot_registration?
+    self.on_the_spot_registration_changed? && self.on_the_spot_registration?
+  end
+
+  validate :on_the_spot_registration_must_be_valid
+  private def on_the_spot_registration_must_be_valid
+    errors.add(:on_the_spot_registration, I18n.t('competitions.errors.on_the_spot_with_past_event_change_deadline')) if enabling_on_the_spot_registration? && event_change_deadline_date&.past?
+  end
+
   # Since Competition.events only includes saved events
   # this method is required to ensure that in any forms which
   # select events, unsaved events are still presented if
--- a/app/models/competition.rb
+++ b/app/models/competition.rb
@@ -2496,7 +2496,7 @@ class Competition < ApplicationRecord
         # These keys all represent timestamps. They may only be edited by non-admins if...
         #   - the original value (pre-edit) has not yet passed
         #   - the new value is in the future (extending deadlines is allowed, shortening them is not)
-        if %w[registration.closingDateTime registration.waitingListDeadlineDate registration.eventChangeDeadlineDate].include?(joined_key)
+        if %w[registration.closingDateTime registration.eventChangeDeadlineDate].include?(joined_key)
           existing_value = current_state_form.dig(*prefixes, key)
 
           if existing_value.present?
--- a/app/models/competition.rb
+++ b/app/models/competition.rb
@@ -2505,7 +2505,9 @@ class Competition < ApplicationRecord
           # Complain if the existing timestamp lies in the past
           #   Note: Some timestamps are less strict than others (see https://github.com/thewca/worldcubeassociation.org/issues/11416)
           #   so we only enforce this validation on a subset of timestamps.
-          if previously_had_value && %w[registration.closingDateTime registration.waitingListDeadlineDate].include?(joined_key)
+          edits_forbidden_if_past = %w[registration.closingDateTime registration.waitingListDeadlineDate].include?(joined_key)
+
+          if previously_had_value && edits_forbidden_if_past
             existing_datetime = DateTime.parse(existing_value)
 
             raise WcaExceptions::BadApiParameter.new(I18n.t('competitions.errors.editing_deadline_already_passed', timestamp: existing_datetime), json_property: joined_key) if existing_datetime.past?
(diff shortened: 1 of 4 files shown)


app/models/competition.rb
```diff
@@ -1195,6 +1195,15 @@ def registration_edits_currently_permitted?
     errors.add(:event_change_deadline_date, I18n.t('competitions.errors.event_change_deadline_after_end_date')) if event_change_deadline_date > end_date.to_datetime.end_of_day
   end
 
+  def enabling_on_the_spot_registration?
+    self.on_the_spot_registration_changed? && self.on_the_spot_registration?
+  end
+
+  validate :on_the_spot_registration_must_be_valid
+  private def on_the_spot_registration_must_be_valid
+    errors.add(:on_the_spot_registration, I18n.t('competitions.errors.on_the_spot_with_past_event_change_deadline')) if enabling_on_the_spot_registration? && event_change_deadline_date&.past?
```
dunkOnIT, line 1204: The safe accessor for `event_change_deadline_date` raises some question marks for me. The implication is that we won't raise an error if event_change_deadline_date is nil - which seems reasonable, I just want to dig a bit deeper.
 
- Can `event_change_deadline_date` ever be nil? 
  -  Yes - whether or not event edits are enabled for the comp
- If so, does that have any implication for OTS behaviour? 
  - Are/should OTS registrations be possible if event edits are disabled?

app/models/competition.rb
```diff
@@ -1195,6 +1195,15 @@ def registration_edits_currently_permitted?
     errors.add(:event_change_deadline_date, I18n.t('competitions.errors.event_change_deadline_after_end_date')) if event_change_deadline_date > end_date.to_datetime.end_of_day
   end
 
+  def enabling_on_the_spot_registration?
+    self.on_the_spot_registration_changed? && self.on_the_spot_registration?
+  end
+
+  validate :on_the_spot_registration_must_be_valid
```
dunkOnIT, line 1202: I think we can get a better name here - what we're really checking is the consistency of the ots date with the event change deadline, right? Perhaps `enforce_edit_deadline_ots_consistency` or something similar? 

(I'm not a huge fan of acronyms in variable names but the tradeoff here feels worth it)

app/models/competition.rb
```diff
@@ -2490,23 +2499,33 @@ def set_form_data(form_data, current_user)
...
-          if existing_value.present?
-            existing_datetime = DateTime.parse(existing_value).utc
+          previously_had_value = existing_value.present?
+          will_have_value = value.present?
 
-            raise WcaExceptions::BadApiParameter.new(I18n.t('competitions.errors.editing_deadline_already_passed', timestamp: existing_datetime), json_property: joined_key) unless existing_datetime >= DateTime.now.utc
+          # Complain if the existing timestamp lies in the past
+          #   Note: Some timestamps are less strict than others (see https://github.com/thewca/worldcubeassociation.org/issues/11416)
+          #   so we only enforce this validation on a subset of timestamps.
+          if previously_had_value && %w[registration.closingDateTime registration.waitingListDeadlineDate].include?(joined_key)
```
dunkOnIT, line 2508: Why not split this list out to a variable/constant declaration? That would mean we can explicitly call it "can be edited if past" or whatever, meaning we don't need to infer the significance of this list

spec/controllers/competitions_controller_spec.rb
dunkOnIT, line 1: I think these tests should be moved into model tests or request specs - [this post](https://www.codewithjason.com/use-controller-request-specs-rails-dont/) lays out why. In summary:
- controller specs are deprecated, and the official recommendation of the Rails and RSpec teams are to use request specs instead (because it provides a more realistic test of an actual REST request) 
- request specs are expensive, and should only be used for general happy/failure paths
- granular behaviour should be defined in model specs - in general, we should be able to define nearly all of our tests there

app/models/competition.rb
```diff
@@ -2490,23 +2499,33 @@ def set_form_data(form_data, current_user)
...
-          if value.present?
-            new_datetime = DateTime.parse(value).utc
+          if previously_had_value && will_have_value
+            new_datetime = DateTime.parse(value)
+            existing_datetime = DateTime.parse(existing_value)
+
+            new_before_existing = new_datetime < existing_datetime
 
-            raise WcaExceptions::BadApiParameter.new(I18n.t('competitions.errors.edited_deadline_not_in_future', new_timestamp: new_datetime), json_property: joined_key) unless new_datetime > DateTime.now.utc
+            # Complain if the new value lies before the old value (the user is trying to move something into the past)
```
dunkOnIT, line 2527: This comment implies that the validation is trying to prevent deadlines being moved to the past. However, the logic seems to trigger if the new deadline is before the old deadline - irrespective of whether it is before/after the current date.

What is the desired behavior here?

dunkOnIT (changes requested):

app/models/competition.rb
gregorbg: There are more checks for the Edit Events deadline directly above (you might have to expand the diff to see them). I was considering merging them into one, but I ultimately decided against it because the two validations write their error properties (their validation "results") into two different fields.

app/models/competition.rb
gregorbg: I was trying to be consistent with the other validation names and patterns directly above, but to be fair I'm not a big fan of those names either.

Would you rather sacrifice consistency for a better name here locally, or sacrifice the PR scope by renaming the other methods as well?

spec/controllers/competitions_controller_spec.rb
gregorbg: Be my guest to move them in a separate PR :) Right now it's about establishing "functionality tests" in line with what we already have.

app/models/competition.rb
gregorbg: Okay, I can see how my comment was misleading. Will change that :)

app/models/competition.rb
zekemackay06: Yes - `event_change_deadline_date` can be nil. This means that competitors can change registered events at any time. For OTS registration to be enabled, this deadline must either be during the competition or removed entirely. Most of the time, to enable OTS registration, delegates remove the deadline altogether.

app/models/competition.rb
dunkOnIT: I'm all for sacrificing consistency, and will note down the renames for a potential future/downtime PR

spec/controllers/competitions_controller_spec.rb
dunkOnIT: Cool, noted for follow-up

dunkOnIT (approved):

PR #11438, opened 2025-04-29
Fixed history button to show 'history'

pranav-027:
The history button on all delegates page redirects to this URL: 
https://www.worldcubeassociation.org/competitions?years=all&delegate=51390&show_admin_details=yes

Fixed so it will redirect to past comps: 
https://www.worldcubeassociation.org/competitions?years=all&delegate=51390&show_admin_details=yes&state=past

### DIFF
--- a/app/webpacker/components/Delegates/DelegatesTable.jsx
+++ b/app/webpacker/components/Delegates/DelegatesTable.jsx
@@ -89,6 +89,7 @@ export default function DelegatesTable({
                     display: 'admin',
                     years: 'all',
                     delegate: delegate.user.id,
+                    state: 'past',
                   })}
                   >
                     {I18n.t('delegates_page.table.history')}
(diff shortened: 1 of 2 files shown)


dunkOnIT:
What is the "History" button? Could you share a screenshot?

pranav-027:
@dunkOnIT 

image

you can find this on https://www.worldcubeassociation.org/delegates#all

dunkOnIT:
Hmm, I'm still struggling to reach this screen. My steps to reproduce are:
- be logged in as 2019HOBB02 (ie me)
- http://localhost:3000/delegates#all

That just yields this screen: 
image

I don't see any further potential for interaction on the page. I've tried looking at the code and it seems like `isAllRegions` needs to be true, but that seems to depend on the value of a hash whose source isn't clear to me.

Let me know what I'm doing wrong!

pranav-027:
@dunkOnIT Almost there Just scroll to the bottom of the page :)

gregorbg (approved):

PR #11440, opened 2025-04-29
Clean up routes from old "attendee ID" concept

gregorbg:
Replaces all registration routes that were still using (competitionId, userId) tuples to work with our internal registration IDs instead.

A redirect for the old format has been added if Delegates click on things in emails, which we unfortunately cannot retroactively manipulate.

### DIFF
--- a/app/views/registrations/edit.html.erb
+++ b/app/views/registrations/edit.html.erb
@@ -1,5 +1,6 @@
 <% provide(:title, t('registrations.edit_registration.title', person: @user.name, comp: @competition.name)) %>
 <%= render layout: 'nav' do %>
-  <%= react_component('RegistrationsV2/RegistrationEdit', { competitionInfo: @competition.to_competition_info ,
+  <%= react_component('RegistrationsV2/RegistrationEdit', { registrationId: @registration.id,
+                                                            competitionInfo: @competition.to_competition_info ,
                                                             user: @user }) %>
 <% end %>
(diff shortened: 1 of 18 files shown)


app/views/registrations/edit.html.erb
```diff
@@ -1,5 +1,6 @@
-<% provide(:title, t('registrations.edit_registration.title', person: @user.person&.name || @user.name, comp: @competition.name)) %>
```
gregorbg, line 1: Note for review: Whenever we link a `Person` back to a `User` (commonly known as "claim your WCA ID"), we also backfill the Person's name and DOB into the User.

In other words, claimed WCA IDs are always in sync between user and person automatically.

dunkOnIT: Makes sense, but do we want to introduce an unrelated change here in this PR?

dunkOnIT (dismissed):

app/views/registrations/edit.html.erb
gregorbg: It's not really an unrelated change, more of a fix for something that was clearly redundant before.
Also, it affects only the page title, i.e. the thing that gets displayed on your browser tab and the window frame, so if for some obscure reason this change messes something up, the impact is minimal.

FinnIckler (approved):
tested pages and the redirect

app/webpacker/components/RegistrationsV2/RegistrationEdit/RegistrationEditor.jsx
```diff
@@ -312,7 +312,7 @@ export default function RegistrationEditor({ competitor, competitionInfo }) {
           {(registration.payment.payment_statuses.includes('succeeded') || registration.payment.payment_statuses.includes('refund')) && (
             <Refunds
               competitionId={competitionInfo.id}
```
dunkOnIT, line 314: Do we still need to be passing competitionId here?

gregorbg: Yes, it's needed for the mutation that is actually issuing the refund

PR #11441, opened 2025-04-29
Fix incorrect report posted date calculation in the competitions admin view

maxidragon:
Despite the fixes in #10783 and #10962, the report posted date is still sometimes calculated incorrectly. This PR aims to finally resolve the issue

### DIFF
--- a/app/webpacker/lib/utils/competition-table.js
+++ b/app/webpacker/lib/utils/competition-table.js
@@ -92,9 +92,7 @@ export function numberOfDaysAfter(competition, refDate) {
   const parsedEndDate = parseDateString(competition.end_date).startOf('day');
   const parsedRefDate = DateTime.fromISO(refDate, { zone: 'utc' }).startOf('day');
 
-  const numberOfDays = parsedRefDate.diff(parsedEndDate, 'days').days;
-
-  return Math.abs(numberOfDays);
+  return parsedRefDate.diff(parsedEndDate, 'days').days;
 }
 
 export function timeDifferenceAfter(competition, refDate) {
--- a/app/webpacker/lib/utils/competition-table.js
+++ b/app/webpacker/lib/utils/competition-table.js
@@ -94,7 +94,7 @@ export function numberOfDaysAfter(competition, refDate) {
 
   const numberOfDays = parsedRefDate.diff(parsedEndDate, 'days').days;
 
-  return Math.max(0, numberOfDays);
+  return Math.abs(numberOfDays);
 }
 
 export function timeDifferenceAfter(competition, refDate) {
--- a/app/webpacker/lib/utils/competition-table.js
+++ b/app/webpacker/lib/utils/competition-table.js
@@ -89,13 +89,12 @@ export function timeDifferenceBefore(competition, refDate) {
 }
 
 export function numberOfDaysAfter(competition, refDate) {
-  const parsedStartDate = parseDateString(competition.end_date).endOf('day');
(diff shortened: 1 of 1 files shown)


app/webpacker/lib/utils/competition-table.js
```diff
@@ -89,13 +89,12 @@ export function timeDifferenceBefore(competition, refDate) {
 }
 
 export function numberOfDaysAfter(competition, refDate) {
-  const parsedStartDate = parseDateString(competition.end_date).endOf('day');
-  const parsedRefDate = DateTime.fromISO(refDate);
+  const parsedEndDate = parseDateString(competition.end_date).startOf('day');
+  const parsedRefDate = DateTime.fromISO(refDate, { zone: 'utc' }).startOf('day');
```
gregorbg, line 93: `fromISO` implies that `refDate` is in ISO8601, which should already contain time zone information. Why do you need to enforce UTC here nonetheless?

app/webpacker/lib/utils/competition-table.js
```diff
@@ -89,13 +89,12 @@ export function timeDifferenceBefore(competition, refDate) {
...
 
-  const numberOfDays = parsedStartDate.diff(parsedRefDate, 'days').days;
+  const numberOfDays = parsedRefDate.diff(parsedEndDate, 'days').days;
 
-  // Floor is used here because we want to show 0 days after the competition if it's the same day
-  return Math.floor(Math.abs(numberOfDays));
+  return Math.max(0, numberOfDays);
```
gregorbg, line 97: Why `max` instead of something like `abs`? I read this as "make sure it's not negative", so what are the circumstances that can even make a negative date appear?

app/webpacker/lib/utils/competition-table.js
maxidragon: I'm not sure, but without enforcing UTC the following values are displayed (results were submitted on the same day and report posted 7 days after):
image

app/webpacker/lib/utils/competition-table.js
```diff
@@ -89,13 +89,12 @@ export function timeDifferenceBefore(competition, refDate) {
...
-  const parsedRefDate = DateTime.fromISO(refDate);
+  const parsedEndDate = parseDateString(competition.end_date).startOf('day');
+  const parsedRefDate = DateTime.fromISO(refDate, { zone: 'utc' }).startOf('day');
 
-  const numberOfDays = parsedStartDate.diff(parsedRefDate, 'days').days;
+  const numberOfDays = parsedRefDate.diff(parsedEndDate, 'days').days;
 
-  // Floor is used here because we want to show 0 days after the competition if it's the same day
-  return Math.floor(Math.abs(numberOfDays));
+  return Math.abs(numberOfDays);
```
gregorbg, line 97: Okay, now this looks cleaner in the code, but you still avoided answering my question: Why can negative values happen in the first place? I mean, we know that it's impossible to post results or submit a report before the competition is over, so `parsedEndDate.diff(parsedRefDate, 'days').days` _should_ be enough to guarantee that the result can never be negative. Why do you need this `abs` kind of safeguard at all?

gregorbg (changes requested):

app/webpacker/lib/utils/competition-table.js
maxidragon: Done

gregorbg (approved):

PR #11442, opened 2025-04-29
Refactor available refund route to general registration payment route

FinnIckler:
image
This is quite simple right now, but nested data in table is a design nightmare so I kept it like that for the sake of simplicity


config/routes.rb
```diff
@@ -99,6 +99,7 @@
     get 'registrations/psych-sheet' => 'registrations#psych_sheet', as: :psych_sheet
     get 'registrations/psych-sheet/:event_id' => 'registrations#psych_sheet_event', as: :psych_sheet_event
     resources :registrations, only: [:index, :update, :create, :edit, :destroy], shallow: true
+    get 'registrations/:registration_id/payment' => 'payment#registration_payments', as: :registration_payments
```
gregorbg, line 102: Nit 1: `/payments` with plural S in the URL string

gregorbg, line 102: Nit 2: You can probably move this under the `resources :registrations` block with a `do`:
```suggestion
    resources :registrations, only: [:index, :update, :create, :edit, :destroy], shallow: true do
      get 'payments' => 'payment#registration_payments', as: :registration_payments
    end
```

gregorbg: Although from the top of my head I wouldn't know how this affects the naming of the route helper. You will have to try that out locally.

app/webpacker/components/RegistrationsV2/RegistrationEdit/Payments.jsx
```diff
@@ -1,26 +1,27 @@
...
+    isLoading: paymentsLoading,
     refetch,
   } = useQuery({
-    queryKey: ['refunds', competitionId, userId],
-    queryFn: () => getAvailableRefunds(competitionId, userId),
+    queryKey: ['payments', registrationId],
+    queryFn: () => getRegistrationPayments(competitionId, registrationId),
     refetchOnWindowFocus: false,
     refetchOnReconnect: false,
     staleTime: Infinity,
```
gregorbg, line 27: Technically unrelated but very, very easy fix: Aren't these options (and the fourth one below, which GitHub doesn't let me comment on because it arbitrarily rules that it's too far away from the changed lines, grr....) part of the default `WCAQueryClientProvider` setup these days?

gregorbg (changes requested):
I think we should merge https://github.com/thewca/worldcubeassociation.org/pull/11440 first (which Duncan probably?? approved by accident, we were talking about another PR privately on Slack, so please don't rush ahead with merging that one preemptively...) and that will lead to a trivial number of merge conflicts (for example, including the `id` in the registration V2 JSON).

Also, regarding the table layout: Have you tried adding a second row if there are refunding payments, and give that a `colSpan` to cover the whole row? You could easily add a list in that one (or even consider giving it more columns to show for example who the refund was issued by).
I won't absolutely insist on this feature, but it would be nice to have a bit more details on the refunds. See if you can find something that works, and if it's crazy hard or always looks :hankey: then report back here and we will merge as-is and think about refund details later :tm:

config/routes.rb
FinnIckler: well, then we should move all those routes there right?

FinnIckler:
image
what about this

app/webpacker/lib/requests/routes.js.erb
```diff
@@ -97,7 +97,7 @@ export const adminFixResultsUrl = (personId, competition_id = undefined, event_i
 
 export const paymentFinishUrl = (competitionId, paymentIntegration) => `<%= CGI.unescape(Rails.application.routes.url_helpers.registration_payment_completion_url("${competitionId}", "${paymentIntegration}", host: EnvConfig.ROOT_URL)) %>`
 
-export const availableRefundsUrl = (competitionId, userId) => `<%= CGI.unescape(Rails.application.routes.url_helpers.refunds_path) %>?attendee_id=${competitionId}-${userId}`
+export const registrationPaymentsUrl = (competitionId, registrationId) => `<%= CGI.unescape(Rails.application.routes.url_helpers.competition_registration_payments_path("${competitionId}", "${registrationId}")) %>`
```
gregorbg, line 100: If I'm reading the `routes.rb` correctly, then `competitionId` is not even needed as a route parameter anymore

FinnIckler: it is, it's under the   `resources :competitions do`

gregorbg: Oh wait, that trick only worked for `resources :registrations` because it's marked as `shallow: true`. In that case, I think you should really consider my other comment above about nesting the route, because otherwise supplying `competitionId` *and* `registrationId` (which implicitly also identifies the competition) is kinda redundant. Nesting would make this route become simpler and profit from the `shallow` routing as well

FinnIckler: ok, I'll nest it

config/routes.rb
```diff
@@ -98,7 +98,9 @@
     post 'registrations/add' => 'registrations#do_add', as: :registrations_do_add
     get 'registrations/psych-sheet' => 'registrations#psych_sheet', as: :psych_sheet
     get 'registrations/psych-sheet/:event_id' => 'registrations#psych_sheet_event', as: :psych_sheet_event
-    resources :registrations, only: [:index, :update, :create, :edit, :destroy], shallow: true
+    resources :registrations, only: [:index, :update, :create, :edit, :destroy], shallow: true do
+      get ':registration_id/payments' => 'payment#registration_payments', as: :payments
```
gregorbg, line 102: Sorry for picking so much on your `routes.rb`, but the fact that it's nested under `resources :registrations` adds a `:registration_id` route param automagically. `get 'payments'` is literally enough to achieve what you want.

gregorbg (changes requested):

gregorbg (approved):
Cool, fine by me now! Do you have any strong preference whether we should merge this one or https://github.com/thewca/worldcubeassociation.org/pull/11440 first?

FinnIckler:
I don't have any preference, but I can review the other PR in like an hour

PR #11444, opened 2025-04-30
Expose refunds asynchronously instead of hard-redirecting

gregorbg:
The React refunds panel was still using the old Rails-esque endpoint that redirected the user and communicated using flash messages. The aim of this PR is to communicate the result directly via JSON asynchronously and let the frontend handle displaying the messages and updating the values.

This has two practical consequences:
1. The refunds endpoint upon success now exposes the payment it just refunded. This allows React to replace the affected rows in-place, instead of triggering a hard reload
2. `RegistrationPayment` entities now get their own custom "V2 JSON" format. Consider it an early, preliminary API spec. Since this only used internally, it will be easy to change in the future whenever required.

### DIFF
--- a/app/models/registration_payment.rb
+++ b/app/models/registration_payment.rb
@@ -29,6 +29,26 @@ class RegistrationPayment < ApplicationRecord
     end
   end
 
+  def to_v2_json
+    payment_provider = CompetitionPaymentIntegration::INTEGRATION_RECORD_TYPES.invert[self.receipt_type]
+
+    available_amount = self.amount_available_for_refund
+    full_amount_ruby = self.amount_lowest_denomination
+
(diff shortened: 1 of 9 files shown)


app/models/registration_payment.rb
```diff
@@ -29,6 +29,37 @@ def payment_status
     end
   end
 
+  def to_v2_json(helpers, refunds: false)
+    payment_provider = CompetitionPaymentIntegration::INTEGRATION_RECORD_TYPES.invert[self.receipt_type]
```
FinnIckler, line 33: why not use key(self.receipt_type) instead of invert?

gregorbg: (a) because `invert` was used before (see diff) and (b) because I only knew `key?` existed but I didn't know about `key` :D

FinnIckler (approved):

PR #11446, opened 2025-04-30
Payment overview in Paymentstep

FinnIckler:
image
Currently I am basically showing nothing, because I need the data structure from #11444, so this is mostly rearranging and white space changes


FinnIckler:
Partial Refund Payment Overview:
image
Uncollapsed Accordion:
image
Full Refund:
image
Full Paid:
image

FinnIckler:
any good way to find out why the payment panel is not rendering the tests? I looked at the screenshots and it looks like this 
screenshot_2025-05-01-11-55-22 896

dunkOnIT:
I really like it! A couple of nits: 

- [Fig 1] Often websites have an arrow indicating that an accordion can be expanded - would this be possible to add? eg, next to "Minor" in [Fig 2]
- Once a user is refunded/partially refunded but not accepted, the default behaviour is for them to see the Payment Information panel, not the registration overview panel. I think this might lead to confusion for users, which we could improve by
  - always defaulting to the Approval panel, even if user is partially refunded but not yet accepted
  - potentially, adding a "View Registration" button at the bottom of Payment Information, further prompting the user to go to the Approval panel (probably only necessary if we don't want to default to the approval panel?)
- Potentially we could show more information under net payment, by showing the total paid less total refunds. eg - "Net Payment:\n$5.00 (United States Dollar) | $10 initial payment - $5 total refunded"

Fig 1:
image

Fig 2: 
image

FinnIckler:
I added the button, but I am not sure if we need to do the other thing. Not sure in which circumstance a user will be refunded, but not approved. 
If a user is refunded it usually means their registration is cancelled/rejected, or they have some kind of payment wave/reduced privilege which would only occur once they are approved.

dunkOnIT:
There are also the cases of partial refunds in the case of RO membership used by several French/Scandanavian countries, which is the main concern I had in mind. This seems like the most likely case where someone would both be refunded, and still wanting to view their registration, so want to make sure it is covered.

dunkOnIT (approved):
Happy with it from a UX perspective! @gregorbg shout if you have any code concerns, otherwise feel free to merge

gregorbg (approved):

PR #11448, opened 2025-04-30
Fetch registration history from separate endpoint

gregorbg:
In the spirit of https://github.com/thewca/worldcubeassociation.org/pull/11442, let's also create a separate endpoint for the history!

(This also saves one silly refetch from the refund endpoint where refunding would effectively reload the whole page just so the registration history could be correct)

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -99,7 +99,7 @@ Rails.application.routes.draw do
     get 'registrations/psych-sheet/:event_id' => 'registrations#psych_sheet_event', as: :psych_sheet_event
     resources :registrations, only: %i[index update create edit destroy], shallow: true do
       get 'payments' => 'payment#registration_payments', as: :payments
-      get 'history' => 'registrations#registration_history', on: :member, as: :history
+      resource :history, only: [:show], controller: :registration_history
     end
     get 'edit/registrations' => 'registrations#edit_registrations'
     get 'register' => 'registrations#register'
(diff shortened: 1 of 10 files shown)


config/routes.rb
```diff
@@ -99,6 +99,7 @@
     get 'registrations/psych-sheet/:event_id' => 'registrations#psych_sheet_event', as: :psych_sheet_event
     resources :registrations, only: [:index, :update, :create, :edit, :destroy], shallow: true do
       get 'payments' => 'payment#registration_payments', as: :payments
+      get 'history' => 'registrations#registration_history', on: :member, as: :history
```
FinnIckler, line 102: what exactly does on :member do? I looked it up, but I don't immediately get why it's necessary here

gregorbg: Simply put, it distinguishes between calling the route parameter `/registrations/:id` and `/registrations/:registration_id`. And the existing controller structure relies a lot on `:id`, so that's what I had to go with.

Unfortunately, this also prompts Rails to change the name of the route handler into some obscure prefix thing, I'm currently trying to find a solution for that,

gregorbg: That won't hurt. But other places throughout the controller which perform pre-checks will explode if it's called `registration_id` instead of `id`

FinnIckler:
I'm fine with the code, but I think we should discuss still refetching the history on every registration change

gregorbg:
> I'm fine with the code, but I think we should discuss still refetching the history on every registration change

I took care of the automated refetching in https://github.com/thewca/worldcubeassociation.org/pull/11448/commits/6bc635e81824a0d3f046460ea96dbe1aec767801

FinnIckler:
Now the manual refresh button is redundant no?

gregorbg:
Hm, depends. Async payments _could_ still come through while the page is open, although I agree that this is a rare use-case.

FinnIckler (approved):

PR #11450, opened 2025-04-30
Clean up Registration V2 JSON payment section

gregorbg:
We weren't really using the `payment_statuses` anymore, except for the most recent one. So I wrote a separate method for that. Unfortunately, the result is not exactly pretty if you want optimized SQL queries, but this is a price that I'm willing to pay since Rails doesn't give us any better APIs.

Also, I decided to rename some fields while I'm at it and remove a dependency on our backend money formatter.


FinnIckler (approved):
Tested and working

PR #11452, opened 2025-04-30
Refactor Registration API routes in V1 API

gregorbg:
I thought "hey some cleanup won't do no harm"... But then the scope of this PR unfortunately exploded :sweat_smile: 

The general idea is to continue in the spirit of https://github.com/thewca/worldcubeassociation.org/pull/11440 and query registrations by ID wherever possible. Howver, this has wide-ranging implications for the wiring inside the controller, as well as for the registration step panel polling/editing.

Leaving this as a draft just to see what CI does, and then I also need to figure out what to do with the pre-loaded registration and whether passing along the whole beast is really the best idea...


app/webpacker/components/RegistrationsV2/Register/index.jsx
```diff
@@ -20,6 +20,7 @@ const editableRegistrationStates = ['accepted', 'pending', 'waiting_list'];
 export default function Index({
   competitionInfo,
   userInfo,
+  registrationId = null,
```
FinnIckler, line 23: why is there a default value?

config/routes.rb
```diff
@@ -345,14 +343,19 @@
...
-        post '/register', to: 'registrations#create'
-        patch '/register', to: 'registrations#update'
-        patch '/bulk_update', to: 'registrations#bulk_update'
-        get '/:competition_id', to: 'registrations#list'
-        get '/:competition_id/admin', to: 'registrations#list_admin', as: :list_admin
-        get '/:competition_id/payment', to: 'registrations#payment_ticket', as: :payment_ticket
+      resources :competitions, only: [] do
+        resources :registrations, only: %i[index show create update], shallow: true do
+          member do
+            get 'payments', to: 'registrations#registration_payments'
```
FinnIckler, line 349: moving this here now means we want this part of the public API, which I am not sure of

gregorbg: Eventually everything has to be part of v1 so that it can be served to Next. Is there a better place for this API, especially considering our long-term infrastructure plan?

FinnIckler: Good point with Next, I think it would make more sense behind an internal API. There will be dozens of routes that we want to expose in a different internal scope in the future when we are moving more and more pages to next. Even in registrations there are routes like the payment completion URL that don't live in v1 and I would argue shouldn't.

spec/requests/api_registrations_spec.rb
```diff
@@ -16,20 +16,19 @@
...
-        post api_v1_registrations_register_path, params: registration_request, headers: headers
+        post api_v1_competition_registrations_path(competition_id: competition.id), params: registration_request, headers: headers
         expect(response.body).to eq({ status: "accepted", message: "Started Registration Process" }.to_json)
         expect(response).to have_http_status(:accepted)
       end
 
       it 'enqueues an AddRegistrationJob' do
         expect {
-          post api_v1_registrations_register_path, params: registration_request, headers: headers
+          post api_v1_competition_registrations_path(competition_id: competition.id), params: registration_request, headers: headers
```
dunkOnIT, line 26: Bit of a nit, but can we not just pass `competition` here, and in other instances? I've tested locally and it works

dunkOnIT:
```suggestion
post api_v1_competition_registrations_path(competition), params: registration_request, headers: headers
```

spec/requests/api_registrations_spec.rb
```diff
@@ -428,20 +427,6 @@
       expect(email.subject).to eq(I18n.t('registrations.mailer.new.mail_subject', comp_name: registration.competition.name))
     end
 
-    it 'raises error if registration doesnt exist' do
```
dunkOnIT, line 431: Looks like we're just deleting this test? Is it being replaced by something?

app/webpacker/components/RegistrationsV2/Register/index.jsx
gregorbg: That's true, but it relies on JavaScript internals. I still think that explicitly marking the `null` default value makes it clear to someone reading the method signature that "competitionInfo and userInfo must be there, registrationId can be dropped if you want".

gregorbg: Also, TypeScript.

config/routes.rb
gregorbg: Then what do we do with this route in this PR?

spec/requests/api_registrations_spec.rb
gregorbg: Oh, I had intermittently deleted this because I was hell-bent on querying registrations only by numeric ID. In that case, this test would have been useless because you cannot even query a non-existing registration if you don't have a numeric ID to query for.

But now I already understood that the "competitionId+userId" tuple querying is unavoidable, so I'll happily reinstate this test.

config/routes.rb
FinnIckler: keep it under `resources :registrations, only: %i[index update create edit destroy], shallow: true` ?

app/webpacker/components/RegistrationsV2/Register/index.jsx
FinnIckler: okay, but then add a comment that you explicitly mark this as optional and that it should be fixed when moving to typescript

spec/requests/api_registrations_spec.rb
gregorbg: Nevermind: I just realized that this test was trying to do a `PATCH` against a non-existing (competitionId, userId)` tuple. That won't work anymore because the `PATCH` route now definitely needs a `registrationId`. We could send a zero and see that this blows up? Not sure whether that helps though

config/routes.rb
gregorbg: That does not sound like a long-term solution :/

app/webpacker/components/RegistrationsV2/Register/index.jsx
gregorbg: I have decided to remove the default value in all places that are React props, but keep it in the routing function call

spec/requests/api_registrations_spec.rb
dunkOnIT: Yeah I think sending a zero or any other reg value we know doesn't exist and asserting that we get a 404 back is the way to go here

FinnIckler (approved):

PR #11453, opened 2025-05-01
Use form state for registration panel components

gregorbg:
Reusing the same kind of hooks and reducer that the competition form currently uses, with a little extra features added.

The goal here is to make the code significantly easier to read, because the entire registration now has a "global state" that can be manipulated by hooks, instead of using local `useState` variables everyhwere. Most importantly, this gets rid of several `useEffect` hooks that were trying to keep the outer context and the local component state in sync.

I think `RegistrationEditor.jsx` is a good starting point because the changes there are the most "pure". Looking at that file will give you a pretty good idea of the overall PR intention.


gregorbg:
I think I resolved all merge conflicts, but it was surprisingly non-trivial. Can you do a second round of (brief) testing, please?

Issue #11445, opened 2025-04-30
Add WCIF `registrantId` to CSV export


dunkOnIT:
@pranav-027 would you be interested in working on this? The user who requested `id` in a previous PR was actually talking about the `registrantId`, not `userId`. 

Note that this is the `registrantId` as defined in the WCIF (unique in the scope of the competition, not the db) - **not** the `registrationId` (unique in the scope of the database)

pranav-027:
Sure you can assign it to me! Also we are not removing userId right? Just adding this registrationId

dunkOnIT:
Thanks! And correct, not removing `userId` - just adding `registrantId`

dunkOnIT:
> Sidenote: Why do you want this to be included in the export in the first place?

Requested by Anto Cam, with the reasoning of wanting to use it for building scorecards from the CSV export

PR #11454, opened 2025-05-01
add wcif registrant_id in CSV

pranav-027:
fixes: https://github.com/thewca/worldcubeassociation.org/issues/11445

Not sure if it's the best implementation. Happy to hear some suggestions!


app/models/competition.rb
```diff
@@ -2923,4 +2923,11 @@ def attempt_auto_close!
     threshold_reached = fully_paid_registrations_count >= auto_close_threshold && auto_close_threshold.positive?
     threshold_reached && update(closing_full_registration: true, registration_close: Time.now)
   end
+
+  def wcif_registrant_id_for(registration)
+    registrations.wcif_ordered.each.with_index(1) do |r, idx|
+      return idx if r.id == registration.id
+    end
+    nil
```
gregorbg, line 2931: I think Ruby has a stdlib method for that
```suggestion
    registrations.wcif_ordered.find_index { it.id == registration.id }
```

gregorbg (changes requested):
Thanks for coming up with an initiative! Unfortunately, handling it like this is sinfully expensive because it requires reading all registrations (normally from the database, but thanks to some clever Rails-internal caching that you get for free, most calls come from memory) for every single individual registration that you're looping over.

I'd much rather take this PR (and the corresponding issue) as an incentive to discuss more "clever" ways of computing and storing registration IDs. Persist something :tm: to the database the moment that a registration is created, and use that as a simple direct field access, instead of computing crucial identifiers on-the-fly by some `order` condition.

app/models/competition.rb
pranav-027: I asked GPT for suggestions to handle the bulk registrations. and it came up with this

pranav-027:
@gregorbg, this is done. 

image

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
```diff
@@ -19,7 +19,7 @@ import {
 import { useConfirm } from '../../../lib/providers/ConfirmProvider';
 
 function V3csvExport(selected, registrations, competition) {
-  let csvContent = `Status,Name,Country,WCA ID,Birth Date,Gender,${competition.event_ids.join(',')},Email,Guests,IP,Registration Date Time (UTC),Payment Date Time(UTC),User Id,Registration Status\n`;
+  let csvContent = `Status,Name,Country,WCA ID,Birth Date,Gender,${competition.event_ids.join(',')},Email,Guests,IP,Registration Date Time (UTC),Payment Date Time(UTC),User Id,Registration Status,Registration Id\n`;
```
maxidragon, line 22:
```suggestion
  let csvContent = `Status,Name,Country,WCA ID,Birth Date,Gender,${competition.event_ids.join(',')},Email,Guests,IP,Registration Date Time (UTC),Payment Date Time(UTC),User Id,Registration Status,Registrant Id\n`;
```

gregorbg (approved):

PR #11455, opened 2025-05-01
Bye Bye HEIC in Delegate Reports

gregorbg:
Seems that `image/*` is still too generous for that one company that doesn't release its file format specifications in the year 2025


pranav-027:
:(

FinnIckler (approved):

PR #11456, opened 2025-05-01
Allow banning a user if their registration is currently rejected

lnzainn:
At the moment, if the user has registered for a competition and their registration has been rejected, then the validation will fail and won't let you ban that user. In addition to cancelled registrations, this PR lets you ban someone whose registration is rejected. 

I tested it locally and it works as expected. 

cc @thewca/wic


app/controllers/api/v0/user_roles_controller.rb
```diff
@@ -76,7 +76,7 @@ def create
       user = User.find(user_id)
       ban_reason = params[:banReason]
       scope = params[:scope]
-      upcoming_comps_for_user = user.competitions_registered_for.not_over.merge(Registration.not_cancelled)
+      upcoming_comps_for_user = user.competitions_registered_for.not_over.joins(:registrations).where.not(registrations: { competing_status: %w[rejected cancelled] }).distinct
```
gregorbg, line 79: Nit 1: You don't need the `joins(:registrations)` association. It is implicitly included in the way that `competitions_registered_for` is defined.

gregorbg: Nit 2: We had to apply a very similar fix (from just "cancelled" to "cancelled or rejected") recently in https://github.com/thewca/worldcubeassociation.org/commit/39723133d836bba54743ffd1374d687083b40601. The snippet there is almost exactly identical to what you're doing here, probably it's worth merging these two approaches into one common method.

(But if you're not comfortable enough with Ruby then let me know and we can do that cleanup in a later PR)

gregorbg (changes requested):
Thanks, some small cosmetic change requests but otherwise all fine

gregorbg (approved):

PR #11459, opened 2025-05-02
Change `admin_comment` to `organizer_comment`

dunkOnIT:
Follow-up PR to #11408, Gregor and I discussed in slack that 11408 should establish consistency in the naming of this field. My suggestion is that we be more consistent in our usage of `admin`, to differentiate between comment made by a _website_ admin and an _administrative comment_

I think that in the context of calling it the "Reg Admin" page, the contraction of "administration" -> "admin" is more obvious. Just hearing "admin comment", it's easier to think of it as a contraction of "administrator", which can cause confusion.

### DIFF
--- a/app/controllers/api/v1/registrations/registrations_controller.rb
+++ b/app/controllers/api/v1/registrations/registrations_controller.rb
@@ -45,7 +45,7 @@ class Api::V1::Registrations::RegistrationsController < Api::V1::ApiController
   def create
     # Currently we only have one lane
     if params[:competing]
-      competing_params = params.permit(:guests, competing: [:status, :comment, { event_ids: [] }, :admin_comment])
+      competing_params = params.permit(:guests, competing: [:status, :comment, { event_ids: [] }, :organizer_comment])
 
       user_id = registration_params['user_id']
       competition_id = registration_params['competition_id']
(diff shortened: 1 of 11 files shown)


dunkOnIT:
Blocked: Need feedback on how to change keys in i18n files

gregorbg:
You're actually introducing two new inconsistencies here:
1. The `comment` field is stored as plural `comments` in the database, but you're storing `admin_comment` as singular in the database
2. The WCIF specifies it as `administrativeComment`, so maybe you could opt for `administrative_comment` as a convention here as well?

Regarding I18N files, just change the keys in all YAML files that have it translated. If there are also *content* changes, then only change the content (and key) in the English file, and let the automated "There are new translations" workflow do its thing.
In other words: If there are purely *key* changes, you can manually apply them in all relevant YAML files. As soon as there is string content that changes in the English base, then all other YAML files don't need to be touched (because they will need to be touched by human translators anyways) regardless of whether there are key changes or not.

dunkOnIT:
> The comment field is stored as plural comments in the database, but you're storing admin_comment as singular in the database

It sounds like there also inconsistency with the plurality in WCIF - which do we prefer? Personally, I'd rather change `comments` to be singular, at least to me the plural implies _multiple comment objects_ which is a bit confusing - especially with how Rails conventions around plurals work

> The WCIF specifies it as administrativeComment, so maybe you could opt for administrative_comment as a convention here as well?

Sounds good! I'll just wait for clarity on the first item before pushing changes

Issue #11462, opened 2025-05-02
Record history by Country broken


FinnIckler:
Just noticed this. Error in new relic is Mysql2::Error: Unknown column 'results.country_id' in 'where clause' so this is renaming related?

dunkOnIT:
Working on this

PR #11463, opened 2025-05-03
Record history by region fix

pranav-027:
fixes: https://github.com/thewca/worldcubeassociation.org/issues/11462


pranav-027:
Changes will be done with tests in https://github.com/thewca/worldcubeassociation.org/pull/11464

PR #11464, opened 2025-05-03
Fix record history by region

### LINKED ISSUE #11462: Record history by Country broken

dunkOnIT:
Fixes #11462 

Currently, [record history by region](https://www.worldcubeassociation.org/results/records?region=_Africa&show=history) is broken in prod. I replicated this issue locally through tests, and fixed it by correcting what appears to be inconsistent table aliases in the SQL queries.

I have added parametrized tests to cover the basic cases that were failing - there is more that can be done here though.

Follow-up PR: This section of the code is severely under-tested. Add more tests against the results API endpoints we use in production. Specifically:
- [ ] Add 3 and 4-parameter tests
- [ ] Add all cases of 2-parameter tests (currently we're only testing combinations of `show` and `region` - but we should ideally test all combinations of 
- [ ] We _possibly_ don't always return errors when an incorrect value is supplied (this may just be an error in my tests at the time of writing them, not an error with the code) - investigate this and raises tests/fix if necessary
- [ ] We don't raise errors if unexpected parameters are supplied (eg, 'event' instead of 'event_id') - perhaps not the end of the world, but leads to false-positives when testing

It is particularly important to make sure the test fails first, to ensure that you are actually triggering the behaviour you expect. (For example, in this PR, my test database was incompletely populated, and so wasn't triggering the region-specific code. I would not have picked this up if not for the fact that this use case was failing in production but not under my testing.)

### DIFF
--- a/app/controllers/results_controller.rb
+++ b/app/controllers/results_controller.rb
@@ -334,10 +334,10 @@ class ResultsController < ApplicationController
     @continent = Continent.c_find(params[:region])
     @country = Country.c_find(params[:region])
     if @continent.present?
-      @region_condition = "AND results.country_id IN (#{@continent.country_ids.map { |id| "'#{id}'" }.join(',')})"
+      @region_condition = "AND result.country_id IN (#{@continent.country_ids.map { |id| "'#{id}'" }.join(',')})"
       @region_condition += " AND record_name IN ('WR', '#{@continent.record_name}')" if @is_histories
     elsif @country.present?
-      @region_condition = "AND results.country_id = '#{@country.id}'"
+      @region_condition = "AND result.country_id = '#{@country.id}'"
       @region_condition += " AND record_name <> ''" if @is_histories
     else
       @region_condition = ""
--- a/app/controllers/results_controller.rb
+++ b/app/controllers/results_controller.rb
@@ -4,7 +4,7 @@ class ResultsController < ApplicationController
   REGION_WORLD = "world"
   YEARS_ALL = "all years"
   SHOW_100_PERSONS = "100 persons"
-  SHOWS = ["mixedslimseparatehistorymixed history"].freeze
+  SHOWS = ['mixed', 'slim', 'separate', 'history', 'mixed history'].freeze
   GENDERS = %w[Male Female].freeze
   SHOW_MIXED = "mixed"
   GENDER_ALL = "All"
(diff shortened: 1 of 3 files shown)


pranav-027:
I tried this approach, it definitely fixes the history and mixed history filter, but breaks the other 'non-history' filters mixed and slim.

Because, in non-history queries, we query the results table directly, and in history queries, we query from a UNION subquery aliased as `result`.  So I just renamed the alias to `results` https://github.com/thewca/worldcubeassociation.org/pull/11463

dunkOnIT:
Ah ok I see! Definitely evidence for the fact that we need tests against these endpoints

gregorbg (changes requested):
The "cleaner" (i.e., more consistent) fix here is actually to change the SQL query to use `results` with plural S as an alias.

Rationale: We're already using `results` as a table name and as an alias for the concise tables in all other queries throughout the controller. It's only the regional record query that (inconsistently) uses `result` in singular form. Changing the filter's expected alias to `result` would fix the regions query but it breaks all other query types in return.

app/controllers/results_controller.rb
```diff
@@ -176,12 +178,11 @@ def records
     params[:show] ||= SHOW_MIXED
     params[:gender] ||= GENDER_ALL
 
-    @shows = [SHOW_MIXED, "slim", "separate", "history", "mixed history"]
```
dunkOnIT, line 179: Moved this definition into a constant on new-line 7

spec/support/test_constants.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+module TestConstants
+  RESULT_TEST_EVENTS = %w[333 333mbf 333fm].freeze
+  RESULT_TEST_REGIONS = ['_Africa', 'South Africa'].freeze
```
gregorbg, line 5: Hm, I wonder who might have come up with these values...?

app/controllers/results_controller.rb
```diff
@@ -4,6 +4,8 @@ class ResultsController < ApplicationController
   REGION_WORLD = "world"
   YEARS_ALL = "all years"
   SHOW_100_PERSONS = "100 persons"
+  SHOWS = ['mixed', 'slim', 'separate', 'history', 'mixed history'].freeze
```
gregorbg, line 7: I think RuboCop doesn't like this...?

gregorbg (approved):

app/controllers/results_controller.rb
dunkOnIT: I was also expecting it to correct it, but I think because `mixed history` has a space it's fine with not using %w? It clearly didn't complain in CI, at any rate :P

PR #11467, opened 2025-05-04
Use unique keys in Users table

kr-matthews:
Probably just the email, without the name and id, would be sufficient?


gregorbg (approved):
Thanks! I just checked and the database does have an SQL `UNIQUE` constraint on the `email` column, so using the email alone as key should indeed be enough.

I'll leave it up to you whether you want to change this or keep the current PR as-is.

gregorbg (approved):

PR #11513, opened 2025-05-06
Fix coverage reporting

dunkOnIT:


spec/spec_helper.rb
```diff
@@ -6,7 +6,10 @@
 require 'webmock/rspec'
 SimpleCov::Formatter::LcovFormatter.config.report_with_single_file = true
 SimpleCov.formatter = SimpleCov::Formatter::LcovFormatter
-SimpleCov.start
+SimpleCov.start do
+  track_files 'app/**/*.rb'
+  add_filter '/spec/'
+end
```
gregorbg, line 12: It seems that SimpleCov has pre-defined configurations for Rails: https://github.com/simplecov-ruby/simplecov?tab=readme-ov-file#use-it-with-any-framework

dunkOnIT: Ah that's convenient!

gregorbg:
By using the instructions at https://github.com/tagliala/coveralls-ruby-reborn?tab=readme-ov-file#github-actions, your local environment will use the default `SimpleCov` format on your local machine test runs, which happens to be nice and beautified HTML that you can open in your browser of choice. Only in the CI will it switch to the "raw" Lcov format.

By inspecting the HTML report, I have been able to determine that our code coverage is indeed better than 3%. So it's not an inherent SimpleCov flaw, only the Lcov itself seems to be either misconfigured or not correctly parsed by Coveralls.

gregorbg:
Interestingly, disabling Spring gave exactly 3 consistent test failures. All of them were in feature specs, and upon closer investigation I was able to conclude that all 3 failures happened because our Rails application server running in the background (through Capybara) was now **too fast**.

Here's what ChatGPT has to say about that:

> # Why Feature Specs Run Faster (and More Reliably) Without Spring 🚀
>
> Even though Spring is designed to speed up Rails by preloading the app, it can **slow down or destabilize feature specs**. Here's why:
>
> - 🧠 **Stale preload**: Spring loads Rails *before* the test suite starts, which can interfere with tools like SimpleCov, test setup, or dynamic config.
> - 🌐 **Capybara server conflicts**: Capybara starts a separate app server (like Puma), which may clash with Spring's loaded environment or DB connections.
> - 🧵 **Thread & connection issues**: Spring can reuse database connections or thread pools in ways that break transactional tests or async behavior.
> - ❗ **Swallowed errors**: Follow-up HTTP requests (e.g., Turbo, background JS) may silently fail or never be noticed when Spring exits too fast.
> - ⏱️ **Cleaner teardown**: Without Spring, the app fully boots and shuts down per run, avoiding leftover state between tests.
>
> **Conclusion**: For unit tests, Spring may help. But for system/feature specs, **disabling Spring often results in faster, cleaner, and more accurate tests**. ✅

gregorbg:
In particular, there were two test failures who first checked "is the loader showing?" and then immediately checked "has the loader gone?" to make sure that (a) loading happens in the first place and (b) the loading completes. Our feature specs are now so fast that the loading finishes before the first check even kicks in. But because of (a), we cannot fully skip both checks, because it might falsely lead us to believe that loading completed when in reality it did not even start yet.

Solution: Test for the presence of more than one (there will always be the trivial "No selection" fallback) row in the dropdown.

The other test failure was when opening the competition form and setting an invalid competition ID (one with spaces included) as admin. The expected behavior is that the backend sends a 200 response which instructs the frontend form to trigger a redirect (Note: The "update competition payload" happens through an async mutation, so the backend cannot hard-redirect). Unfortunately, this redirect takes some time even after making the `window.location.replace` call to initiate the redirect (this is unintuitive on the part of JS, but I'm sure there are good reasons for this).
The accelerated feature spec without Spring now had just barely enough time to try and fetch proximity competitions and eligible Series Competition partners, using the corrupted ID from the test -- which of course led to an error because it intentionally is an invalid ID that doesn't exist.

Solution: Don't propagate the updated ID to the rest of the form if the response holds a redirect (in an ideal world, this shouldn't be possible, because the frontend `window.location` redirect call _does happen first_ in the mutation chain, so by the time the mutation success handler reaches the point where it persists the corrupted ID into storage, the redirect _should_ already have happened. In reality, redirecting takes time.)

gregorbg:
This now reports reasonable coverage data: https://coveralls.io/jobs/164587881 :tada:

FinnIckler (approved):

Issue #4426, opened 2019-08-05
Show 3x3 Multi-Blind points next to the result


Meneghetti:
The 3x3 Multi-Blind results are shown as "Solved/Attempted xx:xx", while the ranking itself is given by the actual "points" of the attempt, which is never shown on the website.

It would be good for the user experience if the points were shown right next to the result. Not only it would make the rules more clear and more intuitive for the regular user, but it would also make the visualization of the results faster for experienced users.

I would suggest that:

Instead of: **32/39** 56:05
It could be: **25** (32/39) 56:05

AlbertoPdRF:
Hi @Meneghetti, thanks for the suggestion!

We are currently actively working on porting that part of the website from PHP to RoR. Once we are done with that, we can think about this!

By the way, I'm CC'ing @lgarron for his opinion.

moralsh:
This, thanks to the porting done, is just a change in one line:

![image](https://user-images.githubusercontent.com/12932500/69826196-6582be80-1212-11ea-9e17-154b8b6cf314.png)

However I'm not sure this is adding too much numbers there, perhaps a better solution could be yo write the points on a tooltip so when your mouse is over the cell you can see the points, I might take another screenshot later if i think of a better way to do it than inserting HTML on that method.

PR #11516, opened 2025-05-06
show points of MBLD results in rankings and records page

pranav-027:
fixes: https://github.com/thewca/worldcubeassociation.org/issues/4426

### DIFF
--- a/app/webpacker/lib/wca-live/attempts.js
+++ b/app/webpacker/lib/wca-live/attempts.js
@@ -127,7 +127,7 @@ export function decodeMbldAttemptResult(value) {
   const solved = points + missed;
   const attempted = solved + missed;
   const centiseconds = seconds === 99999 ? null : seconds * 100;
-  return { solved, attempted, centiseconds };
+  return { solved, attempted, centiseconds, points };
 }
 
 /**
@@ -165,7 +165,7 @@ export function centisecondsToClockFormat(centiseconds) {
 }
 
 function formatMbldAttemptResult(attemptResult) {
-  const { solved, attempted, centiseconds } = decodeMbldAttemptResult(
+  const { solved, attempted, centiseconds, points } = decodeMbldAttemptResult(
     attemptResult,
   );
   const clockFormat = centisecondsToClockFormat(centiseconds);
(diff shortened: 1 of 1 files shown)


app/webpacker/lib/wca-live/attempts.js
```diff
@@ -174,7 +178,7 @@ function formatMbldAttemptResult(attemptResult) {
   // using it here allows us to expand space between mbf results without
   //  expanding the spaces within the individual results
   // see https://github.com/thewca/worldcubeassociation.org/issues/6375
-  return `${solved}/${attempted}\u2002${shortClockFormat}`;
+  return `${points}\u2002(${solved}/${attempted})\u2002${shortClockFormat}`;
```
FinnIckler, line 181: Isn't there a way to do this without using unicode characters like \u2002?

pranav-027: It's already used in the same result. If you can suggest something better happy to search and implement:)

gregorbg:
Hmm... This is not the standard way how we display MBLD results. Is there a wider community consensus on this?

gregorbg (changes requested):

pranav-027:
> Hmm... This is not the standard way how we display MBLD results. Is there a wider community consensus on this?

Yes, the linked issue https://github.com/thewca/worldcubeassociation.org/issues/4426 has all the relevant discussion! 
Here is the link to forum [poll](https://forum.worldcubeassociation.org/t/multi-blind-point-display/13070) as well!

gregorbg:
I am not comfortable introducing this change based on a 4-year-old community poll with 79 anonymous responses.

If we are going to fundamentally change the way MBLD results are displayed, then this could have broad-ranging implications. Essentially, through this PR you are introducing a new, never-before-seen "MBLD format". It will introduce an inconsistency to other places where MBLD results are written, for example on the profile pages.

If we want to keep it consistent, we would need to introduce this "new MBLD" format _everywhere_. And if we really do that, we need to do much, _much_ better than 79 random forum users. Also, if we end up deciding to do this, then does this also imply that on scoresheets at competitions all around the world, you would need to start writing down "points (solved/attempted) time" in handwriting on the scoresheets?

Issue #10890, opened 2025-02-21
Recompute Delegate report discussion URL when competition name Is updated


pranav-027:
**Describe the bug**
The delegate report discussion URL is set when competition is created using the `set_discussion_url` method. But it remains unchanged if the competition name is changed. 

Example: [Mumbai Cubing Sprint 2024](https://www.worldcubeassociation.org/competitions/MumbaiCubingSprint2024)

I think the discussion URL could be recomputed after the name change or delegate_report could be created once the comp is announced.

gregorbg:
Both approaches sound viable, but the first one is a bit more risky since the old URL could already be active if somebody tried to click it. I like the notion of the DR being generated only once the competition is announced more.

In the bigger picture, we would probably want to get rid of the reliance on Google Groups for viewing DRs, but that's a different project of a pretty large scope.

Issue #7076, opened 2022-07-30
Discussion links in delegate reports points to an incorrect name


simonkellly:
**Describe the bug**
When viewing a delegate report such as the one for [Don't DNF Dublin 2022](https://www.worldcubeassociation.org/competitions/dontdnfdublin2022/report) the link to view the discussion for the delegate report points to a URL which specifies the name the competition was originally created as, rather than its actual name

**Expected behavior**
The view discussion link will point to the correct google groups search

From a quick glance, I think it saves this URL some time before the report is submitted and at some point between this time and when it is reported, the competition's name is set to it's actual name

Issue #3595, opened 2018-11-12
Incorrect Search for Competition Reports


capitalcubing:
**Describe the bug**
When viewing the delegate report of a competition on the website, there is an option to view it in google groups for the full discussion. The option searches the competition name in google groups. However, often the name is not correct. (not sure why)

**To Reproduce**
An example:[Caarn Ahmedabad Cube Open Report](https://www.worldcubeassociation.org/competitions/CaarnAhmedabadCubeOpen2018/report). This searches for Cube Carnival instead of Cube Open and that leads to 0 results in google group.

akashrupela:
I am the author. Used another team account by mistake.

AlbertoPdRF:
This was reported in Slack and here's @jonatanklosko's reply:

> We generate the report URL as soon as it's created (doesn't mean sent) (https://github.com/thewca/worldcubeassociation.org/blob/32a6ac7400978b37ced5bbcda843f79ee9ad584d/WcaOnRails/app/models/delegate_report.rb#L13). This happens as soon as someone visits the "Delegate report" page, even without doing any changes (e.g. by accidentally navigating to this page).

> So in this case, the report had been crated before the name changed, that's why the URL is 'outdated'

jfly:
I see two options:

1. Drop the `discussion_url` column from the database, and instead always autogenerate it.
2. Only populate `discussion_url` at the time the report is *posted*, rather than when the report is created in our database.

@pedrosino, do you have some time to work on one of these (or some other clever third idea I haven't thought of?)

AlbertoPdRF:
I'm labeling this as a good second issue.

PR #11518, opened 2025-05-07
delegate_report discussion url to be created only while posting

pranav-027:
fixes: https://github.com/thewca/worldcubeassociation.org/issues/10890, https://github.com/thewca/worldcubeassociation.org/issues/3595 and https://github.com/thewca/worldcubeassociation.org/issues/7076

setting discussion url only at the time of posting

### DIFF
--- a/app/models/delegate_report.rb
+++ b/app/models/delegate_report.rb
@@ -99,7 +99,7 @@ class DelegateReport < ApplicationRecord
     new_posted = ActiveRecord::Type::Boolean.new.cast(new_posted)
     self.posted_at = (new_posted ? Time.now : nil)
     self.posted_by_user_id = current_user&.id
-    self.discussion_url = "https://groups.google.com/a/worldcubeassociation.org/forum/#!topicsearchin/reports/#{URI.encode_www_form_component(competition.name)}" if new_posted && self.discussion_url.blank?
+    self.discussion_url = "https://groups.google.com/a/worldcubeassociation.org/forum/#!topicsearchin/reports/#{URI.encode_www_form_component(competition.name)}"
   end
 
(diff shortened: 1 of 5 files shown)


spec/factories/competitions.rb
```diff
@@ -238,7 +238,7 @@
 
     trait :with_delegate_report do
       after(:create) do |competition|
-        FactoryBot.create(:delegate_report, :posted, competition: competition)
```
gregorbg, line 241: Have you tried setting this as a default in the `delegate_report` factory directly? You could introduce it as a trait, something like `with_report_url` and then reference that here

app/models/delegate_report.rb
```diff
@@ -104,6 +99,7 @@ def posted=(new_posted)
     new_posted = ActiveRecord::Type::Boolean.new.cast(new_posted)
     self.posted_at = (new_posted ? Time.now : nil)
     self.posted_by_user_id = current_user&.id
+    self.discussion_url = "https://groups.google.com/a/worldcubeassociation.org/forum/#!topicsearchin/reports/#{URI.encode_www_form_component(competition.name)}" if new_posted && self.discussion_url.blank?
```
gregorbg, line 102: Why do you introduce the additional behavior that makes sure existing URLs are not overridden? (the part about `if ... && self.discussion_url.blank?`)

The previous implementation _always_ set the discussion URL, no matter what. So I don't see why introducing this new behavior is necessary.

gregorbg (changes requested):

app/models/delegate_report.rb
pranav-027: Makes sense. I have removed it

Issue #11231, opened 2025-04-02
Translation status calculation mislabels registration errors


jonatanklosko:
*Originally posted by @AlexandreOndet in https://github.com/jonatanklosko/internationalize/issues/34.*

> Yesterday I noticed that even if I got the French translation up to date (and synchronized) and that it was merged. [It still shows](https://www.worldcubeassociation.org/translations/status) that some elements needs attention:
>
> ![Image](https://github.com/user-attachments/assets/b2821e6e-64e0-4d71-bd2d-536205a9d6fc)
>
> It appears some lines are not used anymore or something, shouldn't Internationalize remove those when exporting the file?

Notice that the same keys are in the "Missing" and "Unused" sections, so it's most likely a bug.

AlexandreOndet:
The bug is now fixed, can someone close this issue please?

PR #11522, opened 2025-05-08
Fixing translation status calculation mislabels registration errors

AlexandreOndet:
Here is a fix for #11231 
The keys in en.yml must use the `'` in order for the comparison to work when tracking status.
It gives us this now:
1_error
Which is better but not perfect. I still want to figure out why it says that `logo > intro` is outdated because it shoudn't, we actually are up to date.


gregorbg (approved):
Thanks for figuring this one out! Sneaking in the `fr` hash manually... I shall allow it :stuck_out_tongue_winking_eye:

AlexandreOndet:
The last key issue seems to be due to the fact that there's no `#original_hash: ` present in `fr.yml` but only for this specific key. So my guess would be a bug in the way internationalize generates the file.
Maybe @jonatanklosko would have a better understanding of what is happening ?
I tried computing it manually just to see but it's not a real fix since:
1. French is not the only language affected
2. We would need to re add it manually every time which is not really sustainable

PR #11524, opened 2025-05-08
Allow force-resetting cronjobs in the admin panel

gregorbg:
The Results Dump is failing every now and then for mysterious, unknown reasons. Could be deployment-related, but then I would expect the Developer Dump to be affected as well, but it isn't.

This is a "fix" that allows us to reset CJ state without digging into the database, by force-enabling the `Reset` button when the debug section is expanded. A little bit overkill, but should get the job done while we don't have the time (or motivation) to dig super deep into the Redis logs or memory model of Sidekiq...


FinnIckler (approved):
I assume the result export failing is the same issue as the tsv exports sometimes only containing half it's data... but I agree we probably don't need to investigate it too much

PR #11529, opened 2025-05-09
Activate Lint/UnusedMethodArgument cop

FinnIckler:
Didn't want to introduce too many code changes, so I just added _ to most variables


gregorbg (changes requested):
Hm, "just" prepending an underscore to avoid code changes kinda defeats the purpose of the cop. (Because if that's what we're doing, then we can just keep the cop at `Disabled` and call it a day.)

Did you put some more thought into which code changes are reasonable and which ones are silly (because we will definitely get rid of the method soon)?

Also, I want to point out that there's the `AllowUnusedKeywordArguments` configuration option: https://docs.rubocop.org/rubocop/1.75/cops_lint.html#lintunusedmethodargument which I think solves a non-trivial amount of `rubocop:disable` that you had to do

app/models/competition.rb
```diff
@@ -637,7 +637,7 @@ def championship_warnings
     warnings
   end
 
-  def info_for(user)
```
gregorbg, line 640: This one is interesting. The original purpose was to show "info for the given user". That's why the (pre-RuboCop) method kinda implies the name "info for user". I am not sure at which point passing in the user became unnecessary, but presumably it's because we're moving more and more warnings to React.

With all that being said, the method should probably be renamed if this argument is being removed, because it means now there is no user anymore that we're generating information _for_. So the "for" should not appear in the name.

lib/results_validators/persons_validator.rb
```diff
@@ -60,7 +60,7 @@ def self.dob_validations(dob, competition_id = nil, **message_args)
       validation_issues
     end
 
-    def self.name_validations(name, competition_id = nil, **message_args)
+    def self.name_validations(name, competition_id = nil, **_message_args)
```
gregorbg, line 63: Curious that this doesn't seem to be covered by `AllowUnusedKeywordArguments`... Have you double-checked? Or might this be a leftover from the time before you introduced that config option I mentioned?

FinnIckler: sadly no

app/models/competition.rb
```diff
@@ -637,7 +637,7 @@ def championship_warnings
...
+  def info
```
gregorbg, line 640: A tad bit too generic now, what about `info_messages`?

gregorbg (approved):

gregorbg (approved):

PR #11530, opened 2025-05-09
Activate Style/AccessModifierDeclarations cop

FinnIckler:
This moves a bunch of methods around, but it was all autocorrect

PR #11532, opened 2025-05-09
Activate Style/ClassAndModuleChildren cop

FinnIckler:
all autocorrected, boils down to only whitespace changes apart from the module head


FinnIckler:
actually I wanted to close this, not https://github.com/thewca/worldcubeassociation.org/pull/11530

PR #11534, opened 2025-05-09
Added tracking of non en-based locales

AlexandreOndet:
As mentioned in #11522, languages not based on `en` are currently not tracked properly.
In this PR I propose to add a config file to list non standard based locales and to compare these one to those instead of English.

So we go from this:
fr_CA_not_tracked

To this:
fr_CA_tracked

The last two keys seem to have a hash issue, I'm not sure why yet but I don't think it's related to the changes in this PR.


gregorbg:
Thanks for the submission! I just merged your other PR, could you rebase (or otherwise update) this one so that the "numeric to string" key changes disappear from the diff please?

AlexandreOndet:
Done 🙂

config/i18n_config/locale_base.yml
gregorbg, line 1: You don't have to introduce a new file for this purpose. Instead you can just hijack `config/locales/locales.rb`

gregorbg (changes requested):

config/i18n_config/locale_base.yml
AlexandreOndet: that's fair

app/controllers/translations_controller.rb
```diff
@@ -4,10 +4,12 @@ class TranslationsController < ApplicationController
   before_action :authenticate_user!, except: [:index]
 
   def self.compute_bad_i18n_keys
-    english = locale_to_translation('en')
+    base_locales = Locales::AVAILABLE.transform_keys(&:to_s).transform_values { |attrs| attrs[:base_locale] || 'en' }
```
gregorbg, line 7:
```suggestion
    base_locales = Locales::AVAILABLE.stringify_keys.transform_values { |attrs| attrs[:base_locale] || 'en' }
```

gregorbg, line 7:
```suggestion
    base_locales = Locales::AVAILABLE.transform_keys(&:to_s).transform_values { it[:base_locale] || 'en' }
```

gregorbg: For context: `it` can be used as the default parameter name for lambdas or blocks that have only one argument, since Ruby 3.4 (which we are conveniently already using :wink:)

app/controllers/translations_controller.rb
```diff
@@ -4,10 +4,12 @@ class TranslationsController < ApplicationController
...
 
     (I18n.available_locales - [:en]).index_with do |locale|
-      locale_to_translation(locale).compare_to(english)
+      base_locale = base_locales[locale.to_s]
```
gregorbg, line 10: Why did you add the `to_s` here? If you remove it and keep the status quo here, then you might not have to manually "stringify" the keys up in `base_locales` either.

gregorbg (changes requested):

app/controllers/translations_controller.rb
AlexandreOndet: good idea

gregorbg (approved):

PR #11535, opened 2025-05-10
Registration admin page fix for anonymized registration

pranav-027:
The registration admin page breaks if any registration is anonymized. 

image

This happens because of the null value of the country!


app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationTableFooter.jsx
```diff
@@ -25,7 +25,9 @@ export default function RegistrationAdministrationTableFooter({
   ).length;
 
   const countryCount = new Set(
-    registrations.map((reg) => reg.user.country.iso2),
+    registrations
+      .map((reg) => reg.user.country?.iso2)
+      .filter((iso2) => iso2),
```
kr-matthews, line 30:
```suggestion
      .filter(Boolean),
```

pranav-027: @kr-matthews done

PR #11308, opened 2025-04-14
Schedule Editor: Fix sticky & reposition delete drop-zone

kr-matthews:
_Note: view the diff without white space changes._

Problems: Sticky behaviour is weird, and delete drop-zone is often not visible, requiring to drag an event temporarily up the calendar.

[Screencast from 2025-04-13 05:05:57 PM.webm](https://github.com/user-attachments/assets/12601e27-6e31-413f-9c7d-46aa1e30b9d5)

Solution: Fix sticky behaviour and move delete drop-zone into the sticky column.

[Screencast from 2025-04-13 05:06:55 PM.webm](https://github.com/user-attachments/assets/538c03d8-319e-4ac2-9404-3aac22954038)

If your screen is small, you can still access the bottom of the event picker, though only when scrolled down to the bottom. This page/screen already didn't work well on phones, and this is unlikely to ever cause a real problem on a desktop.

[Screencast from 2025-04-13 05:10:40 PM.webm](https://github.com/user-attachments/assets/66938079-7180-4780-ba01-da04d618d8b5)

### Not part of PR

Note: deleting an event via right-clicking it might be nice. Ideally with a pop-up context menu with a delete option. But as a proof-of-concept without the menu, just add this prop to `FullCalendar`:
```
                    // no direct `eventContextMenu` prop or anything like that unfortunately
                    eventDidMount={(arg) => {
                      arg.el.addEventListener('contextmenu', (e) => {
                        e.preventDefault();
                        removeEvent(arg.event);
                      });
                    }}
```

[Screencast from 2025-04-13 05:09:08 PM.webm](https://github.com/user-attachments/assets/db2a6eb1-f2a2-440c-88d7-427198244724)

PR #11538, opened 2025-05-10
Edit Schedule: right-click to delete

kr-matthews:
Taken from what I mentioned on #11308. I could not find a simple way to have a context menu appear, because we aren't rendering the event on the calendar (FullCalendar is, we just provide the content) and therefore cannot wrap it in a `Popup` or anything. I'm open to suggestions though.

It's not clear to users that this functionality now exists. It's a bit jarring maybe, if you weren't expecting it, but it's easy to click no to cancel. Before this, right-click would do nothing.

[Screencast from 2025-05-09 10:20:42 PM.webm](https://github.com/user-attachments/assets/6923e22a-bd09-43d3-b7e3-829d034afa67)

### DIFF
--- a/app/webpacker/components/EditSchedule/EditActivities/index.js
+++ b/app/webpacker/components/EditSchedule/EditActivities/index.js
@@ -520,6 +520,14 @@ function EditActivities({
                     select={addActivityFromCalendarDrag}
                     // allow clicking on existing events to edit them
                     eventClick={editCustomEvent}
+                    // allow right-clicking on existing events to delete them;
+                    //  no direct `eventContextMenu` prop or anything like that unfortunately
+                    eventDidMount={(arg) => {
+                      arg.el.addEventListener('contextmenu', (e) => {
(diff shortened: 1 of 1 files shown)


gregorbg (approved):
It's not a hill that I will definitely die on, but at least I want to ask once:
In the existing `eventClick` listener, is there a way to figure out whether the click that triggered the event was indeed a right-click?

kr-matthews:
> It's not a hill that I will definitely die on, but at least I want to ask once: In the existing `eventClick` listener, is there a way to figure out whether the click that triggered the event was indeed a right-click?

I did not/cannot find any documentation or other sources explaining that this is possible in React FullCalendar, and when I right-click a (calendar) event, our on-click handler is not called at all. (Google's search AI claims it is possible, but it seems to be wrong as far as I can tell.)

gregorbg:
I just played around with this and it seems that `eventClick` doesn't even fire for right-click. Oh well...

Issue #2343, opened 2017-12-29
API specific 404 that returns a JSON error message


coder13:
instead of returning a full html page, we should send back a short and simple JSON error message when someone tries to query a page we don't have under /api.

trungams:
I emailed the WCA software team to say that would like to take on this issue :)

viroulep:
Sure thing @trungams, I've assigned it to you :)
Personally I'll be unavailable for the next 10 days, but definitely feel free to open a pull request as soon as you want some feedback from WST/contributors to this repository; we definitely welcome draft or WIP PR to discuss and comment if you're unsure about anything!

UnsolvedCypher:
@trungams Is this still an issue you are working on?

trungams:
@UnsolvedCypher Hi, sorry I completely forgot about this. Unfortunately, no. I had a WIP at the time but got caught up with life and all. I don't think I will be able to work on this in the near future :disappointed:

UnsolvedCypher:
No worries @trungams!

I'm thinking about the best way to do this, and I'm not sure where the actual controller for the endpoint should go. It can't go in the file with the rest of the API because that is only for v0 (so it would only catch /api/v0/(...) URLs and we want to catch /api/(...) as well). It could go with the static pages though it's not exactly a static "page". Does anyone have thoughts on this?

PR #11541, opened 2025-05-10
JSON error message for not implemented API endpoints

pranav-027:
fixes: https://github.com/thewca/worldcubeassociation.org/issues/2343

### DIFF
--- a/app/controllers/api/api_controller.rb
+++ b/app/controllers/api/api_controller.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+class Api::ApiController < ApplicationController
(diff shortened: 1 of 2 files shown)


FinnIckler:
There is technically already https://github.com/thewca/worldcubeassociation.org/blob/main/app/controllers/api_errors_controller.rb, but we aren't using it in production

pranav-027:
Okay, is there any reason behind it?

FinnIckler:
Because it's low level rails and overwrites the general error page. The general idea was for API only servers to only return json and for rendering servers to html, but we aren't there yet.

Issue #10172, opened 2024-11-02
Typing enter while entering a cut-off ignores the input


kr-matthews:
**Describe the bug**
When editing a cutoff in a competition's edit events form, typing a time followed by enter will ignore the time and close the modal with the old value still in use (0 if there was no old value).

**To Reproduce**
1. Go to a competition you're organizing/delegating and go to Edit > Manage Events
2. Add an event, add a round for that event, click on the "-" box to edit the cutoff.
3. Select Best of 2 and type something into the result box.
4. Type enter.
5. The modal closes and 0 is used as the cutoff result, ignoring the time you had typed.

**Expected behavior**
Enter should use the time and close the modal.

**Additional Context**
There are a lot of modals on this page, make sure they all work properly when typing enter in any input.

kr-matthews:
I'm not sure how to fix this. The `TimeField` (the input) has a `draftInput` value which isn't passed on to the actual value until the input is blurred (because you might want to type 60000 for 6:00.00, but after the first 4 keys the input is 6000 which can't be parsed to a valid time). Pressing enter submits the modal, which causes the time field to be removed before it has a chance to use the draft input value.

Disabling enter (by removing the `onSubmit` from the modal and adding the handling to the modal button instead) would prevent people from accidentally losing the value they typed, but would be annoying - and this modal is used elsewhere too so that might have other consequences.

PR #11542, opened 2025-05-11
Edit time limit modal improvements

ticklemepierce:
https://discord.com/channels/1121050787074805830/1121171516105699408/1365715107795173417

PR for the above.  Favor function reuse over component reuse as the component was built for WCA live and therefore had additional features not needed in edit modal.

### DIFF
--- a/app/webpacker/components/EditEvents/Modals/EditTimeLimitModal/index.js
+++ b/app/webpacker/components/EditEvents/Modals/EditTimeLimitModal/index.js
@@ -5,9 +5,8 @@ import _ from 'lodash';
 import { Form, Label, Radio } from 'semantic-ui-react';
 import { events } from '../../../../lib/wca-data.js.erb';
 import { roundIdToString } from '../../../../lib/utils/wcif';
-import { centisecondsToClockFormat } from '../../../../lib/wca-live/attempts';
+import { centisecondsToClockFormat, clockFormatToCentiseconds, reformatInput } from '../../../../lib/wca-live/attempts';
 import { useDispatch } from '../../../../lib/providers/StoreProvider';
-import TimeField from '../../../EditResult/WCALive/AttemptResultField/TimeField';
 import { updateTimeLimit } from '../../store/actions';
 import ButtonActivatedModal from '../ButtonActivatedModal';
 import TimeLimitDescription from './TimeLimitDescription';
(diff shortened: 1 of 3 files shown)


gregorbg:
I do not have access to that Discord. Can you please summarize your changes and the motivation for your changes here?

kr-matthews:
> I do not have access to that Discord. Can you please summarize your changes and the motivation for your changes here?

I believe this is for #10172.

ticklemepierce:
> > I do not have access to that Discord. Can you please summarize your changes and the motivation for your changes here?
> 
> I believe this is for #10172.

Yes that's correct

app/webpacker/components/EditEvents/Modals/EditTimeLimitModal/index.js
```diff
@@ -51,12 +56,12 @@ export default function EditTimeLimitModal({ wcifEvent, wcifRound, disabled }) {
 
   const Title = useMemo(() => `Time limit for ${roundIdToString(wcifRound.id)}`, [wcifRound.id]);
 
-  const hasUnsavedChanges = () => (
+  const hasUnsavedChanges = useCallback(() => (
     !_.isEqual(timeLimit, { centiseconds, cumulativeRoundIds })
-  );
+  ), [timeLimit, cumulativeRoundIds, draftInput]);
```
gregorbg, line 61: I see that `draftInput` and `centiseconds` are somewhat tied together, but I still don't see how including `draftInput` in this dependency array helps in any way. It should probably be `centiseconds` instead

ticklemepierce: yes I believe that is correct.  Sorry for the delay - have been quite busy

kr-matthews:
> this had fallen off my radar a bit but I can try and look at it again soon. Fixing the underlying component didn't seem worth it to me as the component used here doesn't seemed to be really connected to the usage from WCA live. That could just be that I don't understand the usages well enough but it doesn't seem to me to be a bug with the underlying component so much as trying to use the component for something it wasn't designed for

Do the other uses need the extra functionality? If not, then simplifying the component and making it work everywhere still seems best to me. There should be a good reason if different time inputs across the website aren't using the same component. If you replace the existing component with your new logic, does it work everywhere else that the existing component is used?

PR #11573, opened 2025-05-12
Migrate Scramble Matcher UI

FinnIckler:
The lib code is still a big mess, because it's 1:1 taken from the current implementation of scramble matcher.

The actual Matching drag and drop Table is not done either, as there is still some issue with the auto assign lib.

Screenshots:
image
image
image
image


pranav-027:
Is this going to eliminate the existing scramble matcher?

gregorbg:
This one PR, no. But this plus the three or four follow-up PRs, yes.

FinnIckler:
I removed the buttons and just assign on upload automatically.
image

FinnIckler:
image
new table just having the scramble row draggable

pranav-027:
Is 'attempt-based' matching considered here for MBLD and FMC? 
image

Like this?

gregorbg:
Not yet, we're working on this in separate stages.

gregorbg (approved):
Merging as a baseline for further development, not visible in production yet.

PR #11579, opened 2025-05-15
Activate 5 small Rspec cops

FinnIckler:
These all only produced a couple of lines of diff. There was one duplicated test, not sure what that was about @dunkOnIT

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -357,21 +357,6 @@ RSpec/NamedSubject:
 RSpec/Be:
   Enabled: false
 
-RSpec/ScatteredLet:
-  Enabled: false
-
-RSpec/RepeatedExample:
-  Enabled: false
-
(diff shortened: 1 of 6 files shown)


dunkOnIT:
Looks like I had a similar test for normal users and organizers, but the organizer one was also using the `default_user`. I've just tried to change it in the github UI - I'll fix the tests if I broke them

gregorbg (approved):

gregorbg:
> I'll fix the tests if I broke them

Looks like you did :stuck_out_tongue_winking_eye:

gregorbg (changes requested):

dunkOnIT:
I'm not 100% confident to resolve the merge conflicts (it's not clear to me how they've reappeared)

gregorbg (approved):

PR #11580, opened 2025-05-15
6 more small rspec cops

FinnIckler:
again, super small changes for each cop (one even had no changes needed)

### DIFF
--- a/spec/features/competitions_list_spec.rb
+++ b/spec/features/competitions_list_spec.rb
@@ -15,14 +15,16 @@ RSpec.feature "Competitions list", :js do
       before do
         visit "/competitions?show_admin_details=yes"
 
-        # The delegate dropdown should have three items: The generic 'None'
-        #   as well as two Delegates created by the `let(:competition)` above.
-        expect(page).to have_css("#delegate div.item", visible: :hidden, count: 3)
-
         within(:css, "#delegate") do
           find(".search").set(delegate.name)
           find(".search").send_keys(:enter)
         end
+      end
+
+      # The delegate dropdown should have three items: The generic 'None'
+      #   as well as two Delegates created by the `let(:competition)` above.
+      it "shows the delegate dropdown" do
+        expect(page).to have_css("#delegate div.item", visible: :hidden, count: 3)
       end
 
       it "the delegate is selected within the form" do
--- a/spec/features/eligible_voters_spec.rb
+++ b/spec/features/eligible_voters_spec.rb
@@ -4,7 +4,11 @@ require "rails_helper"
 require "csv"
 
 RSpec.feature "Eligible voters csv" do
-  before { Timecop.freeze(Time.utc(2016, 5, 5, 10, 5, 3)) }
+  before {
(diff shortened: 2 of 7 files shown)


spec/features/competitions_list_spec.rb
```diff
@@ -15,16 +15,18 @@
       before do
         visit "/competitions?show_admin_details=yes"
 
-        # The delegate dropdown should have three items: The generic 'None'
-        #   as well as two Delegates created by the `let(:competition)` above.
-        expect(page).to have_css("#delegate div.item", visible: :hidden, count: 3)
```
gregorbg, line 20: I'm inclined to allow an exception here, because this is our undercover way of assuring _before each individual test_ that the async "load available Delegates" `useQuery` finished loading...

gregorbg: Notably, it's *not* an individual test that is supposed to check "can it load". It is an actual guarantee for all tests that it did indeed load -- without this guarantee, all further assumptions about "selecting a Delegate from the dropdown" don't make much sense

spec/features/eligible_voters_spec.rb
```diff
@@ -4,7 +4,11 @@
 require "csv"
 
 RSpec.feature "Eligible voters csv" do
-  before { Timecop.freeze(Time.utc(2016, 5, 5, 10, 5, 3)) }
+  before {
+    Timecop.freeze(Time.utc(2016, 5, 5, 10, 5, 3))
+    sign_in board_member
+  }
```
gregorbg, line 10: Two lines means one of your other cop PRs will want to turn it into a "proper" `do` block I think

spec/jobs/clear_connected_payment_integrations_spec.rb
```diff
@@ -11,7 +11,7 @@
       expect(competition.competition_payment_integrations.count).to eq(1)
     end
 
-    it 'disconnects all integrations  older than disonnect_delay' do
```
gregorbg, line 14: lol there is a cop for this

gregorbg (changes requested):

gregorbg:
Also, probably the most nitpicky nitpick ever: It's 7 cops, not 6 :stuck_out_tongue:

FinnIckler:
as I said, one of the cops didn't actually do any changes, that's why I said 6

spec/features/competitions_list_spec.rb
FinnIckler: ah okay. I couldn't infer that from the comment

FinnIckler: the tests still pass interestingly

gregorbg: That's just a happy coincidence. It means that during _this one particular test run_, the loading happened to be fast enough for the dropdown to be populated already. So it's literally luck. With the assumption, you can be _sure_ that the dropdown loaded.

gregorbg:
> as I said, one of the cops didn't actually do any changes

But you still _activated_ it :upside_down_face:

PR #11584, opened 2025-05-16
Allow Results submission from WCA Live synced Round Results

FinnIckler:
image
This adds a currently very crude button (I really want to migrate that page to react :upside_down_face: ) to import the Results directly from what was submitted to WCA Live, instead of going through scramble matcher.
I wasn't fully able to test this yet as Round Results is not populated in the dev-dump.


app/controllers/results_submission_controller.rb
```diff
@@ -97,6 +97,75 @@ def upload_json
...
   end
 
+  def import_from_live
+    @competition = competition_from_params
+    return redirect_to competition_submit_results_edit_path if @competition.results_submitted?
+
+    errors = []
+    results_to_import = @competition.rounds.flat_map do |round|
+      round.round_results.map do |result|
+        results_to_import >> InboxResult.new({
```
gregorbg, line 107: You shouldn't be appending anything here anymore. Just `InboxResult.new` directly as return value will do.

app/controllers/results_submission_controller.rb
```diff
@@ -97,6 +97,75 @@ def upload_json
...
+      end
+    end
+
+    person_with_results = results_to_import.map(&:person_id).uniq
+
+    persons_to_import = @competition.registrations.wcif_ordered
+                                    .includes([:user])
+                                    .to_enum
+                                    .with_index(1)
+                                    .select { |r, registrant_id| r.wcif_status == "accepted" && person_with_results.include?(registrant_id) }
```
gregorbg, line 130: Try to merge this `person_with_results.include?` check into the chain much earlier: `@competition.registrations.where(id: person_with_results).wcif_ordered...` and see whether that works

gregorbg: (not sure if it has any uexpected side-effects, but at least on the surface it's more efficient

app/controllers/results_submission_controller.rb
```diff
@@ -97,6 +97,75 @@ def upload_json
     end
   end
 
+  def import_from_live
+    @competition = competition_from_params
+    return redirect_to competition_submit_results_edit_path if @competition.results_submitted?
+
+    errors = []
```
gregorbg, line 104: You can initialize this further down below now

gregorbg (changes requested):

app/controllers/results_submission_controller.rb
FinnIckler: well currently the id here is the wcif_id so that doesn't work. And I don't think we give the raw registration_id in the wcif

gregorbg: Aaah the infamous WCIF order

app/views/results_submission/new.html.erb
```diff
@@ -10,11 +10,14 @@
...
       <ul>
-        <li>Uploading a valid JSON to the website</li>
+        <li>Uploading a valid JSON to the website or use the results submitted via WCA Live</li>
         <li>Then the website will run a number of checks. If everything is correct you'll be able to submit these results to the WRT.</li>
       </ul>
     </p>
     <%= render "submit_json_panel", upload_json: @upload_json, already_has_submitted_result: @results_validator.any_results? %>
+    <%= simple_form_for :submit_from_live, url: competition_import_from_live_path do |f| %>
+      <%= f.submit "Submit WCA Live Results", class: "btn btn-primary" %>
+    <% end %>
```
gregorbg, line 20:
```suggestion
    <%= button_to(
            competition_import_from_live_path(@competition),
            class: 'btn btn-danger',
            method: :post
          ) do %>
        <%= Submit WCA Live Results %>
      <% end %>
```

gregorbg (changes requested):

app/views/results_submission/new.html.erb
FinnIckler: it's 
<%= button_to competition_import_from_live_path,
                  title: "Submit WCA Live Results",
                  class: "btn btn-primary",
                  method: :post do
    %>
      Submit WCA Live Results
    <% end %>

FinnIckler: rails seems to be clever enough to infer the competition

FinnIckler:
I copy pasted some Round Results from prod and made the changes to fix the import.
In the previous version there is a comment that says:
```
# Using this way of setting the attribute saves one SELECT per result
# to validate the competition presence.
# (a lot of time considering all the results to import!)
new_res.competition = competition
```
But I am seeing a bunch of
` CACHE Competition Load (0.0ms)  SELECT `competitions`.* FROM `competitions` WHERE `competitions`.`id` = 'GermanOpen2025' LIMIT 1
`
So this should not be valid anymore with current Rails?

FinnIckler:
Took an older competition with just one event and successfully imported the results.

app/models/inbox_scramble_set.rb
```diff
@@ -24,6 +24,18 @@ def backfill_round_information!
     self.round_type_id = matched_round.round_type_id
   end
 
+  def group_id
```
gregorbg, line 27: The name here is a bit ambiguous. I don't think that the name `group_id` convincingly implies that it is non-numeric. What about something like `alphabetic_group_index`?

gregorbg (approved):

PR #11586, opened 2025-05-18
Special account anonymization fix

pranav-027:
as reported on Slack - Anonmysation for special account isn't working

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1472,7 +1472,7 @@ class User < ApplicationRecord
       last_sign_in_ip: nil,
       # If the account associated with the WCA ID is a special account (delegate, organizer,
       # team member) then we want to keep the link between the Person and the account.
-      wca_id: special_account? ? new_wca_id : nil,
+      wca_id: special_account? ? wca_id : nil,
       current_avatar_id: special_account? ? nil : current_avatar_id,
       country_iso2: special_account? ? country_iso2 : nil,
     )


gregorbg:
Can you please give some context as to why this fix works and why it doesn't break the whole anonymization procedure (especially for "non-special-account" cases)?

pranav-027:
For the usual anonymisation process, we do not keep any link between the user and the person, and we set the wca_id to null.

But as mentioned in the comments on lines 1473 and 1474, we need to keep the link between the user and the person for special accounts.

The issue here was that we were setting `new_wca_id` in the users table, but it is not computed or defined in `user.rb`. Instead, it is computed in the anonymise method of `person.rb`, where we update the `wca_id` with `new_wca_id`.

Therefore, in the user anonymise method, we only need to set the `wca_id`, as it has already been updated in person with the correct `new_wca_id`. And about breaking the normal anonymisation process as changes are being made in conditional check for special accounts the non special anonymisation is untouched

danieljames-dj:
Thanks @pranav-027 for quickly raising the fix, and sorry that I missed this during implementation. As Gregor mentioned, the `wca_id` in `user.rb` will still point to the old WCA ID. I also agree with Gregor's solution of taking the return value from `person.anonymize` and pass it as an argument to `user.anonymize`.

pranav-027:
@gregorbg and @danieljames-dj Sorry for the delay your fix makes much more sense I have implemented it as per your suggestions

app/controllers/tickets_controller.rb
```diff
@@ -180,9 +180,8 @@ def anonymize
     end
 
     ActiveRecord::Base.transaction do
-      person&.anonymize
-
-      users_to_anonymize.each(&:anonymize)
+      new_wca_id = person&.anonymize
+      users_to_anonymize.each { |user_to_anonymize| user_to_anonymize.anonymize(new_wca_id) }
```
gregorbg, line 184:
```suggestion
      users_to_anonymize.each { it.anonymize(new_wca_id) }
```

gregorbg: Cool new Ruby 3.4 language feature: Blocks can take `it` as the implicit default name if there is only one argument

gregorbg (commented):
Code syntax nitpick, otherwise looking good now!

app/controllers/tickets_controller.rb
pranav-027: Done

pranav-027:
@gregorbg, can we merge this? one task is blocked becasue of this

gregorbg (approved):

PR #11618, opened 2025-05-20
Scramble matcher UX Improvements

FinnIckler:
Adds:
- Can't drag if there is only 1 row
- Can't drag if a scramble is missing
- Shows missing scramble in table
- Don't show Rounds Panel when there is only one round

I still want to add an error indicator for missing scrambles and greying out the submit button when there are no changes or there are errors

### DIFF
--- a/app/webpacker/components/ScrambleMatcher/Rounds.jsx
+++ b/app/webpacker/components/ScrambleMatcher/Rounds.jsx
@@ -7,7 +7,17 @@ import Groups from './Groups';
 import { events, roundTypes } from '../../lib/wca-data.js.erb';
 import { useDispatchWrapper } from './reducer';
 
-const scrambleSetToName = (scrambleSet) => `${events.byId[scrambleSet.event_id].name} ${roundTypes.byId[scrambleSet.round_type_id].name} - ${String.fromCharCode(64 + scrambleSet.scramble_set_number)}`;
+const prefixForIndex = (index) => {
+  if (index < 26) {
+    return String.fromCharCode(65 + index);
+  }
+  return (
+    prefixForIndex(Math.floor(index / 26) - 1)
+    + String.fromCharCode(65 + (index % 26))
+  );
+};
+
+const scrambleSetToName = (scrambleSet) => `${events.byId[scrambleSet.event_id].name} ${roundTypes.byId[scrambleSet.round_type_id].name} - ${prefixForIndex(scrambleSet.scramble_set_number - 1)}`;
 
 export default function Rounds({
   wcifRounds,
--- a/app/webpacker/components/ScrambleMatcher/Rounds.jsx
+++ b/app/webpacker/components/ScrambleMatcher/Rounds.jsx
@@ -8,13 +8,9 @@ import { events, roundTypes } from '../../lib/wca-data.js.erb';
 import { useDispatchWrapper } from './reducer';
 
 const prefixForIndex = (index) => {
-  if (index < 26) {
-    return String.fromCharCode(65 + index);
-  }
-  return (
-    prefixForIndex(Math.floor(index / 26) - 1)
-    + String.fromCharCode(65 + (index % 26))
-  );
+  const char = String.fromCharCode(65 + (index % 26));
+  if (index < 26) return char;
+  return prefixForIndex(Math.floor(index / 26) - 1) + char;
 };
 
 const scrambleSetToName = (scrambleSet) => `${events.byId[scrambleSet.event_id].name} ${roundTypes.byId[scrambleSet.round_type_id].name} - ${prefixForIndex(scrambleSet.scramble_set_number - 1)}`;
--- a/app/webpacker/components/ScrambleMatcher/Rounds.jsx
+++ b/app/webpacker/components/ScrambleMatcher/Rounds.jsx
@@ -8,6 +8,7 @@ import { events, roundTypes } from '../../lib/wca-data.js.erb';
 import { useDispatchWrapper } from './reducer';
 
 const scrambleSetToName = (scrambleSet) => `${events.byId[scrambleSet.event_id].name} ${roundTypes.byId[scrambleSet.round_type_id].name} - ${String.fromCharCode(64 + scrambleSet.scramble_set_number)}`;
+
 export default function Rounds({
   wcifRounds,
   matchState,
@@ -24,6 +25,7 @@ export default function Rounds({
       />
     );
   }
+
   return (
     <RoundsPicker
       wcifRounds={wcifRounds}
@@ -33,6 +35,7 @@ export default function Rounds({
     />
   );
 }
+
 function SelectedRoundPanel({
   selectedRound,
   matchState,
@@ -56,6 +59,7 @@ function SelectedRoundPanel({
     (idx) => `${activityCodeToName(selectedRound.id)}, Group ${idx + 1}`,
     [selectedRound.id],
   );
+
   return (
     <>
       <ScrambleMatch
@@ -75,6 +79,7 @@ function SelectedRoundPanel({
     </>
   );
 }
+
 function RoundsPicker({
   wcifRounds,
   matchState,
@@ -86,6 +91,7 @@ function RoundsPicker({
     () => wcifRounds.find((r) => r.id === selectedRoundId),
     [wcifRounds, selectedRoundId],
   );
+
   return (
     <>
       <Header as="h4">
--- a/app/webpacker/components/ScrambleMatcher/Rounds.jsx
+++ b/app/webpacker/components/ScrambleMatcher/Rounds.jsx
@@ -8,40 +8,84 @@ import { events, roundTypes } from '../../lib/wca-data.js.erb';
 import { useDispatchWrapper } from './reducer';
 
 const scrambleSetToName = (scrambleSet) => `${events.byId[scrambleSet.event_id].name} ${roundTypes.byId[scrambleSet.round_type_id].name} - ${String.fromCharCode(64 + scrambleSet.scramble_set_number)}`;
-
 export default function Rounds({
   wcifRounds,
   matchState,
   dispatchMatchState,
   showGroupsPicker = false,
 }) {
+  if (wcifRounds.length === 1) {
+    return (
+      <SelectedRoundPanel
+        selectedRound={wcifRounds[0]}
+        matchState={matchState}
+        dispatchMatchState={dispatchMatchState}
+        showGroupsPicker={showGroupsPicker}
+      />
+    );
+  }
+  return (
+    <RoundsPicker
+      wcifRounds={wcifRounds}
+      matchState={matchState}
+      dispatchMatchState={dispatchMatchState}
+      showGroupsPicker={showGroupsPicker}
+    />
+  );
+}
+function SelectedRoundPanel({
+  selectedRound,
+  matchState,
+  dispatchMatchState,
+  showGroupsPicker = false,
+}) {
+  const onRoundDragCompleted = useCallback(
+    (fromIndex, toIndex) => dispatchMatchState({
+      type: 'moveRoundScrambleSet',
+      roundId: selectedRound.id,
+      fromIndex,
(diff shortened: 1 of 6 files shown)


app/webpacker/components/ScrambleMatcher/ScrambleMatch.jsx
```diff
@@ -68,43 +70,44 @@ export default function ScrambleMatch({
...
+                      isDragDisabled={rowCount === 1 || hasError}
                     >
                       {(providedDraggable, snapshot) => {
                         const definitionIndex = computeOnDragIndex(index, snapshot.isDragging);
 
                         return (
                           <Ref innerRef={providedDraggable.innerRef}>
                             <Table.Row
-                              key={rowData.id}
+                              key={rowData?.id ?? `extra-scramble-set-${index + 1}`}
```
gregorbg, line 93: See above about reusing this "computed" key

app/webpacker/components/ScrambleMatcher/index.jsx
```diff
@@ -71,6 +87,34 @@ function ScrambleMatcher({ wcifEvents, competitionId, initialScrambleFiles }) {
...
       </Message>
+      { missingIds.length > 0 && (
+        <Message error>
+          <Message.Header>Missing Scramble Sets</Message.Header>
+          <Message.List>
+            { missingIds.map((id) => (
+              <Message.Item key={id}>
+                Missing scramble sets for round
+                {' '}
+                {id}
```
gregorbg, line 98: Maybe make this human-readable?

app/webpacker/components/ScrambleMatcher/index.jsx
```diff
@@ -71,6 +87,34 @@ function ScrambleMatcher({ wcifEvents, competitionId, initialScrambleFiles }) {
...
+      )}
+      { missingScrambleIds.length > 0 && (
+        <Message error>
+          <Message.Header>Missing Scrambles</Message.Header>
+          <Message.List>
+            { missingScrambleIds.map((id) => (
+              <Message.Item key={id}>
+                Missing scrambles round
+                {' '}
+                {id}
```
gregorbg, line 112: See above for human-friendly strings

app/webpacker/components/ScrambleMatcher/index.jsx
```diff
@@ -61,6 +61,22 @@ function ScrambleMatcher({ wcifEvents, competitionId, initialScrambleFiles }) {
     mutationFn: () => submitMatchedScrambles(competitionId, matchState),
   });
 
+  const roundIds = useMemo(() => _.flatMap(wcifEvents, 'rounds').map((r) => r.id), [wcifEvents]);
```
gregorbg, line 64: I'm a bit partial to using JS stdlib wherever possible
```suggestion
  const roundIds = useMemo(() => wcifEvents.flatMap((event) => event.rounds).map((r) => r.id), [wcifEvents]);
```

app/webpacker/components/ScrambleMatcher/index.jsx
```diff
@@ -61,6 +61,22 @@ function ScrambleMatcher({ wcifEvents, competitionId, initialScrambleFiles }) {
...
+
+  const missingIds = useMemo(() => {
+    if (!matchState) return [];
+    return _.difference(roundIds, _.keys(matchState));
```
gregorbg, line 68:
```suggestion
    return _.difference(roundIds, Object.keys(matchState));
```

app/webpacker/components/ScrambleMatcher/index.jsx
```diff
@@ -61,6 +61,22 @@ function ScrambleMatcher({ wcifEvents, competitionId, initialScrambleFiles }) {
     mutationFn: () => submitMatchedScrambles(competitionId, matchState),
   });
 
+  const roundIds = useMemo(() => _.flatMap(wcifEvents, 'rounds').map((r) => r.id), [wcifEvents]);
+
+  const missingIds = useMemo(() => {
+    if (!matchState) return [];
```
gregorbg, line 67: When does this happen? The initializer for the reducer uses an actual `reduce` (lol) with an empty `{}` as initial argument. Even that has a "valid" (although empty, hence the quotes) set of keys.

app/webpacker/components/ScrambleMatcher/index.jsx
```diff
@@ -61,6 +61,22 @@ function ScrambleMatcher({ wcifEvents, competitionId, initialScrambleFiles }) {
...
+    return _.difference(roundIds, _.keys(matchState));
+  }, [matchState, roundIds]);
+
+  const missingScrambleIds = useMemo(() => {
+    if (!matchState) return [];
+    return _.filter(roundIds, (roundId) => {
```
gregorbg, line 73:
```suggestion
    return roundIds.filter((roundId) => {
```

app/webpacker/components/ScrambleMatcher/index.jsx
```diff
@@ -61,6 +61,22 @@ function ScrambleMatcher({ wcifEvents, competitionId, initialScrambleFiles }) {
...
+      const matchedRound = matchState[roundId];
+      const wcifRound = _.flatMap(wcifEvents, 'rounds').find((r) => r.id === roundId);
```
gregorbg, line 75:
```suggestion
      const wcifRound = wcifEvents.flatMap((event) => event.rounds).find((r) => r.id === roundId);
```

gregorbg (changes requested):

gregorbg:
After working on the email draft, I think we should not call the button "Submit". It is too easy to confuse that with the "Results *submission* process". Something like "Save changes" (and a `primary` blue color) should be better suited for the time being.

app/webpacker/components/ScrambleMatcher/index.jsx
FinnIckler: it doesn't, but it would give an error that scrambles are missing for all events otherwise, do we want that functionality?

app/webpacker/components/ScrambleMatcher/Rounds.jsx
```diff
@@ -7,39 +7,99 @@ import Groups from './Groups';
...
-const scrambleSetToName = (scrambleSet) => `${events.byId[scrambleSet.event_id].name} ${roundTypes.byId[scrambleSet.round_type_id].name} - ${String.fromCharCode(64 + scrambleSet.scramble_set_number)}`;
+const prefixForIndex = (index) => {
+  if (index < 26) {
+    return String.fromCharCode(65 + index);
+  }
+  return (
+    prefixForIndex(Math.floor(index / 26) - 1)
+    + String.fromCharCode(65 + (index % 26))
+  );
+};
```
gregorbg, line 18: Just like in the Ruby version, I would rather you declare the `String.fromCharCode(65 + (index % 26))` as a variable and return early if below 26.

gregorbg (changes requested):

app/webpacker/components/ScrambleMatcher/index.jsx
```diff
@@ -61,6 +62,23 @@ function ScrambleMatcher({ wcifEvents, competitionId, initialScrambleFiles }) {
     mutationFn: () => submitMatchedScrambles(competitionId, matchState),
   });
 
+  const roundIds = useMemo(() => wcifEvents.flatMap((event) => event.rounds)
+    .map((r) => r.id), [wcifEvents]);
+
+  const missingIds = useMemo(() => {
```
gregorbg, line 68:
```suggestion
  const roundIdsWithoutScrambles = useMemo(() => {
```

app/webpacker/components/ScrambleMatcher/index.jsx
gregorbg: Mh, it is an efficient way of signaling to the user that "nothing has been uploaded yet".
Maybe if you rephrase the wording a little bit, or do a simple `if missingRoundIds.length === roundIds.length then display kind "please upload something" else display actual error` thingy.

But to be honest, in terms of code lines that would be at least as complicated as the "if empty return empty" special case you're handling right now. So I'm not very firm on this and leave it up to you. Just _some_ way of "hey there is no data yet why don't you upload something" would be nice

app/webpacker/components/ScrambleMatcher/index.jsx
gregorbg: (affects lots of usages below)

app/webpacker/components/ScrambleMatcher/index.jsx
```diff
@@ -61,6 +62,23 @@ function ScrambleMatcher({ wcifEvents, competitionId, initialScrambleFiles }) {
...
+    if (!matchState) return [];
+    return _.difference(roundIds, Object.keys(matchState));
+  }, [matchState, roundIds]);
+
+  const missingScrambleIds = useMemo(() => {
```
gregorbg, line 73:
```suggestion
  const roundIdsWithTooFewScrambles = useMemo(() => {
```

app/webpacker/components/ScrambleMatcher/index.jsx
```diff
@@ -71,6 +89,34 @@ function ScrambleMatcher({ wcifEvents, competitionId, initialScrambleFiles }) {
...
+          </Message.List>
+        </Message>
+      )}
+      { missingScrambleIds.length > 0 && (
+        <Message error>
+          <Message.Header>Missing Scrambles</Message.Header>
+          <Message.List>
+            { missingScrambleIds.map((id) => (
+              <Message.Item key={id}>
+                Missing scrambles round
```
gregorbg, line 112:
```suggestion
                Missing scrambles in round
```

app/webpacker/components/ScrambleMatcher/index.jsx
gregorbg: although that is quite a mouthful. If you have a better name, feel free.

app/webpacker/components/ScrambleMatcher/ScrambleMatch.jsx
```diff
@@ -68,43 +70,45 @@ export default function ScrambleMatch({
...
             <Ref innerRef={providedDroppable.innerRef}>
               <Table.Body {...providedDroppable.droppableProps}>
-                {matchableRows.map((rowData, index) => {
+                {_.times(rowCount).map((index) => {
+                  const rowData = matchableRows[index];
                   const isExpected = index < expectedNumOfRows;
                   const isExtra = !isExpected;
 
                   const hasError = isExpected && !rowData;
+                  const fallbackIndex = `extra-scramble-set-${index + 1}`;
```
gregorbg, line 79: Probably you can extract the whole `rowData?.id ?? fallback` thingy into a variable?

app/webpacker/components/ScrambleMatcher/index.jsx
```diff
@@ -61,6 +62,23 @@ function ScrambleMatcher({ wcifEvents, competitionId, initialScrambleFiles }) {
...
+    return roundIds.filter((roundId) => {
+      const matchedRound = matchState[roundId];
+      const wcifRound = wcifEvents.flatMap((event) => event.rounds).find((r) => r.id === roundId);
+      return matchedRound.length < wcifRound.scrambleSetCount;
+    });
+  }, [matchState, roundIds, wcifEvents]);
+
   return (
     <>
       <Message info>
```
gregorbg, line 84: In terms of UX, I believe this whole info message can move down into the upload panel component, and be included there as a simple `Header.Subheader` in a (slightly) condensed form.

```
Scrambles are assigned automatically upon upload. If there is a discrepancy
between the number of scramble sets in the JSON file and the number of groups
in the round you can do manual adjustments below.
```

app/webpacker/components/ScrambleMatcher/index.jsx
FinnIckler: actually, I do like just showing the user that it's all missing scramble sets when you first open it

gregorbg (approved):

PR #11620, opened 2025-05-21
ignore generational suffixes in WCA ID

pranav-027:


lib/finish_unfinished_persons.rb
```diff
@@ -108,8 +109,13 @@ def self.compute_semi_id(competition_year, person_name, available_per_semi = {})
...
 
-    last_name = name_parts[-1]
-    rest_of_name = name_parts[...-1].join
+    if name_parts.length > 1 && GENERATIONAL_SUFFIXES.include?(name_parts[-1])
+      last_name = name_parts[-2]
+      rest_of_name = name_parts[0...-2].join
+    else
+      last_name = name_parts[-1]
+      rest_of_name = name_parts[0...-1].join
+    end
```
gregorbg, line 118: The two if-branches look almost identical. I think what you're trying to do is compute an index which represents the "proper" family name, and then slice on that index. So I think you should do exactly that in code: Determine an integer (which defaults to `-1` and if the generational suffix check is true, gets set to `-2`) and then just slice once based on that index.

gregorbg (commented):
Re testing: We have a whole test suite in the `spec` folder. You are very free to write your own tests, which call `compute_semi_id` or otherwise access your new code. I am pretty sure that we already have some test cases for this flying around somewhere as well

pranav-027:
@gregorbg done with test!

spec/finish_unfinished_persons_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+  describe '.compute_semi_id' do
+    let(:competition_year) { 2023 }
+    let(:available_per_semi) { {} }
+
+    before do
+      allow(Person).to receive_messages(
+        where: Person,
+        order: Person,
+        pick: nil,
+      )
```
gregorbg, line 15: This is not how mocking during tests should work. What exactly is breaking if you remove this `before` block?

spec/finish_unfinished_persons_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+        expect(updated_available).to have_key(semi_id)
+        expect(updated_available[semi_id]).to eq(98) # 99 - 1
+      end
+    end
+
+    context 'when IDs are already taken' do
+      let(:person_name) { 'John Doe' }
+
+      before do
+        allow(Person).to receive(:pick).and_return('2023DOEJ01')
```
gregorbg, line 34: If you want the `Person` entity (which just represents the underlying database table) to return a certain value, then insert that value into the database using `FactoryBot`.

spec/finish_unfinished_persons_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+        expect(semi_id).to start_with('2023')
+        expect(updated_available[semi_id]).to be < 99
+      end
+    end
+
+    context 'when no semi-ID can be generated' do
+      let(:person_name) { 'John Doe' }
+
+      it 'raises an error' do
+        allow(Person).to receive(:pick).and_return('2023DOEJ99')
```
gregorbg, line 49: See above

gregorbg (changes requested):
You're mocking data in the tests without actually mocking data. You should use `FactoryBot` methods to create actual `Person` stubs.

Our test suite makes sure that the database is reset to a "clean" state between tests, i.e. that one dummy `Person` you're inserting for one specific test case is not present anymore when the next test case runs.

lib/finish_unfinished_persons.rb
```diff
@@ -3,6 +3,7 @@
 module FinishUnfinishedPersons
   WCA_ID_PADDING = 'U'
   WCA_QUARTER_ID_LENGTH = 4
+  GENERATIONAL_SUFFIXES = %w[JR JNR SR SNR II III IV].freeze
```
gregorbg, line 6: I am clearly not the expert in name data, but I just want to make sure from a code consistency perspective: Are you sure you want to keep only the caps-lock variants like `JR` and `JNR`?

Don't be afraid to tell me off if I'm overlooking something here, because WRT clearly has a ton more experience in the actual data processing department. But purely from a programmers "lower-case vs upper-case" perspective, I have to point out that a lot of people use `James Pickens Jr.` as their name, not `James Pickens JR`.

How are you dealing with lower-cased and partially lower-cased variants, as well as with abbreviation points at the end like `Jr.`?

spec/lib/finish_unfinished_persons_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+  describe '.compute_semi_id' do
+    let(:competition_year) { 2023 }
+    let(:available_per_semi) { {} }
+
+    context 'with a simple name' do
+      let(:person) { create(:person, name: 'John Smith') }
+
+      it 'generates a semi_id with the correct format' do
+        allow(FinishUnfinishedPersons).to receive(:extract_roman_name).with('John Smith').and_return('John Smith')
+        allow(FinishUnfinishedPersons).to receive(:remove_accents).with('John Smith').and_return('John Smith')
```
gregorbg: Why do you have to mock this behavior? Doesn't that defeat the purpose of testing that the inner workings of these methods do what you would expect them to?

spec/lib/finish_unfinished_persons_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+        expect(available_per_semi[semi_id]).to be >= 0
+      end
+    end
+
+    context 'with a generational suffix' do
+      let(:person) { create(:person, name: 'John Smith Jr') }
+
+      it 'correctly identifies the last name and generates a semi_id' do
+        allow(FinishUnfinishedPersons).to receive(:extract_roman_name).with('John Smith Jr').and_return('John Smith Jr')
+        allow(FinishUnfinishedPersons).to receive(:remove_accents).with('John Smith Jr').and_return('John Smith Jr')
```
gregorbg: (ibid)

spec/lib/finish_unfinished_persons_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+        expect(semi_id).to start_with('2023SMIT')
+      end
+    end
+
+    context 'with an extended generational suffix' do
+      let(:person) { create(:person, name: 'John Smith Jnr') }
+
+      it 'correctly identifies the last name for JNR' do
+        allow(FinishUnfinishedPersons).to receive(:extract_roman_name).with('John Smith Jnr').and_return('John Smith Jnr')
+        allow(FinishUnfinishedPersons).to receive(:remove_accents).with('John Smith Jnr').and_return('John Smith Jnr')
```
gregorbg: (ibid)

spec/lib/finish_unfinished_persons_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+        expect(semi_id).to start_with('2023SMIT')
+      end
+    end
+
+    context 'with accented characters' do
+      let(:person) { create(:person, name: 'José García') }
+
+      it 'handles accented characters correctly' do
+        allow(FinishUnfinishedPersons).to receive(:extract_roman_name).with('José García').and_return('Jose Garcia')
+        allow(FinishUnfinishedPersons).to receive(:remove_accents).with('Jose Garcia').and_return('Jose Garcia')
```
gregorbg: (ibid)

spec/lib/finish_unfinished_persons_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+        expect(semi_id).to start_with('2023GARC')
+      end
+    end
+
+    context 'with a single name part' do
+      let(:person) { create(:person, name: 'Madonna') }
+
+      it 'uses the single name as the last name' do
+        allow(FinishUnfinishedPersons).to receive(:extract_roman_name).with('Madonna').and_return('Madonna')
+        allow(FinishUnfinishedPersons).to receive(:remove_accents).with('Madonna').and_return('Madonna')
```
gregorbg: (ibid)

spec/lib/finish_unfinished_persons_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+        expect(semi_id).to start_with('2023MADO')
+      end
+    end
+
+    context 'with a name containing special characters' do
+      let(:person) { create(:person, name: 'John O\'Connor') }
+
+      it 'sanitizes special characters' do
+        allow(FinishUnfinishedPersons).to receive(:extract_roman_name).with('John O\'Connor').and_return('John OConnor')
+        allow(FinishUnfinishedPersons).to receive(:remove_accents).with('John OConnor').and_return('John OConnor')
```
gregorbg: (ibid)

gregorbg (changes requested):

lib/finish_unfinished_persons.rb
pranav-027: @gregorbg My initial exclusion list was `GENERATIONAL_SUFFIXES = %w[Jr Sr Jr. Sr. II III IV].freeze` only. 

But thankfully :) , after my test failures, I discovered the code on line 110:

```ruby
name_parts = sanitized_roman_name.gsub(/[^a-zA-Z ]/, '').upcase.split
```
This line removes special characters and converts the name to uppercase. So, during generational suffix comparison, `James Pickens Jr.` or `James Pickens JR` will both become `JAMES PICKENS JR` at the time of `last_name` computation

Hope this helps!

spec/lib/finish_unfinished_persons_spec.rb
pranav-027: I have removed the mocks

spec/lib/finish_unfinished_persons_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+        semi_id, = FinishUnfinishedPersons.compute_semi_id(competition_year, person.name, available_per_semi)
+
+        expect(semi_id).to match(/\A2023[A-Z]{4}\z/)
+        expect(semi_id).to start_with('2023SMIT')
+        expect(available_per_semi[semi_id]).to be >= 0
+      end
+    end
+
+    context 'with a generational suffix' do
+      let(:person) { create(:person, name: 'John Smith Jr.') }
```
pranav-027, line 23: @gregorbg test case for your concerns 😁

PR #11623, opened 2025-05-23
Create Records Table

FinnIckler:
Still working on scopes, integrating with results, creating the new check records script and more. But opening this early to get database model feedback.

### DIFF
--- a/app/models/round.rb
+++ b/app/models/round.rb
@@ -243,6 +243,19 @@ class Round < ApplicationRecord
     "#{event.id}-r#{self.number}"
   end
 
+  def end_time
+    schedule.minimum(:end_time)
+  end
+
+  def schedule
+    ScheduleActivity
+      .where(holder_type: 'VenueRoom')
+      .joins("INNER JOIN venue_rooms ON venue_rooms.id = schedule_activities.holder_id")
+      .joins("INNER JOIN competition_venues ON competition_venues.id = venue_rooms.competition_venue_id")
+      .where("competition_venues.competition_id = ?", competition_id)
+      .where("activity_code LIKE ?", "#{wcif_id}%")
+  end
+
   def to_string_map(short: false)
     {
       wcif_id: wcif_id,
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.2].define(version: 2025_05_20_093305) do
+ActiveRecord::Schema[7.2].define(version: 2025_05_23_115703) do
   create_table "active_storage_attachments", charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
     t.string "name", null: false
     t.string "record_type", null: false
@@ -952,6 +952,23 @@ ActiveRecord::Schema[7.2].define(version: 2025_05_20_093305) do
     t.integer "country_rank", default: 0, null: false
     t.index ["event_id"], name: "fk_events"
     t.index ["person_id"], name: "fk_persons"
+  end
+
+  create_table "records", charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
+    t.string "record_type", null: false
+    t.bigint "result_id", null: false
+    t.integer "value", null: false
+    t.string "event_id", null: false
+    t.string "country_id"
+    t.string "continent_id"
+    t.string "record_scope", null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["country_id", "record_scope"], name: "index_records_on_country_id_and_record_scope"
+    t.index ["event_id", "record_scope"], name: "index_records_on_event_id_and_record_scope"
+    t.index ["event_id", "record_type", "record_scope"], name: "index_records_on_event_id_and_record_type_and_record_scope", unique: true
+    t.index ["record_scope"], name: "index_records_on_record_scope"
+    t.index ["result_id"], name: "index_records_on_result_id"
   end
 
   create_table "regional_organizations", charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -780,21 +780,6 @@ ActiveRecord::Schema[7.2].define(version: 2025_05_23_115703) do
     t.datetime "updated_at", precision: nil, null: false
   end
 
-  create_table "manual_payment_integrations", charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
-    t.string "payment_reference", null: false
-    t.text "payment_information", null: false
-    t.datetime "created_at", null: false
-    t.datetime "updated_at", null: false
-  end
-
-  create_table "manual_payment_records", charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
-    t.string "payment_reference"
-    t.integer "amount_iso_denomination", null: false
-    t.string "currency_code", null: false
-    t.datetime "created_at", null: false
-    t.datetime "updated_at", null: false
-  end
-
   create_table "oauth_access_grants", id: :integer, charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
     t.integer "resource_owner_id", null: false
     t.integer "application_id", null: false
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -982,7 +982,7 @@ ActiveRecord::Schema[7.2].define(version: 2025_05_23_115703) do
     t.datetime "updated_at", null: false
     t.index ["country_id", "record_scope"], name: "index_records_on_country_id_and_record_scope"
     t.index ["event_id", "record_scope"], name: "index_records_on_event_id_and_record_scope"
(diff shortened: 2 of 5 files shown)


FinnIckler:
For scopes like 
```
scope :national_records, lambda { |country_id = nil|
    scope = where(record_scope: 'NR')
    country_id.present? ? scope.where(country_id: country_id) : scope
  }
```
we need to decide if we we want to 
- actually only get the 'pure' NRs or
- also want the CRs and WRs that are National Records.

We probably have use cases for both? Not sure

lib/check_regional_records.rb
```diff
@@ -126,6 +126,60 @@ def self.load_ordered_results(event_id, competition_id, value_column, regional_r
...
+                                                      PARTITION BY results.event_id, result_timestamps.round_timestamp
+                                                      ORDER BY results.#{value_column}
+                                                    ) AS best_rank
+                                                  FROM results
+                                                  INNER JOIN result_timestamps ON result_timestamps.result_id = results.id
+                                                  WHERE results.#{value_column} > 0
+                                                  #{event_id.present? ? "AND results.event_id = '#{event_id}'" : ''}
+                                                  #{from_timestamp.present? ? "AND result_timestamps.round_timestamp >= '#{from_timestamp}'" : ''}
+                                                  #{to_timestamp.present? ? "AND result_timestamps.round_timestamp <= '#{to_timestamp}'" : ''}
+                                                )
```
gregorbg, line 165: I think you will be able to push runtimes down by a lot if you pre-merge all of this information into `result_timestamp`.

This will effectively mean that the code you're running here is not 100% live (it is only as recent as the state of the precomputed `results_timestamp` table is) but (a) WRT is already used to that with the existing `regional_records_lookup` and (b) the potential performance gains are worth it IMO.

app/models/regional_record.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  CONTINENT_TO_RECORD_MARKER = {
+    '_Africa' => 'AfR',
+    '_Europe' => 'ER',
+    '_North America' => 'NAR',
+    '_South America' => 'SAR',
+    '_Asia' => 'AsR',
+    '_Oceania' => 'OcR',
+  }.freeze
```
gregorbg, line 82: We're storing this in the database already, there's a `record_name` column in the `continents` table that you can use.

Since we're already quite conservative about caching countries, continents, events etc. in-memory, it will be very cheap (and less redundant!) to use `Continent.c_find(continent_id).record_name` instead.

app/models/round.rb
```diff
@@ -243,6 +243,14 @@ def wcif_id
     "#{event.id}-r#{self.number}"
   end
 
+  def end_time
+    schedule.maximum(:end_time)
+  end
```
gregorbg, line 248: Is this actually being used anywhere?

gregorbg (commented):
I haven't looked at the core algorithm (`check_records_new_table`) in detail yet, but here are already some remarks about the different SQL queries, code reuse, etc.

lib/tasks/result_timestamps.rake
```diff
@@ -0,0 +1,38 @@
...
+          FROM schedule_activities
+          WHERE schedule_activities.round_id = results.round_id
+        ) AS round_timestamp,
+        NOW(),
+        NOW()
+      FROM results WHERE round_id IS NOT NULL
+      ON DUPLICATE KEY UPDATE
+        round_timestamp = VALUES(round_timestamp),
+        updated_at = VALUES(updated_at)
+    SQL
```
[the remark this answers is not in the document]
gregorbg: Suggestion:
```sql
WITH max_times_per_round AS (
    SELECT round_id, MAX(end_time) AS max_end_time
    FROM schedule_activities
    GROUP BY round_id
)
SELECT
    r.id,
    COALESCE(m.max_end_time, c.end_date),
    NOW(),
    NOW()
FROM results r
         LEFT JOIN max_times_per_round m ON m.round_id = r.round_id
         LEFT JOIN competitions c ON c.id = r.competition_id;
```

app/controllers/admin_controller.rb
```diff
@@ -240,7 +240,7 @@ def override_regional_records
                           .permit(:competition_id, :event_id, :refresh_index)
 
     @check_records_request = CheckRegionalRecordsForm.new(action_params)
-    @check_results = @check_records_request.run_check
+    @check_results = @check_records_request.run_check_new
```
gregorbg, line 243: This is okay while testing but should obviously not go live when merging

app/models/regional_record.rb
```diff
@@ -0,0 +1,90 @@
...
+
+class RegionalRecord < ApplicationRecord
+  enum :record_scope, {
+    NR: 0,
+    CR: 1,
+    WR: 2,
+  }
+
+  RECORD_SCOPES = %w[WR CR NR].freeze
+  RECORD_TYPES = %w[single average].freeze
```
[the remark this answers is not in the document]
gregorbg: I think this can be worked around by using `prefix: true`. That way, instead of `RegionalRecord.average` the scope will be called `RegionalRecord.record_type_average`

app/models/regional_record.rb
```diff
@@ -82,9 +81,13 @@ def self.is_record_at_date?(value, event_id, record_type, country_id, timestamp)
...
-
-    'WR'
+    case record_scope
+    when :continental
+      CONTINENT_TO_RECORD_MARKER[continent_id]
+    when :world
+      "WR"
+    when :national
+      "NR"
+    end
```
gregorbg, line 91: Cool! Nitpick: The order of the case branches should match the "natural" order of the enum

db/schema.rb
```diff
@@ -953,6 +952,24 @@
     t.index ["person_id"], name: "fk_persons"
   end
 
+  create_table "records", charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
```
gregorbg, line 955: The table seems to have "moved" in the schema definition file, because the name mysteriously changed?!

db/schema.rb
gregorbg, line 1: Several changes in this file seem like merge whoopsies

db/schema.rb
```diff
@@ -674,8 +674,8 @@
     t.string "event_id", null: false
     t.integer "round_number", null: false
     t.integer "scramble_set_number", null: false
-    t.integer "ordered_index", null: false
```
gregorbg, line 677: (merge whoopsie?)

db/schema.rb
```diff
@@ -674,8 +674,8 @@
...
     t.integer "matched_round_id"
+    t.integer "matched_round_ordered_index"
```
gregorbg, line 678: (merge whoopsie?)

db/schema.rb
```diff
@@ -691,7 +691,6 @@
     t.bigint "inbox_scramble_set_id", null: false
     t.boolean "is_extra", default: false, null: false
     t.integer "scramble_number", null: false
-    t.integer "ordered_index", null: false
```
gregorbg, line 694: (merge whoopsie?)

db/schema.rb
```diff
@@ -1049,7 +1048,7 @@
 
   create_table "registrations", id: { type: :integer, unsigned: true }, charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
     t.string "competition_id", limit: 32, default: "", null: false
-    t.integer "registrant_id", null: false
+    t.integer "registrant_id"
```
gregorbg, line 1051: (merge whoopsie?)

lib/check_regional_records.rb
```diff
@@ -147,33 +147,39 @@ def self.check_records_new_table(event_id = nil, competition_id = nil, from_time
...
-          } if is_record
+          if is_record
+            result = Result.includes(:competition).find(result_id)
+            {
+              computed_marker: computed_marker,
+              competition: result.competition,
+              result:result,
+            }
+          end
+        rescue Exception => e
```
gregorbg, line 179: Hm... Fine for debugging / temporary development, but obviously stuff like that cannot make it to production. Why can errors occur here in the first place, and can you do something in the checking code to make sure that the source of the error gets fixed?

lib/tasks/fill_record_table.rake
```diff
@@ -9,7 +9,7 @@ namespace :records do
     ].each do |records|
       Result::MARKERS.compact.each do |marker|
         is_cr = RegionalRecord::CONTINENT_TO_RECORD_MARKER.value?(marker)
-        record_scope = is_cr ? :CR : marker.to_sym
+        record_scope = is_cr ? :continental : { "WR" => :world, "NR" => :national }[marker]
```
gregorbg, line 12: This mapping of "human-readable tags" to enum constant values could be moved into a constant

gregorbg (changes requested):

db/migrate/20250523115703_add_record_computation_tables.rb
```diff
@@ -0,0 +1,27 @@
...
+      t.date :record_timestamp, null: false
+      t.integer :record_scope, null: false, index: true
+      t.index %i[event_id record_type record_scope]
+      t.index %i[event_id record_scope]
+      t.index %i[country_id record_scope]
+      t.timestamps
+    end
+
+    create_table :result_timestamps do |t|
+      t.references :result, null: false, index: { unique: true }
```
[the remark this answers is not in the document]
FinnIckler: will this not cause a super long migration again?

PR #11645, opened 2025-05-27
Add `registrant_id` to Registration model (& backfill)

dunkOnIT:


app/models/registration.rb
```diff
@@ -46,6 +46,14 @@ class Registration < ApplicationRecord
 
   serialize :roles, coder: YAML
 
+  before_create -> { self.registrant_id = competition.registrations.count + 1 }
```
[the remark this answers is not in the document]
gregorbg: If, for some reason, we import external `registrant_id`s, these should be validated as-is instead of being overwritten

app/models/registration.rb
```diff
@@ -46,6 +46,14 @@ class Registration < ApplicationRecord
...
+
+  validate :unique_registrant_id
+  private def unique_registrant_id
+    errors.add(:registrant_id, message: "A competition cannot have duplicate registrant_id's") if
+      self.competition.registrations.where(registrant_id: registrant_id).where.not(id: self.id).count > 0
+  end
```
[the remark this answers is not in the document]
gregorbg: Rails has a built-in validator for that

lib/tasks/populate_registrant_id.rake
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+namespace :registrant_id do
+  desc 'Backfills all registration.registrant_id values with their WCIF-generated registrant_id'
+  task :populate do
+    Competition.all.each do |comp|
+      puts "Populating registrant ID for #{comp.id}"
+      changes = comp.persons_wcif.map do |person|
```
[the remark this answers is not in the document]
gregorbg: That automatically kicks out all iterations that return `nil` due to the `if person['registration'].present?` check.

spec/models/registration_spec.rb
```diff
@@ -1307,4 +1307,21 @@
...
     end
   end
+
+  describe 'registrant_id' do
+    it 'populates upon create' do
+      expect(registration.registrant_id).to eq(1)
+    end
+
+    it 'allows registrations belonging to different competitions to have the same registrant id' do
+      expect(create(:registration)).to be_valid
```
gregorbg, line 1317: How is this ensured to _definitely_ be a different competition? Different from what?
(Reading the title, you should probably add assumptions that the two `registrant_id`s are indeed equal, and that the competition IDs are not equal)

lib/tasks/populate_registrant_id.rake
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+namespace :registrant_id do
+  desc 'Backfills all registration.registrant_id values with their WCIF-generated registrant_id'
+  task :populate do
```
gregorbg, line 5: More like `backfill`?

gregorbg, line 5: Also, this currently doesn't call the Rails classloader, meaning all models are not autoloaded. I wonder how this ever worked on your machine :stuck_out_tongue_winking_eye:

lib/tasks/populate_registrant_id.rake
```diff
@@ -0,0 +1,28 @@
...
+    Competition.all.each do |comp|
+      puts "Populating registrant ID for #{comp.id}"
+      changes = comp.persons_wcif.map do |person|
+        { id: person['registration']['wcaRegistrationId'], registrant_id: person['registrantId'] } if
+          person['registration'].present?
+      end
+
+      ids = changes.map { it[:id] }
+      existing_ids = Registration.where(id: ids).pluck(:id)
```
gregorbg, line 14:
```suggestion
      existing_ids = Registration.where(id: ids).ids
```

gregorbg (changes requested):
Lots of small change requests throughout the code.

Regarding `upsert_all`: This is a SQL problem. not a Rails problem. You're doing everything correctly from a Rails perspective. However, the `upsert_all` creates an `INSERT INTO ... ON DUPLICATE KEY UPDATE` SQL statement, which will definitely try an `INSERT` first, no matter what. There is no "check for key collision first, and then update or else insert" logic. There is only "Definitely try insert, then if the inserted key collides update instead" logic on an SQL level. So unfortunately you cannot use `upsert_all` here.

The best way in my opinion would be to use the models you're loading for the IDs, and instead of "just" taking the IDs and calling it a day, loop over the models to trigger updates one by one. This can then be wrapped in a transaction per competition to improve performance.

app/models/registration.rb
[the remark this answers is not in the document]
dunkOnIT: Ah I didn't realize we could change the scope of uniqueness!

lib/tasks/populate_registrant_id.rake
```diff
@@ -0,0 +1,24 @@
...
+namespace :registrant_id do
+  desc 'Backfills all registration.registrant_id values with their WCIF-generated registrant_id'
+  task :populate do
+    Competition.in_batches(of: 30) do |batch|
+      ActiveRecord::Base.connection.transaction do
+        batch.each do |comp|
+          puts "Populating registrant ID for #{comp.id}"
+
+          changes = comp.persons_wcif(authorized: true).filter_map do |person|
+            next if person['registration'].nil?
```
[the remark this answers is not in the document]
dunkOnIT: I know that this behaviour is definitely necessary, as we have `persons` in the WCIF who don't have registrations (for example, unregistered organizers/delegates)

This approach may be dangerous, but my thought is that we can pretty easily check the db for `null` entries after the rake task runs. We can also run this on staging beforehand to do a `null` check before hitting prod with it[1]

If you have ideas for a good alternate behaviour I'm all ears, but my suspicion is that if we ended up with `null` entries we would need to debug how they ended up being `null` anyway

[1] Although even if we do it on prod and realize there is an issue that needs to be correct, we can pretty harmlessly reset these values to `null`

lib/tasks/populate_registrant_id.rake
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+namespace :registrant_id do
+  desc 'Backfills all registration.registrant_id values with their WCIF-generated registrant_id'
+  task :populate do
+    Competition.all.each do |comp|
```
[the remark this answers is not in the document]
dunkOnIT: Noted, thanks!

lib/tasks/populate_registrant_id.rake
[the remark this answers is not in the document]
dunkOnIT: Done

db/migrate/20250526141514_add_registrant_id_to_registrations.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddRegistrantIdToRegistrations < ActiveRecord::Migration[7.2]
+  def change
+    add_column :registrations, :registrant_id, :integer, after: :id
```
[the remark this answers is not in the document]
dunkOnIT: I think after `competition_id` makes the most sense, given that `registrant_id` is unique per competition - this makes it very easily to visually scan if we're ever looking at the results of raw database queries (ie, most times I look at registrant_id in the db I'd want to know "registrant_id for which comp?")

Comparatively, user_id isn't as interesting/important so I don't think registrant_id needs to be directly next to it

app/models/registration.rb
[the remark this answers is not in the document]
dunkOnIT: Changed

lib/tasks/populate_registrant_id.rake
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+namespace :registrant_id do
+  desc 'Backfills all registration.registrant_id values with their WCIF-generated registrant_id'
+  task :populate do
+    Competition.in_batches(of: 30) do |batch|
```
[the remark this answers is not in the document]
dunkOnIT: Pretty much - working on the assumptions that:
-  we want maybe 5k records in memory at a time (based purely on vibes and a brief conversation with ChatGPT). 
- each comp has ~200 attendees on average (I just checked and actually this is more like 66 attendees on average - ~1m rows / ~15k competitions - so we could theoretically take this to 50-60 if we agree with my first assumption about targeting ~5k rows

db/migrate/20250526141514_add_registrant_id_to_registrations.rb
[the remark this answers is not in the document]
gregorbg: Agreed!

lib/tasks/populate_registrant_id.rake
gregorbg: That's probably it. Athough I'm not too knowledgeable on the specifics of Spring, but it seems like that's the only possible explanation

app/models/registration.rb
```diff
@@ -46,16 +46,19 @@ class Registration < ApplicationRecord
 
   serialize :roles, coder: YAML
 
+  # Set a `registered_at` timestamp for newly created records,
+  #   but only if there is no value already specified from the outside
+  after_initialize :mark_registered_at, if: :new_record?, unless: :registered_at?
```
gregorbg, line 51: Why did you move this here?

app/models/registration.rb
```diff
@@ -46,16 +46,19 @@ class Registration < ApplicationRecord
...
+  before_create -> { self.registrant_id ||= competition.registrations.count + 1 }
+  validates :registrant_id, uniqueness: { scope: :competition_id }
```
gregorbg, line 53: Can this validation actually screw us over as long as the values are not backfilled yet?

gregorbg: I mean, during the very initial deployment, it will see a bunch of `NULL`s and probably freak out

app_secrets.rb
```diff
@@ -33,7 +33,7 @@ def vault_file(secret_name, file_path, refresh: true)
 
   private def vault_read(secret_name)
     Vault.with_retries(Vault::HTTPConnectionError, Vault::HTTPError) do |attempt, e|
-      puts "Received exception #{e} from Vault - attempt #{attempt}" if e.present?
+      Rails.logger.debug { "Received exception #{e} from Vault - attempt #{attempt}" } if e.present?
```
gregorbg, line 36: Unrelated change

spec/jobs/wca_cronjob_spec.rb
gregorbg, line 1: All unrelated changes

gregorbg (changes requested):

spec/jobs/wca_cronjob_spec.rb
dunkOnIT: Is it possible this somehow crept in from the merge with main? I can guarantee I didn't make these changes, nor do I have anything like it in my local. Happy to remove though

app/models/registration.rb
dunkOnIT: Same comment as elsewhere - the only thing I can think is the merge with main, but then I don't understand why it would be showing up in the diff

app/models/registration.rb
dunkOnIT: All I can say is that it works locally, but I'm happy to remove it/put it behind an environment flag until we're done with backfilling

app_secrets.rb
dunkOnIT: Rubocop??

spec/jobs/wca_cronjob_spec.rb
dunkOnIT, line 1: Rubocop??

app/models/registration.rb
gregorbg: Okay, this one is actually the only one that is not a choke: RuboCop wants `before_create` (which was not present in this class before!) to be placed after `after_initialize`.

Why did you choose to move the existing `after_initialize` up, rather than moving the new `before_create` down?

gregorbg (changes requested):
Have this idea for a fully one-off bulk query. Local testing looked like it does its thing, can you please cross-check? @FinnIckler

FinnIckler:
Just ran it and checked against a competition wcif and it looks correct!

app/models/registration.rb
dunkOnIT: That was Rubocop's autocorrect solution - but I've moved the `after_create` down now instead

PR #11658, opened 2025-05-29
Remove WIC from Delegate Demotion Mailer

slongbehn:
WIC does not need demotion emails. Removing that mailer but keeping all the promotions.


slongbehn:
@pranav-027 Clearly above my head :) Would you mind taking a look? Is it too complicated to remove us from demotions (if I understand the error correctly)?

PR #11661, opened 2025-05-30
React to incoming Stripe refunds in webhook

gregorbg:
Had this old PR laying around. Pushing here so that @dunkOnIT can pick it up.


app/controllers/registrations_controller.rb
```diff
@@ -307,10 +307,12 @@ def stripe_webhook
       logger.warn "Stripe webhook error while parsing basic request. #{e.message}"
       return head :bad_request
     end
+
     # Check if webhook signing is configured.
     if AppSecrets.STRIPE_WEBHOOK_SECRET.present?
       # Retrieve the event by verifying the signature using the raw body and secret.
       signature = request.env['HTTP_STRIPE_SIGNATURE']
+
```
dunkOnIT, line 315: I put these line breaks in kind've instinctively as they seem more readable, can remove

app/controllers/registrations_controller.rb
```diff
@@ -327,25 +329,27 @@ def stripe_webhook
...
     audit_event = StripeWebhookEvent.create_from_api(event)
     audit_remote_timestamp = audit_event.created_at_remote
+    connected_account = ConnectedStripeAccount.find_by(account_id: audit_event.account_id)
+
+    if connected_account.nil?
+      logger.error "Stripe webhook reported event for account '#{audit_event.account_id}' but we are not connected to that account."
+      return head :not_found
+    end
 
     stripe_intent = event.data.object # contains a polymorphic type that depends on the event
```
gregorbg: Rename to something like `stripe_event_payload`

gregorbg: Or `event_data` or whatever, I am open to whatever you think works best. Just no specific vocabulary like `intent`

app/controllers/registrations_controller.rb
```diff
@@ -327,25 +329,27 @@ def stripe_webhook
...
     stored_record = StripeRecord.find_by(stripe_id: stripe_intent.id)
 
     handling_event = StripeWebhookEvent::HANDLED_EVENTS.include?(event.type)
+    incoming_event = StripeWebhookEvent::INCOMING_EVENTS.include?(event.type)
+
+    stored_record ||= StripeRecord.create_from_api(stripe_intent, {}, audit_event.account_id) if incoming_event
 
     if stored_record.nil?
-      logger.error "Stripe webhook reported event on entity #{stripe_intent.id} but we have no matching transaction."
+      logger.error "Stripe webhook reported event on entity #{stripe_intent.id} but we have no matching charge."
```
gregorbg: Rename back to `transaction`, or perhaps `record`

app/controllers/registrations_controller.rb
```diff
@@ -378,6 +382,53 @@ def stripe_webhook
...
+        stored_record.destroy
+
+        logger.error "Stripe webhook reported a refund on charge #{stripe_intent.charge} but we never issued that one"
+        return head :not_found
+      elsif stored_record.parent_record.nil?
+        stored_record.update!(parent_record: original_charge)
+      end
+
+      # NOTE: Feels weird to be checking the type of the event in this block - is there a better way to handle this logic?
+      stored_record.update_from_api(stripe_intent) if event.type == StripeWebhookEvent::REFUND_UPDATED
```
gregorbg: Turn into a `create_or_update` above

app/controllers/registrations_controller.rb
```diff
@@ -378,6 +382,53 @@ def stripe_webhook
...
+      original_charge = StripeRecord.charge.find_by(stripe_id: stripe_intent.charge)
+
+      if original_charge.nil?
+        # We created a record because refund creation is an incoming event,
+        #   but now we figured out that we don't care about it after all.
+        stored_record.destroy
+
+        logger.error "Stripe webhook reported a refund on charge #{stripe_intent.charge} but we never issued that one"
+        return head :not_found
+      elsif stored_record.parent_record.nil?
```
gregorbg: Kinda redundant `if` (Rails is smart enough to skip redundant updates internally)

app/controllers/registrations_controller.rb
```diff
@@ -378,6 +382,53 @@ def stripe_webhook
...
+
+        if original_payment.nil?
+          logger.error "Tried to record refund on charge #{stripe_intent.charge} but we never issued that one"
+          return head :not_found
+        end
+
+        stored_holder.with_lock do
+          already_refunded = stored_record.registration_payment.present?
+
+          refund_succeeded = stored_record.stripe_status_succeeded?
```
gregorbg: This check specifically can happen outside of the lock

app/controllers/registrations_controller.rb
```diff
@@ -516,14 +567,21 @@ def refund_payment
...
-      ruby_money.cents,
-      ruby_money.currency.iso_code,
-      refund_receipt,
-      payment_record.registration_payment.id,
-      current_user.id,
-    )
+    original_payment = payment_record.registration_payment
+
+    registration.with_lock do
+      already_refunded = original_payment.refunding_registration_payments.any?(receipt: payment_record)
```
gregorbg:
```suggestion
      already_refunded = original_payment.refunding_registration_payments.where(receipt: refund_receipt).any?
```

app/models/stripe_record.rb
```diff
@@ -182,6 +182,17 @@ def self.amount_to_ruby(amount_stripe_denomination, iso_currency)
     amount_stripe_denomination.to_i
   end
 
+  # We dont update parameters, account_id or parent record - none of that should be change by receiving an UPDATE to a given refund
+  def update_from_api(api_record)
```
gregorbg: Merge with `create_from_api` below by using `find_or_create_by!`

spec/factories/payment_intents.rb
```diff
@@ -16,7 +16,7 @@
     trait :confirmed do
       confirmed_at { DateTime.now }
       wca_status { 'succeeded' }
-      payment_record { FactoryBot.create(:stripe_record, :payment_intent, stripe_status: 'succeeded') }
+      payment_record { FactoryBot.create(:stripe_record, :successful_pi) }
```
gregorbg:
```suggestion
      payment_record { association(:stripe_record, :successful_pi) }
```

spec/factories/registration_payments.rb
```diff
@@ -5,13 +5,23 @@
     transient do
       skip_auto_accept_hook { false }
       competition { registration.competition }
-      revert_mode_to { nil }
+      payment_intent { create(:payment_intent, :confirmed, holder: registration) }
```
gregorbg:
```suggestion
      payment_intent { association(:payment_intent, :confirmed, holder: registration) }
```

spec/factories/registration_payments.rb
```diff
@@ -5,13 +5,23 @@
...
     end
 
-    registration { nil }
+    registration do
+      refunded_registration_payment.present? ? refunded_registration_payment.registration : nil
```
gregorbg:
```suggestion
      refunded_registration_payment&.registration
```

spec/factories/registration_payments.rb
```diff
@@ -5,13 +5,23 @@
...
     user_id { registration&.user_id }
     amount_lowest_denomination { competition&.base_entry_fee_lowest_denomination }
     currency_code { competition&.currency_code }
+    refunded_registration_payment { nil }
+    receipt do
+      if refunded_registration_payment.present?
+        create(:stripe_record, :refund, parent_record: refunded_registration_payment.receipt)
+      else
+        payment_intent.payment_record.child_records.first
+      end
```
gregorbg:
```suggestion
      if refunded_registration_payment.present?
        association(:stripe_record, :refund, parent_record: refunded_registration_payment.receipt)
      else
        payment_intent.payment_record.child_records.first
      end
```

spec/factories/stripe_records.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :stripe_record do
+    transient do
+      create_child { false }
```
gregorbg: Either better name `create_child_charge` or more flexibility (default `nil`, pass in a trait like `:charge`, `:refund`, etc. on-demand)

app/models/stripe_record.rb
dunkOnIT: Opted not to use `find_or_create_by` as it doesn't _seem_ to give me any real advantages - we still want to use different logic for create/update - at least, that seems easier than needing to handle the presence/absence of `parameters`, `account_id` and `parent_record` (which are only needed upon create)

app/controllers/registrations_controller.rb
dunkOnIT: Fixed

app/controllers/registrations_controller.rb
dunkOnIT: Fixed

app/controllers/registrations_controller.rb
dunkOnIT: Fixed

app/controllers/registrations_controller.rb
dunkOnIT: Fixed

app/controllers/registrations_controller.rb
dunkOnIT: Fixed

spec/factories/payment_intents.rb
dunkOnIT: Fixed

spec/factories/registration_payments.rb
dunkOnIT: Fixed

spec/factories/registration_payments.rb
dunkOnIT: Fixed, now appears on line 18

spec/factories/registration_payments.rb
dunkOnIT: Fixed, now appears on line 20

spec/factories/stripe_records.rb
dunkOnIT: Fixed

app/models/stripe_record.rb
```diff
@@ -182,17 +182,32 @@ def self.amount_to_ruby(amount_stripe_denomination, iso_currency)
...
+        stripe_id: api_record.id,
+        amount_stripe_denomination: api_record.amount,
+        currency_code: api_record.currency,
+        stripe_status: api_record.status,
+        account_id: account_id,
+        parent_record: parent_record,
+      )
+    end
+
+    new_record || existing_record
```
gregorbg: This still feels a bit too "manual". I think a more idiomatic way would be to leverage the `create_or_initialize_by` helper, something along the lines of:
```suggestion
  def self.create_or_update_from_api(api_record, parameters = nil, account_id = nil, parent_record = nil)
    record = StripeRecord.find_or_initialize_by(stripe_id: api_record.id) do |new_record|
      new_record.parameters = parameters
      new_record.account_id = account_id
      new_record.parent_record = parent_record
    end.tap do |record|
      record.update!(
        stripe_record_type: api_record.object,
        stripe_id: api_record.id,
        amount_stripe_denomination: api_record.amount,
        currency_code: api_record.currency,
        stripe_status: api_record.status,
      )
    end
```

gregorbg: Note that I wrote this freestyle directly in the GH comment panel, you should definitely try whether it works as intended.

Per the documentation of `find_or_initialize_by`, it takes one block as optional argument, which is only executed in the "initialize" case.
The other `tap` could be replaced by
```ruby
record = find_or_initialize_by ...

record.update!(...)

record # as explicit return value
```
but I think since it's only one single `update!` operation on the very variable that we want as return value anyways, tapping is warranted and readable.

app/models/stripe_record.rb
```diff
@@ -27,7 +27,7 @@ class StripeRecord < ApplicationRecord
     legacy_payment_intent_registered: "payment_intent_registered",
     legacy_success: "success",
     legacy_failure: "failure",
-  }
+  }, prefix: true
```
gregorbg: Have you made certain that this doesn't break other places in the code? From the top of my head, I am mostly concerned about `ConnectedStripeAccount` and `PaymentIntent` models

gregorbg (commented):
Cannot click "Request Changes" because I originally opened this PR, but please consider my comments as such.

app/models/stripe_record.rb
dunkOnIT: Removed this, I didn't realize that it would cause the non-prefixed methods to no longer work (but it makes sense)

app/models/stripe_record.rb
dunkOnIT: Altered as suggested, ran the relevant tests locally and it all looks good, we'll see what the CI says.

Also notable that I removed `stripe_record_type: api_record.object,` from the `update` portion - as in my mind, we should only be touching that parameter upon create

app/models/stripe_record.rb
```diff
@@ -182,17 +182,20 @@ def self.amount_to_ruby(amount_stripe_denomination, iso_currency)
...
-      stripe_status: api_record.status,
-      account_id: account_id,
-      parent_record: parent_record,
-    )
+  def self.create_or_update_from_api(api_record, parameters = nil, account_id = nil, parent_record = nil)
+    StripeRecord.find_or_initialize_by(stripe_id: api_record.id) do |new_record|
+      new_record.parameters = parameters
+      new_record.account_id = account_id
+      new_record.parent_record = parent_record
+      new_record.stripe_record_type = api_record.object
```
gregorbg: Hypothetically speaking, could we move `account_id` and `stripe_record_type` directly into the `find` call's parentheses parameters?

gregorbg (commented):
I want to approve but GH doesn't let me because it's technically "my" PR.
Please feel free to approve yourself (LOL this is exactly the opposite of what GitHub wanted to achieve with that "you cannot approve your own PRs" feature...)

The one comment that I have is only cosmetic and non-binding.

app/models/stripe_record.rb
dunkOnIT: Oooh interesting, let me try!

dunkOnIT (approved):
image

Issue #11651, opened 2025-05-28
Competitors on waiting list of SICC2025 are not able to edit their registration


pranav-027:
**Describe the bug**
After this https://github.com/thewca/worldcubeassociation.org/pull/11410. Competitors on the waiting list can edit their registrations when `Competitors may NOT edit their registration once it has been accepted` is set. But for the competitions [SICC2025](https://www.worldcubeassociation.org/competitions/SICC2025) competitors are getting `event update deadline has already passed` error while trying to update the events

**To Reproduce**
1. Register for [SICC2025](https://www.worldcubeassociation.org/competitions/SICC2025)
2. Try to update registrations (events, guests, comment)
3. See error

**Expected behavior**
Competitors on waiting list should be able to update their registrations

**Screenshots**

pranav-027:
Hello @dunkOnIT, I was testing this for a newcomer account id - 342200

pranav-027:
I asked a user with a WCA ID to try the same (2023GADG01) and encountered the same error. 

I tried to do the same thing on PuneCubeCarnival2025 and KotaCubeFestival2025 with user 342200. I was able to update my registration on PuneCubeCarnival2025 but not on 342200.  I think its happening only for non admin accounts

pranav-027:
Seems like I have missed on backend changes in the controller in the `user_can_modify_registration` method

PR #11663, opened 2025-05-30
Limit scope of allow_self_edit only to only apply to accepted registrations

dunkOnIT:
Changes were made in the frontend to support this, but the corresponding backend changes weren't implemented.

### DIFF
--- a/app/models/competition.rb
+++ b/app/models/competition.rb
@@ -1150,11 +1150,14 @@ class Competition < ApplicationRecord
     start_date.present? && start_date > Date.new(2021, 6, 24)
   end
 
+  def cannot_edit_accepted_registrations?
+    !self.allow_registration_edits?
+  end
+
   # can registration edits be done right now
   # must be allowed in general, and if the deadline field exists, is it a date and in the future
   def registration_edits_currently_permitted?
-    !started? && self.allow_registration_edits &&
-      (!event_change_deadline_date_required? || event_change_deadline_date.blank? || event_change_deadline_date > DateTime.now)
+    !started? && (!event_change_deadline_date_required? || event_change_deadline_date.blank? || event_change_deadline_date > DateTime.now)
   end
 
   private def dates_must_be_valid
--- a/app/controllers/api/v1/registrations_controller.rb
+++ b/app/controllers/api/v1/registrations_controller.rb
@@ -139,7 +139,10 @@ class Api::V1::RegistrationsController < Api::V1::ApiController
       can_administer_or_current_user?(@competition, @current_user, target_user)
 
     raise WcaExceptions::RegistrationError.new(:forbidden, Registrations::ErrorCodes::USER_EDITS_NOT_ALLOWED) unless
-      @competition.registration_edits_currently_permitted? || @current_user.can_manage_competition?(@competition) || user_uncancelling_registration?(@registration, new_status)
+      !(@registration.competing_status_accepted? && @competition.cannot_edit_accepted_registrations?) &&
+      @competition.registration_edits_currently_permitted? ||
+      @current_user.can_manage_competition?(@competition) ||
+      user_uncancelling_registration?(@registration, new_status)
 
     raise WcaExceptions::RegistrationError.new(:unauthorized, Registrations::ErrorCodes::REGISTRATION_IS_REJECTED) if
       user_is_rejected?(@current_user, target_user, @registration) && !organizer_modifying_own_registration?(@competition, @current_user, target_user)
--- a/spec/requests/api_registrations_spec.rb
+++ b/spec/requests/api_registrations_spec.rb
@@ -468,14 +468,15 @@ RSpec.describe 'API Registrations' do
       expect(response).to have_http_status(:unauthorized)
     end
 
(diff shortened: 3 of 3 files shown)


app/models/competition.rb
```diff
@@ -1150,11 +1150,14 @@ def event_change_deadline_date_required?
...
+  def cannot_edit_accepted_registrations?
+    !self.allow_registration_edits?
+  end
+
   # can registration edits be done right now
   # must be allowed in general, and if the deadline field exists, is it a date and in the future
   def registration_edits_currently_permitted?
-    !started? && self.allow_registration_edits &&
-      (!event_change_deadline_date_required? || event_change_deadline_date.blank? || event_change_deadline_date > DateTime.now)
+    !started? && (!event_change_deadline_date_required? || event_change_deadline_date.blank? || event_change_deadline_date > DateTime.now)
```
gregorbg, line 1160: Why did `self.allow_registration_edits` have to go?

app/models/registration.rb
```diff
@@ -646,4 +646,15 @@ def attempt_auto_accept
...
     { 'user_id' => user_id, 'competing' => { 'status' => status } }
   end
+
+  def user_can_modify?(user)
+    # Managers can always modify a registrations
+    return true if user.can_manage_competition?(self.competition)
+
+    # A registration can be edited by a user if it hasn't been accepted yet, and if edits are allowed.
+    editable_by_user = !(self.accepted? && self.competition.cannot_edit_accepted_registrations?) &&
+                       self.competition.registration_edits_currently_permitted?
```
gregorbg, line 656: I don't understand why this works. The `&& self.competition.registration_edits_currently_permitted?` part at the end is like saying "No matter what the other condition(s) of the registration is/are, you definitely have to **also** (the meaning of `&&`) make sure that edits are currently allowed as per the deadline".

But I think the point of this PR is to _skip_ checking the deadline if you're not accepted yet?!

spec/models/registration_spec.rb
```diff
@@ -1322,4 +1322,39 @@
...
+      competition.event_change_deadline_date = 2.weeks.from_now
+      expect(registration.user_can_modify?(competitor)).to be true
+    end
+
+    it "unless event edit deadline has passed" do
+      registration.competing_status = Registrations::Helper::STATUS_ACCEPTED
+      competition.allow_registration_edits = true
+      competition.event_change_deadline_date = 2.weeks.ago
+      expect(registration.user_can_modify?(competitor)).to be false
+    end
```
gregorbg, line 1358: There is an additional (new) test case to cover here, which is that registration edits of non-accepted registrations should be possible, even when the registration has passed.

app/models/registration.rb
gregorbg: My suggestion would be something along the lines of...

```ruby
editable_by_user = !self.accepted? || (self.competition.allow_registration_edits? && self.competition.registration_edits_currently_permitted?)
```

It's very well possible that I'm overlooking something, but at least from an intuitive standpoint, the query "is it in a 'not-accepted' state? Then do whatever the heck you want. Otherwise, check that edits are permitted and the edit deadline has not yet passed" makes a lot of sense given what we're trying to accomplish.

As a "bonus", it would also save the creation of the extra `def cannot_edit_accepted_registrations?` method.

app/models/competition.rb
dunkOnIT: Because `allow_registration_edits` now only applies to `accepted` registrations - so it doesn't contribute to defining globally whether registration edits are allowed

app/models/registration.rb
dunkOnIT: > the query "is it in a 'not-accepted' state? Then do whatever the heck you want. 

I think this is where our understanding differs. To my understanding, the `events_edit deadline` and `allow_registrations_edits` are separate things. 
- `events_edit_deadline` applies to _all_ registrations irrespective of status (and `registration_edits_currently_permitted?` implements this check in a slightly more nuanced way) 
- `allow_registration_edits` applies _only_ to accepted registrations (there's an argument for renaming the column here that I dismissed because I didn't want to introduce potential for errors in the renaming process, but we could reconsider this)

So in my mind, just because a reg isn't accepted doesn't mean that you can do whatever you want - you're still subject to the edit deadline.

However, I'm open to the alternative perspective of: 
- edits are always permitted on non-accepted registrations (until competition starts?)
- accepted registrations can be edited unless
  - `allow_registration_edits` is false, or the `edit_event_deadline` has passed

I'll take this to WCAT to clarify the scope

gregorbg: I do not feel very strongly about which version should be the "correct one". That's up to WCAT to decide.
However I do know that the original issue for this PR was prompted by the Edit Events deadline already having passed, see the error message in https://github.com/thewca/worldcubeassociation.org/issues/11651#issuecomment-2921343697

dunkOnIT: That is actually a case of a slightly misleading error string - the actual code is just USER_EDITS_NOT_ALLOWED (4001), but in the en.yml it is: 
```
'-4001': "The event edit deadline has passed"
```

PR #11698, opened 2025-06-02
Add support for inactive CPIs

dunkOnIT:
This PR: 
- [ ] Adds an `active` property to competition_payment_integrations
- [ ] Uses that property to render information on the payment setup page that a new CPI should be connected

Potential follow-ups: 
- [ ] Display message to users when CPI is inactive
  - [ ] hydrate the CPI's `active` property into the step/payment panel
- [ ] capture & act on Stripe webhook for deactived integration

### DIFF
--- a/app/models/competition_payment_integration.rb
+++ b/app/models/competition_payment_integration.rb
@@ -33,4 +33,13 @@ class CompetitionPaymentIntegration < ApplicationRecord
     raise ArgumentError.new("Invalid integration name. Allowed values are: #{AVAILABLE_INTEGRATIONS.keys.join(', ')}") unless
       AVAILABLE_INTEGRATIONS.key?(integration_name)
   end
+
+  # We mark an account as inactive if a payment provider informs us that we no longer have access to that account
+  # In this case, the expected workflow is:
+  # 1. Inform the organizer via UI that the account is inactive, and ask them to remove the connected payment integration
+  # 2. Organizer removes CPI and connects a new account
+  # Thus "inactive" is a placeholder state before a CPI is removed - which is why we have no mark_active functionality
+  def mark_inactive
+    self.update(active: false)
+  end
 end
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1288,6 +1288,7 @@ en:
         amount_too_low: "The total amount must be at least the registration fees to be paid."
         amount_rather_high: "You are about to pay at least double the required amount for your registration to proceed."
       errors:
+        no_account_access: "Unable to access the organizer's payment account. Please contact the organization and ask them to confirm their Payment Setup details on the WCA website."
         registration_closed: "Registration has closed - no further payments are allowed."
         already_paid: "The charge was not placed because the registration fees are already paid."
         not_allowed: "You can't pay for that registration."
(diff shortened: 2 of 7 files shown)


app/models/competition_payment_integration.rb
```diff
@@ -33,4 +33,13 @@ def self.validate_integration_name!(integration_name)
     raise ArgumentError.new("Invalid integration name. Allowed values are: #{AVAILABLE_INTEGRATIONS.keys.join(', ')}") unless
       AVAILABLE_INTEGRATIONS.key?(integration_name)
   end
+
+  # We mark an account as inactive if a payment provider informs us that we no longer have access to that account
+  # In this case, the expected workflow is:
+  # 1. Inform the organizer via UI that the account is inactive, and ask them to remove the connected payment integration
+  # 2. Organizer removes CPI and connects a new account
+  # Thus "inactive" is a placeholder state before a CPI is removed - which is why we have no mark_active functionality
+  def mark_inactive
```
gregorbg: Since this is mutating the entity's state in-place, it should be called `mark_inactive!` by convention

db/migrate/20250529074918_add_active_to_cpi.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddActiveToCpi < ActiveRecord::Migration[7.2]
+  def change
+    add_column :competition_payment_integrations, :active, :boolean, after: :connected_account_id, default: true
```
gregorbg: Hm. Our default assumption is that CPIs added to this table are active, and only get **de**-activated in exceptional circumstances. This means calling the column `inactive` with default `false` feels less redundant.

gregorbg: Also, regardless of which way we end up flipping the boolean: Since databases don't have the `?` at the end as a supported field name, we usually call booleans `is_*`.

gregorbg (changes requested):

db/schema.rb
```diff
@@ -1184,7 +1185,7 @@
   end
 
   create_table "schedule_activities", charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
-    t.bigint "venue_room_id", null: false
+    t.bigint "venue_room_id"
```
gregorbg: Unrelated change (but you have merge conflicts anyways...)

gregorbg (changes requested):

config/locales/en.yml
```diff
@@ -1288,6 +1288,7 @@ en:
         amount_too_low: "The total amount must be at least the registration fees to be paid."
         amount_rather_high: "You are about to pay at least double the required amount for your registration to proceed."
       errors:
+        no_account_access: "Unable to access the organizer's payment account. Please contact the organization and ask them to confirm their Payment Setup details on the WCA website."
```
gregorbg:
```suggestion
        no_account_access: "Unable to access the organizer's payment account. Please contact the organization team and ask them to confirm their Payment Setup details on the WCA website."
```

db/schema.rb
gregorbg: Unrelated change still there

spec/models/competition_payment_integration_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+require 'rails_helper'
+
+RSpec.describe CompetitionPaymentIntegration do
+  describe '#mark_inactive' do
+    let!(:competition) { create(:competition, :stripe_connected) }
+    let(:integration) { competition.competition_payment_integrations.first }
+
+    it 'sets to inactive in the database' do
+      integration.mark_inactive!
+      expect(integration.is_inactive?).to be(true)
```
gregorbg, line 12: Where is the method `is_inactive?` being defined?

db/schema.rb
dunkOnIT: Fixed

spec/models/competition_payment_integration_spec.rb
dunkOnIT: It's a property of CompetitionPaymentIntegration

PR #11706, opened 2025-06-02
v1 API camelCase key support

dunkOnIT:
This PR does the following: 
- [x] Converts incoming camelCase to snake_case in the v1 api_controller (with tests asserting behaviour)
- [x] Adds a `render_as_camel_case` method to the v1 api_controller (with tests asserting behaviour)

### Summary
- There aren't any standard Rails-y ways to handle this that I could find (although there are more appropriate ways to do it if we used certain gems which we don't currently use
- But it's pretty easy to roll our own, and that's a common recommendation for our (not uncommon) use case
- Came across two gems that may be of interest to us at some point
  - [JBuilder](https://github.com/rails/jbuilder) (a DSL for building JSON payloads)
  - [active_model_serializers](https://github.com/rails-api/active_model_serializers) (helps with serializing model objects into JSON, as the name suggests. AFAIK we do this "manually" at the moment, and this gem _may_ make it more dev-friendly

### Sources
- https://www.reddit.com/r/rails/comments/10qnxhb/snakecamel_case_conversion_gem/
- https://mnishiguchi-2019.netlify.app/2017/11/29/rails-hash-camelize-and-underscore-keys/
- https://stackoverflow.com/questions/37569439/automatically-convert-hash-keys-to-camelcase-in-jbuilder
- https://discuss.rubyonrails.org/t/rails-with-snakecase-camelcase-conversion-frontend-adapter-pattern/75064/3
- https://stackoverflow.com/questions/1509915/converting-camel-case-to-underscore-case-in-ruby
- Conversation with ChatGPT (which confirmed my general impression from the above sources)

### DIFF
--- a/app/controllers/api/v1/api_controller.rb
+++ b/app/controllers/api/v1/api_controller.rb
@@ -15,6 +15,7 @@ class Api::V1::ApiController < ActionController::API
   skip_before_action :validate_jwt_token
   def test_action
     return head :not_found if Rails.env.production? && EnvConfig.WCA_LIVE_SITE?
+
     params.delete(:action)
     params.delete(:api) # TODO: ChatGPT claims I shouldn't be getting this key - but for now I'm just trying to get the tests passing
     params.delete(:controller)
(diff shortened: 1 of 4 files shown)


gregorbg:
Closing this in favor of continuing to use snake_case for the time being. We can dig the comments with your research up if we ever revisit your decision

PR #11707, opened 2025-06-03
Correctly cast boolean param for applyFix checkbox in validators

gregorbg:
The frontend was using a correctly typed Boolean, but then when passing it to `jsonToQueryString` to transport as URL parameter, it gets automatically cast into a string.

The Rails backend also reads everything from the URL query as string, so it was interpreting `if "false"` (note the quotes marking it as string!) which behaves like `true` because the string is not empty.


gregorbg:
CC @danieljames-dj

dunkOnIT (approved):

PR #11709, opened 2025-06-03
Add round_id foreign keys to results and scrambles table

gregorbg:
Migration step that was originally planned for the previous backfilling PR, but took such a long time that we couldn't run it as part of the normal deployment migrations.

My idea for this PR is to add the code change here on GitHub for consistency, but to actually execute the change manually through the console and insert the corresponding row into `schema_migrations`. That way, we don't run into timeouts of the automated deployment 5-minute blue-green time window.


gregorbg:
I have manually run these queries, merging now in the hope that it just "slides by" during deployment.

Issue #11710, opened 2025-06-03
Paginate banned competitors page UI


danieljames-dj:
The banned competitors page UI (http://localhost:3000/panel-page/banned-competitors) is currently not paginated. Because of this, it will show only 100 results.

The UI should be modified in a way that there can be pagination so that the user can select which page they want to see.

PR #11711, opened 2025-06-03
Paginate banned competitors page

pranav-027:
fixes: https://github.com/thewca/worldcubeassociation.org/issues/11710

Copy pasta from incident log component

### DIFF
--- a/app/webpacker/components/Panel/pages/BannedCompetitorsPage/BannedCompetitors.jsx
+++ b/app/webpacker/components/Panel/pages/BannedCompetitorsPage/BannedCompetitors.jsx
@@ -1,55 +1,73 @@
 import React from 'react';
 import { Icon, Table, Segment } from 'semantic-ui-react';
 import UserBadge from '../../../UserBadge';
+import PaginationFooter from '../../../PaginationFooter';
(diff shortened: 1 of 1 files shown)


danieljames-dj:
@pranav-027 Thanks for taking this up. Backend changes won't be needed. You just need to pass the page number as the argument with normal roles fetching API. For example, let's say you are using '/user_roles?(...)' for fetching list of roles. When you use pagination, you should use '/user_roles?(...)&page=n' (where n is the page number). The backend for the same is already handled. You can try out by opening some URLs (like `/api/v0/user_roles?page=3`) in your browser.

pranav-027:
[screen-capture (4).webm](https://github.com/user-attachments/assets/75849dfb-7035-4f81-86da-0a3a233ef9bc)

Currently looking like this and i found comment by daniel as well in the `user_roles_controller.rb`

```ruby
  def index
    roles = pre_filtered_user_roles

    # Filter & Sort roles.
    roles = UserRole.filter_roles(roles, current_user, params)
    roles = UserRole.sort_roles(roles, params[:sort])

    # Paginating the list by sending only first 100 elements unless mentioned in the API.
    paginate json: roles
  end
```

PR #11714, opened 2025-06-03
Add role column to ticket stakeholder table

danieljames-dj:
This is the implementation of new role column in ticket stakeholders table. The details of the design can be found [here](https://github.com/thewca/worldcubeassociation.org/issues/9463#issuecomment-2780150631).


app/webpacker/components/Tickets/TicketWorkbenches/index.jsx
gregorbg, line 1: I don't think you need to call this `index` if you're not actually exposing an actual component that can be rendered. This kind of lookup feels like it should better go under some local (i.e. within the `Tickets` folder strucutre) `utils` or similar.

db/migrate/20250603154450_change_column_null_for_users_status.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class ChangeColumnNullForUsersStatus < ActiveRecord::Migration[7.2]
```
gregorbg: The name of this migration doesn't make a lot of sense to me

db/migrate/20250603142039_populate_stakeholder_role_for_ticket_stakeholders.rb
```diff
@@ -0,0 +1,12 @@
...
+
+class PopulateStakeholderRoleForTicketStakeholders < ActiveRecord::Migration[7.2]
+  def up
+    TicketStakeholder.where(stakeholder_type: 'User').update_all(stakeholder_role: 'requester')
+    TicketStakeholder.where(stakeholder_type: 'UserGroup').update_all(stakeholder_role: 'actioner')
+  end
+
+  def down
+    TicketStakeholder.update_all(stakeholder_role: nil)
+  end
```
gregorbg: FWIW, you can merge this into one big migration:

```
def change
  add_column ...
  
  reversible do |direction|
    direction.up do
      # stuff from the `def up` body
    end
    
    direction.down do
      # stuff from the `def down` body
    end
  end
  
  change_column_null ...
end
```

app/webpacker/components/Tickets/index.jsx
```diff
@@ -41,18 +41,24 @@ function SkateholderSelector({ stakeholderList, setUserSelectedStakeholder }) {
...
-        sync={sync}
-        currentStakeholder={currentStakeholder}
-      />
+      {TicketWorkbench && (
+        <TicketWorkbench
+          ticketDetails={ticketDetails}
+          sync={sync}
+          currentStakeholder={currentStakeholder}
+        />
+      )}
```
gregorbg, line 61: This will "silently fail" (that means, just stop rendering anything) if the stakeholder has no workbench defined. I think a more clear message along the lines of `{TicketWorkbench ? <TicketWorkbench ... /> : 'Workbench not found'}` would be prudent

gregorbg (changes requested):

db/migrate/20250603142039_populate_stakeholder_role_for_ticket_stakeholders.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class PopulateStakeholderRoleForTicketStakeholders < ActiveRecord::Migration[7.2]
+  def up
+    TicketStakeholder.where(stakeholder_type: 'User').update_all(stakeholder_role: 'requester')
```
pranav-027: Just a question, why don't we add conditions like 'if already not done then do' in database migrations?

db/migrate/20250603142039_populate_stakeholder_role_for_ticket_stakeholders.rb
danieljames-dj: Oh I didn't knew about `reversible`, merging to single migration sounds like a great idea. Done.

app/webpacker/components/Tickets/TicketWorkbenches/index.jsx
danieljames-dj: I'm also bit confused. I now changed to keep this under `Tickets` folder with name `TicketWorkbenches`. This was how `PanelPages` was also done. Let me know if I want to change it.

app/webpacker/components/Tickets/index.jsx
danieljames-dj: Actually workbench may not be there for every case. For example, for edit-person ticket, there is currently no workbench for requester (it might come in future). But I don't want to show any message saying that there is no workbench, because there are many other things in the page for the user - like comments, logs, ticket header, etc. So is it fine to have it in current way?

db/migrate/20250603142039_populate_stakeholder_role_for_ticket_stakeholders.rb
danieljames-dj: I'm not sure if this is common in migrations. But usually if we are sure that this is not yet done, then we don't give that condition.

db/migrate/20250603154450_change_column_null_for_users_status.rb
danieljames-dj: Okay changed to new migration anyway.

gregorbg (approved):

PR #11716, opened 2025-06-04
Add Infra to Nextjs Setup with Docker/ECS

FinnIckler:


next-frontend/src/payload.config.ts
```diff
@@ -9,32 +9,44 @@ import { authConfig } from "@/auth.config";
...
+    dbAdapter = sqliteAdapter({
+      client: {
+        url: process.env.DATABASE_URI || "",
+      },
+    });
+  } else {
+    dbAdapter = mongooseAdapter({
       url: process.env.DATABASE_URI || "",
+    });
+  }
```
gregorbg: This could be moved into a separate `getDatabaseAdapter` function, which gets rid of the `let` neatly

next-frontend/src/payload.config.ts
```diff
@@ -9,32 +9,44 @@ import { authConfig } from "@/auth.config";
...
-  },
-  collections: [Media],
-  globals: [Nav],
-  secret: process.env.PAYLOAD_SECRET || "",
-  typescript: {
-    outputFile: path.resolve(dirname, "payload-types.ts"),
-  },
-  db: sqliteAdapter({
-    client: {
+export default function config() {
```
gregorbg: You may have accidentally introduced a mistake here: Previously, the code was _calling_ `buildConfig` and then returning the _result_ of that call as its export.

Now you're exporting a `function`, which breaks Payload (and probably the typecheck in CI). Make sure that whatever is returned here is actually an object, not a function returning an object.

FinnIckler: yeah, I fixed this locally already

next-frontend/next.config.ts
```diff
@@ -5,6 +5,7 @@ const nextConfig: NextConfig = {
   experimental: {
     optimizePackageImports: ["@chakra-ui/react"],
   },
+  output: "standalone",
```
gregorbg, line 8: What does this line do and why is it necessary?

next-frontend/src/payload.config.ts
```diff
@@ -9,32 +8,65 @@ import { authConfig } from "@/auth.config";
...
+        url: process.env.DATABASE_URI || "",
+      },
+    });
+  }
+}
+
+export default createConfig();
+
+async function createConfig() {
+  return buildConfig({
```
gregorbg: Why not inline this? Like the old `export default buildConfig({ ...`

next-frontend/dockerfile
```diff
@@ -0,0 +1,61 @@
...
+FROM base AS builder
+WORKDIR /app
+COPY --from=deps /app/node_modules ./node_modules
+COPY . .
+
+# Next.js collects completely anonymous telemetry data about general usage.
+# Learn more here: https://nextjs.org/telemetry
+# Uncomment the following line in case you want to disable telemetry during the build.
+ENV NEXT_TELEMETRY_DISABLED=1
+RUN corepack enable
```
gregorbg: You're running this twice in the Dockerfile, might make sense to just run it as part of the base image?

next-frontend/dockerfile
gregorbg, line 1: Please rename to `Dockerfile` with a capital D so GitHub turns on its syntax highlighting :D

next-frontend/dockerfile
```diff
@@ -0,0 +1,61 @@
...
+# Create the types and build
+RUN yarn run chakra:types && yarn run build
+
+# Production image, copy all the files and run next
+FROM base AS runner
+WORKDIR /app
+
+ENV NODE_ENV=production
+# Uncomment the following line in case you want to disable telemetry during runtime.
+ENV NEXT_TELEMETRY_DISABLED=1
```
gregorbg: Also, this can probably be moved to the `base` image definition

next-frontend/dockerfile
```diff
@@ -0,0 +1,61 @@
...
+# As per https://authjs.dev/getting-started/deployment#docker
+ENV AUTH_TRUST_HOST=true
+RUN addgroup --system --gid 1001 nodejs
+RUN adduser --system --uid 1001 nextjs
```
gregorbg: Why is the group called `nodejs` but the user is called `nextjs`?

next-frontend/next.config.ts
FinnIckler: Basically it just creates one `server.js` file that you can use to start the server directly only with node. It vastly reduces Image size because it doesn't include the `node_modules` anymore. I've seen it in basically all official guides and examples

next-frontend/src/payload.config.ts
FinnIckler: true, that works again now

next-frontend/dockerfile
FinnIckler: this is straight from the official next.js example dockerfile, not sure why they put it there twice

next-frontend/dockerfile
FinnIckler: also from the official dockerfile, maybe node is owned by the nodejs group in the alpine image? I can check it out

gregorbg:
TIL: You can override package-internal dependency resolution in Yarn by adding the following top-level block to `package.json`:
```json
"resolutions": {
  "@payloadcms/db-mongodb/mongoose": "^8.15.1"
}
```

FinnIckler:
I had to make the imports for the database adapters dynamic as just importing the sqlite adapter causes the next compiler to include it and requiring the libsqlite library

next-frontend/package.json
```diff
@@ -14,12 +17,15 @@
     "lint:fix": "next lint --fix"
   },
   "dependencies": {
+    "@aws-sdk/credential-providers": "^3.823.0",
     "@chakra-ui/react": "^3.20.0",
     "@emotion/react": "^11.14.0",
+    "@payloadcms/db-mongodb": "^3.40.0",
```
gregorbg:
```suggestion
    "@payloadcms/db-mongodb": "^3.41.0",
```

gregorbg: Plus corresponding `yarn.lock` changes

PR #11718, opened 2025-06-06
Make flash message HTML safe

danieljames-dj:
Currently, there is a case where the flash message is not HTML safe and because of which [this alert](https://github.com/thewca/worldcubeassociation.org/blob/9acfb68f10ef0bdc46d9e46a1538c1116741817a/app/controllers/users_controller.rb#L266) will looks weird.

In this PR, I've made the flash message HTML safe, so that HTML messages can be rendered in flash.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -104,6 +104,7 @@ gem "tzf"
 gem 'playwright-ruby-client', require: 'playwright'
 gem 'hash_diff'
 gem 'tsort'
+gem 'html_safe_flash'
 
 group :development, :test do
   gem 'spring'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -391,6 +391,8 @@ GEM
     highline (3.1.2)
       reline
     hiredis (0.6.3)
+    html_safe_flash (0.2.0)
+      rails (>= 6)
     htmlentities (4.3.4)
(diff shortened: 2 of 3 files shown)


gregorbg (commented):
I don't think there's any way to be sure about one making each and every single flash message assume HTML safety by default.

If you have one message that looks weird, you can mark that one specific message as safe, no?

gregorbg:
My initial instinct was correct: Marking everything as `html_safe` by default is a horrible idea, because it opens us up to vulnerabilities: https://github.com/samvera/hyrax/pull/3229

In short, if a malicious user finds _any_ place in our code, where they can input a string into a textfield, and that exact input is then later used as a flash alert, then we are making ourselves critically vulnerable to XSS.

I don't feel confident to make a call on whether we have any alert message where user input appears or not. And I don't want you to search through our whole code trying to find it out, because even if we're currently safe, then it would be hard to take note of and keep a maintenance on all flash messages in all places of our code everywhere.

gregorbg:
I would be open for solutions such as https://github.com/quadule/html_safe_flash

danieljames-dj:
Okay, installed `html_safe_flash` and it is working fine without any additional changes. Thanks for it.

PR #11721, opened 2025-06-06
Basic Vitest Setup for NextJS

dunkOnIT:
This PR introduces the testing packages + config files as well as a single passing test against the footer

### DIFF
--- a/next-frontend/package.json
+++ b/next-frontend/package.json
@@ -2,6 +2,7 @@
   "name": "next-frontend",
   "version": "0.1.0",
   "private": true,
+  "type": "module",
   "scripts": {
     "dev": "next dev",
     "check:types": "tsc --noEmit",
@@ -10,7 +11,9 @@
     "build:types": "openapi-typescript public/wcaAPI.yaml -o src/lib/wca/wcaSchema.ts",
     "chakra:types": "chakra typegen ./src/theme.ts",
     "lint": "next lint",
-    "lint:fix": "next lint --fix"
+    "lint:fix": "next lint --fix",
+    "test": "vitest --config vite.config.ts",
+    "test:ui": "vitest --ui"
   },
   "dependencies": {
     "@chakra-ui/react": "^3.19.1",
@@ -37,17 +40,23 @@
     "@eslint/eslintrc": "^3",
     "@stylistic/eslint-plugin": "^4.4.0",
     "@tanstack/eslint-plugin-query": "^5.78.0",
+    "@testing-library/dom": "10.4.0",
+    "@testing-library/jest-dom": "6.6.3",
+    "@testing-library/react": "^16.3.0",
     "@types/node": "^22",
     "@types/react": "^19",
     "@types/react-dom": "^19",
     "@types/swagger-ui-react": "^5.18.0",
+    "@vitejs/plugin-react": "4.5.1",
     "eslint": "^9",
     "eslint-config-next": "15.3.3",
     "eslint-config-prettier": "^10.1.5",
     "eslint-plugin-prettier": "^5.4.1",
+    "jsdom": "^26.1.0",
     "openapi-typescript": "^7.8.0",
     "prettier": "^3.5.3",
-    "typescript": "^5"
+    "typescript": "^5",
+    "vitest": "^3.2.1"
   },
   "scarfSettings": {
     "enabled": false
--- a/next-frontend/src/components/Footer.test.tsx
+++ b/next-frontend/src/components/Footer.test.tsx
@@ -1,0 +1,18 @@
+import { render, screen } from "@testing-library/react";
+import { React, act } from "react";
+import { Provider as UiProvider } from "@/components/ui/provider";
+import Footer from "@/components/Footer";
+
+describe("Footer", () => {
+  it("renders without crashing", async () => {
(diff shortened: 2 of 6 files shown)


next-frontend/package.json
```diff
@@ -2,6 +2,7 @@
   "name": "next-frontend",
   "version": "0.1.0",
   "private": true,
+  "type": "module",
```
gregorbg: I think (97% sure from my local testing) that this breaks NextJS compilation.
Try `yarn build` to see whether the build scripts still complete successfully with this line included.

next-frontend/package.json
```diff
@@ -10,7 +11,9 @@
     "build:types": "openapi-typescript public/wcaAPI.yaml -o src/lib/wca/wcaSchema.ts",
     "chakra:types": "chakra typegen ./src/theme.ts",
     "lint": "next lint",
-    "lint:fix": "next lint --fix"
+    "lint:fix": "next lint --fix",
+    "test": "vitest --config vite.config.ts",
+    "test:ui": "vitest --ui --config vite.config.ts"
   },
   "dependencies": {
```
gregorbg, line 21: All of the changes listed here under `dependencies` feel more like `devDependencies`. Can you please double-check?

dunkOnIT: I think they're already in devDependencies, github's line compression just made it look like they weren't?

gregorbg: They were not in the beginning, but they are now :D

dunkOnIT:
> Did you follow any specific setup guide to come up with these changes? Or did you go through all of this by yourself?

No specific guide - this was me + ChatGPT + trying to run a test file until it would pass

next-frontend/package.json
dunkOnIT: Made unnecessary by renaming `vitest.config.ts` to `vitest.config.mts`

next-frontend/package.json
```diff
@@ -37,17 +40,23 @@
     "@eslint/eslintrc": "^3",
     "@stylistic/eslint-plugin": "^4.4.0",
     "@tanstack/eslint-plugin-query": "^5.78.0",
+    "@testing-library/dom": "10.4.0",
+    "@testing-library/jest-dom": "6.6.3",
```
dunkOnIT: This provides frontend testing methods/matchers that are not available in the other testing libraries - and is fairly commonly used alongside Vitest (one ref: https://www.reddit.com/r/reactjs/comments/1fiasoc/should_you_still_use_jestdom_with_vitest/) 

That Reddit link also mentions jest-dom being included with React testing library usually - but as far as I can tell, this is not the case

next-frontend/vitest.setup.ts
gregorbg, line 1: This looks like a "meaningless" setup file. That is, it does define some module functions if they don't exist yet, but the actual content of the definitions is meaningless. Why is this needed in the first place?

dunkOnIT: After some playing around and googling: 
- **Practical effect**: Removing this raises a `TypeError: window.matchMedia is not a function`
- **Is this workaround acceptable?**: It is recommended in the [Jest documentation](https://jestjs.io/docs/manual-mocks#mocking-methods-which-are-not-implemented-in-jsdom) (which Vitest is based on), and [this Vitest issue](https://github.com/vitest-dev/vitest/issues/821) describes a much-hearted solution which follows the same principle
- **Why is it needed?**: I _believe_ the issue here is that `jsdom` doesn't support methods used in normal/non-headless browsers (like `window.matchMedia` - so when they are called, an error is raised. However, these headed browser methods are irrelevant in test, so it is sufficient to just mock them and have meaningless return values

test-results/.last-run.json
gregorbg, line 1: Uhm, this should probably be `gitignore`d?

next-frontend/src/components/Footer.test.tsx
gregorbg, line 1: Is it really good practice to have the test file lay around next to the source file in the "main" source directory?
At least in Rails, we keep tests in a completely separate folder tree, so I would expect this to live outside of `src`, perhaps under `test` or `tests`...?

playwright-report/index.html
gregorbg, line 1: I can see somebody is experimenting with Playwright :P

next-frontend/src/components/Footer.test.tsx
dunkOnIT: Vitest is unopinionated on this, and it's ultimately down to personal preference - having tests live in the same directory as source files is definitely an acceptable practice, and I'm a fan of it for ergonomic reasons when we're dealing with unit/component tests

FinnIckler: I have done it the same way in my nextjs projects. It makes more sense when you have a Folder for each Component though. Basically my setup is like this:
```
/components
   /components/Footer
     /components/Footer.tsx
     /components/Footer.test.tsx
     /components/Footer.stories.tsx (for storybook)
```

playwright-report/index.html
dunkOnIT: :skull: resolved

test-results/.last-run.json
dunkOnIT: Resolved

Issue #11023, opened 2025-03-13
Long names in region dropdown renders poorly


xsrvmy:
On the persons page, the region dropdown is narrow, and long country names produce the result shown in the following screenshot with the text wrapping below the flag, which is probably not desirable:

![Image](https://github.com/user-attachments/assets/6f38c788-6d3b-4b9c-8ba3-6417cd0c07f8)

gregorbg:
Not sure what's a good solution here. Making the dropdown box suuuuuuuuper long just so that we can fit "Democratic People's Republic of Korea" onto one line, while >=90% of all other entries in the whole list are less than 20 characters long, creates a lot of blank space and is probably not desirable either.

xsrvmy:
The name should probably wrap lines but stay to the right of the flag (and the flag vertically centered) rather than wrapping below it which is what looks ugly right now.

dunkOnIT:
> The name should probably wrap lines but stay to the right of the flag (and the flag vertically centered) rather than wrapping below it which is what looks ugly right now.

Agreed

PR #11723, opened 2025-06-07
vertically center RegionSelector country name to flag

Brandon-Reichman:
Fixes https://github.com/thewca/worldcubeassociation.org/issues/11023  Minor display issue in region dropdown where countries with long names were getting wrapped under the flag.

This change makes the name stay to the right of the flag and be vertically centered, as shown below.

wca-region-select-text-wrap

### DIFF
--- a/app/webpacker/stylesheets/override.scss
+++ b/app/webpacker/stylesheets/override.scss
@@ -118,7 +118,7 @@ body, html {
 
   // This is to vertically align region flag and name in RegionSelector dropdown
   .ui.dropdown {
-    .menu.visible.transition[role='listbox'] {
+    .menu[role='listbox'] {
       .item[role='option'] {
         display: flex;
         align-items: center;


Brandon-Reichman:
Thanks! I've moved this styling change to `override.scss`

app/webpacker/stylesheets/override.scss
```diff
@@ -115,4 +115,22 @@ body, html {
       background-color: #e0e0e0 !important;
     }
   }
+
+  // This is to vertically align region flag and name in RegionSelector dropdown
+  .ui.dropdown {
+    .menu.visible.transition[role='listbox'] {
```
gregorbg: I don't think `visible` and `transition` are necessary to identify a menu dropdown.
The fact that you're nesting inside `.ui.dropdown`, combined with the `.menu` and `[role=listbox]` selector, should be more than enough.

gregorbg (changes requested):
Cool, much better now! A small nitpick about the specificity of the CSS selector you chose, but otherwise good to do :D

app/webpacker/stylesheets/override.scss
Brandon-Reichman: Updated!

Issue #11672, opened 2025-06-02
Warning logo for pending report submission not visible when delegate is also a WRT member


pranav-027:
**Describe the bug**
When a user holds both the delegate and WRT member roles, the warning icon indicating a pending report submission in the past competition section of My Competitions page is not displayed.

PR #11724, opened 2025-06-08
fix missing delegate report warning for team members

Brandon-Reichman:
Fixes https://github.com/thewca/worldcubeassociation.org/issues/11672 where members of WRT, WST, WCAT, and Board would not see warning icon for missing delegate reports.
For these users, `permissions.can_administer_competitions.scope` would return `*`, so calling `.includes(competitionId)` would always return false. This has been fixed to check the scope for `*`  or check if the competitionId in included in the scope, similar to the view and edit check.
The way `can_administer_competitions` permission is defined, anyone on one of the teams listed above will see the missing report warning on competitions, regardless if they are a delegate for that competition, which seems like the intention.


gregorbg (approved):
Thanks for the concise and effective fix!

Issue #10927, opened 2025-02-28
Deleting a picture discards unsaved delegate report edits


kr-matthews:
**Describe the bug**
When deleting an uploaded image from a delegate report, all other changes are lost.

**To Reproduce**
1. Go to a competition's unsubmited delegate report.
2. Upload pictures.
3. Save
4. Come back later, make some other edits without saving.
5. Delete an image.
6. All changes in step 4 are gone.

Issue #7400, opened 2022-11-27
Create an "Are you sure you want to leave the page?" popup on unsaved delegate reports


dunkOnIT:
From "New website feature proposal" email thread:

> Hi WST,
>
> I just made the rookie mistake of closing the delegate report by accident without saving it and had to write it all over again. Would it be possible to add an extra layer of safety like the one that comes out when you try to post a report or close the competition creation form without submitting (attached picture)?
> I think this would be a very nice addition to the website.

This seems like a very reasonable suggestion to me.

Nevseros:
If this is going to be added, can the same feature also be added for Incidents log? I just had to rewrite the incident twice due to a missclick, which is my mistake, but still quite disappointing. :(

dunkOnIT:
Sure thing @Nevseros, good catch! I've updated the initial description accordingly.

Nevseros:
Another fail - another page found! New posts and posts edits would benefit from this feature as well.

dunkOnIT:
What portion of the site is the new posts/edit posts feature? I'm guessing the announcements section, given that you've just posted a 2022 survey announcement? :P

Nevseros:
Exactly! 😄

PR #11726, opened 2025-06-08
Add Form Alert For Unsaved Changes

Brandon-Reichman:
Closes https://github.com/thewca/worldcubeassociation.org/issues/7400

Created a JavaScript function that tracks if a form has been changed and triggers the default browser alert of "You have unsaved changes. Are you sure you want to leave?" alert when navigating away. This is done simply by calling `event.preventDefault()` before page unload.

This function is applied to delegate reports, incident logs, and front page announcements.

Resolves https://github.com/thewca/worldcubeassociation.org/issues/10927 as this alert would appear when clicking delete image with unsaved changes.

### DIFF
--- a/app/webpacker/components/Posts/PostForm.jsx
+++ b/app/webpacker/components/Posts/PostForm.jsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useMemo, useState } from 'react';
+import React, { useCallback, useMemo, useState, useEffect } from 'react';
 import {
   Button, Checkbox, Form, FormField, FormGroup, Header, Message,
 } from 'semantic-ui-react';
@@ -83,12 +83,16 @@ export default function PostForm({
     postId,
   ]);
 
+  useEffect(() => {
+    unsavedFormAlert(document.getElementById('post-form'));
+  }, []);
+
   return (
     <>
       <Header>
         {header}
       </Header>
-      <Form onSubmit={onSubmit}>
+      <Form id="post-form" onSubmit={onSubmit}>
         <Form.Input label={I18n.t('activerecord.attributes.post.title')} onChange={setFormTitle} value={formTitle} />
         <FormField>
           <label htmlFor="post-body">{I18n.t('activerecord.attributes.post.body')}</label>
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -293,3 +293,21 @@ $(function() {
     form.submit();
   });
 });
+
+// Tracks changes on form and warns user before navigating away with unsaved changes
+function unsavedFormAlert(form) {
+  if (form) {
+      let unsavedChanges = false;
+      const markUnsaved = () => { unsavedChanges = true; }
+
+      form.addEventListener("change", markUnsaved);
+      form.addEventListener("input", markUnsaved);
+      form.addEventListener("submit", () => { unsavedChanges = false });
+
+      window.addEventListener("beforeunload", (e) => {
+        if (unsavedChanges) {
+          e.preventDefault();
+        }
+      });
+    }
+}
--- a/app/views/delegate_reports/edit.html.erb
+++ b/app/views/delegate_reports/edit.html.erb
@@ -6,7 +6,7 @@
 <%= render layout: 'nav' do %>
   <h2>Editing Delegate Report for <%= @competition.name%></h2>
   <%= simple_form_for @delegate_report, url: delegate_report_path(@competition.id),
-                                        html: { class: "delegate-report" } do |f| %>
+                                        html: { class: "delegate-report", id: "delegate-report-form" } do |f| %>
     <%= f.hidden_field :updated_at %>
 
     <% if @delegate_report.posted? %>
@@ -68,6 +68,7 @@
 
 <script>
   document.addEventListener("DOMContentLoaded", function () {
+    unsavedFormAlert(document.getElementById("delegate-report-form"));
     const checkbox = document.getElementById("wic_feedback_checkbox");
 
     if (checkbox) {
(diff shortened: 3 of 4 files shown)


Brandon-Reichman:
I'm honestly unsure how to resolve the failure `Expected a line break after this opening brace.` as only `, useEffect` was added to that line.

I suppose the other one makes sense calling `unsavedFormAlert` from react, a bit weird, but it worked in local testing.

app/webpacker/components/Posts/PostForm.jsx
```diff
@@ -83,12 +85,31 @@ export default function PostForm({
...
+  useEffect(() => {
+    const form = document.getElementById('post-form');
+
+    if (form) {
+      let unsavedChanges = false;
+      const markUnsaved = () => { unsavedChanges = true; };
+
+      form.addEventListener('change', markUnsaved);
+      form.addEventListener('input', markUnsaved);
+      form.addEventListener('submit', () => { unsavedChanges = false; });
```
gregorbg: You have access to the individual values that are being manipulated by the form (title, body, tags, etc...) through the `useState` and `useInputState` declarations at the top of the form. You can compare them to `post` to see whether something changed.

app/webpacker/components/Posts/PostForm.jsx
```diff
@@ -83,12 +85,31 @@ export default function PostForm({
...
+
+      window.addEventListener('beforeunload', (e) => {
+        if (unsavedChanges) {
+          e.preventDefault();
+        }
+      });
```
gregorbg: If you add a global window listener, you should also return a cleanup from `useEffect` to remove that same listener upon unload. See `components/EditEvents/index.jsx` for an example on how that can be accomplished.

Brandon-Reichman:
I should have known better than to open the PR in the state it was.

Thanks for pointing out `AreYouSure`. It wasn't working with markdown editor inputs, so I had to change `markdown-editor.js` to trigger the change event on the element when editor changes were made. This allows AreYouSure to track these changes.

I've also cleaned how react tracks the form changes.

app/webpacker/components/Posts/PostForm.jsx
```diff
@@ -83,12 +90,23 @@ export default function PostForm({
...
+  const onChange = useCallback(() => {
+    setHasChanged(true);
+  }, [setHasChanged]);
+
+  useEffect(() => {
+    if (!hasChanged) return;
+    window.addEventListener('beforeunload', (event) => {
+      event.preventDefault();
+    });
+  }, [hasChanged]);
```
gregorbg: You're still missing the `return` for unloading the event listener.

app/webpacker/lib/markdown-editor.js
```diff
@@ -131,6 +131,14 @@ $(() => {
       },
     });
 
+    // Manually trigger change to original textarea so are-you-sure can be utilized
+    editor.codemirror.on('change', () => {
+      this.value = editor.value();
+      $(this).trigger('change');
+    });
+    // So edited value does not persist on refresh
+    editor.value(this.defaultValue);
```
gregorbg, line 140: In how far is this related to the current PR? Just looking at your code comment, it feels out of scope. Could you please elaborate?

Brandon-Reichman: The behavior of the value persisting on refresh was introduced with the snippet added above this line. I added this to prevent that new behavior.

app/webpacker/components/Posts/PostForm.jsx
Brandon-Reichman: Added!

app/webpacker/components/Posts/PostForm.jsx
Brandon-Reichman: Thanks for pointing out `components/EditEvents/index.jsx`! I've changed the implementation to be similar to `EditEvents`, which is way cleaner than what I first attempted. :sweat_smile:

app/webpacker/components/Posts/PostForm.jsx
```diff
@@ -26,6 +28,10 @@ export default function PostForm({
   const [postId, setPostId] = useInputState(post?.id ?? null);
   const [unstickAt, setUnstickAt] = useState(post?.unstick_at ?? null);
 
+  const unsavedChanges = useMemo(() => (
+    !_.isEqual(formTitle, formBody, formTags, formIsStickied, postURL, postId, unstickAt)
```
gregorbg: I don't understand this line. Normally, equality compares only two values A and B, so this should be something like `_.isEqual(originalPostFromProps, postThatUserModified)`. But you're throwing in seven different state variables, which of course are never ever going to be equal to one another.

gregorbg: Also, according to the documentation at https://lodash.com/docs/4.17.15#isEqual, this method doesn't even support more than two args.

gregorbg (changes requested):

app/webpacker/components/Posts/PostForm.jsx
Brandon-Reichman: I've fixed this to be using isEqual to correctly compare the initial post state to the edited post. Tested locally and the behavior works as expected.

Brandon-Reichman:
I honestly forgot about this until I recently looked at my branches. @gregorbg are the most recent changes acceptable?

At the very least, I know having this warning on the delegate reports would be very nice.

gregorbg:
I am definitely fine with the `are-you-sure` kind of changes in the "old" forms.
And the manual `isEqual` check is also good.

I still have a bit "weird" gut feeling about the changes to the Markdown editor. Can you explain to me again why exactly these are necessary (or link me to a previous comment if we debated about it before)

PR #11752, opened 2025-06-09
Added waiting list position to CSV

dunkOnIT:
It's not possible to determine the order of the waiting list in the CSV export, so I've added the waiting list position as an additional column.

### DIFF
--- a/app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
+++ b/app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
@@ -44,7 +44,7 @@ function V3csvExport(selected, registrations, competition) {
       },${
         registration.registrant_id
       },${
-        registration.competing.waiting_list_position || null
+        registration.competing.waiting_list_position || ""
       }\n`;
     });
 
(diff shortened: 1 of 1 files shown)


app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
```diff
@@ -43,7 +43,10 @@ function V3csvExport(selected, registrations, competition) {
         registration.competing.registration_status
       },${
         registration.registrant_id
+      },${
+        registration.competing.waiting_list_position || null
```
gregorbg: Isn't the waiting list position just "null or integer" anyways?
Please let me know if I'm missing something, but this reads as "fall back to null if it's null" to me

dunkOnIT: It comes out as `undefined` without the `|| null`. So not much of a difference either way, `null` just feels a bit better

gregorbg: Then just fall back to an empty string

gregorbg (approved):

PR #11754, opened 2025-06-09
Migrate form section of submit results page to React

danieljames-dj:
This PR aims in migrating a part of submit results page. In the follow-up PRs, I'll migrate the remaining part as well.

In this migration, I've changed the backend API function from `create` to `submit_to_wrt`, this is to match the method with action name. Indeed it's still creating `ResultsSubmission` object, but that will probably be only short term. I'm feeling that the object won't be needed anyway. Though I'm not removing it in this PR, as it will make this PR more complicated.


app/controllers/results_submission_controller.rb
```diff
@@ -121,24 +121,24 @@ def import_from_live
     redirect_to competition_submit_results_edit_path
   end
 
-  def create
+  def submit_to_wrt
```
gregorbg: What exactly is the reasoning behind this name change? Rails has quite strong conventions about `create`, `index`, etc. so I would prefer to keep the name. But open to hear your argument if you think that the name change is really justified.

gregorbg: Further reading: https://guides.rubyonrails.org/routing.html#crud-verbs-and-actions

app/controllers/results_submission_controller.rb
```diff
@@ -121,24 +121,24 @@ def import_from_live
...
     # Check inbox, create submission, send email
     @competition = competition_from_params
 
-    submit_results_params = params.require(:results_submission).permit(:message, :confirm_information)
-    submit_results_params[:competition_id] = @competition.id
-    @results_submission = ResultsSubmission.new(submit_results_params)
+    message = params.require(:message)
+    @results_submission = ResultsSubmission.new({
+                                                  message: message,
+                                                  confirm_information: true,
```
gregorbg, line 134: Why is this changing to a constant `true`? You're effectively removing the choice, so why keep the boolean around at all?

app/controllers/results_submission_controller.rb
```diff
@@ -121,24 +121,24 @@ def import_from_live
...
-
-      flash[:success] = "Thank you for submitting the results!"
       @competition.update!(results_submitted_at: Time.now)
-      redirect_to competition_path(@competition)
+      render status: :ok, json: { success: true }
     else
-      flash[:danger] = "Submitted results contain errors."
       @results_validator = @results_submission.results_validator
-      render :new
+      render status: :internal_server_error, json: { error: "Submitted results contain errors." }
```
gregorbg: I don't think `internal_server_error` is correct here. A 500 means "something really unexpected which we didn't anticipate during development at all has happened and the server **crashed hard**".
But you are actually already expecting the case that a result submission might not be valid, so you need a different response code. Leaving it as a small exercise to you to figure out possible alternatives and pick one :)

app/controllers/results_submission_controller.rb
danieljames-dj: Okay I agree with keeping the name. Changed it back.

app/controllers/results_submission_controller.rb
danieljames-dj: Sorry my bad. I'm using `unprocessable_entity` in similar places, will use the same here.

app/controllers/results_submission_controller.rb
danieljames-dj: I actually tried removing that earlier, but it will change lots of files which looked unrelated to this PR. Can I do that in one of the follow-up PRs?

gregorbg: Okay, if you add a comment explaining why you chose to add constant `true` here ;)

app/webpacker/components/CompetitionResultSubmission/FormToWrt/index.jsx
```diff
@@ -0,0 +1,95 @@
...
+
+  const {
+    mutate: submitToWrtMutate,
+    isLoading,
+    isError: isErrorInSubmission,
+    error: errorInSubmission,
+  } = useMutation({
+    mutationFn: submitToWrt,
+    onSuccess: () => setSuccess(true),
+  });
```
gregorbg: The `useMutation` function actually returns a boolean called `isSuccess`, which does pretty much what the name suggests (analogous to `isError` which you're already using)

app/webpacker/components/CompetitionResultSubmission/FormToWrt/index.jsx
```diff
@@ -0,0 +1,95 @@
...
+
+  const [confirmDetails, setConfirmDetails] = useCheckboxState(false);
+  const [message, setMessage] = useInputState();
+
+  const [success, setSuccess] = useState();
+
+  const {
+    mutate: submitToWrtMutate,
+    isLoading,
+    isError: isErrorInSubmission,
```
gregorbg: The naming here is a bit confusing, at least in my opinion. What about differentiating between `isErrorInPreviousUpload` and `isErrorInCurrentUpload` instead? I am also open to other suggestions.

app/controllers/results_submission_controller.rb
danieljames-dj: Added. I made it in a generic way that anybody can do the cleanup, but I'll do this cleanup myself after this.

app/webpacker/components/CompetitionResultSubmission/FormToWrt/index.jsx
danieljames-dj: The names you suggested looks good to me and indeed makes sense. I've applied it.

On a sidenote, I'm expecting `isErrorInPreviousUpload` to go away once the pages are completely migrated. But I don't promise anything for now on this. In case I do, I'll rename the later as well to avoid confusions.

app/webpacker/components/CompetitionResultSubmission/FormToWrt/index.jsx
danieljames-dj: Oh yes, I knew it. I missed it while coding. Changed now.

gregorbg (approved):

PR #11765, opened 2025-06-10
Split Edit Person actioner workbench to multiple files

danieljames-dj:
This is basically cleanup of EditPerson actioner workbench view. Reason for the cleanup is that currently the file is pretty big and adding more features is difficult. I have plans of adding some features within the workbench, so this split up will be very helpful.

Also, for `closeTicket`, I'm still relying on the `useSaveAction`. This is because I want to reuse the same method that will be used [here](https://github.com/thewca/worldcubeassociation.org/blob/b65bbccfb24d41cd480e6e129e140ab6a2dd5e5d/app/webpacker/components/Tickets/TicketHeader/StatusView.jsx#L28). Currently the linked file is also using `useSaveAction`. I'll be merging both that `useSaveAction` and the `useSaveAction` in this PR together in one of the upcoming PRs.


danieljames-dj:
Thanks Kevin, yes I tested it locally and is used only within WRT when ticket requests come.

PR #11776, opened 2025-06-12
Add Playwright for end-to-end tests

dunkOnIT:
# Info

It does seem like `playwright` and `@playwright/test` were conflicting - I removed them both in my local, and reinstalled just `@playwright.test`. I believe this is the correct approach, as Playwright Test includes the Playwright Library, per [their documentation](https://playwright.dev/docs/library):
> Playwright Library provides unified APIs for launching and interacting with browsers, while **Playwright Test provides all this plus a fully managed end-to-end Test Runner and experience.** [emphasis mine]

I've confirmed that comp PDF generation still works locally even with these changes.

# Todo

- [x] Basic setup with Playwright's example tests
- [x] Assess diff / contents of playwright setup
- [x] Ensure that existing playwright dependencies still work
- [x] Write an actually meaningful E2E test
- [x] Get that test to execute in the CI

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -192,9 +192,12 @@ services:
       - SYS_ADMIN
     init: true
     environment:
-      - DEBIAN_FRONTEND=noninteractive
-      - DEBUG=pw:api
-      - DISPLAY=$DISPLAY
+      DEBIAN_FRONTEND: noninteractive
+      DEBUG: pw:api
+      DISPLAY: $DISPLAY
+      PW_TEST_CONNECT_WS_ENDPOINT: ""
+      PW_TEST_BACKEND_HEALTH_URL: http://wca_on_rails:3000/api/v0/healthcheck
+      PW_TEST_FRONTEND_HEALTH_URL: http://nextjs:3001
     ports:
       - "8089:8089"
     working_dir: /app


playwright.config.ts
```diff
@@ -0,0 +1,80 @@
...
+    // },
+    // {
+    //   name: 'Google Chrome',
+    //   use: { ...devices['Desktop Chrome'], channel: 'chrome' },
+    // },
+  ],
+
+  /* Run your local dev server before starting the tests */
+  webServer: {
+    command: 'cd next-frontend && corepack install && yarn install && yarn dev --port 3001',
```
gregorbg: The part about `corepack install && yarn install` _should_ not be necessary.
In local development environment, you're (supposed to be) running the tests on Docker containers which already have Corepack and Yarn set up and all dependencies installed (happens on every container boot), and in the CI these steps should happen in the GitHub Action runner before the actual tests get triggered.

playwright.config.ts
```diff
@@ -0,0 +1,80 @@
+import { defineConfig, devices } from '@playwright/test';
+
+/**
+ * Read environment variables from file.
+ * https://github.com/motdotla/dotenv
+ */
+// import dotenv from 'dotenv';
+// import path from 'path';
+// dotenv.config({ path: path.resolve(__dirname, '.env') });
```
gregorbg: Why are you leaving this snippet in here if you don't use it? You might just as well remove the code in that case

gregorbg: On a semi-related sidenote: Why don't you use this actually? Your GitHub Actions CI env defines a ton of files, around half of which could be bridged in through a `.env.test` file that gets checked in to CI

.github/workflows/system_tests.yml
```diff
@@ -0,0 +1,68 @@
...
+      PORT: 3000
+      DISABLE_SPRING: 1
+      CAPYBARA_APP_HOST: "" # Explicitly set a blank value to override our default Docker container setup
+      NEW_RELIC_ENABLED: false
+      PAYLOAD_SECRET: amazing-payload-secret
+      AUTH_SECRET: amazing-authjs.secret # Add by executing `npx auth`. Read more: https://cli.authjs.dev
+      OIDC_ISSUER: http://wca_on_rails:3000 # Usually your local (Docker-internal!) Rails root URL
+      OIDC_CLIENT_ID: create-your-own-oauth # Create an OAuth application in the monolith with (at least!) `public`, `oidc`, `profile` and `email` scopes
+      OIDC_CLIENT_SECRET: create-your-own-oauth # See above
+      DATABASE_URI: file:./next-frontend.db
```
gregorbg: All of these could be read by `dotenv`, see my other comment in the Playwright config file

.github/workflows/system_tests.yml
```diff
@@ -0,0 +1,68 @@
...
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    env:
+      RACK_ENV: test
+      RAILS_ENV: test
+      NODE_ENV: test
+      PORT: 3000
+      DISABLE_SPRING: 1
+      CAPYBARA_APP_HOST: "" # Explicitly set a blank value to override our default Docker container setup
```
gregorbg: Why is this necessary? Are you even trying to run a Ruby backend server?

.github/workflows/system_tests.yml
```diff
@@ -0,0 +1,68 @@
...
+          node-version: '20'
+          cache: 'yarn'
+      - name: Install fresh Yarn packages
+        run: bin/yarn install
+      - name: Run Overcommit commit hooks
+        run: |
+          gem install overcommit
+          overcommit --sign
+          overcommit --sign pre-commit
+          overcommit --run
```
gregorbg: I don't think our backend linter should run for frontend integration tests. We have a separate GitHub Action that cares about backend tests.

.github/workflows/system_tests.yml
```diff
@@ -0,0 +1,68 @@
...
+      - name: Activate MySQL # as per https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md
+        run: sudo systemctl start mysql.service
+      - name: Populate database with seeds
+        run: bin/rake db:reset
+      - name: Pre-compile assets for frontend tests
+        run: |
+          bin/bundle exec i18n export
+          bin/rake assets:precompile
```
gregorbg: Why are you precompiling our "old" Ruby-React-glue frontend assets, if they're not under test? Will they be under test by Playwright at some point?

I was under the impression that we want to migrate the pages and use Playwright only for the NextJS part of the website, am I missing something?

.gitignore
```diff
@@ -63,3 +63,10 @@ yarn-debug.log*
 
 # Ignore Font files in accordance to Licensing Agreement
 next-frontend/src/app/(wca)/fonts/TTNormsPro/*
+
+# Playwright
+node_modules/
```
gregorbg: This is not Playwright-related. Why are you adding this here? It should already be gitignored through some other file(s)

playwright.config.ts
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * See https://playwright.dev/docs/test-configuration.
+ */
+export default defineConfig({
+  testDir: './system-tests',
+  /* Run tests in files in parallel */
+  fullyParallel: true,
```
gregorbg, line 17: How does this match the other config about `workers` below?

gregorbg: What I mean is, down there is a comment that says "Opt out of parallel tests on CI". But here, you set the parallellism to `true` no matter what

playwright.config.ts
gregorbg, line 1: Have you ever tried out running Playwright tests locally (in your Docker zoo)?
I currently don't see any config that would point it to our pre-build Playwright runtime running inside Docker.

playwright.config.ts
```diff
@@ -0,0 +1,80 @@
...
+    //   use: { ...devices['Desktop Edge'], channel: 'msedge' },
+    // },
+    // {
+    //   name: 'Google Chrome',
+    //   use: { ...devices['Desktop Chrome'], channel: 'chrome' },
+    // },
+  ],
+
+  /* Run your local dev server before starting the tests */
+  webServer: {
```
gregorbg: Also want to point out that here, you _could_ launch multiple web servers as per https://playwright.dev/docs/test-webserver#multiple-web-servers. This would be useful if you want to write integration tests against NextJS pages that need login (because for that, our Rails backend needs to run) but I am fully conscious that this might be out of scope for this initial proof of concept style PR

playwright.config.ts
gregorbg: As per https://playwright.dev/docs/docker#connecting-to-the-server, it would seem that you have to configure this through an env variable. So you may need to touch `docker-compose.yml` and set this variable in the NextJS container, so that anybody running the `playwright test` command from within that container will automagically have it connect against the Playwright Docker image which we expose.

gregorbg (changes requested):

playwright.config.ts
dunkOnIT: Should we be using this `webServer` part of the config at all if we'll be running tests against the docker container locally? 

Seems better to start the server independently of the `yarn playwright test` command in CI given that the server will already be started independently locally. (The only reason this worked for me, I realize now, is because I deliberately took the `nextjs` container down locally so that I could have `playwright` start the server - I think)

playwright.config.ts
dunkOnIT: Now using it!

.github/workflows/system_tests.yml
dunkOnIT: Done

.github/workflows/system_tests.yml
dunkOnIT: The plan is that Playwright will have tests that interact with the Next frontend, which hits the Ruby backend. The tests don't currently do that, but that's why I've left the Railsy stuff in

playwright.config.ts
dunkOnIT: I haven't tried running it _in_ the docker zoo (ie, from within a particular container), but I have run these tests locally with Docker up and they worked without issue

It's not clear to me why we would want to run these tests from inside Docker, as opposed to just running them in the local developer's console

playwright.config.ts
dunkOnIT: This is the autogenerated playwright config (what you get when running `yarn create playwright`) - so not a conscious decision by me, and clearly something playwright considers a sane default. 

Quick consultation with ChatGPT doesn't suggest anything more complicated than it's safer to run tests sequentially in CI (reduce risk of flaky failures), while running in parallel locally makes sense where rapid debugging is possible and speed is more desirable. 

We could consider parallelism in CI in future, but for now I think it makes sense to leave it disabled until we've got a config/setup that we're happy with and has proven stable

.github/workflows/system-tests.yml
gregorbg, line 1: This file looks useless. Does it serve a purpose?

package.json
```diff
@@ -107,7 +108,9 @@
     "not dead"
   ],
   "devDependencies": {
+    "@playwright/test": "1.52.0",
     "@pmmmwh/react-refresh-webpack-plugin": "^0.6.0",
+    "@types/node": "^24.0.1",
```
gregorbg: We already have this package installed further down in this same `devDependencies` block. I suppose that this addition (that I'm currently commenting on) came from a Playwright script, in which case I am surprised that said script did not detect this.

Regardless, this addition is really not needed.

package.json
```diff
@@ -107,7 +108,9 @@
     "not dead"
   ],
   "devDependencies": {
+    "@playwright/test": "1.52.0",
```
gregorbg: Playwright has a newer version these days, I think it's `1.56.1`. Refer to the one being installed in our latest package JSON on `main`.

playwright.config.ts
```diff
@@ -0,0 +1,80 @@
...
+  /* Retry on CI only */
+  retries: process.env.CI ? 2 : 0,
+  /* Opt out of parallel tests on CI. */
+  workers: process.env.CI ? 1 : undefined,
+  /* Reporter to use. See https://playwright.dev/docs/test-reporters */
+  reporter: 'html',
+  /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
+  use: {
+    /* Base URL to use in actions like `await page.goto('/')`. */
+    baseURL: process.env.PLAYWRIGHT_BASE_URL,
```
gregorbg: This name should better reflect that it is the "base url *which is being used during frontend tests*".
We also use Playwright internally for rendering PDFs when someone clicks "Show Competition PDF" on any competition detail page, so just calling this "the base URL for Playwright" (as per its current name) could cause confusion.

dunkOnIT: Changed

package.json
dunkOnIT: Now on 1.57.0

.github/workflows/system-tests.yml
dunkOnIT: Somehow I had two files - I'm now using this one for the test workflow, as dash is consistent with our other workflow names

package.json
dunkOnIT: Should now be removed

dunkOnIT:
I've updated the PR such that it works in local when I run `yarn playwright test`. This is enough for now to start getting a frontend test framework going - once we have some kind of skeleton going there which is worth running in CI, I'll add the `system-tests` workflow again. For now, I've got it on [its own branch](https://github.com/dunkOnIT/worldcubeassociation.org/tree/gh-actions/playwright-test) in my repo

PR #11779, opened 2025-06-12
Move apply_fixes param from generic_validator to positions_validator

danieljames-dj:
The `apply_fixes` field of GenericValidator is not actually supposed to be generic. It's actually required only for PositionsValidator. Hence, I'm moving the parameter to PositionsValidator.

### DIFF
--- a/app/controllers/admin/scrambles_controller.rb
+++ b/app/controllers/admin/scrambles_controller.rb
@@ -35,7 +35,7 @@ module Admin
       scramble = Scramble.new(scramble_params)
       if scramble.save
         # We just inserted a new scramble, make sure we at least give it correct information.
-        validator = ResultsValidators::ScramblesValidator.new(apply_fixes: true)
+        validator = ResultsValidators::ScramblesValidator.new
         validator.validate(competition_ids: [scramble.competition_id])
         json[:messages] = ["Scramble inserted!"].concat(validator.infos.map(&:to_s))
       else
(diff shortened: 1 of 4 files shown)


gregorbg:
Hm. I understand the logical motivation behind this change, but I don't see its actual advantage in the code.
For what it's worth, the code is becoming more complex (constructor override instead of general-purpose "one size fits all" constructor). This can also be seen by the net positive line count (17-9=+8 lines extra) of this diff.[1]

It also blocks us from introducing new validators with automated fixes in the future. The fact that _right now_, only `PositionsValidator` applies fixes is correct, but in the future this might change. And this PR is actually destroying our easy opportunity to invent such a future validator.

One other compromise solution that I would be open to: Remove the `apply_fixes` from **all** validators, including the `PositionsValidator`, and build the "fix it" functionality into a new `ResultsFixing` routine.

[1]: A net-positive git diff is not a bad thing in general. For example, the architecture I suggested in my last paragraph will (almost) certainly add more lines than it removes. But it introduces a new architecture layer and a clean separation of concerns, so the added lines are "worth it". In your case, I don't see how the relatively large percentage of added lines (net add 8 lines, out of 26 lines changed total means almost one third new line bloat) is worth the complexity, because you're not substantially improving the code interfaces or design architecture. You're just moving a boolean flag to a "more hidden" place for what it's worth

danieljames-dj:
Thanks, Gregor for your detailed comment. I completely support you. I really like the compromise solution. Though the compromise solution may not be too straight forward, I have to do some more brainstorming. But I really love to do it. It's in my checklist now.

PR #11783, opened 2025-06-13
Next i18n

FinnIckler:
This adds I18n capabilities to next using the i18next package. I have a buildscript that generates the json files from our translation.yaml.

What this doesn't do yet:
- add a cookie so the backend can now what language to use when server side rendering
- add generating the translations to the build script
- add a way to change languages in the frontend using the cookie (or localStorage for client)

I can invest time to do that now if we think it's high priority.

I followed this guide https://www.locize.com/blog/i18n-next-app-router linked on the official i18next website

### DIFF
--- a/next-frontend/package.json
+++ b/next-frontend/package.json
@@ -29,6 +29,7 @@
     "@payloadcms/live-preview-react": "^3.42.0",
     "@payloadcms/next": "^3.42.0",
     "@payloadcms/richtext-lexical": "^3.42.0",
+    "@payloadcms/storage-s3": "^3.42.0",
     "@tanstack/react-query": "^5.80.7",
     "accept-language-parser": "^1.5.0",
     "aws4": "^1.13.2",
--- a/.github/actions/build-js-image/action.yaml
+++ b/.github/actions/build-js-image/action.yaml
@@ -14,6 +14,13 @@ outputs:
 runs:
   using: 'composite'
   steps:
+    - name: Copy over symlinked data
+      shell: bash
+      run: |
+        rm -rf ./next-frontend/src/lib/staticData
+        cp -rT ./lib/static_data ./next-frontend/src/lib/staticData
+        rm ./next-frontend/src/lib/staticData/countries.real.json
+        cp ./config/wca-states.json ./next-frontend/src/lib/staticData/countries.real.json
     - name: Build and push JS Docker image
       uses: docker/build-push-action@v6
       id: build-js-image
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -492,14 +492,8 @@ class User < ApplicationRecord
     user.senior_delegates.include?(self)
   end
 
-  def age_in_years
-    years_since_birth = Date.today.year - dob.year
-    birthday_happened = Date.today.day >= dob.day && Date.today.month >= dob.month
-    birthday_happened ? years_since_birth : years_since_birth - 1
-  end
-
   def below_forum_age_requirement?
-    age_in_years < FORUM_AGE_REQUIREMENT
+    (Date.today - FORUM_AGE_REQUIREMENT.years) < dob
   end
 
   def forum_banned?
--- a/config/initializers/doorkeeper.rb
+++ b/config/initializers/doorkeeper.rb
@@ -7,7 +7,7 @@ Doorkeeper.configure do
   # For more information go to
   # https://github.com/doorkeeper-gem/doorkeeper/wiki/Using-Scopes
   default_scopes  :public
-  optional_scopes :dob, :email, :manage_competitions, :openid, :profile
+  optional_scopes :dob, :email, :manage_competitions, :openid, :profile, :cms
 
   base_controller 'ApplicationController'
   base_metal_controller 'ApplicationController'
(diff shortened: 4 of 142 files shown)


next-frontend/eslint.config.mjs
```diff
@@ -22,7 +22,7 @@ const eslintConfig = [
   },
   eslintPluginPrettierRecommended,
   ...pluginQuery.configs["flat/recommended"],
-  globalIgnores(["src/types"]),
+  globalIgnores(["src/types","./build-translations.js"]),
```
gregorbg: Why? I mean, this is a code file maintained by us, so we should be able to lint it. No?

next-frontend/build-translations.js
```diff
@@ -0,0 +1,37 @@
...
+  const fileContent = fs.readFileSync(filePath, 'utf8');
+  const parsed = yaml.load(fileContent);
+
+  Object.entries(parsed).forEach(([topLevelKey, content]) => {
+    const flattened = flattenObject(content);
+    const outputPath = path.join(outputDir, `${topLevelKey}.json`);
+    fs.writeFileSync(outputPath, JSON.stringify(flattened, null, 2));
+    console.log(`✔ Wrote ${outputPath}`);
+  });
+});
```
gregorbg: Not sure if I'm a big fan of code being called top-level. Maybe wrap this into a function and invoke that function in the build script or behind a yarn command or something?

next-frontend/package.json
```diff
@@ -13,6 +13,7 @@
     "types:openapi": "openapi-typescript public/wcaAPI.yaml -o src/types/openapi.ts",
     "types:chakra": "chakra typegen ./src/theme.ts --outdir src/types/chakra --strict",
     "types:payload": "payload generate:types",
+    "build:translations": "node build-translations.js",
```
gregorbg: Is this really the way to go about invoking JS snippets, even on local developer machines?
I know a lot of Yarn packages deliver their own `CLI` scripts, and while they may use Node runtimes under the hood, I'm pretty sure they don't manually invoke `node`. Can you spend a little time on researching yarn scripts / script runners please?

next-frontend/src/lib/i18n/get18n.ts
```diff
@@ -0,0 +1,19 @@
+import i18next from "./i18n";
+// import { headerName } from "./settings";
+// import { headers } from "next/headers";
+
+export async function getT(options?: { keyPrefix?: string }) {
+  // const headerList = await headers();
+  const lng = "en";
```
gregorbg: This should definitely go into a global "fallback language" constant that we can use as single source of truth throughout the whole code base whenever we need a fallback language.

next-frontend/src/lib/i18n/get18n.ts
```diff
@@ -0,0 +1,19 @@
+import i18next from "./i18n";
+// import { headerName } from "./settings";
+// import { headers } from "next/headers";
```
gregorbg: Why are these commented-out imports here?

next-frontend/src/lib/i18n/get18n.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export async function getT(options?: { keyPrefix?: string }) {
+  // const headerList = await headers();
+  const lng = "en";
+  if (lng && i18next.resolvedLanguage !== lng) {
+    await i18next.changeLanguage(lng);
+  }
```
gregorbg: Why do you have to check this during every single method call? I think for someone who just casually browses the website in French, this will trigger a `changeLanguage` for every single translation string that is rendered, no?

next-frontend/src/lib/i18n/get18n.ts
```diff
@@ -0,0 +1,19 @@
...
+  return {
+    t: i18next.getFixedT(
+      lng ?? i18next.resolvedLanguage,
```
gregorbg: Not sure how `lng` can ever be `undefined` if you're hard-setting it to `en` above. So I think this `??` will never ever trigger...

next-frontend/tsconfig.json
```diff
@@ -38,7 +38,8 @@
     ".next/types/**/*.ts"
   ],
   "exclude": [
-    "node_modules"
+    "node_modules",
+    "./build-translations.js"
```
gregorbg, line 42: Also made me realize, why is this written in JS in the first place?

FinnIckler:
The comments about getT can all be explained because as mentioned above, the server side language detection is currently not implemented. 
You can imagine it being const lng = lookIntoCookies first and then headers. Then this will make sense

next-frontend/eslint.config.mjs
FinnIckler: just because it is in js

next-frontend/build-translations.js
FinnIckler: why not? this is a function that is invoked with node

next-frontend/tsconfig.json
FinnIckler: because otherwise it would need to be
- Transpiled in some way or
- Ran with ts-node
Which is a bunch more overhead then just having this in js

next-frontend/package.json
FinnIckler: They definitely do, or they have a `#!/usr/bin/env node`. Check out the openapi-typescript cli.js in your node_modules or the one from chakra-ui in cli/bin/index.js

FinnIckler:
I have set up the cookie interaction with the frontend and backend. The language selector in the navbar is a bit messed up (it makes the navbar super tall when you click on it), but the language changing works!

PR #11785, opened 2025-06-13
Add JSON backed static data to next

FinnIckler:
There is a bunch more data in wca-data.js.erb that isn't JSON backed which will need to be migrated later

### DIFF
--- a/next-frontend/src/components/CountryMap.tsx
+++ b/next-frontend/src/components/CountryMap.tsx
@@ -1,119 +1,7 @@
 "use client";
 
 import { Text } from "@chakra-ui/react";
-
-export const countryCodeMapping: Record<string, string> = {
-  // Africa
-  DZ: "Algeria",
-  EG: "Egypt",
-  GH: "Ghana",
-  KE: "Kenya",
-  MA: "Morocco",
-  NG: "Nigeria",
-  ZA: "South Africa",
-  MU: "Mauritius",
-
-  // Asia
-  AF: "Afghanistan",
-  CN: "China",
-  HK: "Hong Kong",
-  IN: "India",
-  ID: "Indonesia",
-  IR: "Iran",
-  IL: "Israel",
-  JP: "Japan",
-  KZ: "Kazakhstan",
-  KR: "Republic of Korea",
-  KW: "Kuwait",
-  LB: "Lebanon",
-  MY: "Malaysia",
-  PH: "Philippines",
-  QA: "Qatar",
-  SA: "Saudi Arabia",
-  SG: "Singapore",
-  TW: "Chinese Taipei",
-  TH: "Thailand",
-  TR: "Turkey",
-  AE: "United Arab Emirates",
-  VN: "Vietnam",
-  AZ: "Azerbaijan",
-  MN: "Mongolia",
-  NP: "Nepal",
-  LK: "Sri Lanka",
-  BT: "Bhutan",
-  BD: "Bangladesh",
-  KG: "Kyrgyzstan",
-
-  // Europe
-  AT: "Austria",
-  BE: "Belgium",
-  BG: "Bulgaria",
-  CZ: "Czech Republic",
-  DK: "Denmark",
-  EE: "Estonia",
-  FI: "Finland",
-  FR: "France",
-  DE: "Germany",
-  GR: "Greece",
-  HU: "Hungary",
-  IS: "Iceland",
-  IE: "Ireland",
-  IT: "Italy",
-  LV: "Latvia",
-  LT: "Lithuania",
-  NL: "Netherlands",
-  NO: "Norway",
-  PL: "Poland",
-  PT: "Portugal",
-  RO: "Romania",
-  RU: "Russia",
-  RS: "Serbia",
-  SK: "Slovakia",
-  SI: "Slovenia",
-  ES: "Spain",
-  SE: "Sweden",
-  CH: "Switzerland",
-  UA: "Ukraine",
-  GB: "United Kingdom",
-  GE: "Georgia",
-  HR: "Croatia",
-  ME: "Montenegro",
-
-  // North America
-  BS: "Bahamas",
-  BB: "Barbados",
-  CA: "Canada",
-  CR: "Costa Rica",
-  CU: "Cuba",
-  DO: "Dominican Republic",
-  SV: "El Salvador",
-  GT: "Guatemala",
-  HN: "Honduras",
-  JM: "Jamaica",
-  MX: "Mexico",
-  PA: "Panama",
-  PR: "Puerto Rico",
-  TT: "Trinidad and Tobago",
-  US: "United States",
-
-  // South America
-  AR: "Argentina",
-  BO: "Bolivia",
-  BR: "Brazil",
-  CL: "Chile",
-  CO: "Colombia",
-  EC: "Ecuador",
-  PY: "Paraguay",
-  PE: "Peru",
-  UY: "Uruguay",
-  VE: "Venezuela",
-
-  // Oceania
-  AU: "Australia",
-  FJ: "Fiji",
-  NZ: "New Zealand",
-  XO: "Multiple Countries (Oceania)",
-};
+import { countries } from "@/lib/staticData/wca-data";
 
 type CountryMapProps = {
   code: string;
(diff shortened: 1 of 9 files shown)


FinnIckler:
I assume I need to copy this data in the build because the symlink will not resolve in the container

next-frontend/src/app/(wca)/competitions/competitionClient.tsx
```diff
@@ -46,15 +46,13 @@ interface CompetitionsListProps {
...
-    label: name,
-    value: code,
-  }));
-
   const { contains } = useFilter({ sensitivity: "base" });
 
   const { collection, filter } = useListCollection({
-    initialItems: countries,
+    initialItems: Object.entries(countries.byIso2).map(([code, country]) => ({
+      label: country.id,
```
gregorbg, line 53: Is this a temporary placeholder kind of thing? Using the ID as display value would probably lead to curious camelCase artifacts like `UnitedStatesOfAmerica` or perhaps even just `USA`...

next-frontend/src/components/persons/ProfileCard.tsx
```diff
@@ -79,7 +79,10 @@ const ProfileCard: React.FC<ProfileData> = ({
         <DataListRoot variant="profileStat">
           <DataListItem label="WCA ID" value={wcaId} />
           {gender !== "o" && <DataListItem label="Gender" value={gender} />}
-          <DataListItem label="Region" value={countryCodeMapping[regionIso2]} />
+          <DataListItem
+            label="Region"
+            value={countries.byIso2[regionIso2].id}
```
gregorbg: Again, using the ID for display here?

next-frontend/src/components/CountryMap.tsx
```diff
@@ -1,127 +1,15 @@
...
+import { countries } from "@/lib/staticData/wca-data";
 
 type CountryMapProps = {
   code: string;
   bold?: boolean;
 };
 
 const CountryMap = ({ code, bold = false }: CountryMapProps) => {
-  const country = countryCodeMapping[code.toUpperCase()] || "Unknown";
+  const country = countries.byIso2[code.toUpperCase()].id || "Unknown";
```
gregorbg, line 12: ID for display (see above)

next-frontend/src/lib/staticData/wca-data.ts
```diff
@@ -0,0 +1,177 @@
...
+  formatIds: string[];
+};
+
+type RoundType = {
+  id: string;
+  rank: number;
+  name: string;
+  cellName: string;
+  final: boolean;
+};
```
gregorbg: Have we ever discussed the design decision of camelizing this part? Seems that it caused you quite some headache (as well as semi-arcane type defintions)...
And we kinda committed to our API just being snake_case, so it would cause confusion to handle this part of the data source in camelCase.
If we use snake_case here as well, would it simplify data loading? At least it would certainly simplify developer experience, because then everything that is a factual API as well as these "quasi-API" data sources all use the same conventions

next-frontend/src/lib/staticData/wca-data.ts
```diff
@@ -0,0 +1,177 @@
...
+  rawEntities: T[],
+): CamelizedKeys<T>[] {
+  return rawEntities.map(camelizeKeys);
+}
+
+// ----- COUNTRIES -----
+const fictionalCountryIds = _.map(countriesFictive, "id");
+
+const countryData = loadStaticData(
+  // @ts-expect-error currently the fictive countries don't have the same properties as the real countries so ts isn't happy
```
gregorbg: What happenend to the union type that you hinted at earlier?

next-frontend/src/lib/staticData/wca-data.ts
```diff
@@ -0,0 +1,177 @@
...
+);
+
+export const continents = {
+  real: _.map(realContinents, extendContinents),
+};
+
+function extendContinents(continent: Continent) {
+  return {
+    ...continent,
+    name: (t: (path: string) => string) => t(`continents.${continent.name}`),
```
gregorbg: I don't think we should keep the names as part of the standard interface

next-frontend/src/lib/staticData/wca-data.ts
```diff
@@ -0,0 +1,177 @@
...
+  byId: _.mapValues(_.keyBy(formatsData, "id"), extendFormats),
+};
+
+function extendFormats(rawFormat: Format) {
+  return {
+    ...rawFormat,
+    name: (t: (path: string) => string) => t(`formats.${rawFormat.id}`),
+    shortName: (t: (path: string) => string) =>
+      t(`formats.short.${rawFormat.id}`),
+  };
```
gregorbg: Don't think it's worth keeping the names

next-frontend/src/lib/staticData/wca-data.ts
```diff
@@ -0,0 +1,177 @@
...
+  official: _.map(_.filter(eventsData, "isOfficial"), extendEvents),
+  byId: _.mapValues(_.keyBy(eventsData, "id"), extendEvents),
+};
+
+export const WCA_EVENT_IDS = Object.values(events.official).map((e) => e.id);
+
+function extendEvents(rawEvent: Event) {
+  return {
+    ...rawEvent,
+    name: (t: (path: string) => string) => t(`events.${rawEvent.id}`),
```
gregorbg: Again, don't think we have to keep the names

next-frontend/src/lib/staticData/wca-data.ts
```diff
@@ -0,0 +1,177 @@
...
+    formats() {
+      return this.formatIds.map((formatId) => formats.byId[formatId]);
+    },
```
gregorbg: I don't think this needs to stay a method, you can just define a property `formats` that loads/links the data directly.

next-frontend/src/lib/staticData/wca-data.ts
```diff
@@ -0,0 +1,177 @@
...
+    recommendedFormat() {
+      return this.formats()[0];
+    },
```
gregorbg: Same as above, I think the choice of making this a method is a veeeeery old legacy decision due to the way things were back then, but now we have a chance to rectify :smile:

gregorbg (commented):
I have some lingering questions about code design which are unfortunately not quite trivial. Curious to hear your thoughts!

next-frontend/src/app/(wca)/competitions/competitionClient.tsx
FinnIckler: yes, this is temporary until the i18n PR is merged

next-frontend/src/lib/staticData/wca-data.ts
FinnIckler: it would definitely help in a way of not needing all of 
```
type CamelizedKeys<T> = {
  [K in keyof T as K extends string ? CamelCase<K> : K]: T[K];
};

type CamelCase<S extends string> = S extends `${infer Head}_${infer Tail}`
  ? `${Lowercase<Head>}${Capitalize<CamelCase<Tail>>}`
  : Lowercase<S>;

function camelizeKeys<T extends Record<string, unknown>>(
  obj: T,
): CamelizedKeys<T> {
  return _.mapKeys(obj, (v, k) => _.camelCase(k)) as CamelizedKeys<T>;
}

function loadStaticData<T extends Record<string, unknown>>(
  rawEntities: T[],
): CamelizedKeys<T>[] {
  return rawEntities.map(camelizeKeys);
}
```
Which means also means all the types will be inferred from json directly. 
Only downside is that we won't be able to copy paste code that relies on it being camelCase

next-frontend/src/lib/staticData/wca-data.ts
FinnIckler: cool, that makes it easier. I had to make it a method because i18n now works different in the server or the client

next-frontend/src/lib/staticData/wca-data.ts
FinnIckler: how else do you have access to the current `formatIds` though? You can't use `this` in properties

gregorbg: Well you have a parameter `rawEvent` being passed to this method, which is equivalent to `this` in the context of this method :upside_down_face:

FinnIckler: oh duh

PR #11810, opened 2025-06-16
Migrate Upload Results section of admin results page to react

danieljames-dj:
"Upload results" is a section through which the results JSON can be uploaded. This section is present in two different pages for two different set of users:
1. For Delegates - In competition page, open 'Submit Results'
2. For WRT - In competition page, open 'Admin Results' -> 'Upload and check results'

This PR is aimed at migrating just the WRT's view. The Delegate's view will be migrated in a later PR because currently anyway both have very separate backend controllers, and if both pages are migrated, it will change two controllers and there will be significant change in backend which I'm trying to avoid for now.

UI before uploading:
Screenshot 2025-06-16 at 16 11 23

Screen recording after upload:

https://github.com/user-attachments/assets/a959a728-c313-42aa-8290-57465114f3ad

Error case:

https://github.com/user-attachments/assets/82bc21d2-0cc6-4727-9f31-8c3957dcd65b

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -192,11 +192,11 @@ class AdminController < ApplicationController
 
     # This makes sure the json structure is valid!
     if upload_json.import_to_inbox
-      competition.update!(results_submitted_at: Time.now) if competition.results_submitted_at.nil? && mark_result_submitted
+      competition.touch(:results_submitted_at) if competition.results_submitted_at.nil? && mark_result_submitted
       render status: :ok, json: { success: true }
     else
       render status: :unprocessable_entity, json: {
-        error: "Error importing JSON file.",
+        error: upload_json.errors.full_messages,
       }
     end
   end
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -28,7 +28,6 @@ class AdminController < ApplicationController
 
   def new_results
     @competition = competition_from_params
-    @upload_json = UploadJson.new
     @results_validator = ResultsValidators::CompetitionsResultsValidator.create_full_validation
     @results_validator.validate(@competition.id)
   end
@@ -181,23 +180,24 @@ class AdminController < ApplicationController
   end
 
   def create_results
-    @competition = competition_from_params
+    competition = competition_from_params
 
     # Do json analysis + insert record in db, then redirect to check inbox
     # (and delete existing record if any)
-    upload_json_params = params.require(:upload_json).permit(:results_file)
-    upload_json_params[:competition_id] = @competition.id
-    @upload_json = UploadJson.new(upload_json_params)
+    upload_json = UploadJson.new({
+                                   results_file: params.require(:results_file),
+                                   competition_id: competition.id,
+                                 })
+    mark_result_submitted = ActiveRecord::Type::Boolean.new.cast(params.require(:mark_result_submitted))
 
     # This makes sure the json structure is valid!
-    if @upload_json.import_to_inbox
-      @competition.update!(results_submitted_at: Time.now) if @competition.results_submitted_at.nil?
-      flash[:success] = "JSON file has been imported."
-      redirect_to competition_admin_upload_results_edit_path
+    if upload_json.import_to_inbox
+      competition.update!(results_submitted_at: Time.now) if competition.results_submitted_at.nil? && mark_result_submitted
+      render status: :ok, json: { success: true }
     else
-      @results_validator = ResultsValidators::CompetitionsResultsValidator.create_full_validation
-      @results_validator.validate(@competition.id)
-      render :new_results
+      render status: :unprocessable_entity, json: {
+        error: "Error importing JSON file.",
+      }
     end
   end
 
--- a/app/webpacker/components/CompetitionResultSubmission/ImportResultsData/index.jsx
+++ b/app/webpacker/components/CompetitionResultSubmission/ImportResultsData/index.jsx
@@ -1,0 +1,59 @@
+import React, { useState } from 'react';
+import { Accordion, Container, Message } from 'semantic-ui-react';
+import UploadResultsJson from '../UploadResultsJson';
+import WCAQueryClientProvider from '../../../lib/providers/WCAQueryClientProvider';
+
+export default function Wrapper({
+  competitionId,
+  alreadyHasSubmittedResult,
+  isWrtViewing,
+}) {
+  return (
+    <WCAQueryClientProvider>
+      <ImportResultsData
+        competitionId={competitionId}
+        alreadyHasSubmittedResult={alreadyHasSubmittedResult}
+        isWrtViewing={isWrtViewing}
+      />
+    </WCAQueryClientProvider>
+  );
+}
+
+function ImportResultsData({
+  competitionId,
+  alreadyHasSubmittedResult,
+  isWrtViewing,
+}) {
+  const [activeAccordion, setActiveAccordion] = useState(!alreadyHasSubmittedResult);
+
+  return (
+    <Container fluid>
+      <Accordion fluid styled>
+        <Accordion.Title
+          active={activeAccordion}
+          onClick={() => setActiveAccordion((prevValue) => !prevValue)}
+        >
+          Import Results Data
+        </Accordion.Title>
+        <Accordion.Content active={activeAccordion}>
+          {alreadyHasSubmittedResult
+            ? (
+              <Message warning>
+                Some results are already there, importing results data again will override all
+                of them!
+              </Message>
+            )
+            : (
+              <Message info>
+                Please start by selecting a JSON file to import.
+              </Message>
+            )}
+          <UploadResultsJson
+            competitionId={competitionId}
+            isWrtViewing={isWrtViewing}
+          />
+        </Accordion.Content>
+      </Accordion>
+    </Container>
+  );
+}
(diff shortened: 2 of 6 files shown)


app/controllers/admin_controller.rb
```diff
@@ -181,23 +180,24 @@ def delete_results_data
...
+                                 })
+    mark_result_submitted = ActiveRecord::Type::Boolean.new.cast(params.require(:mark_result_submitted))
 
     # This makes sure the json structure is valid!
-    if @upload_json.import_to_inbox
-      @competition.update!(results_submitted_at: Time.now) if @competition.results_submitted_at.nil?
-      flash[:success] = "JSON file has been imported."
-      redirect_to competition_admin_upload_results_edit_path
+    if upload_json.import_to_inbox
+      competition.update!(results_submitted_at: Time.now) if competition.results_submitted_at.nil? && mark_result_submitted
```
gregorbg:
```suggestion
      competition.touch(:results_submitted_at) if competition.results_submitted_at.nil? && mark_result_submitted
```

gregorbg: This is the more idiomatic way to execute "mark current time" on a timestamp column. Note that this does not run model validations except the ones that are intentionally defined as `after_touch` (which I think we don't use in our codebase at all)

So if you have some business logic that complains and doesn't allow `results_submitted_at` to be set based on WRT rules or WCAT rules or other real-world rules (that are defined in our code as Rails validations) then this suggestion is not viable.

app/webpacker/components/CompetitionResultSubmission/ImportResultsData/index.jsx
```diff
@@ -0,0 +1,59 @@
...
+          active={activeAccordion}
+          onClick={() => setActiveAccordion((prevValue) => !prevValue)}
+        >
+          Import Results Data
+        </Accordion.Title>
+        <Accordion.Content active={activeAccordion}>
+          {alreadyHasSubmittedResult
+            ? (
+              <Message warning>
+                Some results are already there, importing results data again will override all
```
gregorbg: This message is a bit misleading. It is easily possible to have some results which are uploaded, but not yet posted. I think "some inbox results are already there" is the smallest possible fix, but may lead to confusion with ordinary Delegates who don't know anything about our inbox and posting processes. So what about "Some results have already been uploaded before"?

app/webpacker/components/CompetitionResultSubmission/ImportResultsData/index.jsx
```diff
@@ -0,0 +1,59 @@
...
+                of them!
+              </Message>
+            )
+            : (
+              <Message info>
+                Please start by selecting a JSON file to import.
+              </Message>
+            )}
```
gregorbg: The result is always a `Message` component in any case, so you could do something like:
```jsx
<Message warning={alreadyHasSubmitted} info={!alreadyHasSubmitted}>
  ...
</Message>
```

app/webpacker/components/CompetitionResultSubmission/UploadResultsJson.jsx
```diff
@@ -0,0 +1,42 @@
...
+import Errored from '../Requests/Errored';
+
+export default function UploadResultsJson({ competitionId, isWrtViewing }) {
+  const [resultFile, setResultFile] = useState();
+  const [markResultSubmitted, setMarkResultSubmitted] = useCheckboxState(isWrtViewing);
+
+  const { mutate: uploadResultsJsonMutate, error, isError } = useMutation({
+    mutationFn: () => uploadResultsJson({ competitionId, resultFile, markResultSubmitted }),
+    onSuccess: () => {
+      window.location.reload();
```
gregorbg, line 18: Why?

gregorbg: (For context, I am not a fan of manual page reloads. The success hook should instead set some React state that then triggers a re-render with the updated "has been uploaded" information. Simplest case is a `useState` that is uesd as `hasUploaded`, so I want to understand why you think this isn't possible here)

app/webpacker/components/CompetitionResultSubmission/UploadResultsJson.jsx
```diff
@@ -0,0 +1,42 @@
...
+      {isWrtViewing && (
+        <Form.Checkbox
+          checked={markResultSubmitted}
+          onChange={setMarkResultSubmitted}
+          label="If results are not marked as submitted, mark it as submitted (this is only visible to WRT)"
+        />
+      )}
+      <Form.Button
+        disabled={!resultFile}
+        onClick={uploadResultsJsonMutate}
```
gregorbg: You should make this button `type="submit"` and move the mutation to the `onSubmit` of the form. This is because `onSubmit` will run browser-sided form validations for you (like checking `required` fields for example). 
May not be too useful here because your form isn't exactly complex, but still a good rule of thumb in general to use `onSubmit` whenever you can

app/controllers/admin_controller.rb
```diff
@@ -181,23 +180,24 @@ def delete_results_data
...
+      render status: :ok, json: { success: true }
     else
-      @results_validator = ResultsValidators::CompetitionsResultsValidator.create_full_validation
-      @results_validator.validate(@competition.id)
-      render :new_results
```
gregorbg, line 200: The effect of this line was that whenever a results upload has not been successful, it automatically triggered validations to show _why_ it hasn't been successful.
It seems that you're now losing this feature, not sure whether that's cool for WRT thought?

gregorbg (changes requested):

app/controllers/admin_controller.rb
danieljames-dj: Oh this is nice, changed.

app/controllers/admin_controller.rb
danieljames-dj: Actually the failure here may not be because of validations that is ran by `results_validator`, it's because of other validations in `UploadJson`. I now noticed that I am not showing WRT the reason why this actually failed.

I've now changed it to send the actual error message to frontend. This way, WRT will know why it failed.

Sidenote: I understand that it's not good to have one set of validations inside `UploadJson` and another set of validations inside `ResultsValidator`. I've already noted this down for future improvements.

app/webpacker/components/CompetitionResultSubmission/ImportResultsData/index.jsx
danieljames-dj: Your version looks better, changed it. Thank you.

app/webpacker/components/CompetitionResultSubmission/ImportResultsData/index.jsx
danieljames-dj: Okay changed.

app/webpacker/components/CompetitionResultSubmission/UploadResultsJson.jsx
danieljames-dj: I basically want to re-render the rails HTML part and for that it's necessary. I will be migrating those to React soon, at that time, this reload won't be needed anymore.

I've added a comment explaining the same. Hope this works.

app/webpacker/components/CompetitionResultSubmission/UploadResultsJson.jsx
danieljames-dj: Okay thanks for that, changed.

app/webpacker/components/CompetitionResultSubmission/UploadResultsJson.jsx
gregorbg: Yup, comments are always good for temporary solutions :D

app/controllers/admin_controller.rb
```diff
@@ -181,23 +180,24 @@ def delete_results_data
...
+                                 })
+    mark_result_submitted = ActiveRecord::Type::Boolean.new.cast(params.require(:mark_result_submitted))
 
     # This makes sure the json structure is valid!
-    if @upload_json.import_to_inbox
-      @competition.update!(results_submitted_at: Time.now) if @competition.results_submitted_at.nil?
-      flash[:success] = "JSON file has been imported."
-      redirect_to competition_admin_upload_results_edit_path
+    if upload_json.import_to_inbox
+      competition.touch(:results_submitted_at) if competition.results_submitted_at.nil? && mark_result_submitted
```
gregorbg, line 195: Just curious, in what circumstance do you need to run `import_to_inbox` but then also ignore the `results_submitted_at` after submission has been successful?
In other words, when will `mark_result_submitted` be `false`?

danieljames-dj: In normal cases (ie, when Delegates upload the JSON), `results_submitted_at` won't get changed. Even the form won't have any way to set this to true. It's because the results will be "marked as submitted", only when they fill the form.

But when WRT uploads a result through WRT's page, then the `results_submitted_at` needs to be updated. This is because WRT doesn't have to fill the form.

I understand it's not so good to have two different UIs to do the same thing, also the setting of `results_submitted_at` looks bit hacky, and I definitely have the plans to improve this. Once the migration to React is completed, there are plenty of cleanups that will help improve this.

gregorbg: Okidoki!

FinnIckler:
For the delegate view. Don't we have that new page already in react with the upload from WCA Live view that we added?

danieljames-dj:
@FinnIckler No, it's still in rails HTML: https://github.com/thewca/worldcubeassociation.org/blob/main/app/views/results_submission/new.html.erb

PR #11811, opened 2025-06-16
Make rubymoney support offline

danieljames-dj:
Few months ago, there were some changes made related to RubyMoney, and because of those, currently developers cannot start server when there is no internet connection. This PR aims at silently catching no-internet error during local development.

### DIFF
--- a/config/initializers/ruby_money.rb
+++ b/config/initializers/ruby_money.rb
@@ -21,7 +21,11 @@ if Rails.env.production? && EnvConfig.WCA_LIVE_SITE? && !EnvConfig.ASSETS_COMPIL
   Money.default_bank = mclb
 else
   eu_bank = EuCentralBank.new
-  eu_bank.update_rates
+  begin
+    eu_bank.update_rates
+  rescue SocketError, OpenURI::HTTPError => e
+    Rails.logger.warn "Failed to update EU Central Bank rates: #{e.message}. Using default rates."
+  end
   Money.default_bank = eu_bank
 end
 


gregorbg (commented):
Alternative suggestion, since you know I'm not a friend of `rescue`ing errors:
Introduce an optional environment variable, that can block the code even _trying_ to refetch: `eu_bank.update_rates unless EnvConfig.DEVELOPMENT_OFFLINE_MODE` or similar

danieljames-dj:
Yes that looks like a great suggestion. I've implemented it.

env_config.rb
```diff
@@ -31,6 +31,7 @@
     mandatory :CDN_ASSETS_DISTRIBUTION_ID, :string
     mandatory :REGISTRATION_QUEUE, :string
     mandatory :LIVE_QUEUE, :string
+    optional :DEVELOPMENT_OFFLINE_MODE, :bool, false
```
gregorbg: You shouldn't need this. A variable introduced under this `if Rails.env.production?` branch that has `DEVELOPMENT` in its name is very confusing

gregorbg: You could change your call-site to `unless Rails.env.development? && EnvConfig.OFFLINE?` to make sure you never query a non-existing variable.

gregorbg: We already have several other use-cases of this pattern, see the block of definitions close to `DISABLE_BULLET`. There are 5 or so variables over there that simply "do not exist" in prod mode, and your new one should definitely become one of them.

gregorbg (changes requested):

env_config.rb
danieljames-dj: Okay, changed. I didn't new we can have a env just for development. Thanks for that.

env_config.rb
```diff
@@ -56,6 +56,7 @@
     optional :WRC_WEBHOOK_URL, :string, ''
     optional :REGISTRATION_QUEUE, :string, ''
     optional :LIVE_QUEUE, :string, ''
+    optional :OFFLINE_MODE, :bool, false
```
gregorbg: You can still keep the `DEVELOPMENT_` in the name though, to make clear that this variable only affects developer experience, and not our production servers

danieljames-dj: Done

PR #11818, opened 2025-06-16
Add About Us Page

FinnIckler:
I  went pretty barebones for a start. Everything but the buttons on the call to action can be implemented just using the richt text editor.

Adding     "postinstall": "chakra typegen ./src/theme.ts" to our package.json actually broke the docker build (as it as no access to the src files during dependency install so I haven't tested this yet


next-frontend/src/app/(wca)/faq/page.tsx
```diff
@@ -23,6 +23,10 @@ export default async function FAQ() {
 
   const faqCategories = faqCategoriesResult.docs;
 
+  if (faqCategories.length === 0) {
+    return <Heading> No FAQ categories found, add some!</Heading>;
```
gregorbg: Extraneous space
```suggestion
    return <Heading>No FAQ categories found, add some!</Heading>;
```

next-frontend/src/app/(wca)/about/page.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+export default async function About() {
+  const payload = await getPayload({ config });
+
+  const aboutPage = await payload.findGlobal({ slug: "about-us-page" });
+
+  const aboutItems = aboutPage.blocks;
+
+  if (aboutItems.length === 0) {
+    return <Heading> No About Items, add some!</Heading>;
```
gregorbg: Extraneous space
```suggestion
    return <Heading>No About Items, add some!</Heading>;
```

next-frontend/src/components/about/AboutUsItem.tsx
```diff
@@ -0,0 +1,47 @@
...
+              width: "100%",
+              height: "auto",
+              borderRadius: "1rem",
+              objectFit: "cover",
+            }}
+          />
+        </Box>
+      )}
+    </Flex>
+  );
```
gregorbg: Why are you doing all of the `Flex`ing (including the outer component as well as the inner `flex="1"`) yourself? Chakra usually has very good defaults for this...

next-frontend/src/components/about/AboutUsItem.tsx
```diff
@@ -0,0 +1,47 @@
...
+            src={image.url}
+            alt={image.alt || title}
+            width={800}
+            height={400}
+            style={{
+              width: "100%",
+              height: "auto",
+              borderRadius: "1rem",
+              objectFit: "cover",
+            }}
```
gregorbg: You should not need to wrap this inside `style`. Chakra has an extremely extensive CSS-in-JS mechanism built in that supports `borderRadius`, `fit`, etc. directly on the `Image` component itself.

About `width` and `height`, why are you setting those twice?

gregorbg: I think you may want to look into https://www.chakra-ui.com/docs/components/image#html-width-and-height

next-frontend/src/components/about/CallToAction.tsx
```diff
@@ -0,0 +1,44 @@
...
+            <Button
+              key={i}
+              as="a"
+              colorScheme="blue"
+              variant={i === 0 ? "solid" : "outline"}
+              size="lg"
+            >
+              <a href={button.label} rel={"noopener"}>
+                {button.label}
+              </a>
```
gregorbg: I don't think you need to use low-level `a` tags here manually. Both Chakra and NextJS have a `Link` component, and you should use Chakra's if it's an external link or NextJS' if it's an internal link to one of our own pages within the app

next-frontend/src/globals/About.ts
```diff
@@ -0,0 +1,98 @@
+import { GlobalConfig } from "payload";
+import { markdownConvertedField } from "@/collections/helpers";
+
+export const AboutUsPage: GlobalConfig = {
+  slug: "about-us-page",
+  label: "About Us Page",
+  access: {
+    read: () => true,
+  },
```
gregorbg: Redundant, as this is the default anyways.

next-frontend/src/globals/About.ts
```diff
@@ -0,0 +1,98 @@
...
+  fields: [
+    {
+      name: "blocks",
+      type: "blocks",
+      required: true,
+      blocks: [
```
gregorbg: You can pull the individual block choices out into their own `const` definitions using the `Block` type. Makes the nesting a little less deep and therefor easier to read

gregorbg: Also, purely optional but more FYI: If you pass an `interfaceName` to the individual `Block`s, the codegen will expose that name when generating types, meaning you can reference the type of your blocks when typehinting.

This isn't immediately relevant for your current PR, because your blocks are relatively simple and you can pass around the individual "properties" of your blocks directly to the child components when rendering them.
But if you take a look at the `Home` page code, you will see that you can also pass down the block as a whole to child components using this `interfaceName` reference.

gregorbg (changes requested):
Cool! I like the new version of the About Us page in the second screenshot very very much, just a few nitpicks in the code to figure out

next-frontend/src/components/about/AboutUsItem.tsx
FinnIckler: not sure what you mean, can you point me to any references? I can replace it with a HStack, but I still need the flex=1 if I want to have the image all the way to the right

next-frontend/src/components/about/CallToAction.tsx
FinnIckler: You do, as per Button as Link documentation in https://chakra-ui.com/docs/components/button#as-link

gregorbg: The `a` link over there is just a generic "stand-in" example. Nothing prevents you from using a "clean" `Link` component.

gregorbg: In fact, I am surprised that Chakra isn't using its own link in this case.

next-frontend/src/components/about/CallToAction.tsx
```diff
@@ -0,0 +1,45 @@
...
+          {buttons.map((button, i) => (
+            <Button
+              key={i}
+              as="a"
+              colorScheme="blue"
+              variant={i === 0 ? "solid" : "outline"}
+              size="lg"
+              asChild
+            >
+              <a href={button.label} rel={"noopener"}>
```
gregorbg: Also, you should use `url` instead of label here :sweat_smile: 
```suggestion
              <a href={button.url} rel={"noopener"}>
```

next-frontend/src/components/about/AboutUsItem.tsx
gregorbg: Haven't spun up the code on my local yet, but if your goal is "have the image stick to the very end" then maybe a simple `justify: "space-between"` on the outer stack will do?

FinnIckler: that does work!

next-frontend/src/components/about/AboutUsItem.tsx
```diff
@@ -0,0 +1,41 @@
...
+      </VStack>
+
+      {image?.url && (
+        <Box flexShrink={0} maxW="500px" w="full">
+          <Image
+            src={image.url}
+            alt={image.alt || title}
+            borderRadius={"1rem"}
+            objectFit="cover"
+          />
```
gregorbg: Also, you don't have to resort to only Chakra image or only NextJS image. You can combine them per https://www.chakra-ui.com/docs/components/image#nextjs-image.

Define "as much as you can" (especially styling and theming props) on the outer Chakra image, and then define "raw" parameters like the `src` and `alt` (and potentially everything related to loading optimizations) on the inner NextJS image

FinnIckler: then I need to set `width` and `height` though right? Which is unknown because it is a dynamic image

gregorbg: Do you? I'm not sure...
https://nextjs.org/docs/app/api-reference/components/image#props does not list `width` and `height` as required...

FinnIckler: if you scroll down you will see:
```
You must set both width and height properties unless:

    The image is statically imported.
    The image has the [fill property](https://nextjs.org/docs/app/api-reference/components/image#fill)
```

FinnIckler: I will try if fill works fine though

FinnIckler: nope, fill makes it fill the whole screen

gregorbg: This is interesting... From the documentation of `fill`:
>  A boolean that causes the image to expand to the size of the parent element.

This means that the image is somehow "incorrectly" judging the width of the parent element. Or rather, the parent element's width is off... I will need to try this out locally and play around with it a bit.

FinnIckler: It also says "`The parent element must assign position: "relative", "fixed", "absolute".`", which doesn't look like it does

gregorbg (approved):

PR #11824, opened 2025-06-18
Fix birthday determination bug in `user#age_in_years`

dunkOnIT:
We had a bug in the forum age restriction where a user born in the same year, earlier month and later date caused the system to think their birthday had not yet passed - fixed in this PR, and added a test for it.

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -761,7 +761,7 @@ RSpec.describe User do
     end
   end
 
-  describe '#below_forum_age_requirement?' do
+  describe '#below_forum_age_requirement?', :tag do
     let(:user) { create(:user) }
 
     it 'true when user under 13' do
@@ -775,7 +775,7 @@ RSpec.describe User do
     end
 
     it 'false when user older than 13' do
-      user.dob = Date.today.advance(days: 1, years: -13)
+      user.dob = Date.today.advance(days: -1, years: -13)
       expect(user.below_forum_age_requirement?).to be(false)
     end
   end


dunkOnIT:
No good reason - that's much simpler!

app/models/user.rb
```diff
@@ -492,14 +492,8 @@ def senior_delegate_for?(user)
...
 
-  def age_in_years
-    years_since_birth = Date.today.year - dob.year
-    birthday_happened = Date.today.day >= dob.day && Date.today.month >= dob.month
-    birthday_happened ? years_since_birth : years_since_birth - 1
-  end
-
   def below_forum_age_requirement?
-    age_in_years < FORUM_AGE_REQUIREMENT
+    (Date.today - 13.years) < dob
```
gregorbg: Perhaps this should be `<=` so that you can access the forums on the day of your 13th birthday

gregorbg: Also, it seems you're completely ditching `FORUM_AGE_REQUIREMENT` now, is there a special reason for this? `Date.today - FORUM_AGE_REQUIREMENT.years` should :tm: work

spec/models/user_spec.rb
```diff
@@ -761,29 +761,6 @@
...
-      user.dob = Date.today.advance(years: -10)
-      expect(user.age_in_years).to be(10)
-    end
-
-    it 'doesnt count the year if before user birthday' do
-      user.dob = Date.today.advance(years: -10, days: 1)
-      expect(user.age_in_years).to be(9)
-    end
-  end
-
```
gregorbg, line 786: These very relevant test cases should be migrated to the `below_forum_age_requirement` part below -- unless they already are?
I mean, fake a user with these "edge case" birthdates and then run the `below?` method on them.

gregorbg (changes requested):

app/models/user.rb
dunkOnIT: > Perhaps this should be <= so that you can access the forums on the day of your 13th birthday

That case is already covered in the current logic - see `user_spec#800`

And changed to use the constant :+1:

spec/models/user_spec.rb
dunkOnIT: We already have tests covering `below_forum_age_requirement` - some cases were extra here because of the implementation quirks of `age_in_years`

PR #11825, opened 2025-06-19
Duplicate checker dashboard for Delegates

danieljames-dj:
This is a page where the Delegates can identify potential delegates in the similar way how WRT is identifying.

### DIFF
--- a/app/controllers/results_submission_controller.rb
+++ b/app/controllers/results_submission_controller.rb
@@ -17,7 +17,7 @@ class ResultsSubmissionController < ApplicationController
   end
 
   def last_duplicate_checker_job_run
-    last_job_run = DuplicateCheckerJobRun.where(competition_id: params.require(:competition_id)).first
+    last_job_run = DuplicateCheckerJobRun.find_by(competition_id: params.require(:competition_id))
 
     render status: :ok, json: last_job_run
   end


app/controllers/results_submission_controller.rb
```diff
@@ -12,6 +12,31 @@ def new
...
+    @competition = competition_from_params
+  end
+
+  def potential_duplicates_data
+    competition = competition_from_params
+
+    render status: :ok, json: {
+      duplicate_checker_last_fetch_status: competition.duplicate_checker_last_fetch_status,
+      duplicate_checker_last_fetch_time: competition.duplicate_checker_last_fetch_time,
+      similar_persons: competition.potential_duplicate_people.order(:score).reverse_order,
```
```suggestion
      similar_persons: competition.potential_duplicate_people.order(score: :desc),
```

app/webpacker/components/DuplicateChecker/index.jsx
```diff
@@ -0,0 +1,59 @@
...
+    <WCAQueryClientProvider>
+      <DuplicateChecker competitionId={competitionId} />
+    </WCAQueryClientProvider>
+  );
+}
+
+function DuplicateChecker({ competitionId }) {
+  const { data, isLoading, isError } = useQuery({
+    queryKey: ['potential-duplicates-data'],
+    queryFn: () => getPotentialDuplicatesData({ competitionId }),
```
gregorbg: If `competitionId` is part of the `queryFn` parameters, it *must* be part of the `queryKey` as well.

app/webpacker/components/DuplicateChecker/index.jsx
```diff
@@ -0,0 +1,59 @@
...
+  });
+
+  const queryClient = useQueryClient();
+
+  const { mutate: computePotentialDuplicatesMutate } = useMutation({
+    mutationFn: computePotentialDuplicates,
+    onSuccess: ({ duplicate_checker_last_fetch_status: duplicateCheckerLastFetchStatus }) => {
+      queryClient.setQueryData(
+        ['potential-duplicates-data'],
```
gregorbg: See above re changing the query key

gregorbg (commented):
So far, I feel like the overall idea of the code is fine. I like that you specified the algorithm and the way that a duplicate person is linked to the "original" as well as to the potential duplicate.

The only thing that I am not a fan of (at all) is bloating the `competitions` table even more. Especially with all those `*_last_run` parameters. And `PotentialDuplicatePerson` being linked to a competition also doesn't make much sense to me.

Instead, what if you introduce a new database table where each row corresponds to one run of the job? Like `DuplicateCheckerResult`. When you start the job, the row can be already created using the `fetch_status` enum that currently lives under competition.rb, and when the run is finished the row corresponding to this run is updated.

The individual "jobRun" entities can then be linked to a competition, so that we keep a history of runs, the competition can easily access its "last run" or "most recent run" and then all of the remaining logic happens within that model itself, away from `competition`. The `PotentialDuplicatePerson` can then also be linked to the run, not to the parent competition (but if there is a justified use case where you really, _really_ need the competition to directly access duplicates, you can still use `has_many :duplicate_persons, through: :duplication_job_runs` for pretty much free)

app/webpacker/components/DuplicateChecker/index.jsx
danieljames-dj: Oh yes, I missed it. Thanks for pointing. I've changed it now.

app/webpacker/components/DuplicateChecker/index.jsx
danieljames-dj: Done

danieljames-dj:
Thanks @gregorbg for your suggestion on the schema. I've implemented it now, and marked this PR as ready for review.

app/jobs/compute_potential_duplicates.rb
```diff
@@ -0,0 +1,32 @@
...
+
+class ComputePotentialDuplicates < ApplicationJob
+  def perform(job)
+    persons_cache = Person.select(:id, :wca_id, :name, :dob, :country_id)
+    new_comers = job.competition
+                    .registrations
+                    .includes(:user)
+                    .select { it.wcif_status == "accepted" }
+                    .map(&:user)
+                    .select { it.wca_id.nil? }
```
gregorbg: This should be feasible through one or two clever `has_many` relation on competition. Something along the lines of `has_many :accepted_registrations, -> { competing_status_accepted? }, class_name: "Registration"` and `has_many :accepted_newcomers, -> { where(wca_id: nil) }, class_name: "User", through: :accepted_registrations`

gregorbg: Beware, I wrote this completely freestyle on GitHub without checking the actual code. Probably need to polish / fix the details

app/controllers/results_submission_controller.rb
```diff
@@ -12,6 +12,28 @@ def new
...
   end
 
+  def duplicate_checker
+    @competition = competition_from_params
+  end
+
+  def last_duplicate_checker_job
+    competition = competition_from_params
+
+    render status: :ok, json: competition.duplicate_checker_jobs.order(created_at: :desc).first
```
gregorbg: You can implement this either as a `default_scope { order(created_at: :desc) }` on the `DuplicateCheckerJob` or as a scope on the `has_many :duplicate_checker_jobs, -> { order(...) }` on Competition

app/controllers/results_submission_controller.rb
```diff
@@ -12,6 +12,28 @@ def new
...
+  end
+
+  def compute_potential_duplicates
+    competition = competition_from_params
+
+    job = DuplicateCheckerJob.create!(
+      competition_id: competition.id,
+      status: DuplicateCheckerJob.statuses[:not_started],
+    )
+    ComputePotentialDuplicates.perform_later(job)
```
gregorbg: The `ComputePotentialDuplicates` job should create its `DuplicateCheckerJob` in its own job body, instead of the record being passed in from the outside

app/models/duplicate_checker_job.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+class DuplicateCheckerJob < ApplicationRecord
```
gregorbg: This should probably be called `DuplicateCheckerJobRuns` because you're not serializing the actual **job** itself. You are serializing the information that the job ran (and when) and what its results are

gregorbg: Don't forget to rename the table in the migration and schema.rb accordingly

app/controllers/results_submission_controller.rb
gregorbg: Also, unrelated: I am not a big fan of "abusing" the technical `created_at` and `updated_at` timestamps for business logic. I know that we do that in other places of the code, but those are all legacy use cases that I would like to get rid of.

Why are you not using `start_time` and/or `end_time` here?

danieljames-dj: Adding `default_scope` in `duplicate_checker_job` looks like a good idea. Done.

Regarding using `created_at`, TBH I wanted to use `start_time`, but I was probably out of mind that time 😅 Changed it to `start_time`.

app/controllers/results_submission_controller.rb
danieljames-dj: If I create this job/run inside `ComputePotentialDuplicates`, then I won't have a job to return. The frontend is expecting a job/run so that it will know the status of the same.

I can solve it temporarily by not sending anything to frontend, and let the frontend create a dummy job with status "in progress". But the problem is that if the user refreshes immediately, and if job is not yet started (which is very very rare), then the user will be shown the result of past job which is probably no longer relevant and might confuse the user.

One solution I can think of is, how about having a static method inside `DuplicateCheckerJob` which will create the job and start the async job? I'm not sure if it has much difference from current solution, but I'm open to discussion.

app/jobs/compute_potential_duplicates.rb
danieljames-dj: That's really clever, thanks for that. Done. (made some minor changes)

app/models/duplicate_checker_job.rb
danieljames-dj: Done

app/controllers/results_submission_controller.rb
gregorbg: With the bigger schema migrations in place, I feel less strongly about this now. But you should definitely rename the variable from `job` to something like `job_run` or `run_record` or something. To make clear that the thing you're passing in is not the job itself.

app/controllers/results_submission_controller.rb
```diff
@@ -12,6 +12,28 @@ def new
...
   end
 
+  def newcomer_checks
+    @competition = competition_from_params
+  end
+
+  def last_duplicate_checker_job_run
+    competition = competition_from_params
+
+    render status: :ok, json: competition.duplicate_checker_job_runs.first
```
gregorbg: It's relatively expensive to load the whole competition from the DB, because you don't really need it. You just need it as a "utility" to access the `has_many`, but I think you should be able to do something similar by `DuplicateCheckerJobRuns.find_by(competition_id: params[...])`

gregorbg: (not a complete performance desaster though, if using `competition_from_params` is really the only way, then please explain why and I shall approve it)

app/jobs/compute_potential_duplicates.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+class ComputePotentialDuplicates < ApplicationJob
+  def perform(job)
+    persons_cache = Person.select(:id, :wca_id, :name, :dob, :country_id)
+
+    job.touch(:start_time)
+    job.update!(status: DuplicateCheckerJobRun.statuses[:in_progress])
+    job.competition.accepted_newcomers.each do |user|
```
gregorbg, line 10: An empty line before this statement won't hurt. Group the DB updates together into one "logical" unit, and then start the `accepted_newcomers.each` loop in a separate block

app/jobs/compute_potential_duplicates.rb
```diff
@@ -0,0 +1,26 @@
...
+        PotentialDuplicatePerson.create!(
+          duplicate_checker_job_run_id: job.id,
+          original_user_id: user.id,
+          duplicate_person_id: person.id,
+          algorithm: PotentialDuplicatePerson.algorithms[:jarowinkler],
+          score: score_decimal * 100,
+        )
+      end
+    end
+    job.touch(:end_time)
```
gregorbg, line 24: Again, empty line before here will do wonders for readability

app/jobs/compute_potential_duplicates.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+class ComputePotentialDuplicates < ApplicationJob
+  def perform(job)
+    persons_cache = Person.select(:id, :wca_id, :name, :dob, :country_id)
+
+    job.touch(:start_time)
+    job.update!(status: DuplicateCheckerJobRun.statuses[:in_progress])
+    job.competition.accepted_newcomers.each do |user|
+      name = user.name
```
gregorbg: Do you need this `name` variable in more than one place? If not, then you can access `user.name` inline directly inside `compute_similar_persons`

app/jobs/compute_potential_duplicates.rb
```diff
@@ -0,0 +1,26 @@
...
+      country_id = user.country.id
```
gregorbg: Same as above

app/models/duplicate_checker_job_run.rb
```diff
@@ -0,0 +1,27 @@
...
+  belongs_to :competition
+
+  default_scope -> { order(start_time: :desc) }
+
+  enum :status, {
+    not_started: 'not_started',
+    in_progress: 'in_progress',
+    success: 'success',
+    failed: 'failed',
+  }, prefix: true
```
gregorbg: You can actually specify `default: :not_started` explicitly. Not a big difference in functionality, but makes it more clear in terms of the "lifecycle" design

gregorbg: On second thought, what do you think about naming this `lifecycle` instead of `status`? I don't have any strong feelings on this, so feel free to discard this idea. But `status` is a bit generic...

app/models/duplicate_checker_job_run.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  def similar_persons
+    potential_duplicate_people.order(score: :desc)
```
gregorbg: Again, this is a good candidate for a `default_scope` or at least for a scoped access directly in `has_many :potential_duplicate_people`.

app/models/duplicate_checker_job_run.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  DEFAULT_SERIALIZE_OPTIONS = {
+    methods: %w[similar_persons],
```
gregorbg: If you follow my other comment above, you can then actually change this into `include: %w[potential_duplicate_people]`

app/controllers/results_submission_controller.rb
danieljames-dj: That's a good improvement, implemented it.

app/controllers/results_submission_controller.rb
danieljames-dj: Okay, renamed to `job_run`.

app/jobs/compute_potential_duplicates.rb
danieljames-dj: Done

app/jobs/compute_potential_duplicates.rb
danieljames-dj: Done

app/jobs/compute_potential_duplicates.rb
danieljames-dj: Done

app/jobs/compute_potential_duplicates.rb
danieljames-dj: Done

app/models/duplicate_checker_job_run.rb
danieljames-dj: Added `default`. TBH I didn't like the name `lifecycle` much. I also understand `status` is worse as it is generic. How about `run_status`? (other options I got from LLM: `state`, `execution_status`, `job_status`, `processing_state`, `lifecycle_status`)

app/models/duplicate_checker_job_run.rb
danieljames-dj: Okay, added scope in `has_many`.

app/models/duplicate_checker_job_run.rb
danieljames-dj: Done, thanks

app/models/duplicate_checker_job_run.rb
gregorbg: Yup, `run_status` seems good to me!

db/migrate/20250619163228_create_potential_duplicate_people.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class CreatePotentialDuplicatePeople < ActiveRecord::Migration[7.2]
+  def change
+    create_table :potential_duplicate_people do |t|
```
gregorbg: Feel free to go with `_persons` here instead. You will have to override `self.table_name = "..."` in the model as well.

gregorbg: I know that the Rails default is "people" as a plural, but for historical reasons we already use "Persons" in sooooo many other places that I think it will be good to keep it consistent.

app/models/duplicate_checker_job_run.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+class DuplicateCheckerJobRun < ApplicationRecord
+  has_many :potential_duplicate_people, -> { order(score: :desc) }, dependent: :destroy
```
gregorbg: This should also be `potential_duplicate_persons`, see my other comment in the migration below

(TL;DR: For historical reasons and consistency reasons, we should prefer "Persons")

danieljames-dj: Okay changed.

db/migrate/20250619163228_create_potential_duplicate_people.rb
danieljames-dj: Okay changed.

app/models/duplicate_checker_job_run.rb
danieljames-dj: Okay changed to `run_status`.

app/controllers/results_submission_controller.rb
```diff
@@ -12,6 +12,28 @@ def new
     @results_validator.validate(@competition.id)
   end
 
+  def newcomer_checks
+    @competition = competition_from_params
+  end
+
+  def last_duplicate_checker_job_run
+    last_job_run = DuplicateCheckerJobRun.where(competition_id: params[:competition_id]).first
```
gregorbg: I think `where(...).first` is the same as `find_by(...)`. Can you please check to see whether that respects your `default_scope`? (I'm already 90% sure that it should, just need those extra 10% for confirmation)

app/controllers/results_submission_controller.rb
```diff
@@ -12,6 +12,28 @@ def new
...
+
+    render status: :ok, json: last_job_run
+  end
+
+  def compute_potential_duplicates
+    competition = competition_from_params
+
+    job_run = DuplicateCheckerJobRun.create!(
+      competition_id: competition.id,
+      run_status: DuplicateCheckerJobRun.run_statuses[:not_started],
```
gregorbg: This matches the default value and can therefor be dropped

app/controllers/results_submission_controller.rb
```diff
@@ -12,6 +12,28 @@ def new
...
+  end
+
+  def last_duplicate_checker_job_run
+    last_job_run = DuplicateCheckerJobRun.where(competition_id: params[:competition_id]).first
+
+    render status: :ok, json: last_job_run
+  end
+
+  def compute_potential_duplicates
+    competition = competition_from_params
```
gregorbg: Here's another case where you're loading the whole competition into memory, just so that you can access its ID. Nothing else is done with this `competition` as far as I can tell.
If that's indeed the case (and I'm not missing something) then you can also refactor this to pass the `competition_id` in the job run `create!` below directly from `params`.

app/webpacker/components/NewcomerChecks/index.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import WCAQueryClientProvider from '../../lib/providers/WCAQueryClientProvider';
+import DuplicateChecker from './DuplicateChecker';
+
+export default function Wrapper({ competitionId }) {
```
gregorbg: You can move this whole component (and get rid of the whole file) into the `DuplicateChecker` definition.

Just declare the "real" `DuplicateChecker` component in there without any `export`, and then add this code snippet in there as well as the new `export default`. This is our standard practice when the Wrapper really does nothing more than adding a few contexts.

(If your wrapper has more complicated logic, like creating a reducer, then your approach of a separate file would be more justified)

db/migrate/20250619163228_create_potential_duplicate_persons.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class CreatePotentialDuplicatePersons < ActiveRecord::Migration[7.2]
+  def change
+    create_table :potential_duplicate_persons do |t|
+      t.references :duplicate_checker_job_run, null: false, foreign_key: { to_table: :duplicate_checker_job_runs }
```
gregorbg: I think in this specific case (name of the reference is exactly the table name in plural) Rails _should_ be able to infer the `to_table` automatically and you can just set `foreign_key: true`

db/migrate/20250619163228_create_potential_duplicate_persons.rb
```diff
@@ -0,0 +1,14 @@
...
+      t.references :original_user, type: :integer, null: false, foreign_key: { to_table: :users }
+      t.references :duplicate_person, type: :integer, null: false, foreign_key: { to_table: :persons }
+      t.string :algorithm, null: false
```
gregorbg: This might be misleading for other devs trying to understand your code, because this `algorithm` only affects the name string comparison. But we also compare DOB (and nationality??) which affects the resulting `score`, but is completely unchanged by `algorithm`.
Perhaps something like `name_matching_algorithm` would be more exact? It's longer but it's also more accurate about what actually happens in the code

app/views/competitions/_nav.html.erb
```diff
@@ -79,6 +79,13 @@
           }
         end
         if current_user&.can_upload_competition_results?(@competition)
+          unless @competition.probably_over?
+            nav_items << {
+              text: t('.menu.newcomer_checks'),
+              path: competition_newcomer_checks_path(@competition),
+              icon: "users",
+            }
+          end
```
gregorbg: With this menu item, the feature gets "launched" immediately. Would you be open to hide this feature behind some sort of test flag, or maybe some `unless Rails.env.production? && EnvConfig.WCA_LIVE_SITE?` check, to let a few more people (especially WRT) test this on staging first?

gregorbg (changes requested):
More thoughts and ideas. Sorry for the continuous stream of reviews, but this is a pretty big feature so my thoughts keep trickling in over time

app/controllers/results_submission_controller.rb
danieljames-dj: Yes I already verified that `where(...).first` is respecting the `default_scope`. Should I change to `find_by(...)`? I assume that also respects `default_scope` (but didn't verify for `find_by(...)`).

app/controllers/results_submission_controller.rb
danieljames-dj: Ya good catch. Done. I made it `params.require(:competition_id)`, and the same to the above method as well (it was `params[:competition_id]` earlier)

app/controllers/results_submission_controller.rb
danieljames-dj: Okay done

app/views/competitions/_nav.html.erb
danieljames-dj: Sounds good. How about instead of just staging, open it in prod itself but only to WRT (`can_admin_results?`)? At least the DOB will be different, and can actually use for the competitions where the WRT members delegate.

app/webpacker/components/NewcomerChecks/index.jsx
danieljames-dj: Okay done.

db/migrate/20250619163228_create_potential_duplicate_persons.rb
danieljames-dj: Okay done

db/migrate/20250619163228_create_potential_duplicate_persons.rb
danieljames-dj: Okay done

app/controllers/results_submission_controller.rb
gregorbg: Yeah, I meant that you should replace it by `find_by` and see whether the scoping still works.

app/views/competitions/_nav.html.erb
gregorbg: Yeah that sounds like a good plan!

app/controllers/results_submission_controller.rb
danieljames-dj: Okay done

app/views/competitions/_nav.html.erb
danieljames-dj: Done

gregorbg (approved):

PR #11826, opened 2025-06-19
WC2025 Homepage Changes

dunkOnIT:
## MUST DO
 - [x] responsive design: desktop, tablet, mobile
- [x] Images
  - [x] Bring the cubingusa images into our repo
  - [x] Try replacing logos with Board Rubik's appreciation post
- [x] Add an env_var/server setting we can use to enable/disable the wc2025 partial

## SHOULD DO
- [x] Give WCT a default option to set the live link to the WC2025 promo video
- [ ] We could also use this page as a countdown/hype-generator? 
- [ ] Form
  - [x] Improve clarity/wording of form
  - [ ] Re-fetch the states when hitting an endpoint
- [ ] Diff review
  - [ ] Simplify/clean up CSS
- [x] try and change background image

## TESTS
- [x] Access livestream_management
  - [x] Yes:Board, WCT, WST
  - [x] No: Anyone else
- [x] Fire requests against update/confirm endpoints
  - [x] Yes:Board, WCT, WST
  - [x] No: Anyone else

## Ramblings
- The design stuff: 
  - Flexbox/raw CSS generally works, but takes a bit of hardcoding of the size of the embed in particular
  - Bootstrap works for responsive, but it doesn't play very nicely with the embed
    - Often shows up as smaller than we would like it to be
    - Disappearing completely when smaller than Large; gag me on a spoon lol
  - Bootstrap also doesn't vertically center easily: see [this discussion](https://stackoverflow.com/questions/20547819/vertical-align-with-bootstrap-3)

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -40,10 +40,10 @@ class PostsController < ApplicationController
     @latest_post = Post.order(sticky: :desc, created_at: :desc).first
     @preview = params[:preview] == "1" && current_user&.can_administrate_livestream?
     @video_id = if @preview
-      ServerSetting.find_by(name: ServerSetting::TEST_VIDEO_ID_NAME)&.value
-    else
-      ServerSetting.find_by(name: ServerSetting::LIVE_VIDEO_ID_NAME)&.value
-    end
+                  ServerSetting.find_by(name: ServerSetting::TEST_VIDEO_ID_NAME)&.value
+                else
+                  ServerSetting.find_by(name: ServerSetting::LIVE_VIDEO_ID_NAME)&.value
+                end
   end
 
   def livestream_management


dunkOnIT:
@thewca-bot deploy staging

dunkOnIT:
@thewca-bot deploy staging

dunkOnIT:
@thewca-bot deploy staging

dunkOnIT:
@thewca-bot deploy staging

dunkOnIT:
@thewca-bot deploy staging

dunkOnIT:
@thewca-bot deploy staging

dunkOnIT:
@thewca-bot deploy staging

dunkOnIT:
@thewca-bot deploy staging

dunkOnIT:
@thewca-bot deploy staging

app/controllers/posts_controller.rb
```diff
@@ -37,6 +38,32 @@ def index
 
   def homepage
     @latest_post = Post.order(sticky: :desc, created_at: :desc).first
+    @preview = params[:preview] == "1" && current_user.can_administrate_livestream?
```
gregorbg: This is very dangerous and will probably explode if a user is not logged in, because then `current_user` will be `nil` and calling `can_administrate_livestream?` will raise an exception. You should use nil-safe accessing here.

app/controllers/posts_controller.rb
```diff
@@ -37,6 +38,32 @@ def index
...
+  end
+
+  def livestream_management
+  end
+
+  def update_test_link
+    new_value = params[:new_test_value]
+    test = ServerSetting.find_or_create_by(name: "TEST_wc2025_video_url")
+
+    if test.update!(value: new_value)
```
gregorbg: This check is redundant in the sense that the `else` will never be entered: `update!` with the exclamation sign throws hard errors when something didn't work, so the code execution will abort and render the global "Something went wrong" error page.
Judging by your if-else-skeleton, you probably want the "soft" `update` method instead, which just returns `false` instead of raising the error

gregorbg:
```suggestion
    if test.update(value: new_value)
```

app/controllers/posts_controller.rb
```diff
@@ -37,6 +38,32 @@ def index
...
     @latest_post = Post.order(sticky: :desc, created_at: :desc).first
+    @preview = params[:preview] == "1" && current_user.can_administrate_livestream?
+  end
+
+  def livestream_management
+  end
+
+  def update_test_link
+    new_value = params[:new_test_value]
+    test = ServerSetting.find_or_create_by(name: "TEST_wc2025_video_url")
```
gregorbg: Can the string values like `TEST_wc2025_video_url` (and others) be stored as constants `WC2025_VIDEO_URL` somewhere? In case of doubt, putting them under `ServerSetting` directly shall be fine, just so that you avoid typos in your numerous `find` or `find_by` invocations.

app/controllers/posts_controller.rb
```diff
@@ -37,6 +38,32 @@ def index
...
+      render json: { data: test.reload.value }
+    else
+      render json: { error: test.errors }
+    end
+  end
+
+  # Sets the live link to the value of the current test link
+  def promote_test_link
+    test = ServerSetting.find("TEST_wc2025_video_url")
+    live = ServerSetting.find_or_create_by(name: "wc2025_video_url")
```
gregorbg: strings into constants, see above

app/views/posts/_wc2025.html.erb
```diff
@@ -0,0 +1,73 @@
+<% video_id = @preview ? ServerSetting.find_by(name: 'TEST_wc2025_video_url')&.value : ServerSetting.find_by(name: 'wc2025_video_url')&.value %>
```
gregorbg: Strings into constants, see above

app/webpacker/components/Posts/LivestreamManager.jsx
```diff
@@ -0,0 +1,241 @@
...
+  };
+
+  const submitPromote = () => {
+    promoteTestVideoIdMutation();
+    setConfirmPromoteOpen(false);
+  };
+
+  return (
+    <Container>
+      <h1>Livestream Management</h1>
```
gregorbg: This can use SemUI `Header` component

gregorbg (changes requested):

gregorbg (changes requested):
Now you went a little bit too far with the `ServerSettings` constants :sweat_smile: 

With the `def self.test_video_id` et al, you're losing control over database access methods (ie whether you're actually doing `find`, `find_by`, `find_or_create_by`, etc). I was referring to only the string value of the setting _name_ itself, that can be exposed throughout the code.

Are you sure that (implicitly) using `find_or_create_by` _everywhere_ is really what you want?

dunkOnIT:
`find_or_create_by` _everywhere_ definitely isn't what I wanted :P but I think using `find_or_create_by` everywhere that I'd want to access the value should be fine?

app/controllers/posts_controller.rb
dunkOnIT: Fixed

app/controllers/posts_controller.rb
dunkOnIT: Done (pending discussion)

app/controllers/posts_controller.rb
dunkOnIT: Done (pending discussion)

app/views/posts/_wc2025.html.erb
dunkOnIT: Done (pending discussion)

app/views/posts/_wc2025.html.erb
```diff
@@ -0,0 +1,71 @@
+<% return if @video_id.blank? %>
```
gregorbg: You should do this check one level up, i.e. `render 'wc2025' unless @video_id.blank?`

app/views/posts/homepage.html.erb
```diff
@@ -1,6 +1,12 @@
 <% add_to_packs("homepage") %>
 <div id="homepage">
+  <% if @preview %>
+    <h1 style="text-align: center;">================= HOMEPAGE PREVIEW ====================</h1>
+  <% end %>
+
   <div class="ui fluid container" id="cover-homepage">
+
+    <%= render 'wc2025' %>
```
gregorbg: Do the conditional check here as per my other comment

app/views/posts/_wc2025.html.erb
```diff
@@ -0,0 +1,71 @@
...
+          </div>
+        </div>
+      </div>
+
+      <div class="col-md-8 col-lg-6 column-gap">
+        <div class="embed-responsive embed-responsive-16by9">
+          <iframe
+            class="embed-responsive-item"
+            style="border-radius: 12px; overflow: hidden;"
+            src=<%= "https://www.youtube.com/embed/#{@video_id}?autoplay=1&mute=1&loop=1&playlist=#{@video_id}" %>
```
gregorbg:
```suggestion
            src="<%= "https://www.youtube.com/embed/#{@video_id}?autoplay=1&mute=1&loop=1&playlist=#{@video_id}" %>"
```

gregorbg: Context: HTML expects `src` to be a string demarcated by `"`. And the HTML source code itself cannot magically know that whatever Ruby spits out in the end "accidentally" happens to be a string

PR #11828, opened 2025-06-19
Downgrade @payloadcms/storage-s3 to 3.42

FinnIckler:
If you locally have it installed already you will need to remove the ^ so it will definitely install 3.42 to fix it (because technically 3.43 fulfills ^3.42)

### DIFF
--- a/next-frontend/yarn.lock
+++ b/next-frontend/yarn.lock
@@ -3796,7 +3796,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@payloadcms/storage-s3@npm:3.42.0":
+"@payloadcms/storage-s3@npm:^3.42.0":
   version: 3.42.0
   resolution: "@payloadcms/storage-s3@npm:3.42.0"
   dependencies:


gregorbg:
Right now, you're kinda "enforcing" people to run into errors again because this commit isn't changing the underlying version in the lockfile.

FinnIckler:
that's exactly what I did

gregorbg:
Then maybe you committed the result in the wrong order? Or forgot to push after making some more changes? Right now, my GitHub shows a diff of three lines total for this PR, which does not include downgrading the lockfile.

next-frontend/yarn.lock
```diff
@@ -12737,7 +12692,7 @@ __metadata:
     "@payloadcms/live-preview-react": "npm:^3.42.0"
     "@payloadcms/next": "npm:^3.42.0"
     "@payloadcms/richtext-lexical": "npm:^3.42.0"
-    "@payloadcms/storage-s3": "npm:^3.43.0"
+    "@payloadcms/storage-s3": "npm:3.42.0"
```
gregorbg:
```suggestion
    "@payloadcms/storage-s3": "npm:^3.42.0"
```

next-frontend/yarn.lock
```diff
@@ -3796,17 +3796,17 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@payloadcms/storage-s3@npm:^3.43.0":
-  version: 3.43.0
-  resolution: "@payloadcms/storage-s3@npm:3.43.0"
+"@payloadcms/storage-s3@npm:3.42.0":
```
gregorbg:
```suggestion
"@payloadcms/storage-s3@npm:^3.42.0":
```

gregorbg (approved):
Should be good now! (sorry for the many comments, yarn downgrades are just hard by nature...)

PR #11831, opened 2025-06-19
Next officers/board/translators page

FinnIckler:
Currently looks like this 
image

I noticed we don't have an Email Icon, should I then just use the react icon one?


next-frontend/public/wcaAPI.yaml
```diff
@@ -120,6 +180,103 @@ components:
...
+            updated_at:
+              type: string
+              format: date-time
+            email:
+              type: string
+              format: email
+        class:
+          type: string
+
+    Group:
```
danieljames-dj: Maybe `UserGroup`?

next-frontend/src/types/openapi.ts
```diff
@@ -215,6 +270,51 @@ export interface paths {
...
+                /** Format: date-time */
+                created_at?: string;
+                /** Format: date-time */
+                updated_at?: string;
+                /** Format: email */
+                email?: string;
+            };
+            class?: string;
+        };
+        Group: {
```
danieljames-dj: Same here, `UserGroup` will be better?

FinnIckler:
image

PR #11835, opened 2025-06-20
Migrate Events page to NextJS

gregorbg:
The translations are still a bit rudimentary, but it is otherwise feature-complete.

The main problem with translations is the way we (used to) inject readily built HTML **as string** into another translation string. This warrants a whole different discussion, so opening the PR in its current state and we have to fix that interpolation problem in a different PR


gregorbg:
Doing it that way will avoid having to maintain a hard-coded list of "plural keys". Which is tricky, because different languages have different plural rules. [Arabic is especially (in)famous for complex plural forms](https://www.i18next.com/translation-function/plurals#languages-with-multiple-plurals)

PR #11840, opened 2025-06-20
Next regional organizations page

FinnIckler:
Reusing the loading and errored component from my delegates and officers page. The I18n translate one is also just a copy paste from the current one.

We also don't have I18.tArray and I tried porting it, but it's a huge mess and honestly shouldn't exist haha

image

image

This is without images, as the dev export doesn't have them. I will try out with some on monday


FinnIckler:
hover:
image
normal:
image

gregorbg (approved):

PR #11881, opened 2025-06-26
Teams committees delegates page

FinnIckler:
image
Of course `orientation="vertical"` is one of the things you can't set depending on responsiveness. So the mobile version needs to work differently

This also reuses a lot of code from https://github.com/thewca/worldcubeassociation.org/pull/11831 but that one is getting a bit bloated


FinnIckler:
yeah, that does look better
image

FinnIckler:
Screenshot 2025-07-07 at 15 24 10

FinnIckler:
Screenshot 2025-07-07 at 15 24 10
Three is too much?

maxidragon:
Hmm, I think it looks kinda ugly without a break between cards. Two should be fine

gregorbg (approved):

PR #11882, opened 2025-06-26
Ask for claiming WCA ID after sign up instead of during sign up

danieljames-dj:
Currently during sign up, the user is asked to select the WCA ID and the delegate to claim WCA ID. Overall this flow is bit complex.

In this PR, I'm simplifying the sign up form to just ask "whether they need to claim WCA ID", this way the sign up page will be very simple, and after login, the user will be asked to claim WCA ID if they have that boolean set.

This PR is creating a new field `should_claim_wca_id`, which will increase the number of columns in users table. But this will allow to remove few other columns in `users` table later once the "claiming WCA ID" part is moved to tickets.


app/controllers/application_controller.rb
```diff
@@ -83,6 +83,11 @@ def doorkeeper_unauthorized_render_options(error: nil)
     session[:locale] = nil
     set_locale
 
+    if current_user.should_claim_wca_id?
+      flash[:info] = t('devise.after_sign_in.wca_claim_alert', claim_wca_id_path: profile_claim_wca_id_url).html_safe
```
gregorbg: Does this depend on https://github.com/thewca/worldcubeassociation.org/pull/11718?

app/controllers/application_controller.rb
```diff
@@ -83,6 +83,11 @@ def doorkeeper_unauthorized_render_options(error: nil)
...
+      current_user.update(should_claim_wca_id: false)
```
gregorbg: This `update` call returns a boolean which is currently ignored. You have to put this into an `if` somehow, to reasonably react to the case where the update operation in the DB failed.

gregorbg: Easiest case: Change the outer `if` to `if current_user.should_claim_wca_id? && current_user.update(...)`.
But this is potentially dangerous seeing that it does not allow you to distinguish between "`else` because `current_user.should_claim_wca_id?` is false" and "`else` because `should_claim_wca_id?` is true but the subsequent `update` failed".

spec/features/sign_up_spec.rb
gregorbg, line 1: I'm always a bit unhappy when test cases are removed without replacement. What's the plan here? Are the old test cases truly obsolete? Or do you plan to re-add them as part of a follow-up migration?

gregorbg: I think since the "claim your WCA ID" functionality still exists in principle, these test cases should _not_ become obsolete. Maybe the way how you _reach_ the behaviour that is being tested is wildly different, but the basic "idea" or "spirit" of the test cases should still exist IMO.

Feel free to change my mind if I'm overlooking something obvious.

gregorbg (changes requested):

app/controllers/application_controller.rb
danieljames-dj: For a better user-experience, it will be better if that PR is resolved. Otherwise, the user has to copy the link and paste it in a new tab.

app/controllers/application_controller.rb
danieljames-dj: Is it fine if we do no action in `else` case? If `should_claim_wca_id` is `false`, indeed we don't have anything to do because that is a normal use-case. But I'm confused on the case where the DB write fails. Is it fine to just ignore the failure (and try again during next login) or should we do any action?

Also, when the `should_claim_wca_id` is `true`, won't it be better to show the alert anyway irrespective of whether the write is successful or not?

spec/features/sign_up_spec.rb
danieljames-dj: Yes, the "claim your WCA ID" functionality still exist. But it's no longer connected to sign up page. I'll elaborate: earlier users were able to claim WCA ID using two ways:
1. During sign up
2. Using [claim WCA ID page/form](https://www.worldcubeassociation.org/profile/claim_wca_id)

In this PR, the first option is getting removed. So now users can claim WCA ID only using option 2 and that functionality (luckily) has unit tests here: https://github.com/thewca/worldcubeassociation.org/blob/main/spec/features/claim_wca_id_spec.rb.

For the remaining part, I believe I removed only those portion which are truly obsolete, and tried to have replacements wherever possible. I may not have any follow-up PRs for the sign up page, but I will have follow-up PRs for claiming WCA ID process improvements (not soon though).

app/controllers/application_controller.rb
gregorbg: I am beginning to doubt whether this method is the correct place for checking anyways. Because I just realized that this block of code was actually added to `after_sign_in_path_for`.

That method is a Devise-internal hook for rerouting. It feels like abuse to check our WCA-internal business logic and even perform data writes here. I know that it may "just work", and I am impressed that you even found this hook, but I still don't feel that its appropriate to hijack an internal authentication hook that runs for _every single user_ (out of which, >90% probably don't even need to claim their WCA ID anymore)

I also don't see why injecting this message as a flash is necessary. You could consider doing stuff like looking at our global layout files, and directly do something like
```
<% if current_user.should_claim_wca_id? %>
  <div class="alert">
    Please claim your WCA ID!
  </div>
<% end %>
```
rather easily. And if the body of the `if` becomes too complex, you can even move it out into its own partial like `_claim_wca_id_alert.html.erb`.

And if you do that, then you could even implement the partial as a `react_component` and add more info, maybe a button to say "Huh no I'm fine I already did it" and other stuff like that.

app/views/devise/registrations/new.html.erb
```diff
@@ -5,119 +5,18 @@
...
-                <%= f.input :country_iso2, collection: Country.all_sorted_by(I18n.locale, real: true), value_method: lambda { |c| c.iso2 }, label_method: lambda { |c| c.name } %>
-              </div>
-            </div>
-          </div>
-        </div>
-        <%= f.hidden_field :claiming_wca_id %>
+        <%= f.input :should_claim_wca_id %>
         <%= recaptcha_tags %>
         <br>
-        <%= f.submit t('.sign_up'), class: "btn btn-primary", disabled: true %>
```
gregorbg, line 62: This change reveals a UX weakness of your new approach: The old sign-up page had the submit button disabled by default, and *forced you* to make a conscious choice about "Have you competed before?" -- only if you click one of the tabs, did the submit button become enabled.

Your new suggestion is much less complex in terms of code, and I support that general cleanup idea, but it makes the user experience more "risky". It is much easier to "skip over" or "ignore" or maybe even just "genuinely and honestly miss" a checkbox that has no real impact on the submit button.

Please think about some way to make this distinction (Have you competed before? Yes or No?) more binding again.

spec/features/sign_up_spec.rb
gregorbg: If you don't have time for the follow-up PRs right now, then perhaps we should consider reverting this PR to draft and tackling it later as part of a "claim WCA ID" overhaul. What do you think?

gregorbg (changes requested):
Long story short: I like the idea of simplification for your PR, but I don't like introducing a whole new boolean column to the users table whose only purpose is to show one single alert (and then after showing the alert, immediately gets cleared, and becomes "useless" until the end of time and will be false forever in all user rows).

I'm starting to have some doubts whether the PR should be merged right now, in its current state. Details are in my review comments,

app/controllers/application_controller.rb
danieljames-dj: Okay having it in global layout is a good idea. I didn't try that though, but I believe that will be a better approach. I'll implement it once we have a decision on the sign up page. For now, I'll leave this comment open.

spec/features/sign_up_spec.rb
danieljames-dj: Only drawback of having too much time between this PR and the next "claim WCA ID" improvement PR is that users table will have one additional column for that much time. Did you say so for the same reason or are there any other reasons? (not arguing, just trying to understand :-))

app/views/devise/registrations/new.html.erb
danieljames-dj: Asking the user "yes" or "no" is a good idea. I tried that, but I think I'll be requiring jquery to make sure that the user click that field. Is it fine to bring back jquery for this case alone?

gregorbg: Only to the degree that was already there before your PR.

spec/features/sign_up_spec.rb
gregorbg: The main reason is that you're significantly changing a core part of our user onboarding experience, but then you say that you don't have time for follow-ups.
Claiming WCA ID is a significant feature. If we change this, let's change it properly with remodels and redesigns.

danieljames-dj: Okay, I'll mark this PR as draft, and will discuss the design with you later. I'll try my best not to delay this much. The basic design idea is to create a new ticket_type named `claim_wca_id` and have the assigned Delegate as the only stakeholder. The new ticket type model will have fields like `unconfirmed_wca_id` and `delegate_id_to_handle_wca_id_claim` from the `users` table. This PR is definitely the starting point. But anyway I'll give the design and implementation idea once I have time for continuing this.

danieljames-dj:
Attaching screenshot of the alert (in react):

Screenshot 2025-08-16 at 15 11 31

app/controllers/application_controller.rb
danieljames-dj: Changed to react.

app/views/devise/registrations/new.html.erb
danieljames-dj: Done

spec/features/sign_up_spec.rb
danieljames-dj: I've now marked this as 'ready for review'. This has all the changes that you suggested. The removed test cases are for those UI which are no longer there.

gregorbg (commented):
I still have a few high-level concerns.

1. Who "owns" the sign-in page? That is, which team or committee (or even the Board, god beware) is responsible for deciding how simple (or how complex) the sign in page should be? Who can approve that this new, redesigned flow is fine and desirable?
2. You're introducing a "sleeper boolean": Essentially it's a flag in the DB that may be `true` for some users temporarily, but in the long run, over time, it will be `false` for every user ever in the DB. This feels like a waste of space. Can we improve the overall flow, in a way that we can "properly" trigger claims?

The issue around (2) makes me think a lot about the claim process overall. Is the sign-up process really so horribly complex? Do we know of users who wanted to create an account but gave up because they thought "god this is complicated"?

In the previous (current production) version, we got all the information that we need for the claim during sign-in. With this PR, the sign-up process gets (arguably) easier, but we sacrifice receiving all information early, and we "allow" users to be sloppy and just ignore the popup message. I am not fully convinced yet that this is a desirable. good sign-up flow.

danieljames-dj:
Discussed with Gregor offline and he helped in figuring out solution for creating ticket in the current rails UI. But before implementing that, I'll be needing 'bcc stakeholders' feature: https://github.com/thewca/worldcubeassociation.org/pull/13152. I'll mark this as draft for now, will continue once the other PR is closed.

Issue #11728, opened 2025-06-08
Let posts on the WCA website remain stickied for 2 weeks


lnzainn:
Before the migration, this page https://www.worldcubeassociation.org/posts/new by default stickied posts for 2 weeks if you simply checked the box 'Sticky' at the bottom given that you did not manually change the unstick date. 

I believe this might be a simple feature request - if you check the 'sticky' box, then it should set an end date of 2 weeks automatically rather than it being permanent.

PR #11884, opened 2025-06-29
Let posts remain stickied for 2 weeks by default

pranav-027:
fixes: https://github.com/thewca/worldcubeassociation.org/issues/11728

https://github.com/user-attachments/assets/a14eb295-436e-437e-871f-1577b3c755f8

### DIFF
--- a/app/webpacker/components/Posts/PostForm.jsx
+++ b/app/webpacker/components/Posts/PostForm.jsx
@@ -24,12 +24,7 @@ export default function PostForm({
   ] = useCheckboxState(post?.show_on_homepage ?? true);
   const [postURL, setPostURL] = useInputState(post?.url ?? null);
   const [postId, setPostId] = useInputState(post?.id ?? null);
-  const [unstickAt, setUnstickAt] = useState(() => {
-    if (post?.unstick_at) return post.unstick_at;
-    const now = new Date();
-    now.setDate(now.getDate() + 14);
-    return now.toISOString().split('T')[0];
-  });
+  const [unstickAt, setUnstickAt] = useState(post?.unstick_at ?? null);
 
   const tagOptions = useMemo(
     () => allTags.map((tag) => ({ value: tag, text: tag, key: tag })),
--- a/app/webpacker/components/Posts/PostForm.jsx
+++ b/app/webpacker/components/Posts/PostForm.jsx
@@ -24,7 +24,12 @@ export default function PostForm({
   ] = useCheckboxState(post?.show_on_homepage ?? true);
   const [postURL, setPostURL] = useInputState(post?.url ?? null);
   const [postId, setPostId] = useInputState(post?.id ?? null);
-  const [unstickAt, setUnstickAt] = useState(post?.unstick_at ?? null);
+  const [unstickAt, setUnstickAt] = useState(() => {
+    if (post?.unstick_at) return post.unstick_at;
+    const now = new Date();
+    now.setDate(now.getDate() + 14);
+    return now.toISOString().split('T')[0];
+  });
 
   const tagOptions = useMemo(
     () => allTags.map((tag) => ({ value: tag, text: tag, key: tag })),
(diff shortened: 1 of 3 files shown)


app/webpacker/components/Posts/PostForm.jsx
```diff
@@ -24,7 +24,12 @@ export default function PostForm({
   ] = useCheckboxState(post?.show_on_homepage ?? true);
   const [postURL, setPostURL] = useInputState(post?.url ?? null);
   const [postId, setPostId] = useInputState(post?.id ?? null);
-  const [unstickAt, setUnstickAt] = useState(post?.unstick_at ?? null);
+  const [unstickAt, setUnstickAt] = useState(() => {
+    if (post?.unstick_at) return post.unstick_at;
+    const now = new Date();
+    now.setDate(now.getDate() + 14);
+    return now.toISOString().split('T')[0];
```
gregorbg: We usually like to communicate dates in standard ISO format. Is there a particular reason why you chose to deviate from the ISO format?

app/webpacker/components/Posts/PostForm.jsx
```diff
@@ -24,7 +24,12 @@ export default function PostForm({
   ] = useCheckboxState(post?.show_on_homepage ?? true);
   const [postURL, setPostURL] = useInputState(post?.url ?? null);
   const [postId, setPostId] = useInputState(post?.id ?? null);
-  const [unstickAt, setUnstickAt] = useState(post?.unstick_at ?? null);
+  const [unstickAt, setUnstickAt] = useState(() => {
+    if (post?.unstick_at) return post.unstick_at;
+    const now = new Date();
+    now.setDate(now.getDate() + 14);
```
gregorbg: For "complex" date-time operations, we also have the `luxon` JS library installed. It has a more fluent interface that can do stuff like `DateTime.now().plus({ days: 14 })`

gregorbg (changes requested):
Some specific comments are directly in the code diff. On a more general note, you could achieve the same in the backend by using the `attribute` notation on the model class `Post` itself, which arguably feels a bit cleaner (the source of truth for the data is closer to the model where that data is being stored).
But it would also require passing a serialized version of the mostly empty post to the frontend component, so that's a large bit of mostly-redundant code. I have a slight preference for the cleaner backend solution, but I am willing to accept your frontend solution if the other version is too much work / too verbose to implement.

Reference: https://guides.rubyonrails.org/active_model_basics.html#attributes

gregorbg:
Another plausible solution is hidden in the description of the Issue file here on GitHub: You could replace the current `onChange={setFormIsStickied}` with a manually created method that "wraps" the existing `setFormIsStickied` _and_ also writes a default value to the sticky date, if not already present.

pranav-027:
@gregorbg Implemented solution on backend. 

I wanted to apply the fix only on the backend (in the controller), but decided not to touch code that hasn’t changed. Implementing it in the model makes much more sense. 
Also, I wasn’t able to do it with an attribute, since unstick_at is set conditionally (only when 'sticky' is true).

spec/models/post_spec.rb
```diff
@@ -48,4 +48,12 @@
       expect(Post.find(post.id).tags_array).to match_array %w[wic]
     end
   end
+
+  context "unstick_at behavior" do
+    it "defaults unstick_at to 2 weeks from now when sticky is true and unstick_at is nil" do
+      post = build(:sticky_post, unstick_at: nil)
+      post.valid?
```
gregorbg:
```suggestion
      expect(post).to be_valid
```

gregorbg: I can see that you were only calling this line to trigger validations.
An alternative solution would be to use `create` instead of `build` above, which inserts the mocked record into the (testing) DB. And DB inserts always inherently trigger validations.

gregorbg (changes requested):

gregorbg (approved):

pranav-027:
My 50th commit to this repo 🥳

PR #11922, opened 2025-07-01
Allow admins to enable live auto-accept in production

dunkOnIT:


app/controllers/competitions_controller.rb
```diff
@@ -486,6 +486,16 @@ def unbookmark
     head :ok
   end
 
+  before_action :check_live_auto_accept_permissions, only: %i[create update]
+  private def check_live_auto_accept_permissions
+    currently_enabled = params['action'] == 'create' ? false : competition_from_params.auto_accept_preference_live?
```
dunkOnIT: fix this lol

gregorbg: Nope.

app/webpacker/components/CompetitionForm/FormSections/CompetitorLimit.js
```diff
@@ -9,10 +9,16 @@ export default function CompetitorLimit() {
   const {
     competitorLimit: {
       enabled: hasLimit,
-      autoAcceptEnabled,
+      autoAcceptPreference,
     },
   } = useFormObject();
 
+  const autoAcceptOptions = ['disabled', 'bulk', 'live'].map((status) => ({
```
gregorbg: Pre-filter based on whether the user is admin or not

app/controllers/competitions_controller.rb
```diff
@@ -486,6 +486,16 @@ def unbookmark
     head :ok
   end
 
+  before_action :check_live_auto_accept_permissions, only: %i[create update]
```
dunkOnIT: remove all of this and use the CompetitorLimit form instead

app/webpacker/components/CompetitionForm/FormSections/CompetitorLimit.js
dunkOnIT: filter the options available to the delegate based on whether or not they're an admin

dunkOnIT: also put the values for this in wca-data

gregorbg: Unrelated: Export the (entire!) enum from Rails through `wca-data`, and consume via `Object.keys(...yourExportedThingy...)` here

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationList.jsx
```diff
@@ -423,7 +423,7 @@ export default function RegistrationAdministrationList({ competitionInfo }) {
 
   return (
     <Segment loading={isMutating || isAutoAccepting}>
-      {competitionInfo.auto_accept_registrations && (
+      {competitionInfo.auto_accept_preference === 'bulk' && (
```
gregorbg:
```suggestion
      {competitionInfo.auto_accept_preference === exportedThingy.bulk && (
```

config/locales/en.yml
```diff
@@ -1799,9 +1799,10 @@ en:
           enabled:
             "true": "I want to specify a competitor limit"
             "false": "I DO NOT want to specify a competitor limit"
-          auto_accept_enabled:
-            "true": "Enable 'Bulk Auto Accept' button in Registrations menu"
-            "false": "Disable 'Bulk Auto Accept' button"
+          auto_accept_preference:
+            "disabled": "Disable 'Bulk Auto Accept' button"
+            "bulk": "Enable 'Bulk Auto Accept' button in Registrations menu"
+            "live": "Admin only"
```
gregorbg: ???

gregorbg: Please make it more descriptive. Even if only visible for admins in the first place, those admins should still know how many washing machines they're buying by choosing this option

spec/factories/registrations.rb
```diff
@@ -113,6 +113,7 @@
     end
 
     trait :unpaid do
+      # TODO: What the fuck is going on here? This very much looks like it creates a payment?
```
gregorbg: :upside_down_face:

spec/factories/competitions.rb
```diff
@@ -108,9 +108,18 @@
...
+      competitor_limit { 5 }
+      auto_accept_disable_threshold { 4 }
+    end
+
+    trait :live_auto_accept do
+      use_wca_registration { true }
+      auto_accept_preference { :live }
       competitor_limit_enabled { true }
       competitor_limit_reason { 'test' }
       competitor_limit { 5 }
```
gregorbg: Refactor commonalities into actual common trait

app/models/registration.rb
```diff
@@ -593,10 +590,10 @@ def last_positive_payment
       .first
   end
 
-  delegate :auto_accept_registrations?, to: :competition
+  delegate :auto_accept_preference, :auto_accept_preference_disabled?, :auto_accept_preference_bulk?, :auto_accept_preference_live?, to: :competition
 
-  def attempt_auto_accept
-    failure_reason = auto_accept_failure_reason
+  def attempt_auto_accept(bulk: false)
```
gregorbg: Pass in `live: false` instead of `bulk: false` because that's the potentially more harmful operation

app/models/registration.rb
```diff
@@ -618,9 +615,9 @@ def attempt_auto_accept
     end
   end
 
-  private def auto_accept_failure_reason
+  private def auto_accept_failure_reason(bulk)
     return Registrations::ErrorCodes::OUTSTANDING_FEES if outstanding_entry_fees.positive?
-    return Registrations::ErrorCodes::AUTO_ACCEPT_NOT_ENABLED unless auto_accept_registrations?
+    return Registrations::ErrorCodes::AUTO_ACCEPT_NOT_ENABLED if (bulk && auto_accept_preference_live?) || auto_accept_preference_disabled?
```
gregorbg: This creates odd edge cases for competitions with `bulk`, where a live accept triggers `bulk = false` and then this current check would let it pass

db/migrate/20250620074209_add_auto_accept_preference_to_competitions.rb
```diff
@@ -0,0 +1,19 @@
...
+    add_column :competitions, :auto_accept_preference, :integer, default: 0, null: false
+
+    reversible do |direction|
+      direction.up do
+        execute <<~SQL.squish
+          UPDATE competitions
+          SET auto_accept_preference = CASE
+            WHEN auto_accept_registrations = TRUE THEN 1
+            ELSE 0
+          END
```
gregorbg:
```suggestion
          UPDATE competitions
SET
  auto_accept_preference = auto_accept_registrations;
```

gregorbg: Along with a comment that this works because the order of the new enum values just so happens to match the integer representations `0` and `1` for the old boolean column

db/migrate/20250620074209_add_auto_accept_preference_to_competitions.rb
```diff
@@ -0,0 +1,16 @@
...
+
+class AddAutoAcceptPreferenceToCompetitions < ActiveRecord::Migration[7.2]
+  def change
+    add_column :competitions, :auto_accept_preference, :integer, default: 0, null: false
+
+    reversible do |direction|
+      direction.up do
+        execute <<~SQL.squish
+          UPDATE competitions
+          SET auto_accept_preference = auto_accept_registrations
```
gregorbg, line 14: Brief comment why this works

db/schema.rb
```diff
@@ -1188,7 +1190,7 @@
   end
 
   create_table "schedule_activities", charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
-    t.bigint "venue_room_id", null: false
```
gregorbg: Unrelated change

db/schema.rb
```diff
@@ -349,6 +349,7 @@
   create_table "competition_payment_integrations", charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
     t.string "connected_account_type", null: false
     t.bigint "connected_account_id", null: false
+    t.boolean "is_inactive", default: false, null: false
```
gregorbg: Unrelated change

spec/factories/competitions.rb
gregorbg: Kindly ping

app/models/registration.rb
```diff
@@ -593,10 +590,10 @@ def last_positive_payment
       .first
   end
 
-  delegate :auto_accept_registrations?, to: :competition
+  delegate :auto_accept_preference, :auto_accept_preference_disabled?, :auto_accept_preference_bulk?, :auto_accept_preference_live?, to: :competition
 
-  def attempt_auto_accept
-    failure_reason = auto_accept_failure_reason
+  def attempt_auto_accept(expected_setting)
```
gregorbg: Still not quite happy with the name. At the moment of triggering this attempt, it is the actual real mode. Nothing expected, but the actual real thing.
What you're _expecting_ is for the real mode and the competition setting to match. But the parameter here does not / should not represent that expectation. The parameter should be "what's really happening right now", and you're formulating an expectation based on this in the `failure_reason` method down below.

gregorbg (changes requested):

app/models/registration.rb
dunkOnIT: I don't quite follow the thinking here - in my mind, the source of truth is what the competition setting currently is. The caller function is submitting the value that it _thinks_ the competition setting should be - I do agree that the naming could be better for the argument, though. I'll give it some thought.

gregorbg: The competition setting is certainly one source of truth.

The mode that the current auto accept attempt runs under is another source of truth.

The _expectation_ is that these two truths / facts match up. You're not passing this expectation to the attempt. You're passing a truth to the attempt (this auto accept run is run as bulk / run as live) and later down when you're comparing that truth to the competition setting truth, there's an expectation embedded into that comparison.

dunkOnIT: Renamed to `required_setting` - so the calling function is asserting what setting is needed for its call to succeed

db/migrate/20250620074209_add_auto_accept_preference_to_competitions.rb
dunkOnIT: Added

db/schema.rb
dunkOnIT: Removed

spec/factories/registrations.rb
dunkOnIT: Removed, will investigate separately

db/schema.rb
gregorbg: Still here

spec/factories/registration_payment.rb
```diff
@@ -5,6 +5,7 @@
     transient do
       skip_auto_accept_hook { false }
       competition { registration.competition }
+      revert_to { nil }
```
gregorbg: What is `revert_to`? Did you create this or is it some RSpec shenanigan?

dunkOnIT: I created it - it just holds the auto_accept mode to revert back to in the skip_create_hook logic

gregorbg: Then perhaps `revert_mode_to`?

app/models/competition.rb
```diff
@@ -2960,6 +2963,13 @@ def self.delegate_edits_json_schema
             "mainEventId" => { "type" => %w[string null] },
           },
         },
+        "competitorLimit" => {
+          "type" => "object",
+          "properties" => {
+            "autoAcceptPreference" => { "type" => "string", "enum" => Competition.auto_accept_preferences.keys },
```
gregorbg, line 2975: In its current form, this will allow Delegates to change any auto-accept setting at will. Is this intended?
(And should we keep WCAT in the loop about it?)

db/schema.rb
gregorbg: Stiiiill here :P

app/models/competition.rb
dunkOnIT: I think we can confirm it with WCAT before we launch, but yes this is intended - delegates should be able to opt in/out of auto accept as desired

db/schema.rb
gregorbg: And yet, still here!

db/schema.rb
```diff
@@ -1,4 +1,4 @@
-# This file is auto-generated from the current state of the database. Instead
+# This file is auto-generated from the current state of the database. Insteadschema
```
gregorbg: Plus, you now introduced one bonus whoopsie!

spec/factories/registration_payment.rb
```diff
@@ -20,16 +21,22 @@
...
     end
 
     after(:build) do |_payment, evaluator|
-      evaluator.competition.auto_accept_registrations = false if evaluator.skip_auto_accept_hook
+      evaluator.competition.auto_accept_preference = :disabled if evaluator.skip_auto_accept_hook
     end
 
     after(:create) do |_payment, evaluator|
-      evaluator.competition.auto_accept_registrations = true if evaluator.skip_auto_accept_hook
+      evaluator.competition.auto_accept_preference = evaluator.revert_mode_to if evaluator.skip_auto_accept_hook
```
gregorbg: Just curious: Does `evaluator.competition.auto_accept_preference_previously_was` work?

gregorbg: If yes, that would make the whole `revert_mode_to` thingy obsolete

gregorbg (approved):
Fine with the code now, just a non-critical question out of sheer interest

PR #11963, opened 2025-07-07
Remove `confirm_information` from results_submission model

danieljames-dj:
`confirm_information` is not needed anymore as the validation is now in the react frontend.


danieljames-dj:
Got it, thanks. Will make sure not to remove from next time onwards.

PR #11969, opened 2025-07-09
Migrate Delegate's upload JSON view to React

danieljames-dj:
By this, the 'upload JSON' part will be completely migrated to react.


app/models/user.rb
```diff
@@ -926,7 +926,7 @@ def can_update_competition_series?(competition)
   end
 
   def can_upload_competition_results?(competition)
-    can_submit_competition_results?(competition, upload_only: true)
+    can_admin_results? || can_submit_competition_results?(competition, upload_only: true)
```
gregorbg: Why is this change becoming necessary now?

gregorbg: In particular, I don't see any other occurence of `can_upload_competition_results?` in your whole PR

app/webpacker/components/CompetitionResultSubmission/UploadResultsJson.jsx
```diff
@@ -10,7 +10,12 @@ export default function UploadResultsJson({ competitionId, isWrtViewing }) {
   const [markResultSubmitted, setMarkResultSubmitted] = useCheckboxState(isWrtViewing);
 
   const { mutate: uploadResultsJsonMutate, error, isError } = useMutation({
-    mutationFn: () => uploadResultsJson({ competitionId, resultFile, markResultSubmitted }),
+    mutationFn: () => uploadResultsJson({
+      competitionId,
+      resultFile,
+      markResultSubmitted,
+      storeUploadedJson: !isWrtViewing, // The JSON will be uploaded to database only for Delegates.
```
gregorbg: Will this lead to weird edge-cases when the regular, listed Delegate of a competition also happens to "accidentally" be a WRT member?

app/models/user.rb
danieljames-dj: The `upload_json` method already has a validation that it will work only if `can_upload_competition_results?` is `true`. Earlier, this check was only for Delegates and because of that there were additional checks like whether results are already posted, whether competition is in_progress/over, whether competition is announced, etc. But these are not applicable when WRT is uploading. They should have more power and should be able to do it anytime.

app/webpacker/components/CompetitionResultSubmission/UploadResultsJson.jsx
danieljames-dj: No, the boolean is controlled from the page. If the page is 'submit results' of Delegates, then even for WRT member it will be `false`. It will be `true` only when the view is "admin page".

gregorbg: Then maybe you should rename the variable to make that more clear. Something along the lines of `isWrtAdminView` (or perhaps even just `isAdminView`)

app/models/user.rb
gregorbg: Can you please point me to the exact lines of code where the pre-existing checks that you mention are defined?
I am very wary about updating existing permission methods and want to make 101% sure this doesn't break anything

danieljames-dj: https://github.com/thewca/worldcubeassociation.org/blob/8d5ac7d7397dd3624b81fdf66e74b11f0bdd80c8/app/controllers/results_submission_controller.rb#L7

app/webpacker/components/CompetitionResultSubmission/UploadResultsJson.jsx
danieljames-dj: Done

app/models/user.rb
gregorbg: In that case I think you should adjust `can_submit_competition_results?` instead.
The method `can_upload_competition_results?` only serves as a "convenience alias" to `can_submit` with the optional `upload_only` parameter set to false. All of the core logic happens in `can_submit`, including an already existing check of `can_admin_results?` at some point.

danieljames-dj: Okay added in `can_submit`. I didn't change that method first because the method was already complicated, and now it's more complicated I guess. 😅

gregorbg: "complicated" is a relative term. More lines of code, yes. But also more _meaningful_ lines of code by far.

I think the naming of variables as well as the final return value in `can_submit` do a very good job at explaining what's going on, and they are fairly readable. The fact that the code now gets even more bloated is not your fault, it's the fault of the real-world WCA logic which just happens to be complicated.

app/models/user.rb
```diff
@@ -935,9 +935,8 @@ def can_submit_competition_results?(competition, upload_only: false)
                        else
                          competition.staff_delegates.include?(self)
                        end
-    appropriate_role = can_admin_results? || allowed_delegate
     appropriate_time = competition.in_progress? || competition.probably_over?
-    competition.announced? && appropriate_role && appropriate_time && !competition.results_posted?
+    can_admin_results? || (competition.announced? && allowed_delegate && appropriate_time && !competition.results_posted?)
```
gregorbg: I want to discuss your implementation, because currently I don't agree with the implications.

You make it so that admins can submit results **any** time, which in my opinion doesn't make sense. For example, even WRT members should not be allowed to submit results before the competition started. And they also should not be allowed to submit results, if the competition hasn't been announced yet.

So your current `can_admin_results` acting as an "escape hatch" is too much in my opinion. Please go back to the drawing board and think more calmly about which exact checks you should skip, and under what circumstances WRT members should be allowed to "skip" other checks

danieljames-dj: Okay, I've modified both upload & submit methods. Is this structure fine for you? If yes, I would like to write some unit tests as well for a double confirmation as these two methods are bit critical (at least now).

danieljames-dj: Also, I made the `can_submit_competition_results?` as simple as `can_admin_results? || competition.staff_delegates.include?(self)`. This is because this method is used only in one place and that is to show the form to submit the results to WRT. The page that contains the form already checks if it's an appropriate time to upload results. So it's already covered.

app/models/user.rb
```diff
@@ -926,18 +926,13 @@ def can_update_competition_series?(competition)
...
-    allowed_delegate = if upload_only
-                         competition.delegates.include?(self)
-                       else
-                         competition.staff_delegates.include?(self)
-                       end
-    appropriate_role = can_admin_results? || allowed_delegate
-    appropriate_time = competition.in_progress? || competition.probably_over?
-    competition.announced? && appropriate_role && appropriate_time && !competition.results_posted?
+  def can_submit_competition_results?(competition)
+    can_admin_results? || competition.staff_delegates.include?(self)
```
gregorbg: You're making quite a few substantial changes here. For example, you're not checking anymore whether the competition is over or not, whether it is announced or not, etc. Can you please justify these changes?

gregorbg: My main worry is that you may give `staff_delegates` of a competition access to post results before the competition even started.

app/controllers/results_submission_controller.rb
```diff
@@ -13,25 +13,32 @@ def new
...
-      @competition.uploaded_jsons.create(json_str: @upload_json.results_json_str)
-      redirect_to competition_submit_results_edit_path
-    else
-      @results_validator = ResultsValidators::CompetitionsResultsValidator.create_full_validation
-      @results_validator.validate(@competition.id)
-      render :new
-    end
+    return render status: :unprocessable_entity, json: { error: upload_json.errors.full_messages } unless upload_json.import_to_inbox
+
+    competition.touch(:results_submitted_at) if competition.results_submitted_at.nil? && mark_result_submitted
```
gregorbg: What about the `results_submitted_by` field, by the way?

gregorbg: There is also `def results_posted?` in `competition.rb`, which checks both `submitted_at` AND `submitted_by`.

app/views/results_submission/new.html.erb
```diff
@@ -14,7 +14,11 @@
         <li>Then the website will run a number of checks. If everything is correct you'll be able to submit these results to the WRT.</li>
       </ul>
     </p>
-    <%= render "submit_json_panel", upload_json: @upload_json, already_has_submitted_result: @results_validator.any_results? %>
+    <%= react_component("CompetitionResultSubmission/ImportResultsData", {
+      competitionId: @competition.id,
+      alreadyHasSubmittedResult: @results_validator.any_results?,
+    }) %>
+    <br /> <%# The <br /> tag is used here to add some spacing for better readability. %>
```
gregorbg: That's what `br` tags usually do, not sure if you really need to explain that in a comment

app/models/user.rb
gregorbg: Also, consider this:
Before your refactor, the "philosophy" behind these two methods was that both `can_upload` and `can_submit` were essentially the same check, except that Trainee Delegates should be allowed to "upload only", without actually posting.
Every other piece of logic was shared. The two separate methods existed **only** so that you can make a distinction between "upload but not quite yet post to WRT" for Trainee Delegates specifically.
Your refactor seems to "reinvent the wheel": Now, both methods perform substantially different checks. That doesn't seem to match the spirit of the previous implementation.

app/controllers/results_submission_controller.rb
danieljames-dj: We don't track `results_submitted_by` IIRC, or am I missing something?

For second comment I assume you meant `results_submitted?`, I've changed to that, thanks.

app/views/results_submission/new.html.erb
danieljames-dj: Okay 😬  removed.

app/models/user.rb
danieljames-dj: > you're not checking anymore whether the competition is over or not, whether it is announced or not, etc. Can you please justify these changes?

It's because all those are checked in `can_upload_competition_results?` and the UI which is using `can_submit` is showing the page only if `can_upload`.

But I agree it's fine to do a double check here as well, based on your other comments as well in this thread. So I've changed it bit more. Can you please check if this is fine now?

app/controllers/results_submission_controller.rb
gregorbg: No, I definitely meant `def results_posted?`. See here, copied from the `competition.rb` source code:

```ruby
def results_posted?
  results_posted_at.present? && results_posted_by.present?
end
```

What do you mean by "We don't track `results_submitted_by`? We do have this column in the database, and the code can access it, so I'm not sure what you mean by "we don't track".

app/models/user.rb
gregorbg: Yes, looking much clearer now. I like it!

app/controllers/results_submission_controller.rb
danieljames-dj: Discussed over chat

gregorbg: (nevermind, I obviously confused `submitted` and `posted`)

PR #11971, opened 2025-07-09
Fix newcomer page permission issue

danieljames-dj:
Apologies for another fix PR, my test competition was apparently past competition so didn't figure out another permission issue which was hidden within results_submission_controller

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -940,6 +940,10 @@ class User < ApplicationRecord
     competition.announced? && appropriate_role && appropriate_time && !competition.results_posted?
   end
 
+  def can_check_newcomers_data?(competition)
+    competition.upcoming? && can_admin_results?
+  end
+
   def can_create_poll?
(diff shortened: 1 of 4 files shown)


gregorbg:
If this was not caught by existing tests, perhaps you should add a new test case? :upside_down_face:

gregorbg (approved):

app/models/user.rb
```diff
@@ -940,6 +940,10 @@ def can_submit_competition_results?(competition, upload_only: false)
     competition.announced? && appropriate_role && appropriate_time && !competition.results_posted?
   end
 
+  def can_check_newcomers_data?(competition)
+    competition.upcoming? && can_admin_results?
```
pranav-027, line 944: @danieljames-dj This should include delegates also right ?

danieljames-dj: Not now, will be adding them later after some testing within WRT. I will be posting the details in our channel soon. :-)

PR #11973, opened 2025-07-09
Migrate "import WCA Live results" to react

danieljames-dj:
This will bring the "Import WCA Live Results" even in admin results view, which will allow WRT to perform the same action. Also made sure this is not visible in production.


app/controllers/results_submission_controller.rb
```diff
@@ -127,15 +133,9 @@ def import_from_live
                 end
     end
 
-    if errors.any?
-      flash[:danger] = errors.join("<br/>")
-      @results_validator = ResultsValidators::CompetitionsResultsValidator.create_full_validation
-      @results_validator.validate(@competition.id)
-      return render :new
-    end
+    return render status: :unprocessable_entity, json: { error: errors.join("<br/>") } if errors.any?
```
gregorbg: Why do you need to join the errors with `<br>` tags here already? This was necessary in the old world because it was a `flash` message (which cannot be dynamic).

Now you're moving this to the new React world, where you can just return the individual error messages in a JSON array, and then have React do whatever it wants with this list (for example, loop over them in a `<List>`, or whatver...)

app/webpacker/components/CompetitionResultSubmission/ImportResultsData/ImportWcaLiveResults.jsx
```diff
@@ -0,0 +1,22 @@
...
+  } = useMutation({
+    mutationFn: () => importWcaLiveResults({ competitionId }),
+    onSuccess: onImportSuccess,
+  });
+
+  if (isPending) return <Loading />;
+  if (isError) return <Errored error={error} />;
+
+  return (
+    <Button onClick={importWcaLiveResultsMutate}>Import WCA Live Results</Button>
```
gregorbg: Maybe this button should be `primary`? Otherwise it will be the default SemUI gray which is easy to miss

app/webpacker/lib/wca-data.js.erb
```diff
@@ -245,3 +245,6 @@ export const cronjobs = <%= JobUtils::WCA_CRONJOBS.map(&:to_s).to_h { |item| [it
 
 // ----- RESULTS -----
 export const duplicateCheckerJobRunStatuses = <%= DuplicateCheckerJobRun.run_statuses.to_json %>
+
+// ----- PRODUCTION CONFIG -----
+export const isProduction = <%= Rails.env.production? && EnvConfig.WCA_LIVE_SITE? %>
```
gregorbg: I'm afraid I have to veto exporting `Rails.env` specific stuff, because there won't be any equivalent of exporting this value to NextJS.

Can you think of another way, for example by passing some boolean flag (that determines whether the button should be shown or not) to the props of the component via `react_component` view helper?

gregorbg (changes requested):

app/webpacker/lib/wca-data.js.erb
danieljames-dj: Okay how about making it viewable only in adminView?

app/webpacker/components/CompetitionResultSubmission/ImportResultsData/ImportWcaLiveResults.jsx
danieljames-dj: Done

app/controllers/results_submission_controller.rb
danieljames-dj: Will attend this soon.

app/webpacker/lib/wca-data.js.erb
gregorbg: For now, that makes sense. In the future, we obviously also want to launch it for the public.

app/controllers/results_submission_controller.rb
danieljames-dj: Okay, I am now passing the errors as it is. I'm not doing anything else for now. But all the errors are being displayed separated by comma. Since it's thrown as error, I'm getting error as a single string because of [this](https://github.com/thewca/worldcubeassociation.org/blob/0a4ac3dac63b7b20d4ba403365313890c46261ba/app/webpacker/lib/requests/fetchWithAuthenticityToken.js#L20).

I can actually get this as list if I send 200 with errors in JSON, but I think it won't be necessary as this anyways serve the purpose. I'll convert it that way in future if needed.

app/webpacker/lib/wca-data.js.erb
danieljames-dj: Yes, that will be possible for sure.

app/controllers/results_submission_controller.rb
gregorbg: Cool!

PR #11974, opened 2025-07-09
add regionals to download voters section

Brandon-Reichman:
@thewca/wic required a list of regional delegate contacts and I figured this could just get added to the "Download Voters" section.

---

It seems worth noting that the "WCA leaders and seniors" button will throw an error for me locally saying
`undefined method 'active?' for nil` on

```
team_leaders = RolesMetadataTeamsCommittees.leader.includes(:user, :user_role).select { |role_metadata| role_metadata.user_role.active? }.map(&:user)
```
^ This is not something I touched in this PR.

However, this doesn't seem to be an issue in production. I got the same error, locally, when trying to select the regionals in a similar way, so ended up having to query them in a different way to get it to work.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -95,7 +95,7 @@ class User < ApplicationRecord
   end
 
   def self.regional_voters
-    regional_delegates = RolesMetadataDelegateRegions.regional_delegate.where(user_roles: { end_date: nil }).includes(:user).map(&:user).uniq.compact
+    RolesMetadataDelegateRegions.regional_delegate.joins(:user_role).merge(UserRole.active).includes(:user).map(&:user).uniq.compact
   end
 
   def self.all_discourse_groups


app/models/user.rb
```diff
@@ -94,6 +94,16 @@ def self.leader_senior_voters
     (team_leaders + senior_delegates).uniq.compact
   end
 
+  def self.regional_voters
+    regional_delegates = RolesMetadataDelegateRegions
+                         .where(status: "regional_delegate")
+                         .joins(:user_role)
+                         .where(user_roles: { end_date: nil })
```
gregorbg:
```suggestion
                         .merge(UserRole.active)
```

gregorbg: This is a Rails feature to re-use already defined scopes

app/models/user.rb
```diff
@@ -94,6 +94,16 @@ def self.leader_senior_voters
     (team_leaders + senior_delegates).uniq.compact
   end
 
+  def self.regional_voters
+    regional_delegates = RolesMetadataDelegateRegions
+                         .where(status: "regional_delegate")
```
gregorbg:
```suggestion
                         .regional_delegate
```

gregorbg: Works because the value `regional_delegate` is defined as part of a Rails `enum` on `status`

app/models/user.rb
```diff
@@ -94,6 +94,16 @@ def self.leader_senior_voters
...
+                         .joins(:user_role)
+                         .where(user_roles: { end_date: nil })
+                         .includes(:user)
+                         .map(&:user)
+    regional_delegates.uniq.compact
```
gregorbg: I think you can directly chain this onto the previous query builder:
`.includes(:user).map(&:user).uniq.compact`

app/models/user.rb
```diff
@@ -94,6 +94,10 @@ def self.leader_senior_voters
     (team_leaders + senior_delegates).uniq.compact
   end
 
+  def self.regional_voters
+    regional_delegates = RolesMetadataDelegateRegions.regional_delegate.where(user_roles: { end_date: nil }).includes(:user).map(&:user).uniq.compact
```
gregorbg: It seems that you did not address my previous comment about re-using the scope with `.joins(:user_roles).merge(UserRole.active)`. Did you try it, and did it work?

Brandon-Reichman: Sorry, I seemed to have totally missed that comment. This should now be addressed and I've tested the latest version locally, seems to work correctly.

PR #11978, opened 2025-07-10
Few cleanups in tickets

danieljames-dj:
All are just cleanups - basically using react-query instead of old saveAction and useLoadedData

### DIFF
--- a/app/webpacker/components/Tickets/api/updateStatus.js
+++ b/app/webpacker/components/Tickets/api/updateStatus.js
@@ -1,7 +1,7 @@
 import { fetchJsonOrError } from '../../../lib/requests/fetchWithAuthenticityToken';
 import { actionUrls } from '../../../lib/requests/routes.js.erb';
 
-export default async function updateStatus({ ticketId, status, currentStakeholder }) {
+export default async function updateStatus({ ticketId, status, currentStakeholderId }) {
   await fetchJsonOrError(
     actionUrls.tickets.updateStatus(ticketId),
     {
@@ -11,7 +11,7 @@ export default async function updateStatus({ ticketId, status, currentStakeholde
       },
       body: JSON.stringify({
         ticket_status: status,
-        acting_stakeholder_id: currentStakeholder.id,
+        acting_stakeholder_id: currentStakeholderId,
       }),
     },
   );
--- a/app/webpacker/components/Tickets/index.jsx
+++ b/app/webpacker/components/Tickets/index.jsx
@@ -48,7 +48,11 @@ function Tickets({ id }) {
     isError: isErrorUpdateStatus,
     error: errorUpdateStatus,
   } = useMutation({
-    mutationFn: (status) => updateStatus({ ticketId: id, status, currentStakeholder }),
+    mutationFn: (status) => updateStatus({
+      ticketId: id,
+      status,
+      currentStakeholderId: currentStakeholder.id,
+    }),
     onSuccess: (status) => {
       queryClient.setQueryData(
         ['ticket-details', id],


app/webpacker/components/Tickets/SkateholderSelector.jsx
```diff
@@ -0,0 +1,30 @@
...
+      <p>
+        You are part of more than one stakeholder, please select the stakeholder as which you want
+        to visit the ticket page.
+      </p>
+      <Dropdown
+        options={stakeholderList.map((requesterStakeholder) => ({
+          key: requesterStakeholder.id,
+          text: `${requesterStakeholder.stakeholder.name} (${requesterStakeholder.stakeholder_role})`,
+          value: requesterStakeholder,
+        }))}
```
gregorbg: Since they are derived from the `stakeholderList` prop, it's probably safe to `memo` these. Computing them inline will only induce unnecessary code execution during rerender.

app/webpacker/components/Tickets/index.jsx
```diff
@@ -93,8 +42,33 @@ function Tickets({ id }) {
...
+    mutate: updateStatusMutate,
+    isPending: isPendingUpdateStatus,
+    isError: isErrorUpdateStatus,
+    error: errorUpdateStatus,
+  } = useMutation({
+    mutationFn: (status) => updateStatus({ ticketId: id, status, currentStakeholder }),
+    onSuccess: (status) => {
+      queryClient.setQueryData(
+        ['ticket-details', id],
+        () => ({
```
gregorbg: Your solution is _almost_ perfect 😉 Instead of "grabbing" the `ticketDetails` variable from the outer scope (the return value of the `useQuery` above), the callback actually gets passed the current state of the query client as per the query key you're specifying.

So instead of...
```js
const ticketDetails = useQuery(someCoolQueryKey, queryFn);

queryClient.setQueryData(someCoolQueryKey, () => ({ ...ticketDetails, status }))
```

You should do something along the lines of...
```js
// Ignore that this variable exists! You are free to use it in your component, but not directly within the mutation success hook
const ticketDetails = { ... };

queryClient.setQueryData(someCoolQueryKey, (oldTicketDetails) => ({ ...oldTicketDetails, status }))
```

gregorbg: Further reading: https://tanstack.com/query/latest/docs/reference/QueryClient#queryclientsetquerydata

gregorbg: (I know, this is a very subtle change, and in your current use case it probably doesn't even make any difference. But there is a 1% chance that your current way of doing things can lead to subtle bugs when the original query result is refetched at the exact moment you're writing an update, and those bugs can be extreeeemely hard to track down and debug.)

gregorbg (changes requested):

app/webpacker/components/Tickets/index.jsx
```diff
@@ -79,9 +21,16 @@ export default function Wrapper({ id }) {
 }
 
 function Tickets({ id }) {
+  const queryClient = useQueryClient();
   const {
-    data: ticketDetails, sync, loading, error,
-  } = useLoadedData(actionUrls.tickets.show(id));
+    data: ticketDetails,
+    isLoading: isLoadingTicketDetails,
```
gregorbg: Here you should probably use `isPending`. The flag `isLoading` only covers the first, initial queryFn upon page load.

gregorbg (changes requested):

app/webpacker/components/Tickets/SkateholderSelector.jsx
danieljames-dj: Done

app/webpacker/components/Tickets/index.jsx
danieljames-dj: Okay changed

app/webpacker/components/Tickets/index.jsx
danieljames-dj: Okay understood, thanks for explaining in detail. I've now changed it.

gregorbg: As a sidenote, this is considered good practice in general when using `setQueryData`. Either pass in a constant(!!) value (for example, the return value passed to the `onSuccess` hook itself can be considered constant), or pass in a function that receives the old value and recomputes it into the new value.
Never compute something based on React state (anything that comes from any function called `use...`) and pretend like it's a constant value.

app/webpacker/components/Tickets/api/updateStatus.js
```diff
@@ -0,0 +1,19 @@
+import { fetchJsonOrError } from '../../../lib/requests/fetchWithAuthenticityToken';
+import { actionUrls } from '../../../lib/requests/routes.js.erb';
+
+export default async function updateStatus({ ticketId, status, currentStakeholder }) {
```
gregorbg: This signature is a bit inconsistent about IDs. For the ticket, it expects the ID as a parameter directly, but for the stakeholder, it expects the full object and then extracts the ID by itself.
Is there any particular reason why you need to do it this way?

gregorbg (changes requested):

app/webpacker/components/Tickets/api/updateStatus.js
danieljames-dj: Okay that was an overlook by me. I've changed the parameter to `currentStakeholderId`.

gregorbg (approved):

PR #11979, opened 2025-07-10
Fix ticket creation log

danieljames-dj:
Earlier, while creating the ticket, the log added was "update status" which was not right. So now fixing it by creating a new log type which is to create log.

### DIFF
--- a/app/models/tickets_edit_person.rb
+++ b/app/models/tickets_edit_person.rb
@@ -53,8 +53,7 @@ class TicketsEditPerson < ApplicationRecord
 
       TicketLog.create!(
         ticket_id: ticket.id,
-        action_type: TicketLog.action_types[:status_updated],
-        action_value: TicketsEditPerson.statuses[:open],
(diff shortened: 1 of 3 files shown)


app/models/tickets_edit_person.rb
```diff
@@ -53,8 +53,7 @@ def self.create_ticket(wca_id, changes_requested, requester)
 
       TicketLog.create!(
         ticket_id: ticket.id,
-        action_type: TicketLog.action_types[:status_updated],
-        action_value: TicketsEditPerson.statuses[:open],
```
gregorbg, line 57: Why did you get rid of this line?

danieljames-dj: That is actually optional and used to support the `action_type`. For ticket creation, there is no data to support, so it's left as null. If it was new comment, would have added comment ID, similarly if it was status update (like earlier), the new status could have been the action_value.

gregorbg (approved):

PR #11981, opened 2025-07-10
Suggest use of iOS Password in 2FA Page

dunkOnIT:
Suggested in this email thread: "Listing iOS passwords app as 2fa method"

> As of iOS 18, iOS has a native Passwords app that can be used to receive OTPs. I was just able to enable 2FA for my WCA account using it. Can this method be mentioned in the 2fa tab of the “Edit Profile” page? Currently only Google Authenticator and Microsoft Authenticator are listed as alternatives to Authy. Given that the Passwords app is pre-installed on iOS now, it may be among the most accessible 2fa options to the average person.

Seems reasonable.


config/locales/en.yml
```diff
@@ -615,7 +615,7 @@ en:
           support_desc_html: "The WCA website supports %{two_factor_link} using a one-time password:"
           #context: describes the 2FA options
           options:
-            app: "generated by a dedicated 2FA application on your smartphone (such as Authy or Google Authenticator for instance)."
+            app: "generated by a dedicated 2FA application on your smartphone (such as Authy, Google Authenticator or iOS Passwords)."
```
Rouxles:
```suggestion
            app: "generated by a dedicated 2FA application on your smartphone (such as Authy, Google Authenticator, or iOS Passwords)."
```
Not sure if this is consistent with other WST-related things, but added an Oxford comma:

dunkOnIT: I like it, thanks!

dunkOnIT:
@neelgore ah thanks for the clarification! Updated.

PR #11983, opened 2025-07-11
Create tickets for competition results

danieljames-dj:
There is no significant impact as of now. WRT can still post the results the old way. This is basically the MVP, will eventually add more features.

The immediate improvement for WRT is that now WRT can view the warnings & delegate's message in the website itself.

### DIFF
--- a/app/controllers/admin/results_controller.rb
+++ b/app/controllers/admin/results_controller.rb
@@ -47,8 +47,7 @@ module Admin
 
       ActiveRecord::Base.transaction do
         @updated_competitions.update(posting_user: current_user)
-        ticket_competition_result_ids = @updated_competitions.joins(:tickets_competition_result).pluck('tickets_competition_result.id')
-        TicketsCompetitionResult.where(id: ticket_competition_result_ids)
+        TicketsCompetitionResult.where(competition: @updated_competitions)
                                 .update_all(status: TicketsCompetitionResult.statuses[:warnings_verification])
       end
 
--- a/app/controllers/admin/results_controller.rb
+++ b/app/controllers/admin/results_controller.rb
@@ -46,9 +46,9 @@ module Admin
       return render json: { error: "No competitions to lock." } if @updated_competitions.empty?
 
       ActiveRecord::Base.transaction do
-        @updated_competitions.update(posting_user: current_user)
         TicketsCompetitionResult.where(competition: @updated_competitions)
                                 .update_all(status: TicketsCompetitionResult.statuses[:warnings_verification])
+        @updated_competitions.update(posting_user: current_user)
       end
 
       render json: { message: "Competitions successfully locked, go on posting!" }
--- a/app/controllers/admin/results_controller.rb
+++ b/app/controllers/admin/results_controller.rb
@@ -21,7 +21,10 @@ module Admin
             competitions: @pending_competitions.as_json(
               only: %w[id name results_submitted_at],
               methods: %w[city country_iso2],
-              include: { posting_user: user_attributes },
+              include: {
+                posting_user: user_attributes,
+                result_ticket: {},
+              },
             ),
           }
         end
@@ -41,6 +44,17 @@ module Admin
       #   it was a no-op.
       @updated_competitions = Competition.pending_posting.where(posting_user: nil).where(id: params[:competition_ids])
       return render json: { error: "No competitions to lock." } if @updated_competitions.empty?
+
+      ActiveRecord::Base.transaction do
+        @updated_competitions.update(posting_user: current_user)
+        @updated_competitions.each do |competition|
+          next if competition.tickets_competition_result.nil? # This is to temporarily allow the results which didn't get submitted through tickets.
+
+          competition.tickets_competition_result.update!(
+            status: TicketsCompetitionResult.statuses[:warnings_verification],
+          )
+        end
+      end
 
       json = { error: "Something went wrong." }
       json = { message: "Competitions successfully locked, go on posting!" } if @updated_competitions.update(posting_user: current_user)
--- a/app/controllers/admin/results_controller.rb
+++ b/app/controllers/admin/results_controller.rb
@@ -47,19 +47,12 @@ module Admin
 
       ActiveRecord::Base.transaction do
         @updated_competitions.update(posting_user: current_user)
-        @updated_competitions.each do |competition|
-          next if competition.tickets_competition_result.nil? # This is to temporarily allow the results which didn't get submitted through tickets.
-
-          competition.tickets_competition_result.update!(
-            status: TicketsCompetitionResult.statuses[:warnings_verification],
-          )
-        end
+        ticket_competition_result_ids = @updated_competitions.joins(:tickets_competition_result).pluck('tickets_competition_result.id')
+        TicketsCompetitionResult.where(id: ticket_competition_result_ids)
+                                .update_all(status: TicketsCompetitionResult.statuses[:warnings_verification])
       end
 
-      json = { error: "Something went wrong." }
-      json = { message: "Competitions successfully locked, go on posting!" } if @updated_competitions.update(posting_user: current_user)
-
-      render json: json
+      render json: { message: "Competitions successfully locked, go on posting!" }
     end
 
     def show
--- a/app/controllers/results_submission_controller.rb
+++ b/app/controllers/results_submission_controller.rb
@@ -151,9 +151,9 @@ class ResultsSubmissionController < ApplicationController
     message = params.require(:message)
     results_validator = ResultsValidators::CompetitionsResultsValidator.create_full_validation.validate(competition.id)
 
-    render status: :unprocessable_entity, json: { error: "Submitted results contain errors." } if results_validator.any_errors?
+    return render status: :unprocessable_entity, json: { error: "Submitted results contain errors." } if results_validator.any_errors?
 
-    render status: :unprocessable_entity, json: { error: "There is already a ticket associated, please contact WRT to update this." } if competition.result_ticket.present?
+    return render status: :unprocessable_entity, json: { error: "There is already a ticket associated, please contact WRT to update this." } if competition.result_ticket.present?
 
     CompetitionsMailer.results_submitted(competition, results_validator, message, current_user).deliver_now
 
--- a/app/controllers/results_submission_controller.rb
+++ b/app/controllers/results_submission_controller.rb
@@ -153,8 +153,14 @@ class ResultsSubmissionController < ApplicationController
 
     render status: :unprocessable_entity, json: { error: "Submitted results contain errors." } if results_validator.any_errors?
 
+    render status: :unprocessable_entity, json: { error: "There is already a ticket associated, please contact WRT to update this." } if competition.result_ticket.present?
+
     CompetitionsMailer.results_submitted(competition, results_validator, message, current_user).deliver_now
-    competition.touch(:results_submitted_at)
+
+    ActiveRecord::Base.transaction do
+      competition.touch(:results_submitted_at)
+      TicketsCompetitionResult.create_ticket(competition.id, message, current_user)
+    end
 
     render status: :ok, json: { success: true }
   end
--- a/app/controllers/results_submission_controller.rb
+++ b/app/controllers/results_submission_controller.rb
@@ -159,7 +159,7 @@ class ResultsSubmissionController < ApplicationController
 
     ActiveRecord::Base.transaction do
       competition.touch(:results_submitted_at)
-      TicketsCompetitionResult.create_ticket(competition.id, message, current_user)
+      TicketsCompetitionResult.create_ticket!(competition.id, message, current_user)
     end
 
     render status: :ok, json: { success: true }
--- a/app/models/tickets_competition_result.rb
+++ b/app/models/tickets_competition_result.rb
@@ -19,7 +19,7 @@ class TicketsCompetitionResult < ApplicationRecord
     end
   end
 
-  def self.create_ticket(competition_id, delegate_message, submitted_delegate)
+  def self.create_ticket!(competition_id, delegate_message, submitted_delegate)
     ticket_metadata = TicketsCompetitionResult.create!(
       status: TicketsCompetitionResult.statuses[:submitted],
       competition_id: competition_id,
--- a/app/models/tickets_competition_result.rb
+++ b/app/models/tickets_competition_result.rb
@@ -7,7 +7,7 @@ class TicketsCompetitionResult < ApplicationRecord
     submitted: "submitted",
     warnings_verification: "warnings_verification",
     results_verification: "results_verification",
-    closed: "closed",
+    posted: "posted",
   }
 
   has_one :ticket, as: :metadata
--- a/app/models/tickets_competition_result.rb
+++ b/app/models/tickets_competition_result.rb
@@ -1,0 +1,55 @@
+# frozen_string_literal: true
+
+class TicketsCompetitionResult < ApplicationRecord
+  self.table_name = "tickets_competition_result"
+
+  enum :status, {
+    submitted: "submitted",
+    warnings_verification: "warnings_verification",
+    results_verification: "results_verification",
+    closed: "closed",
+  }
+
+  has_one :ticket, as: :metadata
+
+  def action_user_groups(action)
+    case action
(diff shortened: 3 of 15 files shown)


app/models/tickets_competition_result.rb
```diff
@@ -0,0 +1,55 @@
...
+
+class TicketsCompetitionResult < ApplicationRecord
+  self.table_name = "tickets_competition_result"
+
+  enum :status, {
+    submitted: "submitted",
+    warnings_verification: "warnings_verification",
+    results_verification: "results_verification",
+    closed: "closed",
+  }
```
gregorbg: Shouldn't this more be like `posted` or `published`? "Closed" sounds like "disregarded" or "not useful".

app/webpacker/components/PostingCompetitions/index.js
```diff
@@ -106,6 +107,13 @@ function PostingCompetitionsIndex({
                   >
                     Import results page
                   </Button>
+                  <Button
+                    target="_blank"
+                    color="red"
+                    href={viewUrls.tickets.show(c.result_ticket.id)}
+                  >
+                    Tickets page
+                  </Button>
```
gregorbg: Are you certain that `c.result_ticket.id` will always exist? Thinking about the legacy transition period here...

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/index.jsx
```diff
@@ -0,0 +1,34 @@
...
+import { ticketsCompetitionResultStatuses } from '../../../../lib/wca-data.js.erb';
+import WarningsVerification from './WarningsVerification';
+import { adminImportResultsUrl } from '../../../../lib/requests/routes.js.erb';
+
+export default function CompetitionResultActionerView({ ticketDetails, updateStatus }) {
+  const { ticket: { metadata: { status, competition_id: competitionId } } } = ticketDetails;
+
+  switch (status) {
+    case ticketsCompetitionResultStatuses.submitted:
+      return <>Please lock the competition results from the Posting dashboard.</>;
```
gregorbg: Empty tag. Please use something like `p` or `div` or similar.

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/index.jsx
```diff
@@ -0,0 +1,34 @@
...
+      return (
+        <WarningsVerification
+          ticketDetails={ticketDetails}
+          updateStatus={updateStatus}
+        />
+      );
+
+    case ticketsCompetitionResultStatuses.results_verification:
+      return (
+        <>
```
gregorbg: Empty tag, see above

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/index.jsx
```diff
@@ -0,0 +1,34 @@
...
+          Please finish the remaining steps in
+          {' '}
+          <a href={adminImportResultsUrl(competitionId)}>import results page</a>
+          .
+        </>
+      );
+
+    default:
+      return <>Unknown status</>;
```
gregorbg: Empty tag, see above

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultRequesterView/index.jsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+
+export default function CompetitionResultRequesterView() {
+  return (
+    <>
```
gregorbg: Empty tag, see above

db/migrate/20250710084311_create_tickets_competition_result.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class CreateTicketsCompetitionResult < ActiveRecord::Migration[7.2]
+  def change
+    create_table :tickets_competition_result do |t|
```
gregorbg, line 5: Should this include a reference to the uploaded JSON?

app/controllers/admin/results_controller.rb
```diff
@@ -42,6 +45,17 @@ def start_posting
       @updated_competitions = Competition.pending_posting.where(posting_user: nil).where(id: params[:competition_ids])
       return render json: { error: "No competitions to lock." } if @updated_competitions.empty?
 
+      ActiveRecord::Base.transaction do
+        @updated_competitions.update(posting_user: current_user)
```
gregorbg, line 51: Why is this necessary now? I understand what this line does, but it should already have done that before your PR. Where is the `posting_user` being set before your PR? I don't see a deletion line that would explain a "move from there to here". I only see an added line. And that feels redundant

app/controllers/admin/results_controller.rb
```diff
@@ -42,6 +45,17 @@ def start_posting
...
+        @updated_competitions.each do |competition|
+          next if competition.tickets_competition_result.nil? # This is to temporarily allow the results which didn't get submitted through tickets.
+
+          competition.tickets_competition_result.update!(
+            status: TicketsCompetitionResult.statuses[:warnings_verification],
+          )
+        end
```
gregorbg: This should be possible via a `JOIN`. Something like `@updated_competitions.joins(:ticket_competition_result).update(status: ...)`

gregorbg: But please try out whether that really works, I'm kinda guessing stuff while writing my review here on GitHub

app/controllers/results_submission_controller.rb
```diff
@@ -153,8 +153,14 @@ def create
 
     render status: :unprocessable_entity, json: { error: "Submitted results contain errors." } if results_validator.any_errors?
 
+    render status: :unprocessable_entity, json: { error: "There is already a ticket associated, please contact WRT to update this." } if competition.result_ticket.present?
```
gregorbg: This message is not entirely clear to me. How would the old code behave at this place, and why do you need to change the procedure to manually contact WRT?

app/models/tickets_competition_result.rb
```diff
@@ -0,0 +1,55 @@
...
+  has_one :ticket, as: :metadata
+
+  def action_user_groups(action)
+    case action
+    when :update_status
+      [UserGroup.teams_committees_group_wrt]
+    end
+  end
+
+  def self.create_ticket(competition_id, delegate_message, submitted_delegate)
```
gregorbg:
```suggestion
  def self.create_ticket!(competition_id, delegate_message, submitted_delegate)
```

app/controllers/results_submission_controller.rb
```diff
@@ -153,8 +153,14 @@ def create
...
+
     CompetitionsMailer.results_submitted(competition, results_validator, message, current_user).deliver_now
-    competition.touch(:results_submitted_at)
+
+    ActiveRecord::Base.transaction do
+      competition.touch(:results_submitted_at)
+      TicketsCompetitionResult.create_ticket(competition.id, message, current_user)
```
gregorbg: Will this trigger any new email sendouts?

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultRequesterView/index.jsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+
+export default function CompetitionResultRequesterView() {
```
gregorbg, line 3: If this is basically a "nothing to see here, go on" view, then why do you even offer this view in the first place?

app/controllers/admin/results_controller.rb
danieljames-dj: Oh sorry I forgot to delete the old line. I did it now. Hope it's now clear.

app/controllers/admin/results_controller.rb
danieljames-dj: Didn't work as it was trying to update the competition, but I want to update `ticket_competition_result`.

Anyway, `joins` is a good idea to remove the ones which doesn't have ticket. So I've made it slightly better than current version.

app/controllers/results_submission_controller.rb
danieljames-dj: Ticket gets created when the result is already submitted. So during submission, we ideally won't find any ticket.

But in case we find a ticket during submission, that means somebody has already submitted, and the submission entry is cleared (maybe manually) so that the result can be re-imported. This is very rare case. In that case, it won't make sense to create another ticket for the new submission.

One possible way is to update the status & delegate_message here. But I don't feel so comfortable as it will destroy the previous message and status.

For now, I prefer WRT to handle it manually by setting `submitted_at` to `true` from PMA.

I'm having some long term plans to avoid this manual process. But the idea is very vague and may not be able to explain it in a comment.

app/controllers/results_submission_controller.rb
danieljames-dj: Nope

app/models/tickets_competition_result.rb
danieljames-dj: Once the system is completely migrated to tickets, `posted` may not mean that the process is completed. Even after posting, there can be a case where the WRT asks for some clarifications, or Delegate requests for some additional change, etc.

I'm thinking to store whether the results are posted or not in a boolean named `posted` within `results_competition_result` (not in this PR, but in future). This way, in case WRT has some clarification question or if Delegate wants to make some change, status will be changed to something like `queries_open` or any more meaningful names. But once that is completed, it should eventually reach back to the final state.

The final state can come and go - for example, it can be closed initially, and if some query come up after a week or two, it can change to `queries_open` and once resolved, again `closed`, and so on.

I'm not sure what else can be good status name for the final state. I prefer `close` as it matches with the `edit_person` tickets as well.

app/webpacker/components/PostingCompetitions/index.js
danieljames-dj: Oh right, I missed that case here. Thanks for pointing it out.

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/index.jsx
danieljames-dj: Okay changed.

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/index.jsx
danieljames-dj: Okay changed.

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/index.jsx
danieljames-dj: Changed

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultRequesterView/index.jsx
danieljames-dj: So that Delegates won't get confused if they see this ticket. Chances are less that they will see this as there is no direct link. But just in case they see, thought of giving this.

Also, in a later stage, they will definitely need a workbench here to request for result fixes, or any other requests.

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultRequesterView/index.jsx
danieljames-dj: Changed

db/migrate/20250710084311_create_tickets_competition_result.rb
danieljames-dj: Yes, I actually explored that. But I think I'll do it in a separate PR after this. Also, this table won't have reference. One ticket can have multiple JSONs, so I prefer a separate table to store the JSONs, and reference the latest one through a method or some other way.

app/models/tickets_competition_result.rb
danieljames-dj: Okay applied.

db/migrate/20250710084311_create_tickets_competition_result.rb
gregorbg: Actually, the stored jsons could point to a ticket. Legacy rows just have that pointer set to `null` and that way you should be good to go

app/models/tickets_competition_result.rb
gregorbg: I would argue that once a ticket is closed, it should never be opened again (except in very, very rare circumstances like technical glitch).
If you need a results change, start a new ticket.
If you have a follow-up query that is not yet clarified, don't set the ticket to its final state until the query has been clarified.

danieljames-dj: Okay, I agree with separate ticket for results change. Also closing the ticket only after finishing all follow-up sounds like a good idea. That should be implementable probably.

But coming to the actual question, do you still want the `status` to be `posted`, or are you fine with `closed`? Argument I'm having with `closed` is that it can be a common status for all types of tickets once it's completely resolved.

gregorbg: If you're aiming for such a "done for good" status flag on *all* tickets of *all types*, then you should probably go for an `is_closed` boolean column on the main tickets table.

danieljames-dj: Okay, I'll think about that. For now, if I go with `posted`, once we have a status in future which corresponds to "posted but open queries", can we have a new status that time named `posted_with_open_queries` and later change to `posted` once resolved? ie, are you fine with the status names `posted_with_open_queries` and `posted` as the last two statuses or do you have any other idea?

gregorbg: Yeah that could work

danieljames-dj: Okay changed.

app/controllers/admin/results_controller.rb
```diff
@@ -42,10 +45,14 @@ def start_posting
       @updated_competitions = Competition.pending_posting.where(posting_user: nil).where(id: params[:competition_ids])
       return render json: { error: "No competitions to lock." } if @updated_competitions.empty?
 
-      json = { error: "Something went wrong." }
-      json = { message: "Competitions successfully locked, go on posting!" } if @updated_competitions.update(posting_user: current_user)
+      ActiveRecord::Base.transaction do
+        @updated_competitions.update(posting_user: current_user)
+        ticket_competition_result_ids = @updated_competitions.joins(:tickets_competition_result).pluck('tickets_competition_result.id')
+        TicketsCompetitionResult.where(id: ticket_competition_result_ids)
+                                .update_all(status: TicketsCompetitionResult.statuses[:warnings_verification])
```
gregorbg: Here's one more thing to try: If you define a `belongs_to :competition` directly on `TicketsCompetitionResult`, then doing `TicketsCompetitionResult.where(competition: @updated_competitions).update_all(...)` might work

gregorbg: I am not sure, since we're not doing these composite model approaches very often, but please at least try it out to see whether it works or not

danieljames-dj: Yes, that worked. Great solution, thank you.

gregorbg (approved):

danieljames-dj:
@gregorbg During final testing, I figured out couple of issues, which I have now fixed.

One of them was related to your suggestion (`TicketsCompetitionResult.where(competition: @updated_competitions)`), I actually tested it earlier in `rails console` and it worked. This time, I found that the `@updated_competitions` will get changed once the field is updated.

I'll explain: the init of `@updated_competitions` is as follows:
```
@updated_competitions = Competition.pending_posting.where(posting_user: nil).where(id: params[:competition_ids])
```

Later, in transaction, following two lines were executed:
```
@updated_competitions.update(posting_user: current_user)
TicketsCompetitionResult.where(competition: @updated_competitions)
                                .update_all(status: TicketsCompetitionResult.statuses[:warnings_verification])
```

During second line, the `@updated_competitions` is fetched again that resulted in empty list. This is actually a new learning for me, I never thought after `update`, if it's executed again, it will refetch the first query again.

Anyway, I hope the fix is fine for you.

PR #11986, opened 2025-07-13
Merge users page in WRT panel (replacement of reassign WCA ID script too)

danieljames-dj:
Currently WRT panel has a page reassign WCA ID script to reassign WCA ID from one account to another. This PR aims at introducing a new page to 'merge two users'. This also means that to transfer WCA ID from one account to another, same feature can be used, and hence the existing reassign WCA ID script can be removed.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -4,8 +4,7 @@ class UsersController < ApplicationController
   before_action :authenticate_user!, except: %i[select_nearby_delegate acknowledge_cookies]
   before_action :check_recent_authentication!, only: %i[enable_2fa disable_2fa regenerate_2fa_backup_codes]
   before_action :set_recent_authentication!, only: %i[edit update enable_2fa disable_2fa]
-  before_action -> { redirect_to_root_unless_user(:can_admin_results?) }, only: %i[admin_search]
-  before_action -> { redirect_to_root_unless_user(:can_admin_results?) }, only: %i[merge]
+  before_action -> { redirect_to_root_unless_user(:can_admin_results?) }, only: %i[admin_search merge]
 
   RECENT_AUTHENTICATION_DURATION = 10.minutes.freeze
 


app/models/user.rb
```diff
@@ -1354,15 +1354,20 @@ def self.new_with_session(params, session)
     end
   end
 
+  def special_account_competitions
+    {
+      organized_competitions: organized_competitions.map(&:name),
+      delegated_competitions: delegated_competitions.map(&:name),
+      announced_competitions: competitions_announced.map(&:name),
+      results_posted_competitions: competitions_results_posted.map(&:name),
```
gregorbg: Important lesson about DB efficiency: Your current approach loads a bunch of models into memory, then reads one value from them, and discards them.

All 4 lines are doing the same concept, so I will pick one as an example: `organized_competitions.map(&:name)`.
1. It accesses the `organized_competitions` association, no data loading done yet because Rails doesn't know what will happen with that association yet
    - For example, if you call `some_association.count`, it only fires one `SELECT COUNT` call, without selecting the full individual rows. So Rails always waits what you do with an association before firing any calls, and calling only the association itself like `my_competition.some_association` does pretty much nothing
2. It calls `map` on the association. The method `map` is defined to move over entries of an array, so at this point Rails actually fires a `SELECT` call to retrieve all rows
3. For every single item in the array (which are all `Competition` models) you call the _method_ `name`.
    - The point that you're _calling a method_ is important, because that's the reason you need the `&` at the beginning of `&:name`.
4. The rest of the model is discarded once the `map` terminates. This is effectively wasted data, given the high amount of columns in the `competitions` table

gregorbg: Now here's the better version: `organized_competitions.pluck(:name)`. What it does differently:
1. It targets the same association (but nothing happens yet, see above`
2. It uses `pluck`. Instead of looping over an array of in-memory items, `pluck` is designed to give information to an SQL query (before that query is being executed!)
3. You're passing in `:name` as the name of the column(s) you're interested in.
    - Note that here, you do **not** need the `&` prefix, because you're not referencing a method to call. Instead, you're referencing the _name of a column in the table_, which is done via a straight-forward Rails symbol.
4. The result is a `SELECT name FROM competitions WHERE ...` directly, and the returned result is an array of names.

app/models/user.rb
```diff
@@ -1510,4 +1515,19 @@ def anonymize(new_wca_id = nil)
       country_iso2: special_account? ? country_iso2 : nil,
     )
   end
+
+  def transfer_data_to(user)
```
gregorbg: In the interest of readability, the parameter should probably be called `new_user` or `target_user` or similar.

app/models/user.rb
```diff
@@ -1510,4 +1515,19 @@ def anonymize(new_wca_id = nil)
...
+    competition_delegates.update_all(delegate_id: user.id)
+    competitions_results_posted.update_all(results_posted_by: user.id)
+    competitions_announced.update_all(announced_by: user.id)
+    roles.update_all(user_id: user.id)
+    registrations.update_all(user_id: user.id)
+
+    return if wca_id.blank?
+
+    wca_id_to_be_transferred = wca_id
+    self.update!(wca_id: nil) # Must remove WCA ID before adding it as it is unique in the Users table.
```
gregorbg: Can this backfire with validations? I think a lot of validations that we have depend on "Does this person have a WCA ID?"

And seeing that you already transferred a lot of relevant information (roles etc.) above, some of the validations might "freak out" if you're suddenly trying to nullify a WCA ID

app/webpacker/components/Panel/pages/MergeUsersPage/index.jsx
```diff
@@ -0,0 +1,46 @@
...
+      <Button
+        primary
+        disabled={!firstUserId || !secondUserId || firstUserId === secondUserId}
+        onClick={() => setModalOpen(true)}
+      >
+        Initiate Merge
+      </Button>
+      <Modal open={modalOpen}>
+        <Modal.Content>
+          <MergeUsers firstUserId={firstUserId} secondUserId={secondUserId} />
```
gregorbg, line 41: It feels a bit redundant to pass in the IDs and then fetch the user details again separately inside the `MergeUsers` component. The `AdminWcaSearch` component unfortunately uses `IdWcaSearch` under the hood, but you could modify it so that it takes an additional `idOnly` boolean flag as param, and when you set that to `false` it can patch through the whole model. Doing so would save you two network calls.

app/webpacker/components/Panel/pages/MergeUsersPage/SpecialAccountDetails.jsx
```diff
@@ -0,0 +1,37 @@
...
+
+export default function SpecialAccountDetails({ user }) {
+  const specialAccountCompetitionKeys = _.keys(user.special_account_competitions);
+  if (!user.roles.length && !specialAccountCompetitionKeys.length) {
+    return null;
+  }
+
+  return (
+    <Message>
+      <p>{`The user with email ${user.email} is a special account.`}</p>
```
gregorbg:
```suggestion
      <Message.Header>{`The user with email ${user.email} is a special account.`}</Message.Header>
```

app/webpacker/components/Panel/pages/MergeUsersPage/SpecialAccountDetails.jsx
```diff
@@ -0,0 +1,37 @@
...
+      {specialAccountCompetitionKeys.length && (
+        <p>
+          Following are the competitions which made the account of user a special account:
+          {specialAccountCompetitionKeys.map((specialAccountCompetitionKey) => (
+            <p>
+              {specialAccountCompetitionKey}
+              :
+              {user.special_account_competitions[specialAccountCompetitionKey].join(', ')}
+            </p>
+          ))}
```
gregorbg: This looks like a good use-case for a `List`, perhaps even a nested list for the individual items. Also look at `List.Header` and `List.Description`, for example at https://react.semantic-ui.com/elements/list/#types-divided or https://react.semantic-ui.com/elements/list/#types-tree (but I will leave the final design decision to you)

app/controllers/users_controller.rb
```diff
@@ -45,6 +46,43 @@ def index
...
+    render status: :ok, json: user.as_json(
+      include: %w[roles],
+      methods: %w[special_account_competitions],
+      private_attributes: %w[email],
+    )
+  end
+
+  def merge
+    to_be_maintained_user = User.find(params.require(:toBeMaintainedUserId))
+    to_be_anonymized_user = User.find(params.require(:toBeAnonymizedUserId))
```
gregorbg: The names (for both the Ruby variables as well as the corresponding params) are a bit lengthy. My first instinct would be to just drop the "to be" prefix, and call them "maintained user" and "anonymized user".
But I'm open to other ideas as well, as long as they're shorther than the current code.

gregorbg: Also, "anonymization" has a special meaning in other places in our code (GDPR etc) so is that really what you're going for here? Is that other user indeed being anonymized in the GDPR sense?
If not, you should reconsider the name here as well.

gregorbg (changes requested):

app/controllers/users_controller.rb
danieljames-dj: Actually when I think again, anonymization is not really necessary here. The reason why I was thinking to anonymize is that users should not continue use their old account. Or in other words somehow lock access to old account. But it is actually not helping, the user can still create a new account. So I'll remove the anonymizing part from here.

Coming to the variable name, what about `maintained_user` and `non_maintained_user`? Or rather simply `from_user` and `to_user` will work?

app/models/user.rb
danieljames-dj: I don't think that will be the case, because this was already happening in reassign WCA ID script. Also, will be going through few rounds of testing before the final merging. So if there are any issues, we will catch there.

app/webpacker/components/Panel/pages/MergeUsersPage/index.jsx
danieljames-dj: The component requires few other data as well like competitions_organized, competitions_delegated, roles, etc. So if I make AdminSearch to fetch these, it will do these calls in other places as well.

Also, I'm planning to reuse this component in DuplicatePersons page. If the user clicks 'merge' against one of the duplicate, I'm planning to show this view in a modal. At that time as well, I might not have all the necessary data.

So is it fine to go with the current way?

danieljames-dj (commented):
I didn't attend the coding changes questions yet. I've added some clarification questions, based on that I'll make all the changes in single go. Hope that works. Thank you.

app/controllers/users_controller.rb
gregorbg: I like `from_user` and `to_user` very much! But I will leave the final decision with you, because you know what works best for WRT

app/webpacker/components/Panel/pages/MergeUsersPage/index.jsx
gregorbg: Okay, in case of extra data and the potential for reuse later, then it's fine

app/models/user.rb
gregorbg: Your argument doesn't hold because the old script was doing ` User.where(id: account1_user.id).update_all(wca_id: nil)`. This directly translates to a raw SQL query, which Rails executes without performing model validation checks.
Your new code runs directly on an instance of User using `update!`, which means that validations are run and errors are thrown if validations fail. This was not the case in the past implementation.

More reading: https://www.mintbit.com/blog/update-all-vs-update-in-rails/

app/controllers/users_controller.rb
danieljames-dj: Okay changed to `from_user` and `to_user`.

app/models/user.rb
danieljames-dj: Thanks a lot for the detailed explanation. And sorry for not thinking this way. I've made the change now.

app/models/user.rb
danieljames-dj: Okay changed to `new_user`.

app/models/user.rb
danieljames-dj: Oh I didn't knew `update_all` will not run validations. Thanks for the new learning.

app/webpacker/components/Panel/pages/MergeUsersPage/SpecialAccountDetails.jsx
danieljames-dj: Changed.

app/webpacker/components/Panel/pages/MergeUsersPage/SpecialAccountDetails.jsx
danieljames-dj: Divided list looks like a good idea, implemented it. I'm not sure if nested will be good, because if I list down all competitions by one in each row, I think it will be very big page for special roles. So I think we can have them listed in a line separated by comma.

app/models/user.rb
danieljames-dj: I tested locally by merging two accounts, and one of them was of Sébastien, who have lots of data in website, and it got merged successfully. So I assume it won't backfire validations.

app/controllers/users_controller.rb
```diff
@@ -45,6 +46,42 @@ def index
     end
   end
 
+  def show
```
gregorbg: I don't think that the general-purpose "show" is a good place to expose such detailed, specific (and partially private!) information.
You're making changes to `users_controller`, which is "the global" way of accessing users. If you're making a view for merging, maybe something like `show_for_merge`? Or some other name, at least as long as it's not general-purpose.

app/controllers/users_controller.rb
```diff
@@ -45,6 +46,42 @@ def index
...
+    render status: :ok, json: user.as_json(
+      include: %w[roles],
+      methods: %w[special_account_competitions],
+      private_attributes: %w[email],
+    )
+  end
+
+  def merge
+    to_user = User.find(params.require(:toUserId))
+    from_user = User.find(params.require(:fromUserId))
```
gregorbg: Nitpick: These lines should perhaps be swapped, so that you get a "from --> to" impression when reading as well.

app/controllers/users_controller.rb
```diff
@@ -45,6 +46,42 @@ def index
...
+    if !current_user.results_team? && (to_user.special_account? || from_user.special_account?)
+      return render status: :bad_request,
+                    json: { error: 'One of the account is a special account, please contact WRT to merge them.' }
+    end
+
+    return render status: :bad_request, json: { error: "Cannot merge users with both having a WCA ID" } if to_user.wca_id.present? && from_user.wca_id.present?
+
+    ActiveRecord::Base.transaction do
+      from_user.transfer_data_to(to_user)
+    end
```
gregorbg: A transaction with one statement looks a bit silly. If the changes within `transfer_data_to` should be atomic, then wrap a transaction "internally" within the method body of `def transfer_data_to`.

app/models/user.rb
```diff
@@ -1510,4 +1515,19 @@ def anonymize(new_wca_id = nil)
...
+    competition_organizers.update_all(organizer_id: new_user.id)
+    competition_delegates.update_all(delegate_id: new_user.id)
+    competitions_results_posted.update_all(results_posted_by: new_user.id)
+    competitions_announced.update_all(announced_by: new_user.id)
+    roles.update_all(user_id: new_user.id)
+    registrations.update_all(user_id: new_user.id)
+
+    return if wca_id.blank?
+
+    wca_id_to_be_transferred = wca_id
```
gregorbg: Here is a good place to write `self.wca_id`, to make explicitly clear where this value is coming from

app/models/user.rb
gregorbg: Are you confident that we have enough tests in place, so that future validation conflicts will be detected early on?

app/webpacker/components/Panel/pages/MergeUsersPage/SpecialAccountDetails.jsx
```diff
@@ -0,0 +1,39 @@
...
+        </p>
+      )}
+      <List divided relaxed>
+        {specialAccountCompetitionKeys.length && (
+        <List.Item>
+          {specialAccountCompetitionKeys.map((specialAccountCompetitionKey) => (
+            <List.Content>
+              <List.Header>{specialAccountCompetitionKey}</List.Header>
+              <List.Description>
+                {user.special_account_competitions[specialAccountCompetitionKey].join(', ')}
```
gregorbg, line 33: Okay, the argument about the list growing too long for someone with many competitions is fair. Please add a comment here that you concsciously decided against a nested list / sub-list, so that in the future we don't accidentally "improve" this against WRT's will.

app/webpacker/components/Panel/pages/MergeUsersPage/api/getUserDetails.js
```diff
@@ -0,0 +1,14 @@
...
+import { viewUrls } from '../../../../../lib/requests/routes.js.erb';
+
+export default async function getUserDetails(userId) {
+  const { data } = await fetchJsonOrError(
+    viewUrls.users.show(userId),
+    {
+      headers: {
+        Accept: 'application/json',
+      },
+    },
```
gregorbg: Isn't this automatically part of `fetchJsonOrError`?!

gregorbg (changes requested):

app/controllers/users_controller.rb
danieljames-dj: Okay, `show_for_merge` looks good. Changed to that.

app/controllers/users_controller.rb
danieljames-dj: Oh yes, changed.

app/controllers/users_controller.rb
danieljames-dj: Okay done

app/models/user.rb
danieljames-dj: Okay done

app/models/user.rb
danieljames-dj: I'm around 90% confident. The only risk I am seeing is that in worst case, the action will just fail, and 'merging users' won't be usually critical action, so WST will get some time to fix it.

app/webpacker/components/Panel/pages/MergeUsersPage/SpecialAccountDetails.jsx
danieljames-dj: Okay done

app/webpacker/components/Panel/pages/MergeUsersPage/api/getUserDetails.js
danieljames-dj: Not sure, but without this, it is working now, so removed. I cannot recall exactly why I had this earlier, mostly a miss. Anyway thanks.

app/controllers/users_controller.rb
```diff
@@ -5,6 +5,7 @@ class UsersController < ApplicationController
   before_action :check_recent_authentication!, only: %i[enable_2fa disable_2fa regenerate_2fa_backup_codes]
   before_action :set_recent_authentication!, only: %i[edit update enable_2fa disable_2fa]
   before_action -> { redirect_to_root_unless_user(:can_admin_results?) }, only: %i[admin_search]
+  before_action -> { redirect_to_root_unless_user(:can_admin_results?) }, only: %i[merge]
```
gregorbg: If I'm not mistaken, then these two `before_action` bodies are completely similar. You can merge them into one if you use `only: %i[admin_search merge]`

gregorbg (changes requested):

app/controllers/users_controller.rb
danieljames-dj: Oh yes, thank you

gregorbg (approved):

PR #11987, opened 2025-07-13
Reuse common code between upload-json and import-live

danieljames-dj:
My initial attempt was to remove `upload_json` model. But it's pretty difficult, because lots of things are done there, and those are very very critical things as well. So I feel it's better to start very slow.

As a first step, I'm merging the common code used in both places.

### DIFF
--- a/app/controllers/results_submission_controller.rb
+++ b/app/controllers/results_submission_controller.rb
@@ -47,16 +47,14 @@ class ResultsSubmissionController < ApplicationController
                                    competition_id: competition.id,
                                  })
 
-    mark_result_submitted = ActiveRecord::Type::Boolean.new.cast(params.require(:mark_result_submitted))
-    store_uploaded_json = ActiveRecord::Type::Boolean.new.cast(params.require(:store_uploaded_json))
+    return render status: :unprocessable_entity, json: { error: upload_json.errors.full_messages } unless upload_json.valid?
 
-    # This makes sure the json structure is valid!
-    return render status: :unprocessable_entity, json: { error: upload_json.errors.full_messages } unless upload_json.import_to_inbox
+    temporary_results_data = upload_json.temporary_results_data
+    results_to_import = temporary_results_data[:results_to_import]
+    scrambles_to_import = temporary_results_data[:scrambles_to_import]
+    persons_to_import = temporary_results_data[:persons_to_import]
 
-    competition.touch(:results_submitted_at) if !competition.results_submitted? && mark_result_submitted
-    competition.uploaded_jsons.create!(json_str: upload_json.results_json_str) if store_uploaded_json
-
-    render status: :ok, json: { success: true }
+    import_temporary_results(competition, results_to_import, scrambles_to_import, persons_to_import, upload_json.results_json_str)
   end
 
   def import_from_live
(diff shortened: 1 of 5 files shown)


app/controllers/results_submission_controller.rb
```diff
@@ -122,17 +120,30 @@ def import_from_live
...
     ActiveRecord::Base.transaction do
       InboxPerson.where(competition_id: competition.id).delete_all
       InboxResult.where(competition_id: competition.id).delete_all
       Scramble.where(competition_id: competition.id).delete_all
       InboxPerson.import!(persons_to_import)
-      InboxResult.import!(results_to_import)
       Scramble.import!(scrambles_to_import)
+      InboxResult.import!(results_to_import)
+
+      competition.touch(:results_submitted_at) if params.key?(:mark_result_submitted) && ActiveRecord::Type::Boolean.new.cast(params.require(:mark_result_submitted)) && !competition.results_submitted?
```
gregorbg: Too much happening on one single line

app/controllers/results_submission_controller.rb
```diff
@@ -122,17 +120,30 @@ def import_from_live
...
+
+      competition.uploaded_jsons.create!(json_str: results_json_str) if params.key?(:store_uploaded_json) && ActiveRecord::Type::Boolean.new.cast(params.require(:store_uploaded_json))
```
gregorbg: Again, too much happening on one line

spec/models/upload_json_spec.rb
```diff
@@ -58,10 +58,11 @@
     }.to_json
 
     upload_json = build(:upload_json, competition_id: competition.id, results_json_str: results_json)
+    temporary_results_data = upload_json.temporary_results_data
 
-    expect(upload_json.import_to_inbox).to be true
-    expect(InboxResult.count).to eq 1
-    inbox_result = InboxResult.first
+    expect(upload_json.valid?).to be true
```
gregorbg:
```suggestion
    expect(upload_json).to be_valid
```

gregorbg: Also, this is not a formal replacement of the old test case.
The old case was actually _running_ the import routine, seeing if the act of inserting stuff into the DB also works.
The question "do validations pass" is much more harmless and does not cover nearly as much code.

spec/models/upload_json_spec.rb
```diff
@@ -58,10 +58,11 @@
...
+    expect(temporary_results_data[:results_to_import].count).to eq 1
+    inbox_result = temporary_results_data[:results_to_import].first
```
gregorbg: These lines don't need to be changed if you can actually trigger the import.

app/controllers/results_submission_controller.rb
```diff
@@ -122,17 +120,30 @@ def import_from_live
       end
     end
 
+    import_temporary_results(competition, results_to_import, scrambles_to_import, persons_to_import)
```
gregorbg: What about moving this to some `lib` module? Create a new one in case of doubt... Makes it easier to call in tests, and also keeps the controller code short and sweet.

danieljames-dj: Good idea, moved to lib. Created a new lib named `competition_results` which can be used for methods related to competition results.

app/controllers/results_submission_controller.rb
danieljames-dj: Okay fixed it.

app/controllers/results_submission_controller.rb
danieljames-dj: Fixed it too

spec/models/upload_json_spec.rb
danieljames-dj: Thanks, changed it.

spec/models/upload_json_spec.rb
danieljames-dj: Okay now running import as well.

app/controllers/results_submission_controller.rb
gregorbg: I'd argue that `competition_results` is a bit too broad. Half of the code we have under `lib` could be "related to competition results". Try to keep it concise, something like `results_import` or similar should do.

danieljames-dj: Okay changed to `competition_results_import`, that works?

spec/models/upload_json_spec.rb
```diff
@@ -58,8 +58,16 @@
...
+    CompetitionResultsImport.import_temporary_results(
+      competition,
+      temporary_results_data,
+      mark_result_submitted: false,
+      store_uploaded_json: true,
+      results_json_str: upload_json.results_json_str,
+    )
 
-    expect(upload_json.import_to_inbox).to be true
+    expect(upload_json).to be_valid
```
gregorbg: You should move this test before the actual import. Trying to import from a (potentially) invalid JSON doesn't make any sense.

gregorbg (changes requested):

spec/models/upload_json_spec.rb
danieljames-dj: Done

gregorbg (approved):

gregorbg (approved):

PR #12006, opened 2025-07-14
Permission handling in tickets for status change and commenting

danieljames-dj:
This is an attempt to bring the ticket related permissions to frontend as well, so that user will clearly know what they can do.

### DIFF
--- a/app/controllers/ticket_comments_controller.rb
+++ b/app/controllers/ticket_comments_controller.rb
@@ -17,7 +17,7 @@ class TicketCommentsController < ApplicationController
     ticket = Ticket.find(params.require(:ticket_id))
     acting_stakeholder = TicketStakeholder.find(params.require(:acting_stakeholder_id))
 
-    return head :bad_request unless ticket.user_stakeholders(current_user).map(&:id).include?(acting_stakeholder.id)
+    return head :bad_request unless ticket.user_stakeholders(current_user).pluck(:id).include?(acting_stakeholder.id)
     return head :unauthorized unless acting_stakeholder.actions_allowed.include?(:add_comment)
 
     comment = TicketComment.create!(
--- a/app/controllers/ticket_comments_controller.rb
+++ b/app/controllers/ticket_comments_controller.rb
@@ -17,7 +17,7 @@ class TicketCommentsController < ApplicationController
     ticket = Ticket.find(params.require(:ticket_id))
     acting_stakeholder = TicketStakeholder.find(params.require(:acting_stakeholder_id))
 
-    return head :bad_request unless ticket.user_stakeholders(current_user).pluck(:id).include?(acting_stakeholder.id)
+    return head :bad_request unless ticket.user_stakeholders(current_user).include?(acting_stakeholder)
     return head :unauthorized unless acting_stakeholder.actions_allowed.include?(:add_comment)
 
     comment = TicketComment.create!(
--- a/app/controllers/ticket_comments_controller.rb
+++ b/app/controllers/ticket_comments_controller.rb
@@ -15,17 +15,16 @@ class TicketCommentsController < ApplicationController
 
   def create
     ticket = Ticket.find(params.require(:ticket_id))
-    acting_stakeholder_id = params.require(:acting_stakeholder_id)
+    acting_stakeholder = TicketStakeholder.find(params.require(:acting_stakeholder_id))
 
-    # Currently only stakeholders can create comments.
-    return head :unauthorized unless ticket.can_user_access?(current_user)
-    return head :bad_request unless ticket.user_stakeholders(current_user).map(&:id).include?(acting_stakeholder_id)
+    return head :bad_request unless ticket.user_stakeholders(current_user).map(&:id).include?(acting_stakeholder.id)
+    return head :unauthorized unless acting_stakeholder.actions_allowed.include?(:add_comment)
 
     comment = TicketComment.create!(
       ticket: ticket,
       comment: params.require(:comment),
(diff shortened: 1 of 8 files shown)


app/models/tickets_edit_person.rb
```diff
@@ -11,11 +11,8 @@ class TicketsEditPerson < ApplicationRecord
   has_one :ticket, as: :metadata
   has_many :tickets_edit_person_fields
 
-  def action_user_groups(action)
-    case action
-    when :update_status
-      [UserGroup.teams_committees_group_wrt]
-    end
+  def actions_allowed
+    %i[add_comment update_status]
```
gregorbg, line 17: These action literals should probably be standardized into constants somewhere, to reduce the risk of typos (for example `if actions_allowed.include?(:update_stauts)` or similar) as much as possible

app/models/ticket_stakeholder.rb
```diff
@@ -34,6 +30,14 @@ def user_stakeholder?
     stakeholder_type == "User"
   end
 
+  def actions_allowed
+    ticket.metadata.actions_allowed(self)
+  end
+
+  DEFAULT_SERIALIZE_OPTIONS = {
+    methods: %w[stakeholder actions_allowed],
```
gregorbg: Not really related to the intention of your current PR, but if you're touching `DEFAULT_SERIALIZE_OPTIONS` anways, then you can also move `stakeholder` to `include` instead of `methods`.

app/models/ticket_stakeholder.rb
```diff
@@ -34,6 +30,14 @@ def user_stakeholder?
     stakeholder_type == "User"
   end
 
+  def actions_allowed
+    ticket.metadata.actions_allowed(self)
```
gregorbg: If I understand your idea correctly, then the method `actions_allowed` on the metadata should probably be renamed to something like `actions_allowed_for`, to make clear that the actions relate to a particular stakeholder.

gregorbg: Also, unrelated comment: Can you declare the metadata as a (polymorphic) `has_one` with `through: ticket`? Or does that not work because of polymorphism?

gregorbg (changes requested):

app/models/ticket_stakeholder.rb
danieljames-dj: Yes, `actions_allowed_for` is better, changed it.

I tried around 5-10 minutes on figuring out if there is a way for `has_one`, but couldn't get it. When I give something like `has_one :metadata, through: :ticket, source: :metadata`, I'm getting error that `source_type` is needed but I cannot give it because metadata is polymorphic.

app/models/ticket_stakeholder.rb
danieljames-dj: Okay changed. Thank you.

app/models/tickets_edit_person.rb
danieljames-dj: As per our discussion in chat, I'll do this in a separate PR after this.

app/controllers/ticket_comments_controller.rb
```diff
@@ -15,17 +15,16 @@ def index
 
   def create
     ticket = Ticket.find(params.require(:ticket_id))
-    acting_stakeholder_id = params.require(:acting_stakeholder_id)
+    acting_stakeholder = TicketStakeholder.find(params.require(:acting_stakeholder_id))
 
-    # Currently only stakeholders can create comments.
-    return head :unauthorized unless ticket.can_user_access?(current_user)
-    return head :bad_request unless ticket.user_stakeholders(current_user).map(&:id).include?(acting_stakeholder_id)
+    return head :bad_request unless ticket.user_stakeholders(current_user).map(&:id).include?(acting_stakeholder.id)
```
gregorbg: In the internal code of `user_stakeholders`, you're relying on ActiveRecord `or`, which returns a "clean" association, so I wonder whether `.pluck(:id)` would work here... Please try it out and let me know 🙃

app/controllers/tickets_controller.rb
```diff
@@ -65,10 +65,10 @@ def show
   def update_status
     ticket = Ticket.find(params.require(:ticket_id))
     ticket_status = params.require(:ticket_status)
-    acting_stakeholder_id = params.require(:acting_stakeholder_id)
+    acting_stakeholder = TicketStakeholder.find(params.require(:acting_stakeholder_id))
 
-    return head :unauthorized unless ticket.action_allowed?(:update_status, current_user)
-    return head :bad_request unless ticket.user_stakeholders(current_user).map(&:id).include?(acting_stakeholder_id)
+    return head :bad_request unless ticket.user_stakeholders(current_user).map(&:id).include?(acting_stakeholder.id)
```
gregorbg: (see other comment about potentially using `.pluck(:id)`, feel free to just discard this comment if the other line doesn't work)

app/models/ticket_stakeholder.rb
gregorbg: Okay, then the polymorphism makes it too complicated probably. Let's keep it as-is then!

app/controllers/ticket_comments_controller.rb
danieljames-dj: Yes that is working, thank you. Changed it.

app/controllers/tickets_controller.rb
danieljames-dj: Changed that too

gregorbg (approved):

PR #12009, opened 2025-07-14
Add graphql API for WCIF

FinnIckler:
To solve the Issue with having to return the whole WCIF data on every request and killing venue WiFIs and generating huge data transfer bills this introduces a graphql API for the WCIF endpoint.

I have generated a lot of the code from https://graphql-ruby.org/getting_started and then used the JSON schemas to translate them to the Graphql DSL used by the gem. 

There is still lots to be done for all the nested data, currently only top level attributes can be queried. 
I am opening this for feedback already.

Basic query working in the graphiql editor:
Screenshot 2025-07-14 at 14 17 53

### DIFF
--- a/spec/controllers/server_status_controller_spec.rb
+++ b/spec/controllers/server_status_controller_spec.rb
@@ -105,6 +105,6 @@ RSpec.describe "RegulationsCheck" do
     status, description = check.status_description
 
     expect(status).to eq :danger
-    expect(description).to match(/Error while loading regulations: unexpected character: 'i' at line 1 column 1 from JSON::ParserError/)
+    expect(description).to match(/JSON::ParserError/)
   end
 end


FinnIckler:
Ok, it was actually very easy to handle the nested data. You just needed to add `method: :persons_wcif` and then change the fields to camelCase like they are in the wcif and now it just works!

FinnIckler:
@thewca-bot deploy staging

FinnIckler:
@thewca-bot deploy staging

FinnIckler:
I added authorization by moving to the api controller and using it to check if the user can manage the competition.
In the models I added a visibility check so those fields can't be queried if not authorized.

FinnIckler:
@thewca-bot deploy staging

FinnIckler:
@thewca-bot deploy staging

PR #12010, opened 2025-07-16
Add handling for InvalidAuthenticityToken error

dunkOnIT:
We get emails from users who triggered InvalidAuthenticityToken errors every few days. The most likely culprit for this is leaving the login/sign up page open for too long, and the token having expired by the time they submit the form.

This PR gracefully handles the issue by telling the user the session expired and redirects them back to the last page they were on (with homepage as a fallback). If I'm lucky, I'll never need to see another "I can't log in" support email again in my life :)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -21,6 +21,8 @@ class ApplicationController < ActionController::Base
     ::NewRelic::Agent.add_custom_attributes(HTTP_USER_AGENT: request.user_agent)
   end
 
+  rescue_from ActionController::InvalidAuthenticityToken, with: :handle_invalid_auth_token
+
   def self.locale_counts
     @@locale_counts
   end
@@ -134,5 +136,10 @@ class ApplicationController < ActionController::Base
     def store_user_location!
       # :user is the scope we are authenticating
       store_location_for(:user, request.fullpath)
+    end
+
+    def handle_invalid_auth_token
+      reset_session
+      redirect_back fallback_location: root_path, alert: I18n.t('devise.sessions.expired')
     end
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -21,8 +21,6 @@ class ApplicationController < ActionController::Base
     ::NewRelic::Agent.add_custom_attributes(HTTP_USER_AGENT: request.user_agent)
   end
 
-  rescue_from ActionController::InvalidAuthenticityToken, with: :handle_invalid_auth_token
-
   def self.locale_counts
     @@locale_counts
   end
@@ -136,10 +134,5 @@ class ApplicationController < ActionController::Base
     def store_user_location!
       # :user is the scope we are authenticating
       store_location_for(:user, request.fullpath)
-    end
-
-    def handle_invalid_auth_token
-      reset_session
-      redirect_back fallback_location: root_path, alert: I18n.t('devise.sessions.expired')
     end
 end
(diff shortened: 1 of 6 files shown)


app/controllers/application_controller.rb
```diff
@@ -135,4 +137,9 @@ def store_user_location!
       # :user is the scope we are authenticating
       store_location_for(:user, request.fullpath)
     end
+
+    def handle_invalid_auth_token
+      reset_session
```
gregorbg: What is this `reset_session` and why does it become necessary when handling the error?

gregorbg: In particular, it is my understanding that we're only running into this error handler _because_ the session of a tab that was open too long has expired. This line reads to me like "Oh if the session has expired, then please expire it".

dunkOnIT: It doesn't just expire it, it "initializes a new session object" per [the docs](https://apidock.com/rails/ActionController/Base/reset_session) - my understanding is that without this, the session would just remain expired

spec/requests/api_competitions_spec.rb
```diff
@@ -549,10 +549,14 @@
 
         it "prevents from CSRF attacks" do
           headers["ACCESS_TOKEN"] = "INVALID"
```
gregorbg: Aaaah. This test case exactly reveals what has been my gut all along:
We absolutely cannot merge this feature in its current state. You're basically giving anybody who tries to fake a CSRF token the opportunity to get issued a valid token and retry.

The core idea of this test is that "if you supply an invalid token, then you shall not pass". This is good and necessary, it's called "web security". Re-issuing a token when something invalid gets submitted implicitly breaks our whole security system and cannot go live.

dunkOnIT:
Changed the PR to serve a custom error page for the CSRF error, and also for Competition-specific 404's

app/controllers/errors_controller.rb
```diff
@@ -7,7 +7,15 @@ def show
     @exception = request.env["action_dispatch.exception"]
     @status_code = ActionDispatch::ExceptionWrapper.new(request.env, @exception).status_code
     @request_id = request.env["action_dispatch.request_id"]
-    render error_page(@status_code), status: @status_code
+
+    if @exception.instance_of?(ActiveRecord::RecordNotFound) && @exception.model == "Competition"
+      @id = params['id']
+      render 'competition_not_found'
```
gregorbg:
```suggestion
      render 'competition_not_found', status: :not_found
```

app/controllers/errors_controller.rb
```diff
@@ -7,7 +7,15 @@ def show
...
+    elsif @exception.instance_of?(ActionController::InvalidAuthenticityToken)
+      render 'session_expired'
```
gregorbg:
```suggestion
      render 'session_expired', status: :unprocessable_entity
```

app/views/errors/competition_not_found.html.erb
```diff
@@ -0,0 +1,8 @@
+<div class="dialog" style="display: flex; height: 80vh; justify-content: center; align-items:center">
+  <div>
```
gregorbg, line 2: Why is there a div inside a div here?

app/views/errors/session_expired.html.erb
```diff
@@ -0,0 +1,7 @@
+<div class="dialog" style="display: flex; height: 80vh; justify-content: center; align-items:center">
+  <div>
```
gregorbg, line 2: Same as above: div inside a div

gregorbg (changes requested):

app/views/errors/competition_not_found.html.erb
dunkOnIT: I copied over from how we're already doing [404.html.erb](https://github.com/thewca/worldcubeassociation.org/blob/main/app/views/errors/404.html.erb)

gregorbg (approved):

PR #12017, opened 2025-07-19
Combine ticket log action_types and other actions

danieljames-dj:
This PR mainly involves merging the ticket_log action types with the allowed actions for stakeholders. Also, moved some code in controller to private methods, because the controller will have more actions soon, and those actions can also reuse the method.

### DIFF
--- a/app/controllers/ticket_comments_controller.rb
+++ b/app/controllers/ticket_comments_controller.rb
@@ -14,11 +14,8 @@ class TicketCommentsController < ApplicationController
   end
 
   def create
-    ticket = Ticket.find(params.require(:ticket_id))
-    acting_stakeholder = TicketStakeholder.find(params.require(:acting_stakeholder_id))
-
-    return head :bad_request unless ticket.user_stakeholders(current_user).include?(acting_stakeholder)
-    return head :unauthorized unless acting_stakeholder.actions_allowed.include?(:add_comment)
+    ticket, acting_stakeholder = ticket_and_acting_stakeholder_from_params
+    return if check_ticket_errors(ticket, acting_stakeholder, TicketLog.action_types[:create_comment])
 
     comment = TicketComment.create!(
       ticket: ticket,
@@ -28,5 +25,26 @@ class TicketCommentsController < ApplicationController
     )
 
     render json: comment
+  end
+
+  private def ticket_and_acting_stakeholder_from_params
+    ticket = Ticket.find(params.require(:ticket_id))
+    acting_stakeholder = TicketStakeholder.find(params.require(:acting_stakeholder_id))
+
+    [ticket, acting_stakeholder]
+  end
+
+  private def check_ticket_errors(ticket, acting_stakeholder, action)
+    unless ticket.user_stakeholders(current_user).include?(acting_stakeholder)
+      render status: :bad_request, json: { error: "You are not a stakeholder for this ticket." }
+      return true
+    end
+
+    unless acting_stakeholder.actions_allowed.include?(action)
+      render status: :unauthorized, json: { error: "You are not allowed to perform this action." }
+      return true
+    end
+
+    false
   end
 end
--- a/app/controllers/ticket_comments_controller.rb
+++ b/app/controllers/ticket_comments_controller.rb
@@ -2,6 +2,19 @@
 
 class TicketCommentsController < ApplicationController
   before_action :authenticate_user!
+  before_action -> { check_ticket_errors(TicketLog.action_types[:create_comment]) }, only: [:create]
+
+  private def check_ticket_errors(action_type)
+    @action_type = action_type
+    @ticket = Ticket.find(params.require(:ticket_id))
+    @acting_stakeholder = TicketStakeholder.find(params.require(:acting_stakeholder_id))
+
+    render status: :bad_request, json: { error: "You are not a stakeholder for this ticket." } unless @ticket.user_stakeholders(current_user).include?(@acting_stakeholder)
+
+    render status: :unauthorized, json: { error: "You are not allowed to perform this action." } unless @acting_stakeholder.actions_allowed.include?(@action_type)
+
+    false
+  end
 
   def index
     ticket = Ticket.find(params.require(:ticket_id))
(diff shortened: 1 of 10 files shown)


app/controllers/ticket_comments_controller.rb
```diff
@@ -29,4 +26,25 @@ def create
...
   end
+
+  private def ticket_and_acting_stakeholder_from_params
+    ticket = Ticket.find(params.require(:ticket_id))
+    acting_stakeholder = TicketStakeholder.find(params.require(:acting_stakeholder_id))
+
+    [ticket, acting_stakeholder]
+  end
+
+  private def check_ticket_errors(ticket, acting_stakeholder, action)
```
gregorbg: This is probably best done as part of the Rails filter chain. The idea is to declare a `before_action :check_ticket_errors, only: [:create]` (where the `only` part can be adjusted as needed), because the Rails controller chain automatically halts and returns early, if any `render` or `redirect` has been called during a `before_action`.
It effectively gets rid of the need to return a boolean for you.

gregorbg: Also, when you want to use variables from database calls, like `ticket_and_acting_stakeholder_from_params`, you can do this:
- In your `before_action`, you store these variables into `@ticket` instance variables.
- If the filter chain halts, because some of your business logic checks failed, then Rails will just return early and the response will already go out to the client (ie React code in the browser). The `@ticket` variable is a little bit "useless", compared to a normal `ticket` local variable.
- But if the filter chain does not halt, and all pre-checks pass, then the `@ticket` declared in the `before_action` method body will also be available in the normal controller method body without any additional action needed.

gregorbg: (Sidenote: In an upcoming Rubocop release, the linter will actually fail on these "force to return a dummy boolean just so that I can return early in the controller" methods, because it's not considered good practice in Rails. See some controller files in https://github.com/thewca/worldcubeassociation.org/pull/11962/files for reference)

app/controllers/tickets_controller.rb
```diff
@@ -189,4 +188,25 @@ def anonymize
...
   end
+
+  private def ticket_and_acting_stakeholder_from_params
+    ticket = Ticket.find(params.require(:ticket_id))
+    acting_stakeholder = TicketStakeholder.find(params.require(:acting_stakeholder_id))
+
+    [ticket, acting_stakeholder]
+  end
+
+  private def check_ticket_errors(ticket, acting_stakeholder, action)
```
gregorbg: Same thing about a pseudo-boolean being returned, which should be converted to `before_action` logic.

db/migrate/20250719140801_change_ticket_log_action_type_enum.rb
```diff
@@ -0,0 +1,25 @@
...
+class ChangeTicketLogActionTypeEnum < ActiveRecord::Migration[7.2]
+  def up
+    execute <<-SQL.squish
+      UPDATE ticket_logs
+      SET action_type = CASE action_type
+                        WHEN 'created' THEN 'create_ticket'
+                        WHEN 'status_updated' THEN 'update_status'
+                        ELSE action_type
+                        END;
+    SQL
```
gregorbg: You can write this in Rails language:
```ruby
TicketLog.where(action_type: 'created').update_all(action_type: 'create_ticket')
```

gregorbg: t would require one line (ie one `UPDATE` statement) per type, instead of a compact `CASE`. But it's more maintainable if we ever need to revisit or adjust this migration.

gregorbg: But I am not sure how Rails will react to the fact that you're changing enum values that "don't exist" anymore (because you're hard-replacing them in the model code above). Have you tried the AR approach before?

gregorbg (changes requested):

db/migrate/20250719140801_change_ticket_log_action_type_enum.rb
danieljames-dj: I actually didn't try the AR approach thinking that rails will throw error as the enum value doesn't exist. But it worked as you said. Changed it.

app/controllers/ticket_comments_controller.rb
danieljames-dj: Wow this is actually cool - I knew regarding `before_action` but I didn't go with that just because I have to initialize `ticket` and `acting_stakeholder` two times. But reusing it by initializing them with `@` is a great idea. Changed it.

app/controllers/tickets_controller.rb

db/migrate/20250719140801_change_ticket_log_action_type_enum.rb
gregorbg: Cool! I think this has to do with the fact that `update_all` does not run any form of validations. It's basically a "raw translation" of SQL `UPDATE` statement. Only `update` and `update!` that are called on an individual model go through validations.

app/controllers/ticket_comments_controller.rb
```diff
@@ -2,6 +2,19 @@
...
+  private def check_ticket_errors(action_type)
+    @action_type = action_type
+    @ticket = Ticket.find(params.require(:ticket_id))
+    @acting_stakeholder = TicketStakeholder.find(params.require(:acting_stakeholder_id))
+
+    render status: :bad_request, json: { error: "You are not a stakeholder for this ticket." } unless @ticket.user_stakeholders(current_user).include?(@acting_stakeholder)
+
+    render status: :unauthorized, json: { error: "You are not allowed to perform this action." } unless @acting_stakeholder.actions_allowed.include?(@action_type)
+
+    false
```
gregorbg: This is now unnecessary. `before_action` does not need any explicit return value at all.

app/controllers/tickets_controller.rb
```diff
@@ -4,12 +4,25 @@ class TicketsController < ApplicationController
...
+  private def check_ticket_errors(action_type)
+    @action_type = action_type
+    @ticket = Ticket.find(params.require(:ticket_id))
+    @acting_stakeholder = TicketStakeholder.find(params.require(:acting_stakeholder_id))
+
+    render status: :bad_request, json: { error: "You are not a stakeholder for this ticket." } unless @ticket.user_stakeholders(current_user).include?(@acting_stakeholder)
+
+    render status: :unauthorized, json: { error: "You are not allowed to perform this action." } unless @acting_stakeholder.actions_allowed.include?(@action_type)
+
+    false
```
gregorbg: Unnecessary now, see above

gregorbg (changes requested):

app/controllers/ticket_comments_controller.rb
danieljames-dj: Done

app/controllers/tickets_controller.rb

PR #12043, opened 2025-07-21
Disallow deprecated fields in roles API

danieljames-dj:
This change is because roles API is doing too many DB queries.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1251,10 +1251,21 @@ class User < ApplicationRecord
     # NOTE: doing deep_dup is necessary here to avoid changing the inner values
     # of the freezed variables (which would leak PII)!
     default_options = DEFAULT_SERIALIZE_OPTIONS.deep_dup
+
+    include_email = options&.[](:include_email)
+    exclude_deprecated = options&.[](:exclude_deprecated)
+
     # Delegates's emails and regions are public information.
-    default_options[:methods].push("email", "location", "region_id") if staff_delegate?
+    default_options[:methods].push("email") if include_email || staff_delegate?
+    default_options[:methods].push("location", "region_id") if !exclude_deprecated && staff_delegate?
 
     options = default_options.merge(options || {}).deep_dup
+
+    if exclude_deprecated
+      options[:methods].delete("delegate_status")
+      options[:include].delete("teams")
+    end
+
     # Preempt the values for avatar and teams, they have a special treatment.
     include_avatar = options[:include]&.delete("avatar")
     include_teams = options[:include]&.delete("teams")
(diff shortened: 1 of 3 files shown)


app/models/user.rb
```diff
@@ -1251,10 +1251,21 @@ def serializable_hash(options = nil)
     # NOTE: doing deep_dup is necessary here to avoid changing the inner values
     # of the freezed variables (which would leak PII)!
     default_options = DEFAULT_SERIALIZE_OPTIONS.deep_dup
+
+    include_email = options&.[](:include_email)
+    exclude_deprecated = options&.[](:exclude_deprecated)
```
gregorbg: Hm, I'm not a big fan of this way of nil-safe calls to the `[]` method.
AI suggests the following alternatives:
- `include_email, exclude_deprecated = options&.values_at(:include_email, :exclude_deprecated)`
- `include_email = options&.dig(:include_email)`
- `include_email = options&.fetch(:include_email, false)`

app/models/user.rb
```diff
@@ -1251,10 +1251,21 @@ def serializable_hash(options = nil)
...
-    default_options[:methods].push("email", "location", "region_id") if staff_delegate?
+    default_options[:methods].push("email") if include_email || staff_delegate?
+    default_options[:methods].push("location", "region_id") if !exclude_deprecated && staff_delegate?
 
     options = default_options.merge(options || {}).deep_dup
+
+    if exclude_deprecated
+      options[:methods].delete("delegate_status")
+      options[:include].delete("teams")
+    end
```
gregorbg: This feels a bit redundant with your other code change above, because you're only adding `delegate_status` and `teams` if the condition `if !exclude_deprecated` is already met. Why are you running this here again?

gregorbg (changes requested):

app/models/user.rb
danieljames-dj: Okay I've cleaned it up a bit.

app/models/user.rb
danieljames-dj: First option looks good. Applied

app/models/user.rb
```diff
@@ -1243,18 +1243,27 @@ def url
...
-    # Delegates's emails and regions are public information.
-    default_options[:methods].push("email", "location", "region_id") if staff_delegate?
+
+    include_email, exclude_deprecated = options&.values_at(:include_email, :exclude_deprecated)
+
+    unless exclude_deprecated
+      default_options[:methods].push("location", "region_id") if staff_delegate?
+      default_options[:methods].push("delegate_status")
+      default_options[:include].push("teams")
+
```
gregorbg: extraneous blank line. I'm surprised that our linter did not beat you over the head with this

danieljames-dj: Oops, thanks for catching, removed.

PR #12045, opened 2025-07-21
Migrate results preview page to react

danieljames-dj:
Couple of things to note:
- Two fields like `expand_all` and `unindent_solves` are not used anymore because now I'm expanding all by default, and in the two views where this is being used, it's fine. Same for `unindent_solves`.
- `inbox_result` is probably used only in this component so I assume the changes in `inbox_result.rb` is fine.

Other than these, this is straightforward react migration.


app/models/inbox_result.rb
```diff
@@ -15,4 +17,20 @@ def person
   def person_name
     inbox_person&.name || "<person_id=#{person_id}>"
   end
+
+  def name
+    person_name
+  end
```
gregorbg:
```suggestion
  alias_method :name, :person_name
```

app/webpacker/components/ResultsData/Results/ResultRow.js
```diff
@@ -30,7 +30,9 @@ function ResultRow({
         )}
       </Table.Cell>
       <Table.Cell>
-        <a href={personUrl(result.wca_id)}>{`${result.name}`}</a>
+        {result.wca_id
+          ? <a href={personUrl(result.wca_id)}>{`${result.name}`}</a>
+          : `${result.name}`}
```
gregorbg:
```suggestion
          : result.name}
```

gregorbg: No need to wrap this variable into a string, or am I wrong?

app/webpacker/components/ResultsData/Results/ResultRow.js
```diff
@@ -30,7 +30,9 @@ function ResultRow({
         )}
       </Table.Cell>
       <Table.Cell>
-        <a href={personUrl(result.wca_id)}>{`${result.name}`}</a>
+        {result.wca_id
+          ? <a href={personUrl(result.wca_id)}>{`${result.name}`}</a>
```
gregorbg:
```suggestion
          ? <a href={personUrl(result.wca_id)}>{result.name}</a>
```

gregorbg: Also, no need to wrap this into a string IMO

app/models/inbox_result.rb
```diff
@@ -5,6 +5,8 @@ class InboxResult < ApplicationRecord
 
   # see result.rb for explanation of the scope
   belongs_to :inbox_person, ->(ibr) { where(competition_id: ibr.competition_id) }, primary_key: :id, foreign_key: :person_id, optional: true
+  delegate :country_iso2, to: :person
+  delegate :wca_id, to: :person
```
gregorbg: These two methods are now becoming a part of `DEFAULT_SERIALIZE_OPTIONS`. Which means they will be called on every single InboxResult.

The comment on `def person` implies that loading the person is very cost-prohibitive, so it looks to me like doing this now will create an incredible amount of SQL queries :(

gregorbg: This is part of the reason why the old Ruby templating code uses `persons_by_id` as a lookup, instead of actually accessing each person on the individual result row during iteration.
You will have to come up with a similar solution to reduce the database load.

gregorbg (changes requested):

app/models/inbox_result.rb
danieljames-dj: Oh okay, I did few optimizations now. Hope it's better.

app/models/inbox_result.rb
danieljames-dj: Done

app/webpacker/components/ResultsData/Results/ResultRow.js
danieljames-dj: Okay removed.

app/webpacker/components/ResultsData/Results/ResultRow.js
danieljames-dj: Okay removed.

app/controllers/tickets_controller.rb
```diff
@@ -189,4 +189,10 @@ def anonymize
       new_wca_id: person&.reload&.wca_id, # Reload to get the new wca_id
     }
   end
+
+  def imported_temporary_results
+    competition = Competition.find(params.require(:competition_id))
+
+    render json: competition.inbox_results.includes(:inbox_person)
```
gregorbg, line 196: Have you actually tried out whether this works? The scope `inbox_person` depends on a composite key defined with a `-> () { where() }` style block, Rails usually isn't able to preload these on the fly

gregorbg: For reference: 

Image

gregorbg: Surprisingly, this works. No idea what kind of lazy-loading Rails internal black magic is behind this, but I'll take it as a win. Kudos for figuring that out!

gregorbg (approved):

app/controllers/tickets_controller.rb
danieljames-dj: Thanks Gregor for checking this.

PR #12049, opened 2025-07-22
Use composite keys for InboxPerson lookup

gregorbg:
Small optimization based on #12045.

I have compared the queries generated by Rails manually in the console using individual rows, but let's see whether everything works as expected in the CI as well.

The main advantage is that you can now do stuff like my_competition.inbox_results.join(:inbox_person), which was not possible before. However, this now means that you have to pay attention to accessing the ID directly from an AR model, which would return an array. See `def numeric_id` in InboxPerson for details.

(Remake of #12047)


app/models/inbox_person.rb
```diff
@@ -4,6 +4,10 @@ class InboxPerson < ApplicationRecord
   # for some reason, the ActiveRecord plural for "Person" is "people"…
   self.table_name = 'inbox_persons'
 
+  # For historic reasons, we insert the registrant ID as the SQL table `id` column.
```
danieljames-dj, line 7: I understand that this registrant ID is different from the actual registrant ID (that we use in WCIF). Is this ID named as registrantID anywhere? If not can we name it as personId or something else to avoid confusing with the actual registrantId?

danieljames-dj (approved):
This is great, thanks @gregorbg.

app/models/inbox_person.rb
gregorbg: Please correct me if I'm wrong, but I believe that it is actually the registrantID from WCIF

gregorbg: Yes, confirmed that we are actually using the `registrantId` from WCIF. Scrambles-Matcher uses it to put that into the `id` field of the results JSON we are importing: https://github.com/thewca/scrambles-matcher/blob/7bf1013ef426e1ef5f770a0ffa155bf8ba2241c7/src/logic/import-export-wcif.js#L32-L33

danieljames-dj: Oh my bad. I was under the assumption that `registrantId` is the primary key of `registrations` table. 🤦  Thanks for clarifying.

PR #12050, opened 2025-07-22
Use DocumentDB compatibility layer for Payload

gregorbg:
See https://payloadcms.com/docs/database/mongodb#using-other-mongodb-implementations, recently introduced in Payload 3.48 via https://github.com/thewca/worldcubeassociation.org/pull/12038


gregorbg:
Note for reviewers: The name `compatabilityOptions` actually contains a typo in the original code of Payload

PR #12051, opened 2025-07-22
Few optimizations for roles and groups

danieljames-dj:
This is follow-up PR to https://github.com/thewca/worldcubeassociation.org/pull/12043.

Now, the roles are loading with very less queries (except avatars - for avatars still it's doing one call for each user)

EDIT: Not just avatars, actually metadata is also loading. I tried adding in `includes` but since it's polymorphic association, it cannot do eager loading. The API is somehow doing lazy loading, but unit tests are trying to do eager loading.

### DIFF
--- a/app/controllers/api/v0/user_groups_controller.rb
+++ b/app/controllers/api/v0/user_groups_controller.rb
@@ -40,7 +40,7 @@ class Api::V0::UserGroupsController < Api::V0::ApiController
     # Sorts the list of groups by name.
     groups = groups.sort_by(&:name)
 
-    render json: groups
+    render json: groups, methods: %w[lead_user]
   end
 
   def create
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1248,7 +1248,7 @@ class User < ApplicationRecord
     only: %w[id wca_id name gender
              country_iso2 created_at updated_at],
     methods: %w[url country],
-    include: %w[avatar teams],
+    include: %w[avatar],
   }.freeze
 
   def serializable_hash(options = nil)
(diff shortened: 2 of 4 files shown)


app/models/user.rb
```diff
@@ -1248,7 +1248,7 @@ def url
     only: %w[id wca_id name gender
              country_iso2 created_at updated_at],
     methods: %w[url country],
-    include: %w[avatar teams],
```
gregorbg, line 1251: Is this safe, regarding the default serialization in the public "show user" API?

app/controllers/api/v0/user_groups_controller.rb
```diff
@@ -40,7 +40,7 @@ def index
     # Sorts the list of groups by name.
     groups = groups.sort_by(&:name)
 
-    render json: groups
+    render json: groups, methods: %w[lead_user]
```
gregorbg, line 43: Just curious: How does this specific optimization work?
I saw that `lead_user` was already part of the default serialization options of UserGroup before. So, it was already serialized before this change. Now you're just moving the instruction "Please serialize the lead_user" to a different place in the code, but in the end, the final output of the serialization would be the same as it was before this PR.
Why is this making things more efficient?

danieljames-dj: So this `lead_user` field is required in groups API, but not in roles API. With this change, the `lead_user` will still load in groups API, but will stop loading in roles API.

app/models/user.rb
danieljames-dj: Actually it's being added back below. This is something I should have done along with the previous PR but somehow missed it.

gregorbg (approved):

PR #12054, opened 2025-07-23
Add option to merge in duplicate person dashboard

danieljames-dj:
This feature will allow Delegates to merge the potential duplicates from the newcomers check page

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -77,16 +77,7 @@ class UsersController < ApplicationController
 
     return render status: :bad_request, json: { error: "Cannot merge users with both having a WCA ID" } if to_user.wca_id.present? && from_user.wca_id.present?
 
-    ActiveRecord::Base.transaction do
-      from_user.transfer_data_to(to_user)
-      # After this merge, there won't be any registrations for from_user as all
-      # of them will be transferred to to_user. So any potential duplicates of
-      # from_user is no longer valid. There might be some potential duplicates for
-      # to_user, but they need to be refetched. But refetching here may look
-      # confusing, so removing the potential duplicates of to_user as well.
-      PotentialDuplicatePerson.where(original_user_id: from_user.id).delete_all
-      PotentialDuplicatePerson.where(original_user_id: to_user.id).delete_all
-    end
+    from_user.transfer_data_to(to_user)
 
     render status: :ok, json: { success: true }
   end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -77,7 +77,16 @@ class UsersController < ApplicationController
 
     return render status: :bad_request, json: { error: "Cannot merge users with both having a WCA ID" } if to_user.wca_id.present? && from_user.wca_id.present?
 
-    from_user.transfer_data_to(to_user)
+    ActiveRecord::Base.transaction do
+      from_user.transfer_data_to(to_user)
+      # After this merge, there won't be any registrations for from_user as all
+      # of them will be transferred to to_user. So any potential duplicates of
+      # from_user is no longer valid. There might be some potential duplicates for
+      # to_user, but they need to be refetched. But refetching here may look
+      # confusing, so removing the potential duplicates of to_user as well.
+      PotentialDuplicatePerson.where(original_user_id: from_user.id).delete_all
+      PotentialDuplicatePerson.where(original_user_id: to_user.id).delete_all
+    end
 
     render status: :ok, json: { success: true }
   end
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -74,6 +74,8 @@ class Person < ApplicationRecord
     user.save! if user # User copies data from the person before validation, so this will update him.
   end
 
+  delegate :id, to: :user, prefix: true, allow_nil: true
+
   def update_using_sub_id!(attributes)
     raise unless update_using_sub_id(attributes)
   end
@@ -367,10 +369,6 @@ class Person < ApplicationRecord
     else
       []
     end
-  end
-
-  def user_id
-    user&.id
   end
 
   def serializable_hash(options = nil)
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -369,6 +369,10 @@ class Person < ApplicationRecord
     end
   end
 
+  def user_id
+    user&.id
+  end
+
   def serializable_hash(options = nil)
     json = super(DEFAULT_SERIALIZE_OPTIONS.merge(options || {}))
     json[:class] = self.class.to_s.downcase
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -59,6 +59,7 @@ class User < ApplicationRecord
   belongs_to :current_avatar, class_name: "UserAvatar", inverse_of: :current_user, optional: true
   belongs_to :pending_avatar, class_name: "UserAvatar", inverse_of: :pending_user, optional: true
   has_many :user_avatars, dependent: :destroy, inverse_of: :user
+  has_many :potential_duplicate_persons, dependent: :destroy, foreign_key: :original_user_id, class_name: "PotentialDuplicatePerson"
 
   scope :confirmed_email, -> { where.not(confirmed_at: nil) }
   scope :newcomers, -> { where(wca_id: nil) }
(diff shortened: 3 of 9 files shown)


app/models/person.rb
```diff
@@ -369,6 +369,10 @@ def private_attributes_for_user(user)
     end
   end
 
+  def user_id
+    user&.id
+  end
+
```
gregorbg: You can just use syntactic sugar `delegate :id, to: :user, prefix: true, allow_nil: true`

app/controllers/users_controller.rb
```diff
@@ -77,7 +77,16 @@ def merge
...
-    from_user.transfer_data_to(to_user)
+    ActiveRecord::Base.transaction do
+      from_user.transfer_data_to(to_user)
+      # After this merge, there won't be any registrations for from_user as all
+      # of them will be transferred to to_user. So any potential duplicates of
+      # from_user is no longer valid. There might be some potential duplicates for
+      # to_user, but they need to be refetched. But refetching here may look
+      # confusing, so removing the potential duplicates of to_user as well.
+      PotentialDuplicatePerson.where(original_user_id: from_user.id).delete_all
+      PotentialDuplicatePerson.where(original_user_id: to_user.id).delete_all
```
gregorbg: You should define a `has_many :potential_duplicate_persons` on the `User` model, which will allow you to just use `from_user.potential_duplicate_persons.delete_all`.

gregorbg: If you go with this approach, make sure to specify a suitable `dependent:` strategy on the `has_many` association, because I think Rails's default is to leave the rows in place and just `NULL`ify their foreign keys.

gregorbg: Sidenote: If you have a relation in `user.rb`, then you might just execute these two new calls as part of the `transfer_data_to` method body...

app/models/user.rb
```diff
@@ -1529,19 +1529,17 @@ def anonymize(new_wca_id = nil)
   end
 
   def transfer_data_to(new_user)
-    ActiveRecord::Base.transaction do
```
gregorbg: You can keep the transaction in here, if my other suggestion above turns out to work well

app/webpacker/components/NewcomerChecks/DuplicateChecker.jsx
```diff
@@ -34,6 +36,10 @@ function DuplicateChecker({ competitionId }) {
     },
   });
 
+  const [mergeModalOptions, setMergeModalOptions] = useState({
+    isOpen: false,
+  });
```
gregorbg: Is there a specific reason why you decided to squeeze `isOpen` and `potentialDuplicatePerson` into one object, instead of just keeping them as two separate states?

gregorbg: Also, you could consider just inferring `isOpen` with a `useMemo` that checks whether `potentialDuplicatePerson` is null or not.
If you do that, you don't have any `setIsOpen` any more, and you need to "simulate" closing by setting the duplicatePerson to `null` or `undefined`

app/webpacker/components/NewcomerChecks/MergeModal.jsx
```diff
@@ -0,0 +1,40 @@
...
+    queryClient.setQueryData(
+      ['last-duplicate-checker-job', competitionId],
+      (previousData) => ({
+        ...previousData,
+        potential_duplicate_persons: previousData.potential_duplicate_persons.filter(
+          (person) => person.original_user_id !== originalUser.id,
+        ),
+      }),
+    );
+  };
```
gregorbg: This onSuccess hook is making me reconsider the design of the original mutation (that I had originally approved) a little bit. I cannot comment on the lines of the actual `useMutation` declaration, because GitHub does not consider them "part of this PR", so I need to explain here.

Mutations should be as self-contained as possible. As a rule of thumb, you should be (hypothetically) able to move mutations to a separate `lib` or `utils` file at any moment, without anything breaking.

In your current version, it would break, because your mutation is accessing some `useState` defined outside of the mutation.

Instead, I suggest that you pass in `fromUserId` and `toUserId` as parameters to the mutation (note, you cannot do `mutationFn(fromUserId, toUserId)` because the `mutationFn` is designed to only take exactly 1 parameter, so you have to pass it as `mutationFn({ fromUserId, toUserId })` object!) and let the mutation just make the network call.

The key advantage of this is that the `onSuccess` hook gets passed the same parameters as the original `mutationFn` in its second parameter: The actual signature is something like `onSuccess(mutationFnReturnValue, optionsPassedToMutationFn)`.

You can then leverage this second parameter by reading the `originalUser.id` not from the outer component state, but from the `onSuccess` parameters. And if you do that, you can just move the whole `onSuccess` down into the `useMutation` declaration, and you don't have to "pass it along" through the component render cycle.

gregorbg: Further reading: https://tanstack.com/query/latest/docs/framework/react/guides/mutations#mutation-side-effects

gregorbg (changes requested):

app/webpacker/components/NewcomerChecks/MergeModal.jsx
danieljames-dj: Okay thanks Gregor for detailed explanation. I have made the change, but I still need to get back onSuccess because I'll be needing `competition_id` for the queryKey.

app/webpacker/components/NewcomerChecks/DuplicateChecker.jsx
danieljames-dj: Okay makes sense. Maybe `isOpen` is not needed at all - I can directly use `potentialDuplicatePerson`.

app/models/person.rb
danieljames-dj: Done

app/controllers/users_controller.rb
danieljames-dj: Okay done

app/models/user.rb
danieljames-dj: Ya removed

app/webpacker/components/NewcomerChecks/MergeModal.jsx
```diff
@@ -0,0 +1,40 @@
...
+import MergeUsers from '../Panel/pages/MergeUsersPage/MergeUsers';
+
+export default function MergeModal({ potentialDuplicatePerson, competitionId }) {
+  const queryClient = useQueryClient();
+  const {
+    original_user: originalUser,
+    duplicate_person: duplicatePerson,
+  } = potentialDuplicatePerson;
+
+  const onSuccess = (_, { _fromUserId, _toUserId }) => {
```
gregorbg: Why are there underscores in front of `fromUserId` and `toUserId`?

app/webpacker/components/Panel/pages/MergeUsersPage/MergeUsers.jsx
```diff
@@ -40,12 +43,8 @@ export default function MergeUsers({ firstUserId, secondUserId }) {
     error: mergeError,
     isSuccess,
   } = useMutation({
-    mutationFn: () => {
-      const fromUserId = (
-        firstUserId === toUserId ? secondUserId : firstUserId
-      );
-      return mergeUsers(toUserId, fromUserId);
-    },
+    mutationFn: ({ _fromUserId, _toUserId }) => mergeUsers(_toUserId, _fromUserId),
```
gregorbg: Again, why the underscores?

gregorbg (changes requested):

app/webpacker/components/NewcomerChecks/MergeModal.jsx
danieljames-dj: Changed it, I was already having `fromUserId` field in the MergeUsers, so wanted to use different name and hence used _. Now managed to find a prefix to the parameter name.

app/webpacker/components/Panel/pages/MergeUsersPage/MergeUsers.jsx
danieljames-dj: Changed it, I had it initially because there was already another `fromUserId`, managed to rename it in a better way now.

gregorbg (approved):

PR #12073, opened 2025-07-24
Create tickets once the competition ends

danieljames-dj:
Currently ticket for results are created only when the Delegate submits results. Now it's changed to a bit early stage. Delegate will be asked to click a button asking to 'start result submission process', and that point the ticket will be created.

With this, now we can move the results submission process of Delegates to the ticket (as a beta feature).

### DIFF
--- a/app/controllers/results_submission_controller.rb
+++ b/app/controllers/results_submission_controller.rb
@@ -149,16 +149,41 @@ class ResultsSubmissionController < ApplicationController
 
     return render status: :unprocessable_entity, json: { error: "Submitted results contain errors." } if results_validator.any_errors?
 
-    return render status: :unprocessable_entity, json: { error: "There is already a ticket associated, please contact WRT to update this." } if competition.result_ticket.present?
+    return render status: :unprocessable_entity, json: { error: "There is no ticket associated, please contact WST." } if competition.result_ticket.nil?
 
     CompetitionsMailer.results_submitted(competition, results_validator, message, current_user).deliver_now
 
     ActiveRecord::Base.transaction do
       competition.touch(:results_submitted_at)
-      TicketsCompetitionResult.create_ticket!(competition, message, current_user)
+      competition.tickets_competition_result.update!(
+        delegate_message: message,
+        status: TicketsCompetitionResult.statuses[:submitted],
+      )
+      competition.tickets_competition_result.competition_stakeholder.update!(connection: TicketStakeholder.connections[:cc])
+      TicketStakeholder.create!(
+        ticket_id: competition.result_ticket.id,
+        stakeholder: UserGroup.teams_committees_group_wrt,
+        connection: TicketStakeholder.connections[:assigned],
+        stakeholder_role: TicketStakeholder.stakeholder_roles[:actioner],
+        is_active: true,
+      )
     end
 
     render status: :ok, json: { success: true }
+  end
+
+  def start_result_submission_process
+    @competition = competition_from_params
+
+    if @competition.tickets_competition_result.present?
+      return render status: :bad_request, json: {
+        error: "Ticket for result submission already exists.",
+      }
+    end
+
+    TicketsCompetitionResult.create_ticket!(@competition, current_user.id)
+
+    redirect_to competition_submit_results_edit_path
   end
 
   private def competition_from_params
--- a/app/controllers/results_submission_controller.rb
+++ b/app/controllers/results_submission_controller.rb
@@ -149,7 +149,7 @@ class ResultsSubmissionController < ApplicationController
 
     return render status: :unprocessable_entity, json: { error: "Submitted results contain errors." } if results_validator.any_errors?
 
-    return render status: :unprocessable_entity, json: { error: "There is no ticket associated, please contact WST." } if competition.result_ticket.nil?
+    return render status: :not_found, json: { error: "There is no ticket associated, please contact WST." } if competition.result_ticket.nil?
 
     CompetitionsMailer.results_submitted(competition, results_validator, message, current_user).deliver_now
 
@@ -177,7 +177,7 @@ class ResultsSubmissionController < ApplicationController
 
     if @competition.tickets_competition_result.present?
       return render status: :bad_request, json: {
-        error: "Ticket for result submission already exists.",
+        error: "Ticket for result submission already exists. Please refresh your page.",
       }
     end
 
--- a/app/models/ticket.rb
+++ b/app/models/ticket.rb
@@ -29,9 +29,20 @@ class Ticket < ApplicationRecord
   def can_user_access?(user)
     return false if user.nil?
 
-    ticket_stakeholders.belongs_to_user(user).any? ||
-      ticket_stakeholders.belongs_to_groups(user.active_groups).any? ||
-      ticket_stakeholders.belongs_to_competitions(user.delegated_competitions).any?
+    stakeholder_types = ticket_stakeholders.distinct.pluck(:stakeholder_type)
+
+    stakeholder_types.any? do |stakeholder_type|
+      case stakeholder_type
+      when "User"
+        ticket_stakeholders.belongs_to_user(user).any?
+      when "UserGroup"
+        ticket_stakeholders.belongs_to_groups(user.active_groups).any?
+      when "Competition"
+        ticket_stakeholders.belongs_to_competitions(user.delegated_competitions).any?
+      else
+        false
+      end
+    end
   end
 
   DEFAULT_SERIALIZE_OPTIONS = {
--- a/app/models/ticket.rb
+++ b/app/models/ticket.rb
@@ -30,7 +30,8 @@ class Ticket < ApplicationRecord
     return false if user.nil?
 
     ticket_stakeholders.belongs_to_user(user).any? ||
-      ticket_stakeholders.belongs_to_groups(user.active_groups).any?
+      ticket_stakeholders.belongs_to_groups(user.active_groups).any? ||
+      ticket_stakeholders.belongs_to_competitions(user.delegated_competitions).any?
   end
 
   DEFAULT_SERIALIZE_OPTIONS = {
--- a/app/models/tickets_competition_result.rb
+++ b/app/models/tickets_competition_result.rb
@@ -4,6 +4,7 @@ class TicketsCompetitionResult < ApplicationRecord
   self.table_name = "tickets_competition_result"
 
   enum :status, {
+    pending_import: "pending_import",
     submitted: "submitted",
     locked_for_posting: "locked_for_posting",
     warnings_verified: "warnings_verified",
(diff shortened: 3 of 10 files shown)


app/controllers/results_submission_controller.rb
```diff
@@ -149,18 +149,43 @@ def create
 
     return render status: :unprocessable_entity, json: { error: "Submitted results contain errors." } if results_validator.any_errors?
 
-    return render status: :unprocessable_entity, json: { error: "There is already a ticket associated, please contact WRT to update this." } if competition.result_ticket.present?
+    return render status: :unprocessable_entity, json: { error: "There is no ticket associated, please contact WST." } if competition.result_ticket.nil?
```
gregorbg: Are you sure this is a good, backwards-compatible way of handling things? I think we should deploy the feature first, and give it some "transition time", before beginning to assume that "Oh there is no ticket so something must be horribly wrong"

gregorbg: Unrelated: Now that you have changed the logic behind this error, the status should be `not_found` (404) instead of `unprocessable_entity` (422)

app/controllers/results_submission_controller.rb
```diff
@@ -149,18 +149,43 @@ def create
...
     end
 
     render status: :ok, json: { success: true }
   end
 
+  def start_result_submission_process
+    @competition = competition_from_params
+
+    if @competition.tickets_competition_result.present?
+      return render status: :bad_request, json: {
```
gregorbg, line 179: In the previous code before you raised this PR, you had used `unprocessable_entity` when the ticket was already present, along with a "please contact WRT". What caused you to change your approach by using a different status code and different message content?

app/models/tickets_competition_result.rb
```diff
@@ -23,37 +24,33 @@ def actions_allowed_for(ticket_stakeholder)
...
-    TicketLog.create!(
-      ticket_id: ticket.id,
-      action_type: TicketLog.action_types[:update_status],
-      action_value: ticket_metadata.status,
-      acting_user_id: submitted_delegate.id,
-      acting_stakeholder_id: competition_stakeholder.id,
-    )
+  def self.create_ticket!(competition, user_id)
+    ActiveRecord::Base.transaction do
+      ticket_metadata = TicketsCompetitionResult.create!(
```
gregorbg:
```suggestion
      ticket_metadata = self.create!(
```

app/models/tickets_competition_result.rb
```diff
@@ -4,6 +4,7 @@ class TicketsCompetitionResult < ApplicationRecord
   self.table_name = "tickets_competition_result"
 
   enum :status, {
+    pending_import: "pending_import",
```
gregorbg, line 7: You could consider a `default: :pending_import` on the `enum :status` definition now

app/models/tickets_competition_result.rb
```diff
@@ -23,37 +24,33 @@ def actions_allowed_for(ticket_stakeholder)
...
-      ticket_id: ticket.id,
-      action_type: TicketLog.action_types[:update_status],
-      action_value: ticket_metadata.status,
-      acting_user_id: submitted_delegate.id,
-      acting_stakeholder_id: competition_stakeholder.id,
-    )
+  def self.create_ticket!(competition, user_id)
+    ActiveRecord::Base.transaction do
+      ticket_metadata = TicketsCompetitionResult.create!(
+        status: TicketsCompetitionResult.statuses[:pending_import],
```
gregorbg: This would become redundant if you go with the `default` suggestion on the enum above

app/models/tickets_competition_result.rb
```diff
@@ -23,37 +24,33 @@ def actions_allowed_for(ticket_stakeholder)
...
+        competition_id: competition.id,
```
gregorbg: If you have the full competition model, you can link it directly by just doing `competition: competition`. Not much different in terms of the end result that lands in the database, but in Rails-world you've already "pre-warmed" the association that way.

gregorbg: (that means, `self.create!(competition_id: competition.id).competition.name` will fire another database call to retrieve the competition, whereas `self.create!(competition: competition).competition.name` will not)

app/models/tickets_competition_result.rb
```diff
@@ -23,37 +24,33 @@ def actions_allowed_for(ticket_stakeholder)
...
+
+      ticket = Ticket.create!(metadata: ticket_metadata)
+
+      competition_stakeholder = TicketStakeholder.create!(
+        ticket_id: ticket.id,
+        stakeholder: competition,
+        connection: TicketStakeholder.connections[:assigned],
+        stakeholder_role: TicketStakeholder.stakeholder_roles[:requester],
+        is_active: true,
+      )
```
gregorbg:
```suggestion
      competition_stakeholder = ticket.ticket_stakeholders.create!(
        stakeholder: competition,
        connection: TicketStakeholder.connections[:assigned],
        stakeholder_role: TicketStakeholder.stakeholder_roles[:requester],
        is_active: true,
      )
```

app/models/tickets_competition_result.rb
```diff
@@ -23,37 +24,33 @@ def actions_allowed_for(ticket_stakeholder)
...
+
+      TicketLog.create!(
+        ticket_id: ticket.id,
+        action_type: TicketLog.action_types[:create_ticket],
+        acting_user_id: user_id,
+        acting_stakeholder_id: competition_stakeholder.id,
+      )
```
gregorbg:
```suggestion
      ticket.ticket_logs.create!(
        action_type: TicketLog.action_types[:create_ticket],
        acting_user_id: user_id,
        acting_stakeholder_id: competition_stakeholder.id,
      )
```

db/migrate/20250724025857_make_results_ticket_delegate_message_nullable.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class MakeResultsTicketDelegateMessageNullable < ActiveRecord::Migration[7.2]
+  def change
+    change_column_null :tickets_competition_result, :delegate_message, true
```
gregorbg, line 5: Why are you allowing an empty delegate message now?

app/views/results_submission/new.html.erb
```diff
@@ -6,6 +6,14 @@
...
   <% if @competition.results_submitted? %>
     <p>The results have already been submitted. If you have any more questions or comments please reply to the email sent with the first results submission.</p>
+  <% elsif @competition.tickets_competition_result.nil? %>
+    <%= button_to competition_start_result_submission_process_path,
+                  title: "Start result submission process",
+                  class: "btn btn-primary",
+                  method: :post do
+    %>
+      Start result submission process
+    <% end %>
```
gregorbg, line 16: Didn't you migrate this page to React already?!

app/models/ticket.rb
```diff
@@ -30,7 +30,8 @@ def can_user_access?(user)
     return false if user.nil?
 
     ticket_stakeholders.belongs_to_user(user).any? ||
-      ticket_stakeholders.belongs_to_groups(user.active_groups).any?
+      ticket_stakeholders.belongs_to_groups(user.active_groups).any? ||
+      ticket_stakeholders.belongs_to_competitions(user.delegated_competitions).any?
```
gregorbg: In the long run, we will need a more efficient strategy. This method will keep growing as you introduce more and more stakeholder types.

I think you should first get all (dinstinct) `stakeholder_type` values that apply to this ticket, and then run only the checks relevant for those.

app/models/tickets_competition_result.rb
```diff
@@ -23,37 +24,33 @@ def actions_allowed_for(ticket_stakeholder)
...
+    end
+  end
+
+  def competition_stakeholder
+    ticket.ticket_stakeholders.find_by(stakeholder: competition)
```
gregorbg: I am 70% sure this should be possible via a relation, something along the lines of...
- `has_many :ticket_stakeholders, through: :ticket`
- `has_one :competition_stakeholder, through: :ticket_stakeholders, source: :stakeholder, source_type: 'Competition', foreign_key: :competition_id`

gregorbg: Oh wait, I think my above suggestion will point you to the competition, not to the actual stakeholder entity. So the solution potentially is even simpler:
`has_one :competition_stakeholder, -> { where(stakeholder_type: "Competition") }, through: :ticket, source: :ticket_stakeholders, primary_key: :stakeholder_id, foreign_key: :competition_id`

app/models/tickets_competition_result.rb
danieljames-dj: Okay changed.

app/models/tickets_competition_result.rb
danieljames-dj: Okay removed

app/controllers/results_submission_controller.rb
danieljames-dj: > I think we should deploy the feature first, and give it some "transition time"

Once this is deployed, then users won't be able to see the results-submission page till they click the 'start results submission' process. They will be able to see the results-submission page only if there is a ticket associated with it. This error is just a double-checking, and won't have any impact.

> Unrelated: Now that you have changed the logic [...]

Done

app/controllers/results_submission_controller.rb
danieljames-dj: So earlier, the tickets were getting created during submission. If there is a re-submission (maybe WRT removes existing results on Delegate's request), it tries to create the ticket again which should fail and WRT might be able to help by manually marking the results as submitted after importing temporary results.

Now, the ticket will already be created. This is the action for the button to create the ticket and the button will be visible only if the ticket is not there. So in case this request comes when there is already a ticket, it means the user hasn't refreshed page for sometime and during that time another user created the ticket. So ideally a refresh will definitely fix their issue. I've updated the text asking the user to refresh.

app/models/ticket.rb
danieljames-dj: Not sure I understood what you meant, anyway I've changed based on my guess. Is it correct?

app/models/tickets_competition_result.rb
danieljames-dj: Done

app/models/tickets_competition_result.rb
danieljames-dj: Okay changed.

app/models/tickets_competition_result.rb
danieljames-dj: Changed

app/models/tickets_competition_result.rb
danieljames-dj: Changed

app/views/results_submission/new.html.erb
danieljames-dj: Not yet, only some parts of this page is migrated.

db/migrate/20250724025857_make_results_ticket_delegate_message_nullable.rb
danieljames-dj: Because while creating the ticket we won't have `delegate_message`. `delegate_message` will be created only during `submission` step.

app/models/tickets_competition_result.rb
danieljames-dj: Wow this is amazing, I didn't think that foreign key can be added here.

danieljames-dj:
Discussed with Gregor over Slack, and came to the conclusion that we can create the ticket only when the proper json is uploaded. So I'm closing this PR without merging.

PR #12074, opened 2025-07-25
Show name format and DOB issues in newcomer check page

danieljames-dj:
Usually Delegates will get name issues and DOB issues only during uploading the results JSON for first time. With this, they will get these details even before the competition.

### DIFF
--- a/app/controllers/results_submission_controller.rb
+++ b/app/controllers/results_submission_controller.rb
@@ -26,9 +26,9 @@ class ResultsSubmissionController < ApplicationController
   def newcomer_name_format_check
     competition = competition_from_params
 
-    name_format_issues = competition.accepted_newcomers.map do |user|
+    name_format_issues = competition.accepted_newcomers.flat_map do |user|
       ResultsValidators::PersonsValidator.name_validations(user.name, nil)
-    end.flatten
+    end
 
     render status: :ok, json: name_format_issues
   end
@@ -36,9 +36,9 @@ class ResultsSubmissionController < ApplicationController
   def newcomer_dob_check
     competition = competition_from_params
 
-    dob_issues = competition.accepted_newcomers.map do |user|
+    dob_issues = competition.accepted_newcomers.flat_map do |user|
       ResultsValidators::PersonsValidator.dob_validations(user.dob, nil, name: user.name)
-    end.flatten
+    end
 
     render status: :ok, json: dob_issues
   end
--- a/app/controllers/results_submission_controller.rb
+++ b/app/controllers/results_submission_controller.rb
@@ -4,8 +4,8 @@ require 'fileutils'
 
 class ResultsSubmissionController < ApplicationController
   before_action :authenticate_user!
-  before_action -> { redirect_to_root_unless_user(:can_upload_competition_results?, competition_from_params) }, except: %i[newcomer_checks last_duplicate_checker_job_run compute_potential_duplicates]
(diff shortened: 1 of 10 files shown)


app/controllers/results_submission_controller.rb
```diff
@@ -23,6 +23,28 @@ def last_duplicate_checker_job_run
     render status: :ok, json: last_job_run
   end
 
+  def newcomer_name_format_check
+    competition = competition_from_params
+
+    name_format_issues = []
+    competition.accepted_newcomers.each do |user|
+      name_format_issues += ResultsValidators::PersonsValidator.name_validations(user.name, nil)
+    end
```
gregorbg: Use `map` instead.

gregorbg: Unrelated: What is the `nil` argument at the end?

app/controllers/results_submission_controller.rb
```diff
@@ -23,6 +23,28 @@ def last_duplicate_checker_job_run
...
+    render status: :ok, json: name_format_issues
+  end
+
+  def newcomer_dob_check
+    competition = competition_from_params
+
+    dob_issues = []
+    competition.accepted_newcomers.each do |user|
+      dob_issues += ResultsValidators::PersonsValidator.dob_validations(user.dob, nil, name: user.name)
+    end
```
gregorbg: Use `map` instead

gregorbg: (same as above: extraneous `nil` argument)

app/controllers/results_submission_controller.rb
danieljames-dj: Oh ya map is better, changed.

Re `nil`, the method requires `competition_id` for the cases where the validators are being ran for multiple competitions. The `competition_id` is just to create the validation object and won't have any impact in this case. In case there were multiple competitions, then against each warnings, it can show the corresponding competition with this field.

app/controllers/results_submission_controller.rb
```diff
@@ -23,6 +23,26 @@ def last_duplicate_checker_job_run
     render status: :ok, json: last_job_run
   end
 
+  def newcomer_name_format_check
+    competition = competition_from_params
+
+    name_format_issues = competition.accepted_newcomers.map do |user|
+      ResultsValidators::PersonsValidator.name_validations(user.name, nil)
+    end.flatten
```
```suggestion
    name_format_issues = competition.accepted_newcomers.flat_map do |user|
      ResultsValidators::PersonsValidator.name_validations(user.name, nil)
    end
```

app/controllers/results_submission_controller.rb
```diff
@@ -23,6 +23,26 @@ def last_duplicate_checker_job_run
...
+
+    render status: :ok, json: name_format_issues
+  end
+
+  def newcomer_dob_check
+    competition = competition_from_params
+
+    dob_issues = competition.accepted_newcomers.map do |user|
+      ResultsValidators::PersonsValidator.dob_validations(user.dob, nil, name: user.name)
+    end.flatten
```
```suggestion
    dob_issues = competition.accepted_newcomers.flat_map do |user|
      ResultsValidators::PersonsValidator.dob_validations(user.dob, nil, name: user.name)
    end
```

gregorbg (changes requested):

app/controllers/results_submission_controller.rb
danieljames-dj: Accepted suggestion

app/controllers/results_submission_controller.rb
danieljames-dj: Accepted suggestion

gregorbg (approved):

PR #12075, opened 2025-07-25
Change in schema of ticket logs

danieljames-dj:
This is as per discussion with Gregor. Each log will be for one action, but one action can cause multiple changes. So, decided to move changes to a different table to have 1:N relation instead of 1:1 relation.

One drawback of this change is that the log no longer show to what status it is updated. But since it's internal only, currently it's not significant. I've some plans of improving the UI of logs completely, will be taking it up during that time.


app/controllers/tickets_controller.rb
```diff
@@ -78,13 +78,15 @@ def update_status
 
     ActiveRecord::Base.transaction do
       @ticket.metadata.update!(status: ticket_status)
-      TicketLog.create!(
-        ticket_id: @ticket.id,
+      ticket_log = @ticket.ticket_logs.create!(
         action_type: @action_type,
```
gregorbg, line 82: With this new design, is the enum `action_type` still relevant? I think the kind of action hugely depends on what kind of ticket you're dealing with (for example, a results posting ticket will have very different actions compared to a competition announcement ticket)

gregorbg: I definitely don't want the `action_type` enum to "blow up" over time and get ridiculously large, because every ticket type adds its actions to one central enum.

The easiest solution would just be to drop the column, but that means you lose information on the change.
Another compromise solution would be to augment the enum to have "status change, comment added, metadata changed" and then one new extra column that is **not** an enum, where specific tickets can write whatever the heck they want (or leave it null if it's a "general" action like changing status or updating comment)

app/controllers/tickets_controller.rb
```diff
@@ -78,13 +78,15 @@ def update_status
 
     ActiveRecord::Base.transaction do
       @ticket.metadata.update!(status: ticket_status)
```
gregorbg, line 80: Side question: Here you're assuming that every metadata ever will definitely have a `status` column. So why is that status not part of the original `ticket` table in the first place?

app/models/ticket_log_change.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class TicketLogChange < ApplicationRecord
+  enum :field_name, {
+    status: "status",
+  }
```
gregorbg: There will definitely be fields (for specific metadata of specific ticket flows like nationality change or competition announcement) that cannot be put into a general-purpose enum. It's the same overall thinking as my comments on the `action_type` above.

app/webpacker/components/Tickets/TicketLogs.jsx
```diff
@@ -12,14 +12,14 @@ export default function TicketLogs({ ticketId }) {
     queryFn: () => getLogs({ ticketId }),
   });
 
-  function logText(actionType, actionValue) {
+  function logText(actionType) {
```
gregorbg, line 15: Why are you getting rid of `actionValue`? You're still storing that information...

db/migrate/20250725050517_create_ticket_log_changes.rb
```diff
@@ -0,0 +1,29 @@
...
+class CreateTicketLogChanges < ActiveRecord::Migration[7.2]
+  def change
+    reversible do |dir|
+      dir.up do
+        create_table :ticket_log_changes do |t|
+          t.references :ticket_log, null: false
+          t.string :field_name, null: false
+          t.string :field_value, null: false
+          t.timestamps
+        end
```
gregorbg: `create_table` is inherently reversible and does not need to go into a `reversible` block.

db/migrate/20250725050517_create_ticket_log_changes.rb
```diff
@@ -0,0 +1,29 @@
...
+              field_name: TicketLogChange.field_names[:status],
+              field_value: ticket_log.action_value,
+            )
+          end
+        end
+      end
+
+      dir.down do
+        drop_table :ticket_log_changes
+      end
```
gregorbg: Unnecessary if you move the `create_table` out of the whole `reversible` block, per my comment above.

db/migrate/20250725060536_remove_action_value_from_ticket_logs.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class RemoveActionValueFromTicketLogs < ActiveRecord::Migration[7.2]
+  def change
+    reversible do |dir|
+      dir.up do
+        remove_column :ticket_logs, :action_value, :string
```
gregorbg: This should(?) also be reversible

app/controllers/tickets_controller.rb
danieljames-dj: In our ticket discussion, the reason for moving `status` to metadata table was that "various teams will have various workflows with wildly different statuses". But having `status` in the original `ticket` table is also fine I assume, but the code will be complicated to do validation of the field. Now since it's inside `metadata`, the validations are done by `enum`.

app/controllers/tickets_controller.rb
danieljames-dj: So my plan is to use this `action_type` only for ticket-specific things (like ticket created, comment added, status changed, etc) and not for metadata actions (like results submitted, results posted, etc).

For metadata actions, my plan (actually mostly your idea in an old discussion) is to have the `action_type` value as `metadata_action` and there will be `metadata_action` column to store the corresponding metadata action.

app/models/ticket_log_change.rb
danieljames-dj: Same comment here - the `field_name` will be `metadata`, and we can have another column `metadata_field` to say which metadata column of the ticket was changed.

(`status` is an exception - even though it's in `metadata`, we can consider it as ticket-specific)

app/controllers/tickets_controller.rb
danieljames-dj: Thinking again, do you think it's better to move `status` to the `tickets` table? Since tickets are in very early stage, it won't be much change. The complicated validation method will be a drawback, but I assume that is a good trade-off.

app/webpacker/components/Tickets/TicketLogs.jsx
danieljames-dj: This is because adding all sub-actions can be complicated. I prefer to do it along with other missing informations (in the log - like acting stakeholder, etc) using a good UI library (like [Feed](https://react.semantic-ui.com/views/feed/)) in the follow-up PR.

db/migrate/20250725060536_remove_action_value_from_ticket_logs.rb
danieljames-dj: Oh yes, you are right, changed.

db/migrate/20250725050517_create_ticket_log_changes.rb
danieljames-dj: Okay, agreed. Changed.

db/migrate/20250725050517_create_ticket_log_changes.rb
danieljames-dj: Okay changed.

app/controllers/tickets_controller.rb
gregorbg: First of all: Such a design decision is definitely out of the scope of this PR.
Sorry for opening the discussion here 😁 

Second: Well it depends on what you're trying to achieve with the `status` column. I think that every ticket, no matter what type or what metadata, always has a "common status", like "opened", "in_progress", "closed", or similar (you don't have to use these exact names, they are just to show my general idea).

The metadata status can be more "specific", depending on the use-case, so you definitely should avoid having to do extremely lengthy validations in the common `Ticket` model.

You can have the best of both worlds: Use a general status in the `tickets` table, and _also_ allow assigning a more specific `status` in the metadata. But here's the point: The metadata one doesn't need to be called `status`, it could be called `step` for example. (Or any other name that you think makes sense)

In the upload method, you pass in a `status` which you can be sure always exists on the base ticket (it's the common status explained above) and you can also pass in an arbitrary set of "metadata params" under `metadata` key in the POST payload. This "metadata params" may contain `status` or `step` or really anything else that the frontend deems "worthy" to be included in the status update

app/controllers/tickets_controller.rb
gregorbg: See my other explanation above for the `status` move.

I agree with your (and my, LOL 😂 ) idea of using the `action_type` for general descriptions, and having a specific `metadata_action` type for detail changes.

app/models/ticket_log_change.rb
gregorbg: Agreed, based on my other comments above.

db/migrate/20250725050517_create_ticket_log_changes.rb
```diff
@@ -0,0 +1,25 @@
...
+      t.references :ticket_log, null: false
+      t.string :field_name, null: false
+      t.string :field_value, null: false
+      t.timestamps
+    end
+
+    reversible do |dir|
+      dir.up do
+        TicketLog.find_each do |ticket_log|
+          if ticket_log.update_status?
```
gregorbg: Reading this makes me think that `enum :update_status` in ticket log should probably be a `prefix: true`, so that this can read `if ticket_log.action_type_update_status?`

If that is too long for your taste, then instead of `prefix: true` as a boolean, you have the choice to actually specify a symbol as prefix: `prefix: :action` will give you `ticket_log.action_update_status?` for example

db/migrate/20250725060536_remove_action_value_from_ticket_logs.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class RemoveActionValueFromTicketLogs < ActiveRecord::Migration[7.2]
+  def change
+    remove_column :ticket_logs, :action_value, :string
+
+    reversible do |dir|
+      dir.down do
+        TicketLogChange.find_each do |ticket_log_change|
+          ticket_log_change.ticket_log.update!(action_value: ticket_log_change.field_value) if ticket_log_change.status?
```
gregorbg: `if ticket_log_change.status?` is the same argument as above: The enum should probably be `prefix: true` or `prefix: :field`, depending on your taste.

app/controllers/tickets_controller.rb
danieljames-dj: Okay thanks.

db/migrate/20250725050517_create_ticket_log_changes.rb
danieljames-dj: Actually default prefix makes sense. I've made the change. Thank you.

db/migrate/20250725060536_remove_action_value_from_ticket_logs.rb
danieljames-dj: Added prefix.

app/controllers/tickets_controller.rb
danieljames-dj: Okay understood, thank you. I'll work on this in a later stage. Hope that works.

gregorbg: Yup totally!

gregorbg (approved):

PR #12076, opened 2025-07-25
Hide competitor tab before reg opens

dunkOnIT:
A common complaint from organizers re auto accept is that they don't want to pre-accept organizers/delegates because it leads to the Competitors tab displaying before reg has opened, and them needing to deal with queries from confused prospective competitors.

We had already implemented this, and then we had to revert it because WC2025 would have its competitor list disappear between registration periods. We no longer have that issue, so we're free to re-implement this.

### DIFF
--- a/app/models/competition.rb
+++ b/app/models/competition.rb
@@ -1003,12 +1003,12 @@ class Competition < ApplicationRecord
     true
   end
 
-  def registration_currently_open?
-    use_wca_registration? && !cancelled? && !registration_not_yet_opened? && !registration_past?
-  end
-
   def after_registration_open?
     use_wca_registration? && !cancelled? && !registration_not_yet_opened?
+  end
+
+  def registration_currently_open?
+    after_registration_open? && !registration_past?
   end
 
   def registration_not_yet_opened?
--- a/app/models/competition.rb
+++ b/app/models/competition.rb
@@ -1007,6 +1007,10 @@ class Competition < ApplicationRecord
     use_wca_registration? && !cancelled? && !registration_not_yet_opened? && !registration_past?
   end
 
+  def after_registration_open?
+    use_wca_registration? && !cancelled? && !registration_not_yet_opened?
+  end
+
   def registration_not_yet_opened?
     registration_open && Time.now < registration_open
   end
--- a/app/models/competition.rb
+++ b/app/models/competition.rb
@@ -1011,12 +1011,20 @@ class Competition < ApplicationRecord
     after_registration_open? && !registration_past?
   end
 
+  def after_registration_open?
+    use_wca_registration? && !cancelled? && !registration_not_yet_opened? && any_registrations?
+  end
+
   def registration_not_yet_opened?
     registration_open && Time.now < registration_open
   end
 
   def registration_past?
     registration_close && registration_close < Time.now
+  end
+
+  def can_show_competitors_page?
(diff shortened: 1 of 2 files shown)


app/models/competition.rb
```diff
@@ -1007,6 +1007,10 @@ def registration_currently_open?
     use_wca_registration? && !cancelled? && !registration_not_yet_opened? && !registration_past?
   end
 
+  def after_registration_open?
```
danieljames-dj, line 1006: The method name doesn't match for what it actually is. For example, if a competition is cancelled after registration is open, this will return `false`, but it is still `after_registration_open`. Same for `use_wca_registration`. How about renaming this method to something like `can_show_competitors_page` and that way you can actually bring the `@competition.any_registrations?` as well to inside this method?

gregorbg: I would argue that the act of cancelling the registration implicitly closes registration.
Your argument points out a weakness in our naming scheme: Is it (a) technically _after_ (ie, has the moment in time passed) the timestamp we call `registration_open` in our database , or is it (b) after the concept "the registration has opened for this competition".

For consistency with legacy methods in our code (for example, see `registration_currently_open?` above) I am arguing in favor of version (b) above and letting the method stand.

gregorbg: One avenue for compromise that I can see is to combine this method's result and the `any_registrations?` result into one new `can_show_competitors_page?` method. But I'm not sure how deep we want to dive into permissions territory here.

app/models/competition.rb
```diff
@@ -1007,6 +1007,10 @@ def registration_currently_open?
...
+    use_wca_registration? && !cancelled? && !registration_not_yet_opened?
+  end
```
gregorbg, line 1008: If you move this definition above `registration_currently_open?`, you can reuse it:
```
def after_registration_open?
  ...
end

def registration_currently_open?
  after_registration_open? && !registration_past?
end

Comment (dunkOnIT) [reply app/models/competition.rb:1008]: Changed as suggested

File: app/models/competition.rb (hunk 1 above)
Comment (dunkOnIT) [reply app/models/competition.rb:1006]: I tend to agree with interpretation (b) as Gregor has laid out, but I do agree there's a weakness in our naming here. I don't want to blow the scope of this relatively simple PR with that, though. 

`can_show_competitors_page?` is tempting, especially given that it would give us a nice place to introduce conditional logic for Major Championships if we needed to. Actually, I'm going to do that - given that `after_registration_open?` will only be used in this one place in our code anyway

Comment (dunkOnIT) [reply app/models/competition.rb:1006]: (Ended up deciding to keep `after_registration_open?` because otherwise the logic would have just been duplicated in `can_show_competitors_page?` and `registration_currently_open?`

File: app/models/competition.rb [hunk 3]
```diff
@@ -1015,6 +1019,10 @@ def registration_past?
     registration_close && registration_close < Time.now
   end
 
+  def can_show_competitors_page?
+    after_registration_open? && registrations.competing_status_accepted.competing.any?
```
gregorbg, line 1023: This used to be `any_registrations?`. Why have you changed to the significantly more complex version here?

dunkOnIT: Because we only display accepted and competing registrations in the Competitors tab - we don't want to be displaying an empty Competitors tab if we have one `rejected` registration, or a non-competing `accepted` registration

PR #12080, opened 2025-07-26
Allow to retry Duplicate checker if delayed

danieljames-dj:
Currently if a job is stuck due to some unknown reason, Delegates won't have any way to retry. This PR is providing a "Retry" button.

Also did some cleanups related to the same files.

### DIFF
--- a/app/controllers/results_submission_controller.rb
+++ b/app/controllers/results_submission_controller.rb
@@ -44,6 +44,11 @@ class ResultsSubmissionController < ApplicationController
   end
 
   def compute_potential_duplicates
+    last_job_run = DuplicateCheckerJobRun.find_by(competition_id: params.require(:competition_id))
+    should_fail_last_job_run = last_job_run&.run_status_not_started? || last_job_run&.run_status_in_progress?
+
+    last_job_run.update!(run_status: DuplicateCheckerJobRun.run_statuses[:failed]) if should_fail_last_job_run
+
     job_run = DuplicateCheckerJobRun.create!(competition_id: params.require(:competition_id))
     ComputePotentialDuplicates.perform_later(job_run)
 
--- a/app/controllers/results_submission_controller.rb
+++ b/app/controllers/results_submission_controller.rb
@@ -45,9 +45,9 @@ class ResultsSubmissionController < ApplicationController
 
   def compute_potential_duplicates
     last_job_run = DuplicateCheckerJobRun.find_by(competition_id: params.require(:competition_id))
-    should_fail_last_job_run = last_job_run&.run_status_not_started? || last_job_run&.run_status_in_progress?
+    job_run_running_too_long = last_job_run&.run_status_not_started? || last_job_run&.run_status_in_progress?
 
-    last_job_run.update!(run_status: DuplicateCheckerJobRun.run_statuses[:failed]) if should_fail_last_job_run
+    last_job_run.update!(run_status: DuplicateCheckerJobRun.run_statuses[:long_running_uncertain]) if job_run_running_too_long
 
     job_run = DuplicateCheckerJobRun.create!(competition_id: params.require(:competition_id))
     ComputePotentialDuplicates.perform_later(job_run)
--- a/app/webpacker/components/NewcomerChecks/DuplicateCheckerHeader.jsx
+++ b/app/webpacker/components/NewcomerChecks/DuplicateCheckerHeader.jsx
@@ -3,6 +3,8 @@ import { DateTime } from 'luxon';
 import { Button, Message } from 'semantic-ui-react';
 import { duplicateCheckerJobRunStatuses } from '../../lib/wca-data.js.erb';
 
+// Usually this job should run in less than 5 minutes in most of the cases. It
+// can increase sometimes. A safe definition of "delay" looks like 10 minutes.
 const JOB_RUN_DELAY_MINUTES = 10;
 
 function isJobRunDelayed(lastDuplicateCheckerJobRun) {
@@ -33,21 +35,27 @@ export default function DuplicateCheckerHeader({
         <Button onClick={run}>Retry</Button>
       </Message>
     );
-  } if (lastDuplicateCheckerJobRun.run_status === duplicateCheckerJobRunStatuses.not_started) {
+  }
+
+  if (lastDuplicateCheckerJobRun.run_status === duplicateCheckerJobRunStatuses.not_started) {
     return (
       <Message info>
         Duplicate Checker will start soon. Please check after sometime.
         <Button onClick={refetch}>Refresh</Button>
       </Message>
     );
-  } if (lastDuplicateCheckerJobRun.run_status === duplicateCheckerJobRunStatuses.in_progress) {
+  }
+
+  if (lastDuplicateCheckerJobRun.run_status === duplicateCheckerJobRunStatuses.in_progress) {
     return (
       <Message info>
         Duplicate Checker is currently running. Please check after sometime.
         <Button onClick={refetch}>Refresh</Button>
       </Message>
     );
-  } if (lastDuplicateCheckerJobRun.run_status === duplicateCheckerJobRunStatuses.success) {
+  }
+
+  if (lastDuplicateCheckerJobRun.run_status === duplicateCheckerJobRunStatuses.success) {
     return (
       <Message positive>
         {`Duplicate Checker ran successfully at ${
@@ -55,7 +63,9 @@ export default function DuplicateCheckerHeader({
         <Button onClick={run}>Re-run now</Button>
       </Message>
     );
-  } if (lastDuplicateCheckerJobRun.run_status === duplicateCheckerJobRunStatuses.failed) {
+  }
+
+  if (lastDuplicateCheckerJobRun.run_status === duplicateCheckerJobRunStatuses.failed) {
     return (
       <Message negative>
         Something went wrong. Please try running again.
@@ -63,6 +73,7 @@ export default function DuplicateCheckerHeader({
       </Message>
     );
   }
+
   return (
     <Message positive>
       Duplicate Checker has not yet ran.
--- a/app/webpacker/components/NewcomerChecks/DuplicateCheckerHeader.jsx
+++ b/app/webpacker/components/NewcomerChecks/DuplicateCheckerHeader.jsx
@@ -3,13 +3,14 @@ import { DateTime } from 'luxon';
 import { Button, Message } from 'semantic-ui-react';
 import { duplicateCheckerJobRunStatuses } from '../../lib/wca-data.js.erb';
 
-// As per the specification of server while writing this comment, for each
-// server, it will take approx 2 seconds. The number of newcomers will be
-// usually 150 in very big competitions. It can go up as well. Considering 150,
-// the time taken will be around 5 minutes. I'm ignoring the uncertainties and
-// expecting that number of newcomers will get doubled over next few years. So
-// defining the delay as 10 minutes.
-const JOB_RUN_DELAY_MINUTES = 10;
+// This is the average estimate runtime for a person as of July 2025.
+const ESTIMATE_RUNTIME_SECONDS_PER_PERSON = 2;
+
+// This is the approx number of newcomers in very big competitions as of July 2025.
+const NEWCOMER_COUNT_THRESHOLD = 150;
+
+// This is the number of minutes after which we consider the job to be delayed.
+const JOB_RUN_DELAY_MINUTES = (NEWCOMER_COUNT_THRESHOLD * ESTIMATE_RUNTIME_SECONDS_PER_PERSON) / 60;
 
 function isJobRunDelayed(lastDuplicateCheckerJobRun) {
   const jobNotStartedOrInProgress = (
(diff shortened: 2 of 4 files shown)


app/controllers/results_submission_controller.rb
```diff
@@ -44,6 +44,11 @@ def newcomer_dob_check
   end
 
   def compute_potential_duplicates
+    last_job_run = DuplicateCheckerJobRun.find_by(competition_id: params.require(:competition_id))
```
gregorbg, line 47: How can you make sure that this `find_by` always delivers the _last_ job run, when there can be multiple job runs? The job checker run has a default scope to apply an `ORDER BY` clause, but I'm not sure whether `find_by` yields the first or the last entry in that ordered set.

app/controllers/results_submission_controller.rb
```diff
@@ -44,6 +44,11 @@ def newcomer_dob_check
...
+    should_fail_last_job_run = last_job_run&.run_status_not_started? || last_job_run&.run_status_in_progress?
```
gregorbg: This is a bit confusing to me, I had to read it twice to understand what you're doing.

My main source of confusion is this: If there is a `last_job_run`, and if it is marked as (for example) `in_progress`, then why should it fail?

The confusion is made even stronger based on the fact that this code snippet is executed no matter what. Maybe provide the user with some additional "override previous job run(s)" checkbox, that transmits a boolean to the backend? Or would that be too complicated to explain to the average Delegate?

app/webpacker/components/NewcomerChecks/DuplicateCheckerHeader.jsx
```diff
@@ -1,16 +1,50 @@
...
 import { DateTime } from 'luxon';
 import { Button, Message } from 'semantic-ui-react';
 import { duplicateCheckerJobRunStatuses } from '../../lib/wca-data.js.erb';
 
-export default function DuplicateCheckerHeader({ lastDuplicateCheckerJobRun, run }) {
-  if (
-    lastDuplicateCheckerJobRun.run_status === duplicateCheckerJobRunStatuses.in_progress
-    || lastDuplicateCheckerJobRun.run_status === duplicateCheckerJobRunStatuses.not_started
-  ) {
+const JOB_RUN_DELAY_MINUTES = 10;
```
gregorbg: Why is this a good estimate? Can we be sure things will never get worse in the future? Please add a short (1-2 lines) code comment outlining your rationale behind this number.

app/webpacker/components/NewcomerChecks/DuplicateCheckerHeader.jsx
```diff
@@ -1,16 +1,50 @@
...
+  );
+
+  if (jobRunDelayed) {
+    return (
+      <Message warning>
+        Job running longer than expected. Click &apos;Retry&apos; if it appears stuck.
+        <Button onClick={run}>Retry</Button>
+      </Message>
+    );
+  } if (lastDuplicateCheckerJobRun.run_status === duplicateCheckerJobRunStatuses.not_started) {
```
gregorbg: You should either (a) turn this into an `else if` or (b) add a line break before the subsequent if.

app/webpacker/components/NewcomerChecks/DuplicateCheckerHeader.jsx
```diff
@@ -1,16 +1,50 @@
...
+    return (
+      <Message info>
+        Duplicate Checker will start soon. Please check after sometime.
+        <Button onClick={refetch}>Refresh</Button>
+      </Message>
+    );
+  } if (lastDuplicateCheckerJobRun.run_status === duplicateCheckerJobRunStatuses.in_progress) {
```
gregorbg: Same as above regarding line break or `else if`

app/webpacker/components/NewcomerChecks/DuplicateCheckerHeader.jsx
```diff
@@ -1,16 +1,50 @@
...
     return (
       <Message info>
         Duplicate Checker is currently running. Please check after sometime.
+        <Button onClick={refetch}>Refresh</Button>
       </Message>
     );
   } if (lastDuplicateCheckerJobRun.run_status === duplicateCheckerJobRunStatuses.success) {
```
gregorbg: Same as above regarding line break or `else if`

gregorbg (changes requested):

app/controllers/results_submission_controller.rb
danieljames-dj: I've tested it, this is giving last entry. Also I'm using this same in the above method `last_duplicate_checker_job_run`.

app/controllers/results_submission_controller.rb
danieljames-dj: The checkbox might make it complicated to Delegate I assume. But irrespective of that, when a new run is executed, the results of the previous run is no longer relevant and cannot be fetched by the user as well. It's fine even if we don't 'fail' them because they won't be accessed any longer.

I'm marking it as fail just for the sake of completion. We don't know if the job will get completed in future or not, if it gets completed, it will get changed to success - but doesn't matter for us anyway.

app/webpacker/components/NewcomerChecks/DuplicateCheckerHeader.jsx
danieljames-dj: Okay, added a comment.

app/webpacker/components/NewcomerChecks/DuplicateCheckerHeader.jsx
danieljames-dj: Okay opted for second one.

app/webpacker/components/NewcomerChecks/DuplicateCheckerHeader.jsx
danieljames-dj: Done

app/webpacker/components/NewcomerChecks/DuplicateCheckerHeader.jsx
danieljames-dj: Done

app/controllers/results_submission_controller.rb
gregorbg: Then may I suggest a compromise, which is to introduce a new job status that describes "not sure whether it failed but we think it's running too long so we treat it as failed"?

That way, we can distinguish between jobs that actually failed, and jobs which we deprecated.

app/webpacker/components/NewcomerChecks/DuplicateCheckerHeader.jsx
```diff
@@ -1,34 +1,79 @@
...
 import { Button, Message } from 'semantic-ui-react';
 import { duplicateCheckerJobRunStatuses } from '../../lib/wca-data.js.erb';
 
-export default function DuplicateCheckerHeader({ lastDuplicateCheckerJobRun, run }) {
-  if (
-    lastDuplicateCheckerJobRun.run_status === duplicateCheckerJobRunStatuses.in_progress
-    || lastDuplicateCheckerJobRun.run_status === duplicateCheckerJobRunStatuses.not_started
-  ) {
+// Usually this job should run in less than 5 minutes in most of the cases. It
+// can increase sometimes. A safe definition of "delay" looks like 10 minutes.
```
gregorbg: This comment is essentially saying "I declare that it runs in 5 minutes, and I think that 10 minutes is reasonable".

- _Why_ is 10 minutes reasonable? You are doubling the average runtime. Why not triple it? Why not use factor 1.5?
- Does the runtime of 5 minutes scale according to number of competitors? So does "more competitors" mean "(significantly) more runtime"? Or is it more the case that the initial data fetch takes a constant 4 minutes, and all the comparisons for each competitor happen within 1 minute? This affects the decision about a good estimate.
- What makes your delay "safe"? What are the consequences of restarting a job that we believe failed (because it ran for more than your current 10 minute threshold) but actually didn't fail?

gregorbg (changes requested):

app/controllers/results_submission_controller.rb
danieljames-dj: Okay done

app/webpacker/components/NewcomerChecks/DuplicateCheckerHeader.jsx
danieljames-dj: Hm good points. Actually it depends on number of newcomers. I'm assuming that for 1 newcomer it will take around 2 seconds. If a competition has 150 newcomers, it can go up to 300 seconds (5 minutes). I feel fetching the number of newcomers here and giving a warning will be overkill, so I assumed 150 newcomers as a reasonable choice for big competitions.

What do you suggest here? IMO leaving it as 10 minutes will be better. I've modified the comments to be more clear.

gregorbg: I agree that fetching the number of newcomers just to compute the threshold is overkill.
Then what about adding a visible text next to the "Reset" button, along the lines of "This button is visible because the last check did not complete within {xyz} minutes. If your competitions has a very high number of newcomers (more than 150), it may be possible that your job actually needs more time to run. Please use this button at your own discretion."

danieljames-dj: Done

app/webpacker/components/NewcomerChecks/DuplicateCheckerHeader.jsx
```diff
@@ -1,34 +1,87 @@
...
 import { Button, Message } from 'semantic-ui-react';
 import { duplicateCheckerJobRunStatuses } from '../../lib/wca-data.js.erb';
 
-export default function DuplicateCheckerHeader({ lastDuplicateCheckerJobRun, run }) {
-  if (
-    lastDuplicateCheckerJobRun.run_status === duplicateCheckerJobRunStatuses.in_progress
-    || lastDuplicateCheckerJobRun.run_status === duplicateCheckerJobRunStatuses.not_started
-  ) {
+// As per the specification of server while writing this comment, for each
+// server, it will take approx 2 seconds. The number of newcomers will be
```
gregorbg: I don't understand this.
- "the specification of server" --> which server?
- "it will take approx 2 seconds" --> what is taking 2 seconds?

Do you mean 2 seconds per person, so 2 sec*150=300 sec=5 min?

app/webpacker/components/NewcomerChecks/DuplicateCheckerHeader.jsx
```diff
@@ -1,34 +1,87 @@
...
+  );
+
+  if (jobRunDelayed) {
+    return (
+      <Message warning>
+        {`Job running longer than ${JOB_RUN_DELAY_MINUTES} minutes. Click "Retry" if it appears
+        stuck. (Please note - retry option is available because it's running for more than
+        ${JOB_RUN_DELAY_MINUTES} minutes). If your competitions has a very high number of newcomers
+        (more than 150), it may be possible that your job actually needs more time to run. Please
+        use this button at your own discretion.`}
```
gregorbg: You don't need to use backtick interpolation for the whole message body.
```suggestion
        Job running longer than
        {JOB_RUN_DELAY_MINUTES}
        minutes. Click "Retry" if it appears stuck. (Please note - retry option is available because it's running for more than
        {JOB_RUN_DELAY_MINUTES}
        minutes). If your competitions has a very high number of newcomers
        (more than 150), it may be possible that your job actually needs more time to run. Please use this button at your own discretion.}
```

gregorbg: Also, the phrasing can be made a bit shorter. For example, the whole part in brackets ("Please note...") can become "Job appears stuck, because it has been running longer than XYZ minutes".

gregorbg: Full suggestion:

```
Job appears stuck, because it has been running longer than XYZ minutes. Click "Retry" below if you would like to cancel this job run and start a new check.

Please note: If your competition has an unusually high number of newcomers (more than 150), it may be possible that your job actually needs more time to run. Please use the button at your own discretion.
```

app/webpacker/components/NewcomerChecks/DuplicateCheckerHeader.jsx
```diff
@@ -1,34 +1,87 @@
...
-    lastDuplicateCheckerJobRun.run_status === duplicateCheckerJobRunStatuses.in_progress
-    || lastDuplicateCheckerJobRun.run_status === duplicateCheckerJobRunStatuses.not_started
-  ) {
+// As per the specification of server while writing this comment, for each
+// server, it will take approx 2 seconds. The number of newcomers will be
+// usually 150 in very big competitions. It can go up as well. Considering 150,
+// the time taken will be around 5 minutes. I'm ignoring the uncertainties and
+// expecting that number of newcomers will get doubled over next few years. So
+// defining the delay as 10 minutes.
+const JOB_RUN_DELAY_MINUTES = 10;
```
gregorbg: Also, now that you have explained where this constant comes from, what about changing the code? You can introduce a `const ESTIMATE_RUNTIME_PER_PERSON = 2` and `const NEWCOMER_COUNT_THRESHOLD = 150` and then just multiply them together: `const JOB_RUN_DELAY_MINUTES = NEWCOMER_COUNT_THRESHOLD * ESTIMATE_RUNTIME_PER_PERSON / 60`.

gregorbg: This will make the code "explain itself" to a large degree, and you can reduce your comments substantially (for example, a simple "This is the average estimate for newcomers as of July 2025" will be enough for the `NEWCOMER_COUNT_THRESHOLD` variable)

gregorbg: And you can also reuse the constant values like `150` in the text message that you're showing below to Delegates for the Retry button

gregorbg (changes requested):

app/webpacker/components/NewcomerChecks/DuplicateCheckerHeader.jsx
danieljames-dj: This is really nice. Implemented.

app/webpacker/components/NewcomerChecks/DuplicateCheckerHeader.jsx
danieljames-dj: Okay changed.

app/webpacker/components/NewcomerChecks/DuplicateCheckerHeader.jsx
danieljames-dj: Yes, I've made it more clear now. Hope it's fine now.

gregorbg (approved):

PR #12081, opened 2025-07-27
Edit user modal for duplicate checker page

danieljames-dj:
This is to improve the UX of Duplicate checker page - the users doesn't have to go to edit user page to make changes.

Also, I'm planning to use this same component in the edit user page in future.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -7,6 +7,7 @@ class UsersController < ApplicationController
   before_action :set_recent_authentication!, only: %i[edit update enable_2fa disable_2fa]
   before_action :redirect_if_cannot_edit_user, only: %i[edit update]
   before_action -> { redirect_to_root_unless_user(:can_admin_results?) }, only: %i[admin_search merge]
+  before_action -> { check_edit_access }, only: %i[show_for_edit update_user_data]
 
   RECENT_AUTHENTICATION_DURATION = 10.minutes.freeze
 
@@ -45,6 +46,29 @@ class UsersController < ApplicationController
         }
       end
     end
+  end
+
+  private def check_edit_access
+    @user = User.find(params.require(:id))
+    cannot_edit_reason = current_user.cannot_edit_data_reason_html(@user)
+
+    render status: :unauthorized, json: { error: cannot_edit_reason } if cannot_edit_reason.present?
+  end
+
+  def show_for_edit
+    render status: :ok, json: @user.as_json(
+      private_attributes: %w[dob],
+    )
+  end
+
+  def update_user_data
+    user_details = params.permit(:name, :gender, :dob, :country_iso2)
+    
+    @user.update!(user_details)
+
+    render status: :ok, json: @user.as_json(
+      private_attributes: %w[dob],
+    )
   end
 
   def show_for_merge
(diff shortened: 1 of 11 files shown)


app/controllers/users_controller.rb
```diff
@@ -47,6 +48,29 @@ def index
     end
   end
 
+  private def check_edit_access
+    @user = User.find(params.require(:id))
+    cannot_edit_reason = current_user.cannot_edit_data_reason_html(@user)
```
gregorbg, line 53: Are we sure that the user is definitely always logged in at this point? Otherwise it should be `current_user&.foo(..)`

app/models/user.rb
```diff
@@ -1045,6 +1045,8 @@ def cannot_organize_competition_reasons
   end
 
   def cannot_edit_data_reason_html(user_to_edit)
+    return I18n.t('users.edit.cannot_edit.reason.no_access') unless user_to_edit == self || can_edit_any_user?
```
gregorbg, line 1048: Where does this new check come from? I can understand from the code what this check does, you don't need to explain that in detail. What I want to know is:
1. Why was this check never necessary before?
2. Will this affect other parts in the code (outside of your current PR) that rely on `cannot_edit_data_reason_html`?

app/webpacker/components/EditUser/api/getUserDetails.js
```diff
@@ -0,0 +1,18 @@
...
+    viewUrls.users.showForEdit(userId),
+  );
+
+  const userDetails = {
+    id: data.id,
+    name: data.name,
+    dob: data.dob,
+    gender: data.gender,
+    country_iso2: data.country_iso2,
+  };
```
gregorbg: If your intention is to create an exact copy (as far as I can tell, all keys that you are creating are exactly identical to `data`, then you can also just do `{ ...data }`.

gregorbg: But also: Why are you creating a shallow copy in the first place, instead of directly returning `data`?

gregorbg (changes requested):

app/controllers/users_controller.rb
danieljames-dj: Yes the user will be logged in because there is `before_action :authenticate_user!` before this `before_action`.

app/models/user.rb
danieljames-dj: So this was used in two places earlier:
1. Getting editable fields of user
2. Show "cannot edit" warning in user edit page

For 1, this check was already there before calling this method. I've changed that as well below.

For 2, it will be good to have this check because the user won't be anyway able to edit as these are not added in the editable fields. So this check will be useful.

This answered your both questions I assume.

app/webpacker/components/EditUser/api/getUserDetails.js
danieljames-dj: Basically I don't want the full data to go to the frontend because remaining data are not necessary there (especially it's very huge I assume). I don't see any negative effects for sending the `data` completely as well. Should I send it as `data` instead?

gregorbg: If you don't want the full data to go to the frontend (ie you don't need it at all) then consider creating a dedicated endpoint which only gives you the fields that you need. Bonus: This avoids redundant database queries for stuff like the user avatar, which is quite costly to retrieve.

danieljames-dj: Okay done

PR #12097, opened 2025-07-28
Add Terms and Condition Tables

FinnIckler:


db/migrate/20250728105210_add_t_c_tables.rb
```diff
@@ -0,0 +1,22 @@
...
+      t.datetime :deleted_at, index: true
+
+      t.timestamps
+    end
+
+    create_table :terms_and_conditions_accepts do |t|
+      t.references :terms_and_conditions, null: false, foreign_key: true
+      t.references :user, type: :integer, null: false, foreign_key: true
+
+      t.timestamps
```
gregorbg, line 19: Please provide `accepted_at` as a separate timestamp. I want to be able to track this without relying on the "technical" Rails timestamps

db/migrate/20250728105210_add_t_c_tables.rb
```diff
@@ -0,0 +1,22 @@
...
+    end
+
+    add_index :terms_and_conditions_accepts, %i[terms_and_conditions_id user_id], unique: true, name: 'index_terms_accepts_on_terms_and_user'
```
gregorbg: I think there is `t.index` to use directly within the `create_table` block

db/migrate/20250728105210_add_t_c_tables.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+class AddTCTables < ActiveRecord::Migration[7.2]
+  def change
+    create_table :terms_and_conditions do |t|
+      t.text :content, null: false
+      t.references :competition, type: :string, null: false, foreign_key: true
+      t.datetime :deleted_at, index: true
```
gregorbg:
```suggestion
      t.datetime :revoked_at, index: true
```

gregorbg (changes requested):

db/migrate/20250728105210_add_t_c_tables.rb
FinnIckler: replaced with revoked_at

PR #12099, opened 2025-07-28
Display competitor list regardless of whether using WCA reg or not

gregorbg:
Fixes a critical bug where registrations that are not on our system are not showing registrations, even after successfully uploading a CSV.

The registrations are there, but due to #12076 they are hidden.

In the diff of that PR, you can see that we pulled out "too much" of the previous `registration_currently_open?` method into a new, refactored `after_registration_open?` method. In particular, the check whether a competition is using WCA system or not should not influence whether we consider a competition as "after registration open" or not.

I am not sure how to proceed about the `cancelled` part, so I am leaving this here as a PR for review rather than pushing a hotfix directly to `main`.

### DIFF
--- a/app/models/competition.rb
+++ b/app/models/competition.rb
@@ -1004,11 +1004,11 @@ class Competition < ApplicationRecord
   end
 
   def after_registration_open?
-    use_wca_registration? && !cancelled? && !registration_not_yet_opened?
+    !registration_not_yet_opened?
   end
 
   def registration_currently_open?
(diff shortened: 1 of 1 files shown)


dunkOnIT:
We don't at the moment - in normal circumstances I'd say we should add them, but given that 
- the Competitors tab code is not very volatile, and 
- we're transitioning this page to Next soon

I'd say we can survive without a test until the Next transition

PR #12102, opened 2025-07-28
Add new `wca_status`:`requires_capture` 

dunkOnIT:
This PR adds the `requires_capture` `wca_status` to `PaymentIntent`, and implements it for `StripeRecord`.

I could be crazy, but this seems to be all that's needed. The main issue I see is the relatively strange behaviour of updating the status of the PaymentIntent in `update_status_and_charges` - specifically: 
- Until now, the status wasn't updated unless it was one of the statuses we checked for in the case statement - this seems wrong? 
- `PaymentIntent#update_intent_details` only updated the `error_details` - why were we assuming there would be an error? And if we aren't assuming there would be an error, why are we doing a db update for a `nil` value? 
  - I have updated this method to update the wca_status 
  - I would like to make this method an `else` on the `case` block, but perhaps there is a reason it is deliberately run (as far as I can tell) for every time `update_status_and_charges` is called? 
- `StripeRecord#update_status` also seems to assume there will be an error // writes a `nil` value to the database? Also, the `when... [:charge]` block here seems to not use a safe accessor - which it seems like we should? 
```ruby
    when StripeRecord.stripe_record_types[:payment_intent]
      stripe_error = api_transaction.last_payment_error&.code
    when StripeRecord.stripe_record_types[:charge]
      stripe_error = api_transaction.failure_message
    end
```

These smell like issues that stem from just not thoroughly testing the logic in `update_status_and_charges` - but perhaps there's something I'm missing?


app/models/stripe_record.rb
gregorbg, line 1: I think you can do a similar change for `PaypalRecord` as well

spec/factories/payment_intents.rb
```diff
@@ -23,5 +23,9 @@
       payment_record { association(:stripe_record, :payment_intent, :not_started) }
       wca_status { 'created' }
     end
+
+    trait :pending do
+      payment_record { association(:stripe_record) }
```
gregorbg, line 28: In how far is the `stripe_record` that you're associating here representing the `pending` state?

spec/models/payment_intent_spec.rb
```diff
@@ -91,4 +91,16 @@
       it_behaves_like '#update PaymentIntent to incompatible status', 'legacy_failure', 'failed', 'succeeded'
     end
   end
+
+  describe 'update status and charges' do
+    let(:intent) { create(:payment_intent, :pending) }
+    let(:mock_update) { double("StripeUpdate", status: "requires_capture", last_payment_error: nil) }
```
gregorbg, line 98: What is `"StripeUpdate"`?

app/models/payment_intent.rb
```diff
@@ -85,16 +86,16 @@ def update_status_and_charges(payment_account, api_intent, action_source, source
       end
 
       # Write any additional details now that the main status cycle is complete
-      self.update_intent_details(api_intent)
+      self.update_intent_details(updated_wca_status, api_intent)
```
gregorbg: If you pass this along here, then a lot of `update` statements above are becoming partially redundant (they also execute an update on `wca_status` among others, which could be cleaned up)

gregorbg (changes requested):

spec/factories/payment_intents.rb
dunkOnIT: I don't quite understand the question, but the StripeRecord factory's defaults align with a pending state so I didn't have to pass any other arguments when creating the association

spec/models/payment_intent_spec.rb
dunkOnIT: The name for the mock class - not important unless we're making assertions about class type somewhere in the code

spec/factories/payment_intents.rb
gregorbg: That pretty much resolves my question. The fact that creating an associated `stripe_record` implicitly marks it as `pending` makes total sense, I am debating whether or not you should add a `pending` trait to the `stripe_record` as well to make it clear/explicit.

But nonetheless that's a bit out of the scope of this PR.

spec/models/payment_intent_spec.rb
gregorbg: Okay, then perhaps could you add a comment that the name of the mocked class you chose (`"StripeUpdate"`) is arbitrary and (somewhat) meaningless for this test?

dunkOnIT:
Ok thanks for the info, it all makes a lot more sense! 

The main change I'm suggesting with my latest commit is to simplify the update handling, by only having one `update` run for each call, and doing everything needed in there. At least for me, this is approach is easier to understand as someone not familiar with the code and its intentions. 

Additionally, I think it is important that we aggressively store status changes - not least because we're aggressively storing status changes for the record, and we have strong expectations in the code that the `wca_status` and `stripe_status` mappings align. Even if it doesn't produce errors in our current implementation, it is not intuitive at all that we could have "legitimate" cases where the `stripe_status` and `wca_status` can go out of alignment because we don't persist the updated `wca_status` to the database

app/models/payment_intent.rb
gregorbg, line 1: This change doesn't work the way you think, because you have now pulled the `api_intent.last_payment_error` access out of the switch-case.

The switch-case was there to specifically make sure what kind of API intent we're dealing with. The field `last_payment_error` is part of the Stripe SDK and as such only exists on StripeRecord entites.

Without the `switch` (that you're getting rid of in this commit), you cannot even be sure whether the field `last_payment_error` even exists.

gregorbg (changes requested):

app/models/payment_intent.rb
dunkOnIT: Ah great point - updated the method to provide the error type instead of handling the update

spec/models/payment_intent_spec.rb
dunkOnIT: Done

app/models/payment_intent.rb
```diff
@@ -41,7 +42,7 @@ def update_status_and_charges(payment_account, api_intent, action_source, source
       self.payment_record.update_status(api_intent)
       updated_wca_status = self.determine_wca_status.to_s
 
-      case updated_wca_status
+      case updated_wca_status # This has been updated by assign_common_attributes
```
gregorbg: This comment is not correct anymore

app/models/payment_intent.rb
```diff
@@ -80,22 +79,22 @@ def update_status_and_charges(payment_account, api_intent, action_source, source
           confirmation_source: nil,
           canceled_at: nil,
           cancellation_source: nil,
-          wca_status: updated_wca_status,
         )
       end
 
-      # Write any additional details now that the main status cycle is complete
-      self.update_intent_details(api_intent)
+      update_common_attributes!(api_intent)
```
gregorbg: You should do `self.` here (because there's a whole lot of entities/objects flying around in this method so it helps with disambiguation) and you also actually have to pass in the updated wca_status that the method signature now expects

dunkOnIT: Done

app/models/payment_intent.rb
dunkOnIT: Done

app/models/payment_intent.rb
```diff
@@ -80,22 +79,22 @@ def update_status_and_charges(payment_account, api_intent, action_source, source
           confirmation_source: nil,
           canceled_at: nil,
           cancellation_source: nil,
-          wca_status: updated_wca_status,
         )
       end
 
-      # Write any additional details now that the main status cycle is complete
-      self.update_intent_details(api_intent)
+      self.update_common_attributes!(api_intent)
```
gregorbg: You're still not passing in the `updated_wca_status` here

gregorbg (approved):

gregorbg (approved):

PR #12103, opened 2025-07-29
Migrate "merge temporary results" step to react

danieljames-dj:
This migrates one step of results posting to react and inside ticket. The old way is still there for backward compatibility.

This PR also involves one unrelated change - changing ticket ID to integer in Tickets/index.jsx because everywhere else the id is integer.

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -85,7 +85,7 @@ class AdminController < ApplicationController
   def import_inbox_results
     @competition = competition_from_params
 
-    CompetitionResultsImport.merge_temporary_results(@competition.result_ticket)
+    @competition.tickets_competition_result.merge_temporary_results
 
     load_result_posting_steps do
       render partial: 'import_results_steps'
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -90,7 +90,7 @@ class AdminController < ApplicationController
       return redirect_to competition_admin_import_results_path(@competition)
     end
 
-    @competition.tickets_competition_result.merge_temporary_results
+    @competition.tickets_competition_result.merge_inbox_results
 
     load_result_posting_steps do
       render partial: 'import_results_steps'
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -85,6 +85,11 @@ class AdminController < ApplicationController
   def import_inbox_results
     @competition = competition_from_params
 
+    if @competition.tickets_competition_result.nil?
+      flash[:danger] = "Cannot import results because the competition does not have a results ticket."
+      return redirect_to competition_admin_import_results_path(@competition)
+    end
+
     @competition.tickets_competition_result.merge_temporary_results
 
     load_result_posting_steps do
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -85,7 +85,7 @@ class AdminController < ApplicationController
   def import_inbox_results
     @competition = competition_from_params
 
-    @competition.tickets_competition_result.merge_temporary_results
+    CompetitionResultsImport.merge_temporary_results(@competition.result_ticket)
 
     load_result_posting_steps do
       render partial: 'import_results_steps'
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -85,38 +85,7 @@ class AdminController < ApplicationController
   def import_inbox_results
     @competition = competition_from_params
 
-    ActiveRecord::Base.transaction do
-      result_rows = @competition.inbox_results
-                                .includes(:inbox_person)
-                                .map do |inbox_res|
-        inbox_person = inbox_res.inbox_person
-
-        person_id = inbox_person&.wca_id.presence || inbox_res.person_id
-        person_country = inbox_person&.country
-
-        {
-          pos: inbox_res.pos,
-          person_id: person_id,
-          person_name: inbox_res.person_name,
-          country_id: person_country.id,
-          competition_id: inbox_res.competition_id,
-          event_id: inbox_res.event_id,
-          round_type_id: inbox_res.round_type_id,
-          round_id: inbox_res.round_id,
-          format_id: inbox_res.format_id,
-          value1: inbox_res.value1,
-          value2: inbox_res.value2,
-          value3: inbox_res.value3,
-          value4: inbox_res.value4,
-          value5: inbox_res.value5,
-          best: inbox_res.best,
-          average: inbox_res.average,
-        }
-      end
-
-      Result.insert_all!(result_rows)
-      @competition.inbox_results.destroy_all
-    end
+    @competition.tickets_competition_result.merge_temporary_results
 
     load_result_posting_steps do
       render partial: 'import_results_steps'
--- a/app/models/tickets_competition_result.rb
+++ b/app/models/tickets_competition_result.rb
@@ -59,41 +59,4 @@ class TicketsCompetitionResult < ApplicationRecord
       field_value: ticket_metadata.status,
     )
   end
-
-  def merge_temporary_results
-    ActiveRecord::Base.transaction do
-      result_rows = competition.inbox_results
-                                .includes(:inbox_person)
-                                .map do |inbox_res|
-        inbox_person = inbox_res.inbox_person
-
-        person_id = inbox_person&.wca_id.presence || inbox_res.person_id
-        person_country = inbox_person&.country
-
-        {
-          pos: inbox_res.pos,
-          person_id: person_id,
-          person_name: inbox_res.person_name,
-          country_id: person_country.id,
-          competition_id: inbox_res.competition_id,
-          event_id: inbox_res.event_id,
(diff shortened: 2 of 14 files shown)


app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/TimelineView.jsx
```diff
@@ -19,6 +19,17 @@ const TIMELINE_STATUSES = [
...
+  {
+    status: ticketsCompetitionResultStatuses.merged_temporary_results,
+    description: `WRT will have a rough look at the results to make sure that
+    there is no major flaw that requires aborting the posting process. For
+    example, check if there are big suspicious chunks of DNFs/DNSs and verify
+    that the shape of the results is sound (e.g. results should generally grow
+    “wider” from top to bottom with very few exceptions like DNFs or cutoffs).
+    Once done with the rough look, proceed to click the “Merge Temporary
+    Results” button which will copy data from InboxResults to Results, then
+    clear the data in InboxResults.`,
```
gregorbg: Did you have such a description before? Or did you just come up with it now for the PR?

If it's the latter: Where will the description be shown, and who will usually read it?

app/webpacker/components/Tickets/index.jsx
```diff
@@ -15,7 +15,7 @@ import updateStatus from './api/updateStatus';
 export default function Wrapper({ id }) {
   return (
     <WCAQueryClientProvider>
-      <Tickets id={id} />
+      <Tickets id={parseInt(id, 10)} />
```
gregorbg: This feels a little bit wrong. If you are so certain that it can definitely always be parsed as an integer, then why not just pass in an integer from the outside world in the first place?

Or in other words, under what circumstance is the `id` prop in `Wrapper` **not** an integer, and could you just fix that (instead of manually parsing here)?

app/models/tickets_competition_result.rb
```diff
@@ -58,4 +59,41 @@ def self.create_ticket!(competition, delegate_message, submitted_delegate)
       field_value: ticket_metadata.status,
     )
   end
+
+  def merge_temporary_results
```
gregorbg: I am slightly opposed (but could still be convinced otherwise) to keeping the business logic within the ticket metadata.

Tickets are for describing workflows, assigning responsibilities, tracking changes, etc. They are not made to actually perform the changes.

As a further argument: If you start this way now, it would mean that in the future we have to move ~70% of our whole codebase into tickets (everything around creating, announcing, posting competitions for example), which does not feel right.

Perhaps `lib` would be the best place for this kind of code.

gregorbg: I would be willing to accept a compromise where you leave `def merge_temporary_results` in the ticket, but the method body just calls the lib code.

gregorbg (changes requested):

app/models/tickets_competition_result.rb
danieljames-dj: Okay changed to lib. I feel this can be moved completely to controller once this is the only page where this method is used. That can probably happen once the deprecated page is removed in the future.

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/TimelineView.jsx
danieljames-dj: I copied this description from WRT procedures doc. I have a timeline view on the top of ticket workbench, so WRT members can refer there on the step. Eventually this can even be moved away from procedures doc so that the ticket will be single source of truth, and future new members will find it more easier.

app/webpacker/components/Tickets/index.jsx
danieljames-dj: Okay actually I was under the assumption that in future for next, the ID from URL will always be string. Now I did research, and we can actually make it int by defining type for the prop. So sending it as integer from backend makes sense. Changed.

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/TimelineView.jsx
gregorbg: So to summarize, this is only visible to WRT members and never to Delegates (except those who happen to be WRT member). Correct?

danieljames-dj: Yes correct.

lib/competition_results_import.rb
```diff
@@ -35,4 +35,42 @@ def self.import_temporary_results(competition, temporary_results_data, mark_resu
 
     errors
   end
+
+  def self.merge_temporary_results(ticket)
```
gregorbg: This method can only get `competition` as a parameter. If you need to do ticket-related changes, you should do them outside of this method, at the place where the method is called.

gregorbg (changes requested):

lib/competition_results_import.rb
danieljames-dj: Actually I also need to update the status of the ticket. This is called from two places and I need to do the same thing in both places. So is it fine to pass ticket instead like now?

gregorbg: No, because this is a lib code that deals only with results. Nothing else.

If you need to reuse ticket snippets, then (a) either just live with a little bit of code duplication or (b, preferrably) introduce a shared method somewhere else, outside of "lib-results" that handles the tickety stuff and calls this lib-results code internally.

danieljames-dj: Okay I've done it with help of ticket_competition_result model. Hope this is fine now?

lib/competition_results_import.rb
```diff
@@ -35,4 +35,37 @@ def self.import_temporary_results(competition, temporary_results_data, mark_resu
 
     errors
   end
+
+  def self.merge_temporary_results(competition)
```
gregorbg: You can (and should) still keep this as a transaction. Nested transactions are not forbidden! They are a little bit expensive, so you should not start wrapping every database operation you do in a transaction "just to be safe", but in cases like this, it's definitely justified.

gregorbg: Also, I am not happy about the new name. What made you switch away from the well-established `import_inbox_results`?

config/routes.rb
```diff
@@ -233,6 +233,7 @@
   end
   resources :tickets, only: %i[index show] do
     post 'update_status' => 'tickets#update_status', as: :update_status
+    post 'merge_temporary_results' => 'tickets#merge_temporary_results', as: :merge_temporary_results
```
gregorbg: Same as below about changing the name from `import_inbox_results`

app/controllers/admin_controller.rb
```diff
@@ -85,38 +85,7 @@ def result_inbox_steps
...
-          value5: inbox_res.value5,
-          best: inbox_res.best,
-          average: inbox_res.average,
-        }
-      end
-
-      Result.insert_all!(result_rows)
-      @competition.inbox_results.destroy_all
-    end
+    @competition.tickets_competition_result.merge_temporary_results
```
gregorbg: Can we be absolutely sure that this ticket always exists at this point?
Not just "sure because our flow is designed this way" but also "sure because there's some filter action somewhere that throws a 404 if it's not there", or "sure because the button that triggers this endpoint is hidden on a page that is only visible if the ticket exists"

gregorbg (changes requested):

app/controllers/admin_controller.rb
danieljames-dj: Okay added. Unfortunately the error is not shown clearly, the UI is not rendering properly here (on error). At least it's conveyed to the user that the inboxresults failed. I didn't spend time on error handling because this method is hopefully going away in another week or two.

lib/competition_results_import.rb
danieljames-dj: So the step were Delegates upload results are also known as "import results" (especially for live import). So I thought of using some name other import. But I'm open to discussion - if you feel I should change it back to `import_inbox_results`, I can change it.

config/routes.rb
danieljames-dj: Answered in that comment.

lib/competition_results_import.rb
danieljames-dj: And regarding transaction - okay, changed it to transaction.

gregorbg: Well I don't feel strongly about the "merge", but I do feel strongly about the new introduction of "temporary". We don't yet have a clear definition about what "temporary" means, and it's not commonly used outside of this PR (I think).

In my opinion, "inbox results" is a well established name for results that are "almost real" (ie, validators have passed, and they could be imported if WRT decides they agree). You are free to choose a verb that you like, to distinguish between (a) the Delegate starting the process of putting something into inbox_results and (b) the WRT putting the inbox_results into the proper results.

But both cases A and B should use the term "inbox_results" in my opinion.

gregorbg (changes requested):

lib/competition_results_import.rb
danieljames-dj: Okay renamed to `merge_inbox_results`

PR #12104, opened 2025-07-29
Migrate Submit results page to react

danieljames-dj:
This page migrates to React for the page which Delegates use to submit results.

### DIFF
--- a/app/views/results_submission/new.html.erb
+++ b/app/views/results_submission/new.html.erb
@@ -1,35 +1,12 @@
 <% provide(:title, "Results submission") %>
-<% add_to_packs("markdown_editor") %>
 
 <%= render layout: 'nav' do %>
   <h1><%= yield(:title) %></h1>
 
-  <% if @competition.results_submitted? %>
-    <p>The results have already been submitted. If you have any more questions or comments please reply to the email sent with the first results submission.</p>
-  <% else %>
-    <p>
-      The result submission process has two steps:
-      <ul>
-        <li>Uploading a valid JSON to the website or use the results submitted via WCA Live</li>
-        <li>Then the website will run a number of checks. If everything is correct you'll be able to submit these results to the WRT.</li>
-      </ul>
-    </p>
-    <%= react_component("CompetitionResultSubmission/ImportResultsData", {
-      competitionId: @competition.id,
-      alreadyHasSubmittedResult: @results_validator.any_results?,
-    }) %>
-    <br />
-
-    <%= render "check_results_panel",
-      results_validator: @results_validator,
-      force_collapse: @results_submission
-    %>
-    <% if @current_user.can_submit_competition_results?(@competition) %>
-      <%= react_component("CompetitionResultSubmission/FormToWrt", {
-        competitionId: @competition.id,
-        isErrorInPreviousUpload: @results_validator.any_errors?,
-      }) %>
-    <% end %>
-  <% end %>
-
+  <%= react_component("CompetitionResultSubmission", {
+    competitionId: @competition.id,
+    resultsSubmitted: @competition.results_submitted?,
+    alreadyHasSubmittedResult: @competition.inbox_results.present?,
+    canSubmitResults: @current_user.can_submit_competition_results?(@competition),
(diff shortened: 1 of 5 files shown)


app/views/results_submission/new.html.erb
```diff
@@ -1,35 +1,12 @@
...
-        competitionId: @competition.id,
-        isErrorInPreviousUpload: @results_validator.any_errors?,
-      }) %>
-    <% end %>
-  <% end %>
-
+  <%= react_component("CompetitionResultSubmission", {
+    competitionId: @competition.id,
+    resultsSubmitted: @competition.results_submitted?,
+    alreadyHasSubmittedResult: @competition.inbox_results.present?,
```
gregorbg: The naming of these two parameters is confusing. To a naive developer (ie someone who knows React very well but has not worked with Results code and WRT code before), this looks like both variables kinda do the same thing.

"results submitted" and "has submitted results" sound very, very identical if you turn the variable names into English phrases in your mind.

Please come up with some name to distinguish more clearly what these two values are about.

app/webpacker/components/CompetitionResultSubmission/FormToWrt/index.jsx
```diff
@@ -41,50 +46,55 @@ function FormToWrt({ competitionId, isErrorInPreviousUpload }) {
...
+  if (isValidationPending || isSubmitPending) return <Loading />;
   if (isSuccess) return <Message success>Thank you for submitting the results!</Message>;
   if (isErrorInPreviousUpload) return <Errored error={ERROR_MESSAGE_UPLOADED_RESULTS} />;
   if (isErrorInCurrentUpload) return <Errored error={errorInSubmission} />;
 
   return (
     <Accordion fluid styled>
       <Accordion.Title
-        active={activeAccordion}
+        active
```
gregorbg: Why is this accordion always, unconditionally, active now?

app/webpacker/components/CompetitionResultSubmission/index.jsx
```diff
@@ -0,0 +1,56 @@
...
+      comments please reply to the email sent with the first results submission.
+    </Message>;
+  }
+
+  return (
+    <>
+      The result submission process has two steps:
+      <List bulleted>
+        <List.Item>
+          Uploading a valid JSON to the website or use the results submitted via WCA Live
```
gregorbg: The "use results via WCA Live" feature is not available to the general public yet. Please do not advertise any features that haven't been publicly released.

app/webpacker/components/CompetitionResultSubmission/index.jsx
```diff
@@ -0,0 +1,56 @@
...
+        </List.Item>
+        <List.Item>
+          Submit these results to the WRT after addressing warnings if any.
```
gregorbg:
```suggestion
          Submit these results to the WRT after addressing warnings (if any).
```

app/views/results_submission/new.html.erb
danieljames-dj: Okay changed. I was actually just reusing the existing variable, and didn't notice that they became confusing now. I've anyway changed it to `hasTemporaryResults`, hope it's good now.

app/webpacker/components/CompetitionResultSubmission/FormToWrt/index.jsx
danieljames-dj: Actually now this can be visible every time. Earlier, this was hidden when there was no results. Now this component is shown only when it's actually needed.

danieljames-dj: I had forgot to do the same for `Accordion.Content`. Now fixed that.

app/webpacker/components/CompetitionResultSubmission/index.jsx
danieljames-dj: Okay removed. I just noticed that this was there earlier as well - I added it accidentally earlier.

gregorbg (approved):

PR #12112, opened 2025-07-29
Add validation in uploadJson between registration & inboxPerson

danieljames-dj:
This additional check will make sure that there is a corresponding accepted registration for every entry in inboxPerson.


app/models/upload_json.rb
```diff
@@ -21,6 +21,24 @@ class UploadJson
     else
       errors.add(:results_file, "can't be blank")
     end
+
+    if competition.use_wca_registration?
+      persons_to_import.each do |person|
+        registration = Registration.find_by(competition_id: competition_id, registrant_id: person.id)
```
FinnIckler: Isn't there a way to do this in one query? 
Registration.where(competition_if: competition_id, registrant_id: persons_to_import.pluck(:id)
Out of the loop

gregorbg: Yes, but then you lose the ability to compare every `persons_to_import` uniquely. The point is not only to notice any differences, but to match _which_ person caused those differences.

Your suggestion would make for a cleaner DB query, but then for each of the registrations that you found, you would need to iterate over `persons_to_import` to match it with the (current, while looping) registration.

danieljames-dj: That sounds like a good idea. I can do something like `Registration.where(competition_id: competition_id, registrant_id: persons_to_import.pluck(:id), competing_status: 'accepted')`. I might still want the loop. I've made the change, hope it's better now.

gregorbg: Okay nevermind, Daniel found a good indexing.

PR #12116, opened 2025-07-30
Minor fixes in Newcomer checks page

danieljames-dj:
Following are the fixes done in this PR:
- Typo in DOB checker page
- Show "Merge" button only if the details are exactly same
- Typos in DuplicateCheckerHeader messages
- Show duplicates only if the job is complete (during this change I felt like doing some more related cleanups, did that as well)

### DIFF
--- a/app/webpacker/components/NewcomerChecks/DobChecker.jsx
+++ b/app/webpacker/components/NewcomerChecks/DobChecker.jsx
@@ -17,7 +17,7 @@ export default function DobChecker({ competitionId }) {
   if (isLoading) return <Loading />;
   if (isError) return <Errored error={error} />;
   if (dobChecks.length === 0) {
-    return <Message positive>All DOBs looks good. Thanks for checking.</Message>;
+    return <Message positive>All DOBs are looking good. Thanks for checking.</Message>;
   }
 
   return (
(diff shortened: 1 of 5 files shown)


app/webpacker/components/NewcomerChecks/DobChecker.jsx
```diff
@@ -17,7 +17,7 @@ export default function DobChecker({ competitionId }) {
   if (isLoading) return <Loading />;
   if (isError) return <Errored error={error} />;
   if (dobChecks.length === 0) {
-    return <Message positive>All names looks good. Thanks for checking.</Message>;
+    return <Message positive>All DOBs looks good. Thanks for checking.</Message>;
```
gregorbg:
```suggestion
    return <Message positive>All DOBs are looking good. Thanks for checking.</Message>;
```

app/webpacker/components/NewcomerChecks/SimilarPersonTable.jsx
```diff
@@ -0,0 +1,66 @@
...
+                  {duplicatePerson.wca_id}
+                </a>
+              </Table.Cell>
+              <Table.Cell>{score}</Table.Cell>
+              <Table.Cell>
+                {exactSameDetails && (
+                <Button onClick={() => mergePotentialDuplicate(potentialDuplicatePerson)}>
+                  Merge
+                </Button>
+                )}
```
gregorbg: Just a suggestion for UI: You can always render the button, but use `disabled={!exactSameDetails}`. Also, you should probably show some alternative if the details are _not_ exactly the same, because in some cases it might lead to confusion why the button is disabled/invisible.

gregorbg (changes requested):

app/webpacker/components/NewcomerChecks/SimilarPersonTable.jsx
danieljames-dj: Ya that's a good idea. Done.

PR #12117, opened 2025-07-30
Migrate import registrations to react

danieljames-dj:
This PR aims at migrating the frontend of "import registrations" to react.

### DIFF
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -82,7 +82,7 @@ class RegistrationsController < ApplicationController
 
   def do_import
     competition = competition_from_params
-    file = params[:registrations_import][:registrations_file]
+    file = params[:csv_registration_file]
     required_columns = ["status", "name", "country", "wca id", "birth date", "gender", "email"] + competition.events.map(&:id)
     # Ensure the CSV file includes all required columns.
     headers = CSV.read(file.path).first.compact.map(&:downcase)
@@ -143,11 +143,9 @@ class RegistrationsController < ApplicationController
     new_locked_users.each do |user|
       RegistrationsMailer.notify_registrant_of_locked_account_creation(user, competition).deliver_later
     end
-    flash[:success] = I18n.t("registrations.flash.imported")
-    redirect_to competition_registrations_import_url(competition)
+    render status: :ok, json: { success: true }
   rescue StandardError => e
-    flash[:danger] = e.to_s
-    redirect_to competition_registrations_import_url(competition)
+    render status: :unprocessable_entity, json: { error: e.to_s }
   end
 
   def add
--- a/app/webpacker/components/ImportRegistrations/UploadRegistrationCsv.jsx
+++ b/app/webpacker/components/ImportRegistrations/UploadRegistrationCsv.jsx
@@ -11,7 +11,7 @@ export default function UploadRegistrationCsv({ competitionId, onImportSuccess }
   const {
     mutate: importRegistrationsMutate, isPending, error, isError,
   } = useMutation({
-    mutationFn: () => importRegistrations({ competitionId, csvFile }),
+    mutationFn: importRegistrations,
     onSuccess: onImportSuccess,
   });
 
@@ -19,7 +19,7 @@ export default function UploadRegistrationCsv({ competitionId, onImportSuccess }
   if (isError) return <Errored error={error} />;
 
   return (
-    <Form onSubmit={importRegistrationsMutate}>
+    <Form onSubmit={() => importRegistrationsMutate({ competitionId, csvFile })}>
       <Form.Input
         type="file"
         accept="text/csv"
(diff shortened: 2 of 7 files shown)


app/controllers/registrations_controller.rb
```diff
@@ -143,11 +143,9 @@ def do_import
     new_locked_users.each do |user|
       RegistrationsMailer.notify_registrant_of_locked_account_creation(user, competition).deliver_later
     end
-    flash[:success] = I18n.t("registrations.flash.imported")
-    redirect_to competition_registrations_import_url(competition)
+    render status: :ok, json: { success: true }
   rescue StandardError => e
-    flash[:danger] = e.to_s
-    redirect_to competition_registrations_import_url(competition)
+    render status: :unprocessable_entity, json: { error: e.to_s }
```
gregorbg, line 148: What is the `StandardError` that is being rescued here, and when / under which circumstances can it occur?

app/webpacker/components/ImportRegistrations/index.jsx
```diff
@@ -0,0 +1,43 @@
...
+function ImportRegistrations({ competitionId }) {
+  const [success, setSuccess] = useState();
+  const panes = [
+    {
+      menuItem: 'Upload Registration CSV',
+      render: () => (
+        <Tab.Pane>
+          <UploadRegistrationCsv
+            competitionId={competitionId}
+            onImportSuccess={setSuccess}
```
gregorbg: The way you're passing the success handler here, means that the first argument of the mutation's inner `onSuccess` handler will be used to pass to `setSuccess`.

In other words, the server response (which, by API design of `useMutation`, is the first parameter of `onSuccess`) gets set as the `success` state defined above, and you're "abusing" the fact that the server response is a non-empty object which gets evaluated as "truthy".

I would feel more comfortable if you used some callback like `() => setSuccess(true)`

gregorbg: Or is there a specific reason why you _need_ to capture the server response in the `success` state here?

app/webpacker/components/ImportRegistrations/UploadRegistrationCsv.jsx
```diff
@@ -0,0 +1,38 @@
...
+import Loading from '../Requests/Loading';
+import Errored from '../Requests/Errored';
+import I18n from '../../lib/i18n';
+
+export default function UploadRegistrationCsv({ competitionId, onImportSuccess }) {
+  const [csvFile, setCsvFile] = useState();
+  const {
+    mutate: importRegistrationsMutate, isPending, error, isError,
+  } = useMutation({
+    mutationFn: () => importRegistrations({ competitionId, csvFile }),
```
gregorbg: You're grabbing state from outside the mutation within the mutation function body. This works, but is not good practice. You should pass `competitionId` and `csvFile` as parameters when calling `importRegistrationsMutate` instead.

app/webpacker/components/ImportRegistrations/index.jsx
```diff
@@ -0,0 +1,43 @@
...
+  if (success) {
+    return <Message success>{I18n.t('registrations.flash.imported')}</Message>;
+  }
+
+  return (
+    <>
+      <Message info>
+        {I18n.t('registrations.import.info')}
+      </Message>
+      <Tab panes={panes} />
```
gregorbg, line 40: Why are you rendering a `Tab` with only one single pane?

gregorbg (changes requested):

app/controllers/registrations_controller.rb
danieljames-dj: Actually in the above code, there are lots of `raise` for different cases, and all of them are reaching here.

app/webpacker/components/ImportRegistrations/UploadRegistrationCsv.jsx
danieljames-dj: Okay changed

app/webpacker/components/ImportRegistrations/index.jsx
danieljames-dj: I'll be having more tab panes when there are other ways to import registrations - like WCIF. Or should I add tab only when there is second option? The UI with just one tab is also actually looking fine.

app/webpacker/components/ImportRegistrations/index.jsx
danieljames-dj: Okay, changed to `() => setSuccess(true)`. I am having the `success` handler here because I actually want to show common 'success' for each types of import.

app/controllers/registrations_controller.rb
gregorbg: Okay, I'll accept it for the sake of getting this migration done, but in the future we definitely need to refactor this into some `lib` code that works without `raise`s 🙃

app/webpacker/components/ImportRegistrations/index.jsx
gregorbg: Okay, if it looks alright then it's fine by me

gregorbg (approved):

app/controllers/registrations_controller.rb
danieljames-dj: Thank you. I'll definitely be refactoring the backend and remove all the `raise`s (not sure yet whether I'll use `lib` or not, I haven't thought of it yet).

PR #12127, opened 2025-07-31
Create newcomers page inside ticket

danieljames-dj:
The current "create newcomers" step of results submission is really big - by big I mean lots of things are happening there - like verifying the newcomer data, creating WCA ID for them and assigning them to the corresponding results. But since the core thing happening here is "creation of WCA IDs", I'm naming this step as "created_wca_ids". Pretty sure that this step is going to change a lot in future, at that time maybe we can revisit regarding the name.


app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/CreateWcaIds.jsx
```diff
@@ -0,0 +1,67 @@
...
+        </List.Item>
+        <List.Item>
+          {resultNoWcaIdCount}
+          {' '}
+          competitors in the Results table for this competition still have to be assigned a WCA ID.
+        </List.Item>
+      </List>
+      <List>
+        <List.Item>
+          <Button primary href={viewUrls.admin.completePersons([competitionId])} target="_blank">
```
gregorbg: Does an `href` directly on a button even work? Feels very wrong, I don't think it should be like that :O

Probably more like `<Link as={Button} ... />` or similar. How do we deal with this elsewhere?

danieljames-dj: Unfortunately there is nothing like `Link` in React SUI (I might be wrong, if so please correct me). And this is working fine. We have used `href` directly in the button at a few places including delegates page.

gregorbg: Okay, what about `<Button as="a" ... />`?

danieljames-dj: Yes that works the same way. Changed.

PR #12131, opened 2025-07-31
Adding `status` to RegistrationPayment

dunkOnIT:


app/models/payment_intent.rb
```diff
@@ -7,6 +7,8 @@ class PaymentIntent < ApplicationRecord
   belongs_to :confirmation_source, polymorphic: true, optional: true
   belongs_to :cancellation_source, polymorphic: true, optional: true
 
+  has_one :registration_payment
```
gregorbg: This most certainly has to become a `has_many`. All major providers that we integrate currently model their APIs so that one intent can have many charges. And receipts in the reg payments table are usually linked to charges. So reg payments to receipt charges is 1:1, but receipt charges to their parent intent is 1:N. By extension, the new relation from reg payment to the parent intent also needs to be 1:N

app/models/registration_payment.rb
```diff
@@ -22,6 +23,10 @@ def amount_available_for_refund
     amount_lowest_denomination + refunding_registration_payments.sum(:amount_lowest_denomination)
   end
 
+  def status
```
gregorbg: `status` is a bit too generic IMO.

db/migrate/20250801074513_link_payment_intent_and_reg_payment.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class LinkPaymentIntentAndRegPayment < ActiveRecord::Migration[7.2]
+  def change
+    add_column :registration_payments, :payment_intent_id, :integer
```
gregorbg: There should be an `add_reference` if I'm not mistaken. Maybe the name is slightly different, but in general Rails mutations do have comprehensive support for relations and relation bindings. The documentation will know more.

lib/tasks/migration_utilities.rake
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+namespace :migration_utilities do
```
gregorbg: Too generic. The namespace should be something like `registration_payments` and the task name can be "migrate payment intents" or "backfill payment intent ID" or something the likes.

dunkOnIT: I was wondering whether we want to, instead of creating separate files that tend to have one-off scripts for data migration, we want to group them together under one namespace?

app/models/payment_intent.rb
dunkOnIT: Done

app/models/registration_payment.rb
dunkOnIT: My general thinking is that RegistrationPayment is on the user's side of the "veil of blissful ignorance", and isn't used alongside PaymentRecords or PaymentIntents as frequently - so there's less need to differentiate between (for example) `wca_status` and `stripe_status` for readability's sake. 

This means that the usage will probably end up looking like `.registration_payment.status` a lot of the time, or some variant of `{variable_name}.status`, where the context of what is invoking `.status` make it clear the purpose of the status.

If you'd still prefer specificity we can do that, though: 
- `registration_payment_status` is quite a mouthful
- `payment_status` might be simpler, and appropriate in the contexts where we tend to use RegistrationPayment?

db/migrate/20250801074513_link_payment_intent_and_reg_payment.rb
dunkOnIT: Done

app/models/payment_intent.rb
```diff
@@ -7,6 +7,8 @@ class PaymentIntent < ApplicationRecord
   belongs_to :confirmation_source, polymorphic: true, optional: true
   belongs_to :cancellation_source, polymorphic: true, optional: true
 
+  has_many :registration_payment
```
gregorbg: Nitpick, but important for Rails AR: By convention, `has_many` names should be in plural. I think (but only 40% sure) that this will fail to resolve in its current form.

app/models/registration_payment.rb
gregorbg: payment_status seems like a good middle ground.

db/schema.rb
gregorbg, line 1: Now contains unrelated changes from your other PR

lib/tasks/migration_utilities.rake
gregorbg: We already went quite far down the "one-off files" path, so I'd rather stick to our guns there.

gregorbg (changes requested):

app/models/competition.rb
```diff
@@ -2996,6 +2996,7 @@ def self.delegate_edits_json_schema
   def fully_paid_registrations_count
     registrations
       .joins(:registration_payments)
+      .where(registration_payments: { is_completed: true })
```
gregorbg: Try whether `.merge(RegistrationPayment.completed)` works

app/models/registration_payment.rb
```diff
@@ -11,23 +11,43 @@ class RegistrationPayment < ApplicationRecord
...
+  scope :completed, -> { where(is_completed: true) }
 
   monetize :amount_lowest_denomination,
            as: "amount",
            allow_nil: true,
            with_model_currency: :currency_code
 
+  def should_auto_close?
+    refunded_registration_payment_id.nil? && saved_change_to_is_completed?(to: true)
+  end
```
gregorbg: Where is this code ever being called?

app/models/registration_payment.rb
```diff
@@ -11,23 +11,43 @@ class RegistrationPayment < ApplicationRecord
...
 
   private def auto_accept_hook
     registration.attempt_auto_accept(:live)
   end
 
   private def auto_close_hook
-    registration.consider_auto_close
+    return unless refunded_registration_payment_id.nil? && is_completed?
+
+    registration.consider_auto_close if saved_change_to_is_completed? || previously_new_record?
```
gregorbg, line 43: Why are all of these additional checks becoming relevant to `auto_close_hook`, but not to `auto_accept_hook`?

app/models/registration_payment.rb
```diff
@@ -11,23 +11,43 @@ class RegistrationPayment < ApplicationRecord
...
   monetize :amount_lowest_denomination,
            as: "amount",
            allow_nil: true,
            with_model_currency: :currency_code
 
+  def should_auto_close?
+    refunded_registration_payment_id.nil? && saved_change_to_is_completed?(to: true)
+  end
+
+  def create_uncaptured_payment
```
gregorbg: Where is this method being used, and why are you creating a method in `RegistrationPayment` to create a new `RegistrationPayment`?

app/models/registration_payment.rb
dunkOnIT: Leftover from before I moved the logic to `auto_close_hook` - now removed

app/models/registration_payment.rb
dunkOnIT: I honestly have no idea - removed

app/models/registration_payment.rb
dunkOnIT: Our logic around whether auto accept should trigger is much more thoroughly fleshed out in `attempt_auto_accept`, which is desirable because it lets us do much more granular logging around _why_ a registration didn't get auto-accepted. That logging is important if there's every confusion/dispute around why a certain registration wasn't auto-accepted, which could lead to someone losing a competing slot.

There's an argument that we should do this for auto-close too - but we don't currently have a history feature to take advantage of the way we do with registrations 

Lastly - there is perhaps an argument that we could _at least_ only trigger auto_accept if the payment `is_completed`, but: 
-  even in this case, I'd rather explicitly log that auto accept was tried and failed, which would let us investigate potential issues
- `outstanding_payments` is a check in `attempt_auto_accept`, which relies on paid_entry_fees, which is aware of completed payments 

That said, I do think we should have a test that explicitly confirms that an `is_completed: false` registration_payment triggers a failed attempt_auto_accept - which I've added

spec/requests/api_registrations_spec.rb
gregorbg, line 1: Just curious: What has changed about these tests that they were working with the `0` expectation before but need a `1000` expectation now?

dunkOnIT: The expectation was plain wrong[1] (ie, we should not expect outstanding fees to be 0 after a refund), but not being picked up because we needed to do `registration.reload` to actually pick up the refund. I'm not sure why the change to `paid_entry_fees` in this PR means that it no longer needs this reload, but I assume it has to do with the `.completed` scope not being loaded into memory, where previously it was just using the already in-memory `registration_payment` data

[1] I pretty much always try and force a failure in the test case before making the changes necessary for it to succeed, to avoid situations like this one. Not sure how it slipped through the cracks in this particular instance - at a guess, copy-pasting a bunch of similar test cases and not being careful enough to force failure cases

app/models/registration_payment.rb
```diff
@@ -11,23 +11,27 @@ class RegistrationPayment < ApplicationRecord
 
   delegate :auto_accept_preference_live?, to: :registration
   after_create :auto_accept_hook, if: :auto_accept_preference_live?
```
gregorbg, line 14: Shouldn't this also become a `after_save` then?

app/models/registration_payment.rb
```diff
@@ -11,23 +11,27 @@ class RegistrationPayment < ApplicationRecord
...
-  after_create :auto_close_hook, unless: :refunded_registration_payment_id?
+  after_save :auto_close_hook
```
gregorbg, line 15: And as an afterthought to my above comment, what about an `after_create ..., if: :is_completed?` as well as a separate `after_update ..., if: :saved_change_to_is_completed?`

app/models/registration_payment.rb
```diff
@@ -11,23 +11,27 @@ class RegistrationPayment < ApplicationRecord
...
+  scope :completed, -> { where(is_completed: true) }
 
   monetize :amount_lowest_denomination,
            as: "amount",
            allow_nil: true,
            with_model_currency: :currency_code
 
   def amount_available_for_refund
-    amount_lowest_denomination + refunding_registration_payments.sum(:amount_lowest_denomination)
+    amount_lowest_denomination + refunding_registration_payments.completed.sum(:amount_lowest_denomination)
```
gregorbg: Can this create awkward edge cases? I'm thinking:
- There is a full, completed and confirmed registration payment of 30 USD
- Organizer A initiates a refund of 20 USD because of a reduced entry fee policy for RO peepz.
- The refund takes time to go through on Stripe's side because Stripe has to negotiate something with the bank of the user who originally paid
- Half a day later, organizer B checks whether there are any people to refund thanks to the same RO reduced entry fee policy as before
- The original refund from organizer A in the morning still hasn't gone through, so it's not recorded as `completed` yet in our database (because Stripe is still arguing with the bank that issued the credit card of the original payment)
- Organizer B sees an "available" amount of 30 USD
- Organizer B initiates the refund again, **which is wrong**

app/models/registration_payment.rb
gregorbg: You said:
> That said, I do think we should have a test that explicitly confirms that an is_completed: false registration_payment triggers a failed attempt_auto_accept - which I've added

which is interesting because my comment revolves around `attempt_auto_close`. Do we have sufficient similar test coverage there?

spec/models/registration_payments_spec.rb
```diff
@@ -21,5 +21,10 @@
       create(:registration_payment, :refund, registration: registration)
       expect(registration_payment.amount_available_for_refund).to eq(0)
     end
+
+    it 'does not included is_completed: false refunds' do
+      create(:registration_payment, :refund, is_completed: false, registration: registration)
+      expect(registration_payment.amount_available_for_refund).to eq(1000)
```
gregorbg: This can create "false positive" situations in the UI, see my lengthy comment above

gregorbg: So I'm really not sure whether this `eq(1000)` should even be the desired assumption here. You will see what I mean when you go through my example case

spec/models/registration_spec.rb
gregorbg, line 1: Yikes about all the `sleep`!! 🤮 I assume that this is "necessary" because we rely on the `created_at` for ordering payments. And now you know why I always preach that _any_ business logic relying on Rails' auto-generated `created_at` and `updated_at` is a horrible, horrible idea.

At the risk of scope creep, I think this necessitates an extra `paid_at` column which we can manually control during tests (by simply passing in desired values in the `FactoryBot` calls)

gregorbg (changes requested):

app/models/registration_payment.rb
dunkOnIT: Not until we want to add Live AA support for manual payments I don't think - which I'm hoping will be not too many PR's away

app/models/registration_payment.rb
dunkOnIT: What do you see us gaining from that? My feeling was that given how many conditions are coming into play, it was easier/cleaner to just group them all together in one function

app/models/registration_payment.rb
dunkOnIT: We do, see:
- competition_spec:1917 (which auto_close relies on) 
- registration_spec:1243

app/models/registration_payment.rb
dunkOnIT: I'm pretty sure I agree, will take a bit more time to think it through

spec/models/registration_spec.rb
dunkOnIT: I agree, especially given that a payment can now be completed after it's created. I'm going to create a separate, pre-requisite PR just to keep things a bit clean

dunkOnIT: Actually I kept paid_at in this PR given how linked to is_completed its functionality is (still need to update the existing tests though)

dunkOnIT: Tests updated - there are now a bunch more sleeps, but that is to support testing on the implementation of `paid_at`. Now that we're using `paid_at`, we have `ms` precision and don't need to sleep in the tests which initially raised this concern :P

app/models/registration_payment.rb
```diff
@@ -10,14 +10,25 @@ class RegistrationPayment < ApplicationRecord
...
            as: "amount",
            allow_nil: true,
            with_model_currency: :currency_code
 
+  private def becoming_completed?
+    is_completed && (will_save_change_to_is_completed? || new_record?)
+  end
+
+  private def set_paid_at
+    self.paid_at = Time.now.utc unless self.paid_at.present?
```
gregorbg:
```suggestion
    self.paid_at = current_time_from_proper_timezone unless self.paid_at.present?
```

This is a Rails helper that is also used when calling `touch` on a model

gregorbg: Also, the `unless` part can probably move directly into the `before_save` part. Something along the lines of `unless: :paid_at?`

spec/models/registration_spec.rb
```diff
@@ -1331,4 +1368,121 @@
...
+      expect(@expected_pmt.created_at).to be < @other1.created_at
+      expect(@expected_pmt.updated_at).to be < @other1.created_at
+      expect(@expected_pmt.created_at).to be < @other1.updated_at
+      expect(@expected_pmt.updated_at).to be < @other1.updated_at
+    end
+
+    describe '#last_payment' do
+      context 'payments are loaded' do
+        before do
+          registration.registration_payments.load
```
gregorbg, line 1443: What is this?

gregorbg (changes requested):

spec/models/registration_spec.rb
dunkOnIT: I'm testing the behaviour of `last_payment`, which has branching logic based on whether or not payments are loaded into memory: 
```ruby
def last_payment
    if registration_payments.loaded?
      registration_payments.completed.max_by(&:paid_at)
    else
      registration_payments.completed.order(:paid_at).last
    end
  end
```

app/models/registration_payment.rb
```diff
@@ -10,14 +10,25 @@ class RegistrationPayment < ApplicationRecord
   has_many :refunding_registration_payments, class_name: 'RegistrationPayment', inverse_of: :refunded_registration_payment, foreign_key: :refunded_registration_payment_id, dependent: :destroy
 
   delegate :auto_accept_preference_live?, to: :registration
+  before_save :set_paid_at, if: -> { becoming_completed? && paid_at.blank? }
```
gregorbg:
```suggestion
  before_save :set_paid_at, if: :becoming_completed?, unless: :paid_at?
```

dunkOnIT: Updated

app/models/registration_payment.rb
dunkOnIT: Updated

db/migrate/20250828075341_add_paid_at_to_registration_payment.rb
```diff
@@ -0,0 +1,16 @@
...
+
+class AddPaidAtToRegistrationPayment < ActiveRecord::Migration[7.2]
+  def change
+    add_column :registration_payments, :paid_at, :datetime
+
+    reversible do |direction|
+      direction.up do
+        execute <<~SQL.squish
+          UPDATE registration_payments
+          SET paid_at = created_at
```
gregorbg, line 11: What about some `WHERE paid_at IS NULL` just for good measure? I don't know how our prod server is configured, but some MySQL versions reject general `UPDATE` without `WHERE` statements outright

dunkOnIT: Done

PR #12132, opened 2025-07-31
Next js regulations

FinnIckler:
This requires removing `<% provide(:title, 'WCA Regulations') %>`  and saving it as an html file once we are done with the migration


next-frontend/src/app/(wca)/regulations/history/official/[version]/page.tsx
```diff
@@ -11,7 +11,7 @@ export default function HistoricalRegulation() {
       <AspectRatio>
         <iframe
           width={"100%"}
-          src={`https://regulations.worldcubeassociation.org/history/official/${params.version}/index.html`}
```
Comment (FinnIckler) [inline next-frontend/src/app/(wca)/regulations/history/official/[version]/page.tsx]: I removed the /index.html because that is the default file that cloudfront looks for in the s3 bucket

FinnIckler:
Recreating with new spec

PR #12135, opened 2025-08-02
Remove Board permission for older TNoodle usage

ohexter:
The existing text is not in line with current practice. The Board have agreed with their removal from this page.


FinnIckler:
Weird, that was done automatically with the linter, I'll fix it

PR #12158, opened 2025-08-04
My competitions nextjs

FinnIckler:
Haven't actually done any design yet. This was just a 1:1 conversion with Components.


next-frontend/src/app/(wca)/competitions/mine/page.tsx
gregorbg, line 1: I don't see why this needs to be done as client side component. The AuthJS session can be retrieved server-side as well, and the only other thing is a useMemo which in server-side can just be a normal statement (with `await` if necessary)

next-frontend/src/components/competitions/Mine/PastCompetitionTable.tsx
gregorbg, line 1: Again, I'm rather certain this can be an async server component.

next-frontend/src/components/competitions/Mine/TableCells.tsx
```diff
@@ -0,0 +1,95 @@
...
+export function ReportTableCell({
+  competitionId,
+  isReportPosted,
+  isPastCompetition,
+}: ReportTableCellProps) {
+  const {
+    canViewDelegateReport,
+    canEditDelegateReport,
+    canAdministerCompetition,
+  } = usePermissions()!;
```
gregorbg, line 61: I vaguely remember drafting up a `usePermissions` equivalent for server-side rendering. Not sure if we merged that one already though, please take a look.

gregorbg (changes requested):
WST <3 SSR

FinnIckler:
Just adding a Container and a VStack is already making it look pretty decent.
image
Just a few adjustments in the tables needed

FinnIckler:
image
image
I made the table horizontally scrollable, I think that's better? Will need community feedback

next-frontend/src/app/(wca)/competitions/mine/page.tsx
FinnIckler: well, it's a page that different for everyone, so there is no need for it to be a server component no? No SEO considerations and no reusing of server rendering because it's different for everyone.

PR #12162, opened 2025-08-05
Move abort results to ticket

danieljames-dj:
WRT (very rarely) abort results in case some issues are there with submission and ask the Delegate to resubmit.

Currently this flow is apparently not working because I missed to consider this during the recent works, but luckily we didn't had any case of abort so far.

I've also changed the behavior of how results can be aborted. I feel this current way is bit more cleaner and easy to understand.


app/controllers/results_submission_controller.rb
```diff
@@ -172,13 +172,24 @@ def create
 
     return render status: :unprocessable_entity, json: { error: "Submitted results contain errors." } if results_validator.any_errors?
 
-    return render status: :unprocessable_entity, json: { error: "There is already a ticket associated, please contact WRT to update this." } if competition.result_ticket.present?
+    if competition.tickets_competition_result.present? && !competition.tickets_competition_result.aborted?
+      return render status: :unprocessable_entity, json: {
+        error: "Cannot submit results for this competition as the results ticket is not aborted.",
```
gregorbg: I don't get this. "cannot submit results because the ticket is not aborted" confuses me, because the logical conclusion is that "can only submit results if the ticket is aborted"??!
What am I missing?

danieljames-dj: Okay, I've rephrased the text to be less confusing.

app/controllers/results_submission_controller.rb
```diff
@@ -172,13 +172,24 @@ def create
 
     return render status: :unprocessable_entity, json: { error: "Submitted results contain errors." } if results_validator.any_errors?
 
-    return render status: :unprocessable_entity, json: { error: "There is already a ticket associated, please contact WRT to update this." } if competition.result_ticket.present?
+    if competition.tickets_competition_result.present? && !competition.tickets_competition_result.aborted?
+      return render status: :unprocessable_entity, json: {
+        error: "There is already a ticket associated with this, hence the results can be submitted only if the ticket is in abort state.",
```
gregorbg: Okay, now it makes more sense. To whom is this message usually shown?
If Delegates see this warning message, then they probably won't know what "ticket is in an abort state" means. What about something like "... can be resubmitted only if the [previous] posting has been cancelled by WRT"?

I also find that "abort" is a bit of a strong word, at least in outside communication. If you use this term in WRT internally, then it's fine to call it like that for the tickets enum, but the communication to Delegates does not forcibly need to use the exact same vocabulary.

danieljames-dj: This message actually won't reach anywhere because the UI won't allow to submit results if there is a ticket in non-aborted state.

Your wording actually makes more sense. I've now modified it.

app/controllers/admin_controller.rb
```diff
@@ -36,20 +36,6 @@ def check_competition_results
     @competition = competition_from_params
   end
 
-  def clear_results_submission
-    # Just clear the "results_submitted_at" field to let the Delegate submit
-    # the results again. We don't actually want to clear InboxResult and InboxPerson.
-    @competition = competition_from_params
-
-    if @competition.results_submitted? && !@competition.results_posted?
-      @competition.update(results_submitted_at: nil)
```
gregorbg: Where did these checks go? Does `results_submitted_at` still play a role in your new tickets system?

gregorbg (commented):
You have a merge conflict now, please resolve.

app/controllers/admin_controller.rb
danieljames-dj: Okay after some thoughts I made some changes now.

I was attempting to not reset `results_submitted_at` when the results are cleared, instead just aborting the ticket which allows the Delegate to resubmit. But I see that this requires much more changes - like results-not-submitted nags are depending on `results_submitted_at`, similarly posting dashboard is using the same, etc. So I feel it's too early to go that way.

I've now brought this method back and is reusing this to clear results submission. Sorry for causing confusions.

gregorbg (approved):

PR #12164, opened 2025-08-06
Records API for next

FinnIckler:
This PR adds an API (+ a basic page using it) for the Next Records Page. It's a mix out of the current `/records` v0 API and the old API that is used for the records page. 

This API can be used for the mixed, separate and slim views, but for history I will build add an additional one. Probably `results/records/history`.

Because we want to get away from the `#` kind of routing, we probably don't want a state on the page, but just links to `/results/records/333/germany/slim` or something? 

I will try this out and see if this is feasible.


FinnIckler:
All the views:
image
slim:
image
Separate:
image
history:
image
mixed history:
image

PR #12166, opened 2025-08-07
Fix casing of labels in ticket timeline

pranav-027:
Is there any other way to implement this ?

cc: @danieljames-dj


danieljames-dj:
One other solution is to add it in en.yml. But I don't think we need translation for internal facing website.

@gregorbg Can you please provide your input on the same?

Also, as an information, there is a plan to move the `status` field to `tickets` table (discussion [here](https://github.com/thewca/worldcubeassociation.org/pull/12075#discussion_r2236124433)). In case we go with en.yml, that change will again cause some changes in translations and hence cause extra work to translators.

pranav-027:
ig we are good to merge then @danieljames-dj ?

danieljames-dj (approved):

PR #12177, opened 2025-08-11
Add comments in registration csv

pranav-027:
Helpful in FMC simul locations and state championships

### DIFF
--- a/app/models/registration.rb
+++ b/app/models/registration.rb
@@ -299,7 +299,7 @@ class Registration < ApplicationRecord
       registrant_id: registrant_id,
       competing: {
         event_ids: event_ids,
-        comments: comments || "",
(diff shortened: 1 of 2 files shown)


app/models/registration.rb
```diff
@@ -299,6 +299,7 @@ def to_v2_json(admin: false, pii: false)
       registrant_id: registrant_id,
       competing: {
         event_ids: event_ids,
+        comments: comments || "",
```
gregorbg: Why is the `|| ""` fallback necessary here? If someone did not enter a comment, then representing such a (non-existent) comment as `null` totally makes sense

pranav-027: Makes sense! Fixed

PR #12208, opened 2025-08-11
Actually use stylistic package for linting

gregorbg:
We had installed the `stylistic` linter in our dependencies for quite a while, but we never correctly used it. We registered a plugin namespace for `@stylistic`, but we never actually activated any _rules_ for that namespace.

Instead, this PR introduces the recommended ruleset (which in itself also encapsulates the `plugins` definition, so our manual "useless" definition can go).

As a side-effect, Prettier and Stylistic have different opinions for example when it comes to single-quotes VS double-quoutes. Luckily, there is `eslint-config-prettier`, which configures Prettier to "silence" all other linter rules that would override its preferences (fun fact: We also had already installed this in our dependencies but never correctly used it). So with regards to Prettier, we don't destroy anything, and we only "gain" the good parts from Stylistic.


FinnIckler (approved):
oops...

PR #12214, opened 2025-08-12
Add competition name to result ticket

danieljames-dj:
This is to let WRT know which competition they are posting, this is useful when they are working on multiple results in parallel.

### DIFF
--- a/app/models/tickets_competition_result.rb
+++ b/app/models/tickets_competition_result.rb
@@ -70,7 +70,9 @@ class TicketsCompetitionResult < ApplicationRecord
   end
 
   DEFAULT_SERIALIZE_OPTIONS = {
-    include: %w[competition],
+    include: {
+      competition: { only: %i[id name], methods: [], include: [] },
+    },
   }.freeze
 
   def serializable_hash(options = nil)
--- a/app/models/tickets_competition_result.rb
+++ b/app/models/tickets_competition_result.rb
@@ -13,8 +13,7 @@ class TicketsCompetitionResult < ApplicationRecord
   }
 
   has_one :ticket, as: :metadata
-  belongs_to :competition
-  delegate :name, to: :competition, prefix: true
+  belongs_to :competition, -> { select(:id, :name) }
 
   def actions_allowed_for(ticket_stakeholder)
     if ticket_stakeholder.stakeholder == UserGroup.teams_committees_group_wrt
@@ -71,7 +70,9 @@ class TicketsCompetitionResult < ApplicationRecord
   end
 
   DEFAULT_SERIALIZE_OPTIONS = {
-    methods: %w[competition_name],
+    include: {
+      competition: { only: %i[id name], methods: [], include: [] },
+    },
   }.freeze
 
   def serializable_hash(options = nil)
--- a/app/models/tickets_competition_result.rb
+++ b/app/models/tickets_competition_result.rb
@@ -14,6 +14,7 @@ class TicketsCompetitionResult < ApplicationRecord
 
   has_one :ticket, as: :metadata
   belongs_to :competition
+  delegate :name, to: :competition, prefix: true
 
   def actions_allowed_for(ticket_stakeholder)
     if ticket_stakeholder.stakeholder == UserGroup.teams_committees_group_wrt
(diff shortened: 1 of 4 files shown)


app/webpacker/components/Tickets/TicketHeader/index.jsx
```diff
@@ -18,3 +22,16 @@ export default function TicketHeader({ ticketDetails, currentStakeholder, update
     </Card>
   );
 }
+
+function Heading({ ticketDetails }) {
+  const { ticket: { metadata_type: ticketType, metadata } } = ticketDetails;
+
+  switch (ticketType) {
+    case ticketTypes.edit_person:
+      return 'Edit Person';
```
gregorbg: Why do `edit_person` tickets not have specific titles?
Or in other words, why is it okay to show `"Edit Person"` for this type, but not okay to just show `"Competition Results"` for the new `competition_results` type?

app/models/tickets_competition_result.rb
```diff
@@ -68,4 +69,12 @@ def merge_inbox_results
       self.update!(status: TicketsCompetitionResult.statuses[:merged_inbox_results])
     end
   end
+
+  DEFAULT_SERIALIZE_OPTIONS = {
+    methods: %w[competition_name],
```
gregorbg: This will call/fetch the `competition` association upon every serialization. Are you sure that you pre-loaded via `TicketsCompetitionResult.includes(:competition)`  in all relevant places?

danieljames-dj: Oh I didn't see that the whole `competition` was fetched. I've now fixed it.

app/webpacker/components/Tickets/TicketHeader/index.jsx
danieljames-dj: For competition results, the title helps to identify which competition's ticket are we working on. Hence the competition name in title is necessary. While for edit person, WRT doesn't need any additional information in the title, so giving a generic title. We anyway can know the name of the requestor and other details from the workbench.

gregorbg: Okay, then the follow-up question is: Why does Edit Person have all necessary information in the body, and Post Results doesn't?

It seems that for both ticket types (edit person and post results) there is some kind of "main" information or "identifying" information:
- Edit Person: The person being edited
- Post Results: The competition for which results are being posted

In one case, this "main" information appears in the header, and in the other case, it can appear in the body. That's very inconsistent, why?

gregorbg:
The fix for competition name that you came up with is way too specific. It's not easily maintainable (if you need additional fields in the future) and it's also hard to trace bugs because of the `select` projection directly within the `belongs_to`.

The previous way to do things was fine (the "boring" `belongs_to`) but you just need to make sure that the competition is included via `includes` at the point where the ticket is originally loaded from the DB

app/webpacker/components/Tickets/TicketHeader/index.jsx
danieljames-dj: Okay, I understand that this is inconsistent. I revisited the edit person page and I felt that WCA ID is a necessary requirement. It's currently shown as part of `EditPersonForm`, but that form will soon be removed (I didn't discuss the alternative with you yet, I'll definitely discuss with you once I finish off the higher priority tasks).

For now, for consistency, I've added both generic heading as well as required information (as subheading). Hope this sounds good.

gregorbg (changes requested):
Well, loading the whole competition is okay but _serializing_ the whole thing is over the top.

When I said "the way you did before", then I was talking about the `delegate` approach.

Also, have you made sure that the competition is pre-loaded using `includes` or similar?

danieljames-dj:
> Well, loading the whole competition is okay but _serializing_ the whole thing is over the top.
> 
> When I said "the way you did before", then I was talking about the `delegate` approach.
> 
> Also, have you made sure that the competition is pre-loaded using `includes` or similar?

Oh okay, I didn't do that way as it's not possible with polymorphic association, or did I miss anything?

danieljames-dj:
> > Oh okay, I didn't do that way as it's not possible with polymorphic association, or did I miss anything?
> 
> Can you please point me to the correct controller method where the ticket is being loaded?

https://github.com/thewca/worldcubeassociation.org/blob/9963bbd2ebc2a3bdb6b0ea6d3a2d7415bfb3f89e/app/controllers/tickets_controller.rb#L57

gregorbg:
Thanks, in that case if you're accessing the ticket through a generic `ticket_id` then preloading is indeed tricky.

The thing that I definitely don't want is for you to serialize the _whole_ competition into the metadata, because that will also include Delegates, their avatars, etc. The original `delegates :name, to: :competition` approach should be fine.

danieljames-dj:
Okay, I modified the serialization to only include `id` and `name`. Didn't use `delegate` because I need both `id` and `name` now to provide hyperlink. Hope it's fine.

gregorbg (approved):

PR #12215, opened 2025-08-12
Move some reg import errors to before_action

danieljames-dj:
This PR aims at doing some minor cleanups in registration import. This PR won't clean up the registration import completely - that means, there will still be some `raise` even after this PR is merged. Remaining fixes will be done in upcoming PRs.

### DIFF
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -95,7 +95,7 @@ class RegistrationsController < ApplicationController
     event_column_errors = csv_rows.filter_map do |row|
       competition_events.find do |competition_event|
         cell_value = row[competition_event.event_id.to_sym]
-        next unless %w[1 0].exclude?(cell_value)
+        next if %w[1 0].include?(cell_value)
 
         I18n.t(
           "registrations.import.errors.invalid_event_column",
@@ -105,19 +105,29 @@ class RegistrationsController < ApplicationController
       end
     end
 
-    raw_dobs = csv_rows.pluck(:birth_date)
-    wrong_format_dobs = raw_dobs.reject { |raw_dob| Date.safe_parse(raw_dob)&.to_fs == raw_dob }
-    dob_column_error = I18n.t("registrations.import.errors.wrong_dob_format", raw_dobs: wrong_format_dobs.join(", ")) if wrong_format_dobs.any?
+    dob_column_error = column_check(csv_rows, :birth_date, 'wrong_dob_format', :raw_dobs) do |raw_dob|
+      Date.safe_parse(raw_dob)&.to_fs != raw_dob
+    end
 
-    @emails = csv_rows.pluck(:email)
-    email_duplicates = @emails.select { |email| @emails.count(email) > 1 }.uniq
-    email_duplicate_error = I18n.t("registrations.import.errors.email_duplicates", emails: email_duplicates.join(", ")) if email_duplicates.any?
+    email_duplicate_error = column_check(csv_rows, :email, 'email_duplicates', :emails) do |email, emails|
+      emails.count(email) > 1
+    end
 
-    wca_ids = csv_rows.pluck(:wca_id).reject(&:empty?)
-    wca_id_duplicates = wca_ids.select { |wca_id| wca_ids.count(wca_id) > 1 }.uniq
-    wca_id_duplicate_error = I18n.t("registrations.import.errors.wca_id_duplicates", wca_ids: wca_id_duplicates.join(", ")) if wca_id_duplicates.any?
+    wca_id_duplicate_error = column_check(csv_rows, :wca_id, "wca_id_duplicates", :wca_ids) do |wca_id, wca_ids|
+      wca_id.present? && wca_ids.count(wca_id) > 1
+    end
 
     [event_column_errors, dob_column_error, email_duplicate_error, wca_id_duplicate_error].flatten
+  end
+
+  private def column_check(csv_rows, column_name, error_key, i18n_keyword)
+    column_values = csv_rows.pluck(column_name)
+
+    malformed_values = column_values.select do |value|
+      yield value, column_values
+    end.uniq
+
+    I18n.t("registrations.import.errors.#{error_key}", i18n_keyword => malformed_values.join(", ")) if malformed_values.any?
   end
 
   private def competitor_limit_error(competition, competitor_count)
@@ -178,9 +188,10 @@ class RegistrationsController < ApplicationController
     #   from CSV import to be considered as one "batch". So we mark a timestamp
     #   once, and then reuse it throughout the loop.
     import_time = Time.now.utc
+    emails = @registration_rows.pluck(:email)
     ActiveRecord::Base.transaction do
       @competition.registrations.accepted.each do |registration|
-        registration.update!(competing_status: Registrations::Helper::STATUS_CANCELLED) unless @emails.include?(registration.user.email)
+        registration.update!(competing_status: Registrations::Helper::STATUS_CANCELLED) unless emails.include?(registration.user.email)
       end
       @registration_rows.each do |registration_row|
         user, locked_account_created = user_for_registration!(registration_row)
--- a/spec/requests/registrations_spec.rb
+++ b/spec/requests/registrations_spec.rb
@@ -428,6 +428,19 @@ RSpec.describe "registrations" do
             end.to change { competition.registrations.count }.by(1)
           end
         end
+
+        context "CSV empty file" do
+          it "throws an error" do
+            file = csv_file [
+              ["Status", "Name", "Country", "WCA ID", "Birth date", "Gender", "Email", "333", "444"],
+            ]
+            expect do
+              post competition_registrations_do_import_path(competition), params: { csv_registration_file: file }
+            end.not_to(change { competition.registrations.count })
+            expect(response).to have_http_status(:unprocessable_entity)
+            expect(response.body).to include "The file is empty."
+          end
+        end
       end
     end
   end


app/controllers/registrations_controller.rb
```diff
@@ -34,6 +34,53 @@ class RegistrationsController < ApplicationController
     redirect_to competition_path(competition_from_params) if competition_from_params.use_wca_registration?
   end
 
+  before_action :check_errors_import_registration!, only: %i[do_import]
+  private def check_errors_import_registration!
+    @competition = competition_from_params
+    file = params[:csv_registration_file]
+
+    competition_events = @competition.competition_events
+    required_event_columns = competition_events.map(&:event_id)
```
gregorbg:
```suggestion
    required_event_columns = competition_events.pluck(:event_id)
```

If you're working on an ActiveRecord relation, `pluck` is always your friend for some individual field

app/controllers/registrations_controller.rb
```diff
@@ -34,6 +34,53 @@ class RegistrationsController < ApplicationController
...
+    required_other_columns = ["status", "name", "country", "wca id", "birth date", "gender", "email"]
+
+    headers = CSV.read(file.path).first.compact.map(&:downcase)
+    missing_headers = (required_other_columns + required_event_columns) - headers
+    return render status: :unprocessable_entity, json: { error: I18n.t("registrations.import.errors.missing_columns", columns: missing_headers.join(", ")) } if missing_headers.any?
+
+    filtered_csv_rows = CSV.read(
```
gregorbg: Why are you doing `CSV.read` here again? Wouldn't it be enough to load the CSV once (with all proper settings and everything) and then perform all necessary checks on the already loaded CSV?

app/controllers/registrations_controller.rb
```diff
@@ -34,6 +34,53 @@ class RegistrationsController < ApplicationController
...
+      headers: true,
+      header_converters: :symbol,
+      skip_blanks: true,
+      converters: ->(string) { string&.strip },
+    ).select { |row| row[:status] == "a" }
+
+    event_column_error = nil
+    @registration_rows = []
+
+    filtered_csv_rows.each do |row|
```
gregorbg: Wow, there's a lot of stuff going on in this loop all at once. I am not a fan of breaking twice, and also not a fan of using `<<` to build a new array (it should just be `map` in the first place).

Can you maybe split this into smaller sub-steps, which can rely on `map` and `filter` and such?
I think a more granular pipeline, which maps the rows first (using `to_hash` and such) and then loops once again over the mapped result to throw errors if it's not 0/1 will be much easier to read. The price you pay is that you have to loop over the list once, but even our very very biggest competitions ever have "only" 2000 competitors, which is a ludicrously small number of entries to iterate over for modern processors.

Also feel free to dump this whole big loop construct into your favorite LLM to get some suggestions about a functional, immutable rewrite.

app/controllers/registrations_controller.rb
```diff
@@ -81,17 +128,9 @@ def import
...
   def do_import
-    competition = competition_from_params
-    file = params[:csv_registration_file]
-    required_columns = ["status", "name", "country", "wca id", "birth date", "gender", "email"] + competition.events.map(&:id)
-    # Ensure the CSV file includes all required columns.
-    headers = CSV.read(file.path).first.compact.map(&:downcase)
-    missing_headers = required_columns - headers
-    raise I18n.t("registrations.import.errors.missing_columns", columns: missing_headers.join(", ")) if missing_headers.any?
+    competition = @competition
+    registration_rows = @registration_rows
```
gregorbg: These two statements feel a little bit silly. Why are you not referencing `@competition` directly in those lines which previously accessed `competition`?

gregorbg (changes requested):

app/controllers/registrations_controller.rb
danieljames-dj: Okay done

app/controllers/registrations_controller.rb
danieljames-dj: Okay LLM suggested a good approach with `CSV.foreach`. Hope this is better now.

app/controllers/registrations_controller.rb
danieljames-dj: Okay done

app/controllers/registrations_controller.rb
danieljames-dj: I didn't remove earlier as I didn't want to change any other part of this method. But now I moved some more part to `before_action`, so I removed these two lines.

spec/requests/registrations_spec.rb
```diff
@@ -29,6 +29,7 @@
       it "renders an error when there are missing columns" do
         file = csv_file [
           ["Status", "Name", "WCA ID", "Birth date", "Gender", "Email", "444"],
+          ["a", "Sherlock Holmes", "", "2000-01-01", "m", "sherlock@example.com", "1"],
```
gregorbg, line 32: Why did you add this to the test case?
The idea of your refactor should be that the previous test cases all work the same way they worked before.

spec/requests/registrations_spec.rb
```diff
@@ -401,16 +402,18 @@
 
         context "registrant accepted in the database, but not in the CSV file" do
           it "deletes the registration" do
-            registration = create(:registration, :accepted, competition: competition)
+            registration1 = create(:registration, :accepted, competition: competition, events: %w[333])
+            registration2 = create(:registration, :accepted, competition: competition, events: %w[444])
             file = csv_file [
               ["Status", "Name", "Country", "WCA ID", "Birth date", "Gender", "Email", "333", "444"],
+              ["a", registration1.user.name, registration1.user.country.id, "", registration1.user.dob, registration1.user.gender, registration1.user.email, "1", "0"],
```
gregorbg, line 409: Same as my other comment: Why are you adding data here?

spec/requests/registrations_spec.rb
```diff
@@ -401,16 +402,18 @@
 
         context "registrant accepted in the database, but not in the CSV file" do
           it "deletes the registration" do
-            registration = create(:registration, :accepted, competition: competition)
+            registration1 = create(:registration, :accepted, competition: competition, events: %w[333])
+            registration2 = create(:registration, :accepted, competition: competition, events: %w[444])
```
gregorbg, line 406: And why are you now distinguishing two different registrations here?

app/controllers/registrations_controller.rb
```diff
@@ -34,6 +34,112 @@ class RegistrationsController < ApplicationController
...
+      headers ||= row.headers.map { |header| header.to_s.downcase }
+      rows << row if row[:status] == "a"
+    end
+
+    { headers: headers, rows: rows }
+  end
+
+  private def validate_required_headers(headers)
+    competition_events = @competition.competition_events
+    required_event_columns = competition_events.pluck(:event_id)
```
gregorbg:
```suggestion
  private def validate_required_headers(headers, competition)
    required_event_columns = competition.competition_events.pluck(:event_id)
```

gregorbg: Regarding the parameters: It's not good practice to rely on the fact that `@competition` "magically" exists. In controller methods, you use this approach to "share" variables between `before_action` checks and the actual controller method body, because you don't have any other choice. But when you're calling helper methods from within the `before_action`, then you do have the choice to explicitly pass them as parameters, and you should definitely always do that.

app/controllers/registrations_controller.rb
```diff
@@ -34,6 +34,112 @@ class RegistrationsController < ApplicationController
...
+    required_other_columns = %w[status name country wca_id birth_date gender email]
+    required_columns = required_other_columns + required_event_columns
+
+    missing_columns = required_columns - headers
+
+    I18n.t("registrations.import.errors.missing_columns", columns: missing_columns.join(", ")) if missing_columns.any?
+  end
+
+  private def validate_event_columns(csv_rows)
+    competition_events = @competition.competition_events
```
gregorbg: See my other remark above about passing in vs. magically expecting `@competition`

app/controllers/registrations_controller.rb
```diff
@@ -34,6 +34,112 @@ class RegistrationsController < ApplicationController
...
+        %w[1 0].exclude?(cell_value)
+      end
+
+      if invalid_event
+        error = I18n.t("registrations.import.errors.invalid_event_column", value: row[invalid_event.event_id.to_sym], column: invalid_event.event_id)
+        break
+      end
+    end
+
+    error
```
gregorbg: You are using this to return the _first_ error that was found. This is consistent with the old code, but maybe not very desirable from a user perspective: Imagine you have two accidental errors in your large CSV, which you didn't realize. You try to upload, get an error report, fix it, try to upload again, get an error report again.

With some `filter_map` magic on the outer block, you should be able to (a) get rid of `break` and the empty `error = nil` initializer, as well as (b) return all invalid errors at once.

app/controllers/registrations_controller.rb
```diff
@@ -34,6 +34,112 @@ class RegistrationsController < ApplicationController
...
+      }
+    end
+
+    raw_dobs = @registration_rows.pluck(:birth_date)
+    wrong_format_dobs = raw_dobs.reject { |raw_dob| Date.safe_parse(raw_dob)&.to_fs == raw_dob }
+    return unless wrong_format_dobs.any?
+
+    render status: :unprocessable_entity, json: {
+      error: I18n.t("registrations.import.errors.wrong_dob_format", raw_dobs: wrong_format_dobs.join(", ")),
+    }
```
gregorbg: Is there a chance that you can come up with a common helper method for this repetitive block?

app/controllers/registrations_controller.rb
```diff
@@ -34,6 +34,112 @@ class RegistrationsController < ApplicationController
...
+      header_converters: :symbol,
+      skip_blanks: true,
+      converters: ->(string) { string&.strip },
+    ) do |row|
+      headers ||= row.headers.map { |header| header.to_s.downcase }
+      rows << row if row[:status] == "a"
+    end
+
+    { headers: headers, rows: rows }
+  end
```
gregorbg: Hm... In my previous comment, I had suggested something functional and immutable. This approach is neither functional (it uses iterative `foreach`) nor is it immutable (it uses empty initializers and `||=` as well as `<<` assignments). If LLM cannot help you, then try the good old "reading the documentation" 😉

danieljames-dj: Okay my bad, I didn't had details conversation with LLM. Did now and updated. Hope it's better now.

app/controllers/registrations_controller.rb
danieljames-dj: Okay thanks for explanation, done.

app/controllers/registrations_controller.rb
danieljames-dj: Done

spec/requests/registrations_spec.rb
danieljames-dj: I actually added a new error which will return 'empty file' error, and hence the previous CSV will fail with 'empty file' error. Hence adding one row for the testcase to work as expected.

spec/requests/registrations_spec.rb
danieljames-dj: Same reason as above - there was only one registration, and the testcase wanted to test whether the accepted registrants not in csv will become cancelled if that entry is not passed. The testcase was sending a csv with just header. To make the testcase work as expected, I'm adding one more registration.

spec/requests/registrations_spec.rb
danieljames-dj: Same reason as above

app/controllers/registrations_controller.rb
danieljames-dj: I can do that. Probably with some method like below:

```
  private def render_unprocessable_entity_error(error)
    render status: :unprocessable_entity, json: { error: error }
  end
```

Is this what you meant? (thought of confirming this with you before executing as this change is there in multiple places)

app/controllers/registrations_controller.rb
danieljames-dj: Okay done, thanks

app/controllers/registrations_controller.rb
gregorbg: Not just the render. I mean the whole logic of plucking a value, applying a transform, and returning an i18n message in case the transform returns "invalid".

In other words: Right now you have three structurally similar blocks. Ideally, you would only have one block and three lines which call this shared block.

danieljames-dj: Okay I've attempted for it, not so confident though.

spec/requests/registrations_spec.rb
gregorbg: Then what about a test case that makes sure an empty CSV file errors out as expected?

app/controllers/registrations_controller.rb
```diff
@@ -34,6 +34,98 @@ class RegistrationsController < ApplicationController
...
+    wrong_format_dobs = raw_dobs.reject { |raw_dob| Date.safe_parse(raw_dob)&.to_fs == raw_dob }
+    dob_column_error = I18n.t("registrations.import.errors.wrong_dob_format", raw_dobs: wrong_format_dobs.join(", ")) if wrong_format_dobs.any?
+
+    @emails = csv_rows.pluck(:email)
+    email_duplicates = @emails.select { |email| @emails.count(email) > 1 }.uniq
+    email_duplicate_error = I18n.t("registrations.import.errors.email_duplicates", emails: email_duplicates.join(", ")) if email_duplicates.any?
+
+    wca_ids = csv_rows.pluck(:wca_id).reject(&:empty?)
+    wca_id_duplicates = wca_ids.select { |wca_id| wca_ids.count(wca_id) > 1 }.uniq
+    wca_id_duplicate_error = I18n.t("registrations.import.errors.wca_id_duplicates", wca_ids: wca_id_duplicates.join(", ")) if wca_id_duplicates.any?
```
gregorbg: When I was talking about unifying these checks before, I had something like this in mind...

```ruby
def column_check(csv_rows, column_name, error_key, i18n_keyword)
  column_values = csv_rows.pluck(column_name)

  malformed_values = column_values.select do |value|
    yield value, column_values
  end.uniq

  I18n.t("registrations.import.errors.#{error_key}", i18n_keyword => malformed_values.join(", ")) if malformed_values.any?
end
```

And then the call in the actual method checks can become
```ruby
column_check(:birth_date, "wrong_dob_format", :raw_dobs) { |dob| Date.safe_parse(raw_dob)&.to_fs != raw_dob  }
column_check(:email, "email_duplicates", :emails) { |email, emails| emails.count(email) > 1 }
column_check(:wca_id, "wca_id_duplicates", :wca_ids) { |wca_id, wca_ids| !wca_id.empty? && wca_ids.count(wca_id) > 1 }
```

gregorbg: The only small afterthough here is that you would have to declare `@emails` separately after running the check, perhaps even within the body of the actual `do_import` controller method. But that won't hurt I think.

gregorbg (changes requested):

app/controllers/registrations_controller.rb
danieljames-dj: Okay this is interesting. I think I'm using `yield` for first time. Implemented.

spec/requests/registrations_spec.rb
danieljames-dj: Done

gregorbg (approved):

PR #12217, opened 2025-08-13
Serve Nextjs image assets through cloudfront

FinnIckler:
I already ran terraform and created the cloudfront distribution, so files can be accessed like this https://assets-nextjs.worldcubeassociation.org/_MG_2425.jpg already.

So this should just work once deployed?

### DIFF
--- a/next-frontend/src/payload.config.ts
+++ b/next-frontend/src/payload.config.ts
@@ -45,7 +45,11 @@ async function plugins() {
       ...defaultPlugins,
       s3Storage({
         collections: {
-          media: true,
+          media: {
+            generateFileURL: (file) => {
+              return `${process.env.MEDIA_BUCKET_CDN!}/${file.prefix}/${file.filename}`;
+            },
+          },
         },
         bucket: process.env.MEDIA_BUCKET!,
         config: {
--- a/infra/wca_on_rails/production/nextjs.tf
+++ b/infra/wca_on_rails/production/nextjs.tf
@@ -26,6 +26,10 @@ locals {
       value = aws_s3_bucket.next-media.id
     },
     {
+      name = "MEDIA_BUCKET_CDN"
+      value = "https://assets-nextjs.worldcubeassociation.org"
+    },
+    {
       name  = "OIDC_ISSUER"
       value = "https://www.worldcubeassociation.org/"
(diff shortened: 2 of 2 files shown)


next-frontend/src/payload.config.ts
```diff
@@ -45,7 +45,11 @@ async function plugins() {
       ...defaultPlugins,
       s3Storage({
         collections: {
-          media: true,
+          media: {
+            generateFileURL: (file) => {
+              return `${process.env.MEDIA_BUCKET_CDN!}/${file.prefix}/${file.filename}`;
```
gregorbg: If the file that you linked in your initial description is reachable at https://assets-nextjs.worldcubeassociation.org/_MG_2425.jpg, then why do you need to use two path segments (one for `prefix`, one for `filename`) here? The sample URL only contains one path segment after the CDN domain.

FinnIckler: currently, we don't set a prefix, but if we start doing multiple collections with different prefixes, we would need it.
I assumed that https://assets-nextjs.worldcubeassociation.org//_MG_2425.jpg would also be reachable, but I just noticed that it isn't. I just wanted to make it future proof, but I guess we'll need to change it if/when we want to introduce prefixes.

gregorbg: Why not just start introducing a (currently quite meaningless, but future-proof) prefix for `media` right away?

FinnIckler: well we would need to move all the currently uploaded files. If that is fine, than sure!

next-frontend/src/payload.config.ts
```diff
@@ -45,7 +45,12 @@ async function plugins() {
       ...defaultPlugins,
       s3Storage({
         collections: {
-          media: true,
+          media: {
+            prefix: "media",
```
gregorbg, line 49: Does this mean we have to manually specify that every collection has a prefix that is the same name as the collection?

That would be kinda 🤦 on Payload's part...

FinnIckler: well the prefix can be whatever so I do understand that they make you choose it manually

gregorbg (approved):

PR #12218, opened 2025-08-13
comment csv overflow fix

pranav-027:
missed in https://github.com/thewca/worldcubeassociation.org/pull/12177

### DIFF
--- a/app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
+++ b/app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
@@ -45,9 +45,9 @@ function V3csvExport(selected, registrations, competition) {
         registration.registrant_id
       },${
         registration.competing.waiting_list_position || ''
-      },"${
+      },${
         escapeCsv(registration.competing.comments)
-      }"\n`;
+      }\n`;
     });
 
   const blob = new Blob([csvContent], { type: 'text/csv;charset=utf-8;' });
--- a/app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
+++ b/app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
@@ -46,7 +46,7 @@ function V3csvExport(selected, registrations, competition) {
       },${
         registration.competing.waiting_list_position || ''
       },"${
-        registration.competing.comments || ''
+        escapeCsv(registration.competing.comments)
       }"\n`;
     });
 
@@ -61,6 +61,16 @@ function V3csvExport(selected, registrations, competition) {
 
   document.body.removeChild(link);
   URL.revokeObjectURL(url);
+}
+
+function escapeCsv(value) {
+  if (value == null) return '';
+  const str = String(value);
+  const escaped = str.replace(/"/g, '""');
+  if (/[",\n\r]/.test(str)) {
+    return `"${escaped}"`;
+  }
+  return escaped;
 }
 
 function csvExport(selected, registrations, competition) {
--- a/app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
+++ b/app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
@@ -18,6 +18,13 @@ import {
 } from '../../../lib/utils/registrationAdmin';
 import { useConfirm } from '../../../lib/providers/ConfirmProvider';
 
+function escapeCsv(value) {
+  if (!value) return '';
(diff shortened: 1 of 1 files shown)


gregorbg:
@simonkellly made a very good point that I had not considered when I approved the original PR: This is a user-controlled comment, so we need to be _very_ careful about escaping.

It is plausible, even for a non-malicious user, to include a `"` in their comments. (`Dear Organizer, the payment has been sent via Paypal from my dad's email which is "foo@bar.com". Thank you!`). This would hugely confuse the "manual" wrapping with `"` that you're doing in your PR.

What does the Internet say about masking values for CSV export? I'm not super-keen on including a new NPM dependency for this, and I don't think we need to be _craaaazy_ careful about code injection or something, but some basic degree of care is definitely called for

FinnIckler:
I think it's more a communications issue to tell organizers in what encoding the comment will be saved

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
```diff
@@ -63,6 +63,16 @@ function V3csvExport(selected, registrations, competition) {
   URL.revokeObjectURL(url);
 }
 
+function escapeCsv(value) {
+  if (value == null) return '';
+  const str = String(value);
+  const escaped = str.replace(/"/g, '""');
```
gregorbg: Why are you replacing one quote with two quotes? Is that really the way to escape quotes in CSV?

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
```diff
@@ -63,6 +63,16 @@ function V3csvExport(selected, registrations, competition) {
   URL.revokeObjectURL(url);
 }
 
+function escapeCsv(value) {
+  if (value == null) return '';
```
gregorbg: What about `if (!value)`?

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
gregorbg: If yes, please provide an additional code comment explaining this behavior.

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
```diff
@@ -63,6 +63,16 @@ function V3csvExport(selected, registrations, competition) {
...
+  const str = String(value);
+  const escaped = str.replace(/"/g, '""');
+  if (/[",\n\r]/.test(str)) {
+    return `"${escaped}"`;
+  }
+  return escaped;
```
gregorbg: Why not always just wrap the thing in quotes and call it a day?

gregorbg (changes requested):

gregorbg:
Sidenote: I would be open to a simple(!) CSV implementation along the lines of https://csv.js.org/stringify/api/sync/, if that makes your life easier.

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
pranav-027: Yes Per RFC 4180, double quotes in CSV are escaped by doubling them:  `"Hello"  ->  ""Hello""`

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
pranav-027: done

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationActions.jsx
pranav-027: Done

pranav-027:
> Sidenote: I would be open to adding an external library with simple(!) CSV implementation along the lines of https://csv.js.org/stringify/api/sync/, if that makes your life easier.

I tried this not sure how to add a new node dependency. Should i just run yarn install in rails shell ?

FinnIckler:
The docker setup will automatically run yarn install when booting up. So you just need to add it and run docker compose up

pranav-027:
> The docker setup will automatically run yar
registration.2.csv)
n install when booting up. So you just need to add it and run docker compose up

Okay thanks! for now i have implemented it without librabry and seems to work fine 
Example CSV - 
[SICC2025-registration (2).csv](https://github.com/user-attachments/files/21822084/SICC2025-

If the external library is needed i'll implement that!

PR #12221, opened 2025-08-14
Move events merged data to tickets UI

danieljames-dj:
Screenshot:

Screenshot 2025-08-14 at 11 40 02

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -122,7 +122,7 @@ class AdminController < ApplicationController
   end
 
   def delete_results_data
-    competition = competition_from_params(associations: %i[rounds results scrambles])
+    competition = competition_from_params(associations: [:results, :scrambles, { rounds: %i[results scrambles] }])
 
     model = params.require(:model)
 
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -122,7 +122,7 @@ class AdminController < ApplicationController
   end
 
   def delete_results_data
-    @competition = competition_from_params
+    @competition = competition_from_params(associations: %i[rounds results scrambles])
 
     model = params.require(:model)
 
@@ -134,9 +134,9 @@ class AdminController < ApplicationController
 
       case model
       when Result.name
-        Result.where(round_id: round_id).destroy_all
+        @competition.rounds.find(round_id).results.destroy_all
       when Scramble.name
-        Scramble.where(round_id: round_id).destroy_all
+        @competition.rounds.find(round_id).scrambles.destroy_all
       else
         return render status: :bad_request, json: { error: "Invalid model: #{model}" }
       end
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -130,22 +130,19 @@ class AdminController < ApplicationController
       @competition.results.destroy_all
       @competition.scrambles.destroy_all
     else
-      event_id = params.require(:event_id)
-      round_type_id = params.require(:round_type_id)
+      round_id = params.require(:roundId)
 
       case model
       when Result.name
-        Result.where(competition_id: @competition.id, event_id: event_id, round_type_id: round_type_id).destroy_all
+        Result.where(round_id: round_id).destroy_all
       when Scramble.name
-        Scramble.where(competition_id: @competition.id, event_id: event_id, round_type_id: round_type_id).destroy_all
+        Scramble.where(round_id: round_id).destroy_all
       else
-        raise "Invalid table: #{params[:table]}"
+        return render status: :bad_request, json: { error: "Invalid model: #{model}" }
       end
     end
 
-    load_result_posting_steps do
-      render partial: 'import_results_steps'
-    end
+    render status: :ok, json: { success: true }
   end
 
   def fix_results
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -122,21 +122,21 @@ class AdminController < ApplicationController
   end
 
   def delete_results_data
-    @competition = competition_from_params(associations: %i[rounds results scrambles])
+    competition = competition_from_params(associations: %i[rounds results scrambles])
 
     model = params.require(:model)
 
     if model == 'All'
-      @competition.results.destroy_all
-      @competition.scrambles.destroy_all
+      competition.results.destroy_all
+      competition.scrambles.destroy_all
     else
-      round_id = params.require(:roundId)
+      round = competition.rounds.find(params.require(:roundId))
 
       case model
       when Result.name
-        @competition.rounds.find(round_id).results.destroy_all
+        round.results.destroy_all
       when Scramble.name
-        @competition.rounds.find(round_id).scrambles.destroy_all
+        round.scrambles.destroy_all
       else
         return render status: :bad_request, json: { error: "Invalid model: #{model}" }
       end
--- a/app/controllers/tickets_controller.rb
+++ b/app/controllers/tickets_controller.rb
@@ -252,8 +252,10 @@ class TicketsController < ApplicationController
       {
         round_id: round.id,
         round_name: round.name,
-        result_count: round.results.count,
-        scramble_count: round.scrambles.count,
+        count: {
+          result: round.results.count,
+          scramble: round.scrambles.count,
+        },
       }
     end
 
--- a/app/controllers/tickets_controller.rb
+++ b/app/controllers/tickets_controller.rb
@@ -243,4 +243,20 @@ class TicketsController < ApplicationController
 
     render status: :ok, json: { success: true }
   end
+
+  def events_merged_data
+    ticket = Ticket.find(params.require(:ticket_id))
+    competition = ticket.metadata.competition
+
+    rounds_data = competition.rounds.sort_by { |r| [r.event_id, r.round_type.rank] }.map do |round|
+      {
+        round_id: round.id,
+        round_name: round.name,
+        result_count: round.results.count,
+        scramble_count: round.scrambles.count,
+      }
+    end
+
+    render status: :ok, json: rounds_data
+  end
 end
--- a/app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/MiscActions/EventsMergedData/EventsMergedDataContent.jsx
+++ b/app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/MiscActions/EventsMergedData/EventsMergedDataContent.jsx
@@ -42,14 +42,19 @@ export default function EventsMergedDataContent({ ticketDetails }) {
           if (model === 'All') {
             return {
               ...roundData,
-              result_count: 0,
-              scramble_count: 0,
+              count: {
+                results: 0,
+                scrambles: 0,
+              },
             };
           }
           if (roundData.round_id === roundId) {
             return {
               ...roundData,
-              [`${model.toLowerCase()}_count`]: 0,
+              count: {
+                ...roundData.count,
+                [model.toLowerCase()]: 0,
+              },
             };
           }
           return roundData;
@@ -69,34 +74,30 @@ export default function EventsMergedDataContent({ ticketDetails }) {
         more data will override the existing inbox data, but not the merged results data. You
         may remove the currently merged results data using the interface below.
       </Message>
-      <Table celled striped compact="very">
+      <Table celled striped compact="very" definition>
         <Table.Header>
           <Table.Row>
             <Table.HeaderCell>Event Round</Table.HeaderCell>
-            <Table.HeaderCell>Has Merged Results</Table.HeaderCell>
-            <Table.HeaderCell>Has Merged Scrambles</Table.HeaderCell>
+            <Table.HeaderCell colSpan="2">Has Merged Results</Table.HeaderCell>
+            <Table.HeaderCell colSpan="2">Has Merged Scrambles</Table.HeaderCell>
           </Table.Row>
         </Table.Header>
         <Table.Body>
           {eventsMergedData.map((roundData) => (
             <Table.Row key={roundData.round_id}>
-              <Table.Cell collapsing>{roundData.round_name}</Table.Cell>
-              <Table.Cell collapsing>
-                <DataActioner
-                  roundData={roundData}
-                  model="Result"
-                  deleteMutate={deleteResultsDataMutate}
-                  competitionId={competitionId}
-                />
-              </Table.Cell>
-              <Table.Cell collapsing>
-                <DataActioner
-                  roundData={roundData}
-                  model="Scramble"
-                  deleteMutate={deleteResultsDataMutate}
-                  competitionId={competitionId}
-                />
-              </Table.Cell>
+              <Table.Cell>{roundData.round_name}</Table.Cell>
+              <DataActioner
+                roundData={roundData}
+                model="Result"
+                deleteMutate={deleteResultsDataMutate}
+                competitionId={competitionId}
+              />
+              <DataActioner
+                roundData={roundData}
+                model="Scramble"
+                deleteMutate={deleteResultsDataMutate}
+                competitionId={competitionId}
+              />
             </Table.Row>
           ))}
         </Table.Body>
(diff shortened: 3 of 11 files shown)


app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/MiscActions/EventsMergedData/EventsMergedDataContent.jsx
```diff
@@ -0,0 +1,148 @@
...
+function DataActioner({
+  roundData, model, deleteMutate, competitionId,
+}) {
+  const [showConfirm, setShowConfirm] = useState();
+  const count = model === 'Result' ? roundData.result_count : roundData.scramble_count;
+
+  if (count > 0) {
+    return (
+      <>
+        <>
```
gregorbg: Two nested fragments are very unusual and almost certainly redundant. Why does one single fragment not work?

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/MiscActions/EventsMergedData/index.jsx
```diff
@@ -0,0 +1,33 @@
...
+
+export default function EventsMergedData({ ticketDetails }) {
+  const [isModalOpen, setIsModalOpen] = useState();
+
+  return (
+    <>
+      <Popup
+        trigger={(
+          <div>
+            {/* Button wrapped in a div because disabled button does not fire mouse events */}
```
gregorbg: But the button for this particular modal never ever becomes `disabled` though??

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/MiscActions/EventsMergedData/EventsMergedDataContent.jsx
```diff
@@ -0,0 +1,148 @@
...
+                />
+              </Table.Cell>
+            </Table.Row>
+          ))}
+        </Table.Body>
+      </Table>
+      Please be careful removing data! Data in the above table is live.
+      Remove all results and scrambles only, does not affect persons:
+      {' '}
+      <Button color="red" as="a" onClick={() => setShowConfirm(true)}>X ALL</Button>
```
gregorbg: Why is this an `as="a"` if it has no href attribute?

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/MiscActions/EventsMergedData/EventsMergedDataContent.jsx
```diff
@@ -0,0 +1,148 @@
...
+            return {
+              ...roundData,
+              result_count: 0,
+              scramble_count: 0,
+            };
+          }
+          if (roundData.round_id === roundId) {
+            return {
+              ...roundData,
+              [`${model.toLowerCase()}_count`]: 0,
```
gregorbg: Feel free to redesign the API data format returned by the backend if it makes your life in React easier.

app/controllers/tickets_controller.rb
```diff
@@ -243,4 +243,20 @@ def post_results
...
+  def events_merged_data
+    ticket = Ticket.find(params.require(:ticket_id))
+    competition = ticket.metadata.competition
+
+    rounds_data = competition.rounds.sort_by { |r| [r.event_id, r.round_type.rank] }.map do |round|
+      {
+        round_id: round.id,
+        round_name: round.name,
+        result_count: round.results.count,
+        scramble_count: round.scrambles.count,
```
gregorbg: See my other comment about this API design not being ideal for React key access. Maybe something in the shape of `counts { results, scrambles }` could be more helpful?

app/controllers/admin_controller.rb
```diff
@@ -130,22 +130,19 @@ def delete_results_data
       @competition.results.destroy_all
       @competition.scrambles.destroy_all
     else
-      event_id = params.require(:event_id)
-      round_type_id = params.require(:round_type_id)
+      round_id = params.require(:roundId)
```
gregorbg: If the data is definitely linked via `rounds`, then feel free to actually load the round, and use its `has_many` associations to delete.

gregorbg:
The `collapsing` evidently didn't work, the left column with the round/event names is still waaaaay to wide in your last screenshot.

Please consult the SemUI React documentation

app/controllers/admin_controller.rb
danieljames-dj: Done

app/controllers/tickets_controller.rb
danieljames-dj: Okay done

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/MiscActions/EventsMergedData/EventsMergedDataContent.jsx
danieljames-dj: Done

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/MiscActions/EventsMergedData/EventsMergedDataContent.jsx
danieljames-dj: Oops, removed.

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/MiscActions/EventsMergedData/EventsMergedDataContent.jsx
danieljames-dj: Ya it was not needed. I've anyway changed the view a bit, so this may not be relevant anymore.

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/MiscActions/EventsMergedData/index.jsx
danieljames-dj: Oh yes, it's not required. I've removed the `div` now.

app/controllers/admin_controller.rb
```diff
@@ -122,30 +122,27 @@ def delete_inbox_data
...
 
       case model
       when Result.name
-        Result.where(competition_id: @competition.id, event_id: event_id, round_type_id: round_type_id).destroy_all
+        @competition.rounds.find(round_id).results.destroy_all
```
gregorbg: You can extract this into a `round = @competition.rounds.find(round_id)` outside of the `case` block.

app/controllers/admin_controller.rb
```diff
@@ -122,30 +122,27 @@ def delete_inbox_data
   end
 
   def delete_results_data
-    @competition = competition_from_params
+    @competition = competition_from_params(associations: %i[rounds results scrambles])
```
gregorbg: I think (please correct me if I'm overlooking something) that it should be `{ rounds: [:results, :scrambles] }`. You're not calling `results` or `scrambles` directly on the competition itself, but rather you're calling them through `round`.

gregorbg: Sidenote: Why is this `@competition` and not just `competition` without the `@`?
Pure JSON backend controllers that don't render their own erb views usually don't need `@` variables.

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/MiscActions/EventsMergedData/EventsMergedDataContent.jsx
gregorbg: But I can still see it on GitHub? Did you forget to commit and/or push?

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/MiscActions/EventsMergedData/EventsMergedDataContent.jsx
```diff
@@ -0,0 +1,152 @@
...
+      />
+    </>
+  );
+}
+
+function DataActioner({
+  roundData, model, deleteMutate, competitionId,
+}) {
+  const [showConfirm, setShowConfirm] = useState();
+  const count = model === 'Result' ? roundData.count.result : roundData.count.scramble;
```
gregorbg:
```suggestion
  const count = roundData.count[model.toLowerCase()];
```

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/MiscActions/EventsMergedData/EventsMergedDataContent.jsx
```diff
@@ -0,0 +1,152 @@
...
+      <Table.Cell>
+        <Button
+          color="red"
+          onClick={() => setShowConfirm(true)}
+          compact
+          disabled={count === 0}
+        >
+          X
+        </Button>
+      </Table.Cell>
```
gregorbg: Have you tried putting this button into the "Yes/No cell" above, and using `floated="right"`?

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/MiscActions/EventsMergedData/EventsMergedDataContent.jsx
```diff
@@ -0,0 +1,152 @@
...
+}) {
+  const [showConfirm, setShowConfirm] = useState();
+  const count = model === 'Result' ? roundData.count.result : roundData.count.scramble;
+
+  return (
+    <>
+      <Table.Cell>{count > 0 ? 'Yes' : 'No'}</Table.Cell>
+      <Table.Cell>
+        <Button
+          color="red"
```
gregorbg:
```suggestion
          negative
```

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/MiscActions/EventsMergedData/index.jsx
```diff
@@ -0,0 +1,30 @@
...
+  const [isModalOpen, setIsModalOpen] = useState();
+
+  return (
+    <>
+      <Popup
+        trigger={(
+          <Button onClick={() => setIsModalOpen(true)}>
+            Events Merged Data
+          </Button>
+      )}
```
gregorbg: The indentation on this line looks off. Might just be the GH viewer, though

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/MiscActions/EventsMergedData/EventsMergedDataContent.jsx
```diff
@@ -0,0 +1,152 @@
...
+        <Table.Header>
+          <Table.Row>
+            <Table.HeaderCell>Event Round</Table.HeaderCell>
+            <Table.HeaderCell colSpan="2">Has Merged Results</Table.HeaderCell>
+            <Table.HeaderCell colSpan="2">Has Merged Scrambles</Table.HeaderCell>
+          </Table.Row>
+        </Table.Header>
+        <Table.Body>
+          {eventsMergedData.map((roundData) => (
+            <Table.Row key={roundData.round_id}>
```
gregorbg, line 87: Try `verticalAlign="middle"` for slightly better text flow

gregorbg (changes requested):

app/controllers/admin_controller.rb
danieljames-dj: Regarding association, TBH I didn't think your way. So both actually works because results & scrambles are having `has_many` association with both competitions and rounds, mainly because both `results` and scrambles` have `competition_id` and `round_id` (which maybe redundant?).

In this case, I'm opting for the existing association that I created, just because the most probable use-case is going to be using the `results` and `scrambles` from `competition` (when model is `All`) than using them from `rounds` (when model is `result` or `scramble`). The reason is that WRT very very rarely have the use-case of deleting data of just one event. At least 99% use-cases in the past have been to delete all data of a competition.

I hope this is fine, but I'm open to use the other way if you feel that is better.

Regarding `@competition`, I've changed it now. This method was earlier being called from rails, now no longer it's being used.

app/controllers/admin_controller.rb
danieljames-dj: Okay done

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/MiscActions/EventsMergedData/EventsMergedDataContent.jsx
danieljames-dj: Thanks, adding it in `Row` didn't work, but adding in `Cell` worked.

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/MiscActions/EventsMergedData/EventsMergedDataContent.jsx
danieljames-dj: Oops I think I removed it from the other button ("X") and not this one ("X ALL"). Removed from this too now.

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/MiscActions/EventsMergedData/EventsMergedDataContent.jsx
danieljames-dj: Done

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/MiscActions/EventsMergedData/EventsMergedDataContent.jsx
danieljames-dj: Done (this time did for other button too)

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/MiscActions/EventsMergedData/EventsMergedDataContent.jsx
danieljames-dj: Yes, but I felt this view was better. The other view had a bit more space in between and wasn't able to reduce it by adding `compact`.

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/MiscActions/EventsMergedData/index.jsx
danieljames-dj: Yes it was off. No idea why eslint formatter didn't fix this. Fixed it now.

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/MiscActions/EventsMergedData/EventsMergedDataContent.jsx
danieljames-dj: Actually it's not like way better, it's just slightly better. Having `floated="right"` is also fine if merging header columns is not a good practice.

app/controllers/admin_controller.rb
gregorbg: It is not redundant at all. Both models `competitions` and `rounds` have the association, but it means different things: If you access `some_competition.results`, you're loading all results for the whole competition, while `some_round.results` loads only the results of that one round.

Regarding use-case: That makes sense, but luckily `includes` is quite smart at being lazy: It won't actually fire the pre-loaded call until you really need it. So perhaps `associations: [:results, :scrambles, { rounds: [:results, :scrambles] }]` might work?

danieljames-dj: Ya that is also working, and that makes sense. Changed.

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/MiscActions/EventsMergedData/EventsMergedDataContent.jsx
```diff
@@ -0,0 +1,152 @@
...
+    onSuccess: (_, { model, roundId }) => {
+      queryClient.setQueryData(
+        ['events-merged-data', ticketId],
+        (previousData) => previousData.map((roundData) => {
+          if (model === 'All') {
+            return {
+              ...roundData,
+              count: {
+                results: 0,
+                scrambles: 0,
```
gregorbg: Here you're using plural (result**s** and scramble**s**) but up in the Rails controller, where the data originally comes from, it's just `result` and `scramble`.

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/MiscActions/EventsMergedData/EventsMergedDataContent.jsx
```diff
@@ -0,0 +1,152 @@
...
+              },
+            };
+          }
+          if (roundData.round_id === roundId) {
+            return {
+              ...roundData,
+              count: {
+                ...roundData.count,
+                [model.toLowerCase()]: 0,
```
gregorbg, line 56: Same here as above: Be careful about plural vs. singluar nouns.

gregorbg (changes requested):

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/MiscActions/EventsMergedData/EventsMergedDataContent.jsx
danieljames-dj: Oh yes, I missed that here. Fixed it now. Thanks for catching.

app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/MiscActions/EventsMergedData/EventsMergedDataContent.jsx
danieljames-dj: `model` is already singular here.

PR #12227, opened 2025-08-14
Rankings next

FinnIckler:
image

I made some changes to make the FilterBox component and shared methods reusable across both records and rankings. Still need to implement the By Region view and the URL changes (in both PRs actually)


FinnIckler:
image
was much easier to implement the show filters this time around

PR #12253, opened 2025-08-18
Add Logo page to Next

FinnIckler:
image
All pretty straightforward. Not sure how we handle the logo being on the right background though, maybe just have it as part of the image collection?


next-frontend/src/app/(wca)/logo/download.tsx
```diff
@@ -0,0 +1,29 @@
...
+        onCheckedChange={(e) => setAcceptedGuidelines(!!e.checked)}
+      >
+        <Checkbox.HiddenInput />
+        <Checkbox.Control />
+        <Checkbox.Label>
+          <I18nHTMLTranslate i18nKey="logo.headings.download_logo_assets.accept_terms_and_conditions" />
+        </Checkbox.Label>
+      </Checkbox.Root>
+      <Button disabled={!acceptedGuidelines} asChild>
+        <a href="https://assets.worldcubeassociation.org/assets/9963bbd/files/WCA%20Logo%20Assets.zip">
```
gregorbg: Is the link meant to be hardcoded? At least we could put it into Payload (as a field of the `logo-page` global) and pass it down here as a param

next-frontend/src/app/(wca)/logo/page.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  if (logoItems.length === 0) {
+    return <Heading>No Logo Items, add some!</Heading>;
+  }
+
+  const { t } = await getT();
+
+  return (
+    <Container>
+      <VStack gap="8" width="full" pt="8" alignItems="left">
```
gregorbg: I am sure that most of these properties are unnecessary. For example, `width="full"` is already the default behavior if you don't specify `width` at all.

next-frontend/src/app/(wca)/logo/page.tsx
```diff
@@ -0,0 +1,79 @@
...
+                  <Heading size="2xl">{item.title}</Heading>
+                  <Text>{item.caption}</Text>
+                  <HStack w="full">
+                    {item.images.map((value) => {
+                      const image = value.image as Media;
+                      return (
+                        <Image
+                          src={image.url!}
+                          alt={item.caption}
+                          key={image.url}
```
gregorbg: I think image entities have an `id` property which could be better used as `key` here

next-frontend/src/app/(wca)/logo/page.tsx
```diff
@@ -0,0 +1,79 @@
...
+                  />
+                </Fragment>
+              );
+            }
+            case "logoVariant": {
+              return (
+                <Fragment key={item.id}>
+                  <Heading size="2xl">{item.title}</Heading>
+                  <Text>{item.caption}</Text>
+                  <HStack w="full">
```
gregorbg, line 58: You could give this `HStack` a "neutral" background (for example `bg="bg"`, which looks silly but is actually a semantic token that follows the dark/light theme). That would partially solve the issue of image upload backgrounds

FinnIckler: it wouldn't, I tried that out as one of my solutions, but that still makes one of the logos not visible (the dark one in dark mode and the light one in light mode)

next-frontend/src/app/(wca)/logo/download.tsx
```diff
@@ -0,0 +1,33 @@
...
+        onCheckedChange={(e) => setAcceptedGuidelines(!!e.checked)}
+      >
+        <Checkbox.HiddenInput />
+        <Checkbox.Control />
+        <Checkbox.Label>
+          <I18nHTMLTranslate i18nKey="logo.headings.download_logo_assets.accept_terms_and_conditions" />
+        </Checkbox.Label>
+      </Checkbox.Root>
+      <Button disabled={!acceptedGuidelines} asChild>
+        <a href={logoDownloadLink}>
```
gregorbg: Why are you using `a` instead of a Chakra `Link` component?

gregorbg (approved):

PR #12254, opened 2025-08-18
Use existing singular plural logic in registration admin page

pranav-027:
image

Implementation is already existing on the public competitiors page. Did some copy pasta from `Competitors.jsx` and `PreTableInfo.jsx`


danieljames-dj:
The error looks really weird. It's saying that one key is not used, but you are not removing it's usages. Will wait for someone else to check.

PR #12256, opened 2025-08-19
Split `es` into `es-419` and `es-ES`

dunkOnIT:
With Guido actively contributing a Latin American Spanish translation, we need to split the `es` file into two, as follows: 
* `es-ES` (source: latest commit before Guido's contributions - used [this file](https://github.com/thewca/worldcubeassociation.org/blob/4161ae12a7cadd3d9a0e985649a8afc1e82c7eef/config/locales/es.yml), based on [the history](https://github.com/thewca/worldcubeassociation.org/commits/8ea83f9155ecc379220a1ee2923c9292eb6e5787/config/locales/es.yml)
* `es-419` (source: Guido's latest contribution in #12230 - which we will close in favour of this PR)

Everything done except for the flag_id for es-419 (see comment below) - and I'll need to change translator groups on prod once this is merged


config/locales/available.json
```diff
@@ -23,9 +23,13 @@
...
   },
-  "es": {
+  "es-ES": {
     "flag_id": "es",
-    "name": "Español"
+    "name": "Español (España)"
+  },
+  "es-419": {
+    "flag_id": "LatAm",
+    "name": "Español (América Latina)"
```
GuidoDipietro, line 32: This is just slightly inconsistent with how other split locales are named, in that they don't use parenthesis. It would maybe be a good idea to use the same formatting everywhere

GuidoDipietro:
Thanks for the work @dunkOnIT! I hope this will make stuff much smoother moving forward, and let me know how I can assist if at all.

config/locales/available.json
gregorbg: We can take care of that in a follow-up PR, as soon as we reached a consensus who should be the "main" Spanish. I am unwilling to let only the sheer "number of speakers" decide this question.

gregorbg (approved):

config/locales/available.json
dunkOnIT: Other options per Guido's suggestion in the other PR would we: 
- Español de España (for Spain-Spanish)
- Español latinoamericano (for Latin American Spanish) 

Although I like that brackets guarantee more of a consistent alphabetical ordering - I can imagine other languages that put descriptor words first might result in languages not being next to each other on the list, potentially causing confusion for users 

So I'd say if we change anything, we should change French and Portuguese (I assume zh-TW and zh-TN are beyond this concern)

PR #12257, opened 2025-08-19
Make 'Upload from WCA Live' button almost publicly available

gregorbg:
This will let people use our new scramble upload panel if they want to. There will be a reasonable transition period, during which the old "scrambles-matcher" way of doing things will also still be there.

In terms of code, I have made the following adjustments for better usability:
- The button in the "Upload Results" panel now also has short explanatory message box
- The component to upload your TNoodle files has been renamed to "Upload Scrambles" (from a previous "Match Scrambles")
- The WCA Live button now displays how many files you've already uploaded. If the count is 0, the button is disabled.


gregorbg:
The rest of this page is not translated either, so I just went with English.

danieljames-dj (approved):

dunkOnIT (changes requested):
Two suggested changes:
- Change wording for the first bullet under "The results submission process has two steps:" from "Uploading a valid JSON to the website" -> "Importing results (either by uploading a valid JSON from Scrambles Matcher, or importing directly from WCA Live)"
  - We could change these bullets to an ordered list to get actual numbering if we want
- Indicate that the import feature is still in beta: 
  - Heading "Import WCA Live Results" -> "(Beta) Import WCA Live Results"
  - Add a bullet under "Please Note" which reads: "This feature is still in beta - please report any errors or issues to the WCA Software Team [link to contact form]"

I think it's safer to bill it as a "Beta" given that we haven't exhaustively tested it on our end - I don't think it will meaningfully diminish use, given that it offers a huge convenience boost for delegates. But I do think it's important to signpost a higher-than-usual element of uncertainty

PR #12258, opened 2025-08-19
Add metadata_action to logs

danieljames-dj:
This is as per the discussion earlier while changing schema of ticket logs

### DIFF
--- a/app/models/ticket_log.rb
+++ b/app/models/ticket_log.rb
@@ -20,7 +20,7 @@ class TicketLog < ApplicationRecord
 
     errors.add(:metadata_action, "must be present when action_type is metadata_action") if action_type_metadata_action? && metadata_action.nil?
 
-    return if metadata_action.nil? || ticket.metadata_type.safe_constantize::ACTION_TYPE.key?(metadata_action)
+    return if metadata_action.nil? || ticket.metadata_type.safe_constantize::ACTION_TYPE.key?(metadata_action.to_sym)
 
     errors.add(:metadata_action, "is not a valid action")
   end
--- a/app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/MergeInboxResults.jsx
+++ b/app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/MergeInboxResults.jsx
@@ -7,7 +7,7 @@ import Loading from '../../../Requests/Loading';
 import Errored from '../../../Requests/Errored';
 import { ticketsCompetitionResultStatuses } from '../../../../lib/wca-data.js.erb';
 
-export default function MergeInboxResults({ ticketDetails }) {
+export default function MergeInboxResults({ ticketDetails, currentStakeholder }) {
   const { ticket: { id, metadata: { competition_id: competitionId } } } = ticketDetails;
 
   const queryClient = useQueryClient();
@@ -17,7 +17,7 @@ export default function MergeInboxResults({ ticketDetails }) {
     isError,
     error,
   } = useMutation({
-    mutationFn: () => mergeInboxResults({ ticketId: id }),
+    mutationFn: mergeInboxResults,
     onSuccess: () => {
       queryClient.setQueryData(
         ['ticket-details', id],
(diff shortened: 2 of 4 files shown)


app/models/tickets_competition_result.rb
```diff
@@ -26,6 +30,14 @@ def actions_allowed_for(ticket_stakeholder)
     end
   end
 
+  def metadata_actions_allowed_for(ticket_stakeholder)
+    if ticket_stakeholder.stakeholder == UserGroup.teams_committees_group_wrt
+      [TicketsCompetitionResult::ACTION_TYPE[:merge_inbox_results]]
```
gregorbg: You don't need to specify `TicketsCompetitionResult::` since you're currently within that class itself.

app/webpacker/components/Tickets/TicketLogs.jsx
```diff
@@ -12,12 +12,16 @@ export default function TicketLogs({ ticketId }) {
...
+  function logText(actionType, metadataAction) {
     switch (actionType) {
       case ticketLogActionTypes.create_ticket:
         return 'Ticket created.';
       case ticketLogActionTypes.update_status:
         return 'Status updated.';
+      case ticketLogActionTypes.create_comment:
+        return 'Comment added.';
+      case ticketLogActionTypes.metadata_action:
+        return `Metadata action ${metadataAction} performed.`;
```
gregorbg: This is not necessarily easy to understand for non-WST users

gregorbg (changes requested):
The changes that you made so far make sense, but you did not touch `actions_allowed_for` at all. The idea is to get these actions "away" from the metadata, back into the main ticket, right? So all code snippets that relate to `actions` (not the new `metadata_actions`) don't need to happen in `metadata` anymore. Or not?

app/models/tickets_competition_result.rb
danieljames-dj: Okay removed

app/webpacker/components/Tickets/TicketLogs.jsx
danieljames-dj: Okay, I've removed "metadata". Currently logs is not so clean UX-wise, but I'll be improving the UX with better texts later.

danieljames-dj (commented):
> The idea is to get these actions "away" from the metadata, back into the main ticket

I'm not sure if I understood this part correctly. I'll discuss with you over chat.

db/migrate/20250816050817_add_metadata_action_to_ticket_logs.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddMetadataActionToTicketLogs < ActiveRecord::Migration[7.2]
+  def change
+    add_column :ticket_logs, :metadata_action, :string
```
gregorbg: Please use `after: :action_type` (or pick another column that seems more reasonable to you, but don't just throw it to the end of the table randomly)

gregorbg (changes requested):

db/migrate/20250816050817_add_metadata_action_to_ticket_logs.rb
danieljames-dj: Done

app/models/ticket_log.rb
```diff
@@ -13,6 +14,17 @@ class TicketLog < ApplicationRecord
     include: %w[ticket_log_changes],
   }.freeze
 
+  validate :validate_metadata_action
+  def validate_metadata_action
+    errors.add(:metadata_action, "can only be set when action_type is metadata_action") if action_type != TicketLog.action_types[:metadata_action] && metadata_action.present?
```
gregorbg: You can use `if !self.action_type_metadata_action?` which is autogenerated by Rails thanks to your `prefix: true` at the enum declaration

app/models/ticket_log.rb
```diff
@@ -13,6 +14,17 @@ class TicketLog < ApplicationRecord
...
+
+    errors.add(:metadata_action, "must be present when action_type is metadata_action") if action_type == TicketLog.action_types[:metadata_action] && metadata_action.nil?
```
gregorbg: Use helper method instead of manually comparing action types, see other comment

danieljames-dj: Done

app/models/ticket_log.rb
danieljames-dj: Done

PR #12260, opened 2025-08-19
Add edit button for name warnings in newcomer check page

danieljames-dj:
Changing the way issues are created for name issues. If this is fine, will do the same for DOB as well in a follow-up PR.

### DIFF
--- a/app/webpacker/components/NewcomerChecks/NameFormatChecker.jsx
+++ b/app/webpacker/components/NewcomerChecks/NameFormatChecker.jsx
@@ -29,7 +29,7 @@ export default function NameFormatChecker({ competitionId, setUserIdToEdit }) {
         </Table.Row>
       </Table.Header>
       <Table.Body>
-        {nameFormatChecks.map(({ id, issues }) => issues.map(({ id: issueId, args }) => (
+        {nameFormatChecks.flatMap(({ id, issues }) => issues.map(({ id: issueId, args }) => (
           <Table.Row>
             <Table.Cell>{I18n.t(`validators.persons.${issueId}`, args)}</Table.Cell>
             <Table.Cell>


app/webpacker/components/NewcomerChecks/NameFormatChecker.jsx
```diff
@@ -1,32 +1,45 @@
...
-    </List>
+    <Table celled>
+      <Table.Header>
+        <Table.Row>
+          <Table.HeaderCell>Warning message</Table.HeaderCell>
+          <Table.HeaderCell>Action</Table.HeaderCell>
+        </Table.Row>
+      </Table.Header>
+      <Table.Body>
+        {nameFormatChecks.map(({ id, name, issues }) => issues.map((issue) => (
```
gregorbg: Shouldn't the first `map` be a `flatMap`?!

app/webpacker/components/NewcomerChecks/index.jsx
```diff
@@ -14,20 +16,49 @@ export default function Wrapper({ competitionId }) {
...
+            person.original_user_id === user.id
+              ? { ...person, original_user: user }
+              : person
+          ))),
+      }),
+    );
+
+    if (queryClient.getQueryData(['newcomer-name-format-checks', competitionId]).some((check) => check.id === user.id)) {
+      queryClient.invalidateQueries(['newcomer-name-format-checks', competitionId]);
+    }
```
gregorbg, line 39: (1) Why are you only invalidating under certain conditions?
(2) As far as I can tell, the old hook in `NewcomerChecks/SimilarPersons.jsx` did the same `setQueryData` as above, but did not do any invalidation at all. Why is it becoming necessary now?

lib/results_validators/persons_validator.rb
```diff
@@ -60,42 +60,64 @@ def self.dob_validations(dob, competition_id = nil, **message_args)
       validation_issues
     end
 
-    def self.name_validations(name, competition_id = nil, **_message_args)
-      validation_issues = []
-      roman_readable = PersonsValidator.roman_readable_part(name)
-      split_name = roman_readable.split
+    def self.name_validations(name, competition_id = nil)
```
gregorbg: From the looks of it, the only purpose of this method is to take "raw" validation issues and wrap them into "nice" `ValidationError` and `ValidationWarning` objects. This seems overkill, especially if you have to start doing this in other validators as well.

gregorbg: I would instead argue that you can leave this validator completely as-is, and "unpack" the `ValidationError` and `ValidationWarning` models at the place where you're using/serializing them. In this case, that would be `results_submission_controller.rb`. Is that possible?

app/controllers/results_submission_controller.rb
```diff
@@ -24,8 +24,16 @@ def last_duplicate_checker_job_run
...
+    name_format_issues = competition.accepted_newcomers.filter_map do |user|
+      issues = ResultsValidators::PersonsValidator.name_issues(user.name)
+
+      if issues.present?
+        {
+          id: user.id,
+          name: user.name,
+          issues: issues,
+        }
+      end
```
```suggestion
    name_format_issues = competition.accepted_newcomers.filter_map do |user|
      ResultsValidators::PersonsValidator.name_issues(user.name).presence
    end.map do |issues|
      {
        id: user.id,
        name: user.name,
        issues: issues,
      }
```

app/webpacker/components/NewcomerChecks/NameFormatChecker.jsx
danieljames-dj: Actually no, because the format of the `nameFormatChecks` was changed.

app/webpacker/components/NewcomerChecks/index.jsx
danieljames-dj: The one in SimilarPersons was `last-duplicate-checker-job`, and that is still not invalidated. The one I'm invalidating is `newcomer-name-format-checks`, because the user has changed the name from "Name Checks" page and frontend won't know if there are any more warnings, hence invalidating so that another server call will happen.

lib/results_validators/persons_validator.rb
danieljames-dj: Okay done.

app/controllers/results_submission_controller.rb
```diff
@@ -24,8 +24,16 @@ def last_duplicate_checker_job_run
   def newcomer_name_format_check
     competition = competition_from_params
 
-    name_format_issues = competition.accepted_newcomers.flat_map do |user|
-      ResultsValidators::PersonsValidator.name_validations(user.name, nil)
+    name_format_issues = competition.accepted_newcomers.filter_map do |user|
+      issues = ResultsValidators::PersonsValidator.name_validations(user.name)
+
+      if issues.present?
```
danieljames-dj, line 30: @gregorbg Sorry I had to revert your suggestion because I'll be requiring `user` object to build this, and hence `presence` didn't work as expected.

app/webpacker/components/NewcomerChecks/NameFormatChecker.jsx
```diff
@@ -1,32 +1,45 @@
...
-    </List>
+    <Table celled>
+      <Table.Header>
+        <Table.Row>
+          <Table.HeaderCell>Warning message</Table.HeaderCell>
+          <Table.HeaderCell>Action</Table.HeaderCell>
+        </Table.Row>
+      </Table.Header>
+      <Table.Body>
+        {nameFormatChecks.map(({ id, issues }) => issues.map(({ id: issueId, args }) => (
```
gregorbg: I still don't understand your argument. Yes, the format has changed from before, I get that. But why does that mean that a "map inside map" is okay?

I am mostly confused because this line returns an array *of an array* of React nodes. I am genuinely surprised that this even works without crashing.

danieljames-dj: Thanks for explaining, I've changed to `flatMap`.

gregorbg (approved):

PR #12275, opened 2025-08-23
Remember toggles on reg admin page

kr-matthews:
One of the medium priorities left on #10542.

Like `useStoredState`, this stores in local storage, so is specific to the device.

I think the only potential concern is that we intentionally made birthdays hidden by default on this page a few years ago (before it got migrated to React). I'm not sure this is a major concern, but if it is then I could allow passing a sanitizer function which processes the value that comes out of storage (which could toggle birthday to off in this case).

### DIFF
--- a/app/webpacker/lib/hooks/useStoredReducer.js
+++ b/app/webpacker/lib/hooks/useStoredReducer.js
@@ -11,7 +11,7 @@ import { getJsonItem, setJsonItem } from '../utils/localStorage';
 export default function useStoredReducer(reducer, initialState, key) {
   function augmentedReducer(state, action) {
     const newState = reducer(state, action);
-    localStorage.setItem(key, JSON.stringify(newState));
+    setJsonItem(key, newState);
     return newState;
   }
 
--- a/app/webpacker/lib/hooks/useStoredReducer.js
+++ b/app/webpacker/lib/hooks/useStoredReducer.js
@@ -1,0 +1,33 @@
+import { useReducer } from 'react';
+
+/**
+ * This functions like the useReducer hook, but it fetches the state stored in
+ * local storage, via the given key, on subsequent uses.
+ *
+ * Do NOT call this twice with the same key - updating one of such a pair
+ * will not update the other's state.
+ */
+export default function useStoredReducer(reducer, initialState, key) {
+  let storedState;
+  try {
+    storedState = JSON.parse(localStorage.getItem(key));
+  } catch {
+    storedState = null;
+  }
+
+  function augmentedReducer(state, action) {
+    const newState = reducer(state, action);
+    localStorage.setItem(key, JSON.stringify(newState));
+    return newState;
+  }
+
+  const [state, dispatch] = useReducer(augmentedReducer, initialState, (value) => {
+    if (storedState === null) {
+      localStorage.setItem(key, JSON.stringify(value));
+      return value;
+    }
+    return storedState;
+  });
+
+  return [state, dispatch];
+}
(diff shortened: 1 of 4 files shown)


app/webpacker/lib/hooks/useStoredReducer.js
```diff
@@ -0,0 +1,33 @@
...
+ * Do NOT call this twice with the same key - updating one of such a pair
+ * will not update the other's state.
+ */
+export default function useStoredReducer(reducer, initialState, key) {
+  let storedState;
+  try {
+    storedState = JSON.parse(localStorage.getItem(key));
+  } catch {
+    storedState = null;
+  }
```
gregorbg: Under which circumstances can an error occur? I would prefer a "is `localStorage.getItem(key)` present"-style check over a carte blanche catch approach

app/webpacker/lib/hooks/useStoredReducer.js
```diff
@@ -0,0 +1,33 @@
...
+
+  function augmentedReducer(state, action) {
+    const newState = reducer(state, action);
+    localStorage.setItem(key, JSON.stringify(newState));
+    return newState;
+  }
```
gregorbg: Would it be worth wrapping this in a `useCallback`?

app/webpacker/lib/hooks/useStoredReducer.js
```diff
@@ -0,0 +1,33 @@
...
+
+  const [state, dispatch] = useReducer(augmentedReducer, initialState, (value) => {
+    if (storedState === null) {
+      localStorage.setItem(key, JSON.stringify(value));
+      return value;
+    }
+    return storedState;
```
gregorbg: As far as I can tell, you should be able to move the whole initialization block for `storedState` into this callback, as the variable `storedState` is only ever used in here.

app/webpacker/lib/hooks/useStoredReducer.js
kr-matthews: `JSON.parse` will throw an exception if it can't parse its argument, which in theory could be anything since it's coming from local storage. The try-catch isn't for `localStorage.getItem`.

app/webpacker/lib/hooks/useStoredReducer.js
kr-matthews: I don't think there's any benefit. It's not being used in any dependency arrays that I'm aware of. Reducers need to be pure (not depend on any state), but it doesn't matter if the reference is a new function every time.

app/webpacker/lib/hooks/useStoredState.js
```diff
@@ -8,14 +8,14 @@ import { useState } from 'react';
...
-    storedState = null;
-  }
-
   const [state, setState] = useState(() => {
+    let storedState;
+    try {
+      storedState = JSON.parse(localStorage.getItem(key));
+    } catch {
+      storedState = null;
+    }
```
gregorbg: Now that you have also moved this code block (in addition to the new `storedReducer`) you're suddenly making a good point that the whole "load something JSON-esque from local storage or return null if it fails" logic should be its own independent snippet

app/webpacker/lib/utils/localStorage.js
```diff
@@ -0,0 +1,14 @@
+export function getJsonItem(key) {
+  const item = localStorage.getItem(key);
+  let parsedItem;
+  try {
+    parsedItem = JSON.parse(item);
+  } catch {
+    parsedItem = null;
+  }
+  return parsedItem;
```
gregorbg:
```suggestion
  try {
    return JSON.parse(item);
  } catch {
    return null;
  }
```

kr-matthews: Yeah I had this thought earlier today too.

PR #12299, opened 2025-08-25
Add Common Queries to WIC Panel

Brandon-Reichman:
WIC has been using the statistics site for 3 queries that aid in investigations. The stats site often doesn't work, and has outdated data.

I've created 3 panel pages, and added them to the WIC Panel, to help WIC easily access this data.

The three pages are
- Competitor Registrations
- Organized Competitions
- Delegated Competitions


app/controllers/admin/persons_controller.rb
```diff
@@ -49,6 +49,66 @@ def results
...
+                           .where.not(competitions: { announced_at: nil })
+                           .where(competitions: { cancelled_at: nil })
+                           .select(
+                             'competitions.*,
+          competitions.id   AS competition_id,
+          competitions.name AS competition_name',
+                           )
+
+      render json: competitions.as_json(
+        only: %w[competition_id competition_name city_name country_id start_date competing_status],
```
pranav-027: I guess `competing_status` isnt needed here

app/controllers/admin/persons_controller.rb
```diff
@@ -49,6 +49,66 @@ def results
...
```
pranav-027: and here

app/controllers/admin/persons_controller.rb
```diff
@@ -49,6 +49,66 @@ def results
...
+    def registrations
+      person_wca_id = params.require(:wcaId)
+
+      person = Person.current.find_by!(wca_id: person_wca_id)
+      registrations = person.user.registrations
+                            .joins(:competition)
+                            .where('competitions.start_date > ?', Date.current)
+                            .order('competitions.start_date DESC')
+                            .select(
+                              'registrations.*,
```
pranav-027: Is it necessary to use `registrations.*` here or would it be possible to query only the fields needed for the response?

app/webpacker/components/Panel/pages/CompetitorRegistrationPage/index.jsx
```diff
@@ -0,0 +1,106 @@
...
+
+async function getCompetitions({ wcaId }) {
+  const { data } = await fetchJsonOrError(
+    viewUrls.persons.registrations(wcaId),
+  );
+  return data || {};
+}
+
+function CompetitorRegistrationPage() {
+  const [formValues, setFormValues] = useState({});
```
gregorbg: Why are you creating a whole `formValues` object if your only purpose is to store a WCA ID? You might just as well do `const [wcaId, setWcaId] = useState();`...

app/webpacker/components/Panel/pages/CompetitorRegistrationPage/index.jsx
```diff
@@ -0,0 +1,106 @@
...
+    queryFn: () => getCompetitions({ wcaId }),
+    enabled: !!wcaId,
+  });
+
+  const handleWcaIdChange = (_, { value: newWcaId }) => setFormValues(
+    (prev) => ({
+      ...prev,
+      wcaId: newWcaId,
+    }),
+  );
```
gregorbg: If you have a `setWcaId` from `useState` above, you can use our custom `useInputUpdater` hook to wrap it in a way that SemUI understands.

app/webpacker/components/Panel/pages/CompetitorRegistrationPage/index.jsx
```diff
@@ -0,0 +1,106 @@
...
+              <Table.HeaderCell>Country</Table.HeaderCell>
+              <Table.HeaderCell>Date</Table.HeaderCell>
+              <Table.HeaderCell>Status</Table.HeaderCell>
+            </Table.Row>
+          </Table.Header>
+
+          <Table.Body>
+            {competitionsList.map((row) => (
+              <Table.Row key={row.competition_id}>
+                <Table.Cell><a href={`/competitions/${row.competition_id}`}>{row.competition_name}</a></Table.Cell>
```
gregorbg: We have link helpers which you can (and should) import from `lib/requests/routes.js.erb`. I believe you need `competitionUrl(row.competition_id)`.

app/models/user.rb
```diff
@@ -753,6 +756,9 @@ def self.panel_list
           panel_pages[:downloadVoters],
           panel_pages[:bannedCompetitors],
           panel_pages[:delegateProbations],
+          panel_pages[:competitorRegistration],
+          panel_pages[:organizedCompetitions],
+          panel_pages[:delegatedCompetitions],
```
gregorbg: I don't see the need to put these into three separate panel pages

app/controllers/admin/persons_controller.rb
```diff
@@ -49,6 +49,66 @@ def results
...
+      person = Person.current.find_by!(wca_id: person_wca_id)
+      competitions = person.user.delegated_competitions
+                           .order('competitions.start_date DESC')
+                           .where.not(competitions: { announced_at: nil })
+                           .where(competitions: { cancelled_at: nil })
+                           .select(
+                             'competitions.*,
+          competitions.id   AS competition_id,
+          competitions.name AS competition_name',
+                           )
```
gregorbg:
```suggestion
                           .over.visible.not_cancelled
                           .order(start_date: :desc)
```

gregorbg: I don't think your renaming is necessary if you're only loading the competition models and nothing else. You can just use the existing `id` and `name` fields in that case.

app/controllers/admin/persons_controller.rb
```diff
@@ -49,6 +49,66 @@ def results
...
+      person = Person.current.find_by!(wca_id: person_wca_id)
+      competitions = person.user.organized_competitions
+                           .order('competitions.start_date DESC')
+                           .where.not(competitions: { announced_at: nil })
+                           .where(competitions: { cancelled_at: nil })
+                           .select(
+                             'competitions.*,
+          competitions.id   AS competition_id,
+          competitions.name AS competition_name',
+                           )
```
gregorbg: Exact same remarks as for `delegated_competitions`

app/controllers/admin/persons_controller.rb
```diff
@@ -49,6 +49,66 @@ def results
       )
     end
 
+    def registrations
+      person_wca_id = params.require(:wcaId)
+
+      person = Person.current.find_by!(wca_id: person_wca_id)
+      registrations = person.user.registrations
+                            .joins(:competition)
+                            .where('competitions.start_date > ?', Date.current)
```
gregorbg:
```suggestion
                            .merge(Competition.not_over)
```

app/controllers/admin/persons_controller.rb
```diff
@@ -49,6 +49,66 @@ def results
...
+          competitions.id   AS competition_id,
+          competitions.name AS competition_name,
+          competitions.city_name,
+          competitions.country_id,
+          competitions.start_date',
+                            )
+
+      render json: registrations.as_json(
+        only: %w[competition_id competition_name city_name country_id start_date competing_status],
+      )
```
gregorbg: Rails serialization actually can do nested shenanigans:

```ruby
registrations.as_json(
  only: %w[...],
  include: {
    competitions: { only: %w[id, name, city_name, ...] }
  }
)
```

gregorbg (changes requested):
Thanks for your submission! The use-case makes perfect sense, however I cannot help but wonder why you're strictly separating the three queries into three panel pages?

You could just have one panel page with a WCA ID search box and let it display all three "statistics". Or you could maybe as a compromise have something like an Accordion or Tabs element to separate between the three queries' on the same page. 

But I don't think what you're doing in this PR is computationally expensive enough to justify running it so separately.

Brandon-Reichman:
Thanks for the detailed review. I've put this all into one page and updated the back-end queries.

app/controllers/admin/persons_controller.rb
```diff
@@ -49,6 +49,51 @@ def results
...
+      registrations = person.user.registrations
+                            .joins(:competition)
+                            .merge(Competition.not_over)
+                            .order(start_date: :asc)
+
+      render json: registrations.as_json(
+        only: %w[competing_status],
+        include: {
+          competition: {
+            only: %w[id name city_name country_id start_date],
```
gregorbg, line 65: Competitions by default serialize the Delegates and Organizers, which is rather costly. If you're not interested in that information (for the purpose of your PR) then you should explicitly put `include: []` to override the default behavior (it's not pretty but Rails works this way...)

app/webpacker/components/Panel/pages/HelpfulQueriesPage/index.jsx
```diff
@@ -0,0 +1,225 @@
...
+import { useQuery } from '@tanstack/react-query';
+import { IdWcaSearch } from '../../../SearchWidget/WcaSearch';
+import SEARCH_MODELS from '../../../SearchWidget/SearchModel';
+import { fetchJsonOrError } from '../../../../lib/requests/fetchWithAuthenticityToken';
+import { viewUrls, competitionUrl } from '../../../../lib/requests/routes.js.erb';
+
+const hasWcaId = (val) => typeof val === 'string' && val.trim().length > 0;
+
+const statusColor = (s) => {
+  switch ((s || '').toLowerCase()) {
```
gregorbg: What about `s?.toLowerCase()`?

app/webpacker/components/Panel/pages/HelpfulQueriesPage/index.jsx
```diff
@@ -0,0 +1,225 @@
...
+import {
+  Form, Header, Label, Loader, Message, Segment, Tab, Table,
+} from 'semantic-ui-react';
+import { useQuery } from '@tanstack/react-query';
+import { IdWcaSearch } from '../../../SearchWidget/WcaSearch';
+import SEARCH_MODELS from '../../../SearchWidget/SearchModel';
+import { fetchJsonOrError } from '../../../../lib/requests/fetchWithAuthenticityToken';
+import { viewUrls, competitionUrl } from '../../../../lib/requests/routes.js.erb';
+
+const hasWcaId = (val) => typeof val === 'string' && val.trim().length > 0;
```
gregorbg: I genuinely believe that we have no other occurence of `typeof` in our whole frontend code, and I'd like to keep it that way. What is the intention behind this type check? Or in other words, why is `val && val.trim().length > 0` or similar not good enough?

app/webpacker/components/Panel/pages/HelpfulQueriesPage/index.jsx
```diff
@@ -0,0 +1,225 @@
...
+  const { data } = await fetchJsonOrError(viewUrls.persons.registrations(wcaId));
+  const arr = data || [];
+  return arr.map((r) => ({
+    id: r.competition.id,
+    name: r.competition.name,
+    city_name: r.competition.city_name,
+    country_id: r.competition.country_id,
+    start_date: r.competition.start_date,
+    status: r.competing_status,
+  }));
```
gregorbg: Why do you need to map this down to such a specific format? You can just access `row.competition.id` or `row.competition.country_id` down in the table code...

gregorbg: If you really, _really_ need to flatten the data like this, then `{ ...r, ...r.competition }` will "fold" all competition keys to the top level automatically for you (although I'd still argue that returning the raw data itself is fine)

app/webpacker/components/Panel/pages/HelpfulQueriesPage/index.jsx
```diff
@@ -0,0 +1,225 @@
...
+            name="wcaId"
+            control={IdWcaSearch}
+            model={SEARCH_MODELS.person}
+            multiple={false}
+            value={wcaId}
+            onChange={(_, { value }) => setWcaId(value || '')}
+          />
+        </Form>
+      </Segment>
+      <Tab panes={panes} />
```
gregorbg: I would suggest wrapping this in your own custom component that gets passed `wcaId`, so that you can do the "has a WCA ID been entered? If no, display a sad alert message, if yes display the tab with the three panes that you care about" logic centrally in one single place

app/webpacker/components/Panel/pages/HelpfulQueriesPage/index.jsx
```diff
@@ -0,0 +1,225 @@
...
+            ))}
+          </Table.Body>
+        </Table>
+      )}
+    </>
+  );
+}
+
+function HelpfulQueriesPage() {
+  const [wcaId, setWcaId] = useState();
```
gregorbg: Suggest to use `useInputState` here, which is a "convenience wrapper around a normal `useState`" style hook that we defined ourselves in `lib/hooks/useInputState.js`. It will make it so that the `setWcaId` function automatically "understands" the params passed by the Semantic UI `Form.Field` component

app/webpacker/components/Panel/pages/HelpfulQueriesPage/index.jsx
```diff
@@ -0,0 +1,225 @@
...
+      <Segment>
+        <Form>
+          <Form.Field
+            label="WCA ID"
+            name="wcaId"
+            control={IdWcaSearch}
+            model={SEARCH_MODELS.person}
+            multiple={false}
+            value={wcaId}
+            onChange={(_, { value }) => setWcaId(value || '')}
```
gregorbg: Just pass `onChange={setWcaId}` directly as a raw param here, if you use `useInputState` as suggested in my other comment above

app/webpacker/components/Panel/pages/HelpfulQueriesPage/index.jsx
```diff
@@ -0,0 +1,225 @@
...
+    queryFn: () => fetchDelegated(wcaId),
+    enabled,
+  });
+
+  if (!enabled) return <Message info content="Select a WCA ID to load delegated competitions." />;
+  if (!isFetching && data.length === 0) return <Message content="No delegated competitions found." />;
+
+  return (
+    <>
+      <Loader active={isFetching} inline />
```
gregorbg: You can return early with a `if (isFetching) { return <Loader ...props... /> }` style query, which has the advantage that implicitly, `data` will exist after you passed that early return.

app/webpacker/components/Panel/pages/HelpfulQueriesPage/index.jsx
```diff
@@ -0,0 +1,225 @@
...
+    enabled,
+  });
+
+  if (!enabled) return <Message info content="Select a WCA ID to load organized competitions." />;
+  if (!isFetching && data.length === 0) return <Message content="No organized competitions found." />;
+
+  return (
+    <>
+      <Loader active={isFetching} inline />
+      {data.length > 0 && (
```
gregorbg: You can merge the `data.length` part by turning the `data.length > 0 && (<Table />)` part into a `data.length > 0 ? Table : Message` ternary expression. Especially if you leverage an early return with `isFetching` as explained in another review comment of mine.

Brandon-Reichman:
A bit slower turnover this time. :sweat_smile: I've addressed everything you've mentioned.

app/webpacker/components/Panel/pages/HelpfulQueriesPage/index.jsx
```diff
@@ -0,0 +1,212 @@
...
+import { IdWcaSearch } from '../../../SearchWidget/WcaSearch';
+import SEARCH_MODELS from '../../../SearchWidget/SearchModel';
+import { fetchJsonOrError } from '../../../../lib/requests/fetchWithAuthenticityToken';
+import { viewUrls, competitionUrl } from '../../../../lib/requests/routes.js.erb';
+import useInputState from '../../../../lib/hooks/useInputState';
+
+const hasWcaId = (val) => val?.trim()?.length > 0;
+
+const statusColor = (s) => {
+  switch ((s?.toLowerCase() || '')) {
```
gregorbg: Tiny nitpick: I still don't see why the `|| ''` part is necessary here. If for some reason `s` is null or undefined, then `s?.toLowerCase()` will become null and that `null` value will trigger the `default` branch the same way that an empty string literal `''` would.

gregorbg (approved):

app/webpacker/components/Panel/pages/HelpfulQueriesPage/index.jsx
Brandon-Reichman: I've updated this

PR #12308, opened 2025-08-25
Use Active Record to set foreign key checks to false in dev dump

FinnIckler:
This could be it

### DIFF
--- a/lib/database_dumper.rb
+++ b/lib/database_dumper.rb
@@ -1227,9 +1227,9 @@ module DatabaseDumper
 
     # Turn of foreign key checking to avoid errors when dumping data caused by foreign keys referencing not yet
     # existing rows.
-    DatabaseDumper.mysql("SET unique_checks=0", dump_db_name)
-    DatabaseDumper.mysql("SET foreign_key_checks=0", dump_db_name)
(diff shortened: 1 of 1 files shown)


lib/database_dumper.rb
```diff
@@ -1259,6 +1259,11 @@ def self.with_dumped_db(dump_config_name, dump_sanitizers, dump_ts_name = nil, d
       end
 
       ActiveRecord::Base.connection.execute("INSERT INTO #{dump_db_name}.server_settings (name, value, created_at, updated_at) VALUES ('#{dump_ts_name}', UNIX_TIMESTAMP(), NOW(), NOW())") if dump_ts_name.present?
+
+      # Turn these back on. We do establish a new connection again in the ensure block, but just in case this carries over
+      ActiveRecord::Base.connection.execute("SET unique_checks=0")
+      ActiveRecord::Base.connection.execute("SET foreign_key_checks=0")
+      ActiveRecord::Base.connection.execute("SET autocommit=0")
```
gregorbg: This should all be 1s, not 0s

gregorbg (changes requested):

gregorbg (approved):

PR #12312, opened 2025-08-25
Only set foreign_key_checks to false in dev dump

FinnIckler:
I think we just tanked the database by not committing the results and other big tables immediately when exporting it. 

image

These other two options were necessary in the dev dump load, but not here


gregorbg:
I can see that the `autocommit` is immediately relevant to the memory load. What about the unique key checks?

PR #12313, opened 2025-08-25
Move Next Competition Tabs into separate routes

FinnIckler:
I wanted to add my live pages into the tabs menu and noticed that it's all in one page which didn't really seem like the way to go. This also solves the routing by using separate routes which are rendered with RSC. The main / page can then be used to redirect to the other pages based on the old hash routing.

There is currently no data sharing, but that can be fixed with caching.

Thoughts?

### DIFF
--- a/next-frontend/src/app/(wca)/competitions/[competitionId]/page.tsx
+++ b/next-frontend/src/app/(wca)/competitions/[competitionId]/page.tsx
@@ -6,7 +6,6 @@ import {
   Heading,
   HStack,
   SimpleGrid,
-  Tabs,
   Text,
   Link as ChakraLink,
   VStack,


FinnIckler:
Moving 
The Tabs.Trigger around the component like 
```
<Tabs.Trigger value="general" asChild>
        <Link
          href={route({
           pathname: "/competitions/[competitionId]",
           query: { competitionId },
              })}
          >
          General
        </Link>
</Tabs.Trigger>
```
Instead of 
```
<Link
  href={route({
    pathname: "/competitions/[competitionId]",
    query: { competitionId },
  })}
>
  <Tabs.Trigger value="general">General</Tabs.Trigger>
</Link>
```
Reveals a big issue. If we want it to work like I outlined (RSC + separate Routes for each) we need to make the Tab List a client component, because otherwise whenever you want to reload the page the `defaultValue` is set, which means it doesn't render the correct Tab. 

You cannot use `usePathname` in a server component to infer the correct Value.
I will try it out, but if it will not work then we might need to look for a different solution

FinnIckler:
ok, it seems like this works very easily

next-frontend/src/components/ui/loading.tsx
```diff
@@ -1,11 +1,13 @@
...
+import { Box, Progress } from "@chakra-ui/react";
 
 export default function Loading() {
   return (
-    <Progress.Root value={null}>
-      <Progress.Track>
-        <Progress.Range />
-      </Progress.Track>
-    </Progress.Root>
+    <Box w="100%">
```
gregorbg: I'm very unhappy about this Box hack. I assume you have already investigated more thoroughly, but why is it necessary in the first place?

If anything, then I would be willing to accept some `w="100%"` being passed to `Progress.Root` directly. But even then, it still feels like that's a bug/oversight in Chakra...

FinnIckler: It was actually related to the` <Tabs.Content> `not being part of the layout

PR #12315, opened 2025-08-26
Remove import results button from posting dashboard & admin page

danieljames-dj:
This change is to force WRT to post results through tickets, and use old import-results page only if it's really necessary. The old import-results page is still accessible through competition nav.

We can keep this for few days (maybe a week or so) and if we don't receive any complaints from WRT, the import-results page can be removed completely.


gregorbg (approved):
Approving from a technical perspective. It is up to WRT when exactly this feature is actually ready to go live.

PR #12317, opened 2025-08-26
Fixes in ticket comments

danieljames-dj:
Doing couple of fixes in ticket comments:
1. By the introduction of competition stakeholder, the comments stopped sending email because stakeholder no longer had email object. This is now fixed.
2. When "create comment" button was clicked, it was not going to loading state because we were not using `isPending`. This is also now fixed.

### DIFF
--- a/app/models/ticket_comment.rb
+++ b/app/models/ticket_comment.rb
@@ -9,19 +9,19 @@ class TicketComment < ApplicationRecord
     include: %w[acting_user],
   }.freeze
 
-  private def stakeholders_to_email_notify
+  private def stakeholder_emails_to_notify
     stakeholders = ticket.ticket_stakeholders.includes(:stakeholder)
 
     stakeholders = stakeholders.reject { |stakeholder| stakeholder.id == acting_stakeholder_id } if acting_stakeholder.user_stakeholder?
 
-    stakeholders.map(&:stakeholder)
+    stakeholders.flat_map(&:emails)
   end
 
   after_create :notify_stakeholders
   private def notify_stakeholders
-    recipient_emails = stakeholders_to_email_notify.map(&:email)
+    recipient_emails = stakeholder_emails_to_notify
 
-    TicketsMailer.notify_create_ticket_comment(self, recipient_emails).deliver_later
+    TicketsMailer.notify_create_ticket_comment(self, recipient_emails).deliver_now # TODO: Change to deliver_later
   end
 
   def author_text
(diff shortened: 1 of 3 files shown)


app/models/ticket_comment.rb
```diff
@@ -9,19 +9,19 @@ class TicketComment < ApplicationRecord
...
     stakeholders = stakeholders.reject { |stakeholder| stakeholder.id == acting_stakeholder_id } if acting_stakeholder.user_stakeholder?
 
-    stakeholders.map(&:stakeholder)
+    stakeholders.flat_map(&:emails)
   end
 
   after_create :notify_stakeholders
   private def notify_stakeholders
-    recipient_emails = stakeholders_to_email_notify.map(&:email)
+    recipient_emails = stakeholder_emails_to_notify
```
gregorbg: This line now seems redundant. Why are you not accessing `stakeholder_emails_to_notify` directly in the places where you used to access `recipient_emails`?

app/models/ticket_comment.rb
```diff
@@ -9,19 +9,19 @@ class TicketComment < ApplicationRecord
...
 
-    TicketsMailer.notify_create_ticket_comment(self, recipient_emails).deliver_later
+    TicketsMailer.notify_create_ticket_comment(self, recipient_emails).deliver_now # TODO: Change to deliver_later
```
gregorbg: What is keeping you from using `deliver_later`?

app/models/ticket_stakeholder.rb
```diff
@@ -34,6 +34,18 @@ def user_stakeholder?
     stakeholder_type == "User"
   end
 
+  def competition_stakeholder?
+    stakeholder_type == "Competition"
+  end
+
+  def emails
+    if competition_stakeholder?
+      stakeholder.delegates.map(&:email)
```
gregorbg: Would `pluck(:email)` work here?

danieljames-dj: Oh yes, I missed that. Changed

app/models/ticket_comment.rb
danieljames-dj: Oh right, changed it.

app/models/ticket_comment.rb
danieljames-dj: I missed to revert after testing. Changed it now. sorry.

gregorbg (approved):

PR #12320, opened 2025-08-27
Show in edit person ticket if old value is in sync with person

danieljames-dj:
This is because there are chances that `old_value` can be different from the current person data - especially when the person details got updated through some other way - like direct PMA edit, or through edit person form.

If WRT is informed about the out-of-sync, it will help them to understand that something has changed since the request initially happened.

### DIFF
--- a/app/webpacker/components/Tickets/TicketWorkbenches/EditPersonActionerView/OldDataSyncInfo.jsx
+++ b/app/webpacker/components/Tickets/TicketWorkbenches/EditPersonActionerView/OldDataSyncInfo.jsx
@@ -1,0 +1,56 @@
+import { useMutation, useQueryClient } from '@tanstack/react-query';
+import React, { useMemo } from 'react';
+import { Button, Message } from 'semantic-ui-react';
+import syncEditPersonRequest from '../../api/editPerson/syncEditPersonRequest';
+import Loading from '../../../Requests/Loading';
+import Errored from '../../../Requests/Errored';
+
+export default function OldDataSyncInfo({ ticketDetails, currentStakeholder }) {
+  const {
+    ticket: {
+      id,
+      metadata: {
+        tickets_edit_person_fields: ticketsEditPersonFields,
+        person,
+      },
+    },
+  } = ticketDetails;
+
+  const queryClient = useQueryClient();
+  const {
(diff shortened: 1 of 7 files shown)


danieljames-dj:
Ya definitely WRT likes to have the latest value for old value, but not every time. The out-of-sync can be considered as a signal for WRT to understand that something has changed since the request came up.

I understand this might cause an extra work for WRT - that they need to check what had happened in between. But in many cases, it's worth knowing that because there are high chances that there can be a conflict.

For example, let's assume John Doe wants to add a middle name "A." or "B.". He first sent an email to WRT asking to add middle name "A.", and then raised a request through ticket to add middle name "B.". Let's assume somebody attended the email first, and fixed the name using 'edit person' form. Now when they check the ticket, the ticket will say that request is from "John Doe" to "John B. Doe", but his name was already changed once earlier. Now WRT to ask for a clarification what exactly John wants. At the ticket, if WRT wants to know that there is some conflicts, it's better if it's informed.

On a side note: this is indeed a temporary solution. Once we have the full system in-place, then these can be removed, and even "old value" column can also be removed.

app/webpacker/components/Tickets/TicketWorkbenches/EditPersonActionerView/OldDataSyncInfo.jsx
```diff
@@ -0,0 +1,56 @@
...
+    isError,
+    error,
+  } = useMutation({
+    mutationFn: syncEditPersonRequest,
+    onSuccess: () => {
+      // We've invalidated the ticket details because we can't confirm that the updated value
+      // matches the current person's data. To be safe, we're refetching the information, as
+      // the person's details may have changed since the last time they were accessed.
+      queryClient.invalidateQueries(['ticket-details', id]);
+    },
```
gregorbg: You can send the serialized ticket (in its current, brand-new updated form) as a success response from the `sync_edit_person_request` controller method. Then you can use `setQueryData` instead of `invalidateQueries` here, which saves you one network request 👍

danieljames-dj: Makes sense, done.

PR #12323, opened 2025-08-28
Add features to edit the requested changes in edit person ticket

danieljames-dj:
There can be cases where WRT needs to modify the request a bit instead of rejecting it. For example, if the user accidentally gives the name as "Jonh" instead of "John", and if ID clearly mentioned that the name is "John", then it will be good if WRT is allowed to edit the change field.

Currently the advantage of this changing will be that the ticket will be updated with what exactly is going to happen. In future, we will have action buttons like "Execute as fix" and "Execute as update". Once we have that buttons, then we will be using this value in database to perform the action.

### DIFF
--- a/app/webpacker/components/Tickets/TicketWorkbenches/EditPersonActionerView/EditPersonFieldEditor.jsx
+++ b/app/webpacker/components/Tickets/TicketWorkbenches/EditPersonActionerView/EditPersonFieldEditor.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useCallback } from 'react';
 import { Form } from 'semantic-ui-react';
 import useInputState from '../../../../lib/hooks/useInputState';
 import I18n from '../../../../lib/i18n';
@@ -15,16 +15,18 @@ export default function EditPersonFieldEditor({
   actionMutate,
 }) {
   const [newValue, setNewValue] = useInputState(oldValue);
+
+  const formSubmitHandler = useCallback(() => actionMutate({
+    ticketId,
+    actingStakeholderId,
+    fieldName,
+    oldValue,
+    newValue,
+    editPersonFieldId: id,
+  }), [actingStakeholderId, actionMutate, fieldName, id, newValue, oldValue, ticketId]);
+
   return (
-    <Form onSubmit={() => actionMutate({
-      ticketId,
-      actingStakeholderId,
-      fieldName,
-      oldValue,
-      newValue,
-      editPersonFieldId: id,
-    })}
-    >
+    <Form onSubmit={formSubmitHandler}>
       <FormInput
         fieldName={fieldName}
         newValue={newValue}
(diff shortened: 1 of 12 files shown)


app/webpacker/components/Tickets/TicketWorkbenches/EditPersonActionerView/EditPersonFieldEditor.jsx
```diff
@@ -0,0 +1,80 @@
...
+  const [newValue, setNewValue] = useInputState(oldValue);
+  return (
+    <Form onSubmit={() => actionMutate({
+      ticketId,
+      actingStakeholderId,
+      fieldName,
+      oldValue,
+      newValue,
+      editPersonFieldId: id,
+    })}
```
gregorbg: You should pull this out into a callback before the `return`

app/webpacker/components/Tickets/TicketWorkbenches/EditPersonActionerView/EditPersonRequestedChanges.jsx
```diff
@@ -0,0 +1,178 @@
...
+        }) => setEditPersonFieldActionDetails({
+          action: 'update',
+          id,
+          fieldName,
+          oldValue,
+        })}
+        deleteChange={(editPersonFieldId) => setEditPersonFieldActionDetails({
+          action: 'delete',
+          editPersonFieldId,
+        })}
```
gregorbg: Please move them into separate callbacks instead of declaring these "in-line"

app/webpacker/components/Tickets/TicketWorkbenches/EditPersonActionerView/EditPersonRequestedChanges.jsx
```diff
@@ -0,0 +1,178 @@
...
+            ...oldTicketDetails.ticket,
+            metadata: {
+              ...oldTicketDetails.ticket.metadata,
+              tickets_edit_person_fields: [
+                ...oldTicketDetails.ticket.metadata.tickets_edit_person_fields,
+                newEditField,
+              ],
+            },
+          },
+        }),
```
gregorbg: This kind of "given an old ticket, update some specific part of it" code appears three times within the same file. You absolutely have to come up with some helper code to deduplicate, please. And do consider using that helper code as well in the ~50 other places throughout your ticket frontend that do something similar :P

gregorbg (changes requested):

app/webpacker/components/Tickets/TicketWorkbenches/EditPersonActionerView/EditPersonFieldEditor.jsx
danieljames-dj: Done

app/webpacker/components/Tickets/TicketWorkbenches/EditPersonActionerView/EditPersonRequestedChanges.jsx
danieljames-dj: Done

app/webpacker/components/Tickets/TicketWorkbenches/EditPersonActionerView/EditPersonRequestedChanges.jsx
danieljames-dj: Okay, I've moved the codes in this file to a separate method. The same method can be used for `updateStatus`, but I feel it will go out of context. I'll do that in one of the follow-up PRs where I'll anyway be changing schema of `status` and hence there will be bunch of changes related to that. Hope this works for you.

danieljames-dj (commented):
Attended the comments

gregorbg (approved):

PR #12327, opened 2025-08-29
Make round reference in Results table non-null

gregorbg:
See title (sorry but this is really obvious if you look at the diff as well)


gregorbg:
FWIW We are already automatically setting `round_id` for every results row since June, when https://github.com/thewca/worldcubeassociation.org/pull/11709 was merged.

That PR also included a backfilling script, which has been run successfully. As of writing this comment, running `SELECT * FROM results WHERE round_id IS NULL` yields an empty result.

next-frontend/src/types/nextjs-routes.d.ts
gregorbg, line 1: merge with main to remove this changeset from the diff

PR #12328, opened 2025-08-29
Next competition podiums

FinnIckler:
image

Looks very plain right now, maybe we put some of the cards on the top of this like it is with the current page

### DIFF
--- a/next-frontend/src/components/competitions/Cards.tsx
+++ b/next-frontend/src/components/competitions/Cards.tsx
@@ -463,7 +463,7 @@ export function InfoCard({
             </Card.Header>
             <Card.Body>
               <Text>{competitionInfo.city}, </Text>
-              <CountryMap code={competitionInfo.country_iso2} bold />
+              <CountryMap code={competitionInfo.country_iso2} bold t={t} />
             </Card.Body>
           </Card.Root>
 


next-frontend/src/app/(wca)/competitions/[competitionId]/layout.tsx
```diff
@@ -9,10 +10,13 @@ export default async function CompetitionLayout({
   params: Promise<{ competitionId: string }>;
 }) {
   const { competitionId } = await params;
+  const competitionInfo = (await getCompetitionInfo(competitionId))!.data;
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/layout.tsx]:
```
const { data, error } = await getCompetitionInfo(competitionId);

if (error) return <SomeFailedComponent />

// data is implicitly not-null from here on because TypeScript is smart
```

Comment (gregorbg) [reply next-frontend/src/app/(wca)/competitions/[competitionId]/layout.tsx]: Sidenote: You're already doing it correctly in the `podiums/page.tsx`

next-frontend/src/app/(wca)/competitions/[competitionId]/podiums/page.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  if (!competitionInfo) {
+    return <Text>Competition does not exist</Text>;
+  }
+
+  const { t } = await getT();
+
+  const podiumsRequest = await getPodiums(competitionId);
+
+  const podiumResults = podiumsRequest.data!;
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/podiums/page.tsx]: Please do it cleanly like with `getCompetitionInfo` above

gregorbg (changes requested):

PR #12331, opened 2025-09-01
Add 'aggressive retry' flag to cronjob config

gregorbg:
Normally, we don't want cronjobs that errored out to be re-scheduled, because Sidekiq has its own retry mechanism.

However, in the case of the arrogant Google Admin SDK, which enforces draconius rate limits but at the same time does not offer any way of batching requests, we do want to hard-retry upon failure without the exponential backoff of Sidekiq retries. So we override the default mechanism to make sure that our mailing lists stay in sync.


gregorbg:
Closing because we "solved" the problem for the time being by reducing the cronjob timing.

PR #12332, opened 2025-09-01
Fix `signed_id` issue in Delegate Reports

dunkOnIT:
Sometimes users will experience the following error when trying to work with a Delegate Report: 
```bash 
ActionView::Template::Error (Cannot get a signed_id for a new record):
```

Gregor explains the cause in detail [here](https://github.com/rails/rails/issues/50234) - in brief, a user uploading a valid image on a delegate report that fails validations causes an image to be attached but not persisted, yielding the error. 

The easy fix suggested in that issue discussion is to check `.persisted?` before trying to render the image, which is what this PR does.

### DIFF
--- a/app/views/delegate_reports/edit.html.erb
+++ b/app/views/delegate_reports/edit.html.erb
@@ -28,13 +28,15 @@
         <% if @delegate_report.setup_images.attached? %>
           <div class="row">
             <% @delegate_report.setup_images.each do |image| %>
-              <div class="col-sm-1 thumb">
-                <%= image_tag image.variant(:preview).processed %>
+              <% if image.blob.persisted? %>
+                <div class="col-sm-1 thumb">
+                  <%= image_tag image.variant(:preview).processed %>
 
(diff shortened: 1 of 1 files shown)


gregorbg (commented):
This prevents the page from crashing upon display, but it does not solve the underlying issue.

If there is a `setup_image` which is _not_ marked as `persisted?`, then it means that "Somebody tried to upload this image at some point, but because of reasons :tm: it did not get saved to our storage disk (S3 in this case)". This commonly happens when validations on the whole `DelegateReport` model fail, and it leaves behind a "phantom image": The database structure for marking an upload as "this upload exists" is there, but the actual bytes backing it are not there.

This can cause headaches in case of Delegate Reports because we have a "minimum number of required images", which is based on "number of images that the database _thinks_ exist". There is no way (to my knowledge) of counting "number of images which really exist and are uploaded".

The sad reality is that this is a workaround for displaying the page, not a real fix for the issue. The fix has to come from Rails-land.

PR #12333, opened 2025-09-01
Add verify warnings as action & add to misc actions

danieljames-dj:
Earlier "verify warnings" internally calling "update status" action. With this PR, it will call the new "verify actions", and hence reduce dependency on update status action.

### DIFF
--- a/app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/MiscActions.jsx
+++ b/app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/MiscActions.jsx
@@ -2,8 +2,9 @@ import React from 'react';
 import { Header, List, Segment } from 'semantic-ui-react';
 import AbortProcess from './MiscActions/AbortProcess';
 import EventsMergedData from './MiscActions/EventsMergedData';
+import WarningsAndMessageButton from './MiscActions/WarningsAndMessageButton';
 
-export default function MiscActions({ ticketDetails, updateStatus }) {
+export default function MiscActions({ ticketDetails }) {
   return (
     <Segment>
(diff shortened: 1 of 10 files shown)


app/webpacker/components/Tickets/TicketWorkbenches/CompetitionResultActionerView/MiscActions.jsx
```diff
@@ -2,23 +2,28 @@ import React from 'react';
...
-export default function MiscActions({ ticketDetails, updateStatus }) {
+export default function MiscActions({ ticketDetails }) {
   return (
     <Segment>
       <Header>Misc Actions</Header>
       <List>
         <List.Item>
           <AbortProcess
             ticketDetails={ticketDetails}
-            updateStatus={updateStatus}
```
gregorbg, line 14: Just curious, why can this line go away and what does it have to do with the current PR?

danieljames-dj: Hm actually not related. `AbortProcess` never needed `updateStatus` - I introduced `updateStatus` for this class in https://github.com/thewca/worldcubeassociation.org/pull/12162, but in the same PR mostly I've removed the dependency and forgot to remove from here.

In this PR, I wanted to remove `updateStatus` completely from `CompetitionResultActionerView`, and since I found it no longer used in `AbortProcess`, I did it in same PR. Should I move it to a separate PR?

gregorbg: Nah it's fine. Just wanted to understand the diff :D

PR #12359, opened 2025-09-01
Add Country translations to translation.jsons using third party library

FinnIckler:
I decided to this route (adding it to the json) instead of using the library directly in the code for the following reasons (I started with using it in the code directly):

- Much easier to overwrite language names
- saves a bunch of space not shipping the country json (which has so much more data in it)
- The library already [anticipates](https://github.com/michaelwittig/node-i18n-iso-countries?tab=readme-ov-file#installing) the space issue so we would need to have a file that calls `countries.registerLocale(...));` 20 times. We can't do this dynamically because dynamic importing would need async components.
- Much cleaner implementation by using the t function
- This is basically how we do it in rails (library injects something into the translations)

### DIFF
--- a/next-frontend/build-translations.js
+++ b/next-frontend/build-translations.js
@@ -30,11 +30,6 @@ function addCountryNames(translation, lang) {
   }
   const iso3166CountryCodes = Object.keys(countries.getAlpha2Codes());
 
-  // Load the language file for the country names
-  countries.registerLocale(
-    require(`i18n-iso-countries/langs/${iso639LanguageCode}.json`),
-  );
-
   iso3166CountryCodes.forEach((iso3166CountryCode) => {
     const languageKey = `countries.${iso3166CountryCode}`;
     if (!translation[languageKey]) {
--- a/next-frontend/build-translations.js
+++ b/next-frontend/build-translations.js
@@ -5,6 +5,7 @@ const fs = require("fs");
 const path = require("path");
 const yaml = require("js-yaml");
 const languages = Object.keys(require("./src/lib/i18n/locales/available.json"));
+const countries = require("i18n-iso-countries");
 
 // Recursively flatten a nested object using dot notation
 function flattenObject(obj, prefix = "") {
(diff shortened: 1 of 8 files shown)


next-frontend/build-translations.js
```diff
@@ -19,6 +20,26 @@ function flattenObject(obj, prefix = "") {
   }, {});
 }
 
+// Add Country Names if not translated already
+function addCountryNames(translation, lang) {
+  const supportedLanguages = countries.getSupportedLanguages();
+  if (supportedLanguages.indexOf(lang) === -1) {
```
gregorbg:
```suggestion
  if (!supportedLanguages.includes(lang)) {
```

next-frontend/build-translations.js
```diff
@@ -19,6 +20,26 @@ function flattenObject(obj, prefix = "") {
   }, {});
 }
 
+// Add Country Names if not translated already
+function addCountryNames(translation, lang) {
```
gregorbg, line 24: This method could benefit from a little bit of documentation inside its method body. I am mainly confused about all the different variations of `code` appearing here and there

next-frontend/src/components/CountryMap.tsx
```diff
@@ -1,16 +1,22 @@
...
 
 type CountryMapProps = {
   code: string;
   bold?: boolean;
 };
 
 const CountryMap = ({ code, bold = false }: CountryMapProps) => {
-  const country = countries.byIso2[code.toUpperCase()].id || "Unknown";
-  return <Text fontWeight={bold ? "bold" : "normal"}>{country}</Text>;
+  const { t } = useT();
```
gregorbg: This means that you are forced to use `CountryMap` as a client component. In other situations I have tried passing in `t` as a prop, which version do you think makes more sense here?

FinnIckler: I think passing t as a prop is better

next-frontend/build-translations.js
```diff
@@ -19,6 +20,34 @@ function flattenObject(obj, prefix = "") {
...
+
+  if (!supportedLanguages.includes(iso639LanguageCode)) {
+    return translation;
+  }
+  const iso3166CountryCodes = Object.keys(countries.getAlpha2Codes());
+
+  // Load the language file for the country names
+  countries.registerLocale(
+    require(`i18n-iso-countries/langs/${iso639LanguageCode}.json`),
+  );
```
gregorbg: I am not sure whether this `registerLocale` logic is required, because the code you're writing here is running on the server. Quote from the library README:
> Install it using npm: `npm install i18n-iso-countries`
> ```
> var countries = require("i18n-iso-countries");
> ```
> If you use i18n-iso-countries with Node.js, you are done.

gregorbg (changes requested):

next-frontend/build-translations.js
gregorbg: Right afterwards, the README also specifies:

> If you use the package in a browser environment, you have to register the languages you want to use [...]

but we are not using the package directly in a browser environment anyways

gregorbg (approved):

PR #12360, opened 2025-09-01
Allow delegates to upload a competition's scrambles

dunkOnIT:
Previously, permissions only allowed people who `can_admin_results` to access the Upload Scrambles page - this made it hard for our users to test, so I've fixed it to use the same permissions as Results Upload

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -4,8 +4,9 @@ require 'csv'
 
 class AdminController < ApplicationController
   before_action :authenticate_user!
-  before_action -> { redirect_to_root_unless_user(:can_admin_results?) }, except: %i[all_voters leader_senior_voters regional_voters]
+  before_action -> { redirect_to_root_unless_user(:can_admin_results?) }, except: %i[all_voters leader_senior_voters regional_voters upload]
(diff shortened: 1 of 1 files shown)


app/controllers/admin_controller.rb
```diff
@@ -4,8 +4,9 @@
 
 class AdminController < ApplicationController
   before_action :authenticate_user!
-  before_action -> { redirect_to_root_unless_user(:can_admin_results?) }, except: %i[all_voters leader_senior_voters regional_voters]
+  before_action -> { redirect_to_root_unless_user(:can_admin_results?) }, except: %i[all_voters leader_senior_voters regional_voters upload]
   before_action -> { redirect_to_root_unless_user(:can_see_eligible_voters?) }, only: %i[all_voters leader_senior_voters regional_voters]
+  before_action -> { redirect_to_root_unless_user(:can_upload_competition_results?, competition_from_params) }, except: :upload
```
gregorbg: Did you mean `only: :upload`?

dunkOnIT: Well this begs the question of how it worked at all when I tested locally with a delegate account

gregorbg: Because using `except` disables the check for the `upload` route, meaning it was completely unprotected and anyone could've accessed it

dunkOnIT: Ah of course and I'd excepted it from the other check as well

PR #12363, opened 2025-09-02
Add action buttons to approve edit person requests

danieljames-dj:
This was discussed with WRT and WRT is happy with this change.

This this change, WRT no longer needs to fill the "edit person" form, instead directly execute the action using the buttons like "Execute as Fix" or "Execute as Update"


app/controllers/api/v0/wrt/persons_controller.rb
```diff
@@ -26,29 +26,17 @@ def update
...
-    elsif params[:method] == "update"
-      if person.update_using_sub_id(edit_params)
-        render status: :ok, json: { success: "Successfully updated #{person.name}." }
-      else
-        render status: :unprocessable_entity, json: { error: "Error while updating #{person.name}." }
-      end
-    else
-      render status: :unprocessable_entity, json: { error: "Unknown method #{params[:method]}." }
-    end
+    render status: :ok, json: { success: true }
```
danieljames-dj, line 39: Sending proper message on person fixes/updates is not much useful. So returning success if it's success. Also not manually throwing `unprocessable_entity` anymore. Instead letting the code throw error if something goes wrong.

app/models/tickets_edit_person.rb
```diff
@@ -78,6 +80,12 @@ def self.create_ticket(wca_id, changes_requested, requester)
     end
   end
 
+  def out_of_sync
```
gregorbg: This method returns a boolean, so it should be called `out_of_sync?` per Ruby conventions

gregorbg (changes requested):

app/models/tickets_edit_person.rb
danieljames-dj: Oh yes, done.

gregorbg (approved):

PR #12365, opened 2025-09-02
Bugfix/helpful queries permission

Brandon-Reichman:
I initially put these functions in the admin/person_controller and it works great when **I** am logged in, but only me. Any other WIC member will not get results on any search. This is because everything in the admin/persons_controller is available to users who `can_admin_results?` giving me access as a WRT member, but not any other WIC members.

Just adding these functions to the AdminController's `before_action` (similar to the voter functions available to WIC) worked, but rubocop complained about them not being defined in the class. So I moved them to the admin_controller and changed the routes accordingly.

I've tested these changes locally with my user and multiple other WIC users and it's working for everyone.

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -211,6 +211,52 @@ class AdminController < ApplicationController
     voters User.regional_voters, "regional-wca-voters"
   end
 
+  def registrations
+    person_wca_id = params.require(:wcaId)
+
+    person = Person.current.find_by!(wca_id: person_wca_id)
+    registrations = person.user.registrations
+                          .joins(:competition)
+                          .merge(Competition.not_over)
+                          .order(start_date: :asc)
+
+    render json: registrations.as_json(
+      only: %w[competing_status],
+      include: {
+        competition: {
+          only: %w[id name city_name country_id start_date],
+          include: [],
+        },
+      },
+    )
+  end
+
+  def organized_competitions
+    person_wca_id = params.require(:wcaId)
+
+    person = Person.current.find_by!(wca_id: person_wca_id)
+    competitions = person.user.organized_competitions
+                         .over.visible.not_cancelled
+                         .order(start_date: :desc)
+
+    render json: competitions.as_json(
+      only: %w[id name city_name country_id start_date],
+    )
+  end
+
+  def delegated_competitions
+    person_wca_id = params.require(:wcaId)
+
+    person = Person.current.find_by!(wca_id: person_wca_id)
+    competitions = person.user.delegated_competitions
+                         .over.visible.not_cancelled
+                         .order(start_date: :desc)
+
+    render json: competitions.as_json(
+      only: %w[id name city_name country_id start_date],
+    )
+  end
+
   private def voters(users, filename)
     csv = CSV.generate do |line|
       users.each do |user|
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -4,8 +4,8 @@ require 'csv'
 
 class AdminController < ApplicationController
   before_action :authenticate_user!
-  before_action -> { redirect_to_root_unless_user(:can_admin_results?) }, except: %i[all_voters leader_senior_voters regional_voters]
-  before_action -> { redirect_to_root_unless_user(:can_see_eligible_voters?) }, only: %i[all_voters leader_senior_voters regional_voters]
+  before_action -> { redirect_to_root_unless_user(:can_admin_results?) }, except: %i[all_voters leader_senior_voters regional_voters registrations organized_competitions delegated_competitions]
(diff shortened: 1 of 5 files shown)


app/controllers/admin_controller.rb
```diff
@@ -4,8 +4,8 @@
 
 class AdminController < ApplicationController
   before_action :authenticate_user!
-  before_action -> { redirect_to_root_unless_user(:can_admin_results?) }, except: %i[all_voters leader_senior_voters regional_voters]
-  before_action -> { redirect_to_root_unless_user(:can_see_eligible_voters?) }, only: %i[all_voters leader_senior_voters regional_voters]
+  before_action -> { redirect_to_root_unless_user(:can_admin_results?) }, except: %i[all_voters leader_senior_voters regional_voters registrations organized_competitions delegated_competitions]
+  before_action -> { redirect_to_root_unless_user(:can_see_eligible_voters?) }, only: %i[all_voters leader_senior_voters regional_voters registrations organized_competitions delegated_competitions]
```
gregorbg: Is `can_see_eligible_voters?` really the correct permission to check for your new features?

Brandon-Reichman: Honestly no. I realized it was weird, but I think I'm a bit unfamiliar with these permissions and couldn't find an appropriate one. Do you have any suggestion? I'm thinking that it should be open enough so that if the panel page potentially gets added to another teams panel, they would already be able to access these. Would something like `can_access_panels` be right?

gregorbg: CC @danieljames-dj

danieljames-dj: Thanks @gregorbg for cc-ing me.

IMO these functions should be part of `app/controllers/users_controller` because all these details are actually connected to the user and not the person.

In the frontend, you can use `AdminWcaSearch` so that WIC can search with name or email address (I think even searching with WCA ID will work) and in backend you can get the corresponding `userId`.

Now coming to permission, we do have `can_access_panel?(:wic)`, but that method mentions that you have access to the panel, using it here might restrict the access of this page through "Admin panel".

Here you are actually trying to access some details of a user, and it can be **any** users, so the permission should most probably be `can_view_all_users?`. One argument against that will be that `can_view_all_users?` doesn't mean you can access any data of a user, but I would say it's fine.

Brandon-Reichman:
Thanks @danieljames-dj I've moved these functions to the users_controller, which seems to by default check for `can_view_all_users`. All of the data from these queries can be aggregated by anyone searching the website hard enough, so I'm not worried about the permissions going to the wrong users. 

I've also updated the routes and references of `wcaId` to `userId`.

I tried to use `AdminWcaSearch`, but anyway I tried to put it in caused `JSON.parse: unexpected character at line 1 column 1 of the JSON data` With no stack trace. However, just changing the model from `SEARCH_MODELS.person` to `SEARCH_MODELS.user` seems to let me search users without wca id's. This would also be useful for WIC so thanks for pointing that out.

app/webpacker/components/Panel/pages/HelpfulQueriesPage/index.jsx
```diff
@@ -148,63 +148,63 @@ function DelegatedPane({ wcaId }) {
...
+          {!hasUserId(userId)
+            ? <Message info content="Select a User to load delegated competitions." />
+            : <DelegatedPane userId={userId} />}
         </Tab.Pane>
       ),
     },
-  ]), [wcaId]);
+  ]), [userId]);
 
   return <Tab panes={panes} />;
```
danieljames-dj, line 170: Since you are using `hasUserId` in many places, you can try doing something like following before this line:

```
if (!hasUserId(userId)) return <Message info content="Select a User to load delegated competitions." />;
```

This way, you don't have to check if `userId` exists at many places.

Also, you can check if it's possible to replace the `!hasUserId(userId)` directly with `!userId` (it should most probably work)

danieljames-dj (commented):
Backend code LGTM. I've left a cleanup comment for the frontend.

app/webpacker/components/Panel/pages/HelpfulQueriesPage/index.jsx
Brandon-Reichman: Thanks @danieljames-dj, I've simplified the missing user id check in this section. I also tested just using `!userId` and it seems to work everywhere. So I've removed `hasUserId()`

danieljames-dj: Thanks @Brandon-Reichman. Now since the `Tab` is getting rendered only when `userId` is present, you can remove the `userId` check in all the above methods.

For example, in `DelegatedPane`, you don't have to set `enabled` so that it will be enabled by default. Same for other two components. And even in the API calls, since it will get called only if `userId` is available, you can avoid `userId` checks there.

danieljames-dj (approved):
Thanks Brandon. LGTM. Leaving this to @gregorbg for final review.

Brandon-Reichman:
@gregorbg Do you approve the latest changes?

Issue #12173, opened 2025-08-07
Allow Front Page Posts to be Approved/Queued by WCT


slongbehn:
**Is your feature request related to a problem? Please describe.**
WCT has to manage the timing of posts onto the WCA front page manually, informing leaders when their posting date is. While this is fine, it can make things fall through the cracks. Additionally, it doesn't stop a poster from forgetting to submit to WCT and posting anyway. Malicious actors can also post without any type of approval, and it could take time for someone to notice and take action.

**Describe the solution you'd like**
Ideally, those with posting privileges sends the post via the WCA website but it does not post. Instead, WCT is able to review it and set a posting date so that it is posted on that date by the person it's submitted by. This prevents mistakes and malicious posts.

**Describe alternatives you've considered**

- WCT currently manages it manually, and leaders need to set reminders when their posting date is coming up. This is how it currently operates, and is fine if this is too software intensive.
- The poster could also queue the post for the date based on WCT's timeline if the "approval" portion is too much of an extra software step

**Additional context**
The [WCA Website Posting policy](https://documents.worldcubeassociation.org/documents/policies/internal/Website%20Posting.pdf) outlines:
_Board members, Committees/Teams (excluding the WCAT and WRT), and/or Delegates wishing to post must contact the WCA Communication Team via email with their request and a draft of the information they would like included_.

FinnIckler:
this is very easy to do in the CMS we have in the new nextjs website, which is were post handling will be done in the future.

PR #12369, opened 2025-09-03
Nextjs post migration

### LINKED ISSUE #14258: Import posts from old website

### LINKED ISSUE #14408: Removed "Announced by..." in the Posts

### LINKED ISSUE #14441: Use custom names/titles instead of ID's for collection items (announcements, testimonials, etc)

FinnIckler:
This is a Proof of Concept how we can migrate the existing Posts to Nextjs. 

Open Questions:
- Do we need to add authentication
- Does the Announcement Collection replace the Posts or are these separate entities.
- Is the featured Announcements the replacement for posts being stickied? 

We can talk about this next week, but this works in principle. I have also made sure to test how we can implement #12173 in the future and this seems to be one way of doing it


next-frontend/src/collections/Post.ts
```diff
@@ -0,0 +1,58 @@
+import { CollectionConfig } from "payload";
+import { markdownConvertedField } from "@/collections/helpers";
+
+export const Posts: CollectionConfig = {
```
FinnIckler: Merge this with Announcements

next-frontend/src/app/(wca)/api/wca/import-posts/route.ts
```diff
@@ -0,0 +1,53 @@
...
+  await Promise.all(
+    body.map(async (post) =>
+      payload.create({
+        collection: "posts",
+        data: {
+          title: post.title,
+          "author name": post.author_name,
+          slug: post.slug,
+          sticky: post.sticky,
+          // @ts-expect-error payload still marks this as string even though it is richText
```
FinnIckler: look into this more

next-frontend/src/collections/Post.ts
```diff
@@ -0,0 +1,58 @@
...
+      name: "sticky",
+      type: "checkbox",
+    },
+    {
+      name: "unstickAt",
+      type: "date",
+      required: false,
+    },
+    {
+      name: "author name",
```
FinnIckler: rename to AuthorName

next-frontend/src/collections/Post.ts
```diff
@@ -0,0 +1,58 @@
...
+      type: "text",
+      required: true,
+    },
```
gregorbg: This should be a relation to the users table

lib/tasks/next_tasks.rake
```diff
@@ -0,0 +1,17 @@
...
+    task :import, :environment do
+      connection = Faraday.new(
+        url: EnvConfig.NEXTJS_URL,
+        headers: {
+          'Content-Type' => 'application/json',
+        },
+        &FaradayConfig
+      )
+      connection.post("/api/wca/import-posts") do |req|
+        req.body = Post.all.sample(1).to_json(teaser_only: false)
```
gregorbg: authors need to be included here

FinnIckler:
Okay, this is now done.
This still follows the current workflow:
- User adds a new post
- User adds that post to the home page

No automation there. I think that should be a different PR after a discussion with the stakeholders

FinnIckler:
Fixes https://github.com/thewca/worldcubeassociation.org/issues/14258

gregorbg:
> * Do we need to add authentication

Yes.

> * Does the Announcement Collection replace the Posts or are these separate entities.

It will eventually replace Posts.

> * Is the featured Announcements the replacement for posts being stickied?

Yes

next-frontend/src/app/(wca)/api/wca/import-posts/route.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  // Cache the users so we don't have to do a lookup for each post
+  const users: Record<string, User> = {};
+
+  for (let i = 0; i < body.length; i++) {
+    const post = body[i];
+    const publishedBy = users[post.author.email];
+    if (!publishedBy) {
+      users[post.author.email] = await findOrCreateUser(post.author, payload);
+    }
```
gregorbg, line 77: With the "Published By" tag gone completely, do we now even need this information at all anymore?

Sure, the Payload schema currently requires the information, but that schema is written by us. So we could remove it entirely if we wanted... Guess it's worth confirming that with WCT

next-frontend/src/collections/Announcements.ts
```diff
@@ -3,7 +3,18 @@ import { markdownConvertedField } from "@/collections/helpers";
...
+    defaultColumns: ["title", "publishAt", "publishedBy"],
+  },
+  defaultSort: "-publishAt",
   fields: [
+    {
+      name: "slug",
+      type: "text",
+      required: true,
+      unique: true,
+    },
```
gregorbg, line 17: Can we/should we auto-generate this?

next-frontend/src/collections/Announcements.ts
```diff
@@ -15,21 +26,51 @@ export const Announcements: CollectionConfig = {
...
     },
     {
+      name: "sticky",
+      type: "checkbox",
+    },
+    {
+      name: "unstickAt",
+      type: "date",
+      required: false,
+    },
```
gregorbg, line 51: We don't really have this timing behavior anymore. It used to be that any post was potentially visible on the homepage, so they had this `publishAt` as a workaround to dynamically say "hide if it's newer than the timestamp of the user trying to access the homepage".

But now, with the current Payload setup, WCT or whoever has full control over which posts are being displayed in the first place, so we should definitely reconsider the posting workflow

next-frontend/src/collections/Announcements.ts
```diff
@@ -15,21 +26,51 @@ export const Announcements: CollectionConfig = {
...
+          return user?.roles?.includes("wct") === true;
+        },
+        update: ({ req: { user } }) => {
+          return user?.roles?.includes("wct") === true;
+        },
+        read: () => {
+          return true;
+        },
+      },
+    },
```
gregorbg, line 74: Is this part of the old, existing `"posts"` schema in Rails?

gregorbg (commented):
There are no major concerns about your code in itself, but I'm starting to have larger doubts about the whole "posting workflow" in the old vs new homepage.

Stuff like `sticky` and `unstickAt` are all conventions which have "grown this way" because that's the fastest, easiest solution some developer came up with at a time when we did not have a full CMS.

Now we have a full CMS, so we should make sure that we design the posting system right from the get-go.

next-frontend/src/app/(wca)/api/wca/import-posts/route.ts
FinnIckler: This is a bit older, but https://github.com/thewca/worldcubeassociation.org/issues/12173 says they still would want that information.

FinnIckler:
Yes, this PR did intentionally not implement any workflow changes. I am not sure what is the best way, considering that it might take quite a while to get the ideal workflow from WCT. We could wait until we implement the new workflow and then try to do some kind of conversion from the old posts to the new announcements, but that might delay getting the homepage done?

next-frontend/src/collections/Announcements.ts
FinnIckler: I think admins can still create and update, but otherwise yes

next-frontend/src/collections/Announcements.ts
FinnIckler: yeah we don't really need sticky either right?

PR #12371, opened 2025-09-03
Fixes related to post results page of tickets

danieljames-dj:
Fixing basically two errors:
1. Ticket page going blank on posting - reason is that frontend expects `posted_user` and `results_posted_at`, which was not available. So sending the ticket as response now, so that frontend gets the latest data.
2. During error, there was no `I18n.t`, because of which we wouldn't have known if errors have come (we will know there are some errors, but won't know exact errors, and luckily these are rare errors so this wasn't impacted so far).

### DIFF
--- a/lib/competition_results_import.rb
+++ b/lib/competition_results_import.rb
@@ -72,13 +72,13 @@ module CompetitionResultsImport
   end
 
   def self.post_results_error(comp)
-    return t('competitions.messages.computing_auxiliary_data') if ComputeAuxiliaryData.in_progress?
+    I18n.t('competitions.messages.computing_auxiliary_data') if ComputeAuxiliaryData.in_progress?
 
-    return t('competitions.messages.no_results') unless comp.results.any?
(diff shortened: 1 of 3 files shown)


lib/competition_results_import.rb
```diff
@@ -72,13 +72,13 @@ def self.merge_inbox_results(competition)
   end
 
   def self.post_results_error(comp)
-    return t('competitions.messages.computing_auxiliary_data') if ComputeAuxiliaryData.in_progress?
+    I18n.t('competitions.messages.computing_auxiliary_data') if ComputeAuxiliaryData.in_progress?
```
gregorbg: missing a `return`

gregorbg (changes requested):

lib/competition_results_import.rb
danieljames-dj: Oops thanks for the catch, fixed.

gregorbg (approved):

PR #12372, opened 2025-09-04
Fix frontend qualifications

dunkOnIT:
Unqualified events weren't being disabled in the frontend due to a typo in accessing the `qualification_wcif` property


FinnIckler (approved):
Same typo as in the other file too 🙃

PR #12408, opened 2025-09-08
[MPI 1] Backend: Connect Manual Payment Integration

dunkOnIT:


spec/requests/competitions_spec.rb
```diff
@@ -3,9 +3,9 @@
 require "rails_helper"
 
 RSpec.describe "competitions" do
-  describe "PATCH #update_competition" do
-    let!(:competition) { create(:competition, :with_delegate, :future, :visible, :auto_accept, :with_valid_schedule) }
+  let!(:competition) { create(:competition, :with_delegate, :future, :visible, :auto_accept, :with_valid_schedule) }
```
dunkOnIT, line 6: This is just moving the factory definition up a level so that `GET #connect_payment_integration` can also access it

app/controllers/competitions_controller.rb
```diff
@@ -235,28 +239,46 @@ def payment_integration_setup
...
+    else
+      competition.competition_payment_integrations.build(connected_account: integration_reference)
+    end
 
     if competition.save
       flash[:success] = t('payments.payment_setup.account_connected', provider: t("payments.payment_providers.#{payment_integration}"))
     else
       flash[:danger] = t('payments.payment_setup.account_not_connected', provider: t("payments.payment_providers.#{payment_integration}"))
     end
-
```
gregorbg: Why did this poor empty line have to die?

app/models/connected_stripe_account.rb
```diff
@@ -145,7 +145,7 @@ def self.generate_onboarding_link(competition_id)
     client.auth_code.authorize_url(oauth_params)
   end
 
-  def self.connect_account(oauth_return_params)
+  def self.connect_integration(oauth_return_params)
```
gregorbg, line 148: Should an equivalent change also be made in the Paypal connector?

app/models/competition.rb
```diff
@@ -1003,6 +1003,16 @@ def using_payment_integrations?
     competition_payment_integrations.any? && paid_entry?
   end
 
+  def payment_integration_type
+    return nil unless using_payment_integrations?
+
+    CompetitionPaymentIntegration::AVAILABLE_INTEGRATIONS.key(competition_payment_integrations.first.connected_account_type)
+  end
+
+  def using_manual_payment?
```
dunkOnIT: change this to `manual_connected`

app/models/competition.rb
```diff
@@ -1003,6 +1003,16 @@ def using_payment_integrations?
     competition_payment_integrations.any? && paid_entry?
   end
 
+  def payment_integration_type
```
dunkOnIT: use `payment_integration_types` instead

app/models/competition.rb
```diff
@@ -1828,6 +1838,12 @@ def top_level_activities
     competition_venues.includes(venue_rooms: { schedule_activities: [:child_activities] }).map(&:top_level_activities).flatten
   end
 
+  def manual_payment_details
+    return nil unless using_manual_payment?
+
+    competition_payment_integrations.first.connected_account.account_details
```
dunkOnIT: Use `payment_account_for` instead

app/models/competition.rb
```diff
@@ -1862,8 +1878,8 @@ def to_competition_info
                force_comment_in_registration use_wca_registration external_registration_page guests_entry_fee_lowest_denomination guest_entry_status
                information events_per_registration_limit guests_enabled auto_accept_preference auto_accept_disable_threshold],
       methods: %w[url website short_name city venue_address venue_details latitude_degrees longitude_degrees country_iso2 event_ids
-                  main_event_id number_of_bookmarks using_payment_integrations? uses_qualification? uses_cutoff? competition_series_ids registration_full?
-                  part_of_competition_series? registration_full_and_accepted?],
+                  main_event_id number_of_bookmarks using_payment_integrations? payment_integration_type manual_payment_details uses_qualification?
+                  uses_cutoff? competition_series_ids registration_full? part_of_competition_series? registration_full_and_accepted?],
```
gregorbg: Undo because this is communicated as part of a separate "steps config" now

app/models/competition.rb
gregorbg, line 1: All new helper methods, especially those mentioning something like `competition_payment_integrations.first` should go, or be reduced to using existing helper methods along lines 2700psq

app/controllers/competitions_controller.rb
```diff
@@ -235,21 +239,40 @@ def payment_integration_setup
...
+    end
 
-    competition.competition_payment_integrations.new(connected_account: account_reference)
+    connector = CompetitionPaymentIntegration::AVAILABLE_INTEGRATIONS[payment_integration].safe_constantize
+    integration_reference = connector&.connect_integration(params)
+
+    raise ActionController::RoutingError.new("No integration reference submitted") if integration_reference.blank?
+
+    if payment_integration == :manual && competition.manual_payment_details.present?
+      existing_cpi = competition.competition_payment_integrations.first
```
gregorbg: These two lines should also be changed to reflect the newly updated helper methods in `competition.rb`. Especially the second line here is really grossly oversimplifying things, although I am probably to blame for this one

app/controllers/competitions_controller.rb
```diff
@@ -235,21 +240,40 @@ def payment_integration_setup
...
+    connector = CompetitionPaymentIntegration::AVAILABLE_INTEGRATIONS[payment_integration].safe_constantize
+    integration_reference = connector&.connect_integration(params)
+
+    raise ActionController::RoutingError.new("No integration reference submitted") if integration_reference.blank?
+
+    # Small hack: We allow de-facto updates by "re-connecting" manual payment in the UI.
+    #   This is done to allow edits to a Manual CPI, but coding a proper `PATCH` form
+    #   would break the mold of the usual OAuth flow with "proper" payment providers.
+    if payment_integration == :manual && competition.payment_account_for(:manual)&.account_details.present?
+      existing_cpi = competition.competition_payment_integrations.first
```
gregorbg: See my previous comment (that got lost in the recent commits I suppose). This should not use `first` anymore, probably you can put the `account_details` from the `if` check into a separate variable and recycle it here?!

dunkOnIT: Oop didn't realize we were doing this in the controller too, I only checked the model

config/locales/en.yml
```diff
@@ -2308,6 +2313,9 @@ en:
         paypal:
           display_name: "Name"
           primary_email: "Email"
+        manual:
+          payment_info: "Payment instructions (will be displayed on Competition Page)"
+          payment_reference: "What information should the user provide for you to identify their payment? (eg, payment reference number / account ID / etc)"
```
gregorbg: This should be adjusted to follow the new column names

app/webpacker/components/ManualPaymentSetup/index.js
```diff
@@ -0,0 +1,46 @@
...
+  const [paymentReferenceLabel, setPaymentReferenceLabel] = useInputState(accountDetails?.payment_reference_label);
+
+  return (
+    <>
+      <Header>
+        {I18n.t('payments.payment_setup.manual_payments_header')}
+      </Header>
+      <Form method="GET" action={connectPaymentIntegrationUrl(competitionId, 'manual')}>
+        <Form.Field
+          label={I18n.t('payments.payment_setup.account_details.manual.payment_info')}
```
gregorbg: Update the i18n key

app/webpacker/components/ManualPaymentSetup/index.js
```diff
@@ -0,0 +1,46 @@
...
+          control={MarkdownEditor}
+          value={paymentInstructions}
+          onChange={setPaymentInstructions}
+          imageUploadEnabled
+        />
+        {/* Transport the Markdown covertly through Base64 to maintain line breaks */}
+        <input value={utf8ToBase64(paymentInstructions)} name="payment_instructions" hidden />
+        <Form.Input
+          label={I18n.t('payments.payment_setup.account_details.manual.payment_reference')}
```
gregorbg: Update the i18n key

PR #12429, opened 2025-09-11
[MPI 2] Add Manual Payment Records

dunkOnIT:
This PR:
- Adds the ManualPaymentRecord model
- Implements a payment flow for manual payments, through a combination of:
  - changes to `payment_ticket` and `payment_completion` 
  - implementation of `prepare_intent` on the ManualPaymentRecord model
- Supports the ability for the user to change their payment reference by re-submitting `payment_completion`

## Key Questions

- When do we create a RegistrationPayment? 
  - currently I'm creating it in the `payment_ticket`, but it could also be created upon the submission of `payment_completion` (which is the implementation I had initially) 
  - Creating it in the `ticket` makes it possible for organizers to mark users as paid who haven't yet submitted their payment reference (although we could also handle this in the `capture_payment` logic, which is introduced in the next PR)
  - `PaymentIntent#update_status_and_charges`, as well as `Registration#record_payment` get a little messier if we create the RegPayment upon payment_completion, as we need to add a more conditional clauses for if a manual payment already exists - but there are probably ways to get around this

## MPI 1.5 Considerations

This PR introduces an inconsistency in the relationship between a PaymentIntent ("WCA" implied whenever I write it like this) and its associated PaymentRecord - in Stripe or PayPal, a PaymentIntent would be associated with the PaymentRecord which represents the "parent" object for the overall payment lifecycle - a Stripe Payment Intent, or a PayPal Order. 

However, in this PR, we are associating the PaymentIntent with the equivalent of a "charge" record[1], and because our code relies on the assumption that there is an underlying provider record for the PaymentIntent to be associated with, we are creating the PaymentRecord at the same time as creating the PaymentIntent. This is awkward - we should either: 
1. create the PaymentRecord at a later stage; only when the "charge" takes place
2. create an intermediary PaymentRecord which corresponds to a PaymentIntent

I'm partial to (1) because (2):
  - creates an unnecessary record in our database
  - makes us rely on the assumption that there will _always_ be some sort of "parent"-equivalent payment record - or force us to improvise one if there isn't. _(more on this shortly)_

The change (1) requires is us allowing the creation of PaymentIntents without associated payment records. This can be done by: 
- making the `belongs_to: :payment_record` association `optional: true`, and 
- changing the `wca_status_consistency` validation to allow there to be no associated payment record only if the `wca_status` is `created`

I have pushed a WIP version of this to the PR.

I had previously thought that this would also require us to implement some generalized charge-surfacing logic in the PaymentIntent - but inspection of our current payment flow reveals that this is handled in the `find_payment_from_request`-style logic in the provider account models. If needed we can refactor/expand on these - and perhaps moving forward with this implementation will reveal that need. 

We may also need to refactor around `registrations_controller:448` - as this relies explicitly on the assumption that a PI will have an underlying PaymentRecord representing that PI

[1] We could consider some internal terminology for this - "payment event", or "payment attempt"

### What do PaymentRecords represent? 

I see our payment process as follows: 
1. Provider gives us an update (payment completion/webhook)
2. We internalise that provider logic (determine & persist PaymentRecord updates)
3. We translate provider-specific logic into general WCA payments logic (PaymentIntent updates) 
4. The generalised WCA payments logic triggers WCA business logic (create RegistrationPayments)

Based on this understanding, I think the value of PaymentRecords lies in their being _as flexible as possible_ - we should try to mimic provider structures/logic, not try to force conformity to our payments model. Conformity to our payments model is "the provider does Some Stuff (:TM), and we can abstract that into a certain set of properties on the PaymentIntent that our business logic can act from". 

### Refactor opportunities 

- `remote` (in `retrieve_remote`) -> `provider` - eg, `retrieve_provider_record` or similar
- `find_payment_from_request` (in the ConnectedProviderAccount models - eg, ConnectedStripeAccount) -> `find_payment_intent` (as this seems to be what it is actually doing - 

## MPI PR Overview
I'm busy reorganizing PR's and it's a lot to keep track of - here's the breakdown: 

### Canonical PR's: 
- MPI 1: #12408 Added ManualPaymentIntegration
- MPI 2: This PR 
- [MPI 3](https://github.com/dunkOnIT/worldcubeassociation.org/pull/5): IN PROGRESS - Adds the user-facing frontend for manual payments 
- MPI 4: TODO - RegAdmin table changes - displaying payment reference info in the table, and adding the ability to capture manual payments (also adds a route for bulk capturing manual registration payments)
- MPI 5: TODO - RegEdit changes - correctly display manual payments in Payment History, and allow the organizer to mark them as paid/unpaid as necessary (also adds a route for uncapturing registration payments)

### Historical / Reference PR's: 
Any of these still open will be closed as/when they are no longer needed.  
- #12159 - has the full commit history for backend changes, as part of that history was squashed when creating MPI 2 (useful for the commit history)
- #12278 - contains all frontend changes, as well as the full commit history from ALL MPI changes until that point (I branched off parent branches, instead of rebasing onto them). In the process of being split up into MPI 3/4/5, then I'll close it

### Future PR's: 
- MPI 100: TODO - Enable Live Auto Accept for manual payments
- MPI 101: TODO - Expose payment capture endpoints to 3rd parties via API

### DIFF
--- a/app/models/manual_payment_integration.rb
+++ b/app/models/manual_payment_integration.rb
@@ -53,7 +53,7 @@ class ManualPaymentIntegration < ApplicationRecord
   end
 
   def retrieve_payments(payment_intent)
-    yield payment_intent.payment_record
+    yield payment_intent.payment_record if block_given?
   end
 
   def self.generate_onboarding_link(competition_id)
--- a/app/models/payment_intent.rb
+++ b/app/models/payment_intent.rb
@@ -84,7 +84,9 @@ class PaymentIntent < ApplicationRecord
           wca_status: updated_wca_status,
         )
       when PaymentIntent.wca_statuses[:requires_capture]
-        yield api_intent if block_given?
+        payment_account.retrieve_payments(self) do |payment|
+          yield payment if block_given?
+        end
       end
 
       self.update_common_attributes!(api_intent, updated_wca_status)


app/controllers/competitions_controller.rb
```diff
@@ -279,7 +279,6 @@ def connect_payment_integration
     else
       flash[:danger] = t('payments.payment_setup.account_not_connected', provider: t("payments.payment_providers.#{payment_integration}"))
     end
-
```
gregorbg, line 282: RIP blank line

app/models/manual_payment_integration.rb
```diff
@@ -7,6 +7,37 @@ def self.manual_payments_disabled?
     Rails.env.production? && EnvConfig.WCA_LIVE_SITE?
   end
 
+  def prepare_intent(registration, amount_iso, currency_iso, paying_user)
+    existing_intent = registration.payment_intents.first
```
gregorbg: Why only the first?

app/models/manual_payment_integration.rb
```diff
@@ -7,6 +7,37 @@ def self.manual_payments_disabled?
...
+      return existing_intent
+    end
+
+    self.create_intent(registration, amount_iso, currency_iso, paying_user)
+  end
+
+  private def create_intent(registration, amount_iso, currency_iso, paying_user)
+    manual_record = ManualPaymentRecord.create(amount_iso_denomination: amount_iso, currency_code: currency_iso, manual_status: :created)
+    # We create a registration payment with the payment ticket instead of upon payment completion
+    # so that organizrs can mark a registrant as paid even if the registrant hasn't submitted a payment reference yet
```
gregorbg: Typo
```suggestion
    # so that organizers can mark a registrant as paid even if the registrant hasn't submitted a payment reference yet
```

app/models/competition.rb
```diff
@@ -1862,8 +1862,8 @@ def to_competition_info
                force_comment_in_registration use_wca_registration external_registration_page guests_entry_fee_lowest_denomination guest_entry_status
                information events_per_registration_limit guests_enabled auto_accept_preference auto_accept_disable_threshold],
       methods: %w[url website short_name city venue_address venue_details latitude_degrees longitude_degrees country_iso2 event_ids
-                  main_event_id number_of_bookmarks using_payment_integrations? uses_qualification? uses_cutoff? competition_series_ids registration_full?
-                  part_of_competition_series? registration_full_and_accepted?],
+                  main_event_id number_of_bookmarks using_payment_integrations? payment_integration_type manual_payment_details uses_qualification?
+                  uses_cutoff? competition_series_ids registration_full? part_of_competition_series? registration_full_and_accepted?],
```
gregorbg: Still no changes to the serialization here please

.github/workflows/ruby-test.yml
dunkOnIT, line 1: Will remove soon - but wanted to see what you'd think of a change like this? It allows us to have CI run in non-main branches, which was relevant as the PRs on my fork were diff-ing against their preceeding branches

app/models/manual_payment_record.rb
```diff
@@ -0,0 +1,43 @@
...
+    WCA_TO_MANUAL_PAYMENT_STATUS_MAP.find { |_key, values| values.include?(self.manual_status) }.first
+  end
+
+  def retrieve_remote
+    self
+  end
+
+  def retrieve_payments
+    self
+  end
```
gregorbg: What is this method and why is it necessary? I think the CPI connectors have a `def retrieve_payments(payment_intent)` instead.

.github/workflows/ruby-test.yml
gregorbg: Hm. I am open to it but not as part of the current manual CPI PR series

spec/requests/api_registrations_spec.rb
dunkOnIT, line 1: I moved all the pre-existing tests within a `stripe_connected` context blocked, which has seriously confused Github in the first part of this diff

app/models/payment_intent.rb
```diff
@@ -102,12 +102,16 @@ def update_common_attributes!(api_intent, updated_wca_status)
 
     def wca_status_consistency
```
gregorbg, line 107: I feel like we could refactor this. Alias the `stripe_status` and `paypal_status` etc attributes to something like `remote_status` (Rails has both `alias_method` as well as `alias_attribute`) and agree on a consistent naming scheme for the constant map. Then we can write the same exact block of code for any `payment_record_type`.

db/migrate/20250909120729_add_manual_payment_records_table.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class AddManualPaymentRecordsTable < ActiveRecord::Migration[7.2]
+  def change
+    create_table :manual_payment_records do |t|
+      t.string :payment_reference
```
gregorbg: Why is this nullable?

dunkOnIT: My thinking at the time was that organizers might not always _need_ a user to submit a payment reference, but I agree that having it not be nullable is better

app/models/payment_intent.rb
dunkOnIT: Done - I wonder if we should go a step further and rename the columns themselves, instead of just aliasing them? Or is there a good reason to keep the column names different?

app/controllers/api/v1/registrations_controller.rb
```diff
@@ -263,11 +263,21 @@ def validate_payment_ticket_request
   end
 
   def payment_ticket
+    # TODO: GB Remove this default after deployment, once we're sure all existing competitions' requests
```
dunkOnIT, line 266: Should we remove this now?

app/controllers/api/v1/registrations_controller.rb
```diff
@@ -263,11 +263,21 @@ def validate_payment_ticket_request
...
+    #   have been fully processed (~30 minutes after the AWS deployment cycle is complete should be more than sufficient)
+    payment_integration_type = params[:payment_integration_type] || :stripe
+
+    return render json: { client_secret: 'manual' } if payment_integration_type == 'manual'
```
dunkOnIT, line 270: I wanted to refer to some kind of constant/enum value for this but couldn't find one - we could change the submission value of `payment_integration_type`, but I didn't want to touch the frontend/expand the scope of this PR further

app/controllers/registrations_controller.rb
```diff
@@ -463,7 +455,8 @@ def payment_completion
     # Some API gateways like Stripe provide the client_secret as a kind of "checksum" (or fraud protection)
     #   back to us upon redirect. Other providers (like PayPal…) unfortunately don't.
     #   So we only compare this secret value with our stored intent record if it's actually provided to us
-    if secret_check.present? && stored_intent.client_secret != secret_check
+    validation_token = params[:payment_intent_client_secret]
+    if validation_token.present? && stored_intent.client_secret != validation_token
```
dunkOnIT: I think `validation_token` is a bit more general and easier to grok than `client_secret` in this context - we could rename the column, or have a `validate_token` method in PaymentIntent - but I didn't want to be too sweeping in the changes, I think the rename already brings a lot of clarity

app/controllers/registrations_controller.rb
```diff
@@ -463,7 +455,8 @@ def payment_completion
     # Some API gateways like Stripe provide the client_secret as a kind of "checksum" (or fraud protection)
     #   back to us upon redirect. Other providers (like PayPal…) unfortunately don't.
     #   So we only compare this secret value with our stored intent record if it's actually provided to us
-    if secret_check.present? && stored_intent.client_secret != secret_check
+    validation_token = params[:payment_intent_client_secret]
```
dunkOnIT: This was being returned by the `find_payment` method for some reason - it doesn't need to be, the method was just reading the value from the `params`

app/models/manual_payment_integration.rb
```diff
@@ -7,18 +7,57 @@ def self.manual_payments_disabled?
     Rails.env.production? && EnvConfig.WCA_LIVE_SITE?
   end
 
+  def prepare_intent(registration, amount_iso, currency_iso, paying_user, payment_reference: nil)
+    existing_intent = registration.manual_payment_intent
+    if existing_intent.present?
+      reference_to_write = payment_reference.presence || existing_intent.payment_record.payment_reference
```
dunkOnIT, line 13: This line is used to update the payment_reference - it's the only time where we get access to the `params` which contain the reference in the `payment_completion` flow

app/models/manual_payment_integration.rb
```diff
@@ -7,18 +7,57 @@ def self.manual_payments_disabled?
     Rails.env.production? && EnvConfig.WCA_LIVE_SITE?
   end
 
+  def prepare_intent(registration, amount_iso, currency_iso, paying_user, payment_reference: nil)
```
dunkOnIT: `payment_reference` is an optional parameter because `prepare_intent` is shared across all CompetitionPaymentIntegrations

app/models/registration.rb
```diff
@@ -214,12 +218,17 @@ def record_payment(
     receipt,
     user_id
   )
+    # Don't create multiple registration payments for a manual payment
```
dunkOnIT, line 222: We could also have logic in `update_status_and_charges` determining when to `yield` - but that logic risks getting clunky (especially when we add in status regressions from uncapturing a payment in MPI 5). This seems like the cleanest way of handling it, although I'm not thrilled. 

The other option would be a `uniqueness` validation, but then to avoid the error being thrown we would need to check for the presence of other manual RegistrationPayments anywhere that `record_payment` is called - so it still seems better to do this check in `record_payment`

app/models/manual_payment_integration.rb
dunkOnIT: But it is set to `null: false` in the database, so I'm wary of any circumstance where this would actually be called without specifying a `payment_reference`

app/controllers/api/v1/registrations_controller.rb
gregorbg: By all means!

gregorbg: That is, you only need to remove the `|| stripe` fallback. You can probably change it to `params.require` too while you're at it

app/controllers/api/v1/registrations_controller.rb
gregorbg: What exactly did you want to turn into a constant here, and what kept you from "just doing it"?
Sorry I can't quite reconcile the explanations of your comment with this specific line of code.

app/controllers/registrations_controller.rb
gregorbg: It does need to be handled the way that it was handled prior to your changes.
The name of the param that it's being transported in is determined *by Stripe*. Only the Stripe connector/CPI would know that it needs to look up this param. PayPal may have other params or no params at all, and YourFunnyPaymentGateway123 may also compute the param differently.

TL;DR Reading from this specific param key is Stripe-specific and does not generalize to other payment methods

app/controllers/registrations_controller.rb
gregorbg: Then please put it on a follow-up list or some kind of TODO, but keep to the existing nomenclature within the scope of this PR.

app/models/payment_intent.rb
gregorbg: Within the `PaymentRecord` tables, we want to stay as close to the gateways as possible. Enforcing a common naming scheme would mean controlling a property that should instead be controlled by the gateways.

app/controllers/registrations_controller.rb
dunkOnIT: Ah of course I'm with you

app/controllers/api/v1/registrations_controller.rb
dunkOnIT: EDIT: Ignore this, see my latest comment in this thread.

=====

I wanted to turn `'manual'` in `payment_integration_type == 'manual'` into a constant, but I couldn't find a source that already provided that - ideally, it would come from the ones already defined in `models/competition_payment_integration.rb`: 

```ruby
  AVAILABLE_INTEGRATIONS = {
    paypal: 'ConnectedPaypalAccount',
    stripe: 'ConnectedStripeAccount',
    manual: 'ManualPaymentIntegration',
  }.freeze
```

I think the best thing would be to expose the values of this hash to the frontend, and have the frontend submit them (eg, { payment_integration_type: "ManualPaymentIntegration" }) - but I'm concerned with the scope creep that represents for this PR, so hardcoding the string seems good enough for now

app/controllers/registrations_controller.rb
dunkOnIT: Upon further consideration, I think the fact that this variable only has to do with Stripe seems a better argument for not expecting it to be returned from `find_payment_intent...` - as it forces us to return this value from that method for _all_ integrations - even when it is irrelevant to them. 

It still seems better to me to handle this in the method body, probably with an explicitly check for the payment 

Additionally - I think we should explicitly check the payment provider. Otherwise, it seems to me that the secret check could be bypassed by simply not including it

app/controllers/registrations_controller.rb
dunkOnIT: Done

app/controllers/api/v1/registrations_controller.rb
dunkOnIT: Done

app/controllers/api/v1/registrations_controller.rb
dunkOnIT: Oops, turns out that we're only adding this code in this PR - I thought it was already merged in. Reverting changes, I'll make my suggested changes in MPI 3, which is where we're touching exactly this frontend code.

app/controllers/api/v1/registrations_controller.rb
dunkOnIT: Not done - see below comment chain

app/controllers/registrations_controller.rb
```diff
@@ -463,9 +455,13 @@ def payment_completion
...
-    if secret_check.present? && stored_intent.client_secret != secret_check
-      flash[:error] = t("registrations.payment_form.errors.generic.secret_invalid", provider: t("payments.payment_providers.#{payment_integration}"))
-      return redirect_to competition_register_path(competition)
+    # TODO: Refactor `secret_check` to `validate_token`, and consider making this a PaymentIntent method
+    if payment_integration == :stripe
+      secret_check = params[:payment_intent_client_secret]
+      unless secret_check.present? && stored_intent.client_secret == secret_check
+        flash[:error] = t("registrations.payment_form.errors.generic.secret_invalid", provider: t("payments.payment_providers.#{payment_integration}"))
+        return redirect_to competition_register_path(competition)
+      end
```
gregorbg, line 464: I don't see the benefit of this approach. You're introducing a fancy-snowflake special case for Stripe, so you have to check "are we even running on Stripe?". The answer to this question is already "baked into" the `payment_account` instance, so I wonder why you're not using it (the way it was used before).

app/controllers/api/v1/registrations_controller.rb
gregorbg: I don't follow... Please link me to the exact comment that justifies why you cannot remove this default value computation here.

app/models/connected_paypal_account.rb
```diff
@@ -54,16 +54,13 @@ def find_payment(record_id)
...
     # Provided by ourselves when we pass the order body from the PayPal frontend SDK
     #   to our own backend redirect route
     order_id = params[:orderID]
 
     # We expect that the record here is a top-level `order` object in PayPal's API model
-    stored_order = PaypalRecord.paypal_order.find_by(paypal_id: order_id)
-
-    # PayPal doesn't have nice checksums with orders unfortunately :/
-    [stored_order, nil]
+    PaypalRecord.paypal_order.find_by(paypal_id: order_id).payment_intent
```
gregorbg, line 63: This change depends on the other discussion above. Do we want to discard checksums just because one provider (in this case PayPal) doesn't happen to provide them?

app/models/connected_stripe_account.rb
```diff
@@ -100,15 +100,14 @@ def find_payment(record_id)
...
     intent_id = params[:payment_intent]
-    intent_secret = params[:payment_intent_client_secret]
 
     # We expect that the record here is a top-level PaymentIntent in Stripe's API model
-    stored_record = StripeRecord.payment_intent.find_by(stripe_id: intent_id)
-
-    [stored_record, intent_secret]
+    # The first invocation of `payment_intent` is the payment_intent scope on StripeRecord
+    # The second is getting the associated PaymentIntent object from the found StripeRecord
+    StripeRecord.payment_intent.find_by(stripe_id: intent_id).payment_intent
```
gregorbg, line 110: (see above about being cautious with this change)

app/models/manual_payment_integration.rb
```diff
@@ -7,18 +7,49 @@ def self.manual_payments_disabled?
...
+    # Because there is no outgoing request to a payment provider, we did not create a PaymentIntent during the payment_ticket step
+    # Thus, we need to create the PaymentIntent and associated PaymentRecord now instead
+
+    registration = Registration.find(params[:registration_id])
+    entry_fee = registration.competition.base_entry_fee_lowest_denomination
+    currency_code = registration.competition.currency_code
+    user = registration.user
+    reference = params[:payment_reference]
+
+    prepare_intent(registration, entry_fee, currency_code, user, reference)
```
gregorbg:
```suggestion
    self.prepare_intent(registration, entry_fee, currency_code, user, reference)
```

app/models/payment_intent.rb
```diff
@@ -75,14 +76,15 @@ def update_status_and_charges(payment_account, api_intent, action_source, source
...
-        # Reset by the gateway
         self.update!(
           confirmed_at: nil,
           confirmation_source: nil,
           canceled_at: nil,
           cancellation_source: nil,
           wca_status: updated_wca_status,
         )
+      when PaymentIntent.wca_statuses[:requires_capture]
+        yield api_intent if block_given?
```
gregorbg: You're not yielding the correct information here. In the "successful" case above, we're querying the individual _charges_ and yielding them to the outside block. Now you're yielding the overarching payment intent, which is substantially different from yielding charges.

The outer block does not know which of the switch case branches of this method body yielded to it, so this will not behave correctly in most cases.

app/models/registration.rb
```diff
@@ -214,12 +219,17 @@ def record_payment(
...
+    return if receipt.instance_of?(ManualPaymentRecord) &&
+              registration_payments.where(receipt_type: CompetitionPaymentIntegration::INTEGRATION_RECORD_TYPES[:manual]).any?
```
gregorbg, line 224: Why are you returning even before generating a history entry? This feels like you're short-circuiting too much.

I see that your above comment explains why you're short-circuiting _here_ and not somewhere else in the code, but my question is why you want/need this short-circuiting _at all_.

gregorbg (changes requested):

app/controllers/api/v1/registrations_controller.rb
dunkOnIT: > turns out that we're only adding this code in this PR - I thought it was already merged in. 

https://github.com/thewca/worldcubeassociation.org/pull/12429/files#r2382362561

app/controllers/registrations_controller.rb
dunkOnIT: In my view, we were overloading the `find_payment_from_request` method by having it _also_ assign the `secret_check` variable[1]. This is compounded in refactoring to `find_payment_intent_from_request`. 

If a stripe-specific logic chain isn't desirable, then we could implement a `payment_account.secret_check` method instead - but given that this only applies to one of the payment integrations we support at the moment, it didn't seem necessary to generalise it into a method that all CompetitionPaymentIntegrations must use. 

If we add a bunch more payment integrations, and end up having 4/6 that use some form of secret check/token validation, it'll make a lot of sense to have this method in all CPI's - but if we end up at 1/6 (ie, only Stripe has it), then we're returning `null` from a lot of different methods when we could instead be handling this one, Stripe-specific case inline.

[1] This is the only place in our code I've seen us return this tuple-like value, and nothing in the method naming suggests that it would return anything apart from the `payment` object

app/models/registration.rb
dunkOnIT: Two reasons: 
1. For a manual payment, we create an (incomplete) RegistrationPayment upon submission of the payment reference - ie, in a `requires_capture` status. When the payment is eventually captured, `update_status_and_charges` will yield again in the `succeeded` block - and we need to make sure that it doesn't create a _second_ RegistrationPayment.
2. In general, we don't want multiple RegistrationPayments for a manual payment[1] - so a check like this always makes sense before creating a RegistrationPayment. (As mentioned in the comment above, it could be additionally supported with a `uniqueness` validation)

[1] at least at the moment, the correct course of action is _always_ to update the existing RegistrationPayment for a manual payment, not create a new one

PR #12431, opened 2025-09-11
Change `auto_accept_preference` database dump value

dunkOnIT:
There is a validation which raises an error if you try and save a competition which has a non-0 (disabled) `auto_accept_preference` and no payment integration connected.

Currently, when dumping the db, auto_accept_preference is set to copy - which means that we end up with comps in the db dump which have auto accept enabled - which means that any time you try and save a change to such a comp in the db dump, the validation will fail.

This PR resolves the issue by using the db default (0 / disabled) for dumping `auto_accept_preference`

### DIFF
--- a/db/migrate/20250911125658_remove_auto_accept_registrations_column.rb
+++ b/db/migrate/20250911125658_remove_auto_accept_registrations_column.rb
@@ -2,6 +2,6 @@
 
 class RemoveAutoAcceptRegistrationsColumn < ActiveRecord::Migration[7.2]
   def change
-    remove_column :competitions, :auto_accept_registrations, :boolean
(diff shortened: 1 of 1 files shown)


db/migrate/20250911125658_remove_auto_accept_registrations_column.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class RemoveAutoAcceptRegistrationsColumn < ActiveRecord::Migration[7.2]
+  def change
+    remove_column :competitions, :auto_accept_registrations, :boolean
```
gregorbg: Super duper nitpick, but you should probably also specify `default: false, null: false`

PR #12455, opened 2025-09-16
Add chakra based next datepicker

FinnIckler:
Simple placeholder datepicker using only Chakra components. I did some research looking through the custom implementations in the https://github.com/chakra-ui/chakra-ui/issues/580 but it's all outdated using chakra v2. 

I also tried AI generated (bolt/chatgpt) but it also generated v2 code so I took a little bit of everything to create this.

This should be enough for our current usecases, if we want more we can think about building something on top of https://github.com/deseretdigital/dayzed which I have seen used.

Nothing selected:
image

Range selected:
image

Single Selected:
image

### DIFF
--- a/next-frontend/src/components/DatePicker.tsx
+++ b/next-frontend/src/components/DatePicker.tsx
@@ -219,19 +219,14 @@ export const DatePicker: React.FC<DatePickerProps> = ({
       positioning={{ placement: "bottom-start" }}
     >
       <Popover.Trigger asChild>
-        <Button
-          disabled={disabled}
-          justifyContent="flex-start"
-          w="full"
-          maxW="300px"
-        >
+        <Button disabled={disabled} justifyContent="flex-start">
           <LuCalendar size={16} />
           <Text truncate>{formatDisplayValue()}</Text>
         </Button>
       </Popover.Trigger>
       <Portal>
         <Popover.Positioner>
-          <Popover.Content width="320px" p={0}>
+          <Popover.Content width="sm" p={0}>
             <Popover.Arrow />
             <Popover.Body p={4}>
               <VStack gap={4}>


next-frontend/src/components/DatePicker.tsx
```diff
@@ -0,0 +1,336 @@
...
+      positioning={{ placement: "bottom-start" }}
+    >
+      <PopoverTrigger asChild>
+        <Button
+          disabled={disabled}
+          justifyContent="flex-start"
+          w="full"
+          maxW="300px"
+        >
+          <AiFillCalendar size={16} />
```
gregorbg: We don't use `Ai` icons in any other places of the code as far as I'm concerned. Are there any viable alternatives in either our own icon pack, or Lucide?

next-frontend/src/components/DatePicker.tsx
```diff
@@ -0,0 +1,336 @@
...
+          day === "Sun"
+            ? 7
+            : ["Mon", "Tue", "Wed", "Thu", "Fri", "Sat"].indexOf(day) + 1,
+        ),
+        "EEE",
+        { locale: currentLocale },
+      ),
+  );
+
+  return (
```
gregorbg, line 215: You're inconsistent between dot-notation references (for example, `Popover.Positioner`) and fully-qualified imports (for example, `PopoverTrigger` instead of `Popover.Trigger`) throughout the JSX skeleton. Please use dot-notation wherever possible.

next-frontend/src/app/(wca)/competitions/competitionClient.tsx
```diff
@@ -46,6 +49,14 @@ interface CompetitionsListProps {
...
   competitions,
 }: CompetitionsListProps) {
+  const [date, setDate] = useState<
+    | Date
+    | {
+        start: Date | null;
+        end: Date | null;
+      }
+    | null
+  >(null);
```
gregorbg: The type of this state definitely needs a reusable reference, perhaps even exported from `DatePicker` directly

next-frontend/src/components/DatePicker.tsx
```diff
@@ -0,0 +1,336 @@
...
+  endOfWeek,
+} from "date-fns";
+import * as dateLocales from "date-fns/locale";
+import { AiFillCalendar } from "react-icons/ai";
+import { BiChevronLeft, BiChevronRight } from "react-icons/bi";
+
+export interface DateRange {
+  start: Date | null;
+  end: Date | null;
+}
```
gregorbg, line 37: Are these JS-native `Date` objects? We tend to prefer Luxon `DateTime` because we heavily rely on FullCalendar, and that one unfortunately doesn't have native JS-`Date` support.

next-frontend/src/components/DatePicker.tsx
```diff
@@ -0,0 +1,336 @@
...
+  value,
+  onChange,
+  mode = "single",
+  locale,
+  placeholder,
+  minDate,
+  maxDate,
+  disabled = false,
+}) => {
+  const [currentMonth, setCurrentMonth] = useState(new Date());
```
gregorbg, line 60: Why is the month represented as an entire `Date`?

next-frontend/src/components/DatePicker.tsx
```diff
@@ -0,0 +1,336 @@
...
+          ? "solid"
+          : isInRange || isTempInRange
+            ? "outline"
+            : "ghost",
+      colorPalette:
+        isSelected || isRangeStart || isRangeEnd || isTempStart || isTempEnd
+          ? "blue"
+          : isInRange || isTempInRange
+            ? "blue"
+            : "gray",
```
gregorbg: Why do you modify the color palette so heavily? This seems like it should rather be a `disabled` prop or similar. Since this is a method for returning button props, I can definitely say that buttons do have a `disabled` prop.

next-frontend/src/components/DatePicker.tsx
```diff
@@ -0,0 +1,336 @@
...
+      (minDate && date < minDate) || (maxDate && date > maxDate);
+
+    return {
+      size: "sm",
+      variant:
+        isSelected || isRangeStart || isRangeEnd || isTempStart || isTempEnd
+          ? "solid"
+          : isInRange || isTempInRange
+            ? "outline"
+            : "ghost",
```
gregorbg: Same as for the `colorPalette` modification: Seems like you're just simulating a "non-selectable" button state... You're even using the exact same booleans without properly re-using them

next-frontend/src/components/DatePicker.tsx
```diff
@@ -0,0 +1,336 @@
...
+            : "gray",
+      bg: !isCurrentMonth
+        ? "transparent"
+        : isToday(date) &&
+            !isSelected &&
+            !isRangeStart &&
+            !isRangeEnd &&
+            !isTempStart &&
+            !isTempEnd
+          ? "blue.50"
```
gregorbg: Should probably be `colorPalette.50`

next-frontend/src/components/DatePicker.tsx
```diff
@@ -0,0 +1,336 @@
...
+          : undefined,
+      color: !isCurrentMonth ? "gray.400" : undefined,
+      opacity: isDisabledDate ? 0.4 : 1,
+      cursor: isDisabledDate ? "not-allowed" : "pointer",
+      _hover: isDisabledDate ? {} : { bg: "blue.50" },
```
gregorbg: Again, this all hints at the fact that you should just use the `disabled` property directly and let Chakra internally take care of all of this.

next-frontend/src/components/DatePicker.tsx
```diff
@@ -0,0 +1,336 @@
...
+          2024,
+          0,
+          day === "Sun"
+            ? 7
+            : ["Mon", "Tue", "Wed", "Thu", "Fri", "Sat"].indexOf(day) + 1,
+        ),
+        "EEE",
+        { locale: currentLocale },
+      ),
+  );
```
gregorbg: This definitely needs a comment, I have no idea what it does. Nor do I know why the `day === "Sun"` check is necessary to return 7. I assume it has to do something with indexing the days of the week, but that may hugely differ by culture (some consider Sunday the "start of the week", others do Monday, some do Thursday)

next-frontend/src/components/DatePicker.tsx
```diff
@@ -0,0 +1,336 @@
...
+      open={isOpen}
+      onOpenChange={({ open }) => setIsOpen(open)}
+      positioning={{ placement: "bottom-start" }}
+    >
+      <PopoverTrigger asChild>
+        <Button
+          disabled={disabled}
+          justifyContent="flex-start"
+          w="full"
+          maxW="300px"
```
gregorbg: A bit arbitrary, can you use semantic tokens like `"small"` or `"xl"` or whatever here?

next-frontend/src/components/DatePicker.tsx
```diff
@@ -0,0 +1,336 @@
...
+        >
+          <AiFillCalendar size={16} />
+          <Text truncate>{formatDisplayValue()}</Text>
+        </Button>
+      </PopoverTrigger>
+      <Portal>
+        <Popover.Positioner>
+          <PopoverContent width="320px" p={0}>
```
gregorbg: Again, arbitrary number where a semantic token should probably be possible

next-frontend/src/components/DatePicker.tsx
```diff
@@ -0,0 +1,336 @@
...
+              <VStack gap={4}>
+                {/* Month Navigation */}
+                <HStack w="full" justify="space-between" align="center">
+                  <IconButton
+                    aria-label="Previous month"
+                    size="sm"
+                    variant="ghost"
+                    onClick={() => setCurrentMonth(subMonths(currentMonth, 1))}
+                  >
+                    <BiChevronLeft size={16} />
```
gregorbg: We don't rely on `Bi` anywhere else in the code. Use our own pack or Lucide, please

next-frontend/src/components/DatePicker.tsx
```diff
@@ -0,0 +1,336 @@
...
+                      locale: currentLocale,
+                    })}
+                  </Text>
+                  <IconButton
+                    aria-label="Next month"
+                    size="sm"
+                    variant="ghost"
+                    onClick={() => setCurrentMonth(addMonths(currentMonth, 1))}
+                  >
+                    <BiChevronRight size={16} />
```
gregorbg: Again, `Bi` icon

next-frontend/src/components/DatePicker.tsx
```diff
@@ -0,0 +1,336 @@
...
+                        textAlign="center"
+                        py={1}
+                      >
+                        {day}
+                      </Text>
+                    ))}
+                  </Grid>
+
+                  {/* Calendar days */}
+                  <Grid templateColumns="repeat(7, 1fr)" gap={1}>
```
gregorbg: In case of such a "straight-forward" template, `SimpleGrid` should be sufficient

next-frontend/src/components/DatePicker.tsx
FinnIckler: it's because the `date-fns` like `subMonths`, `addMonths` and `startOfMonth` all operate on dates

next-frontend/src/components/DatePicker.tsx
FinnIckler: it was because January 1 - 7 is a monday - sunday. I refactored it now with custom start of the week

next-frontend/src/components/DatePicker.tsx
FinnIckler: this is all because I am using date-fns. I can see how complicated it is to switch to luxon for all of it

PR #12458, opened 2025-09-17
Add Chakra based RegionSelector

FinnIckler:
I used this in 3 PRs so far, so I decided to just open a PR for it and also fix the issues with missing region headers.

image
image

The width can scale of course

### DIFF
--- a/next-frontend/src/app/(wca)/competitions/competitionClient.tsx
+++ b/next-frontend/src/app/(wca)/competitions/competitionClient.tsx
@@ -1,8 +1,6 @@
 "use client";
 
 import {
-  useFilter,
-  useListCollection,
   Container,
   VStack,
   Heading,
@@ -33,9 +31,10 @@ import CompRegoCloseDateIcon from "@/components/icons/CompRegoCloseDateIcon";
 import EventIcon from "@/components/EventIcon";
 import Flag from "react-world-flags";
 
-import countries from "@/lib/wca/data/countries";
 import { WCA_EVENT_IDS } from "@/lib/wca/data/events";
 import type { components } from "@/types/openapi";
+import RegionSelector from "@/components/RegionSelector";
+import { useT } from "@/lib/i18n/useI18n";
 
 type CompetitionIndex = components["schemas"]["CompetitionIndex"];
 
@@ -46,16 +45,6 @@ interface CompetitionsListProps {
 export default function CompetitionsClient({
   competitions,
 }: CompetitionsListProps) {
-  const { contains } = useFilter({ sensitivity: "base" });
-
-  const { collection, filter } = useListCollection({
-    initialItems: Object.entries(countries.byIso2).map(([code, country]) => ({
-      label: country.id,
-      value: code,
-    })),
-    filter: contains,
-  });
-
   const marks = [
     { value: 0, label: "closest" },
     { value: 25, label: "close" },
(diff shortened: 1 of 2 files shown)


next-frontend/src/app/(wca)/competitions/competitionClient.tsx
```diff
@@ -83,40 +74,11 @@ export default function CompetitionsClient({
...
-                                <Combobox.ItemIndicator />
-                              </Combobox.Item>
-                            ))}
-                          </Combobox.Content>
-                        </Combobox.Positioner>
-                      </Portal>
-                    </Combobox.Root>
+                    <RegionSelector
+                      label={t("region")}
+                      onRegionChange={(e) => console.log(e)}
```
gregorbg, line 79: I'm sure this is not what you intended 🙃

next-frontend/src/components/RegionSelector.tsx
```diff
@@ -0,0 +1,164 @@
...
+const continentOptions = (t: TFunction) =>
+  Object.values(continents.real)
+    .toSorted((a, b) =>
+      t(`continents.${a.id}`).localeCompare(t(`continents.${b.id}`)),
+    )
+    .map((continent) => ({
+      key: continent.id,
+      label: t(`continents.${continent.name}`),
+      value: continent.id,
+    }));
```
gregorbg: You're piping the continents through the `t` localization just for sorting, then forget the localized string, and apply it again in the next `map` step.

What about `map`ping first, and then afterwards doing `toSorted` with `a.label` (and `b.label, respectively)?

next-frontend/src/components/RegionSelector.tsx
```diff
@@ -0,0 +1,164 @@
...
+      flag: (
+        <Flag
+          code={country.iso2}
+          fallback={country.id}
+          width={32}
+          height={25}
+        />
+      ),
+      value: country.iso2,
+    }));
```
gregorbg: Same as above about "wasting" a `t` invocation in preemptive sort.

next-frontend/src/components/RegionSelector.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+  return (
+    <VStack alignItems="start">
+      <Text textStyle="label">{label}</Text>
+      <Combobox.Root
+        collection={collection}
+        onInputValueChange={(e) => filter(e.inputValue)}
+        onValueChange={(e) => onRegionChange(e.value[0])}
+        width="100%"
+        colorPalette="blue"
```
gregorbg: Why do we want region selectors to always be blue?

next-frontend/src/components/RegionSelector.tsx
```diff
@@ -0,0 +1,164 @@
...
+          </Combobox.IndicatorGroup>
+        </Combobox.Control>
+        <Portal>
+          <Combobox.Positioner>
+            <Combobox.Content justifyContent="flex-start">
+              <Combobox.Empty>No items found</Combobox.Empty>
+              {collection.items.map((item) => (
+                <Combobox.Item item={item} key={item.key}>
+                  {item.flag}
+                  {item.content ? item.content : item.label}
```
gregorbg:
```suggestion
                  {item.content ?? item.label}
```

next-frontend/src/app/(wca)/competitions/competitionClient.tsx
FinnIckler: well it is, because the real use will come in the /competitions PR

next-frontend/src/components/RegionSelector.tsx
FinnIckler: good question, I think that was just how slate built it. just removed it

PR #12460, opened 2025-09-18
Allow Documents to be reordered

FinnIckler:
Introduces a new DocumentPage global to make the documents orderable

### DIFF
--- a/next-frontend/src/app/(wca)/documents/page.tsx
+++ b/next-frontend/src/app/(wca)/documents/page.tsx
@@ -23,7 +23,7 @@ export default async function Documents() {
 
   const documentRelation = documentsResult.documents;
 
-  if (!documentRelation || documentRelation?.length === 0) {
+  if (documentRelation.length === 0) {
     return <Heading>No documents found. Add some</Heading>;
   }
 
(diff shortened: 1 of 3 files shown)


next-frontend/src/globals/Documents.ts
```diff
@@ -0,0 +1,21 @@
+import { GlobalConfig } from "payload";
+
+export const DocumentsPage: GlobalConfig = {
+  slug: "documents-page",
+  label: "Documents Page",
+  fields: [
+    {
+      type: "array",
```
gregorbg, line 8: I think you should give `required: true` to this top-level array as well

next-frontend/src/types/payload.ts
```diff
@@ -1671,4 +1703,4 @@ export interface Auth {
 
 declare module 'payload' {
   export interface GeneratedTypes extends Config {}
-}
+}
```
gregorbg: Mysterious newline deletion

next-frontend/src/app/(wca)/documents/page.tsx
```diff
@@ -12,26 +12,30 @@ import { getPayload } from "payload";
...
-    limit: 0,
+  const documentsResult = await payload.findGlobal({
+    slug: "documents-page",
   });
 
-  const documents = documentsResult.docs;
+  const documentRelation = documentsResult.documents;
 
-  if (documents.length === 0) {
+  if (!documentRelation || documentRelation?.length === 0) {
```
gregorbg: Using `required` as per my other comment would simplify this check

PR #12462, opened 2025-09-18
Add preview functionality for the Homepage

FinnIckler:
While we had
```
livePreview: {
      url: "/",
}
```
set, that actually didn't do anything. It just showed the current Version of the page.

To really introduce Previews, you have to activate [drafts through versioning](https://payloadcms.com/docs/versions/overview#versions-enabled-drafts-disabled) and then set up [draft mode](https://nextjs.org/docs/app/guides/draft-mode) via an API.

I have only activated it for the Home Page so far, but I think every global page could make use of it.

Note: this will most likely break the sqlite database for developers due to changes in the schema (can be reset using `payload migrate:fresh`). I really think we should be switching to mongodb locally and offer a payload import to work with.


gregorbg:
I did that literally 20 seconds ago :joy:

PR #12485, opened 2025-09-22
Auto-accept next competitor when an Accepted registration cancels

dunkOnIT:
It was recently reported that Live Auto Accept is not auto-accepting the waiting list leader when an Accepted user cancels their registration - instead, auto accept only triggers upon the next paid registration, causing a user who registers and pays later than the waiting list leader to get an accepted spot ahead of them. 

I'm pretty sure this functionality _used to_ exist, but seems like it was taken out when I refactored to bulk auto accept - and I never added it back in. 

In either case, we need to do the following: 
1. trigger auto accept on the waiting list leader when an Accepted user's registration is cancelled
2. prevent auto-accept of a pending user when there are users on the waiting list

(2) is the quick-fix, so I'll submit a PR for that separately, but (1) raises some interesting implementation questions that I want to tease out in this PR. 

## How To Trigger The Auto-Accept
Two options: 
1.  `after_save` hook that checks if the status was changed away from `accepted`
2. Check the status change in `lanes/competing.rb#update!`

(1) feels cleaner/more idiomatic, but (2) seems more consistent with out existing logic - notably, status change emails are triggered from within the competing lane, not on the model. Thus, I think we should do (2) - keeping all of the update-triggered logic in one place. 

## What Action to Take in Order to Auto-Accept
Again, two options: 
1. Call `competition.waiting_list.first.attempt_auto_accept`
2. Call `competition.bulk_auto_accept`

(1) is the straightforward/obvious option - but there are potentially issues with concurrency if multiple users cancel their registration at the same time [1], and we try to access the same waiting_list_leader. This seems exceedingly unlikely given that it is not the mad dash of registration opening - but perhaps it's still something to consider.

The solution to this concurrency concern, I think, would be to call `bulk_auto_accept` _every time_ this hook is triggered. Although now that I write this, I'm not sure it would get around the concurrency issues - surely we could still have the same waiting list leader being called. 

If we _did_ want to call `bulk_auto_accept`, then I would suggest adding a check in the function that takes into account how many open slots it's trying to fill - and only querying pending registrations if there are still open slots after the waiting list is taken into account, to save on redundant DB calls in the case that an organizer cancels multiple registrations at once.

[1] organizers cancelling multiple registrations is also a case I though about, but in this case we handle the cancellations sequentially so our current implementation doesn't fall afoul of concurrency concerns - I'm not sure if we should be implementation-agnostic when considering such concerns, though


spec/factories/registrations.rb
```diff
@@ -120,7 +120,7 @@
       after(:create) do |registration|
         payment = FactoryBot.build(:registration_payment, registration: registration, user: registration.user,
                                                           amount_lowest_denomination: registration.competition.base_entry_fee_lowest_denomination)
-        payment.save(validate: false)
+        RegistrationPayment.insert!(payment.attributes.except("id", "created_at", "updated_at"))
```
gregorbg: I don't like manual inserts this way. Why is it (seemingly) necessary here?

app/models/registration.rb
```diff
@@ -75,6 +75,7 @@ class Registration < ApplicationRecord
   validates :guests, numericality: { less_than_or_equal_to: DEFAULT_GUEST_LIMIT, if: :guests_unrestricted?, frontend_code: Registrations::ErrorCodes::UNREASONABLE_GUEST_COUNT }
 
   after_save :mark_registration_processing_as_done
+  after_save :trigger_bulk_auto_accept, if: -> { competing_status_previously_changed?(from: 'accepted', to: 'cancelled') }
```
gregorbg: What about when someone is being rejected by organizers / WIC? From the perspective of other competitors who are on the waiting list, this should trigger AA as well for them to pick the now-free spot.

spec/factories/registrations.rb
dunkOnIT: Because `save` still triggers `after_create` hooks - I was somehow under the impression that `validate: false` prevented this, but it certainly does not. Using `insert` appears to be the only way to bypass the `after_create` hook.

I could use another approach now that we have `is_completed` - creating a payment as `is_completed: false`, and then using `update_columns` (I think that's the function name - the one that updates a column directly) - but `update_columns` seems quite similar to the `insert` pattern. (With, I concede, the notable exception that it only bypasses validations/hooks for _one_ column, instead of the entire model)

app/models/registration.rb
dunkOnIT: Great point - changed it to trigger on _any_ change away from `accepted`

spec/factories/registrations.rb
dunkOnIT: I talked myself into changing it :P

app/models/registration.rb
```diff
@@ -614,7 +620,8 @@ def attempt_auto_accept(mode)
 
   private def auto_accept_failure_reason(mode)
     return Registrations::ErrorCodes::OUTSTANDING_FEES if outstanding_entry_fees.positive?
-    return Registrations::ErrorCodes::AUTO_ACCEPT_NOT_ENABLED if auto_accept_preference_disabled? || (auto_accept_preference.to_sym != mode)
+    return Registrations::ErrorCodes::AUTO_ACCEPT_NOT_ENABLED if auto_accept_preference_disabled? ||
+                                                                 (auto_accept_preference == 'bulk' && auto_accept_preference.to_sym != mode)
```
gregorbg: ...doesn't this specific part of the expression (the && in brackets) just boil down to `mode != :bulk`?

app/models/registration.rb
```diff
@@ -75,6 +75,7 @@ class Registration < ApplicationRecord
   validates :guests, numericality: { less_than_or_equal_to: DEFAULT_GUEST_LIMIT, if: :guests_unrestricted?, frontend_code: Registrations::ErrorCodes::UNREASONABLE_GUEST_COUNT }
 
   after_save :mark_registration_processing_as_done
+  after_save :trigger_bulk_auto_accept, if: -> { competing_status_previously_changed?(from: 'accepted') }
```
gregorbg: Please jog my memory: Why is it a good idea to always trigger *bulk* regardless of whether the AA preference is set to bulk or live? (and does this check whether AA is enabled at all??)

dunkOnIT: > (and does this check whether AA is enabled at all??)

Good point - it'll get stopped by the `attempt_auto_accept` checks, but the hook should only trigger if AA is enabled :sweat_smile: 

> Why is it a good idea to always trigger bulk regardless of whether the AA preference is set to bulk or live?

I am choosing to trigger bulk auto accept, instead of just auto-accepting a waiting list leader, as this handles edge cases where we might "miss" a trigger and end up with permanently open spots in the accepted list which the organizer needs to manually handle.

Put slightly differently, `live` auto accept is usually triggered by action on a particular registration - so we can have 100% confidence that we're targeting the right registration. This accept-if-cancelled logic is trying to _fill open spots_, not change the state of one particular registration - so it seems more edge-case-proof to go with a "fill all spots" logic than a "change this one registration" logic.

app/models/registration.rb
dunkOnIT: I'm not sure what that's doing there in the first place - removed and checking if any tests fail

app/models/registration.rb
dunkOnIT: (Added the check for if auto accept is enabled)

app/models/registration.rb
```diff
@@ -75,6 +75,9 @@ class Registration < ApplicationRecord
   validates :guests, numericality: { less_than_or_equal_to: DEFAULT_GUEST_LIMIT, if: :guests_unrestricted?, frontend_code: Registrations::ErrorCodes::UNREASONABLE_GUEST_COUNT }
 
   after_save :mark_registration_processing_as_done
+  after_save :trigger_bulk_auto_accept, if: lambda { |_lambda|
```
gregorbg:
```suggestion
  after_save :trigger_bulk_auto_accept, if: -> {
```

app/models/registration.rb
```diff
@@ -75,6 +75,9 @@ class Registration < ApplicationRecord
...
+    competition.auto_accept_preference == 'live' && competing_status_previously_changed?(from: 'accepted')
```
gregorbg:
```suggestion
    competition.auto_accept_preference_live? && competing_status_previously_changed?(from: 'accepted')
```

app/models/registration.rb
```diff
@@ -553,6 +556,11 @@ def serializable_hash(options = nil)
     super(DEFAULT_SERIALIZE_OPTIONS.merge(options || {}))
   end
 
+  # Allows us to trigger bulk_auto_accept from within an instance of Registration
+  def trigger_bulk_auto_accept
+    self.class.bulk_auto_accept(competition)
```
gregorbg:
```suggestion
    self.class.bulk_auto_accept(self.competition)
```

gregorbg: Not strictly necessary, but helps to distinguish between the current registration instance's competition, and the `bulk_auto_accept` parameter (which is also named `competition`)

dunkOnIT: Done

app/models/registration.rb
dunkOnIT: Done

app/models/registration.rb
dunkOnIT: Done

dunkOnIT: Turns out this is a rubocop: 
```bash
app/models/registration.rb:78:45: C: [Corrected] Style/Lambda: Use the lambda method for multiline lambdas.
  after_save :trigger_bulk_auto_accept, if: -> {
                                            ^^
```

gregorbg (approved):

PR #12486, opened 2025-09-22
Pending registrations should auto-accept only Waiting List if it is non-empty

dunkOnIT:
Hotfix for the issue described in #12485:
> It was recently reported that Live Auto Accept is not auto-accepting the waiting list leader when an Accepted user cancels their registration - instead, auto accept only triggers upon the next paid registration, causing a user who registers and pays later than the waiting list leader to get an accepted spot ahead of them.
> I'm pretty sure this functionality used to exist, but seems like it was taken out when I refactored to bulk auto accept - and I never added it back in.

This fix just prevents a pending registration from being accepted directly onto the accepted list if there are any entries on the waiting list - which handles the most immediate issue. It also reverses the logic of status assignment - from defaulting to `accepted`, to defaulting to `waiting_list`, which seems the more prudent approach.

### DIFF
--- a/app/models/registration.rb
+++ b/app/models/registration.rb
@@ -614,6 +614,8 @@ class Registration < ApplicationRecord
     return Registrations::ErrorCodes::AUTO_ACCEPT_NOT_ENABLED if auto_accept_preference_disabled? || (auto_accept_preference.to_sym != mode)
     return Registrations::ErrorCodes::INELIGIBLE_FOR_AUTO_ACCEPT unless competing_status_pending? || waiting_list_leader?
     return Registrations::ErrorCodes::AUTO_ACCEPT_DISABLE_THRESHOLD if competition.auto_accept_threshold_reached?
+    # Pending registrations can still be accepted onto the waiting list, so we only raise an error for already-waitlisted registrations
+    return Registrations::ErrorCodes::COMPETITOR_LIMIT_REACHED if competing_status_waiting_list? && competition.registration_full_and_accepted?
 
     Registrations::ErrorCodes::REGISTRATION_NOT_OPEN unless competition.registration_currently_open?
   end
@@ -628,10 +630,16 @@ class Registration < ApplicationRecord
   end
 
   private def build_auto_accept_payload
-    status = if competition.registration_full_and_accepted? && competing_status_pending?
-               Registrations::Helper::STATUS_WAITING_LIST
-             else
+    # The default case should be that a user's status will be waiting list, unless specific criteria for acceptance are met
+    # We can only accept a registration in one of these cases:
+    # 1. There must be space on the Accepted list, and
+    # 2. They're first on the waiting list, or
+    # 3. The waiting list is empty and they're on the pending list
+    status = if !competition.registration_full_and_accepted? &&
+             (waiting_list_leader? || competing_status_pending? && waiting_list.empty?)
                Registrations::Helper::STATUS_ACCEPTED
+             else
+               Registrations::Helper::STATUS_WAITING_LIST
              end
 
     # String keys because this is mimicing a params payload
--- a/spec/models/registration_spec.rb
+++ b/spec/models/registration_spec.rb
@@ -708,7 +708,7 @@ RSpec.describe Registration do
       expect(waiting_list_reg.reload.competing_status).to eq('accepted')
     end
 
-    context 'auto-accept does not succed' do
+    context 'auto-accept does not succeed' do
       it 'if a waitlisted registration is not first in the waiting list' do
         create_list(:registration, 3, :waiting_list, competition: auto_accept_comp)
         waiting_list_reg = create(:registration, :waiting_list, competition: auto_accept_comp)
@@ -719,6 +719,14 @@ RSpec.describe Registration do
         waiting_list_reg.attempt_auto_accept(:live)
         expect(waiting_list_reg.reload.competing_status).to eq('waiting_list')
         expect(waiting_list_reg.registration_history.last[:changed_attributes][:auto_accept_failure_reasons]).to eq("-7003")
+      end
+
+      it 'does not accept a pending registration ahead of a waitlisted one' do
+        create(:registration, :waiting_list, competition: auto_accept_comp)
+        create(:registration_payment, :skip_create_hook, registration: reg, competition: auto_accept_comp)
+
+        reg.attempt_auto_accept(:live)
+        expect(reg.reload.competing_status).to eq('waiting_list')
       end
 
       it 'if registration_payment.is_completed: false' do
@@ -839,7 +847,6 @@ RSpec.describe Registration do
       end
       let!(:prevented_reg) { create(:registration, competition: limited_comp) }
 
-      # Fails because waiting_list_position persists when it shouldnt; #11173 should fix
       it 'if competitor limit is reached and first on waiting list' do
         create_list(:registration, 5, :accepted, competition: limited_comp)
 
@@ -848,7 +855,7 @@ RSpec.describe Registration do
         expect(waiting_list_reg.reload.competing_status).to eq('waiting_list')
 
         waiting_list_reg.attempt_auto_accept(:live)
-        expect(waiting_list_reg.registration_history.last[:changed_attributes][:auto_accept_failure_reasons]).to eq(['Competitor limit reached.'].to_s)
+        expect(waiting_list_reg.registration_history.last[:changed_attributes][:auto_accept_failure_reasons]).to eq('-4006')
         expect(waiting_list_reg.reload.competing_status).to eq('waiting_list')
       end
 


app/models/registration.rb
```diff
@@ -614,6 +614,8 @@ def attempt_auto_accept(mode)
     return Registrations::ErrorCodes::AUTO_ACCEPT_NOT_ENABLED if auto_accept_preference_disabled? || (auto_accept_preference.to_sym != mode)
     return Registrations::ErrorCodes::INELIGIBLE_FOR_AUTO_ACCEPT unless competing_status_pending? || waiting_list_leader?
     return Registrations::ErrorCodes::AUTO_ACCEPT_DISABLE_THRESHOLD if competition.auto_accept_threshold_reached?
+    # Pending registrations can still be accepted onto the waiting list, so we only raise an error for already-waitlisted registrations
+    return Registrations::ErrorCodes::COMPETITOR_LIMIT_REACHED if competing_status_waiting_list? && competition.registration_full_and_accepted?
```
dunkOnIT, line 621: This check wasn't present before because it was handled by the validations instead - now, I'm moving it into auto accept because I'm changing the default status we assign in auto_accept to waiting list.

Previously, if the comp was full and a waitlisted registration had .attempt_auto_accept run on it: 
- it would get assigned the `accepted` status
- the "competition full" validation would prevent the status change - this validation error is what the tests used to check for 

Now, however, the following would happen: 
- it would get assigned the `waiting_list` status
- the "competition full" validation would not trigger, because the reg isn't being accepted
- we would try and process the update, attempting to change its status (again) to `waiting_list` 

It seems better to explicitly log an error saying "we tried to auto-accept this waitlisted registration but the competitor limit is reached so we couldn't process that update"

app/models/registration.rb
```diff
@@ -628,10 +630,16 @@ def attempt_auto_accept(mode)
...
   private def build_auto_accept_payload
-    status = if competition.registration_full_and_accepted? && competing_status_pending?
-               Registrations::Helper::STATUS_WAITING_LIST
-             else
+    # The default case should be that a user's status will be waiting list, unless specific criteria for acceptance are met
+    # We can only accept a registration in one of these cases:
+    # 1. There must be space on the Accepted list, and
+    # 2. They're first on the waiting list, or
+    # 3. The waiting list is empty and they're on the pending list
+    status = if !competition.registration_full_and_accepted? &&
```
gregorbg: What about introducing a custom method in `competition.rb` (along the lines of `def registration_spot_available?`, name tentative) that returns this very negation? Just to improve readability

gregorbg: Unrelated: The `if` can become a bit less "dense" and easier to parse if you start by `return STATUS_WAITING_LIST if competition.registration_full_and_accepted?`.
You're currently packing all different cases into one "master-if", I'd rather have the code process the different branches/possibilities step-by-step.

app/models/registration.rb
```diff
@@ -628,10 +630,16 @@ def attempt_auto_accept(mode)
...
+                (waiting_list_leader? || (competing_status_pending? && waiting_list.empty?))
```
gregorbg: Does `waiting_list_leader?` internally check the competing status? It seems unintuitive that one part of the `||` performs an explicit check on `competing_status_pending?` and the other part (seemingly) assumes `competing_status_waiting_list?` secretly under the hood.

app/models/registration.rb
gregorbg: (and doing so would ironically invalidate my other comment above about a readability-helper method...)

app/models/registration.rb
dunkOnIT: Yes it does check waiting list status

gregorbg: That is good for data integrity, but on the surface still confusing when reading this (very neuralgic) part of our code

app/models/registration.rb
dunkOnIT: I like the readability change! Just pushed it

app/models/registration.rb
```diff
@@ -628,10 +630,16 @@ def attempt_auto_accept(mode)
...
   private def build_auto_accept_payload
-    status = if competition.registration_full_and_accepted? && competing_status_pending?
-               Registrations::Helper::STATUS_WAITING_LIST
-             else
+    # The default case should be that a user's status will be waiting list, unless specific criteria for acceptance are met
+    # We can only accept a registration in one of these cases:
+    # 1. There must be space on the Accepted list, and
+    # 2. They're first on the waiting list, or
+    # 3. The waiting list is empty and they're on the pending list
+    return Registrations::Helper::STATUS_WAITING_LIST if competition.registration_full_and_accepted?
```
gregorbg: Now the code comment block may need to be adjusted / realigned with the existing code

app/models/registration.rb
```diff
@@ -628,10 +630,16 @@ def attempt_auto_accept(mode)
...
+    status = if waiting_list_leader? || (competing_status_pending? && waiting_list.empty?)
                Registrations::Helper::STATUS_ACCEPTED
+             else
+               Registrations::Helper::STATUS_WAITING_LIST
              end
```
gregorbg: In the spirit of my previous comment, about explicitly checking the competing status in every part of the `||` logic: If you do that, it effectively turns into a decision tree based on competing status.

My first intuition was to turn this into a `switch competing_status`. But I see now that the individual branches may get convoluted because you have to perform additional logic checks even after confirming that the competing_status is such-and-such. Nonetheless, I still like the overall idea of "apply certain logic based on what the current competing_status of the registration trying to be auto-accepted is". Can you play around with this idea for ~10 minutes (feel free to consult LLMs, although the prompting might get tricky to make it understand our business logic and use-case) and see whether you can come up with something?

gregorbg: After submitting, I had this idea: (rough draft)
```ruby
def can_be_accepted?
  case self.competing_status
  when Registration.competing_statuses[:waiting_list]
    return self.waiting_list_leader?
  when Registration.competing_statuses[:waiting_list]
    return self.waiting_list.empty?
  else
    return false
  end
end
```

and then use it as `status = self.can_be_accepted? ? STATUS_ACCEPTED : STATUS_WAITING_LIST`

dunkOnIT: Just pushed - I like the `case` statement, I think it cleans things up dramatically. Also refactored the function calls so that we can early return just the status value

app/models/registration.rb
```diff
@@ -627,14 +631,18 @@ def attempt_auto_accept(mode)
...
-    # String keys because this is mimicing a params payload
-    { 'user_id' => user_id, 'competing' => { 'status' => status } }
+    # Only Pending and Waitlisted registrations can be auto-accepted
+    case competing_status
+    when Registrations::Helper::STATUS_WAITING_LIST
+      # Only the waiting list leader can be accepted from the waiting list
+      Registrations::Helper::STATUS_ACCEPTED if waiting_list_leader?
+    when Registrations::Helper::STATUS_PENDING
+      # Pending registrations should go to waiting list unless there is no waiting list
+      waiting_list.empty? ? Registrations::Helper::STATUS_ACCEPTED : Registrations::Helper::STATUS_WAITING_LIST
```
gregorbg: How does the check `waiting_list.empty?` cover the case "there is no waiting list" (mentioned in your comment)?

app/models/registration.rb
```diff
@@ -593,7 +593,9 @@ def attempt_auto_accept(mode)
       return { succeeded: false, info: failure_reason }
     end
 
-    update_payload = build_auto_accept_payload
+    # String keys because this is mimicing a params payload
+    update_payload = { 'user_id' => user_id, 'competing' => { 'status' => auto_accept_competing_status } }
```
gregorbg: Hmm... After your recent refactor, `auto_accept_competing_status` can now return `nil`. (Either when the `when` case doesn't include the current registration's status, like `cancelled` or `rejected`, or when the check within the `case` branch doesn't match, like the `if waiting_list_leader?` returning false).

Do we want this? What happens if you pass in an actual `nil` value to this payload, does the database even allow setting the status to `nil`?

dunkOnIT: Hmm, I've refactored to return `waiting_list` as a default value, although it makes it slightly clunky. 

```ruby
private def auto_accept_competing_status
    # The only possible status is waiting list if registration is full
    return Registrations::Helper::STATUS_WAITING_LIST if competition.registration_full_and_accepted?

    # Only Pending and Waitlisted registrations can be auto-accepted
    case competing_status
    when Registrations::Helper::STATUS_WAITING_LIST
      # We can only accept the waiting list leader - a non-waiting-list-leader registration should never make it here
      waiting_list_leader? ? Registrations::Helper::STATUS_ACCEPTED : Registrations::Helper::STATUS_WAITING_LIST
    when Registrations::Helper::STATUS_PENDING
      # Pending registrations should go to waiting list unless there is no one on the waiting list
      waiting_list.empty? ? Registrations::Helper::STATUS_ACCEPTED : Registrations::Helper::STATUS_WAITING_LIST
    else
      # Waiting list is our default status - but registrations with other statuses should never make it to here
      Registrations::Helper::STATUS_WAITING_LIST
    end
  end
```
(posting here because trying to install overcommit broke my ability to push commits)

gregorbg: What you're trying to do can simply be achieved by applying `STATUS_WAITING_LIST` as a nil-safe fallback value at the point in the code where the return value of `auto_accept_competing_status` is being used

app/models/registration.rb
```diff
@@ -593,17 +593,20 @@ def attempt_auto_accept(mode)
...
-    update_payload = build_auto_accept_payload
-    auto_accepted_registration = Registrations::RegistrationChecker.apply_payload(self, update_payload, clone: false)
+    new_competing_status = eligible_for_accepted_status? ? Registrations::Helper::STATUS_ACCEPTED : Registrations::Helper::STATUS_WAITING_LIST
+    # String keys because this is mimicing a params payload
+    update_payload = { 'user_id' => user_id, 'competing' => { 'status' => new_competing_status } }
+
+    updated_registration = Registrations::RegistrationChecker.apply_payload(self, update_payload, clone: false)
 
-    if auto_accepted_registration.valid?
+    if updated_registration.valid?
```
dunkOnIT, line 602: "accept" is becoming overloaded here - so I'm changing the name to clarify that this is the result of the update operation, not limited to registrations which have received the `accepted` status specifically

app/models/registration.rb
```diff
@@ -627,14 +632,16 @@ def attempt_auto_accept(mode)
...
-               Registrations::Helper::STATUS_WAITING_LIST
-             else
-               Registrations::Helper::STATUS_ACCEPTED
-             end
+  private def eligible_for_accepted_status?
+    return false if competition.registration_full_and_accepted?
 
-    # String keys because this is mimicing a params payload
-    { 'user_id' => user_id, 'competing' => { 'status' => status } }
+    case competing_status
```
dunkOnIT, line 638: I refactored down to this implementation, and only realized afterwards that it's exactly what you suggested initially - well played :P

PR #12526, opened 2025-10-01
Improve registration event diff

### LINKED ISSUE #12252: Improve event diff in Registration History and Update Request
Image

kr-matthews:
Mostly resolves #12252. The backend is only sending the toggled events for the registration history view, so this is the best I can do on the frontend alone. If someone wants to update the backend to alter the data being sent, I could build on that to show the added and removed events separately. Additionally, here seems to be a bug where the backend is only sending the added events and ignoring the removed events. This behaviour is happening on main, so not related to my changes.

[Screencast from 2025-09-30 06:43:33 PM.webm](https://github.com/user-attachments/assets/3c08c969-4fca-4bff-8878-43d74c123cc5)

Requesting to change events via contact form:

[Screencast from 2025-09-30 06:05:25 PM.webm](https://github.com/user-attachments/assets/a285fe52-a7b8-4640-95d0-e0162ac7e024)

I'm not sure whether `registration.competing.event_ids` is the appropriate way to get the originally selected events.

### DIFF
--- a/app/webpacker/components/RegistrationsV2/RegistrationEdit/RegistrationHistory.jsx
+++ b/app/webpacker/components/RegistrationsV2/RegistrationEdit/RegistrationHistory.jsx
@@ -62,15 +62,15 @@ export default function RegistrationHistory({ registrationId }) {
               <Table.Cell>
                 {Object.entries(entry.changed_attributes).map(
                   ([k, v]) => (
-                    <>
+                    <React.Fragment key={k}>
                       {k === 'event_ids' ? (
-                        <span key={k}>
+                        <span>
                           Toggled events
                           {' '}
                           <EventIcons ids={v} />
                         </span>
                       ) : (
-                        <span key={k}>
+                        <span>
                           Changed
                           {' '}
                           {k}
@@ -81,7 +81,7 @@ export default function RegistrationHistory({ registrationId }) {
                         </span>
                       )}
                       <br />
-                    </>
+                    </React.Fragment>
                   ),
                 )}
               </Table.Cell>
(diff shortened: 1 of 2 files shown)


app/webpacker/components/RegistrationsV2/RegistrationEdit/RegistrationHistory.jsx
```diff
@@ -69,17 +62,26 @@ export default function RegistrationHistory({ registrationId }) {
...
-                    <span key={k}>
-                      Changed
-                      {' '}
-                      {k}
-                      {' '}
-                      to
-                      {' '}
-                      {formatHistoryColumn(k, v)}
-                      {' '}
+                    <>
```
gregorbg: I think this `React.Fragment` should be the one to get the `key`, not the individual child spans.

app/webpacker/components/RegistrationsV2/RegistrationEdit/RegistrationHistory.jsx
```diff
@@ -98,3 +100,7 @@ export default function RegistrationHistory({ registrationId }) {
     </>
   );
 }
+
+function EventIcons({ ids }) {
+  return events.official.map((e) => ids.includes(e.id) && <EventIcon key={e.id} id={e.id} style={{ cursor: 'unset' }} />);
+}
```
gregorbg, line 108: That's a lot of React shenanigans happening on one line. My gut is unsure whether this is still okay or too much -- do you commonly see this pattern elsewhere?

My slightly more preferred approach would be a `filter` based on `includes`, and then a separate `map` for the already-filtered list. But I'm 50/50 on which version looks cleaner.

kr-matthews: I've moved the map function body onto its own line. If this were a 'regular' function then I think the filter & map separately would be good. But for component returns, it's common to just use a map since `false` will be ignored when rendered.

PR #12527, opened 2025-10-01
Optimize attempt backfill

FinnIckler:
Query performance: `Query OK, 26155354 rows affected (3 min 26.72 sec)`
I ran the integrity check and found two missmatches:
Result Id 7253786 and 7096277 which both had attempts that were deleted (set to 0) later. Which caused the
```
def create_or_update_attempts
    attempts = (1..5).filter_map do |n|
      value = public_send(:"value#{n}")

      { value: value, attempt_number: n, result_id: id } unless value.zero?
    end
    ResultAttempt.upsert_all(attempts)
  end
```
not to work because it doesn't deal with deleted attempts.
Running the integrity check just takes a couple seconds so just deleting those attempts and rerunning the `create_or_update_attempts` method for those is easy

### DIFF
--- a/lib/tasks/move_results_to_attempt_tables.rake
+++ b/lib/tasks/move_results_to_attempt_tables.rake
@@ -3,7 +3,18 @@
 namespace :results do
   desc "Migrates all results to attempts"
   task migrate_attempts: [:environment] do
-    Result.find_each(&:create_or_update_attempts)
+    ActiveRecord::Base.connection.execute <<-SQL.squish
+    INSERT IGNORE INTO result_attempts (value, attempt_number, result_id)
+    SELECT value1, 1, id FROM results WHERE value1 != 0
+    UNION ALL
+    SELECT value2, 2, id FROM results WHERE value2 != 0
+    UNION ALL
+    SELECT value3, 3, id FROM results WHERE value3 != 0
+    UNION ALL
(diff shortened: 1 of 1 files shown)


FinnIckler:
We should also add an index on value before we do this though as it will greatly decrease the time to query the attempt table

lib/tasks/move_results_to_attempt_tables.rake
```diff
@@ -3,7 +3,43 @@
 namespace :results do
   desc "Migrates all results to attempts"
   task migrate_attempts: [:environment] do
-    Result.find_each(&:create_or_update_attempts)
+    ActiveRecord::Base.connection.execute <<-SQL.squish
+    INSERT IGNORE INTO result_attempts (value, attempt_number, result_id)
```
gregorbg, line 7: What does the `IGNORE` part do here?

FinnIckler: Well currently already have a lot of results that have attempts already, so IGNORE makes it that on conflict (we have a uniqueness constraint on result_id and attempt_number) it does nothing

FinnIckler: another option would be to do another query to figure out which result_ids already have attempts, but that would make it a good more complicated

PR #12531, opened 2025-10-02
Use Attempts instead of value1-5 when displaying results

FinnIckler:
Requires the attempts backfill to be run.

This also affects competition results, but if I interpret these associations correctly, results aren't even preloaded in competitions#show?

```
associations = {
      competition_venues: {
        venue_rooms: [:schedule_activities],
      },
      # FIXME: this part is triggerred by the competition menu generator when generating the psychsheet event list, should we care?
      competition_events: {
        # NOTE: we hit this association through competition.has_fees?, which then calls 'has_fee?' on each competition_event, which then use the competition to get the currency.
        competition: [],
        event: [],
        # NOTE: we eventually hit the rounds->competition->competition_event in the TimeLimit 'to_s' method when having cumulative limit across rounds
        rounds: {
          competition: { rounds: [:competition_event] },
          competition_event: [],
        },
      },
      rounds: {
        # Used by TimeLimit, but this is a weird includes...
        competition: { rounds: [:competition_event] },
      },
    }
    @competition = competition_from_params(includes: associations)
```

### DIFF
--- a/app/controllers/results_controller.rb
+++ b/app/controllers/results_controller.rb
@@ -83,7 +83,6 @@ class ResultsController < ApplicationController
         JOIN results ON results.id = value_and_id % 1000000000
         ORDER BY value, person_name
       SQL
-
     elsif @is_results
       # rubocop:disable Style/ConditionalAssignment
       #   for better readability of the individual indentations of the SQL queries
@@ -156,7 +155,6 @@ class ResultsController < ApplicationController
           #{@gender_condition}
         ORDER BY value, results.country_id, start_date, person_name
       SQL
-
     else
       flash[:danger] = t(".unknown_show")
       return redirect_to rankings_path
--- a/db/migrate/20260119131442_add_records_indices_to_result_attempts.rb
+++ b/db/migrate/20260119131442_add_records_indices_to_result_attempts.rb
@@ -1,7 +1,11 @@
 # frozen_string_literal: true
 
 class AddRecordsIndicesToResultAttempts < ActiveRecord::Migration[8.1]
+  # rubocop:disable Rails/BulkChangeTable
   def change
     add_index :result_attempts, :value
+    add_index :results, %i[average person_name competition_id round_type_id], name: "results_n_results_average_speedup"
+    add_index :results, %i[best person_name competition_id round_type_id], name: "results_n_results_single_speedup"
   end
+  # rubocop:enable Rails/BulkChangeTable
 end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -1123,6 +1123,8 @@ ActiveRecord::Schema[8.1].define(version: 2026_01_19_131442) do
     t.integer "value3", default: 0, null: false
     t.integer "value4", default: 0, null: false
     t.integer "value5", default: 0, null: false
+    t.index ["average", "person_name", "competition_id", "round_type_id"], name: "results_n_results_average_speedup"
+    t.index ["best", "person_name", "competition_id", "round_type_id"], name: "results_n_results_single_speedup"
     t.index ["competition_id", "updated_at"], name: "index_Results_on_competitionId_and_updated_at"
     t.index ["competition_id"], name: "Results_fk_tournament"
     t.index ["country_id"], name: "_tmp_index_Results_on_countryId"


app/models/concerns/resultable.rb
```diff
@@ -206,11 +206,15 @@ def missed_combined_round_cutoff?
...
-                      SolveTime.new(event_id, :single, value5)].freeze
+    @solve_times ||= if is_a?(InboxResult) # TODO: Remove this when we switch from InboxResult to LiveAttempts
+                       [SolveTime.new(event_id, :single, value1),
+                        SolveTime.new(event_id, :single, value2),
+                        SolveTime.new(event_id, :single, value3),
+                        SolveTime.new(event_id, :single, value4),
+                        SolveTime.new(event_id, :single, value5)].freeze
+                     else
+                       result_attempts.map { |r| SolveTime.new(event_id, :single, r.value) }.freeze
+                     end
```
gregorbg: You can just override `def solve_times` in IBR instead of doing an `is_a?` check here

spec/factories/results.rb
```diff
@@ -121,6 +121,14 @@
       person { FactoryBot.create(:person) }
     end
 
+    after(:build) do |result, _options|
+      result.result_attempts = (1..5).filter_map do |n|
+        value = result.public_send(:"value#{n}")
+
+        ResultAttempt.new({ value: value, attempt_number: n }) unless value.zero?
+      end
```
gregorbg: I'd reckon you can reuse the snippet from your other PR here, so perhaps that other one should get merged first (and then just `map` into `ResultAttempt.new` here)

gregorbg (changes requested):

spec/factories/results.rb
FinnIckler: I copied it for now

FinnIckler:
I am beginning to go into the territory that I have to rewrite the validators and I just started with the cutoff one.
I had to change the definition of `<=>` in solve time because it can now happen in `results_similar_to` that one result has less solve times than to other.

dunkOnIT:
@thewca-bot deploy staging

app/helpers/results_helper.rb
```diff
@@ -4,13 +4,20 @@ module ResultsHelper
...
       classes << "trimmed" if result.trimmed_indices.include?(i)
       classes << "best" if i == result.best_index
       classes << "worst" if i == result.worst_index
       content_tag :td, solve_time.clock_format, class: classes.join(' ')
     end.reduce(:+)
+
+    # Currently there are always 5 solves in any results, even for mean of 3 or Bo1
+    # That's why we always need to fill it up with 5 tds
+    missing_solves = 5 - result.solve_times.length
+    return completed_solves + Array.new(missing_solves, (content_tag :td)).reduce(:+) if missing_solves != 0
```
gregorbg:
```suggestion
    return completed_solves + Array.new(missing_solves, content_tag(:td)).reduce(:+) if missing_solves != 0
```

app/models/concerns/resultable.rb
```diff
@@ -184,11 +184,11 @@ def missed_combined_round_cutoff?
...
-                      SolveTime.new(event_id, :single, value2),
-                      SolveTime.new(event_id, :single, value3),
-                      SolveTime.new(event_id, :single, value4),
-                      SolveTime.new(event_id, :single, value5)].freeze
+    @solve_times ||= result_attempts.map { |r| SolveTime.new(event_id, :single, r.value) }.freeze
+  end
+
+  def worst_index
+    sorted_solves_with_index.max[1]
   end
```
gregorbg: This method body mysteriously moved around and I don't see why

app/models/inbox_result.rb
```diff
@@ -15,6 +15,10 @@ def person_name
     inbox_person&.name || "<person_id=#{person_id}>"
   end
 
+  def solve_times
+    @solve_times ||= attempts.filter_map { |attempt| SolveTime.new(event_id, :single, attempt) if attempt != 0 }.freeze
```
gregorbg: Try to reference `SolveTime::SKIPPED_VALUE` here instead of the literal `0`

spec/models/result_spec.rb
```diff
@@ -431,9 +423,10 @@
     context "check number of non-zero solves" do
       def result_with_n_solves(n, options)
         result = build(:result, options)
-        (1..5).each do |i|
+        (1..5).map do |i|
```
gregorbg: Why are you turning this into a `map` if you're not using the result of the `map` call?

spec/models/result_spec.rb
```diff
@@ -431,9 +423,10 @@
...
           result.send :"value#{i}=", i <= n ? 42 : 0
         end
+        result.result_attempts = result.result_attempts_attributes.map(&ResultAttempt.method(:new))
         result
```
gregorbg:
```suggestion
        result_attempts = (1..5).map do |i|
          [:"value#{i}", i <= n ? 42 : 0]
        end
        build(:result, options, **result_attempts)
```

spec/lib/results_validators/individual_results_validator_spec.rb
```diff
@@ -71,10 +71,11 @@
...
                                        cutoff: cutoff, event_id: "444", round: round44)
-        res_over_with_results.update!(value3: res_over_with_results.value2,
-                                      value4: res_over_with_results.value2,
-                                      value5: res_over_with_results.value2,
-                                      average: res_over_with_results.value2)
+        res_over_with_results.update(value3: res_over_with_results.value2,
+                                     value4: res_over_with_results.value2,
+                                     value5: res_over_with_results.value2,
+                                     average: res_over_with_results.value2)
+        res_over_with_results.create_or_update_attempts if result_kind == :result
```
gregorbg: Why does the `update` call immediately above not trigger the correct "backfilling" callbacks already? (In other words: Why do you have to trigger this manually here yourself?)

spec/lib/results_validators/individual_results_validator_spec.rb
```diff
@@ -197,6 +198,7 @@
         res_mbf = create(result_kind, :mbf, competition: competition1, round: round_333mbf)
         # 8 points in 60:02 (ie: reached the time limit and got +2)
         res_mbf.update(value2: 910_360_200)
+        res_mbf.result_attempts.reload if result_kind == :result
```
gregorbg: Same idea as my question above: This should trigger the update hooks in the model, why is it necessary to _also_ call a manual reload here?

app/jobs/add_live_result_job.rb
```diff
@@ -13,15 +13,11 @@ def perform(results, round_id, registration_id, entered_by)
...
-      value1: results[0],
-      value2: results[1] || 0,
-      value3: results[2] || 0,
-      value4: results[3] || 0,
-      value5: results[4] || 0,
       event_id: event.id,
       round_type_id: round.round_type_id,
       round_id: round.id,
       format_id: format.id,
+      result_attempts: results.map.with_index(1) { |r, index| ResultAttempt.new({ attempt_number: index, value: r }) },
```
gregorbg:
```suggestion
      result_attempts: results.map.with_index(1) { |r, index| ResultAttempt.new(attempt_number: index, value: r) },
```

app/controllers/live_controller.rb
```diff
@@ -48,15 +48,11 @@ def update_result
...
-      value1: results[0],
-      value2: results[1],
-      value3: results[2],
-      value4: results[3] || 0,
-      value5: results[4] || 0,
       event_id: round.event.id,
       round_type_id: round.round_type_id,
       round_id: round.id,
       format_id: round.format_id,
+      result_attempts: results.map.with_index(1) { |r, index| ResultAttempt.new({ attempt_number: index, value: r }) },
```
gregorbg:
```suggestion
      result_attempts: results.map.with_index(1) { |r, index| ResultAttempt.new(attempt_number: index, value: r) },
```

app/models/result.rb
```diff
@@ -15,7 +15,7 @@ class Result < ApplicationRecord
 
   has_many :result_attempts, dependent: :destroy
 
-  after_update_commit :create_or_update_attempts
+  after_update :create_or_update_attempts
 
   def create_or_update_attempts
```
gregorbg, line 45: Here's a wild shot: `self.result_attempts.reset` at the very end of this method should (in theory) nuke the in-memory state of already loaded stuff and make a lot of `reload if result_kind == :result` in the test suite unnecessary

spec/lib/results_validators/individual_results_validator_spec.rb
FinnIckler: I think it's because we never commit the results in tests, so `after_update_commit` doesn't trigger. Don't ask me why `after_update` doesn't work though. I will try your reset approach

app/models/result.rb
FinnIckler: it only fixes one of the two issues

spec/lib/results_validators/individual_results_validator_spec.rb
FinnIckler: this issue was not fixed when introducing the reload

spec/models/result_spec.rb
gregorbg: You marked this as "Resolved" but without any meaningful interaction.
Please explain (a) why my suggestion doesn't make sense/ doesn't work or (b) accept it/ implement it.

FinnIckler: I was sure I pressed the button, but I guess github ignored it

FinnIckler: oh it's because git says  Applying suggestions on deleted lines is currently not supported. 
so I will do this myself

PR #12532, opened 2025-10-02
Make mongodb output to a file in dev

FinnIckler:
Silences the noisy docker compose output of mongodb

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -92,6 +92,10 @@ services:
   wca_payload_db:
     image: mongo:8
     container_name: "payload_db"
+    command:
+      - --quiet
+      - --logpath
+      - /var/log/mongodb/mongod.log
     environment:
       MONGO_INITDB_ROOT_USERNAME: root
       MONGO_INITDB_ROOT_PASSWORD: root


gregorbg (commented):
With the amount of logs that Mongo generates, this file could get very big very fast. Maybe we should clean it up / delete it on boot? That way, you can still trace log messages even after shutdown, but the next time you do some development work, the previous logs will be cleaned up

docker-compose.yml
```diff
@@ -92,6 +92,10 @@ services:
   wca_payload_db:
     image: mongo:8
     container_name: "payload_db"
+    command:
+      - --quiet
+      - --logpath
+      - /var/log/mongodb/mongod.log
```
gregorbg, line 98: Also, typo in the filename ;)
```suggestion
      - /var/log/mongodb/mongodb.log
```

gregorbg (changes requested):

docker-compose.yml
FinnIckler: it actually stands for mongo daemon. The name of the command is `mongod`

FinnIckler:
We can also just set `--logappend=false` which will overwrite the log

gregorbg (approved):

PR #12533, opened 2025-10-02
Use attempts table in results posting

FinnIckler:
I think rails should just populate the results_id field like this?

### DIFF
--- a/lib/competition_results_import.rb
+++ b/lib/competition_results_import.rb
@@ -67,7 +67,7 @@ module CompetitionResultsImport
       end
 
       new_results = Result.insert_all!(result_rows)
-      attempts = new_results.map { it.result_attempts_payload(result_id: it.id) }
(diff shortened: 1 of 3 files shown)


gregorbg (approved):
Code LGTM, please test with an actual submission in local before merging.

FinnIckler:
I tested it and it doesn't work because 
`Result.insert_all!(result_rows)`
is a low level insert call where you can't do associations

lib/competition_results_import.rb
```diff
@@ -69,6 +69,8 @@ def self.merge_inbox_results(competition)
       Result.insert_all!(result_rows)
       competition.inbox_results.destroy_all
     end
+    attempts = competition.reload.results.map { it.result_attempts_attributes(result_id: it.id) }.flatten
```
gregorbg:
```suggestion
    attempts = competition.reload.results.flat_map { it.result_attempts_attributes(result_id: it.id) }
```

PR #12560, opened 2025-10-06
Use SSL to connect in mysql CLI

FinnIckler:
seems like updating the docker image made this necessary

### DIFF
--- a/lib/database_dumper.rb
+++ b/lib/database_dumper.rb
@@ -1311,7 +1311,7 @@ module DatabaseDumper
 
   def self.mysql_cli_creds
     config = ActiveRecord::Base.connection_db_config.configuration_hash
-    "--user=#{config[:username]} --password=#{config[:password] || "''"} --port=#{config[:port]} --host=#{config[:host]} #{"--ssl-ca=/rails/rds-cert.pem" if Rails.env.production?}"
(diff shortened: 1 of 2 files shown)


FinnIckler:
well tests still fail (which make sense because we don't need the certificate in test (or dev), a bit weird of an error though?
` Mysql2::Error: Query was empty`
I would assume it would fail with 'can't find certificate' or something

gregorbg:
> I forgot the space at the end!

thank god we have test coverage for exports :P

lib/database_dumper.rb
```diff
@@ -1311,7 +1311,8 @@ def self.public_results_dump(dump_filename, tsv_folder)
 
   def self.mysql_cli_creds
     config = ActiveRecord::Base.connection_db_config.configuration_hash
-    "--user=#{config[:username]} --password=#{config[:password] || "''"} --port=#{config[:port]} --host=#{config[:host]}"
+    ssl_config = Rails.env.production? ? "--ssl-ca=/rails/rds-cert.pem" : ""
+    "--user=#{config[:username]} --password=#{config[:password] || "''"} --port=#{config[:port]} --host=#{config[:host]} #{ssl_config}"
```
gregorbg:
```suggestion
    "--user=#{config[:username]} --password=#{config[:password] || "''"} --port=#{config[:port]} --host=#{config[:host]} #{"--ssl-ca=/rails/rds-cert.pem" if Rails.env.production?}"
```

gregorbg (changes requested):

PR #12567, opened 2025-10-06
Sync Terraform with manual AWS Changes

FinnIckler:
I enabled the `deployment_circuit_breaker` by hand on saturday. The priority changes are a bit weird, they have been there for months (I used them to make room for next), but my linux terraform doesn't seem to pick them up when running terraform plan/apply.


FinnIckler:
no, it just means when we do outside changes not via terraform, they are not part of the changeset on my linux machine. At least not these ones? I could try out some others

PR #12606, opened 2025-10-16
Use person_id instead of wca_id in AdvancementConditionsValidator

dunkOnIT:
Email thread ref: "Perth Time Trial 2025 Results error"

By mapping by `wca_id` instead of `person_id`, we introduced the possibility where a new user with a null `wca_id` in a subsequent round who _did_ make the advancement criterion (where the advancement criteria is a target result, not a ranking) would cause all newcomers in the previous round to have the validation run for them as well, leading to the following cryptic error: 
> Round 333-2: ,,,, competed but did not meet the attempt result advancement condition (Average of 5 < 20.00 advances to next round). Please make sure the advancement condition reflects what was used during the competition, and remove the results if needed.

Note the presence of ",,,," which suggests mapping of a null value. 

This PR introduces three changes:
* test to reproduce this error
* iterate over `person_id` instead of `wca_id`
* when building error string, use `name` if `person_id` isn't available (and delegate `name` to `inbox_person`)

I'm a little concerned that because `Result` doesn't have `person_id`, this might raise an error - but I'm going to let the tests run and cross that bridge if it burns. 

Also, I've put it on my backlog to refactor the tests, and then the AdvancementConditionsValidator code - both need a bit of TLC to be more readable and maintainable.

### DIFF
--- a/spec/lib/results_validators/advancement_conditions_validator_spec.rb
+++ b/spec/lib/results_validators/advancement_conditions_validator_spec.rb
@@ -108,7 +108,7 @@ RSpec.describe ResultsValidators::AdvancementConditionsValidator do
         expected_errors << RV::ValidationError.new(ACV::COMPETED_NOT_QUALIFIED_ERROR,
                                                    :rounds, competition2.id,
                                                    round_id: "222-f",
-                                                   ids: fake_person.wca_id,
+                                                   ids: "#{fake_person.name} (#{fake_person.wca_id})",
                                                    condition: first_round.advancement_condition.to_s(first_round))
       end
       expected_errors << RV::ValidationError.new(ACV::ROUND_9P1_ERROR,


lib/results_validators/advancement_conditions_validator.rb
```diff
@@ -94,11 +94,11 @@ def run_validation(validator_data)
 
                 # Check that no one proceeded if they shouldn't have
                 if condition.instance_of? AdvancementConditions::AttemptResultCondition
-                  current_persons = results_by_round_type_id[round_type_id].map(&:wca_id)
+                  current_persons = results_by_round_type_id[round_type_id].map(&:person_id)
                   people_over_condition = previous_results.filter do |r|
                     sort_by_column = r.format.sort_by == "single" ? :best : :average
                     current_persons.include?(r.wca_id) && r.send(sort_by_column) > condition.attempt_result
```
gregorbg: If `current_persons` now consists of lots of `person_id`s, then you should not look up the `wca_id` here. How the heck did the tests pass on this?

lib/results_validators/advancement_conditions_validator.rb
```diff
@@ -94,11 +94,11 @@ def run_validation(validator_data)
...
-                  end.map(&:wca_id)
+                  end.map { it.wca_id.presence || it.name }
```
gregorbg: This makes the output format rather inconsistent. Maybe something like "Always use the name and additionally append the WCA ID in brackets if any" would work better?

gregorbg (commented):
This is a very reasonable change, and it makes me wonder why it didn't explode any earlier.

For reference, both `results` and `inbox_results` tables actually use a `person_id` column on the database level (which _is_ also available in Rails). The accessor `wca_id` is a manually created convenience method:
- In the case of `results`, it simply is an alias to `person_id`, which will *always* contain a WCA ID because `results` only holds those Results which are completely handled and posted by WRT, so WCA IDs for newcomers will already have been generated at this point in time.
- In the case of `inbox_results`, it delegates `wca_id` to the corresponding `inbox_person`, which may or may not be `null`. That's exactly where the error came from.

I have a few minor concerns, but the overall intention of the PR is correct and good to go.

lib/results_validators/advancement_conditions_validator.rb
dunkOnIT: Oh wow, good catch. It passes because `current_persons.include(r.wca_id)` is always false, and my new test only checks that this error _doesn't_ get raised. I thought the other tests covered this, but apparently not. This is why you _always_ include a failure case when testing, lol

lib/results_validators/advancement_conditions_validator.rb
```diff
@@ -94,11 +94,13 @@ def run_validation(validator_data)
...
                 if condition.instance_of? AdvancementConditions::AttemptResultCondition
-                  current_persons = results_by_round_type_id[round_type_id].map(&:wca_id)
+                  current_persons = results_by_round_type_id[round_type_id].map(&:person_id)
                   people_over_condition = previous_results.filter do |r|
                     sort_by_column = r.format.sort_by == "single" ? :best : :average
-                    current_persons.include?(r.wca_id) && r.send(sort_by_column) > condition.attempt_result
-                  end.map(&:wca_id)
+                    current_persons.include?(r.person_id) && r.send(sort_by_column) > condition.attempt_result
+                  end.map do
+                    "#{it.name}#{" (#{it.wca_id})" if it.wca_id.present? }"
```
dunkOnIT: This could have been: 
- a 2-liner (determining the suffix separately from concatenating the final string), or
- a single-line block

But this approach seemed the best compromise between compactness and readability

lib/results_validators/advancement_conditions_validator.rb
```diff
@@ -94,11 +94,13 @@ def run_validation(validator_data)
...
                 if condition.instance_of? AdvancementConditions::AttemptResultCondition
-                  current_persons = results_by_round_type_id[round_type_id].map(&:wca_id)
+                  current_persons = results_by_round_type_id[round_type_id].map(&:person_id)
                   people_over_condition = previous_results.filter do |r|
                     sort_by_column = r.format.sort_by == "single" ? :best : :average
-                    current_persons.include?(r.wca_id) && r.send(sort_by_column) > condition.attempt_result
-                  end.map(&:wca_id)
+                    current_persons.include?(r.person_id) && r.send(sort_by_column) > condition.attempt_result
+                  end.map do
+                    "#{it.name}#{" (#{it.wca_id})" if it.wca_id.present?}"
```
dunkOnIT: This could have been:
- a 2-liner (determining the suffix separately from concatenating the final string), or
- a single-line block

But this approach seemed the best compromise between compactness and readability

lib/results_validators/advancement_conditions_validator.rb
dunkOnIT: Fixed, and added a test for it

lib/results_validators/advancement_conditions_validator.rb
dunkOnIT: Done, and added a test that checks both output formats

lib/results_validators/advancement_conditions_validator.rb
```diff
@@ -94,11 +94,13 @@ def run_validation(validator_data)
...
                 if condition.instance_of? AdvancementConditions::AttemptResultCondition
-                  current_persons = results_by_round_type_id[round_type_id].map(&:wca_id)
+                  current_persons = results_by_round_type_id[round_type_id].map(&:person_id)
                   people_over_condition = previous_results.filter do |r|
                     sort_by_column = r.format.sort_by == "single" ? :best : :average
-                    current_persons.include?(r.wca_id) && r.send(sort_by_column) > condition.attempt_result
-                  end.map(&:wca_id)
+                    current_persons.include?(r.person_id) && r.send(sort_by_column) > condition.attempt_result
+                  end.map do |competitor|
+                    "#{competitor.name}#{" (#{competitor.wca_id})" if competitor.wca_id.present?}"
```
dunkOnIT, line 102: This could have been:
- a 2-liner (determining the suffix separately from concatenating the final string), or
- a single-line block

But this approach seemed the best compromise between compactness and readability

lib/results_validators/advancement_conditions_validator.rb
```diff
@@ -94,11 +94,13 @@ def run_validation(validator_data)
...
                 # Check that no one proceeded if they shouldn't have
                 if condition.instance_of? AdvancementConditions::AttemptResultCondition
-                  current_persons = results_by_round_type_id[round_type_id].map(&:wca_id)
+                  current_persons = results_by_round_type_id[round_type_id].map(&:person_id)
                   people_over_condition = previous_results.filter do |r|
                     sort_by_column = r.format.sort_by == "single" ? :best : :average
-                    current_persons.include?(r.wca_id) && r.send(sort_by_column) > condition.attempt_result
-                  end.map(&:wca_id)
+                    current_persons.include?(r.person_id) && r.send(sort_by_column) > condition.attempt_result
+                  end.map do |competitor|
```
dunkOnIT, line 101: `person` is the intuitive name, but using `competitor` instead to avoid mislabelling, as the object can be (as I understand) either `person` (in case of Result) or `inbox_person` (in case of InboxResult)

PR #12609, opened 2025-10-17
Install TTNormsPro proprietary font on production machines

gregorbg:
This checks in a securely encrypted copy of our styleguide's font files. We cannot use the originals because of licensing restrictions :/

The ciphertext has been generated by the vault `transit` engine. It is also used to decrypt the font files on the production machine in the Docker entrypoint. Note that they are **not** being decrypted during build time, because the containers are build on GitHub Actions which does not have Vault access.


gregorbg:
Thank you!

PR #12648, opened 2025-10-24
Support Multiple National Championship Podium Assignments

dunkOnIT:
Email thread: "Re: Results for FMC Iberian Championship 2025"

A competition serving as a national championship for multiple countries is supported in the code and UI. However, in the national championship podium determination code, the results were not being split up according to country - meaning that podium places were not being correctly assigned, or assigned at all past the third result in the `final_results` filter. 

This PR fixes this issue by grouping by both country_id AND event_id, and adds tests to validate its effectiveness.

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -185,7 +185,7 @@ class Person < ApplicationRecord
             .joins(:event)
             .order("events.rank, pos")
             .includes(:format, :competition)
-            .group_by { { event_id: it.event_id, country_id: it.country_id} }
+            .group_by { { event_id: it.event_id, country_id: it.country_id } }
             .each_value do |final_results|
               previous_old_pos = nil
               previous_new_pos = nil


app/models/person.rb
```diff
@@ -185,7 +185,7 @@ def championship_podiums_with_condition
             .joins(:event)
             .order("events.rank, pos")
             .includes(:format, :competition)
-            .group_by(&:event_id)
+            .group_by { { event_id: it.event_id, country_id: it.country_id } }
```
gregorbg, line 188: ...maaaaaybe possible on SQL level????

```suggestion
            .group(:event_id, :country_id)
```

dunkOnIT: Confirmed as not possible due to this error: 
```bash
 1) Person#championship_podiums ignores DNF results on the podium
     Failure/Error:
       yield(competition.results)
         .final
         .succeeded
         .joins(:event)
         .order("events.rank, pos")
         .includes(:format, :competition)
         .group(:event_id, :country_id)
         .each do |final_results|
           previous_old_pos = nil
           previous_new_pos = nil
     
     ActiveRecord::StatementInvalid:
       Mysql2::Error: Expression #1 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'wca_test.results.id' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by
```

PR #12654, opened 2025-10-26
Show Competitor Tab if registration date is in the future and there are accepted non-organizer/delegate registrations

dunkOnIT:
WCAT requested a way to have the Competitor tab still show up even if the registration open date is in the future, provided registration has opened at least once in the past. Their use case is when organizers want to run a second registration period due to spots still being available - in this case, WCAT will adjust the registration_open date forward, causing the competitor tab to disappear in our current logic. 

The long-term solution for this is support for multiple registration periods. The short-term solution is either: 
- add a `show_competitor_tab` field to the Competition model that WCAT can toggle, or
- display the competitor tab is there are any accepted registrations for non-organizers/delegates - thanks to Cailyn for this very practical suggestion! (presumably the initial registration period will not be so abysmal that literally no one besides organizers/delegates register - and if that does happen, it probably doesn't matter that the Competitor tab isn't visible before the next registration period anyway)

**Note:** This PR also removes a clause that wouldn't show the competitor tab even if registration was open. I think that clause is more likely to lead to users being confused ("I can't see the competitor tab, is something wrong?") as opposed to the pretty obvious case of the list just being empty.

### DIFF
--- a/spec/models/competition_spec.rb
+++ b/spec/models/competition_spec.rb
@@ -1998,28 +1998,26 @@ RSpec.describe Competition do
       expect(competition.can_show_competitors_page?).to be(true)
     end
 
-    context 'before registration opens' do
-      before do
-        competition.registration_open = 1.day.from_now
-      end
+    context 'before registration opens', :zxc do
+      let(:not_open) { create(:competition, :registration_not_opened, :with_organizer, :with_delegate) }
 
       it 'is false with no accepted registrations' do
-        expect(competition.can_show_competitors_page?).to be(false)
+        expect(not_open.can_show_competitors_page?).to be(false)
       end
 
       it 'is false if the only accepted registrations are for organizers/delegates' do
-        delegate = competition.delegates.first
-        organizer = competition.organizers.first
+        delegate = not_open.delegates.first
+        organizer = not_open.organizers.first
 
-        create(:registration, :accepted, competition: competition, user: delegate)
-        create(:registration, :accepted, competition: competition, user: organizer)
+        create(:registration, :accepted, competition: not_open, user: delegate)
+        create(:registration, :accepted, competition: not_open, user: organizer)
 
-        expect(competition.can_show_competitors_page?).to be(false)
+        expect(not_open.can_show_competitors_page?).to be(false)
       end
 
-      it 'unless there are accepted non-delegate/organizer registrations' do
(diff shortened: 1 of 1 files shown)


spec/models/competition_spec.rb
```diff
@@ -1977,4 +1977,50 @@ def change_and_check_activities(new_start_date, new_end_date)
...
+    it 'is true after registration has closed irrespective of registrations' do
+      competition.registration_close = 1.day.ago
+      expect(competition.registrations.competing_status_accepted.competing.count).to be(0)
+      expect(competition.can_show_competitors_page?).to be(true)
+    end
+
+    context 'before registration opens' do
+      before do
+        competition.registration_open = 1.day.from_now
+      end
```
gregorbg: You can just `let` a new `:competition` here, with all the factory properties and scopes that you need.
The test cases under this `context` will automatically detect and use the `competition` variable that is "closest" to them -- that's one of the strengths of using `let` 🙃

spec/models/competition_spec.rb
```diff
@@ -1977,4 +1977,50 @@ def change_and_check_activities(new_start_date, new_end_date)
...
+        delegate = competition.delegates.first
+        organizer = competition.organizers.first
+
+        create(:registration, :accepted, competition: competition, user: delegate)
+        create(:registration, :accepted, competition: competition, user: organizer)
+
+        expect(competition.can_show_competitors_page?).to be(false)
+      end
+
+      it 'unless there are accepted non-delegate/organizer registrations' do
```
gregorbg: The title is misleading here

spec/models/competition_spec.rb
dunkOnIT: Done!

spec/models/competition_spec.rb
dunkOnIT: Changed

kr-matthews:
Would it be good to have a warning somewhere on the edit competition form that removing an organizer or delegate will reveal the registration list if they've registered?

dunkOnIT:
I think it's a good thought regarding edge cases, but I think it points at an underlying issue: That someone removed as an organizer/delegate shouldn't be able to pre-register in the first place - so if they have an existing registration, it should get cancelled. I'll reach out to WCAT and ask what behaviour they want in this case.

gregorbg (approved):

PR #12685, opened 2025-10-27
Prevent 500 error when a competition with a blank `start_date` is saved

dunkOnIT:
We were comparing registration dates to `start_date` without a presence check, which would cause a 500 error from the server when a comp without a `start_date` was saved. This PR fixes it, although we also need a frontend fix to show an error in the case that the server returns an error - as it was not displaying the error.

### DIFF
--- a/app/models/competition.rb
+++ b/app/models/competition.rb
@@ -1202,9 +1202,8 @@ class Competition < ApplicationRecord
     errors.add(:end_date, I18n.t('competitions.errors.span_too_many_days', max_days: MAX_SPAN_DAYS)) if number_of_days > MAX_SPAN_DAYS
   end
 
-validate :registration_dates_must_be_valid, if: :start_date?
-private def registration_dates_must_be_valid
-
+  validate :registration_dates_must_be_valid, if: :start_date?
+  private def registration_dates_must_be_valid
     errors.add(:refund_policy_limit_date, I18n.t('competitions.errors.refund_date_after_start')) if refund_policy_limit_date? && refund_policy_limit_date > start_date
 
     return unless registration_period_required? && [registration_open, registration_close].all?(&:present?)


app/models/competition.rb
```diff
@@ -1206,10 +1206,10 @@ def registration_edits_currently_permitted?
   private def registration_dates_must_be_valid
     errors.add(:refund_policy_limit_date, I18n.t('competitions.errors.refund_date_after_start')) if refund_policy_limit_date? && refund_policy_limit_date > start_date
```
gregorbg, line 1207: Here's already an implicit requirement `start_date`

app/models/competition.rb
```diff
@@ -1206,10 +1206,10 @@ def registration_edits_currently_permitted?
...
 
-    if registration_period_required? && registration_open.present? && registration_close.present? &&
-       (registration_open >= start_date || registration_close >= start_date)
-      errors.add(:registration_close, I18n.t('competitions.errors.registration_period_after_start'))
-    end
+    return unless registration_period_required? && [registration_open, registration_close, start_date].all?(&:present?)
```
gregorbg: If we're certain that `start_date` is always needed anyways, just introduce an independent `validates :start_date, presence: true`.

spec/models/competition_spec.rb
```diff
@@ -1611,6 +1611,12 @@ def change_and_check_activities(new_start_date, new_end_date)
   context "new competition is invalid when" do
     let!(:new_competition) { build(:competition, :with_delegate, :future, :visible, :with_valid_schedule) }
 
+    it 'there is no start date' do
+      new_competition.start_date = nil
+      expect { new_competition.valid? }.not_to raise_error
```
gregorbg: This is redundant because the next assumption below would make the test fail if the validation crashes

app/models/competition.rb
dunkOnIT: As discussed on the call, we already have this in `dates_must_be_valid`: 
```ruby
if start_date.nil? && end_date.nil?
      if confirmed_or_visible?
        errors.add(:start_date, I18n.t('common.errors.invalid'))
        errors.add(:end_date, I18n.t('common.errors.invalid'))
      end
      return
    end

    return errors.add(:start_date, I18n.t('common.errors.invalid')) if start_date.blank?
    return errors.add(:end_date, I18n.t('common.errors.invalid')) if end_date.blank?
```

The problem is that failed validations don't stop the validation process, so a `nil` `start_date` still triggers an error when it reaches the next validation

spec/models/competition_spec.rb
dunkOnIT: Fixed

app/models/competition.rb
dunkOnIT: Good point - added a guard clause above it

app/models/competition.rb
```diff
@@ -1204,12 +1204,14 @@ def registration_edits_currently_permitted?
 
   validate :registration_dates_must_be_valid
   private def registration_dates_must_be_valid
+    return unless start_date.present?
```
gregorbg:
```suggestion
validate :registration_dates_must_be_valid, if: :start_date?
private def registration_dates_must_be_valid
```

PR #12690, opened 2025-10-28
Add bo5 as the allowed format for 333bld from 2026 onwards

FinnIckler:
Let's see how many tests fail

### DIFF
--- a/app/models/competition.rb
+++ b/app/models/competition.rb
@@ -57,7 +57,7 @@ class Competition < ApplicationRecord
            allow_nil: true,
            with_model_currency: :currency_code
 
-  validate :start_date, :cant_change_across_regulation_boundaries, if: -> { start_date_was.present? && events.pluck(:id).include?('333bf') }
+  validate :start_date, :cant_change_across_regulation_boundaries, if: -> { start_date_was.present? && event_ids.include?('333bf') }
   private def cant_change_across_regulation_boundaries
     errors.add(:start_date, "You can't change the start date across Regulation boundaries.") if
       (start_date_was.year == 2025 && start_date.year == 2026) ||


app/models/round.rb
```diff
@@ -58,8 +58,18 @@ def format
     self.number = 0
   end
 
+  # Competitions before 2026 have to use Mo3 for 333bf, but after 2026 they need to use Ao5
+  private def uses_wrong_333bf_format?
```
dunkOnIT: We only want to enforce this once the regulations are actually announced - that leaves us 3 options: 
1. Don't merge this change until 1st Dec
2. Add a feature flag that we toggle with env variables or server_settings or similar
3. Add a date-based guard clause that returns unless `Date.today > new_regulations_announcement_date`, where announcement date is 2025-12-01

app/models/round.rb
```diff
@@ -58,8 +58,18 @@ def format
...
+    new_regulations_start_date = Date.new(2026, 1, 1)
+    return false if event.id != "333bf"
```
dunkOnIT: This guard clause can be the first line in the function, can't it?

dunkOnIT:
Would it be possible to make the following changes/added features too? 

## Frontend
- [ ] Don't show BO5 as an option - _even for 2026 comps_ - until Dec 1st (reg announcement)
  - other option: we don't merge this until Dec 1st - but I think it would be nice to have it already in the codebase
- [ ] Once regs are announcement, only show BO5/MO3 based on when the comp takes place - eg, a BO5 2026-starting comp can't change to 2025, or vice-versa. 
  - [ ] ALTERNATIVE: If it's not possible to limit the options displayed, it would be great to add some kind of UI hint (like a question mark that the user can hover over) explaining the 2026-BO5, 2025-MO3 limitations (as they will get an error from the backend if they choose the wrong format)

## Backend
- [ ] Add a validation preventing the user from changing `start_date` from 2025 to 2026 or vice-versa if 3bld is an event 
  - The error message returned by this should tell the user to contact WCAT[1], if we display an error message to the user (as opposed to a generic "An error occured")
  - This is because our validations will prevent a user from changing the start_date OR the event format, requiring manual intervention from admins. This is preferable to organizers being able to silently run an illegal format

[1] WCAT will then contact us - this is an unlikely-enough case that I'd propose we just handle the changes manually, although if it's easy to build permissions into the validation (I don't _think_ it is given that we don't have access to current_user in `models/competition.rb`) then we can do that

app/models/round.rb
FinnIckler: it can, maybe I should make the new_regulations_start_date a constant

FinnIckler:
I just noticed that the formats.json is not automatically loaded into the db on restart, so we need to run the rake task too.

gregorbg:
It is definitely being loaded in Prod

FinnIckler:
where in the code is that done?

gregorbg:
https://github.com/thewca/worldcubeassociation.org/blob/main/lib%2Ftasks%2Fstatic_data.rake#L27-L31

FinnIckler:
I just tested it and the current state works for adding the event and the validations work too. For limiting what is displayed this would introduce a ton of temporary code which I kinda want to avoid...

app/models/round.rb
FinnIckler: the issue with 2 and 3 is that the bo5 format still exists and could be selected

dunkOnIT:
Forgot to comment, sorry! This is QA'd and ready for review.

app/models/competition.rb
```diff
@@ -57,6 +57,15 @@ class Competition < ApplicationRecord
            allow_nil: true,
            with_model_currency: :currency_code
 
+  validate :start_date, :cant_change_across_regulation_boundaries, if: lambda { |_lambda|
```
gregorbg:
```suggestion
  validate :start_date, :cant_change_across_regulation_boundaries, if: -> { 
```

app/models/competition.rb
```diff
@@ -57,6 +57,15 @@ class Competition < ApplicationRecord
...
+    start_date_changed? && start_date_was.present? # start_date_was will be nil when `start_date` changes from `nil` to `true`
```
gregorbg: Food for though: I think `start_date_was.present?` implies that `start_date_changed?`. In other words, how can `start_date_changed?` not be true when the value `start_date_was` is `present`?

This suspicion needs to be tried out and validated though. If it turns out to be correct, then checking `start_date_was.present?` is all that is needed.

app/models/round.rb
```diff
@@ -58,8 +58,18 @@ def format
     self.number = 0
   end
 
+  # Competitions before 2026 have to use Mo3 for 333bf, but after 2026 they need to use Ao5
+  NEW_REGULATIONS_START_DATE = Date.new(2026, 1, 1)
```
gregorbg: "New Regulations" is way too generic as a name for this field

app/models/round.rb
```diff
@@ -58,8 +58,18 @@ def format
...
+  private def uses_wrong_333bf_format?
+    return false if event.id != "333bf"
+    return format_id != "5" if competition.start_date >= NEW_REGULATIONS_START_DATE
+
+    format_id != "3"
+  end
```
gregorbg: These are a lot of cascading checks and early returns. May I suggest something along the lines of:

```suggestion
  private def uses_wrong_333bf_format?
    expected_format_id = competition.start_date >= NEW_REGULATIONS_START_DATE ? "5" : "3"

    event.id == "333bf" && format_id != expected_format_id
  end
```

gregorbg: If you are concerned about the database load of `competition.start_date` then I might be inclined to agree to a middle ground where you keep `return if event.id != "333bf"` but still use the `expected_format_id` approach.

dunkOnIT: Updated with the guard clause and `expected_format_id` approach per your compromise suggestion

app/models/round.rb
dunkOnIT: ideal would `2026_REGULATIONS...` but we can't start with a numeral so I did `REGULATIONS_2026...`

app/models/competition.rb
dunkOnIT: Tested with your suggested change and it worked! Also asked ChatGPT for a second opinion, and it also thinks that `start_date_changed?` is redundant in this check

app/models/competition.rb
dunkOnIT: Fixed

gregorbg (approved):

dunkOnIT:
@thewca-bot deploy staging

app/models/concerns/resultable.rb
```diff
@@ -118,7 +118,9 @@ def should_compute_average?
     #    - All events that allow "mean of 3" no longer allow "best of 3".
     #  - May 1, 2019
     #    - 444bf and 555bf mean are officially recognized
-    format_id == "a" || format_id == "m" || (format_id == "3" && %(333ft 333fm 333bf 444bf 555bf).include?(event_id))
+    #  - January 1, 2026
+    #    - Bo5/Ao5 becomes the format for 333bld
+    format_id == "a" || format_id == "m" || format_id == "5" || (format_id == "3" && %(333ft 333fm 333bf 444bf 555bf).include?(event_id))
```
gregorbg: This needs to check not only `format_id == "5"` but rather the combination of `5 && 3bld`. See how `3` does it.
No need to check the date in my opinion, we make sure with other checks in other places that results with `5` can only exist after 2026

app/models/competition.rb
```diff
@@ -57,6 +57,13 @@ class Competition < ApplicationRecord
            allow_nil: true,
            with_model_currency: :currency_code
 
+  validate :start_date, :cant_change_across_regulation_boundaries, if: -> { start_date_was.present? && events.pluck(:id).include?('333bf') }
```
gregorbg: Can `event_ids` instead of `events.pluck(:id)` work?

dunkOnIT: Looks like no? Tried in the Rails console: 
```bash
wca-on-rails(dev)> comp.events.pluck(:id)
  Event Pluck (2.1ms)  SELECT `events`.`id` FROM `events` INNER JOIN `competition_events` ON `events`.`id` = `competition_events`.`event_id` WHERE `competition_events`.`competition_id` = 'WC2025' ORDER BY `events`.`rank` ASC, `competition_events`.`event_id` ASC
=> ["333", "222", "444", "555", "666", "777", "333bf", "333fm", "333oh", "clock", "minx", "pyram", "skewb", "sq1", "444bf", "555bf", "333mbf"]
wca-on-rails(dev)> comp.events.event_ids
/usr/local/bundle/3.4.2/ruby/3.4.0/gems/irb-1.15.2/lib/irb.rb:406:in 'full_message': undefined method 'event_ids' for an instance of ActiveRecord::Associations::CollectionProxy (NoMethodError)
```

gregorbg: You misunderstood -- in your example it should be `comp.event_ids`

gregorbg: You misunderstood -- in your example it should be `comp.event_ids`

dunkOnIT: Aha yep that works - changed

PR #12695, opened 2025-10-29
Use result_attempts in attempts method

FinnIckler:
let's see how much will blow up!


FinnIckler:
huh it doesn't hahaha

PR #12696, opened 2025-10-30
Stop writing to value1...5 columns

FinnIckler:
I will have to merge in https://github.com/thewca/worldcubeassociation.org/pull/12531 so that not a bunch of unrelated tests fail. 

Tasks:
- [x] Rankings by person
- [x] Records/Rankings need access to attempts
- [x] Editing Results
- [x] Change Factory
- [x] CAD tables?

I´m glad we got rid of the 5 subquery query. Takes about 45 seconds on my machine for 3x3. I looked into optimizations, but  this is as far as I got after about an hour of work. These will be procomputed for the big filters anyway.

### DIFF
--- a/lib/results_validators/scrambles_validator.rb
+++ b/lib/results_validators/scrambles_validator.rb
@@ -19,7 +19,7 @@ module ResultsValidators
 
     def competition_associations
       {
-        scrambles: [],
+        scrambles: [:round],
       }
     end
 


FinnIckler:
I think I have found a decent solution for showing the result attempts in rankings/records. Basically instead of getting the whole result in the `@query ` we just get the ids. Which we can then use to get Active Record Result objects to serialize.

app/models/person.rb
```diff
@@ -247,7 +248,7 @@ def records
   end
 
   def completed_solves_count
-    results.pluck("value1, value2, value3, value4, value5").flatten.count(&:positive?)
+    result_attempts.where("value > 0").count
```
gregorbg: The `where value > 0` should become a `scope` on `ResultAttempt` (perhaps called "completed", in the spirit of this method name?) and it can also use the `where(value: 0...)` notation.

gregorbg: If memory serves, Rails is a bit picky with open-ended ranges so depending on the implementation you might have to do something like `where.not(value: ..0)` or similar to "hack" a `>` operator in SQL

db/migrate/20251029153518_remove_value_n_indices.rb
gregorbg, line 1: Is this really already be supposed to be part of this same PR? I thought we agreed on a "soft removal" first

db/migrate/20251030105922_add_index_to_result_attempts_value.rb
gregorbg, line 1: This may be useful to have already in the other PR that introduces/uses the `result_attempts` productively

db/schema.rb
```diff
@@ -1127,12 +1124,13 @@
...
-    t.index ["event_id", "value3"], name: "index_Results_on_eventId_and_value3"
-    t.index ["event_id", "value4"], name: "index_Results_on_eventId_and_value4"
-    t.index ["event_id", "value5"], name: "index_Results_on_eventId_and_value5"
+    t.index ["event_id", "id"], name: "idx_results_event_id_id"
     t.index ["event_id"], name: "Results_fk_event"
+    t.index ["event_id"], name: "index_Results_on_eventId_and_value1"
+    t.index ["event_id"], name: "index_Results_on_eventId_and_value2"
+    t.index ["event_id"], name: "index_Results_on_eventId_and_value3"
+    t.index ["event_id"], name: "index_Results_on_eventId_and_value4"
+    t.index ["event_id"], name: "index_Results_on_eventId_and_value5"
```
gregorbg: These indices should also be dropped when `value1..5` are dropped

db/migrate/20251029153518_remove_value_n_indices.rb
FinnIckler: yes, this will be removed after all the tests pass so we know that there is no hidden usages of value1...5 anywhere

db/migrate/20251030105922_add_index_to_result_attempts_value.rb
FinnIckler: it's only needed in this PR

db/migrate/20251029153518_remove_value_n_indices.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class RemoveValueNIndices < ActiveRecord::Migration[8.1]
+  def change
+    change_table :results, bulk: true do |t|
+      t.remove_index %i[event_id value1]
```
FinnIckler, line 6: I assume this will not affect any speed on the records and rankings page computation?

gregorbg: Correct. The indexes were there for our old `UNION ALL` folding, which had already been deleted from the code in a previous PR.

This migration is really just "housekeeping"

PR #12698, opened 2025-10-31
Show championship podiums for non-333 events

dunkOnIT:
I misunderstood how the `group_by` block worked, and didn't add test coverage for non-333 events, so this went undetected.

The fix is to return a hash-key in both the true/false case of the ternary that is returning the group_by parameter.

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -189,7 +189,6 @@ class Person < ApplicationRecord
               # Group by country_id in the case of a national championship to cover the case of a comp acting as multiple national championships
               championship_type == :national ? { event_id: results.event_id, country_id: results.country_id } : { event_id: results.event_id }
             end
-            # .tap { |grouped| byebug }
             .each_value do |final_results|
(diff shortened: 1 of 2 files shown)


app/models/person.rb
```diff
@@ -187,8 +187,9 @@ def championship_podiums_with_condition(championship_type)
             .includes(:format, :competition)
             .group_by do |results|
               # Group by country_id in the case of a national championship to cover the case of a comp acting as multiple national championships
-              championship_type == :national ? { event_id: results.event_id, country_id: results.country_id } : :event_id
+              championship_type == :national ? { event_id: results.event_id, country_id: results.country_id } : { event_id: results.event_id }
             end
+            # .tap { |grouped| byebug }
```
FinnIckler: hello debug statement

dunkOnIT: :blush:

FinnIckler: good bye debug statement

PR #12727, opened 2025-11-03
Auto Accept: Don't validate for connected account when comp is over

dunkOnIT:
We auto-disconnect payment integrations after 30 days - this means that currently it isn't possible to edit competition details after the CPI gets disconnected, impeding duties of teams like WRT and WCAT. 

This PR skips the validation is the end_date of the comp has passed.

### DIFF
--- a/spec/models/competition_spec.rb
+++ b/spec/models/competition_spec.rb
@@ -1646,7 +1646,7 @@ RSpec.describe Competition do
         end
 
         it 'integrated payment not enabled when competition is confirmed' do
-          confirmed_comp = build(:competition, :confirmed, :bulk_auto_accept)
+          confirmed_comp = build(:competition, :confirmed, :bulk_auto_accept, :future)
           expect(confirmed_comp).not_to be_valid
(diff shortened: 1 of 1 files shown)


spec/models/competition_spec.rb
```diff
@@ -1635,6 +1635,7 @@ def change_and_check_activities(new_start_date, new_end_date)
   describe "validate auto accept fields" do
     let(:competition) { create(:competition, use_wca_registration: true) }
 
+
```
gregorbg: Double newline

spec/models/competition_spec.rb
```diff
@@ -1651,6 +1652,7 @@ def change_and_check_activities(new_start_date, new_end_date)
           expect(confirmed_comp.errors[:auto_accept_preference]).to include("You must enable a payment integration (eg, Stripe) in order to use auto-accept")
         end
 
+
```
gregorbg: Double newline

dunkOnIT: Done

spec/models/competition_spec.rb
dunkOnIT: Done

PR #12730, opened 2025-11-03
Add created_at and updated_at to attempts backfilling

FinnIckler:
I will use this to run backfilling on staging so don't merge before I deployed to staging


FinnIckler:
@thewca-bot deploy staging

PR #12731, opened 2025-11-04
Move api v1 to fetchWithAuthenticityToken

FinnIckler:
Now that we are using oidc in Next, we don't need to run another authentication method.

Steps Taken:
- have the V1 Controller inherit from `ApplicationController` so we have access to `current_user` and doorkeeper
- rename `validate_jwt_token` to require_user
- Move all the calls of `fetchWithJWTToken` to `fetchWithAuthenticityToken`

Another side effect of this, is that ROs that use 3rd Party systems like New Zealand, Taiwan, Malaysia, etc. could now just send requests to this API using their already existing flows (if they link WCA accounts already that is).

Maybe we should reach out?


gregorbg (commented):
I think you can now also safely remove the `jwt_token` property from tjhe relevant spec's factories, as well as the helper that generates an on-the-fly JWT during testing

PR #12742, opened 2025-11-07
Dual rounds model

FinnIckler:
Some Observations:
1. Advancement is just calculated on WCA Live level, for the wca_live integration we do have `recompute_advancing`, which needs to be changed, but otherwise the only other way it is checked is the `AdvancementConditionsValidator`
2. Do we introduce a new AdvancementConditionId for this?
3. If no, how to we allow the creation of linked rounds through wcif patches?
4. Should I include the changes to local_pos and global_pos like we discussed in this PR?


FinnIckler:
For the validator, it currently does `results_by_round_type_id = results_for_event.group_by(&:round_type_id)` which can probably be easiest replace by something like `competition.rounds_with_linked.index_by(:round_type_id).transform_values(&:results)` where rounds_with_linked would be rounds or linked_rounds together.

app/models/linked_rounds.rb
```diff
@@ -0,0 +1,27 @@
...
+    errors.add(:rounds, "must all belong to the same event") if event_ids.size > 1
+
+    return unless round_numbers.size > 1
+
+    errors.add(:rounds, "must all belong to the same round number")
+  end
+
+  def results
+    rounds.flat_map(&:results)
+  end
```
gregorbg: This can be replaced more efficiently by a `has_many :results, through: :rounds`

app/models/linked_rounds.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+class LinkedRounds < ApplicationRecord
```
gregorbg: This class (along with its filename) should be singular `LinkedRound` by convention

gregorbg: Observe also how our other models are called `Round` and `Result`, instead of `Rounds` or `Results`. Only the table on SQL level is created with plural terminology.

app/models/linked_rounds.rb
```diff
@@ -0,0 +1,27 @@
...
+  has_many :rounds
+
+  validate :all_round_consistency
+
+  private def all_round_consistency
+    return if rounds.empty?
+
+    competition_ids = rounds.map(&:competition_id).uniq
+    event_ids       = rounds.map(&:event_id).uniq
+    round_numbers   = rounds.map(&:number).uniq
```
gregorbg: This is hyper-nitpick, but there's actually a more optimal way to do this. I will use `competition_ids` as an example, but the same should apply to the others:
- You add a `has_many :competitions, through: :rounds`.
- You specify it to only list _distinct_ competitions by doing `has_many :competitions, -> { distinct }, through: :rounds`
- This gives you the magic helper `competition_ids`.
- You can then declare a Rails-native validation `validates :competition_ids, length: { is: 1, message: "must all belong to the same competition" }

app/models/linked_rounds.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    errors.add(:rounds, "must all belong to the same competition") if competition_ids.size > 1
+
+    errors.add(:rounds, "must all belong to the same event") if event_ids.size > 1
+
+    return unless round_numbers.size > 1
+
+    errors.add(:rounds, "must all belong to the same round number")
```
gregorbg: I don't understand this validation. The idea is that you explicitly link _different_ rounds together, so for example having one round be `number=1` and the other "linked" round be `number=2` is the expected and desired approach for this.
Why are you erroring out if there is more than one distinct round number?

db/migrate/20251107120000_add_linked_rounds_model.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class AddLinkedRoundsModel < ActiveRecord::Migration[7.2]
+  def change
+    create_table :linked_rounds do |t|
+      t.string :wcif_id
+      t.timestamps
+    end
+
+    add_reference :rounds, :linked_rounds, index: true
```
gregorbg: This should also be singular `linked_round` methinks. You can also try `foreign_key: true`

gregorbg: According to the documentation at https://apidock.com/rails/ActiveRecord/ConnectionAdapters/SchemaStatements/add_reference, the first argument is a table name (and table names are plural) but the second one is the reference name and that should be singular because it is a simple `belongs_to`.

gregorbg (changes requested):

db/migrate/20251107120000_add_linked_rounds_model.rb
FinnIckler: interesting that this just worked then

app/models/linked_rounds.rb
FinnIckler: Ah, I thought they would have the same number (like they both be round 1)

FinnIckler: But I guess it makes sense if they aren't because of the 4 round limit

app/models/linked_rounds.rb
FinnIckler: true, I was basing this all of competition_series but I guess that's an exception because the plural and singular are the same...

app/models/linked_round.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class LinkedRound < ApplicationRecord
+  has_many :rounds
+  has_many :results, through: :rounds
+  has_many :competitions, through: :rounds
+  has_many :events, through: :rounds
+
+  validates :competition_ids, length: { is: 1, message: "must all belong to the same competition" }
+  validates :event_ids, length: { is: 1, message: "must all belong to the same event" }
```
gregorbg: You're missing the `-> { distinct }` scope in the `has_many` relations above

app/models/linked_round.rb
```diff
@@ -0,0 +1,13 @@
...
+
+  validate :all_round_consistency
```
gregorbg: This is now orphaned

app/models/linked_round.rb
gregorbg: On second thought, I think you can actually achieve the same by just piping through `competition_event`. This is a model that encapsulates "a WCA event happening at a competition", and it nicely checks both properties at once. So perhaps `has_many :competition_events, -> { distinct }, through: :rounds` should be enough to check.

gregorbg (changes requested):

app/models/linked_round.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class LinkedRound < ApplicationRecord
+  has_many :rounds
+  has_many :results, through: :rounds
+  has_many :competitions, through: :rounds
```
gregorbg: I don't think this is necessary anymore, correct me though if I'm wrong

app/models/linked_round.rb
```diff
@@ -0,0 +1,10 @@
...
+  has_many :competition_events, -> { distinct }, through: :rounds
+
+  validates :competition_event_ids, length: { is: 1, message: "must all belong to the same competition" }, if: -> { competition_event_ids.present? }
```
gregorbg: Why the `if` part?

FinnIckler: because if you create a `LinkedRound` without any round attached yet this will fail and don't we have to create the `LinkedRound` first because we need to know its id to set it for the rounds?

app/models/linked_round.rb
FinnIckler: yes, currently not

lib/advancement_conditions/dual_round_condition.rb
```diff
@@ -0,0 +1,17 @@
...
+      "dual"
+    end
+
+    def to_s(_round, short: false)
+      I18n.t("advancement_condition#{'.short' if short}.dual")
+    end
+
+    def max_advancing(results)
+      results.count
+    end
```
gregorbg: Depends on how you want to use this.
At this early prototyping stage, my thoughts are more along the lines of...
- R1 and R2 are linked as a "dual round"
- R1 -> R2 has a percent advancement criterion, with 100% being set
- R2 -> whatever comes after it is the one that has the dual condition
 
In essence, the dual round condition is not the one between the dual rounds themselves, but rather the one at the end marking "hey this _was_ a dual round, you need to consider both rounds".
With this idea in mind, the total number of results is then *not* correct for `max_advancing`

FinnIckler: The problem with this approach is that you then can't specify what advancement condition the round actually should have. I was thinking R1 has the actual Advancement condition of the round and R2 is marked as a dual round and basically says "combine results and look at advancement condition of previous round"

PR #12770, opened 2025-11-11
Use Rounds directly in ACV

FinnIckler:
Precondition to Dual Round support

### DIFF
--- a/lib/results_validators/advancement_conditions_validator.rb
+++ b/lib/results_validators/advancement_conditions_validator.rb
@@ -39,107 +39,114 @@ module ResultsValidators
         competition = competition_data.competition
         comp_start_date = competition.start_date
 
-        results_by_round_id = competition_data.results.group_by(&:round_id)
+        results_by_event_id = competition_data.results.group_by(&:event_id)
+        results_by_event_id.each do |event_id, results_for_event|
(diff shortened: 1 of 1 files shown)


lib/results_validators/advancement_conditions_validator.rb
```diff
@@ -35,133 +35,111 @@ def competition_associations
...
-            if previous_round_type_id
-              # Get the actual Round from the website: they are populated for all
-              # competitions and we can check both what actually happens and what
-              # was set to happen.
-              previous_round = competition.find_round_for(event_id, previous_round_type_id)
-
-              # It can happen that the WCIF contains rounds which have been deleted on the WCA website.
-              # Most famously, this occurs when the organizers planned with four rounds but less than 100 competitors
-              # sign up and they don't realize it immediately. Then, the competition proceeds with four rounds but
-              # upon results submission the final round is deleted from the competition_events table.
```
gregorbg, line 89: I wouldn't remove this specific comment. Only the "loading" part above can be removed safely.

FinnIckler: but it doesn't fit to previous_round.present? anymore right?

PR #12774, opened 2025-11-12
Add dual round advancement condition

FinnIckler:
I included Frontend changes to set a dual round and display it.

I moved merging results to a static method on LinkedRounds because we will have to do it that way in the ACV. 

How I am currently displaying it in the event pages. Suggestions welcome
Screenshot 2025-11-12 at 10 30 08
Screenshot 2025-11-12 at 10 30 37

### DIFF
--- a/lib/advancement_conditions/linked_round_condition.rb
+++ b/lib/advancement_conditions/linked_round_condition.rb
@@ -2,6 +2,15 @@
 
 module AdvancementConditions
   class LinkedRoundCondition < AdvancementCondition
+
+    attr_accessor :nested_advancement_condition, :linked_round_ids
+
+    def initialize(nested_advancement_condition, linked_round_ids)
+      self.nested_advancement_condition = nested_advancement_condition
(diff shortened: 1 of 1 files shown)


app/models/linked_round.rb
```diff
@@ -2,8 +2,44 @@
...
+    sorted_results = best_result_per_person.sort_by { |result| result.should_compute_average? ? result.to_solve_time(:average) : result.to_solve_time(:best) }
+    # Overwrite pos for display purposes for now (including handling ties)
+    last_time = nil
+    last_pos = 0
+    tie_count = 0
+
+    sorted_results.each_with_index do |result, index|
+      time = result.should_compute_average? ? result.to_solve_time(:average) : result.to_solve_time(:best)
+
+      if last_time.present? && time == last_time
```
FinnIckler: I know that this isn't technically correct yet as a tie depends on the event (for ao5 we need to have tied single and average for example).

PR #12776, opened 2025-11-13
WCA Live Integration by person Nextjs

FinnIckler:
Prototype frontend:
Screenshot 2025-11-13 at 10 47 05

I also included a rake task to migrate RoundResults to LiveResults for testing. Very handy for testing.

This should be the easiest one to spec out

### DIFF
--- a/app/models/live_result.rb
+++ b/app/models/live_result.rb
@@ -41,7 +41,7 @@ class LiveResult < ApplicationRecord
 
   def potential_score
     rank_by = round.format.sort_by == 'single' ? 'best' : 'average'
-    complete? ? to_solve_time(rank_by) : SolveTime.new(event_id, rank_by, BEST_POSSIBLE_SCORE)
+    complete? ? self[rank_by.to_sym] : BEST_POSSIBLE_SCORE
   end
 
   def should_recompute?


app/models/live_result.rb
```diff
@@ -41,7 +41,7 @@ def to_solve_time(field)
 
   def potential_score
     rank_by = round.format.sort_by == 'single' ? 'best' : 'average'
-    complete? ? self[rank_by.to_sym] : BEST_POSSIBLE_SCORE
+    complete? ? to_solve_time(rank_by) : SolveTime.new(event_id, rank_by, BEST_POSSIBLE_SCORE)
```
gregorbg: This changes the return value of the method entirely. I assume this is intended, but why?

next-frontend/src/components/live/LiveResultsTable.tsx
```diff
@@ -0,0 +1,17 @@
+import { components } from "@/types/openapi";
+
+const advancingColor = "0, 230, 118";
```
gregorbg: Colors like this should be defined in the global Chakra theme, and then referenced via CSS-in-JS

next-frontend/src/app/(wca)/competitions/[competitionId]/live/competitors/[registrationId]/page.tsx
```diff
@@ -0,0 +1,88 @@
...
+                return (
+                  <Table.Row key={`${roundId}-${key}`}>
+                    <Table.Cell>
+                      <Link
+                        href={`/competitions/${competitionId}/live/rounds/${roundId}`}
+                      >
+                        Round {roundId}
+                      </Link>
+                    </Table.Cell>
+                    <Table.Cell width={1} style={rankingCellStyle(result)}>
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/live/competitors/[registrationId]/page.tsx]: You can directly pass `backgroundColor="advancing"` or similar, if you work it into the theme (as indicated in my other comment)

next-frontend/src/app/(wca)/competitions/[competitionId]/live/competitors/[registrationId]/page.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const { name, results } = personResultRequest.data;
+
+  const resultsByEvent = _.groupBy(results, "event_id");
+
+  return (
+    <Container>
+      <Heading size="5xl">{name}</Heading>
+      {_.map(resultsByEvent, (eventResults, key) => (
+        <>
+          <Heading size="2xl">{events.byId[key].name}</Heading>
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/live/competitors/[registrationId]/page.tsx]: Rather than heading sizes, you should work with `textStyle`s defined in our Chakra theme. We have `h1` until `h3`, and `s1` and `s2`. Play around and see which one suits you best.

next-frontend/src/types/openapi.ts
gregorbg, line 1: Still not too crazy about the amount of (seemingly) unrelated diff in this file. Have you tried deleting and entirely re-generating the file?

And if that still yields significant changes: Please try the same (deleting and regenerating) on `main`, and push a hotfix (that can then be merged into this PR) if there are apparent changes.

app/models/live_result.rb
FinnIckler: because otherwise sorting will not be correct. I could also call SolveTime.new on the result of the method instead, but it's currently only used once

FinnIckler: I am not sure if there is a situation where we want this result not as solvetime

next-frontend/src/types/openapi.ts
FinnIckler: yeah I will check that

FinnIckler:
I will make a new PR with the chakra theme changes

next-frontend/src/types/openapi.ts
FinnIckler: Weirdly enough regenerating on main and on here both produces no changeset

next-frontend/src/components/live/LiveResultsTable.tsx
FinnIckler: For this I will just use the colour that's closest from the chakra colours

app/models/live_result.rb
FinnIckler: this method should be part of the `global_pos` PR though

PR #12779, opened 2025-11-14
WCA Live nextjs prerequisites 

FinnIckler:
Just a couple prerequisites that will make it easier to have multiple PRs open at the same time.
Includes:
- The rake task to import synced WCA Live results to new LiveResult Model
- A bunch of OpenAPI schemas that all the different Routes use
- installing wca-helpers and actioncable

### DIFF
--- a/next-frontend/yarn.lock
+++ b/next-frontend/yarn.lock
@@ -5894,13 +5894,6 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@wca/helpers@npm:^1.1.5":
-  version: 1.1.5
-  resolution: "@wca/helpers@npm:1.1.5"
-  checksum: 10c0/8a626f2213549586c4cfeb130f04dc8c63ea3f44e6804f29282e0aa973bb17b28cfb3b108b3758b2c677077413c8d9644546f97856abecab3045dc94c287c06f
-  languageName: node
-  linkType: hard
-
 "@zag-js/accordion@npm:1.27.0":
   version: 1.27.0
   resolution: "@zag-js/accordion@npm:1.27.0"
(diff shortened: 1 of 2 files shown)


next-frontend/yarn.lock
```diff
@@ -12463,11 +12485,13 @@ __metadata:
     "@types/js-cookie": "npm:^3"
     "@types/newrelic": "npm:^9.14.8"
     "@types/node": "npm:^24"
+    "@types/rails__actioncable": "npm:^8"
     "@types/react": "npm:^19"
     "@types/react-dom": "npm:^19"
     "@types/react-world-flags": "npm:^1"
     "@types/swagger-ui-react": "npm:^5.18.0"
     "@vitejs/plugin-react": "npm:^5.1.0"
+    "@wca/helpers": "npm:^1.1.5"
```
gregorbg: What do you need this package for?

FinnIckler: I am using it for the types like `CurrentEventId`. But I would also argue that we should be using it for functions like `parseActivityCode` instead of our codebase to keep a single source of truth that people can rely on no?

gregorbg: Right now we should build our own independent thing, and then look how to "properly" use the wca-helpers package later.
I don't want us to use an old, only veeeeeeeeery loosely maintained package "just for the sake of using it".
Ideally, the wca-helpers package should be "reborn" out of the code that we're building in NextJS now.

gregorbg: Right now we should build our own independent thing, and then look how to "properly" use the wca-helpers package later.
I don't want us to use an old, only veeeeeeeeery loosely maintained package "just for the sake of using it".
Ideally, the wca-helpers package should be "reborn" out of the code that we're building in NextJS now.

FinnIckler: ok

PR #12782, opened 2025-11-14
WCA Live Nextjs Round Results

FinnIckler:
Prototype frontend:
Screenshot 2025-11-14 at 14 52 58

There is definitely a lot of features still needed for this one, most notably actually live updating the results from the websocket connection and the projected results views.

### DIFF
--- a/next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/page.tsx
+++ b/next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/page.tsx
@@ -1,35 +1,24 @@
-"use client";
+"use server";
 
-import { useParams } from "next/navigation";
-import useAPI from "@/lib/wca/useAPI";
-import Loading from "@/components/ui/loading";
 import { Container, Heading, VStack } from "@chakra-ui/react";
 import LiveResultsTable from "@/components/live/LiveResultsTable";
 import { parseActivityCode } from "@/lib/wca/wcif/rounds";
+import { getResultByRound } from "@/lib/wca/live/getResultsByRound";
 
-export default function ResultPage() {
-  const { roundId, competitionId } =
-    useParams<"/competitions/[competitionId]/live/rounds/[roundId]">();
+export default async function ResultPage({
+  params,
+}: {
+  params: Promise<{ roundId: string; competitionId: string }>;
+}) {
+  const { roundId, competitionId } = await params;
 
-  const api = useAPI();
+  const resultsRequest = await getResultByRound(competitionId, roundId);
 
-  const { data: resultsRequest, isLoading } = api.useQuery(
-    "get",
-    "/v1/competitions/{competitionId}/live/rounds/{roundId}",
-    {
-      params: { path: { roundId, competitionId } },
-    },
-  );
-
-  if (isLoading) {
-    return <Loading />;
-  }
-
-  if (!resultsRequest) {
+  if (!resultsRequest.data) {
     return <p>Error loading Results</p>;
   }
 
-  const { results, id, competitors } = resultsRequest;
+  const { results, id, competitors } = resultsRequest.data;
 
   return (
     <Container bg="bg">
--- a/next-frontend/src/components/live/LiveResultsTable.tsx
+++ b/next-frontend/src/components/live/LiveResultsTable.tsx
@@ -1,4 +1,3 @@
-import { useMemo } from "react";
 import _ from "lodash";
 import events from "@/lib/wca/data/events";
 import { Link, Table } from "@chakra-ui/react";
@@ -54,16 +53,14 @@ export default function ResultsTable({
   const resultsByRegistrationId = _.keyBy(results, "registration_id");
   const event = events.byId[eventId];
 
-  const sortedCompetitors = useMemo(() => {
-    return _.orderBy(
-      competitors,
-      [
-        (competitor) => customOrderBy(competitor, resultsByRegistrationId),
-        (competitor) => customOrderBy(competitor, resultsByRegistrationId),
-      ],
-      ["asc", "asc"],
-    );
-  }, [competitors, resultsByRegistrationId]);
+  const sortedCompetitors = _.orderBy(
+    competitors,
+    [
+      (competitor) => customOrderBy(competitor, resultsByRegistrationId),
+      (competitor) => customOrderBy(competitor, resultsByRegistrationId),
+    ],
+    ["asc", "asc"],
+  );
 
   const solveCount = event.recommendedFormat.expected_solve_count;
   const attemptIndexes = [...Array(solveCount).keys()];
--- a/next-frontend/src/lib/wca/live/getResultsByRound.ts
+++ b/next-frontend/src/lib/wca/live/getResultsByRound.ts
@@ -1,0 +1,13 @@
+import { serverClient } from "@/lib/wca/wcaAPI";
+import { cache } from "react";
+
+export const getResultByRound = cache(
+  async (competitionId: string, roundId: string) => {
+    return await serverClient.GET(
+      "/v1/competitions/{competitionId}/live/rounds/{roundId}",
+      {
+        params: { path: { competitionId, roundId } },
+      },
+    );
+  },
+);


next-frontend/package.json
```diff
@@ -34,6 +34,7 @@
     "@payloadcms/storage-s3": "3.63.0",
     "@rails/actioncable": "^8.1.100",
     "@tanstack/react-query": "^5.90.7",
+    "@wca/helpers": "^1.1.5",
```
gregorbg: :/

app/models/round.rb
```diff
@@ -167,13 +167,13 @@ def accepted_registrations_with_wcif_id
...
                    .accepted
-                   .map { it.as_json({ include: [user: { only: [:name], methods: [], include: [] }] }).merge("registration_id" => r.registrant_id) }
+                   .map { it.as_json({ include: [user: { only: [:name], methods: [], include: [] }] }).merge("registration_id" => it.registrant_id) }
     else
       advancing = previous_round.live_results.where(advancing: true).pluck(:registration_id)
 
       Registration.includes(:user)
                   .find(advancing)
-                  .map { it.as_json({ include: [user: { only: [:name], methods: [], include: [] }] }).merge("registration_id" => r.registrant_id) }
+                  .map { it.as_json({ include: [user: { only: [:name], methods: [], include: [] }] }).merge("registration_id" => it.registrant_id) }
```
gregorbg: How did this ever work with `r`? (Did it even?)

app/models/round.rb
```diff
@@ -270,6 +270,21 @@ def to_wcif
     }
   end
 
+  def to_live_json(only_podiums: false)
```
gregorbg, line 255: How is this different from the "normal" WCIF?

next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/page.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const api = useAPI();
+
+  const { data: resultsRequest, isLoading } = useQuery({
+    queryKey: roundResultsKey(roundId, competitionId),
+    queryFn: () =>
+      api.GET("/v1/competitions/{competitionId}/live/rounds/{roundId}", {
+        params: { path: { roundId, competitionId } },
+      }),
+    select: (data) => data.data,
+  });
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/page.tsx]: You can now use the typed query helper from my other `openapi-react-query` PR which was just merged

next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/page.tsx
```diff
@@ -0,0 +1,52 @@
...
+  if (!resultsRequest) {
+    return <p>Error loading Results</p>;
+  }
+
+  const { results, id, competitors } = resultsRequest;
+
+  return (
+    <Container bg="bg">
+      <VStack align="left">
+        <Heading size="5xl">Live Results</Heading>
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/page.tsx]: Use `textStyle="h1"` or `h2` please

next-frontend/src/components/live/LiveResultsTable.tsx
```diff
@@ -0,0 +1,223 @@
...
+import _ from "lodash";
+import { components } from "@/types/openapi";
+import events from "@/lib/wca/data/events";
+import { Link, Table } from "@chakra-ui/react";
+import {
+  centisecondsToClockFormat,
+  formatAttemptResult,
+} from "@/lib/wca/wcif/attempts";
+
+const advancingColor = "0, 230, 118";
```
gregorbg: Put this into the theme, as per my other comment on #12776

next-frontend/src/components/live/LiveResultsTable.tsx
```diff
@@ -0,0 +1,223 @@
...
+
+const customOrderBy = (
```
gregorbg, line 11: How does the backend do similar tasks? If it never does: How does the current WCA Live do similar tasks?

next-frontend/src/components/live/LiveResultsTable.tsx
```diff
@@ -0,0 +1,223 @@
...
+    borderRadius: "4px",
+    fontWeight: 600,
+    fontSize: "0.6em",
+    position: "absolute",
+    top: "0px",
+    right: "0px",
+    transform: "translate(110%, -40%)",
+    color: "rgb(255, 255, 255)",
+    backgroundColor: "",
+  };
```
gregorbg: This is a *lot* of styles, I doubt they are all necessary.

next-frontend/src/components/live/LiveResultsTable.tsx
```diff
@@ -0,0 +1,223 @@
...
+  }
+
+  if (result?.advancing_questionable) {
+    return { backgroundColor: `rgba(${advancingColor}, 0.5)` };
+  }
+
+  return {};
+};
+
+export const recordTagStyle = (tag: string) => {
```
gregorbg: This way of manually injecting untyped styles at runtime is undesired in ChakraUI. Please (1) clean the styles up per my other comment and (2) add them into our theme. You could have them as a new "variant" (cf. https://www.chakra-ui.com/docs/theming/recipes#defining-the-recipe and existing definitions in our theme) and use the record tags as values for the variant types.

next-frontend/src/components/live/LiveResultsTable.tsx
```diff
@@ -0,0 +1,223 @@
...
+    display: "block",
+    lineHeight: 1,
+    padding: "0.3em 0.4em",
+    borderRadius: "4px",
+    fontWeight: 600,
+    fontSize: "0.6em",
+    position: "absolute",
+    top: "0px",
+    right: "0px",
+    transform: "translate(110%, -40%)",
```
gregorbg: This is especially shady. Why do you need inline transforms? If you want to move the component around within the row, think about [Float](https://www.chakra-ui.com/docs/components/float).
To the naked eye (haven't spun up your code at all locally) this looks like "Badge + Float + some **minimal** custom styling on the badge"

app/models/round.rb
FinnIckler: it contains the `live_results` instead of the `round_results`, it also contains the actual `round_id` and not only the `wcif_id` (because I am using that in my routes).
Ideally we want to have these changes as part of the normal wcif in the future, but this is the only way to change it non destructively

next-frontend/src/components/live/LiveResultsTable.tsx
FinnIckler: The backend just recomputes the positions. WCA Live also does it like that. 
For data being sorted when a new result comes in WCA Live solves this issue by sending all the data and replacing it. 
We discussed this before that we want to use a "send the id we want to sort it after" or something to correctly sort new data. 

So I could remove this for now and rely on the position and then introduce that logic when I put in the automatic update

app/models/round.rb
gregorbg: Then what about using some kind of "object spreading"? Ruby can do this...
```ruby
{
  **self.to_wcif,
  actual_round_id: self.id,
  results: self.live_results # would override the `results` key from `to_wcif`, just like in JS
}
```

next-frontend/src/components/live/LiveResultsTable.tsx
gregorbg: Yeah seems best not to do too much in one PR

app/models/round.rb
FinnIckler: it did not

next-frontend/src/components/live/LiveResultsTable.tsx
FinnIckler: yeah just using the `recordTagBadge` component that I made works fine

FinnIckler:
Screenshot 2025-11-20 at 11 41 24
This is how the badges look

next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/page.tsx
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/page.tsx:1]: I don't see why you have to run this page on the client.
As it currently stands, this can easily be an `async` server page

next-frontend/src/components/live/LiveResultsTable.tsx
```diff
@@ -0,0 +1,153 @@
...
+  const sortedCompetitors = useMemo(() => {
+    return _.orderBy(
+      competitors,
+      [
+        (competitor) => customOrderBy(competitor, resultsByRegistrationId),
+        (competitor) => customOrderBy(competitor, resultsByRegistrationId),
+      ],
+      ["asc", "asc"],
+    );
+  }, [competitors, resultsByRegistrationId]);
```
gregorbg: No need for this manual `useMemo` now that we have React Compiler 😉

PR #12783, opened 2025-11-16
Add developer mailing list opt-in

dunkOnIT:
Adds a preference which allows users to opt-in to a developer mailing list. This is a pretty straightforward/unsophisticated solution to the problem of having an email list - if this grows to contain too many users, or we want to start sending mails more regularly, we might want to consider alternative solutions.

For now, for the purpose of sending notifications about stuff like WCIF schema changes, export schema changes and API updates (when we release a public API), it should do the job.

### DIFF
--- a/app/views/users/edit.html.erb
+++ b/app/views/users/edit.html.erb
@@ -167,7 +167,6 @@
             <%= f.input :receive_developer_mails %>
           </div>
 
-
           <%= f.submit t('.save'), class: "btn btn-primary" %>
         <% end %>
       </div>
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -480,7 +480,7 @@ en:
         registration_notifications_enabled: "This only applies for competitions you organize or delegate, and can be overridden on a per-competition basis."
         unconfirmed_wca_id: ""
         wca_id: ""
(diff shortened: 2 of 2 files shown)


app/views/users/edit.html.erb
```diff
@@ -164,8 +164,10 @@
           <div id="notifications">
             <%= f.input :results_notifications_enabled %>
             <%= f.input :registration_notifications_enabled %>
+            <%= f.input :receive_developer_mails %>
           </div>
 
+
```
gregorbg: What is this extra empty line doing here❓

config/locales/en.yml
```diff
@@ -479,6 +480,7 @@ en:
         registration_notifications_enabled: "This only applies for competitions you organize or delegate, and can be overridden on a per-competition basis."
         unconfirmed_wca_id: ""
         wca_id: ""
+        receive_developer_mails: "This is limited to notifications for major version changes to resources like data exports, WCIF schema and public API's. Recommended for developers maintaing software that relies on these tools."
```
gregorbg: Not a native speaker, but shouldn't it be just `APIs` without the apostrophe?

dunkOnIT: Wow TIL

app/views/users/edit.html.erb
dunkOnIT: Removed

gregorbg (approved):

PR #12784, opened 2025-11-16
Add hook-based implementation of AttemptResultField

gregorbg:
This uses the existing fields from Webpacker-world as a blueprint, but with two major changes:
1. TypeScript typings! The fields are now "raw inputs" (ie `<Input ... />` components) that you can do whatever you want with: Wrap them in `Form`s, don't wrap them, reuse them, ...
2. The common behavior and shared patters that used to be duplicated code is now pulled out into shared hooks.

The idea of how to use these things is still largely the same as before: You have a "draft" value that the user is currently working on (while actively typing) and once the user is finished typing and blurs the field (by hitting tab or clicking some other component on the website) it tries to "commit" the value -- if it can parse it (because `abc` is not a valid result time).
This replicates the existing scoretaking "experience" where scoretakers tab from one field to the next. Shortcuts like `D` for DNF for example are duly being considered!

The usage is currently detailed in `/dashboard`, feel free to play around with it. Just change the `eventId` prop (and maybe reload the page because for whatever reason the NextJS HMR doesn't pick that one up on my machine).

I am not too happy with the MBLD input, because by nature it needs to be three fields but that makes them significantly less flexible / composable UI-wise. Open for suggestions on how to improve things.

Note that I have played around with `react-imask` and `Maskito` quite a fair bit, to try if standardized input maskings fit our bill. TL;DR: They don't. Longer story: They don't support "filling in" patterns from right to left (from centiseconds -> seconds -> minutes -> hours, like "up the unit ladder"), only LTR straight-forward pattern parsing. And they also don't have the notion of a "draft" value and instead hijack the whole input behavior via their own `ref`. Which I think is against the spirit of modern React composability.

### DIFF
--- a/next-frontend/src/app/(wca)/dashboard/AttemptResultField.tsx
+++ b/next-frontend/src/app/(wca)/dashboard/AttemptResultField.tsx
@@ -18,7 +18,7 @@ import _ from "lodash";
 import {
   DNF_VALUE,
   DNS_VALUE,
-  formatAttemptResult,
+  centisecondsToClockFormat,
   MultiBldResult,
   decodeMbldResult,
   SKIPPED_VALUE,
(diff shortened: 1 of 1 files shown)


FinnIckler (approved):
Looks good!

PR #12785, opened 2025-11-17
Reg admin: fix/clean up Paid On column

kr-matthews:
I've tested locally, but, I never managed to get a competition with payments working locally, so I still need someone else to fully test a competition using payment (with a stripe account hooked up).

In particular:

- for competitions without payment, only registered-on date is used
- for competitions with payment, only the payment date is used - if not available, it just says 'not paid'
- if the competitor was refunded, the date shown is the refund date, not the payment date, so I added a `*`
- hovering over the cell will provide a full date (with time, time zone) in all situations, and has text to explain whether it's the registration date, payment date, or refund date if the competition uses payments


app/webpacker/components/RegistrationsV2/RegistrationAdministration/AdministrationTableRow.jsx
```diff
@@ -21,39 +21,50 @@ import RegionFlag from '../../wca/RegionFlag';
...
+  const wasRefunded = paymentStatus === 'refund';
+  const trigger = (
+    <span>
+      {hasPaid
+        ? `${formatDate(updatedAt, withFullDate)}${wasRefunded ? '*' : ''}`
+        : I18n.t('registrations.list.not_paid')}
+    </span>
+  );
 
+  const content = (() => {
```
gregorbg, line 52: Why is this wrapped in a parameter-less callback?

kr-matthews: To get rid of the `let content` while still keeping the `if-return` pattern. Otherwise it would be a nested ternary, which is less readable.

PR #12821, opened 2025-11-17
Add local pos and global pos to live Results

FinnIckler:
Moves some code out of #12774 and adds the `local_pos` and `global_pos` columns. Ideally I would like to have a nice all SQL update to `global_pos` like we have with `local_pos`, but that would be quite the query (grouping the results by person, finding the minimum of each group and then ranking over them).

### DIFF
--- a/app/models/scramble.rb
+++ b/app/models/scramble.rb
@@ -2,7 +2,7 @@
 
 class Scramble < ApplicationRecord
   belongs_to :competition
-  belongs_to :round, optional: true
+  belongs_to :round
 
   validates :group_id, format: { presence: true, with: /\A[A-Z]+\Z/, message: "Invalid scramble group name" }
   validates :event_id, presence: true
@@ -11,13 +11,10 @@ class Scramble < ApplicationRecord
   validates :scramble_num, numericality: { presence: true, greater_than: 0 }
   validates :is_extra, inclusion: { presence: true, in: [true, false] }
 
-  # See the explanation in `resultable.rb` which has a validation with the same name.
-  validate :linked_round_consistent, if: :round_id?
-  def linked_round_consistent
-    errors.add(:competition, "Should match '#{round.competition_id}' of the linked round, but is '#{competition_id}'") unless competition_id == round.competition_id
-    errors.add(:round_type, "Should match '#{round.round_type_id}' of the linked round, but is '#{round_type_id}'") unless round_type_id == round.round_type_id
-    errors.add(:event, "Should match '#{round.event_id}' of the linked round, but is '#{event_id}'") unless event_id == round.event_id
-  end
+  delegate :competition_id, :round_type_id, :event_id, to: :round, prefix: true
+  validates :competition_id, comparison: { equal_to: :round_competition_id }
+  validates :round_type_id, comparison: { equal_to: :round_round_type_id }
+  validates :event_id, comparison: { equal_to: :round_event_id }
 
   def round_type
     RoundType.c_find(round_type_id)
--- a/db/migrate/20251117142333_change_scramble_to_round_reference_not_null.rb
+++ b/db/migrate/20251117142333_change_scramble_to_round_reference_not_null.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class ChangeScrambleToRoundReferenceNotNull < ActiveRecord::Migration[7.2]
+  def change
+    change_column_null :scrambles, :round_id, false
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.2].define(version: 2025_11_17_124011) do
+ActiveRecord::Schema[7.2].define(version: 2025_11_17_142333) do
   create_table "active_storage_attachments", charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
     t.string "name", null: false
     t.string "record_type", null: false
@@ -1261,7 +1261,7 @@ ActiveRecord::Schema[7.2].define(version: 2025_11_17_124011) do
     t.string "competition_id", limit: 32, null: false
     t.string "event_id", limit: 6, null: false
     t.string "round_type_id", limit: 1, null: false
-    t.integer "round_id"
+    t.integer "round_id", null: false
     t.string "group_id", limit: 3, null: false
     t.boolean "is_extra", null: false
     t.integer "scramble_num", null: false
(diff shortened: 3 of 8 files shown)


app/models/concerns/resultable.rb
```diff
@@ -179,6 +179,17 @@ def missed_combined_round_cutoff?
     @sorted_solves_with_index ||= solve_times.each_with_index.reject { |s, _| s.skipped? }.sort.freeze
   end
 
+  def tied_with?(other_result)
+    false if other_result.nil?
+    if %w[a m].include?(result.format_id)
```
gregorbg: `format`s have a `sort_by` column which can be even less technical debt than your current approach of hard-coding certain formats.

FinnIckler: actually just working on that because it makes combining results easier as well

app/models/linked_round.rb
```diff
@@ -3,7 +3,24 @@
...
+
+  def self.combine_results(round_results, format)
+    rank_by = format.rank_by_column
+    results_by_registration_id = round_results.group_by(&:registration_id)
+    persons = results_by_registration_id.keys
+    best_result_per_person = persons.map do |person|
+      results_by_registration_id[person].min_by { |result| result.to_solve_time(rank_by) }
+    end
+
+    best_result_per_person.sort_by { |result| result.to_solve_time(rank_by) }
```
gregorbg: Does this consider ties?

app/models/round.rb
```diff
@@ -153,6 +153,8 @@ def previous_round
   def accepted_registrations
     if number == 1
       registrations.accepted
+    elsif linked_round.present? # A linked round can currently only be Rounds Number 1+2
```
gregorbg: This is a quite strong assumption based on current Regulations. It may change in the future, is there a more general implementation that suits this idea?

app/models/round.rb
```diff
@@ -164,6 +166,8 @@ def accepted_registrations_with_wcif_id
       registrations.includes(:user)
                    .accepted
                    .map { it.as_json({ include: [user: { only: [:name], methods: [], include: [] }] }).merge("registration_id" => r.registrant_id) }
+    elsif linked_round.present? # A linked round can currently only be Rounds Number 1+2
```
gregorbg: This branch feels completely redundant, or am I missing something?

db/migrate/20251117124011_add_local_pos.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddLocalPos < ActiveRecord::Migration[7.2]
+  def change
+    change_table :live_results, bulk: true do |t|
+      t.rename :ranking, :local_pos
+      t.integer :global_pos
```
gregorbg:
```suggestion
      t.integer :global_pos, after: :local_pos
```

app/models/linked_round.rb
FinnIckler: it doesn't but it doesn't matter because ties are only necessary in the `global_pos` computation later

app/models/round.rb
FinnIckler: different kind of serialization

app/models/linked_round.rb
FinnIckler: the order of the results is correct, regardless of ties

app/models/linked_round.rb
```diff
@@ -3,7 +3,28 @@
...
 
   validates :competition_event_ids, length: { maximum: 1, message: "must all belong to the same competition" }
+
+  def merged_live_results
+    LinkedRound.combine_results(live_results, formats.first)
+  end
+
+  def first_round_in_link
+    rounds.min_by(&:number)
+  end
```
gregorbg, line 18: I would suggest applying a `-> { order(:number) }` scope directly to the `has_many` above. Then this kind of query will just turn into `rounds.first`.

app/models/live_result.rb
```diff
@@ -88,8 +91,12 @@ def notify_users
     end
 
     def recompute_advancing
+      has_linked_round = round.linked_round.present?
+      advancement_determining_column = has_linked_round ? 'global_pos' : 'local_pos'
```
gregorbg: In case of non-linked rounds, `global_pos` should exactly equal `local_pos`. So simply always picking `global_pos` should be fine

app/models/linked_round.rb
gregorbg: You may be correct, but I am not convinced yet. Please indulge me in your thinking.

app/models/live_result.rb
```diff
@@ -48,10 +48,13 @@ def should_recompute?
     saved_change_to_best? || saved_change_to_average?
   end
 
-  def recompute_ranks
-    rank_by = round.format.sort_by == 'single' ? 'best' : 'average'
+  def recompute_positions
+    # For linked rounds we need to merge the results and calculate a new global pos
+    recompute_global_pos if round.linked_round.present?
```
gregorbg: As per my other comment, we should consider always recomputing `global_pos`. In the case of non-linked rounds, the act of "recomputing" is as simple as copying over `local_pos` exactly

app/models/round.rb
```diff
@@ -150,9 +150,16 @@ def previous_round
     Round.joins(:competition_event).find_by(competition_event: competition_event, number: number - 1)
   end
 
+  def should_consider_previous_round_results?
```
gregorbg: This is quite a mouthful. The least that you can do is to remove the `should_`, it's common Ruby conventions to rely only on the `?` mark at the end to imply "some kind of auxiliary verb like can or must or has or should"

app/models/round.rb
```diff
@@ -150,29 +150,22 @@ def previous_round
...
+    # All linked rounds except the first one in a chain of linked rounds
+    linked_round.present? && linked_round.first_round_in_link.id != id
+  end
+
   def accepted_registrations
     if number == 1
       registrations.accepted
+    elsif should_consider_previous_round_results?
+      previous_round.accepted_registrations
```
gregorbg: You successfully fixed _half_ of my previous comment: The check is now Dual-agnostic, but the body of the `if` still looks at strictly the *immediate* predecessor. Using your new `linked_round.first_round_in_link.registrations` should do fine methinks.

app/models/live_result.rb
FinnIckler: but local_pos is only calculated after global_pos

FinnIckler: It might just work the other way around as well

app/models/live_result.rb
FinnIckler: done

app/models/linked_round.rb
FinnIckler: you are correct, I do need the `secondary_rank_by` for cases where single break ties

dunkOnIT:
> But I am letting @dunkOnIT think about that :P

On my list for after results export!

PR #12829, opened 2025-11-19
v2.0.0 Results Export

dunkOnIT:
## Overview

About as ready as I can get it - there is a chance some of the S3 stuff fails in prod, as I can't test that locally/on staging, but otherwise I have tested this locally and I'm generally happy with the shape of it. 

This PR does the following: 
- adds support for different versions of the results exports, including marking versions and end-of-life dates
- adds v2 of the results export, including changes to schema, readme, db configs
- generate results dumps for all "live" (non-EOL) versions
- updates the /export/results page to feature v2 data, point at v2 links, and explain the deprecation
- maintains the existing v1 permalinks until 2026-01-01, so existing tools should not be affected
  - endpoint changes are supported by tests

I've run it locally and I'm happy that it does what it should - with the caveat that I haven't exhaustively tested the output data beyond general sanity checks of whether the right tables are present and that the shape of their data is accurate.

## TODO
- [x] v2 export Changes
  - [x] Drop value1-5 from `results`
  - [x] Include `result_attempts` 
  - [x] Figure out where `PUBLIC_RESULTS_VERSION` is used, and determine if
    - [ ] it should be v1 or v2
    - [ ] if we need a second constant for the one 

- [x] Export job
  - [x] Add a v2 export schema
  - [x] Generate the v2 export like at all
  - [x] Should we change the name of the dump config? Or can we use results export for both? 
  - [x] Only generate the v2 export after 2026-01-01

- [x] Cronjob changes
  - [x] Generate v2 as part of the result exports - figure out if we want to do both as part of one command, or do them separately. Separately is probably better from a job management perspective? At least, they should be separate jobs, but triggered by the same cron entry point, and that entry point can be updated to remove v1 from it

- [x] Page updates
  - [x] Everything should be presented with v2 being the new default
  - [x] Format change explanation - v1 support will be discontinued on 2026-01-01, devs are encouraged to update before then
  - [x] Link to mailing list

- [x] Update the README
  - [x] Versioning information
  - [x] Subscribe to mailing list
  - [x] Updated info re structure of tables

- [x] Link changes
  - [x] Have a new link to the v2 file
  - [x] Raise an error in the v1 link from and including 2026-01-01

## Rollout process
1. Get this PR done so that we have a reference for the v2.0.0 format available on the website
1. Communicate the impending change to developers
  1. Announce on all relevant platforms 
  2. Reach out specifically to devs who maintain popular tools to ensure they're aware of the incoming change 
1. v1 will remain default until and including 2025-12-31 
1. We will increment to v2.0.0 on 2026-01-01, and stop providing v1.0.0 at this time
  1. There is an argument for continuing to provide v1.0.0 until we have the first H2H results in the db (at this point we lose backwards compatibility) - but I prefer making the breaking change earlier, provided we are able to give ~1 month of notice

### DIFF
--- a/app/views/database/_v2_public_results_readme.md.erb
+++ b/app/views/database/_v2_public_results_readme.md.erb
@@ -13,9 +13,25 @@ Opt in to the WCA Software Mailing List via [your profile preferences](https://w
 ## V2.0.0 Changelog
 
 - Solved snake_case/CamelCase inconsistency by changing all table names to snake_case (this matches the WCA codebase's convention)
-- Added `attempt_results` SQL table / TSV export file
-- `value1-5` in `results` will now be `NULL`
-- See [results and attempt_results](#results-and-attemptresults) for more information on how this data interacts
+- In addition to the above change in naming convention, the following column names have been changed:
+  - `competitions.wcaDelegate` -> `competitions.delegates`
+  - `competitions.organiser` -> `competitions.organizers`
+  - `competitions.latitude` -> `competitions.latitude_microdegrees`
+  - `competitions.longitude` -> `competitions.longitude_microdegrees`
+  - `persons.id` -> `persons.wca_id`
+  - `persons.subid` -> `persons.sub_id`
+  - `scrambles.scramble_id` -> `scrambles.id`
+  - ``
+- The following columns have been removed:
+  - `continents.latitude`
+  - `continents.longitude`
+  - `continents.zoom`
+  - `events.cellName`
+  - `results.value{1-5}`
+
+- Changes to recording of attempts
+  - Added `attempt_results` SQL table / TSV export file, and removed `value1-5` in `results` table
+  - See [results and attempt_results](#results-and-attemptresults) for more information on how this data interacts
 
 ## Description
 


dunkOnIT:
@thewca-bot deploy staging

dunkOnIT:
@thewca-bot deploy staging

dunkOnIT:
@thewca-bot deploy staging

lib/db_dump_helper.rb
```diff
@@ -59,57 +59,59 @@ def self.public_s3_file_size(file_name)
     bucket.object(file_name).content_length
   end
 
-  def self.resolve_results_export(file_type, export_timestamp = DumpPublicResultsDatabase.successful_start_date)
-    base_name = DbDumpHelper.result_export_file_name(file_type, export_timestamp)
+  def self.resolve_results_export(file_type, version, export_timestamp = DumpPublicResultsDatabase.successful_start_date)
+    return self.result_export_file_name(file_type, version, Time.new(2025, 11, 25)) unless Rails.env.production?
```
dunkOnIT, line 63: This allows for testing, similar to how we do it in `self.public_s3_file_size`

lib/db_dump_helper.rb
```diff
@@ -59,57 +59,59 @@ def self.public_s3_file_size(file_name)
...
+
+    base_name = DbDumpHelper.result_export_file_name(file_type, version, export_timestamp)
 
     "#{DbDumpHelper::RESULTS_EXPORT_FOLDER}/#{base_name}"
   end
 
-  def self.cached_results_export_info(file_type, export_timestamp = DumpPublicResultsDatabase.successful_start_date)
-    Rails.cache.fetch("database-export-#{export_timestamp}-#{file_type}", expires_in: 1.day) do
-      file_name = DbDumpHelper.resolve_results_export(file_type, export_timestamp)
+  def self.cached_results_export_info(file_type, version, export_timestamp = DumpPublicResultsDatabase.successful_start_date)
```
dunkOnIT, line 70: `version` added before `export_timestamp`, as `export_timestamp` is optional/has a default value

app/controllers/api/v0/api_controller.rb
```diff
@@ -221,15 +221,18 @@ def export_public
...
     _, sql_filesize = DbDumpHelper.cached_results_export_info("sql", timestamp)
     _, tsv_filesize = DbDumpHelper.cached_results_export_info("tsv", timestamp)
+    current_version_key = DatabaseDumper.current_results_export_version
+    current_version_number = DatabaseDumper::RESULTS_EXPORT_VERSIONS[current_version_key][:metadata][:export_format_version]
 
     render json: {
       export_date: timestamp&.iso8601,
-      sql_url: "#{sql_permalink_url}.zip",
+      export_version: current_version_number,
+      sql_url: Rails.application.routes.url_helpers.results_permalink_url(:v2, 'sql'),
```
gregorbg:
```suggestion
      sql_url: results_permalink_url(:v2, 'sql'),
```

app/controllers/api/v0/api_controller.rb
```diff
@@ -221,15 +221,18 @@ def export_public
...
       sql_filesize_bytes: sql_filesize,
-      tsv_url: "#{tsv_permalink_url}.zip",
+      tsv_url: Rails.application.routes.url_helpers.results_permalink_url(:v2, 'tsv'),
```
gregorbg:
```suggestion
      tsv_url: results_permalink_url(:v2, 'tsv'),
```

db/results_dump_v2_schema.rb
```diff
@@ -0,0 +1,157 @@
...
+    t.integer "average", default: 0, null: false
+    t.string "personName", limit: 80
+    t.string "personId", limit: 10, default: "", null: false
+    t.string "personCountryId", limit: 50
+    t.string "formatId", limit: 1, default: "", null: false
+    t.integer "value1", default: 0, null: false
+    t.integer "value2", default: 0, null: false
+    t.integer "value3", default: 0, null: false
+    t.integer "value4", default: 0, null: false
+    t.integer "value5", default: 0, null: false
```
gregorbg: Why are you still including these? I mean, we're going through the hassle of introducing a breaking change. Why would we carry deadwood with us through this major change?

db/results_dump_v2_schema.rb
```diff
@@ -0,0 +1,157 @@
...
+  create_table "championships", id: false, charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
+    t.integer "id", default: 0, null: false
+    t.string "competition_id", null: false
+    t.string "championship_type", null: false
+  end
+
+  create_table "eligible_country_iso2s_for_championship", charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
+    t.string "championship_type", null: false
+    t.string "eligible_country_iso2", null: false
+  end
```
gregorbg, line 148: Same as above: if we're introducing (very) breaking changes already, then why on earth would we need to "carry over" old inconsistent conventions into the shiny new format?

Either make everything `CamelCase` or everything `snake_case`. But going through the transition pain to v2 without fixing the ob(li)vious inconsistencies from v1? No.

gregorbg (changes requested):

app/controllers/api/v0/api_controller.rb
dunkOnIT: Throws an error unless we do an `includes Rails.application.routes.url_helpers` somewhere

app/controllers/api/v0/api_controller.rb
dunkOnIT: Throws an error unless we do an includes Rails.application.routes.url_helpers somewhere

db/results_dump_v2_schema.rb
dunkOnIT: Aha, I didn't understand the sanitizers-schema relationship at the time I left these in. Removed

db/results_dump_v2_schema.rb
dunkOnIT: Ah, didn't add the table name changes to my todo - changed now, including the necessary updates to README and sanitizers

app/controllers/api/v0/api_controller.rb
gregorbg: That is odd, because in the diff you can clearly see it was using a "naked" `sql_permalink_url` before.
And evidently, that did work for over 2 years **without** specifying `Rails.application.routes.url_helpers.sql_permalink_url`. So there's something else going on here...

gregorbg: I just tried locally and it _does_ work with only `results_permalink_url`, without all of the verbose helper prefixing.
image

dunkOnIT: Well seems like I was dead wrong, changed

db/results_dump_v2_schema.rb
gregorbg: This is only partially resolved now: You still have a big inconsistency in looots of column names between `camelCase` and `snake_case`

lib/database_dumper.rb
```diff
@@ -1205,6 +1203,274 @@ def self.actions_to_column_sanitizers(columns_by_action)
     }.freeze,
   }.freeze
 
+  V2_RESULTS_SANITIZERS = {
```
gregorbg, line 1206: General remark: If the `source_table` is equal to the key of the sanitizer, then you don't have to override

gregorbg: I know all these `source_table`s are probably coming from the first draft of the PR when the names were indeed different, but now it's time to clean up :P

dunkOnIT: Removed

db/results_dump_v2_schema.rb
dunkOnIT: Updated

lib/database_dumper.rb
```diff
@@ -1205,6 +1203,259 @@ def self.actions_to_column_sanitizers(columns_by_action)
...
+    }.freeze,
+    "persons" => {
+      column_sanitizers: actions_to_column_sanitizers(
+        copy: %w[
+          name
+          gender
+        ],
+        fake_values: {
+          "id" => "wca_id",
+          "subid" => "sub_id",
```
dunkOnIT: It was already `subid` in the v1 export - should we use this opportunity to change it to `sub_id`?

lib/database_dumper.rb
```diff
@@ -1205,6 +1203,259 @@ def self.actions_to_column_sanitizers(columns_by_action)
...
+        ],
+        fake_values: {
+          "city_name" => "city_name",
+          "country_id" => "country_id",
+          "venue_address" => "venue_address",
+          "venue_details" => "venue_details",
+          "cell_name" => "cell_name",
+          "cancelled" => "(competitions.cancelled_at IS NOT NULL AND competitions.cancelled_by IS NOT NULL)",
+          "event_specs" => "REPLACE(GROUP_CONCAT(DISTINCT competition_events.event_id), \",\", \" \")",
+          "wca_delegate" => "GROUP_CONCAT(DISTINCT(CONCAT(\"[{\", users_delegates.name, \"}{mailto:\", users_delegates.email, \"}]\")) SEPARATOR \" \")",
```
dunkOnIT: Should we use this opportunity to pluralize this field? ie, `delegates`

lib/database_dumper.rb
```diff
@@ -1205,6 +1203,259 @@ def self.actions_to_column_sanitizers(columns_by_action)
...
+          "organiser" => "GROUP_CONCAT(DISTINCT(CONCAT(\"[{\", users_organizers.name, \"}{mailto:\", users_organizers.email, \"}]\")) SEPARATOR \" \")",
```
dunkOnIT: Should we pluralize this field?

lib/database_dumper.rb
gregorbg: Feel free!

lib/database_dumper.rb
gregorbg: Feel free!

lib/database_dumper.rb
gregorbg: Yes, and probably "americanize" it by changing to `organizers`

dunkOnIT: Done

lib/database_dumper.rb
dunkOnIT: Done

lib/database_dumper.rb
dunkOnIT: Done

app/controllers/database_controller.rb
```diff
@@ -1,37 +1,67 @@
...
+    @sql_path, @sql_filesize = DbDumpHelper.cached_results_export_info("sql", :v2)
+    @tsv_path, @tsv_filesize = DbDumpHelper.cached_results_export_info("tsv", :v2)
 
     @sql_filename = File.basename(@sql_path)
     @tsv_filename = File.basename(@tsv_path)
   end
 
   def sql_permalink
-    url, = DbDumpHelper.cached_results_export_info("sql")
+    if Date.today > Date.parse("2025-12-31")
```
gregorbg: Could you move the hard-coded `Date` into a constant?

app/controllers/database_controller.rb
```diff
@@ -1,37 +1,67 @@
...
+      }, status: :gone
+    end
+
+    url, = DbDumpHelper.cached_results_export_info("sql", :v1)
     redirect_to url, status: :moved_permanently, allow_other_host: true
   end
 
   def tsv_permalink
-    url, = DbDumpHelper.cached_results_export_info("tsv")
+    if Date.today > Date.parse("2025-12-31")
```
gregorbg: constant

app/controllers/database_controller.rb
```diff
@@ -1,37 +1,67 @@
...
+
+    url, = DbDumpHelper.cached_results_export_info("tsv", :v1)
+    redirect_to url, status: :moved_permanently, allow_other_host: true
+  end
+
+  def results_permalink
+    version = params.require(:version).to_sym
+    file_type = params.require(:file_type)
+
+    if Date.today > Date.parse("2025-12-31") && version == :v1
```
gregorbg: constant

db/results_dump_v2_schema.rb
```diff
@@ -0,0 +1,152 @@
...
+    t.string "id", limit: 32, default: "", null: false
+    t.string "name", limit: 50, default: "", null: false
+    t.string "city_name", limit: 50, default: "", null: false
+    t.string "country_id", limit: 50, default: "", null: false
+    t.text "information", size: :medium
+    t.integer "year", limit: 2, default: 0, null: false, unsigned: true
+    t.integer "month", limit: 2, default: 0, null: false, unsigned: true
+    t.integer "day", limit: 2, default: 0, null: false, unsigned: true
+    t.integer "end_month", limit: 2, default: 0, null: false, unsigned: true
+    t.integer "end_day", limit: 2, default: 0, null: false, unsigned: true
```
gregorbg, line 25: You can introduce `end_year` as well, now that we're doing a breaking change.

db/results_dump_v2_schema.rb
```diff
@@ -0,0 +1,152 @@
...
+    t.text "event_specs", size: :long
+    t.text "delegates", size: :medium
+    t.text "organizers", size: :medium
+    t.string "venue", limit: 240, default: "", null: false
+    t.string "venue_address"
+    t.string "venue_details"
+    t.string "external_website", limit: 200
+    t.string "cell_name", limit: 45, default: "", null: false
+    t.integer "latitude"
+    t.integer "longitude"
```
gregorbg: These should specify their units. In the newer `wcif_venues` table, we use `latitude_microdegrees`.
You need to find out (a) whether these fields here (and in the original `competitions` table) are actually stored in microdegrees and (b) depending on your findings add a `_degrees` or `microdegrees` suffix

db/results_dump_v2_schema.rb
```diff
@@ -0,0 +1,152 @@
...
+  end
+
+  create_table "continents", id: false, charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
+    t.string "id", limit: 50, default: "", null: false
+    t.string "name", limit: 50, default: "", null: false
+    t.string "record_name", limit: 3, default: "", null: false
+    t.integer "latitude", default: 0, null: false
+    t.integer "longitude", default: 0, null: false
```
gregorbg: See remark about units above.

db/results_dump_v2_schema.rb
```diff
@@ -0,0 +1,152 @@
...
+    t.integer "zoom", limit: 1, default: 0, null: false
```
gregorbg: Also, in addition to the units debate, do we really need this in the *results* export?
(a) we don't provide similar data for countries
(b) how is this even relevant for computing anything about _results_?

gregorbg: For competitions, I can still understand that we give coordinates. Because (at least on a global scale) competitions can be "pinpointed" exactly. Continents cannot. Determining the "center" of a continent is by definition completely arbitrary and therefor completely useless

lib/database_dumper.rb
```diff
@@ -1205,6 +1203,259 @@ def self.actions_to_column_sanitizers(columns_by_action)
...
+          venue
+          latitude
+          longitude
+        ],
+        fake_values: {
+          "city_name" => "city_name",
+          "country_id" => "country_id",
+          "venue_address" => "venue_address",
+          "venue_details" => "venue_details",
+          "cell_name" => "cell_name",
```
gregorbg: Can't this just be a part of `copy` nowadays?

lib/database_dumper.rb
```diff
@@ -1205,6 +1203,259 @@ def self.actions_to_column_sanitizers(columns_by_action)
...
+    "persons" => {
+      column_sanitizers: actions_to_column_sanitizers(
+        copy: %w[
+          name
+          gender
+        ],
+        fake_values: {
+          "id" => "wca_id",
+          "sub_id" => "sub_id",
+          "country_id" => "country_id",
```
gregorbg: These two can go under `copy` now

lib/database_dumper.rb
```diff
@@ -1205,6 +1203,259 @@ def self.actions_to_column_sanitizers(columns_by_action)
...
+          "information" => "REGEXP_REPLACE(information, '[[:space:]]+', ' ')",
+        },
+      ),
+    }.freeze,
+    "scrambles" => {
+      column_sanitizers: actions_to_column_sanitizers(
+        copy: %w[
+          scramble
+        ],
+        fake_values: %w[
```
gregorbg: `fake_values` without actually telling what should be faked. How was this not caught during testing?

lib/database_dumper.rb
```diff
@@ -1205,6 +1203,259 @@ def self.actions_to_column_sanitizers(columns_by_action)
...
+    "continents" => {
+      column_sanitizers: actions_to_column_sanitizers(
+        copy: %w[
+          id
+          latitude
+          longitude
+          name
+          zoom
+        ],
+        fake_values: %w[
```
gregorbg: `fake_values` without actually telling what should be faked. How was this not caught during testing?

lib/database_dumper.rb
```diff
@@ -1205,6 +1203,259 @@ def self.actions_to_column_sanitizers(columns_by_action)
...
+      ),
+    }.freeze,
+    "countries" => {
+      column_sanitizers: actions_to_column_sanitizers(
+        copy: %w[
+          id
+          iso2
+          name
+        ],
+        fake_values: %w[
```
gregorbg: `fake_values` without actually telling what should be faked. How was this not caught during testing?

lib/database_dumper.rb
```diff
@@ -1205,6 +1203,259 @@ def self.actions_to_column_sanitizers(columns_by_action)
...
+    "events" => {
+      column_sanitizers: actions_to_column_sanitizers(
+        copy: %w[
+          id
+          format
+          name
+          rank
+        ],
+        fake_values: {
+          # Copy over column to keep backwards compatibility
```
gregorbg: Now we're introducing a major version break we don't have to worry about any backwards compatibility anymore

lib/database_dumper.rb
```diff
@@ -1205,6 +1203,259 @@ def self.actions_to_column_sanitizers(columns_by_action)
...
+    }.freeze,
+    "round_types" => {
+      column_sanitizers: actions_to_column_sanitizers(
+        copy: %w[
+          id
+          final
+          name
+          rank
+        ],
+        fake_values: %w[
```
gregorbg: `fake_values` without actually telling what should be faked. How was this not caught during testing?

lib/database_dumper.rb
```diff
@@ -1205,6 +1203,259 @@ def self.actions_to_column_sanitizers(columns_by_action)
...
+          country_rank
+        ],
+      ),
+    }.freeze,
+    "ranks_average" => {
+      column_sanitizers: actions_to_column_sanitizers(
+        copy: %w[
+          best
+        ],
+        fake_values: %w[
```
gregorbg: `fake_values` without actually telling what should be faked. How was this not caught during testing?

lib/database_dumper.rb
```diff
@@ -1205,6 +1203,259 @@ def self.actions_to_column_sanitizers(columns_by_action)
...
+          updated_at
+        ],
+      ),
+    }.freeze,
+    "ranks_single" => {
+      column_sanitizers: actions_to_column_sanitizers(
+        copy: %w[
+          best
+        ],
+        fake_values: %w[
```
gregorbg: `fake_values` without actually telling what should be faked. How was this not caught during testing?

lib/database_dumper.rb
```diff
@@ -1205,6 +1203,259 @@ def self.actions_to_column_sanitizers(columns_by_action)
...
+        ],
+        fake_values: {
+          "competition_id" => "competition_id",
+          "event_id" => "event_id",
+          "round_type_id" => "round_type_id",
+          "person_name" => "person_name",
+          "person_id" => "person_id",
+          "format_id" => "format_id",
+          "regional_single_record" => "regional_single_record",
+          "regional_average_record" => "regional_average_record",
```
gregorbg: These can go to `copy` nowadays

lib/database_dumper.rb
dunkOnIT: One response for all the comments:
- What I meant to do was take these to `copy`, as it seemed like they were mostly just taking the value of the column anyway? 
- Why tests didn't fail: We don't have automated tests against the actual values of the db dump

db/results_dump_v2_schema.rb
dunkOnIT: Done. Shouldn't it be `limit: 4`, though? I see that `year` also has limit 2, which is why I haven't changed it in the line, but it does perplex me: 
```ruby
    t.integer "year", limit: 2, default: 0, null: false, unsigned: true
```

lib/database_dumper.rb
dunkOnIT: Done

lib/database_dumper.rb
dunkOnIT: [Addressed here](https://github.com/thewca/worldcubeassociation.org/pull/12829#discussion_r2596039920)

lib/database_dumper.rb
dunkOnIT: [Addressed here](https://github.com/thewca/worldcubeassociation.org/pull/12829#discussion_r2596039920)

lib/database_dumper.rb
dunkOnIT: Done

lib/database_dumper.rb
dunkOnIT: [Addressed here](https://github.com/thewca/worldcubeassociation.org/pull/12829#discussion_r2596039920)

lib/database_dumper.rb
dunkOnIT: [Addressed here](https://github.com/thewca/worldcubeassociation.org/pull/12829#discussion_r2596039920)

lib/database_dumper.rb
dunkOnIT: [Addressed here](https://github.com/thewca/worldcubeassociation.org/pull/12829#discussion_r2596039920)

lib/database_dumper.rb
dunkOnIT: Done

lib/database_dumper.rb
dunkOnIT: Done

db/results_dump_v2_schema.rb
dunkOnIT: 100% agree, removed

db/results_dump_v2_schema.rb
dunkOnIT: Removed per other discussion

db/results_dump_v2_schema.rb
dunkOnIT: Changed to `_microdegrees`

app/controllers/database_controller.rb
dunkOnIT: Done

app/controllers/database_controller.rb
dunkOnIT: Done

app/controllers/database_controller.rb
dunkOnIT: Done

lib/database_dumper.rb
gregorbg: Okay, I don't think we need actual content tests against the values of the DB dump. That would be far too complicated.

What makes me curious is: The code using `fake_values` operates over the entries on a key-value basis. I.e. it think that every entry has a key and a value. I am surprised that the code which generates a dump (which in its most basic form -- does the code run and generate _something_ -- is definitely under test!) does not explode violently when "skipping" the fact that `fake_values` is supposed to be a dict/hash

db/results_dump_v2_schema.rb
gregorbg: This does not(!) specify the amount of digits directly, but rather the amount of storage space.
The connection is quite arbitrary (mostly due to historical reasons), see https://www.jdeen.com/blog/list-of-rails-data-types-for-activerecord-migrations under the special sub-heading "Integers"

gregorbg: In this case, an `unsigned limit 2` translates to "unsigned Smallint" in MySQL, which can capture values up to 65535

lib/database_dumper.rb
dunkOnIT: > I am surprised that the code which generates a dump (which in its most basic form -- does the code run and generate something -- is definitely under test!)

Is it under test, though? We have code that generates the _developer dump_ in `database_dumper_spec:46`, but I can't see anywhere that we actually run a dump for the results export. 

I'll focus on WCIF v1.1 first, and then see if I can replicate the dev dump test for the results exports.

PR #12833, opened 2025-11-21
Remove find_round_for method

gregorbg:
See title.

This cleanup is possible now that #12817 is in

### DIFF
--- a/app/controllers/api/v0/competitions_controller.rb
+++ b/app/controllers/api/v0/competitions_controller.rb
@@ -86,16 +86,16 @@ class Api::V0::CompetitionsController < Api::V0::ApiController
   def event_results
     competition = competition_from_params(associations: [:rounds])
     event = Event.c_find!(params[:event_id])
-    rounds = competition.results
-                        .includes(:round)
-                        .where(event_id: event.id)
-                        .group_by(&:round)
-                        .sort_by { |round, _| -round.number }
-                        .map do |round, results|
+    rounds = competition.rounds
+                        .includes(:results)
+                        .where(competition_events: { event: event })
+                        .except(:order)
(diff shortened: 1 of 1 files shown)


gregorbg:
Bonus: Also remove the backfilling Rake task. `round_id` is now `NOT NULL` everywhere, so trying to backfill null values is silly

app/controllers/api/v0/competitions_controller.rb
```diff
@@ -86,18 +86,15 @@ def podiums
...
-    results_by_round = competition.results
-                                  .where(event_id: event.id)
-                                  .group_by(&:round_type)
-                                  .sort_by { |round_type, _| -round_type.rank }
-    rounds = results_by_round.map do |round_type, results|
-      # I think all competitions now have round data, but let's be cautious
-      # and assume they may not.
-      # round data.
-      round = competition.find_round_for(event.id, round_type.id)
+    rounds = competition.results
```
FinnIckler: I actually also changed this code recently and I did
```
rounds_for_event = competition.rounds
                                  .filter { |round| round.event_id == event.id }
                                  .sort_by { |round| -round.round_type.rank }
rounds = rounds_for_event.map do |round|
  {
    id: round.id,
    roundTypeId: round.round_type.id,
    results: round.results.sort_by { |r| [r.pos, r.person_name] },
  }
```
Which does look much shorter, not sure if it is quicker though

gregorbg: Thanks for the idea, I think I know how to make this quicker :D

gregorbg: Went with the "middle ground" approach now :D

PR #12834, opened 2025-11-21
Only use asymmetric OIDC encryption keys in prod

FinnIckler:
The current key in `.env.development` can only be used for symmetric encryption, so let's just switch to that for development and test.

### DIFF
--- a/env_config.rb
+++ b/env_config.rb
@@ -86,8 +86,6 @@ EnvConfig = SuperConfig.new(raise_exception: !is_compiling_assets) do
 
   mandatory :DUMP_HOST, :string
 
-  mandatory :OIDC_ALGORITHM, :string
-
   # ROOT_URL is used when generating full urls (rather than relative urls).
   # Trick to discover the port we're set to run on from
   # https://stackoverflow.com/a/48069920/1739415.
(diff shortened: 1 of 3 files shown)


config/initializers/doorkeeper_openid_connect.rb
```diff
@@ -7,7 +7,7 @@
   end
   signing_key AppSecrets.OIDC_SECRET_KEY
   # This is an asymmetric encryption, meaning the clients validating the JWT Token use the Public key instead
```
gregorbg: The comment should be adjusted accordingly

env_config.rb
```diff
@@ -86,6 +86,8 @@
 
   mandatory :DUMP_HOST, :string
 
+  mandatory :OIDC_ALGORITHM, :string
```
gregorbg: Move this closer to `OIDC_ISSUER` please

infra/wca_on_rails/production/rails.tf
```diff
@@ -18,6 +18,10 @@ locals {
       name  = "OIDC_ISSUER"
       value = var.ROOT_URL
     },
+    {
+      name  = "OIDC_ALGORITHM"
+      value = "rs256"
```
gregorbg: This should be uppercase for consistency

infra/wca_on_rails/staging/rails.tf
```diff
@@ -16,6 +16,10 @@ locals {
       name  = "OIDC_ISSUER"
       value = var.ROOT_URL
     },
+    {
+      name  = "OIDC_ALGORITHM"
+      value = "rs256"
```
gregorbg: uppercase, see above

PR #12867, opened 2025-11-25
WCA Live nextjs useResultsSubscription

FinnIckler:
image

Probably makes sense to make the connection indicator more prominent.
Also I couldn't use `useEffectEvent` yet because we are still on React 18 :cry: 
I will make a separate PR to remove the '/api' part out of the environment variable and introduce a routes file that has the cable route and api route.

### DIFF
--- a/next-frontend/src/app/(wca)/layout.tsx
+++ b/next-frontend/src/app/(wca)/layout.tsx
@@ -16,11 +16,11 @@ export const metadata: Metadata = {
 const devFont = Rubik({ subsets: ["latin"] });
 
 const computeFont = async () => {
-  // if (process.env.PROPRIETARY_FONT === "TTNormsPro") {
-  //   const { ttNormsPro } = await import("@/styles/fonts");
-  //
-  //   return ttNormsPro;
-  // }
+  if (process.env.PROPRIETARY_FONT === "TTNormsPro") {
+    const { ttNormsPro } = await import("@/styles/fonts");
+
+    return ttNormsPro;
+  }
 
   return devFont;
 };


FinnIckler:
This does not clear up how we will handle live updates when it comes to recomputing positions/advancing. We probably will need some kind of changeset protocol?

next-frontend/src/lib/hooks/useResultsSubscription.ts
```diff
@@ -0,0 +1,49 @@
...
+        received: onReceived,
+        initialized: () => setConnectionState(CONNECTION_STATE_INITIALIZED),
+        connected: () => setConnectionState(CONNECTION_STATE_CONNECTED),
+        disconnected: () => setConnectionState(CONNECTION_STATE_DISCONNECTED),
+      },
+    );
+
+    return () => {
+      subscription.unsubscribe();
+      return;
```
gregorbg: This statement seems a bit useless...?

next-frontend/src/components/live/ConnectionPulse.tsx
```diff
@@ -0,0 +1,39 @@
...
+  );
+
+  const connectionColor = CONNECTION_COLORS[connectionState];
+
+  const isConnected = connectionState === CONNECTION_STATE_CONNECTED;
+
+  const shouldShow = !isConnected || animationPulse;
+
+  return (
+    <Box
```
gregorbg: What about using https://www.chakra-ui.com/docs/components/status?

gregorbg: You could even consider using the label feature there to actually show a text (that indicates to users better what this blinking thing actually _means_). But in the scope of the current PR, that's optional decoration of course :D

next-frontend/src/components/live/ConnectionPulse.tsx
```diff
@@ -0,0 +1,39 @@
...
+  connectionState,
+  animationDuration = PULSE_DURATION,
+}: {
+  connectionState: ConnectionState;
+  animationDuration?: number;
+}) {
+  const animationPulse = usePerpetualState(
+    (prev) => !prev,
+    animationDuration * 1.5,
+  );
```
gregorbg: We had needed this "self-toggling" boolean because of the way that SemUI's hide-and-show components worked.

Now with Chakra, we have _much_ better inline type-safe CSS animation support: https://www.chakra-ui.com/docs/styling/style-props/transitions#animation. You should be able to use a `infinite` timing spec combined with an out-of-the-box animation in the Chakra theme: https://www.chakra-ui.com/docs/theming/animations

gregorbg: I'm not sure how disabling the animation would look like, but the most "crude" solution that I could think of is to just do `const animation = isConnected ? 'pulse' : undefined`

next-frontend/src/lib/hooks/usePerpetualState.tsx
gregorbg, line 1: Sidenote: If you can figure out a clean theme-based animation as per my other comments, then this file and hook are not needed at all in your current PR

next-frontend/src/lib/hooks/useResultsSubscription.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export default function useResultsSubscription(
+  roundId: number,
+  onReceived: (data: components["schemas"]["LiveResult"]) => void,
+) {
+  const [connectionState, setConnectionState] = useState<ConnectionState>(
+    CONNECTION_STATE_INITIALIZED,
+  );
+
+  useEffect(() => {
```
gregorbg, line 28: For the record: `useEffect` is okay here because it is managing an external connection (the websocket) outside of React's lifecycle (which is the original intention of `useEffect`), _and_ because it has a cleanup function that it can `return`.

next-frontend/src/lib/hooks/useResultsSubscription.ts
FinnIckler: it has to return  void

FinnIckler:
https://github.com/user-attachments/assets/d9cbcb25-dc5f-47a4-a640-d144d9542a3e

next-frontend/src/components/live/ConnectionPulse.tsx
```diff
@@ -0,0 +1,31 @@
...
+  const connectionColor = CONNECTION_COLORS[connectionState];
+
+  return (
+    <Status.Root colorPalette={connectionColor}>
+      <Status.Indicator
+        animationName="ping"
+        animationDuration="1s"
+        animationTimingFunction="cubic-bezier(0, 0, 0.2, 1)"
+        animationIterationCount="infinite"
+      />
```
gregorbg, line 28: Cool! How do you stop the animation though based on the current `connectionState`?

In the video that you posted the dot is indeed changing color (and text), but it seems to be always blinking. I think only a (slow) blink when it's actually connected and live (green) makes the most sense.

next-frontend/src/components/live/ConnectionPulse.tsx
```diff
@@ -0,0 +1,31 @@
...
+  const { t } = useT();
+
+  const connectionColor = CONNECTION_COLORS[connectionState];
+
+  return (
+    <Status.Root colorPalette={connectionColor}>
+      <Status.Indicator
+        animationName="ping"
+        animationDuration="1s"
+        animationTimingFunction="cubic-bezier(0, 0, 0.2, 1)"
```
gregorbg: Why is this custom Bézier timing necessary? Does it work by skipping this key and just specifying `animationName` only?

gregorbg: Take a look at this: https://github.com/chakra-ui/chakra-ui/blob/650bfd1fcfe8cbf78ae3c4066e3cd219ad5bc894/apps/compositions/src/lib/keyframe-doc.tsx#L30

The result can be seen at https://www.chakra-ui.com/docs/theming/animations. Maybe something more "trivial" like `ease-in-out` might work?

next-frontend/src/lib/hooks/usePerpetualState.tsx
gregorbg: You marked my previous comment as "Resolved", yet the file is still here. Please delete it now that it's unused.

next-frontend/src/lib/hooks/useResultsSubscription.ts
gregorbg: Even then, just calling `subscription.unsubscribe()` as the first and only line of code within the `=> {}` block should do. I insist that the `return;` itself is unnecessary.

next-frontend/src/app/(wca)/layout.tsx
gregorbg, line 1: Ehehe :/

FinnIckler: removed

PR #12871, opened 2025-11-26
Add Past Competitions tab to Helpful Queries

Brandon-Reichman:
adds a tab for past competitions to the helpful queries panel, for WIC.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -438,7 +438,7 @@ class UsersController < ApplicationController
 
     competitions = user.person.competitions
                        .over.visible.not_cancelled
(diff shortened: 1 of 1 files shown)


app/controllers/users_controller.rb
```diff
@@ -429,6 +429,23 @@ def delegated_competitions
     )
   end
 
+  def past_competitions
+    user_id = params.require(:userId)
+
+    user = User.find(user_id)
+
+    return Competition.none unless user.person
```
gregorbg, line 437: TIL that `MyModel.none` is a thing in Rails 😄

gregorbg (approved):

PR #12872, opened 2025-11-26
Add and display mocked head-to-head round value

dunkOnIT:
This PR:
- adds a `mock_h2h` column on the `round` table
- adds a scope for querying this property, as well as a `h2h_events` method in `competition.rb` which gives the `ids` of all events marked as having h2h finals
- exposes `h2h_events` in the `competition_info` JSON, so that the frontend can access it without requiring changes to WCIF
- changes the `EventsTable` to display "Head-to-Head Finals" instead of "Finals" where appropriate 

image

This PR _does not_: 
- enforce validations on what round can be marked as H2H, as toggling H2H will happen in the database, so it will bypass Rails code in any case
  - this is acceptable given the very limited number of competitions H2H will be used on, particularly in this initial mock-support stage

If this approach is acceptable, next steps are: 
- change "Finals" text on Calendar to "Head-to-Head Finals"
- deploy to staging for WMCT UAT approval

### DIFF
--- a/app/webpacker/components/EventsTable/index.jsx
+++ b/app/webpacker/components/EventsTable/index.jsx
@@ -1,4 +1,4 @@
-import React, { useCallback } from "react";
+import React, { useCallback } from 'react';
 import {
   Table,
   TableBody,
@@ -18,20 +18,17 @@ import {
 } from '../../lib/utils/wcif';
 
 export default function EventsTable({ competitionInfo, wcifEvents }) {
-  const determineRoundLabel = useCallback(
-    (index, numRounds, round) => {
-      const cutoff = Boolean(round.cutoff)
-      const roundId = round.id
-      const roundTypeId = getRoundTypeId(index, numRounds, cutoff);
-      const roundTypeName = I18n.t(`rounds.${roundTypeId}.cell_name`)
+  const determineRoundLabel = useCallback((index, numRounds, round) => {
+    const cutoff = Boolean(round.cutoff);
+    const roundId = round.id;
+    const roundTypeId = getRoundTypeId(index, numRounds, cutoff);
+    const roundTypeName = I18n.t(`rounds.${roundTypeId}.cell_name`);
 
-      if (competitionInfo.h2h_rounds.includes(roundId) && roundTypeId === 'f') {
-        return `${I18n.t('formats.h')} ${roundTypeName}`;
-      } else {
-        return roundTypeName
-      }
-    }, [competitionInfo.h2h_rounds]
-  )
+    if (competitionInfo.h2h_rounds.includes(roundId) && roundTypeId === 'f') {
+      return `${I18n.t('formats.h')} ${roundTypeName}`;
+    }
+    return roundTypeName;
+  }, [competitionInfo.h2h_rounds]);
 
   return (
     <div style={{ overflowX: 'scroll' }}>


app/models/competition.rb
```diff
@@ -3053,4 +3053,12 @@ def attempt_auto_close!
     threshold_reached = fully_paid_registrations_count >= auto_close_threshold && auto_close_threshold.positive?
     threshold_reached && update(closing_full_registration: true, registration_close: Time.now)
   end
+
+  def h2h_events
+    competition_events
+      .joins(:rounds)
+      .merge(Round.h2h)
+      .distinct
+      .pluck(:event_id)
```
gregorbg: From the top of my head, `self.rounds.h2h.pluck(:event_id).uniq` should do

config/locales/en.yml
```diff
@@ -139,6 +139,9 @@ en:
     "h":
       name: "Qualification round"
       cell_name: "Qualification round"
+    "h2h":
+      name: "Head-to-Head Final"
+      cell_name: "Head-to-Head Final"
```
gregorbg: Are we sure that (even in the long run, when we have "proper" support) finals are the only rounds that can appear as H2H?

And even if that's the case from a *current 2026 Regulations* perspective: Wouldn't it make more sense to define H2H as a separate "add-on" string, and glue it in front of whatever the round type originally was?

db/migrate/20251126083044_add_mock_h2_h_to_round_table.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddMockH2HToRoundTable < ActiveRecord::Migration[7.2]
+  def change
+    add_column :rounds, :mock_h2h, :boolean, default: false, null: false
```
gregorbg: Inside the database we have established the convention of using `is_` or `has_` prefixes.
Rails wants to use `?` suffixes, but those are not allowed as MySQL column names, so we went with the prefix approach instead.

app/models/competition.rb
gregorbg: On second thought: I am not too big of a fan of "hiding" the information about H2H *rounds* behind the corresponding *events*. You effectively lose information about _which_ round of this event was marked as H2H. And I don't see a reason why we want to lose / get rid of that information.

Suggestion: You change this into `h2h_rounds` generating wcif_ids for all rounds that are marked as `is_h2h_mock`. Something like `self.rounds.h2h.map(&:wcif_id)` should really do. In the frontend, the round knows its WCIF ID, so a simple `includes` in JS will do

app/models/competition.rb
```diff
@@ -1868,7 +1868,7 @@ def to_competition_info
                information events_per_registration_limit guests_enabled auto_accept_preference auto_accept_disable_threshold],
       methods: %w[url website short_name city venue_address venue_details latitude_degrees longitude_degrees country_iso2 event_ids
                   main_event_id number_of_bookmarks using_payment_integrations? uses_qualification? uses_cutoff? competition_series_ids registration_full?
-                  part_of_competition_series? registration_full_and_accepted?],
+                  part_of_competition_series? registration_full_and_accepted? h2h_events],
```
gregorbg: Please add a comment specifying that this is temporary and should be removed from serialization once we have "proper" support.

It is really easy (especially in NextJS land) to serialize a ton of properties but never use them in the OpenAPI spec. Hard to detect but expensive on our database.

db/migrate/20251126083044_add_mock_h2_h_to_round_table.rb
dunkOnIT: Done

app/models/competition.rb
dunkOnIT: Done

config/locales/en.yml
dunkOnIT: Good point - I've added the key to `formats` instead

app/models/competition.rb
dunkOnIT: Changed the method body, and now returning round_id instead of event_id

app/webpacker/components/EventsTable/index.jsx
```diff
@@ -18,6 +18,15 @@ import {
...
 
 export default function EventsTable({ competitionInfo, wcifEvents }) {
+  function determineRoundLabel(index, numRounds, cutoff, roundId) {
+    const roundLabel = getRoundTypeId(index, numRounds, cutoff);
+
+    if (competitionInfo.h2h_rounds.includes(roundId) && roundLabel === 'f') {
+      return `${I18n.t('formats.h')} ${I18n.t('rounds.f.cell_name')}`;
+    }
+    return I18n.t(`rounds.${roundLabel}.cell_name`);
+  }
```
gregorbg: The old SemUI Webpacker code does not have React-Compiler installed, so you should wrap this in a `useCallback` to keep the reference stable

app/webpacker/components/EventsTable/index.jsx
```diff
@@ -59,7 +68,9 @@ export default function EventsTable({ competitionInfo, wcifEvents }) {
                   {events.byId[event.id].name}
                 </TableCell>
               )}
-              <TableCell>{I18n.t(`rounds.${getRoundTypeId(i + 1, event.rounds.length, Boolean(round.cutoff))}.cell_name`)}</TableCell>
+              <TableCell>
+                {determineRoundLabel(i + 1, event.rounds.length, Boolean(round.cutoff), round.id)}
```
gregorbg: You could just pass `round` and let the `determineRoundLabel` method internally read the `cutoff` and `id` properties as required.

app/webpacker/components/EventsTable/index.jsx
```diff
@@ -18,6 +18,15 @@ import {
 } from '../../lib/utils/wcif';
 
 export default function EventsTable({ competitionInfo, wcifEvents }) {
+  function determineRoundLabel(index, numRounds, cutoff, roundId) {
+    const roundLabel = getRoundTypeId(index, numRounds, cutoff);
```
gregorbg: You should call this `roundTypeId`, that's established "WCA vocabulary". (And besides it is the return value of a method literally called "get the roundTypeId")

app/webpacker/components/EventsTable/index.jsx
```diff
@@ -18,6 +18,15 @@ import {
...
+
+    if (competitionInfo.h2h_rounds.includes(roundId) && roundLabel === 'f') {
+      return `${I18n.t('formats.h')} ${I18n.t('rounds.f.cell_name')}`;
+    }
+    return I18n.t(`rounds.${roundLabel}.cell_name`);
```
gregorbg: You could extract the call that does `rounds.$roundLabel.cell_name` into a variable and then do `if ... return "${I18n h2h} variable" else return variable`

app/webpacker/components/EventsTable/index.jsx
```diff
@@ -18,6 +18,15 @@ import {
 } from '../../lib/utils/wcif';
 
 export default function EventsTable({ competitionInfo, wcifEvents }) {
+  function determineRoundLabel(index, numRounds, cutoff, roundId) {
+    const roundLabel = getRoundTypeId(index, numRounds, cutoff);
+
+    if (competitionInfo.h2h_rounds.includes(roundId) && roundLabel === 'f') {
```
gregorbg: In how far is the "should be final" safeguard justified?

dunkOnIT: Completely according to the [current proposed regulations](https://github.com/thewca/wca-regulations-january-2026/pull/62/files):
> 9b7b) The Head to Head format may only be used for the final round of an event.

But I agree with your general considerations around not hard-limiting H2H to finals only / being aware that the format could be made available for other round types

app/webpacker/components/EventsTable/index.jsx
dunkOnIT: Changed

dunkOnIT:
@thewca-bot deploy staging

PR #12873, opened 2025-11-26
Nextjs Profile Tabs

FinnIckler:
image
image

Also cleans up a lot of the Slate code. 
I have not done Championship Podium Tabs and Map Tabs yet. For the Podiums, the openapi spec says that is in the API, but it isn't? For the map, we don't have a library in nextjs yet.

### DIFF
--- a/next-frontend/src/components/persons/PersonalRecordsTable.tsx
+++ b/next-frontend/src/components/persons/PersonalRecordsTable.tsx
@@ -7,7 +7,6 @@ import { LuShare2 } from "react-icons/lu";
 import { components } from "@/types/openapi";
 import events from "@/lib/wca/data/events";
 import { formatAttemptResult } from "@/lib/wca/wcif/attempts";
-import _ from "lodash";
 
 interface RecordsProps {
   records: components["schemas"]["PersonInfo"]["personal_records"];
--- a/next-frontend/src/components/persons/ResultsTab.tsx
+++ b/next-frontend/src/components/persons/ResultsTab.tsx
@@ -11,10 +11,28 @@ interface ResultsTabProps {
 }
 
 const ResultsTab: React.FC<ResultsTabProps> = ({ wcaId }) => {
+  const [eventId, setEventId] = useState("333");
+
+  return (
+    <VStack>
+      <EventSelector
+        title=""
+        selectedEvents={[eventId]}
+        onEventClick={(e) => setEventId(e)}
+        hideAllButton={true}
+        hideClearButton={true}
+      />
+      <Results wcaId={wcaId} eventId={eventId} />
+    </VStack>
+  );
+};
+
+const Results: React.FC<{ wcaId: string; eventId: string }> = ({
+  wcaId,
+  eventId,
+}) => {
   const api = useAPI();
   const { t } = useT();
-
-  const [eventId, setEventId] = useState("333");
 
   const { data: resultsQuery, isLoading } = api.useQuery(
     "get",
@@ -32,18 +50,7 @@ const ResultsTab: React.FC<ResultsTabProps> = ({ wcaId }) => {
     return <Text>Failed fetching results</Text>;
   }
 
-  return (
-    <VStack>
-      <EventSelector
-        title=""
-        selectedEvents={[eventId]}
-        onEventClick={(e) => setEventId(e)}
-        hideAllButton={true}
-        hideClearButton={true}
-      />
-      <ByCompetitionTable results={resultsQuery} t={t} />
-    </VStack>
-  );
+  return <ByCompetitionTable results={resultsQuery} t={t} />;
 };
 
 export default ResultsTab;


app/controllers/api/v0/persons_controller.rb
```diff
@@ -40,6 +40,12 @@ def competitions
     render json: person.competitions
   end
 
+  def records
+    person = Person.current.find_by!(wca_id: params[:wca_id])
+    render json: person.results
+                       .where("regional_single_record IS NOT NULL OR regional_average_record IS NOT NULL").as_json
```
gregorbg:
```suggestion
    person_results = person.results
    render json: person_results
                       .where.not(regional_single_record: nil).or(person_results.where.not(regional_average_record: nil))
```

next-frontend/src/components/persons/ChampionshipPodiums.tsx
```diff
@@ -1,7 +1,16 @@
...
 
-const ChampionshipPodiumsTab: React.FC = () => {
+interface ChampionshipPodiumProps {
+  championship_podiums: {
+    world?: Record<string, never>[];
+    continental?: Record<string, never>[];
+    greater_china?: Record<string, never>[];
+    national?: Record<string, never>[];
+  };
+}
```
gregorbg: Can we "infer" this from the OpenAPI types?

next-frontend/src/components/persons/RecordsTable.tsx
```diff
@@ -0,0 +1,66 @@
...
+      {"WR" in recordsByType && (
+        <>
+          <Heading textStyle="h2">{t("persons.show.world_records")}</Heading>
+          <RecordsByEvent recordResults={recordsByType["WR"]} t={t} />
+        </>
+      )}
+      {["ER", "NAR", "SAR", "ASR", "OCR"].map((region) => {
+        return (
+          region in recordsByType && (
+            <>
```
gregorbg: Needs to be a `React.Fragment` so that it can specify `key={region}`

next-frontend/src/lib/dates/format.ts
```diff
@@ -0,0 +1,31 @@
+export function formatDateRange(start: string, end: string): string {
```
gregorbg: Can this use Luxon?

next-frontend/src/app/(wca)/persons/[wcaId]/page.tsx
```diff
@@ -12,6 +12,22 @@ import MapTab from "@/components/persons/MapTab";
...
+  event: string;
+  snr: number;
+  scr: number;
+  swr: number;
+  single: string;
+  average: string;
+  anr: number;
+  acr: number;
+  awr: number;
+}
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/persons/[wcaId]/page.tsx]: Is it possible to (easily, without a ton of effort) reshape this so that it's closer to what OpenAPI already returns anyways?

gregorbg (changes requested):
I love cleanup! Now let's clean up some more 😄

app/controllers/api/v0/persons_controller.rb
FinnIckler: that's what I had initially, but I didn't really think the nested where was cleaner

next-frontend/src/components/persons/ChampionshipPodiums.tsx
FinnIckler: done, but this is still not actually in the openAPI response

next-frontend/src/components/persons/RecordsTable.tsx
FinnIckler: Yes and no, from a React perspective it does, but a user will never have more than one kind of Continental Record at a time. I am still adding it though.

next-frontend/src/app/(wca)/persons/[wcaId]/page.tsx
Comment (FinnIckler) [reply next-frontend/src/app/(wca)/persons/[wcaId]/page.tsx]: yeah it wasn't that hard

next-frontend/src/lib/dates/format.ts
FinnIckler: I will just port the function we are using currently which uses luxon

next-frontend/src/components/persons/ResultsTab.tsx
```diff
@@ -1,18 +1,37 @@
...
 const ResultsTab: React.FC<ResultsTabProps> = ({ wcaId }) => {
-  return (
-    <>
-      <Heading>Results</Heading>
-      <ResultsTable wcaId={wcaId} />
-    </>
+  const api = useAPI();
+  const { t } = useT();
+
+  const [eventId] = useState("333");
```
gregorbg: Huh? A `useState` without setter is _very_ fishy...

app/controllers/api/v0/persons_controller.rb
gregorbg: Strong bias towards my Rails-y version, simply because it is easier to refactor and maintain. It's "type-safer" than raw strings (Ruby isn't exactly type safe in the first place, but this declarative approach is at least _relatively_ better than hard-coded strings)

next-frontend/src/components/persons/ResultsTab.tsx
FinnIckler: yeah this is still from the slate code, I think this was supposed to simulate the event picker? I will add an actual one

next-frontend/src/components/persons/PersonalRecordsTable.tsx
```diff
@@ -4,44 +4,16 @@ import { Tooltip } from "@/components/ui/tooltip";
...
-  single: string;
-  average: string;
-  anr: number;
-  acr: number;
-  awr: number;
-}
+import { components } from "@/types/openapi";
+import events from "@/lib/wca/data/events";
+import { formatAttemptResult } from "@/lib/wca/wcif/attempts";
+import _ from "lodash";
```
gregorbg: You still have this (now) unused import flying around

PR #12879, opened 2025-11-28
Add `H2H` format display

dunkOnIT:
- Adds "H2H" in the `Format` column of the table
- Removes "Head-to-Head" from the round name, per Ruben's request in Slack DM's
- Also refactors the table rendering/H2H mocking functions to make the code a little cleaner.

Github makes the diff look much more dramatic than it is because of indentation changes in the table body.

image

### DIFF
--- a/app/webpacker/components/EventsTable/index.jsx
+++ b/app/webpacker/components/EventsTable/index.jsx
@@ -65,9 +65,7 @@ export default function EventsTable({ competitionInfo, wcifEvents }) {
                     {events.byId[event.id].name}
                   </TableCell>
                 )}
-                <TableCell>
-                  <TableCell>{I18n.t(`rounds.${roundTypeId}.cell_name`)}</TableCell>
-                </TableCell>
+                <TableCell>{I18n.t(`rounds.${roundTypeId}.cell_name`)}</TableCell>
                 <TableCell>
                   {round.cutoff && `${formats.byId[round.cutoff.numberOfAttempts].shortName} / `}
                   {isH2hFinal ? I18n.t('formats.short.h') : formats.byId[round.format].shortName}


config/locales/en.yml
```diff
@@ -97,14 +97,14 @@ en:
     "3": "Best of 3"
     "a": "Average of 5"
     "m": "Mean of 3"
-    "h": "Head-to-Head"
```
gregorbg, line 100: Why does this have to go? The round name is still supposed to be displayed as "Head to Head Final", is it not?

dunkOnIT: No, it isn't - Ruben requested that we just leave it as "Final"

app/webpacker/components/EventsTable/index.jsx
```diff
@@ -64,51 +54,56 @@ export default function EventsTable({ competitionInfo, wcifEvents }) {
...
+                  <TableCell rowSpan={event.rounds.length}>
+                    {events.byId[event.id].name}
+                  </TableCell>
                 )}
-              </TableCell>
-              {competitionInfo['uses_cutoff?'] && (
                 <TableCell>
-                  {round.cutoff
-                    && cutoffToString(round)}
+                  <TableCell>{I18n.t(`rounds.${roundTypeId}.cell_name`)}</TableCell>
```
gregorbg: In my diff it looks like you've wrapped a TableCell within a TableCell, which does not make much sense. What is going on here?

gregorbg: Image

gregorbg (changes requested):

app/webpacker/components/EventsTable/index.jsx
dunkOnIT: Fixed, thank you! Brought in the original line from the old diff, not realizing that I'd since made that cell definition

PR #12910, opened 2025-12-01
Nextjs competition tabs

FinnIckler:
What am I doing wrong when using the Markdown component? 
image

It seems like it half works, but stuff like tables don't render

### DIFF
--- a/next-frontend/openapi/schemas/competitions/Tabs.yaml
+++ b/next-frontend/openapi/schemas/competitions/Tabs.yaml
@@ -3,6 +3,7 @@ required:
   - id
   - competition_id
   - content
+  - name
   - display_order
 properties:
   id:
--- a/next-frontend/src/components/competitions/TabMenu.tsx
+++ b/next-frontend/src/components/competitions/TabMenu.tsx
@@ -67,7 +67,10 @@ export default function TabMenu({
             <Link
               href={route({
                 pathname: "/competitions/[competitionId]/tabs/[tabName]",
-                query: { competitionId: competitionInfo.id, tabName },
+                query: {
+                  competitionId: competitionInfo.id,
+                  tabName: encodeURIComponent(tabName),
+                },
               })}
             >
               <Text maxW="44">{tabName}</Text>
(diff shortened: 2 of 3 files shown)


next-frontend/openapi/schemas/competitions/Tabs.yaml
```diff
@@ -0,0 +1,17 @@
+type: object
+required:
+  - id
+  - competition_id
+  - content
+  - display_order
```
gregorbg, line 7: Why is `name` not required?

next-frontend/src/app/(wca)/competitions/[competitionId]/tabs/[tabName]/loading.tsx
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/tabs/[tabName]/loading.tsx:1]: What is this file being used for?

Comment (gregorbg) [reply next-frontend/src/app/(wca)/competitions/[competitionId]/tabs/[tabName]/loading.tsx:1]: I mean, I know it's a NextJS-dedicated loading page, but why? Why are we only introducing it now that the custom tabs are being rendered? Why was it not necessary before, why is it suddenly necessary now?

next-frontend/openapi/schemas/competitions/Tabs.yaml
FinnIckler: I just forgot it

next-frontend/src/app/(wca)/competitions/[competitionId]/tabs/[tabName]/loading.tsx
Comment (FinnIckler) [reply next-frontend/src/app/(wca)/competitions/[competitionId]/tabs/[tabName]/loading.tsx:1]: We are actually using this page in all Tabs! You need it when you have a multi layered server side rendered page like this. When I click on a tab in the menu on the left, this tells next what to render while it loads (because we don't have access to a Loading boolean)

next-frontend/src/components/competitions/TabMenu.tsx
```diff
@@ -57,15 +58,25 @@ export default function TabMenu({
...
+            <Link
+              href={route({
+                pathname: "/competitions/[competitionId]/tabs/[tabName]",
+                query: {
+                  competitionId: competitionInfo.id,
+                  tabName: encodeURIComponent(tabName),
+                },
+              })}
+            >
+              <Text maxW="44">{tabName}</Text>
```
gregorbg, line 76: The `44` seems a bit arbitrary, can you use any tokens like `sm` or `xs` instead?

FinnIckler: well the smallest token there is is `3xs` which is 14rm. `44` is 11rem though and it doesn't look good with `3xs`

PR #12918, opened 2025-12-02
Scrollable delegate list in next

FinnIckler:
image

My idea to solve the overflow issue with too many delegates. Thoughts? Maybe we should do the same with the organizers

### DIFF
--- a/next-frontend/src/components/competitions/OrganizerCard.tsx
+++ b/next-frontend/src/components/competitions/OrganizerCard.tsx
@@ -15,13 +15,11 @@ import { MarkdownProse } from "@/components/Markdown";
 import WcaDocsIcon from "@/components/icons/WcaDocsIcon";
 import { LuChevronLeft, LuChevronRight } from "react-icons/lu";
 
-function OrganizerCarousel({
-  organizers,
-}: {
-  organizers:
-    | components["schemas"]["Person"][]
-    | components["schemas"]["Organizer"][];
-}) {
+type OrganizerPerson =
+  | components["schemas"]["Person"]
+  | components["schemas"]["Organizer"];
+
+function OrganizerCarousel({ organizers }: { organizers: OrganizerPerson[] }) {
   return (
     <Carousel.Root
       orientation="horizontal"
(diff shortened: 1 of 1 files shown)


FinnIckler:
I am?
https://github.com/thewca/worldcubeassociation.org/pull/12918/files#diff-a1e381d616556f0096cb981464934f394d3d198d29272716d10d3bf17972d389R30

next-frontend/src/components/competitions/OrganizerCard.tsx
```diff
@@ -0,0 +1,117 @@
...
+import { MarkdownProse } from "@/components/Markdown";
+import WcaDocsIcon from "@/components/icons/WcaDocsIcon";
+import { LuChevronLeft, LuChevronRight } from "react-icons/lu";
+
+function OrganizerCarousel({
+  organizers,
+}: {
+  organizers:
+    | components["schemas"]["Person"][]
+    | components["schemas"]["Organizer"][];
```
gregorbg: I think defining a (local, ie non-exported) type alias like `type OrganizerPerson = Person | Organizer` and then doing `OrganizerPerson[]` here is more intuitive.

gregorbg: In other words, do the union `|` first, then perform _one_ `[]` on the union. As opposed to union-ing arrays.

next-frontend/src/components/competitions/OrganizerCard.tsx
```diff
@@ -0,0 +1,117 @@
...
+            <LuChevronLeft />
+          </IconButton>
+        </Carousel.PrevTrigger>
+
+        <Carousel.NextTrigger asChild>
+          <IconButton size="xs" variant="ghost">
+            <LuChevronRight />
+          </IconButton>
+        </Carousel.NextTrigger>
+      </Carousel.Control>
```
gregorbg, line 61: Have you tried composing the arrows directly into the "main view", like here: https://www.chakra-ui.com/docs/components/carousel#arrows ?

gregorbg: The idea is to fit the navigation arrows direclty left and right of the main view, instead of squeezing then _under_ there.

gregorbg (changes requested):

PR #12921, opened 2025-12-03
Nextjs Map component

### LINKED ISSUE #15410: Add Map functionality to /competitions

FinnIckler:
image
image


FinnIckler:
image

I did some exploratory work with map-gl-libre. It's easy to use, but I think using Chakra Components directly on the map is making it really slow. This might be fixable by using another Layer, memo-ing it or just only sticking to the build in components for Marker and Popup

FinnIckler:
there are types generated in `.next/types` though? Which are definitely used to infer types at runtime in our setup

FinnIckler:
Screenshot 2026-08-20 at 16 27 56

I switched to a Map GL layer

FinnIckler:
@gregorbg, the MapGL Layer did fix the performance issues

next-frontend/src/components/map/Map.tsx
```diff
@@ -0,0 +1,182 @@
...
+const PIN_LAYER_ID = "competition-pins";
+const PIN_ICON_ID = "competition-pin";
+const PIN_ICON_PIXEL_RATIO = 2;
+const PIN_ICON_SIZE = 64;
+const PIN_DISPLAY_SIZE = PIN_ICON_SIZE / PIN_ICON_PIXEL_RATIO;
+
+// Lucide's `map-pin` as a filled silhouette, with the centre hole cut out by
+//   `evenodd`. maplibre loads it as an SDF, which is what lets one image be
+//   recoloured per competition by the `icon-color` expression below.
+const PIN_ICON_SVG = `<svg xmlns="http://www.w3.org/2000/svg" width="${PIN_ICON_SIZE}" height="${PIN_ICON_SIZE}" viewBox="0 0 24 24"><path fill="#000" fill-rule="evenodd" d="M20 10c0 4.993-5.539 10.193-7.399 11.799a1 1 0 0 1-1.202 0C9.539 20.193 4 14.993 4 10a8 8 0 0 1 16 0Zm-5 0a3 3 0 1 0-6 0 3 3 0 0 0 6 0Z"/></svg>`;
```
gregorbg, line 45: Okay for now, I would still like to (later ™️ ) think about whether we can somehow dynamically retrieve this from `LuMapPin` or something

next-frontend/src/components/map/Map.tsx
```diff
@@ -0,0 +1,182 @@
...
+  const router = useRouter();
+  const [hoveredCompetitionId, setHoveredCompetitionId] = useState<string>();
+
+  // maplibre paints in WebGL and cannot resolve the CSS variables that Chakra
+  //   tokens compile to, so the pin palette has to be read as literal colours.
+  //   That rules out the `solid` tokens, which resolve to a variable reference,
+  //   so these name the rungs that `brandSolid` points red and blue at.
+  const [upcomingColor, pastColor, haloColor] = useToken("colors", [
+    "red.600",
+    "blue.700",
```
gregorbg, line 57: I would not be surprised if Chakra has a kind of "resolve this" hook. But who knows, not worth to put in the effort rn

next-frontend/src/components/persons/CompetitionsTab.tsx
```diff
@@ -30,7 +30,7 @@ const CompetitionsTab: React.FC<CompetitionsTabProps> = ({ wcaId }) => {
   }
 
   if (!competitionQuery) {
-    return <Text>Failed fetching results</Text>;
+    return <Text>Failed fetching competitions</Text>;
```
gregorbg: Can I haz `OpenApiError` component?

next-frontend/src/components/persons/MapTab.tsx
```diff
@@ -1,12 +1,35 @@
...
+    },
   );
+
+  if (isLoading) {
+    return <Loading />;
+  }
+
+  if (!competitionQuery) {
+    return <Text>Failed fetching competitions</Text>;
+  }
```
gregorbg: Same as above: Use standardized error component

next-frontend/src/components/persons/MapTab.tsx
```diff
@@ -1,12 +1,35 @@
+"use client";
```
gregorbg: Why not make this an async server component? Instead of `useAPI`, you can just fetch from the URL directly and then I see no other hooks

gregorbg (changes requested):

next-frontend/src/components/persons/MapTab.tsx
FinnIckler: done

next-frontend/src/components/persons/CompetitionsTab.tsx
FinnIckler: I got you

gregorbg (approved):

PR #12925, opened 2025-12-04
Set the max width of tables to 100% of container in competition tabs

FinnIckler:
This fixes the tables in the tabs not taking the full width. For the competitor tab, this happens only if there are enough events/long competitor names to fill that space. This is a separate issue which I will fix separately


gregorbg (commented):
Chakra has an alias `full` which I feel like is more expressive

PR #12926, opened 2025-12-04
Wrap Event List on Competition Page

FinnIckler:
image

### DIFF
--- a/next-frontend/src/components/competitions/Cards.tsx
+++ b/next-frontend/src/components/competitions/Cards.tsx
@@ -10,7 +10,6 @@ import {
   Stat,
   Badge,
   Wrap,
-  WrapItem,
 } from "@chakra-ui/react";
 import BookmarkIcon from "@/components/icons/BookmarkIcon";
 import CompRegoOpenDateIcon from "@/components/icons/CompRegoOpenDateIcon";


next-frontend/src/components/competitions/Cards.tsx
```diff
@@ -352,20 +354,20 @@ export function EventCard({
...
         <Card.Title textStyle="s4">Events List</Card.Title>
-        <HStack gap="4">
+        <Wrap gap="4">
           {competitionInfo.event_ids.map((event_id) => (
             <EventIcon
               key={event_id}
               eventId={event_id}
               boxSize="8"
               color={
-                event_id === competitionInfo.main_event_id && event_id !== "333"
```
gregorbg, line 362: Also, why did you remove this condition? Is it part of the bug observations spreadsheet? And if yes, why? What's the point of marking 3x3 green for ~95% of our competitions?

FinnIckler: It isn't, I just thought it would be confusing if no main event was listed at all

gregorbg: 333 is a very fair default assumption.

gregorbg: 333 is a very fair default assumption.

next-frontend/src/components/competitions/Cards.tsx
```diff
@@ -9,6 +9,8 @@ import {
   HStack,
   Stat,
   Badge,
+  Wrap,
+  WrapItem,
```
gregorbg: Now this import is unused, as the CI will undoubtedly tell you

PR #12927, opened 2025-12-04
Give Competition Images a max size

FinnIckler:
image
image
Also moves the image card next to the venue details as per bug sheet

### DIFF
--- a/next-frontend/src/components/competitions/Cards.tsx
+++ b/next-frontend/src/components/competitions/Cards.tsx
@@ -130,6 +130,7 @@ export function AdditionalInformationCard({
         <MarkdownProse
           as={Card.Description}
           content={competitionInfo.information}
+          imageProps={{ maxW: "sm" }}
           textStyle="body"
         />
       </Card.Body>


next-frontend/src/components/Markdown.tsx
```diff
@@ -25,6 +25,7 @@ export const MarkdownProse: PolymorphicComponent<
               {children}
             </ChakraLink>
           ),
+          img: (img) => <Image src={img.src} maxW="sm" alt={img.alt} />,
```
gregorbg: Globally implementing this for all Markdown parsed sections is not a good idea imo. What if you give the whole Markdown field (at the place where you're inserting it) a `maxWidth`?

FinnIckler: That will not solve the problem. I want the text to still fill all the space, but just the images to have a max size

FinnIckler: maybe as an option in the markdown component?

gregorbg: Yup. Mind if I hijack your PR for just one commit? I know the Chakra type system quite well

FinnIckler: go ahead!

gregorbg: There you go 😉

PR #12928, opened 2025-12-04
Fix Podium Table

FinnIckler:
Adds rounded corners to the tables and handles events without averages.

### DIFF
--- a/next-frontend/src/components/results/ResultsTable.tsx
+++ b/next-frontend/src/components/results/ResultsTable.tsx
@@ -25,6 +25,7 @@ export function ResultsTable({
   const event = events.byId[eventId];
 
   const solveCount = event.recommendedFormat.expected_solve_count;
+  const anyAverages = results.some((r) => r.average !== 0);
 
   return (
     <Table.ScrollArea rounded="md" maxW="full">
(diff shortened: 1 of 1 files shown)


next-frontend/src/components/results/ResultsTable.tsx
```diff
@@ -27,67 +27,71 @@ export function ResultsTable({
...
-      </Table.Header>
+    <Table.ScrollArea rounded="md" maxW="full">
+      <Table.Root>
+        <Table.Header>
+          <Table.Row>
+            <Table.ColumnHeader>#</Table.ColumnHeader>
+            {isAdmin && <Table.ColumnHeader>Edit</Table.ColumnHeader>}
+            <Table.ColumnHeader>Competitor</Table.ColumnHeader>
+            <Table.ColumnHeader>Best</Table.ColumnHeader>
+            {solveCount > 2 && <Table.ColumnHeader>Average</Table.ColumnHeader>}
```
gregorbg: I feel like `average != 0` is the semantically more sound check

next-frontend/src/components/results/ResultsTable.tsx
```diff
@@ -27,67 +27,71 @@ export function ResultsTable({
...
+                    })}
+                  >
+                    {competitorResult.name}
+                  </Link>
+                </Table.Cell>
+                <Table.Cell style={{ position: "relative" }}>
+                  {formatAttemptResult(competitorResult.best, eventId)}{" "}
+                  {recordTagBadge(competitorResult.regional_single_record)}
+                </Table.Cell>
+                {solveCount > 2 && (
```
gregorbg: See above about good check conditions

next-frontend/src/components/results/ResultsTable.tsx
gregorbg: If you don't have any concrete results that you can check in the header, do some `listOfResults.some()` shenanigans to see if _any_ result in your currently displayed dataset has an average

gregorbg: (Just for reference: There are competitions that hold more than one MultiBLD attempt, i.e. they use `Best of 2` or `Best of 3` format. These will have an `expected_solve_count` larger than 1 but still no average to be displayed)

gregorbg (changes requested):

PR #12930, opened 2025-12-04
Add Title Metadata to each page

FinnIckler:
There were 2 titles that were hardcoded in Rails currently and are not localized.
I could add them.

### DIFF
--- a/next-frontend/src/app/(wca)/competitions/[competitionId]/layout.tsx
+++ b/next-frontend/src/app/(wca)/competitions/[competitionId]/layout.tsx
@@ -13,8 +13,13 @@ export async function generateMetadata({
 }: TitleProps): Promise<Metadata> {
   const { competitionId } = await params;
 
+  const { data: competitionInfo, error } =
+    await getCompetitionInfo(competitionId);
+
+  if (error || !competitionInfo) return { title: "Competition Not Found" };
+
   return {
-    title: `${competitionId}`,
+    title: `${competitionInfo.name}`,
   };
 }
 


next-frontend/src/app/(wca)/persons/[wcaId]/page.tsx
```diff
@@ -12,6 +12,25 @@ import MapTab from "@/components/persons/MapTab";
...
+type TitleProps = {
+  params: Promise<{ wcaId: string }>;
+};
+
+export async function generateMetadata({
+  params,
+}: TitleProps): Promise<Metadata> {
+  const { wcaId } = await params;
+
+  const { data: personDetails, error } = await getResultsByPerson(wcaId);
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/persons/[wcaId]/page.tsx]: Do you have to use the endpoint that returns a whole bunch of results, just so that you can figure out the name of the person?

next-frontend/src/app/(wca)/layout.tsx
```diff
@@ -9,8 +9,10 @@ import RandomBackground from "@/components/RandomBackground";
 import { Rubik } from "next/font/google";
 
 export const metadata: Metadata = {
-  title: "WCA Website",
-  description: "WST x SLATE",
+  title: {
+    template: "%s | World Cube Association",
+    default: "",
```
gregorbg, line 14: What does this `default` key do?

next-frontend/src/app/(wca)/competitions/[competitionId]/layout.tsx
```diff
@@ -2,6 +2,21 @@ import { Container, Text } from "@chakra-ui/react";
...
+  params: Promise<{ competitionId: string }>;
+};
+
+export async function generateMetadata({
+  params,
+}: TitleProps): Promise<Metadata> {
+  const { competitionId } = await params;
+
+  return {
+    title: `${competitionId}`,
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/layout.tsx]: I don't know whether the competition ID is accurate enough to use as title. What about fetching the `competitionInfo` and using the actual name? NextJS should be smart enough to "recycle" the same fetch for the actual page body

gregorbg (changes requested):
Feel free to translate the previously hard-coded titles!

next-frontend/src/app/(wca)/persons/[wcaId]/page.tsx
Comment (FinnIckler) [reply next-frontend/src/app/(wca)/persons/[wcaId]/page.tsx]: oh, this method is named wrong. It actually gives the personInfo for displaying everything in the profile page

Comment (FinnIckler) [reply next-frontend/src/app/(wca)/persons/[wcaId]/page.tsx]: same method that is used in the persons page

next-frontend/src/app/(wca)/layout.tsx
FinnIckler: You have to set a default sadly. It would mean that the default title is "| World Cube Association". it's not ideal, but we are overwriting it everywhere

next-frontend/src/app/(wca)/competitions/[competitionId]/layout.tsx
Comment (FinnIckler) [reply next-frontend/src/app/(wca)/competitions/[competitionId]/layout.tsx]: yeah it is. good point

PR #12932, opened 2025-12-04
Use simple grid in delegates page

FinnIckler:
I don't know what I was trying to do with the first version of this, but this is so much cleaner
Screenshot 2025-12-04 at 20 38 05

There is definitely still a backend issue regarding which groups contain delegates.
Even on prod, trying to get all African Delegates with
https://www.worldcubeassociation.org/api/v0/user_roles?parentGroupId=8&isActive=true&sort=location%2Cname
does not lead to any results.

### DIFF
--- a/next-frontend/src/app/(wca)/delegates/page.tsx
+++ b/next-frontend/src/app/(wca)/delegates/page.tsx
@@ -1,5 +1,4 @@
 import {
-  Center,
   Container,
   Heading,
   Link,


next-frontend/src/app/(wca)/delegates/page.tsx
gregorbg: You have a CI warning 🙃

PR #12933, opened 2025-12-05
Split EventPicker into Single and Multi

FinnIckler:
Screenshot 2025-12-05 at 16 41 26
The single event picker can even wrap!

### DIFF
--- a/next-frontend/src/components/EventSelector.tsx
+++ b/next-frontend/src/components/EventSelector.tsx
@@ -74,6 +74,7 @@ export function SingleEventSelector({
               colorPalette="green"
               disabled={disabled}
               value={eventId}
+              maxW="16"
             >
               <RadioCard.ItemHiddenInput />
               <RadioCard.ItemControl>
@@ -106,8 +107,6 @@ export function MultiEventSelector({
   wrap = false,
 }: MultiEventSelectorProps) {
   const { t } = useT();
-
-  const Container = wrap ? HStack : Wrap;
 
   return (
     <Tooltip
@@ -153,7 +152,11 @@ export function MultiEventSelector({
             )}
           </ButtonGroup>
         </Fieldset.Legend>
-        <Container>
+        <CheckboxGroup
+          disabled={disabled}
+          flexDirection="row"
+          flexWrap={wrap ? "wrap" : undefined}
+        >
           {eventList.map((eventId) => {
             const currentEventSelected = selectedEvents.includes(eventId);
             const currentEventDisabled = eventsDisabled.includes(eventId);
@@ -173,6 +176,7 @@ export function MultiEventSelector({
                 size={eventButtonsCompact ? "sm" : undefined}
                 checked={currentEventSelected}
                 onCheckedChange={() => onEventClick(eventId)}
+                maxW="16"
               >
                 <CheckboxCard.HiddenInput />
                 <CheckboxCard.Control>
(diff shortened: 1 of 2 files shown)


next-frontend/src/components/EventSelector.tsx
```diff
@@ -31,7 +34,53 @@ interface EventSelectorProps {
...
+  return (
+    <RadioCard.Root
+      disabled={disabled}
+      orientation="vertical"
+      align="center"
+      value={selectedEvent}
+      onValueChange={(e) => onEventClick(e.value ?? "")}
+    >
+      {title && <RadioCard.Label>{title}</RadioCard.Label>}
+      <Wrap>
```
gregorbg: I am torn between "Oh nice" and "Hm, this forces the developer into a wrap situation even if they don't want to".

next-frontend/src/components/EventSelector.tsx
```diff
@@ -7,13 +7,16 @@ import {
   CheckboxCard,
   CheckboxGroup,
   Fieldset,
+  HStack,
```
gregorbg: You have CI warnings 🙃

next-frontend/src/components/EventSelector.tsx
```diff
@@ -31,7 +34,53 @@ interface EventSelectorProps {
...
+  onEventClick,
+  disabled,
+}: SingleEventSelectorProps) {
+  return (
+    <RadioCard.Root
+      disabled={disabled}
+      orientation="vertical"
+      align="center"
+      value={selectedEvent}
+      onValueChange={(e) => onEventClick(e.value ?? "")}
```
gregorbg: What about `e.value!`? What are situations where you can provoke a radio selection to be `undefined`?

next-frontend/src/components/EventSelector.tsx
```diff
@@ -31,7 +34,53 @@ interface EventSelectorProps {
...
 }
 
-export default function EventSelector({
+interface SingleEventSelectorProps {
+  title: string;
+  eventList?: string[];
+  selectedEvent: string;
+  onEventClick: (eventId: string) => void;
+  disabled?: boolean;
+  eventButtonsCompact?: boolean;
```
gregorbg, line 44: You're defining this prop but never using it. Try using it with the radio control buttons.

FinnIckler:
you could use the justify prop to fix that

next-frontend/src/components/EventSelector.tsx
```diff
@@ -49,17 +50,22 @@ export function SingleEventSelector({
   selectedEvent,
   onEventClick,
   disabled,
+  eventButtonsCompact = false,
+  wrap = false,
 }: SingleEventSelectorProps) {
+  const Container = wrap ? Fragment : Wrap;
```
gregorbg: Why `Fragment` and not `HStack` like before?
Should this also apply to `MultiEventSelector`?

FinnIckler: Changed to `HStack`. For `MultiEventSelector` that would need to replace `CheckboxGroup`, which I think should just work?

gregorbg: Oh, interesting. It might be that it "just works", but `CheckboxGroup` still feels like the cleaner alternative. Can you see if `CheckboxGroup` reacts amicably to `flexWrap` or similar properties?

PR #12935, opened 2025-12-05
Fix link color readability on dark backgrounds

gregorbg:
Randomly composed a tone with Gemini.

## Before
image

## After
image


gregorbg:
Yes, that's a Chakra convention.

PR #12964, opened 2025-12-08
Add register page skeleton to NextJS

gregorbg:
The internal data flow of the registration panel is quite complicated, so the submission logic needs to follow in a separate PR. Also, this alleviates us from the need of handling the `POST` traffic, the polling and the Stripe components for the beta tests.

image
image
image

### DIFF
--- a/next-frontend/src/app/(wca)/competitions/[competitionId]/register/page.tsx
+++ b/next-frontend/src/app/(wca)/competitions/[competitionId]/register/page.tsx
@@ -44,7 +44,6 @@ export default async function RegisterPage({
 
   const competitionInfo = competitionInfoResponse.data;
 
-  // @ts-expect-error TODO: Fix this
   const stepConfig = await fetchConfig(session.accessToken, competitionId);
 
   if (stepConfig.error) {
--- a/next-frontend/src/components/EventSelector.tsx
+++ b/next-frontend/src/components/EventSelector.tsx
@@ -11,6 +11,7 @@ import {
   VisuallyHidden,
   HStack,
   Wrap,
+  Stack,
 } from "@chakra-ui/react";
 import { useT } from "@/lib/i18n/useI18n";
 import { Tooltip } from "@/components/ui/tooltip";
@@ -72,7 +73,7 @@ export function SingleEventSelector({
 interface MultiEventSelectorProps {
   eventList?: string[];
   selectedEvents: string[];
-  onEventClick?: (eventId: string) => void;
+  onEventClick: (eventId: string) => void;
   disabled?: boolean;
   eventButtonsCompact?: boolean;
   maxEvents?: number;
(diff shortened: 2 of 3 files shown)


next-frontend/src/app/(wca)/competitions/[competitionId]/(results)/results/all/FilteredResults.tsx
```diff
@@ -34,8 +34,6 @@ export default function FilteredResults({
         selectedEvents={[activeEventId]}
         onEventClick={setActiveEventId}
         eventList={competitionInfo.event_ids}
-        hideAllButton
```
Comment (FinnIckler) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/(results)/results/all/FilteredResults.tsx]: unrelated?

next-frontend/src/app/(wca)/competitions/[competitionId]/register/StepPanel.tsx
```diff
@@ -0,0 +1,206 @@
...
+      <Steps.CompletedContent>
+        <StepSummary
+          form={registrationForm}
+          competitionInfo={competitionInfo}
+        />
+      </Steps.CompletedContent>
+
+      <ButtonGroup size="sm" variant="surface" asChild>
+        <Group grow>
+          {/*
```
Comment (FinnIckler) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/register/StepPanel.tsx:165]: commented out

next-frontend/src/app/(wca)/competitions/[competitionId]/(results)/results/all/FilteredResults.tsx
Comment (FinnIckler) [reply next-frontend/src/app/(wca)/competitions/[competitionId]/(results)/results/all/FilteredResults.tsx]: maybe move these changes to a separate PR?

next-frontend/src/app/(wca)/competitions/[competitionId]/register/StepPanel.tsx
Comment (gregorbg) [reply next-frontend/src/app/(wca)/competitions/[competitionId]/register/StepPanel.tsx:165]: Intentionally. I wanted to keep a note of how a back button _would_ look in principle, but we currently don't need one.

next-frontend/src/app/(wca)/competitions/[competitionId]/(results)/results/all/FilteredResults.tsx
Comment (gregorbg) [reply next-frontend/src/app/(wca)/competitions/[competitionId]/(results)/results/all/FilteredResults.tsx]: Not quite. But it is somewhat contingent on your https://github.com/thewca/worldcubeassociation.org/pull/12933.

The idea here is that in "my" refactor as part of the current PR, The hide/show property of the buttons is derived based on "do you even pass a `onHideClick` handler?". In other words, the selector internally determines `hideAllButton = onHideClick !== undefined`.

Comment (gregorbg) [reply next-frontend/src/app/(wca)/competitions/[competitionId]/(results)/results/all/FilteredResults.tsx]: I will admit though that the refactor contained in this PR is ~3 weeks old at this point.

FinnIckler (approved):

PR #12991, opened 2025-12-12
Move recompute methods into round.rb and introduce LiveResult.compute_average_and_best

FinnIckler:
for `compute_average_and_best` we could probably also set it using a hook every time it gets updated?

### DIFF
--- a/app/models/live_result.rb
+++ b/app/models/live_result.rb
@@ -47,16 +47,13 @@ class LiveResult < ApplicationRecord
 
   def self.compute_average_and_best(attempts, round)
     r = Result.new(
-      value1: attempts[0].result,
-      value2: attempts[1]&.result || 0,
-      value3: attempts[2]&.result || 0,
-      value4: attempts[3]&.result || 0,
-      value5: attempts[4]&.result || 0,
       event_id: round.event.id,
       round_type_id: round.round_type_id,
       round_id: round.id,
       format_id: round.format_id,
+      result_attempts: attempts.map(&:to_result_attempt),
     )
+
     [r.compute_correct_average, r.compute_correct_best]
   end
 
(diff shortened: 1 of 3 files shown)


app/models/round.rb
```diff
@@ -181,6 +181,115 @@ def accepted_registrations
...
+                  (lr.#{rank_by} <= 0) ASC,
+                  lr.#{rank_by} ASC
+                  #{", lr.#{secondary_rank_by} <= 0, lr.#{secondary_rank_by} ASC" if secondary_rank_by}
+              ) AS rownum
+            FROM live_results lr
+            WHERE lr.round_id IN (#{round_ids})
+              AND lr.best != 0
+        ) x
+        WHERE rownum = 1
+    ) AS person_best
```
gregorbg: Indentation around this area seems off (while writing this comment, I am not even sure which exact lines I should mark for comment)

app/models/live_result.rb
```diff
@@ -52,6 +45,21 @@ def best_possible_solve_times
     end
   end
 
+  def self.compute_average_and_best(attempts, round)
+    r = Result.new(
+      value1: attempts[0].result,
+      value2: attempts[1]&.result || 0,
+      value3: attempts[2]&.result || 0,
+      value4: attempts[3]&.result || 0,
+      value5: attempts[4]&.result || 0,
```
gregorbg: Can you leverage the new result_attempts structure here?

FinnIckler: not until the PR is merged that uses them in the `solve_time` method

FinnIckler: done that now

app/models/round.rb
gregorbg, line 1: I trust that the code here has only been _moved_ without changing anything. Currently I don't have the capacity to check every line here, so this review is granted based on trusting that you can c/p properly :P

PR #13015, opened 2025-12-15
Also slightly reduce RAM for staging

FinnIckler:
for some reason staging ram was already 20MB higher than production


FinnIckler:
Screenshot 2025-12-15 at 12 21 13
done, I think we might need one more deployment for the binpacker to kick in, but that'll happen naturally during the week

PR #13031, opened 2025-12-16
Fix wrong ISO 3166-1 code for Macau in database export readme

oyyq99999:
MC is for Monaco lol


gregorbg (approved):
Thanks! I have confirmed that it is actually `MO` in the data itself. So it was really just the README where we got it wrong.

PR #13032, opened 2025-12-16
Fix New Relic Instrumentation

FinnIckler:
This broke since I setup new relic and this is from the official new relic blog


FinnIckler:
Where did you see that? We are not using next start to start our server, because we are using the standalone output mode. See our docker file and the docker entrypoint

PR #13033, opened 2025-12-16
Store JSON when submitting via WCA Live stored results

gregorbg:
Turns out, we already had a pipeline for storing the JSONs that Delegates submit. (They get forwarded to WRT as an email attachment).

This PR connects the new "Use WCA Live results" feature to this existing pipeline, by inserting the authorized WCIF as "uploaded JSON". The point is that this contains all relevant information (persons, results) at the time of submission.

Scrambles are not included in this WCIF, but if you use "WCA Live stored results", then you *have* to upload scrambles to our new Scrambles Matcher anyways. They will be stored and persisted in there.


app/controllers/scramble_files_controller.rb
danieljames-dj, line 1: I'm not sure about the changes in this file (and the related scramble related files). So my review in these files will be limited.

PR #13035, opened 2025-12-16
Use different font in development in next webpack config

coder13:
## Summary

Fix issue with prod fonts trying to be loaded in development.

image

This issue is fixed by loading a different font via the nextjs webpack config.

### DIFF
--- a/next-frontend/next.config.ts
+++ b/next-frontend/next.config.ts
@@ -26,6 +26,13 @@ const nextConfig: NextConfig = {
   },
   output: "standalone",
   productionBrowserSourceMaps: true,
+  webpack(config) {
+    if (!process.env.PROPRIETARY_FONT) {
+      config.resolve.alias["@/styles/fonts"] =
+        require("path").resolve(__dirname, "src/styles/fonts/empty-font.ts");
+    }
+    return config;
+  },
   async rewrites() {
     return [
       {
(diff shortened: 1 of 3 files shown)


next-frontend/next.config.ts
```diff
@@ -26,6 +26,13 @@ const nextConfig: NextConfig = {
   },
   output: "standalone",
   productionBrowserSourceMaps: true,
+  webpack(config) {
+    if (!process.env.PROPRIETARY_FONT) {
+      config.resolve.alias["@/styles/fonts"] =
+        require("path").resolve(__dirname, "src/styles/fonts/empty-font.ts");
```
FinnIckler, line 32: I assume this is not possible to do via import?

coder13: Are you referring specifically to `require("path")`. I hadn't thought of it, yeah should be able to change to import I think.

FinnIckler: yes, please use the import then. We should only be using `require` if it's needed for a commonjs module as per `@typescript-eslint/no-require-imports`

gregorbg: Kindly ping @coder13 :)

gregorbg: Pinging again @coder13, getting this config in would be a great help to de-clutter some of our CI scripts!

coder13: So sorry, I do **not** get github notifications. Please annoy me on slack if you need my quick response.

PR #13040, opened 2025-12-17
Wca live next podiums page

FinnIckler:
Screenshot 2025-12-17 at 11 45 38

### DIFF
--- a/next-frontend/src/app/(wca)/competitions/[competitionId]/live/competitors/[registrationId]/page.tsx
+++ b/next-frontend/src/app/(wca)/competitions/[competitionId]/live/competitors/[registrationId]/page.tsx
@@ -2,7 +2,7 @@ import { Container, Heading, Link, Table } from "@chakra-ui/react";
 import { getResultByPerson } from "@/lib/wca/live/getResultByPerson";
 import _ from "lodash";
 import events from "@/lib/wca/data/events";
-import { rankingCellColour } from "@/components/live/LiveResultsTable";
+import { rankingCellColorPalette } from "@/components/live/LiveResultsTable";
 import { formatAttemptResult } from "@/lib/wca/wcif/attempts";
 import { Fragment } from "react";
 
@@ -67,7 +67,8 @@ export default async function PersonResults({
                     </Table.Cell>
                     <Table.Cell
                       width={1}
-                      backgroundColor={rankingCellColour(result)}
+                      layerStyle="fill.deep"
+                      colorPalette={rankingCellColorPalette(result)}
                     >
                       {global_pos}
                     </Table.Cell>


gregorbg:
> There is already an advancing questionable definition in the chakra theme so we should be able to just change the font colour there too

The problem with the colors is that you're "only" overriding the background color and nothing else.

Chakra is pretty good at theming and contrast colors. But it can only do that for you if you allow it to make use of the full color palette. Your current approach is "definitely use this one hex color, don't bother about anything else". So Chakra doesn't have a chance to also pick a matching font color or black/white contrast modes.

I think a better approach is to change the `rankingCellColor` method from a specific color to rather return a color *scheme* like `green` or `yellow`. You can then choose a layer style according to https://www.chakra-ui.com/docs/theming/layer-styles, for example `layerStyle="fill.muted"`. This in combination with `colorPalette="green"` will Chakra to choose a matching tone and contrast colors as defined in our streamlined theme.

For example, this is how `layerStyle="fill.muted"` looks like:
image

FinnIckler:
fill.deep would be the closest to the existing live colours, but these all look good to me

next-frontend/src/app/(wca)/competitions/[competitionId]/live/podiums/page.tsx
```diff
@@ -0,0 +1,57 @@
...
+}: {
+  params: Promise<{ competitionId: string }>;
+}) {
+  const { competitionId } = await params;
+
+  const { error: podiumError, data: rounds } =
+    await getLivePodiums(competitionId);
+
+  if (podiumError) {
+    return <Text>Error fetching Podiums</Text>;
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/live/podiums/page.tsx]: This can probably use the recently created openapi error component

next-frontend/src/app/(wca)/competitions/[competitionId]/live/podiums/page.tsx
```diff
@@ -0,0 +1,57 @@
...
+    (r) => parseActivityCode(r.id).eventId,
+  );
+
+  return (
+    <Container>
+      <Heading textStyle="h1">Podiums</Heading>
+      {WCA_EVENT_IDS.map((e) => {
+        const rounds = roundsByEventId[e];
+        if (!rounds) return;
+        const finalRound = rounds[0];
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/live/podiums/page.tsx]: Why is the _first_ element of the list also the final round?

next-frontend/src/app/(wca)/competitions/[competitionId]/live/podiums/page.tsx
```diff
@@ -0,0 +1,57 @@
...
+        const eventId = parseActivityCode(finalRound.id).eventId;
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/live/podiums/page.tsx]: Shouldn't this always be equal to `e` from the `map` loop above?

next-frontend/src/app/(wca)/competitions/[competitionId]/live/podiums/page.tsx
```diff
@@ -0,0 +1,57 @@
...
+            {finalRound.results.length > 0 ? (
+              <LiveResultsTable
+                results={finalRound.results}
+                competitionId={competitionId}
+                competitors={finalRound.competitors}
+                eventId={eventId}
+                showEmpty={false}
+              />
+            ) : (
+              "Podiums to be determined"
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/live/podiums/page.tsx]: This should be localized properly

next-frontend/src/components/live/LiveResultsTable.tsx
```diff
@@ -92,6 +92,7 @@ export default function ResultsTable({
             <Table.Row key={competitor.id}>
               <Table.Cell
                 width={1}
+                layerStyle="fill.deep"
```
gregorbg, line 95: This needs to also reference a `colorPalette` (instead of `backgroundColor` below) as discussed.

next-frontend/src/app/(wca)/competitions/[competitionId]/live/podiums/page.tsx
Comment (FinnIckler) [reply next-frontend/src/app/(wca)/competitions/[competitionId]/live/podiums/page.tsx]: the list only has one element. It should actually just be an index by

PR #13041, opened 2025-12-17
add back instrumentation.ts

FinnIckler:
We actually still need this because otherwise nextjs will optimize the newrelic import away because it's not used anywhere


FinnIckler:
Another solution would be to manually copy over the newrelic folder into the node_modules as part of the dockerfile, but I think this is a cleaner way?

PR #13044, opened 2025-12-17
Live next competitor page

FinnIckler:
Basically enables a "live" mode for the competitor table that links to the live results. We should probably add a search bar like in the current wca live page?

### DIFF
--- a/next-frontend/src/app/(wca)/competitions/[competitionId]/competitors/page.tsx
+++ b/next-frontend/src/app/(wca)/competitions/[competitionId]/competitors/page.tsx
@@ -25,7 +25,10 @@ export default async function Competitors({
   return (
     <TabCompetitors
       id={competitionId}
-      isLive={hasPassed(competitionInfo.start_date)}
+      isLive={
+        hasPassed(competitionInfo.start_date) &&
+        !hasPassed(competitionInfo.end_date)
+      }
     />
   );
 }


next-frontend/src/app/(wca)/competitions/[competitionId]/competitors/page.tsx
```diff
@@ -1,11 +1,23 @@
...
 }: {
   params: Promise<{ competitionId: string }>;
 }) {
   const { competitionId } = await params;
 
-  return <TabCompetitors id={competitionId} />;
+  const { data: competitionInfo, error } =
+    await getCompetitionInfo(competitionId);
+
+  if (error || !competitionInfo) return { title: "Competition Not Found" };
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/competitors/page.tsx]: This looks like a whoopsie

next-frontend/src/app/(wca)/competitions/[competitionId]/competitors/page.tsx
```diff
@@ -1,11 +1,23 @@
...
+
+  return (
+    <TabCompetitors
+      id={competitionId}
+      isLive={hasPassed(competitionInfo.start_date)}
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/competitors/page.tsx]: You should probably also check for the competition end date _not_ having passed as well

next-frontend/src/app/(wca)/competitions/[competitionId]/live/competitors/page.tsx
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/live/competitors/page.tsx:1]: Why are you introducing this new route as a pure redirect?
Is this an existing page that needs to be implemented to keep deep-link backwards compatibility?

Comment (FinnIckler) [reply next-frontend/src/app/(wca)/competitions/[competitionId]/live/competitors/page.tsx:1]: it is, but just on the wca live side so maybe we don't actually need it

next-frontend/src/app/(wca)/competitions/[competitionId]/competitors/page.tsx
Comment (FinnIckler) [reply next-frontend/src/app/(wca)/competitions/[competitionId]/competitors/page.tsx]: well.. the competitors tab is not shown if the competition end date has passed

Comment (gregorbg) [reply next-frontend/src/app/(wca)/competitions/[competitionId]/competitors/page.tsx]: That does not keep anyone from bookmarking a link to this page while it's still visible, and then trying to open it 10 days later

next-frontend/src/app/(wca)/competitions/[competitionId]/live/competitors/page.tsx
Comment (gregorbg) [reply next-frontend/src/app/(wca)/competitions/[competitionId]/live/competitors/page.tsx:1]: I would actually argue that we don't. Because the use of stable competition IDs is a new concept, WCA Live used to use integer IDs for that. So we're breaking any and all old links anyways.

PR #13045, opened 2025-12-17
Better error page in nextjs

FinnIckler:
I think I have moved all pages that was checking their errors over. 

The page looks like this:
Screenshot 2025-12-17 at 19 58 08

If we want actual nice formatting in the code block we would need to install one of those adapters https://chakra-ui.com/docs/components/code-block

### DIFF
--- a/next-frontend/src/app/(wca)/competitions/[competitionId]/(results)/layout.tsx
+++ b/next-frontend/src/app/(wca)/competitions/[competitionId]/(results)/layout.tsx
@@ -1,4 +1,4 @@
-import { GridItem, SimpleGrid, Text } from "@chakra-ui/react";
+import { GridItem, SimpleGrid } from "@chakra-ui/react";
 import { InfoCard } from "@/components/competitions/Cards";
 import { MarkdownFirstImage } from "@/components/MarkdownFirstImage";
 import { getCompetitionInfo } from "@/lib/wca/competitions/getCompetitionInfo";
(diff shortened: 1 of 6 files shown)


next-frontend/src/components/ui/errored.tsx
```diff
@@ -1,13 +1,65 @@
...
+                <Accordion.ItemTrigger>
+                  {t("errors.next_frontend.details")}
+
+                  <Accordion.ItemIndicator />
+                </Accordion.ItemTrigger>
+                <Accordion.ItemContent>
+                  <Accordion.ItemBody>
+                    <CodeBlock.Root
+                      meta={{ wordWrap: true }}
+                      code={JSON.stringify({
```
gregorbg: For now, just `JSON.stringify(yourStuff, null, 2)` should do I think. As a compromise.

next-frontend/src/components/ui/errored.tsx
```diff
@@ -1,13 +1,65 @@
...
+  Alert,
+  CodeBlock,
+  Container,
+  Float,
+  IconButton,
+} from "@chakra-ui/react";
+import { TFunction } from "i18next";
 
-export default function Errored({ error }: { error: string }) {
+export default function Errored({
```
gregorbg: That name is a bit generic for me. What about `OpenapiError` or something?

next-frontend/src/components/ui/openapiError.tsx
```diff
@@ -1,13 +1,65 @@
...
+} from "@chakra-ui/react";
+import { TFunction } from "i18next";
 
-export default function Errored({ error }: { error: string }) {
+export default function Errored({
+  t,
+  response,
+}: {
+  response: Response;
+  t: TFunction;
```
gregorbg, line 16: Is there a chance you can use the actual `error` that comes back from any Openapi call? The one that is mutually exclusive with `data`...

FinnIckler: We currently don't define any in the open API spec so they are all never. I could go and add them all to the spec though

PR #13049, opened 2025-12-18
Allow using staging as a backend for nextjs

FinnIckler:
I assume this won't deploy though because just importing `import { sqliteAdapter } from "@payloadcms/db-sqlite";` will lead the server to crash because of needed libs if I remember correctly. We fixed that with some dynamic imports last time, but that would mean we have to go to a async config file again.

Thoughts?


next-frontend/src/payload.config.ts
```diff
@@ -136,7 +137,9 @@ export default buildConfig({
   typescript: {
     outputFile: path.resolve(dirname, "types/payload.ts"),
   },
-  db: mongooseAdapter(dbOptions()),
+  db: process.env.DATABASE_URI
+    ? mongooseAdapter(dbOptions())
+    : sqliteAdapter({ client: { url: "file:./payload.db" } }),
```
gregorbg, line 143: I think this should toggle based on a dedicated env variable (perhaps `env.DATABASE_BACKEND === 'mongodb'` or even some form of `env.WCA_STAGING_BACKEND` being set to any truthy value?) instead of "guessing" whether `DATABASE_URI` is set or not.

If you do implement this change, then you can actually pass `env.DATABASE_URI` to `client.url` in the Sqlite adapter config (of course assuming that you pass it a sensible value in `dev:staging` runner in `package.json`)

next-frontend/package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
+    "dev:staging": "WCA_BACKEND_API_URL=\"https://staging.worldcubeassociation.org/api/\" NEXT_PUBLIC_WCA_FRONTEND_API_URL=\"https://staging.worldcubeassociation.org/api\" DATABASE_URI=\"\" next dev",
```
coder13, line 6: Did you forget to add `--webpack`?

PR #13055, opened 2025-12-19
Nextjs Live Homepage

FinnIckler:
Screenshot 2025-12-19 at 16 47 25

I do still need the round_id to correctly link. What should I do, also fetch rounds? Three fetches feels pretty over the top. Or should I just make a new endpoint

### DIFF
--- a/next-frontend/src/components/competitions/Schedule/LiveView.tsx
+++ b/next-frontend/src/components/competitions/Schedule/LiveView.tsx
@@ -28,7 +28,6 @@ import {
 } from "@/lib/wca/dates";
 import EventIcon from "@/components/EventIcon";
 import { route } from "nextjs-routes";
-import { TFunction } from "i18next";
 import { useT } from "@/lib/i18n/useI18n";
 
 interface LiveViewProps {
@@ -50,6 +49,13 @@ export default function LiveView({
   const browserTimezone = Intl.DateTimeFormat().resolvedOptions().timeZone;
   const [timeZone, setTimeZone] = useState(browserTimezone);
 
+  const collection = createListCollection({
+    items: [...timeZones, browserTimezone].map((t) => ({
+      value: t,
+      label: t,
+    })),
+  });
+
   const dates = getDatesBetweenInclusive(
     firstStartTime,
     lastStartTime,
@@ -64,21 +70,18 @@ export default function LiveView({
   return (
     <VStack align="left">
       <Select.Root
-        collection={createListCollection({
-          items: [...timeZones, browserTimezone].map((t) => ({
-            value: t,
-            label: t,
-          })),
-        })}
-        width="320px"
+        collection={collection}
+        width="3/12"
         value={[timeZone]}
         onValueChange={(e) => setTimeZone(e.value[0])}
       >
         <Select.HiddenSelect />
-        <Select.Label>Select Timezone</Select.Label>
+        <Select.Label>{t("competitions.schedule.time_zone")}</Select.Label>
         <Select.Control>
           <Select.Trigger>
-            <Select.ValueText placeholder="Select Timezone" />
+            <Select.ValueText
+              placeholder={t("competitions.schedule.time_zone")}
+            />
           </Select.Trigger>
           <Select.IndicatorGroup>
             <Select.Indicator />
@@ -87,9 +90,9 @@ export default function LiveView({
         <Portal>
           <Select.Positioner>
             <Select.Content>
-              {timeZones.map((timezone) => (
-                <Select.Item item={timezone} key={timezone}>
-                  {timezone}
+              {collection.items.map((timezone) => (
+                <Select.Item item={timezone} key={timezone.label}>
+                  {timezone.value}
                   <Select.ItemIndicator />
                 </Select.Item>
               ))}
@@ -144,8 +147,8 @@ export default function LiveView({
                         </Button>
                       </Card.Body>
                       <Card.Footer>
-                        {getSimpleTimeString(activity.startTime)} -{" "}
-                        {getSimpleTimeString(activity.endTime)}
+                        {getSimpleTimeString(activity.startTime, timeZone)} -{" "}
+                        {getSimpleTimeString(activity.endTime, timeZone)}
                       </Card.Footer>
                     </Card.Root>
                   );


next-frontend/src/app/(wca)/competitions/[competitionId]/live/page.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export default async function LiveOverview({
+  params,
+}: {
+  params: Promise<{ competitionId: string }>;
+}) {
+  const { competitionId } = await params;
+  const { t } = await getT();
+
+  const { error, data: wcifSchedule } = await getSchedule(competitionId);
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/live/page.tsx]: The `error` here should be aliased to `wcifScheduleError` for consistency.

next-frontend/src/app/(wca)/competitions/[competitionId]/live/page.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  if (error) {
+    return <Errored error={`${error.data.id} not found`} />;
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/live/page.tsx]: Are we absolutely certain that no matter what happens, the error always has the `.data.id` property? I feel like the answer to this question should definitely be No, so what was your thinking here?

Can you use the new(-ish) centralized openapi error component instead?

Comment (gregorbg) [reply next-frontend/src/app/(wca)/competitions/[competitionId]/live/page.tsx]: Wait, `Errored` doesn't even exist anymore on `main` :O

next-frontend/src/app/(wca)/competitions/[competitionId]/live/page.tsx
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  const { error: wcifEventsError, data: wcifEvents } =
+    await getEvents(competitionId);
+
+  if (wcifEventsError) {
+    return <Errored error={`${wcifEventsError.data.id} not found`} />;
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/live/page.tsx]: (same remarks as above)

next-frontend/src/components/competitions/Schedule/LiveView.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+  const dates = getDatesBetweenInclusive(
+    firstStartTime,
+    lastStartTime,
+    timeZone,
+  );
+
+  // Show the first date that has not passed, if all of them have, show the last date
+  const defaultDate =
+    dates.filter((d) => !hasPassed(d.toISO()!))[0] ?? dates[dates.length - 1];
```
gregorbg: The comment is helpful here, otherwise I'd be completely lost. But you can still consider moving `dates[dates.length - 1]` to a separate variable, especially given that you're already using it a few lines above

next-frontend/src/components/competitions/Schedule/LiveView.tsx
```diff
@@ -0,0 +1,106 @@
...
+                            pathname:
+                              "/competitions/[competitionId]/live/rounds/[roundId]",
+                            query: {
+                              competitionId,
+                              roundId: activity.id.toString(),
+                            },
+                          })}
+                        >
+                          <HStack>
+                            <EventIcon eventId={eventId} />
```
gregorbg: You can use `fontSize="2xl"` to make the icon more prominent

gregorbg (changes requested):

next-frontend/src/app/(wca)/competitions/[competitionId]/live/page.tsx
Comment (FinnIckler) [reply next-frontend/src/app/(wca)/competitions/[competitionId]/live/page.tsx]: the not found error does always have an id! But yes, I will move this to the new component

next-frontend/src/components/competitions/Schedule/LiveView.tsx
FinnIckler: above I am using `activities[activities.length - 1]`, but I can still put it as a separate variable for clarity

next-frontend/src/app/(wca)/competitions/[competitionId]/live/page.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+  const uniqueTimeZones = [
+    ...new Set(wcifSchedule.venues.map((venue) => venue.timezone)),
+  ];
+
+  return (
+    <LiveView
+      competitionId={competitionId}
+      activities={allActivitiesSorted}
+      timeZone={uniqueTimeZones[0]}
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/live/page.tsx]: Wild thought: Why not pass all available timezones to the component and offer a dropdown in the top-right corner of the UI?

next-frontend/src/components/competitions/Schedule/LiveView.tsx
```diff
@@ -0,0 +1,109 @@
...
+interface LiveViewProps {
+  timeZone: string;
+  competitionId: string;
+  activities: components["schemas"]["WcifActivity"][];
+  wcifEvents: components["schemas"]["WcifEvent"][];
+  t: TFunction;
+}
+
+export default function LiveView({
+  timeZone,
```
gregorbg: I just realized: This uses the first available venue timezone of the competition by default. Current WCA Live uses the browser's "home" timezone.

In my mind, the latter makes more sense because users browsing WCA Live want to know when they can spectate and expect live results from the position they're currently watching from.

As suggested above, you could offer a dropdown to pick from all available venue timezones, but this dropdown should definitely also have a "Home" option for the browser timezone, regardless of whether that's originally included in the venue timezones or not.

gregorbg (changes requested):
I just noticed something substantial about the display of the dates in the overview panel

next-frontend/src/components/competitions/Schedule/LiveView.tsx
```diff
@@ -0,0 +1,163 @@
+"use client";
```
gregorbg, line 1: I don't think you need to client-render the whole page just because of one dropdown.

Specifically, if you pull the part of the component related to the timezone dropdown into its own component, you can mark only that component as `use client` and leave the rest of the skeleton as server-side rendered.

(We might even think about sharing that "timezone picker" component with the schedule view later)

FinnIckler: I cannot, I need the timezone state in the actual schedule itself

FinnIckler: Because otherwise the schedule wouldn't change if I pick another timezone

PR #13057, opened 2025-12-22
Sanity Check through Rails

FinnIckler:
TODO:
- remove category from table and use the sanity_check_category table
- add all queries

### DIFF
--- a/app/jobs/sanity_check_category_job.rb
+++ b/app/jobs/sanity_check_category_job.rb
@@ -5,7 +5,7 @@ class SanityCheckCategoryJob < WcaCronjob
 
   def perform(sanity_check_category)
     sanity_check_category.sanity_checks.find_each do |sanity_check|
-      query_result = ActiveRecord::Base.connection.exec_query sanity_check.query
+      query_result = sanity_check.run_query
 
       sanity_check.sanity_check_results.create!(
         query_results: query_result,


db/migrate/20251222101010_create_sanity_check_statistics_table.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class CreateSanityCheckStatisticsTable < ActiveRecord::Migration[7.0]
+  def change
+    create_table :sanity_check_results do |t|
+      t.json :query_results, null: false
+      # FK to cronjob_statistics.name (string PK) TODO: add index
+      t.string :cronjob_statistic_name, null: false
```
gregorbg: `t.references` takes `type: :string` and `index: true` options.

Unfortunately, non-standard names (which you're using here: You're calling it `_name` instead of `_id`) are not supported by `t.references`. So you have to go with `t.references :cronjob_statistic` and live with the fact that it's gonna be `cronjob_statistic_id`. But in my view that's okay because the ID column over there is a string, so no need to worry.

If Rails somehow doesn't infer the table name from `cronjob_statistic` correctly, you can use `foreign_key: { to_table: :cronjob_statistics }` to point it in the right direction. You should try first without this override though, methinks Rails is smart enough for that.

app/controllers/admin_controller.rb
```diff
@@ -14,6 +14,16 @@ def merge_people
     @merge_people = MergePeople.new
   end
 
+  def sanity_check
+    @categories = SanityCheckCategory.all
+  end
+
+  def run_sanity_check
+    sanity_check_category = SanityCheckCategory.find(params.require(:sanity_check_category_id))
+    SanityCheckCategoryJob.perform_later({ name: sanity_check_category.name, category_id: sanity_check_category.id })
```
gregorbg: Why do you need to pass the name separately? Just pass the ID, when the cronjob runs it can by itself look up the category based on the ID param and then read its name on-the-fly

app/jobs/sanity_check_category_job.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class SanityCheckCategoryJob < WcaCronjob
+  def perform(options)
```
gregorbg: As per my comment above, I think you only need `category_id`.

Fun fact: Rails can correctly serialize whole AR models as args (internally, it serializes the model into a string representation that only encodes the ID, and then when the job is picked up it does `SomeModel.find(id_that_was_automagically_serialized)` under the hood)

More fun fact: You can pass in multiple args and Rails ActiveJob will deserialize them for you in the correct order. So `def perform(category_id, some_other_arg, foo_bar)` is possible and will "just work".

app/controllers/admin_controller.rb
gregorbg: Ah, I see now why you're passing the `name` separately, it's so that the runner statistics can pick up the ID. Let me think about a solution for that...

app/jobs/run_all_sanity_checks.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class RunAllSanityChecks < WcaCronjob
```
gregorbg: The name of any job class should end in `*Job`. May I suggest `AllSanityChecksJob`? The "run" is kinda implied because every job can be run

app/models/sanity_check_result.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class SanityCheckResult < ApplicationRecord
+  belongs_to :sanity_check
+  has_one :sanity_check_category, through: :sanity_check
+  has_many :sanity_check_exclusions, through: :sanity_check
+  belongs_to :cronjob_statistic, primary_key: :name, foreign_key: :cronjob_statistic_name
```
gregorbg: Prize question: Is this separate `belongs_to` (plus its underlying database column) even necessary in the first place?

Every sanity check result knows its category (through the corresponding sanity check) and the category can directly reference a `cronjob_statistic`.
Having this extra column which is (normally) always filled by the `sanity_check_category`'s name feels redundant (because the category is already well-defined)

db/migrate/20251222101010_create_sanity_check_statistics_table.rb
gregorbg: See also my other comment on whether this column is even necessary in the first place

app/models/sanity_check_result.rb
FinnIckler: oh true, I can just do a find by name

app/jobs/all_sanity_checks_job.rb
```diff
@@ -5,5 +5,9 @@ def perform
     SanityCheckCategory.find_each do |sanity_check_category|
       SanityCheckCategoryJob.perform_later(sanity_check_category)
     end
+    # Enqueue Results job after, these are run on a single thread one after another
+    SanityCheckCategory.select(:email_to).uniq.each do |email_to|
+      SanityCheckResultsJob.perform_later(email_to)
```
gregorbg: Why do you defer this to a new job (which, cumbersomly, needs to be run off the cronjob queue artificially) when you could just do `deliver_later` right here instead?

FinnIckler: that would put it on the mailer_queue which would not wait for the other jobs to finish?

app/models/sanity_check_result.rb
```diff
@@ -0,0 +1,19 @@
...
+class SanityCheckResult < ApplicationRecord
+  belongs_to :sanity_check
+  has_one :sanity_check_category, through: :sanity_check
+  has_many :sanity_check_exclusions, through: :sanity_check
+
+  delegate :topic, :comments, to: :sanity_check
+
+  def cronjob_statistic
+    CronjobStatistic.find(sanity_check_category.id)
+  end
```
gregorbg, line 18: As far as I understand, we query the CJ statistics by name and not by ID

gregorbg: This might be worth introducing a `def statistic_id` in the `SanityCheckCategory` model though and use that across the board (in the ActiveJob runners, here, in the mailers, etc.)

app/models/sanity_check_result.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  def results_without_exclusions
+    query_results.filter do |result|
+      !sanity_check_exclusions.exists?(exclusion: result.to_json)
```
gregorbg: Have you thought about the problem where the column order of existing sanity check exclusions is different in the Java adapter?

app/jobs/all_sanity_checks_job.rb
gregorbg: Wild thought: What about introducing a whole new "sanity check queue"?

app/models/sanity_check_result.rb
FinnIckler: yes. The solution Sebastien came up with is to just change all Select * queries to the correct order. Which I think is an easy win

app/jobs/sanity_check_results_job.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class SanityCheckResultsJob < ApplicationJob
+  # We want to make sure this runs not in parallel but after the sanity checks
+  QUEUE_NAME = :wca_jobs
```
gregorbg: Is overriding this specific class variable really enought?
I thought you needed some `ActiveJob` DSL like `enqueue_as :wca_jobs` or smth

gregorbg (changes requested):

app/models/sanity_check_result.rb
FinnIckler: oops, true

app/models/sanity_check_result.rb
gregorbg: Interesting definition of "easy". Will this fix the issue though? If there are already-existing exclusions for any sanity check, then they have a non-defined (not well-defined) column order, so you cannot do a raw JSON comparison...

FinnIckler: well there are actually only a few `select *` queries. These can just be replaced by the `Select columns, in, order, of, saved, in, exclusions`.
Let me try it out really quick

app/jobs/sanity_check_results_job.rb
FinnIckler: A new sanity check queue sounds like a good idea

app/models/sanity_check_result.rb
gregorbg: This idea still assumes that *all* exclusions thus far have the exact same column order.

Even if we go with your approach, and change all queries to a "well-defined" `SELECT` order: Who can guarantee that `.to_json` on a Ruby level will respect this order?

What I'm trying to say is: JSON objects do not have a defined order. That's just how JSON works. End of story. Any solution proposed so far ignores this fact and (on some level) just "hopes" that the order will be correct. I would like to think about solutions which do not fundamentally rely on a broken assumption.

app/models/sanity_check_exclusion.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  # We have to do a partial equals here, as historically some of the exclusions have less
+  # data than others
+  def partially_equals?(exclusion_hash, query_result_hash)
+    exclusion_hash.each do |key, value|
+      return false unless value.to_s == query_result_hash[key].to_s
+    end
+
+    true
```
gregorbg:
```suggestion
    exclusion_hash.all? { |key, value| value.to_s == query_result_hash[key]&.to_s }
```

gregorbg: Image

gregorbg (changes requested):

app/models/sanity_check_exclusion.rb
gregorbg: Sidenote, why do you need the `to_s`? If the answer is "because it's coming from the Java implementation" then I have good news: Java needs `toString` because it's strongly typed so it cares about casting the comparison operands to the same type. And `toString` is often just the "cheapest" way of doing that.
Ruby `==` doesn't care about the types of the operand.

app/jobs/sanity_check_results_job.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class SanityCheckResultsJob < ApplicationJob
+  # We want to make sure this runs not in parallel but after the sanity checks
+  queue_as :sanity_checks
```
gregorbg: You can reference `SanityCheckCategoryJob::QUEUE_NAME` here

gregorbg: Also, line break :D

app/models/sanity_check.rb
```diff
@@ -4,6 +4,9 @@ class SanityCheck < ApplicationRecord
   include StaticData
 
   belongs_to :sanity_check_category
+  has_many :sanity_check_results
+  has_many :sanity_check_exclusions
+  has_one :latest_result
```
gregorbg: This is a `belongs_to` 🙃

gregorbg: Also, this will need a `class_name: "SanityCheckResult"` to function properly, because Rails cannot infer from the name alone what the type of this property should be

app/models/sanity_check_category.rb
```diff
@@ -4,6 +4,8 @@ class SanityCheckCategory < ApplicationRecord
   include StaticData
 
   has_many :sanity_checks
+  has_many :sanity_check_results, through: :sanity_checks
+  has_many :latest_results, through: :sanity_checks
```
gregorbg, line 8: Not sure if this also needs a `class_name: "SanityCheckResults"` or whether it is enough to specify that on the original relation... Please try it out to see how this works

app/models/sanity_check_result.rb
```diff
@@ -0,0 +1,25 @@
...
+  delegate :topic, :comments, to: :sanity_check
+
+  after_create :update_latest_result
+
+  def update_latest_result
+    sanity_check.update(latest_result_id: self.id)
+  end
+
+  def cronjob_statistic
+    CronjobStatistic.find(sanity_check_category.name)
```
gregorbg: You should reuse the `parameterize.underscore` thingy from `folder_handle` here, or perhaps even create a new `parameterize.camelcase` version and expose it as `def statistic_name` in the category.
You can then reuse that method here as well as over in the method override of the cronjob

gregorbg (changes requested):

app/jobs/all_sanity_checks_job.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class AllSanityChecksJob < WcaCronjob
+  def perform
+    SanityCheckCategory.find_each do |sanity_check_category|
+      SanityCheckCategoryJob.perform_later(sanity_check_category)
+    end
+    # Enqueue Results job after, these are run on a single thread one after another
+    SanityCheckCategory.select(:email_to).uniq.each do |email_to|
```
gregorbg:
```suggestion
    SanityCheckCategory.distinct.pluck(:email_to).each do |email_to|
```

gregorbg: This executes one `SELECT DISTINCT email_to FROM ...`. Note that the order is critical here because `pluck` immediately resolves and returns a Ruby array. You have to pipe the `distinct` into the query chain before it gets "terminated" by `pluck`.

app/controllers/admin_controller.rb
```diff
@@ -14,6 +14,26 @@ def merge_people
...
+    SanityCheckCategoryJob.perform_later(sanity_check_category)
+    flash[:success] = "Sanity check job enqueued for category #{sanity_check_category.name}."
+    redirect_to sanity_check_path
+  end
+
+  def add_exclusion
+    sanity_check_id = params.require(:sanity_check_id)
+    exclusion = params.require(:exclusion_json)
+
+    SanityCheckExclusion.create(exclusion: exclusion, sanity_check_id: sanity_check_id)
```
gregorbg: This can fail silently. `create` returns a boolean to indicate whether the operation was successful. Either you...
(a) store that return value in a variable and decide which flash message to display (success or "something went wrong") or...
(b) you use `create!` and let the website crash due to AR raising an error when the creation fails

app/jobs/sanity_check_results_job.rb
gregorbg, line 1: With the synchronous `group_by` approach in the "all sanity checks job", is it now even necessary to have this as a separate job anymore?

You can just call `SanityCheckMailer` directly in the job after all checks have been run synchronously in the `perform_now` loop

PR #13083, opened 2025-12-23
Add Sanity Check Queries and Categories as Static Data

FinnIckler:
All the Data for my PR at #13057 so it can be more easily reviewed by WRT. All the data was generated by using a PMA JSON export.

### DIFF
--- a/spec/models/sanity_check_spec.rb
+++ b/spec/models/sanity_check_spec.rb
@@ -23,5 +23,17 @@ RSpec.describe SanityCheck do
         expect { ActiveRecord::Base.connection.execute explain_query }.not_to raise_error
       end
     end
+
+    it "references all SQL files that are in the sanity_check_sql folder" do
+      # `pluck` does not work here because `query` is not a DB column!
+      used_queries = SanityCheck.all.map(&:query)
+
+      sc_folder = Rails.root.join("lib/sanity_check_sql")
+      sql_files = sc_folder.glob('**/*.sql')
+
+      stored_queries = sql_files.map(&:read)
+
+      expect(stored_queries).to match_array(used_queries)
+    end
   end
 end


app/models/sanity_check.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+class SanityCheck < ApplicationRecord
+  include StaticData
+
+  belongs_to :sanity_check_category
+
+  def self.data_file_handle
+    self.name.pluralize.underscore.to_s
+  end
```
gregorbg: Why are you overriding this? The default implementation in `StaticData` is already
```ruby
  def self.data_file_handle
    self.name.pluralize.underscore
  end
```

app/models/sanity_check.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  # Overwrite method to handle .sql files
+  def self.all_raw_sanitized
+    column_symbols = column_names.map(&:to_sym)
+
+    all_raw.map do |attributes|
+      attrs = attributes.symbolize_keys
+      attrs[:query] = Rails.root.join("lib", "sanity_check_sql", attrs[:query_file]).read
+      delete attrs[:query_file]
```
gregorbg: Not a fan of `delete` :/
Especially since you are using `slice` below, which is already effectively "removing" all columns that you don't care about. So I don't see the point of doing this `delete` operation here

app/models/sanity_check_category.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class SanityCheckCategory < ApplicationRecord
+  include StaticData
+
+  has_many :sanity_checks
+
+  def self.data_file_handle
+    self.name.pluralize.underscore.to_s
+  end
```
gregorbg: Same as above, I don't see why this override is necessary

lib/static_data/sanity_checks.json
gregorbg, line 1: A question about all query files: Why do the integer IDs need to be included in the file name/path?

Currently, you're referring to all `query_files` with the same pattern: `1 - person_data_irregularities/1 - name_special_characters.sql`. Why not just `person_data_irregularities/name_special_characters.sql`?

app/models/sanity_check.rb
```diff
@@ -0,0 +1,24 @@
...
+  include StaticData
+
+  belongs_to :sanity_check_category
+
+  def self.data_file_handle
+    self.name.pluralize.underscore.to_s
+  end
+
+  # Overwrite method to handle .sql files
+  def self.all_raw_sanitized
```
gregorbg: Here's a wild shot: You maybe don't need to override this loader function at all, and let the database only hold the name of the file, as well as let the sanity check category hold a "folder handle" (or similar, name tentative).

You could then load the file on-the-fly and stop storing the query in the database itself altogether.
If you want to cache the file loading, just use `@query_str ||= File.read(...)` or similar.

gregorbg: This would require a few schema changes, they should be straight-forward though

gregorbg: I am in favor of this new approach because it clarifies the "source of truth" for the actual queries: In the current case, the whole SQL query is loaded into the database at server boot time. This could confuse WRT if they at some point change a query directly in the database (for example, because they created a more optimized version) but then before the actual Sanity Check runs, another deployment "kills" the changes that were made only in the DB (but not in the source files).

lib/static_data/sanity_check_categories.json
```diff
@@ -0,0 +1,77 @@
+[
+  {
+    "id": 1,
+    "name": "Personal Data Irregularities",
```
gregorbg: Nit: In the database it's `Person Data Irregularities`

lib/static_data/sanity_check_categories.json
```diff
@@ -0,0 +1,77 @@
...
+    "email_to": "wrt@worldcubeassociation.org"
+  },
+  {
+    "id": 6,
+    "name": "Invalid Scrambles",
+    "email_to": "wrt@worldcubeassociation.org"
+  },
+  {
+    "id": 7,
+    "name": "Irregularities",
```
gregorbg: Should be `Competition Data Irregularities`

lib/static_data/sanity_checks.json
gregorbg, line 1: Unrelated: You're storing a lot of `comments` with empty strings, which are actually `null` in the database.
Try to either (a) explicitly encode `"comments": null` or (b, preferred if it works) just dropping the `comments` key from the sanity checks which don't have comments

gregorbg: Had another fundamentally different idea, see the model code for `SanityCheck` above

FinnIckler: I included them so they are sorted correctly. Meaning in the order that they are in the WRT sheet. Which makes it much easier for WRT to check the queries. This was when I would assume WRT would be checking if I copy pasted the queries out of the DB export correctly

app/models/sanity_check.rb
FinnIckler: whoops. Didn't know there was a default

app/models/sanity_check.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class SanityCheck < ApplicationRecord
+  include StaticData
+
+  belongs_to :sanity_check_category
+
+  def query
+    @query ||= Rails.root.join("lib", "sanity_check_sql", "#{sanity_check_category.id} - #{sanity_check_category.name.gsub(/\s+/, '').underscore}", "#{self.id} - #{self.query_file}").read
```
gregorbg: Too much happening in one line. First construct the folder name (the `"#category_id - #category_name"` part) and the file name (the `"#id - #query_file"` part) separately, then join them using `Rails.root.join`, and then finally do `@query ||= joined_file.read`.

gregorbg: Also, the `gsub`ed name of the category should be moved into its own method in the `SanityCheckCategory` model, so that you can just call `sanity_check_category.camel_name` here (method name tentative)

lib/static_data/sanity_checks.json
gregorbg: Okay having the files be sorted in the folder so that WRT can access them more easily makes sense

spec/models/sanity_check_spec.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe SanityCheck do
+  context "SQL Files" do
+    it "Can read all files" do
+      SanityCheckCategory.load_json_data!
+      SanityCheck.load_json_data!
```
gregorbg: This call should be moved to `db/seeds`, similar to `events.seeds.rb` and others that are already there

gregorbg: See also the `elibigle_iso2_for_championship` thingy as a reference on how to declare the order in which seeds need to be run (sanity checks can only be imported _after_ the categories are imported)

app/models/sanity_check_category.rb
```diff
@@ -4,4 +4,8 @@ class SanityCheckCategory < ApplicationRecord
   include StaticData
 
   has_many :sanity_checks
+
+  def camel_case_name
```
gregorbg: Random idea that just came to mind: Why not just have the following directly inside this `SanityCheckCategory` model?

```
def folder_handle
  "#{self.id} - #{self.name.parametrize.underscore}"
end
```

...where depending on your taste and appetite for complexity, the latter part could move into its own `def parameter_name` or similar. But I won't insist on that.

app/models/sanity_check.rb
```diff
@@ -6,6 +6,8 @@ class SanityCheck < ApplicationRecord
   belongs_to :sanity_check_category
 
   def query
-    @query ||= Rails.root.join("lib", "sanity_check_sql", "#{sanity_check_category.id} - #{sanity_check_category.name.gsub(/\s+/, '').underscore}", "#{self.id} - #{self.query_file}").read
+    category_folder = "#{sanity_check_category.id} - #{sanity_check_category.camel_case_name.underscore}"
+    file_name = "#{self.id} - #{self.query_file}"
```
gregorbg: Same idea as above: Move this into `def file_handle` right here in the model, then you can shorten `def query` again to `@query ||= Rails.root.join(..., sanity_check_category.folder_handle, self.file_handle).read`

app/models/sanity_check_category.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class SanityCheckCategory < ApplicationRecord
+  include StaticData
+
+  has_many :sanity_checks
+
+  def folder_handle
+    "#{self.id} - #{self.name.gsub(/\s+/, '').underscore}"
```
gregorbg: Why not `self.name.parametrize.underscore`?

FinnIckler: I tried that, but I got
`expected no Exception, got #<NoMethodError: undefined method 'parametrize' for an instance of String>`

FinnIckler: oh it's `parameterize`

gregorbg: Oh sorry my bad! :sweat_smile:

spec/models/sanity_check_spec.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe SanityCheck do
+  context "SQL Files" do
+    it "Can read all files" do
+      SanityCheck.find_each do |sanity_check|
+        expect { sanity_check.query }.not_to raise_error
```
gregorbg, line 9: After you have established that this does not raise an error, I would suggest also adding another `expect(sanity_check.query.presence).not_to be_nil` just to make sure that the query is also actually populated with some non-empty string content

PR #13087, opened 2025-12-26
Schedule Editor: Add effects to auto-expand calendar bounds

kr-matthews:
A minor improvement which reduces the need to go into the calendar settings and manually adjust the bounds.

https://github.com/user-attachments/assets/f2bdf515-f6c3-4a74-8d54-a981435994ce


gregorbg (commented):
I am quite allergic to using React effects for coupling state together. See https://react.dev/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state which is a simplified version of what you're trying to achieve.

Have you considered the following alternative solutions?
1. Have `calendarStart` and `calendarEnd` be fully automatic, derived values that you cannot manipulate manually (essentially turn them into a `useMemo` instead of a `useState`
2. Think about what can even trigger these effects in the first place, i.e. changes to the activities of the underlying WCIF. Identify all points in the code where these changes can be triggered (the action that adds an activity, removes an activity, moves an activity, etc.) and in those triggers, call `setCalendarStart` and `setCalendarEnd` instead of wrapping them in effect bodies

Sidenote: There is also `const setReasonableCalendarBounds` in this file already, could that possibly be reused?

PR #13112, opened 2026-01-04
Update CountryBandDetails table to match the new 2026 dues update

danieljames-dj:
This PR is to update the dues values as per the new values which were supposed to be in-effect from Jan 1.

### DIFF
--- a/db/migrate/20260104183856_update_country_band_details.rb
+++ b/db/migrate/20260104183856_update_country_band_details.rb
@@ -37,9 +37,9 @@ class UpdateCountryBandDetails < ActiveRecord::Migration[8.1]
     CountryBandDetail.where(number: 2).where(start_date: '2026-01-01').delete_all
     CountryBandDetail.where(number: 3).where(start_date: '2026-01-01').delete_all
     CountryBandDetail.where(number: 4).where(start_date: '2026-01-01').delete_all
-    CountryBandDetail.where(number: 1).where(start_date: '2018-01-01').update_all(end_date: nil)
-    CountryBandDetail.where(number: 2).where(start_date: '2018-01-01').update_all(end_date: nil)
-    CountryBandDetail.where(number: 3).where(start_date: '2018-01-01').update_all(end_date: nil)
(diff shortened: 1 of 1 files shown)


db/migrate/20260104183856_update_country_band_details.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  def down
+    CountryBandDetail.where(number: 1).where(start_date: '2026-01-01').delete_all
+    CountryBandDetail.where(number: 2).where(start_date: '2026-01-01').delete_all
+    CountryBandDetail.where(number: 3).where(start_date: '2026-01-01').delete_all
+    CountryBandDetail.where(number: 4).where(start_date: '2026-01-01').delete_all
+    CountryBandDetail.where(number: 1).where(start_date: '2018-01-01').update_all(end_date: nil)
+    CountryBandDetail.where(number: 2).where(start_date: '2018-01-01').update_all(end_date: nil)
+    CountryBandDetail.where(number: 3).where(start_date: '2018-01-01').update_all(end_date: nil)
+    CountryBandDetail.where(number: 4).where(start_date: '2018-01-01').update_all(end_date: nil)
```
gregorbg: Can you identify these "old bands" by using `where(end_date: '2025-12-31')`?

gregorbg (changes requested):
One code change regarding the `down` part is requested. The data itself (from the WFC email) looks good to me.

I would still like a third pair of eyes to cross-check that the raw data itself (the percentages, the raw USD amounts themselves, and the dates) are correct. Summoning @dunkOnIT, see the e-mail that Daniel forwarded to your WCA inbox for reference.

db/migrate/20260104183856_update_country_band_details.rb
danieljames-dj: Done

dunkOnIT:
This PR changes the `CountryBandDetails` - ie, what does Band 3 mean. It isn't changing which countries are _in_ band 3 - which is defined in the `CountryBands` table

PR #13114, opened 2026-01-05
Edit regional delegate in regions page

danieljames-dj:
There was a request from Board to have regional delegate edit access for the senior delegates.

The roles UIs are bit messed up, so adding new features like this can be a bit tricky. I've created a component `UserRoles/CreateModal` which can be placed in all the places where a new role needs to be created. This will help to avoid too many places where roles are added in different ways (like it's happening now).

For now, I'm using this new component only in the 'Region Manager' page (replaced already existing feature) and 'Regions' page (new feature). But in future this component can be easily replaced everywhere else which I'll be doing in any of the follow-up PRs.

### DIFF
--- a/app/webpacker/components/Panel/views/UserRoles/api/createUserRole.js
+++ b/app/webpacker/components/Panel/views/UserRoles/api/createUserRole.js
@@ -1,6 +1,5 @@
-import _ from 'lodash';
 import { fetchJsonOrError } from '../../../../../lib/requests/fetchWithAuthenticityToken';
-import { apiV0Urls } from '../../../../../lib/requests/routes.js.erb';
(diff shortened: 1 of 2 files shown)


app/webpacker/components/Panel/views/UserRoles/api/createUserRole.js
```diff
@@ -0,0 +1,24 @@
...
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify(_.omitBy({
+        userId,
+        groupId,
+        status,
+        location,
+      }, _.isNull)),
```
gregorbg: Can you share some parts of the logic from `requests.js.erb` method `jsonToQueryString`?

danieljames-dj: Done

Issue #13111, opened 2026-01-04
V2 results export requires significantly more disk space (and time) than the V1 results export


Logiqx:
## Summary

Working with the V2 results export requires significantly more disk space (and time) than the V1 results export.

The combined size of the ZIP and uncompressed SQL is now 2.6 times larger; increasing from 1.47 GB to 3.75 GB.

The storage required for the database tables is also 2 times larger; increasing from 1.35 GB to 2.70 GB.

These increases impact download times, load times, and storage requirements.

This is not ideal when using micro instances for automation within cloud environments.

## Details

The majority of tables in the V2 export are essentially the same size as V1. It is the results that are much larger..

The results table (V1) was 0.76 GB, but results and result_attempts (V2) are 0.58 GB and 1.53 GB respectively.

Loading result_attempts will also impact the size of the transaction log, perhaps double the original V1 database.

The sizes of the fields in result_attempts are as follows: 

id				bigint(20)	8 bytes
value			int(11)		4 bytes
attempt_number	int(11)		4 bytes
result_id			bigint(20)	8 bytes
created_at		datetime(6)	8 bytes
updated_at		datetime(6)	8 bytes

## Thoughts

It seems somewhat unlikely that the id and timestamp fields will be useful to third parties for statistical / personal purposes. The PK does not facilitate table joins, and the timestamps are operational details.

Perhaps it would make sense to limit the result export to result_id + attempt_number + value, but include id and timestamps in the developer export?

This would reduce individual record sizes from 40 bytes to 16 bytes. The attempt_number field could also be a tinyint (1 byte) for further savings.

## Issues

I have reworked the Senior Rankings to use the V2 results export, but the deployment has failed due to storage that is available to my AWS micro instances.

I could look to increase the storage by a further 5 GB but that will cost me and feels avoidable, when considering the thoughts / suggestions.

Hopefully you will consider modifying the results export for the benefit of third parties like myself, and perhaps the WCA as well?

FinnIckler:
I think this sounds mostly reasonable, not sure if we shoot ourself in the foot with the tinyint though as that means we could never do more than 127 attempts per result. Currently Head2Head results are capped at a certain point (5 sets * 7 points = 35 attempts maximum) but that could change. I am in favour of looking more closely on our datatypes in general though.

dunkOnIT:
[...]

PR #13129, opened 2026-01-06
Results Export v2 Size Reduction

dunkOnIT:
First pass at addressing the size constraints described in #13111

This PR: 
- Adds a `local` argument to `db:dump:public_results`, making it easier to run/test the results export locally
- Makes the following changes to the exported `result_attempts` table in the v2 export:
  - Removes `created_at` and `updated_at` timestamps
  - Sets `attempt_number` to an unsigned `tinyint` - giving us up to 255 attempts per round, which seems adequate for the foreseeable future
  - Removes the `id` column from `result_attempts` (we match based on the `result_id` column, so it isn't needed)

### DIFF
--- a/spec/lib/database_dumper_spec.rb
+++ b/spec/lib/database_dumper_spec.rb
@@ -90,14 +90,14 @@ RSpec.describe "DatabaseDumper" do
         all_live = {
           v1: {
             metadata: {
-              export_format_version: '1.0.0',
+              export_format_version: 'v1.0.0',
               version_label: 'deprecated',
               end_of_life_date: '3026-01-01',
             },
           },
           v2: {
             metadata: {
-              export_format_version: '2.0.1',
+              export_format_version: 'v2.0.2',
               version_label: 'current',
               end_of_life_date: nil,
             },
@@ -112,14 +112,14 @@ RSpec.describe "DatabaseDumper" do
         v1_dead = {
           v1: {
             metadata: {
-              export_format_version: '1.0.0',
+              export_format_version: 'v1.0.0',
               version_label: 'deprecated',
               end_of_life_date: '1026-01-01',
             },
           },
           v2: {
             metadata: {
-              export_format_version: '2.0.1',
+              export_format_version: 'v2.0.2',
               version_label: 'current',
               end_of_life_date: nil,
             },


db/results_dump_v2_schema.rb
```diff
@@ -87,12 +87,10 @@
     t.integer "country_rank", default: 0, null: false
   end
 
-  create_table "result_attempts", charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
+  create_table "result_attempts", id: false, charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
     t.integer "value", null: false
-    t.integer "attempt_number", null: false
+    t.column "attempt_number", "tinyint unsigned", null: false
```
gregorbg: I think you can realize something similar via `t.integer "foo_column", limit: 2`. The number being passed to `limit` is the storage size in bytes.

gregorbg: The `unsigned` part can be realized via `t.integer "foo_column", unsigned: true`

lib/tasks/db.rake
```diff
@@ -43,10 +43,13 @@ namespace :db do
...
     desc 'Generates a partial dump of our database containing only results and relevant stuff for statistics.'
-    task public_results: :environment do
+    # task public_results: :environment do
+    task :public_results, [:local] => [:environment] do |_, args|
+      local = args[:local].present? || false
+
       DatabaseDumper.results_export_live_versions.each do |v|
-        puts "Dumping results for version: #{v}"
-        DbDumpHelper.dump_results_db(v)
+        puts "Dumping results for version: #{v}. Dumping locally? #{local}"
```
gregorbg:
```suggestion
        puts "Dumping results#{local ? ' locally' : ''} for version: #{v}"
```

lib/tasks/db.rake
```diff
@@ -43,10 +43,13 @@ namespace :db do
     end
 
     desc 'Generates a partial dump of our database containing only results and relevant stuff for statistics.'
-    task public_results: :environment do
+    # task public_results: :environment do
+    task :public_results, [:local] => [:environment] do |_, args|
+      local = args[:local].present? || false
```
gregorbg:
```suggestion
      local = args[:local].present?
```

gregorbg: The method `.present?` **always** returns a boolean, so I don't know why `|| false` would be necessary

lib/db_dump_helper.rb
```diff
@@ -76,35 +76,41 @@ def self.cached_results_export_info(file_type, version, export_timestamp = DumpP
...
-  def self.dump_results_db(version, export_timestamp = DateTime.now)
-    Dir.mktmpdir do |dir|
-      FileUtils.cd dir do
-        tsv_folder_name = "TSV_export"
-        FileUtils.mkpath tsv_folder_name
+  def self.dump_results_db(version, export_timestamp = DateTime.now, local: false)
+    target_dir =  if local
+                    path = "#{RESULTS_EXPORT_FILENAME}_#{version}_#{export_timestamp.strftime('%Y%m%dT%H%M%SZ')}"
+                    Dir.mkdir(path)
+                    path
```
gregorbg:
```suggestion
"#{RESULTS_EXPORT_FILENAME}_#{version}_#{export_timestamp.strftime('%Y%m%dT%H%M%SZ')}".tap { Dir.mkdir(it) }
```

gregorbg (changes requested):

db/results_dump_v2_schema.rb
dunkOnIT: Changed, with the exception that I've made it `limit: 1` as `tinyint` is 1 byte

lib/tasks/db.rake
dunkOnIT: Ah, that was a holdover from 
```rails
      reload = args[:reload] || false
```

lib/tasks/db.rake
dunkOnIT: Changed

lib/db_dump_helper.rb
dunkOnIT: Ooh that's nice

lib/tasks/db.rake
```diff
@@ -45,10 +45,10 @@ namespace :db do
     desc 'Generates a partial dump of our database containing only results and relevant stuff for statistics.'
     # task public_results: :environment do
     task :public_results, [:local] => [:environment] do |_, args|
-      local = args[:local].present? || false
+      local = args[:local].present?
 
       DatabaseDumper.results_export_live_versions.each do |v|
-        puts "Dumping results for version: #{v}. Dumping locally? #{local}"
+        puts "Dumping results #{local ? 'locally' : ''} for version: #{v}."
```
gregorbg: Hyper nitpick: If you're _not_ running locally, and the `#{}` part evaluates to an empty string, then you will have two spaces (one in front of the `#{}` interpolation and one directly after).
My original suggestion was `"foo#{local ? "yes" : "no"} bar"` intentionally, although I admit that in the end we won't care much about this one log line

lib/db_dump_helper.rb
```diff
@@ -76,35 +76,39 @@ def self.cached_results_export_info(file_type, version, export_timestamp = DumpP
...
-    Dir.mktmpdir do |dir|
-      FileUtils.cd dir do
-        tsv_folder_name = "TSV_export"
-        FileUtils.mkpath tsv_folder_name
+  def self.dump_results_db(version, export_timestamp = DateTime.now, local: false)
+    target_dir =  if local
+                    "#{RESULTS_EXPORT_FILENAME}_#{version}_#{export_timestamp.strftime('%Y%m%dT%H%M%SZ')}".tap { Dir.mkdir(it) }
+                  else
+                    Dir.mktmpdir
+                  end
```
gregorbg: Now that you have this one-liner simplification in place, you can consider turning the whole thing into a ternary: `local ? foo_bar : Dir.mktmpdir`. Not sure how to set line breaks to make it look pretty though, but I'm sure Rubocop can figure something out 😉

lib/tasks/db.rake
dunkOnIT: This was a rubocop change - do we want to override it, take a different tack, or leave as it?

lib/db_dump_helper.rb
dunkOnIT: Changed - I thought the line looked a bit too long as a ternary which is why I initially left it as the if-block, but happy to make it a ternary if you're happy with the readability!

gregorbg (commented):
In the README changelog for `v2.0.1`, you explicitly mention:
```md
- Fixed missing `results.id` and `result_attempts.id` keys
```

Now in this PR you're removing the `result_attempts.id` again. Does this not contradict the changelog? Should we bump the version once more?

dunkOnIT:
Excellent point! Just done that, and amended references to v2.0.1

I also picked up that the stubbed versions in `database_dumper_spec` were using the wrong format (1.0.0 instead of v1.0.0) - this is corrected, but those values were never under test[1] (the value of their key was), which is why it didn't cause any test failures.

[1] and don't need to be under test

PR #13131, opened 2026-01-07
Sanity check Tests: person data irregularities

FinnIckler:
Was able to do almost all of them using a simple shared example. I had to do `update_columns` a couple times to bypass rails validations, so these cases can not actually be reached through the rails code.

### DIFF
--- a/spec/models/sanity_check_spec.rb
+++ b/spec/models/sanity_check_spec.rb
@@ -45,7 +45,7 @@ RSpec.describe SanityCheck do
             irregular_people = irregular_people.map do |name|
               create(:person, name: name)
             end
-            result_ids = run_query(sanity_check.query).pluck("id")
+            result_ids = sanity_check.run_query.pluck("id")
 
             expect(result_ids).to match_array(irregular_people.map(&:id))
           end
@@ -54,7 +54,7 @@ RSpec.describe SanityCheck do
             valid_people = valid_people.map do |name|
               create(:person, name: name)
             end
-            result_ids = run_query(sanity_check.query).to_a
+            result_ids = sanity_check.run_query.to_a
 
             expect(result_ids & valid_people.map(&:id)).to be_empty
           end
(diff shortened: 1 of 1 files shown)


lib/sanity_check_sql/1 - person_data_irregularities/61 - first_names_with_abbreviation.sql
```diff
@@ -1,3 +1,3 @@
-SELECT wca_id, name, country_id
+SELECT id, wca_id, name, country_id
```
gregorbg, line 1: Is this change consistent with the other queries of category 1?
We have a bunch of other sanity checks that do Regex'ing on `persons` table names, do they all select the `id` field?

spec/models/sanity_check_spec.rb
```diff
@@ -36,4 +36,154 @@
...
+        { id: 5, irregular_people: ["John Doe (abc)", "John doe (a黄)"], valid_people: ["John Doe (黄)"] },
+        { id: 6, irregular_people: ["John Doe (黄"], valid_people: ["John Doe (黄)"] },
+        { id: 7, irregular_people: [
+          "John Doe (黄) ",
+          "John Doe ",
+          " John Doe",
+          "John Doe (黄 )",
+          "John Doe ( 黄)",
+        ], valid_people: ["Jane Doe"] },
+        { id: 8,  irregular_people: ["д (John)", "β (Jane)"], valid_people: ["Ja Do (ณั)", "Ja Do (黄)", "Jo Do (文)", "Ja Do (혁)", "Jo Do (星)"] },
```
gregorbg, line 99: An interesting edge case in `irregular_people` that you should consider here: There are certain Cyrillic or Greek letters that visually look like they could be Latin alphabet letters (but they have a different UTF-8 code point). Those should be tested by using a name which is "mostly Latin" but contains one "hidden" Cyrillic letter (or Greek letter, whatever example you can come up with I don't care)

spec/models/sanity_check_spec.rb
```diff
@@ -36,4 +36,154 @@
...
+
+          expect(result_ids).to contain_exactly(competition.id)
+        end
+      end
+    end
+  end
+
+  def run_query(query)
+    ActiveRecord::Base.connection.exec_query(query)
+  end
```
gregorbg: Based on #13057 you should add
```ruby
def execute_query
  ActiveRecord::Base.connetion.exec_query self.query
end
```
to the AR `SanityCheck` model. The `SanityCheckCategoryJob` over in the other PR can then use this method, as well as all of your tests here and in subsequent PRs as well.

lib/sanity_check_sql/1 - person_data_irregularities/61 - first_names_with_abbreviation.sql
FinnIckler: yes, that's why I changed it. It is the only query regarding names that does not select the id field

spec/models/sanity_check_spec.rb
FinnIckler: I was thinking the same thing already, but wasn't sure in what PR to put it

PR #13132, opened 2026-01-07
Sanity check tests: irregular results

FinnIckler:
These are the tests against the original non 1..5 queries. I will build upon the PR to switch them out with the new queries

### DIFF
--- a/spec/models/sanity_check_spec.rb
+++ b/spec/models/sanity_check_spec.rb
@@ -125,4 +125,150 @@ RSpec.describe SanityCheck do
       end
     end
   end
+
+  context "Person Data Irregularities" do
+    context "Wrong names" do
+      RSpec.shared_examples 'correct sanity check' do |sanity_check, irregular_people, valid_people|
+        context sanity_check.topic.to_s do
+          it "correctly finds all irregular names" do
+            irregular_people = irregular_people.map do |name|
+              create(:person, name: name)
+            end
+            result_ids = sanity_check.run_query.pluck("id")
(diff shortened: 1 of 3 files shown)


lib/sanity_check_sql/2 - irregular_results/14 - wrong_number_of_results.sql
```diff
@@ -1,4 +1,4 @@
-SELECT *
+SELECT *, r.id as result_id
```
gregorbg, line 1: Why is this aliasing necessary if you're already doing `SELECT *`?

FinnIckler: because the id column is taken by format.id

spec/models/sanity_check_spec.rb
```diff
@@ -36,4 +36,84 @@
...
+        result_ids = sanity_check.run_query.pluck("id")
+
+        expect(result_ids).to contain_exactly(r.id)
+      end
+    end
+
+    context "wrong number of results" do
+      let(:sanity_check) { SanityCheck.find(14) }
+
+      it "Correctly find irregular results" do
```
gregorbg: The sanity check is called "wrong number of results", but you're only testing "less than the expected amount". What about "more than the expected amount", i.e. an `mo3` result with 4 values?

FinnIckler: good point. Added!

PR #13133, opened 2026-01-07
Sanity Check Tests: Duplicate Results

FinnIckler:
Just a single sanity check

### DIFF
--- a/spec/models/sanity_check_spec.rb
+++ b/spec/models/sanity_check_spec.rb
@@ -43,9 +43,9 @@ RSpec.describe SanityCheck do
       competition = create(:competition)
       round = create(:round, competition: competition)
       create(:result, competition: competition, round: round, event_id: "333",
-                      value1: 100, value2: 100, value3: 100, value4: 100, value5: 100, average: 100, best: 100)
+                      value1: 100, value2: 200, value3: 300, value4: 400, value5: 500, average: 300, best: 100)
       create(:result, competition: competition, round: round, event_id: "333",
-                      value1: 100, value2: 100, value3: 100, value4: 100, value5: 100, average: 100, best: 100)
+                      value1: 100, value2: 200, value3: 300, value4: 400, value5: 500, average: 300, best: 100)
 
       result_ids = sanity_check.run_query.pluck("competitions")
       expect(result_ids).to contain_exactly(competition.id)


spec/models/sanity_check_spec.rb
```diff
@@ -36,4 +36,17 @@
...
     end
   end
+
+  context "Duplicate Results" do
+    it "Correctly finds duplicate results" do
+      sanity_check = SanityCheck.find(18)
+      competition = create(:competition)
+      round = create(:round, competition: competition)
+      create(:result, competition: competition, round: round, event_id: "333")
+      create(:result, competition: competition, round: round, event_id: "333")
```
gregorbg: You should also mock a person to make sure the result has been achieved by the same person

gregorbg: Side note: Is this even considered by the sanity check itself?

FinnIckler: no, it's just at the same competition

gregorbg: Okay, but then what is making this result a "duplicate" in the first place?
If you consider only the two `create` lines here by themselves, then you're essentially saying "create two 333 results for the same round of the same competition". For reference, the most recent World Championship had around one thousand results for the same 333 round of the same competition. So doing this is not inherently wrong.
What hidden implementation detail of our factories are turning these results into an actual *duplicate*?

FinnIckler: oh, it's because factories always have the same values 1..5 by default. I will make that explciit

spec/models/sanity_check_spec.rb
```diff
@@ -37,6 +37,21 @@
...
 
+  context "Duplicate Results" do
+    it "Correctly finds duplicate results" do
+      sanity_check = SanityCheck.find(18)
+      competition = create(:competition)
+      round = create(:round, competition: competition)
+      create(:result, competition: competition, round: round, event_id: "333",
+                      value1: 100, value2: 100, value3: 100, value4: 100, value5: 100, average: 100, best: 100)
+      create(:result, competition: competition, round: round, event_id: "333",
+                      value1: 100, value2: 100, value3: 100, value4: 100, value5: 100, average: 100, best: 100)
```
gregorbg: Can you make the individual values a little more "exciting"?
Something like `100, 200, 300, 400, 500`  yields `best=100` and `average=300`, but at the same time it makes clear that "duplicate" does not mean "duplicate attempts within the same result" (as your repeated 100 might imply to the untrained eye) but rather means "exactly identical independent results across the same round"

PR #13134, opened 2026-01-07
Sanity Check Tests: WCA Id Irregularities

FinnIckler:
Also just one sanity check


spec/models/sanity_check_spec.rb
```diff
@@ -36,4 +36,16 @@
       expect(stored_queries).to match_array(used_queries)
     end
   end
+
+  context "WCA Id Irregularities" do
+    it "Correctly finds year not matching to first competition year" do
+      sanity_check = SanityCheck.find(19)
+      person = create(:person)
+      person.update_columns(wca_id: "1982TEST01")
+      create(:result, person: person)
```
gregorbg: You should explicitly create a competition (which is explicitly not 1982) to have a more controlled set of test data

PR #13139, opened 2026-01-08
Assign registrant_id based on max registrant_id

dunkOnIT:
After manually deleting registrations for NAC2026, we had an incident during NAC2026 registrations where the `uniqueness` validation on `registrant_id` was failing. This is because `registrant_id` was being assigned based on the count of registrations - but by manually deleting registrations, we reduced the count, and put ourselves in a position where we were trying to assign the same `registrant_id` to multiple values. 

As Cailyn suggested, the solution here is to increment from the max `registrant_id`, not to use the count.

### DIFF
--- a/app/models/registration.rb
+++ b/app/models/registration.rb
@@ -66,9 +66,8 @@ class Registration < ApplicationRecord
   before_create :ensure_registrant_id
 
   private def ensure_registrant_id
-    self.registrant_id ||= competition.registrations.count + 1
-    # max_registrant_id = competition.registrations.maximum(:registrant_id) || 0
-    # self.registrant_id ||= max_registrant_id + 1
+    max_registrant_id = competition.registrations.maximum(:registrant_id) || 0
+    self.registrant_id ||= max_registrant_id + 1
   end
 
   validates :guests, numericality: { greater_than_or_equal_to: 0 }


app/models/registration.rb
```diff
@@ -66,7 +66,7 @@ class Registration < ApplicationRecord
   before_create :ensure_registrant_id
 
   private def ensure_registrant_id
-    self.registrant_id ||= competition.registrations.count + 1
+    self.registrant_id ||= competition.registrations.any? ? competition.registrations.maximum(:registrant_id) + 1 : 1
```
gregorbg: This (mostly) generates two SQL queries: One for the `any?` and (if that returned `true`, which it will in 99% of cases) one more for the `maximum`.

I've just tried out `Registration.where(competition_id: 'DefinitelyNotExistingComp1970').maximum(:registrant_id)` to simulate the case of "no registrations exist yet". The result of the `maximum` call is `nil`. So you can always just fire one `maximum` and use `|| 0` to coalesce to zero. Stuff that into a variable, and _then_ add one to guarantee that you will always just fire one query no matter what.

gregorbg (changes requested):

spec/models/registration_spec.rb
gregorbg, line 1: I think you need to consider another test case which more closely models what happened at NAC: Create exactly the right amount (and right order with respect to `registrant_id`) so that the old `count` approach _would_ crash, but the new max approach doesn't.

That way, we will be safe(r) in the future if we ever decide to change the implementation again.

dunkOnIT:
Another argument against "filling gaps": `registrant_id` is used in tools like Groupifier, DD, and printed on physical materials like scorecards or lanyards. "Filling gaps" might cause collision with existing assignments in other software/physical tools

spec/models/registration_spec.rb
dunkOnIT: Test case added as requested. The test case: 
- Creates registrations up to a total of 5
- Deletes `registrant_id: 4`, leaving 4 registrations where the last one will have a `registrant_id` of 5
- Creates a new registration, and asserts that
  - Assigning a `registrant_id` doesn't cause a clash
  - Its `registrant_id` is 6, as would be expected in the new code

You can test this by running the test on `models/registration_spec.rb:1445`, and replacing `ensure_registrant_id` with its original body: 
```ruby
  private def ensure_registrant_id
    self.registrant_id ||= competition.registrations.count + 1
  end
```

That will cause the test to fail, but the new one succeeds

PR #13145, opened 2026-01-09
Sanity check tests: duplicate scrambles

FinnIckler:


spec/models/sanity_check_spec.rb
```diff
@@ -36,4 +36,30 @@
...
     end
   end
+
+  context "Duplicate_scrambles" do
+    it "Duplicate Scrambles within competition id" do
+      sanity_check = SanityCheck.find(20)
+      competition = create(:competition)
+      round = create(:round, competition: competition)
+      create(:scramble, competition: competition, round: round)
+      create(:scramble, competition: competition, round: round)
```
gregorbg: These two lines, isolated by themselves, do not indicate to me that the generated data would be "wrong". You're just creating two scrambles for the same round of the same competition. But that's not inherently wrong: In reality, one round of one competition will have 25+ scrambles, or even 100+ for major champs.
You need to make more explicit what the "duplication" here is.

spec/models/sanity_check_spec.rb
```diff
@@ -36,4 +36,30 @@
...
+
+      expect(result_ids).to contain_exactly(competition.id)
+    end
+
+    it "Duplicate Scrambles across competition" do
+      sanity_check = SanityCheck.find(21)
+      competition_1 = create(:competition)
+      competition_2 = create(:competition)
+      create(:scramble, competition: competition_1)
+      create(:scramble, competition: competition_2)
```
gregorbg: Same deal as above. This data generation alone by itself is not invalid (it "happens to be" invalid because of the specific implementation detail of our factories, but we shouldn't rely on those)

spec/models/sanity_check_spec.rb
```diff
@@ -37,6 +37,32 @@
     end
   end
 
+  context "Duplicate_scrambles" do
+    it "Duplicate Scrambles within competition id" do
```
gregorbg, line 41: Your title implies "within the whole competition [potentially even across rounds]" but your test _actually_ covers "within the same round of a competition". Which one does the sanity check actually consider? If it's both, then have tests for both. If it's only one of them, make sure the title and the test contents are consistent.

FinnIckler: it groups the scrambles by competition_id not round_id. So I will switch it to be two different rounds

gregorbg: Well if it groups by `competition_id` then it means "doesn't matter if it's the same round_id or not". In other words, it can detect both cases: One for where there is a duplicate scramble within the same round, but it can _also_ detect when that duplicate scramble is in another, independent round of the same competition.
I think both cases should be covered by your test.

PR #13151, opened 2026-01-09
Ticket to claim WCA ID

danieljames-dj:
This PR is adding new ticket type `claim_wca_id`. With this PR, the WCA ID claims will happen within tickets.

Also, in this PR, we are making the tickets and user's columns (`unconfirmed_wca_id` and `delegate_id_to_handle_wca_id_claim`) in sync. In the follow-up PR, these two columns can be removed from `users` table as they are no longer needed.


app/controllers/accounts/registrations_controller.rb
```diff
@@ -1,14 +1,21 @@
...
   before_action :check_captcha, only: [:create]
-  # rubocop:enable Rails/LexicallyScopedActionFilter
+
   protected def after_update_path_for(_resource)
     edit_user_registration_path
   end
 
+  def create
+    super do |resource|
+      next unless resource.persisted?
```
gregorbg, line 14: Where did you get this line from?

app/controllers/accounts/registrations_controller.rb
```diff
@@ -1,14 +1,21 @@
...
+      next if resource.delegate_to_handle_wca_id_claim.nil?
+
+      TicketsClaimWcaId.create_ticket!(resource)
```
gregorbg: What about...
```suggestion
      TicketsClaimWcaId.create_ticket!(resource) if resource.delegate_to_handle_wca_id_claim.present?
```

gregorbg: Does RuboCop complain about that?

app/views/users/edit.html.erb
```diff
@@ -83,7 +83,10 @@
                       </a>
                     </span>
                     <div class="input-group-btn">
-                      <%= link_to t('.approve'), confirm_wca_id_path(userId: @user.id, wcaId: @user.unconfirmed_wca_id), method: :post, data: { confirm: t('.approve_confirm', wca_id: @user.unconfirmed_wca_id) }, class: "btn btn-default", id: "approve-wca-id" %>
+                      <% claim_ticket = TicketsClaimWcaId.open.find_by(user: @user)&.ticket %>
+                      <% if claim_ticket %>
+                        <%= link_to t('.go_to_claim_wca_id_ticket'), ticket_path(claim_ticket), class: "btn btn-primary", target: "_blank" %>
+                      <% end %>
```
gregorbg: The old "confirm" button (in the left, deleted side of the diff view) was always present, no matter what.
Now you're only rendering the link in certain conditions. So it changed from "always some content" to "maybe some content". Why? Does this break the UI in cases where the content/ticket is not there?

app/webpacker/components/Tickets/TicketWorkbenches/ClaimWcaIdActionerView/ApproveView.jsx
```diff
@@ -0,0 +1,65 @@
...
+        positive
+        onClick={() => setShowApproveConfirm(true)}
+        disabled={disabled}
+      >
+        Approve
+      </Button>
+
+      <Confirm
+        open={showApproveConfirm}
+        content="You are about to approve the WCA ID claim. This will also remove all other claims for this WCA ID. Are you sure you want to continue?"
```
gregorbg, line 53: Is there a chance to show how many other claims there are? If it gets too complicated (ie, you would need to design a completely new endpoint for it) then don't bother...

README.md
```diff
@@ -1,3 +1,8 @@
+# TODO
+
+1. Modify tickets_controller.rb - everything else committed so far is good.
+2. Close tickets when a claim is cleared.
```
gregorbg: Whoops 🙃

gregorbg (changes requested):

README.md
danieljames-dj: Whoops, removed.

app/controllers/accounts/registrations_controller.rb
danieljames-dj: I don't remember - maybe while testing, I had to add this? Anyway I have added comment explaining why we need it.

app/controllers/accounts/registrations_controller.rb
danieljames-dj: I tried applying it. Hopefully it won't complain. I made it separate because there are two cases where I need to skip.

app/views/users/edit.html.erb
danieljames-dj: Ah good catch. So earlier as well there is a condition - this will be shown only if `@user.unconfirmed_wca_id.present?`.

I've now changed that as well. Hope it makes sense now.

app/webpacker/components/Tickets/TicketWorkbenches/ClaimWcaIdActionerView/ApproveView.jsx
danieljames-dj: I think we need to write an API to fetch that - can do it, won't be a big deal.

danieljames-dj: I just realized I didn't verify whether the other tickets are actually getting closed. I'll check that as well.

app/controllers/accounts/registrations_controller.rb
gregorbg: I still don't see it in the code changes. Please double-check whether you committed your change as intended.

app/controllers/accounts/registrations_controller.rb
gregorbg: Thank you, very helpful!

app/controllers/tickets_controller.rb
```diff
@@ -380,6 +431,7 @@ def join_as_bcc_stakeholder
     is_active = ActiveRecord::Type::Boolean.new.cast(params.require(:is_active))
 
     ActiveRecord::Base.transaction do
+      @user.update!(**User::CLEAR_WCA_ID_CLAIM_ATTRIBUTES)
```
gregorbg: Why does this specific, WCA ID claim related action happen in a generic, general-purpose BCC stakeholder endpoint?!

app/views/users/edit.html.erb
```diff
@@ -70,7 +70,8 @@
         <% if current_user.can_view_all_users? %>
           <div class="row" id="wca_id">
             <div class="col-xs-6">
-              <% if @user.unconfirmed_wca_id.present? %>
+              <% claim_ticket = TicketsClaimWcaId.open.find_by(user: @user)&.ticket %>
+              <% if claim_ticket %>
```
gregorbg:
```suggestion
              <% if claim_ticket.present? %>
```

app/webpacker/components/Tickets/TicketWorkbenches/ClaimWcaIdActionerView/ApproveView.jsx
gregorbg: Kindly ping :)

app/webpacker/components/Tickets/TicketWorkbenches/ClaimWcaIdActionerView/index.jsx
```diff
@@ -0,0 +1,58 @@
...
+      </Message>
+    );
+  }
+
+  if (status === ticketStatuses.claim_wca_id.closed) {
+    return (
+      <>
+        This request is already handled.
+      </>
+    );
```
gregorbg: Why do you wrap this in an empty fragment tag?

```suggestion
    return "This request is already handled.";
```

db/migrate/20260212074157_backfill_claim_wca_id_tickets_for_existing_users.rb
```diff
@@ -0,0 +1,20 @@
...
+  end
+
+  def down
+    TicketsClaimWcaId.find_each do |tickets_claim_wca_id|
+      ticket = tickets_claim_wca_id.ticket
+      ticket.ticket_comments.destroy_all
+      ticket.ticket_logs.destroy_all
+      ticket.ticket_stakeholders.destroy_all
+      tickets_claim_wca_id.destroy
+      ticket.destroy
```
gregorbg: If you add some smart `dependent: :destroy` or `dependent: :delete_all` (exercise: Consult Rails documentation about what is the difference between the two) then these two lines will be enough, and Rails will do all the "cascading" for you

gregorbg: Depending on the support for polymorphics, even one single `ticket.destroy` will be enough (if you can cascade into the polymorphic metadata with a `dependent` option as well, which I'm not sure about)

gregorbg (changes requested):

app/controllers/accounts/registrations_controller.rb
danieljames-dj: Oops I think I missed it. I've applied suggestion now.

app/views/users/edit.html.erb
danieljames-dj: Applied suggestion

app/webpacker/components/Tickets/TicketWorkbenches/ClaimWcaIdActionerView/index.jsx
danieljames-dj: Ya empty fragment was not needed, committed suggestion.

app/controllers/tickets_controller.rb
danieljames-dj: Oops I accidentally added it. Removed it now. I really don't remember why I did it though.

danieljames-dj (commented):
I'm changing this to draft for now. I feel there are some gaps which needs to be handled.

db/migrate/20260212074157_backfill_claim_wca_id_tickets_for_existing_users.rb
danieljames-dj: This is now done.

app/webpacker/components/Tickets/TicketWorkbenches/ClaimWcaIdActionerView/ApproveView.jsx
danieljames-dj: I've addressed the request in https://github.com/thewca/worldcubeassociation.org/pull/14188, will rebase this PR once that PR is ready.

PR #13152, opened 2026-01-10
Add bcc stakeholder in tickets for admins

danieljames-dj:
Currently tickets can only be accessed by stakeholders. I had plan to extend it to relevant users in future including super admins (WST admins + WRT Seniors), but looks like I need to do it now because any delegates should be able to access a "claim WCA ID ticket" (claim WCA ID ticket is WIP). This is because this request is "transferrable" by any delegates. For example, I claimed WCA ID with a person X as Delegate, but the person X is not responding to my messages or emails. Suddenly I met another delegate person Y, and I want person Y to claim my WCA ID. Currently person Y can just go to my profile and handle the claim even though it's assigned to person X. So when it comes to tickets, the person Y should be able to open the ticket, and transfer the ticket to themselves and handle the claim.

### DIFF
--- a/app/webpacker/components/Tickets/index.jsx
+++ b/app/webpacker/components/Tickets/index.jsx
@@ -48,7 +48,7 @@ function Tickets({ id }) {
 
   if (isPendingTicketDetails) return <Loading />;
   if (isErrorTicketDetails) {
-    if (errorTicketDetails?.response?.status === 401 && errorTicketDetails?.message?.includes('No access to ticket')) {
+    if (errorTicketDetails.message.includes('No access to ticket')) {
       return <SelfRoleAssigner ticketId={id} />;
     }
     return <Errored />;


gregorbg:
I like the vibe of your second suggestion, but I have two points of feedback/questions:

1. What do you mean by "still assign them an eligible role without creating a stakeholder"? How would that look like, using the concrete example of WCA ID claims?
2. I don't think `check_ticket_errors` is the right place to implicitly create roles. I don't like implicitly doing important things, I didn't like it in your initial version (as you first submitted the PR) and I don't like it in your second proposal either. What about the UI offering some kind of "I want to take charge of this!" button, which creates the stakeholder entity when you click it, and _after_ that mutation has been successful, the UI then lets you do actual actions on the ticket?

danieljames-dj:
This is now ready for review.

app/webpacker/components/Tickets/SelfRoleAssigner.jsx
```diff
@@ -0,0 +1,98 @@
...
+  if (isError) return <Errored error={error} />;
+
+  return (
+    <>
+      <Message warning>
+        <Message.Header>No Active Roles</Message.Header>
+        <p>You don&apos;t have any roles currently, but you can assign a role to yourself.</p>
+        <Button primary onClick={() => setIsAssigning(true)}>
+          Click here to assign a role for yourself
+        </Button>
```
gregorbg: I think the average Delegate (looking at later when we have "Claim WCA ID" as tickets, which any Delegate should be able to access) does not understand what "You don't have any roles" means.

A more transparent way would be to say "You currently don't have access to this ticket" or "You are currently not assigned to this ticket" or "You currently don't have permission to perform any actions on this ticket" or anything similar like this.

gregorbg: In the modal below, I can more easily accept the term "roles", because then the Delegate understands "Oh I was not linked with this ticket before, now I have to choose _how_ I want to be linked to the ticket".

So in the second step, using "role" is okay. But in the initial header that first announces the issue, already using "role" might feel a bit confusing.

app/controllers/ticket_stakeholders_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+
+    return head :unauthorized unless ticket.metadata.eligible_roles_for_bcc(current_user).include?(stakeholder_role)
+
+    stakeholder = TicketStakeholder.create!(
+      ticket_id: ticket.id,
+      stakeholder: current_user,
+      connection: connection,
+      stakeholder_role: stakeholder_role,
+      is_active: is_active,
+    )
```
gregorbg: Will this create a log entry for the ticket?

app/webpacker/components/Tickets/SelfRoleAssigner.jsx
danieljames-dj: Okay I've changed it (still included role, but made a clarification first). Hope this works.

app/controllers/ticket_stakeholders_controller.rb
danieljames-dj: No, for stakeholders, I haven't started adding logs. I understand the reason of having a log for this entry - so that we can know if anybody was added to this ticket by checking the logs.

I'm actually having some issues in creating logs for the stakeholders. Indeed it's a separate issue, but will mention here and maybe I'll address it in a follow-up PR.

The issue I'm facing is, whenever a stakeholder is created, I need to add log. Something like:

```
TicketStakeholder.create!(...)
TicketLogs.create!(...)
```

But I don't find this strong enough because we can easily miss adding the log entry. Do you have any other suggestion for the same?

I cannot use `after_create` because I won't have `requester` object.

LLM suggested to have `attr_accessor :requester` in the model, and do some validation like `validates :requester, presence: true, on: :create`. This way, while creating stakeholder, we need to mandatorily add `requester` while creating the object and I can use `after_create` for creating the logs.

Do you agree with this LLM solution or do you have any other idea?
(This is applicable even to the logs of ticket itself so I prefer some common solution which can be used in both Ticket and TicketStakeholder).

gregorbg: I am having a hard time to accept this argument that "we can easily miss adding the log entry". Actually, you can easily miss a lot of things. You are the programmer, you are responsible that your web application does what you want it to do.

Your concrete suggestion is not suitable for this PR, because it introduces a precedent that others might copy and start using in their own code too. Before we do that, I want to _thoroughly_ consider all the options that we have. This PR is not the place for such thorough discussions.

Also, if you do end up implementing this as a model hook, then there's always the danger that there is some edge case where you _don't_ want the hook to fire... What then? It's incredibly hard to tell Rails to "save/update this model but skip this one specific hook". You can skip _all_ hooks easily, but then you also skip validations which is not desirable. So this is really a non-trivial decision to make.

For the sake of this PR, please create the log entries manually where appropriate.

app/controllers/ticket_stakeholders_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+class TicketStakeholdersController < ApplicationController
+  before_action :authenticate_user!
+
+  def create
+    ticket = Ticket.find(params.require(:ticket_id))
+    connection = params.require(:connection)
+    stakeholder_role = params.require(:stakeholderRole)
+    is_active = ActiveRecord::Type::Boolean.new.cast(params.require(:isActive))
+
+    return head :unauthorized unless ticket.metadata.eligible_roles_for_bcc(current_user).include?(stakeholder_role)
```
gregorbg: I don't understand this whole `eligible_roles_for_bcc` concept. I understand BCC roles, and I understand how you want to use them, but the `eligible_for` part specifically confuses me.

Is it roles that the currently logged in user is allowed to create? If so, why do you restrict it only to BCC roles specifically? Admins should also be allowed to create non-BCC stakeholders, for example. (This is further increased by the fact that the controller method is called `TicketStakeholdersController#create`, which is very generic and open. It does not mention `create_bcc` or similar anywhere in the route name).

app/models/ticket.rb
```diff
@@ -30,7 +30,8 @@ def can_user_access?(user)
     return false if user.nil?
 
     ticket_stakeholders.belongs_to_user(user).any? ||
-      ticket_stakeholders.belongs_to_groups(user.active_groups).any?
+      ticket_stakeholders.belongs_to_groups(user.active_groups).any? ||
+      metadata.eligible_roles_for_bcc(user).any?
```
gregorbg: This addition makes the already expensive check even more expensive :(

The above two checks (the two which already existed before this PR) make sense to me because they directly relate to how the user may be directly linked to the current ticket. Either he's a related user or he's part of a related group. So the user can access.

But the "which BCC roles can this user create? Are there any BCC roles that this user can create?" logic confuses me. Isn't the point of BCC roles that someone creates this BCC role and _then_ after it is created the new BCC role is what gives the user access through one of the two already existing checks?

gregorbg: To put my question more simply: What part in your "user journey" breaks, if you don't add this line here in this PR?

app/controllers/ticket_stakeholders_controller.rb
danieljames-dj: > Is it roles that the currently logged in user is allowed to create

Yes

> If so, why do you restrict it only to BCC roles specifically? Admins should also be allowed to create non-BCC stakeholders, for example.

There are two concepts - roles and connections. `role` defines what actions they can do in tickets (eg: actioner, requester), and `connection` is like "assigned", "cc", "bcc", etc. "assigned" means the next action item will be on that one or multiple users. "cc" and "bcc" means they are part of the ticket and has a role. The difference between "cc" and "bcc" is that if we are sending any email notifications, we won't be considering "bcc" roles because they are mostly not part of the ticket.

Now coming to your question: we are restricting to only bcc roles to avoid letting them part of the ticket notifications. If they are actually supposed to be part of the ticket and get notifications, there will be other ways for them to join as cc or assignee. I can give some examples of 'other way': let's consider 'competition announcement ticket'. We have two scenerio here: 1. Board want to read the discussion happening, 2. WCAT want Board to read the discussion. For case 1, Board member can just go to the ticket and read the discussion like any other competition's tickets. But the Board member is not expecting an email notification just because they visited. Now coming to case 2, here WCAT will be doing an action "Get review from Board" (or something similar) and that action will add Board as a "cc" connection. This time definitely they should start receiving email notifications of that ticket till they are marked as "inactive".

Now coming to second part - "admins should also be allowed to create non-bcc". This won't help them in anyway. Without even the "non-bcc" role, "bcc" admins can see and do anything other people in the tickets can. So "bcc" role should be fine. Also once we start listing "stakeholders" in headers, seeing an admin in the stakeholders might look confusing/weird.

>  It does not mention create_bcc or similar anywhere in the route name

Actually this is a generic API to create stakeholder. Now thinking again, I feel this should be "add_bcc_connection" action. I'll explain this in the other comment where I asked suggestion on stakeholder.

app/controllers/ticket_stakeholders_controller.rb
danieljames-dj: Oh actually while typing the above comment, I realized something. We had decided on a concept like at a time, we will have only one action and for that action, there will be multiple changes. For example, when a ticket is created, the "create ticket" is the action and "adding stakeholders" is a "change".

I think for this as well the same thing matters. How about I create a new action "add_bcc_connection"? Adding stakeholders will be part of the "changes" within that. Do you agree with this?

app/models/ticket.rb
danieljames-dj: Okay I understand that this is bit expensive. Also this method is used in other places, so this might not be a good place to do it. I've moved this check to controller. Hope it works.

app/controllers/ticket_stakeholders_controller.rb
gregorbg: If the change is triggered by the same (one!) action, then yeah sure :)

app/controllers/ticket_stakeholders_controller.rb
gregorbg: Your concrete explanation with "WCAT and Board" on a competition announcement ticket makes a lot of sense, thanks!

I don't quite follow how this leads to your conclusion though. In your current implementation of `eligible_roles_for_bcc`, you pass in a user as argument. But the only use-case for that user is to check whether it's an admin or not, and only admins get an _actual_ result.
This feels to me like a "hidden permission check" or "undercover permission check". Can you please elaborate on the `eligible` part a bit more? Eligible for what? Eligible by whom?

```ruby
def eligible_roles_for_bcc(user)
  return [] unless user.admin?

  [
    TicketStakeholder.stakeholder_roles[:actioner],
    TicketStakeholder.stakeholder_roles[:requester],
  ]
end
```

app/models/ticket.rb
gregorbg: This moved the check to another place, but it does not fully clarify my concerns. This may be related to my other comment in another thread above, but here's what I'm understanding:

- The endpoint `show` in `tickets_controller` is for displaying _any_ ticket, no matter the metadata.
- Of course, it needs to check whether the user calling that endpoint has permissions to see the ticket. Obvious, no problem here.
- The permission is currently checked (even after your change!) based on three different factors (combined by an `||` boolean logic)
  1. Is the ticket linked to the user directly as a "user stakeholder"? --> Obvious, all good
  2. Is the ticket linked to the user through "group stakeholders"? (ie user is a member of WCAT, and WCAT is a group stakeholder of the ticket) --> obvious, all good
  3. [added in this PR] Is there any BCC role that the user _could_ potentially create, **but has not created it yet**

That third point is confusing the hell out of me. Why is that (3) giving access to the ticket already? I thought the point was that you get the BCC role added, and then **the already persisted existence of the role** is what grants you access via (1) or (2)...?

app/controllers/ticket_stakeholders_controller.rb
danieljames-dj: In the current case, only admins are allowed. Or in other words, in edit person ticket and results ticket, only admins are allowed to add themselves as the bcc role.

But for upcoming ticket types this won't be the case. For example:

- competition_announcement ticket
If Board wants to do some changes in the ticket as a 'actioner', they can do that. In that case, the `eligible_roles_for_bcc` method will return [:actioner] for Board.

- claim_wca_id ticket
Any delegate can come to this ticket and add themselves as the a bcc role `actioner`, so that they can transfer the claim request to themselves or to other delegates.

app/models/ticket.rb
gregorbg: pinging this comment @danieljames-dj

danieljames-dj: Oh okay, I think I understood your question - you are asking why I'm giving access to the user before they actually join the ticket using bcc role.

I actually thought of returning ticket even in that case just because it's more easier. Also I thought that the changes required in frontend will be too much, but it was easy.

Does this make sense now?

app/models/ticket_log.rb
```diff
@@ -4,6 +4,7 @@ class TicketLog < ApplicationRecord
   enum :action_type, {
     create_ticket: "create_ticket",
     create_comment: "create_comment",
+    join_as_bcc_stakeholder: "join_as_bcc_stakeholder",
```
gregorbg, line 7: Why is this action so specific? A simple "create stakeholder" could be enough.

app/controllers/tickets_controller.rb
```diff
@@ -70,9 +78,13 @@ def show
       end
       format.json do
         ticket = Ticket.find(params.require(:id))
+        bcc_eligible_roles = ticket.metadata.eligible_roles_for_bcc(current_user)
 
-        # Currently only stakeholders can access the ticket.
-        return head :unauthorized unless ticket.can_user_access?(current_user)
+        unless ticket.can_user_access?(current_user)
+          return render json: { eligible_roles_for_bcc: bcc_eligible_roles } if bcc_eligible_roles.any?
```
gregorbg: I don't like this, because it means that the same endpoint returns totally different JSON depending on who accesses it, and when they access it. This would be a nightmare to document in a public API.

gregorbg: I know you need this functionality, but then it has to be a separate endpoint.

app/models/ticket_log.rb
danieljames-dj: For every action in ticket, I'm trying to make the `action` as something that says "what action is done". Here, definitely the action is 'creating a stakeholder', but in other actions it can come as a 'change', and not 'action'. For example, let's assume for edit person ticket, delegate requests for a review from a delegate. Then the action is going to be `request_delegate_review` and one of the change within that will be to add delegate as the stakeholder.

app/controllers/tickets_controller.rb
danieljames-dj: Oh ya that's a better idea, implemented.

app/webpacker/components/Tickets/api/getTicketDetails.js
```diff
@@ -1,9 +1,17 @@
...
-  const { data } = await fetchJsonOrError(
+  const response = await fetchWithAuthenticityToken(
     actionUrls.tickets.show(ticketId),
   );
+
+  if (!response.ok) {
+    const error = new Error(`${response.status}: ${response.statusText}`);
+    error.status = response.status;
+    throw error;
+  }
```
gregorbg: Ehm... What is this? `fetchJsonOrError` literally has the error handling built into its name, why are you changing that here?

gregorbg: The thing that confuses me is that I'm seeing such a snippet in your PRs for the first time. Literally every other case of `fetchJsonOrError` works just fine like that, so I have very strong doubts that this is necessary here.

app/webpacker/components/Tickets/SelfRoleAssigner.jsx
```diff
@@ -0,0 +1,102 @@
...
+        </Modal.Content>
+        <Modal.Actions>
+          <Button
+            onClick={() => setIsAssigning(false)}
+            content="Cancel"
+          />
+          <Button
+            primary
+            disabled={!selectedRole}
+            onClick={createBccStakeholder}
```
gregorbg, line 95: When will the modal be closed? The `createBccStakeholder` is pretty much just forwarding directly to the mutation, and the mutation does not do any kind of `setIsAssigning(false)` in the success handler...

config/routes.rb
```diff
@@ -235,12 +235,14 @@
     post 'merge_inbox_results' => 'tickets#merge_inbox_results', as: :merge_inbox_results
     post 'post_results' => 'tickets#post_results', as: :post_results
     get 'edit_person_validators' => 'tickets#edit_person_validators', as: :edit_person_validators
+    get 'eligible_roles' => 'tickets#eligible_roles', as: :eligible_roles
```
gregorbg: The route name is a little bit generic. Eligible _for what_? The name should imply that it's about roles which you (the currently logged in user) are allowed to create new stakeholders for

app/webpacker/components/Tickets/api/getTicketDetails.js
danieljames-dj: I actually found an issue with `fetchJsonOrError`, that's why didn't use it.

In the frontend code, I needed a way to check the `response.status` - and will proceed with fetching `eligible_roles_for_bcc` only if I get `401`. But `fetchJsonOrError` will fail with following error message:

```
SyntaxError: Failed to execute 'json' on 'Response': Unexpected end of JSON input
    at fetchWithAuthenticityToken.js:16:1
...
```

Basically it's trying to convert the `response` to JSON using `response.json()` but it's failing.

After some more research, I figured out that I won't get this error if I actually send a JSON body along with status from the backend. I've fixed it now.

app/webpacker/components/Tickets/SelfRoleAssigner.jsx
danieljames-dj: On success, it will just invalidate `ticket-details`, making it fetch ticket details again, and this time it won't reach `SelfRoleAssigner` again (as there is going to be new stakeholder object), so modal object will be removed.

config/routes.rb
danieljames-dj: Okay, how about `eligible_roles_for_bcc`? Will that be fine?

gregorbg: Sounds good!

app/webpacker/components/Tickets/api/getTicketDetails.js
gregorbg: This is a good practice (sending JSON along with the response) to tell the frontend _what exactly_ went wrong. I don't think you should just listed to 401, because the 401 could be caused by a whole host of problems.
Rather, your backend should return something meaningful and unique in that specific JSON rendered on that specific line, so that the frontend can know "aha, this is not just any 401, this is *the particular* 401 that I want to catch!"

danieljames-dj: Okay, makes sense, done.

config/routes.rb
danieljames-dj: Changed

app/webpacker/components/Tickets/index.jsx
```diff
@@ -31,17 +32,27 @@ function Tickets({ id }) {
...
+      return stakeholders[0];
     }
     return null;
-  }, [ticketDetails?.requester_stakeholders, userSelectedStakeholder]);
+  }, [stakeholders, userSelectedStakeholder]);
 
   if (isPendingTicketDetails) return <Loading />;
-  if (isErrorTicketDetails) return <Errored error={errorTicketDetails} />;
+  if (isErrorTicketDetails) {
+    if (errorTicketDetails?.response?.status === 401 && errorTicketDetails?.message?.includes('No access to ticket')) {
```
gregorbg: Why do you do a whole bunch of `?.` null safe access here? This if-branch is only entered when you know that an error occured, so accessing the error null-safe feels redundant.

Also, you can just do the `errorTicketDetails.message` check. No need in my opinion to verify the 401

danieljames-dj: Regarding null safe - I agree it's not necessary as we already know that `errorTicketDetails` exists. I don't have much opinion on checking 401 or not, I'm fine with either way, I've removed it now.

gregorbg (approved):

PR #13157, opened 2026-01-11
Revise setup images instructions for clarity

r-oxana01:
Updated the setup images message in the Delegate Reports template to specify the need for current competition images and asks delegates to explain why they might not have current photos. Used to directly address resuage of old setup photos in reports.


config/locales/en.yml
```diff
@@ -505,7 +505,7 @@ en:
...
-        setup_images: "You MUST include at least %{image_count} images which illustrate the scrambling and competition areas (other images are also welcome). You can select multiple files at once. You can delete images only when enough other images have been uploaded (if you want to replace an image, you need to upload the new one first, and then delete the old one)."
+        setup_images: "You MUST include at least 2 images which illustrate the scrambling and competition areas from this current competition, prefferably during the competition (other images are also welcome). You can select multiple files at once. You can delete images only when enough other images have been uploaded (if you want to replace an image, you need to upload the new one first, and then delete the old one). If you do not have current pictures of the competition area and/or the scrambling area while in use, please explain why below."
```
sarahstrong314:
```suggestion
        setup_images: "You MUST include at least %{image_count} images which illustrate the scrambling and competition areas from this current competition, preferably during the competition (other images are also welcome). You can select multiple files at once. You can delete images only when enough other images have been uploaded (if you want to replace an image, you need to upload the new one first, and then delete the old one). If you do not have current pictures of the competition area and/or the scrambling area while in use, please explain why."
```

sarahstrong314: The text is displayed twice, once in the venue section and once under the upload button, so it's unclear what "below" refers to.

sarahstrong314: `%{image_count}` is needed since it's a variable stored in another file

dunkOnIT:
Sarah is correct that %{image_count} should not be replaced with `2`. Once that change is reverted, we can approve.

PR #13163, opened 2026-01-12
Replace value1..5 in sanity checks: attempt numbers

FinnIckler:
Also adds the new Attempt Number Gap Sanity Check from Sebastien

### DIFF
--- a/spec/factories/results.rb
+++ b/spec/factories/results.rb
@@ -123,7 +123,7 @@ FactoryBot.define do
 
     after(:build) do |result|
       result.result_attempts_attributes.each do |at|
-        result.result_attempts.build(attempt_number: at[:attempt_number], value: at[:value])
+        result.result_attempts.build(**at)
       end
     end
   end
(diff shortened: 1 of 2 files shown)


lib/sanity_check_sql/2 - irregular_results/13 - no_first_solve.sql
```diff
@@ -1,3 +1,4 @@
-SELECT *
-FROM results
-WHERE value1 = 0
+SELECT result_id
```
gregorbg, line 1: This will most likely affect exclusions, because the old sanity check is doing `SELECT *` and that gives a whole zoo of columns which all need to be included in the results of the new sanity check.

The other way round it would not be a problem, ie adding stuff to the output of an already existing check. You are removing stuff, so what about exclusions?

gregorbg (changes requested):

lib/sanity_check_sql/2 - irregular_results/13 - no_first_solve.sql
FinnIckler: no exclusions for that sanity check as mentioned. These are something that should never happen and if they do, should be fixed

PR #13164, opened 2026-01-12
Replace value1..5 in sanity checks: Wrong number of results

FinnIckler:


spec/factories/results.rb
```diff
@@ -120,5 +120,11 @@
     country_id { person.country_id }
     regional_single_record { nil }
     regional_average_record { nil }
+
+    after(:build) do |result|
+      result.result_attempts_attributes.each do |at|
+        result.result_attempts.build(attempt_number: at[:attempt_number], value: at[:value])
```
gregorbg:
```suggestion
        result.result_attempts.build(**at)
```

gregorbg: Also, didn't we have hooks for this on the model level?

lib/sanity_check_sql/2 - irregular_results/14 - wrong_number_of_results.sql
```diff
@@ -1,7 +1,12 @@
...
-  AND f.expected_solve_count <>
-      IF(value1 <> 0, 1, 0) + IF(value2 <> 0, 1, 0) + IF(value3 <> 0, 1, 0) + IF(value4 <> 0, 1, 0) +
-      IF(value5 <> 0, 1, 0)
+WITH num_solves AS (
+  SELECT result_id, COUNT(*) AS actual_solves
+  FROM result_attempts
+  GROUP BY result_id
+)
+
+SELECT r.id, r.person_id, r.competition_id, r.event_id, r.round_type_id, f.expected_solve_count, t2.actual_solves
```
gregorbg, line 7: Is the list of attributes returned here equivalent to the `SELECT *, r.id AS result_id` from the old query?
Asking to make sure that exclusions are not adversely affected

gregorbg (changes requested):

lib/sanity_check_sql/2 - irregular_results/14 - wrong_number_of_results.sql
FinnIckler: good point. We do need a "exclusion still works" test for each of these

spec/factories/results.rb
FinnIckler: just for after update!

FinnIckler: and you can also not call the method in the after build hook, because it's not persisted to the database yet so `attempts = self.result_attempts_attributes(result_id: self.id)` will still be nil here

lib/sanity_check_sql/2 - irregular_results/14 - wrong_number_of_results.sql
FinnIckler: well actually, there are 0 exclusions for these, because these are things that should be fixed, not excluded

PR #13192, opened 2026-01-13
Replace value1..5 in sanity checks: Different Result Types

FinnIckler:


spec/models/sanity_check_spec.rb
```diff
@@ -237,11 +237,11 @@
         round2 = create(:round, competition: competition2, event_id: "333oh")
 
         mo3_with_missing = create(:result, :mo3, event_id: "666", round: round1, competition: competition1)
-        mo3_with_missing.update_columns(value3: 0)
+        mo3_with_missing.result_attempts.find_by!(attempt_number: 3).delete
```
gregorbg, line 269: Your other PR(s) use the approach `r.result_attempts.where(attempt_number: 3).delete_all`. Please try to be consistent 😇

spec/models/sanity_check_spec.rb
```diff
@@ -237,11 +237,11 @@
...
         create(:result, :mo3, event_id: "666", round: round1, competition: competition1)
 
         bo5_with_missing = create(:result, round: round2, competition: competition2)
-        bo5_with_missing.update_columns(value5: 0)
+        bo5_with_missing.result_attempts.find_by!(attempt_number: 5).delete
```
gregorbg, line 273: Consistency, see above

gregorbg (changes requested):

spec/models/sanity_check_spec.rb
FinnIckler: yeah I will change that to use find_by! + delete instead

spec/models/sanity_check_spec.rb
FinnIckler: changed the other PRs

lib/sanity_check_sql/2 - irregular_results/15 - different_result_types.sql
gregorbg, line 1: This query looks fishy. The CTE in the beginning should only extract the result ID and the count of attempts for that result ID, other properties -- if interesting/relevant -- should be `JOIN`ed in via the `result_id` later on.

And what is `COUNT(DISTINCT num_attempts) AS different_result_types` all about? It may be the naming that is off here, but I cannot figure out why the "count of different numbers of attempts" has something to do with "result **types**".

FinnIckler: Sebastien rewrote this one based on the Eleanor query and I understand the different result types as different formats, meaning if there is a bo3 in an ao5. The query has been called different result types in the db for who knows as long

FinnIckler: And I don't necessarily agree with the other data being joined at the end, the final query works on `num_attempts_per_result`, why join twice?

gregorbg: Because you're currently violating `FULL_GROUP_BY` 🙃

gregorbg: And forgive me if I'm missing something, but I cannot see the term `different_result_types` in the old query at all

gregorbg (changes requested):

lib/sanity_check_sql/2 - irregular_results/15 - different_result_types.sql
FinnIckler: If I am violating that, I would get an error no?

FinnIckler: different_result_types used to be num_results

gregorbg: > different_result_types used to be num_results

That means, we're breaking all existing exclusions. Doesn't it?

FinnIckler: no exclusions exist

gregorbg: Okay. Per my chat with LLM, the grouping works in this particular case, because you're grouping by `result_id` which points to the primary key of `results`, and the MySQL engine is smart enough to recognize that "because this is the primary key, I can guarantee that the values will be unique".

But I would still argue it makes the query harder to read. I suggest the following:

```sql
WITH relevant_results AS (
    SELECT id, competition_id, event_id, round_type_id
    FROM results
    WHERE RIGHT(competition_id, 4) >= '2013'
),
attempt_counts AS (
    SELECT ra.result_id, COUNT(*) AS num_attempts
    FROM result_attempts ra
    INNER JOIN relevant_results rr ON ra.result_id = rr.id
    GROUP BY ra.result_id
)
SELECT 
    r.competition_id,
    r.event_id,
    r.round_type_id,
    COUNT(DISTINCT ac.num_attempts) AS different_result_types
FROM relevant_results r
JOIN attempt_counts ac ON ac.result_id = r.id
GROUP BY 
    r.competition_id, 
    r.event_id, 
    r.round_type_id
HAVING different_result_types > IF(round_type_id IN ('c', 'd', 'e', 'g', 'h'), 2, 1);
```

PR #13195, opened 2026-01-13
Replace value1..5 in sanity checks: Only DNS Results

FinnIckler:


lib/sanity_check_sql/2 - irregular_results/16 - only_dns_results.sql
gregorbg, line 1: Line breaks are your friend :)

gregorbg: Also, is there any reason why the old query did `ABS(1 + valueN) = 1` instead of `valueN = -2`??

FinnIckler: yes, because it also triggers if some of the values are 0

FinnIckler: and where do I need a line break?

gregorbg: > yes, because it also triggers if some of the values are 0

I guess we implicitly catch the "values are 0" case through the other "is the numbering of attempts consistent" sanity checks...?

> and where do I need a line break?

You actually need two line break**s** 😉 One before the `FROM` and one before the `HAVING`

FinnIckler: yes, especially through the invalid values for attempt_results check.

PR #13196, opened 2026-01-13
Replace value1..5 in sanity checks: Non Zero Average for Less Than 3 Results

FinnIckler:


lib/sanity_check_sql/2 - irregular_results/17 - non_zero_average_for_less_than_3_results.sql
gregorbg, line 1: This new query is currently just plain wrong, you most likely missed a line at the end while copy-pasting.
You definitely need some form of `WHERE average <> 0`.

FinnIckler:
test pass because there is no test for "it doesn't trigger for a correct missed cutoff result"

spec/factories/results.rb
gregorbg, line 1: This file looks like it needs a merge

PR #13197, opened 2026-01-13
Replace value 1 5 sanity checks duplicate results

FinnIckler:
I am struggling getting a test case working for 333mbo, maybe you can't create these with our factories?

This one definitely needs an exclusion check, because we have 3 exclusions for this in the db


lib/sanity_check_sql/3 - duplicate_results/18 - at_least_two_duplicate_results.sql
```diff
@@ -1,17 +1,27 @@
...
-       value2,
-       value3,
-       value4,
-       value5,
+WITH grouped_attempts AS (SELECT result_id,
+                                 MAX(CASE WHEN attempt_number = 1 THEN value END) AS value1,
+                                 MAX(CASE WHEN attempt_number = 2 THEN value END) AS value2,
+                                 MAX(CASE WHEN attempt_number = 3 THEN value END) AS value3,
+                                 MAX(CASE WHEN attempt_number = 4 THEN value END) AS value4,
+                                 MAX(CASE WHEN attempt_number = 5 THEN value END) AS value5,
```
gregorbg: Why in all heavens name are we abusing the new schema to replicate/emulate the old `valueN` schema?

FinnIckler: Eleanor wrote this query, but I assume this is because it needs to have the same columns in the query for exclusions

spec/models/sanity_check_spec.rb
```diff
@@ -172,6 +173,27 @@
...
+
+    it "Doesn't trigger for fmc or multiblind old style" do
+      competition = create(:competition, event_ids: ["333fm"])
+      round = create(:round, competition: competition, event_id: "333fm", format_id: "m")
+      create(:result, competition: competition, round: round, event_id: "333fm",
+                      value1: 21, value2: 22, value3: 23, value4: 0, value5: 0, best: 21, average: 2200, format_id: "m")
+      create(:result, competition: competition, round: round, event_id: "333fm",
+                      value1: 21, value2: 22, value3: 23, value4: 0, value5: 0, best: 21, average: 2200, format_id: "m")
+
+      # Currently getting the error: Validation failed: Format '1' is not allowed for '333mbo'
```
gregorbg, line 185: `333mbo` is currently not listed in `preferred_formats.json` at all. That's why you're getting this error.

As a "compromise solution" I would suggest creating the round as modern `333mbf` and then updating the round ID of the corresponding `competition_event` via an update call that doesn't trigger validations.

lib/sanity_check_sql/3 - duplicate_results/18 - at_least_two_duplicate_results.sql
gregorbg: Then I guess we may have to adjust the exclusions manually. Or resort to something like `GROUP_CONCAT`, which would allow us to list all solves (and it would be 5 or 3 in most cases). But I cannot allow the "there are only 5 solves" assumption to return!

FinnIckler: rewrote the query to use GROUP_CONCAT

spec/models/sanity_check_spec.rb
gregorbg: Now that you've fixed it, this two-line comment can go :)

spec/factories/results.rb
gregorbg, line 1: Merge `main` please

PR #13199, opened 2026-01-13
Replace value 1..5 sanity checks timelimit violations

FinnIckler:
48 uses Eleanor's unoptimized query and 49 sadly doesn't work at all because of GROUP BY restrictions. WIll work on this tomorrow.

### DIFF
--- a/lib/sanity_check_sql/12 - consistency_of_results_and_rounds_data/49 - timelimit_violations_cumulative.sql
+++ b/lib/sanity_check_sql/12 - consistency_of_results_and_rounds_data/49 - timelimit_violations_cumulative.sql
@@ -1,27 +1,10 @@
-WITH round_numbers AS (
+WITH aggregated_attempts AS (
   SELECT
-    t0.*,
-    ROW_NUMBER() OVER (
-      PARTITION BY t0.competition_id, t0.event_id
-      ORDER BY rt.`rank`
-      ) AS round
-  FROM (
-         SELECT DISTINCT
-           r.competition_id,
-           r.event_id,
-           r.round_type_id
-         FROM results r
-       ) t0
-         JOIN round_types rt ON t0.round_type_id = rt.id
-),
-     aggregated_attempts AS (
-       SELECT
-         result_id,
-         SUM(IF(value > 0, value, 0)) AS total_time
-       FROM result_attempts
-       GROUP BY result_id
-     )
-
+    result_id,
+    SUM(IF(value > 0, value, 0)) AS total_time
+  FROM result_attempts
+  GROUP BY result_id
+)
 SELECT r.person_id,
        r.competition_id,
        JSON_EXTRACT(ro.time_limit, '$.cumulativeRoundIds') AS rounds,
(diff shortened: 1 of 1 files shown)


spec/models/sanity_check_spec.rb
```diff
@@ -67,15 +67,15 @@
                                average: 300, best: 300)
 
       # Apply Timelimit violations
-      result.update_columns(value1: 302)
+      result.result_attempts.find_by!(attempt_number: 1).update(value: 302)
```
gregorbg: This `update` call will run validations, and thus probably fail silently with a `false`(?)

spec/models/sanity_check_spec.rb
```diff
@@ -67,15 +67,15 @@
...
 
-      result_ids = sanity_check.run_query.pluck("value1", "value2", "value3")
+      result_ids = sanity_check.run_query.pluck("value")
 
-      expect(result_ids).to contain_exactly([302, 300, 300])
+      expect(result_ids).to contain_exactly(302)
```
gregorbg, line 74: Why is the expected output of the sanity check changing if the test data isn't changing?

spec/models/sanity_check_spec.rb
```diff
@@ -67,15 +67,15 @@
...
     end
 
     it "Correctly identifies timelimit violations (cumulative), single round" do
-      sanity_check = SanityCheck.find(48)
+      sanity_check = SanityCheck.find(49)
```
gregorbg, line 78: What the...? Was the wrong sanity check under test before? If so, why did that test work?

spec/models/sanity_check_spec.rb
```diff
@@ -86,9 +86,9 @@
                                average: 300, best: 300)
 
       # Apply Timelimit violations
-      result.update_columns(value1: 302)
+      result.result_attempts.find_by!(attempt_number: 1).update(value: 302)
```
gregorbg: Same remark about the `update` call running validations as above

spec/models/sanity_check_spec.rb
```diff
@@ -112,9 +112,9 @@
                                average: 300, best: 300, person: person)
 
       # Apply Timelimit violations
-      result.update_columns(value1: 302)
+      result.result_attempts.find_by!(attempt_number: 1).update(value: 302)
```
gregorbg: Same about `update` running validations

gregorbg (changes requested):

spec/models/sanity_check_spec.rb
FinnIckler: No, check the WRT slack channel. Basically this query was named wrong (saying it only checked non cumulative cutoffs, but it did still check single round cumulative cutoffs) and Eleanor correctly fixed that with the new queries

spec/models/sanity_check_spec.rb
FinnIckler: because the new query is not outputting value1..n anymore

spec/factories/results.rb
gregorbg, line 1: This file wants a merge

spec/models/sanity_check_spec.rb
gregorbg: Seems like you missed this, should be `update_columns` too...?

PR #13200, opened 2026-01-14
Head-to-Head Results: Manual Import Support

dunkOnIT:
## Overview

The main objective of this PR is to get a prototype results import and display process working, so that we can collaborate with WRT and WMCT on these. That's also why the backend/frontend PR's are combined - so that I can deploy to staging and get WRT/WMCT input. 

There's also then some not-amazing stuff happening in the code - some hacky/inelegant stuff going on. Very happy to improve the code quality now that we have something workable.

Also, there are no validations/tests for 
- the CSV file that gets imported, or 
- the H2H results that get imported

That's something I'm aiming to work on in the week to come - until we have those, it'll rely on WRT's manual attention/sanity checks for the first competition's results.

## Questions
- [x] How on the model level do we actually progress from live_result/attempts to persisted results? Will a h2h_attempt belong_to both a live_attempt and a result_attempt? 
- [x] h2h_competitors is a little weird - it ends up having a result per user _per match_. This is unintuitive - you'd sort of expect a table of all the participants in the h2h round, and then to link to them from the `h2h_match`. But, of course, we can't do this because there are two competitors, and we don't want a `competitor_a`, `competitor_b` situation. So, I think this join table does make sense - perhaps it's a naming thing. `h2h_match_competitors` is probably the way to go
- [ ] Should I be doing anything with `live_result_history_entries`? I'm not at the moment

## Phase 1: Basic Import (just get something that works)
This will get us to the point where we have the technical capacity to import H2H results - though it will require manual validation of the CSV file, and WRT's alertness when reviewing/posting results, as there are no automated validations in place yet
- [x] Add models
- [x] Rake task to import into live results/attempts
  - [x] Write the job
  - [x] Import into my db and do a sanity check
  - [ ] Validate CSV import?
  - [x] Happy-path tests?
- [x] Rake task to post results from live results/attempts (doesn't need to be a full results posting process - we can rely on WRT to do the checking necessary)
- [x] Have results posting validators ignore rounds marked as H2H
  - [x] Find someone's actual results file, and test with it
- [x] Figure out how to populate `pos` for `results`

## Phase 2: Display the results
- [x] Show in Rankings/Results pages
- [x] Show on Competition Results page
- [x] Show on Scrambles page

## Phase 3: Add validations 
- [ ] Consider making `h2h_attempt` belong_to `scramble`
- [ ] Add validations for H2H results 
- [ ] Maaaybe basic validations on CSV input? 
 - [ ] Captured vs expected: Matches, sets, competitors, attempts, points
 - [ ] Competitors in each match must be unique, in the round, and have a logical/possible progression path

## Phase 4: Results posting
- [ ] Add a proper (with validations etc) results posting process for live_results/attempts -> results/result_attempts
- [ ] Add the ability to edit results with GUI

### DIFF
--- a/lib/tasks/h2h_results_import.rake
+++ b/lib/tasks/h2h_results_import.rake
@@ -55,7 +55,7 @@ namespace :h2h_results do
         )
 
         # And finally we can save the attempt's scramble
-        scramble_set = InboxScrambleSet.find_or_create_by!(matched_round_id: Round.find(round_id), scramble_set_number: scramble_set_number) do |set|
+        scramble_set = InboxScrambleSet.find_or_create_by!(matched_round_id: round_id, scramble_set_number: scramble_set_number) do |set|
           set.competition = set.matched_round.competition
           set.event = set.matched_round.event
           set.round_number = set.matched_round.number


.gitignore
dunkOnIT, line 1: Will remove this when the PR is no longer in draft

spec/lib/tasks/h2h_results_import_spec.rb
dunkOnIT: These tests are for convenience in development. There's a case for either commenting them out or removing them entirely when we (or, ideally, marking them as disabled in RSpec, but I'm not sure we have that kind of tag filtering at the moment)

app/models/concerns/resultable.rb
dunkOnIT, line 1: These can be uncommented once #12531 has been merged

spec/lib/tasks/h2h_results_import_spec.rb
```diff
@@ -0,0 +1,202 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require 'rake'
+
+RSpec.xdescribe "import:h2h_data", type: :task do
```
dunkOnIT: Note the `xdescribe`, which marks this as `Pending` and thus doesn't run it in our CI. We can remove these tests once we're no longer using the Rake task, if we want to remove them at that point

app/helpers/competitions_helper.rb
```diff
@@ -59,14 +59,21 @@ def people_to_sentence(results)
 
   def winners(competition, main_event)
     top_three = competition.results.where(event: main_event).podium.order(:pos)
+    h2h_finals = top_three.first.format_id == "h"
```
dunkOnIT, line 62: This pattern of inferring the round format from the first result will appear a lot in the code. It doesn't feel _great_, as obviously I'm inferring from a result not from a round - but I'm trying wherever possible to use the data we already have instead of injecting even more data, and the idea that all results from a given round will share the same round format seems reliable enough to use it for the kind of inference I'm doing here - at least as far as "get the code to a prototype state" goes

FinnIckler: is it even guarenteed that the first of these will be the main (i,e head to head) event?

dunkOnIT: Based on the query in line 61, it seems like they should all be from the main event?

dunkOnIT:
@thewca-bot deploy staging

app/webpacker/components/ResultsData/ViewData.js
```diff
@@ -66,9 +70,10 @@ function ResultsView({
           eventId={eventId}
           round={round}
           newEntryUrlFn={newEntryUrlFn}
-          DataRowHeader={DataRowHeader}
+          DataRowHeader={round?.results?.[0].format_id === 'h' ? H2hRowHeader : DataRowHeader}
```
gregorbg: Do rounds themselves not have `format_id`s?

app/webpacker/components/ResultsData/ViewData.js
```diff
@@ -66,9 +70,10 @@ function ResultsView({
...
           DataRowBody={DataRowBody}
           adminMode={adminMode}
+          isH2hRound={round?.results?.[0].format_id === 'h'}
```
gregorbg: Should be able to deduce that from the round, depending on what output format that round is in

lib/results_validators/advancement_conditions_validator.rb
```diff
@@ -43,9 +43,12 @@ def run_validation(validator_data)
         results_by_event_id.each do |event_id, results_for_event|
           results_by_round_id = results_for_event.group_by(&:round_id)
 
+          # We are filtering out H2H finals as their results are loaded via a different process until we transition
+          # all results to be loaded via the live_results/live_attempts pipeline. See #13200 for more information.
           rounds_without_deprecated_types = competition.rounds
                                                        .filter { it.event_id == event_id }
                                                        .reject { IGNORE_ROUND_TYPES.include?(it.round_type_id) }
+                                                       .reject { it.is_h2h_mock == true }
```
gregorbg:
```suggestion
                                                       .reject { it.is_h2h_mock? }
```

gregorbg: Auto-generated by Rails as a magic method

lib/results_validators/events_rounds_validator.rb
```diff
@@ -68,12 +68,9 @@ def check_events_match(competition, results)
...
         # Check for rounds which are entirely missing results.
-        # This is treated as a warning because under certain circumstances, you can plan for a round but not hold it.
-        #   The most "popular" reason for this is Regulation 9m (https://www.worldcubeassociation.org/regulations/#9m)
-        #   which effectively means that you wanted to hold three rounds of 3BLD but not enough people showed up.
-        # This scenario happens often enough that we allow Delegates to fix it "by themselves"
-        #   without WCAT intervention. That's why it's only a soft warning (as opposed to events being a hard error).
-        expected = competition.rounds.map(&:human_id)
+        # We are filtering out H2H finals as their results are loaded via a different process until we transition
+        #   all results to be loaded via the live_results/live_attempts pipeline. See #13200 for more information.
+        expected = competition.rounds.reject { it.is_h2h_mock == true }.map(&:human_id)
```
gregorbg:
```suggestion
        expected = competition.rounds.reject { it.is_h2h_mock? }.map(&:human_id)
```

db/migrate/20260113045553_add_h2_h_models.rb
```diff
@@ -0,0 +1,33 @@
...
+      t.integer :match_number, limit: 1, null: false
+      t.timestamps
+    end
+
+    create_table :h2h_match_competitors do |t|
+      t.references :h2h_match, null: false, foreign_key: true
+      t.references :user, type: :integer, null: false, foreign_key: true
+      t.timestamps
+    end
+    add_index :h2h_match_competitors, %i[h2h_match_id user_id], unique: true
```
gregorbg: This can be pulled into the `create_table` block as `t.index %i[...], unique: true

db/migrate/20260113045553_add_h2_h_models.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+class AddH2HModels < ActiveRecord::Migration[8.1]
+  def change
+    create_table :h2h_matches do |t|
+      t.references :round, type: :integer, null: false, foreign_key: true
+      t.integer :match_number, limit: 1, null: false
```
gregorbg, line 7: Is this a reasonable limitation? Can we imagine a future with a two-digit amount of rounds?

db/migrate/20260113045553_add_h2_h_models.rb
```diff
@@ -0,0 +1,33 @@
...
+    create_table :h2h_match_competitors do |t|
+      t.references :h2h_match, null: false, foreign_key: true
+      t.references :user, type: :integer, null: false, foreign_key: true
+      t.timestamps
+    end
+    add_index :h2h_match_competitors, %i[h2h_match_id user_id], unique: true
+
+    create_table :h2h_sets do |t|
+      t.references :h2h_match, null: false, foreign_key: true
+      t.integer :set_number, limit: 1, null: false
```
gregorbg, line 20: Again, is this a reasonable restriction?

db/migrate/20260113045553_add_h2_h_models.rb
```diff
@@ -0,0 +1,33 @@
...
+      t.timestamps
+    end
+
+    create_table :h2h_attempts do |t|
+      t.references :h2h_set, null: false, foreign_key: true
+      t.references :live_attempt, foreign_key: true
+      t.references :result_attempt, foreign_key: true
+      t.references :h2h_match_competitor, null: false, foreign_key: true
+      t.integer :set_attempt_number, limit: 1, null: false
```
gregorbg, line 29: Is this a reasonable restriction?

lib/tasks/h2h_results_import.rake
```diff
@@ -0,0 +1,186 @@
...
+  desc "Import H2H data from CSV"
+  task :import, [:file_path] => :environment do |_t, args|
+    file_path = args[:file_path]
+
+    # TODO: Add test that confirms support for multiple rounds in one file
+    ActiveRecord::Base.transaction do
+      CSV.foreach(file_path, headers: true) do |row|
+        puts "considering #{row}"
+
+        round_id = row['round_id'].to_i
```
gregorbg, line 15: What kind of ID are we expecting here? The numeric database primary key?

gregorbg: I'm a little bit baffled by the `to_i`

lib/tasks/h2h_results_import.rake
```diff
@@ -0,0 +1,186 @@
...
+        match_number   = row['match_number'].to_i
+        set_number     = row['match_set_number'].to_i
+        row['set_attempt_number'].to_i
+        registration_id = row['registration_id'].to_i
+        final_pos = row['final_position'].to_i
+        value = row['time_seconds'].sub(".", "").to_i
```
gregorbg, line 21: Is that a good practice for parsing? What if it's an "over one minute" time? I think we already have parsers in `SolveTime` or somewhere else

lib/tasks/h2h_results_import.rake
```diff
@@ -0,0 +1,186 @@
...
+        scramble_set_number = row['scramble_set_number'].to_i
+        scramble_number = row['scramble_number'].to_i
+        is_extra = row['is_extra'].to_i
+        scramble = row['scramble']
+
+        # First find/create the H2H infrastructure models
+        result = LiveResult.find_or_create_by!(registration_id: registration_id, round_id: round_id) do |lr|
+          lr.average = 0
+          lr.best = value
```
gregorbg, line 30: Huh?! Why does that make sense?

lib/tasks/h2h_results_import.rake
```diff
@@ -0,0 +1,186 @@
...
+
+        set = H2hSet.find_or_create_by!(h2h_match_id: match.id, set_number: set_number)
+
+        # Now we can create the attempt-specific records
+        live_attempt = LiveAttempt.create!(
+          value: value,
+          attempt_number: result.live_attempts.count + 1,
+          live_result: result,
+        )
+        puts live_attempt.errors unless live_attempt.valid?
```
gregorbg: I doubt you would ever reach this line because you're using `create!` above, which explodes loudly if something is not valid (and thus stops the whole code execution)

lib/tasks/h2h_results_import.rake
```diff
@@ -0,0 +1,186 @@
...
+
+        match = H2hMatch.find_or_create_by!(round_id: round_id, match_number: match_number)
+        competitor = H2hMatchCompetitor.find_or_create_by!(h2h_match_id: match.id, user_id: Registration.find(registration_id).user.id)
+
+        set = H2hSet.find_or_create_by!(h2h_match_id: match.id, set_number: set_number)
+
+        # Now we can create the attempt-specific records
+        live_attempt = LiveAttempt.create!(
+          value: value,
+          attempt_number: result.live_attempts.count + 1,
```
gregorbg, line 46: This is "hacky" at best. Can you not try to read all CSV rows first, then group them by some column(s) so that you can iterate over "the set of results per person"?

lib/tasks/h2h_results_import.rake
```diff
@@ -0,0 +1,186 @@
...
+          lr.best = value
+          lr.last_attempt_entered_at = DateTime.now
+          lr.global_pos = final_pos
+          lr.local_pos = final_pos
+        end
+
+        result.update!(best: value) if value.positive? && result.best > value
+
+        match = H2hMatch.find_or_create_by!(round_id: round_id, match_number: match_number)
+        competitor = H2hMatchCompetitor.find_or_create_by!(h2h_match_id: match.id, user_id: Registration.find(registration_id).user.id)
```
gregorbg:
```suggestion
        competitor = H2hMatchCompetitor.find_or_create_by!(h2h_match_id: match.id, user_id: Registration.find(registration_id).user_id)
```

gregorbg: No need to load the whole user into memory just to access its ID

lib/tasks/h2h_results_import.rake
```diff
@@ -0,0 +1,186 @@
...
+        )
+        puts live_attempt.errors unless live_attempt.valid?
+
+        h2h_attempt = H2hAttempt.create!(
+          h2h_set: set,
+          live_attempt: live_attempt,
+          h2h_match_competitor: competitor,
+          set_attempt_number: H2hAttempt.where(h2h_set_id: set, h2h_match_competitor_id: competitor).count + 1,
+        )
+        puts h2h_attempt.errors unless h2h_attempt.valid?
```
gregorbg: Again, this line seems redundant given the fact you're using `create!`

lib/tasks/h2h_results_import.rake
```diff
@@ -0,0 +1,186 @@
...
+              digit = number % 26
+              digit = 26 if digit.zero?
+
+              result << (digit + 64).chr
+              number -= digit
+              number /= 26
+            end
+
+            result.reverse
+          end
```
gregorbg: There's a helper method for this in `InboxScrambleSet#alphabetic_group_index`

lib/tasks/h2h_results_import.rake
dunkOnIT: Hmm, we _can_, I'm just reticent to change the implementation now that WRT has seen and signed off on the results in staging, and given that we don't have good tests to make sure that any new implementation works the same as the old one

dunkOnIT: I recognize it's not the most efficient approach, but the job runs so quickly that the inefficiency didn't seem relevant to me when writing the job - doesn't seem like it would cause issues for us on prod unless we had several thousand rows of H2H results being imported at one time - and we know we have a pretty low ceiling on how many H2H rows we'll get at any one time from a competition

app/webpacker/components/ResultsData/ViewData.js
dunkOnIT: They do not - see the console.log output: 
image

dunkOnIT: (That `roundTypeId` is just whether the round is 1st/2nd/final round)

app/webpacker/components/ResultsData/ViewData.js
dunkOnIT: See above, `round` doesn't have the requisite data - it isn't the full `round` object

db/migrate/20260113045553_add_h2_h_models.rb
dunkOnIT: [Regulation 9m](https://www.worldcubeassociation.org/regulations/#9m) specifies the maximum number of rounds as 4 - so this seems safe, barring radical change to either the WCA's competition structure, or what `round` means in our codebase. 

In either case, I think migrating the `limit` on this db column will be the least of our concerns 😅

db/migrate/20260113045553_add_h2_h_models.rb
dunkOnIT: Done

db/migrate/20260113045553_add_h2_h_models.rb
dunkOnIT: I think so for the foreseeable future - regs define `3` as the maximum number of sets. That would need to increase a lot for this to increase, and even if it does hit double-digits, we would need to write a code change to accommodate that anyway (even if just changing a `max_sets` constant), so we can include a migration in there to increase the `limit` property here

db/migrate/20260113045553_add_h2_h_models.rb
dunkOnIT: Perhaps the most likely to change - but again, 7 is the maximum number of attempts in a set per the regulations. I'd say we change it via a migration when the regulations tell us to

lib/results_validators/advancement_conditions_validator.rb
dunkOnIT: Done

lib/results_validators/events_rounds_validator.rb
dunkOnIT: Done

lib/tasks/h2h_results_import.rake
dunkOnIT: Yep, numeric database key. `to_i` is there to fail loudly if we get non-integer input

lib/tasks/h2h_results_import.rake
dunkOnIT: Because this is the first time we're seeing the result, so best will be whatever the value is - block only run on `create`

lib/tasks/h2h_results_import.rake
dunkOnIT: Even if the time is over a minute, we _should_ be getting the value in seconds - and I check this in the CSV - so converting from clock time to centiseconds isn't an issue. 

I clean the data in the incoming CSV to ensure that it always has 2 decimal places - could write a more complex parser to handle for those cases, but ensuring the integrity of the CSV in Google Sheets seems like the best thing in the short-term, before more validations are added in a follow-up PR

I couldn't find a parser for this case - I checked in SolveTime, as well as grepping for `clock_time`, and `centiseconds`/`to_centiseconds` etc. We do have functions that convert centiseconds to ClockTime and other formats, though - not sure if you were thinking of those maybe?

lib/tasks/h2h_results_import.rake
dunkOnIT: True :D

lib/tasks/h2h_results_import.rake
dunkOnIT: I realize now (looking at the code in context) that you were talking about attempt numbering, not efficiency. I still think the concerns about changing the numbering at this stage apply - although I concede that I can export the results of the job before/after and compare those results to ensure that nothing changed in _this run_, at least.

lib/tasks/h2h_results_import.rake
dunkOnIT: Correct

lib/tasks/h2h_results_import.rake
dunkOnIT: Correct

lib/tasks/h2h_results_import.rake
dunkOnIT: Amazing, changed

app/webpacker/components/ResultsData/ViewData.js
gregorbg: Then you should look at the source of the data, which can be traced through the developer console (by just seeing which URL it hits) and then opening the corresponding Rails controller.

In this case I believe it is `controllers/api/v0/competitions_controller`, method `event_results` and `event_scrambles`. You can add the H2H mock information there and then access it here in the frontend.

gregorbg: There is also `controllers/admin/results_controller` method `show_events_data`, which TBH I'm not too sure where it's used or whether H2H information is needed there. Just wanted to point it out.

app/controllers/api/v0/competitions_controller.rb
```diff
@@ -100,6 +100,7 @@ def event_results
       {
         id: round.id,
         roundTypeId: round.round_type_id,
+        isH2hMock: round.is_h2h_mock,
```
gregorbg: Perhaps this should use the `is_h2h_mock?` magic method for proper boolean casting?

lib/tasks/h2h_results_import.rake
```diff
@@ -0,0 +1,167 @@
...
+          value: value,
+          attempt_number: result.live_attempts.count + 1,
+          live_result: result,
+        )
+
+        h2h_attempt = H2hAttempt.create!(
+          h2h_set: set,
+          live_attempt: live_attempt,
+          h2h_match_competitor: competitor,
+          set_attempt_number: H2hAttempt.where(h2h_set_id: set, h2h_match_competitor_id: competitor).count + 1,
```
gregorbg:
```suggestion
          set_attempt_number: H2hAttempt.where(h2h_set: set, h2h_match_competitor: competitor).count + 1,
```

lib/tasks/h2h_results_import.rake
```diff
@@ -0,0 +1,167 @@
...
+        )
+
+        # And finally we can save the attempt's scramble
+        scramble_set = InboxScrambleSet.find_or_create_by!(matched_round: Round.find(round_id), scramble_set_number: scramble_set_number) do |set|
```
gregorbg:
```suggestion
        scramble_set = InboxScrambleSet.find_or_create_by!(matched_round_id: round_id, scramble_set_number: scramble_set_number) do |set|
```

lib/tasks/h2h_results_import.rake
gregorbg: Yeah I didn't realize that you're actually forcing organizers to denote "1 minute 15 seconds" as `75.xy`. I thought they would write `1:15.xy` into the CSV

gregorbg (changes requested):
Tiny bit of cosmetics, otherwise good to go!

lib/tasks/h2h_results_import.rake
dunkOnIT: Just noting (as mentioned in the call) that organizers are not inputting data directly into this CSV. 

There is a dedicated scoretaking spreadsheet that Alvaro gives the organizers where they enter results - and then, Alvaro has a process that converts those results into this CSV format.

No one is being forced to manually converted minutes into seconds - the spreadsheet process is handling that.

lib/tasks/h2h_results_import.rake
dunkOnIT: Changed

app/controllers/api/v0/competitions_controller.rb
```diff
@@ -100,6 +100,7 @@ def event_results
                           {
                             id: round.id,
                             roundTypeId: round.round_type_id,
+                            isH2hMock: round.is_h2h_mock,
```
gregorbg:
```suggestion
                            isH2hMock: round.is_h2h_mock?,
```

PR #13201, opened 2026-01-14
Remove outdated information from scrambles page

pranav-027:
Correct me if  I am wrong here but I have been generating scrambles since years now and I have never seen this folder in my directory


gregorbg:
Older versions do create this folder, so maybe you could leave the paragraph there and just add a few words to make clear that this behavior depends on the version you're using?

PR #13202, opened 2026-01-14
Sanity Check: Invalid Cumulative Time Limit

FinnIckler:
New Sanity Check from Eleanor, because the wrong data currently causes false positives for the cumulative timelimit across round sanity check.


lib/static_data/sanity_check_categories.json
gregorbg, line 1: Why does this need to go into an entirely new category (with only one entry)?
We already have `9 - suspicious round data`...

lib/static_data/sanity_checks.json
```diff
@@ -474,5 +474,12 @@
     "topic": "Invalid Attempt Result Values",
     "comments": null,
     "query_file": "invalid_attempt_result_values.sql"
+  },
+  {
+    "id": "70",
+    "sanity_check_category_id": "9",
+    "topic": "Invalid Time Cumulative Limit",
```
gregorbg: Shouldn't it be "Invalid Cumulative Time Limit"...?

FinnIckler: yeah..

PR #13204, opened 2026-01-15
Support USD to USD conversion in Money::Bank::CurrencylayerBank

danieljames-dj:
Currently the dues calculation is failing for USD, because the library is not allowing to convert USD to USD as the exchange rate is not available.

Added the patch fix for the same.

Alternative considered: Modify dues_calculator.rb to check if currency code is USD, but it will make the code more complicated. Also I feel converting from USD to USD should be allowed by the library ideally.

### DIFF
--- a/config/initializers/monkey_patches.rb
+++ b/config/initializers/monkey_patches.rb
@@ -162,20 +162,4 @@ Rails.configuration.to_prepare do
       end
     end
   end
-
-  # Temporary fix until https://github.com/phlegx/money-currencylayer-bank/pull/20
-  # to allow exchanging from a currency to itself
-  Money::Bank::CurrencylayerBank.class_eval do
-    # 1. Create an alias for the original method so we don't lose it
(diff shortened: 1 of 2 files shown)


danieljames-dj:
I actually considered editing `dues_calculator`, but I feel since ruby-money supports this, I wish CurrencylayerBank also supported this and hence touched the moneypatch instead.

`safe_exchange_to` also looks interesting - I can actually create a file named `money_helper` inside `helpers` directory and use that. Should I go with that solution instead?

Also I assume if I go with that, I won't no longer be able to do like `money_object.exchange_to` instead should use `exchange_to(money_object)`. Or I should define the method in monkeypatch. Would like to get your advice on which way to proceed.

danieljames-dj:
Oh that was a great catch @gregorbg, thanks for that. I just realized that we were throwing error if the exchange rate is not found and that was the reason. I've now fixed it.

PR #13205, opened 2026-01-16
Leverage round table joining in record checking script

gregorbg:
See title.

### DIFF
--- a/spec/models/sanity_check_spec.rb
+++ b/spec/models/sanity_check_spec.rb
@@ -490,6 +490,7 @@ RSpec.describe SanityCheck do
     end
   end
 
+  # rubocop:disable Layout/LineLength
   context "incorrect record assignments", :clean_db_with_truncation do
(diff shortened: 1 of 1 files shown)


gregorbg:
Tests added :upside_down_face:

spec/models/sanity_check_spec.rb
```diff
@@ -489,4 +489,125 @@
       end
     end
   end
+
+  # rubocop:disable Layout/LineLength
+  context "incorrect record assignments", :clean_db_with_truncation do
```
FinnIckler, line 494: :clean_db_with_truncation still here because of your development, or does it have to be here?

gregorbg: It has to be there in order to run `AuxiliaryDataComputation.compute_concise_results` with all of its table rotation logic and stuff. That doesn't work with the default `SAVEPOINT` mechanics of our RSpec suite.

FinnIckler (approved):

PR #13209, opened 2026-01-17
Download wfc export in english

danieljames-dj:
WFC had a case where some exports had the dues estimate in the form "123,45" instead of "123.45" and this affects some of our formula. So this change is to force the WFC export in english


danieljames-dj:
> ...which definitely needs to be a separate PR. The English change is a fix which I am (was) willing to accept "just like that". But your other change is a feature which needs tests and a separate discussion. It's completely bugged in the way you're trying to "sneak it in" here and will not behave the way you wanted it to behave.

Done

gregorbg (approved):
Thank you!

PR #13211, opened 2026-01-17
Exempt first 5 competitions from dues

danieljames-dj:
The WCA Dues if exempt from dues for the first 5 competitions in a country.

### DIFF
--- a/spec/models/competition_spec.rb
+++ b/spec/models/competition_spec.rb
@@ -1239,10 +1239,7 @@ RSpec.describe Competition do
   describe "is exempt from dues" do
     let(:four_by_four) { Event.find "444" }
     let(:fmc) { Event.find "333fm" }
-
-    before do
-      create_list(:competition, 5, country_id: "Canada", city_name: "Vancouver, British Columbia", start_date: 2.years.ago, end_date: 2.years.ago + 2.days)
-    end
+    let!(:initial_competitions) { create_list(:competition, 5, country_id: "Canada", city_name: "Vancouver, British Columbia", start_date: 2.years.ago, end_date: 2.years.ago + 2.days) }
 
     it "is false when competition has no championships" do
       competition = create(:competition, events: [four_by_four], championship_types: [], country_id: "Canada", city_name: "Vancouver, British Columbia")
@@ -1299,17 +1296,19 @@ RSpec.describe Competition do
       expect(comp_6.exempt_from_wca_dues?).to be true
     end
 
-    it "is false for the 6th competition if it starts after the first 5" do
+    it "is false for the 6th competition if it starts after the first 5 and true for the first 5 competitions" do
       comp_6 = create(:competition, country_id: "Canada", city_name: "Vancouver, British Columbia", start_date: Date.today, end_date: Date.today + 2.days)
 
       expect(comp_6.exempt_from_wca_dues?).to be false
+      expect(initial_competitions).to all(be_exempt_from_wca_dues)
     end
 
-    it "is false if 6th competition is happening after many years" do
-      create_list(:competition, 5, country_id: "Korea", start_date: 7.years.ago, end_date: 7.years.ago + 2.days)
+    it "is false if 6th competition is happening after many years and true for first 5 competitions" do
+      first_competitions = create_list(:competition, 5, country_id: "Korea", start_date: 7.years.ago, end_date: 7.years.ago + 2.days)
       comp_6 = create(:competition, country_id: "Korea", start_date: Date.today, end_date: Date.today + 2.days)
 
       expect(comp_6.exempt_from_wca_dues?).to be false
+      expect(first_competitions).to all(be_exempt_from_wca_dues)
     end
 
     it "is true for 6th competition if one of the first 5 competitions are multi-national FMC competition" do


spec/models/competition_spec.rb
```diff
@@ -1240,6 +1240,10 @@ def change_and_check_activities(new_start_date, new_end_date)
     let(:four_by_four) { Event.find "444" }
     let(:fmc) { Event.find "333fm" }
 
+    before do
+      create_list(:competition, 5, country_id: "Canada", city_name: "Vancouver, British Columbia", start_date: 2.years.ago, end_date: 2.years.ago + 2.days)
+    end
+
```
gregorbg: If you do `let!` instead of `let` (note the exclamation mark in the end) then the code block will be executed immediately

gregorbg: It's basically a "clean syntax" version of the `before` block you're doing

spec/models/competition_spec.rb
```diff
@@ -1240,6 +1240,10 @@ def change_and_check_activities(new_start_date, new_end_date)
     let(:four_by_four) { Event.find "444" }
     let(:fmc) { Event.find "333fm" }
 
+    before do
+      create_list(:competition, 5, country_id: "Canada", city_name: "Vancouver, British Columbia", start_date: 2.years.ago, end_date: 2.years.ago + 2.days)
```
gregorbg: Why are you creating the data for Canada in this `before` block, but lots of sample data for other countries within the individual test bodies themselves?

spec/models/competition_spec.rb
```diff
@@ -1279,6 +1283,42 @@ def change_and_check_activities(new_start_date, new_end_date)
...
+      comp_6 = create(:competition, country_id: "Germany", start_date: 1.month.ago, end_date: 1.month.ago + 3.days)
+
+      expect(comp_5.exempt_from_wca_dues?).to be true
+      expect(comp_6.exempt_from_wca_dues?).to be true
+    end
+
+    it "is false for the 6th competition if it starts after the first 5" do
+      comp_6 = create(:competition, country_id: "Canada", city_name: "Vancouver, British Columbia", start_date: Date.today, end_date: Date.today + 2.days)
+
+      expect(comp_6.exempt_from_wca_dues?).to be false
```
gregorbg, line 1302: I think you should also test whether the `exempt_from_wca_dues?` for the first five competitions in the "Canada list" return `true`.

spec/models/competition_spec.rb
```diff
@@ -1279,6 +1283,42 @@ def change_and_check_activities(new_start_date, new_end_date)
...
+    end
+
+    it "is false if 6th competition is happening after many years" do
+      create_list(:competition, 5, country_id: "Korea", start_date: 7.years.ago, end_date: 7.years.ago + 2.days)
+      comp_6 = create(:competition, country_id: "Korea", start_date: Date.today, end_date: Date.today + 2.days)
+
+      expect(comp_6.exempt_from_wca_dues?).to be false
```
gregorbg, line 1310: Again, you should also check whether all 5 previous Korean competitions return `true`.

spec/models/competition_spec.rb
```diff
@@ -1279,6 +1283,42 @@ def change_and_check_activities(new_start_date, new_end_date)
...
+    end
+
+    it "is true for 6th competition if one of the first 5 competitions are multi-national FMC competition" do
```
gregorbg, line 1314: Is this what WFC wants?

spec/models/competition_spec.rb
danieljames-dj: Oh that's nice, changed to `let!`, thanks,

spec/models/competition_spec.rb
danieljames-dj: I was not able to create this with `let`, that's why I used `before`. But your suggestion of `let!` helped. Regarding country, I just randomly selected from other examples in this file.

spec/models/competition_spec.rb
danieljames-dj: Done

spec/models/competition_spec.rb
danieljames-dj: Done

spec/models/competition_spec.rb
danieljames-dj: Yes, WFC is not considering multi-national FMC competition as a competition of that country (just like how the website is not considering it as a competition of that country).

I've anyway posted a message in WFC group for double confirmation.

spec/models/competition_spec.rb
```diff
@@ -1279,6 +1280,44 @@ def change_and_check_activities(new_start_date, new_end_date)
...
+
+      expect(comp_5.exempt_from_wca_dues?).to be true
+      expect(comp_6.exempt_from_wca_dues?).to be true
+    end
+
+    it "is false for the 6th competition if it starts after the first 5 and true for the first 5 competitions" do
+      comp_6 = create(:competition, country_id: "Canada", city_name: "Vancouver, British Columbia", start_date: Date.today, end_date: Date.today + 2.days)
+
+      expect(comp_6.exempt_from_wca_dues?).to be false
+      expect(initial_competitions).to all(be_exempt_from_wca_dues)
```
gregorbg, line 1303: Just curious: Where is this `be_exempt_from_wca_dues` method defined?

danieljames-dj: That’s defined in `competition.rb` as `exempt_from_wca_dues`. I’m leveraging the RSpec 'be' matcher convention.

PR #13250, opened 2026-01-21
Infrastructure changes: Allow nextjs to talk to rails over the internal network

FinnIckler:
Includes a couple terraform maintenance changes (updating the lambda bundler command and updating terraform version to support b/g deployment).

We accomplish this by turning on [AWS Service Connect](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/service-connect.html) for both nextjs (client only) and rails (server and client). This allows nextjs to talk to the rails container using the rails.local DNS name.

You can imagine this as nextjs and rails being in the same docker compose, but load balancing between all the different IPs.

For deployment we can now trigger one the same way as with the other services instead of implicitly after pushing the new image.
The only thing I haven't migrated yet is the deployment notification, as this needs a lambda function now (I think).

### DIFF
--- a/infra/wca_on_rails/production/rails.tf
+++ b/infra/wca_on_rails/production/rails.tf
@@ -148,6 +148,7 @@ locals {
           dump_replica_host: "dev-dump-worldcubeassociation-dot-org.comp2du1hpno.us-west-2.rds.amazonaws.com"}))
     }
   ]
+  rails_internal_dns = "rails.local"
 }
 
 data "aws_iam_policy_document" "task_assume_role_policy" {
@@ -317,7 +318,7 @@ resource "aws_ecs_task_definition" "this" {
 }
 
 resource "aws_service_discovery_private_dns_namespace" "this" {
-  name = "rails.local"
+  name = local.rails_internal_dns
   vpc  = var.shared.vpc_id
 }
 
@@ -386,7 +387,7 @@ resource "aws_ecs_service" "this" {
       discovery_name = "rails-cluster"
       client_alias {
         port = 3000
-        dns_name = "rails.local"
+        dns_name = local.rails_internal_dns
       }
     }
   }


infra/wca_on_rails/main.tf
```diff
@@ -6,7 +6,7 @@ terraform {
   required_providers {
     aws = {
       source  = "hashicorp/aws"
-      version = ">= 6.14.1"
+      version = ">= 6.28.0"
```
gregorbg, line 9: Will this put us into weird "AMI image memory changed by 3MB, letting our deployments fail" situations again?

(Sidenote: What came of that issue which you posted to AWS? I forgot to bookmark it, so now I cannot find it)

infra/wca_on_rails/production/rails.tf
```diff
@@ -336,21 +370,28 @@ resource "aws_ecs_service" "this" {
...
-    ]
+  service_connect_configuration {
+    enabled = true
+    namespace = aws_service_discovery_private_dns_namespace.this.name
+    service {
+      port_name = "rails-port"
+      discovery_name = "rails-cluster"
+      client_alias {
+        port = 3000
+        dns_name = "rails.local"
```
gregorbg: Is it possible (and reasonable) to cross-reference the `resource "aws_service_discovery_private_dns_namespace"` here?

The idea is to reduce the number/amount of hard-coded `"rails.local"` strings

infra/wca_on_rails/main.tf
FinnIckler: nah, that isn't dependent on the terraform version

infra/wca_on_rails/production/rails.tf
FinnIckler: I can put it into a variable

PR #13254, opened 2026-01-21
Fix edit person "execute as update" feature in tickets

danieljames-dj:
There was a bug in the code because of which the 'execute as update' action  was not working in tickets. I've also wrote some unit tests for the same.

### DIFF
--- a/spec/factories/tickets_edit_person.rb
+++ b/spec/factories/tickets_edit_person.rb
@@ -3,7 +3,7 @@
 FactoryBot.define do
   factory :tickets_edit_person do
     status { :open }
-    wca_id { FactoryBot.create(:user_with_wca_id).wca_id }
+    person { association(:person) }
 
     trait :default_stakeholders do
       after(:create) do |edit_person_ticket|


spec/controllers/tickets_controller_spec.rb
```diff
@@ -182,4 +182,76 @@
...
+    before do
+      sign_in wrt_member
+    end
+
+    context "when the person's data is out of sync" do
+      let(:edit_name_ticket) { create(:edit_name_ticket) }
+
+      before do
+        edit_name_ticket.person.update!(name: "John Doe")
+      end
```
gregorbg, line 198: It feels redundant to immediately update a ticket which you've just (mock-)created. Why not directly create it with the desired name?

The idea is that inside your factory for `tickets_edit_person`, instead of `wca_id { FactoryBot.create(:user_with_wca_id).wca_id }` you rely on the fact that the actual Rails model has a `belongs_to :person` defined. So inside the factory you can do `person { association(:user_with_wca_id) }`.

I am 70% confident that this `person` initializer is enough, and FactoryBot will initialize `wca_id` "under the hood" for you, because that's the `foreign_key` that defines the `belongs_to` in the model. I am not sure though and you will have to check this.

Anyways, once you've set up this `person { ...whatever... }` in the factory, you can `let(:john_doe) { create(:user_with_wca_id, name: "John Doe") }` here, and then `let(:edit_name_ticket) { create(:edit_name_ticket, person: john_doe) }`.

danieljames-dj: Actually I do have that. While the ticket is created, it gets created with a default name. I'm trying to override that name "after" the ticket is created, so that the "execute" will fail with `out_of_sync` error.

gregorbg: Is it important to change the name _after_ the creation? My suggestion is to create a person for the ticket with the name "John Doe", and then pass that "John Doe person" to the ticket right in the moment of creation.

The background is that it is much more clean to define `person { association :user_with_wca_id }` in the factory than it is to do `wca_id { create(:user_with_wca_id).wca_id }`. Is there something I'm missing why this approach won't work for you?

danieljames-dj: > Is there something I'm missing why this approach won't work for you?

So I'm trying to reproduce what happens in a real world scenario. In real world, following is the sequential order:

- A user requested for a name change from "John Doe" to "John A. Doe". At this point, a ticket will be created with `old_value` = the current person name in database
- After 1 hour, a WRT member manually changed the name to "John B. Doe" using PMA
- After another 1 hour, a WRT member tries to approve the data assuming the request is from "John Doe" (old value) to "John A. Doe" (new value), but the name has already changed to "John B. Doe" that the ticket is not aware about. So this should throw error.

I understand that your suggestion will do the exact same thing, but I wanted the unit-test to behave the same way as it happens - like first do action 1, then action 2 and then action 3. In your case, action 1 & action 2 are somehow combined.

I hope this clarifies why I prefer to stick with my approach.

gregorbg: Okay, in this case at least you can still use the `association` approach in the factory but leave everything here in the "spec body" as-is.

gregorbg: As a general rule of thumb, using `FactoryBot.create` (or `FactoryBot.build`) directly in factories should be the exception rather than the rule, and is generally only okay in `after_build` or `after_create` hooks to create associated stuff (like mocking a schedule for a competition that you just created with the `:with_valid_schedule` trait).
But within the "raw properties" of the factory object under construction, `FactoryBot.create` and `FactoryBot.build` should raise some red flags.

FWIW, I know we're doing this a lot in legacy factories, but that's probably because the `association` approach wasn't so well developed back then.

danieljames-dj: Okay understood, I've made the change.

PR #13260, opened 2026-01-23
increase AWS service connect request timeout to 60 seconds

FinnIckler:
The default aws service connect time out is 15 seconds, this increases it to 60 so long running wcif syncs don't die. Increasing it to more than 60 seconds doesn't make sense because we have 60 seconds timeout at the Rails and Load Balancer level

### DIFF
--- a/infra/wca_on_rails/production/rails.tf
+++ b/infra/wca_on_rails/production/rails.tf
@@ -291,6 +291,7 @@ resource "aws_ecs_task_definition" "this" {
           hostPort = 3000
           containerPort = 3000
           protocol      = "tcp"
+          appProtocol = "http"
         },
       ]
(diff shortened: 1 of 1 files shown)


infra/wca_on_rails/production/rails.tf
```diff
@@ -291,6 +291,7 @@ resource "aws_ecs_task_definition" "this" {
           hostPort = 3000
           containerPort = 3000
           protocol      = "tcp"
+          appProtocol = "http"
```
gregorbg, line 294: Where does the encryption come into play? Of course we want http*S* on the end user level

FinnIckler: TLS is terminated on our external load balancer

PR #13261, opened 2026-01-23
Save synced Live Results in LiveResults model

### LINKED ISSUE #13691: WCA Live -> ILR Forwards Compatibility
- [ ] Populate LiveResults model when WCIF is patched (this will allow ILR to pick up WCA Live results)

FinnIckler:
opening in draft so I can check it out from my laptop


app/models/live_result.rb
```diff
@@ -75,6 +75,49 @@ def values_for_sorting
     end
   end
 
+  def to_inbox_result
+    attempt_values = result.attempts.map(&:value)
```
gregorbg: Shouldn't this be `live_attempts`? What is this `result.` method (or relation??) that you're accessing here?

app/models/live_result.rb
```diff
@@ -75,6 +75,49 @@ def values_for_sorting
...
+    InboxResult.new({
+                      competition: competition,
+                      person_id: result.person_id,
+                      pos: result.ranking,
+                      event_id: round.event_id,
+                      round_type_id: round.round_type_id,
+                      round_id: round.id,
+                      format_id: round.format_id,
+                      best: result.best,
+                      average: result.average,
```
gregorbg: Potentially copy-pasta, see my other comment

app/models/round.rb
```diff
@@ -329,6 +329,48 @@ def self.parse_wcif_id(wcif_id)
     { event_id: event_id, round_number: round_number }
   end
 
+  def load_live_results(results_wcif)
```
gregorbg: Per the fact that this uses `insert_all!` internally, you should also call this method `load_live_results!`

app/models/round.rb
```diff
@@ -329,6 +329,48 @@ def self.parse_wcif_id(wcif_id)
...
+    live_results.destroy_all
```
gregorbg: Also, due to this `destroy_all`, the whole method (or at least the integral parts of it) need to be wrapped in one big transaction

gregorbg (changes requested):

app/models/live_result.rb
FinnIckler: attempts is an alias_method of live_attempts. But yeah, it should just be attempts

app/models/live_result.rb
FinnIckler: what other comment?

gregorbg: https://github.com/thewca/worldcubeassociation.org/pull/13261/changes#r2764041175. You're calling a lot of `result.` accessors, and I cannot figure out where they're coming from.

You're already in the `live_result.rb` model, so it should be `self.person_id`, `self.ranking`, `self.event_id`, etc. Not `result.person_id` etc. Or am I missing something?

app/models/round.rb
```diff
@@ -371,7 +412,7 @@ def to_wcif
       "cutoff" => cutoff&.to_wcif,
       "advancementCondition" => advancement_condition&.to_wcif,
       "scrambleSetCount" => self.scramble_set_count,
-      "results" => round_results.map(&:to_wcif),
+      "results" => live_results.map(&:to_wcif),
```
gregorbg: Since you're now accessing an actual table relation (instead of just a field that was arbitrarily serialized into one column) you should probably look out for `includes` calls around the code, at least in the API controller that directly serves WCIF, and add `live_results` at the appropriate spot in there

app/models/live_result.rb
```diff
@@ -113,6 +113,49 @@ def self.compute_diff(before_result, after_result)
...
 
+  def to_inbox_result
+    attempt_values = attempts.map(&:value)
+    InboxResult.new({
+                      competition: competition,
+                      person_id: registration.person_id,
+                      pos: local_pos,
+                      event_id: round.event_id,
+                      round_type_id: round.round_type_id,
+                      round_id: round.id,
```
```suggestion
                      round_id: round_id,
```

app/models/round.rb
gregorbg: As far as I can tell, the `!` has not made it into the method name yet

app/models/round.rb
```diff
@@ -368,6 +368,50 @@ def self.parse_wcif_id(wcif_id)
...
+        {
+          registration_id: registrations_by_wcif_id[result_wcif["personId"]].id,
+          round_id: self.id,
+          last_attempt_entered_at: now,
+          best: result_wcif["best"],
+          average: result_wcif["average"],
+          global_pos: result_wcif["ranking"],
+          local_pos: result_wcif["ranking"],
+          created_at: now,
+          updated_at: now,
```
gregorbg: The documentation of `insert_all` lists a `record_timestamps` option with the following description:
> By default, automatic setting of timestamp columns is controlled by the model's record_timestamps config, matching typical behavior.
>
> To override this and force automatic setting of timestamp columns one way or the other, pass :record_timestamps:

To me, this implies unless you _explicitly_ turn timestamp writing OFF (either on the model level or here within the `insert_all` call), it will be `ON` by default. So you won't need to manually pass something

app/models/round.rb
```diff
@@ -368,6 +368,50 @@ def self.parse_wcif_id(wcif_id)
     ScheduleActivity.parse_activity_code(wcif_id)
   end
 
+  def load_live_results(results_wcif)
+    self.transaction do
+      live_results.destroy_all
+      registrations_by_wcif_id = competition.registrations.index_by(&:registrant_id)
```
gregorbg: Are the registrations prt of the `includes` in the controller method of the `PATCH` endpoint?

app/models/round.rb
```diff
@@ -368,6 +368,50 @@ def self.parse_wcif_id(wcif_id)
...
+      # Reload to get the generated IDs
+      results_by_registration_id = live_results.reload.index_by(&:registration_id)
+
+      attempts_to_load = results_wcif.flat_map do |result_wcif|
+        registration_id = registrations_by_wcif_id[result_wcif["personId"]].id
+        live_result_id = results_by_registration_id[registration_id].id
+
+        result_wcif["attempts"].map.with_index(1) do |attempt, attempt_number|
+          LiveAttempt.build_with_history_entry(attempt["result"], attempt_number, 1)
+                     .attributes
```
gregorbg: I am 99% certain that calling `attributes` here entirely discards the `_with_history_entry` part. You're building a proper AR model class with a history entry correctly attached as a child, but calling `attributes` turns the parent `LiveAttempt` record into a pure hash which (obviously) does not understand anything about the concept of "associations".

app/models/round.rb
```diff
@@ -368,6 +368,50 @@ def self.parse_wcif_id(wcif_id)
...
+                     .merge(
+                       live_result_id: live_result_id,
+                       created_at: now,
+                       updated_at: now,
```
gregorbg: See my other comment about time not needing to be passed in

app/models/round.rb
```diff
@@ -368,6 +368,50 @@ def self.parse_wcif_id(wcif_id)
...
+          best: result_wcif["best"],
+          average: result_wcif["average"],
+          global_pos: result_wcif["ranking"],
+          local_pos: result_wcif["ranking"],
+          created_at: now,
+          updated_at: now,
+        }
+      end
+
+      LiveResult.insert_all!(results_to_load)
```
gregorbg: Should this really be an insert? Feels more like an `upsert`...

app/models/round.rb
```diff
@@ -368,6 +368,50 @@ def self.parse_wcif_id(wcif_id)
...
+                     .attributes
+                     .merge(
+                       live_result_id: live_result_id,
+                       created_at: now,
+                       updated_at: now,
+                     )
+        end
+      end
+
+      LiveAttempt.insert_all!(attempts_to_load) if attempts_to_load.any?
```
gregorbg: See my other remark about `upsert`ing

app/models/round.rb
```diff
@@ -377,7 +421,6 @@ def self.wcif_to_round_attributes(event, wcif, round_number, total_rounds)
       cutoff: Cutoff.load(wcif["cutoff"]),
       advancement_condition: AdvancementConditions::AdvancementCondition.load(wcif["advancementCondition"]),
       scramble_set_count: wcif["scrambleSetCount"],
-      round_results: RoundResults.load(wcif["results"]),
```
gregorbg: Do we want to keep writing this for a week or so as a "transition period" so that we have a safety mechanism to fall back on?

spec/requests/api_competitions_spec.rb
```diff
@@ -509,8 +510,8 @@
           rounds = competition.reload.competition_events.find_by(event_id: "333").rounds
           expect(rounds.length).to eq 1
           expect(rounds.first.scramble_set_count).to eq 2
-          expect(rounds.first.round_results.length).to eq 1
-          expect(rounds.first.round_results.first.attempts.map(&:result)).to eq [456, 745, 657, 465, 835]
+          expect(rounds.first.live_results.length).to eq 1
+          expect(rounds.first.live_results.first.attempts.map(&:value)).to eq [456, 745, 657, 465, 835]
```
gregorbg: Can you do something with `pluck` here? It had to be `map` before, because it was a pure JSON-loaded in-memory model. But now we have an AR database model, so why not make use of it? :D

gregorbg (changes requested):

app/models/round.rb
FinnIckler: the error is eaten by the transaction block though?

app/models/round.rb
FinnIckler: we just destroyed all live results, so this should be an insert

app/models/round.rb
gregorbg: If it is, then that's probably not good... At least not for this specific scenario.

app/models/round.rb
gregorbg: Okay then (a) Is there a reason why we destroy + insert instead of upserting? (b) If the answer to A is Yes, then move the `destroy_all` directly towards the `insert_all` along with a comment explaining your thinking around question A

app/models/round.rb
```diff
@@ -401,6 +401,136 @@ def self.parse_wcif_id(wcif_id)
...
+      end
+
+      LiveResultHistoryEntry.insert_all!(histories_to_generate) if histories_to_generate.any?
+
+      results_which_changed = histories_to_generate.pluck(:live_result_id)
+      LiveResult.where(id: results_which_changed).update_all(last_attempt_entered_at: database_now)
+    end
+
+    # Sync up all of the attempts and histories `upsert_all` shenanigans
+    self.live_results.reset
```
FinnIckler, line 531: do we also need to do `self.linked_round&.live_results.reset` ?

spec/models/competition_wcif_spec.rb
```diff
@@ -672,28 +672,117 @@
...
-        {
-          "personId" => 1,
-          "ranking" => 10,
-          "attempts" => [{ "result" => 456, "reconstruction" => nil }] * 5,
-          "best" => 456,
-          "average" => 456,
-        },
-        {
-          "personId" => 2,
+    context "round results" do
```
FinnIckler, line 675: I think this is fine. Maybe @jonatanklosko can weigh in if this covers all the wca live syncs results to rails cases?

jonatanklosko: Looks good!

PR #13262, opened 2026-01-23
Make live round_results route work by wcif_id

FinnIckler:
Have not tested with this https://github.com/thewca/worldcubeassociation.org/pull/13055 but will do on monday

### DIFF
--- a/app/models/round.rb
+++ b/app/models/round.rb
@@ -380,7 +380,7 @@ class Round < ApplicationRecord
     {
       **self.to_wcif,
       "round_id" => id,
-      "competitors" => accepted_registrations.includes(:user).map { it.as_json({ methods: %i[user_name registrant_id] })},
+      "competitors" => accepted_registrations.includes(:user).map { it.as_json({ methods: %i[user_name], only: %i[id user_id registrant_id] }) },
       "results" => only_podiums ? live_podium : live_results,
     }
   end


app/controllers/api/v1/live/live_controller.rb
```diff
@@ -3,9 +3,15 @@
 class Api::V1::Live::LiveController < Api::V1::ApiController
   skip_before_action :require_user, only: %i[round_results by_person podiums]
   def round_results
-    round_id = params.require(:round_id)
+    competition = Competition.find(params.require(:competition_id))
+
+    event_id, number = params.require(:round_id).split("-")
```
gregorbg: What about changing the URL schema to `/competition/event/number`?

Not saying you have to do it, but just an open suggestion for discussion :) My thinking is that if you use slashes, the Rails engine does the "split by slash" parsing for you for free

FinnIckler: I think that is nice, but I do also think that keeping compatibility with wcif ids is more important?

app/controllers/api/v1/live/live_controller.rb
```diff
@@ -3,9 +3,17 @@
...
+    activity_code =  ScheduleActivity.parse_activity_code(params.require(:round_id))
+
+    event_id, number = activity_code.values_at(:event_id, :round_number)
+
+    return render json: { status: "round not found" }, status: :not_found if event_id.nil? || number.nil?
+
+    competition_event = competition.competition_events.find_by(event_id: event_id)
 
-    round = Round.includes(live_results: %i[live_attempts round event]).find(round_id)
+    round = competition.includes(rounds: [:'live_results:', :'%i[live_attempts', :round, :'event]']).rounds.find_by(competition_event_id: competition_event.id, number: number)
```
gregorbg: Something with quoting the strings here went horribly wrong

gregorbg: Rails actually has complex finders. You can do:
```ruby
my_round = Round.includes(:competition_event, :other_stuff_you_want_to_include).find_by(competition_event: { competition_id: 'FooBarOpen2015', event_id: '333' }, number: 1)
```

app/models/round.rb
```diff
@@ -282,7 +282,9 @@ def to_live_json(only_podiums: false)
     {
       **self.to_wcif,
       "round_id" => id,
-      "competitors" => accepted_registrations.map { it.as_json({ include: [user: { only: [:name], methods: [], include: [] }] }).merge("registration_id" => it.registrant_id) },
+      "competitors" => accepted_registrations.includes([:user]).map do |registration|
```
```suggestion
      "competitors" => accepted_registrations.includes(:user).map do |registration|
```

app/models/round.rb
```diff
@@ -282,7 +282,9 @@ def to_live_json(only_podiums: false)
...
+        registration.as_json({ include: [user: { only: [:name], methods: [], include: [], skip_email: true, exclude_deprecated: true }], methods: [] }).merge("registration_id" => registration.registrant_id)
```
gregorbg: Now that we have `registrant_id` as a stable column in the model, we can probably just include it in the `DEFAULT_SERIALIZE_OPTIONS` for the registration class. I'd be willing to let that slip through in this PR, so you can get rid of the `merge` 😉

app/models/user.rb
gregorbg, line 1: Veto. I know you want to skip the expensive "is this guy a Delegate" checks, but this needs to be optimized separately.

In the serialization above, it looks like you _only_ need the user name, and absolutely nothing else. Since a registration already knows its user via a `belongs_to`, you can just add `delegate :name, to: :user, prefix: true` or similar to `registration.rb` and add `methods: [...foo..., :user_name]` to the serialization process of the registration.

Doing so will only access the one property/column of the user entity, and it will not trigger the whole "serialization routine".

app/models/round.rb
FinnIckler: i can just add it to the methods block

gregorbg: Uh, does that work for columns?

gregorbg: I always thought attributes must go under `only`, as `methods` is designed for stuff that is not an attribute/column

FinnIckler: every column is a method, but I can change that for clarity sake

PR #13267, opened 2026-01-26
Make results uniqueness index actually unique

gregorbg:
Quote from my code comment:

>  At the time of writing this migration, we already have an index on (round_id, person_id) but for historic reasons, it is not `UNIQUE`. We want to change that, but MySQL does not have syntax for "taking this existing index and making it unique". So we drop the old index and immediately re-create it with the `UNIQUE` option set.

Before we deploy this, we need to make absolutely sure there are no collisions on prod. There should be none, because we already have a model-level validation in Rails, but just to be sure:
```sql
SELECT round_id, person_id, COUNT(id) uniq
FROM results
GROUP BY round_id, person_id
HAVING uniq > 1;
```


gregorbg:
Zero results on prod as of this comment, deploying now...

PR #13268, opened 2026-01-26
Nextjs WCA Live: Open round creates empty results

FinnIckler:
We decided to just use the same approach as WCA Live right now. if https://github.com/thewca/worldcubeassociation.org/pull/13055 is merged I can make a quick prototype round admin page that calls the init_round method

### DIFF
--- a/app/models/live_result.rb
+++ b/app/models/live_result.rb
@@ -14,6 +14,8 @@ class LiveResult < ApplicationRecord
 
   belongs_to :round
 
+  scope :not_empty, -> { where.not(best: 0) }
+
   alias_attribute :result_id, :id
 
   has_one :event, through: :round
--- a/app/models/round.rb
+++ b/app/models/round.rb
@@ -292,7 +292,7 @@ class Round < ApplicationRecord
   end
 
   def competitors_live_results_entered
-    live_results.where.not(best: 0).count
+    live_results.not_empty.count
   end
 
   def score_taking_done?


lib/tasks/import_live_results_from_round_results.rake
```diff
@@ -23,22 +23,22 @@ namespace :live_results do
         format = round.format
         results = round_result.attempts
 
-        attempts = results.map.with_index(1) do |r, i|
-          LiveAttempt.build_with_history_entry(r.result, i, 1)
+        attempts = results.map.with_index(1) do |rr, i|
+          LiveAttempt.build_with_history_entry(rr.result, i, 1)
```
gregorbg, line 27: Was this rename intentional? If so, what does `rr` stand for?

lib/tasks/import_live_results_from_round_results.rake
```diff
@@ -23,22 +23,22 @@ namespace :live_results do
...
-          value4: results[3]&.result || 0,
-          value5: results[4]&.result || 0,
           event_id: event.id,
           round_type_id: round.round_type_id,
           round_id: round.id,
           format_id: format.id,
         )
 
+        r.result_attempts = results.map.with_index(1) do |rr, i|
+          ResultAttempt.new(value: rr.result, attempt_number: i)
```
gregorbg, line 38: (see above regarding `rr` as variable name)

app/models/round.rb
```diff
@@ -170,17 +170,29 @@ def consider_previous_round_results?
...
 
+  def init_round
+    empty_results = advancing_registrations.map do |r|
+      { registration_id: r.id, round_id: id }
+    end
+    LiveResult.insert_all!(empty_results)
+  end
+
+  def accepted_registrations
+    registration_ids = live_results.pluck(:registration_id)
```
gregorbg: Why is the set of accepted registrations defined as "the registrations for which there are already live results"?

In my understanding, it should be the other way around: There are accepted registrations, and these are then used to create live results. Essentially, I would expect `registrations.accepted` to appear here.

gregorbg: In particular, the method `accepted_registrations` is used a lot outside of "WCA Live world". It is crucial not to intermix anything here, which is why I like the `advancing_` rename above.

But then this new method confuses me.

gregorbg: Sidenote: If this is really, _really_ necessary, then I would recommend:
```ruby
has_many :live_registrations, through: :live_results, source: :registration

# ...other code

def something_like_accepted_registrations
  live_registration_ids # Rails gives you the _ids helper automagically via `has_many`
end
```

lib/tasks/import_live_results_from_round_results.rake
FinnIckler: round result

app/models/round.rb
FinnIckler: accepted_registrations from round.rb is not used anywhere except for this

FinnIckler: this is not a new method, it was defined as what previously was advancing_registrations

FinnIckler: let me just rename it to competitors

app/models/live_result.rb
```diff
@@ -113,7 +113,7 @@ def recompute_advancing
       round_results = advancement_determining_results.where.not(global_pos: nil)
       round_results.update_all(advancing: false, advancing_questionable: false)
 
-      missing_attempts = round.total_accepted_registrations - round_results.count
+      missing_attempts = round.total_competitors - round_results.count
```
FinnIckler: this is now always 0 so missing attempts should be calculated with `round.total_competitors - round_results` which are not empty

FinnIckler: ok, maybe this is already accounted for as `round_results = advancement_determining_results.where.not(global_pos: nil)`

app/models/round.rb
gregorbg: Okay, I like that name better. But please also consider my other remark about `has_many through`

config/routes.rb
```diff
@@ -374,6 +374,7 @@
         if WcaLive.enabled?
           namespace :live do
             get '/rounds/:round_id' => 'live#round_results', as: :live_round_results
+            get '/rounds/:round_id/open' => "live#open_round", as: :live_round_open
```
gregorbg: Feel like this should be a `POST` or `PUT`?

FinnIckler: true, it just made it easier to trigger in the browser

app/models/round.rb
```diff
@@ -170,27 +171,38 @@ def consider_previous_round_results?
...
+  def init_round
+    empty_results = advancing_registrations.map do |r|
+      { registration_id: r.id, round_id: id, average: 0, best: 0, last_attempt_entered_at: Time.now.utc }
+    end
+    LiveResult.insert_all!(empty_results)
+  end
+
+  def competitors
+    live_registrations
+  end
```
gregorbg: Will this be correct when follow-up rounds are opened?

FinnIckler: yes? these results are scoped to a round so they are no duplicates. They do not take account quitters, but that's for another PR

app/models/round.rb
```diff
@@ -170,27 +171,38 @@ def consider_previous_round_results?
...
     end
   end
 
-  def total_accepted_registrations
-    accepted_registrations.count
+  def init_round
+    empty_results = advancing_registrations.map do |r|
+      { registration_id: r.id, round_id: id, average: 0, best: 0, last_attempt_entered_at: Time.now.utc }
+    end
+    LiveResult.insert_all!(empty_results)
```
gregorbg, line 189: Does the `insert_all` method have a return value?

app/controllers/api/v1/live/live_controller.rb
```diff
@@ -29,4 +29,26 @@ def podiums
 
     render json: final_rounds.map { |r| r.to_live_json(only_podiums: true) }
   end
+
+  def open_round
+    competition = Competition.find(params.require(:competition_id))
+
+    event_id, number = params.require(:round_id).split("-")
```
gregorbg: Can you use the Activity Code parser from the `ScheduleActivity` class?

app/controllers/api/v1/live/live_controller.rb
```diff
@@ -29,4 +29,26 @@ def podiums
...
+
+    round = competition.rounds.find_by!(competition_event_id: competition_event.id, number: number)
+
+    # TODO: Move these to actual error codes at one point
+    # Also think about if we should auto open all round ones at competition day start and not have this check
+    return render json: { status: "previous round has empty results" }, status: :bad_request unless round.number == 1 || round.previous_round.score_taking_done?
+
+    return render json: { status: "round already open" }, status: :bad_request if round.live_results.any?
+
+    round.init_round
```
gregorbg: Ideally, this should have some kind of return value. Even something as trivial as a Boolean or an int returning the number of rows inserted.

(See also my other comment about the return value of `insert_all`. If that turns out to have any return value, then you can just directly use that here without any further ado)

app/models/round.rb
```diff
@@ -170,27 +171,38 @@ def consider_previous_round_results?
...
       advancing = previous_round.live_results.where(advancing: true).pluck(:registration_id)
       Registration.find(advancing)
     end
   end
 
-  def total_accepted_registrations
-    accepted_registrations.count
+  def init_round
+    empty_results = advancing_registrations.map do |r|
+      { registration_id: r.id, round_id: id, average: 0, best: 0, last_attempt_entered_at: Time.now.utc }
```
gregorbg: For getting the current timestamp inside a Rails model, there's the wonderfully eloquent method `self.current_time_from_proper_timezone` (inherited for free by the AR base class)

app/models/round.rb
gregorbg: Then why not just call the `has_many` association `competitors` or perhaps `live_competitors`?

Otherwise, you could just use `live_registrations` everywhere in all places you're using `competitors` right now, although I don't prefer that because "registrations" are conceptually something that happens "before the competition" in most people's minds. Rounds during the competition (with the exception of the very first maybe) don't have "registrations" in most people's minds

app/controllers/api/v1/live/live_controller.rb
```diff
@@ -29,4 +29,29 @@ def podiums
...
     render json: final_rounds.map { |r| r.to_live_json(only_podiums: true) }
   end
+
+  def open_round
+    competition = Competition.find(params.require(:competition_id))
+
+    activity_code =  ScheduleActivity.parse_activity_code(params.require(:round_id))
+
+    event_id = activity_code[:event_id]
+    number = activity_code[:round_number]
```
gregorbg:
```suggestion
    activity_code =  ScheduleActivity.parse_activity_code(params.require(:round_id))

    event_id, number = activity_code.values_at(:event_id, :round_number)
```

gregorbg: Not quite as powerful as JS, but yay destructuring :D

app/models/round.rb
FinnIckler: it returns an https://apidock.com/rails/ActiveRecord/Result so you could do something like what I just pushed

gregorbg: Cool!

app/controllers/api/v1/live/live_controller.rb
```diff
@@ -29,4 +29,28 @@ def podiums
...
+
+    # TODO: Move these to actual error codes at one point
+    # Also think about if we should auto open all round ones at competition day start and not have this check
+    return render json: { status: "previous round has empty results" }, status: :bad_request unless round.number == 1 || round.previous_round.score_taking_done?
+
+    return render json: { status: "round already open" }, status: :bad_request if round.live_results.any?
+
+    result = round.init_round
+
+    render json: { success: result.affected_rows > 0 }
```
gregorbg: From the top of my head, I'm not sure whether this "`> 0` means success" assumption is always fully valid. Right now, I cannot think of a case where it's _not_ valid, but competitions tend to have edge cases. And we have a **lot** of competitions these days :D

May I suggest something like:
- Returning `status: "ok"` always, based on "the code reached this point without crashing"
- Also adding more details to the JSON, for example, `added_rows: result.affected_rows`?

gregorbg: (Future music: The frontend could report "Yay, #{n} competitors progressed to the next round" as a success message or smth)

app/controllers/api/v1/live/live_controller.rb
```diff
@@ -29,4 +29,28 @@ def podiums
...
+
+    activity_code =  ScheduleActivity.parse_activity_code(params.require(:round_id))
+
+    event_id, number = activity_code.values_at(:event_id, :round_number)
+
+    return render json: { status: "round not found" }, status: :not_found if event_id.nil? || number.nil?
+
+    competition_event = competition.competition_events.find_by(event_id: event_id)
+
+    round = competition.rounds.find_by!(competition_event_id: competition_event.id, number: number)
```
gregorbg: We should merge #13262 first, so you can "recycle" the round loading magic, ideally by moving it into `Round.find_by_wcif_id` or smth

app/models/round.rb
```diff
@@ -279,16 +291,12 @@ def recompute_local_pos
     SQL
   end
 
-  def total_accepted_registrations
-    accepted_registrations.count
-  end
-
   def competitors_live_results_entered
-    live_results.count
+    live_results.where.not(best: 0).count
```
gregorbg: Can this be a scope on `LiveResult`?

FinnIckler: I spend way to long coming up with a name, something like "not_incomplete" sounds bad, but "completed" isn't quite accurate. So I just did "not_empty"

gregorbg: We usually use the terminology "skipped" and "not_skipped" for zero results, but that still comes from the old valueN era.
These days, you cannot "skip" value4 and value5 anymore, so I am totally fine with the name that you came up with.

gregorbg: We usually use the terminology "skipped" and "not_skipped" for zero results, but that still comes from the old valueN era.
These days, you cannot "skip" value4 and value5 anymore, so I am totally fine with the name that you came up with.

gregorbg: We usually use the terminology "skipped" and "not_skipped" for zero results, but that still comes from the old valueN era.
These days, you cannot "skip" value4 and value5 anymore, so I am totally fine with the name that you came up with.

PR #13297, opened 2026-01-28
Add locked_by and quit_by columns to live_results

FinnIckler:
I think this is a good comprise between a full table for locks and quits?

### DIFF
--- a/app/models/live_result.rb
+++ b/app/models/live_result.rb
@@ -50,8 +50,8 @@ class LiveResult < ApplicationRecord
     end
   end
 
-  def mark_as_quit(quit_by_user)
-    update(quit_by_id: quit_by_user.id, advancing: false, advancing_questionable: false)
+  def mark_as_quit!(quit_by_user)
+    update!(quit_by_id: quit_by_user.id, advancing: false, advancing_questionable: false)
   end
 
   def locked?
--- a/app/models/round.rb
+++ b/app/models/round.rb
@@ -377,16 +377,18 @@ class Round < ApplicationRecord
   end
 
   def quit_from_round!(registration_id, quitting_user)
-    result = live_results.find_by!(registration_id: registration_id)
+    ActiveRecord::Base.transaction do
+      result = live_results.find_by!(registration_id: registration_id)
 
-    is_quit = result.destroy
+      is_quit = result.destroy
 
-    return is_quit ? 1 : 0 if first_round?
+      return is_quit ? 1 : 0 if first_round?
 
-    # We need to also quit the result from the previous round so advancement can be correctly shown
-    previous_round_results = previous_round.linked_round.present? ? previous_round.linked_round.live_results : previous_round.live_results
+      # We need to also quit the result from the previous round so advancement can be correctly shown
+      previous_round_results = previous_round.linked_round.present? ? previous_round.linked_round.live_results : previous_round.live_results
 
-    previous_round_results.where(registration_id: registration_id).count { |r| r.mark_as_quit(quitting_user) }
+      previous_round_results.where(registration_id: registration_id).count { |r| r.mark_as_quit!(quitting_user) }
+    end
   end
 
   def wcif_id


app/controllers/api/v1/live/live_controller.rb
```diff
@@ -32,19 +32,35 @@ def podiums
...
+    return render json: { status: "unauthorized" }, status: :unauthorized unless @current_user.can_manage_competition?(competition)
     # Also think about if we should auto open all round ones at competition day start and not have this check
     return render json: { status: "previous round has empty results" }, status: :bad_request unless round.number == 1 || round.previous_round.score_taking_done?
 
     return render json: { status: "round already open" }, status: :bad_request if round.live_results.any?
 
     result = round.init_round
 
-    render json: { status: "ok", added_rows: result.affected_rows }
+    locked = round.previous_round.lock_results(@current_user) if round.number != 1
```
gregorbg: This approach is dangerous. Per Dual Rounds, any one given round does not have one deterministic "previous round" anymore. I think it is better to hide all the complexity of the (current) `init_round` and `lock_results` behind one "do the things that need to happen when the user clicks the button in the UI" method and call that one here.

gregorbg: In terms of return value (right now, you have two return values but the new method can obviously only have one), you could return a nilable integer. Nil means something with locking the previous round went wrong, whereas a successful lock just returns the number of rows affected.

app/models/round.rb
```diff
@@ -353,6 +353,10 @@ def self.wcif_to_round_attributes(event, wcif, round_number, total_rounds)
     }
   end
 
+  def lock_results(current_user)
+    live_results.update_columns(locked_by_id: current_user.id)
```
gregorbg: Why are you using `update_columns` instead of something like `update`? The former circumvents hooks and validations, IIRC. Is this intended?

app/controllers/api/v1/live/live_controller.rb
FinnIckler: ok, done. For linked rounds, do we open them both when one is opened?

app/models/live_result.rb
```diff
@@ -14,8 +14,14 @@ class LiveResult < ApplicationRecord
 
   belongs_to :round
 
+  belongs_to :quit_by, class_name: 'User', optional: true
+
+  belongs_to :locked_by, class_name: 'User', optional: true
+
   scope :not_empty, -> { where.not(best: 0) }
 
+  scope :without_quitters, -> { where(quit_by_id: nil).or(where.not(best: 0)) }
```
gregorbg: In which case do you need the second `or(where.not(best: 0))` part?

gregorbg: Sidenote: `.or(not_empty)` should work

app/models/round.rb
```diff
@@ -44,7 +44,10 @@ def format
   has_many :wcif_extensions, as: :extendable, dependent: :delete_all
 
   has_many :live_results, -> { order(:global_pos) }
-  has_many :live_competitors, through: :live_results, source: :registration
+  has_many :live_results_without_quitters,
+           -> { without_quitters.order(:global_pos) },
+           class_name: "LiveResult"
```
gregorbg: Hm, it feels like this shouldn't be a separate relation. Why does `some_round.live_results.without_quitters` (ie, applying the `without_quitters` scope to the `live_results` relation) not work?

app/models/round.rb
```diff
@@ -182,6 +185,13 @@ def advancing_registrations
     end
   end
 
+  def open_and_lock_previous(current_user)
+    init_round
+    return 0 if number == 1 || linked_round.present?
```
gregorbg: Why are you ignoring the return value of `init_round`? And what does the return value `0` represent?

gregorbg: Why are you returning early in case of a linked round? The linked round's individual rounds would _all_ need to be locked, and returning early feels a bit premature...

app/models/round.rb
```diff
@@ -193,18 +203,18 @@ def total_competitors
     live_competitors.count
   end
 
-  def recompute_live_columns
+  def recompute_live_columns(locked)
```
gregorbg: I wouldn't be so overly specific about calling the parameter "locked". There may come more use-cases later in the future that need to skip the advancement criteria.

Instead, I suggest `def recompute_live_columns(skip_advancing: false)`, and calling it like `recompute(skip_advancing: has_been_locked)` elsewhere.

app/models/round.rb
```diff
@@ -353,6 +363,25 @@ def self.wcif_to_round_attributes(event, wcif, round_number, total_rounds)
     }
   end
 
+  def lock_results(current_user)
```
gregorbg: This should probably be `locking_user` or something similar

app/models/round.rb
```diff
@@ -353,6 +363,25 @@ def self.wcif_to_round_attributes(event, wcif, round_number, total_rounds)
...
+    results_to_lock = linked_round.present? ? linked_round.live_results : live_results
+
+    results_to_lock.update_all(locked_by_id: current_user.id)
+  end
+
+  def quit_from_round!(registration_id, current_user)
```
gregorbg: Again, something like `quitting_user` or similar is better than default Devise nomenclature

app/models/round.rb
```diff
@@ -353,6 +363,25 @@ def self.wcif_to_round_attributes(event, wcif, round_number, total_rounds)
...
+    result = live_results.find_by!(registration_id: registration_id)
+
+    result.mark_as_quit(current_user)
+
+    return 1 if number == 1 || linked_round.present?
```
gregorbg: What does this return value represent?

app/models/round.rb
```diff
@@ -353,6 +363,25 @@ def self.wcif_to_round_attributes(event, wcif, round_number, total_rounds)
...
+  def lock_results(current_user)
+    results_to_lock = linked_round.present? ? linked_round.live_results : live_results
+
+    results_to_lock.update_all(locked_by_id: current_user.id)
+  end
+
+  def quit_from_round!(registration_id, current_user)
+    result = live_results.find_by!(registration_id: registration_id)
+
+    result.mark_as_quit(current_user)
```
gregorbg: Here, you are ignoring the return value of `mark_as_quit`. Further down below, you actually care about it and run a `count { it == true }` on it. Either you ignore it both times, or you care about it both times.

app/models/live_result.rb
FinnIckler: We do still want to show quitters for previous rounds.
Currently when someone is quit from round 2 who advanced from round 1 I mark them as quit in round 2 and 1, but I still want to show them in round 1. There is a case to be made to only mark them as quit in round 1 as those results will already be locked so only flipping the advancing will be necessary?

app/models/round.rb
FinnIckler: because I use it in a scope one line down in the 
`has_many :live_competitors, through: :live_results_without_quitters, source: :registration`

app/models/round.rb
FinnIckler: The return value is how many results where locked and it's 0 if there is no previous round to lock

FinnIckler: This is about locking the *previous* round as the title suggests. Linked Rounds do not have previous rounds

app/models/round.rb
FinnIckler: how many results have been marked as quit

app/models/round.rb
FinnIckler: I do not ignore it, I return 1 the line after

FinnIckler: This is because update doesn't return the number of rows updated, but just true or false. I guess I can do something like I just commited

app/controllers/api/v1/live/live_controller.rb
```diff
@@ -32,19 +32,36 @@ def podiums
...
+    registration_id = params.require(:registration_id)
+
+    return render json: { status: "unauthorized" }, status: :unauthorized unless @current_user.can_manage_competition?(competition)
+
+    round = Round.find_by_wcif_id!(wcif_id, competition.id)
+    result = round.live_results.find_by!(registration_id: registration_id)
+
+    return render json: { status: "Cannot quit competitor with results" }, status: :bad_request if result.live_attempts.any?
+
+    quit_count = round.quit_from_round!(registration_id, @current_user)
```
gregorbg, line 63: When do we ever expect this method to return a value greater than 1?
In other words, would a boolean "quit or not quit" make more sense?

app/models/live_result.rb
```diff
@@ -14,6 +14,10 @@ class LiveResult < ApplicationRecord
 
   belongs_to :round
 
+  belongs_to :quit_by, class_name: 'User', optional: true
+
+  belongs_to :locked_by, class_name: 'User', optional: true
```
gregorbg, line 18: This is one of the very rare cases where I will actually suggest deleting a blank line :P

app/models/live_result.rb
```diff
@@ -47,6 +51,10 @@ def best_possible_solve_times
     end
   end
 
+  def mark_as_quit(current_user)
```
gregorbg: More like `quitting_by` or `quit_by_user` or something

app/models/live_result.rb
```diff
@@ -84,6 +92,6 @@ def notify_users
     end
 
     def trigger_recompute_columns
-      round.recompute_live_columns
+      round.recompute_live_columns(skip_advancing: locked_by.present?)
```
gregorbg: For economic reasons, you can just check `locked_by_id.present?` to avoid actually loading the whole `User` into memory (just for checking whether the user exists or not).

gregorbg: On that note, Rails has auto-magic alias `my_field?` for any database column, which is equivalent to `my_field.present?`

So you could just do `alias_method :locked?, :locked_by_id?` and then use `skip_advancing: self.locked?` down here

app/models/round.rb
gregorbg: I do not think this is a fair assumption. Maybe under the current 2026 regs, but not under code design.

app/models/round.rb
```diff
@@ -182,6 +182,13 @@ def advancing_registrations
     end
   end
 
+  def open_and_lock_previous(current_user)
+    init_round
```
gregorbg: The terminology between "open" and "init" is inconsistent

app/models/round.rb
gregorbg: At the risk of missing something: Why are you ignoring the return value of `init_round`?

Your explanation above is about the return value of this current `open_and_lock_previous` method. But inside this method, you're calling `init_round` and you completely return the return value. That's what I am specifically curious about.

app/models/round.rb
```diff
@@ -353,6 +360,25 @@ def self.wcif_to_round_attributes(event, wcif, round_number, total_rounds)
...
+    result = live_results.find_by!(registration_id: registration_id)
+
+    is_quit = result.destroy
+
+    return is_quit ? 1 : 0 if number == 1 || linked_round.present?
+
+    # We need to also quit the result from the previous round so advancement can be correctly shown
+    previous_round_results = previous_round.linked_round.present? ? previous_round.linked_round.live_results : previous_round.live_results
+
+    previous_round_results.where(registration_id: registration_id).map { it.mark_as_quit(quitting_user) }.count { it == true }
```
gregorbg: The unconventional `map { update }.count { it == true }` notation just made me realize that this is firing one UPDATE call per row. Can we merge this into some kind of `update_all` thingy, which should on Rails DB adapter level also return the number of rows affected...

gregorbg: I know this makes code reuse much harder, but in this case I would argue for efficiency and load optimization

gregorbg: Also, on an unrelated side note, how should we react if any of those `update` calls (that you're currently executing separately) returns `false`?

Right now, they would be silently ignored and "simply not be part of the count", which feels _very_ dangerous.

app/controllers/api/v1/live/live_controller.rb
FinnIckler: I was already thinking about bulk quits, but also you need to be quit from two rounds when the previous round was a dual round

app/models/round.rb
FinnIckler: update_all gives us even less control over it though? That doesn't run any hooks or validations

app/models/round.rb
FinnIckler: because init_round returns the number of results initialized, but we want to return the number of results locked. You had a previous comment that I can't return both and I should return that instead

app/controllers/api/v1/live/live_controller.rb
gregorbg: Fair enough.

app/models/round.rb
gregorbg: Fair point! I am now much less in favor of `update_all`. But the `map { ...do some update... }.count { it == true }` still feels awkward. What does the LLM say about this?

app/models/round.rb
```diff
@@ -353,6 +362,25 @@ def self.wcif_to_round_attributes(event, wcif, round_number, total_rounds)
...
+
+    results_to_lock.update_all(locked_by_id: locking_user.id)
+  end
+
+  def quit_from_round!(registration_id, quitting_user)
+    result = live_results.find_by!(registration_id: registration_id)
+
+    is_quit = result.destroy
+
+    return is_quit ? 1 : 0 if number == 1 || (linked_round.present? && linked_round.first_round_in_series)
```
gregorbg: This is duplicated with some logic in `open_and_lock_previous` now, please extract.

app/models/round.rb
```diff
@@ -182,7 +182,16 @@ def advancing_registrations
     end
   end
 
-  def init_round
+  def open_and_lock_previous(locking_user)
+    open_round!
+    return 0 if number == 1 || (linked_round.present? && linked_round.first_round_in_link.number == 1)
+
+    round_to_lock = linked_round.present? ? linked_round.first_round_in_link.previous_round : previous_round
```
gregorbg, line 189: Imagine we're in a constellation of `R1 (R2 R3) R4` where the brackets mean "linked". From an organization team perspective, we may choose to open both rounds separately, because they run independently schedule-wise (for example, there's lunch break between R2 and R3).

In that case, opening R2 first will correctly point to R1 and lock it. After/During lunch, trying to open R3 will _also_ point to R1 and try to lock it again, which seems odd...

I think in terms of "visual" UX, we should give organizers/scoretakers the option to open linked rounds independently. But in the backend, we need to pay attention that we don't accidentally "double-lock" previous rounds.

app/models/round.rb
gregorbg: Also, where is `first_round_in_series` even being defined?

app/models/round.rb
```diff
@@ -182,7 +182,16 @@ def advancing_registrations
     end
   end
 
-  def init_round
+  def open_and_lock_previous(locking_user)
+    open_round!
```
gregorbg: You pushed a rename commit on this line, so GitHub might have wrongfully "neglected" my previous comment. So at the risk of repeating myself: Why are you ignoring the return value of `open_round!`?

app/models/round.rb
FinnIckler: in linked_round.rb

app/models/round.rb
FinnIckler: because open_round! returns the number of results initialized, but we want to return the number of results locked. You had a previous comment that I can't return both and I should return that instead

app/models/round.rb
FinnIckler: but I did typo here

app/models/round.rb
gregorbg: So which one is more interesting to the user? As an organizer, I would be more interested to see that "32 competitors made it to the next round" rather than "the 70 results which you just entered are now locked".

I don't remember the exact context of my previous comment, but I'm sure it was something along the lines of "you cannot return them both in a mutually exclusive way". What you can actually do is...

```
open_count = open_round!
locked_count = lock_competitors

[open_count, locked_count]

# ..and then in the controller...

opened, locked = round.open_and_lock_previous
```

app/models/round.rb
FinnIckler: there is 
`previous_round_results.where(registration_id: registration_id).count { |r| r.mark_as_quit(quitting_user) }`
or we just do 
`previous_round_results.where(registration_id: registration_id).update(quit_by_id: quit_by_user.id, advancing: false, advancing_questionable: false)`
which runs validations and runs one update per row, but it does still only return one boolean

app/models/round.rb
```diff
@@ -353,6 +362,33 @@ def self.wcif_to_round_attributes(event, wcif, round_number, total_rounds)
...
+    result = live_results.find_by!(registration_id: registration_id)
+
+    is_quit = result.destroy
+
+    return is_quit ? 1 : 0 if first_round?
+
+    # We need to also quit the result from the previous round so advancement can be correctly shown
+    previous_round_results = previous_round.linked_round.present? ? previous_round.linked_round.live_results : previous_round.live_results
+
+    previous_round_results.where(registration_id: registration_id).count { |r| r.mark_as_quit(quitting_user) }
```
gregorbg: Okay, this is more succinct now and feels less "wrong". One question remains: How do we deal with the `false` values where the update failed?

FinnIckler: What would an organizer like to know in that case? I assume the registration_id where quitting failed?

gregorbg: Hm... I don't have a quick, intuitive answer to this yet. From a bird's eye perspective, I think that if one of the "quits" failed, the whole UX process of "quitting this competitor [as triggered by button click]" should fail.

And then you would ideally report the error on the row that failed. But that could be cumbersome to code.

Anyways, I think that some level of transaction here (if one fails, roll all of them back) would be prudent.

FinnIckler: something like this for now and then we think about it when I do the UI?

app/models/round.rb
```diff
@@ -353,6 +362,35 @@ def self.wcif_to_round_attributes(event, wcif, round_number, total_rounds)
...
+
+    results_to_lock.update_all(locked_by_id: locking_user.id)
+  end
+
+  def first_round?
+    number == 1 || (linked_round.present? && linked_round.first_round_in_link.number == 1)
+  end
+
+  def quit_from_round!(registration_id, quitting_user)
+    ActiveRecord::Base.transaction do
```
gregorbg, line 380: Yay! Fun fact: You can do `self.transaction`. Every _instance_ of an AR model is also eligible to create/open transactions.

There's not much difference in terms of execution (the `some_random_model.transaction do` internally simply points to `self.class.transaction do yield`) but it does remove a little bit of `ActiveRecord::Base` arcane-ness from the code.

PR #13303, opened 2026-01-30
Diff Protocol for WCA Live Result Web Socket updates

FinnIckler:
I am still writing tests for it, but I think I got the scaffolding done. I am basically sending this:

```
{
updated: [
      {
        registration_id: 123,
        global_pos: 3,
        local_pos: 3,
        advancing: true,
        live_attempts: {
          updated: [{ id: 456, attempt_number: 3, value: 1523 }]
        }
      }
    ],
created: [
  { registration_id: 999, ... }
],
deleted: [122] 
}
```

With updated only sending changed attributes, created sending full ones and deleted just sending ids. 
Deleted probably needs to be changed to work on quitters instead though.

### DIFF
--- a/spec/lib/live/state_diff_spec.rb
+++ b/spec/lib/live/state_diff_spec.rb
@@ -15,7 +15,7 @@ RSpec.describe Live::DiffHelper do
       result = round.live_results.find_by!(registration_id: registration_1.id)
       expect do
         result.update(best: 100)
-      end.to have_broadcasted_to(WcaLive.broadcast_key(round.id))
+      end.to have_broadcasted_to(Live::Config.broadcast_key(round.id))
         .from_channel(ApplicationCable::Channel)
     end
 


lib/live/helper.rb
```diff
@@ -37,5 +41,9 @@ def self.compute_created(before_hash, after_hash)
       created = created_ids.map { |id| after_hash[id] }
       created.presence
     end
+
+    def self.state_hash(live_state)
+      Digest::MD5.hexdigest(live_state.to_json)
```
gregorbg: Why not SHA1 though?

FinnIckler: I thought it would be slower but I think sha-1 is fine

app/controllers/live_controller.rb
```diff
@@ -14,7 +14,8 @@ def add_result
     round_id = params.require(:round_id)
     registration_id = params.require(:registration_id)
 
-    return render json: { status: "result already exist" }, status: :unprocessable_content if LiveResult.exists?(round_id: round_id, registration_id: registration_id)
+    # We create empty results when a round is open
+    return render json: { status: "round is not open" }, status: :unprocessable_content unless LiveResult.exists?(round_id: round_id, registration_id: registration_id)
```
gregorbg: Can you detect the difference between "Round has not been opened yet" and "Round is actually open but the competitor you're trying to add a result for was quit"?

app/models/live_attempt.rb
```diff
@@ -42,6 +42,23 @@ def self.build_with_history_entry(value, attempt_number, acting_user)
     )
   end
 
+  def self.attempt_changed?(before_attempts, after_attempts)
```
gregorbg: The method name is singular, but the arguments' names imply plural

app/models/live_attempt.rb
```diff
@@ -42,6 +42,23 @@ def self.build_with_history_entry(value, attempt_number, acting_user)
...
+    before_hash = (before_attempts || []).index_by { |a| a[:id] }
+    after_hash = (after_attempts || []).index_by { |a| a[:id] }
```
gregorbg: Why not use `{ it[:id] }`?

app/models/live_attempt.rb
```diff
@@ -42,6 +42,23 @@ def self.build_with_history_entry(value, attempt_number, acting_user)
...
+    after_hash.each do |id, after_attempt|
+      before_attempt = before_hash[id]
+      return true if before_attempt.nil? || before_attempt != after_attempt
+    end
+
+    # Deleted attempts
+    deleted = before_hash.keys - after_hash.keys
+    return true if deleted.any?
+
+    false
```
gregorbg: This should be equivalent to directly returning `deleted.any?`

app/models/live_result.rb
```diff
@@ -77,13 +75,45 @@ def values_for_sorting
...
+  end
+
+  private
+
+    def trigger_recompute_and_notify
+      before_state = round.live_state
+
       round.recompute_live_columns
+      # We need to reload because live results are changed directly on SQL level for more optimized queries
+      round.live_results.reload
```
gregorbg: The for what it's worth, why not just (a) reload the whole round or (b) call `round.live_results.reset` within `recompute_live_columns`? The latter is kinda like a "reload without loading", just invalidating so that the next time someone tries to access the relation it is definitely loaded fresh

lib/live/helper.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+require 'digest/sha1'
+
+module Live
+  module Helper
```
gregorbg: I'm not too delighted about the name, it's too generic. You can keep the module and everything it does, just think about a better name please 🙃

lib/live/diff_helper.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+require 'digest/sha1'
```
gregorbg: Is this actually required? For what it's worth, on my `bin/rails c` I can call `Digest::SHA1.hexdigest` without any `require` whatsoever... And `bin/rails` uses the exact same bootstrapping as the actual Rails application would, so the only reason for this line to be necessary is if the _production_ env bootstrapping is significantly different from the _local_ env bootstrapping

app/models/live_attempt.rb
```diff
@@ -42,6 +42,23 @@ def self.build_with_history_entry(value, attempt_number, acting_user)
...
+
+    # Updated or created attempts
+    after_hash.each do |id, after_attempt|
+      before_attempt = before_hash[id]
+      return true if before_attempt.nil? || before_attempt != after_attempt
+    end
+
+    # Deleted attempts
+    deleted = before_hash.keys - after_hash.keys
+    return true if deleted.any?
```
gregorbg: On second thought, these look an awful lot like `Helper.compute_deleted` and `Helper.compute_created`. Can you reuse the code?

lib/live/diff_helper.rb
```diff
@@ -0,0 +1,49 @@
...
+      after_hash.each do |id, after_result|
+        before_result = before_hash[id]
+        next unless before_result # Skip new results
+
+        result_diff = LiveResult.compute_diff(before_result, after_result)
+        updates << result_diff if result_diff.present?
+      end
+
+      updates.presence
+    end
```
gregorbg, line 25:
```suggestion
    def self.compute_updated(before_hash, after_hash)
      after_hash.filter_map do |id, after_result|
        before_result = before_hash[id]
        next unless before_result
    
        LiveResult.compute_diff(before_result, after_result).presence
      end.presence
    end
```

next-frontend/openapi/schemas/live/LiveRound.yaml
```diff
@@ -12,7 +12,10 @@ allOf:
           $ref: './LiveCompetitor.yaml'
       round_id:
         type: number
+      hash:
```
gregorbg: That name is correct but a bit generic. What about `state_hash` or `round_checksum` or anything along those lines? Less of "what this data _is_ computationally" and more of "what this data _does_ logically"

app/controllers/live_controller.rb
FinnIckler: not without the other PR being merged. But I should be able to get all the quit competitor for the previous round and check

app/models/live_attempt.rb
FinnIckler: I don't think we need a method for hash_1.keys - hash_2.keys

app/models/live_attempt.rb
FinnIckler: why not &:id :P

FinnIckler: ah  because it's an access...

next-frontend/openapi/schemas/live/LiveRound.yaml
FinnIckler: round_checksum doesn't really make sense because it is on the round object. but state_hash sounds good

app/jobs/add_live_result_job.rb
```diff
@@ -12,11 +12,9 @@ def perform(results, round_id, registration_id, entered_by)
...
     average, best = LiveResult.compute_average_and_best(attempts, round)
 
-    LiveResult.create!(registration_id: registration_id,
-                       round: round,
-                       live_attempts: attempts,
-                       last_attempt_entered_at: Time.now.utc,
-                       best: best,
-                       average: average)
+    # There always exist empty results
+    live_result = round.live_results.find_by!(registration_id: registration_id)
```
gregorbg, line 16: But what if not?

Daring suggestion (feel free to disagree / argue!): You should pass in the unique ID of the LiveResult itself that you want to "add" (as in, technically it is already added to the DB, so you're not really adding it. But you're _adding_ attempt(s) to that result)

app/models/live_attempt.rb
```diff
@@ -42,6 +42,21 @@ def self.build_with_history_entry(value, attempt_number, acting_user)
...
+    # Updated or created attempts
+    after_hash.each do |id, after_attempt|
+      before_attempt = before_hash[id]
+      return true if before_attempt.nil? || before_attempt != after_attempt
+    end
+
+    # Deleted attempts
+    deleted = before_hash.keys - after_hash.keys
+    deleted.any?
+  end
```
gregorbg, line 47: Fun fact: LLM suggests
```
def self.attempts_changed?(before_attempts, after_attempts)
  before_hash = (before_attempts || []).index_by { it[:id] }
  after_hash = (after_attempts || []).index_by { it[:id] }

  # If the sets of key-value pairs aren't identical, something changed.
  before_hash != after_hash
end
```
because it argues that Ruby is already performing (reasonably) deep comparison on hashes. If you're only interested in the "if changes", not in the "which changes", this seems appropriate.

gregorbg: The very dare-devil approach would be...
```
def self.attempts_changed?(before, after)
  Set.new(Array(before)) != Set.new(Array(after))
end
```

app/models/live_result.rb
```diff
@@ -77,13 +75,45 @@ def values_for_sorting
...
-  private
+  LIVE_STATE_SERIALIZE_OPTIONS = {
+    only: %w[advancing advancing_questionable average average_record_tag best global_pos local_pos registration_id single_record_tag],
+    methods: %w[],
+    include: [live_attempts: { only: %i[id value attempt_number] }],
+  }.freeze
 
-    def notify_users
-      ActionCable.server.broadcast(WcaLive.broadcast_key(round_id), as_json)
+  def live_state
```
gregorbg: I quite liked the `to_` prefix TBH, because it matches Ruby's `to_json`

gregorbg: And it matches our `to_wcif` et al

app/models/round.rb
```diff
@@ -291,6 +291,10 @@ def recompute_local_pos
     SQL
   end
 
+  def live_state
```
gregorbg: See my other comment about `to`

app/models/live_result.rb
```diff
@@ -77,13 +75,45 @@ def values_for_sorting
...
+  end
+
+  private
+
+    def trigger_recompute_and_notify
+      before_state = round.live_state
+
       round.recompute_live_columns
+      # We need to reload because live results are changed directly on SQL level for more optimized queries
+      round.live_results.reset
```
gregorbg: Why did you decide against moving this line into the body of `recompute_live_columns`?

gregorbg (changes requested):

app/jobs/add_live_result_job.rb
FinnIckler: if it doesn't the controller fails!

FinnIckler: but yeah, giving the registration_id into the job is a good idea as well

app/models/live_attempt.rb
FinnIckler: well the tests do not pass

app/jobs/add_live_result_job.rb
FinnIckler: I think I'd rather do that in the adding/updating a Result Frontend PR that has the update route in it as well

gregorbg: Okidoki!

app/models/live_attempt.rb
gregorbg: In that case please take ~20 minutes to jot some example cases and figure out why (not). I am pretty confident that at least the first version should work, preferrably even the second.

app/controllers/live_controller.rb
gregorbg: The other PR is approved now, please merge it first and then revisit this to see how difficult it would be to actually implement. If the solution is somewhat manageable :tm: then please add it before merging this one.

app/models/live_attempt.rb
FinnIckler: turned out there was still a couple bugs. Now it works

app/controllers/live_controller.rb
```diff
@@ -14,7 +14,16 @@ def add_result
...
-    return render json: { status: "result already exist" }, status: :unprocessable_content if LiveResult.exists?(round_id: round_id, registration_id: registration_id)
+    # We create empty results when a round is open
+    result_exists = LiveResult.exists?(round_id: round_id, registration_id: registration_id)
+
+    unless result_exists
+      round = Round.find(round_id)
+      return render json: { status: "round is not open" }, status: :unprocessable_content unless round.live_results.any?
+      return render json: { status: "user was quit from this round" }, status: :unprocessable_content if round.number != 1 && round.previous_round.live_results.where.not(quit_by_id: nil).where(registration_id: registration_id).any?
+
+      return render json: { status: "user is not part of this round" }, status: :unprocessable_content
```
gregorbg: Topic for debate: I think it is enough to report that "user is not part of this round". This is a "natural" consequence after being quit, and it saves us from the long and cumbersome `if` check here.

(And also, that if-check would need to become even more cumbersome once we start considering linked rounds)

app/models/live_result.rb
```diff
@@ -88,13 +86,43 @@ def values_for_sorting
...
-    def notify_users
-      ActionCable.server.broadcast(WcaLive.broadcast_key(round_id), as_json)
+  def to_live_state
+    serializable_hash(LIVE_STATE_SERIALIZE_OPTIONS)
+  end
+
+  def self.compute_diff(before_result, after_result)
+    diff = { "registration_id" => after_result["registration_id"] }
+
+    LIVE_STATE_SERIALIZE_OPTIONS[:only].map.each do |field|
```
gregorbg: I think the `.map.each` is redundant... Simply `.each` should do the trick

spec/requests/live_add_spec.rb
```diff
@@ -34,13 +35,13 @@
       expect(result.average).to eq 333
     end
 
-    it "Can't add result if it already exist" do
+    it "Can't add result if round isn't open yet" do
```
gregorbg, line 38: This now begs for the test case "if round is open but competitor is not part of that round"

app/controllers/live_controller.rb
gregorbg: Do organizers have a legitimate need, or is there a relevant use-case where we need to distinguish between "not part of the round because quit" and "not part of the round because they never progressed in the first place"?

FinnIckler: probably not. I was thinking of skipping it too

app/models/live_result.rb
FinnIckler: oh yeah, not sure what happened there

lib/live/diff_helper.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+module Live
```
gregorbg, line 3: Re-reading your code made me realize: There is also `lib/wca_live.rb` already, could these two utility classes somehow be merged?

gregorbg: Or at least made part of the same module, instead of having one module "Live" and one module "WcaLive"?

app/models/live_result.rb
```diff
@@ -88,13 +86,43 @@ def values_for_sorting
...
+  def to_live_state
+    serializable_hash(LIVE_STATE_SERIALIZE_OPTIONS)
+  end
+
+  def self.compute_diff(before_result, after_result)
+    diff = { "registration_id" => after_result["registration_id"] }
+
+    LIVE_STATE_SERIALIZE_OPTIONS[:only].each do |field|
+      diff[field] = after_result[field] if before_result[field] != after_result[field]
     end
```
```suggestion
    changed_vals = after.slice(*LIVE_STATE_SERIALIZE_OPTIONS[:only])
        .reject { |k, v| before[k] == v }

    diff = changed_vals.merge(registration_id: after_result["registration_id"])
```

app/models/live_result.rb
```diff
@@ -88,13 +86,43 @@ def values_for_sorting
...
-    def trigger_recompute_columns
+    # Include new attempts if they have changed, it's too much of a hassle to
+    # replace single values in the frontend.
+    diff["live_attempts"] = after_result["live_attempts"] if LiveAttempt.attempts_changed?(
+      before_result["live_attempts"],
+      after_result["live_attempts"],
+    )
+
+    # Only return if there are actual changes
+    diff.keys.size > 1 ? diff : nil
```
```suggestion
    diff if diff.except(:registration_id).present?
```

lib/live/diff_helper.rb
```diff
@@ -0,0 +1,42 @@
...
+        before_result = before_hash[id]
+        next unless before_result
+
+        LiveResult.compute_diff(before_result, after_result).presence
+      end.presence
+    end
+
+    def self.compute_deleted(before_hash, after_hash)
+      deleted_ids = before_hash.keys - after_hash.keys
+      deleted_ids.presence
```
gregorbg:
```suggestion
      (before_hash.keys - after_hash.keys).presence
```

lib/live/diff_helper.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+
+    def self.compute_created(before_hash, after_hash)
+      created_ids = after_hash.keys - before_hash.keys
+      created = created_ids.map { |id| after_hash[id] }
+      created.presence
```
gregorbg:
```suggestion
      after_hash.slice(*created_ids).values.presence
```

next-frontend/src/components/live/LiveUpdatingResultsTable.tsx
```diff
@@ -2,20 +2,33 @@
...
+
+  for (const result of previousResults) {
+    if (deletedSet.has(result.registration_id)) continue;
 
-  return [...resultsWithoutNewResult, newResult];
+    const update = updatesMap.get(result.registration_id);
+    results.push(update ? { ...result, ...update } : result);
+  }
+
+  return results.concat(created);
```
gregorbg:
```suggestion
  const diffedResults = previousResults
    .filter((res) => !deletedSet.has(res.registration_id))
    .map((res) => {
      const update = updatesMap.get(result.registration_id);
      return update ? { ...res, ...update } : res;
    });
    
    return [...reconciled, ...created];
```

gregorbg (changes requested):

next-frontend/src/components/live/LiveUpdatingResultsTable.tsx
FinnIckler: This does two passes on the previousResults array, I intentionally tried to write it so it does only one pass

lib/live/diff_helper.rb
FinnIckler: I moved them into the same module

next-frontend/src/components/live/LiveUpdatingResultsTable.tsx
gregorbg: Why is that a problem? Even smartphones these days have processors that can handle millions of rows, and we can have ~2000 at the very most.

Also, it actually does _one_ pass on the `previousResults` array, which results in a new, filtered array, and then does another pass on that :D

FinnIckler: okay

PR #13304, opened 2026-02-02
DevDump Size Reduction

dunkOnIT:
This PR: 
- Blanks `round.round_results` in dev dump (round results have been cleared from the prod db so this is not immediately necessary, but it's an easy 1-line change that gives insurance against round_results filling up again)
- Skips all rows when dumping the CAD tables
- Performs CAD at the end of the dev dump load

### DIFF
--- a/lib/tasks/db.rake
+++ b/lib/tasks/db.rake
@@ -163,7 +163,7 @@ namespace :db do
 
           # Run the CAD jobs so that results are available
           LogTask.log_task "Populating CAD tables" do
-            ComputeAuxiliaryData.new.perform
+            AuxiliaryDataComputation.compute_everything
           end
         end
       end


lib/tasks/db.rake
```diff
@@ -160,6 +160,11 @@ namespace :db do
             owner_id: User.find_by!(wca_id: "2005FLEI01").id,
             owner_type: "User",
           )
+
+          # Run the CAD jobs so that results are available
+          LogTask.log_task "Populating CAD tables" do
+            ComputeAuxiliaryData.new.perform
```
gregorbg: This is the _background job_ which triggers CAD actions. Basically a "Sidekiq adapter" for letting Sidekiq manage CAD.

You want the jobs to be run directly here on-the-fly, so no need to talk to the Sidekiq adapter. Use `AuxiliaryDataComputation.compute_everything` and call it a day.

PR #13306, opened 2026-02-02
Attempt at fixing test flakes

gregorbg:
1. The spec with the "Edit Tabs" was using a `within` call. My trace suggests that this `within` does not wait for the page to load, so I inserted a silly `expect` just to make sure the page has fully loaded.
2. The 2FA one is more tricky to track down because it writes session IDs in the background. I would assume it depends on the DB insertion order of the two `let`s, so I tried to tame this by only triggering the insert on demand of the first test case


FinnIckler:
still gotta merge in the other PR

PR #13337, opened 2026-02-03
Nextjs WCA Live round admin page

FinnIckler:
Still haven't routed the buttons to the routes, but opening for feedback, especially if I should include the `openable` and `clearable` fields in the route. These can be computed in the frontend, but I think especially for 3rd Party developers that don't know the rules these are good to have.

Frontend still looks pretty crappy, especially the text alignment
image

### DIFF
--- a/app/controllers/api/v1/live/live_controller.rb
+++ b/app/controllers/api/v1/live/live_controller.rb
@@ -7,13 +7,12 @@ class Api::V1::Live::LiveController < Api::V1::ApiController
   def add_or_update_result
     results = params.expect(attempts: [%i[value attempt_number]])
     round_id = params.require(:round_id)
-    competition_id = params.require(:competition_id)
+    competition = Competition.find(params.require(:competition_id))
     registration_id = params.require(:registration_id)
 
-    round = Round.find_by_wcif_id!(round_id, competition_id)
+    round = Round.find_by_wcif_id!(round_id, competition.id)
 
-    # TODO: add require_managed! from round admin PR
-    require_user
+    require_manage!(competition)
 
     # We create empty results when a round is open
     live_result = round.live_results.find_by(registration_id: registration_id)


app/models/round.rb
```diff
@@ -434,6 +438,18 @@ def to_live_json(only_podiums: false)
     }
   end
 
+  def to_live_admin_json
+    {
+      **self.to_wcif,
+      "open" => open?,
+      "locked" => live_results.any?(&:locked?),
+      "openable" => !open? && (number == 1 || (previous_round.open? && previous_round.score_taking_done?)),
```
gregorbg: Wouldn't `score_taking_done?` be enough? I don't understand why the previous round needs to be open in order for this round to be openable.

In particular, the only way (that I can see*) for the previous round to _not_ be open is precisely because it was locked when the current round was opened.

*please tell me if I'm missing something

next-frontend/openapi/schemas/live/LiveRoundAdmin.yaml
```diff
@@ -0,0 +1,16 @@
...
+  - type: object
+    properties:
+      open:
+        type: boolean
+      locked:
+        type: boolean
+      openable:
+        type: boolean
+      clearable:
+        type: boolean
```
gregorbg: Here's a wild idea: What about a kind of "lifecycle" enum? Instead of separate booleans, you have one string which can take fixed, pre-defined values like `closed` (or `pending`), `open` and `locked` (or `done`).

Most of these booleans can be "derived" from the strings by simply checking "is it currently in lifecycle state XY?". The things about openable and clearable need information about the previous round though.

I am undecided whether the backend should already encode this "information from previous rounds" into the payload, or whether we should rely on the frontend to do that (because at least in the admin overview, it definitely also has the other rounds's JSONs laying around)

app/models/round.rb
```diff
@@ -434,6 +438,18 @@ def to_live_json(only_podiums: false)
...
+      "clearable" => open? && (number == 1 || previous_round&.locked?),
```
gregorbg: Why is "locked" enough? I would expect `score_taking_done?` here

gregorbg: On second thought, I realized that I argued something similar (but slightly opposite direction) for `openable` above. In general, maybe we should think about the conceptual notion of "when can a round be opened" and "when can a round be cleared" first, and _then_ think about the best way to implement that?

app/models/round.rb
FinnIckler: it's because score_taking_done? will return true when a round isn't open

FinnIckler: because there are no live_results so it thinks 0/0 => is done

app/models/round.rb
FinnIckler: locked always implies score_taking_done?. But score_taking_done? doesn't imply locked

app/models/round.rb
gregorbg: Then that mostly feels like an implementation error in scoretaking_done. Or (if necessary, but currently I can't see why or how) we have two separate "are all results entered, even if there are none" and "are the results _actually, meaningfully_ entered" versions of `scoretaking_done`

app/models/round.rb
```diff
@@ -372,6 +372,22 @@ def lock_results(locking_user)
     results_to_lock.update_all(locked_by_id: locking_user.id)
   end
 
+  STATE_LOCKED = "locked"
+  STATE_OPEN = "open"
+  STATE_READY = "ready"
+  STATE_PENDING = "pending"
+
+  def lifecycle_state
+    return STATE_LOCKED if live_results.first&.locked?
```
gregorbg: Grrrr... If anything, we should use `all` instead of `first`. But then in case of "score taking done", we don't use this approach and we do "actual count == expected count" instead. So we should agree on one and stick with that

FinnIckler: so something like `live_results.locked.count == total_competitors` ?

next-frontend/openapi/schemas/live/LiveRoundAdmin.yaml
gregorbg, line 1: Please see whether you can define `state` as *one* enum and map them to different sub-types using the `discriminator` key: https://swagger.io/docs/specification/v3_0/data-models/inheritance-and-polymorphism/

FinnIckler: I tried to do it like this
```
allOf:
  - $ref: '../wcif/WcifRound.yaml'
  - type: object
    discriminator:
      propertyName: state
      mapping:
        open: './AdminRounds/OpenRound.yaml'
        locked: './AdminRounds/LockedRound.yaml'
        pending: './AdminRounds/PendingReadyRound.yaml'
        ready: './AdminRounds/PendingReadyRound.yaml'
    oneOf:
      - $ref: './AdminRounds/OpenRound.yaml'
      - $ref: './AdminRounds/LockedRound.yaml'
      - $ref: './AdminRounds/PendingReadyRound.yaml'
```
With a yaml looking like this
```
type: object
required:
  - state
  - total_competitors
  - competitors_live_results_entered
properties:
  state:
    type: string
    enum: ["open"]
  total_competitors:
    type: number
  competitors_live_results_entered:
    type: number

```
but the generates a type like this
`LiveRoundAdmin: components["schemas"]["WcifRound"] & Omit<components["schemas"]["OpenRound"] | components["schemas"]["LockedRound"] | components["schemas"]["PendingReadyRound"], "state">;`
Which omits state explicitly. I don't know why and LLM didn't really work either

FinnIckler: I will keep it in stash and revisit it later

gregorbg: Found a way to make this work. It's definitely unintuitive though and maaay(??) be a bug/quirk in the Redocly parser. Let's discuss the details on Friday...

FinnIckler: does it though? the type checker says no

gregorbg: Now the type checker says Yes 😜 

The solution I had to apply is super weird though. Directly doing `$ref` (without the seemingly(!!) redundant `allOf`) makes the Openapi -> TS compiler cut some corners that it shouldn't. I can give you the details on Friday.

FinnIckler: cool! I think I am done with the PR so you can review it whenever

app/controllers/api/v1/live/live_controller.rb
```diff
@@ -1,7 +1,9 @@
 # frozen_string_literal: true
 
 class Api::V1::Live::LiveController < Api::V1::ApiController
-  skip_before_action :require_user, only: %i[round_results by_person podiums]
+  protect_from_forgery with: :null_session
```
gregorbg, line 4: Why are you introducing CSRF tokens?

app/controllers/api/v1/live/live_controller.rb
```diff
@@ -11,6 +13,13 @@ def round_results
     render json: round.to_live_json
   end
 
+  def admin
```
gregorbg: Then perhaps something like `admin_rounds`? Just "live admin" does not tell me a lot about what the endpoint actually does

app/controllers/api/v1/api_controller.rb
```diff
@@ -28,4 +33,8 @@ def render_error(http_status, error, data = nil)
   rescue_from ActiveRecord::RecordNotFound do |e|
     render json: { error: e.to_s, data: { model: e.model, id: e.id } }, status: :not_found
   end
+
+  rescue_from WcaExceptions::NotPermitted do |e|
+    render json: { status: e.error_details }, status: e.status
```
gregorbg: There is a common parent class called `WcaExceptions::ApiException`. The v0 controllers handle exceptions like this:

```
rescue_from WcaExceptions::ApiException do |e|
  render status: e.status, json: { error: e.to_s }.reverse_merge(e.error_details.compact)
end
```

app/models/round.rb
```diff
@@ -191,6 +191,11 @@ def open_and_lock_previous(locking_user)
     [open_count, round_to_lock.lock_results(locking_user)]
   end
 
+  def clear_round!
+    live_results.destroy_all
+    open_round!
```
gregorbg: This should probably happen inside a transaction

app/models/round.rb
```diff
@@ -378,6 +383,27 @@ def lock_results(locking_user)
...
+
+    STATE_PENDING
+  end
+
+  def open?
+    live_results.any?
+  end
+
+  def locked?
+    live_results.locked == total_competitors
```
gregorbg: (a) isn't there a `count` missing here?
(b) won't this suffer from the same "zero-mistake" that you're fixing for `score_taking_done?` straight above?

app/models/round.rb
```diff
@@ -441,6 +467,26 @@ def to_live_json(only_podiums: false)
     }
   end
 
+  def to_live_admin_json
```
gregorbg: Just curious, is there anything in here that's so confidential we can only reveal it to admins?
The number of competitors is public, John Doe can count the number of rounds when looking at any open round. The number of entered results is also effectively public.

TL;DR Consider merging this into the `to_live_json`

config/routes.rb
```diff
@@ -375,9 +375,11 @@
           namespace :live do
             get '/rounds/:round_id' => 'live#round_results', as: :live_round_results
             put '/rounds/:round_id/open' => "live#open_round", as: :live_round_open
+            put '/rounds/:round_id/clear' => "live#clear_round", as: :live_round_clear
             put '/rounds/:round_id/:registration_id' => 'live#quit_competitor', as: :quit_competitor_from_round
             get '/podiums' => 'live#podiums', as: :live_podiums
             get '/registrations/:registration_id' => 'live#by_person', as: :get_live_by_person
+            get '/admin' => 'live#admin', as: :live_admin
```
gregorbg: Related to my other comment above: This can just be a `/rounds`. Not sure why we need to call it "admin".

next-frontend/src/app/(payload)/payload/importMap.js
gregorbg, line 1: Unrelated change

gregorbg: We should probably just push this individually as a hotfix to master, because it will affect literally everybody working on Next frontend

app/controllers/api/v1/live/live_controller.rb
FinnIckler: I am not introducing them, I am turning them off. Because now we have PUT requests which are protected

app/models/round.rb
FinnIckler: well the issue is that to_live_json contains all the results and and all the competitors. Which is gonna be fairly expensive to call for all rounds when we don't need it

gregorbg: Ah, then I think we should consider one `to_live_results_json` and `to_live_info_json` (names tentative, I especially don't like the "info" part...)

But my point still is: There is nothing in the "slim" JSON (the one without the results) that is inherently "admin".

gregorbg: Alternatively: `to_live_json(include_results: true)`

FinnIckler: to_live_json already has(only_podiums: true) and I don't want to add another argument. I'd rather rename the methods

next-frontend/src/app/(wca)/competitions/[competitionId]/live/admin/page.tsx
```diff
@@ -0,0 +1,118 @@
...
+  params: Promise<{ competitionId: string }>;
+}) {
+  const { competitionId } = await params;
+  const { t } = await getT();
+
+  const session = await auth();
+
+  if (!session) {
+    return <p>Please Log in</p>;
+  }
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/live/admin/page.tsx:33]: Do we not have a translation string for that?

Comment (FinnIckler) [reply next-frontend/src/app/(wca)/competitions/[competitionId]/live/admin/page.tsx:33]: We do, but I want to do a translation pass on the whole wca live later separately

PR #13344, opened 2026-02-04
Improve connection handling during dev dump reload

gregorbg:
Addresses recent staging import problems

CC @dunkOnIT

### DIFF
--- a/lib/database_dumper.rb
+++ b/lib/database_dumper.rb
@@ -1419,7 +1419,7 @@ module DatabaseDumper
     ActiveRecord::Base.connection.execute("SET foreign_key_checks=0")
 
     LogTask.log_task "Populating sanitized tables in '#{dump_db_name}'" do
-      ActiveRecord::Base.connection.execute(<<~SQL)
+      ActiveRecord::Base.connection.execute(<<~SQL.squish)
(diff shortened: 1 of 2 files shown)


dunkOnIT:
Efforts to use this PR fail here:
```ruby
ActiveRecord::Base.connection.exec("SOURCE #{DbDumpHelper::DEVELOPER_EXPORT_SQL}")
```

Minor issue - `.exec` is not valid AR syntax. However, changing to `.execute` yields this error:
> Mysql2::Error: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'SOURCE wca-developer-database-dump.sql' at line 1 (Mysql2::Error)

Talking to ChatGPT suggests that we have to pipe that command via MySQL, not AR. I dipped my toes in by trying to change it to `DatabaseDumper.mysql()`, which got me past that line of code but nuked my db in the process - I wasn't using `reload`[1]. 

I'm now reloading my dev db, and will modify the dev dump with a `local` parameter[2] and use that + `LIMIT 100` to generate a dev dump I can use to test the new `load`.

[1] I figured that if I just had a file that inserted a line, I shouldn't use `reload` - which would create an unpopulated temp db - and should instead just have it applied directly on the existing database. That logic was probably flawed :P 
[2] similar to how I did for the results dump

dunkOnIT:
@thewca-bot deploy staging

PR #13348, opened 2026-02-04
WCA Live Nextjs Data Entry

FinnIckler:
Have not yet ported over the tests from the other PR.
Now that updating and creating results is the same thing when it comes to data entry, do we even have a POST and a PATCH endpoint. I have left them pointing to the same controller action for now, but that does feel silly. Not sure what makes more sense for 3rd party devs.

### DIFF
--- a/next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/AddResults.tsx
+++ b/next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/AddResults.tsx
@@ -1,6 +1,6 @@
 "use client";
 import { components } from "@/types/openapi";
-import { Grid, GridItem } from "@chakra-ui/react";
+import { SimpleGrid, GridItem } from "@chakra-ui/react";
 import AttemptsForm from "@/components/live/AttemptsForm";
 import { Format } from "@/lib/wca/data/formats";
 import LiveResultsTable from "@/components/live/LiveResultsTable";
@@ -26,7 +26,7 @@ export default function AddResults({
   const { pendingLiveResults } = useLiveResults();
 
   return (
-    <Grid templateColumns="repeat(16, 1fr)" gap="6">
+    <SimpleGrid columns={16} gap={6}>
       <GridItem colSpan={4}>
         <LiveResultAdminProvider
           format={format}
@@ -61,6 +61,6 @@ export default function AddResults({
           title={`${events.byId[eventId].name} - ${roundNumber}`}
         />
       </GridItem>
-    </Grid>
+    </SimpleGrid>
   );
 }


next-frontend/openapi/schemas/live/SubmitLiveResult.yaml
```diff
@@ -0,0 +1,12 @@
+SubmitLiveResult:
```
FinnIckler: need to remove the name here

app/controllers/api/v1/live/live_controller.rb
```diff
@@ -1,7 +1,35 @@
 # frozen_string_literal: true
 
 class Api::V1::Live::LiveController < Api::V1::ApiController
+  protect_from_forgery with: :null_session
```
gregorbg, line 4: Same as in my other PR: Why is a CSRF check suddenly necessary here?

(If you answer the review comments on the other PR first, feel free to just write "see other PR" here 😉 )

app/controllers/api/v1/live/live_controller.rb
```diff
@@ -1,7 +1,35 @@
...
   skip_before_action :require_user, only: %i[round_results by_person podiums]
+
+  def add_or_update_result
+    results = params.permit(attempts: %i[value attempt_number])[:attempts]
```
gregorbg: I think in Rails 8 the modern equivalent is...
```suggestion
    results = params.expect(attempts: [%i[value attempt_number]])
```

The outer array brackets (yes, double array fully intended!) are for Rails 8 to distinguish between the payloads `{ attempts: { value: 123, attempt_number: 5 } }` and `{ attempts: [ { value: 123, attempt_number: 5 }, { value: 456, attempt_number: 1 } ] }`

app/models/live_result.rb
```diff
@@ -24,8 +22,8 @@ class LiveResult < ApplicationRecord
 
   DEFAULT_SERIALIZE_OPTIONS = {
     only: %w[global_pos local_pos registration_id round_id best average single_record_tag average_record_tag advancing advancing_questionable entered_at entered_by_id],
-    methods: %w[event_id attempts result_id],
-    include: %w[],
+    methods: %w[event_id result_id],
+    include: %w[live_attempts],
```
gregorbg: Why are you changing this from `attempts` to `live_attempts`?

app/models/live_attempt.rb
```diff
@@ -28,15 +28,15 @@ def to_result_attempt
...
+  def self.build_with_history_entry(value, attempt_number, acting_user_id)
     LiveAttempt.build(
       value: value,
       attempt_number: attempt_number,
       live_attempt_history_entries: [
         LiveAttemptHistoryEntry.build(
           value: value,
           entered_at: Time.now.utc,
-          entered_by: acting_user,
+          entered_by: acting_user_id,
```
gregorbg, line 41:
```suggestion
          entered_by_id: acting_user_id,
```

config/cable.yml
```diff
@@ -1,5 +1,6 @@
 development:
-  adapter: async
+  adapter: redis
+  url: <%= ENV.fetch("REDIS_URL") { "redis://wca_redis:6379/3" } %>
```
gregorbg: You should introduce a separate `ACTIONCABLE_REDIS_URL` in `env_config`, similar to how we already have `CACHE_REDIS_URL` and `SIDEKIQ_REDIS_URL` there.

Since production is not using redis for ActionCable, you can declare the variable as `optional` if `env_config` is running in production mode, that way you don't have to do any Terraform changes

next-frontend/src/app/(payload)/payload/importMap.js
gregorbg, line 1: Again unrelated, as per my comment on the other PR

gregorbg (changes requested):
I have not taken a look at the frontend code yet. But I already noticed this PR does an awful lot, would you mind splitting it up?
1. Use Redis for AC in local
2. Change the way that AddResultsJob works
3. Everything else from this PR

app/controllers/api/v1/live/live_controller.rb
FinnIckler: in this case it is the PATCH requests that are protected by CSRF and have to be turned off. They are always on!

app/models/live_result.rb
FinnIckler: I thought my serialization had to be the same name as what's coming from the diff protocol, but now that we are planning to remap for compression it doesn't matter anyway

next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/AddResults.tsx
```diff
@@ -0,0 +1,190 @@
...
+  results: components["schemas"]["LiveResult"][];
+  eventId: string;
+  roundId: string;
+  competitionId: string;
+  competitors: components["schemas"]["LiveCompetitor"][];
+}) {
+  const event = events.byId[eventId];
+  const solveCount = event.recommendedFormat.expected_solve_count;
+
+  const [registrationId, setRegistrationId] = useState<number | null>(null);
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/AddResults.tsx]: Using `undefined` here lets you just do `useState<number>();` I think

next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/AddResults.tsx
```diff
@@ -0,0 +1,190 @@
...
+  const [attempts, setAttempts] = useState<number[]>(
+    zeroedArrayOfSize(solveCount),
+  );
+  const [error, setError] = useState<string>("");
+  const [success, setSuccess] = useState<string>("");
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/AddResults.tsx]: In these cases, TS will definitely be smart enough to infer the `<>` types

next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/AddResults.tsx
```diff
@@ -0,0 +1,190 @@
...
+        setAttempts(zeroedArrayOfSize(solveCount));
+        setError("");
+
+        setTimeout(() => setSuccess(""), 3000);
+      },
+      onError: () => {
+        setError("Failed to update results. Please try again.");
+      },
+    },
+  );
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/AddResults.tsx]: You should try to de-duplicate this as much as reasonable

next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/AddResults.tsx
```diff
@@ -0,0 +1,190 @@
...
+  const [success, setSuccess] = useState<string>("");
+
+  const api = useAPI();
+
+  const handleRegistrationIdChange = useCallback(
+    (value: number) => {
+      setRegistrationId(value);
+      const alreadyEnteredResults = results.find(
+        (r) => r.registration_id === value,
+      );
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/AddResults.tsx]: How will this hook (which I see is getting `results` passed in as a prop) find out about "newly entered" results that were submitted during the lifetime of this component?

Use-case: Enter competitor A, submit successfully, enter competitor B, submit successfully, realize a typo and go back to A to fix something.
--> How will the (more or less) freshly submitted results for A be accommodated?

next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/AddResults.tsx
```diff
@@ -0,0 +1,190 @@
...
+        },
+        body: {
+          attempts: attempts.map((attempt, index) => ({
+            value: attempt,
+            attempt_number: index + 1,
+          })),
+          registration_id: registrationId,
+        },
+      });
+    }
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/AddResults.tsx]: You should probably also deduplicate this

next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/page.tsx
```diff
@@ -0,0 +1,46 @@
...
+  return (
+    <Container bg="bg">
+      <PermissionCheck
+        requiredPermission="canAdministerCompetition"
+        item={competitionId}
+      >
+        <VStack align="left">
+          <Heading textStyle="h1">Live Results</Heading>
+          <AddResults
+            results={results!}
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/page.tsx]: Why this exclamation mark? Under what circumstances can `results` be `undefined`, and why is it safe to skip/ignore this case?

next-frontend/src/components/live/AttemptsForm.tsx
```diff
@@ -0,0 +1,101 @@
...
+  success,
+  header,
+  isPending,
+}: AttemptsFormProps) {
+  const { collection, filter } = useListCollection({
+    initialItems: competitors.map((c) => ({
+      ...c,
+      label: `${c.user_name} (${c.registrant_id})`,
+      value: c.id,
+    })),
```
gregorbg: According to https://ark-ui.com/docs/collections/list-collection section "Working with custom objects" (sorry they don't have permalinks on that page...) you can also set the `itemToString` and `itemToValue` properties on the list collection instead of having to "reshape" your collection items

next-frontend/src/components/live/AttemptsForm.tsx
```diff
@@ -0,0 +1,101 @@
...
+              ))}
+            </Combobox.Content>
+          </Combobox.Positioner>
+        </Portal>
+      </Combobox.Root>
+      {_.times(solveCount).map((index) => (
+        <AttemptResultField
+          eventId={eventId}
+          key={index}
+          value={attempts[index] ?? 0}
```
gregorbg: What exactly (in terms of stack trace) explodes if you don't have this `?? 0` fallback?

next-frontend/src/lib/hooks/useResultsSubscription.ts
```diff
@@ -27,7 +27,7 @@ export type DiffProtocolResponse = {
 };
 
 export default function useResultsSubscription(
-  roundId: number,
+  roundId: string,
```
gregorbg, line 49: Just curious: What is this change about? (I'm sure you have a reason, I just can't see it from pure code review alone)

FinnIckler: we don't work with numeric ids anymore but wcif ids. Sadly this affects almost all the prs that touch the live updating so it's duplicated throughout

FinnIckler: There probably should be a separate pr that renames all these into wcif_id and cleans up a bit

next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/AddResults.tsx
Comment (FinnIckler) [reply next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/AddResults.tsx]: that is true, that is currently only handled in LiveUpdatingResultsTable. So this needs to be refactored

next-frontend/src/components/live/AttemptsForm.tsx
FinnIckler: https://ark-ui.com/docs/collections/list-collection#working-with-custom-objects they do :P

FinnIckler:
I have moved the control flow around so I can have access to the updated live results in the AddResults component.

This caused some annoying code reuse and need to move things around which I should probably fix as a refactor PR first.

I also added a simple Version of adding pending results as feedback in the UX when you click submit. They still show up on the bottom which is not ideal, but I don't know what is better. Having them show up on the top, on the bottom, or at their "correct spot" which would mean we need to calculate average/best in the frontend which is meh.

next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/AddResults.tsx
```diff
@@ -0,0 +1,174 @@
...
+          value: attempt,
+          attempt_number: index + 1,
+        })),
+        registration_id: registrationId,
+      },
+    });
+  };
+
+  return (
+    <Grid templateColumns="repeat(16, 1fr)" gap="6">
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/AddResults.tsx]: Chakra has `SimpleGrid` which should be enough for most use cases

next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/AddResults.tsx
```diff
@@ -0,0 +1,174 @@
...
+              PDF
+            </Link>
+          </Button>
+          <Button asChild>
+            <Link
+              href={`/competitions/${competitionId}/live/rounds/${roundId}/double-check`}
+            >
+              Double Check
+            </Link>
+          </Button>
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/AddResults.tsx]: The `Link`s in here should be typed.

next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/AddResults.tsx
```diff
@@ -0,0 +1,174 @@
...
+            competitors={competitors}
+          />
+        )}
+        <LiveUpdatingResultsTable
+          eventId={eventId}
+          formatId={format.id}
+          competitionId={competitionId}
+          competitors={competitors}
+          isAdmin
+          title=""
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/AddResults.tsx]: This looks redundant, why is it necessary to manually set an empty string?

next-frontend/yarn.lock
gregorbg, line 1: Looks unrelated

next-frontend/src/providers/LiveResultProvider.tsx
```diff
@@ -75,16 +85,40 @@ export function LiveResultProvider({
...
+    (liveResult: PendingLiveResult) => {
+      updatePendingResults((pending) => [
+        ...pending,
+        ...applyDiffToLiveResults(
+          results.filter(
+            (r: LiveResult) => r.registration_id == liveResult.registration_id,
+          ),
+          [liveResult],
+          [],
+          [],
```
gregorbg: Consider using these as default values for `applyDiff`

next-frontend/src/components/live/LiveResultsTable.tsx
gregorbg, line 1: Does this changeset guarantee that pending results are displayed at the bottom of the table?

next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/AddResults.tsx
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/AddResults.tsx:1]: What about moving some (if not most) of this preamble, including the mutations, into their own separate `LiveResultsAdminContext`?

This new context could then also hold the `pendingResults`, although I'm not sure how much of a hassle it is to cross-link with websocket updates that come in. Maybe ActionCable supports adding a subscription to an already open channel? Just food for thought, not a strict requirement to have the pending results also in there.

But having a separate context for the mutations and stuff does seem like a good idea just so that you can avoid passing down all those props to `AttemptsForm`...

Comment (gregorbg) [reply next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/AddResults.tsx:1]: Image

next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/AddResults.tsx
Comment (FinnIckler) [reply next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/AddResults.tsx]: I am just going to remove them as they are going to be added as a component in https://github.com/thewca/worldcubeassociation.org/pull/13408/changes#diff-70453c07a5f8d60bb31eaf55f4f1c38ca3f5a711153994736365f42ae54ca5ad

next-frontend/src/components/live/LiveResultsTable.tsx
FinnIckler: they are in a separate table

next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/AddResults.tsx
Comment (FinnIckler) [reply next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/AddResults.tsx]: not in this case, because I want to especially have a 25% and 75% split

next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/AddResults.tsx
Comment (FinnIckler) [reply next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/AddResults.tsx:1]: Added a LiveResultAdminProvider

next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/AddResults.tsx
Comment (FinnIckler) [reply next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/AddResults.tsx]: Maybe something to work on for the mobile view?

Comment (gregorbg) [reply next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/AddResults.tsx]: Then take a look at https://www.chakra-ui.com/docs/components/simple-grid#column-span

gregorbg (changes requested):

next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/AddResults.tsx
Comment (FinnIckler) [reply next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/AddResults.tsx]: done

PR #13349, opened 2026-02-04
Show newcomer check page in sidenav always

danieljames-dj:
Now for past competitions, it will show something like:

Screenshot 2026-02-04 at 22 28 39

Along with this change, also did a minor change - "no newcomer" message is now not shown while the job is running.

### DIFF
--- a/spec/requests/results_submission_spec.rb
+++ b/spec/requests/results_submission_spec.rb
@@ -90,4 +90,42 @@ RSpec.describe ResultsSubmissionController do
       end
     end
   end
+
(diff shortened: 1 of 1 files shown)


spec/models/user_spec.rb
gregorbg, line 1: Are there test cases for hitting the controller and making sure it returns 400 (or other appropriate errors) when the competition is already passed?

danieljames-dj: No, added now 😬

PR #13352, opened 2026-02-06
Compress the Objects send over the diff protocol

FinnIckler:
This also removes sending over local_pos and global_pos. I tried giving the values that will be send over the most the shortest names. 

I have not done tie handling yet in the order by (I mean I have, but not in the mapping them to their positions)

### DIFF
--- a/next-frontend/src/types/live.ts
+++ b/next-frontend/src/types/live.ts
@@ -4,6 +4,7 @@ import { PartialExcept } from "@/lib/types/objects";
 export type LiveResult = components["schemas"]["RoundLiveResult"] | components["schemas"]["ByPersonLiveResult"];
 export type BaseLiveResult = components["schemas"]["BaseLiveResult"];
 export type LiveCompetitor = components["schemas"]["LiveCompetitor"];
+export type LiveAttempt = components["schemas"]["LiveAttempt"];
 export type LiveRound = components["schemas"]["LiveRound"];
 export type PendingLiveResult = PartialExcept<
   LiveResult,
--- a/app/assets/javascripts/users.js
+++ b/app/assets/javascripts/users.js
@@ -6,22 +6,6 @@ onPage('users#edit, users#update', function() {
     $('.form-group.user_avatar').toggle(!toDelete);
   }).trigger("change");
 
-  var $approve_wca_id = $('#approve-wca-id');
-  var $unconfirmed_wca_id = $("#user_unconfirmed_wca_id");
-  var $unconfirmed_wca_id_profile_link = $("a#unconfirmed-wca-id-profile");
-  $approve_wca_id.on("click", function(e) {
-    $("#user_wca_id").val($unconfirmed_wca_id.val());
-    $unconfirmed_wca_id.val('');
-    $unconfirmed_wca_id.trigger('input');
-  });
-  $unconfirmed_wca_id.on("input", function(e) {
-    var unconfirmed_wca_id = $unconfirmed_wca_id.val();
-    $approve_wca_id.prop("disabled", !unconfirmed_wca_id);
-    $unconfirmed_wca_id_profile_link.parent().toggle(!!unconfirmed_wca_id);
-    $unconfirmed_wca_id_profile_link.attr('href', "/persons/" + unconfirmed_wca_id);
-  });
-  $unconfirmed_wca_id.trigger('input');
-
   // Change the 'section' parameter when a tab is switched.
   $('a[data-toggle="tab"]').on('show.bs.tab', function() {
     var section = $(this).attr('href').slice(1);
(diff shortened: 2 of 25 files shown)


lib/live/diff_helper.rb
```diff
@@ -36,5 +36,39 @@ def self.compute_created(before_hash, after_hash)
...
+    COMPRESSION_MAP = {
+      "advancing" => "ad",
+      "advancing_questionable" => "adq",
+      "average" => "a",
+      "best" => "b",
+      "average_record_tag" => "art",
+      "single_record_tag" => "srt",
+      "registration_id" => "r",
+      "live_attempts" => "la",
+      "id" => "id",
```
FinnIckler: we don't even need to send id for live_attempts actually

gregorbg: greetings from the Dual Rounds discussion 😛

lib/live/diff_helper.rb
```diff
@@ -36,5 +36,39 @@ def self.compute_created(before_hash, after_hash)
...
+      "value" => "v",
+      "attempt_number" => "an",
+    }
+
+    # To send even less data, we shorten the quite long attribute names
+    def self.compress_payload(diff)
+      diff["created"] = compress_diff(diff["created"]) if diff["created"].present?
+      diff["updated"] = compress_diff(diff["updated"]) if diff["updated"].present?
+      diff
```
gregorbg: What about...
```suggestion
compressed_diff = diff.slice("created", "updated").transform_values { compress_diff(it) }

{ **diff, **compressed_diff }
```

next-frontend/openapi/schemas/live/BaseLiveResult.yaml
```diff
@@ -23,6 +23,8 @@ properties:
     type: boolean
   advancing_questionable:
     type: boolean
+  event_id:
+    type: string
```
gregorbg: The `LIVE_STATE_SERIALIZABLE_OPTIONS` in the backend does not contain `event_id`. So why is it fine to include it in the base type?

next-frontend/src/lib/live/applyDiffToLiveResults.ts
```diff
@@ -2,26 +2,64 @@ import {
...
+  decompressDiff,
+  decompressFullResult,
+  decompressPartialResult,
+} from "@/lib/live/decompressDiff";
+import { parseActivityCode } from "@/lib/wca/wcif/rounds";
+
+const augmentResults = (
+  r: BaseLiveResult,
+  roundWcifId: string,
+  eventId: string,
```
gregorbg: In this very PR, if you really add `event_id` to the YAML type of `BaseLiveResult`, then it feels redundant to "augment" it by also passing in an `eventId`. And even more redundant given the fact that `roundWcifId` also encapsules the `eventId`

next-frontend/src/types/live.ts
```diff
@@ -3,9 +3,10 @@ import { PartialExcept } from "@/lib/types/objects";
...
 
 export type LiveResult = components["schemas"]["RoundLiveResult"] | components["schemas"]["ByPersonLiveResult"];
+export type BaseLiveResult = components["schemas"]["BaseLiveResult"];
 export type LiveCompetitor = components["schemas"]["LiveCompetitor"];
 export type LiveRound = components["schemas"]["LiveRound"];
 export type PendingLiveResult = PartialExcept<
   CompressedLiveResult,
-  "registration_id" | "live_attempts"
+  "r" | "la"
 >;
```
gregorbg, line 12: Why is this typed based on the Compressed version? It's not sent over the websocket cable, so no harm in using the full keys. Or what am I missing?

gregorbg (changes requested):
Just a bit of confusion about the type hierarchy still... Otherwise looking good!

next-frontend/openapi/schemas/live/BaseLiveResult.yaml
FinnIckler: oh. yeah that should be removed then

next-frontend/src/types/live.ts
FinnIckler: I pipe pending live results through applyDiff so I have to either do the conversion in addPendingResults or when calling addPendingResults. Both works. I choose doing it when calling so I needed the type change. I can move it though

next-frontend/openapi/schemas/live/BaseLiveResult.yaml
FinnIckler: event_id should be completely removed, as it is part of the wcif_id

next-frontend/src/types/live.ts
gregorbg: I think we should treat pending results as "full" results as much as possible. Or in other words, narrow down to the `applyDiff` format only at the very last moment

gregorbg (approved):

Issue #13350, opened 2026-02-06
Add name validator to ensure it doesn't contain double quotes


maxidragon:
Add a name validator that checks whether the name field doesn't contain double quotes (") and would raise a warning during results submission.

PR #13356, opened 2026-02-07
added name validator to ensure it doesn't contain special character

pranav-027:
fixes: https://github.com/thewca/worldcubeassociation.org/issues/13350

### DIFF
--- a/spec/lib/results_validators/persons_validator_spec.rb
+++ b/spec/lib/results_validators/persons_validator_spec.rb
@@ -171,10 +171,6 @@ RSpec.describe PV do
                                      event_id: "333oh",
                                      round: round_333_oh)
         res_bad_parenthesis.person.update(name: "Bad Parenthesis Guy（test）")
-        res_special_chars = create(:inbox_result,
-                                   competition: competition1,
-                                   event_id: "333oh",
-                                   round: round_333_oh)
         res_special_chars1 = create(:inbox_result,
                                     competition: competition1,
                                     event_id: "333oh",


lib/results_validators/persons_validator.rb
```diff
@@ -74,6 +75,9 @@ def self.name_validations(name, competition_id = nil, **_message_args)
       # Check for wrong parenthesis type.
       validation_issues << ValidationError.new(WRONG_PARENTHESIS_TYPE_ERROR, :persons, competition_id, name: name) if /[（）]/.match?(name)
 
+      # Check for special characters in name.
+      validation_issues << ValidationWarning.new(SPECIAL_CHARACTERS_IN_NAME_WARNING, :persons, competition_id, name: name) if /[^[:alpha:]\s\-'.()]/.match?(name)
```
gregorbg, line 82: This RegExp is quite dense, can you please add a (short) code comment to explain how you created it and why it works?

spec/lib/results_validators/persons_validator_spec.rb
```diff
@@ -170,6 +171,11 @@
                                      event_id: "333oh",
                                      round: round_333_oh)
         res_bad_parenthesis.person.update(name: "Bad Parenthesis Guy（test）")
+        res_special_chars = create(:inbox_result,
+                                   competition: competition1,
+                                   event_id: "333oh",
+                                   round: round_333_oh)
+        res_special_chars.person.update(name: "John \"Doe\"")
```
gregorbg: Is this the only case of "special characters" that you want to test? It seems like this is just a "name may not contain quotes" test case, but your check is actually called "may not contain special characters of any kind". So perhaps you should add one or two more test cases with special characters (that WRT is actually encountering in the wild)

spec/lib/results_validators/persons_validator_spec.rb
```diff
@@ -243,6 +249,12 @@
           RV::ValidationWarning.new(PV::EMPTY_GENDER_WARNING,
                                     :persons, competition1.id,
                                     name: res_whitespace.person.name),
+          RV::ValidationWarning.new(PV::SPECIAL_CHARACTERS_IN_NAME_WARNING,
+                                    :persons, competition1.id,
+                                    name: res_special_chars.person.name),
+          RV::ValidationWarning.new(PV::SPECIAL_CHARACTERS_IN_NAME_WARNING,
+                                    :persons, competition1.id,
+                                    name: res_bad_parenthesis.person.name),
```
gregorbg, line 273: Why is this guy being detected by the new check? Should this guy be detected by your new check?

lib/results_validators/persons_validator.rb
pranav-027: done

spec/lib/results_validators/persons_validator_spec.rb
pranav-027: added more tests

spec/lib/results_validators/persons_validator_spec.rb
pranav-027: This guy was created to check the error for full-width parentheses, which we don’t accept in our database. Since they are special characters they are also being detected by my special-character check.

The existing validation for invalid parentheses raises an error, while my check only raises a warning. I think it’s okay to keep both, even if it’s a bit redundant.

gregorbg (approved):

Issue #13263, opened 2026-01-24
Contact forms should send in English locale


danieljames-dj:
Since the contact forms are to contact the WCA staffs, this should be in English. Currently it's sent in the user's locale.

Change should be done mostly here: https://github.com/thewca/worldcubeassociation.org/blob/1f6d7c7e5ad3644d488074ea5968b0b28554b123/app/controllers/contacts_controller.rb#L4-L12

LLM suggested following to make it deliver in English:
```
  private def maybe_send_contact_email(contact)
    if !contact.valid?
      render status: :bad_request, json: { error: "Invalid contact object created" }
    # Wrap the delivery in the English locale block
    elsif I18n.with_locale(:en) { contact.deliver }
      render status: :ok, json: { message: "Mail sent successfully" }
    else
      render status: :internal_server_error, json: { error: "Mail delivery failed" }
    end
  end
```

This needs to be tested as well as we don't know if this is hallucination. If anybody interested to work on this, please comment.

pranav-027:
Should we exclude the competition contact form from this?
Organizers of a specific region might expect their local language.

danieljames-dj:
I agree on excluding competition contact form.

PR #13358, opened 2026-02-07
Enforce English locale in contacts form

pranav-027:
fixes: https://github.com/thewca/worldcubeassociation.org/issues/13263

### DIFF
--- a/app/controllers/contacts_controller.rb
+++ b/app/controllers/contacts_controller.rb
@@ -6,7 +6,7 @@ class ContactsController < ApplicationController
   private def maybe_send_contact_email(contact, force_locale: nil)
     if !contact.valid?
       render status: :bad_request, json: { error: "Invalid contact object created" }
-    elsif force_locale ? I18n.with_locale(force_locale) { contact.deliver } : contact.deliver
+    elsif I18n.with_locale(force_locale) { contact.deliver }
       render status: :ok, json: { message: "Mail sent successfully" }
     else
       render status: :internal_server_error, json: { error: "Mail delivery failed" }
@@ -192,7 +192,7 @@ class ContactsController < ApplicationController
       # Convert flash to a flash.now, since we're about to render, not redirect.
       flash.now[:recaptcha_error] = flash[:recaptcha_error]
       render fail_view
-    elsif force_locale ? I18n.with_locale(force_locale) { @contact.deliver } : @contact.deliver
+    elsif I18n.with_locale(force_locale) { @contact.deliver }
       flash[:success] = I18n.t('contacts.messages.success')
       redirect_to success_url
     else


app/controllers/contacts_controller.rb
```diff
@@ -186,7 +187,7 @@ def dob_create
       # Convert flash to a flash.now, since we're about to render, not redirect.
       flash.now[:recaptcha_error] = flash[:recaptcha_error]
       render fail_view
-    elsif @contact.deliver
+    elsif I18n.with_locale(:en) { @contact.deliver }
```
gregorbg: You should be consistent here about passing in an (optional) override locale the same way that you're doing for `maybe_send_contact_email` above.

It is not needed _right now_, but it will be good to have it ready in the future and to keep our code consistent.

app/controllers/contacts_controller.rb
```diff
@@ -1,10 +1,10 @@
 # frozen_string_literal: true
 
 class ContactsController < ApplicationController
-  private def maybe_send_contact_email(contact)
+  private def maybe_send_contact_email(contact, force_locale: :en)
     if !contact.valid?
       render status: :bad_request, json: { error: "Invalid contact object created" }
-    elsif contact.deliver
+    elsif force_locale ? I18n.with_locale(force_locale) { contact.deliver } : contact.deliver
```
gregorbg: You can actually pass `nil` to `with_locale`, and it will just "skip" the forcing part:

Image

app/controllers/contacts_controller.rb
```diff
@@ -21,6 +21,7 @@ class ContactsController < ApplicationController
         request: request,
         logged_in_email: current_user&.email || 'None',
       ),
+      force_locale: nil,
```
gregorbg: I think it will be better to explicitly force `en` in the places where needed, rather than "un-focing" `en` in the places where it's not needed.

Another developer using this code will interpret the `force_locale` as "I can force a locale if I want to, otherwise it behaves normally". Not as "I can force a locale if I don't want English".

The consequence of this change will be that you're mentioning `:en` tag in a lot of places, so you can consider extracting it into a `CONTACT_DEFAULT_LANGUAGE` constant in this controller for code reuse.

gregorbg (changes requested):

app/controllers/contacts_controller.rb
pranav-027: Done

app/controllers/contacts_controller.rb
pranav-027: Done

app/controllers/contacts_controller.rb
pranav-027: done

pranav-027: done

PR #13363, opened 2026-02-08
Compute duplicates using exact_first_last_dob algorithm too

danieljames-dj:
This PR implements a new algorithm for newcomer check - exact_first_last_dob. This checks for cases where the first name, last name and DOB are same. This won't be counted in `jarowinkler` algorithm, because if the name of persons are "John Doe" and "John Middle Name Doe", the `jarowinkler` algorithm is going to return low score.

Currently WRT is having a very big query which we run every year or two and manually handle the duplicate cases. This PR will allow WRT to close the investigation forever.

### DIFF
--- a/lib/finish_unfinished_persons.rb
+++ b/lib/finish_unfinished_persons.rb
@@ -109,9 +109,11 @@ module FinishUnfinishedPersons
     sanitized_roman_name = self.remove_accents roman_name
     name_parts = sanitized_roman_name.gsub(/[^a-zA-Z ]/, '').upcase.split
 
-    name_parts.pop if name_parts.length > 1 && GENERATIONAL_SUFFIXES.include?(name_parts[-1])
-
-    name_parts
+    if name_parts.length > 1 && GENERATIONAL_SUFFIXES.include?(name_parts[-1])
+      name_parts[...-1]
+    else
+      name_parts
+    end
   end
 
   def self.compute_semi_id(competition_year, person_name, available_per_semi = {})
--- a/spec/lib/finish_unfinished_persons_spec.rb
+++ b/spec/lib/finish_unfinished_persons_spec.rb
@@ -3,28 +3,28 @@
 require 'rails_helper'
 
 RSpec.describe FinishUnfinishedPersons, type: :module do
-  describe '.name_parts_without_suffix' do
-    {
-      'John Smith' => %w[JOHN SMITH],
-      'John Smith Jr' => %w[JOHN SMITH],
-      'John Smith Jr.' => %w[JOHN SMITH],
-      'John Smith Sr' => %w[JOHN SMITH],
-      'John Smith Jnr' => %w[JOHN SMITH],
-      'John Smith Snr' => %w[JOHN SMITH],
-      'John Smith II' => %w[JOHN SMITH],
-      'John Smith III' => %w[JOHN SMITH],
-      'John Smith IV' => %w[JOHN SMITH],
-      'Jr' => %w[JR],
-      'José García' => %w[JOSE GARCIA],
-      'Takeshi Yamada (山田武)' => %w[TAKESHI YAMADA],
-      "John O'Connor" => %w[JOHN OCONNOR],
-      'John Silver' => %w[JOHN SILVER],
-      'John Michael David Smith Jr' => %w[JOHN MICHAEL DAVID SMITH],
-    }.each do |input, expected|
-      it "returns #{expected} for '#{input}'" do
-        expect(described_class.name_parts_without_suffix(input)).to eq(expected)
-      end
+  shared_examples 'name_parts_without_suffix' do |input, expected|
+    it "returns #{expected} for '#{input}'" do
+      expect(described_class.name_parts_without_suffix(input)).to eq(expected)
     end
+  end
+
+  describe '.name_parts_without_suffix' do
+    it_behaves_like 'name_parts_without_suffix', 'John Smith', %w[JOHN SMITH]
+    it_behaves_like 'name_parts_without_suffix', 'John Smith Jr', %w[JOHN SMITH]
+    it_behaves_like 'name_parts_without_suffix', 'John Smith Jr.', %w[JOHN SMITH]
+    it_behaves_like 'name_parts_without_suffix', 'John Smith Sr', %w[JOHN SMITH]
+    it_behaves_like 'name_parts_without_suffix', 'John Smith Jnr', %w[JOHN SMITH]
+    it_behaves_like 'name_parts_without_suffix', 'John Smith Snr', %w[JOHN SMITH]
+    it_behaves_like 'name_parts_without_suffix', 'John Smith II', %w[JOHN SMITH]
+    it_behaves_like 'name_parts_without_suffix', 'John Smith III', %w[JOHN SMITH]
+    it_behaves_like 'name_parts_without_suffix', 'John Smith IV', %w[JOHN SMITH]
+    it_behaves_like 'name_parts_without_suffix', 'Jr', %w[JR]
+    it_behaves_like 'name_parts_without_suffix', 'José García', %w[JOSE GARCIA]
+    it_behaves_like 'name_parts_without_suffix', 'Takeshi Yamada (山田武)', %w[TAKESHI YAMADA]
+    it_behaves_like 'name_parts_without_suffix', "John O'Connor", %w[JOHN OCONNOR]
+    it_behaves_like 'name_parts_without_suffix', 'John Silver', %w[JOHN SILVER]
+    it_behaves_like 'name_parts_without_suffix', 'John Michael David Smith Jr', %w[JOHN MICHAEL DAVID SMITH]
   end
 
   describe '.compute_semi_id' do


app/jobs/compute_potential_duplicates.rb
```diff
@@ -8,20 +8,72 @@ def perform(job)
...
+        matches[person.id] = 100 if user_first_name.casecmp?(person_first_name) && user_last_name.casecmp?(person_last_name)
+      end
+
+      matches
+    end
+
+    def extract_first_last_name(full_name)
+      roman_name = FinishUnfinishedPersons.extract_roman_name(full_name)
+      sanitized_name = FinishUnfinishedPersons.remove_accents(roman_name)
+      name_parts = sanitized_name.gsub(/[^a-zA-Z ]/, '').split
```
gregorbg: How can there be non-Roman (a-zA-Z) characters, if you already called `extract_roman_name` above?

app/jobs/compute_potential_duplicates.rb
```diff
@@ -8,20 +8,72 @@ def perform(job)
...
+
+      return [nil, nil] if name_parts.length < 2
+
+      last_name_index = FinishUnfinishedPersons::GENERATIONAL_SUFFIXES.include?(name_parts[-1].upcase) ? -2 : -1
+      return [nil, nil] if last_name_index == -2 && name_parts.length < 3
```
gregorbg: This is quite dense to read. I think what you want to do:
1. Strip "generational suffixes" from the end of the array
2. Then take the first and last element after (1)

Ruby has `[1, 2, 3].drop_while { it < 3 }`, but unfortunately that only works from the front of the array. You would conceptually need a "drop from the back of the array. I see two options (pseudo-code):
- `name_parts.reverse.drop_while { it is generational suffix }.reverse`
- `name_parts.take_while { it is NOT generational suffix }`

Not sure which one fits better, but either way you will be able to just return `[name_parts.first, name_parts.last]` afterwards.

app/jobs/compute_potential_duplicates.rb
```diff
@@ -8,20 +8,72 @@ def perform(job)
...
       end
     end
 
     job.touch(:end_time)
     job.update!(run_status: DuplicateCheckerJobRun.run_statuses[:success])
   end
+
+  private
+
+    def find_duplicates(algorithm, user, persons_cache)
```
gregorbg, line 32: Why do you need to return the duplicates as hashes? You're taking great effort here to return the results as a hash (indexed by person ID) and then above in the `name_matching_algorithms.each_key` loop, you effectively turn the hash back into single rows.

Why not just return the "single-rows" format of `compute_similar_persons` immediately?

app/jobs/compute_potential_duplicates.rb
```diff
@@ -8,20 +8,72 @@ def perform(job)
...
+      persons_cache.each do |person|
+        next if person.dob.blank? || person.dob != user.dob
+
+        person_first_name, person_last_name = extract_first_last_name(person.name)
+        next if person_first_name.blank? || person_last_name.blank?
+
+        matches[person.id] = 100 if user_first_name.casecmp?(person_first_name) && user_last_name.casecmp?(person_last_name)
+      end
+
+      matches
```
gregorbg: As per my other comment about the hash format above, this could potentially become a simple `persons_cache.filter` statement.

app/jobs/compute_potential_duplicates.rb
```diff
@@ -8,20 +8,72 @@ def perform(job)
...
+    def find_exact_first_last_dob_matches(user, persons_cache)
+      return {} if user.dob.blank?
+
+      user_first_name, user_last_name = extract_first_last_name(user.name)
+      return {} if user_first_name.blank? || user_last_name.blank?
+
+      matches = {}
+
+      persons_cache.each do |person|
+        next if person.dob.blank? || person.dob != user.dob
```
gregorbg: Does `dob.blank?` ever happen in our DB?

gregorbg (changes requested):

app/jobs/compute_potential_duplicates.rb
danieljames-dj: Hm single rows (array) is better. My thought process was that I just need `person_id` and `score`, so since there is just two, I thought of going with map, and use the same thing in newer algorithms. One thing I was concerned with array was that it's an array of two elements, and ideally it should have been something like `pair` of C++. But I agree array is better than map for this case, so changed it.

app/jobs/compute_potential_duplicates.rb
danieljames-dj: Oh yes, done

app/jobs/compute_potential_duplicates.rb
danieljames-dj: Yes, we do have some old persons with empty DOB (for the competitors of very old competition)

app/jobs/compute_potential_duplicates.rb
danieljames-dj: My bad, it was not required. Removed it.

app/jobs/compute_potential_duplicates.rb
danieljames-dj: Oh `take_while` is really nice, used it. Thank you.

danieljames-dj: Oh actually `take_while`/`drop_while` won't work I guess, because I need to check for generational suffix only in the last name, and not on the names before it. But I got some ideas, let me try it out.

danieljames-dj: I've done it in another way now

app/jobs/compute_potential_duplicates.rb
danieljames-dj: I had to bring it back because I'm now re-using the "splitting the name to parts" and the previous way had this. I'm not sure why it's needed there, but since that code is there for a while, I'm not changing that part.

lib/finish_unfinished_persons.rb
```diff
@@ -104,15 +104,21 @@ def self.string_similarity(string_a, string_b)
...
   end
 
-  def self.compute_semi_id(competition_year, person_name, available_per_semi = {})
+  def self.name_parts_without_suffix(person_name)
     roman_name = self.extract_roman_name person_name
     sanitized_roman_name = self.remove_accents roman_name
     name_parts = sanitized_roman_name.gsub(/[^a-zA-Z ]/, '').upcase.split
 
-    last_name_index = name_parts.length > 1 && GENERATIONAL_SUFFIXES.include?(name_parts[-1]) ? -2 : -1
+    name_parts.pop if name_parts.length > 1 && GENERATIONAL_SUFFIXES.include?(name_parts[-1])
```
gregorbg: Mutating in-place is dangerous. It may work here, but I want to avoid `pop` and similar methods just as a general rule of thumb.

I was thinking about this for a moment but I cannot come up with anything "prettier" than my `.reverse.drop_while.reverse` approach -- which I admit is not really pretty.
Can you please talk to an LLM of your choice about what would be a good immutable way of implementing this `name_parts_without_suffix` method? (And don't forget to critically judge the answer LLM gives you 😉)

app/jobs/compute_potential_duplicates.rb
```diff
@@ -8,20 +8,60 @@ def perform(job)
...
+      return [] if user_first_name.blank? || user_last_name.blank?
+
+      persons_cache.filter do |person|
+        next false if person.dob.blank? || person.dob != user.dob
+
+        person_first_name, person_last_name = extract_first_last_name(person.name)
+        next false if person_first_name.blank? || person_last_name.blank?
+
+        user_first_name.casecmp?(person_first_name) && user_last_name.casecmp?(person_last_name)
+      end.map { |person| [person.id, 100] }
```
gregorbg, line 57: I think in order to stay consistent with the `jarowinkler` case above, this method should only return the persons (ie only the `filter` part) directly, and then the `find_duplicates` part is responsible for mapping it to a `person.id` and `100` score tuple.

(See how `FinishUnfinishedPersons.compute_similar_persons` also just returns a list of persons, and you map them to "score tuples" above)

spec/lib/finish_unfinished_persons_spec.rb
```diff
@@ -3,6 +3,30 @@
...
+      'Takeshi Yamada (山田武)' => %w[TAKESHI YAMADA],
+      "John O'Connor" => %w[JOHN OCONNOR],
+      'John Silver' => %w[JOHN SILVER],
+      'John Michael David Smith Jr' => %w[JOHN MICHAEL DAVID SMITH],
+    }.each do |input, expected|
+      it "returns #{expected} for '#{input}'" do
+        expect(described_class.name_parts_without_suffix(input)).to eq(expected)
+      end
+    end
+  end
```
gregorbg: Oh and by the way I love these tests! But instead of the `each`, you can consider RSpec "Shared Examples": https://rspec.info/features/3-13/rspec-core/example-groups/shared-examples/

(Not required to do it this way, but I thought it might be nice to know)

app/jobs/compute_potential_duplicates.rb
danieljames-dj: The `find_duplicates` is expected to return only person ID and score, and hence I was thinking why not just return those instead?

I understand `jarowinkler` algorithm is working differently, but I'll be more interested to bring the logic of `jarowinkler` within this class (once the `FinishUnfinishedPersons` is no longer required (ie, when `inbox_persons` is removed)) and have that method as well return the details in this same format - `person_id` and `score`.

Does it sound good or do you think retuning `person` is better?

spec/lib/finish_unfinished_persons_spec.rb
danieljames-dj: Oh okay this is nice, I've implemented it. I'm not sure whether by "Not required to do it this way" you meant it's optional or not to do it. Anyway I have it in my commit history, let me know if I need to revert it. `shared_examples` is definitely interesting but I feel both are equally good.

lib/finish_unfinished_persons.rb
danieljames-dj: Okay done

gregorbg (approved):

PR #13401, opened 2026-02-09
Action Cable dev setup

FinnIckler:


config/cable.yml
```diff
@@ -1,10 +1,11 @@
...
+  adapter: redis
+  url: <%= EnvConfig.ACTIONCABLE_REDIS_URL %>
 
 test:
   adapter: test
 
 production:
   adapter: redis
-  url: <%= ENV.fetch("REDIS_URL") { "redis://localhost:6379/1" } %>
+  url: <%= EnvConfig.ACTIONCABLE_REDIS_URL %>
```
gregorbg, line 10: Hm. In its current setup, this will yield an empty string. Not sure if prod explodes due to this, even if we're not _actually_ using AC on there right now

gregorbg: (compare to how it previously had some dummy value as fallback)

gregorbg: Not sure though what a good "temporary dummy value" is anyways...

docker-compose.yml
```diff
@@ -21,6 +21,7 @@ x-app: &app
 x-app-environment: &appEnv
   <<: *env
   SIDEKIQ_REDIS_URL: redis://wca_redis:6379/1
+  ACTIONCABLE_REDIS_URL: redis://wca_redis:6379/2
```
gregorbg, line 24: Not sure if all Rails apps need to know about this. Potentially specifying it under `wca_on_rails` service is enough, similar to `CACHE_REDIS_URL`

docker-compose.yml
gregorbg, line 1: While we're at this, could you expose Redis's port to the outside host machine please?
Literally every single one of our containers that has _any_ port mapping (MySQL, Mongo, Rails, even bloody Webpacker/Shakapacker) uses `ports` and binds its port to the host.

Only Redis does not. And I feel that is (a) very inconsistent and (b) also harmful in the context of WCA Live because if you want your host machine's Ruby environment `bin/rails c` to quickly fire a debug message into the broadcast channel, it will not work.

config/cable.yml
FinnIckler: well it currently explodes as Duncan noticed when he tried to create LiveResults on staging. I am making the second PR right now with the the changes to the LiveResult job. That should be merged first so it doesn't explode.

docker-compose.yml
FinnIckler: they do. Sidekiq needs to know

PR #13403, opened 2026-02-09
Move Diff Transmition to UpdateLiveResultJob

FinnIckler:


app/models/live_attempt.rb
```diff
@@ -46,12 +46,12 @@ def self.attempts_changed?(before_attempts, after_attempts)
...
   end
 
-  def update_with_history_entry(value, acting_user)
+  def update_with_history_entry(value, acting_user_id)
     self.update(value: value)
     self.live_attempt_history_entries.create(
       value: value,
       entered_at: Time.now.utc,
-      entered_by: acting_user,
+      entered_by: acting_user_id,
```
FinnIckler, line 54: this column is actually just entered_by and not entered_by_id. I think our idea behind this was, that this can also be other entities and not just users in the future

spec/lib/live/state_diff_spec.rb
```diff
@@ -9,16 +9,6 @@
   context "Round Diff" do
     let(:round) { create(:round, event_id: "333", number: 1, competition: competition) }
 
-    it 'broadcasts diff to ActionCable' do
```
FinnIckler, line 12: moved this test to the add spec now as it is triggered there

app/controllers/api/v1/live/live_controller.rb
```diff
@@ -1,7 +1,36 @@
...
+    round = Round.find_by_wcif_id!(round_id, competition_id)
+
+    # TODO: add require_managed! from round admin PR
+    require_user
+
+    # We create empty results when a round is open
+    live_result = round.live_results.find_by(registration_id: registration_id)
+    result_exists = live_result.present?
+
+    unless result_exists
```
gregorbg: Why don't you do the `.present?` check inline? That's a common pattern in lots of our controllers

app/jobs/update_live_result_job.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+class UpdateLiveResultJob < ApplicationJob
+  self.queue_adapter = :shoryuken if Live::Config.sqs_queued?
+  queue_as EnvConfig.LIVE_QUEUE if Live::Config.sqs_queued?
+
+  def perform(results, live_result_id, entered_by_id)
```
gregorbg: I would put the live result (ID) first. It's the entity that everything this job depends upon, and _then_ you can list the values/attributes that are being modified on this live result.

gregorbg: On that note, ActiveJob can serialize any ActiveModel class so just passing in the `LiveResult` should(?) be fine

app/jobs/update_live_result_job.rb
```diff
@@ -0,0 +1,41 @@
...
+          previous_attempt
+        else
+          previous_attempt.update_with_history_entry(r[:value], entered_by_id)
+        end
+      else
+        LiveAttempt.build_with_history_entry(r[:value], r[:attempt_number], entered_by_id)
+      end
+    end
+
+    round = Round.find(live_result.round_id)
```
gregorbg: Doesn't `LiveResult` have a `belongs_to` association that you can go through?

app/models/live_attempt.rb
gregorbg: Interesting... Even if we want to have it be "other things", the Rails polymorphic convention is to have it be `_id` and `_type` suffixes

app/models/live_result.rb
```diff
@@ -111,16 +105,4 @@ def self.compute_diff(before_result, after_result)
     # Only return if there are actual changes
     diff if diff.except("registration_id").present?
   end
-
-  private
-
-    def trigger_recompute_and_notify
-      before_state = round.to_live_state
-
-      round.recompute_live_columns(skip_advancing: locked?)
```
gregorbg, line 120: Could we leave at least the recomputing part as a hook?
1. It does not depend on any external infrastructure (the way that broadcasting does)
2. It also makes sense that when any results data is changed, the advancement **must** (from a business logic perspective) be recomputed
3. [My personally strongest argument] It saves you from a gazillion `recompute_live_columns` calls in the tests

FinnIckler: true. I will revert that

app/models/live_attempt.rb
FinnIckler: no, we wanted to go even further. Non Rails entities like entered by Electronic Scoretaking Software xyz

app/models/live_result.rb
```diff
@@ -115,12 +115,6 @@ def self.compute_diff(before_result, after_result)
   private
 
     def trigger_recompute_and_notify
```
gregorbg: The name of this hook does not reflect what it does (anymore)

gregorbg (changes requested):

PR #13405, opened 2026-02-09
Improvise import registrations via CSV with preview

danieljames-dj:
Implement 'preview' for import registrations page

Screenshot 2026-02-26 at 06 56 31


app/controllers/registrations_controller.rb
```diff
@@ -71,10 +76,23 @@ class RegistrationsController < ApplicationController
...
+        wcaId: row[:wca_id]&.upcase,
+        countryIso2: Country.c_find(row[:country]).iso2,
+        gender: row[:gender],
+        birthdate: row[:birth_date],
+        email: row[:email]&.downcase,
+        registration: {
+          eventIds: event_ids,
+          status: "accepted",
+          isCompeting: true,
+          registeredAt: Time.now.utc,
```
gregorbg: This means that the individual rows will each have a few microseconds difference due to the `map` loop. I would recommend declaring one `import_time = Time.now.utc` variable above, and reusing that in every loop.

app/controllers/registrations_controller.rb
```diff
@@ -21,22 +21,52 @@ class RegistrationsController < ApplicationController
...
+    registration_rows = params.require(:registrations)
+
+    return render status: :unprocessable_content, json: { error: "Expected array of registrations" } unless registration_rows.is_a?(Array)
+
+    errors = [
+      validate_registrations(registration_rows, @competition),
+      competitor_limit_error(@competition, registration_rows.length),
+    ].compact.flatten
+
+    return render status: :unprocessable_content, json: { error: errors.compact.join(", ") } if errors.any?
```
gregorbg: Why do you use `compact` here again? `errors` is defined as `[...].compact.flatten` already.

app/webpacker/components/ImportRegistrations/RegistrationPreview.jsx
```diff
@@ -0,0 +1,99 @@
...
+}) {
+  const {
+    mutate: importMutate, isPending, isError, error,
+  } = useMutation({
+    mutationFn: importRegistrations,
+    onSuccess: onImportSuccess,
+  });
+
+  const tableRegistrations = useMemo(
+    () => (registrations || []).map(transformRegistration),
```
gregorbg: When can `registrations` be undefined? Perhaps just declare `registrations = []` in the component function header above (thus making sure it always has a value) or think about why the value that you pass in could ever be undefined.

app/webpacker/components/ImportRegistrations/RegistrationPreview.jsx
```diff
@@ -0,0 +1,99 @@
...
+    <Modal
+      open={!!registrations}
+      onClose={onClose}
+      closeOnEscape
+      size="fullscreen"
+    >
+      <Modal.Header>Preview Registration Data</Modal.Header>
+      <Modal.Content scrolling>
+        {isError && <Errored error={error} />}
+        <RegistrationsAdministrationTable
```
gregorbg, line 76: Does it make sense to even show the `RegAdminTable` if there is an error? My review perspective is that it should be `isError ? <Errored /> : <RegAdminTable />`, so please tell me which use-case I'm overlooking :D

app/webpacker/components/Registrations/Administration/AdministrationTableRow.jsx
```diff
@@ -119,29 +120,38 @@ export default function TableRow({
...
+              </Table.Cell>
+            )}
 
             <Table.Cell>
-              {wcaId ? (
+              {wcaId && (
                 <a href={personUrl(wcaId)}>{wcaId}</a>
-              ) : (
+              )}
+              {!wcaId && userId && (
```
gregorbg, line 179: Why did you split up the ternary condition? Either a WCA ID or a user ID should always be present, no?

gregorbg: Or otherwise, what is being displayed if neither one is present?

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationTableFooter.jsx
```diff
@@ -45,9 +46,17 @@ export default function RegistrationAdministrationTableFooter({
     }),
   );
 
+  const getColSpan = () => {
+    if (isReadOnly) return 2;
+    if (withPosition) return 5;
+    return 4;
+  };
+
+  const firstColSpan = getColSpan();
```
gregorbg: Why are you storing this in a separate variable if it's only used once? Just do `colSpan={getColSpan()}`

app/controllers/registrations_controller.rb
```diff
@@ -218,6 +273,10 @@ def do_import
     render status: :unprocessable_content, json: { error: e.to_s }
   end
 
+  def validate_and_convert_registrations
+    render status: :ok, json: @registration_rows
+  end
```
gregorbg: This can be very confusing for someone who reads the code for the first time: Where the heck is the validation happening?

I know the answer, but others might not. You have two options:
1. (preferred!) Move the `before_action` directly above the `def` of this method, so that it's physically close in the code and the reader notices "Ah there's a check running right before reaching this controller method"
2. If (1) is not possible, at least add a code comment referencing the `before_action` checks that run

app/controllers/registrations_controller.rb
```diff
@@ -202,7 +256,8 @@ def do_import
         registration.assign_attributes(competing_status: Registrations::Helper::STATUS_ACCEPTED) unless registration.accepted?
         registration.registration_competition_events = []
         registration_row[:event_ids].each do |event_id|
-          registration.registration_competition_events.build(competition_event_id: event_id)
+          competition_event = @competition.competition_events.find { |ce| ce.event_id == event_id }
+          registration.registration_competition_events.build(competition_event_id: competition_event.id)
         end
```
gregorbg: I think you can be way more efficient here. I know a fully coded-out solution, but as an exercise I will only give you some hints 😉 

Rails `.build` method can take an array of attributes. So instead of `[{}, {}, {}].each { |attr| reg.something.build(**attr) }` you can actually do `reg.something.build([{}, {}, {}])` directly and it will build three entities "under the hood" for you.

So what you need is a way to query all `competition_event_id`s at once, given a batch of `event_id`s. Try filling in the blanks: `all_competition_event_ids = @competition.competition_events.where(...FILL ME...).ids`

app/controllers/registrations_controller.rb
```diff
@@ -105,29 +148,40 @@ class RegistrationsController < ApplicationController
...
+      event_column_errors,
+      validate_dob_formats(csv_rows.pluck(:birth_date)),
+      validate_no_duplicates(csv_rows.pluck(:email), 'email_duplicates', :emails),
+      validate_no_duplicates(csv_rows.pluck(:wca_id).compact_blank, 'wca_id_duplicates', :wca_ids),
+    ].flatten.compact
   end
 
-  private def column_check(csv_rows, column_name, error_key, i18n_keyword)
-    column_values = csv_rows.pluck(column_name)
+  private def validate_registrations(registration_rows, competition)
```
gregorbg, line 181: Oh no, where did our nice `column_check` method go? I appreciate the additional complexity, but I still think there's a decent amount of code reuse. Can you look into refactoring this a little bit please?

If it doesn't work out so well, please explain your thinking in two-three sentences about why you think this is the best, cleanest code you can come up with for your use case.

app/controllers/registrations_controller.rb
```diff
@@ -21,22 +21,52 @@ class RegistrationsController < ApplicationController
...
   end
 
-  before_action :competition_must_not_be_using_wca_registration!, only: %i[import do_import]
+  before_action :competition_must_not_be_using_wca_registration!, only: %i[import do_import validate_and_convert_registrations]
   private def competition_must_not_be_using_wca_registration!
     redirect_to competition_path(competition_from_params) if competition_from_params.use_wca_registration?
   end
 
   before_action :validate_import_registration, only: %i[do_import]
   private def validate_import_registration
```
gregorbg, line 38: I am a little bit confused between this method and the new `validate_and_parse_registration_data` method below.

Can you please clarify when each method is being called, and why they use "similar but not quite the same" parsing routines?

gregorbg (changes requested):

app/webpacker/components/RegistrationsV2/RegistrationAdministration/RegistrationAdministrationTableFooter.jsx
danieljames-dj: Okay chnaged

app/webpacker/components/Registrations/Administration/AdministrationTableRow.jsx
danieljames-dj: There can be cases where the imported registration so far doesn't have neither WCA ID not user ID. This column was allowing to view their corresponding person/user, and if none are available, I think we can leave it blank.

I've now updated the screenshot in description, hope that explains.

app/webpacker/components/ImportRegistrations/RegistrationPreview.jsx
danieljames-dj: Image

As in the screenshot, if the error is shown along with the table, it might be helpful for the user to check if there is anything is wrong in the table.

(I assume you misunderstood this as "fetching error" instead of "importing error")

app/webpacker/components/ImportRegistrations/RegistrationPreview.jsx
danieljames-dj: Oh actually I was actually rendering this component even before it's uploaded - my mistake. I've now made it to render only if registration if available.

(There won't be edge case where they upload empty registration, it will throw error anyway)

app/controllers/registrations_controller.rb
danieljames-dj: `validate_import_registration` - While actual import
`validate_and_parse_registration_data` - Validation for preview

Will it make sense to rename second one to `validate_and_parse_registration_preview` or do you have better name suggestion here?

app/controllers/registrations_controller.rb
danieljames-dj: Okay went ahead with 1.

app/controllers/registrations_controller.rb
danieljames-dj: Oh yes, I see that it's was there already. I removed the other occurrence as well.

app/controllers/registrations_controller.rb
danieljames-dj: Done

app/controllers/registrations_controller.rb
danieljames-dj: The `column_check` method was doing `csv_rows.pluck` and creating i18n message. I can indeed re-use `column_check` within `validate_no_duplicates`, but it won't let me re-use `validate_no_duplicates` within `validate_registrations` method. Same for `validate_dob_formats`. So I split `column_check` to `validate_no_duplicates` and `validate_dob_formats`.

app/controllers/registrations_controller.rb
danieljames-dj: Okay I got the answer. I also see that same was used in one more place in the file, so moved them to an independent private method.

Thank you.

app/controllers/registrations_controller.rb
```diff
@@ -21,27 +21,61 @@ class RegistrationsController < ApplicationController
...
+      {
+        name: row[:name],
+        wca_id: row[:wcaId],
+        country: country.id,
+        gender: row[:gender],
+        birth_date: row[:birthdate],
+        email: row[:email],
+        event_ids: row.dig(:registration, :eventIds) || [],
+      }
+    end
```
gregorbg: Here, you're using a format with snake_case, like for example `wca_id`. In another method below. you're also assigning to a variable called `@registration_rows`, but you use a different format. Why?

app/controllers/registrations_controller.rb
gregorbg: Mh, yeah I like your suggestion with the `_preview`!

app/controllers/registrations_controller.rb
```diff
@@ -67,14 +101,29 @@ class RegistrationsController < ApplicationController
...
+        gender: row[:gender],
+        birthdate: row[:birth_date],
+        email: row[:email]&.downcase,
+        registration: {
+          eventIds: event_ids,
+          status: "accepted",
+          isCompeting: true,
+          registeredAt: import_time,
+        },
+      }
```
gregorbg: The result here is later assigned to a variable called `@registration_rows`, and it uses camelCase here, like `wcaId` or `countryIso2` for example.

This is inconsistent with my other comment above, where another variable called `@registration_rows` suddenly uses snake_case like `wca_id` or `birth_date`. Please help me understand this inconsistency.

gregorbg (changes requested):
Cool, almost there!

app/controllers/registrations_controller.rb
danieljames-dj: Okay changed.

app/controllers/registrations_controller.rb
danieljames-dj: This one is for importing registration. We will convert to snake_case because existing code is using snake_case. In the other place I changed to camelCase because that API is to convert the CSV to WCIF format, and WCIF format is camelCase, also since it's passed to frontend, camelCase looks bit more clear.

app/controllers/registrations_controller.rb
danieljames-dj: This one is converting to WCIF.

gregorbg (commented):
Thank you for your explanations!
They make total sense, so now you have to think about refactoring your code, so that it conveys this meaning better.

Right now, I see two instances of code stuffing things into a `@registration_rows` variable. And despite the identical name, these are vastly different usages. So you should come up with better variable names.

Also, the `parse_csv_file` does not say what it "parses into". But the target format of WCIF is actually pretty well defined, so you should think of a short and sweet name that conveys it better there as well 😄

app/models/registration.rb
```diff
@@ -264,6 +262,22 @@ def add_history_entry(changes, actor_type, actor_id, action, timestamp = Time.no
...
+    self.registration_competition_events = []
+
+    event_ids = registration_data.dig(:registration, :eventIds) || []
+    self.registration_competition_events.build(
+      competition.competition_events.where(event_id: event_ids).ids.map { |id| { competition_event_id: id } },
+    )
+
+    save!
+    add_history_entry({ event_ids: self.event_ids }, "user", creator.id, source)
+  end
```
gregorbg, line 297: We already have `lib/registrations/registration_checker.rb#apply_payload` for this, please use it.

See also `lib/registrations/lanes/competing.rb` for examples of how it should be used, and how to store all relevant changes as history_entry

app/controllers/registrations_controller.rb
```diff
@@ -237,21 +277,25 @@ def do_add
...
+        wcaId: params[:registration_data][:wca_id],
+        countryIso2: Country.c_find(params[:registration_data][:country]).iso2,
+        gender: params[:registration_data][:gender],
+        birthdate: params[:registration_data][:birth_date],
+        email: params[:registration_data][:email],
+        comments: params[:registration_data][:comments],
+        registration: {
+          eventIds: params[:registration_data][:event_ids] || [],
+        },
+      }
```
gregorbg: I am not too happy about this reshaping of data. Why does `params[:registration_data]` not come in a "good" format already?

gregorbg (changes requested):

app/controllers/registrations_controller.rb
danieljames-dj: I'm trying to reshape this to the WCIF format. Regarding why `params[:registration_data]` not coming in this format - I don't know, but how about taking that up in a later PR because there are lots of places I have to make that change?

app/models/registration.rb
danieljames-dj: Done

gregorbg: Can you take a look at `lib/registrations/lanes/competing.rb` method `def self.update!` and see whether that works for you?

The more code we can reuse, the better.

danieljames-dj: Yes I can use that, I've updated. But I had to make some changes in `update!`, which I am not sure if it's worth. I'm nearly 60-40 in this change (60% towards the change to update!, but 40% I think it's not necessary). Let me know if I need to revert.

app/controllers/registrations_controller.rb
```diff
@@ -21,28 +21,48 @@ class RegistrationsController < ApplicationController
...
     redirect_to competition_path(competition_from_params) if competition_from_params.use_wca_registration?
   end
 
   before_action :validate_import_registration, only: %i[do_import]
   private def validate_import_registration
     @competition = competition_from_params
+
+    @registrations = params.require(:registrations)
+
+    return render status: :unprocessable_content, json: { error: "Expected array of registrations" } unless @registrations.is_a?(Array)
```
gregorbg: Why don't you just use something like `Array.wrap(@registrations)` to make sure that it is an array?

gregorbg: You can also play around with `params.expect`, which is new in Rails 8 these days:
https://blog.saeloun.com/2024/12/10/rails-8-adds-parametersexpect-to-safely-filter-and-require-params/

app/controllers/registrations_controller.rb
gregorbg: Then please move this into a `csv_to_wcif` or similar kind of helper function :)

app/controllers/registrations_controller.rb
gregorbg: Well you already have some kind of "convert data into WCIF" helper above. So I think you should try to reuse it at least within the controller.

My point is that right now, you're even copy-pasting keys like `"countryIso2"` or `"birthdate"` all around the place. At the very least, these keys should be moved to one shared helper method, to avoid typos and similar.

lib/registrations/lanes/competing.rb
```diff
@@ -48,7 +48,7 @@ def self.update!(update_params, registration, acting_entity_id)
           #   from when we called `event_ids` in the `apply_payload` method above.
           registration.events.reset
 
-          history_action_type = Registrations::Helper.action_type(update_params, registration.user_id, acting_entity_id)
+          history_action_type = history_action || Registrations::Helper.action_type(update_params, registration.user_id, acting_entity_id)
```
gregorbg: Have you tried updating the `Helper.action_type` method, instead of "simply" passing in an override value?

gregorbg (commented):
Still a few bits and pieces regarding the backend, but I think we can carry this over the finish line!

app/controllers/registrations_controller.rb
danieljames-dj: Makes sense, changed

lib/registrations/lanes/competing.rb
danieljames-dj: I don't know if I understood it correctly, but can you please check the latest commit if this is what you meant?

app/controllers/registrations_controller.rb
danieljames-dj: Done

app/controllers/registrations_controller.rb
```diff
@@ -67,17 +103,55 @@ class RegistrationsController < ApplicationController
...
+      country: row[:country],
+      gender: row[:gender],
+      birth_date: row[:birth_date],
+      email: row[:email],
+      event_ids: event_ids,
+      status: "accepted",
+      is_competing: true,
+      registered_at: import_time,
+    )
+  end
```
gregorbg: Why do you move this into a separate `csv_to_wcif` method, instead of just calling `build_wcif_data` directly?

gregorbg: Also, you can just use `build_wcif_data(**row, event_ids: event_ids, status: "accepted", is_competing: true, registered_at: import_time)`

app/controllers/registrations_controller.rb
```diff
@@ -237,21 +316,23 @@ def do_add
...
+      registration_data = build_wcif_data(
+        name: params[:registration_data][:name],
+        wca_id: params[:registration_data][:wca_id],
+        country: params[:registration_data][:country],
+        gender: params[:registration_data][:gender],
+        birth_date: params[:registration_data][:birth_date],
+        email: params[:registration_data][:email],
+        comments: params[:registration_data][:comments],
+        event_ids: params[:registration_data][:event_ids] || [],
+      )
```
gregorbg: This looks a lot like `build_wcif_data(**params[:registration_data])`

gregorbg (changes requested):

app/controllers/registrations_controller.rb
danieljames-dj: Oh yes, I started with `csv_to_wcif` alone, then for the other use-case created `build_wcif_data`. Thinking again, just `build_wcif_data` is enough, changed it now.

app/controllers/registrations_controller.rb
danieljames-dj: Ah good catch, I tried, it was not straightforward, but managed to do it in a cleaner way.

app/controllers/registrations_controller.rb
```diff
@@ -67,17 +103,51 @@ class RegistrationsController < ApplicationController
...
+        wca_id: row[:wca_id],
+        country: row[:country],
+        gender: row[:gender],
+        birth_date: row[:birth_date],
+        email: row[:email],
+        event_ids: event_ids,
+        status: "accepted",
+        is_competing: true,
+        registered_at: import_time,
+      )
```
gregorbg, line 123: Just curious, have you tried...

```suggestion
      build_wcif_data(
        **row,
        event_ids: event_ids,
        status: "accepted",
        is_competing: true,
        registered_at: import_time,
      )
```

danieljames-dj: But row has other columns as well - like 222, 333, etc and it causes ArgumentError: error. Anyway I got another way to avoid this, and I've made it similar to that way now.

gregorbg: If you are concerned about "dangling extra arguments", you can also consider `wcif_row = row.slice(:name, :wca_id, ...etc...)` to filter out only the keys that you want, Then, you can still spread `**wcif_row` without the additional `**_` ignore flag in the method signature

app/controllers/registrations_controller.rb
danieljames-dj: I missed to reply this, this was done.

app/controllers/registrations_controller.rb
danieljames-dj: Done

gregorbg (approved):

PR #13407, opened 2026-02-10
Paginate Banned Competitors

Brandon-Reichman:
fixes: https://github.com/thewca/worldcubeassociation.org/issues/11710

The hard part was getting over 100 users banned in my local database :sweat_smile: 

I mostly just copied what Parnav did in https://github.com/thewca/worldcubeassociation.org/pull/11711. However, as discussed in that PR, the api call is only returning 100 rows and the front end is only paginating over that. Daniel mentioned that you can pass the page number like `/api/v0/user_roles?page=3` which works, however the total number of rows there could be is unknown and page options can not be accurately displayed on the front end.

This endpoint seems to be used all over the application so I did not want to modify the return structure by adding the total number of results to `users_roles_controller@index`

The simplest solution seems to be increasing the perPage limit on these specific calls, which are defaulted to 100.
This is what I did by simply adding 500 to the banned competitor requests...
```
apiV0Urls.userRoles.list({
    isActive: true,
    groupType: groupTypes.banned_competitors,
}, 'startDate', 500)
```
This seems like a band-aid solution as the problem could arise again if this limit is passed, but the only other solution I can think of is creating a separate endpoint for banned competitors that would include the total result count and can be properly paginated. Otherwise, 500 should _hopefully_ be sufficient enough for a while.

Please let me know if this is not an acceptable solution for this and if you have a different suggestion.

### DIFF
--- a/app/webpacker/components/Panel/pages/BannedCompetitorsPage/index.jsx
+++ b/app/webpacker/components/Panel/pages/BannedCompetitorsPage/index.jsx
@@ -8,26 +8,33 @@ import Errored from '../../../Requests/Errored';
 import BannedCompetitors from './BannedCompetitors';
 import useLoggedInUserPermissions from '../../../../lib/hooks/useLoggedInUserPermissions';
 import BannedCompetitorForm from './BannedCompetitorForm';
+import usePagination from '../../../../lib/hooks/usePagination';
 
 export default function BannedCompetitorsPage() {
+  const bannedPagination = usePagination(50);
+  const pastBannedPagination = usePagination(50);
   const {
     data: bannedCompetitorRoles,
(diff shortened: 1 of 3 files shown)


app/webpacker/components/Panel/pages/BannedCompetitorsPage/index.jsx
gregorbg, line 1: Our `useLoadedData` hook returns a `headers` object. These response headers contain information on the "total number of entries", so you can indeed properly paginate.

Feel free to look at `webpacker/components/IncidentsLog/index.jsx` as an example of how to access this information.

gregorbg (approved):

PR #13408, opened 2026-02-10
Next WCA Live PDF view 

FinnIckler:
Ported the PDF template from WCA Live. While I did this I noticed that we didn't actually include the country as a table column yet so I had to also introduce that.

This is how the pdf looks:

image

And here is an example file:
[results-8.pdf](https://github.com/user-attachments/files/25214677/results-8.pdf)


FinnIckler:
image
Managed to match the 24 rows per page from wca live pretty easily

PR #13410, opened 2026-02-11
WCA Live Table Fixes

FinnIckler:
While building the PDF and Presentation views I noticed a lot of things still missing/wrong on the Table:
1. Now orders the results correctly on the client with ties
2. Adds the country column
3. Correctly shows the right stat columns depending on the format
4. Pads the attempts so there are always the right amount of Table Cells in the Table (will be important for projected averages and just good HTML practice in general)

### DIFF
--- a/next-frontend/src/lib/live/statColumnsForFormat.ts
+++ b/next-frontend/src/lib/live/statColumnsForFormat.ts
@@ -1,19 +1,19 @@
 import { Format } from "@/lib/wca/data/formats";
 
-const statMap = new Map(
-  (
-    [
-      {
-        name: "average",
-        recordTagField: "average_record_tag",
-        field: "average",
-      },
-      { name: "single", recordTagField: "single_record_tag", field: "best" },
-    ] as const
-  ).map((stat) => [stat.name, stat]),
-);
+const statMap = {
+  average: {
+    name: "average",
+    recordTagField: "average_record_tag",
+    field: "average",
+  },
+  single: {
+    name: "single",
+    recordTagField: "single_record_tag",
+    field: "best",
+  },
+} as const;
 
 export const statColumnsForFormat = (format: Format) =>
   [format.sort_by, format.sort_by_second]
     .filter(Boolean)
-    .map((s) => statMap.get(s as "average" | "single")!);
+    .map((s) => statMap[s as keyof typeof statMap]);


app/models/round.rb
```diff
@@ -435,7 +435,7 @@ def to_live_json(only_podiums: false)
     {
       **self.to_wcif,
       "round_id" => id,
-      "competitors" => live_competitors.includes(:user).map { it.as_json({ methods: %i[user_name], only: %i[id user_id registrant_id] }) },
+      "competitors" => live_competitors.includes(:user).map { it.as_json({ methods: %i[name country_iso2], only: %i[id user_id registrant_id] }) },
```
gregorbg, line 438: I think you had introduced the `user_name` alias before, just for this WCA Live serialization. Now I realize there had already been an alias, so you can delete the new prefixed one...?

next-frontend/src/lib/live/statColumnsForFormat.ts
```diff
@@ -0,0 +1,17 @@
...
+};
+
+const statMap: Stat[] = [
+  { name: "average", recordTagField: "average_record_tag", field: "average" },
+  { name: "single", recordTagField: "single_record_tag", field: "best" },
+];
+
+export const statColumnsForFormat = (format: Format) =>
+  [format.sort_by, format.sort_by_second]
+    .filter((s) => s)
```
gregorbg:
```suggestion
    .filter(Boolean)
```

next-frontend/src/lib/live/statColumnsForFormat.ts
```diff
@@ -0,0 +1,17 @@
...
+type Stat = {
+  name: "average" | "single";
+  recordTagField: "average_record_tag" | "single_record_tag";
+  field: "average" | "best";
+};
+
+const statMap: Stat[] = [
+  { name: "average", recordTagField: "average_record_tag", field: "average" },
+  { name: "single", recordTagField: "single_record_tag", field: "best" },
+];
```
gregorbg: If you're already calling it `statMap`, why don't you actually turn it into a map or object?
That would also get rid of the annoying `find!` below...

gregorbg: See also `{ ... } as const` notation in TypeScript for narrow type definitions: https://www.totaltypescript.com/concepts/as-const

next-frontend/src/lib/live/orderResults.ts
```diff
@@ -0,0 +1,62 @@
...
+
+      if (!isTied) {
+        currentRank = index + 1;
+      }
+
+      result.global_pos = currentRank;
+    }
+  });
+
+  return sortedResults;
```
gregorbg: What is the purpose of this block? I can see you are recomputing `globalPos`, but why? Are you not happy with what the backend gives you?

gregorbg: Regardless, strong disagreement on `let` and on mutability. If it is absolutely necessary to recompute, then you should do the recomputation in an immutable way, ie loop and create new entries with the recomputed global pos rather than mutating the existing entries

app/models/round.rb
FinnIckler: ah yeah, I deleted it in the original PR but then forgot

next-frontend/src/lib/live/orderResults.ts
FinnIckler: We gotta recompute because we will not be sending over `global_pos` over the websockets as per optimizations that we talked about

FinnIckler: but I will make the immutability changes

next-frontend/src/lib/live/statColumnsForFormat.ts
FinnIckler: sadly the as const trick doesn't work on importing the json, so it isn't as pretty

next-frontend/src/lib/live/orderResults.ts
```diff
@@ -0,0 +1,61 @@
...
+    const isTied =
+      result[rankBy] === prev[rankBy] &&
+      (!secondaryRankBy || result[secondaryRankBy] === prev[secondaryRankBy]);
+
+    if (!isTied) {
+      currentRank = index + 1;
+    }
+
+    return { ...result, global_pos: currentRank };
+  });
```
gregorbg: Still has a `let` :P

next-frontend/src/lib/live/statColumnsForFormat.ts
```diff
@@ -0,0 +1,19 @@
...
+    [
+      {
+        name: "average",
+        recordTagField: "average_record_tag",
+        field: "average",
+      },
+      { name: "single", recordTagField: "single_record_tag", field: "best" },
+    ] as const
+  ).map((stat) => [stat.name, stat]),
+);
```
gregorbg: Why do you not instantiate it as...
```
const statMap = {
  single: {
    name: ...,
  },
  average: {
    name: ...,
  }
} as const;
```

FinnIckler: oh I thought you mean a literal js Map

Issue #7843, opened 2023-04-19
Require user to type the ID of the competition they want to delete


dunkOnIT:
Given the amount of data that can be in a comp page, and the danger of mindlessly clicking through confirmations, it might be worth requiring the user to explicitly confirm in a popup the name of the competition they're deleting when they click "Delete" in the competition form. 

_I really wish I hadn't discovered a need for this issue_

PR #13411, opened 2026-02-11
Asking deletgates to enter comp-id before deleting competition

pranav-027:
fixes: https://github.com/thewca/worldcubeassociation.org/issues/7843

### DIFF
--- a/app/webpacker/lib/providers/ConfirmProvider.js
+++ b/app/webpacker/lib/providers/ConfirmProvider.js
@@ -69,7 +69,6 @@ export default function ConfirmProvider({ children }) {
     }
   }, [resolve, handleClose, options.requireInput, inputValue]);
 
-
   const isOpen = resolveReject.length === 2;
 
   return (


app/webpacker/lib/providers/ConfirmProvider.js
```diff
@@ -4,19 +4,24 @@
...
   content: 'Are you sure? This action cannot be undone.',
   confirmButton: 'Yes',
   cancelButton: 'No',
+  requireInput: null, // If set, user must type this exact string to confirm
 };
 
 export default function ConfirmProvider({ children }) {
   const [options, setOptions] = useState(DefaultOptions);
   const [resolveReject, setResolveReject] = useState([]);
+  const [inputValue, setInputValue] = useState('');
```
gregorbg: We have our own `useInputState` hook which is configured exactly for SemUI `Input` elements. It returns the same kind of "getter and setter" tuple that a normal `useState` would, but you can pass the setter directly to the `<Input>`'s `onChange` and it "just works" 🪄

app/webpacker/lib/providers/ConfirmProvider.js
```diff
@@ -4,19 +4,24 @@
...
+  const [inputError, setInputError] = useState(false);
```
gregorbg: I don't think this has to be an independent state. It can be a memo derived from "is the `inputValue` identical to the `requireInput`?"

app/webpacker/lib/providers/ConfirmProvider.js
```diff
@@ -43,19 +52,78 @@ export default function ConfirmProvider({ children }) {
...
+
+  const isOpen = resolveReject.length === 2;
+
+  // If requireInput is set, use a custom Modal with input field
+  if (options.requireInput) {
+    return (
+      <>
+        <ConfirmationContext.Provider value={confirm}>
+          {children}
+        </ConfirmationContext.Provider>
```
gregorbg: This part of the return value is exactly identical to the "old" one with just the classic `<Confirm>`. So I suggest you merge them into one "big" return with a ternary `{options.requireInput ? () : ()}` block to determine the body that should get rendered after the ContextProvider.

gregorbg (changes requested):

app/webpacker/lib/providers/ConfirmProvider.js
pranav-027: Done

app/webpacker/lib/providers/ConfirmProvider.js
pranav-027: Done

PR #13412, opened 2026-02-11
Extend newcomers check page to Delegates

danieljames-dj:


app/models/user.rb
```diff
@@ -948,7 +948,7 @@ def can_submit_competition_results?(competition)
   end
 
   def can_check_newcomers_data?
-    can_admin_results?
+    can_admin_results? || any_kind_of_delegate?
```
gregorbg: How does this check for Delegates associated with a competition? The name "any" implies that French Delegates can check newcomers for Brazilian competitions, which is not desirable. You should only be able to check Newcomers for competitions that you're actually a listed Delegate for. (See also my other general remark about tests)

gregorbg (changes requested):

app/models/user.rb
danieljames-dj: Okay, I was thinking that since delegates anyway have access to all the user's personal details, they can see other competition's newcomers page as well if needed. But when I think again, it's risky as if somebody plays around by running the checker for different competitions, then it won't be good.

I've made the changes, also updated tests.

Issue #7063, opened 2022-07-25
Make contact form's email have a hyperlink to the competition


Jambrose777:
From @c-goodyear:

> Would it be possible to get the competition name part of these emails made into hyperlinks to the competition page itself? It would be a lot faster to action requests like this if there was a clickable link within the email, especially for Delegates with multiple competitions with registrations open!

**Describe the solution you'd like**
In the email sent from the contact form to a competition's organizer, include a link to the competition.

Issue #7930, opened 2023-05-24
Add link to registration in website comment for competition


simonkellly:
**Is your feature request related to a problem? Please describe.**
Quite frequently, we get emails through the contact form for competitions. These are almost always people looking to delete their registration, or adjust the events they are signed up for. When actioning these requests, one has to go and find the competition page, go to the registrations, find that persons registration, and then finally make the change that they are looking to make.

**Describe the solution you'd like**
It would be ideal to have a link to the person's current registration if it exists for the competition they are making an enquiry about, similar to the one which exists when getting notified about a person's registration

RamanTheDeveloper:
Hi, has this issue been assigned to someone?

dunkOnIT:
Hi @RamanTheDeveloper, it hasn't - would you like to work on it?

RamanTheDeveloper:
Hey, yes please.

dunkOnIT:
Done! Let me know if you need any help.

RamanTheDeveloper:
Thanks. Just to get the issue straight, you would like to have a link towards a person's registration added to the email that is sent when an event is changed or registration is deleted?

dunkOnIT:
Not quite - we want a link to the person's registration added in emails submitted to organizers via the  [contact form](https://www.worldcubeassociation.org/contact/website), the user has the option to contact the organizers of a specific competition. (Users can contact organizers with this form with the "Contact organizers of a specific competition" option). 

I'm not sure we can link to an individual's registration - but if it is possible, if the user submitting the contact form is logged in, we want to add a link to their registration in the email. 

For cases where they aren't logged in, let's add a link to the list of registrations for the competition anyway - I think that will be useful for organizers quite often.

FinnIckler:
There is a way to link to an individual registration, it currently looks like this: https://www.worldcubeassociation.org/registrations/{registration_id}/edit

RamanTheDeveloper:
What mailer class is used to send the email with the contact form information?

PR #13413, opened 2026-02-11
Add competition url in competition contact form

pranav-027:
fixes: https://github.com/thewca/worldcubeassociation.org/issues/7063 and https://github.com/thewca/worldcubeassociation.org/issues/7930

image

### DIFF
--- a/app/models/contact_competition.rb
+++ b/app/models/contact_competition.rb
@@ -7,8 +7,13 @@ class ContactCompetition < ContactForm
 
   attr_accessor :competition_id
 
+  delegate :url, to: :competition, prefix: true, allow_nil: true
+
+  def competition
+    @competition ||= Competition.find_by(id: competition_id)
+  end
+
   def to_email
-    competition = Competition.find_by(id: competition_id)
     if competition.present?
       ValidateEmail.valid?(competition.contact) ? competition.contact : competition.managers.map(&:email)
     else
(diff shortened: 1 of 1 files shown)


app/models/contact_competition.rb
```diff
@@ -19,4 +19,11 @@ def to_email
   def subject
     Time.now.strftime("[WCA Website] Comment for #{Competition.find_by(id: competition_id)&.name} by #{name} on %d %b %Y at %R")
   end
+
+  def competition_url
+    competition = Competition.find_by(id: competition_id)
+    return nil if competition.blank?
+
+    Rails.application.routes.url_helpers.competition_url(competition, host: EnvConfig.ROOT_URL)
+  end
```
gregorbg: There is already `def url` in `competition.rb`, so essentially this whole thing boils down to `competition&.url`.

Rails has strong idioms for these simple method definitions that essentially just "forward" to another object: Instead of calling `def`, you can call...

```suggestion

  delegate :url, to: :competition, prefix: true, allow_nil: true
```

config/locales/en.yml
```diff
@@ -577,6 +577,7 @@ en:
         document: "Document"
       contact_wrt:
         wca_id: "WCA ID"
+    competition_link_html: "Click <a href='%{url}'>here</a> to visit the competition page"
```
gregorbg, line 580: Have you done a quick skim to see whether you can "recycle" any existing translations?

gregorbg (changes requested):

config/locales/en.yml
pranav-027: Yeah I just did and didnt anything useful

PR #13414, opened 2026-02-11
Next WCA Live Projector view

### LINKED ISSUE #14819: Add projector mode

FinnIckler:


next-frontend/src/components/live/LiveUpdatingResultsTable.tsx
```diff
@@ -53,6 +55,16 @@ export default function LiveUpdatingResultsTable({
...
 
+  const [inProjectorMode, setInProjectorMode] = useState(false);
+  const enableProjectorView = useCallback(
+    () => setInProjectorMode(true),
+    [setInProjectorMode],
+  );
+  const disableProjectorView = useCallback(
+    () => setInProjectorMode(false),
+    [setInProjectorMode],
+  );
```
gregorbg: We have the React Compiler enabled in NextJS so these memoizations are definitely redundant

next-frontend/src/components/live/ResultsProjector.tsx
```diff
@@ -0,0 +1,322 @@
...
+  );
+
+  const nonemptyResultsRef =
+    useRef<components["schemas"]["LiveResult"][]>(nonemptyResults);
+  useEffect(() => {
+    nonemptyResultsRef.current = nonemptyResults;
+  });
+
+  useEffect(() => {
+    const nonemptyResults = nonemptyResultsRef.current;
```
gregorbg: What is the point of this ref if you could just access `nonEmptyResults` from the props directly?

next-frontend/src/components/live/ResultsProjector.tsx
```diff
@@ -0,0 +1,322 @@
...
+      }, DURATION.HIDING);
+      return () => clearTimeout(timeout);
+    }
+    throw new Error(`Unrecognized status: ${String(status)}`);
+  }, [status, forecastView]);
+
+  const competitorsByRegistrationId = _.keyBy(competitors, "id");
+
+  return (
+    <Dialog.Root open={true} size="full">
```
gregorbg, line 107: Why did you choose to turn this into a dialog? In the top level `LiveUpdatingResultsTable`, you could be like "If projector, then render this stuff as main content, else render the normal stuff as main content".

Or does the explicit Dialog have any advantages which I'm missing right now?

gregorbg (changes requested):

next-frontend/src/components/live/ResultsProjector.tsx
FinnIckler: the dialog makes it much easier to have everything fullscreen

next-frontend/src/components/live/ResultsProjector.tsx
FinnIckler: wca live does it like this, I honestly don't know why. maybe @jonatanklosko knows. But I can also remove it

jonatanklosko: If we access `nonemptyResults` in the `useEffect`, the linter will warn that it should be added to the `useEffect` dependency array, however that would reset the timeouts every time the results change, which is not what we want. One option would be to explicitly ignore the warning by a magic comment, but that's definitely not ideal, since it's easier to miss other dependencies in the future, and in general it is expected that values used in `useEffect` are indeed in the dependency array.

On latest React there is an alternative way to handle this by using `useEffectEvent`. `useEffectEvent` effectively wraps a function, allows it to be called from `useEffect` and that function always closes over the latest props. So it would be something like this:

```ts
const hidingDoneEvent = useEffectEvent(() => {
  setStatus(STATUS.SHOWING);
  setTopResultIndex((topResultIndex) => {
    const newIndex = topResultIndex + getNumberOfRows();
    return newIndex > nonemptyResults.length ||
      (forecastView &&
        !nonemptyResults[topResultIndex].advancing &&
        !nonemptyResults[newIndex].advancing)
      ? 0
      : newIndex;
  });
});

useEffect(() => {
  if (status === STATUS.HIDING) {
    const timeout: NodeJS.Timeout = setTimeout(hidingDoneEvent, DURATION.HIDING);
    return () => clearTimeout(timeout);
  }
}, [...])
```

We could change all `setTimeouts` to use effect events.

It may seem like much, but it's likely it's easier to understand than the ref usage.

FinnIckler: ahh makes sense, we have another place were we want to use `useEffectEvent` but we are stuck on React 18 until we can upgrade to Next 16 (just waiting on some bugfixes on their side)

next-frontend/src/components/live/ResultsProjector.tsx
```diff
@@ -0,0 +1,322 @@
...
+          },
+          forecastView ? DURATION.FORECAST_SHOWN : DURATION.SHOWN,
+        );
+        return () => clearTimeout(timeout);
+      } else {
+        return;
+      }
+    }
+    if (status === STATUS.SHOWING) {
+      const timeout: NodeJS.Timeout = setTimeout(() => {
```
jonatanklosko: Btw. `NodeJS.Timeout` is rather weird to appear in client code, since in the browser the timeout is actually a number (while in nodejs it is an actual `NodeJS.Timeout` object). A more accurate type would be `ReturnType<typeof setTimeout>` (some people define an alias for this). But honestly here I would simply skip the explicit type, since it works perfectly fine without it. In general relying on type inference within function bodies avoids noise, which makes the code easier to read most of the time.

next-frontend/src/components/live/ResultsProjector.tsx
jonatanklosko: FWIW you can define a polyfill for `useEffectEvent` and use that:

```ts
import { useCallback, useInsertionEffect, useRef } from "react";

// TODO: remove once released in React. Also remove the result of
// useEffectEvent from useEffect deps in all places using it, since
// it will no longer warn.

/**
 * Polyfill for the experimental useEffectEvent [1].
 *
 * [1]: https://react.dev/learn/separating-events-from-effects
 */
export default function useEffectEvent<T extends unknown[], U>(
  callback: (...args: T) => U,
): typeof callback {
  const ref = useRef<typeof callback | null>(null);

  useInsertionEffect(() => {
    ref.current = callback;
  }, [callback]);

  return useCallback((...args: T) => {
    const latestFn = ref.current!;
    return latestFn(...args);
  }, []);
}
```

next-frontend/src/components/live/ResultsProjector.tsx
FinnIckler: I did an AI jsx -> tsx conversion of the UseEffect so that's why it has that I guess.

next-frontend/src/components/live/ResultsProjector.tsx
```diff
@@ -0,0 +1,322 @@
...
+                          }}
+                          key={result.registration_id}
+                          style={fadeStyle({
+                            isVisible,
+                            transition: {
+                              enter: { duration: DURATION.SHOWING / 1000 },
+                              exit: { duration: DURATION.HIDING / 1000 },
+                            },
+                            style:
+                              status === STATUS.SHOWING
```
FinnIckler: @jonatanklosko while we have you here, I assume this `transitionDelay` should create the nice "fade out one at a time" effect? I couldn't get it to work in chakra like it was in mui.

jonatanklosko: Yes! I'm not sure what the issue is though, the style looks good to me. What is the behaviour you see, they animate all at once, or not animate at all (both in and out)?

jonatanklosko: Oh to be clear, the expected behaviour is to "fade out all at once, then fade in one at a time".

FinnIckler: https://github.com/user-attachments/assets/11cdc673-ad60-46ec-b356-0b52a37572de

it just fade outs all at once and then doesn't fade in

jonatanklosko: What happens if you change `key={result.registration_id}` to `key={index}`? I suspect that when the key changes and new DOM node is created, it starts with opacity: 1, so there is no prior state to transition from. It's actually interesting that it works in MUI with the `Fade` component, but perhaps they do something extra to account for that.

If that theory is correct, I don't think `key={index}` is going to fully solve it though, because if the last page has less entries, and then transitions back to the first page, there will be "new" row elements and they will show up immediately.

I would try to use a CSS `animation` instead, with animation delay. Something like this:

```css
@keyframes fadeIn {
  from { opacity: 0; }
  to { opacity: 1; }
}

@keyframes fadeOut {
  from { opacity: 1; }
  to { opacity: 0; }
}
```

```ts
{
  animation: isVisible 
    ? `fadeIn ${transition?.enter?.duration ?? 1}s ease-in-out`
    : `fadeOut ${transition?.exit?.duration ?? 1}s ease-in-out`,
  animationFillMode: 'forwards',
  ...style,
}
```

```ts
style: status === STATUS.SHOWING
  ? { animationDelay: `${index * 150}ms` }
  : {}
```

FinnIckler: I will try that. I think mui does some [React.cloneElement shenanigans](https://github.com/mui/material-ui/blob/1ee94a27b931d20b23dab5dc55b35421ba4b8c55/packages/mui-material/src/Fade/Fade.js#L24). I also tried doing that but it did feel too complicated and not in line with us using a framework:

```
const Fade = ({
  children,
  in: isVisible,
  transition,
  style,
}: {
  children: React.ReactElement;
  in: boolean;
  transition?: { enter?: { duration?: number }; exit?: { duration?: number } };
  style?: React.CSSProperties;
}) => {
  return React.cloneElement(children, {
    style: {
      opacity: isVisible ? 1 : 0,
      transition: `opacity ${
        isVisible
          ? transition?.enter?.duration ?? 1
          : transition?.exit?.duration ?? 1
      }s ease-in-out`,
      ...style,
      ...children.props.style,
    },
  });
};
```

FinnIckler: key={index} works but does exactly what you described, meaning it screws up the last transition.

The animation approach does this:

https://github.com/user-attachments/assets/a344ccd7-0818-4822-a53e-45dff79ba8e6

jonatanklosko: So there are two issues when showing new page:

1. Rows are visible until the animation starts. Perhaps we need to set `opacity` to `0`, and then the animation kicks in:

   ```diff
   {
   +  opacity: isVisible ? 0 : 1,
     animation: isVisible 
       ? `fadeIn ${transition?.enter?.duration ?? 1}s ease-in-out`
       : `fadeOut ${transition?.exit?.duration ?? 1}s ease-in-out`,
     animationFillMode: 'forwards',
     ...style,
   }
   ```

2. The SHOWING state lasts 1s and the animation delay has 150ms increments. So after 7 rows do the "incremental" effect, the SHOWING would end, and the rest would animate in. It could be improved by a bit longer SHOWING, and perhaps slightly shorter increments.

FinnIckler: Really close now, just have a "blip" of all rows now before they fade in

https://github.com/user-attachments/assets/d7234bea-3aa1-488d-a8d7-49153385dcc7

jonatanklosko: Hmm, it is actually the old rows blipping before the new rows appear. I don't know why that would happen though :< When we update the status to SHOWING, we update the index, so on the next render we render a new slice with different `key`, so the old rows should be gone. To debug perhaps see what happens if you set HIDING timeout to 500ms (do they blip for longer? maybe something conflicts with animation forwards) and 1500ms.

FinnIckler: changing the hiding timeout does not affect the blip length!

jonatanklosko: Very weird. One more thing you check is to console.log the status and index in the component body and see if SHOWING happens with the new index immediately. It's two separate setStates, but they should be batched.

FinnIckler:
Let's try to get this out for the first bigger (50+?) competition that will be part of the beta. I think it's in an ok state with the animation

next-frontend/src/components/live/ResultsProjector.tsx
FinnIckler: I revived the PR and also fixed the issue

FinnIckler:
Fixes #14819

next-frontend/src/components/live/Cells.tsx
gregorbg, line 1: Can you use a `Table.ColumnGroup` per https://www.chakra-ui.com/docs/components/table#column-group that is only rendered when `isProjector` is true?

next-frontend/src/components/live/ResultsProjector.tsx
```diff
@@ -0,0 +1,278 @@
...
+                            fontSize="1.5rem"
+                            pr={2}
+                            textAlign="right"
+                            bg={
+                              result.advancing
+                                ? "green"
+                                : result.advancing_questionable
+                                  ? "yellow"
+                                  : undefined
+                            }
```
gregorbg: I feel like we have a helper function for this nowadays

next-frontend/src/components/live/ResultsProjector.tsx
```diff
@@ -0,0 +1,278 @@
...
+                      ).includes(status);
+
+                      return competitor.results.map((result) => (
+                        <Table.Row
+                          whiteSpace="nowrap"
+                          css={{
+                            "&:last-child td": { border: 0 },
+                          }}
+                          key={`${result.registration_id}-${result.round_wcif_id}`}
+                          style={fadeStyle({
```
gregorbg: According to my CTRL+F, this is the only place in the code where `fadeStyle` is being called. So why did you extract it into a helper method in the first place?

next-frontend/src/components/live/ResultsProjector.tsx
```diff
@@ -0,0 +1,278 @@
...
+  style?: React.CSSProperties;
+}): React.CSSProperties => ({
+  opacity: isVisible ? 0 : 1,
+  // Longhand only — mixing the `animation` shorthand with `animationDelay`
+  // makes React warn when the delay is conditionally added/removed.
+  animationName: isVisible ? "fadeIn" : "fadeOut",
+  animationDuration: `${(isVisible ? transition?.enter?.duration : transition?.exit?.duration) ?? 1}s`,
+  animationTimingFunction: "ease-in-out",
+  animationFillMode: "forwards",
+  animationDelay: "0ms",
```
gregorbg: This is all stuff which has equivalents in the Chakra theming engine. Please use them over raw `style` CSS

next-frontend/src/components/live/ResultsProjector.tsx
```diff
@@ -0,0 +1,278 @@
...
+  ...style,
+});
+
+type Status = "showing" | "shown" | "hiding" | "paused";
+
+const DURATION = {
+  SHOWN: 10 * 1000,
+  SHOWING: 3000,
+  HIDING: 1000,
+} as const;
```
gregorbg: You can generalize this to be a `Record` which has `type Status` (directly above) as keys, and then use whatever is in the `status` React state to access the timeout value

next-frontend/src/theme.ts
gregorbg, line 1: Chakra already has native keyframes `fade-in` and `fade-out` delivered out of the box, no need to add them by yourself

gregorbg: cf. https://www.chakra-ui.com/docs/theming/animations#keyframes

gregorbg (changes requested):

next-frontend/src/components/live/ResultsProjector.tsx
FinnIckler: I did it, it does require a paused value, but I added a comment

next-frontend/src/components/live/ResultsProjector.tsx
```diff
@@ -0,0 +1,254 @@
...
+      <Portal>
+        <Dialog.Backdrop />
+        <Dialog.Positioner>
+          <Dialog.Content
+            maxH="100vh"
+            overflow="hidden"
+            m={0}
+            borderRadius={0}
+            w="100%"
+            h="100%"
```
gregorbg:
```suggestion
            w="full"
            h="full"
```

next-frontend/src/components/live/ResultsProjector.tsx
```diff
@@ -0,0 +1,254 @@
...
+    }
+  }, [status]);
+
+  return (
+    <Dialog.Root open={true} size="full">
+      <Portal>
+        <Dialog.Backdrop />
+        <Dialog.Positioner>
+          <Dialog.Content
+            maxH="100vh"
```
gregorbg: How is this different from `100%` (or `full`, for what it's worth)?

next-frontend/src/components/live/ResultsProjector.tsx
```diff
@@ -0,0 +1,254 @@
...
+        }, DURATION[status]);
+        return () => clearTimeout(timeout);
+      } else {
+        return;
+      }
+    }
+    if (status === "showing") {
+      const timeout = setTimeout(() => {
+        setStatus("shown");
+      }, DURATION[status]);
```
gregorbg: This is a bit of an idiom throughout the whole `useEffect` (basically appears three times, the third time with some minor changes/adjustments), can you refactor now that you have `const DURATION` above?

next-frontend/src/components/live/ResultsProjector.tsx
```diff
@@ -0,0 +1,254 @@
...
+          <Dialog.Content
+            maxH="100vh"
+            overflow="hidden"
+            m={0}
+            borderRadius={0}
+            w="100%"
+            h="100%"
+          >
+            <Dialog.Header
+              bg="blue.600"
```
gregorbg: Why are you making this specifically blue?

next-frontend/src/components/live/ResultsProjector.tsx
```diff
@@ -0,0 +1,254 @@
...
+                  size="lg"
+                  onClick={disableProjectorView}
+                >
+                  <FaTimes />
+                </IconButton>
+              </Flex>
+            </Dialog.Header>
+
+            <Dialog.Body p={0}>
+              <Table.Root size="lg" css={{ tableLayout: "fixed" }}>
```
gregorbg: Do you absolutely need the `css` parts?

gregorbg: The example at https://www.chakra-ui.com/docs/components/table#column-group which uses `Table.ColumnGroup` does not need `tableLayout: "fixed"`

next-frontend/src/components/live/ResultsProjector.tsx
```diff
@@ -0,0 +1,254 @@
...
+                      // linked round; the projector shows only their best one
+                      // rather than the combined breakdown.
+                      const result = competitor.results[0];
+
+                      return (
+                        <Table.Row
+                          whiteSpace="nowrap"
+                          css={{
+                            "&:last-child td": { border: 0 },
+                          }}
```
gregorbg: What does this do? Why is this necessary?

next-frontend/src/components/live/ResultsProjector.tsx
FinnIckler: this is just a remnants from the wca live code conversion, i'll delete it

next-frontend/src/components/live/ResultsProjector.tsx
FinnIckler: it's actually not even needed because I already do size="full" in the Dialog.Root

next-frontend/src/components/live/ResultsProjector.tsx
FinnIckler: seems like I don't

next-frontend/src/components/live/ResultsProjector.tsx
FinnIckler: it strips the divider of the bottom row, but I don't think we need it. This was another remnant of the wca live view

next-frontend/src/components/live/ResultsProjector.tsx
FinnIckler: done

gregorbg (approved):

PR #13415, opened 2026-02-12
Skip main event warning if the event is last event of the competition

pranav-027:
fixes: https://github.com/thewca/worldcubeassociation.org/issues/13343

### DIFF
--- a/app/models/competition.rb
+++ b/app/models/competition.rb
@@ -1791,7 +1791,7 @@ class Competition < ApplicationRecord
       .joins(rounds: :schedule_activities)
       .reorder('schedule_activities.end_time')
       .last
-      .event_id
+      &.event_id
   end
 
   # See https://github.com/thewca/worldcubeassociation.org/wiki/wcif


app/models/competition.rb
```diff
@@ -1864,6 +1864,14 @@ def top_level_activities
     competition_venues.includes(venue_rooms: { schedule_activities: [:child_activities] }).map(&:top_level_activities).flatten
   end
 
+  def last_event_of_competition
```
gregorbg: The name `last_event` implies that this returns a full "event". You can either:
- (a) Actually return a `CompetitionEvent` or `ScheduleActivity` or
- (b) rename the method to imply that it only returns an event ID

app/models/competition.rb
```diff
@@ -1864,6 +1864,14 @@ def top_level_activities
...
+    competition_events
+      .joins(rounds: :schedule_activities)
+      .order('schedule_activities.end_time DESC')
+      .limit(1)
+      .pick(:event_id)
```
gregorbg: If you want to stick with your query, then I think you can try...

```suggestion
    competition_events
      .joins(rounds: :schedule_activities)
      .order('schedule_activities.end_time DESC')
      .first
      .event_id # Perhaps skip this line and just return the last competition event?
```

gregorbg: Otherwise, one alternative approach that I can think of is `schedule_activities.where.not(round_id: nil).order(:end_time).last`. This returns a schedule activity, which does not immediately have an event ID. But because we filtered for `round_id IS NOT NULL`, we kinda implicitly know that there will be an ActivityCode that is "parseable" into an event ID.

It's admittedly quite hacky, so only consider this if other solutions fail.

app/models/competition.rb
pranav-027: Done

app/models/competition.rb
pranav-027: hate to admit but had to use the hack afterall

gregorbg:
Why did you decide to fundamentally alter the approach of loading the last activity's event ID? At least to me, your new approach seems fundamentally less efficient

gregorbg:
Also, Simon's PR had test cases. Please confer with him on how and when you can port these over.

simonkellly:
Would agree that the previous way of checking the last event was better and cleaner to read.

Feel free to copy/paste the tests from my pr https://github.com/thewca/worldcubeassociation.org/pull/13470/changes, and if you have any issues let me know

pranav-027:
@gregorbg its done changed working as expected now!

```ruby
competition_events
      .joins(rounds: :schedule_activities)
      .order('schedule_activities.end_time DESC')
      .first
      .event_id
```

Query:
```sql
SELECT `competition_events`.* 
FROM `competition_events` INNER JOIN `rounds` ON `rounds`.`competition_event_id` = `competition_events`.`id` 
INNER JOIN `schedule_activities` ON `schedule_activities`.`parent_activity_id` IS NULL AND `schedule_activities`.`round_id` = `rounds`.`id` 
WHERE `competition_events`.`competition_id` = 'FMCPune2024'
 ORDER BY `competition_events`.`event_id` ASC, schedule_activities.end_time DESC LIMIT 1;
```

**--->**

```ruby
competition_events
      .joins(rounds: :schedule_activities)
      .reorder('schedule_activities.end_time DESC')
      .first
      .event_id
```

Query:
```sql
SELECT `competition_events`.* FROM `competition_events` 
INNER JOIN `rounds` ON `rounds`.`competition_event_id` = `competition_events`.`id` 
INNER JOIN `schedule_activities` ON `schedule_activities`.`parent_activity_id` IS NULL AND `schedule_activities`.`round_id` = `rounds`.`id` 
WHERE `competition_events`.`competition_id` = 'FMCPune2025' 
ORDER BY schedule_activities.end_time DESC LIMIT 1
```

app/models/competition.rb
```diff
@@ -1786,6 +1786,14 @@ def top_level_activities
     competition_venues.includes(venue_rooms: { schedule_activities: [:child_activities] }).map(&:top_level_activities).flatten
   end
 
+  def last_event_id_of_competition
+    competition_events
+      .joins(rounds: :schedule_activities)
+      .reorder('schedule_activities.end_time DESC')
+      .first
```
gregorbg: Random thought: Shouldn't `reorder('foo DESC').first` be the same as `reorder('foo').last`? I think Rails actually realizes a `last` call internally by changing to `DESC` under the hood...

Try it out, if it does not work the way I remember then I will green-light your PR as-is

pranav-027: Works as expected! changed

gregorbg:
That simply means that we didn't have a chance to deploy and roll out your code yet.
The PR being merged does not mean that it's immediately available in prod.

PR #13416, opened 2026-02-12
Dual Round View

FinnIckler:
I will continue to work on this tonight, still need to do the changes to applyDiff. But I am opening already to get some eyes on the logic of `mergeAndOrderResults`

### DIFF
--- a/next-frontend/src/components/live/Cells.tsx
+++ b/next-frontend/src/components/live/Cells.tsx
@@ -2,17 +2,16 @@ import { Format } from "@/lib/wca/data/formats";
 import { Link, Table } from "@chakra-ui/react";
 import { Stat, statColumnsForFormat } from "@/lib/live/statColumnsForFormat";
 import { rankingCellColorPalette } from "@/lib/live/rankingCellColorPalette";
-import { components } from "@/types/openapi";
 import { padSkipped } from "@/lib/live/padSkipped";
 import { formatAttemptResult } from "@/lib/wca/wcif/attempts";
 import { recordTagBadge } from "@/components/results/TableCells";
-import { LiveResult } from "@/types/live";
+import { LiveAttempt, LiveCompetitor, LiveResult } from "@/types/live";
 
 export function LiveTableHeader({
-  isDual = false,
+  isLinked = false,
   format,
 }: {
-  isDual?: boolean;
+  isLinked?: boolean;
   format: Format;
 }) {
   const solveCount = format.expected_solve_count;
@@ -25,7 +24,7 @@ export function LiveTableHeader({
       <Table.Row>
         <Table.ColumnHeader textAlign="right">#</Table.ColumnHeader>
         <Table.ColumnHeader>Competitor</Table.ColumnHeader>
-        {isDual && <Table.ColumnHeader>Round</Table.ColumnHeader>}
+        {isLinked && <Table.ColumnHeader>Round</Table.ColumnHeader>}
         <Table.ColumnHeader>Country</Table.ColumnHeader>
         {attemptIndexes.map((num) => (
           <Table.ColumnHeader key={num} textAlign="right">
@@ -49,10 +48,7 @@ export function LivePositionCell({
 }: {
   position: number | string;
   rowSpan?: number;
-  advancingParams: {
-    advancing: boolean;
-    advancing_questionable: boolean;
-  };
+  advancingParams: Pick<LiveResult, "advancing_questionable" | "advancing">;
 }) {
   return (
     <Table.Cell
@@ -76,7 +72,7 @@ export function LiveCompetitorCell({
   isAdmin?: boolean;
   rowSpan?: number;
   competitionId: string;
-  competitor: { name: string; id: number };
+  competitor: Pick<LiveCompetitor, "id" | "name">;
 }) {
   return (
     <Table.Cell rowSpan={rowSpan}>
@@ -100,7 +96,7 @@ export function LiveAttemptsCells({
   competitorId,
 }: {
   format: Format;
-  attempts: components["schemas"]["LiveAttempt"][];
+  attempts: LiveAttempt[];
   eventId: string;
   competitorId: number;
 }) {
@@ -140,10 +136,8 @@ export function LiveStatCells({
     <Table.Cell
       key={`${competitorId}-${stat.name}`}
       textAlign="right"
-      style={{
-        position: "relative",
-        fontWeight: shouldHighlight(statIndex) ? "bold" : "normal",
-      }}
+      position="relative"
+      fontWeight={shouldHighlight(statIndex) ? "bold" : "normal"}
     >
       {formatAttemptResult(result[stat.field], eventId)}{" "}
       {!isAdmin && recordTagBadge(result[stat.recordTagField])}
--- a/next-frontend/src/components/live/LiveResultsTable.tsx
+++ b/next-frontend/src/components/live/LiveResultsTable.tsx
@@ -23,7 +23,7 @@ export default function LiveResultsTable({
   competitors,
   isAdmin = false,
   showEmpty = true,
-  showDualRoundsView = false,
+  showLinkedRoundsView = false,
 }: {
   resultsByRegistrationId: LiveResultsByRegistrationId;
   formatId: string;
@@ -32,7 +32,7 @@ export default function LiveResultsTable({
   competitors: LiveCompetitor[];
   isAdmin?: boolean;
   showEmpty?: boolean;
-  showDualRoundsView?: boolean;
+  showLinkedRoundsView?: boolean;
 }) {
   const competitorsByRegistrationId = _.keyBy(competitors, "id");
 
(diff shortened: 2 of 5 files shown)


FinnIckler:
image
Changed what we talked about in the meeting, also did a lot of refactors. Now it's time to dive into the live update changes

next-frontend/src/lib/live/statColumnsForFormat.ts
maxidragon, line 5:
```suggestion
    name: "Average",
```

maxidragon, line 10:
```suggestion
    name: "Best",
```

FinnIckler:
Thanks, but compared to WCA live these will actually be localized in a future PR so no need to change it now

gregorbg:
Hm. Good draft, but we need to take it further (and simplify). Right now, we do two (or depending on how flexibly you count, three) things to communicate advancement:

- Order of the rows within a competitor (the advancing average goes first, no matter whether it was in the first or second round)
- Background color of the row

In particular, these two concepts kinda "defeat" each other: With the current ordering algorithm in place, it will always be the first/upper row which is coded blue. That makes the blue coding quite redundant IMO.

And then there's also the overall approach of printing an average in bold: I get that this is coming from the "classical" round view table, but in this particular case I wonder whether it makes sense to always print every average in bold. The third option which I hinted at at the top could be to use bold typeface for the "proceeding" average.

I am not very partial towards either approach. Other "hybrid" approaches (like blue background plus bold typefacing of only the average cell that actually proceeds) are conceivable.
But I am however very partial towards styling UI elements carefully and not overloading things. The last most recent screenshot with the blue markers is definitely way too much.

app/jobs/update_live_result_job.rb
```diff
@@ -35,6 +35,10 @@ def perform(live_result, results, entered_by_id)
 
     diff = Live::DiffHelper.add_forecast_stats(diff, round)
 
-    ActionCable.server.broadcast(Live::Config.broadcast_key(round.wcif_id), diff)
+    rounds_to_broadcast_to = round.linked_round&.wcif_ids || [round.wcif_id]
+
+    rounds_to_broadcast_to.each do |wcif_id|
+      ActionCable.server.broadcast(Live::Config.broadcast_key(wcif_id), diff.merge({ "wcif_id" => round.wcif_id }))
+    end
```
gregorbg: I think this is unnecessary now that we've adjusted the frontend to support listening to multiple round channels at once

app/models/round.rb
```diff
@@ -359,9 +359,13 @@ def time_limit_is_too_slow?
     time_limit != TimeLimit::UNDEF_TL && time_limit.cumulative_round_ids.empty? && self.event.fast_event? && time_limit.centiseconds > 60_000
   end
 
-  def self.find_by_wcif_id!(wcif_id, competition_id)
+  def self.find_by_wcif_id!(wcif_id, competition_id, includes: [])
     event_id, number = Round.parse_wcif_id(wcif_id).values_at(:event_id, :round_number)
-    Round.includes(:competition_event, live_results: %i[live_attempts event]).find_by!(competition_event: { competition_id: competition_id, event_id: event_id }, number: number)
+
+    base_includes = [:competition_event]
+    all_includes  = base_includes + Array(includes)
```
```suggestion
    all_includes = [:competition_event, *Array.wrap(includes)]
```

next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/page.tsx
```diff
@@ -25,15 +27,43 @@ export default async function ResultPage({
...
 
-  const { competitors, format } = data;
+  const { competitors, format, id, linked_round_ids } = data;
+
+  if (linked_round_ids) {
+    const linkedRounds = await Promise.all(
+      linked_round_ids
+        .filter((wcif_id) => wcif_id !== id)
+        .map((wcif_id) => getResultByRound(competitionId, wcif_id)),
+    );
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/page.tsx:37]: What if one of them throws an error or is otherwise not successful?

next-frontend/src/components/live/LiveUpdatingResultsTable.tsx
```diff
@@ -2,36 +2,58 @@
...
 }: {
-  eventId: string;
+  roundWcifId: string;
   formatId: string;
   competitionId: string;
   competitors: components["schemas"]["LiveCompetitor"][];
   title: string;
   isAdmin?: boolean;
   showEmpty?: boolean;
+  isDualRound?: boolean;
```
gregorbg: Should be more like `isLinkedRound` to keep our lingo consistent

next-frontend/src/lib/live/rankingCellColorPalette.ts
gregorbg, line 1: This file feels familiar, have you merged yet?

next-frontend/src/components/live/Cells.tsx
```diff
@@ -0,0 +1,152 @@
...
+  position,
+  rowSpan,
+  advancingParams,
+}: {
+  position: number | string;
+  rowSpan?: number;
+  advancingParams: {
+    advancing: boolean;
+    advancing_questionable: boolean;
+  };
```
gregorbg: Please slice this from an actual OpenAPI type definition, just as you do in the "ranking cell color" helper.
You could even consider sharing the "sliced" type definition through a helper.

next-frontend/src/components/live/Cells.tsx
```diff
@@ -0,0 +1,152 @@
...
+}) {
+  return (
+    <Table.Cell
+      width={1}
```
gregorbg, line 55: What does this `1` do, what unit does it represent?

next-frontend/src/components/live/Cells.tsx
```diff
@@ -0,0 +1,152 @@
...
+}
+
+export function LiveAttemptsCells({
+  format,
+  attempts,
+  eventId,
+  competitorId,
+}: {
+  format: Format;
+  attempts: components["schemas"]["LiveAttempt"][];
```
gregorbg: Type alias

next-frontend/src/components/live/Cells.tsx
```diff
@@ -0,0 +1,152 @@
...
+export function LiveCompetitorCell({
+  isAdmin = false,
+  rowSpan,
+  competitionId,
+  competitor,
+}: {
+  isAdmin?: boolean;
+  rowSpan?: number;
+  competitionId: string;
+  competitor: { name: string; id: number };
```
gregorbg: Can this be sliced from an actual OpenAPI definition?

next-frontend/src/components/live/Cells.tsx
```diff
@@ -0,0 +1,152 @@
...
+import { Stat, statColumnsForFormat } from "@/lib/live/statColumnsForFormat";
+import { rankingCellColorPalette } from "@/lib/live/rankingCellColorPalette";
+import { components } from "@/types/openapi";
+import { padSkipped } from "@/lib/live/padSkipped";
+import { formatAttemptResult } from "@/lib/wca/wcif/attempts";
+import { recordTagBadge } from "@/components/results/TableCells";
+import { LiveResult } from "@/types/live";
+
+export function LiveTableHeader({
+  isDual = false,
```
gregorbg: `isLinked`

next-frontend/src/components/live/LiveResultsTable.tsx
```diff
@@ -1,76 +1,71 @@
...
 export default function LiveResultsTable({
   resultsByRegistrationId,
-  eventId,
   formatId,
+  roundWcifId,
   competitionId,
   competitors,
   isAdmin = false,
   showEmpty = true,
+  showDualRoundsView = false,
```
gregorbg: `showLinkedRoundsView`

next-frontend/src/components/live/LiveUpdatingResultsTable.tsx
```diff
@@ -1,37 +1,59 @@
...
+  roundWcifId: string;
   formatId: string;
   competitionId: string;
   competitors: LiveCompetitor[];
   title: string;
   isAdmin?: boolean;
   showEmpty?: boolean;
+  isLinkedRound?: boolean;
 }) {
+  const [showDualRoundsView, setShowDualRoundsView] = useState(isLinkedRound);
```
gregorbg: `showLinkedRoundsView`

next-frontend/src/components/live/Cells.tsx
```diff
@@ -0,0 +1,152 @@
...
+  };
+
+  return stats.map((stat, statIndex) => (
+    <Table.Cell
+      key={`${competitorId}-${stat.name}`}
+      textAlign="right"
+      style={{
+        position: "relative",
+        fontWeight: shouldHighlight(statIndex) ? "bold" : "normal",
+      }}
```
gregorbg: Hm, the `fontWeight` can definitely be moved out of the `style` tag and directly be defined on the `Cell` as a prop. What about the `relative`? I feel like Chakra also has a `position={something}` helper, although it may not be called `position` exactly...

gregorbg: My point is: You should avoid `style={}` and use Chakra's theming engine with (top-level) props directly on the component as much as possible.

next-frontend/src/lib/live/padSkipped.ts
```diff
@@ -3,9 +3,9 @@ import { components } from "@/types/openapi";
 export const padSkipped = (
   attempts: components["schemas"]["LiveAttempt"][],
   expectedNumberOfAttempts: number,
-) => {
+): components["schemas"]["LiveAttempt"][] => {
```
gregorbg: Type alias :)

next-frontend/src/components/live/Cells.tsx
FinnIckler: It comes from here https://chakra-ui.com/docs/theming/sizes

Issue #11297, opened 2025-04-09
[Ban Competitor Form] Add Start Date Field


pranav-027:
There have been multiple instances where the ban action gets delayed due to certain required prerequisites, and the WRT has to manually update the ban's start_date in the database. It would be helpful to include a start date field in the ban competitor form.

slongbehn:
Thanks for looking into this, but wouldn't #11051 resolve this to begin with? The reason I'm not able to ban them "on time" is because their registrations aren't deleted fast enough.

slongbehn:
@pranav-027 Just wanted to bump this again - I think this will be resolved still with #11051 and therefore does not need to be an added feature.

gregorbg:
This ain't Pranav's fault. The delay is on our part, sorry! We're still figuring out how exactly to handle the "reject under any circumstances, even if rejecting would otherwise be invalid" situation.

slongbehn:
No worries; I did not mean to imply fault, only that the issue seemed not necessary. But if this needs to be a medium-term solution that's fine as well; it used to be this way!

PR #13419, opened 2026-02-14
Add start date field in banned competitor form

pranav-027:
fixes: https://github.com/thewca/worldcubeassociation.org/issues/11297

I'll be resuming on https://github.com/thewca/worldcubeassociation.org/pull/11051 again but this seems like a resonable thing to do currently

image

image

### DIFF
--- a/app/webpacker/components/Panel/pages/BannedCompetitorsPage/BannedCompetitorForm.jsx
+++ b/app/webpacker/components/Panel/pages/BannedCompetitorsPage/BannedCompetitorForm.jsx
@@ -38,7 +38,7 @@ export default function BanendCompetitorForm({
     save(apiV0Urls.userRoles.create(), {
       userId: formValues?.user?.id,
       groupType: groupTypes.banned_competitors,
-      startDate: formValues?.startDate || new Date().toISOString().split('T')[0],
+      startDate: formValues?.startDate,
       endDate: formValues?.endDate,
       banReason: formValues?.banReason,
       scope: formValues?.scope,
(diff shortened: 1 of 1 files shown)


app/webpacker/components/Panel/pages/BannedCompetitorsPage/BannedCompetitorForm.jsx
```diff
@@ -48,6 +50,7 @@ export default function BanendCompetitorForm({
 
   const editBannedCompetitor = () => {
     save(apiV0Urls.userRoles.update(banActionRole.id), {
+      startDate: formValues?.startDate || new Date().toISOString().split('T')[0],
```
gregorbg: What is that ` || new Date().toISOString().split('T')[0]` about?
1. If that's really the smartest way to get the current day as a string (which I kinda doubt), then extract it into a helper method to make it more obvious what the purpose of this hack is.
2. Why even use it at all? The backend falls back to `start_date = params[:startDate].presence || Date.today` if nothing is supplied, so I wonder why the frontend is _also_ trying to set a value.

pranav-027: It actually makes sense to just pass null value if nothing is entered and handle the fallback on backend only! I'll remove this from frontend

pranav-027: @gregorbg  Done!

PR #13457, opened 2026-02-16
Live updating table refactors

FinnIckler:
Some refactors needed for https://github.com/thewca/worldcubeassociation.org/pull/13348 and https://github.com/thewca/worldcubeassociation.org/pull/13352/

### DIFF
--- a/next-frontend/src/lib/live/decompressDiff.ts
+++ b/next-frontend/src/lib/live/decompressDiff.ts
@@ -19,11 +19,14 @@ export function decompressDiff(
 export function decompressDiff(
   diff: DiffedLiveResult | CompressedLiveResult,
 ): PartialLiveResultWithRegistrationId | components["schemas"]["LiveResult"] {
-  return _.omitBy(
-    {
-      ...diff,
-      attempts: diff.live_attempts,
-    },
-    _.isUndefined,
-  ) as PartialLiveResultWithRegistrationId;
+  return {
+    registration_id: diff.registration_id,
+    ..._.omitBy(
+      {
+        ...diff,
+        attempts: diff.live_attempts,
+      },
+      _.isUndefined,
+    ),
+  };
 }


next-frontend/src/lib/live/orderResults.ts
```diff
@@ -6,14 +6,12 @@ export const orderResults = (
...
 ) => {
-  const validResults = results.filter((result) => result.best !== 0);
-
   const stats = statColumnsForFormat(format);
 
   const rankBy = stats[0].field;
   const secondaryRankBy = stats[1].field;
 
-  const sortedResults = [...validResults].sort((a, b) => {
+  const sortedResults = results.sort((a, b) => {
```
gregorbg: This is dangerous because `sort` mutates in-place. I think you want `toSorted`.

next-frontend/src/lib/live/decompressDiff.ts
gregorbg, line 1: Can you introduce a type alias for this intersection type?

next-frontend/src/lib/live/applyDiffToLiveResults.ts
```diff
@@ -0,0 +1,32 @@
...
+    .filter((res) => !deletedSet.has(res.registration_id))
+    .map((res) => {
+      const update = updatesMap.get(res.registration_id);
+      return update ? { ...res, ...update } : res;
+    });
+
+  return diffedResults.concat(
+    created.map(
+      // Created will always be a full result
+      (d) => decompressDiff(d) as components["schemas"]["LiveResult"],
```
gregorbg: Can we make TS understand this without manually casting? Maybe generics? (Aka whatever comes in comes out)

gregorbg (changes requested):

next-frontend/src/lib/live/orderResults.ts
FinnIckler: ahh okay. that's why it was like that before

next-frontend/src/lib/live/applyDiffToLiveResults.ts
FinnIckler: you can actually using function overloads!

next-frontend/src/lib/live/decompressDiff.ts
```diff
@@ -0,0 +1,29 @@
...
+export function decompressDiff(
+  diff: DiffedLiveResult | CompressedLiveResult,
+): PartialLiveResultWithRegistrationId | components["schemas"]["LiveResult"] {
+  return _.omitBy(
+    {
+      ...diff,
+      attempts: diff.live_attempts,
+    },
+    _.isUndefined,
+  ) as PartialLiveResultWithRegistrationId;
```
gregorbg: Still not too wild about the forced cast here. You're making sure that whatever `omitBy` does, the result is _definitely_ interpreted as `PartialLiveResultWithRegistrationId`. Then why is the `| components["schemas"]["LiveResult"]` union in the return type declaration even necessary?

Your cast basically screams "IT IS A `PartialLiveResultWithRegistrationId` STOP ASKING". But your return type says "it could also be this other thing". That seems redundant, so what am I missing here?

FinnIckler: well that's just an issue with lodash. It doesn't know that `registration_id` is definitely there in the omit by so I do the cast. I just changed the code to be a little more ugly, but more explicit.

PR #13462, opened 2026-02-17
Extend newcomer page access till results submission

danieljames-dj:
This also improves success message as per discussion with @gregorbg over chat.

### DIFF
--- a/spec/requests/results_submission_spec.rb
+++ b/spec/requests/results_submission_spec.rb
@@ -116,7 +116,8 @@ RSpec.describe ResultsSubmissionController do
     end
 
     context "when competition is ongoing and results not yet submitted" do
-      let(:ongoing_comp) { create(:competition, :announced, :ongoing) }
+      let(:comp_delegate) { create(:delegate) }
+      let(:ongoing_comp) { create(:competition, :announced, :ongoing, delegates: [comp_delegate]) }
 
       it "allows access for WRT user" do
         sign_in wrt_user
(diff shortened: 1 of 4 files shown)


spec/requests/results_submission_spec.rb
```diff
@@ -115,16 +115,28 @@
       end
     end
 
-    context "when competition is not upcoming (past competition)" do
-      let(:past_comp) { create(:competition, :announced, :past) }
+    context "when competition is ongoing and results not yet submitted" do
+      let(:ongoing_comp) { create(:competition, :announced, :ongoing) }
+
+      it "allows access for WRT user" do
```
gregorbg, line 122: What about the case "allows access for a Delegate of this current competition"?

spec/requests/results_submission_spec.rb
```diff
@@ -115,16 +115,28 @@
...
+        sign_in wrt_user
+
+        get competition_newcomer_name_format_check_path(ongoing_comp.id)
+
+        expect(response).to have_http_status(:ok)
+      end
+    end
+
+    context "when competition has results submitted" do
+      let(:results_submitted_comp) { create(:competition, :announced, results_submitted_at: Time.now) }
```
gregorbg: I think (but only 70% sure) we have a factory trait like `create(:competition, :results_submitted)` or similar (might be called `with_submitted_results` or smth, look up the factory file)

danieljames-dj: Oh ya there is `with_valid_submitted_results`, used it.

spec/requests/results_submission_spec.rb
danieljames-dj: Added

PR #13463, opened 2026-02-17
Fix couple of bugs in newcomers page

danieljames-dj:
Following two fixes are applied in this PR:
1. Duplicate person was not parsing `country_iso2`, hence it was taking the same from corresponding `user`. If there is no `user`, then this person will have `null` `country_iso2` and it impacts the newcomers page for some cases.
2. `exactSameDetails` was not considering same ~~DOB~~ gender.

### DIFF
--- a/app/models/potential_duplicate_person.rb
+++ b/app/models/potential_duplicate_person.rb
@@ -18,7 +18,7 @@ class PotentialDuplicatePerson < ApplicationRecord
       },
       duplicate_person: {
         private_attributes: %w[dob],
-        methods: %w[country country_iso2 user_id],
+        methods: %w[country user_id],
       },
     },
   }.freeze
--- a/app/webpacker/components/NewcomerChecks/SimilarPersonTable.jsx
+++ b/app/webpacker/components/NewcomerChecks/SimilarPersonTable.jsx
@@ -43,7 +43,7 @@ export default function SimilarPersonTable({
           const exactSameDetails = (
             originalUser.name === duplicatePerson.name
                   && originalUser.dob === duplicatePerson.dob
-                  && originalUser.country_iso2 === duplicatePerson.country_iso2
+                  && originalUser.country.iso2 === duplicatePerson.country.iso2
                   && originalUser.gender === duplicatePerson.gender
           );
           return (


app/webpacker/components/NewcomerChecks/SimilarPersonTable.jsx
```diff
@@ -44,6 +44,7 @@ export default function SimilarPersonTable({
             originalUser.name === duplicatePerson.name
                   && originalUser.dob === duplicatePerson.dob
                   && originalUser.country_iso2 === duplicatePerson.country_iso2
+                  && originalUser.gender === duplicatePerson.gender
```
gregorbg, line 47: Your PR description says:
> exactSameDetails was not considering same DOB.

But this code is clearly checking _gender_, while the check for DOB was already there. So what's your actual intention? Either your description or your code is wrong.

On a sidenote, why is checking for gender relevant? Or in other words, how realistic is it that name, DOB and country are a perfect match, but it's actually really two different persons with just a different gender? Does that happen even _once_ in our whole `persons` table?

gregorbg (changes requested):

app/webpacker/components/NewcomerChecks/SimilarPersonTable.jsx
danieljames-dj: Oh actually my description was wrong, I actually meant gender.

> why is checking for gender relevant

Definitely this is an edge case - happens very rarely, but I've seen cases where people registered with wrong gender accidentally. Anyway if a case like that comes, then the 'merge' button will be enabled, but when clicked it will fail in the backend saying person details are not same. To avoid that case, I feel a check in frontend too will help them understand that the person details are not same.

On a sidenote - we are not showing gender in the newcomers page, will it be worth showing gender as well so that they will definitely know what is different?

gregorbg: Okay now I'm a little bit confused. You're saying that it can happen person A competed already as Male, forgot about it, and created a new account to register for a new competition as Female?

If that understanding is correct, then we _want_ to catch person A in the `persons` table as a duplicate exactly **because** the genders are **not** the same. So I am not sure how your "only consider as duplicate if genders are the same" helps here, because your last comment describes a clear duplicate situation where genders are NOT the same.

Also, if my scenario above is correct, then we _do_ want to merge these two persons, so that A can compete based on the identity they already provided at their first, earlier comp. Why does the merge button fail in a case where we do want to merge?

danieljames-dj: > So I am not sure how your "only consider as duplicate if genders are the same" helps here

I think I didn't mention this, or maybe I might have accidentally said somewhere 😅

I'll explain in more detail. In newcomers page, we show potential duplicates based on the algorithms (the two algorithms that are existing doesn't take gender into account). In UI, when the duplicates are shown, the user will also be provided with a "merge" button, because if the delegate knows for sure that they are duplicates, they can execute the action themselves.

But our backend for "merging two users" have a logic that will throw error if the details are not matching - because backend won't know which data to take while merging (for example, if user A is male and user B is female, backend doesn't know what the actual gender is).

In that case in the frontend, we disable the "merge" button with a message "Merging is disabled as the user details does not match". At this point, user needs to take an action based on the case:

1. If the registered user's data is wrong, edit that so that they will get the "merge" option enabled.
2. If the duplicate person's data is wrong, they need to request WRT to change the duplicate person's data while submitting results, and WRT will merge them while submitting results.


Now coming to this case, in frontend we are disabling the merge if the details are "matching", but that "matching" logic is currently wrong as we are not considering gender.

gregorbg: My quote about...
> So I am not sure how your "only consider as duplicate if genders are the same" helps here

...came from your code in this PR.

Your current PR makes it so that "duplicates only count as duplicates if the genders match". But when I asked you why that's necessary, you justified your change by saying "happens very rarely, but I've seen cases where people registered with wrong gender accidentally".

So your explanation about the "merge" fail and the background of why the merge fails makes perfect sense. I get that, no need to explain any further.

What still confuses me: The code in this PR is changed, to make sure that the UI only displays a duplicate as a duplicate if the gender also matches. But your explanation says that the actual, real-world "whoops I registered with the wrong information" duplicate happens when the gender does NOT match.

danieljames-dj: > The code in this PR is changed, to make sure that the UI only displays a duplicate as a duplicate if the gender also matches

Sorry for confusion, that's not the case. The duplicate will be shown irrespective of whether they have exact same details or not. The duplicate is judged only based on the algorithm, and not on whether their details are same.

But across every potential duplicate row, we will have a "merge" button which should be enabled only if the details match. If the details are not matching, they won't be able to action the merge anyway, that's why we are disabling the "merge" button.

app/models/potential_duplicate_person.rb
```diff
@@ -18,7 +18,7 @@ class PotentialDuplicatePerson < ApplicationRecord
       },
       duplicate_person: {
         private_attributes: %w[dob],
-        methods: %w[country user_id],
+        methods: %w[country country_iso2 user_id],
```
gregorbg: Hm. You're already serializing the whole `country` into the json. This is an object which in turn should already serialize its own ISO2. So in the frontend, instead of `duplicate_person.country_iso2`, I think that `duplicate_person.country.iso2` should work...

gregorbg (changes requested):

app/webpacker/components/NewcomerChecks/SimilarPersonTable.jsx
gregorbg: Okay wow, sorry that was a complicated one :sweat_smile:

app/models/potential_duplicate_person.rb
danieljames-dj: Ya that will work, changed.

PR #13465, opened 2026-02-17
Compute Best Possible and Worst Possible Average on the backend if needed

FinnIckler:
Adding the check to Round Results only has one extra query so it's pretty cheap

### DIFF
--- a/app/models/live_result.rb
+++ b/app/models/live_result.rb
@@ -26,7 +26,7 @@ class LiveResult < ApplicationRecord
 
   DEFAULT_SERIALIZE_OPTIONS = {
     only: %w[global_pos local_pos registration_id round_id best average single_record_tag average_record_tag advancing advancing_questionable entered_at entered_by_id],
-    methods: %w[event_id attempts result_id best_and_worst_possible_average],
+    methods: %w[event_id attempts result_id forecast_statistics],
     include: %w[],
   }.freeze
 
@@ -114,7 +114,7 @@ class LiveResult < ApplicationRecord
     diff if diff.except("registration_id").present?
   end
 
-  def best_and_worst_possible_average
+  def forecast_statistics
     # use .length on purpose here as otherwise we would use one query per row
     LiveResult.compute_best_and_worse_possible_average(live_attempts.as_json, round) if live_attempts.length < round.format.expected_solve_count
   end
@@ -133,9 +133,8 @@ class LiveResult < ApplicationRecord
         }
       end
 
-      attempts = padded.map { |l| LiveAttempt.new(l) }
-      avg, = LiveResult.compute_average_and_best(attempts, round)
-      avg
+      attempts = padded.map { LiveAttempt.new(it) }
+      LiveResult.compute_average_and_best(attempts, round).first
     end
   end
 
--- a/lib/live/diff_helper.rb
+++ b/lib/live/diff_helper.rb
@@ -42,7 +42,7 @@ module Live
     end
 
     def self.forecast_for(updated_result, round)
-      return updated_result unless result["live_attempts"].length < round.format.expected_solve_count
+      return updated_result unless updated_result["live_attempts"].length < round.format.expected_solve_count
 
       updated_result.merge(LiveResult.compute_best_and_worse_possible_average(updated_result["live_attempts"], round))
     end
--- a/next-frontend/openapi/schemas/live/LiveResult.yaml
+++ b/next-frontend/openapi/schemas/live/LiveResult.yaml
@@ -23,7 +23,7 @@ properties:
     type: integer
   best:
     type: integer
-  best_and_worst_possible_average:
+  forecast_statistics:
     type: object
     properties:
       best_possible_average:
--- a/next-frontend/src/types/openapi.ts
+++ b/next-frontend/src/types/openapi.ts
@@ -852,7 +852,7 @@ export interface components {
             global_pos: number;
             local_pos: number;
             best: number;
-            best_and_worst_possible_average?: {
+            forecast_statistics?: {
                 best_possible_average?: number;
                 worst_possible_average?: number;
             };


app/models/live_result.rb
```diff
@@ -113,6 +114,43 @@ def self.compute_diff(before_result, after_result)
     diff if diff.except("registration_id").present?
   end
 
+  def best_and_worst_possible_average
+    # use .length on purpose here as otherwise we would use one query per row
+    LiveResult.compute_best_and_worse_possible_average(live_attempts.as_json, round) if live_attempts.length < round.format.expected_solve_count
+  end
+
+  def self.compute_best_and_worse_possible_average(live_attempts, round)
```
gregorbg, line 122: This contains a lot of structurally similar code. The "best" parts and "worst" parts are essentially the same, can you refactor for code reuse?

app/models/live_result.rb
```diff
@@ -113,6 +114,43 @@ def self.compute_diff(before_result, after_result)
     diff if diff.except("registration_id").present?
   end
 
+  def best_and_worst_possible_average
+    # use .length on purpose here as otherwise we would use one query per row
+    LiveResult.compute_best_and_worse_possible_average(live_attempts.as_json, round) if live_attempts.length < round.format.expected_solve_count
```
gregorbg, line 119: Why do you call `as_json` here?

next-frontend/openapi/schemas/live/LiveResult.yaml
```diff
@@ -23,6 +23,13 @@ properties:
     type: integer
   best:
     type: integer
+  best_and_worst_possible_average:
+    type: object
+    properties:
+      best_possible_average:
+        type: integer
+      worst_possible_average:
+        type: integer
```
gregorbg: Why did you choose to model it this way?
For global and local pos, you do *not* provide a `global_and_local_pos: { global_pos: 123, local_pos: 456 }` style object. But for BPA/WPA, you do, and I find that confusing -- unless I'm missing something?

lib/live/diff_helper.rb
```diff
@@ -36,5 +36,16 @@ def self.compute_created(before_hash, after_hash)
...
+    def self.add_forecast_stats(diff, round)
+      diff["updated"] = diff["updated"].map do |updated|
+        if updated["live_attempts"].length < round.format.expected_solve_count
+          updated.merge(LiveResult.compute_best_and_worse_possible_average(updated["live_attempts"], round))
+        else
+          updated
+        end
+      end
+      diff
+    end
```
gregorbg, line 48: Not a big fan of this manual "looping" approach. Ruby has `merge` to override certain hash properties, what about something along the lines of...

```ruby
def self.add_forecast_stats(diff, round)
  diff.merge("updated" => Array.wrap(diff["updated"]).map { forecast_for(it, round) })
end

def self.forecast_for(updated_result, round)
  return updated_result unless result["live_attempts"].length < round.format.expected_solve_count

  updated_result.merge(LiveResult.compute_best_and_worse_possible_average(updated_result["live_attempts"], round))
end
```

app/models/live_result.rb
FinnIckler: because it needs to be a hash

next-frontend/openapi/schemas/live/LiveResult.yaml
FinnIckler: because that's how I have it in the serialize method hash. I don't know if it's possible to do this without having separate methods for worst and best possible

lib/live/diff_helper.rb
FinnIckler: that is nice. Will change to that

app/models/live_result.rb
```diff
@@ -113,6 +114,31 @@ def self.compute_diff(before_result, after_result)
...
+      padded = live_attempts + Array.new(missing_count) do |i|
+        {
+          "attempt_number" => live_attempts.length + i + 1,
+          "value" => score,
+        }
+      end
+
+      attempts = padded.map { |l| LiveAttempt.new(l) }
+      avg, = LiveResult.compute_average_and_best(attempts, round)
+      avg
```
```suggestion
      LiveResult.compute_average_and_best(attempts, round)[0]
```

gregorbg: Or even `compute(...).first`

next-frontend/openapi/schemas/live/LiveResult.yaml
gregorbg: Then might I at least suggest a less specific name for the parent properties? Like `forecast_statistics` or similar?

app/models/live_result.rb
```diff
@@ -113,6 +114,31 @@ def self.compute_diff(before_result, after_result)
...
+    LiveResult.compute_best_and_worse_possible_average(live_attempts.as_json, round) if live_attempts.length < round.format.expected_solve_count
+  end
+
+  def self.compute_best_and_worse_possible_average(live_attempts, round)
+    missing_count = round.format.expected_solve_count - live_attempts.length
+
+    {
+      "best_possible_average" => BEST_POSSIBLE_SCORE,
+      "worst_possible_average" => WORST_POSSIBLE_SCORE,
+    }.transform_values do |score|
```
gregorbg, line 128: Cool! 🤩

app/models/live_result.rb
```diff
@@ -113,6 +114,31 @@ def self.compute_diff(before_result, after_result)
...
+      padded = live_attempts + Array.new(missing_count) do |i|
+        {
+          "attempt_number" => live_attempts.length + i + 1,
+          "value" => score,
+        }
+      end
+
+      attempts = padded.map { |l| LiveAttempt.new(l) }
```
```suggestion
      attempts = padded.map { LiveAttempt.new(it) }
```

gregorbg: We should look into enforcing the RuboCop for `it`...

FinnIckler: sadly it doesn't exist. LLM said there is Style/RedundantBlockArguments but that's not real

PR #13466, opened 2026-02-17
Add merge WCA ID functionality for users

danieljames-dj:


app/controllers/users_controller.rb
```diff
@@ -109,6 +110,20 @@ def merge
     render status: :ok, json: { success: true }
   end
 
+  def merge_wca_id
+    user = User.find(params.require(:userId))
+    wca_id = params.require(:wcaId)
+    person = Person.find_by(wca_id: wca_id)
+
+    return render status: :bad_request, json: { error: "WCA ID #{wca_id} does not exist" } if person.nil?
```
gregorbg:
```suggestion
    return render status: :not_found, json: { error: "WCA ID #{wca_id} does not exist" } if person.nil?
```

app/controllers/users_controller.rb
```diff
@@ -109,6 +110,20 @@ def merge
...
+    return render status: :bad_request, json: { error: "User already has a WCA ID: #{user.wca_id}" } if user.wca_id.present?
+    return render status: :bad_request, json: { error: "WCA ID #{wca_id} is already assigned to user #{person.user.id}" } if person.user.present?
+
+    user.update!(wca_id: wca_id)
```
gregorbg: Is this really the way you handle WCA ID claims on a raw database level? I vaguely remember that there's a lot more to it like dummy account removal, avatar transfer, etc...

app/webpacker/components/NewcomerChecks/MergeModal.jsx
```diff
@@ -26,13 +23,26 @@ export default function MergeModal({
...
+  const onWcaIdMergeSuccess = () => {
+    queryClient.setQueryData(
+      ['last-duplicate-checker-job', competitionId],
+      (previousData) => ({
+        ...previousData,
+        potential_duplicate_persons: previousData.potential_duplicate_persons.filter(
+          (person) => person.original_user_id !== originalUser.id,
+        ),
+      }),
+    );
```
gregorbg: From my understanding, this code snippet removes the "freshly merged user" from the list of potential duplicates. Please help me to understand why this makes sense in the user journey.

What happens if a person has been merged successfully, and the page gets hard-reloaded? (Either by the Delegate pressing F5 or just by closing the browser window, shutting down your computer, having lunch, and going back to duplicate checks later during the day). Wouldn't that make the merged user "magically re-appear" because he is still part of the duplicate persons as far as the backend is concerned?

app/webpacker/components/Panel/views/MergeWcaIdToUser/index.jsx
```diff
@@ -0,0 +1,37 @@
...
+export default function MergeWcaIdToUser({
+  userId, wcaId, onSuccess,
+}) {
+  const {
+    mutate: mergeWcaIdMutation,
+    isPending,
+    isError,
+    error,
+  } = useMutation({
+    mutationFn: () => mergeWcaIdToUser(userId, wcaId),
```
gregorbg: `mutationFn` should not grab variables outside of the `useMutation` scope. Try to pass them in as parameters.

It may seem silly, but it makes the underlying React renders more stable.

app/controllers/users_controller.rb
```diff
@@ -109,6 +110,20 @@ def merge
     render status: :ok, json: { success: true }
   end
 
+  def merge_wca_id
```
gregorbg: In terms of naming (for this controller method and all corresponding routes): Shouldn't this be more like `assign_wca_id`?

gregorbg: Below in the code, you're aborting if the user already has a WCA ID. So this method can only succeed if the user does not have a WCA ID yet. And that does not feel like a "merge" to me...

gregorbg: On that note, aren't you completely circumventing your claim WCA ID ticket logic here?

Imagine (in a perfect world, where we already have your ticketing system in full use), that some other John Doe user unrelated to this API call has a pending claim on the WCA ID that you're assigning in this controller. The assignment in the controller would go through, but John Doe will not be notified that his claim has now gone stale.

app/controllers/users_controller.rb
danieljames-dj: Those happen here: https://github.com/thewca/worldcubeassociation.org/blob/6abc6465a8c3008001b7758756f886e0262c4f51/app/models/user.rb#L323C7-L323C61

Currently how it is happening is, delegate will go to user edit page, and add the WCA ID to the WCA ID field and save it.

app/controllers/users_controller.rb
danieljames-dj: Yes, `assign_wca_id` makes sense to me.

Ya your example is correct, the same is applicable for the current save as well. Maybe I'll have a method in `user.rb` where I'll also handle the pending claims.

danieljames-dj: Also the new implementation have some common changes with my other cleanup PR, but that can be ignored for now. I'll make sure to rebase the one that get approved later.

app/webpacker/components/Panel/views/MergeWcaIdToUser/index.jsx
danieljames-dj: Oh ya I remember, done.

app/webpacker/components/NewcomerChecks/MergeModal.jsx
danieljames-dj: Oh I actually missed to remove them while assigning. I did it now.

app/models/user.rb
```diff
@@ -1578,6 +1578,22 @@ def transfer_data_to(new_user)
     end
   end
 
+  def assign_wca_id(wca_id)
+    return if wca_id.blank?
+    raise "User #{id} already has WCA ID #{self.wca_id}" if self.wca_id.present?
+
+    stale_claims = User.where(unconfirmed_wca_id: wca_id).to_a
```
gregorbg: This feels like it could be dangerous, because at this point in time, it will also find the one user that you're actually legitimately trying to assign the WCA ID to. Is this your intention? In that case, it's not dangerous.

But otherwise, you should do something like `User.where.not(id: self.id).where(...)`

app/models/user.rb
```diff
@@ -1578,6 +1578,22 @@ def transfer_data_to(new_user)
...
+
+    ActiveRecord::Base.transaction do
+      update!(wca_id: wca_id)
+      User.where(id: stale_claims.map(&:id))
+          .update_all(unconfirmed_wca_id: nil, delegate_id_to_handle_wca_id_claim: nil)
```
gregorbg, line 1591: I think (not sure, please check) that `update_all` has an `update_all!` equivalent. If yes, then please use it to be consistent with your `update!` call above.

danieljames-dj: A quick google search and LLM search said that `update_all` does not have an `update_all!` equivalent in Rails. Also couldn't find any occurrences for `update_all!`.

app/models/user.rb
danieljames-dj: It was not intentional, good catch. Current behavior would have thrown a claim cancellation email to the user which is actually bad. I've made the change as you suggested.

app/models/user.rb
gregorbg: Okay, then I was confused a bit. Because for example for `insert_all`, I know for sure that there's definitely also `insert_all!`. But maybe updates are a bit of an exception...

app/models/user.rb
```diff
@@ -1578,6 +1578,22 @@ def transfer_data_to(new_user)
     end
   end
 
+  def assign_wca_id(wca_id)
+    return if wca_id.blank?
+    raise "User #{id} already has WCA ID #{self.wca_id}" if self.wca_id.present?
+
+    stale_claims = User.where(unconfirmed_wca_id: wca_id).where.not(id: id).to_a
```
gregorbg: Why do you force it `to_a` here?
- The first usage within the transaction below you can just do `stale_claims.pluck(:id)` or even `stale_claims.ids` (which is a magic Rails alias for `pluck(:id)`
- The second usage after the transaction where you loop using `each` will be smart enough to load the models into memory and iterate "under the hood" even without an explicit `to_a`

gregorbg (changes requested):
I think a lot of changes here were independently merged in another PR. Please confirm the merge conflicts.

app/models/user.rb
danieljames-dj: That's a good idea. But when I checked with LLM, it said that `to_a` is necessary because ActiveRecord is doing lazy loading, and without `to_a`, `stale_claims` will be executed only when used. So in the second usage, when the `slaim_claims` gets loaded it will be empty (this particular portion was LLM).

I then manually added logs and checked and I was actually able to see `stale_claims` was not empty in the last line. I also remember something like this in roles long ago where I didn't face lazy loading issue in local.

I'm a bit bad at the core theory of rails. Can you please help me which way I should take? If I go without `to_a`, and if the `stale_claims` actually becomes empty as LLM is suspecting, then I think we won't even know if it actually worked, it will silently fail.

gregorbg: Rails resolves queries as late as possible, and only keeps them cached in memory after it has been fetched.
- `User.where(unconfirmed_wca_id: wca_id)` is just building some filter condition on `User`. It is not yet necessary to actually _execute_ the SQL query that contains all those filters.
- `.where.not(id: id)` is just adding another filter condition, which means Rails still does not need to fetch. It just adds to the list of conditions that it will use when it fetches in the future (in other languages/frameworks, this is often represented by the concept of a "query builder").
- `.to_a` is like explicitly forcing Rails to load stuff into memory, saying "I don't care about your clever caching, I want this in memory NOW!!". From here on, the array in your memory stays fixed, even if the underlying data in your database has changed.

For your PR, this has the following practical implications:
1. In the case of `pluck(:some_column)`, you're actually only reading a slice of certain values, and Rails does not cache `pluck` results at all -- because the resulting values are not representing full models, so there is nothing that can be cached.
`stale_claims.ids` is just a magic alias to `stale_claims.pluck(:id)`, so Rails will execute `SELECT id FROM ...` directly, return that result, and `stale_claims` itself won't have anything loaded or cached in memory.
2. `.each` actually forces Rails to resolve, so at line 1591 (at the time of writing this comment) it will load a bunch of `User`s into memory and iterate over them. If two lines down, you execute `stale_claims.each { do_something_else }`, it will use the cached users from the first `each`.

If you try stuff out in the Rails console (`bin/rails c`) you have to be careful though: The Rails console is designed in a way that it pretty-prints the output to the user, which is not the default behavior of a Rails program.
So if you write `foo_query = User.where.not(id: 7121)` in a normal Ruby code like your task script here, an SQL query will **NOT** be fired. Only later, if you do something like `foo_query.each { ... }` or `foo_query.map { ... }` you will force Rails to "resolve" the SQL.
However in the Rails console, writing `foo_query = User.where.not(id: 7121)` will _immediately_ fire an SQL query, because the Rails console wants to pretty-print the result of that query for you. And in order to pretty-print the users represented by `foo_query`, of course it needs to fetch them from the DB.
**HOWEVER** this "pretty-print" is designed in a way that prevents Rails from directly caching. You can see that the output query is marked as `fetching for pp`, which is short for "pretty print".

Image

danieljames-dj: Thanks a lot for detailed explanation Gregor, I had some doubts but thanks for clearing them over chat. I've made the change.

app/models/user.rb
```diff
@@ -1578,6 +1578,23 @@ def transfer_data_to(new_user)
     end
   end
 
+  def assign_wca_id(wca_id)
+    return if wca_id.blank?
+    raise "User #{id} already has WCA ID #{self.wca_id}" if self.wca_id.present?
+
+    stale_claims = User.where(unconfirmed_wca_id: wca_id).where.not(id: id)
+    stale_claim_users = stale_claims.to_a
```
gregorbg: On second thought, what about `stale_claims_before_update`? This captures your intent of "actively executing the query right now" even better.

app/models/user.rb
```diff
@@ -1578,6 +1578,23 @@ def transfer_data_to(new_user)
...
+
+    ActiveRecord::Base.transaction do
+      update!(wca_id: wca_id)
+      User.where(id: stale_claims.ids)
+          .update_all(unconfirmed_wca_id: nil, delegate_id_to_handle_wca_id_claim: nil)
+      potential_duplicate_persons.delete_all
```
gregorbg, line 1592: `delete_all` fires a raw, "brutal" SQL `DELETE FROM` without respecting Rails hooks, validations or `dependent: destroy` follow-ups. Is this what you want?

[If No: Then `destroy_all` is your friend, it runs through the Rails model level]

app/models/user.rb
danieljames-dj: Ya better, changed.

app/models/user.rb
danieljames-dj: `potential_duplicate_person` currently doesn't have any dependency. But that doesn't mean that it will not have in future. What is the rule we follow here? One of the older PR had `potential_duplicate_persons.delete_all` (in the method right above this method). I wanted to make both uniform.

Do you think it's better to use `destroy_all` always or do we have some use-cases (other than bulk-delete) where we actually need to use `delete_all`?

gregorbg: There is no general rule; you always need to consider what your needs are.
- `delete_all` is **fast**, because it just fires one single SQL query and lets the database delete stuff. But for this high execution speed, you have to pay the price that Rails validations and `dependent` options on `has_many` are not respected.
- `destroy_all` is more **consistent** with your application logic, it follows your "business logic rules" the way that they are represented in your model code and validations. But you have to pay the price of looping over all entities and executing `DELETE WHERE id=...` statements on the SQL 1-by-1, row by row, making the process much slower.

You always have to decide on a case-by-case basis what makes more sense for you. If you're deleting only a single, concrete record (like something that came from `MyModel.find(...)` as opposed to many rows that came from `MyModel.where(...)`), then you aaaaalmost always want `my_comp.destroy` -- only use `my_comp.delete` if you _really_ know what you're doing.

In the case of bulk deletions like right here, it's much more individual case-by-case decision.

gregorbg (approved):

PR #13467, opened 2026-02-17
Task to cleanup stale WCA ID claims

danieljames-dj:
Currently we have so many claims which are stale because the WCA ID is already assigned to another user. This task is to cleanup those cases.

I'm also working on moving WCA ID claims to tickets (https://github.com/thewca/worldcubeassociation.org/pull/13151). Once that PR is completed, this task won't be needed anymore.

My plan is to run this script before getting that PR merged, so that we won't have stale tickets.

This task cannot be run as part of migration because we need to send email notifications for each claim cancelling.

### DIFF
--- a/app/views/wca_id_claim_mailer/notify_user_of_claim_cancelled.html.erb
+++ b/app/views/wca_id_claim_mailer/notify_user_of_claim_cancelled.html.erb
@@ -2,7 +2,11 @@
 
 <p>You have previously requested to claim the WCA ID <%= @unconfirmed_wca_id %> for your account.</p>
 
+<% if @cleanup %>
 <p>We are currently cleaning up our records and reviewing these requests. We found that the WCA ID you asked for, was already taken by another account (<%= wca_id_link @unconfirmed_wca_id %>) before we could process your request.</p>
+<% else %>
+<p>The WCA ID you asked for (<%= wca_id_link @unconfirmed_wca_id %>), is now taken by another account before we could process your request.</p>
+<% end %>
 
 <p>It means that this WCA ID is no longer available, so we have cancelled your claim.</p>
 
--- a/lib/tasks/cleanup_unclaimed_wca_ids.rake
+++ b/lib/tasks/cleanup_unclaimed_wca_ids.rake
@@ -14,7 +14,7 @@ namespace :cleanup do
     stale_claims.limit(limit).each do |user|
       wca_id = user.unconfirmed_wca_id
       if user.update_columns(unconfirmed_wca_id: nil, delegate_id_to_handle_wca_id_claim: nil)
-        WcaIdClaimMailer.notify_user_of_claim_cancelled(user, wca_id).deliver_later
+        WcaIdClaimMailer.notify_user_of_claim_cancelled(user, wca_id, cleanup: true).deliver_later
         puts "Cleared stale claim for user #{user.id} (#{user.email}) - WCA ID #{wca_id}"
       else
         puts "Failed to clear stale claim for user #{user.id} (#{user.email}) - WCA ID #{wca_id}: #{user.errors.full_messages.join(', ')}"
--- a/app/mailers/wca_id_claim_mailer.rb
+++ b/app/mailers/wca_id_claim_mailer.rb
@@ -11,9 +11,10 @@ class WcaIdClaimMailer < ApplicationMailer
     )
   end
 
-  def notify_user_of_claim_cancelled(user, unconfirmed_wca_id)
+  def notify_user_of_claim_cancelled(user, unconfirmed_wca_id, cleanup: false)
     @user = user
     @unconfirmed_wca_id = unconfirmed_wca_id
+    @cleanup = cleanup
     mail(
       to: user.email,
       subject: "Your WCA ID claim for #{unconfirmed_wca_id} has been cancelled",


app/views/wca_id_claim_mailer/notify_user_of_claim_cancelled.html.erb
```diff
@@ -0,0 +1,4 @@
+<p>Hello <%= @user.name %>,</p>
+
+<p>As part of a cleanup of unclaimed WCA IDs, we found that your pending claim for WCA ID <%= wca_id_link @unconfirmed_wca_id %> is no longer valid. This WCA ID was already assigned to another account before your claim could be processed, so your claim has been cancelled.</p>
```
gregorbg: This is quite "dense" information to process. I can really feel that this mail has been written by a WRT expert who deeply understands the claim process. But for a non-native English speaking cuber (or maybe "no English at all" speaking cuber who relies on Google Translate!) this will be _very_ hard to process...

Can you come up with a little bit "easier" explanation?

lib/tasks/cleanup_unclaimed_wca_ids.rake
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+namespace :cleanup do
+  desc "Removes stale WCA ID claims where the WCA ID has already been assigned to another user"
+  task :unclaimed_wca_ids, [:limit] => [:environment] do |_, args|
+    already_claimed_wca_ids = User.not_dummy_account.where.not(wca_id: nil).select(:wca_id)
```
gregorbg:
```suggestion
    already_claimed_wca_ids = User.not_dummy_account.where.not(wca_id: nil).pluck(:wca_id)
```

lib/tasks/cleanup_unclaimed_wca_ids.rake
```diff
@@ -0,0 +1,23 @@
...
+
+    stale_claims = User.where.not(unconfirmed_wca_id: nil).where(unconfirmed_wca_id: already_claimed_wca_ids)
```
gregorbg:
```suggestion
    stale_claims = User.where(unconfirmed_wca_id: already_claimed_wca_ids)
```

gregorbg: The "not null" check feels a little bit redundant, because you're already checking that the unconfirmed WCA ID must be in a list of `[a, b, c]`, where you know that a, b and c cannot be null.

lib/tasks/cleanup_unclaimed_wca_ids.rake
```diff
@@ -0,0 +1,23 @@
...
+
+    total = stale_claims.count
+    limit = (args[:limit] || total).to_i
+    puts "Found #{total} stale WCA ID claims, processing #{limit}"
+
+    stale_claims.limit(limit).each do |user|
```
gregorbg, line 14: If you're afraid about runtime complexity, you can do `find_each` which loads rows in batches of 100 or 1000

lib/tasks/cleanup_unclaimed_wca_ids.rake
```diff
@@ -0,0 +1,23 @@
...
+      wca_id = user.unconfirmed_wca_id
+      WcaIdClaimMailer.notify_user_of_claim_cancelled(user, wca_id).deliver_later
+      user.update_columns(unconfirmed_wca_id: nil, delegate_id_to_handle_wca_id_claim: nil)
```
gregorbg: This is extremely dangerous: You deliver the mail first, but then what if the `update_columns` statement fails? You need to conditionally tie these two events together.

See whether `update_columns` has a return value. Most Rails update methods either explode (so if your code reaches the next line, you can be sure that the update went through) or they return a boolean indicating whether the update was successful. This method does not end in a `!` so my guess is that it's the boolean type.

Anyways, you need to find some way of verifying that the update went through, and only if you definitely were able to verify then also send the email.

lib/tasks/cleanup_unclaimed_wca_ids.rake
```diff
@@ -0,0 +1,23 @@
...
+      puts "Cleared stale claim for user #{user.id} (#{user.email}) - WCA ID #{wca_id}"
+    end
+
+    puts "Successfully cleared #{[limit, total].min} stale claims" if total.positive?
```
gregorbg: Why do you use a `min` here? You should probably already apply that during the calculation of `limit` above

lib/tasks/cleanup_unclaimed_wca_ids.rake
gregorbg: In this context, also check out https://guides.rubyonrails.org/active_record_callbacks.html#per-transaction-callback

It allows you to perform stuff as part of a transaction, and within that transaction also register "things that you should do only once the transaction has finished"

app/views/wca_id_claim_mailer/notify_user_of_claim_cancelled.html.erb
danieljames-dj: Okay I've attempted to simplify it bit more with help of LLM.

lib/tasks/cleanup_unclaimed_wca_ids.rake
danieljames-dj: Done

lib/tasks/cleanup_unclaimed_wca_ids.rake
danieljames-dj: okay, makes sense. Removed

lib/tasks/cleanup_unclaimed_wca_ids.rake
danieljames-dj: I actually wanted to try with limit 1 first (with `bundle exec rake "cleanup:unclaimed_wca_ids[1]"`), and wanted to check DB if everything worked as expected. That's why I wanted the `limit` field. Overall there are 1223 cases (also answered the main question of your review here).

Following is the query I used to get the details:
```
SELECT 
    claimer.id AS claiming_user_id,
    claimer.name AS claiming_user_name,
    claimer.unconfirmed_wca_id,
    owner.id AS current_owner_id,
    owner.name AS current_owner_name
FROM 
    users AS claimer
JOIN 
    users AS owner ON claimer.unconfirmed_wca_id = owner.wca_id
WHERE 
    claimer.unconfirmed_wca_id IS NOT NULL;
```

lib/tasks/cleanup_unclaimed_wca_ids.rake
danieljames-dj: Okay done. My concern was what happens if mail sending fails - because of some per minute limit or something like that.

lib/tasks/cleanup_unclaimed_wca_ids.rake
danieljames-dj: This was to handle the case where we give higher limit - for example, giving limit as 10000 when there are only 100 stale claims.

But handling this above was better, I've done that now.

lib/tasks/cleanup_unclaimed_wca_ids.rake
FinnIckler: 1223 is a lot of emails to send all at once. We do have a sending limit of 50 emails a second, so we should make sure that that is not exceeded

danieljames-dj: Oh okay, thanks for pointing that, @FinnIckler. How about we use `wait` parameter of `deliver_later` ([link](https://api.rubyonrails.org/classes/ActionMailer/MessageDelivery.html#method-i-deliver_later))?

I'm also thinking won't this be also impacting the case where we are sending "results posted" notification?
https://github.com/thewca/worldcubeassociation.org/blob/2d79400838400d8e2c4b362084f37808a209542f/lib/competition_results_import.rb#L91

FinnIckler: mh. good point. Maybe sidekiq is smart enough to batch these itself or we just don't have enough sidekiq threads for this to be an issue.

app/views/wca_id_claim_mailer/notify_user_of_claim_cancelled.html.erb
```diff
@@ -0,0 +1,9 @@
+<p>Hello <%= @user.name %>,</p>
+
+<p>You have requested in worldcubeassociation.org to claim a WCA ID in the past.</p>
```
gregorbg:
```suggestion
<p>You have previously requested to claim the WCA ID <%= @unconfirmed_wca_id %> for your account.</p>
```

app/views/wca_id_claim_mailer/notify_user_of_claim_cancelled.html.erb
```diff
@@ -0,0 +1,9 @@
...
+
+<p>We are currently cleaning up our records and reviewing these requests. We found that the ID you asked for (<%= wca_id_link @unconfirmed_wca_id %>) was already taken by another account before we could process your request.</p>
```
gregorbg:
```suggestion
<p>We are currently cleaning up our records and reviewing these requests. We found that the WCA ID you asked for, was already taken by another account (<%= wca_id_link @unconfirmed_wca_id %>) before we could process your request.</p>
```

app/views/wca_id_claim_mailer/notify_user_of_claim_cancelled.html.erb
```diff
@@ -0,0 +1,9 @@
...
+
+<p>Because this ID is not available, we have cancelled your claim.</p>
```
gregorbg:
```suggestion
<p>It means that this WCA ID is no longer available, so we have cancelled your claim.</p>
```

lib/tasks/cleanup_unclaimed_wca_ids.rake
```diff
@@ -0,0 +1,26 @@
...
+    limit = [(args[:limit] || total).to_i, total].min
+    puts "Found #{total} stale WCA ID claims, processing #{limit}"
+
+    stale_claims.limit(limit).each do |user|
+      wca_id = user.unconfirmed_wca_id
+      if user.update_columns(unconfirmed_wca_id: nil, delegate_id_to_handle_wca_id_claim: nil)
+        WcaIdClaimMailer.notify_user_of_claim_cancelled(user, wca_id).deliver_later
+        puts "Cleared stale claim for user #{user.id} (#{user.email}) - WCA ID #{wca_id}"
+      else
+        puts "Failed to clear stale claim for user #{user.id} (#{user.email}) - WCA ID #{wca_id}"
```
gregorbg: In this case, you can also display `user.errors` in the output to see what went wrong.

danieljames-dj: Done

gregorbg (approved):

PR #13469, opened 2026-02-18
Live Double Check Page

FinnIckler:
Requires the attempts form from https://github.com/thewca/worldcubeassociation.org/pull/13348 so I am opening this in draft for now

### DIFF
--- a/spec/requests/live_quit_spec.rb
+++ b/spec/requests/live_quit_spec.rb
@@ -133,6 +133,7 @@ RSpec.describe "WCA Live API" do
                                          "average_record_tag" => nil,
                                          "registration_id" => registrations.third.id,
                                          "single_record_tag" => nil,
+                                         "last_attempt_entered_at" => anything,
                                          "live_attempts" => [] }].map { (Live::DiffHelper.compress_payload it).merge({ "user" => user }) }))
     end
   end


next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/double-check/page.tsx
```diff
@@ -0,0 +1,52 @@
...
+}: {
+  params: Promise<{ roundId: string; competitionId: string }>;
+}) {
+  const { roundId, competitionId } = await params;
+
+  const resultsRequest = await getResultByRound(competitionId, roundId);
+
+  if (!resultsRequest.data) {
+    return <p>Error loading Results</p>;
+  }
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/double-check/page.tsx]: This can use the OpenAPI error component

next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/double-check/DoubleCheck.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+  const onPrevious = () => {
+    handleRegistrationIdChange(results[currentIndex - 1].registration_id);
+    setCurrentIndex((oldIndex) => oldIndex - 1);
+  };
+
+  const onNext = () => {
+    handleRegistrationIdChange(results[currentIndex + 1].registration_id);
+    setCurrentIndex((oldIndex) => oldIndex + 1);
+  };
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/double-check/DoubleCheck.tsx]: Would you be open to using a pagination component for this? https://www.chakra-ui.com/docs/components/pagination

Comment (gregorbg) [reply next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/double-check/DoubleCheck.tsx]: I'm thinking along the lines of using `pageSize={1}` to let the user scroll through the sheets one by one

Comment (gregorbg) [reply next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/double-check/DoubleCheck.tsx]: Either you don't render the page indicator (the stuff between the prev/next buttons) at all, or you go with something like https://www.chakra-ui.com/docs/components/pagination#compact or https://www.chakra-ui.com/docs/components/pagination#count-text. I am also open to other ideas and suggestions

next-frontend/src/components/live/AttemptsForm.tsx
```diff
@@ -44,16 +45,25 @@ export default function AttemptsForm({
       parseInt(filterText, 10) === item.registrant_id,
   });
 
+  const selectedCompetitor = competitors.find((c) => c.id === registrationId);
+  const inputDisplayValue = selectedCompetitor
+    ? `${selectedCompetitor.name} (${selectedCompetitor.registrant_id})`
```
gregorbg: This notion of "glue together the name and the registrant_id in brackets" is now happening so much all over the file that it should probably go into a helper function

next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/double-check/page.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const { data, error, response } = await getResultByRound(
+    competitionId,
+    roundId,
+  );
+
+  if (error) {
+    return <OpenapiError response={response} t={t} />;
+  }
+
+  const { results, id, competitors, format } = data;
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/double-check/page.tsx]: You're treating it as a format ID further down below, so why is it called just `format`?

next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/double-check/DoubleCheck.tsx
```diff
@@ -0,0 +1,113 @@
...
+                <br />
+                {events.byId[eventId].name} - {roundNumber}
+              </Card.Header>
+              <Card.Title>Double-check</Card.Title>
+              <Card.Description>
+                Here you can iterate over results ordered by entry time (newest
+                first). When doing double-check you can place a scorecard next
+                to the form to quickly compare attempt results. For optimal
+                experience make sure to always put entered/updated scorecard at
+                the top of the pile.
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/double-check/DoubleCheck.tsx:105]: How do you actually keep track of the time order?

Comment (FinnIckler) [reply next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/double-check/DoubleCheck.tsx:105]: good point, that's what we have `last_attempt_entered_at` for. Do I just order by that by default or do we want a `order_by` query param?

next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/double-check/page.tsx
Comment (FinnIckler) [reply next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/double-check/page.tsx]: that's what wcif calls it

Comment (gregorbg) [reply next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/double-check/page.tsx]: Oh. 😅

next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/double-check/DoubleCheck.tsx
Comment (gregorbg) [reply next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/double-check/DoubleCheck.tsx:105]: Hm, having a query parameter sounds cleaner in terms of data driven design, but it would effectively require us to do a lot of re-fetching... Some pages may require the data in a different order, while some pages may require time-order.
So for now :tm: you can go with the frontend applying its own sorting as needed.

Comment (FinnIckler) [reply next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/double-check/DoubleCheck.tsx:105]: the frontend already does it's ranking based order itself and currently we don't actually return `last_attempt_entered_at` so if we would return it conditionally that would give us another type which I don't really like. 

Currently we return it in `result_id` order, why not just return it `last_attempt_entered_at` by default?

Comment (FinnIckler) [reply next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/double-check/DoubleCheck.tsx:105]: done

Comment (FinnIckler) [reply next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/double-check/DoubleCheck.tsx:105]: let's see how many tests I will need to change having added a new attribute

next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/double-check/page.tsx
```diff
@@ -0,0 +1,59 @@
...
+  );
+
+  if (error) {
+    return <OpenapiError response={response} t={t} />;
+  }
+
+  const { results, id, format } = data;
+
+  const sortedResults = results.toSorted((a, b) =>
+    b.last_attempt_entered_at.localeCompare(a.last_attempt_entered_at),
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/double-check/page.tsx]: This is not a normal way to order date strings :P
Please parse them and sort them "properly"

Comment (FinnIckler) [reply next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/double-check/page.tsx]: the whole point of iso dates is that you can order them like this no?

Comment (gregorbg) [reply next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/admin/double-check/page.tsx]: That may be, but it relies too much on maintaining the string format well and also trusting the individual browser's implementation of `localeCompare`.

I know both have their standards (particularly ECMAScript) but parsing them (and turning them into numbers like you're doing here, which is wonderful!) just feels better.

PR #13477, opened 2026-02-19
Add a Context for LiveResults

FinnIckler:
This is very useful for both the Adding Results and Dual Round Results pages.

### DIFF
--- a/next-frontend/src/providers/LiveResultProvider.tsx
+++ b/next-frontend/src/providers/LiveResultProvider.tsx
@@ -8,7 +8,6 @@ import useResultsSubscription, {
   DiffProtocolResponse,
 } from "@/lib/hooks/useResultsSubscription";
 import { applyDiffToLiveResults } from "@/lib/live/applyDiffToLiveResults";
-import { components } from "@/types/openapi";
 import { useQuery, useQueryClient } from "@tanstack/react-query";
 
 interface LiveResultContextType {


next-frontend/src/providers/LiveResultProvider.tsx
```diff
@@ -0,0 +1,114 @@
...
+} from "@/lib/hooks/useResultsSubscription";
+import { applyDiffToLiveResults } from "@/lib/live/applyDiffToLiveResults";
+
+interface LiveResultContextType {
+  liveResults: LiveResult[];
+  updateLiveResults: React.Dispatch<React.SetStateAction<LiveResult[]>>;
+  stateHash: string;
+  connectionState: ConnectionState;
+  updateStateHash: React.Dispatch<React.SetStateAction<string>>;
+  refetch: () => Promise<void>;
```
gregorbg: Why do you expose the state setters as part of the hash context? I do not see any use-case where a consumer of the context would want to manually update the state hash, for example.

next-frontend/src/providers/LiveResultProvider.tsx
```diff
@@ -0,0 +1,110 @@
...
+  const [liveResults, updateLiveResults] =
+    useState<LiveResult[]>(initialResults);
+  const [stateHash, updateStateHash] = useState<string>(initialHash);
+  const api = useAPI();
+
+  const { refetch } = api.useQuery(
+    "get",
+    "/v1/competitions/{competitionId}/live/rounds/{roundId}",
+    {
+      enabled: false,
```
gregorbg: I don't like using `enabled: false`, because it effectively "degrades" the whole query to an over-exaggerated HTML client.

You have the `initialResults` and `initialHash` already, so you can use `initialData` instead. If the shape of the data does not fit what you need, either (a) use `select` to reshape what the query client expects or (b) change the shape of the data you pass into the provider

next-frontend/src/providers/LiveResultProvider.tsx
```diff
@@ -0,0 +1,110 @@
...
+    },
+  );
+
+  const refetchResults = useCallback(async () => {
+    const { data } = await refetch();
+    if (data) {
+      updateLiveResults(data.results);
+      updateStateHash(data.state_hash);
+    }
+  }, [refetch]);
```
gregorbg: With regards to my "query client holds the state instead of being disabled" comment above: You can load the query client with `useQueryClient` and wrap the setter logic (your `update*` functions) into a `useCallback` that pipes the data that you're trying to update back to the query cache layer.

React Query's `myQueryClient.setQueryData(queryKey, foo)` also supports the `myQueryClient.setQueryData(queryKey, (oldFoo) => doSomething(oldFoo))` in-place updating the way that "classic" React state setters do

gregorbg: If you don't know which key to update, because the `useAPI` manages that "internally":
https://openapi-ts.dev/openapi-react-query/query-options

gregorbg (approved):

PR #13484, opened 2026-02-20
Add WEAT to probations change emails

maxidragon:
WEAT has access to probations panel and spreadsheet and receiving notifications would be helpful for us


gregorbg:
Code changes so far are looking good to me, but you will need to adjust the tests as well.
Currently the test cases in `role_change_mailer_spec` are not expecting WEAT to be there, so the CI fails.

PR #13511, opened 2026-02-23
Refactor LiveResultProvider to allow multiple round ids

FinnIckler:
Still making small changes (like properly fixing the shared yaml types and making the table work natively with ResultsByRegistrationId), but opening already to show how it's gonna work

### DIFF
--- a/next-frontend/src/components/live/PendingResultsTable.tsx
+++ b/next-frontend/src/components/live/PendingResultsTable.tsx
@@ -4,21 +4,6 @@ import { formatAttemptResult } from "@/lib/wca/wcif/attempts";
 import { components } from "@/types/openapi";
 import formats from "@/lib/wca/data/formats";
 import { padSkipped } from "@/lib/live/padSkipped";
-import { useLiveResults } from "@/providers/LiveResultProvider";
-
-export const rankingCellColorPalette = (
-  result: components["schemas"]["LiveResult"],
-) => {
-  if (result?.advancing) {
-    return "green";
-  }
-
-  if (result?.advancing_questionable) {
-    return "yellow";
-  }
-
-  return "";
-};
 
 export default function PendingResultsTable({
   pendingLiveResults,


next-frontend/openapi/schemas/competitions/BaseWcifRound.yaml
gregorbg, line 1: Please fiddle "wcif" into the name somewhere, to make clear that this is basically a WCIF-inspired version

next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/page.tsx
```diff
@@ -30,7 +30,10 @@ export default async function ResultPage({
   return (
     <Container bg="bg">
       <VStack align="left">
-        <LiveResultProvider initialRound={data} competitionId={competitionId}>
+        <LiveResultProvider
+          initialRounds={[data]}
+          competitionId={competitionId}
+        >
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/live/rounds/[roundId]/page.tsx]: Can you make a convenience alias that allows passing in one round just as `initialRound` prop? I don't mind giving it a different name...

Or rather, keep the single round provider at `LiveResultsProvider` and rename the "big one" to `MultiLiveResultsProvider` or smth

next-frontend/src/components/live/LiveUpdatingResultsTable.tsx
```diff
@@ -23,7 +24,9 @@ export default function LiveUpdatingResultsTable({
   isAdmin?: boolean;
   showEmpty?: boolean;
 }) {
-  const { connectionState, liveResults } = useLiveResults();
+  const { connectionState, liveResultsByRegistrationId } = useLiveResults();
+
+  const liveResults = _.flatMap(liveResultsByRegistrationId);
```
gregorbg: What does this "empty" flat map do? In my mind, you always need an action that iterates over every entry.

app/models/live_result.rb
```diff
@@ -13,6 +13,8 @@ class LiveResult < ApplicationRecord
 
   belongs_to :round
 
+  delegate :wcif_id, to: :round
```
gregorbg: Please use `prefix: true` and update usages accordingly. Live Results themselves do not have individual "WCIF IDs", so `some_live_result.wcif_id` is easy to misunderstand.

next-frontend/src/app/(wca)/competitions/[competitionId]/live/podiums/page.tsx
```diff
@@ -44,7 +44,10 @@ export default async function PodiumsPage({
             </Heading>
             {finalRound.results.length > 0 ? (
               <LiveResultsTable
-                results={finalRound.results}
+                resultsByRegistrationId={_.groupBy(
+                  finalRound.results,
+                  "registration_id",
+                )}
```
Comment (gregorbg) [inline next-frontend/src/app/(wca)/competitions/[competitionId]/live/podiums/page.tsx]: Can you move this into the preamble of the component? This is a rather complex calculatoion regarding the fact that it substantially changes the type/shape of the data, so doing this impromptu inline is not a good idea IMO

next-frontend/src/components/live/LiveResultsTable.tsx
```diff
@@ -72,64 +77,66 @@ export default function LiveResultsTable({
       </Table.Header>
 
       <Table.Body>
-        {sortedResults.map((result) => {
-          const competitor =
-            competitorsByRegistrationId[result.registration_id];
-          const hasResult = result.attempts.length > 0;
+        {sortedResultsByCompetitor.map((competitorWithResults) => {
```
gregorbg: The `ByCompetitor` suffix implies that this is a map, but actually it's an array. This is confusing regarding the naming scheme. Please either turn it into an actual map, or change the name to make it more obvious that it's a list.

next-frontend/src/components/live/LiveResultsTable.tsx
```diff
@@ -44,7 +45,11 @@ export default function LiveResultsTable({
 
   const format = formats.byId[formatId];
 
-  const sortedResults = orderResults(results, format);
+  const sortedResultsByCompetitor = mergeAndOrderResults(
```
gregorbg: Nit: The method is called `order` but the variable is called `sorted`. Please harmonize to either `sortResults` and `sorted`, or to `orderResults` and `ordered`

next-frontend/src/components/live/LiveResultsTable.tsx
```diff
@@ -72,64 +77,66 @@ export default function LiveResultsTable({
       </Table.Header>
 
       <Table.Body>
-        {sortedResults.map((result) => {
-          const competitor =
-            competitorsByRegistrationId[result.registration_id];
-          const hasResult = result.attempts.length > 0;
+        {sortedResultsByCompetitor.map((competitorWithResults) => {
+          return competitorWithResults.results.map((r) => {
+            const hasResult = r.attempts.length > 0;
```
gregorbg: Again, the naming here is confusing. I think you want to express the idea of "a tuple of a competitor and a result", using "with" in the sense of "and".
But when I was reading the code, I scratched my head a lot at first, because "with" can also mean "definitely has something", as in "with or without". So I was wondering why you would check if the competitor has results when the premise already is that it's a competitor _with_ results...

next-frontend/src/components/live/LiveResultsTable.tsx
```diff
@@ -72,64 +77,66 @@ export default function LiveResultsTable({
       </Table.Header>
 
       <Table.Body>
-        {sortedResults.map((result) => {
-          const competitor =
-            competitorsByRegistrationId[result.registration_id];
-          const hasResult = result.attempts.length > 0;
+        {sortedResultsByCompetitor.map((competitorWithResults) => {
+          return competitorWithResults.results.map((r) => {
```
gregorbg: renaming `result` to `r` in the loop artificially induces a lot of diff in this file, is there a reason why you shortened it?

next-frontend/src/lib/hooks/useResultsSubscription.ts
```diff
@@ -43,34 +43,63 @@ export type DiffProtocolResponse = {
...
+    Object.fromEntries(
+      roundIds.map((id) => [id, CONNECTION_STATE_INITIALIZED]),
+    ),
   );
 
+  // Stable ref so onReceived doesn't retrigger the effect
+  const onReceivedRef = useRef(onReceived);
+  useLayoutEffect(() => {
+    onReceivedRef.current = onReceived;
+  });
```
gregorbg: Can you move this into our own custom polyfill of `useEffectEvent`? That way, we will have a plug-and-play easy refactor once NextJS rolls out 16.2 and we can finally get React 19

next-frontend/src/lib/hooks/useResultsSubscription.ts
```diff
@@ -43,34 +43,63 @@ export type DiffProtocolResponse = {
...
+          connected: () =>
+            setConnectionStates((prev) => ({
+              ...prev,
+              [roundId]: CONNECTION_STATE_CONNECTED,
+            })),
+          disconnected: () =>
+            setConnectionStates((prev) => ({
+              ...prev,
+              [roundId]: CONNECTION_STATE_DISCONNECTED,
+            })),
```
gregorbg: Please refactor this "update state for one specific round" snippet into shared code

next-frontend/src/lib/hooks/useResultsSubscription.ts
```diff
@@ -43,34 +43,63 @@ export type DiffProtocolResponse = {
...
+        },
+      ),
     );
 
-    return () => {
-      subscription.unsubscribe();
-    };
-  }, [roundId, onReceived]);
+    return () => subscriptions.forEach((s) => s.unsubscribe());
+  }, [roundIds]); // roundIds should be stable (useMemo at call site)
```
gregorbg: We also have React Compiler running already, so that's not much of a concern anyways ;-)

next-frontend/src/lib/hooks/useResultsSubscription.ts
```diff
@@ -43,34 +43,63 @@ export type DiffProtocolResponse = {
...
 
-  return connectionState;
+  // Aggregate: worst state wins
+  const values = Object.values(connectionStates);
+  return values.includes(CONNECTION_STATE_DISCONNECTED)
+    ? CONNECTION_STATE_DISCONNECTED
+    : values.every((s) => s === CONNECTION_STATE_CONNECTED)
+      ? CONNECTION_STATE_CONNECTED
+      : CONNECTION_STATE_INITIALIZED;
```
gregorbg: Can you associate the different states with actual different "level integers" and then just call one single `min`? That would make your intent way more obvious and in line with the code comment

next-frontend/src/lib/live/mergeAndOrderResults.ts
```diff
@@ -0,0 +1,44 @@
...
+): CompetitorWithResults[] => {
+  const orderedResultsByRegistrationId = _.mapValues(
+    resultsByRegistrationId,
+    (results) => orderResults(results, format),
+  );
+
+  const orderedResults = orderResults(
+    Object.values(_.map(orderedResultsByRegistrationId, (r) => r[0])),
+    format,
+  );
```
gregorbg: Why are you ordering twice?

next-frontend/src/providers/LiveResultProvider.tsx
```diff
@@ -22,72 +32,100 @@ const LiveResultContext = createContext<LiveResultContextType | undefined>(
...
-    },
+    ),
   );
 
-  const { refetch, data } = useQuery({
-    ...queryOptions,
-    initialData: initialRound,
+  const results = useQueries({
+    queries: queries.map((q, i) => ({ ...q, initialData: initialRounds[i] })),
+  });
```
gregorbg, line 95: I think you can directly pass the `initialData` to `api.queryOptions` in a second, separate parameter after the OpenAPI options object

next-frontend/src/providers/LiveResultProvider.tsx
```diff
@@ -22,72 +32,100 @@ const LiveResultContext = createContext<LiveResultContextType | undefined>(
...
+  const resultsRef = useRef(results);
+  useLayoutEffect(() => {
+    resultsRef.current = results;
   });
 
-  const { results, state_hash } = data;
+  const roundIds = useMemo(
+    () => initialRounds.map((r) => r.id),
+    [initialRounds],
+  );
```
gregorbg: I don't think you need a memo for that thanks to React Compiler

next-frontend/src/providers/LiveResultProvider.tsx
```diff
@@ -22,72 +32,100 @@ const LiveResultContext = createContext<LiveResultContextType | undefined>(
...
 
   return (
     <LiveResultContext.Provider
       value={{
-        liveResults: results,
-        stateHash: state_hash,
-        refetch,
+        liveResultsByRegistrationId,
+        refetch: () => results.forEach((r) => r.refetch()),
```
gregorbg: You should use `Promise.all` here

next-frontend/src/providers/LiveResultProvider.tsx
```diff
@@ -22,72 +32,100 @@ const LiveResultContext = createContext<LiveResultContextType | undefined>(
...
 
-  const connectionState = useResultsSubscription(initialRound.id, onReceived);
+  const connectionState = useResultsSubscriptions(roundIds, onReceived);
+
+  // Merge results by registration_id for the context
+  const liveResultsByRegistrationId = useMemo(
+    () =>
+      _.groupBy(
+        resultsRef.current.flatMap((r, i) =>
+          (r.data.results ?? []).map((res) => ({
```
gregorbg: When can `results` ever be undefined to trigger the `??`? You're passing in `initialData` to every query, so we should be all good...

next-frontend/src/providers/LiveResultProvider.tsx
```diff
@@ -22,72 +32,100 @@ const LiveResultContext = createContext<LiveResultContextType | undefined>(
...
+            ...res,
+            // To differentiate between results for Dual Rounds
+            wcif_id: initialRounds[i].id,
+          })),
+        ),
+        "registration_id",
+      ),
+    [initialRounds],
```
gregorbg: You have successfully kept `results` out of the dependencies here, but this leads to another new problem: This exact dependency array you see here `[initialRounds]` keeps the memo stable until eternity. Even if a websocket update comes in, the `initialRounds` variable most likely hasn't changed, so the update is ignored by definition.
You would never ever receive updates in child components.

next-frontend/src/components/live/LiveUpdatingResultsTable.tsx
FinnIckler: seems like you were reviewing some old version of the code :P

next-frontend/src/lib/live/mergeAndOrderResults.ts
FinnIckler: One is ordering the results *within* the competitor's results. One is now ordering all those best results

next-frontend/src/components/live/LiveResultsTable.tsx
FinnIckler: `competitorsWithTheirPotentiallyEmptySortedResults` is a bit of a mouthful

FinnIckler: I could remove the competitor from the equation and just have `orderedResultsByRegistrationId` and then look up the competitor dynamically?

next-frontend/src/providers/LiveResultProvider.tsx
FinnIckler: I don't think refetching is necessary from outside anyway

next-frontend/src/lib/hooks/useResultsSubscription.ts
```diff
@@ -43,34 +50,53 @@ export type DiffProtocolResponse = {
...
+  );
+
+  const changeConnectionState = useCallback(
+    (roundId: string, connectionState: ConnectionState) => {
+      setConnectionStates((prev) => ({
+        ...prev,
+        [roundId]: connectionState,
+      }));
+    },
+    [],
```
gregorbg, line 76: This probably needs `setConnectionStates` as a dependency

next-frontend/src/providers/LiveResultProvider.tsx
```diff
@@ -29,65 +31,100 @@ export function LiveResultProvider({
...
+    ...api.queryOptions(
+      "get",
+      "/v1/competitions/{competitionId}/live/rounds/{roundId}",
+      {
+        params: { path: { roundId: round.id, competitionId } },
       },
-    },
-  );
+    ),
+    initialData: initialRounds[i],
```
gregorbg: Huh? Shouldn't this be `initialData: round`? And shouldn't this be wrapped inside brackets?
I'm expecting `api.queryOptions("get", "someRoute", { openApiParams }, { initialData: foo, otherTanstackParams })`.

Might just be the woozy GitHub diff, though.

gregorbg (changes requested):

next-frontend/src/lib/hooks/useResultsSubscription.ts
FinnIckler: it does not. 
```
1️⃣ React guarantees setState identity stability

The setter returned by useState (in your case setConnectionStates) is stable across renders. Its reference never changes, so it does not need to be included in useCallback dependencies.

From React’s guarantees:

The dispatch function returned by useState is stable and won’t change between renders.
```

next-frontend/src/providers/LiveResultProvider.tsx
FinnIckler: i mean initialRounds[i] is the same as round here, but of course it is cleaner

next-frontend/src/lib/hooks/useResultsSubscription.ts
gregorbg: Interesting! It does make it look weird though that there's a `useCallback` with an empty dependency array...

gregorbg: What about just not wrapping it into a `useCallback` at all and letting React Compiler do its thing?

next-frontend/src/lib/live/mergeAndOrderResults.ts
gregorbg: Okay. Totally makes sense why you need to do that, but hard to read especially during the second pass. As far as I can tell, what you need in the second pass is only the _competitors_ ordered by their fastest result...

next-frontend/src/lib/live/mergeAndOrderResults.ts
```diff
@@ -0,0 +1,44 @@
...
+import _ from "lodash";
+import { Format } from "@/lib/wca/data/formats";
+import { orderResults } from "@/lib/live/orderResults";
+import { LiveResultsByRegistrationId } from "@/providers/LiveResultProvider";
+import { LiveResult } from "@/types/live";
+
+type CompetitorWithResults = components["schemas"]["LiveCompetitor"] & {
+  global_pos: number;
+  advancing: boolean;
+  advancing_questionable: boolean;
```
gregorbg: Can you `Pick<>` these three out of the OpenAPI schema? To make it safe for future spec refactors

next-frontend/src/lib/live/mergeAndOrderResults.ts
gregorbg: As in, once you have that `registration_id`, you can then use it to fetch the "fastest" result of that competitor within the loop. I think that would make the code easier to read.

gregorbg: I think something along these lines would make your already existing approach more readable:

```ruby
  const orderedResultsByRegistrationId = _.mapValues(
    resultsByRegistrationId,
    (results) => orderResults(results, format),
  );

  const bestResultsPerCompetitor = Object.values(orderedResultsByRegistrationId).map(
    (results) => results[0]
  );
  
  const globallyOrderedResults = orderResults(bestResultsPerCompetitor, format);
```

next-frontend/src/providers/LiveResultProvider.tsx
```diff
@@ -29,65 +31,100 @@ export function LiveResultProvider({
...
+    ...api.queryOptions(
+      "get",
+      "/v1/competitions/{competitionId}/live/rounds/{roundId}",
+      {
+        params: { path: { roundId: round.id, competitionId } },
       },
-    },
-  );
+    ),
+    initialData: round,
```
gregorbg, line 75: It seems my previous comment somehow vanished. Why is it not...

```suggestion
    ...api.queryOptions(
      "get",
      "/v1/competitions/{competitionId}/live/rounds/{roundId}",
      {
        params: { path: { roundId: round.id, competitionId } },
      },
      {
        initialData: round,
      },
    ),
```

next-frontend/src/providers/LiveResultProvider.tsx
```diff
@@ -29,65 +31,100 @@ export function LiveResultProvider({
...
+        "registration_id",
+      ),
+      stateHashesByRoundId: Object.fromEntries(
+        results.map((r) => [r.data.id, r.data.state_hash]),
+      ),
+    }),
   });
 
-  const { results, state_hash } = data;
+  const roundIds = initialRounds.map((r) => r.id);
```
gregorbg: I think you can move this down closer to `const connectionState`, because it is not immediately needed in `onReceived` (as far as my CTRL+F can tell...)

next-frontend/src/providers/LiveResultProvider.tsx
```diff
@@ -29,65 +31,100 @@ export function LiveResultProvider({
...
 
-  const onReceived = useCallback(
-    (result: DiffProtocolResponse) => {
+  const onReceived = useEffectEvent(
```
gregorbg: Shouldn't the `useEffectEvent` wrapping happen _inside_ the `useResultsSubscription` method body?
As in, you pass in a callable to the hook, and then within the body of that hook (where the `useEffect` immediately resides) whatever callable is being passed in is wrapped?

gregorbg: From https://react.dev/learn/separating-events-from-effects#limitations-of-effect-events:

> Effect Events are very limited in how you can use them: [...] **Never pass them to other components or Hooks.**

next-frontend/src/providers/LiveResultProvider.tsx
```diff
@@ -29,65 +31,100 @@ export function LiveResultProvider({
...
+    ),
+    initialData: round,
+  }));
 
-  const { refetch, data } = useQuery({
-    ...queryOptions,
-    initialData: initialRound,
+  const { liveResultsByRegistrationId, stateHashesByRoundId } = useQueries({
+    queries,
+    combine: (results) => ({
```
gregorbg: Please call this `queryResults` or similar, to distinguish it from the idea of "WCA results [achieved with a StackMat timer at a competition]"

next-frontend/src/providers/LiveResultProvider.tsx
```diff
@@ -29,65 +31,100 @@ export function LiveResultProvider({
...
+      liveResultsByRegistrationId: _.groupBy(
+        results.flatMap((r, i) =>
+          r.data.results.map((res) => ({
+            ...res,
+            // To differentiate between results for Dual Rounds
+            round_wcif_id: initialRounds[i].id,
```
gregorbg, line 86: Would this happen to be available in `r.data` somewhere? I vaguely remember you adding `round_wcif_id` to the serialization of `LiveResult` in another PR...

gregorbg: Oh, it was actually this very PR! So why is this "manual override" necessary then in the first place?

next-frontend/src/providers/LiveResultProvider.tsx
FinnIckler: because we pass it to a useQuery which takes all the options in one object!

next-frontend/src/providers/LiveResultProvider.tsx
FinnIckler: because we do not send it over in the Web Socket. This can also be done in the onReceived instead, not sure what is cleaner

next-frontend/src/lib/hooks/useResultsSubscription.ts
gregorbg: You resolved this comment but I do not see any change in the code, nor can I see any replies (justifying why you didn't change the code)

next-frontend/src/providers/LiveResultProvider.tsx
gregorbg: Okay, merging it in here is totally fine for the time being! 👍

next-frontend/src/providers/LiveResultProvider.tsx
gregorbg: Ah, the actual reason is far more subtle!

```tsx
const queries = initialRounds.map((round) => api.queryOptions(
    "get",
    "/v1/competitions/{competitionId}/live/rounds/{roundId}",
    {
      params: { path: { roundId: round.id, competitionId } },
    },
    {
      initialData: round,
    },
  ));
```

This will work on the surface, BUT: Tanstack Query has a special type helper to say "`data` returned by a query can never be `undefined` if you supply `initialData`". This special type inference gets "lost" if you pipe `initialData` through `api.queryOptions`.
It's definitely a bug on their side, and you found a good way to workaround that bug :P

next-frontend/src/lib/hooks/useResultsSubscription.ts
FinnIckler: if you try that eslint complains
`ESLint: The 'changeConnectionState' function makes the dependencies of useEffect Hook (at line 98) change on every render. Move it inside the useEffect callback. Alternatively, wrap the definition of 'changeConnectionState' in its own useCallback() Hook. (react-hooks/exhaustive-deps)`

gregorbg: How interesting! This means that ESLint does not recognize React Compiler optimizations...

FinnIckler: https://www.npmjs.com/package/eslint-plugin-react-hooks there seems to be a specific set for react compiler rules

gregorbg: We have that plugin running as part of `eslint-nextjs-config`. In fact, the error message that you quoted above comes from exactly this plugin...

gregorbg: We can leave the callback as-is and I will green-light the PR after you fixed the CI errors. But it's still quite curious that this doesn't work as intended

FinnIckler: Yes. What I am saying is that we need to turn on/off specific errors for the react compiler versions? Or do you think nextjs does that

gregorbg: I would expect NextJS to do that, but maybe that's another Next v16 kinda issue. I would keep my feet still until 16.2 comes out and we can upgrade, and if the issue still persists after then we still have plenty of time to worry :D

FinnIckler: I am actually still fixing an interaction with adding and removing pending results in the new setup

lib/live/diff_helper.rb
```diff
@@ -42,7 +42,7 @@ def self.add_forecast_stats(diff, round)
     end
 
     def self.forecast_for(updated_result, round)
-      return updated_result unless updated_result["live_attempts"].length < round.format.expected_solve_count
+      return updated_result if updated_result["live_attempts"].blank? || updated_result["live_attempts"].length == round.format.expected_solve_count
```
gregorbg: What is the edge case you're catching here? A zero-entry array would have worked under the old `<` condition as well, so I think what you really want is `nil?` instead...

next-frontend/src/components/live/LiveUpdatingResultsTable.tsx
```diff
@@ -23,16 +24,21 @@ export default function LiveUpdatingResultsTable({
...
         <ConnectionPulse connectionState={connectionState} />
       </HStack>
+      <PendingResultsTable
+        formatId={formatId}
+        eventId={eventId}
+        competitors={competitors}
+      />
       <LiveResultsTable
-        results={liveResults}
+        resultsByRegistrationId={liveResultsByRegistrationId}
```
gregorbg, line 43: Hm, I'm a bit unhappy about the inconsistent way the context is handled:
- In case of `PendingResultsTable`, you just render the component and internally, it grabs what it needs from the context
- In case of `LiveResultsTable`, you have to _actively_ grab some results from the context and pass them to the component.

Would it be possible to harmonize approaches, so that you do either always A or always B?

lib/live/diff_helper.rb
FinnIckler: yes it is nil. I am catching cases where advancing has changed for a result in the diff, but live_attempts has not

next-frontend/src/components/live/LiveUpdatingResultsTable.tsx
FinnIckler: Well PendingResultsTable can *only* be used inside of a LiveResult Context. But LiveResultsTable can be used outside of it, when we want to display non updating static live results

FinnIckler: I can change it though, don't have any hard preference

FinnIckler: changed

PR #13518, opened 2026-02-24
Differentiate live result types

FinnIckler:
No actual code changes

### DIFF
--- a/next-frontend/src/lib/live/decompressDiff.ts
+++ b/next-frontend/src/lib/live/decompressDiff.ts
@@ -4,9 +4,12 @@ import {
 } from "@/lib/hooks/useResultsSubscription";
 import _ from "lodash";
 import { LiveResult } from "@/types/live";
+import { PartialExcept } from "@/lib/types/objects";
 
-type PartialLiveResultWithRegistrationId = Partial<LiveResult> &
-  Pick<LiveResult, "registration_id">;
+type PartialLiveResultWithRegistrationId = PartialExcept<
+  LiveResult,
+  "registration_id"
+>;
 
 export function decompressDiff(diff: CompressedLiveResult): LiveResult;
 export function decompressDiff(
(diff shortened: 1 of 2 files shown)


next-frontend/src/lib/live/decompressDiff.ts
```diff
@@ -2,23 +2,19 @@ import {
...
-import { components } from "@/types/openapi";
 import _ from "lodash";
+import { LiveResult } from "@/types/live";
 
-type PartialLiveResultWithRegistrationId = Partial<
-  components["schemas"]["LiveResult"]
-> &
-  Pick<components["schemas"]["LiveResult"], "registration_id">;
+type PartialLiveResultWithRegistrationId = Partial<LiveResult> &
+  Pick<LiveResult, "registration_id">;
```
gregorbg: This smells like `PartialExcept`

next-frontend/src/lib/live/rankingCellColorPalette.ts
```diff
@@ -0,0 +1,17 @@
+export function rankingCellColorPalette({
+  advancing,
+  advancing_questionable,
+}: {
+  advancing: boolean;
+  advancing_questionable: boolean;
+}) {
```
gregorbg: Can you type this via a `Pick<>` on the appropriate live result hierarchy item?

next-frontend/openapi/schemas/live/BaseLiveResult.yaml
gregorbg, line 1: It seems like the only two decendants are `ByPersonLiveResult` and `RoundLiveResult`. Both seem to include `round_wcif_id`, `global_pos` and `local_pos`. So why not move that up to the base class?

FinnIckler: I intentionally left them out because that are the values that are not included in the WebSocket type

FinnIckler: I can then use the BaseLiveResult in the compress diff PR

PR #13521, opened 2026-02-25
H2H Import Fix: Handle case where first solve is a DNF

dunkOnIT:
The logic for setting `result.best` only update the `best` value if `value > best`. This didn't work in the case where the first result for a user was a DNF (-1). This PR adds this case by also checking for if the `best` is negative (a DNF), and replacing it with `value` if the value is positive.

### DIFF
--- a/lib/tasks/h2h_results_import.rake
+++ b/lib/tasks/h2h_results_import.rake
@@ -33,8 +33,11 @@ namespace :h2h_results do
           lr.local_pos = final_pos
         end
 
-        # negative result.best means it is DNF, so we also ovrwrite the `best` if we have a non-negative `value` and a negative `best`
-        result.update!(best: value) if value.positive? && (result.best > value || result.best.negative?)
+        # negative result.best means it is DN*, so we need to consider that the new value either beat the current best in a traditional way OR the current best until now is a DN* penalty
+        is_time_better = value.positive? && (value < result.best || result.best.negative?)
+        # When the result is already a penalty, DNF beats DNS. In terms of numbers, this means `-1` is "better" than `-2`, even though mathematically `-1 > -2`
+        is_penalty_better = value.negative? && result.best.negative? && value > result.best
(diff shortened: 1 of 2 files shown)


lib/tasks/h2h_results_import.rake
```diff
@@ -33,7 +33,8 @@ namespace :h2h_results do
           lr.local_pos = final_pos
         end
 
-        result.update!(best: value) if value.positive? && result.best > value
+        # negative result.best means it is DNF, so we also ovrwrite the `best` if we have a non-negative `value` and a negative `best`
+        result.update!(best: value) if value.positive? && (result.best > value || result.best.negative?)
```
gregorbg: Hyper-nitpick: If `DNF` already made you stumble, then in the long term `DNS` (as `-2` value) will, too. Maybe you say now "Huh how is somebody supposed to get a DNS in a Head to Head round where both competitors solve at the same time" but believe me: In the long run, around the world, across all competitions that are held each weekend, it will eventually happen one day.

By that token, we do consider a DNF "better" than a DNS: Look at result `3893755` as a random example, there are thousands of others where all attempts are `-1` or `-2`, but the `best` field is always `-1`. And to finally convince you:
- `SELECT COUNT(*) FROM results WHERE best = -1` = `176887`
- `SELECT COUNT(*) FROM results WHERE best = -2` = `0`

So my suggestion is...
```suggestion
        # negative result.best means it is DN*, so we need to consider that the new value either beat the current best in a traditional way OR the current best until now is a DN* penalty
        is_time_better = value.positive? && (value < result.best || result.best.negative?)
        # When the result is already a penalty, DNF beats DNS. In terms of numbers, this means `-1` is "better" than `-2`, even though mathematically `-1 > -2`
        is_penalty_better = value.negative? && result.best.negative? && value > result.best
        result.update!(best: value) if is_time_better || is_penalty_better
```

dunkOnIT: Updated per your suggestion, thanks!

PR #13523, opened 2026-02-25
fix the special characters validators to allow Unicode chars

pranav-027:
Continued work of: https://github.com/thewca/worldcubeassociation.org/pull/13522

cc: @dunkOnIT @thewca/wrt

### DIFF
--- a/spec/lib/results_validators/persons_validator_spec.rb
+++ b/spec/lib/results_validators/persons_validator_spec.rb
@@ -316,52 +316,52 @@ RSpec.describe PV do
 
         # Test names with apostrophes (common in many countries)
         res_apostrophe1 = create(:inbox_result, competition: competition1, event_id: "333oh", round: round_333oh)
-        res_apostrophe1.person.update(name: "Adam D'Aloia")
+        res_apostrophe1.person.update!(name: "Adam D'Aloia")
 
         res_apostrophe2 = create(:inbox_result, competition: competition1, event_id: "333oh", round: round_333oh)
-        res_apostrophe2.person.update(name: "Sebastian O'Mahony-Hagan")
+        res_apostrophe2.person.update!(name: "Sebastian O'Mahony-Hagan")
 
         # Test names with Unicode scripts (Chinese, Thai, Devanagari, Arabic, Russian, Tamil, etc.)
         res_chinese = create(:inbox_result, competition: competition1, event_id: "333oh", round: round_333oh)
-        res_chinese.person.update(name: "Shayibai Halimulati (莎伊拜·哈力木拉提)")
+        res_chinese.person.update!(name: "Shayibai Halimulati (莎伊拜·哈力木拉提)")
 
         res_thai = create(:inbox_result, competition: competition1, event_id: "333oh", round: round_333oh)
-        res_thai.person.update(name: "Kanneti Sae Han (คันธ์เนตี แซ่ห่าน)")
+        res_thai.person.update!(name: "Kanneti Sae Han (คันธ์เนตี แซ่ห่าน)")
 
         res_devanagari = create(:inbox_result, competition: competition1, event_id: "333oh", round: round_333oh)
-        res_devanagari.person.update(name: "Ram Thakkar (राम ठक्कर)")
+        res_devanagari.person.update!(name: "Ram Thakkar (राम ठक्कर)")
 
         res_arabic = create(:inbox_result, competition: competition1, event_id: "333oh", round: round_333oh)
-        res_arabic.person.update(name: "Farham Farajizadeh (فرهام فرجی‌زاده)")
+        res_arabic.person.update!(name: "Farham Farajizadeh (فرهام فرجی‌زاده)")
 
         res_russian = create(:inbox_result, competition: competition1, event_id: "333oh", round: round_333oh)
-        res_russian.person.update(name: "Nikita Razzamazov (Никита Раззама́зов)")
+        res_russian.person.update!(name: "Nikita Razzamazov (Никита Раззама́зов)")
 
         res_tamil = create(:inbox_result, competition: competition1, event_id: "333oh", round: round_333oh)
-        res_tamil.person.update(name: "Akash Sreedharan (ஆகாஷ் ஸ்ரீதரன்)")
+        res_tamil.person.update!(name: "Akash Sreedharan (ஆகாஷ் ஸ்ரீதரன்)")
 
         res_bengali = create(:inbox_result, competition: competition1, event_id: "333oh", round: round_333oh)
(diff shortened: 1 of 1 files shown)


spec/lib/results_validators/persons_validator_spec.rb
```diff
@@ -309,6 +309,84 @@
...
+        res_baybayin.person.update(name: "John Edison Ubaldo (ᜇ᜔ᜌᜓ︀ᜈ᜔ ᜁᜇᜒᜐᜓ︀ᜈ᜔ ᜂᜊᜎ᜔ᜇᜓ︀)")
+
+        res_vietnamese = create(:inbox_result, competition: competition1, event_id: "333oh", round: round_333oh)
+        res_vietnamese.person.update(name: "Trần Tử Kiên")
+
+        # Test names with forward slash (names in Malaysia especially among Malaysians of Indian origin)
+        # A/L = “Anak Lelaki” (Malay) → means “son of”
+        # A/P = “Anak Perempuan” → means “daughter of”
+        res_slash = create(:inbox_result, competition: competition1, event_id: "333oh", round: round_333oh)
+        res_slash.person.update(name: "Sanjev A/L Loganathan")
```
pranav-027: I found two instances of this in database its being discussed within WRT

gregorbg: Has a conclusion been reached here?

pranav-027: Nope changing PR to draft once we have a conclusion. I'll ping you

pranav-027: Not allowing slashes

db/migrate/20260304054202_normalize_person_name_quotes_to_ascii_apostrophe.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class NormalizePersonNameQuotesToAsciiApostrophe < ActiveRecord::Migration[8.1]
+  def up
+    Person.where("name REGEXP ?", "[’‘]").find_each do |person|
```
gregorbg: According to the CTRL-F of my browser, these two apostrophes are the same symbol. Can you please double-check to verify?

db/migrate/20260304054202_normalize_person_name_quotes_to_ascii_apostrophe.rb
```diff
@@ -0,0 +1,19 @@
...
+
+      next if new_name == person.name
+
+      person.execute_edit_person_request("fix", { name: new_name })
+    end
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
+  end
```
gregorbg: If this is the case, then your data sanity fix should probably not go into a migration.

For one-off scripts like this, use a Rake task under `lib/tasks` (where you will find many existing tasks as example, maybe don't look at `db.rake` because that's the most complex and complicated one) and after deployment, a WST Senior Member will execute the task for you.

db/migrate/20260304054202_normalize_person_name_quotes_to_ascii_apostrophe.rb
```diff
@@ -0,0 +1,19 @@
...
+class NormalizePersonNameQuotesToAsciiApostrophe < ActiveRecord::Migration[8.1]
+  def up
+    Person.where("name REGEXP ?", "[’‘]").find_each do |person|
+      new_name = person.name
+                       .tr("’", "'")
+                       .tr("‘", "'")
+
+      next if new_name == person.name
+
+      person.execute_edit_person_request("fix", { name: new_name })
```
gregorbg: You're passing in `"fix"`, which means that under the hood, the `execute_edit_person_request` method simply fires an SQL `UPDATE` query and nothing else.

This makes me think whether, instead of a Rake task (as per my other comment in this review) the whole thing could just be done in pure SQL instead and WRT can execute it on their own behalf through PhpMyAdmin...

danieljames-dj: @gregorbg I suggested to go with migration instead of direct SQL. The reason is that we need to do the same change in `users` and `persons` table. Doing it via SQL is also possible, but if it can be run through the "fix" code, then I feel it might make things easy. Let us know what will be the ideal approach here.

db/migrate/20260304054202_normalize_person_name_quotes_to_ascii_apostrophe.rb
pranav-027: image
 
 Yes I can confirm that its diffrent

db/migrate/20260304054202_normalize_person_name_quotes_to_ascii_apostrophe.rb
pranav-027: I'll go with SQL query only

db/migrate/20260304054202_normalize_person_name_quotes_to_ascii_apostrophe.rb
pranav-027: I'll consider SQL query only! I tried that on my local and did bit of sanity testing everything seems to be working fine

spec/lib/results_validators/persons_validator_spec.rb
```diff
@@ -309,6 +309,72 @@
...
+        res_vietnamese = create(:inbox_result, competition: competition1, event_id: "333oh", round: round_333oh)
+        res_vietnamese.person.update(name: "Trần Tử Kiên")
+
+        # Test names with bullet
+        res_bullet = create(:inbox_result, competition: competition1, event_id: "333oh", round: round_333oh)
+        res_bullet.person.update(name: "Arabel•la Puig Villacorta")
+
+        # Test name with Dutch "in 't"
+        res_dutch = create(:inbox_result, competition: competition1, event_id: "333oh", round: round_333oh)
+        res_dutch.person.update(name: "Jelle in 't Veld")
```
gregorbg: I do want to kindly point out that `update` calls in Rails can fail. They are documented to return a boolean that indicates whether the update was successful or not.

That's probably not a big issue here, because we don't run unicode symbol validations on an individual model level here that could make the inserts actually fail, but I still want to point it out as a learning.

If you're executing a "fire and forget" update which you assume should "just work", then `update!` with the exclamation mark is what you need. That one throws an exception and thereby interrupts the test execution altogether if the update is unsuccessful

pranav-027: Done!

gregorbg (approved):

Issue #13418, opened 2026-02-13
key not found with WCA Results Export


2016WUJI01:
**Describe the bug**
```
ERROR:  insert or update on table "result_attempts" violates foreign key constraint "result_attempts_result_id_fkey"
DETAIL:  Key (result_id)=(4581808) is not present in table "results".
```

**To Reproduce**
1. When I import result_attempts.tsv after result.tsv, it show me ` violates foreign key constraint`
2. I try to select in result table with (id = 4581808), and not found this id

@thewca/wrt Have you encountered this bug?

gregorbg:
Hm. When you say "import TSV", what are you importing the data into? TSV files themselves have no "foreign keys".

What surprises me is that our production SQL database has this foreign key, and it is _not_ violated. So the data is definitely there. But from the top of my head, I cannot explain why it is missing in the export :O

Issue #13208, opened 2026-01-16
Add a sanity check which ensures there are no orphaned result_attempts


Eiim:
As of the most recent export, there are 105,966 attempts in `result_attempts` which have a `result_id` value that doesn't correspond to any row in `results`. They're present both in the database dump and the results export. It seems like they mostly have result_ids that would suggest they're fairly recent (last year or two). If they correspond to results that are intended to be deleted from the database, then it seems like they also should be deleted from `result_attempts`. Otherwise, I'm not sure what to make of these. If they're intentionally included, I think a mention/explanation in the result export README would be warranted given how many there are.

PR #13524, opened 2026-02-25
Add foreign keys to LiveAttempts

FinnIckler:
This will fix the issue with orphaned attempts like https://github.com/thewca/worldcubeassociation.org/issues/13208 and https://github.com/thewca/worldcubeassociation.org/issues/13418 

I currently have added deleting the orphaned result to the migration. Maybe we want to run this manually?

If we do run this first, we need to run both

```
      DELETE FROM live_attempt_history_entries
      WHERE live_attempt_id IN (
        SELECT id FROM live_attempts
        WHERE live_result_id IS NOT NULL
          AND live_result_id NOT IN (SELECT id FROM live_results)
      )
```
and
```
      DELETE FROM live_attempts
      WHERE live_result_id IS NOT NULL
        AND live_result_id NOT IN (SELECT id FROM live_results)
```

because there is currently no on_delete: cascade for the history entries


db/migrate/20260225164242_add_cascading_foreign_keys_to_live_results.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+class AddCascadingForeignKeysToLiveResults < ActiveRecord::Migration[8.1]
+  def change
+    # First, make sure we cascade for live_attempt_history_entries so when we delete orphaned attempts
+    # it will also delete the orphaned live_attempt_history_entries
+    remove_foreign_key :live_attempt_history_entries, :live_attempts
+    add_foreign_key :live_attempt_history_entries, :live_attempts, column: :live_attempt_id, on_delete: :cascade
```
gregorbg: I think Rails is able to guess the `column` itself, as it follows the "name of to_table in singular with `_id` attached" nomenclature precisely

db/migrate/20260225164242_add_cascading_foreign_keys_to_live_results.rb
```diff
@@ -0,0 +1,20 @@
...
+
+    # Clean up orphaned live_attempts (live_result_id points to a non-existent live_result)
+    execute <<~SQL
+      DELETE FROM live_attempts
+      WHERE live_result_id IS NOT NULL
+        AND live_result_id NOT IN (SELECT id FROM live_results)
+    SQL
+
+    remove_index :live_attempts, :live_result_id
+    add_foreign_key :live_attempts, :live_results, column: :live_result_id, on_delete: :cascade
```
gregorbg: Again, I think Rails can guess the `column` value itself.

db/migrate/20260225164242_add_cascading_foreign_keys_to_live_results.rb
```diff
@@ -0,0 +1,20 @@
...
+    # it will also delete the orphaned live_attempt_history_entries
+    remove_foreign_key :live_attempt_history_entries, :live_attempts
+    add_foreign_key :live_attempt_history_entries, :live_attempts, column: :live_attempt_id, on_delete: :cascade
+
+    # Clean up orphaned live_attempts (live_result_id points to a non-existent live_result)
+    execute <<~SQL
+      DELETE FROM live_attempts
+      WHERE live_result_id IS NOT NULL
+        AND live_result_id NOT IN (SELECT id FROM live_results)
+    SQL
```
gregorbg: You can wrap this into an `up_only do ... end` block

db/migrate/20260225164242_add_cascading_foreign_keys_to_live_results.rb
gregorbg: I do not see why you need to manually remove the existing index. `add_foreign_key` is not documented to add any new index, and when applying the migration on my machine ™️ it actually does not create one. Is it possible at all that the confusion around the indexing might come from an earlier, work-in-progress version of this migration?

FinnIckler: It does though? Otherwise we won't see this in schema.rb
`t.index ["live_result_id"], name: "fk_rails_b8d7c73463"`

FinnIckler: yeah it just worked on the laptop

PR #13525, opened 2026-02-25
Preload live_attempts in recompute_advancing

FinnIckler:
Removes doing one query for each count in `complete?`

### DIFF
--- a/app/models/live_result.rb
+++ b/app/models/live_result.rb
@@ -81,7 +81,8 @@ class LiveResult < ApplicationRecord
   end
 
   def complete?
-    live_attempts.filter { it.value != 0 }.length == round.format.expected_solve_count
+    # Use length hear to not fire a COUNT query for every row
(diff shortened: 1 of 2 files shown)


app/models/live_result.rb
```diff
@@ -81,7 +81,7 @@ def should_recompute?
   end
 
   def complete?
-    live_attempts.where.not(value: 0).count == round.format.expected_solve_count
+    live_attempts.filter { it.value != 0 }.length == round.format.expected_solve_count
```
gregorbg: Just curious for context here: When do we ever create individual live **attempts** with value zero?
I know we will have "empty" live **results** with no attempts and `best` equal to zero a lot, and that's totally fine. I am astonished however that the way you're filtering here (as well as in the old version, FWIW) implies that there can be individual attempts attached to a result with value zero...

PR #13530, opened 2026-02-27
Add Admin Competitor Results Menu

FinnIckler:
Still some open questions:
- Should I send a delete applyDiff onSuccess or do we want the web socket to handle that? (I would say onSuccess for better UX)
- WCA Live shows who would advance in the previous round when quitting, should I already implement that?
  - If yes: how? I would probably just send a request on showing the dialog with a specific route for this prompt? WCA live has the benefit of graphql here

### DIFF
--- a/spec/requests/live_update_spec.rb
+++ b/spec/requests/live_update_spec.rb
@@ -46,8 +46,8 @@ RSpec.describe "WCA Live API" do
       expect do
         put api_v1_competition_live_clear_competitor_in_round_path(competition.id, round.wcif_id, registration.id)
       end.to have_broadcasted_to(Live::Config.broadcast_key(round.wcif_id))
-               .from_channel(ApplicationCable::Channel)
-               .with(hash_including("updated" => [Live::DiffHelper.compress_payload({ "registration_id" => registration.id, "average" => 0, "best" => 0, "live_attempts" => [], "bpa" => 1, "wpa": -1})]))
+        .from_channel(ApplicationCable::Channel)
+        .with(hash_including("updated" => [Live::DiffHelper.compress_payload({ "registration_id" => registration.id, "average" => 0, "best" => 0, "live_attempts" => [], "bpa" => 1, wpa: -1 })]))
 
       result = LiveResult.find_by(round_id: round.id, registration_id: registration.id)
       expect(result).to be_present


config/routes.rb
```diff
@@ -380,7 +379,7 @@
             get '/rounds/:round_id' => 'live#round_results', as: :live_round_results
             put '/rounds/:round_id/open' => "live#open_round", as: :live_round_open
             put '/rounds/:round_id/clear' => "live#clear_round", as: :live_round_clear
-            put '/rounds/:round_id/:registration_id' => 'live#quit_competitor', as: :quit_competitor_from_round
+            put '/rounds/:round_id/:registration_id/quit' => 'live#quit_competitor', as: :quit_competitor_from_round
```
gregorbg: Why not a `DELETE` on the previously established `/rounds/:round_id/:registration_id` path?

next-frontend/src/providers/LiveResultAdminProvider.tsx
```diff
@@ -84,6 +90,19 @@ export function LiveResultAdminProvider({
     },
   );
 
+  const { mutate: mutateQuit, isPending: isPendingQuit } = api.useMutation(
+    "put",
+    "/v1/competitions/{competitionId}/rounds/{round_id}/{:registration_id}/quit",
+    {
+      onSuccess: () => {
+        // Do we remove the competitor here or do we wait for the web socket update?
```
gregorbg: Which web socket message would we be waiting on? Your question kinda implies that the action of quitting triggers a background job that handles the quitting asynchronously. But from all I can tell, the quitting code is completely synchronous so if the request completes successfully, you know that the competitor already _has been_ quit (also in the backend)

gregorbg: I really don't see where quitting triggers any kind of delayed websocket messaging

FinnIckler: Yes, quitting is done synchronously. But we still need to send websocket requests to 
1. Inform Clients that the Competitor is no longer advancing in the previous round and
2. Inform Clients that the Competitor is no longer part of the current round

This is currently not implemented, but I am planning to do this as part of this PR

config/routes.rb
FinnIckler: not a bad idea

app/models/round.rb
```diff
@@ -419,19 +419,50 @@ def first_round?
...
+      if first_round?
+        is_quit ? 1 : 0
+      else
+
+        # Set the last person that didn't advance to advancing if wanted
+        # We need to do this before marking the competitor as quit as otherwise
+        # the quit competitor could be the "best non-advancing" competitor
+        if should_advance_next
+          # The scope is ordered by global_pos by default
+          to_advance = previous_round.live_results.where.not(advancing: true).first
```
FinnIckler: While writing the Live Requirement list I noticed that this is too simple. We actually need to check the advancement condition here

FinnIckler:
Introducing the `next_qualifying_to_round` was a bit of headache and I had to go back into the elixir codebase to figure out how to do it. I added a bunch of tests and am fairly confident with the implementation (except for Dual Rounds, I need to make a PR to easily create these via factories).

app/controllers/api/v1/live/live_controller.rb
```diff
@@ -5,7 +5,9 @@ class Api::V1::Live::LiveController < Api::V1::ApiController
   skip_before_action :require_user!, only: %i[round_results by_person podiums]
 
   def add_or_update_result
-    results = params.expect(attempts: [%i[value attempt_number]])
+    # can't use .expect here because [] is valid for attempts
+    # TODO: Think about if a synchronous route instead would make sense instead
+    results = params.permit(attempts: %i[value attempt_number]).fetch(:attempts, [])
```
gregorbg: Please clarify, what exact (seemingly valid) payload are you sending that breaks `expect`?

app/controllers/api/v1/live/live_controller.rb
```diff
@@ -106,16 +108,20 @@ def open_round
 
   def quit_competitor
     competition = Competition.find(params.require(:competition_id))
+    wcif_id = params.require(:round_id)
     registration_id = params.require(:registration_id)
+    should_advance_next = params[:advance_next]
```
gregorbg: Do you need to parse this into a boolean explicitly? If it's JSON body then it's fine like that, otherwise even in GET query URLs the values `true` and `false` in JS will be received by Ruby as `"true"` and `"false"` string literals

app/jobs/update_live_result_job.rb
```diff
@@ -21,6 +22,10 @@ def perform(live_result, results, entered_by_id)
       end
     end
 
+    previous_attempts.each do |attempt_number, attempt|
+      attempt.destroy unless new_attempt_numbers.include?(attempt_number)
+    end
```
gregorbg: Can this benefit from your new foreign keys? Something like `live_result.live_attempts.where.not(attempt_number: new_attempt_numbers).delete_all`

app/models/round.rb
```diff
@@ -202,7 +202,7 @@ def clear_round!
 
   def open_round!
     empty_results = advancing_registrations.map do |r|
-      { registration_id: r.id, round_id: id, average: 0, best: 0, last_attempt_entered_at: current_time_from_proper_timezone }
+      LiveResult.empty_result_attributes(r.id, id)
```
```suggestion
      LiveResult.empty_result_attributes(r.id, self.id)
```

gregorbg: To clarify the ambiguity of which ID is which

next-frontend/src/providers/LiveResultAdminProvider.tsx
```diff
@@ -84,6 +87,17 @@ export function LiveResultAdminProvider({
     },
   );
 
+  const { mutate: mutateQuit, isPending: isPendingQuit } = api.useMutation(
+    "delete",
+    "/v1/competitions/{competitionId}/live/rounds/{roundId}/{registrationId}",
+    {
+      onSuccess: () => {},
```
gregorbg: If you really don't want to do anything, then you don't have to specify `onSuccess` at all

app/models/round.rb
```diff
@@ -422,19 +426,84 @@ def first_round?
...
+
+    return [] unless prev_results.exists?(advancing: true)
+
+    already_quit_ids = prev_results.where.not(quit_by_id: nil).pluck(:id)
+
+    first_advancing = prev_results.where(advancing: true).first
+
+    candidate_ids = prev_results
+                    .where(advancing: false)
+                    .where.not(id: already_quit_ids)
```
gregorbg: Isn't this essentially just a `where(quit_by_id: nil)`?

app/models/round.rb
```diff
@@ -422,19 +426,84 @@ def first_round?
     number == 1 || (linked_round.present? && linked_round.first_round_in_link.number == 1)
   end
 
-  def quit_from_round!(registration_id, quitting_user)
-    ActiveRecord::Base.transaction do
+  # Port from https://github.com/thewca/wca-live/blob/main/lib/wca_live/scoretaking/advancing.ex#L143
+  # Basically this just removes the number one placed competitor and then sees who of the non-advancing
+  # competitors would make it
+  def next_qualifying_to_round
```
gregorbg: If you define some scopes on `live_results`, this can be made much more readable. I'm thinking along the lines of
- `scope :advancing`
- `scope :not_advancing`
- `scope :quit`
- `scope :not_quit`

app/models/round.rb
```diff
@@ -422,19 +426,84 @@ def first_round?
...
+    return [] if first_round?
+
+    prev_round = previous_round
+    prev_results = prev_round.linked_round.present? ? prev_round.linked_round.live_results : prev_round.live_results
+
+    return [] unless prev_results.exists?(advancing: true)
+
+    already_quit_ids = prev_results.where.not(quit_by_id: nil).pluck(:id)
+
+    first_advancing = prev_results.where(advancing: true).first
```
gregorbg: How do you guarantee that this `first` respects the correct order of `prev_results`?

app/models/round.rb
```diff
@@ -422,19 +426,84 @@ def first_round?
...
+
+    candidate_ids = prev_results
+                    .where(advancing: false)
+                    .where.not(id: already_quit_ids)
+                    .pluck(:id)
+
+    return [] if candidate_ids.empty?
+
+    ignored_ids = ([first_advancing.id] + already_quit_ids)
```
```suggestion
    ignored_ids = ([first_advancing.id] | already_quit_ids)
```

gregorbg: Set union

app/models/round.rb
```diff
@@ -422,19 +426,84 @@ def first_round?
...
+
+    advancement_determining = prev_results
+                              .where.not(id: ignored_ids)
+
+    qualifying_index = prev_round.max_qualifying(advancement_determining)
```
gregorbg: How does this interact with linked rounds, where the `prev_round` may have some form of (yet to be determined) `all` advCondition?

app/models/round.rb
```diff
@@ -219,30 +219,34 @@ def recompute_live_columns(skip_advancing: false)
     live_results.reset
   end
 
+  def max_qualifying(advancement_determining_results)
+    # Our Regulations allow at most 75% of competitors to proceed
+    max_qualifying = (live_results.count * 0.75).floor
```
gregorbg: This will break with Dual Rounds

app/models/round.rb
```diff
@@ -422,19 +426,84 @@ def first_round?
...
+    ignored_ids = ([first_advancing.id] + already_quit_ids)
+
+    advancement_determining = prev_results
+                              .where.not(id: ignored_ids)
+
+    qualifying_index = prev_round.max_qualifying(advancement_determining)
+
+    hypothetically_advancing_ids = advancement_determining
+                                   .take(qualifying_index)
+                                   .filter_map(&:id)
```
gregorbg: Now you lost me. What I understand:
- The `advancement_determining` is a set of all competitors, except for the first/best/fastest advancing guy and all already quit guys.
- You then stuff all of these into `max_qualifying`, to determine which number/amount N of these competitors could advance
- Then you take exactly those N competitors. So far, so god.

But then what is this `filter_map` about? Every row in `advancement_determining` has an ID, so what is the `filter` actually filtering?

app/models/round.rb
```diff
@@ -422,19 +426,84 @@ def first_round?
...
+
+    prev_results.where(id: hypothetically_advancing_ids & candidate_ids)
```
gregorbg: Continuing from my last comment: Now you know "who except the first guy" is advancing. What is the intersection with `candidate_ids` doing?

app/models/round.rb
```diff
@@ -422,19 +426,84 @@ def first_round?
...
+    end
+
+    # Now that the transaction has finished we can notify the clients
+    recompute_advancing
+    live_results.reset
+    after_quit_state = to_live_state
+    Live::DiffHelper.broadcast_compressed_diff(before_quit_state, after_quit_state, self)
+
+    unless first_round?
+      previous_round.live_results.reset
```
gregorbg: In what way have the results of the previous round been affected?

gregorbg: i.e. what makes this reset necessary?

app/models/round.rb
```diff
@@ -422,19 +426,84 @@ def first_round?
...
+    return quit_count if quit_count.zero?
 
-      # We need to also quit the result from the previous round so advancement can be correctly shown
-      previous_round_results = previous_round.linked_round.present? ? previous_round.linked_round.live_results : previous_round.live_results
+    if should_advance_next
+      to_advance = next_qualifying_to_round.first
 
-      previous_round_results.where(registration_id: registration_id).count { |r| r.mark_as_quit!(quitting_user) }
+      if to_advance.present?
+        to_advance.update(advancing: true)
```
gregorbg: What if this `update` operation fails? You're currently using the "quiet" version without `!`.

app/models/round.rb
```diff
@@ -422,19 +426,84 @@ def first_round?
...
+        live_results.create(**LiveResult.empty_result_attributes(to_advance.registration_id, id))
```
```suggestion
        live_results.create(**LiveResult.empty_result_attributes(to_advance.registration_id, self.id))
```

app/controllers/api/v1/live/live_controller.rb
FinnIckler: attempts: [] for clearing the attempts

app/jobs/update_live_result_job.rb
FinnIckler: yeah, true

app/models/round.rb
FinnIckler: prev_results has a default scope of ordered by glopal pos

app/models/round.rb
FinnIckler: that will depend on how we define max_advancing on the advancement condition class

app/models/round.rb
FinnIckler: mh.. yes, only in the cases where there are different amount of competitors in r1 vs. r2. so we will need to have some kind of min between them

app/models/round.rb
FinnIckler: yeah the filter map can be removed. I think I had it in there in an earlier stage of the elixir conversion

app/models/round.rb
FinnIckler: This is equivalent to
```
for result <- hypothetically_qualifying,
            result.id in candidates_to_qualify_ids,
            do: result.person
```
in the elixir code. 
Basically (and maybe @jonatanklosko correct me if I am wrong):
- `advancement_determining` results (which is `qualifying_results_ignoring` in the elixir) considers all the results, regardless if they are already advancing or not. We have to union by candidate_ids here because now we want to know who from the now advancing people can actually advance (meaning they didn't advance already)

app/models/round.rb
FinnIckler: we change the advancing of a competitor in the previous round

app/models/round.rb
FinnIckler: I'll change that to update!

app/models/round.rb
FinnIckler: but you are right it isn't necessary because it is going through the query cache

app/models/round.rb
jonatanklosko: > We have to union by candidate_ids here because now we want to know who from the now advancing people can actually advance (meaning they didn't advance already)

Correct!

app/models/round.rb
jonatanklosko: @FinnIckler if I understand this correctly `advancement_determining` is a subset of all results, which we then use to determine `max_qualifying`. Doesn't this mean that percentage advancement condition will give a smaller value than it should?

app/models/round.rb
```diff
@@ -422,19 +426,84 @@ def first_round?
...
+
+    ignored_ids = ([first_advancing.id] + already_quit_ids)
+
+    advancement_determining = prev_results
+                              .where.not(id: ignored_ids)
+
+    qualifying_index = prev_round.max_qualifying(advancement_determining)
+
+    hypothetically_advancing_ids = advancement_determining
+                                   .take(qualifying_index)
```
jonatanklosko: "Take N" is problematic with tied results. It may happen that you quit a result and the next **ranking number** from the previous round has two tied competitors. In that case both should advance, unless they don't fit the strict 75%, in which case **neither** of them should advance (and a "Take N" approach would likely advance one of them).

To handle such cases, the [current WCA Live code](https://github.com/thewca/wca-live/blob/9c6e24d4614d901f83f7fefade48b4f393a25bd4/lib/wca_live/scoretaking/advancing.ex#L203-L219) basically simulates a scenario where the competitors we want to ignore got DNF, and then it runs the usual advancement/qualify logic on that simulated result set, applying all the usual rules. To be clear, I am not saying that it's necessarily the the best approach, but rather that the code needs consider the same edge cases that the advancement itself does.

(I may be missing something in the logic though, so if you account for that, then all good)

app/models/round.rb
FinnIckler: yes, good point

app/models/round.rb
FinnIckler: this is the usual advancement logic :upside_down_face: so that also currently incorrectly handles ties

app/controllers/api/v1/live/live_controller.rb
FinnIckler: this is a DELETE, which supports json bodies.

spec/lib/live/state_diff_spec.rb
```diff
@@ -140,7 +140,7 @@
 
     it "has a compression key defined for each serialization attribute" do
       create(:live_result, round: round)
-      keys = []
+      keys = %w[best_possible_average worst_possible_average]
```
gregorbg: Why? Either you have them in the live state's keys or you don't. Manually adding things here feels like you're defeating the purpose of the test...

gregorbg: Also, even if there is a good reason for adding these two manually, why does it need to be in this PR? You're not introducing BPA/WPA right here

next-frontend/src/providers/LiveResultProvider.tsx
```diff
@@ -133,6 +138,9 @@ export function MultiRoundResultProvider({
           (r) => !updated.map((u) => u.r).includes(r.registration_id),
         ),
       );
+      updatePendingQuitCompetitors((currentlyQuitCompetitors) =>
+        currentlyQuitCompetitors.difference(new Set(deleted)),
```
gregorbg, line 142: Isn't `deleted` representing the _results_ that were deleted? I'm confused why you can use a set difference on the set of _competitors_ that were deleted...

next-frontend/src/providers/LiveResultProvider.tsx
```diff
@@ -25,6 +25,8 @@ export type LiveResultsByRegistrationId = Record<string, LiveResult[]>;
 interface LiveResultContextType {
   liveResultsByRegistrationId: LiveResultsByRegistrationId;
   pendingLiveResults: LiveResult[];
+  pendingQuitCompetitors: Set<number>;
+  addPendingQuitCompetitor: (registrationId: number) => void;
```
gregorbg: You're kinda "squeezing" them between the pending getter and the pending setter. Please either do `pendingGet, pendingSet, quitGet, quitSet` or `pendingGet, quitGet, pendingSet, quitSet`.

Currently this looks like `pendingGet, quitGet, quitSet, pendingSet`, which confuses me.

next-frontend/openapi/paths/competitions/live/QuitCompetitor.yaml
```diff
@@ -0,0 +1,42 @@
...
+        type: number
+  requestBody:
+    required: false
+    content:
+      application/json:
+        schema:
+          type: object
+          properties:
+            should_advance_next:
+              type: boolean
```
gregorbg: A suggestion on API design regarding mutability of "who proceeds": Big competitions are highly async, so a simple boolean flag is not really "transactional".

What about actually passing a (nullable!) integer ID of who exactly should proceed, and if the backend receives something not-null, it assumes (a) yes, this person should proceed but also (b) checks _again_ after the `next_if_quit` route to make sure this is _really_ the competitor who is currently eligible to advance next. If not, return a corresponding error code.

The idea here is to avoid situations with parallel score taking where the order of the non-proceeding competitors changed between the call to `next_if_quit` and the actual `DELETE` action being fired.

app/controllers/api/v1/live/live_controller.rb
gregorbg: Have you done a bit of research whether this is actually intended? As long as you don't skip the `attempts` key altogether, an empty array _should_ satisfy `expect`, at least from an "intuitive API design" standpoint. There's a slim chance that this is actually a Rails bug...

gregorbg: How can I reproduce this error (the error that would happen if you switched this back to `expect` notation) locally?

app/controllers/api/v1/live/live_controller.rb
```diff
@@ -106,17 +108,35 @@ def open_round
...
+  def next_if_quit
+    competition = Competition.find(params.require(:competition_id))
+    wcif_id = params.require(:round_id)
+    registration_id = params.require(:registration_id)
+
+    require_manage!(competition)
+
+    round = Round.find_by_wcif_id!(wcif_id, competition.id, includes: [:live_results])
+
+    to_advance = round.first_round? ? [] : round.previous_round.next_advancing_without(registration_id).take(1)
```
gregorbg: This `take(1)` confuses me. What if there's an exact tie on the next eligible advancing?

app/models/round.rb
```diff
@@ -417,21 +414,66 @@ def first_round?
...
+
+    first_advancing = relevant_results.advancing.first
+
+    candidate_ids = relevant_results.not_advancing.not_quit.pluck(:id)
 
-      is_quit = result.destroy
+    return [] if candidate_ids.empty?
 
-      return is_quit ? 1 : 0 if first_round?
+    ignored_ids = ([first_advancing.id, competitor_being_quit] | already_quit_ids)
```
gregorbg: Is RuboCop forcing the outermost round brackets?
If yes, then put the first `[...]` literal into its own variable, and then do `ignored_ids = my_variable | already_quit_ids` here.

app/models/round.rb
```diff
@@ -417,21 +414,66 @@ def first_round?
...
+    to_advance = previous_round.next_advancing_without(registration_id).first if should_advance_next
+
+    transaction do
+      quit_count = Live::DiffHelper.broadcast_changes(self) do
+        result = live_results.find_by!(registration_id: registration_id)
+        destroyed = result.destroy
+        live_results.create(**LiveResult.empty_result_attributes(to_advance.registration_id, self.id)) if to_advance.present?
+        recompute_advancing
+        live_results.reset
+        destroyed ? 1 : 0
```
gregorbg: In reality, will it ever happen that `destroyed` is `false` and we're just cool with it?

You're using `find_by!`, which makes sense because if you want to quit Mr. X, you're implicitly assuming that Mr. X currently is part of the round. But then, if you have a `result` from `find_by!` which definitely exists (per the exclamation mark), why are you using soft-delete?

next-frontend/src/providers/LiveResultProvider.tsx
FinnIckler: because both are represented by registration ids

spec/lib/live/state_diff_spec.rb
FinnIckler: yeah, the way is just to create an incomplete result instead.

app/controllers/api/v1/live/live_controller.rb
FinnIckler: good point

app/models/round.rb
FinnIckler: I will switch it to destroy!

next-frontend/src/providers/LiveResultProvider.tsx
FinnIckler: basically when a competitor is deleted we send a websocket update that a result is deleted, I can then use that the quit is no longer pending

next-frontend/openapi/paths/competitions/live/QuitCompetitor.yaml
FinnIckler: it did require some finagling of the methods, but it is done

app/controllers/api/v1/live/live_controller.rb
FinnIckler: let me add a test for this

FinnIckler: yeah, this test fails with expect:

```
    it "Clears a live result correctly" do
      sign_in delegate

      competition = create(:competition, event_ids: ["333"], delegates: [delegate])
      round = create(:round, competition: competition, event_id: "333")
      registration = create(:registration, :accepted, competition: competition)
      create(:live_result, round: round, registration: registration)
      live_request = {
        attempts: [],
        registration_id: registration.id,
      }

      patch api_v1_competition_live_update_results_path(competition.id, round.wcif_id), params: live_request
      expect(response).to be_successful

      perform_enqueued_jobs

      result = LiveResult.find_by(round_id: round.id, registration_id: registration.id)
      expect(result).to be_present

      expect(result.live_attempts).to be_empty
      expect(result.best).to eq 0
      expect(result.average).to eq 0
    end
```

app/jobs/update_live_result_job.rb
```diff
@@ -13,22 +13,16 @@ def perform(live_result, results, entered_by_id)
...
-    history_ordered_results = new_attempts.order(:attempt_number).pluck(:value)
-    live_result.live_result_history_entries.create!(entered_by_id: entered_by_id, action_type: :scoretaking, attempt_details: history_ordered_results)
-
-    after_state = round.to_live_state
-    diff = Live::DiffHelper.round_state_diff(before_state, after_state)
-
-    diff = Live::DiffHelper.add_forecast_stats(diff, round)
-
-    ActionCable.server.broadcast(Live::Config.broadcast_key(round.wcif_id), diff)
+      live_result.update!(live_attempts: new_attempts, best: best, average: average, last_attempt_entered_at: Time.now.utc)
```
gregorbg: I think this line got inserted accidentally during merge. Almost an exact copy is at L20

app/controllers/api/v1/live/live_controller.rb
FinnIckler: fixed this by adding a separate route

